PR #4, opened 2016-07-10
Make ClientConfig Sync

brson:
Replacing RefCell with Mutex, mark some boxes as Send + Sync.

I had to do this to make rustls work with hyper.  [Here's where I did the hyper integration](https://github.com/brson/rustup.rs/blob/rustls/src/download/src/lib.rs#L289).


ctz:
Thanks! :+1:

PR #7, opened 2016-07-15
Handle non-UTF-8 PEM input gracefully

inejge:
Some CA bundles in PEM format have content that is not valid UTF-8. This patch handles that kind of input without panicking.


ctz:
Thanks! :+1:

PR #8, opened 2016-07-19
issue 6: handle absent CA file

yanns:
fix https://github.com/ctz/rustls/issues/6:
when the CA file cannot be read, provide an user friendly error message, and help the user fixing the issue.


ctz:
Thanks!

PR #9, opened 2016-07-19
Enlarge deframer internal buffer

inejge:
Bump `MessageDeframer.chunk` size from 2K to 16K to avoid an excessive number of `read()` syscalls for large transfers. In my testing with rustup it easily gave about 2.5x the throughput.

### DIFF
--- a/src/msgs/deframer.rs
+++ b/src/msgs/deframer.rs
@@ -19,7 +19,7 @@ pub struct MessageDeframer {
   buf: Vec<u8>,
 
   /// A buffer into which we read.
-  chunk: [u8; 2048]
+  chunk: [u8; 16384]
 }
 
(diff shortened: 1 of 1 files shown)


inejge:
> (removed diff, see PR #11)
>
> I think that will fix both issues at the same time. Any thoughts?

The patch in #11 is much better thought out than my little klugde, of course. I've tested it in rustup and it works flawlessly.

Now, about my previous "testing": apologies for claims of dramatic improvements in throughput, they are an artifact of running under strace, which ate a lot more CPU with the larger amount of `read()` calls. (That'll teach me, I hope.) However, there _is_ a measurable reduction in system time with the larger buffer, so I suppose it doesn't hurt to keep it.

Anyway, this PR can be closed.

PR #10, opened 2016-07-20
Add a Session trait to allow being generic over client/server TLS sessions

sciyoshi:
@ctz this is useful for writing P2P protocols on top of TLS, where there is no distinction between client/server after the TLS session is established. Note that this is (minorly) backwards-incompatible, since it requires importing the trait directly to use its methods.

I've also added `send_close_notify` to the trait, since [it can (and should) be sent by both parties](https://tools.ietf.org/html/rfc5246#section-7.2.1).

### DIFF
--- a/examples/tlsclient.rs
+++ b/examples/tlsclient.rs
@@ -19,6 +19,8 @@ use docopt::Docopt;
 
 extern crate rustls;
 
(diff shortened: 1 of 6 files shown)


ctz:
Thanks!

PR #13, opened 2016-08-19
Add target/ to .gitignore.

briansmith:
This is useful for people that work on rustls using the defaults
settings for the compiler, where things get built into target/.


ctz:
Thanks!

PR #14, opened 2016-08-19
Use *ring* and webpki from crates.io.

briansmith:
The maintainer of *ring* and webpki intends to update the leftmost
non-zero version number of *ring* and webpki whenever a
backward-incompatible change is made, similar to what has been done
with untrusted on crates.io, so these dependencies should be relatively
safe in the case of known/intended breakage. See
http://doc.crates.io/specifying-dependencies.html#specifying-dependencies-from-cratesio
for how this works.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,18 +13,8 @@ untrusted = "0.3.1"
 time = "0.1.35"
 base64 = "~0.2.0"
 log = "0.3.6"
-
-[replace]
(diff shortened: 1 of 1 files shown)


briansmith:
I published webpki 0.2.1 on crates.io with the invalid serial number fix and your tests (thanks!). I updated this PR to reference *ring* 0.3.0 and webpki 0.2.1.

ctz:
Thanks! :+1:

PR #29, opened 2016-09-24
Remove unnecessary indirection with &Vec<T>, use &[T].

frewsxcv:


ctz:
Thanks! This was before I understood what I was doing wrt. vecs/slices.

PR #36, opened 2016-11-25
Laying ground work to remove more allocations.

valarauca:
I'm trying to remove some of the dependency on the short lived small vectors where I can.

This patch really just cleans up `SessionID` and shrinks `Random`'s size. I want to make `PayLoad` be a slice in the future, and try to get `VecBuf` so it's using internal slices. I'm not sure we'll see.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -121,7 +121,7 @@ impl MessageCipher for GCMMessageCipher {
       Message {
         typ: msg.typ,
         version: msg.version,
-        payload: MessagePayload::opaque(buf)
+        payload: MessagePayload::opaque(buf.as_slice())
       }
     )
   }
@@ -172,7 +172,7 @@ impl MessageCipher for GCMMessageCipher {
     Ok(Message {
       typ: typ,
       version: version,
-      payload: MessagePayload::opaque(result)
+      payload: MessagePayload::opaque(result.as_slice())
     })
   }
 }
@@ -271,7 +271,7 @@ impl MessageCipher for ChaCha20Poly1305MessageCipher {
       Message {
         typ: msg.typ,
         version: msg.version,
-        payload: MessagePayload::opaque(buf)
+        payload: MessagePayload::opaque(buf.as_slice())
       }
     )
   }
@@ -310,7 +310,7 @@ impl MessageCipher for ChaCha20Poly1305MessageCipher {
     Ok(Message {
       typ: typ,
       version: version,
-      payload: MessagePayload::opaque(buf)
+      payload: MessagePayload::opaque(buf.as_slice())
     })
   }
 }
--- a/src/msgs/base.rs
+++ b/src/msgs/base.rs
@@ -3,7 +3,7 @@ use msgs::codec::{Codec, Reader};
 
 /// An externally length'd payload
 #[derive(Debug, Clone, PartialEq)]
-pub struct Payload( pub Vec<u8>);
+pub struct Payload(pub Vec<u8>);
 
 impl Codec for Payload {
   fn encode(&self, bytes: &mut Vec<u8>) {
--- a/src/msgs/base.rs
+++ b/src/msgs/base.rs
@@ -3,7 +3,7 @@ use msgs::codec::{Codec, Reader};
 
 /// An externally length'd payload
 #[derive(Debug, Clone, PartialEq)]
-pub struct Payload(pub Vec<u8>);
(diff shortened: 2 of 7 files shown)


valarauca:
I'll debug on OSX to see what went wrong.

ctz:
Sorry, I haven't had a chance to look at this properly yet. But I wouldn't worry about the OSX build; travis ci isn't reliable on OSX.

ctz:
Hm, actually this will regress performance a bit by adding another copy into the encryption/decryption path. (After encryption/decryption, the thing passed into `MessagePayload::opaque` was moved into the payload. Now it's dropped to a slice and recopied.) I think we perhaps need an owning `MessagePayload::Opaque` and a borrowing `MessagePayload::Opaque`. Thoughts?

valarauca:
I think that would be the better option. So like `MessagePayload::opaque` mirrors `std::borrow::cow`? Or should it use cow internally?

ctz:
Yes I think that'll work. The key thing is that we can go from a decryption result to an appdata `Message` without a copy.

ctz:
Oh, and the other big win will be making the callers of `send_plain` not create a Vec which is thrown away after encryption. This is all going to involve lots of lifetime specifiers everywhere, which is why I haven't done it yet :)

briansmith:
There is always a copy involved. It's just that sometimes the copy is hidden in the encrypt/decrypt operation since the destination and the source are the same. *ring* currently only supports in-place decryption/encryption, but I believe it needs to add an out-of-place encryption API, at least, in order to minimize copies with common APIs. For example, consider the example in the documentation where there is a `client.write(b"GET / HTTP/1.0\r\n\r\n")` call. With out-of-place encryption it would be possible to avoid an explicit copy.

I'm not sure if this would help with what you describe, or if that issue is different.

valarauca:
>which is why I haven't done it yet :)

Your not kidding I've restructured the [`Payload` and `Codec`](https://github.com/valarauca/codecbase/blob/master/codec.rs) types. I had to relearn a few things about the lifetime system in the process. Going to attempt to integrate this over the next few day, but I'm also traveling so it may take longer.

In the long run a few types will need to be created slightly differently to avoid hangups. 

>There is always a copy involved. It's just that sometimes the copy is hidden in the encrypt/decrypt 

Very true. TLS is compression, padding, AND encryption. 

Ideally (read as: unlikely, but a goal) there shouldn't be allocations outside of the decompression+decryption, or compression+encryption phase. Even then Slab Collections can likely prevent allocating.

valarauca:
So I'm nearly there.

There is just about 40-ish borrow checker errors left.

Took about 2 weeks of editing to add all the lifetime annotations. 

Now I'm trying to fix the few suborn issues I have left.  Hopefully I'll have it passes tests this weekend and can submit the PR. I think there will only be 4 allocations total (outside of server keys, server names, and protocol names). 2 on encryption, 1 for handshake joining, and 1 for network buffering. Everything else is borrowed.

---

Most the lifetime issues are going to be solved by re-implementing `Cow` to work with `Box<[u8]>` and `&'a [u8]` instead of `Vec<u8>` and `&'a [u8]`. The issue is in a lot of scenarios a `Payload` needs to be converted to a `Reader`, and to do this without allocating seems to be write unsafe code, or write a new version of `Cow`. 

---

So far I've added zero unsafe code.

ctz:
Great, zero unsafe code is a requirement!

What do the performance figures look like?

valarauca:
How should I do preformance testing? We should standardize, or you dictate an approach so we don't approach this with separate methodologies. 

Like maybe a stand alone application to just route across 127.0.01(?)

---

I'll start merging in the changes peice meal. Dropping the whole diff and once maybe a bit hard to review. There's minor changes to every file, some changes (especially in msg/ become major).

That approach will regress preformance (Initially). But it'll make the changes understandable, traceable, and we can ensure no tests get dropped during the process. 

Could a branch be made to test against?

ctz:
Compare the results of running `./target/release/examples/bench` between say ec24b51 and the tip of your branch. For reference previous results on my hardware are at  https://github.com/ctz/rustls/wiki/Performance-measurements but obviously these won't be comparable on your hardware.

Doing this on a branch I can see would be good. There are 3-4k new lines on the TLS1.3 branch, for example. I don't expect you to track that of course; but I will need to make equivalent changes when this hits master.

Thanks!

PR #39, opened 2016-12-07
Adding Certificate/PrivateKey types everywhere.

### LINKED ISSUE #37: Add types to pemfile
I believe you want to add `struct Certificate(Vec<u8>)` and `struct SecretKey(Vec<u8>)`  to module `pemfile`, in order to make it clearer what these things are, and what arguments other functions are expecting (I believe this to be 0-cost in Rust, as they have the same memory representation as `Vec<u8>`. Correct me if I'm wrong).

Then `pemfile::certs` would return `Vec<Certificate>` and `pemfile::rsa_private_key` would return `SecretKey`.

pijul:
This is meant to fix #37, and make code reviews easier.
Cool project, by the way.

### DIFF
--- a/examples/bench.rs
+++ b/examples/bench.rs
@@ -79,7 +79,7 @@ fn drain(d: &mut Session, expect_len: usize) {
   }
 }
 
-fn get_chain() -> Vec<Vec<u8>> {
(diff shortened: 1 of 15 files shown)


ctz:
This is great, thank you!

PR #41, opened 2016-12-15
Using more recent version of ring and webpki

pijul:
Without this pull request, rustls is not usable in code that also uses more recent version of ring because of name clashes at link. For example, pijul needs both rustls and thrussh.

Also, a number of deprecated features have been dropped from webpki for security reasons, which probably means rustls should also stop using them.


ctz:
Rebased and merged with test fixes as e9bac07. Thanks!

PR #42, opened 2016-12-21
Pre-life Time annotation structure changes

valarauca:
These are the main structural changes I faced when adding lifetime annotations.

Added them first to main the merge easier.

# Cipher/Message

### Problem:

- Post lifetimes `Message` becomes `Message<'a>` and sending/receiving a lifetime to a trait object got _really_ awkward. 
- In some cases `Message`'s already `Opaque` vector is cloned internally. 
- That was my fault in many places from my previous patch.

### Solution:

- Pass a structure in by reference that holds `ContentType` and `ProtocolVersion`.  [code](https://github.com/valarauca/rustls/blob/f2d350ab04b1df3b17bc395f6246f78a6c9d3026/src/msgs/message.rs#L89,#L109)
- Pass/Receive a `Vec<u8>` from encryption. [code](https://github.com/valarauca/rustls/blob/f2d350ab04b1df3b17bc395f6246f78a6c9d3026/src/cipher.rs#L19,#L22)
- Remove vector duplications within encryption/decryption step. 
- Implemented the trait `MessagePayload::From<Vec<u8>>` to allow for `Message::Opaque` to be constructed without allocations (undoing some of the performance losses from my previous patch) [code](https://github.com/valarauca/rustls/blob/f2d350ab04b1df3b17bc395f6246f78a6c9d3026/src/msgs/message.rs#L74,#L79)

### Comments:

- I think passing/receiving a `Vec<u8>` makes more sense as it gives more flexibility if compression is used in the future. 
- The above changes resulted in fairly _massive_ performance gains. Nearly 5-10% in handshaking and 20-30% in cipher thoughtput. 

# Handler/Expectation/ServerSession/ClientSession

### Problem

- `ServerSessionImpl` and `ClientSessnionImpl` will have lifetime annotations added. `Handler` is `'static` while the buffers inside of `*SessionImpl` are not. 
- So `Handler` type needs to have the `'static` assigned lifetime assigned to half itself, and created a _different_ lifetime based on an arg passed to it. I just couldn't make that work.

### Solution:

- I'm using `ConnState` to dispatch to the proper expectation and handler. 
- [Here is the server code](https://github.com/valarauca/rustls/blob/f2d350ab04b1df3b17bc395f6246f78a6c9d3026/src/server.rs#L343,#L366)
- [Here is the client code](https://github.com/valarauca/rustls/blob/f2d350ab04b1df3b17bc395f6246f78a6c9d3026/src/client.rs#L305,#L338)

### Comments:

- This made the [handshake module](https://github.com/valarauca/rustls/blob/f2d350ab04b1df3b17bc395f6246f78a6c9d3026/src/handshake.rs). Pretty much _all_ constants now.
- Some functions in `server_hs` and `client_hs` are public now, the ones originally passed in `Hander` type.
- Overall this caused a ~1-0.5% performance regression

# Performance Report

- Handshaking performance +4-5% 
- Cipher thought-put +20-30%

# Closing Remarks

I added `Result<(),TLSError>` to the `send_close_notify` as there are internal errors it may trigger. This isn't breaking anything. But there are a fair number of warnings now. 

I can remove this and re-submit if you'd like.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -2,7 +2,7 @@ use ring;
 use std::io::Write;
 use msgs::codec;
 use msgs::codec::Codec;
-use msgs::message::{Message, MessagePayload};
+use msgs::message::BrokenMessage;
 use msgs::fragmenter::MAX_FRAGMENT_LEN;
 use error::TLSError;
 use session::SessionSecrets;
@@ -17,8 +17,8 @@ fn xor(accum: &mut [u8], offset: &[u8]) {
 
 /// Objects with this trait can encrypt and decrypt TLS messages.
 pub trait MessageCipher {
-  fn decrypt(&self, m: Message, seq: u64) -> Result<Message, TLSError>;
-  fn encrypt(&self, m: Message, seq: u64) -> Result<Message, TLSError>;
+  fn decrypt(&self, msg: &BrokenMessage, seq: u64, data: Vec<u8>) -> Result<Vec<u8>, TLSError>;
+  fn encrypt(&self, msg: &BrokenMessage, seq: u64, data: Vec<u8>) -> Result<Vec<u8>, TLSError>;
 }
 
 impl MessageCipher {
@@ -83,9 +83,8 @@ const GCM_EXPLICIT_NONCE_LEN: usize = 8;
 const GCM_OVERHEAD: usize = GCM_EXPLICIT_NONCE_LEN + 16;
 
 impl MessageCipher for GCMMessageCipher {
-  fn decrypt(&self, mut msg: Message, seq: u64) -> Result<Message, TLSError> {
-    let payload = try!(msg.take_opaque_payload().ok_or(TLSError::DecryptError));
-    let mut buf = payload.0;
+  fn decrypt(&self, msg: &BrokenMessage, seq: u64, data: Vec<u8>) -> Result<Vec<u8>, TLSError> {
+    let mut buf = data;
 
     if buf.len() < GCM_OVERHEAD {
       return Err(TLSError::DecryptError);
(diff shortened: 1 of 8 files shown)


ctz:
This is first on my to-do list once I'm back from holiday. Thanks!🎄

valarauca:
Okay I just can't this damned lifetime annotation crate to compile. I've been at it a month. I just don't get the problem. The main issue is the Message<'a> structure. The inplace mutation to modify types of a message when decoding is actually impossible without `unsafe`

ctz:
This looks good, especially the changes around `Handler`. 

At the moment the tls1.3 branch is involving quite a lot of other changes which will conflict with this. So now is not a great time for these changes.

So its a 'yes' to these changes, but it might take some time for them to make it onto master. I'm happy to do that work; I don't expect you to spend time chasing a moving target. Thanks!

valarauca:
I can try re-work to re-submit against the TLS1.3 branch.

Also my original work of changing `Payload` to `Cow<'a, [u8]>` failed. 

At it's core there would have to be an `unsafe` split off function. So that a `Vec<u8>` could become `(Vec<u8>, Vec<u8>)` without allocation. I'm going to submit an RFC for such a function because I can't think of an earthly reason why it would violate rust's safety contract.

briansmith:
> At it's core there would have to be an unsafe split off function. So that a Vec<u8> could become (Vec<u8>, Vec<u8>) without allocation.

If `v: Vec<_>` then you can use `let (part1, part2): (&mut [u8], &mut [u8]) = v.split_at_mut(index)`, if you only really need splices.

However, a `split_at_mut` variant that actually returned `(Vec<u8>, Vec<u8>)` without reallocation would be unsafe because each `Vec`s would think it soley owns the underlying array pointer.

valarauca:
@briansmith Slices don't drop the underlying vector when they call `drop()`.

>without reallocation would be unsafe because each Vecs would think it solely owns the underlying array pointer.

100% of the std::collection::Vec library uses unsafe at it's core. I'm talking about submitting an RFC to the standard library not rolling my own. I'm not going to add unsafe to this project.

The real blocking issues is `reallocate_inplace` drops the underlying buffering if _part_ of it is reallocated. So you can't call `reallocate_inplace` on the same buffer twice.

ctz:
> Cipher/Message

92d745d takes the spirit of your changes. It removes the copies going through defragmentation into decryption. Decryption still has a full allocation, because we can't write the decryption result over the user's data in the end.

Speed on my machine is now approx 17.5gbps per core, from about 6gbps before.

PR #48, opened 2017-02-08
added dangerous feature for accepting self signed certs

rustysec:
The `const` is already there and needs to be manually set for testing. Using a feature, we can avoid editing source and this could be passed along by projects using `rustls` as a dependency. Sometimes it is necessary to accept a self signed certificate (internal networks, ephemeral certificate creation, etc).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ webpki = "0.9.2"
 [features]
 default = ["logging"]
 logging = ["log"]
+dangerous = []
 
 [dev-dependencies]
 log = "0.3.6"
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -12,7 +12,10 @@ use x509;
 use key;
 use std::io;
 
-/// Disable all verifications, for testing purposes.
+#[cfg(feature = "dangerous")]
+const DANGEROUS_DISABLE_VERIFY: bool = true;
+
+#[cfg(not(feature = "dangerous"))]
 const DANGEROUS_DISABLE_VERIFY: bool = false;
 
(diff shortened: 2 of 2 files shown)


mcpherrinm:
I'd be pretty nervous about making this a compile-time option:  It wouldn't be visible from code inspection alone whether this is set - you need to know the options you've built with, or testing.  I can see the value in having this option, but I'm wary of this PR.

In fact, I think your message is a great example why:  People turn off certificate validation at the first sign of trouble.  If you want to accept a self-signed certificate in an internal network, you probably know what the certificate should be, so add it to a RootCertStore and connect with verification on.

rustysec:
While I agree with you in principal, it's not always that simple. If you are bringing a device online for a short period of time and need to enable TLS between 100k endpoints (using `rustls`), adding the cert to the local store is not always an option. There are risks here, for sure, but it's already an option in the code. And if there are crates using `rustls` as a dependency (`hyper-rustls`) it's currently a little cumbersome setup with the feature enabled.

While this probably isn't a feature that will be widely used, it is something I see value in for various workflows. There are workarounds, but this is the most ergonomic in some scenarios. This PR won't make or break anything for me.

ctz:
I appreciate the contribution here, but I'm not really comfortable having something like this in the public API. A fork of rustls with `DANGEROUS_DISABLE_VERIFY` set to true might be a better option, or perhaps some extra API for TOFU-like TLS usages.

rustysec:
Upon further consideration, I concur. Thanks for the discussion, love the `rustls`.

PR #59, opened 2017-03-19
TLS1.3 draft 19

ctz:
This can be merged when this draft is more widespread. So far it's tested with OpenSSL trunk.


ctz:
It's also going to need a draft19 boringssl/bogo.

Issue #60, opened 2017-04-10
Client authentication is expecting `id-kp-serverAuth` instead of `id-kp-clientAuth`


tjamaan:
Hello,

I have a server and a client that need to establish a TLS connection. I created two certs: one for the server with EKU=serverAuth and one for the client with EKU=clientAuth.

When I turned client authentication on I found that the server is rejecting clients with error `WebPKIError(RequiredEKUNotFound)`. I traced the issue to `webpki::verify_cert::check_eku`. The function was looking for EKU [43, 6, 1, 5, 5, 7, 3, 1] (a.k.a `id-kp-serverAuth`) instead of EKU [43, 6, 1, 5, 5, 7, 3, 2] (a.k.a  `id-kp-clientAuth`).

Note:
When the server and client both have EKU=serverAuth, client authentication works.
When the server has EKU=clientAuth and client has EKU=serverAuth client authentication fails on client side (`RequiredEKUNotFound`).

ctz:
Sorry, this is a genuine bug coming from reusing too much of webpki's API for client auth.

It'll need a small amount of change on the webpki side to fix.

tilal6991:
I've sent a PR for the required change on the webpki side: https://github.com/briansmith/webpki/pull/43

PR #61, opened 2017-04-13
Verify client certs with correct EKU

ctz:
Addresses issue #60


ctz:
Incorporated into #111

PR #64, opened 2017-05-04
Allows disabling of hostname verification.

nicklan:
This PR would allow a user to call: `danger_disable_host_name_verification` on a ClientConfig to disable host-name verification.

I know this has been sort-of submitted as a PR before and rejected but I do think there are use cases for this and this tries to make it clear that there is a risk to doing this.  A few things to consider:

- This forces user to call danger_disable_host_name_verification to make it clear this has risks
- The hyper-openssl has something very similar: https://docs.rs/hyper-openssl/0.2.6/hyper_openssl/struct.OpensslClient.html#method.danger_disable_hostname_verification
- curl has the --insecure option for this

My use case is building a tool that integrates with existing infrastructure, where in some cases the existing things are configured to use a cert that doesn't verify.  I imagine this isn't an isolated case (hence all the other tools that support this), and as long as it's very clear what you're doing it's nice to support.

### DIFF
--- a/admin/capture-certdata
+++ b/admin/capture-certdata
@@ -71,6 +71,6 @@ if __name__ == '__main__':
         print '    anchors.add_trust_anchors(&webpki_roots::ROOTS);'
         print '    bench(100, "verify_server_cert(%s)", ' % name
         print '          || (),'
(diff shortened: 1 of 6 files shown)


ctz:
Thanks! This is effectively (but not exactly) landed on master.

PR #67, opened 2017-05-07
Update dependencies to work with *ring* 0.9.

briansmith:
I am not sure why rustc just started complaining about the
array -> slice coercions in src/verifybench.rs. I just updated the file
to make the errors go away.

This depends on https://github.com/ctz/webpki-roots/pull/8.


ctz:
Thanks, 0.7 published.

PR #69, opened 2017-05-11
Fixes #68 -- first attempt at an API for disable client certificate verification

### LINKED ISSUE #68: Disable cert verification for client certs?
I want to start by saying that I recognize that in many senses this goes against the spirit of this (excellent) library, so if there's no interest, I'll be understanding.

I've got a use case where I'd like to be able to accept truly any client cert. I've also had use cases in previous contexts where I'd really want to be able to bring my own cert verification as a typical verifier would present a challenge, the most recent of these was wanting to use SCVP for verification instead of doing it myself.

alex:
Curious what your thoughts on this API is. It requires you to go through two hoops to enable it, and if libraries enable it, the use of `features` will propagate that to things which depend on them, hopefully meaning libraries will not be able to do bad things without their users being aware.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ webpki = "0.12"
 [features]
 default = ["logging"]
 logging = ["log"]
+dangerous_configuration = []
 
 [dev-dependencies]
 log = "0.3.7"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -238,6 +238,8 @@ pub use client::{ClientConfig, ClientSession};
 pub use client::ResolvesClientCert;
 pub use server::{StoresServerSessions, ServerSessionMemoryCache};
 pub use server::{ServerConfig, ServerSession};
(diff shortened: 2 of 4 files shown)


ctz:
Looks good. I'd like to generalise this a bit to cover the other PR too. I've used this as the basis for:

https://github.com/ctz/rustls/tree/jbp-dangerous-apis

(this isn't fully done yet, namely all the tests which relied on `DANGEROUS_DISABLE_VERIFY` are broken)

alex:
The name `i_know_what_i_am_doing` weirds me out a bit, but otherwise it looks reasonable. I guess your plan is to not include the "noop verified", which is ok by me!

I'm glad the feature option is working out.

ctz:
> The name i_know_what_i_am_doing weirds me out a bit

Yeah, that's a bit uppity. It's now just `dangerous()`.

alex:
Closing this in favor of #70

PR #70, opened 2017-05-13
Add dangerous_configuration feature

ctz:
This kind of follows on from PR #48, and incorporates #64 and #69.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -212,6 +212,7 @@ impl ClientConfig {
         }
     }
 
+    #[doc(hidden)]
     pub fn get_verifier(&self) -> &'static verify::ServerCertVerifier {
         self.verifier
     }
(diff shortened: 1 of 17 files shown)


src/client.rs
```diff
@@ -188,6 +189,9 @@ pub struct ClientConfig {
     /// Supported versions, in no particular order.  The default
     /// is all supported versions.
     pub versions: Vec<ProtocolVersion>,
+
+    /// How to verify the server certificate chain.
+    verifier: &'static verify::ServerCertVerifier,
```
alex: Hmm, why `&'static` vs binding the lifetime of the `verifier` to `self`?

ctz: Good point, that's probably too restrictive. I'll put it in a `Box`.

I don't want to have a lifetime relative to `ClientConfig` because that will leak the need for an explicit lifetime all the way up the object tree into the public API, which IMO is ugly for such an otherwise-heavyweight object.

alex: `Box` also works, the ability to have some state on the verifier can be important for some use cases :-)

alex:
Just wanted to note that I've ported my code from my branch over to this API and it is 👍

PR #71, opened 2017-05-21
Update Google certificate

pantsman0:
The Google certificate `src/testdata/cert-google.0.der` as it was expired.

This updates it to the cert received when accessing [https://www.google.com](https://www.google.com/) as of May 21, 2017.


ctz:
Thanks!

PR #85, opened 2017-06-15
Use deque to avoid shifting entire vector

sevagh:
Hello,

We ran into a performance issue regarding very large shifts. Traced it to `vec.remove(0)` with a large internal vector, which a VecDeque can help with.

The benchmarks (`cargo run --example bench`) haven't been affected much by this change, the numbers are largely the same.

Should I write a benchmark which actually tests the vecbuf for large vectors?

### DIFF
--- a/src/vecbuf.rs
+++ b/src/vecbuf.rs
@@ -1,19 +1,20 @@
 use std::io::Read;
 use std::io;
 use std::cmp;
+use std::collections::VecDeque;
 
 /// This is a byte buffer that is built from a vector
 /// of byte vectors.  This avoids extra copies when
 /// appending a new byte vector, at the expense of
 /// more complexity when reading out.
 pub struct ChunkVecBuffer {
-    chunks: Vec<Vec<u8>>,
+    chunks: VecDeque<Vec<u8>>,
     limit: usize,
 }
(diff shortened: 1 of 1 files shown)


ctz:
Thanks!

I guess this was a result of lots of small fragments?

RAttab:
The cause was some slowdown on the consumer side which caused a build up in the queue (not sure about the fragment size). This lead to a feedback loop where the queue got large enough that the time to dequeue a single element (`vec.remove(0)`) was longer then the rate at which we received events (around 5-10k/sec). At this point the queue just got longer and longer and the process spent all its time in `memcpy`. Broken stack frames not-withstanding, this made it rather difficult to hunt down the consumer slowdowns.

On a related note, since this data-structure is basically implementation a queue, I would recommend imposing an upper bound on the number of elements it can contain. I need to dig into rustls and rust stream libraries a bit more, but this could be implemented by not reading new fragments when the deque reaches capacity. This would have the advantage of both avoiding resizes on the deque ring buffer (memory allocations and `memcpy`) and of providing a back-pressure mechanism to the event producer (assuming that there aren't other unbounded buffers on the stream being consumed).

PR #94, opened 2017-08-17
Use `ok_or_else()` in server-side certificate resolution.

briansmith:
Replace one use of the (anti-)pattern:

```
if x.is_none() {
   return Err(...)
};
x = x.unwrap();
```

The new code replaces a dynamic check for avoiding a panic with a
static assurance that there is no panic, in particular avoiding using
`unwrap()`.

This is purely a stylistic change and shouldn't have any semantic
effect. There are probably other instances that should be changed too.
I changed this particular one just because I am interested in improving
the code in this area.


ctz:
Thanks!

PR #96, opened 2017-08-20
Add SNI to server session & expose it to the application.

briansmith:

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -15,3 +15,9 @@ pub struct PrivateKey(pub Vec<u8>);
 /// `rustls::pemfile::certs` function can be used to parse a PEM file.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Certificate(pub Vec<u8>);
+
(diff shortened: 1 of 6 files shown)


briansmith:
I expect to write some BoGo tests for this, but they aren't ready yet for non-technical reasons. I expect I'll submit them directly upstream.

ctz:
looks like this is failing tests because bogo uses fictional SNI and weird certificates. i'm going to regenerate all the certs before running tests

briansmith:
@ctz Would it be better if I just updated BoGo to use reasonable certificates by default?

briansmith:
I will rebase this on top of #99. In the meantime, I'll work on updating BoGo to use certs that will work with this.

ctz:
> Would it be better if I just updated BoGo to use reasonable certificates by default?

Yes, that would be ideal.

ctz:
In the meantime I have this merged, albeit with the check moved after HRR for TLS1.3 and some extra tests.

PR #97, opened 2017-08-22
client: make ClientSessionImpl Sync-compatible

Keruspe:
Otherwise rustls can be tough to use in some multi threaded environment.

### DIFF
--- a/src/client/hs.rs
+++ b/src/client/hs.rs
@@ -65,7 +65,7 @@ macro_rules! extract_handshake_mut(
 );
 
 type CheckResult = Result<(), TLSError>;
-type StateResult = Result<Box<State + Send>, TLSError>;
+type StateResult = Result<Box<State + Send + Sync>, TLSError>;
 
 pub trait State {
     fn check_message(&self, m: &Message) -> CheckResult;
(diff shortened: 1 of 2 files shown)


briansmith:
What kind of threading model are you trying to support? IMO it makes more sense for the session state to be `Send` only, and then require the layer above that manages sessions to ensure that only one thread at a time accesses the session state. Actually, I think it might be better to do a bit more than that, but requiring all the client-side state to be `Sync` seems too much, without more explanation.

Keruspe:
I am actually not that familiar with the Sync trait.

I am using tokio-rustls to get a tls stream used to create a client, then I may have several threads, the client creating one channel per thread (and using a Arc<Mutex<>> to guard the stream).

I managed to workaround that by using tokio-core's "handle.spawn" instead of creating an std::thread. 

It looks like I wasn't the only one hit by that: https://github.com/stepancheg/rust-tls-api/commit/b7b396f8d655a4f7c5058c91469fe81401d9240e

Note that before 548db1998bfafaa9ea355119900f92035870393c it worked fine

ctz:
Yes, you're right that ClientSession/ServerSession should both be both Send and Sync.

PR #99, opened 2017-08-28
0.11.0: Update dependencies.

briansmith:
* Ensure Rust 1.20 compatibility, possibly breaking compatibility with older versions (it seems to work with Rust 1.19 at least).

* Prepare for better client authentication, breaking API due to webpki and webpki-roots API changes.


briansmith:
This won't work until all of these prerequisites are done:

* webpki-roots 0.13 is released: https://github.com/ctz/webpki-roots/pull/10
* sct.rs 0.2 is released: https://github.com/ctz/sct.rs/pull/1
* ct-logs 0.2 is released: https://github.com/ctz/ct-logs/pull/1

PR #107, opened 2017-09-01
Give the certificate resolver a webpki::DNSNameRef.

briansmith:
Allow the resolver to rely on the fact that the name is a valid DNS
name. In particular it allows it to know that the DNS name is given
in the IDN ASCII (punycode) encoding.


briansmith:
This builds on top of #106. Only the third commit is unique to this PR.

Issue #101, opened 2017-08-30
Invalid SNI extension generated in client hello when caller does not pass in a valid DNS name


briansmith:
Consider the case where the calling application passes in "" as the DNS name to connect to. Then Rustls will generate an SNI extension with an empty DNS name entry, which is not valid. (Further, there's no way the default certificate validation can work as it requires a valid DNS name.)

Similarly, consider the case where the calling application passes in a stringified IP address, e.g. "1.2.3.4" or "[1.2.3.4]" or "::1" or "[::1]". No SNI extension should be generated in these cases since “Literal IPv4 and IPv6 addresses are not permitted in "HostName"” according to https://tools.ietf.org/html/rfc6066#section-3. (Further, certificate validation cannot work in these cases either since there's no support for iPAddress subjectAltName entries.)

This issue was filed based on the discussion in https://github.com/quininer/tokio-rustls/pull/14#issuecomment-325512055.

briansmith:
I will send a PR for this.

briansmith:
Fixed in #108.

PR #108, opened 2017-09-01
Take the DNS hostname as a webpki::DNSNameRef in ClientSession.

briansmith:
Make it clear that the name passed into ClientSession is a DNS name
and cannot be empty (""), an IP address, an internationalized domain
name, etc. It has to be a DNS hostname because it is put into the SNI
extension in the client hello.


briansmith:
This builds on top of #106. Only the last two commits are unique to this PR.

briansmith:
Fixes #101.

briansmith:
This is built on top of #107 now. Only the last two commits here are specific to this PR.

PR #110, opened 2017-09-02
More flexible and safer client authentication API.

briansmith:
Stop making "no client authentication" the silent default, because that
is not a safe default. Instead, require the user to make an explicit
choice of whether/how to do client authentication.

Previously, the default client authentication setting was the least
safe, so there's was no need to make the ability to plug in a client
auth implementation a “dangerous” feature. With this change, the
ability to provide one's own client authentication implementation
is available in the default configuration.

Some servers need more flexibility in doing client authentication than
was previously provided. Now all the choices for client authentication
are made by `ClientCertVerifier`.


briansmith:
This is built on top of #108. Only the last commit here is specific to this PR.

PR #111, opened 2017-09-02
Validate client certificates using the id-kp-clientAuth EKU.

briansmith:


briansmith:
This is #61 rebased on top of #110 to fix #60 . Only the last two commits in this PR are specific to this PR.  I didn't change the "Rebuild test certs with appropriate EKU" commit at all.

PR #113, opened 2017-09-04
Avoid cloning peer cert. chain in `get_peer_certificates()`.

briansmith:
If the caller wants a copy, it can make its own copy.


briansmith:

PR #114, opened 2017-09-04
Avoid cloning ALPN protocol in `get_alpn_protocol()`.

briansmith:
The caller can clone it if it wants to.


briansmith:

PR #116, opened 2017-09-04
Expose OCSP response as &[u8], not &Vec<u8>.

briansmith:


briansmith:

Issue #123, opened 2017-10-14
Can SNI be disabled on a client or ignored on the server?


bvinc:
I'm currently experimenting with TLS on other protocols besides HTTP, so I have no use for SNI.  Imagine that I'm using no name-based virtual hosting of any kind and can just work with IP addresses.

Currently, on the client side, rustls appears to always take in a hostname string and use it in the SNI extension in the handshake.

Currently, on the server side, if it receives the SNI extension, then it verifies that the certificate used matches the SNI hostname that the client gave, or else gives the error "The server certificate is not valid for the given SNI name".

Both behaviors combined means that I have to invent a string on the client side, and complicate the certificates used on the server side.

Ideally, I want one of the following:
1.  To be able to disable SNI in the client.
2.  To be able to ignore/disable SNI in the server, since it's just a suggestion.

Would you be interested in either of these two features if I made a PR?  If so, what would the API look like?  Would they be "dangerous"?

Relevant section of RFC 6066:
```
   It is RECOMMENDED that clients include an extension of type
   "server_name" in the client hello whenever they locate a server by a
   supported name type.

   A server that receives a client hello containing the "server_name"
   extension MAY use the information contained in the extension to guide
   its selection of an appropriate certificate to return to the client,
   and/or other aspects of security policy.  In this event, the server
   SHALL include an extension of type "server_name" in the (extended)
   server hello.  The "extension_data" field of this extension SHALL be
   empty.
```

bvinc:
The more I think about this, the more I think #2 is the right solution.

I think I should be able to call ServerConfig::set_single_cert(...) and be able to serve a certificate, even if it doesn't match the SNI name that the client gave me.  The error `The server certificate is not valid for the given SNI name` is just a sanity check that isn't valid for all use cases.  It should be perfectly safe to delete or to be able to disable, this error.

briansmith:
> Imagine that I'm using no name-based virtual hosting of any kind and can just work with IP addresses.

Rustls can't support that scenerio because it doesn't support connecting to a server that doesn't have a domain name. This is made clearer in the current master branch, where the hostname input is a `DNSName` instead of a `&str`.

Adding support for IP addresses is a lot of work. It requires, in particular, adding IP address support to the webpki certificate validation library. And it requires that the CA actually issue certificates for IP addresses and that the CA actually encodes the IP addresses as IpAddress subjectAltName entries.

> To be able to disable SNI in the client

[...]

PR #126, opened 2017-10-23
Don't verify SNI names match the certificate on the server side

bvinc:
I talked about this in issue #123.  @briansmith didn't seem to think this change is a good idea.  I just wanted to lay out what change I think is best, and why I think it's reasonable.

I realize that this change makes absolutely no sense to anyone doing HTTPS with normal PKI certificate verification.  But I feel that to be a generic TLS library, there must be some way to support having a server that sends whatever certificate is necessary and allow the client to verify it in whichever way it wants.

This change should make the following possible:

*  A server can send certificates without sub_alt_names (client can verify other ways)
*  A protocol that's IP based, instead of name-based (where any name that resolves works)
*  A protocol that can *optionally* use names provided by the client (or choose not to)
*  An HTTPS server may actually *want* to send an invalid certificate that causes an understandable client-side error "name doesn't match", rather than just failing the TLS negotiation.
*  `ServerConfig::set_single_cert(...)` can actually be used now to do what it was intended to do (always serve the same cert no matter what SNI says)

Downsides:

*  99% of the time, this error just indicates that a server just chose the wrong certificate, or the certificate used wasn't created correctly.  The previous code catches this case a little earlier instead of later on the client side.

### DIFF
--- a/src/server/hs.rs
+++ b/src/server/hs.rs
@@ -912,26 +912,14 @@ impl ExpectClientHello {
         })?;
 
         // Reject syntactically-invalid end-entity certificates.
-        let end_entity_cert = webpki::EndEntityCert::from(
-              untrusted::Input::from(end_entity_cert.as_ref())).map_err(|_| {
+        webpki::EndEntityCert::from(untrusted::Input::from(end_entity_cert.as_ref())).map_err(|_| {
             sess.common.send_fatal_alert(AlertDescription::InternalError);
             TLSError::General(
                 "end-entity certificate in certificate chain is syntactically invalid".to_string())
         })?;
 
         if let Some(ref sni) = sni {
-            // If SNI was offered then the certificate must be valid for
-            // that hostname. Note that this doesn't fully validate that the
-            // certificate is valid; it only validates that the name is one
-            // that the certificate is valid for, if the certificate is
-            // valid.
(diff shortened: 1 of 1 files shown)


ctz:
I'd like not to get rid of the current behaviour for the 99% of users to whom it's relevant and potentially helpful.  I'd be happy to accept a new bool member of `ServerConfig` which elides the cert well-formedness and SNI cross-check though. WDYT?

bvinc:
Personally, I would lean towards just removing it for 2 reasons:

1.  Complexity.  Less code, less behavior to worry about and maintain, less options in ServerConfig, less documentation about ServerConfig.
2.  Surprise.  I honestly never would have expected my TLS library to be checking the names on my certificates.  If I wrote code to always send a certain certificate, or used `set_single_cert(...)`, I would have expected it to actually send the certificate, and not fail the TLS handshake.  Maybe that's just me though?

But I actually don't have any strong opinions on this.  If you think it should go into ServerConfig, I'll update the PR.  So you think this single boolean should disable the other well-formedness checks too? (empty cert chain, syntactically-invalid end-entity certs)  I'm not sure if I would have much use for that, but maybe it makes sense to group those together.

ctz:
I've merged this to master, but the organisation is different. You will want to set the `opaque_certificate` flag on the `CertifiedKey` your `ResolvesServerCert` returns.

bvinc:
I like it!

Issue #130, opened 2017-11-01
Ignore SNI values entirely when using set_single_cert


matthauck:
It was a surprise to me when using `set_single_cert` when the rustls refused to handshake when the client-presented SNI did not match a subjectAltName in my certificate. It seem better to me to allow the client to decide to refuse to handshake than have a server that refuses to even present its server certificate if SNI does not match, in particular when `set_single_cert` is used for the `ServerConfig`.

sanmai-NL:
I wonder what the security implications are of such a change in design. One thing is clear though, I also had trouble debugging thsi simple issue (#129), and @P-E-Meunier too (#127), all in recent days. @ctz may want to address the challenge library consumers appear to have here. I'm personally looking into whether the docs can be improved.

matthauck:
The design decision here essentially is whether the server has the responsibility (or right?) to make this decision on behalf of the client or not. 

In my opinion:
- The job of the server is to tell the client its identity
- The job of the client is to decide whether to trust the server's identity

Refusing to even `Server Hello` because the SNI doesn't match prevents the client from making this decision, presumably trying to take the high ground since so many clients out there don't validate certs? 

I think that doing this choice only for `set_single_cert` makes sense since that `ServerConfig` is not asking the library to choose from a set of identities based on SNI, it just has a single identity and should always return it. 

I think this is in line with the RFC:

> It is NOT RECOMMENDED to send a warning-level unrecognized_name(112) alert,
because the client's behavior in response to warning-level alerts is
unpredictable.  If there is a mismatch between the server name used
by the client application and the server name of the credential
chosen by the server, this mismatch will become apparent when the
client application performs the server endpoint identification, at
which point the client application will have to decide whether to
proceed with the communication.

(https://tools.ietf.org/html/rfc6066#section-3)

(Another minor point here from re-reading this is that `unrecognized_name` alert is probably preferable to InternalError)

ctz:
> Refusing to even Server Hello because the SNI doesn't match prevents the client from making this decision, presumably trying to take the high ground since so many clients out there don't validate certs?

It's more to make a server misconfiguration more obvious to the server admin.  Now, failing during a handshake is not an ideal time, but is also somewhat unavoidable given it uses the SNI value the client sends.

I wonder if this arrangement would be better, though:

1. keep `server::ResolvesServerCert` as is (a low-level trait for totally reconfiguring certificate selection logic); make the main library not second-guess or check the results.
[...]

PR #131, opened 2017-11-01
Add debug log when certificate does not match SNI

matthauck:
This would have it easier to figure out what was behind #130.


ctz:
Ah, this is a sensible thing to do, yes. Thanks!

Issue #129, opened 2017-10-31
v0.11.0: Failure to establish connections given custom cert chain


sanmai-NL:
A `rustls` (v0.11.0) based `hyper` server fails to establish TLS after switching from the sample keypair to a custom root and sub CA-signed cert chain. The connections are initiated from Chromium and Firefox. I wonder what the error is since the error message is quite opaque client side, and server side no error is thrown apparently.

```sh
  cargo +stable run --example tlsclient -- --http --verbose -p 8082 127.0.0.1
```

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/tlsclient --http --verbose -p 8082 127.0.0.1`
INFO:rustls::client_hs: No cached session for "127.0.0.1"
INFO:rustls::client_hs: Not resuming any session
DEBUG:rustls::client_hs: Sending ClientHello Message {
    typ: Handshake,
    version: TLSv1_0,
    payload: Handshake(
        HandshakeMessagePayload {
            typ: ClientHello,
            payload: ClientHello(
                ClientHelloPayload {
                    client_version: TLSv1_2,
                    random: Random(
                        [
                            214,
                            231,
                            164,
                            59,
                            32,
                            8,
                            26,
                            132,
                            202,
                            54,
                            28,
                            44,
                            215,
                            58,
                            187,
                            89,
                            209,
                            197,
                            22,
                            121,
                            216,
                            137,
                            28,
                            35,
                            90,
                            115,
                            118,
                            73,
                            102,
                            50,
                            227,
                            136
                        ]
                    ),                                                                                                                                                    
                    session_id: SessionID,                                                                                                                                
                    cipher_suites: [                                                                                                                                      
                        TLS13_CHACHA20_POLY1305_SHA256,                                                                                                                   
                        TLS13_AES_256_GCM_SHA384,                                                                                                                         
                        TLS13_AES_128_GCM_SHA256,                                                                                                                         
                        TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,                                                                                                    
                        TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,                                                                                                      
                        TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,                                                                                                          
                        TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,                                                                                                          
[...]

Issue #127, opened 2017-10-27
Handshake fails with BadDER


P-E-Meunier:
Hi!

My Tokio-based PostgreSQL crate https://nest.pijul.com/pmeunier/pleingres was originally based on Rustls, but I couldn't connect to a Google Cloud PostgreSQL database.

I replaced tokio-rustls with tokio-openssl, and it started working.

The error originated from webpki, and was BadDER. The CA certificate was from Google, and there was no other DER-encoded file involved (besides what's in webpki-roots maybe?).

Now that I got a working version, I can reproduce that bug if needed. With OpenSSL, I needed a function that doesn't check the hostname (the host I want to connect to has no hostname and is supposed to be accessed by its IP address).

P-E-Meunier:
The strange thing though, is that rustls doesn't complain when adding the certificate (`add_pem_file()`), it crashes during the negotiation.

The certificate is
```
-----BEGIN CERTIFICATE-----
MIIDITCCAgmgAwIBAgIBADANBgkqhkiG9w0BAQUFADBIMSMwIQYDVQQDExpHb29n
bGUgQ2xvdWQgU1FMIFNlcnZlciBDQTEUMBIGA1UEChMLR29vZ2xlLCBJbmMxCzAJ
BgNVBAYTAlVTMB4XDTE3MTAyNzE0NTAwNVoXDTE5MTAyNzE0NTEwNVowSDEjMCEG
A1UEAxMaR29vZ2xlIENsb3VkIFNRTCBTZXJ2ZXIgQ0ExFDASBgNVBAoTC0dvb2ds
ZSwgSW5jMQswCQYDVQQGEwJVUzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoC
ggEBAL/4CZtILHnNHrYpoTTgTQdWLT8l4gT4XBN4p+usg6g+8LwBwj/X8EPvH6rN
EcKt9yIIrVjooFEWjLMeQAwDHuUWATRpzLfE63J8uQDq4CgUFkkrhx/3FA+iKP96
wNB8F+BcxG3Fg5jWQ4ZNxJUYhXm+VNOu6pftyIVstrTzAs/E7+P50zcmjeXpattF
yeIIb9xmE7Kzs3YBp9QvzlE/KZ4xv3CuGNzjSE28E3OsNw02fhobxMX6MLQIkrWu
[...]

PR #132, opened 2017-11-01
Document SAN requirement on certs

sanmai-NL:
See #130, #129, #127.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -308,6 +308,10 @@ impl ServerConfig {
     /// Sets a single certificate chain and matching private key.  This
     /// certificate and key is used for all subsequent connections,
     /// irrespective of things like SNI hostname.
+    /// ***Note that the end-entity certificate must have the 
+    /// [Subject Alternative Name](https://en.wikipedia.org/wiki/Subject_Alternative_Name) 
+    /// extension to describe, e.g., the valid DNS name. The `commonName` field is 
+    /// disregarded.***
     ///
     /// `cert_chain` is a vector of DER-encoded certificates.
     /// `key_der` is a DER-encoded RSA private key.


src/server/mod.rs
```diff
@@ -308,6 +308,10 @@ impl ServerConfig {
     /// Sets a single certificate chain and matching private key.  This
     /// certificate and key is used for all subsequent connections,
     /// irrespective of things like SNI hostname.
+    /// ***Note that the end-entity certificate must have the 
+    /// [Subject Alternative Name](https://en.wikipedia.org/wiki/Subject_Alternative_Name) 
+    /// extension to describe, e.g., the valid DNS name. The `commonName` field is 
+    /// disregarded.***
```
briansmith: Please add a reference to [RFC 6125 Section 4.1](https://tools.ietf.org/html/rfc6125#section-4.1).

src/server/mod.rs
```diff
@@ -308,6 +308,10 @@ impl ServerConfig {
     /// Sets a single certificate chain and matching private key.  This
     /// certificate and key is used for all subsequent connections,
     /// irrespective of things like SNI hostname.
+    /// ***Note that the end-entity certificate must have the 
+    /// [Subject Alternative Name](https://en.wikipedia.org/wiki/Subject_Alternative_Name)
```
briansmith: Let's avoid using Wikipedia as a reference. RFC 6125 is a much better reference.

Also, I don't think this needs to be called out in bold or whatever `***` indicates.

sanmai-NL: Since multiple people were confused about this requirement within a short time frame (see PR OP), I thought the emphasis markup was warranted. Removed it now.

PR #139, opened 2017-12-07
Add Session::get_negotiated_ciphersuite()

kevinis:
rustls::SessionCommon already has a public get_suite().This PR adds a new API get_negotiated_ciphersuite() for ClientSession and ServerSession to access negotiated ciphersuite. 

This API panics if called prior to TLS negotiation, because SessionCommon:suite is None and SessionCommon::get_suite() unwraps it (see src/session.rs:375). 

The simpleclient example is modified to print the ciphersuite.

### DIFF
--- a/examples/simpleclient.rs
+++ b/examples/simpleclient.rs
@@ -7,6 +7,8 @@ extern crate rustls;
 extern crate webpki;
 extern crate webpki_roots;
 
+use rustls::Session;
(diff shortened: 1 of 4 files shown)


ctz:
Good idea. I have a TLS 1.3 draft 22 branch on the boil and will merge this after that.

PR #144, opened 2017-12-23
Make hostname argument optional

gh2o:
TLS transactions in non-Internet contexts often don't use hostnames. Support these usage scenarios by making the hostname optional, which disables the SNI extension and disables checking for the DNS name in the server certificate.

### DIFF
--- a/src/client/common.rs
+++ b/src/client/common.rs
@@ -54,11 +54,11 @@ pub struct HandshakeDetails {
     pub randoms: SessionRandoms,
     pub using_ems: bool,
     pub session_id: SessionID,
-    pub dns_name: webpki::DNSName,
+    pub dns_name: Option<webpki::DNSName>,
 }
 
 impl HandshakeDetails {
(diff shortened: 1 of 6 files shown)


ctz:
Disabling hostname validation so readily is against one of the main design goals of this library.

Instead, you can achieve your goals already:

* Set `enable_sni` to false in your `ClientConfig`.
* Provide an implementation of the `rustls::verify::ServerCertVerifier` trait and provide that to `ClientConfig::dangerous().set_certificate_verifier()`. That will require building rustls with the `dangerous_configuration` feature.

I hope that helps; let me know if you have problems.

gh2o:
Would it make sense to put this behind `dangerous()` too? Reimplementing the provided `ServerCertVerifier` seems like it's quite a bit of code duplication; it might make more sense to just skip DNS name validation based on a bool in ClientConfig.

gh2o:
Will open as separate PR.

PR #146, opened 2017-12-27
Add "dangerous" option to disable hostname verification

gh2o:
TLS transactions in non-Internet contexts often don't use hostnames. Support these usage scenarios by making the hostname verification of the server certificate optional if the dangerous_configuration feature is enabled.

### DIFF
--- a/src/client/hs.rs
+++ b/src/client/hs.rs
@@ -1147,7 +1147,11 @@ impl State for ExpectTLS13CertificateVerify {
             .get_verifier()
             .verify_server_cert(&sess.config.root_store,
                                 &self.server_cert.cert_chain,
-                                self.handshake.dns_name.as_ref(),
+                                if sess.config.verify_hostname {
+                                    Some(self.handshake.dns_name.as_ref())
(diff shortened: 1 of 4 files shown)


ctz:
I'm afraid this is still problematic for me. The ultimate aim of all the certificate checking is to see that one of the root certificates signed a certificate for example.com. Without the hostname check, all the of the other remaining checks don't achieve much. The supported way to remove all these checks is to provide a `rustls::verify::ServerCertVerifier` which does nothing. At which point, you may consider some kind of trust-on-first-use style pinning of the whole certificate chain for these devices.

I know that implementing a do-nothing `ServerCertVerifier` is kind of annoying, but in this project I'm trying to make it easy to do common things, and not-easy to do uncommon/unsafe things. So I think the cost of ~10 lines of code in every project that wants to do something uncommon is OK.

gh2o:
That's a fair rationale. The reason I wanted to do this is to avoid having to have a direct dependency on webpki (since I am still validating the cert chain).

PR #159, opened 2018-03-31
Prevent Stream::read() from prematurely returning 0

tchebb:
Previously, it was possible for `Stream::read()` to return 0 in cases where the underlying socket had not signalled EOF. This broke functions such as `read_exact()`, which expects 0 to be returned only on EOF.

The bug occurred when only a partial TLS packet was available on the underlying socket. This caused `complete_io()` to return successfully, since it read some data, but failed to get more plaintext, since a partial packet cannot be decoded.


ctz:
Thanks!

Issue #161, opened 2018-04-03
Implement support for QUIC


djc:
I have started work on a pure-Rust implemenation of the QUIC protocol (current draft). I would like to use `rustls` for all TLS-related stuff. Are you interested in having that support implemented? QUIC currently specifies TLS 1.3 only, but there are some extensions that are needed. In particular, so far I've noticed the `TransportParameters` extension as described [in the draft spec](https://tools.ietf.org/html/draft-ietf-quic-transport-10#section-7.4). I can work on this, but would appreciate your support/guidance where necessary.

ctz:
Neat. Yes.

Perhaps we could have traits for client and server starting like:

```
mod quic {
pub trait QuicClient {
  fn get_transport_params(&self) -> msgs::quic::ClientTransportParameters;
  fn process_server_transport_params(&self, params: &msgs::quic::ServerTransportParameters) -> Result<(), TLSError>;
}

pub trait QuicServer {
  fn process_client_transport_params(&self, params: &msgs::quic::ClientTransportParameters) -> Result<msgs::quic::ServerTransportParameters, TLSError>;
}
}
```

... then extended as required for stuff like the Cookie extension (which AIUI is how source address ownership is established, and how stateless retries work).

These `msgs::quic` types can be plumbed into the existing `msgs::handshake::ClientExtension` etc types, but we can keep `TransportParameterId` opaque as a `u16`? There's no need for rustls to track that enum.

It probably makes most sense for these trait impls to be per-session (rather than per-ClientConfig/ServerConfig) which should make it easiest to keep the right context around to fulfil these calls. That will mean we'll need extra QUIC-specific ClientSession/ServerSession constructor-like functions.

All this can go under a `quic` feature, as most users won't benefit from the extra API surface.

ctz:
Related, see #151

djc:
Sounds good so far. I've submitted #162 for the basic handling of `TransportParameters`.

What do you think is the best way to plug into the handshake infrastructure? For example, I don't think we'd want to duplicate most of the stuff in `emit_client_hello_for_retry()`, but I'm not sure to what degree the QUIC side will differ.

Ralith:
I'm also working on [a QUIC implementation](https://github.com/Ralith/quicr) (perhaps me and @djc can combine efforts?), currently using OpenSSL largely due to #151. My implementation is successfully completing the cryptographic handshake.

[...]

PR #162, opened 2018-04-04
Implement support for the QUIC TransportParameters extension

djc:
See #161. Here's my initial take at implementing basic support for TransportParameters.


ctz:
Thanks! :)

Ralith:
How does external code actually submit/access values for this extension?

djc:
That hasn't been implemented yet, but I have a local branch I'll submit soon.

PR #163, opened 2018-04-04
[WIP] Use Codec trait for unsigned integer types

djc:
Have you thought about doing something like this? I noticed that you have this nice `Codec` trait, but then `u8` .. `u64` don't implement that and have functions instead. It seems nice to use the trait everywhere, which groups the implementations and would enable using the `Vec` stuff also for things that have one of these types as the item type. (Similarly, it seems you might be able to leverage the type system to abstract `Codec` over the `Vec` type more generally, and potentially do away with the `PayloadU*` type in favor of `Codec` implementations and good type bounds.)


ctz:
Yes, this is an excellent idea. I'll run with this.

djc:
Cool! I also don't mind doing some more of the legwork tomorrow, let me know.

ctz:
I've just pushed this and other same thing for other types. It's cleaned up quite a bit of other code, so thanks!

I haven't made use of this yet to reduce vectors-of-Codec-types boilerplate -- feel free to try that?

djc:
Hmm, that's quite a bit harder than I thought... The `u24` type being different makes it hard to neatly abstract over it, and I'm also kind of missing some const generics things. Better to mostly leave it as is for now, I guess.

PR #166, opened 2018-04-08
Add support for newer TLS 1.3 drafts

djc:


ctz:
I've added some extra commits to this which move the bogo test suite toward draft-28 support.

djc:
Great! Looks like there's some more work to do with the `online` tests.

djc:
It looks to me like the bogo test suite is somehow not getting draft 28 support -- it looks like that is optional in their client anyway, and maybe should be enabled separately?

From the `topsites` tests, it also looks like those remotes just don't support draft 28 yet.

ctz:
Clarification: the draft-26 change was only intended to affect the server-end. I've dropped those changes from here and pushed jbp-tls13-draft-28.

djc:
Sounds good! What's the plan for merging that to master?

ctz:
This is useful to have ready, but for testing reasons I'm going to wait until boringssl moves to draft-28 as the primary TLS1.3 version before merging this.

djc:
Okay! If you want, you can just close this PR.

est31:
>  I'm going to wait until boringssl moves to draft-28 as the primary TLS1.3 version before merging this.

This is the case now, as of [this boringssl commit](https://boringssl.googlesource.com/boringssl/+/fec83fc78d489a8d444c3a99c51101fdf54d5b71)

ctz:
I've merged this, so thanks for the hard work that went in. We needed a separate patch against bogo to make all the TLS1.3 tests use draft 28, but in the end that was easy to do sed.

djc:
Are you planning a release for when TLS 1.3 comes out? Would be nice to have.

PR #171, opened 2018-05-08
Add support for the SSLKEYLOGFILE environment variable

djc:
The format is documented here:

https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format


djc:
Not sure if you want a feature gate for this? This has already helped me debug my QUIC implementation, and I remember somebody on IRC asking about this a long time ago. I know it's enabled by default in for example Firefox release builds.

ctz:
Thanks, this is a useful feature. I'm going to merge this with a few changes:

- put it behind a trait, so it can be tested easily
- put it at the Server/ClientConfig level. As it currently stands each session (say in the same single server process) each makes its own `KeyLogFile`, opening the same file a bunch of times.

PR #172, opened 2018-05-15
Update *ring* and webpki dependencies.

briansmith:
*ring* 0.13.0 will be released soon. There have been *many* changes
between *ring* 0.13.0-alpha and 0.13.0-alpha2 so there will be a
pre-release testing period.


ctz:
Thanks!

PR #178, opened 2018-06-16
Add `BulkAlgorithm` to the public API.

briansmith:
Allow users of Rustls to use `BulkAlgorithm` to filter the cipher
suites in their configurations.


ctz:
Merged as d6f5c32

PR #179, opened 2018-06-16
Do less unwrapping.

briansmith:


briansmith:
I did this while I was working on something else that might end up being a dead end.

PR #182, opened 2018-07-02
Send only info log for CloseNotify

sorpaas:
According the RFC 5246 section 7.2 (https://tools.ietf.org/html/rfc5246#section-7.2), CloseNotify is a normal alert that notifies that the connection is being closed. Setting this to `warn!` makes it appear fairly often in end programs, but serves no actual "warning". This PR changes the CloseNotify log level to `info!`.


ctz:
Thanks!

PR #187, opened 2018-07-22
Raw handshake I/O for QUIC

Ralith:
This works towards supporting [the new QUIC TLS interface](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#rfc.section.3), wherein QUIC subsumes the TLS record layer, taking over responsibility for fragmentation, encryption/decryption, and communication of alerts.

While `read_hs` as written should handle incoming data correctly, I'm unsure how to proceed with handling outgoing handshake data and alerts. The current implementation performs encoding of alerts, encryption, and fragmentation eagerly, all of which are undesired.

Ideally I'd like to defer these operations to being performed on demand, so that the QUIC interface can easily pass the data out as-is, while the regular interface performs the transformations necessary for conventional use. This seems likely to be a more invasive change, so before I proceed I'd like feedback on the idea, and input on how to approach implementing it.

This interface will likely need to be further extended with helpers to perform [the necessary key derivation](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#rfc.section.5.1), which uses different labels than that of TLS.

### DIFF
--- a/src/client/hs.rs
+++ b/src/client/hs.rs
@@ -20,7 +20,7 @@ use msgs::codec::Codec;
 use msgs::persist;
 use msgs::ccs::ChangeCipherSpecPayload;
 use client::ClientSessionImpl;
-use session::SessionSecrets;
+use session::{SessionSecrets, Protocol};
 use key_schedule::{KeySchedule, SecretKind};
 use cipher;
 use suites;
@@ -30,6 +30,8 @@ use rand;
 use ticketer;
 use error::TLSError;
 use handshake::{check_message, check_handshake_message};
+#[cfg(feature = "quic")]
+use quic;
 
 use client::common::{ServerCertDetails, ServerKXDetails, HandshakeDetails};
 use client::common::{ClientHelloDetails, ReceivedTicketDetails, ClientAuthDetails};
@@ -202,6 +204,10 @@ struct ExpectServerHello {
 struct ExpectServerHelloOrHelloRetryRequest(ExpectServerHello);
 
 fn emit_fake_ccs(hs: &mut HandshakeDetails, sess: &mut ClientSessionImpl) {
+    #[cfg(feature = "quic")] {
+        if let Protocol::Quic = sess.common.protocol { return; }
+    }
+
     if hs.sent_tls13_fake_ccs {
         return;
     }
@@ -592,6 +598,14 @@ impl ExpectServerHello {
                                 &read_key);
         sess.common.get_mut_key_schedule().current_server_traffic_secret = read_key;
 
+        #[cfg(feature = "quic")] {
+            let key_schedule = sess.common.key_schedule.as_ref().unwrap();
+            sess.common.quic.hs_secrets = Some(quic::Secrets {
+                client: key_schedule.current_client_traffic_secret.clone(),
+                server: key_schedule.current_server_traffic_secret.clone(),
+            });
+        }
+
         Ok(())
     }
 
@@ -698,13 +712,6 @@ impl State for ExpectServerHello {
         // Extract ALPN protocol
         if !sess.common.is_tls13() {
             process_alpn_protocol(sess, server_hello.get_alpn_protocol())?;
-        }
-
-        // Reject QUIC if TLS1.2 is in use.
-        if !sess.common.is_tls13() &&
-            server_hello.find_extension(ExtensionType::TransportParameters).is_some() {
-            sess.common.send_fatal_alert(AlertDescription::UnsupportedExtension);
-            return Err(TLSError::PeerMisbehavedError("server wants to do quic+tls1.2".to_string()));
         }
 
         // If ECPointFormats extension is supplied by the server, it must contain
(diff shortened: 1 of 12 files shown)


src/key_schedule.rs
```diff
@@ -40,13 +41,17 @@ pub struct KeySchedule {
...
     pub current_server_traffic_secret: Vec<u8>,
     pub current_exporter_secret: Vec<u8>,
 }
 
+const TLS13_LABEL_PREFIX: &[u8] = b"tls13 ";
+const QUIC_LABEL_PREFIX: &[u8] = b"quic ";
+
 impl KeySchedule {
-    pub fn new(hash: &'static digest::Algorithm) -> KeySchedule {
+    pub fn new(hash: &'static digest::Algorithm, quic: bool) -> KeySchedule {
```
djc: Maybe just take the prefix directly here instead of a bool?

djc: In fact, I'm not even clear whether it makes sense for `quic` or `prefix` to be part of the `KeySchedule` state; it seems strictly an argument for derivation. As long as all the derive callers have access to the right prefix, that might be a cleaner solution.

src/client/hs.rs
```diff
@@ -980,6 +980,8 @@ impl ExpectTLS13Certificate {
 }
 
 impl State for ExpectTLS13Certificate {
+    fn details(&self) -> &HandshakeDetails { &self.handshake }
```
ctz: I think that instead of exposing these parts of the handshake layer for every handshake state, we can move the transcript "upwards" out of `HandshakeDetails` into `ClientSessionImpl`? Or, if we need to do the same for server-side support, into `SessionCommon`

src/key_schedule.rs
ctz: I'd agree with the key schedule knowing the protocol. But I think I'd always prefer a very simple enum to a bool for this kind of thing, because bools are vague at call sites. Then we can give the enum an `Into<&'static [u8]>` implementation to get the actual prefix bytes.

Ralith:
Thanks for the feedback! I'll dig into the proposed changes later today.

Ralith:
Applied the requested changes. I'll now start filling out the remaining key derivation functionality.

Still hoping to hear some thoughts on the refactoring needed for getting outgoing messages and alerts prior to encryption and fragmentation.

Ralith:
Tests currently failing on `QUICTransportParams-{Client,Server}-Rejected-TLS12`. I'm not sure where those are coming from; I thought I removed them. Those tests make no sense to me, because QUIC does not use TLS1.2 at all: a QUIC implementation is required to reject attempts to negotiate a TLS 1.2 connection regardless of the presence of transport parameters.

ctz:
> Still hoping to hear some thoughts on the refactoring needed for getting outgoing messages and alerts prior to encryption and fragmentation.

If I were doing this I'd extract a trait from the necessary functions on `SessionCommon` (like `send_msg`, `send_msg_encrypt`, etc.) and move data they require (the sequence numbers, message fragmenter, encryption/decryption things, etc.) into a struct that implements that trait.

Ralith:
> I'd extract a trait from the necessary functions

And then have a no-op implementation for QUIC? Hmm, that could work. I wonder if it would be cleaner  to instead defer all encryption and fragmentation as late as possible (i.e. perform it on-demand inside `SessionCommon::write_tls`), and thereby keep a common code path and dodge any virtual dispatch.

djc:
> I wonder if it would be cleaner to instead defer all encryption and fragmentation as late as possible (i.e. perform it on-demand inside SessionCommon::write_tls), and thereby keep a common code path and dodge any virtual dispatch.

This sounds like it would map better to what QUIC is doing these days.

Ralith:
After some discussion with @djc we realized there's a much less invasive approach to handling outgoing messages. I think the necessary API is now fully prototyped; next steps should be to try to use it in practice, and gather any further feedback.

Ralith:
It would be nice if `TLSError` had `pub fn alert(&self) -> Option<AlertDescription>` or similar; in that case, `QuicExt::take_alert` could probably be removed in favor of inspecting the return value of `QuicExt::read_hs`.

Ralith:
Travis seems to be failing because `QUICTransportParams-Client-TLS13` constructs configurations that accept TLS versions other than 1.3. Is that intended?

Ralith:
This now seems to be working for Quinn, though I remain mystified as to why travis is testing a TLS1.3 property with TLS1.2.

Ralith:
Quinn is now employing the full breadth of this API to successfully communication with other implementations speaking the current draft, which is fortunately less invasive than draft-13 was. I think we're ready for another round of review.

ctz:
This looks all reasonable to me, code-wise. The things I'd like to see are some basic tests that exercise the new public API (in `tests/api.rs` or perhaps `tests/quic.rs`). Mainly so I can be sure I don't break this code in the future without resorting to building and testing dependent projects.

Ralith:
Good thought, I'll bang that out.

Reviewing this code, I've also been thinking about making `SessionCommon::quic` an `Option<quic::State>` that only exists if the feature is enabled, rather than the current approach of having a separate `Protocol` enum. `Protocol` made more sense when `KeySchedule` had to be aware of the state, but that's no longer necessary.

Ralith:
This test covers everything except `get_early_secret`, which I don't see a good way to test until rustls itself supports enabling 0-RTT on the server

Ralith:
I managed to fill in the necessary parts for QUIC 0-RTT support without too much disruption. I also added the transport parameters to client session storage under QUIC, because this is always required and is very simple for rustls to provide.

ctz:
This'll get merged tomorrow.

Ralith:
Rebased, and added some last-minute internal fixes isolated into the final 3 commits for easy review.

Note that most of the server-sized 0-RTT changes here are generally applicable to TLS1.3, and should be pulled out of the feature guards if/when conventional TLS1.3 0-RTT is implemented.

Still dunno what's going on with the bogo tests trying to combine TLS1.2 with QUIC, but presumably it's a trivial fix.

PR #195, opened 2018-09-23
KTLS Support

quininer:
This PR allows the we to get session secrets and sequence numbers, in order to use rustls as a handshake for KTLS.


ctz:
Out of interest is there a downstream crate using this?

I think it might be preferable to feature-gate these APIs for documentation purposes. What do you think?

quininer:
Do you mean add a feature-gate and add more documentation?

quininer:
I am not satisfied with this PR, I will open an issue.

PR #201, opened 2018-11-10
Move HandshakeHash into SessionCommon

Ralith:
This allows for easy access from the QUIC interface for key derivation, reduces some duplication between client/server data structures, and should not have any semantic impact. Merging this separately will drastically reduce the difficulty of keeping #187 rebased as development on master continues.


ctz:
Thanks!

PR #205, opened 2018-12-30
Expose ALPN protocols as bytes instead of String (fixes #115)

### LINKED ISSUE #115: ALPN API is defined in terms of UTF-8 strings, but ALPN isn't necessarily UTF-8.
The API uses the &str and String types, which means it only works with ALPN protocol identifiers that are valid UTF-8 strings. However, the ALPN protocol specifies that they are actually octet strings with no encoding.

I suggest that we keep the string-based API, but document & test what happens when the peer sends us non-UTF-8 strings.

Testing: should test that we treat a protocol name that isn't valid UTF-8 the same way we treat a protocol name that is valid UTF-8 but not recognized.

Documentation: If the API remains UTF-8 then we'll never support, nor ever negotiate, a protocol with a non-UTF-8 name. We should document this so that people don't worry that Rustls is overlooking the possibility of non-UTF-8-named protocols.

djc:


ctz:
Merged, thanks!

PR #212, opened 2019-01-09
Use *ring* 0.14, along with new versions of ct-logs, sct.rs, webpki, and webpki-roots.

briansmith:


lucab:
I was wondering about travis failure, and it turns out this depends on https://github.com/ctz/sct.rs/pull/3 and a new `sct` release.

briansmith:
Yes, this will fail CI unless/until the updated dependent packages webpki-roots, sct.rs, and ct-logs are released.

PR #216, opened 2019-01-14
Reexport `WriteEarlyData`

jonas-schievink:
Otherwise, the struct can't be named and doesn't show up in generated docs,
which makes the `bytes_left` method very hard to discover.


ctz:
Thanks!

PR #218, opened 2019-01-18
Don't send duplicate fatal alerts on bad client version

Ralith:
This avoids providing an incorrect alert code to QUIC (which assumes at most one alert) and skips unnecessary work for regular TLS.


ctz:
Thanks, raised #222

PR #226, opened 2019-01-27
Don't advertise SHA-1 support.

khuey:
1be8320fc9b636b3b9afd7d14d38ebc807d95484 removed support for using SHA-1 but it didn't remove advertising support for SHA-1. Empirically, connecting to sns.us-east-2.amazonaws.com has been found to select PKCS1_SHA1 if it is advertised.

Cc @briansmith


src/msgs/handshake.rs
```diff
@@ -281,8 +281,6 @@ impl SupportedMandatedSignatureSchemes for SupportedSignatureSchemes {
             SignatureScheme::RSA_PKCS1_SHA512,
             SignatureScheme::RSA_PKCS1_SHA384,
             SignatureScheme::RSA_PKCS1_SHA256,
-
-            SignatureScheme::RSA_PKCS1_SHA1,
```
briansmith: :sigh: Sorry I missed this.

IMO, this is a symptom of a bigger problem in Rustls: Rustls defines enum values for every algorithm, cipher suite, TLS version, etc., but only implements a subset of them. AFAICT it would be much better if each enum had its values limited to the set actually implemented by Rustls. Then this bug wouldn't have occurred.

ctz:
Thanks, I've merged this. Sorry for not catching that earlier.

At the same time I've moved the various lists of supported algorithms into the module that actually deals with that. It was a mistake that policy-like information like this was in a very much mechanism orientated module like `msgs::handshake`.

ctz:
(Will push a point release with this fix in too.)

ctz:
https://crates.io/crates/rustls/0.15.1

khuey:
Thanks.

PR #231, opened 2019-02-03
0-RTT bugfixes

Ralith:
Fixes a handshake failure when a 0-RTT capable server receives a non-0-RTT resumption.


Ralith:
Interop testing uncovered what appears to be a bug in rustls's pre-existing obfuscated_ticket_age computation, which likely affects client 0-RTT use in TLS1.3 as well.

ctz:
Thanks!

PR #241, opened 2019-03-08
Migrate examples into a subcrate

erickt:
Fuchsia doesn't support yet mio, and there is unfortunately no way to run the rustls on Fuchsia as the crate is currently setup. Unfortunately there is no way to make dev-dependencies optional (see
 https://github.com/rust-lang/cargo/issues/1596), so this patch migrates the examples into a subcrate so we can compile the rustls tests without mio.

Note that this is built upon #239, I broke them apart since I figure #239 was uncontroversial, but you might have concerns about this PR.


ctz:
This is quite tricky, because some of the integration-level tests use the example programs (and hence mio). I'll merge changes based on this, though, with just the mio-using stuff in a subcrate.

erickt:
Yeah I figured that one out after I did a cargo clean and saw my builds were failing. Im in progress in a PR that's factoring out those tests into a subcrate.

PR #248, opened 2019-05-08
Fix the unused_doc_comment warning

thomwiggers:
This moves the comment into the macro, allowing to document the generated
enums.

Goes with a PR in the [ctz/tls-hacking](/ctz/tls-hacking) repo


ctz:
Thanks! I think this warning is quite new on nightly, since I only started seeing it about a week ago.

PR #254, opened 2019-05-27
stream: add socket accessors

Keruspe:
The other TLS libraries all provide those methods on their
stream implementations, this helps make things more similar


ctz:
Thanks!

PR #255, opened 2019-05-27
Update minimal dependencies versions

Keruspe:
Doesn't build (at least with a recent rustc) with minimal versions without those bumps.

Add CI target to make sure we update the minimal bounds when needed.


ctz:
Thanks!

PR #257, opened 2019-06-11
Avoid Error fmt

thomwiggers:
Rust change https://github.com/rust-lang/rust/pull/60897 changed the output, making this test fail on nightly


ctz:
Thank you!

PR #258, opened 2019-06-11
Fix warnings about trait objects

thomwiggers:
Trait objects should now have ``dyn`` in types.

Test will fail as per #257


ctz:
Thanks!

PR #267, opened 2019-07-11
Remove QUIC-specific key logging labels

Ralith:
Wireshark decided these were no longer appropriate since most of the divergence between conventional TLS and QUIC crypto was removed.


Ralith:
CI failure is https://github.com/rust-lang/rust/pull/62574

Ralith:
Rebased. Hopefully we can get this in before there's too much more churn?

Ralith:
Rebased with some additional cleanup.

PR #274, opened 2019-07-22
Expose resumption ciphersuite

Ralith:
Needed for QUIC to select the correct 0-RTT header protection cipher, fixing an interop issue with the quant implementation. Please sanity-check the adjustment to the "server varied selected ciphersuite" error detection.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -374,6 +374,7 @@ pub struct ClientSessionImpl {
     pub state: Option<Box<dyn hs::State + Send + Sync>>,
     pub server_cert_chain: CertificatePayload,
     pub early_data: EarlyData,
+    pub resumption_ciphersuite: Option<&'static SupportedCipherSuite>,
 }
 
 impl fmt::Debug for ClientSessionImpl {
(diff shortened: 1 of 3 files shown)


Ralith:
Okay, there's clearly some subtleties to the interaction between `set_suite` and server ciphersuite change validation that I don't fully grasp. @ctz, could you provide any guidance here?

djc:
Just to make it explicit, clearly we hope this could make it into the/an upcoming release. :)

Ralith:
I changed the approach here to one that bypasses the set_suite machinery entirely. I'm not sure it's the most elegant approach, but it's very simple and avoids interacting with the pre-existing logic entirely, and therefore fixes the test failures.

ctz:
Thanks. I may change the public API here slightly in the future but I'll let you know if/when that happens. Mainly because a) it's a bit weird that `get_negotiated_suite()` is filled in before any traffic has been exchanged with the server, and b) the value it returns can change if the server rejects 0RTT and opts for a full handshake with different ciphersuite.

Ralith:
Yeah, I thought about introducing a new function for it, but ultimately I just erred on the approach with less API impact. It would be trivial to adapt Quinn to use such an API.

PR #278, opened 2019-07-23
Simplify construction of AAD and Nonce values.

briansmith:


briansmith:
Note: Instead of doing this nonce stuff, it would be straightforward to use `OpeningKey` and `SealingKey` with a custom `NonceSequence` implementation based on the `make_tls13_nonce()` I wrote here. However, my understanding is that you explicitly do prefer to use `LessSafeKey` instead, so I avoided including those changes here.

ctz:
Thanks!

PR #279, opened 2019-07-25
Start using the *ring* 0.16.4 HDKF API in the TLS 1.3 key schedule

briansmith:
Eliminate some of the inefficiency of the current key schedule  implementation by reducing `Vec<u8>` usage and reducing the number of times an `hmac::Key` is constructed for a given value. Further changes should be done, but those require changing the QUIC API.

### DIFF
--- a/src/key_schedule.rs
+++ b/src/key_schedule.rs
@@ -1,6 +1,6 @@
 /// Key schedule maintenance for TLS1.3
 
-use ring::{hkdf::{self, KeyType as _}, hmac, digest};
+use ring::{aead, hkdf::{self, KeyType as _}, hmac, digest};
 use crate::error::TLSError;
 use crate::cipher::{Iv, IvLen};
 use crate::msgs::base::PayloadU8;
@@ -220,9 +220,10 @@ impl From<hkdf::Okm<'_, PayloadU8Len>> for PayloadU8 {
     }
 }
 
-pub fn derive_traffic_key(secret: &hkdf::Prk, len: usize) -> Vec<u8> {
-    let payload: PayloadU8 = hkdf_expand(secret, PayloadU8Len(len), b"key", &[]);
-    payload.into_inner()
+pub fn derive_traffic_key(secret: &hkdf::Prk, aead_algorithm: &'static aead::Algorithm)
+    -> aead::UnboundKey
+{
+    hkdf_expand(secret, aead_algorithm, b"key", &[])
 }
 
 pub(crate) fn derive_traffic_iv(secret: &hkdf::Prk) -> Iv {
@@ -232,7 +233,7 @@ pub(crate) fn derive_traffic_iv(secret: &hkdf::Prk) -> Iv {
 #[cfg(test)]
 mod test {
     use super::{KeySchedule, SecretKind, derive_traffic_key, derive_traffic_iv};
-    use ring::hkdf;
+    use ring::{aead, hkdf};
 
     #[test]
     fn smoke_test() {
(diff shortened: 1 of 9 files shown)


src/key_schedule.rs
```diff
@@ -48,104 +36,85 @@ impl SecretKind {
...
     /// Derive a secret of given `kind` using the hash of the empty string
     /// for the handshake hash.  Useful only for
     /// `SecretKind::ResumptionPSKBinderKey` and
     /// `SecretKind::DerivedSecret`.
-    pub fn derive_for_empty_hash(&self, kind: SecretKind) -> Vec<u8> {
-        let mut empty_hash = [0u8; digest::MAX_OUTPUT_LEN];
-        empty_hash[..self.hash.output_len]
-            .clone_from_slice(digest::digest(self.hash, &[]).as_ref());
-
-        self.derive(kind, &empty_hash[..self.hash.output_len])
```
briansmith: This code looked weird to me. Maybe I'm misunderstanding it. Please double-check.

briansmith:
OK, I think this is as far as I'll take this.

Sometime in the near future, I'd like to change the QUIC interface so that we don't expose the raw bytes of these secrets to QUIC.

I'd like to change the way logging is done so that we don't construct a `Vec<u8>` for any key material unless a non-no-op log is in use. That way, in the typical case, the key material is never exposed within Rustls (it'd be all protected by *ring*'s key encapsulation), except during the initial key agreement (which is the next thing I'll look into encapsulating).

But, those two changes should happen separately from this PR.

src/key_schedule.rs
```diff
@@ -208,93 +162,94 @@ impl KeySchedule {
...
+    use ring::{aead, hkdf};
 
     #[test]
     fn smoke_test() {
         let fake_handshake_hash = [0u8; 32];
 
-        let mut ks = KeySchedule::new(&digest::SHA256);
-        ks.input_empty(); // no PSK
-        ks.derive(SecretKind::ResumptionPSKBinderKey, &fake_handshake_hash);
+        let mut ks = KeySchedule::new_with_empty_secret(hkdf::HKDF_SHA256);
```
briansmith: I'm not sure this `smoke_test()` test is even useful anymore, especially considering the existence of the other test. In particular, it's testing `derive_bytes` only but `derive_bytes` is the worst function to use.

src/suites.rs
```diff
@@ -156,6 +156,8 @@ pub struct SupportedCipherSuite {
     /// in a deterministic and safe way.  GCM needs this,
     /// chacha20poly1305 works this way by design.
     pub explicit_nonce_len: usize,
+
+    pub(crate) hkdf_algorithm: ring::hkdf::Algorithm,
```
djc: Unless/until rustls returns a `Prk` directly for the QUIC secrets, can we get a getter for this?

djc:
> Sometime in the near future, I'd like to change the QUIC interface so that we don't expose the raw bytes of these secrets to QUIC.

I've been looking at this. I think quinn could make do with getting a `Prk` directly. I suppose one of the sticky points here is how `derive_next()` will be implemented, which I don't think maps cleanly to the current `hkdf` API?

src/key_schedule.rs
ctz: Yes, it was weird. For no good reason.

ctz:
Thanks for your work on this.

ctz:
One future thing I'd like to do here is make the internals of `KeySchedule` internally have session types to prevent accidental or illegal derivation of secrets at the wrong time. eg, it should be impossible to derive a handshake encryption secret before having input the DH shared secret. I think these changes make that easier to do.

briansmith:
> One future thing I'd like to do here is make the internals of `KeySchedule` internally have session types to prevent accidental or illegal derivation of secrets at the wrong time. eg, it should be impossible to derive a handshake encryption secret before having input the DH shared secret. I think these changes make that easier to do.

FWIW, I have already implemented this, when designing the new HKDF API. Let's sync up.

PR #282, opened 2019-07-26
Better encapsulate key material during TLS 1.3 key schedule

briansmith:


src/client/tls13.rs
```diff
@@ -826,8 +828,12 @@ impl hs::State for ExpectFinished {
...
             let key = sess.common
                 .get_key_schedule()
-                .derive_bytes(SecretKind::ClientHandshakeTrafficSecret,
-                              &st.handshake.hash_at_client_recvd_server_hello);
+                .derive_logged_secret(
+                    SecretKind::ClientHandshakeTrafficSecret,
+                    &st.handshake.hash_at_client_recvd_server_hello,
+                    &*sess.config.key_log,
+                    sess.common.protocol.labels().client_handshake_traffic_secret,
+                    &st.handshake.randoms.client);
```
briansmith: With this change, this secret is logged earlier than it was previous. I don't know if this makes any difference.

src/key_schedule.rs
```diff
@@ -232,25 +233,7 @@ pub(crate) fn derive_traffic_iv(secret: &hkdf::Prk) -> Iv {
 mod test {
     use super::{KeySchedule, SecretKind, derive_traffic_key, derive_traffic_iv};
     use ring::{aead, hkdf};
-
-    #[test]
-    fn smoke_test() {
```
briansmith: I removed this test because I don't see its value given the other test. It was testing a function that's been removed.

src/quic.rs
```diff
@@ -9,17 +9,16 @@ use crate::key_schedule;
...
 
 use std::sync::Arc;
-use ring::hkdf::{self, KeyType as _};
+use ring::hkdf;
 use webpki;
-use crate::msgs::base::PayloadU8;
 
 /// Secrets used to encrypt/decrypt traffic
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Clone, Debug)]
```
briansmith: `Prk` doesn't implement `PartialEq` or `Eq` as of now.

briansmith: The way the code is structured requires `Prk` to implement `Clone` so I changed *ring* to do that. I hope we can eventually restructure things to avoid this cloning.

tests/api.rs
```diff
@@ -19,8 +19,6 @@ use rustls::TLSError;
 use rustls::sign;
 use rustls::{ALL_CIPHERSUITES, SupportedCipherSuite};
 use rustls::KeyLog;
-#[cfg(feature = "quic")]
-use rustls::quic::{self, QuicExt, ClientQuicExt, ServerQuicExt};
```
briansmith: Somebody needs to update the QUIC tests for this test. I just commented them out as I don't have the time or patience for it. Somebody will have to take this over from me.

tests/api.rs
```diff
@@ -1951,6 +1951,6 @@ fn assert_lt(left: usize, right: usize) {
 #[test]
 fn session_types_are_not_huge() {
     // Arbitrary sizes
-    assert_lt(mem::size_of::<ServerSession>(), 1536);
-    assert_lt(mem::size_of::<ClientSession>(), 1536);
+    assert_lt(mem::size_of::<ServerSession>(), 2300);
+    assert_lt(mem::size_of::<ClientSession>(), 2300);
```
briansmith: AFAICT, this is due to storing the secrets inline as `Prk` instead of on the heap inside `Vec<u8>`s. I think this is a good change.

As @ctz alluded to in the previous PR, we should change the key schedule to use session types. I think this will reduce the memory consumption here. If we need an immediate avoidance of the size increase, we could store the whole key schedule as a `Box<KeySchedule>` instead.

And/or, we could look into creating compressed key types in *ring*, which store just the raw key values instead of the expanded key. That wouldn't be something I can do immediately though.

briansmith:
@djc Please take a look. Note that in the very newest version of *ring* (released today), you can derive `HeaderProtectionKey` in the type-safe and encapsulated way, which you couldn't do before (due to my oversight).

@ctz If you want to avoid exposing the `ring::Prk`s in the QUIC API, I suggest you create some wrapper types that expose an API you're more comfortable with. Eventually, we'll implement the session types and *that* will fully encapsulate and protect the secrets from any kind of misuse from the QUIC implementation. But that won't happen immediately.

Anyway, I'll need somebody to take over this PR from me, as I won't have any time after this moment to work on it, except to answer questions people may have.

src/key_schedule.rs
```diff
@@ -89,10 +90,16 @@ impl KeySchedule {
         hkdf_expand(&self.current, key_type, kind.to_bytes(), hs_hash)
     }
 
-    pub fn derive_bytes(&self, kind: SecretKind, hs_hash: &[u8]) -> Vec<u8> {
-        let payload: PayloadU8 =
-            self.derive(PayloadU8Len(self.algorithm.len()), kind, hs_hash);
-        payload.into_inner()
+    pub fn derive_logged_secret(&self, kind: SecretKind, hs_hash: &[u8],
+                                key_log: &dyn KeyLog, log_label: &str, client_random: &[u8; 32])
```
briansmith: I think that eventually, instead of taking `log_label` as a parameter, this function should compute it from `kind` inside the `if_key_log.will_log()` branch. (I shouldn't have added the `label` parameter to `will_log()`; it's main purpose is differentiating the no-logging implementation from logs that do actual logging, and that doesn't need to be done label-by-label.)

tests/api.rs
briansmith: To expand on why session types help: Really, we should be storing the traffic secrets in the session and the rest of the keys should be stored in the handshake. That way, when we finish the handshake we'll free the memory for all the non-traffic keys in the key schedule. This should shrink the overall size (heap + non-heap) of the session.

This also illustrates why a compressed `Prk` type would be useful to add to *ring*: especially after the handshake, we'll rarely use the traffic secret, so we should store them in the minimal form, instead of storing them in a fork that's optimized for frequent use. It will be a while before I do this though. I'm planning to implement bigger and more important size optimizations in *ring* first.

tests/api.rs
Ralith: See https://github.com/briansmith/rustls/pull/1

briansmith: > See [briansmith#1](https://github.com/briansmith/rustls/pull/1)

Awesome! It seems reasonable to me and I merged into into this PR.

tests/api.rs
```diff
@@ -1744,50 +1758,19 @@ fn quic_handshake() {
...
-            0xdd, 0x2c, 0x96, 0x1b, 0x3f, 0x52, 0x87, 0xa6, 0xd1, 0x46, 0x7e, 0xe0, 0xae, 0xab,
-            0x33, 0x72, 0x4d, 0xbf,
-        ],
-        server: vec![
-            0x42, 0xdc, 0x97, 0x21, 0x40, 0xe0, 0xf2, 0xe3, 0x98, 0x45, 0xb7, 0x67, 0x61, 0x34,
-            0x39, 0xdc, 0x67, 0x58, 0xca, 0x43, 0x25, 0x9b, 0x87, 0x85, 0x06, 0x82, 0x4e, 0xb1,
-            0xe4, 0x38, 0xd8, 0x55,
-        ],
-    };
-    let updated = client.update_secrets(&initial.client, &initial.server);
```
briansmith: @Ralith Please check out the code coverage report. You removed this use of `update_secrets`, which seemed to be the only test code using it, so its coverage went to zero. I think there may be other code coverage issues.

briansmith: I see in the commit where you mention this that "This is redundant to TLS key schedule tests." Does that mean that the `update_secrets` function is also redundant and can/should be removed? It seems like we either need to test it or remove it.

Ralith: Ah, forgot that part was still isolated from the TLS impl. IIRC it's in principle the same thing as `KeySchedule::derive_next`, and ideally there wouldn't be two distinct functions. However, some refactoring to QuicExt will be necessary for that to be feasible. In the mean time, I brought the test back in https://github.com/briansmith/rustls/pull/2.

tests/api.rs
ctz: Yes, I have a medium-term plan to move more stuff from the `SessionCommon` type into the handshake session types. This also means (eg) post-handshake TLS1.3 sessions don't have any memory used by TLS1.2-specific types.

ctz:
Thanks!

PR #290, opened 2019-08-13
Allow changing client cert verifier in ServerConfig

kevinis:
Hello @ctz! 

I'm working on an implementation of OpenSSL's `SSL_CTX_set_verify()` API.  It would be great if `rustls::ServerConfig` could support `set_certificate_verifier()` like `rustls::ClientConfig` does.

In this PR, I (almostly) copied the `danger` module from `client/mod.rs` and added `set_certificate_verifier()` under `DangerousServerConfig`. I was wondering if you wouldn't mind reviewing this PR. And thank you for your time!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,3 +301,5 @@ pub use crate::verify::{ServerCertVerifier, ServerCertVerified,
 #[cfg(feature = "dangerous_configuration")]
 pub use crate::client::danger::DangerousClientConfig;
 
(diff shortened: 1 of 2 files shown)


ctz:
Thanks, I've merged this with some minor changes. The `dangerous()` thing wasn't needed here because the verifier is passed into the constructor and isn't considered a very sensitive thing to change. Thanks!

PR #295, opened 2019-08-26
StreamOwned: fix get_mut signature

Keruspe:


ctz:
Thanks, that's a weird mistake to have made. And I'm surprised rustc or clippy don't warn on the pointless and immediately discarded &mut.

PR #296, opened 2019-09-02
Add WriteV implementation for any Write objects

jpopesculian:
Uses a write_vectored as a convenience function for implementing a WriteV adapter for any write object. It's a convenience function here, but I thought maybe this is food for thought in terms of changing `write_tls` to always use the `write_vectored` functionality if possible?


ctz:
Thanks!

PR #304, opened 2019-09-14
ClientHello to resolve server certs

jeromegn:
This should make it easier to extend the data available to `ResolvesServerCert::resolver` without future breaking changes.

Since you've already made a breaking change (but didn't release yet) to the same function, I figured it would be a good time to sneak in a refactor.

Any thoughts? I wasn't sure where to put this struct.


ctz:
Thanks for working on this; it looks good. Please can we have the new `ClientHello` struct live next to the declaration of `ResolvesServerCert`?

jeromegn:
Done. Somehow this fixed travis, but I think it's just being flaky?

rustls/src/server/mod.rs
```diff
@@ -95,26 +96,47 @@ pub trait ProducesTickets : Send + Sync {
...
+    sigschemes: &'a [SignatureScheme],
+    alpn: Option<&'a[&'a[u8]]>,
+}
+
+impl<'a> ClientHello<'a> {
+    /// Creates a new ClientHello
+    pub fn new(server_name: Option<webpki::DNSNameRef<'a>>, sigschemes:  &'a [SignatureScheme],
+    alpn: Option<&'a[&'a[u8]]>)->Self {
+        ClientHello {server_name, sigschemes, alpn}
+    }
```
Ralith: If the goal is to ensure forwards-compatibility, perhaps this shouldn't be public?

ctz: Good idea.

ctz:
> travis, but I think it's just being flaky

Yes, travis is regularly flaky :( Especially the macos machines.

jeromegn:
Good call re: `new` visibility.

Thanks for the quick merge!

PR #305, opened 2019-09-19
Send an alert on ALPN mismatch

Ralith:
Required by RFC7301:

In the event that the server supports no protocols that the client
advertises, then the server SHALL respond with a fatal
"no_application_protocol" alert.

Note that no error is returned if the server's protocol list is empty or if the client does not advertise a protocol list. These cases are both interpreted as the extension not being in use.

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -1772,9 +1772,10 @@ fn tls13_stateless_resumption() {
     assert_eq!(storage.takes(), 0);
 }
 
-#[test]
 #[cfg(feature = "quic")]
-fn quic_handshake() {
+mod test_quic {
+    use super::*;
+
     // Returns the sender's next secrets to use, or the receiver's error.
     fn step(send: &mut dyn Session, recv: &mut dyn Session) -> Result<Option<quic::Secrets>, TLSError> {
         let mut buf = Vec::new();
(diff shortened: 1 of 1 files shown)


Ralith:
In the rustls-mio test I see `// Like openssl we don't fail a handshake for no ALPN overlap.`. So I guess this is intentional? This is a violation of the RFC and, if preserved, will force Quinn to awkwardly work around rustls's behavior.

ctz:
> So I guess this is intentional?

Unfortunately, yes. Ultimately I think this is RFC7301 conforming because, at the point there is determined to be no ALPN overlap, the server acts as if it didn't support ALPN all along.

[BoringSSL](https://bugs.chromium.org/p/chromium/issues/detail?id=497770#c1), OpenSSL and golang have this behaviour. GnuTLS has both behaviours controlled with the `GNUTLS_ALPN_MAND` option.

Since rustls-in-quic-mode unequivocally knows that ALPN is required, perhaps the hard-fail behaviour can be done conditionally on this feature?

Ralith:
ALPN is not necessarily required in QUIC:

>  *Unless another mechanism is used for agreeing on an application protocol*, endpoints MUST use ALPN for this purpose. *When using ALPN*, endpoints MUST immediately close a connection (see Section 10.3 in [QUIC-TRANSPORT]) if an application protocol is not negotiated with a no_application_protocol TLS alert.

This is a bit ambiguous, but at least in the case when both endpoints are using it we can be pretty confident in the intended behavior. Updated to apply that iff QUIC is in use.

ctz:
Thanks!

Issue #104, opened 2017-08-30
Server does not reject client hello SNI extension containing multiple hostnames


briansmith:
See https://www.ietf.org/mail-archive/web/tls/current/msg19426.html and RFC 6066: "The ServerNameList MUST NOT contain more than one name of the same name_type."

I don't think we need to enforce the RFC 6066 rule for non-hostname types. However, we should consider at least enforcing that there is not more than one entry of type `host_name`/`HostName` entry.

Perhaps we should follow what BoringSSL does and reject any SNI extension that doesn't contain exactly one (valid) HostName entry.

PR #307, opened 2019-09-23
server rejects SNI containing multiple hostnames

thekuwayama:
Hi!

This PR is related to #104 issue.

This makes that server rejects ClientHello SNI containing multiple hostnames.
The server aborts the handshake with a `decode_error` alert, received ClientHello SNI containing multiple hostnames.


ctz:
Thanks, I've merged this with a little refactoring and testing improvements rolled in.

PR #309, opened 2019-09-24
Select RootCertStore based on SNI for Client Auth

gorup:
Hi! I posted an issue a while ago about using SNI to determine which RootStore to choose [here](https://github.com/ctz/rustls/issues/230#issue-404947222). 

This pull request includes the work to accomplish this. Please let me know what it would take to get it merged!

Its not fully polished and would likely need more testing and better documentation. Also, I expect the wonky change to add new fn's to `ClientCertVerifier` and effectively deprecate the old ones to be a sticking point, but I chose that path because it was quick and I'm in no way attached to that choice.

Thanks!


gorup:
I'd also like to discuss the ergonomics of the 'resolve' method returning 'Option<&RootCertStore>`. 

The issue I'm having is that I'm writing a resolver that uses a `RwLock<HashMap<DNSName, RootCertStore>>` internally, allowing me to change the root stores that are trusted dynamically. The borrow checker doesn't allow me to return a simple `Option<&RootCertStore>`, as the reference to the root store is borrowed from a local variable created in the resolve function, specifically the `RwLock.read()` which lets me access the map returns a local var which is the readguard. A little complicated, but here is an example:

```rust
pub struct DBBackedRootStoreResolver {
    roots: Arc<RwLock<HashMap<DNSName, RootCertStore>>>,
}

impl DBBackedRootStoreResolver {
    // .. code to have the DB update the map asynchronously..
}

impl ResolvesClientRoot for DBBackedRootStoreResolver {
    fn resolve(&self, server_name: Option<&webpki::DNSName>) -> Option<&rustls::RootCertStore> {
        block_on(async move {
            match server_name {
                None => None,
                Some(nameref) => {
                    let certs = self.roots.read().await;
                    let name = nameref.to_owned();
                    certs.get(&name) // ERROR here, returning reference of borrowed value of `certs`, which is a local var that is thrown away on return
                }
            }
        })
    }
}

```

If the `resolve` fn instead expects an `Arc<RootCertStore>`, then I can fix the above example and simply store an arc in my map and clone away. Thoughts?

gorup:
Hi @ctz I know the build is failing but the failures are pretty easy to fix and I just wanted to discuss the contents of this PR before addressing these issues. Any thoughts on potentially merging this? Thanks

ctz:
Sorry for the delay on this one; I've been on holiday the past 10 days. I will endeavour to look at this at the weekend.

ctz:
Thanks for this work. A few notes:

- There are breaking changes in the next release of rustls, so please make the new `_sni` members of `ClientCertVerifier` replace the existing ones (no need for the suffix).
- I'm not really sold on the `ResolvesClientRoot` being a useful separate trait, since it's used by one `ClientCertVerifier` implementation and crates downstream of rustls are expected to replace this wholesale.
- I'm also, I'm afraid, not sold on `AllowAnyAuthenticatedClientForSNIResolvedRoot` being in the base rustls crate at the moment. It seems quite esoteric. But, with that said, I *am* very much sold on ensuring that this can be done in crates downstream of rustls.

So perhaps we can proceed by reducing the scope of this PR to passing SNI into `ResolvesClientRoot` and merging that first?

gorup:
Hey, sounds good! I'll get the work for `ClientCertVerifier` up. I have the public stuff done just working on tests for it among some other stuff I have going on today/tomorrow.

gorup:
Updated with the smaller change set.

One interesting thing, check out the change I made to `tls12.rs`. It seems as though before my change, we would not return a `NoCertificatesPresented` error when auth was mandatory but not provided. `tls13.rs` does return that error, and I changed `tls12.rs` to match 1.3. Is there some spec difference between 1.2 and 1.3 that would be the reason for this change, or was it an omission?

I found out about it when I was writing the tests and 1.3 succeeded but 1.2 failed. Thanks!

gorup:
Check my previous message about a change to `tls12.rs`, but this recent change (besides adding a test) changes the error type thrown when we cannot resolve a cert to 'UnrecognizedName' (or UnrecognisedName). See https://www.ietf.org/rfc/rfc3546.txt for info on that.

With the error change, we actually diverge from how we treat the similar situation when we cannot decide on a cert that the Server should present. In that case, we throw an AccessDenied, error (`hs.rs:625`) e.g.

```rust
            let certkey = sess.config.cert_resolver.resolve(client_hello);
            certkey.ok_or_else(|| {
                sess.common.send_fatal_alert(AlertDescription::AccessDenied);
                TLSError::General("no server certificate chain resolved".to_string())
            })?
```

Thoughts on this as well as the `tls12.rs` change? Thanks!

gorup:
ping

gorup:
HI! I'm hoping to get this merged into master. I am using my fork in a project I'd like to launch soon and I'd prefer to just use master :)

Two points I mentioned above that should be known:

- Check out the change I made to tls12.rs. It seems as though before my change, we would not return a NoCertificatesPresented error when auth was mandatory but not provided. tls13.rs does return that error, and I changed tls12.rs to match 1.3. Is there some spec difference between 1.2 and 1.3 that would be the reason for this change, or was it an omission? I found out about it when I was writing the tests and 1.3 succeeded but 1.2 failed.

- I throw an 'UnrecognizedName' error when I cannot resolve a cert store to use (see [this RFC](https://www.ietf.org/rfc/rfc3546.txt) about that) . We treat the similar situation when we cannot decide on a cert that the Server should present in a different way. In that case, we throw an AccessDenied, error (hs.rs:625) e.g.

            let certkey = sess.config.cert_resolver.resolve(client_hello);
            certkey.ok_or_else(|| {
                sess.common.send_fatal_alert(AlertDescription::AccessDenied);
                TLSError::General("no server certificate chain resolved".to_string())
            })?
    Should I update the error in hs.rs to be UnrecognizedName too? Or just leave it to maintain the status quo?

Let me know if you have questions or comments, thanks!

ctz:
I'm in the process of merging this now, with some minor edits. To answer those questions:

> Check out the change I made to tls12.rs. It seems as though before my change, we would not return a NoCertificatesPresented error when auth was mandatory but not provided

Yeah, I think this was an omission. It only really effects what error the client sees though, so it wasn't a security bug (`verify_client_cert` being passed an empty list will reject it with `NoCertificatesPresented `

> I throw an 'UnrecognizedName' error when I cannot resolve a cert store to use

`UnrecognizedName` wouldn't be my first choice for this, since it's very SNI-specific. I'll make this `AccessDenied` like the other alerts for client auth failure.

PR #311, opened 2019-10-04
Save SNI earlier in handshake

gorup:
Pretty straightforward. I want to have the SNI saved earlier so that I can access it even if something later down the handshake fails.

My use case is a little complicated, but here it goes. I have a multi-tenant TLS proxy, meaning multiple domains can use me to proxy their TLS connections. I use SNI to determine which server cert my proxy will present, using `ResolvesServerCertUsingSNI`. Alternatively, users of this proxy can have the TLS session be end-to-end, specifically that my TLS proxy is really just a TCP proxy, and I forward every byte between the two parties including the bytes that I read as part of the ClientHello.

So, if the handshake fails, mostly due to ResolvesServerCertUsingSNI not finding a pair, I check to see if the SNI is present in the 'end-to-end' endpoint list. If it is, I just forward the client hello through to the other connection which does the TLS handshake.

Without this change, I was unable to get to the SNI value in the case the ResolvesServerCertUsingSNI comes up empty, which in my application does NOT mean we should drop the connection and completely stop doing things.

Let me know if you have any other questions or if you would like me to add tests!

Note: I manually tested this using the proxy service I am working on and it works fine.

### DIFF
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -615,6 +615,7 @@ impl State for ExpectClientHello {
             },
             None => None,
         };
+        save_sni(sess, sni.clone());
 
         // We communicate to the upper layer what kind of key they should choose
         // via the sigschemes value.  Clients tend to treat this extension
@@ -688,11 +689,10 @@ impl State for ExpectClientHello {
 
         if sess.common.is_tls13() {
             return self.into_complete_tls13_client_hello_handling()
-                .handle_client_hello(sess, sni, certkey, &m);
+                .handle_client_hello(sess, certkey, &m);
         }
 
         // -- TLS1.2 only from hereon in --
-        save_sni(sess, sni.clone());
         self.handshake.transcript.add_message(&m);
 
         if client_hello.ems_support_offered() {
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -615,7 +615,6 @@ impl State for ExpectClientHello {
             },
             None => None,
         };
-        save_sni(sess, sni.clone());
 
         // We communicate to the upper layer what kind of key they should choose
         // via the sigschemes value.  Clients tend to treat this extension
@@ -689,10 +688,11 @@ impl State for ExpectClientHello {
 
         if sess.common.is_tls13() {
             return self.into_complete_tls13_client_hello_handling()
-                .handle_client_hello(sess, certkey, &m);
+                .handle_client_hello(sess, sni, certkey, &m);
         }
 
         // -- TLS1.2 only from hereon in --
+        save_sni(sess, sni.clone());
         self.handshake.transcript.add_message(&m);
 
         if client_hello.ems_support_offered() {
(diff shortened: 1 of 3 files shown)


gorup:
I ran `cargo test` but the additional tests failed. They failed because (apparently) the `save_sni` is called > 1 times, as to where before my change it only got executed 1x. This then throws an exception in [`set_sni`](https://github.com/ctz/rustls/blob/69b96b14b10a18baa4569b1d80956187b922113e/rustls/src/server/mod.rs#L474-L478) (which `save_sni` calls).

I think it would be best to, in `save_sni`, do a check that looks up the existing sni if set, and if set, checks that they are equal. If they are equal, great, don't call `set_sni` and avoid the assertion, if they are NOT equal, then throw some PeerMisbehavedError saying we received (somehow?) different SNIs.

Alternatively we can just remove the assertion. I'm unsure what the assertion prevents, as we could have actually changed the SNI in the different messages we sent and we would just stick with the last one we received. With my first proposed change, we would accept one and only one throughout instead of last-in-wins.

Thoughts? Thanks

ctz:
Can you write a test against master that expresses what you want to happen here?

gorup:
This branch is now updated vs the current master and only contains the failing test. If you check it out and revert the revert of my change (`git revert 2d7bee1`), then the test will succeed, although fail other tests down the line per my note above.

I wanted to save the SNI 'early' (more specifically as soon as possible) so that even if something in the TLS handshake fails, we can still see the SNI if it existed. Currently, if the TLS handshake fails during `processs_new_packets` due to not resolving to a CertifiedKey (in `ResolvesServerCert`), upon calling `get_sni_hostname` on the session, we get a `None`. There was an SNI, it was detected and used in the call to the `ResolvesServerCert` impl being used, but its not available to see later. This may still be worth doing so that users can see the SNI that did not have a resolved-to key, but I can work around it.



I'd be happy to get this done, and do the work to make the other tests succeed per my above comment, but again, I have a workaround if you'd like to simply ignore this for now :)

---
Additional Background if you're bored and like L7 load balancers:

I am writing a proxy service/load balancer that can operate in two modes. Just think of a standard load balancer that operates in two modes. One is Hosted, where I have some certs stored and a custom `ResolvesServerCert` implementation that chooses the key/cert to use. That works all fine and dandy. Second it can work in Passthrough mode, where the TLS connection is NOT terminated at my load balancer service, and I forward all of the packets of the TCP connection to the servers of this endpoint. I discovered that the SNI is not saved when the resolver fails to resolve when I was using my resolver that was only aware of 'Hosted' endpoints. When a client connects with the SNI of a Passthrough endpoint, my resolver doesn't resolve. This errors out, and get_sni_hostname returns None(), although I know it was actually a set and part of the ClientHello.

I planned on recovering from this resolver failure by looking at the SNI, seeing if its in my list of Passthrough endpoints, and if so, continue on. That was my first idea, and if I went with that, I would be blocked by this. But I didn't like this and didn't want to be blocked, so I decided to simply use a 'dummy' server config that always resolves so we can create a dummy session and pop open the SNI and see it. From there, I use that SNI to see if its Hosted or Passthrough using my own resolver thing. For Hosted, I just use a ServerConfig I have for Hosted endpoints, create a new session and continue on. If my resolver thing says the endpoint is Passthrough, I just don't do anything with the dummy session, I keep the raw bytes, and just proxy TCP bytes. So instead of the failure-recover-act model, I now have tempSession-detectType-act model which is cleaner IMO.

ctz:
Thanks! Sorry about the delay with this getting this merged.

PR #312, opened 2019-10-08
Add dangerous server cert verifier which uses pinned DNS name for cert validation

gorup:
Hello again!

As commit message states, this allows users to have a client config which will trust the server based on if its certificates are valid for the pinned DNS name.

This is useful when you want to use the SNI to indicate to the target to the server, but the server does not have a certificate valid for the target. Specifically useful for TLS proxies. Let me know if tests would be needed seeing as this is a `dangerous`-only provided struct.

### DIFF
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -298,4 +298,5 @@ pub use crate::verify::{ServerCertVerifier, ServerCertVerified,
     ClientCertVerifier, ClientCertVerified};
 #[cfg(feature = "dangerous_configuration")]
 pub use crate::client::danger::DangerousClientConfig;
-
+#[cfg(feature = "dangerous_configuration")]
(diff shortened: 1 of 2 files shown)


ctz:
Since this can be done entirely on top of the existing API I'm minded not to accept this. It doesn't seem generally useful, and is a quite application-specific behaviour?

ctz:
If you do move this to a higher-level crate, note that `WebPKIVerifier` is public so you can call its `verify_server_cert` behaviour in your own `verify_server_cert` to get the behaviour you have here.

gorup:
unless I'm mistaken, WebPKIVerifier is not public? Its not re-exported when Dangerous is on, and not present otherwise. 

Should I put up a pull-request w/ just adding `WebPKIVerifier` to be exported when Dangerous is on?

ctz:
Yes please do!

PR #314, opened 2019-10-16
Expose WebPKIVerifier under dangerous feature

gorup:
As mentioned in [a closed issue](https://github.com/ctz/rustls/pull/312#issuecomment-542419246), `WebPKIVerifier` should be exposed so users of the `dangerous` flag can potentially wrap this verifier.


ctz:
Thanks!

PR #317, opened 2019-10-25
Upgrade base64 dependency to 0.11

alex:


alex:
Failures don't appear to be caused by this PR.

PR #319, opened 2019-11-06
Expose ClientSession EarlyData is_enabled function

daareiza:
In order to implement 0-RTT on rustls wrappers (like tokio-rustls) it is needed to know if current client session early data is accepted by the server (it's value is set on [client/tls13.rs#L268](https://github.com/ctz/rustls/blob/master/rustls/src/client/tls13.rs#L268]).

This PR just expose the EarlyData `is_enabled` function on the `ClientSession` implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target/
 *.gcda
 *.gcno
 *.info
+.idea
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -634,6 +634,11 @@ impl ClientSession {
     pub fn is_early_data_accepted(&self) -> bool {
         self.imp.early_data.is_accepted()
     }
+
+    /// Returns True if the client is enabled to send early data.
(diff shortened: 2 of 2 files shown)


Ralith:
Couldn't you do `.early_data().is_some()`?

daareiza:
> Couldn't you do `.early_data().is_some()`?

yeah, I didn't think about that possibility, although that will return a new writer even if the wrapper only wants to know if it is enabled/supported or not.

Ralith:
That's just a reference to `self`, and with the `Option` the generated code will just be checking whether the pointer is null, so I think that's fine.

daareiza:
> That's just a reference to `self`, and with the `Option` the generated code will just be checking whether the pointer is null, so I think that's fine.

:O , thanks man!!!, so this PR is somehow useless now that you point that out.

daareiza:
Closed in favor of `.early_data().is_some()` per @Ralith suggestion.

PR #320, opened 2019-11-18
Update QUIC key update label for draft 24

Ralith:


djc:
Alternatively, we can move this code into Quinn so that rustls no longer has to know QUIC labels:

https://github.com/djc/quinn/pull/499

Ralith:
That's a good solution for the time being, but I'd prefer not to have the non-total `match` we need there to look up the HKDF. Would it make sense for rustls to make `SupportedCipherSuite.hkdf_algorithm` public?

I think @briansmith ultimately wants us to move to getting keys directly from rustls rather than having any knowledge of secrets, let alone HKDFs, in the application code, which I'd be okay with though I haven't been super motivated to do the work for it.

ctz:
Thanks!

PR #323, opened 2019-12-04
remove unnecessary dependency from rustls-mio

bluejekyll:
Noticed this pulling in the older bytes and some other dependencies. Looks like it's no longer used.


ctz:
Thanks!

PR #324, opened 2019-12-08
Remove unused TLSError::InvalidDNS variant

link2xt:


ctz:
Good spot, thank you!

PR #325, opened 2019-12-12
Center Ferris and overview text

hlb8122:
Ferris looking bit cramped there on the left.

[Futures](https://github.com/rust-lang/futures-rs) `README.md` opts to center the logo and overview text which I think looks nice.

Disadvantage is that using HTML tags rather than standard Markdown which may not be supported by all Markdown renderers.

Screenshot from 2019-12-12 13-04-15


ctz:
Thanks!

PR #58, opened 2017-03-19
WIP: DTLS Support

manuels:
It took me quite some time to find the best way to make the code wort with `TLSMessage` and `DTLSMessage`, but I think I finally found a good solution.
This pull request is here to document the work in progress. Feel free to comment on the code. I will rebase it from time to time until DTLS implementation is finished.

It will close the issue #40.

PR #326, opened 2019-12-28
[WIP/DRAFT] Continuation on DTLS support.

TimonPost:
This PR is a continuation of #58. That PR is heavily outdated and works with rustls from 3 years ago.
There is a lot of noise in the changes of this PR and rebasing is going to be a nightmare and probably not possible, since a lot of types are renamed and the architecture is changed since that PR.

The goals for this PR:
- [X] Make sure all changes are exported from #58 
- [ ] Migrate changes to newest rustls
- [ ] Implement DTLS 1.3

This version is still in a draft, however, until the official release we can experiment with different implementations and drafts. 

I hope to finish this PR, however, I never dove deep into TLS mechanics, so it may have some speed delay. But that should not be the biggest concern here since the draft is also already open for some years now :). And if it will not be finished, then it is at least a nice baseline for other implementations.


TimonPost:
Sorry, I have to drop this PR due to new year resolutions haha, if anyone want to continue, just copy-paste the code. All changes of #58 are merged.

PR #327, opened 2019-12-28
Code Formatting

TimonPost:
- ran cargo fmt
- Merged imports with cargo fmt merge_imports enabled.
Why? To me, it looks way cleaner to have imports merged instead of repeating import paths. 

No breaking changes are made. Only formatting of existing code.


rustls-mio/examples/simple_0rtt_client.rs
```diff
@@ -1,12 +1,14 @@
 use std::sync::Arc;
 
-use std::io::{stdout, Read, Write};
-use std::net::TcpStream;
+use std::{
+    io::{stdout, Read, Write},
+    net::TcpStream,
+};
```
burdges: This is an unfortunate change.

burdges:
Security projects should never just merge pull requests that do formatting changes.  Raise an issue and see if the core maintainers want to fmt, but..

rustfmt is actually really horrible and harms expressiveness.  It's okay when folks code to rustfmt, like by avoiding method chaining with well named temporaries, but often you want more nuance.

TimonPost:
> Security projects should never just merge pull requests that do formatting changes

Why is that? That doesn't make much sense to me...

> rustfmt is actually really horrible and harms expressiveness. It's okay when folks code to rustfmt, like by avoiding method chaining with well-named temporaries, but often you want more nuance.

Agree and disagree, like, having people taking manual care of formatting is going to be less inconsistent then if you let a tool do it for you. Rustfmt is the rust format a lot of other rust projects are using as well. I do not see any reason why you would not want a codebase to adhere to the accepted format.

briansmith:
> Security projects should never just merge pull requests that do formatting changes. If the core maintainers want to fmt then they should fmt themselves

The best way to review a change like this is to run the `cargo fmt ...` command that the submitter claimed to use to produce the PR, and then verify that the results are identical to what the submitter's PR produced.

> rustfmt is actually really horrible and harms expressiveness.

There are definitely downsides to rustfmt but I think the upsides are bigger so I support doing it and checking it in CI/CD.

TimonPost:
We can also spit this librarie up int two parts:

1. Run cargo fmt
2. Run cargo fmt with `merge_imports` enabled

Then we can reduce the number of changes, and focus on only the formatting and then the namespace mergings.

ctz:
I've done some rustfmt cleanups in the past here, but I'm waiting for 2.0 before using it again[1]. So I'm going to close this pending that.

[1]: Support for predictable formatting of call chains was removed for 1.0 stabilisation reasons, I think. There are numerous upstream bugs about this.

PR #292, opened 2019-08-16
Add Ed25519 support

potatosalad:
This is a work-in-progress/draft pull request for adding Ed25519 support.

See #52 for more information.

_Notes_

* Some `bogo` related tests are still failing.
* The OpenSSL related commands (like `test-ca/build-a-pki.sh`) depend on [OpenSSL version 1.1.1](https://www.openssl.org/news/openssl-1.1.1-notes.html).  On macOS, for example, prefixing `cargo test` commands with `OPENSSL_DIR="$(brew --prefix openssl@1.1)"` will allow the tests to pass.

Issue #52, opened 2017-02-20
X448 / Ed448 / Ed25519 support


Darkspirit:
X448 is a perfect replacement for secp521r1 to gain "level 8" security here: https://www.keylength.com/en/3/

[RFC8080: Edwards-Curve Digital Security Algorithm (EdDSA) for DNSSEC](https://tools.ietf.org/html/rfc8080)

[btls](https://gitlab.com/ilari_l/btls) has both, but no TLS 1.3.

briansmith:
*ring* supports Ed25519 and webpki supported Ed25519 certificates.

Note that https://tools.ietf.org/html/rfc8422 says:

>   All EdDSA computations MUST be performed according to [RFC8032] or
   its successors.  Data to be signed/verified is run through the EdDSA
   algorithm with no hashing (EdDSA will internally run the data through
   the "prehash" function PH).  The context parameter for Ed448 MUST be
   set to the empty string.

I believe that Rustls already has to do things in a way that would automatically work for Ed25519 since *ring* exposes an API that reflects what RFC8422 requires implementations to do. So, perhaps we just need to wire up the key loading and add some tests?

briansmith:
Also, Go implements this as of Go 1.13 (See https://go-review.googlesource.com/c/go/+/177698/4//COMMIT_MSG), so we should be able to do interop testing against Go even if BoringSSL and BoGo haven't been updated for Ed25519 yet.

ctz:
I'm going to close this since ed25519 is now supported. x448 and ed448 are for another day.

PR #340, opened 2020-02-15
Add Ed25519 support

potatosalad:
See #292 for the original draft pull request.

This is a pull request for adding Ed25519 support.

See #52 for more information.

_Notes_

* Some `bogo` related tests are still failing.
* The OpenSSL related commands (like `test-ca/build-a-pki.sh`) depend on [OpenSSL version 1.1.1](https://www.openssl.org/news/openssl-1.1.1-notes.html).  On macOS, for example, prefixing `cargo test` commands with `OPENSSL_DIR="$(brew --prefix openssl@1.1)"` will allow the tests to pass.


ctz:
Thanks for working on this!

I had a quick look at the remaining bogo failures: at least _some_ of these are happening because the test wants to see ed25519 working with TLS1.2. That currently won't work because all the implemented ciphersuites (say, `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256`) will be removed from consideration by the server, leaving zero TLS1.2 ciphersuites to select from.

I suggest we keep this as a TLS1.3-only feature, and perhaps just skip these tests?

kim:
@ctz Thanks for the pointer. 

I only found one test case which is expected to succeed. Since I can't seem to get the bogo suite to run locally, have to wait until new build logs are availabe...

potatosalad:
@kim I pushed up a few other bogo related exemptions.  Locally, everything now passes for me except `Ed25519DefaultDisable-NoAccept` and `Ed25519DefaultDisable-NoAdvertise`.

I'm not sure whether the config settings are appropriate or not or if they are a sign of an actual implementation bug.

kim:
> I suggest we keep this as a TLS1.3-only feature

Do you mean that it should be an error to use an ed25519 key if the protocols include anything below 1.3?

ctz:
This is merged now. Thanks a lot for the hard work on this!

PR #343, opened 2020-02-19
Enable documentation on docs.rs for items behind feature flags

aloucks:
This crate has several private traits in the public API. Navigation and discoverability is improved by enabling `docs.rs` documentation for these traits.


ctz:
Thanks! I wasn't even aware this was possible!

PR #346, opened 2020-03-06
docs: prefer the IETF wording for cipher suite

efx:
I also took the liberty of linking to the RFC page for curious readers.


ctz:
Thanks!

Issue #331, opened 2020-01-05
Using rustls for machine-to-machine w/ private CA and no DNS


jadamcrain:
I couldn't tell if this was possible from scanning the docs. All of the client examples appear to take a `DNSNameRef`.  Lots of m2m environments (e.g. industrial) don't use DNS at all in favor of static IP addresses, and therefore don't use DNS names as the CN or provide the `SubjectAlternativeName` extension at all.

Is it possible to use Rustls w/o the client performing hostname validation via the server CN and/or extensions?

I also have a requirement to parse a custom X.509 extension that contains authz info. Is this possible as well?

briansmith:
> therefore don't use DNS names as the CN or provide the `SubjectAlternativeName` extension at all.

What do these things use as an ID?

jadamcrain:
non-domain name `CN` or a list of allowed IP addresses in the `SAN` extension.

This example X.509 cert is from the Modbus TLS security standard. 

![client_cert](https://user-images.githubusercontent.com/305813/71855442-e6e13280-3095-11ea-9f50-209d88ab43fc.png)

jadamcrain:
Just checking in on this. Did #329 address this fully? Is there anything I can do in terms of a contribution to move this forward?

sehz:
I also like to move this forward

jadamcrain:
It appears like this might be possible since `0.18.0` release with the `dangerous()` configuration to set client/server verification not-based on Web-PKI.

1) It seems non-ideal (and extra dangerous) to make everyone bring their own x.509 parser for these applications. 

2) I've looked at `x509-signature` crate. I would like to see a more complete x.509 representation to work with. It appears like it doesn't fully parse a lot of things.

Just updating this ticket. I will do some playing around with `x509-signature` and our own x509 library to see which path we end up taking for our application.  We might end up contributing a more complete m2m/non-web verifier to the ecosystem.

IniterWorker:
Hi folks, 

The same need here,

Is there a way to workaround temporarily. I'd to use this crate for a project. But, I must comply with m2m without a DNS server.

I am struggling with "how to not provide a DNS Name server" when you are using the "dangerous" feature.

Any hints are welcome,

B. r

Ralith:
As a hacky work-around, you can make up whatever hostname you want, so long as what the client supplies to `connect` matches the certificate.

jadamcrain:
[...]

PR #350, opened 2020-03-13
Support certificates that webpki cannot handle

### LINKED ISSUE #329: Support for fully user-defined certificate verification
libp2p needs to be able to verify certificates that would normally be considered invalid (they contain a critical extension that webpki does not, and should not, know about).  It also needs to return data from this process to the application.  The easiest way to do this is for rustls to not even try to parse the certificate itself.  Instead, it should just treat it as bytes.

DemiMarie-parity:
This is needed for libp2p and other non-web uses.

Fixes #329.


burdges:
Any parallel with #331 here?

DemiMarie-parity:
> Any parallel with #331 here?

This PR would unblock #331, at least to the extent that the relevant certificates have at least one extension.

PR #353, opened 2020-04-02
Allow fully replacing the builtin certificate verifier

DemiMarie-parity:
The first commit allows the builtin signature verification to be replaced by custom code. The second commit allows it to be omited entirely.


DemiMarie-parity:
@ctz would you mind helping me debug the CI? There seems to be a problem with the bogo parts of the CI.

PR #354, opened 2020-04-04
Improve tls_read() performance

nviennot:
Changing a for loop in deframer with memmove() reduces CPU utilization by **30%** when downloading a 1G file over https.

### DIFF
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -146,13 +146,9 @@ impl MessageDeframer {
              * +----------+----------+----------+
              * 0          ^ self.used
              */
-            let used_after = self.used - taken;
 
-            for i in 0..used_after {
-                self.buf[i] = self.buf[i + taken];
-            }
-
-            self.used = used_after;
+            self.buf.copy_within(taken..self.used, 0);
+            self.used = self.used - taken;
         } else if taken == self.used {
             self.used = 0;
         }


Darkspirit:
This increases minimum Rust version from [1.36.0](https://github.com/ctz/rustls/blob/bdaf35b2cc6c6679d3c40f70bdbb9cb0c83ba421/.azure-pipelines.yml#L2-L3) to [1.37.0](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within).

nviennot:
Is there a way to make it optional with a feature, like a polyfill?

For example if the user has rust 1.37.0, use the memmove version, otherwise fall back on the for loop version

nviennot:
With https://github.com/ctz/rustls/pull/356 https://github.com/ctz/rustls/pull/355 improving bench with MTU setting, I'm getting the following numbers when running with an MTU of 10k.

Before:
```
» cargo run --release --example bench bulk TLS13_AES_128_GCM_SHA256 $((300*1024*1024)) 10000
bulk    TLSv1_3 TLS13_AES_128_GCM_SHA256        send    1411.03 MB/s
bulk    TLSv1_3 TLS13_AES_128_GCM_SHA256        recv    808.14  MB/s
```

After:
```
» cargo run --release --example bench bulk TLS13_AES_128_GCM_SHA256 $((300*1024*1024)) 10000
bulk    TLSv1_3 TLS13_AES_128_GCM_SHA256        send    1431.42 MB/s
bulk    TLSv1_3 TLS13_AES_128_GCM_SHA256        recv    1566.24 MB/s
```

ctz:
Great job, thanks!

PR #355, opened 2020-04-05
bench: allow MTU to be specified for the server bulk bench

nviennot:
This is useful for testing the receiver when the MTU is different from 16k.


ctz:
Thanks!

PR #361, opened 2020-04-20
stream: implement write_vectored

Keruspe:
I'm not sure if this is something we want to do, but this allows queuing more data internally without flushing to the socket inbetween buffers


ctz:
Thanks! I had missed the vectored IO was part of std now. We can remove stuff like WriteVAdapter now.

PR #362, opened 2020-04-26
Higher-level, misuse-resistant QUIC API

### LINKED ISSUE #287: QUIC API shouldn't expose HKDF PRKs
Instead of exposing the raw PRKs, the QUIC API should expose an API that limits which HKDF info is allowed to be used for each key. In particular, it the API should prevent non-QUIC labels from being used. I suggest in particular that for each type of key that needs to be derived for QUIC use, there should be a function which hard-codes the parameters to the HKDF expansion.

This would help make it clearer that non-QUIC keys won't be leaked from Rustls when no key log is being used.

Ralith:
Fixes #287. Strengthens encapsulation of secrets within rustls, and therefore reduces danger of security vulnerabilities due to incorrect use.

~My original plan was to expose the *ring* keys and leave the rest to downstream, in the interest of minimizing the amount of code living in rustls to reduce audit load and exposure to changes in QUIC. However, it became apparent that the encryption/decryption logic was both short and necessary for unit tests, and seems very convenient for downstream besides.~

I've kept the `hkdf::Salt` used to generate initial keys out of rustls, which is the main forwards-compatibility hazard and which hopefully doesn't introduce significant danger of incorrect use, since incorrect value will render a QUIC implementation unable to interoperate with any others.

I also considered going further by moving key lifecycle management into rustls. I decided against this on the basis that a QUIC implementation often needs concurrent access to a variety of different keys to handle reordering and to validate key updates, and also needs to control the eviction of keys from memory, which would require a noticably more complex API.

~Leaving as a draft until I've validated the ergonomics and correctness of these changes by implementing them in quinn. Feedback very much desired regardless, however.~

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -1,15 +1,16 @@
 /// This module contains optional APIs for implementing QUIC TLS.
 use crate::client::{ClientConfig, ClientSession, ClientSessionImpl};
-use crate::msgs::enums::{ContentType, ProtocolVersion, AlertDescription};
+use crate::error::TLSError;
+use crate::key_schedule::hkdf_expand;
+use crate::msgs::enums::{AlertDescription, ContentType, ProtocolVersion};
 use crate::msgs::handshake::{ClientExtension, ServerExtension};
 use crate::msgs::message::{Message, MessagePayload};
 use crate::server::{ServerConfig, ServerSession, ServerSessionImpl};
-use crate::error::TLSError;
-use crate::key_schedule::hkdf_expand;
-use crate::session::{SessionCommon, Protocol};
+use crate::session::{Protocol, SessionCommon};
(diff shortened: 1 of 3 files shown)


Ralith:
Removed encryption logic in favor of exposing keys and IV directly following discussion with @djc.

djc:
I think this is a good change, that substantially decreases coupling between rustls and Quinn. As far as I recall, this implementation has been pretty stable over the lifetime of Quinn so should be light on changes. To be explicit, both @Ralith and I will be available to maintain this part of rustls.

Ralith:
Looks like coverage CI is broken:
> error: unknown debugging option: `no-landing-pads`

Ralith:
A quinn branch using this change has been drafted at https://github.com/djc/quinn/pull/740. This is now ready to proceed.

PR #367, opened 2020-05-21
QUIC: Don't set legacy_session_id, fix key update

Ralith:
[QUIC TLS §8.4](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#name-prohibit-tls-middlebox-comp):
> A server SHOULD treat the receipt of a TLS ClientHello with a non-empty legacy_session_id field as a connection error of type PROTOCOL_VIOLATION.

So, let's not send those.


Ralith:
Added a fix for a bug introduced in #362.

PR #370, opened 2020-06-01
update ring

eugene-babichenko:
The latest version adds support for more platforms.


Ralith:
There shouldn't be much benefit to rustls specifying a newer patch version--you can just `cargo update` and get it for free in any sensible project without any effort.

ctz:
Yes the only reason to make this sort of update would be if we depended on something released in .14, which we don't.

PR #386, opened 2020-07-12
fixes #376 -- document platform support in the library docs

### LINKED ISSUE #376: Portability
It should be noted in some obvious place that this library is not portable (and therefore should not be used if the user aims for their application to be portable), due to its dependency on `ring`, which only provides support for `x86(_64)`, `aarch64` and some variants of 32-bit ARM due to requiring arch specific assembly implementations from BoringSSL without providing portable fallbacks for them.

alex:


ctz:
Thanks!

PR #395, opened 2020-08-07
Make examples better

realaltffour:
- make TLS client example a little more consistent with the TLS server example.
- make TLS client have a cleaner implementation code block


ctz:
Merged, thanks!

PR #399, opened 2020-09-08
add export_keying_material support in QUIC mode

kwantam:
This pull request makes it possible to call `export_keying_material` when rustls is used for QUIC.

Prior to applying the proposed patch, the `export_keying_material` method in the State impl for ExpectQUICTraffic is the default one, which returns `TLSError::HandshakeNotComplete`. After applying, export_keying_material correctly calls the KeyScheduleTraffic implementation of `export_keying_material`.

This PR also wires up the existing `do_exporter_test` in the `test_quic` module.

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -1046,4 +1046,11 @@ impl hs::State for ExpectQUICTraffic {
         self.0.handle_new_ticket_tls13(sess, nst)?;
         Ok(self)
     }
+
+    fn export_keying_material(&self,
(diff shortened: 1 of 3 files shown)


kwantam:
(probably you already know this, but the tests that fail are failing on master without this PR, too)

ctz:
> (probably you already know this, but the tests that fail are failing on master without this PR, too)

Yeah, looks like beta/nightly is miscompiling things; these tests are really basic parsing tests that pass with a stable compiler.

ctz:
Thanks!

PR #406, opened 2020-10-07
Improve OwnedTrustAnchor to/from TrustAnchor interop

chrisvittal:
Explicitly make OwnedTrustAnchor part of the public API. Furthermore,
implement From<webpki::TrustAnchor> for OwnedTrustAnchor and
Into<webpki::TrustAnchor> for &OwnedTrustAnchor.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -19,7 +19,8 @@ pub struct OwnedTrustAnchor {
 }
 
 impl OwnedTrustAnchor {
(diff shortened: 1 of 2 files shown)


chrisvittal:
Note: The nature of `pub` in rust means that since the `Vec` is public in `RootCerts`, `to_trust_anchor` is already a 'public' interface.

ctz:
Thanks!

PR #416, opened 2020-10-24
Show ALPN upgrade protocol as string in debug logs

dermesser:
This is an almost trivial request, but it would have helped me as a TLS noob to figure out what the ALPN upgrade does here.

The chosen conversion avoids allocations. Downside: If the ALPN string is ever not valid UTF-8, it will be invisible in the logs. However, [RFC 7301](https://tools.ietf.org/html/rfc7301#section-6) gives some expectation that the ALPN protocol strings are and will consist of characters in the ASCII subset.

Before:
```
[2020-10-24T16:00:32Z DEBUG rustls::client::hs] ALPN protocol is Some([104, 50])
```

After: 
```
[2020-10-24T16:00:32Z DEBUG rustls::client::hs] ALPN protocol is Some("h2")
```

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -360,7 +360,7 @@ pub fn process_alpn_protocol(sess: &mut ClientSessionImpl,
         !sess.config.alpn_protocols.contains(sess.alpn_protocol.as_ref().unwrap()) {
         return Err(illegal_param(sess, "server sent non-offered ALPN protocol"));
     }
-    debug!("ALPN protocol is {:?}", sess.alpn_protocol);
+    debug!("ALPN protocol is {:?}", sess.alpn_protocol.as_ref().map(|us| std::str::from_utf8(us).unwrap_or("")));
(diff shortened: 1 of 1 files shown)


Ralith:
We shouldn't report that the ALPN is "" when it's not.

djc:
Maybe a utf8_lossy() conversion would be fitting here?

dermesser:
Good point. This results in an allocation, but -- if I understand the `debug!()` respectively `log!()` macros correctly -- only at a sufficient debugging level. For the case that the ALPN string is not a printable string, it is also printed in the way that rustls did before this PR.

djc:
LGTM. This is kind of fixing a regression from when I changed ALPN protocols to be stored as bytes instead of str.

ctz:
Thanks, I've merged this with a minor change -- reusing the fmt::Debug handler we have for printing certificate contents. This prints u8 slices as ascii with C-style escapes if needed.

PR #417, opened 2020-11-02
Implement ResolvesServerCert for Arc<T>

NeoLegends:
Sounds like a useful implementation, in particular since the trait only requires `&self`. If this was added we could remove some additional boilerplate from our code.

### DIFF
--- a/rustls/src/server/mod.rs
+++ b/rustls/src/server/mod.rs
@@ -102,6 +102,24 @@ pub trait ResolvesServerCert : Send + Sync {
     fn resolve(&self, client_hello: ClientHello) -> Option<sign::CertifiedKey>;
 }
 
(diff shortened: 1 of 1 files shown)


briansmith:
Hi, I don't understand why this is useful for Rustls or how it can significantly reduce boilerplate in your code. Also, there are a lot of things besides `ResolvesServerCert` that follow the similar pattern within Rustls, so it isn't clear why we'd do this for just this one trait but not the others.

Could you show an example of how it cuts down a significant amount of boilerplate in your code?

briansmith:
@NeoLegends Are you still interested in this? Did you see my question above? Should we close this?

NeoLegends:
Hey! I apologize for the radio silence. This didn't turn out to be as useful as I thought it would be. This can be closed.

PR #428, opened 2020-12-07
don't allocate a new vec in every `ChunkVecBuffer::write_to`

hawkw:
This commit changes the implementation of `ChunkVecBuffer::write_to` so
that a vector is not allocated in every call. This should hopefully
reduce the overhead of writing TLS data to a stream.

Instead of using a `Vec`, a fixed-size array of 64 `IoSlice`s is used.
If the number of queued chunks exceeds 64, we may write fewer bytes in
one `write_to` call, but not allocating and dropping a vec in every call
is probably worth it. My intuition is that 64 `IoSlice`s is plenty and
it shouldn't be exceeded often, but I'll admit that it was chosen fairly
arbitrarily.

Another potential approach could be to use a single vec that lives in
the `ChunkVecBuffer`, and is `clear`ed when consuming written data,
retaining the allocated capacity. That way, the existing allocation is
reused, and we only allocate when the `VecDeque` of queued chunks has
grown longer than its the previous maximum length. However, this is not
presently possible, because an `IoSlice` is borrowed, and we can't add a
`Vec<IoSlice<'a>>` to the `ChunkVecBuffer`, since the lifetime `'a` here
would be the lifetime of the `&mut self` receiver of `write_to` --- so
the slices would not live long enough to be assigned to such a vec.

Therefore, I think using a stack array is the right approach for now.

### DIFF
--- a/rustls/src/vecbuf.rs
+++ b/rustls/src/vecbuf.rs
@@ -115,7 +115,78 @@ impl ChunkVecBuffer {
             return Ok(0);
         }
 
-        let mut bufs = [io::IoSlice::new(&[]); 64];
+        // XXX: `rustls`'s MSRV is 1.40.0, where `IoSlice` doesn't implement
+        // `Copy`. When the MSRV is Rust 1.44.0, this could be the much more
+        // concise
+        // ```
+        // let mut bufs = [io::IoSlice::new(&[]); 64];
+        // ```
+        let mut bufs = [
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
+            io::IoSlice::new(&[]),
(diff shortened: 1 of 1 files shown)


rustls/src/vecbuf.rs
```diff
@@ -117,13 +115,83 @@ impl ChunkVecBuffer {
...
 
-        let used = wr.write_vectored(
-            &self
-                .chunks
-                .iter()
-                .map(|ch| io::IoSlice::new(ch))
-                .collect::<Vec<io::IoSlice>>(),
-        )?;
+        // XXX: `rustls`'s MSRV is 1.40.0, where `IoSlice` doesn't implement
+        // `Copy`. When the MSRV is Rust 1.44.0, this could be the much more
```
briansmith, line 119: We should update the MSRV required for the current `main` branch. There have been some really important improvements, especially related to arrays and min_const_generics and friends. I think we should increase the MSRV at least to 1.44.0 and beyond as necessary, to allow us to do things like this with better code.

djc:
Might make sense to align with tokio on their 6-month old MSRV scheme (which would get us to 1.45).

rustls/src/vecbuf.rs
briansmith: I proposed increasing the MSRV to 1.47.0 in PR #463.

briansmith:
@hawkw could you please update this PR to use the simpler version of the code, now that the MSRV of Rustls has been increased?

hawkw:
> @hawkw could you please update this PR to use the simpler version of the code, now that the MSRV of Rustls has been increased?

Will do, thanks!

hawkw:
Hmm, I'm trying to fix the failing CI rustfmt check, and I noticed that running `cargo fmt` on this PR results in changing a large number of files that weren't modified in this branch. Rustfmt prints the message
```
Warning: Unknown configuration option `chain_width`
```

Is there a specific toolchain or rustfmt version I should be using?

djc:
You can use the admin/fmt script.

ctz:
Thanks! Merged as 62d15df9

PR #439, opened 2021-01-21
Update docs for process_new_packets.

jsha:
As a user, it was not immediately clear to me what methods may be called
after process_new_packets returns an error. Based on reading the code,
I've updated the doc comment with my understanding. My understanding
could be wrong! Happy to update if so.

### DIFF
--- a/rustls/src/session.rs
+++ b/rustls/src/session.rs
@@ -53,9 +53,16 @@ pub trait Session: quic::QuicExt + Read + Write + Send + Sync {
     fn write_tls(&mut self, wr: &mut dyn Write) -> Result<usize, io::Error>;
 
     /// Processes any new packets read by a previous call to `read_tls`.
+    ///
     /// Errors from this function relate to TLS protocol errors, and
     /// are fatal to the session.  Future calls after an error will do
-    /// no new work and will return the same error.
+    /// no new work and will return the same error. After an error is
+    /// received from process_new_packets, you should not call read_tls
+    /// any more (it will fill up buffers to no purpose). However, you
+    /// may call the other methods on the session, including write,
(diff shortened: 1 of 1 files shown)


rustls/src/session.rs
```diff
@@ -53,9 +53,16 @@ pub trait Session: quic::QuicExt + Read + Write + Send + Sync {
...
     /// Errors from this function relate to TLS protocol errors, and
     /// are fatal to the session.  Future calls after an error will do
-    /// no new work and will return the same error.
+    /// no new work and will return the same error. After an error is
+    /// received from process_new_packets, you should not call read_tls
+    /// any more (it will fill up buffers to no purpose). However, you
+    /// may call the other methods on the session, including write,
+    /// send_close_notify, and write_tls. Most likely you will want to
+    /// call write_tls to send any alerts queued by the error and then
+    /// close the connection.
```
briansmith, line 65: "you may call" doesn't mean anything. Presumably, "you may call" any function that you can call and Rustls should ensure that that is safe to do, perhaps just by returning an error that the session is invalid.

I don't think rustls should claim that calls to `write`, `send_close_notify`, etc. will do what they're expected to do after `process_new_packets` returns an error.

jsha: Good point. Perhaps "it is safe to call..."? Or a PR making all other methods short-circuit when the error field is set?

It seems that at least `write_tls` is safe to call after `process_new_packets` returns an error, and is indeed appropriate, because it's potentially needed in order to send an alert:

https://github.com/ctz/rustls/blob/a49f1117cad9082f1d0ccfd2ff45d3a336462f76/rustls/src/session.rs#L190-L200

Issue #440, opened 2021-01-22
When verifying certificates, a single time should be used


briansmith:
When verifying SCTs, we use two different calls to SystemTime::now(). One is used for the call into webpki:

```
fn try_now() -> Result<webpki::Time, TLSError> {
    webpki::Time::try_from(std::time::SystemTime::now())
        .map_err(|_| TLSError::FailedToGetCurrentTime)
}
```

and another is used for the call into sct.rs:
```
fn unix_time_millis() -> Result<u64, TLSError> {
    std::time::SystemTime::now()
        .duration_since(std::time::UNIX_EPOCH)
        .map(|dur| dur.as_secs())
        .map_err(|_| TLSError::FailedToGetCurrentTime)
        .and_then(|secs| {
            secs.checked_mul(1000)
                .ok_or(TLSError::FailedToGetCurrentTime)
        })
}
```

I think it would be better to have a single time for which we're validating the certificate, i.e. a single call to `SystemTime::now()` for all aspects of certificate validation. I think this would be more logically correct. It would also be more efficient. (I also think webpki should just do the SCT validation so Rustls doesn't need to, and so Rustls can support SCTs within certificates; see https://github.com/briansmith/webpki/issues/177.) 

Also note that there is a third call to `SystemTime::now()` per handshake, which is used for ticket rolling. I think in terms of correctness, this duplication is not as important as the duplication between SCT and other certificate validation, but it isn't optimal w.r.t. performance.

briansmith:
I think this should be closed because #442 fixes it. Is there anything remaining @djc?

djc:
#442 didn't include the ticketer, maybe we should keep this open for that?

briansmith:
The ticketer isn't about verifying certificates so maytbe we should just have a separate issue about that if people think it is important. I don't really think it is that urgent after looking into it.

PR #442, opened 2021-01-23
Use single time for verifications (see #440)

djc:
As suggested by @briansmith in #440. Criterion suggests this is a 2.4% improvement on read_tls with EWOULDBLOCK.

It does change the verifier traits, so would be breaking for everyone implementing those traits.

### DIFF
--- a/rustls/src/verifybench.rs
+++ b/rustls/src/verifybench.rs
@@ -18,272 +18,190 @@ fn duration_nanos(d: Duration) -> u64 {
     ((d.as_secs() as f64) * 1e9 + (d.subsec_nanos() as f64)) as u64
 }
 
-fn bench<Fsetup, Ftest, S>(count: usize, name: &'static str, f_setup: Fsetup, f_test: Ftest)
-where
-    Fsetup: Fn() -> S,
-    Ftest: Fn(S),
-{
-    let mut times = Vec::new();
-
-    for _ in 0..count {
-        let state = f_setup();
-        let start = Instant::now();
-        f_test(state);
-        times.push(duration_nanos(Instant::now().duration_since(start)));
-    }
-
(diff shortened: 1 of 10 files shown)


rustls/src/verifybench.rs
```diff
@@ -55,7 +54,7 @@ fn test_reddit_cert() {
         || (),
         |_| {
             let dns_name = webpki::DNSNameRef::try_from_ascii_str("reddit.com").unwrap();
-            V.verify_server_cert(&anchors, &chain[..], dns_name, &[])
+            V.verify_server_cert(&anchors, &chain[..], dns_name, &[], fixed_time())
```
briansmith, line 57: `fixed_time()` should be called before `bench()` so that its execution time isn't included in the measurement.

rustls/src/verifybench.rs
```diff
@@ -74,7 +73,7 @@ fn test_github_cert() {
         || (),
         |_| {
             let dns_name = webpki::DNSNameRef::try_from_ascii_str("github.com").unwrap();
-            V.verify_server_cert(&anchors, &chain[..], dns_name, &[])
+            V.verify_server_cert(&anchors, &chain[..], dns_name, &[], fixed_time())
```
briansmith, line 76: Ditto.

briansmith:
LGTM modulo the nit about the benchmarking. Not sure how to deal with the versioning issue w.r.t. the breaking interface change.

rustls/src/verifybench.rs
djc: Fixed.

djc:
Also added a commit to use a small abstraction for verifybench in order to reduce code duplication and make it more obvious that all of these are doing the same thing with different input.

ctz (approved):

ctz:
Thanks!

PR #443, opened 2021-01-23
Make randomness acquisition fallible (fixes #441)

### LINKED ISSUE #441: Change randomness functions to return result
https://github.com/ctz/rustls/blob/a49f1117cad9082f1d0ccfd2ff45d3a336462f76/rustls/src/rand.rs#L2-L4

> /// The single place where we generate random material
> /// for our own use.  These functions never fail,
> /// they panic on error.

Because ring's SystemRandom is implemented in terms of fallible system operations, it would be good to propagate errors and return them as Result. This is helpful in presenting an API that never panics, which makes the C API much easier.

djc:


jsha (commented):
Thanks for trying this out! What do you think about doing acquisition of session randoms higher in the stack, e.g. in `ServerSession::new()`, and then passing them down the stack as constructor parameters? This would still result in changes to a lot of method signatures, but it would allow a lot of constructors that currently can't return Error to stay that way. It would also nicely express the semantic of "making this thing is infallible, _so long as_ you can feed it some random bytes."

That would still leave fallibility in the ticket generation case, though, which I don't think we can do much about (unless we are willing to generate extra randomness just in case we later have to make a ticket).

rustls/src/rand.rs
```diff
@@ -1,25 +1,26 @@
...
+use crate::TLSError;
 /// The single place where we generate random material
 /// for our own use.  These functions never fail,
 /// they panic on error.
 use ring::rand::{SecureRandom, SystemRandom};
 
 /// Fill the whole slice with random material.
-pub fn fill_random(bytes: &mut [u8]) {
-    SystemRandom::new().fill(bytes).unwrap();
+pub fn fill_random(bytes: &mut [u8]) -> Result<(), TLSError> {
```
quininer, line 9: Can we have a separate `GetRandomFailed` error type?  This can avoid size of return value too large.

briansmith:
This looks pretty good to me. I'm not sure why you don't like it @djc. I was expecting it to end up looking worse, and also there's room for improvement. I'll comment more in the issue.

djc:
> What do you think about doing acquisition of session randoms higher in the stack, e.g. in `ServerSession::new()`, and then passing them down the stack as constructor parameters? This would still result in changes to a lot of method signatures, but it would allow a lot of constructors that currently can't return Error to stay that way. It would also nicely express the semantic of "making this thing is infallible, so long as you can feed it some random bytes."

Yup, will definitely do that on the next pass, I think it would help a lot.

> Can we have a separate `GetRandomFailed` error type? This can avoid size of return value too large.

Will also include this.

briansmith:
> Thanks for trying this out! What do you think about doing acquisition of session randoms higher in the stack, e.g. in ServerSession::new(), and then passing them down the stack as constructor parameters?

Ideally the application is able to do something like this for TCP-based TLS:

* Syscall #1: Send SYN (client) or SYN/ACK (server)
* Syscall #2: Acquire the random bytes it needs while waiting for SYN/ACK (client) or ClientHello (server).
* Syscall #3: Send ACK + ClientHello (client) or ServerHello (server).

In particular, we don't want our first segment of the TCP handshake to be blocked on sending the random bytes.

FWIW, I'm planning to extend the *ring* API with a way to coalesce the random generation of the ECDHE private key with the generation of the hello randoms in one (system) call to the PRNG.

Note that Rustls also has this line:
```
let ours = ring::agreement::EphemeralPrivateKey::generate(alg, &rng).unwrap();
```
That `unwrap()` should be replaced too, as the failure would be due to PRNG failure.

So, as far as rearranging code goes, I would try to rearrange things so the hello random and the ECDHE private key are generated close to each other.

rustls/src/client/mod.rs
```diff
@@ -662,9 +663,14 @@ impl ClientSession {
     /// we behave in the TLS protocol, `hostname` is the
     /// hostname of who we want to talk to.
     pub fn new(config: &Arc<ClientConfig>, hostname: webpki::DNSNameRef) -> ClientSession {
+        Self::try_new(config, hostname).unwrap()
```
briansmith, line 666: This function should be marked `#[deprecated]` or even better, removed. The documentation should explain why it is deprecated.

rustls/src/quic.rs
```diff
@@ -335,6 +335,15 @@ pub trait ClientQuicExt {
         hostname: webpki::DNSNameRef,
         params: Vec<u8>,
     ) -> ClientSession {
+        Self::try_new_quic(config, hostname, params).unwrap()
```
briansmith, line 338: Ditto.

rustls/src/quic.rs
```diff
@@ -360,6 +370,11 @@ pub trait ServerQuicExt {
     /// in that it takes an extra argument, `params`, which contains the
     /// TLS-encoded transport parameters to send.
     fn new_quic(config: &Arc<ServerConfig>, params: Vec<u8>) -> ServerSession {
+        Self::try_new_quic(config, params).unwrap()
```
briansmith, line 373: Ditto.

rustls/src/server/mod.rs
```diff
@@ -579,9 +581,14 @@ impl ServerSession {
     /// Make a new ServerSession.  `config` controls how
     /// we behave in the TLS protocol.
     pub fn new(config: &Arc<ServerConfig>) -> ServerSession {
-        ServerSession {
-            imp: ServerSessionImpl::new(config, vec![]),
-        }
+        Self::try_new(config).unwrap()
```
briansmith, line 584: Ditto.

rustls/src/client/mod.rs
djc: You mean `new()`? Isn't it reasonable for many (types of) callers to continue to just panic for this type of failure? It seems like so far it really hasn't caused to many issues. I guess I could see the point that callers should be forced to make that decision for themselves.

rustls/src/server/mod.rs
```diff
@@ -371,10 +372,11 @@ impl ServerSessionImpl {
             state: Some(Box::new(hs::ExpectClientHello::new(
                 server_config,
                 extra_exts,
+                SessionRandoms::for_server()?,
```
briansmith, line 375: IIUC, the API breaking change here is due to the refactoring where the randoms are generated early, during construction of the session, right? IMO, this is the second strike against this idea. It seems better to delay the generation of the randoms until they are needed. Since those code paths are already fallable to deal with I/O (IIUC), then no API change would be needed.

djc: Fair enough, I'll look into that next.

rustls/src/client/mod.rs
briansmith: > You mean `new()`? Isn't it reasonable for many (types of) callers to continue to just panic for this type of failure? It seems like so far it really hasn't caused to many issues. I guess I could see the point that callers should be forced to make that decision for themselves.

Yes. To clarify, I think we should probably try to avoid using the CSPRNG in the `new` functions at all so that `try_new` isn't necessary; i.e. so `new()` becomes truly infallible. Faliing that, we should just remove the `new()` functions. Failing that, we should deprecated them and document why they are deprecated.

More generally, the API that Rustls should be designed so that Rustls will never panic itself. There might be cases where we have `unwrap()` in the code, but those should be unreachable code, and we should be confident that they are unreachable.

(I also think that eventually we should remove the `unwrap()`s from the example code.)

djc: I agree that we should only retain tightly audited `unwrap()` calls, but there quite a lot of `unwrap()` calls right now. I don't mind spending some time on this, but it would be a bit easier to do so if the work could be funded. @jsha would that be an option for your project?

jsha: I'll ask!

briansmith: > I agree that we should only retain tightly audited `unwrap()` calls, but there quite a lot of `unwrap()` calls right now.

I filed issue #447 to track the effort of eliminating all the potential for panics within Rustls.

ctz:
One use of unwrap() in the codebase here is to avoid untestable error handling code -- which runs for the very first time in the field in hard-to-reproduce conditions. To put it another way -- how can we arrange to test that the error handling for any one of these fallible calls works properly?

briansmith:
> One use of unwrap() in the codebase here is to avoid untestable error handling code -- which runs for the very first time in the field in hard-to-reproduce conditions. To put it another way -- how can we arrange to test that the error handling for any one of these fallible calls works properly?

AFAICT, there is no new code path introduced with this PR. The only difference is that instead of callers catching the error as a panic using `std::panic::catch_unwind`, they would catch it using normal `Result` handling. Within Rustls, we're only bubbling up the error, never trying to handle it.

On Linux, we could write tests that intercept syscalls and force the failure conditions using ptrace. BoringSSL's test suite has that. I can look into exposing that via *ring* for *ring* and other *ring*-based tests, including Rustls, to use. However, IMO that doesn't need to block this improvement.

rustls/src/server/mod.rs
djc: This has been solved (for the server side).

djc:
@ctz FWIW, Brian's point here seems true to me:

> AFAICT, there is no new code path introduced with this PR. The only difference is that instead of callers catching the error as a panic using `std::panic::catch_unwind`, they would catch it using normal `Result` handling. Within Rustls, we're only bubbling up the error, never trying to handle it.

The only thing happening here is propagating failures explicitly in the type system. For careful API users, all this does is make potential failures more explicit.

ctz:
> (I'd like some sort of pronouncement: should I continue maintaining this so you can at some point review/merge it, or is this something you definitely don't want to do?)

Sorry for dropping this. Thanks for this work -- I would indeed like to get this merged (and, really in preference to other PRs that will make it harder to merge this). Could you rebase?

djc:
Rebased.

ctz:
Hm, this seems to have regressed some bogo tests, for example TLS13-HelloRetryRequest-Client-TLS-Sync-ImplicitHandshake: see https://github.com/ctz/rustls/runs/2089338754?check_suite_focus=true#step:8:1675

I think this is due to `emit_client_hello_for_retry` now generating its own randoms every time, which means HelloRetryRequest gets a different one to the original ClientHello.

djc:
Ah, that's bad. Will take a look at fixing it.

briansmith:
PR #533 was intended to help ensure that wouldn't happen.

PR #445, opened 2021-01-26
Add TLS Certificate Compression (RFC8879)

sharksforarms:
Hello RustTLS maintainers! Thanks for the awesome library. It has been a pleasure to develop with.

This PR implements [RFC8879](https://tools.ietf.org/html/rfc8879)

Issue: https://github.com/ctz/rustls/issues/436

## Questions
- `certificate_compression` feature, should this be changed into separate features for each algorithm? Users may only want to support a number of algorithms without the added dependencies.
- I'd like to include arguments in tlsclient/tlsserver examples, with the current DocOpt impl, this would be hacky to integrate. Any appetite to using [structopt](https://github.com/TeXitoi/structopt) instead? If so, I can create a PR for this.
- Any guidance in regards to testing would be appreciated. I'm not sure what the best strategy is. Smoke tests in `tests/api.rs` is possible, although is there any way to assert on plaintext messages vs encrypted opaque blobs?

## TODO
- Minimum version errors in CI, they don't seem to indicate exactly what's wrong, but it seems to be related to zstd. I can't seem to reproduce this locally... most likely ubuntu package differences...

<details>
  <summary>How to test w/ tlsclient/tlsserver</summary>
  
Modify Client/Server configs

For example,
```rust
    config.certificate_compression_algorithms =
        Some(vec![rustls::CertificateCompressionAlgorithm::Brotli]);
```

Generate certs w/ `subjectAltName` (this seems to be required in rustls?)

`cd rustls/rustls-mio`

```
openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -out localhost.cert -keyout localhost.key -subj '/CN=localhost' -config openssl.cfg
```

```
# openssl.cfg
[req]
distinguished_name=dn
x509_extensions=ext
[ dn ]
CN=localhost
[ ext ]
subjectAltName = @alt_names
[alt_names]
DNS.1 = localhost
IP.1 = 127.0.0.1
IP.2 = ::1
```

Server
```
RUST_LOG=trace SSLKEYLOGFILE=$HOME/keylog cargo run --example tlsserver -- --certs ./localhost.cert --key ./localhost.key --port 8083 --verbose http
```

Client
```
RUST_LOG=trace cargo run --example tlsclient -- --cafile ./localhost.cert -p 8083 --verbose --http localhost
```

- Open wireshark
- Observe ClientHello, see new certificate_compression extension (plain text), with 1 compression option (brotli, as configured in tlsclient.rs)
- Observe ServerHello (this will be encrypted)
- Right click on ServerHello -> Protocol Preferences -> (Pre)-Master Secret Log Filename -> Click Browse under "(Pre)-Master Secret Log Filename", select `~/keylog`
- Observe decrypted ServerHello and `CompressedCertificate` extention (brotli)
</details>

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -277,6 +277,17 @@ fn emit_client_hello_for_retry(
         exts.push(ClientExtension::PresharedKeyModes(psk_modes));
     }
 
+    if support_tls13 {
+        if let Some(cert_comp_algos) = &sess
+            .config
+            .certificate_compression_algorithms
+        {
+            exts.push(ClientExtension::CompressCertificate(
+                cert_comp_algos.clone(),
+            ))
+        }
+    }
+
     if !sess.config.alpn_protocols.is_empty() {
         exts.push(ClientExtension::Protocols(ProtocolNameList::from_slices(
             &sess
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -277,14 +277,17 @@ fn emit_client_hello_for_retry(
         exts.push(ClientExtension::PresharedKeyModes(psk_modes));
     }
 
-    if support_tls13 {
-        if let Some(cert_comp_algos) = &sess
-            .config
-            .certificate_compression_algorithms
-        {
-            exts.push(ClientExtension::CompressCertificate(
-                cert_comp_algos.clone(),
-            ))
+    #[cfg(feature = "certificate_compression")]
+    {
+        if support_tls13 {
+            if let Some(cert_comp_algos) = &sess
+                .config
+                .certificate_compression_algorithms
+            {
+                exts.push(ClientExtension::CompressCertificate(
+                    cert_comp_algos.clone(),
+                ))
+            }
         }
     }
 
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1,7 +1,8 @@
-use crate::key;
 use crate::msgs::base::{Payload, PayloadU8, PayloadU16, PayloadU24};
 use crate::msgs::codec;
 use crate::msgs::codec::{Codec, Reader};
+use crate::msgs::enums::CertificateCompressionAlgorithm;
+use crate::msgs::enums::ContentType;
 use crate::msgs::enums::ECCurveType;
 use crate::msgs::enums::PSKKeyExchangeMode;
 use crate::msgs::enums::{CertificateStatusType, ClientCertificateType};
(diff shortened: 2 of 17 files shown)


rustls/src/client/hs.rs
```diff
@@ -277,6 +277,20 @@ fn emit_client_hello_for_retry(
...
     }
 
+    #[cfg(feature = "certificate_compression")]
+    {
+        if support_tls13 {
+            if let Some(cert_comp_algos) = &sess
+                .config
+                .certificate_compression_algorithms
+            {
+                exts.push(ClientExtension::CompressCertificate(
```
sharksforarms, line 287: Add check for `cert_comp_algos.is_empty()`, in the case of `Some(vec![])`

ctz:
This looks like its going in a reasonable direction. Thanks!

> Any guidance in regards to testing would be appreciated.

A good first step would be removing the lines mentioning "CertCompression" from the "DisabledTests" part of bogo/config.json.  To run these locally you'll need golang installed, but past that just sitting in bogo/ and running the "runme" shell script should suffice.

rustls/src/msgs/handshake.rs
```diff
@@ -1426,6 +1454,121 @@ impl CertificateEntry {
...
+        Some(CompressedCertificatePayload {
+            algorithm: CertificateCompressionAlgorithm::read(r)?,
+            uncompressed_length: codec::u24::read(r)?,
+            compressed_certificate_message: PayloadU24::read(r)?,
+        })
+    }
+}
+
+impl CompressedCertificatePayload {
+    pub fn compress_with(
```
ctz, line 1489: I'd loosely prefer to see this code live elsewhere in the crate -- the msgs module is trying to be a type-safe expression of the TLS protocol types, but without any processing/manipulation of them.

ctz: Perhaps it should live somewhere in `crate::compression::`?

sharksforarms:
> This looks like its going in a reasonable direction. Thanks!
> 
> > Any guidance in regards to testing would be appreciated.
> 
> A good first step would be removing the lines mentioning "CertCompression" from the "DisabledTests" part of bogo/config.json. To run these locally you'll need golang installed, but past that just sitting in bogo/ and running the "runme" shell script should suffice.

For the certificate compression tests, bogo seems to use an arbitrary/custom algorithm to test, probably just to test the feature?

For example, the test cases send a compression of `0xff02` and `0xff01` (which my branch doesn't support...)

They "install" it at testing time here:
https://github.com/google/boringssl/blob/3d8b8c3df27cc780a49607be6a27f2c518f5c082/ssl/test/test_config.cc#L1365

And use them here:
https://github.com/google/boringssl/blob/3d8b8c3df27cc780a49607be6a27f2c518f5c082/ssl/test/runner/runner.go#L15900

Not sure how rustls should handle this...

Something like:
`certificate_compression_algorithms: Vec<dyn AlgImpl>`?

That way it remains user configurable and more algorithms can be added (In the case of the tests?)

Do we want users to be able to implement custom compression algorithms?

PR #454, opened 2021-02-02
Remove `unwrap()`s in tls13::prepare_resumption.

briansmith:


rustls/src/client/hs.rs
```diff
@@ -171,17 +171,12 @@ struct ExpectServerHelloOrHelloRetryRequest(ExpectServerHello);
...
-                suite.can_resume_to(&resuming_suite)
-            } else {
-                true
-            }
-        }
-        None => false,
+    if let Some(suite) = sess.common.get_suite() {
+        suite.can_resume_to(&resuming_suite)
+    } else {
+        true
```
djc, line 179: I generally prefer to use this pattern for things like this. It's a little more compact and I find the different outcomes easier to parse visually. Also works great for early return/continue.

```suggestion
    match sess.common.get_suite() {
        Some(suite) => suite.can_resume_to(&resuming_suite),
        None => true,
    }
```

rustls/src/client/tls13.rs
```diff
@@ -301,55 +301,44 @@ pub fn prepare_resumption(
...
-        if sess.config.enable_early_data && max_early_data_size > 0 && !doing_retry {
-            sess.early_data
-                .enable(max_early_data_size as usize);
-            exts.push(ClientExtension::EarlyData);
-        }
+    let resuming_session = if let Some(resuming_session) = handshake.resuming_session.as_ref() {
+        resuming_session
+    } else {
+        return false;
+    };
```
djc, line 308: Would make a similar change here.

rustls/src/client/tls13.rs
```diff
@@ -301,55 +301,44 @@ pub fn prepare_resumption(
...
-                resuming.get_obfuscated_ticket_age(ticketer::timebase()),
-                resuming.cipher_suite,
-            )
-        };
+    let resuming_suite = match sess.find_cipher_suite(resuming_session.cipher_suite) {
+        Some(resuming_suite) if hs::compatible_suite(sess, resuming_suite) => resuming_suite,
+        _ => {
+            return false;
+        }
+    };
```
djc, line 315: And here you already did it like that (though stable rustfmt would format as `_ => return false,`, which is preferable, IMO).

briansmith: I think we should just let whoever can rustfmt stuff do the formatting. I can't guess what rustfmt 2 would do, especially I don't know what the brace style would be.

rustls/src/client/hs.rs
briansmith: Thanks. I updated the PR with this change.

rustls/src/client/tls13.rs
briansmith: Done.

PR #455, opened 2021-02-02
Cache current time at start of handshake.

briansmith:
Cache the current time at the start of the handshake and use that time
in various places in the handshake. A couple other uses of
`SystemTime::now()` remain; I hope to eliminate them in a future commit
after some other refactoring is done.

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -16,6 +16,7 @@ use crate::suites;
 use webpki;
 
 use std::mem;
+use std::time::SystemTime;
 
 pub struct ServerCertDetails {
     pub cert_chain: CertificatePayload,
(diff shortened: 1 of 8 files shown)


briansmith:
The remaining use of `SystemTime::now()` is in `ticketer::timebase()`. I am planning to, at least, remove the use of `ticketer::timebase()` in `client::tls13::prepare_resumption` in a future PR, after PR #454 is completed.

briansmith:
@djc WDYT?

djc:
I think I would be hesitant to make this trade-off without a better sense of what the costs of the syscall are compared to the entire handshake process, especially once networking is involved. Also, your newly added comment states getting a timestamp is slow on "some platforms" -- if you have more specific knowledge about this, it might be useful to just write it down.

IMO what would be useful is to hoist the `SystemTime::now()` calls to a higher level, for example, somewhere in or around `process_new_packets()` after it has figured there is enough data to process a packet. You already have some of that here, but I think it might go higher-level still (I don't know if people have seen "Functional core, imperative shell", but I certainly find that a compelling architecture for most things).

briansmith:
> I think I would be hesitant to make this trade-off without a better sense of what the costs of the syscall are compared to the entire handshake process, especially once networking is involved.

In many cases, there are no syscalls involved. But, in some cases, there are. On Linux if vDSO is used then you're golden. In some virtualized environments, it has been slow, but also it has been optimized in recent versions of the VM hosts.

> IMO what would be useful is to hoist the SystemTime::now() calls to a higher level, for example, somewhere in or around process_new_packets() after it has figured there is enough data to process a packet. You already have some of that here, but I think it might go higher-level still (I don't know if people have seen "Functional core, imperative shell", but I certainly find that a compelling architecture for most things).

In order to send the client hello, one needs to know what time it is if there's a session in the session cache (to see if it is expired). Similarly for sending the server hello. So you basically need to know the time at the start of the handshake.

I would prefer to thread all the bits of state individually through the state machine at a much finer granularity than we currently do.

In any case, I started working on this because I was hoping to eliminate `ticketer::timebase()` and minimize the number of times we do `SystemTime`-to-`u64` conversions, to get rid of the `unwrap()`. I think it probably isn't worth pursuing this right now, so I'll close this and come back to it another time, maybe.

PR #456, opened 2021-02-02
Hash hs unwrap

djc:
This reduces the number of `unwrap()` calls from 6 to 2 and removes one `debug_assert!()`. I believe for these remaining calls, we should propagate them as a local error type (maybe `InvalidState`) to the caller. Or, even better, maybe `HandshakeDetails` should become more typestate-like, evolving its use of types as it goes through the handshake process, so that we could express the progression of available data in the type system as well.

@briansmith any thoughts?

### DIFF
--- a/rustls/src/hash_hs.rs
+++ b/rustls/src/hash_hs.rs
@@ -93,8 +93,8 @@ impl HandshakeHash {
 
     /// Hash or buffer a byte slice.
     fn update_raw(&mut self, buf: &[u8]) -> &mut Self {
-        if self.ctx.is_some() {
-            self.ctx.as_mut().unwrap().update(buf);
+        if let Some(ctx) = &mut self.ctx {
+            ctx.update(buf);
         }
 
         if self.ctx.is_none() || self.client_auth_enabled {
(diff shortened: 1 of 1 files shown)


rustls/src/hash_hs.rs
```diff
@@ -15,10 +15,7 @@ use std::mem;
 /// This is disabled in cases where client auth is not possible.
 pub struct HandshakeHash {
     /// None before we know what hash function we're using
-    alg: Option<&'static digest::Algorithm>,
-
-    /// None before we know what hash function we're using
-    ctx: Option<digest::Context>,
+    ctx: Option<(&'static digest::Algorithm, digest::Context)>,
```
briansmith, line 18: `digest::Context::algorithm()` provides access to the algorithm so you should remove the first part of this tuple.

rustls/src/hash_hs.rs
```diff
@@ -107,12 +101,13 @@ impl HandshakeHash {
...
-            ctx.update(&self.buffer);
-            ctx
-        } else {
-            self.ctx.as_ref().unwrap().clone()
+        let mut ctx = match &self.ctx {
+            Some((_, ctx)) => ctx.clone(),
+            None => {
+                let mut ctx = digest::Context::new(hash);
+                ctx.update(&self.buffer);
+                ctx
```
briansmith, line 109: This is duplicating PR #451. I agree your way of using `ctx.clone()` is better assuming it is correct. However, I think that should be done in a separate PR. In this PR, I would keep the logic the same but remove the unwraps. In this specific instance, I would just revert this change in favor of PR #451.

rustls/src/hash_hs.rs
briansmith, line 18: `digest::Context::algorithm()` can be used to get the digest algorithm, so the first part of the tuple isn't needed and is bad because it makes us worry that it is inconsistent. We should remove the first part of the tuple and just leave the `Context`.

briansmith (changes requested):

djc:
Ah, silly to duplicate your work. Retargeted some additional improvements at your branch in https://github.com/briansmith/rustls/pull/4.

PR #457, opened 2021-02-02
Remove the need to convert digest algorithms to HMAC algorithms.

briansmith:


ctz:
Merged as 9e59e2c46a2bb315a0a0df3679af069f29acc163

PR #458, opened 2021-02-02
Avoid allocating a `Vec` for each digest created during the handshake.

briansmith:
Avoid heap allocation at every time where a digest is computed; that
happens frequently in the TLS 1.3 handshake. Use a stronger type to
hold the digest where practical.

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -247,12 +247,13 @@ pub fn start_handshake_traffic(
     // the two halves will have different record layer protections.  Disallow this.
     hs::check_aligned_handshake(sess)?;
 
-    handshake.hash_at_client_recvd_server_hello = handshake.transcript.get_current_hash();
+    let hash_at_client_recvd_server_hello = handshake.transcript.get_current_hash();
+    handshake.hash_at_client_recvd_server_hello = Some(hash_at_client_recvd_server_hello);
 
     let _maybe_write_key = if !sess.early_data.is_enabled() {
         // Set the client encryption key for handshakes if early data is not used
         let write_key = key_schedule.client_handshake_traffic_secret(
-            &handshake.hash_at_client_recvd_server_hello,
+            &hash_at_client_recvd_server_hello,
             &*sess.config.key_log,
             &handshake.randoms.client,
         );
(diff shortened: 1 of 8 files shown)


rustls/src/server/tls13.rs
```diff
@@ -1032,7 +1033,7 @@ impl hs::State for ExpectFinished {
         let read_key = self
             .key_schedule
             .client_application_traffic_secret(
-                &self.handshake.hash_at_server_fin,
+                self.handshake.hash_at_server_fin.as_ref().unwrap(),
```
briansmith, line 1036: It's possible to avoid this unwrap but it requires refactoring some parts of the state machine. I have a PR pending, for after this is merged, to thread this digest through the state machine so that this unwrap goes away. That PR eliminates `HandshakeDetails::hash_at_server_fin` completely.

briansmith:
I simplified this PR to reduce the size of the diff.

rustls/src/client/tls13.rs
```diff
@@ -479,9 +480,11 @@ impl hs::State for ExpectEncryptedExtensions {
                 let write_key = self
                     .key_schedule
                     .client_handshake_traffic_secret(
-                        &self
+                        self
                             .handshake
-                            .hash_at_client_recvd_server_hello,
+                            .hash_at_client_recvd_server_hello
+                            .as_ref()
+                            .unwrap(),
```
briansmith, line 487: I have a PR on top of this one to remove this `unwrap()`, by threading `hash_at_client_recvd_server_hello` through the handshake state machine directly.

rustls/src/client/tls13.rs
```diff
@@ -989,8 +992,8 @@ impl hs::State for ExpectFinished {
             let key = st
                 .key_schedule
                 .client_handshake_traffic_secret(
-                    &st.handshake
-                        .hash_at_client_recvd_server_hello,
+                    st.handshake
+                        .hash_at_client_recvd_server_hello.as_ref().unwrap(),
```
briansmith, line 996: I have a queued PR on top of this one to remove this `unwrap()`, by threading `hash_at_client_recvd_server_hello` through the handshake state machine directly.

rustls/src/client/tls13.rs
djc: Why not just add the changes to this PR as a separate commit?

briansmith: I think that's out of scope for this PR. Note that if this `unwrap()` were to fail, then previously we were calculating the wrong value. And similarly for the other two `unwrap()`s I added in this PR.

ctz:
Merged as faa3acfe75fd9e41686d5c470aac1c4d6c85c846

PR #459, opened 2021-02-02
Reduce the number of times we allocate a `Vec<u8>` to hold an HMAC tag.

briansmith:
The serialization framework currently stores bytes in `Vec` so not all
of the allocations could be eliminated, but the ones that can were.


djc (approved):

ctz:
Merged as a57894c73d3eb030a443178a3c447b137b7f2b0a

PR #462, opened 2021-02-03
Change the certificate validation API to guarantee end-entity is present.

briansmith:
Rustls already handles internally the case where no certificate is
present in the Certificate message. Every implementation of the
certificate verification APIs currently needs to deal with the
possibility that the end-entity certificate wasn't provided, or
dangerously assume that the list of certificates it is given is never
empty. Make it clear that it will be given at least the end-entity
certificate by splitting the end-entity certificate out into its own
parameter. This makes it easier to correctly and safely implement the
certificate verifier interfaces.

In the previous API, the ordering of the certificates in the chain was
not documented; the API required knowledge of the TLS specification and
an assumption that Rustls does the "obvious" thing to know that the
end-entity certificate is at the beginning. The new API avoids any
doubt about this. Also, there is now documentation that guarantees that
the intermediates are in the same order that they were in the handshake.

This also makes static analysis easier because it eliminates a use of
the panic-prone `[]` slice indexing operator.

### DIFF
--- a/rustls/src/server/tls12.rs
+++ b/rustls/src/server/tls12.rs
@@ -63,25 +63,26 @@ impl hs::State for ExpectCertificate {
                 TLSError::General("client rejected by client_auth_mandatory".into())
             })?;
 
-        if cert_chain.is_empty() {
-            if !mandatory {
-                debug!("client auth requested but no certificate supplied");
-                self.handshake
-                    .transcript
-                    .abandon_client_auth();
-                return Ok(self.into_expect_tls12_client_kx(None));
-            }
-            sess.common
-                .send_fatal_alert(AlertDescription::CertificateRequired);
-            return Err(TLSError::NoCertificatesPresented);
-        }
-
         trace!("certs {:?}", cert_chain);
+
+        let (end_entity, intermediates) = match cert_chain.split_first() {
+            None => {
+                if !mandatory {
+                    debug!("client auth requested but no certificate supplied");
+                    self.handshake.transcript.abandon_client_auth();
+                    return Ok(self.into_expect_tls12_client_kx(None));
+                }
+                sess.common
+                    .send_fatal_alert(AlertDescription::CertificateRequired);
+                return Err(TLSError::NoCertificatesPresented);
+            }
+            Some(chain) => chain,
+        };
 
         let now = std::time::SystemTime::now();
         sess.config
             .verifier
-            .verify_client_cert(cert_chain, sess.get_sni(), now)
+            .verify_client_cert(end_entity, intermediates, sess.get_sni(), now)
             .or_else(|err| {
                 hs::incompatible(sess, "certificate invalid");
                 Err(err)
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -771,24 +771,25 @@ impl hs::State for ExpectCertificate {
                 TLSError::General("client rejected by client_auth_mandatory".into())
             })?;
 
-        if cert_chain.is_empty() {
-            if !mandatory {
-                debug!("client auth requested but no certificate supplied");
-                self.handshake
-                    .transcript
-                    .abandon_client_auth();
-                return Ok(self.into_expect_finished());
-            }
+        let (end_entity, intermediates) = match cert_chain.split_first() {
+            None => {
+                if !mandatory {
+                    debug!("client auth requested but no certificate supplied");
+                    self.handshake.transcript.abandon_client_auth();
+                    return Ok(self.into_expect_finished());
+                }
 
-            sess.common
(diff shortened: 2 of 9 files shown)


rustls/src/server/tls12.rs
```diff
@@ -63,25 +63,26 @@ impl hs::State for ExpectCertificate {
...
+            None => {
+                if !mandatory {
+                    debug!("client auth requested but no certificate supplied");
+                    self.handshake.transcript.abandon_client_auth();
+                    return Ok(self.into_expect_tls12_client_kx(None));
+                }
+                sess.common
+                    .send_fatal_alert(AlertDescription::CertificateRequired);
+                return Err(TLSError::NoCertificatesPresented);
+            }
```
djc, line 78: Maybe write it like this? Seems slightly easier to follow.

```suggestion
            None if !mandatory => {
                    debug!("client auth requested but no certificate supplied");
                    self.handshake.transcript.abandon_client_auth();
                    return Ok(self.into_expect_tls12_client_kx(None));
            }
            None => {
                sess.common
                    .send_fatal_alert(AlertDescription::CertificateRequired);
                return Err(TLSError::NoCertificatesPresented);
            }
```

rustls/src/server/tls13.rs
```diff
@@ -771,24 +771,25 @@ impl hs::State for ExpectCertificate {
...
-            sess.common
-                .send_fatal_alert(AlertDescription::CertificateRequired);
-            return Err(TLSError::NoCertificatesPresented);
-        }
+                sess.common
+                    .send_fatal_alert(AlertDescription::CertificateRequired);
+                return Err(TLSError::NoCertificatesPresented);
+            }
+            Some(chain) => chain,
+        };
```
djc, line 787: Same here.

djc (commented):
Looks great! Some style suggestion below.

rustls/src/server/tls12.rs
briansmith: I'd rather keep the current structure for now, because it is intentionally the same structure that was used in the code previously. Maybe we can make that kind of refactoring in a separate commit. I want to keep the focus of this commit on eliminating the use of the array indexing and improving the public API.

djc: Fair enough.

rustls/src/verify.rs
```diff
@@ -319,29 +324,22 @@ type CertChainAndRoots<'a, 'b> = (
...
+    intermediates: &'a [Certificate],
     roots: &'b RootCertStore,
-    presented_certs: &'a [Certificate],
 ) -> Result<CertChainAndRoots<'a, 'b>, TLSError> {
-    if presented_certs.is_empty() {
-        return Err(TLSError::NoCertificatesPresented);
-    }
-
     // EE cert must appear first.
-    let cert = webpki::EndEntityCert::from(&presented_certs[0].0).map_err(TLSError::WebPKIError)?;
```
jsha, line 330: This is the only use of the `slice[]` indexing operator that I can see, and it seems like it could be replaced by

```
let ee = presented_certs.first().ok_or(TLSError::WebPKIError)?.0;
```

Are there other cases I'm missing? It may be good to change the API for clarity reasons, but it doesn't seem that the API change is needed in order to avoid the indexing operator.

briansmith: It is true there are other ways to avoid the slicing. However, I think this is the best way, as it resolves several issues with the API. I have updated the commit message and the PR description with more details about the improvements to the API, and I've improved the doc comments in the code.

rustls/src/verify.rs
```diff
@@ -78,13 +78,18 @@ impl ClientCertVerified {
 /// Something that can verify a server certificate chain, and verify
 /// signatures made by certificates.
 pub trait ServerCertVerifier: Send + Sync {
-    /// Verify a the certificate chain `presented_certs` against the roots
-    /// configured in `roots`.  Make sure that `dns_name` is quoted by
-    /// the top certificate in the chain.
+    /// Verify a the end-entity certificate `end_entity` is valid for the
```
jsha, line 81:
```suggestion
    /// Verify the end-entity certificate `end_entity` is valid for the
```

rustls/src/verify.rs
```diff
@@ -181,13 +185,20 @@ pub trait ClientCertVerifier: Send + Sync {
         sni: Option<&webpki::DNSName>,
     ) -> Option<DistinguishedNames>;
 
-    /// Verify a certificate chain. `presented_certs` is the certificate chain from the client.
+    /// Verify a the end-entity certificate `end_entity` is valid for the
```
jsha, line 188:
```suggestion
    /// Verify the end-entity certificate `end_entity` is valid for the
```

jsha (commented):
Looks good to me!

PR #471, opened 2021-02-04
Clarify that only one key share is sent per client hello.

briansmith:
Refactor the code to eliminate an unneeded `unwrap()` and an unneeded
one-element vector allocation.

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -103,33 +103,26 @@ pub fn choose_kx_groups(
     // - if not, we might have a hint of what the server supports
     // - if not, send just X25519.
     //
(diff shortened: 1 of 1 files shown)


briansmith:
Besides these changes, there are a lot more simplifications we could make to the code. Are there any near-term plans to implement support for multiple key shares per client hello? If not, then I will follow up this PR with another PR to simplify the key share logic much more.

ctz:
One reason to send two could be if a popular server implements just the MTI key exchange algorithm (sadly, nistp256) and ignores the "SHOULD support" for X25519. However, I don't have any data for that either way. Sending one is probably fine, I think, and I'd like to keep that as X25519 absent solid data.

PR #474, opened 2021-02-05
Use the `find` combinator on iterators

huitseeker:
and other minor refactorings of pattern-matches on `Option` / `Result` that have a direct combinator equivalent.

Reviewed from tool suggestions produced with [comby-rust](https://github.com/huitseeker/comby-rust).


briansmith:
LGTM!

Issue #380, opened 2020-06-28
CloseNotify alert received when using rustls/examples/simpleclient.rs


volfco:
I'm trying to figure out how to write my own HTTP client using `rustls/examples/simpleclient.rs` as a base. 

The code in master works when using it against `google.com:443` or oracle.com, but fails on other popular sites such as reddit.com, twitter.com, github.com with the following error (this was against twitter.com):

```
Current ciphersuite: TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: ConnectionAborted, error: "CloseNotify alert received" }', src/main.rs:40:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Not sure where in the pipeline this error is happening.

ctz:
This is a normal and expected error: I've ensure it's mentioned in the docs as of a5e14882. TLDR: it signifies clean connection closure by the peer. close_notify does not have universal support in TLS libraries/applications, so that's why you see variance between hosts. In your application you should respond by tearing down the underlying TCP connection.

volfco:
I opened this to fix the example code so it's handled cleanly. I'm just trying to wrap my head around all of this, which is why I called out the example.

PR #476, opened 2021-02-08
Add a simple mTLS example

uggla:
Hello,

First thanks for all the hard work on rustls. It is great !

Here is a small example doing mTLS with rustls.
I spent some time understanding how to load certs from pem file and mainly with #380 issue.
So this is an example for beginner like me to have a straight working example.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -16,6 +16,7 @@ log = { version = "0.4.4", optional = true }
 ring = "0.16.19"
 sct = "0.6.0"
 webpki = "0.21.4"
+rustls-pemfile = "0.2.0"
 
 [features]
 default = ["logging"]
@@ -28,7 +29,6 @@ env_logger = "0.8.2"
 log = "0.4.4"
 webpki-roots = "0.21"
 criterion = "0.3.0"
-rustls-pemfile = "0.2.0"
 base64 = "0.13.0"
 
 [[example]]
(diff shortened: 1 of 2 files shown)


briansmith:
First, @uggla, are you still interested in this?

I'm curious about what @djc and @ctz think about this.

I wonder if we should avoid having a bunch of little demo apps that each demonstrate a different facet, and instead demonstrate this kind of thing by adding flags to `tlsclient` and `tlsserver`.

djc:
I don't like example proliferation. This might be a better as an additional option to the existing client, and perhaps additional clarifications in the reference documentation if we think the current approach to mTLS is hard to learn.

uggla:
Hello @briansmith, @djc

I'm used to looking at examples and I like if I can find one really close to what I need.
However, if you judge this is not a proper approach to document things and prefer another way, I'll be happy to help.

briansmith:
I looked at tlsclient and I see it already has:
```
    --auth-key KEY      Read client authentication key from KEY.
    --auth-certs CERTS  Read client authentication certificates from CERTS.
                        CERTS must match up with KEY.
```
And I looked at tlsserver and it has:
```
    --auth CERTFILE     Enable client authentication, and accept certificates
                        signed by those roots provided in CERTFILE.
```

Given this, I'd lean towards trying to clean up tlsclient and tlsserver code to make it easier to read, and making the comments in it much more educational/tutorial in nature, so that those two programs can serve as the example.

Also, pretty much everybody is using tokio-rustls or equivalent, AFAICT, so IMO more people would benefit from a demo that uses tokio-rustls--perhaps in that repo.

WDYT?

djc:
Yes, let's close this in favor of improving existing examples/documentation.

PR #482, opened 2021-02-10
Remove unwrap in client find_session

ab-cgh:


rustls/src/client/hs.rs
```diff
@@ -77,21 +77,22 @@ fn find_session(
...
-    let key_buf = key.get_encoding();
-
-    let maybe_value = sess
-        .config
-        .session_persistence
-        .get(&key_buf);
+    let key_buf = {
+        let key = persist::ClientSessionKey::session_for_dns_name(dns_name);
+        key.get_encoding()
+    };
```
briansmith, line 83: Seems like this change to the scope of `key` is out of scope since it isn't needed for removing the `unwrap()` and it isn't clearly better (or worse) than what was there before. I would revert this change and maybe add this to a TODO for discussing code style changes.

rustls/src/client/hs.rs
```diff
@@ -77,21 +77,22 @@ fn find_session(
...
 
-    if maybe_value.is_none() {
-        debug!("No cached session for {:?}", dns_name);
-        return None;
-    }
+    let config = &sess.config.session_persistence;
+  
+    let value = match config.get(&key_buf) {
```
briansmith, line 87: I would have made this change instead:
```
    let value = sess
        .config
        .session_persistence
        .get(&key_buf)
        .or_else(|| {
            debug!("No cached session for {:?}", dns_name);
            None
        })?;
```

rustls/src/client/hs.rs
```diff
@@ -77,21 +77,22 @@ fn find_session(
...
+        Some(v) => v,
+        None => {
+            debug!("No cached session for {:?}", dns_name);
+            return None;
+        }
+    };
 
-    let value = maybe_value.unwrap();
-    let mut reader = Reader::init(&value[..]);
```
briansmith, line 94: switching from `&value[..]` to `value.as_slice()` is also an out-of-scope change that should be reverted.

If we want to switch the coding pattern here, we should probably do so globally. Probably we should just do what Clippy recommends. Note that `as_slice()` is also unneeded here; just `&value` is enough. But, let's save that for another day.

briansmith (changes requested):

rustls/src/client/hs.rs
ab-cgh: This has been reverted.

rustls/src/client/hs.rs
ab-cgh: Reverted for now as well.

rustls/src/client/hs.rs
ab-cgh: Sure, this works as well. I made the change to keep less intermediate variables.

briansmith (approved):

djc (approved):
Looks good, thanks!

PR #484, opened 2021-02-10
Remove unwrap in deframe_one

ab-cgh:


rustls/src/msgs/hsjoiner.rs
```diff
@@ -90,16 +90,15 @@ impl HandshakeJoiner {
...
             let mut rd = codec::Reader::init(&self.buf);
-            let payload = HandshakeMessagePayload::read_version(&mut rd, version);
-
-            if payload.is_none() {
-                return false;
-            }
-
+            let payload = match HandshakeMessagePayload::read_version(&mut rd, version) {
+                Some(p) => p,
+                None => return false,
```
briansmith, line 95: Normally I do:
```
=> {
    return false;
}
```

Although that's partly because rustfmt forces me to, I think.

ab-cgh: Stable rustfmt didn't change this for me. I don't think rustfmt 2 will either, but I'm unable to install it currently to test that.

briansmith (approved):

rustls/src/msgs/hsjoiner.rs
briansmith: I see some other instances in the codebase where this formatting is used so no need to change anything.

djc (approved):
Looks good, thanks!

PR #485, opened 2021-02-10
Remove unwrap in buf_contains_message

ab-cgh:


briansmith (approved):

rustls/src/msgs/hsjoiner.rs
```diff
@@ -78,8 +78,8 @@ impl HandshakeJoiner {
         self.buf.len() >= HEADER_SIZE
             && self.buf.len()
                 >= (codec::u24::decode(&self.buf[1..4])
-                    .unwrap()
-                    .0 as usize)
+                    .map(|s| s.0 as usize))
+                    .unwrap_or(0)
```
djc, line 82: It's not obvious to me that `unwrap_or(0)` is desirable here, can someone enlighten me?

ab-cgh: This will still fail the comparison, since `0+HEADER_SIZE`  will be less then `self.buf.len()`. To get to this addition, we already verified that what we are looking for must be larger then a header from the comparison on the left of the `&&`, `self.buf.len() >= HEADER_SIZE`.

briansmith: I think this needs const generics to allow `&buf [1..4]` to have type `&[u8; 3]`; in the case we could change `u24::decode` to take a `&[u8; 3]` and make it infallible. Until then I guess maybe we should just defer this.

briansmith: If you really want to be conservative and avoid any functions or operators that panic, here's one way: PR #492. It uses similar ideas to `untrusted`; using `untrusted` would also work.

briansmith:
I think this should be closed because the suggested change is wrong and PR #498 exists now.

briansmith (changes requested):
[Clearing my original approval.]

ab-cgh:
That sounds good to me, closing.

PR #487, opened 2021-02-10
Remove unwraps from tls13 handle_client_hello

ab-cgh:
This one involved a bit of binding refactoring. I think it matches the intent better though, since [this comment](https://github.com/ctz/rustls/blob/main/rustls/src/server/tls13.rs#L590) mentions "key share" and not "group".

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -575,46 +575,39 @@ impl CompleteClientHelloHandling {
             return Err(hs::illegal_param(sess, "client sent duplicate keyshares"));
         }
 
-        let share_groups: Vec<NamedGroup> = shares_ext
-            .iter()
-            .map(|share| share.group)
-            .collect();
-
         let supported_groups = suites::KeyExchange::supported_groups();
-        let chosen_group = supported_groups
+
+        let chosen_share = supported_groups
             .iter()
-            .find(|group| share_groups.contains(group))
-            .cloned();
+            .find_map(|group| shares_ext.iter().find(|share| share.group == *group));
 
-        if chosen_group.is_none() {
-            // We don't have a suitable key share.  Choose a suitable group and
-            // send a HelloRetryRequest.
-            let retry_group_maybe = supported_groups
-                .iter()
-                .find(|group| groups_ext.contains(group))
-                .cloned();
-            self.handshake
-                .transcript
-                .add_message(chm);
+        let chosen_share =  match chosen_share {
+            Some(s) => s,
+            None => {
+                // We don't have a suitable key share.  Choose a suitable group and
+                // send a HelloRetryRequest.
+                let retry_group_maybe = supported_groups
+                    .iter()
+                    .find(|group| groups_ext.contains(group))
+                    .cloned();
 
-            if let Some(group) = retry_group_maybe {
-                if self.done_retry {
-                    return Err(hs::illegal_param(sess, "did not follow retry request"));
+                self.handshake
+                    .transcript
+                    .add_message(chm);
+
+                if let Some(group) = retry_group_maybe {
+                    if self.done_retry {
+                        return Err(hs::illegal_param(sess, "did not follow retry request"));
+                    }
+
+                    self.emit_hello_retry_request(sess, group);
+                    self.emit_fake_ccs(sess);
+                    return Ok(self.into_expect_retried_client_hello());
                 }
 
-                self.emit_hello_retry_request(sess, group);
-                self.emit_fake_ccs(sess);
-                return Ok(self.into_expect_retried_client_hello());
+                return Err(hs::incompatible(sess, "no kx group overlap with client"));
             }
-
-            return Err(hs::incompatible(sess, "no kx group overlap with client"));
-        }
-
-        let chosen_group = chosen_group.unwrap();
-        let chosen_share = shares_ext
-            .iter()
-            .find(|share| share.group == chosen_group)
-            .unwrap();
+        };
 
         let mut chosen_psk_index = None;
         let mut resumedata = None;


rustls/src/server/tls13.rs
```diff
@@ -581,40 +581,39 @@ impl CompleteClientHelloHandling {
...
-            let retry_group_maybe = supported_groups
+        let chosen_share =  match chosen_share {
+            Some(s) => s,
+            None => {
+                // We don't have a suitable key share.  Choose a suitable group and
+                // send a HelloRetryRequest.
+                let retry_group_maybe = supported_groups
                 .iter()
                 .find(|group| groups_ext.contains(group))
                 .cloned();
```
djc, line 598: I think these will now need an extra bit of indentation.

rustls/src/server/tls13.rs
```diff
@@ -581,40 +581,39 @@ impl CompleteClientHelloHandling {
...
 
         let supported_groups = suites::KeyExchange::supported_groups();
-        let chosen_group = supported_groups
+
+        let chosen_share = supported_groups
             .iter()
-            .find(|group| share_groups.contains(group))
-            .cloned();
+            .find(|group| share_groups.contains(group)) // Find the chosen group
+            .and_then(|g| shares_ext.iter().find(|k| k.group == *g)); // Then its share
```
djc, line 588: I'd prefer using `group` over `g` and `share` over `k` -- this code is dense enough as it is...

djc (commented):
Some style nits here. There's actually also a behavioral change where we're now treating more types of incompatibility with the same error handling. For that, I'd like Brian or Joe to have a look as well.

rustls/src/server/tls13.rs
```diff
@@ -581,40 +581,39 @@ impl CompleteClientHelloHandling {
...
 
-        if chosen_group.is_none() {
-            // We don't have a suitable key share.  Choose a suitable group and
-            // send a HelloRetryRequest.
-            let retry_group_maybe = supported_groups
+        let chosen_share =  match chosen_share {
```
briansmith, line 590: `.unwrap_or_else`?

briansmith: Sorry, never mind that comment.

rustls/src/server/tls13.rs
ab-cgh: Names changed, good point.

rustls/src/server/tls13.rs
ab-cgh: Good catch, thanks. Addressed.

ab-cgh:
I didn't see the behavior change originally. I had thought any group, if it existed, would also have a share associated with it. If it didn't have one, it would of hit the `.unwrap()` later down. AFAIK this change keeps the original behavior of if no group exists, then it takes the route of sending a `HelloRetryRequest`.

```
                            Original:
try to get group ------------------------------------------------------------------
                                                                                  v 
    Modified:                    | --> None --> skip group->share search -------> |                                  | --> None --> send HelloRetryRequest --> ...
try to get group --> iterator ----                                                | --> match on groups existence -->|          Original:
                                 | --> Some --> search for share in group too --> |                                  | --> Some --> unwrap on group->share search
                                                                                                                                Modified:
                                                                                                                            Some --> no unwrap needed
```

briansmith (commented):
PTAL at a slightly different alternative: PR #489.

ab-cgh:
I like the implementation in #489 as well. I'm happy to close this PR in favor if it if desired. I have no strong feelings either way.

rustls/src/server/tls13.rs
```diff
@@ -581,40 +581,39 @@ impl CompleteClientHelloHandling {
...
-            // We don't have a suitable key share.  Choose a suitable group and
-            // send a HelloRetryRequest.
-            let retry_group_maybe = supported_groups
-                .iter()
-                .find(|group| groups_ext.contains(group))
-                .cloned();
-            self.handshake
-                .transcript
-                .add_message(chm);
+            .find(|group| share_groups.contains(group)) // Find the chosen group
```
briansmith, line 587: I think you can use `find_map(|group| shares_ext.iter().find(|share| share.group == *group)` here to avoid doing two lookups.

rustls/src/server/tls13.rs
```diff
@@ -581,40 +581,39 @@ impl CompleteClientHelloHandling {
...
+                self.handshake
+                    .transcript
+                    .add_message(chm);
 
-                self.emit_hello_retry_request(sess, group);
-                self.emit_fake_ccs(sess);
-                return Ok(self.into_expect_retried_client_hello());
-            }
+                if let Some(group) = retry_group_maybe {
+                    if self.done_retry {
```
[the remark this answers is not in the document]
briansmith: > It would be clearer to check self.done_retry error out before we try to calculate retry_group_maybe.

We should probably do that reorganization to make it clearer, but my suggestion is out of scope for this PR.

briansmith:
I close my other PR because I noticed it is subtly wrong, and because it was trying to do things that are out of scope of this. I think the `find_map` suggestion above is the only thing I would change.

briansmith:
> I didn't see the behavior change originally. I had thought any group, if it existed, would also have a share associated with it.

I'm not understanding. The `share_groups` collection is derived from the `share_exts` collection so it is indeed true that every item in `shared_groups` exists in `share_exts`. If you take my `find_map` suggestion then I think you can eliminate `share_groups`.

rustls/src/server/tls13.rs
ab-cgh: Good idea, and change applied. This worked nicely.

ab-cgh:
Your suggestion did get rid of `share_groups`, but that explanation was an attempt at showing that the final logic did not change with this PR from what it originally was due to removing `chosen_group`.

briansmith:
LGTM. I think this is much clearer than the original code.

djc:
Yup, this looks good, thanks!

PR #489, opened 2021-02-10
Simplify TLS 1.3 handle_client_hello.

briansmith:
See the individual commit messages for details. It might be useful to review them commit-by-commit.

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -583,34 +583,33 @@ impl CompleteClientHelloHandling {
         let supported_groups = suites::KeyExchange::supported_groups();
         let chosen_group = supported_groups
             .iter()
-            .find(|group| share_groups.contains(group))
-            .cloned();
+            .find(|group| share_groups.contains(group));
 
-        if chosen_group.is_none() {
-            // We don't have a suitable key share.  Choose a suitable group and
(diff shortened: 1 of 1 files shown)


rustls/src/server/tls13.rs
```diff
@@ -575,46 +575,35 @@ impl CompleteClientHelloHandling {
...
-            self.handshake
-                .transcript
-                .add_message(chm);
+            .find(|share| supported_groups.contains(&share.group));
 
-            if let Some(group) = retry_group_maybe {
+        let chosen_share = match chosen_share {
+            Some(share) => share,
+            None => {
                 if self.done_retry {
```
briansmith, line 586: I loved the check for `self.done_retry` earlier to avoid wasted work in the error case.

djc: Previously this check was conditional on `retry_group_maybe` being `Some`. Is it correct to make it no longer conditional on that check? Does the error message still make sense in that case? Or should this check for `retry_group_maybe.is_some() && self.done_retry`?

briansmith (commented):
I'm going to close this in favor of the other PR's implementation since this implementation subtly and unintentionally changes which share we prefer. I will post some improvements on top of #487.

PR #492, opened 2021-02-10
Clarify that message header decoding doesn't panic.

briansmith:
Add checking to `u24::decode()` to avoid array indexing errors.
Add similar checks to HandshakeJoiner::buf_contains_message.

Note that all of these checks are redundant but they allow us to avoid
slice indexing and unwrap in favor of type-safe access. When Rust has
generic consts we will probably be able to simplify this a lot.


djc:
I did yet another take on this in #498, please have a look.

briansmith:
I'm going to close this as I think that we need a more general solution to clarify this kind of code throughout the codebase. This isn't as urgent for my needs as clarifying the correctness of the state machine.

PR #493, opened 2021-02-10
Refactor client hello retrying to avoid unwrapping

ab-cgh:
Refactor `ExpectServerHelloOrHelloRetryRequest::handle_hello_retry_request` to avoid unwrapping


djc:
Thanks!

PR #497, opened 2021-02-11
Refactor `ExtensionProcessing::process_common` to not unwrap

ab-cgh:


rustls/src/server/hs.rs
```diff
@@ -265,6 +253,26 @@ impl ExtensionProcessing {
...
+            if can_use {
+                self.send_sct = true;
+
+                if !sess.common.is_tls13() {
+                    self.exts.push(ServerExtension::make_sct(list));
+                    return;
+                }
+            }
+            // Restore list if we were not allowed to use it
+            server_key.sct_list = Some(list);
```
djc, line 272: I'm not sure what the separate function is buying us here (just the early return?) and I'd like it to be more obvious that a taken-but-unused sct list is always put back. Would you like to do another take?

ab-cgh: I moved it into a function so I could use the early return in combination with the fallback list reset. I tried keeping it inline but I was not able to find a clean way to set the `send_sct` flag with the same rules as before without either cloning `server_key.sct_list` or having multiple places the list would need put back.

Is the list generally large enough that cloning it would be expensive? That would be cleaner then whats presented here.

briansmith: I would like to see an attempt that avoids factoring out the function. Having the separate function makes it harder to understand the code at a higher level because we have to search for callers to that function to see that there is only this one.

djc: I'm pretty sure there's a way to do this without cloning.

ab-cgh: PTAL at the new attempt. The new function has been removed and its now more clear when the taken list is put back.

briansmith:
@ab-cgh While reviewing your PR I had another idea; PTAL at #505.

ab-cgh:
@briansmith I took a look over it and it looks good to me. As it also gets rid of two flags being passed through the entire state machine, I think that I would prefer it be merged over this PR.

rustls/src/server/hs.rs
```diff
@@ -243,18 +243,17 @@ impl ExtensionProcessing {
...
+            if can_use_sct {
+                if let Some(list) = server_key.take_sct_list() {
+                    self.send_sct = true;
+
+                    if sess.common.is_tls13() {
+                        // Restore the SCT list if we can't use it for a ServerExtension.
+                        server_key.sct_list = Some(list);
+                    } else {
+                        self.exts.push(ServerExtension::make_sct(list));
+                    }
```
djc, line 256: Optional nit: I prefer writing these as a match expression these days, but feel free to go either way on this.

rustls/src/server/hs.rs
```diff
@@ -243,18 +243,17 @@ impl ExtensionProcessing {
...
-                && server_key.has_sct_list()
-            {
-                self.send_sct = true;
-
-                if !sess.common.is_tls13() {
-                    let sct_list = server_key.take_sct_list().unwrap();
-                    self.exts
-                        .push(ServerExtension::make_sct(sct_list));
+            let can_use_sct = !for_resume && hello.find_extension(ExtensionType::SCT).is_some();
+            if can_use_sct {
```
djc, line 247: Now we can just inline as `if !for_resume && hello.find_extension(ExtensionType::SCT).is_some() {`, right?

djc (commented):
This is much better than the previous version! Some style nits left:

briansmith (changes requested):
I think we should go with #505 and close this one, based on the above discussion.

ab-cgh:
👍 Closing this in favor of #505

PR #499, opened 2021-02-11
Refactor negotiated_version usage to avoid unwrapping

ab-cgh:
This PR reworks the handling of the negotiated version. Inspiration was taken from https://github.com/ctz/rustls/pull/460, so instead of unwrapping an expected value it is now passed through the state machine.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -455,11 +455,13 @@ impl ExpectServerHello {
 
     fn into_expect_tls12_new_ticket_resume(
         self,
+        negotiated_version: ProtocolVersion,
         secrets: SessionSecrets,
         certv: verify::ServerCertVerified,
         sigv: verify::HandshakeSignatureValid,
     ) -> NextState {
         Box::new(tls12::ExpectNewTicket {
+            negotiated_version,
             secrets,
             handshake: self.handshake,
             resuming: true,
@@ -470,11 +472,13 @@ impl ExpectServerHello {
 
     fn into_expect_tls12_ccs_resume(
         self,
+        negotiated_version: ProtocolVersion,
         secrets: SessionSecrets,
         certv: verify::ServerCertVerified,
         sigv: verify::HandshakeSignatureValid,
     ) -> NextState {
         Box::new(tls12::ExpectCCS {
+            negotiated_version,
             secrets,
             handshake: self.handshake,
             ticket: ReceivedTicketDetails::new(),
@@ -484,8 +488,9 @@ impl ExpectServerHello {
         })
     }
 
-    fn into_expect_tls12_certificate(self) -> NextState {
+    fn into_expect_tls12_certificate(self, negotiated_version: ProtocolVersion) -> NextState {
         Box::new(tls12::ExpectCertificate {
+            negotiated_version,
             handshake: self.handshake,
             server_cert: self.server_cert,
             may_send_cert_status: self.may_send_cert_status,
@@ -511,10 +516,10 @@ impl State for ExpectServerHello {
             server_hello.legacy_version
         };
 
-        match server_version {
+        let version = match server_version {
             TLSv1_3 if tls13_supported => {
-                sess.common.negotiated_version = Some(TLSv1_3);
-            }
+                TLSv1_3
+            },
             TLSv1_2 if sess.config.supports_version(TLSv1_2) => {
                 if sess.early_data.is_enabled() && sess.common.early_traffic {
                     // The client must fail with a dedicated error code if the server
@@ -523,7 +528,6 @@ impl State for ExpectServerHello {
                         "server chose v1.2 when offering 0-rtt".to_string(),
                     ));
                 }
-                sess.common.negotiated_version = Some(TLSv1_2);
 
                 if server_hello
                     .get_supported_versions()
@@ -534,6 +538,8 @@ impl State for ExpectServerHello {
                         "server chose v1.2 using v1.3 extension",
                     ));
                 }
+
+                TLSv1_2
             }
             _ => {
                 sess.common
@@ -568,6 +574,8 @@ impl State for ExpectServerHello {
             ));
         }
 
+        sess.common.negotiated_version = Some(version);
+
         // Extract ALPN protocol
         if !sess.common.is_tls13() {
             process_alpn_protocol(sess, server_hello.get_alpn_protocol())?;
@@ -598,7 +606,6 @@ impl State for ExpectServerHello {
             return Err(illegal_param(sess, "server varied selected ciphersuite"));
         }
 
-        let version = sess.common.negotiated_version.unwrap();
         if !sess
             .common
             .get_suite_assert()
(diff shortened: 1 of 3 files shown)


rustls/src/session.rs
```diff
@@ -433,7 +433,7 @@ pub enum MiddleboxCCS {
 }
 
 pub struct SessionCommon {
-    pub negotiated_version: Option<ProtocolVersion>,
+    pub negotiated_version: ProtocolVersion,
```
briansmith, line 436: The term "negotiated version" refers to the version that was agreed upon. At the time where `SessionCommon` is constructed, we don't know the negotiated version, so the old code was correct so that we can return the right value `None` from `get_protocol_version()`.

In the cases where we are unwrapping this, we should instead thread the known (non-Option) negotiated version through the state machine, similar to https://github.com/ctz/rustls/pull/460.

rustls/src/client/hs.rs
```diff
@@ -511,8 +511,8 @@ impl State for ExpectServerHello {
 
         match server_version {
             TLSv1_3 if tls13_supported => {
-                sess.common.negotiated_version = Some(TLSv1_3);
-            }
+                sess.common.negotiated_version = TLSv1_3
```
briansmith, line 514: Here we should continue to set `sess.common.negotiated_version = Some(negotiated_version)` because this is what is reported by `get_protocol_version()`.

However, also from this point, we should keep passing the negotiated version through to the states that need it in each `into_expecting()` function simlar to https://github.com/ctz/rustls/pull/460.

briansmith (changes requested):

rustls/src/session.rs
ab-cgh: Thank you for the reference! The new changes follow this approach much closer.

rustls/src/client/hs.rs
ab-cgh: I refactored the setting of `sess.common.negotiated_version` as part of my latest changes, but it is still explicitly set.

ab-cgh:
@briansmith PTAL at the latest changes.

djc:
I think the commit message needs updating to cover the new strategy?

djc:
(Sorry, inadvertently closed this.)

ab-cgh:
My bad. The commit and PR have been updated to reflect what it looks like now.

briansmith:
> This PR reworks the handling of the negotiated version. Inspiration was taken from #460, so instead of unwrapping an expected value it is now passed through the state machine.

It would be clearer to remove `HandshakeDetails::version` as part of this. Also, we should prioritize merging #502 over merging this, and then rebase this on top of master after this is merged, because #502 fixes real bugs.

rustls/src/client/hs.rs
```diff
@@ -484,8 +488,9 @@ impl ExpectServerHello {
         })
     }
 
-    fn into_expect_tls12_certificate(self) -> NextState {
+    fn into_expect_tls12_certificate(self, negotiated_version: ProtocolVersion,) -> NextState {
```
djc, line 491: Style nit: let's ditch the trailing comma here?

djc (commented):
Almost there!

rustls/src/server/hs.rs
```diff
@@ -202,7 +202,7 @@ impl ExtensionProcessing {
                 if let Some(resume) = resumedata {
                     if sess.config.max_early_data_size > 0
                         && hello.early_data_extension_offered()
-                        && resume.version == sess.common.negotiated_version.unwrap()
+                        && Some(resume.version) == sess.common.negotiated_version
```
briansmith, line 205: Can we split the server side into a separate PR?

I think we shouldn't be using `sess.common.negotiated_version` here. Instead, should pass a non-`Option` copy of the negotiated version so that it is clear that version negotiation definitely happened before this point; i.e. `sess.common.negotiated` version is never `None` here.

(I think the `unwrap()` was actually better than this, because that made it clearer.)

briansmith (changes requested):

briansmith:
I think it would be good for @ctz to look at this after the server side is split off into a separate PR, to make sure we have agreement on the approach.

rustls/src/client/hs.rs
ab-cgh: Removed.

rustls/src/server/hs.rs
ab-cgh: Reverted. I will look at a better server side implementation elsewhere.

rustls/src/client/hs.rs
```diff
@@ -598,7 +610,6 @@ impl State for ExpectServerHello {
             return Err(illegal_param(sess, "server varied selected ciphersuite"));
         }
 
-        let version = sess.common.negotiated_version.unwrap();
```
briansmith, line 601: Why not do this here:
```
sess.common.negotiated_version = Some(version);
```
Instead of twice above?

rustls/src/client/tls12.rs
```diff
@@ -822,7 +845,7 @@ impl hs::State for ExpectFinished {
         // Hash this message too.
         st.handshake.transcript.add_message(&m);
 
-        save_session(&st.secrets, &mut st.handshake, &mut st.ticket, sess);
+        save_session(&st.secrets, &mut st.handshake, &mut st.ticket, sess, st.negotiated_version);
```
briansmith, line 848: I know it looks messy to thread `negotiated_version` through to this point just for saving the session, but I think in the near future we'll improve it so that what is currently in `st.ticket`, `st.handshake.session_id`, `st.negotiated_version`, and maybe more will all be held in a single `SessionDataToSave` structure that gets threaded through. So, I'm happy to see this incremental step done.

briansmith (changes requested):

rustls/src/client/hs.rs
ab-cgh: I thought I had this in a previous commit revision but it appears to have been lost somehow. These have now been refactored.

briansmith (approved):
LGTM! I'm excited about building the last(?) set of cleanup PRs on top of this next week.

rustls/src/client/hs.rs
```diff
@@ -455,11 +455,13 @@ impl ExpectServerHello {
 
     fn into_expect_tls12_new_ticket_resume(
         self,
+        negotiated_version: ProtocolVersion,
```
ctz, line 458: By this point we're absolutely certain we're talking TLS1.2, therefore there's no need to keep that information separately here, or below.

rustls/src/client/tls12.rs
```diff
@@ -748,9 +771,8 @@ fn save_session(
 
     let scs = sess.common.get_suite_assert();
     let master_secret = secrets.get_master_secret();
-    let version = sess.get_protocol_version().unwrap();
     let mut value = persist::ClientSessionValue::new(
-        version,
+        negotiated_version,
```
ctz, line 775: This can be statically replaced with `ProtocolVersion::TLSv1_2`.

ctz (commented):
Unless I am mistaken, I think the threading through of the protocol version through the TLS1.2 state machine is unnecessary?

briansmith:
> Unless I am mistaken, I think the threading through of the protocol version through the TLS1.2 state machine is unnecessary?

You are right! If we make the changes you suggest, we can avoid threading the field through the state machine. That will simplify this PR considerably.

briansmith:
@ab-cgh I rebased this on main to account for my changes breaking yours and simplified it like @ctz asked for in #516.

ab-cgh:
@briansmith Thank you. I hadn't had a chance to get back to this yet to implement the requested refactor.

djc:
Now that #516 is merged, I think we can close this?

ab-cgh:
Yep, no problem.

PR #500, opened 2021-02-11
Codec: Add bounds checking to decode functions.

briansmith:
The callers probably already ensured the bounds are correct, but check
ourselves.


briansmith:
@djc I think this improves upon the relevant part of each of #498 and #492. I'll rebase #492 on top of this.

djc (approved):
Nice, this is definitely better than my previous attempt.

PR #501, opened 2021-02-11
Codec: Simplify integer encoding functions & clarify assumptions made.

briansmith:
It isn't clear whether the `pub_*` functions are intended to accept an output
buffer larger than needed, but they definitely don't work when the output
buffer is smaller than needed. Clarify this.

Instead of implementing big-endian encoding manually, use the standard library
implementations.

Removing the use of `try_into()` in these functions is deferred until later.
This is just making the panicking explicit instead of implicit.


djc:
LGTM.

PR #502, opened 2021-02-12
Client: Randomize resumption session ID and ensure we use same session for retries.

### LINKED ISSUE #470: RFC 8446 Section D.4 fake session ID compatibility hack is missing during resumption
In PR #464, @dhobsd pointed out that Rustls doesn't implement the RFC 8446 Section D.4 fake session ID workaround:

> The client always provides a non-empty session ID in the ClientHello, as described in the legacy_session_id section of Section 4.1.2.

I am not sure how important it is to implement this workaround. Maybe there are no more compatibility issues? I would prefer to not implement it unless somebody shares compelling and recent data indicating the the compatibility issue still exists.

/cc @davidben

briansmith:
Calculate the value of `ClientHello.legacy_session_id` only in
`emit_initial_client_hello` so that we are sure it will never change.
Ensure that we generate a random value for it whenever we can legally do so.

Only look up the session in the session cache during the initial client hello.
During retries, use the same (lack of) session as was used in the initial client
hello, instead of searching the session cache again. Previously doing a cache
lookup during retries caused us to fail to use the same session when the session
cache contents changed between the initial and retry hello; the retry was racing
against modifications to the session cache.

There may be other problems where our retry client hellos do not match our
initial client hellos as required. Further work should be done to refactor the
code to move move logic from `emit_client_hello_for_retry` into
`emit_initial_client_hello` and/or `handle_hello_retry_request`.

This fixes the TLS13SessionID-TLS13 Bogo test, so enable it.

Fixes #470 and the race originally reported in PR ##464.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -146,6 +146,18 @@ impl InitialState {
                 .set_client_auth_enabled();
         }
         let hello_details = ClientHelloDetails::new();
+        self.handshake.resuming_session = find_session(sess, self.handshake.dns_name.as_ref());
+        if let Some(resuming) = &mut self.handshake.resuming_session {
+            if resuming.version == ProtocolVersion::TLSv1_2 {
+                random_sessionid_for_ticket(resuming);
+            }
+            self.handshake.session_id = resuming.session_id;
+        }
+        // https://tools.ietf.org/html/rfc8446#appendix-D.4
+        // https://tools.ietf.org/html/draft-ietf-quic-tls-34#ref-TLS13
+        if self.handshake.session_id.is_empty() && !sess.common.is_quic() {
+            self.handshake.session_id = random_sessionid();
+        }
         emit_client_hello_for_retry(sess, self.handshake, hello_details, None, self.extra_exts)
     }
 }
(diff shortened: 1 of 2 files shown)


dhobsd (approved):
This is a much cleaner implementation of what I was going for in #464.

rustls/src/client/hs.rs
```diff
@@ -146,6 +146,18 @@ impl InitialState {
...
+            if resuming.version == ProtocolVersion::TLSv1_2 {
+                random_sessionid_for_ticket(resuming);
+            }
+            self.handshake.session_id = resuming.session_id;
+        }
+        // https://tools.ietf.org/html/rfc8446#appendix-D.4
+        // https://tools.ietf.org/html/draft-ietf-quic-tls-34#ref-TLS13
+        if self.handshake.session_id.is_empty() && !sess.common.is_quic() {
+            self.handshake.session_id = random_sessionid();
+        }
```
djc, line 160: This is a logic change because this is now always executed, because it was previously only done in the `resuming_session.is_some()` case. If that is intentional, I would prefer to have that change in a separate commit or PR with a separate justification.

briansmith: > This is a logic change because this is now always executed, because it was previously only done in the `resuming_session.is_some()` case. If that is intentional, I would prefer to have that change in a separate commit or PR with a separate justification.

The justification is in the linked part of the TLS 1.3 spec: "The client always provides a non-empty session ID in the ClientHello, as described in the legacy_session_id section of Section 4.1.2."

It might be possible to separate out the change where we only search the session cache one instead of twice from the change that initializes the session ID, but I do think it makes sense to keep all the session resumption logic (session ID setting and session lookup) together in one PR.

djc: I'm fine with keeping it in one PR. I was thinking the same code could just be in the `else` branch of the `if let`, and then move it in another commit to separate the refactoring and functional change.

briansmith: > I'm fine with keeping it in one PR. I was thinking the same code could just be in the else branch of the if let, and then move it in another commit to separate the refactoring and functional change.

I understand where your coming from. However, I'd like to avoid the intermediate step of wondering "is this refactoring equivalent to the previous logic?" especially because the old logic was wrong, and instead just forget about the previous logic and concentrate on "Is this new logic correct?"

briansmith: @davidben Besides checking the length of the `legacy_session_id` is not 0, it seems like Bogo should also verify that when doing TLS 1.2 ticket-based resumption, the session_id in the resumption client hello is different from the session_id in the previous client hello, presumably randomized. WDYT?

djc: I feel like I'm unlikely to persuade you, but I consistently find separating functional changes from refactoring changes to make reviews easier. Getting the refactoring out of the way first means I can focus on the functional change rather than keeping track of both in my head at the same time.

briansmith: OK, I'll close this in favor of #506 and then a follow-up that does the session ID changes.

PR #503, opened 2021-02-12
Client: Thread hash_at_client_recvd_server_hello through the state machine.

briansmith:
Clarify the exact lifetime of this value, removing the `unwraps()` currently in the code.

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -37,6 +37,7 @@ use crate::client::hs;
 
 use ring::constant_time;
 use webpki;
+use ring::digest::Digest;
 
 // Extensions we expect in plaintext in the ServerHello.
 static ALLOWED_PLAINTEXT_EXTS: &[ExtensionType] = &[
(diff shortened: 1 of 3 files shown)


rustls/src/client/tls13.rs
```diff
@@ -241,7 +242,6 @@ pub fn start_handshake_traffic(
     hs::check_aligned_handshake(sess)?;
 
     let hash_at_client_recvd_server_hello = handshake.transcript.get_current_hash();
-    handshake.hash_at_client_recvd_server_hello = Some(hash_at_client_recvd_server_hello);
 
     let _maybe_write_key = if !sess.early_data.is_enabled() {
```
briansmith, line 246: In this branch I had previously tried to do some experiments to construct this write_key once as `handshake_write_key` and then thread `handshake_write_key` through instead. However, `key_schedule.client_handshake_traffic_secret(` is stateful and the state gets out of sync with `sess.common.record_layer` in some cases if I do that without making other changes, so I reverted all of that and just left this branch as threading the hash.

PR #504, opened 2021-02-12
Prevent panic while panicking in tests/api.rs

thomwiggers:
While playing around with rustls I broke the `client_cert_resolve` test and if the `ResolvesClientCert` implementation for `ClientCheckCertResolve` panics, then it will panic again in the `drop` call leading to an illegal instruction error.


djc:
Seems reasonable, thanks!

PR #505, opened 2021-02-12
Simplify the handling of SCT and OCSP sending in the server.

briansmith:
If we're not going to send one of these, then drop it as early as
possible to reduce memory overhead, so that at the time of sending,
`CertifiedKey` will only contain it if it is to be sent. This makes
the boolean flags that previously controlled this redundant, so
remove them.

In the process, remove some unwrapping.


briansmith:
I updated this PR from the initial version after #497 was closed.

`CertifiedKey::has_sct_list` is now unused within Rustls, but I didn't remove it because it is is part of the public API.

PR #506, opened 2021-02-12
Client: Lookup session once per connection.

briansmith:
Only do the lookup of the session during the initial client hello.
Otherwise, if we do a session lookup during retries, the session cache
may have changed and we might get back a different session, e.g. because
a different connection modified the session cache in the interim.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -136,6 +136,12 @@ impl InitialState {
     }
 
     fn emit_initial_client_hello(mut self, sess: &mut ClientSessionImpl) -> NextState {
+        // During retries "the client MUST send the same ClientHello without
+        // modification" with only a few exceptions as noted in
+        // https://tools.ietf.org/html/rfc8446#section-4.1.2,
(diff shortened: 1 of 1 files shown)


rustls/src/client/hs.rs
```diff
@@ -136,6 +136,12 @@ impl InitialState {
     }
 
     fn emit_initial_client_hello(mut self, sess: &mut ClientSessionImpl) -> NextState {
+        // During retries :the client MUST send the same ClientHello without
+        // modification" with only a few exceptions as noted in
+        // https://tools.ietf.org/html/rfc8446#section-4.1.2,
+        // Calculate all inputs to the client hellos that might otherwise
+        // change between the initial and retry hellos here to enforce this.
```
briansmith, line 143: If/when this is merged, I will file a follow-up issue to track the remaining work of moving logic from `emit_client_hello_for_retry` to this function.

ctz (approved):
#507 should unlock the ability to defend this with a test.

Issue #477, opened 2021-02-08
X25519 is hard-coded as the default client key share algorithm


briansmith:
I think the intent is that the choice of cipher suites is controlled by the user of Rustls. Then the user should be able to control which key exchange algorithms are enabled. Then they might not enable X25519, in which case even attempting to use X25519 would be wrong.

This is how the code looks in `choose_kx_groups`:
```
    let group = retryreq
        .and_then(HelloRetryRequest::get_requested_key_share_group)
        .or_else(|| find_kx_hint(sess, handshake.dns_name.as_ref()))
        .unwrap_or_else(|| NamedGroup::X25519);
```

Perhaps instead we should choose the first key exchange algorithm in the list of enabled key exchange algorithms, and document that this ordering matters and is used for this purpose?

briansmith:
We should have a test where Rustls is configured to not have X25519 enabled to verify that this works.

ctz:
This was fixed in #507

PR #507, opened 2021-02-13
Allow selecting supported groups in config

thomwiggers:
Attempts to implement #477. I could probably use some more pointers on the tests.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -764,7 +764,7 @@ impl ExpectServerHelloOrHelloRetryRequest {
         check_aligned_handshake(sess)?;
 
         let cookie = hrr.get_cookie();
-        let req_group = hrr.get_requested_key_share_group();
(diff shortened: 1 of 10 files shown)


ctz:
Thanks for your work on this! I want to suggest we make the public API a little different, so that in the future a program which configures only one group can have the code for the others discarded a link-time.

That'd mean, I think, that the public API looks like `pub supported_kx_groups: Vec<&'static SupportedKxGroup>`, and then `SupportedKxGroup` is a type similar to `SupportedCipherSuite`:

```
pub struct SupportedKxGroup {
  pub name: NamedCurve,
  agreement_algorithm: &'static ring::agreement::Algorithm,
}
```

With the library exporting all the available values (values of this type shouldn't be instantiable outside the library).

This means `named_group_to_ecdh_alg` in `suites.rs` can die. This doesn't get us all the way to achieving the above goal, but I have some changes on a branch which would get the rest of the way: add a builder-type api for making a ServerConfig/ClientConfig.

ctz:
Merged with the noted changes as #532

Issue #509, opened 2021-02-14
Cipher suite ordering: AES vs Chacha20


Matthias247:
After some testing with [quinn](https://github.com/quinn-rs/quinn) we noticed that rustls most highly preferrenced cipher suite `TLS13_CHACHA20_POLY1305_SHA256` is quite a bit slower than `TLS13_AES_128_GCM_SHA256` and `TLS13_AES_256_GCM_SHA384` on typical desktop and server hardware due to hardware acceleration for AES.

Where a quic loopback transfer with `TLS13_CHACHA20_POLY1305_SHA256` maxed out at 350MB/s, it reached 520MB/s using `TLS13_AES_128_GCM_SHA256` (+50%), and about 500MB/s using `TLS13_AES_256_GCM_SHA384`.

Can the cipher suite list in [ALL_CIPHERSUITES](https://github.com/ctz/rustls/blob/9c546cf7f6366af75c60110554b55922f1aab300/rustls/src/suites.rs#L440-L452) be reordered to prefer AES cipher suites? 

I feel like there will be quite a lot of people running this with default config, and might not detect that they are not reaching peak performance and efficiency.

PR #510, opened 2021-02-14
Adjust default cipher suite order:

ctz:
- prefer aes to not-aes (see #509)
- prefer ecdsa to rsa (in fact this doesn't make much of a difference,
  unless your peer has both kinds of cert).
- prefer aes-256 to aes-128

Add a rationale section to the manual for this.


Matthias247:
Thanks for addressing this so quickly!

PR #511, opened 2021-02-15
Remove CertifiedKey::has_sct_list()

djc:
This became unused in #505. It was originally added 4 years ago in
df361f832bd466f1da263d65bf0a6df64254d3a5, so it seems there was no
specific reason for making it public. In addition, we already have
some breaking API changes lined up, so might as well remove this.


ctz (approved):

ctz:
lgtm!

PR #513, opened 2021-02-16
Client: Move session ID initialization to `emit_initial_client_hello`.

briansmith:
The session ID is not allowed to change between the initial client hello and
a retry client hello.


djc:
LGTM.

PR #515, opened 2021-02-16
Reduce usage of `SessionCommon::get_suite_assert()`.

briansmith:
`SessionCommon::get_suite_assert()` unwraps an `Option` containing the agreed-upon cipher suite. In many cases within the handshake state machine, it is easy to retrieve this value without unwrapping, so do so. This is an incremental step towards removing all uses of the function.

See each individual commit message for a description of each commit.


djc:
Nice!

PR #518, opened 2021-02-17
Client: Clarify when server cert details are filled in and fully available.

briansmith:
Clarify when the full `ServerCertDetails` and related inputs are available *with their correct and complete values*. Now the point at which `ServerCertDetails` is constructed is the earliest point in which the validation could happen.

Drop the `ServerCertDetails` immediately after certificate validation to free up memory. The TLS 1.3 state machine was holding onto the `ServerCertDetails` for many messages longer than needed.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -30,7 +30,7 @@ use crate::ticketer;
 use crate::verify;
 
 use crate::client::common::{ClientHelloDetails, ReceivedTicketDetails};
-use crate::client::common::{HandshakeDetails, ServerCertDetails};
+use crate::client::common::HandshakeDetails;
 use crate::client::{tls12, tls13};
 
 use webpki;
@@ -187,9 +187,6 @@ struct ExpectServerHello {
     handshake: HandshakeDetails,
     early_key_schedule: Option<KeyScheduleEarly>,
     hello: ClientHelloDetails,
-    server_cert: ServerCertDetails,
-    may_send_cert_status: bool,
-    must_issue_new_ticket: bool,
 }
 
 struct ExpectServerHelloOrHelloRetryRequest {
@@ -410,9 +407,6 @@ fn emit_client_hello_for_retry(
         handshake,
         hello,
         early_key_schedule,
-        server_cert: ServerCertDetails::new(),
-        may_send_cert_status: false,
-        must_issue_new_ticket: false,
     };
 
     if support_tls13 && retryreq.is_none() {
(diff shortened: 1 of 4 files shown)


rustls/src/client/hs.rs
```diff
@@ -187,9 +187,6 @@ struct ExpectServerHello {
     handshake: HandshakeDetails,
     early_key_schedule: Option<KeyScheduleEarly>,
     hello: ClientHelloDetails,
-    server_cert: ServerCertDetails,
-    may_send_cert_status: bool,
-    must_issue_new_ticket: bool,
```
briansmith, line 192: Obviously, at the time we're waiting for `ServerHello`, these fields have not been filled in.

rustls/src/client/hs.rs
```diff
@@ -741,15 +742,15 @@ impl State for ExpectServerHello {
...
+                return if must_issue_new_ticket {
                     Ok(self.into_expect_tls12_new_ticket_resume(secrets, certv, sigv))
                 } else {
                     Ok(self.into_expect_tls12_ccs_resume(secrets, certv, sigv))
                 };
             }
         }
 
-        Ok(self.into_expect_tls12_certificate(scs))
+        Ok(self.into_expect_tls12_certificate(scs, may_send_cert_status, must_issue_new_ticket, server_cert_list_list))
```
briansmith, line 753: The danger here is that `may_send_cert_status` and `must_issue_new_ticket` are both `bool`. We're relying on the state machine tests to make sure we don't accidentally transpose them in this call.

rustls/src/client/tls12.rs
```diff
@@ -60,35 +64,40 @@ impl hs::State for ExpectCertificate {
...
     ) -> hs::NextStateOrError {
-        let cert_chain =
+        let server_cert_chain =
             require_handshake_msg!(m, HandshakeType::Certificate, HandshakePayload::Certificate)?;
         self.handshake
             .transcript
             .add_message(&m);
 
-        self.server_cert.cert_chain = cert_chain.clone();
+        // TODO(perf): Avoid this clone of a large object.
```
briansmith, line 73: This is now issue #519.

rustls/src/client/tls13.rs
```diff
@@ -390,7 +390,6 @@ fn validate_encrypted_extensions(
 pub struct ExpectEncryptedExtensions {
     pub handshake: HandshakeDetails,
     pub key_schedule: KeyScheduleHandshake,
-    pub server_cert: ServerCertDetails,
```
briansmith, line 393: `server_cert` is empty at this point so there's no point in threading it through yet.

Issue #466, opened 2021-02-04
Implement RFC 8446 Appendix C.4 in the default session cache.


dhobsd:
I implemented an RFC8446 [appendix C.4]-capable session cache on the current `StoresClientSession` trait, but it doesn't quite feel right. I wanted to open this issue present the challenges I faced and the implementation I ended up with. I feel that the challenges presented may warrant some API changes -- even if implemented internally -- as what I ended up with isn't particularly satisfying (while also being -- I think -- relatively good given the constraints).

RFC8446 [section 4.6.1] recommends that TLS 1.3 servers send multiple session resumption tickets to clients. In [appendix C.4], it's subsequently recommended that clients use tickets at most once to avoid session tracking. The current implementation of `ClientSessionMemoryCache` does not do this, and some properties of `StoresClientSession` (and use of the cache in general) make doing so difficult.

`ClientSessionMemoryCache` iterates keys of a `std::collections::HashMap` to evict. This process executes in an arbitrary order, and it's possible to evict the just-inserted key. This results in effectively a 50% chance of reusing the same ticket, even when a new ticket was provided. The cache only stores a single ticket, while servers may send multiple ([section 4.6.1] mentions several additional reasons clients might want to have extra tickets available). Finally, failed negotiation attempts on a session reuse the last-known ticket.

The `StoresClientSession` trait is intentionally opaque about what's stored in it. Unfortunately, this doesn't allow the handler of the session cache determine whether the supplied `ClientSessionValue` is for TLS 1.3, or for RFC5077 session resumption in a previous version. This wouldn't be so bad, but the same cache is overloaded to also store TLS 1.3 key exchange hints. 

Problematically, the keyspace for TLS 1.3 tickets and RFC5077 sessions are shared, so a `StoresClientSession` implementation must first decode the supplied value as a `ClientSessionValue` and peek at the version to see whether it should be managed per C.4 or not. Although the keyspace isn't shared with key exchange hints, we still have to re-parse the value as a `NamedGroup` when it fails to parse as a `ClientSessionValue` because there's no way to gather information about what sort of `ClientSessionKey` we get to differentiate the keyspace at all.

[...]

PR #523, opened 2021-02-18
Deduplicate server ticket emitter code

djc:


rustls/src/server/tls13.rs
```diff
@@ -880,23 +880,37 @@ impl ExpectFinished {
...
         handshake: &mut HandshakeDetails,
         sess: &mut ServerSessionImpl,
         key_schedule: &KeyScheduleTraffic,
     ) {
         let nonce = rand::random_vec(32);
         let plain = get_server_session_value(handshake, key_schedule, sess, &nonce).get_encoding();
-        let ticket = match sess.config.ticketer.encrypt(&plain) {
-            Some(t) => t,
-            None => return,
+
```
briansmith, line 890: I think it would be a bit clearer if we just wrote:

```
let stateless = sess.config.ticketer.enabled();
if stateless {
...
```

rustls/src/server/tls13.rs
```diff
@@ -880,23 +880,37 @@ impl ExpectFinished {
...
+        let (ticket, lifetime, stateful) = if sess.config.ticketer.enabled() {
+            let ticket = match sess.config.ticketer.encrypt(&plain) {
+                Some(t) => t,
+                None => return,
```
briansmith, line 894: We should have a comment about why it is OK to return early here without forwarding the error.

rustls/src/server/tls13.rs
```diff
@@ -880,23 +880,37 @@ impl ExpectFinished {
...
+            };
+            (ticket, sess.config.ticketer.get_lifetime(), false)
+        } else {
```
briansmith, line 897: I wonder if we should keep supporting stateful tickets in TLS 1.3. Maybe TLS 1.3 should be stateless only? Would implementing the server side of #466 push us in the direction of stateless only? After all, we don't want to store 10x the state on the server to be able to give the client 10 sessions.

rustls/src/server/tls13.rs
djc: Fair point, fixed.

rustls/src/server/tls13.rs
djc: That seems orthogonal to these changes. Can you suggest some wording?

rustls/src/server/tls13.rs
djc: That seems out of scope for this PR, which is just about deduplicating some code. File a separate issue?

briansmith: If we drop the stateful resumption then we'd just do that instead of reviewing this PR, right? That's why I brought it up now. It seems like we're likely to be refactoring this to just immediately(?) throw it away.

djc: I suppose so. On the other hand, while there has been little discussion about dropping stateful ticket support, it still seems worthwhile to me to reduce the amount of duplicate code, potentially making future refactoring easier.

rustls/src/server/tls13.rs
briansmith: We can do it in another PR.

rustls/src/server/tls13.rs
briansmith: It does seem like dropping the stateful session support would not be clearly a good idea anyway so I agree we should just review this.

rustls/src/server/tls13.rs
```diff
@@ -880,23 +880,38 @@ impl ExpectFinished {
...
+            if !stored {
+                trace!("resumption not available; not issuing ticket");
+                return;
+            }
+            let stateful_lifetime = 24 * 60 * 60; // this is a bit of a punt
+            (id, stateful_lifetime)
         };
-        let ticket_lifetime = sess.config.ticketer.get_lifetime();
 
-        let age_add = rand::random_u32(); // nb, we don't do 0-RTT data, so whatever
```
briansmith, line 896: I don't understand the "nb, we don't do 0-RTT data, so whatever" comment that was removed.

briansmith (approved):
I went ahead and reviewed this. It was a bit difficult to review so I did it by trying to write the PR from scratch, and then verifying that what I did and what you did is the same (only trivial differences), with no functional changes. So LGTM.

It would be good for @ctz to make sure there is no significance to the "nb, we don't do 0-RTT data, so whatever"

rustls/src/server/tls13.rs
ctz: This comment is out of date: it was written before rustls-in-quic-mode did 0-RTT data.

Issue #480, opened 2021-02-09
SCTs should be passed to the cert verifier so the cert verifier can enforce CT


briansmith:
Consider the case where one is using a platform-provided certificate verifier, or another custom certificate verifier, where the certificate verifier has its own CT enforcement. Rustls doesn't pass the SCTs from the handshake to the certificate verifier so CT doesn't work in this scenario.

Further, when the cert verifier has support for SCTs embedded into X.509 certificates, the built-in CT support may cause verification to fail when the cert verifier's CT verification would succeed.

In general, CT verification cannot be separated from certificate verification.

Proposal: Move all the CT support from the handshake logic to the default implementations of the certificate verification traits. Remove the `ct_logs` client configuration option; instead the user will need to configure CT in the certificate verifier implementation it chooses.

This is related to #479 but somewhat orthogonal.

ctz:
Pretty sure this is addressed now.

PR #526, opened 2021-02-19
Refactor `ClientConfig` verifier configuration to better support custom certificate verifiers

briansmith:
When an alternate certificate verifier is used, Rustls old CT log and root
configuration was likely irrelevant to it. In particular, a system
certificate verifier will likely have its own settings set at the system
and/or user level. Move those fields out of `ClientConfig` to ensure they
are not used in that situation, and to avoid users wrongly expecting the
custom verifier to use these settings.

Change the API for constructing `ClientConfig` so that one can still
configure the list of CT logs and roots for the default verifier only.

To avoid starting an explosion in the number of `ClientConfig`
constructors, require the user to pass in the desired cipher suites to
each constructor. Document that the new `DEFAULT_CIPHERSUITES` should be
used to select the default values.

Hide the number of cipher suites in `ALL_CIPHERSUITES` from the type by
making it a slice, so that adding a cipher suite to `ALL_CIPHERSUITES` is
no longer an API breaking change.

Pass the SCTs received from the server to the verifier during verification.

Do not do any SCT verification in the handshake, beyond what the
verifier is now expected to do.

Fixes issue #480.

### DIFF
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -12,7 +12,6 @@ use crate::key::Certificate;
 use crate::log::{debug, trace, warn};
 use crate::msgs::enums::SignatureScheme;
 use crate::msgs::handshake::DigitallySignedStruct;
-use crate::msgs::handshake::SCTList;
 use ring::digest::Digest;
 
 type SignatureAlgorithms = &'static [&'static webpki::SignatureAlgorithm];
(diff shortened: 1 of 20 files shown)


briansmith:
I suggest we sort out the `ClientConfig` and then do the corresponding `ServerConfig` changes in a separate PR, which I would commit to providing in the future.

On top of this change, we should then change rustls-native-certs to pass the SCTs to the system verifier.

rustls/src/verify.rs
```diff
@@ -80,18 +79,20 @@ impl ClientCertVerified {
...
+    ///
+    /// `scts` contains the Signed Certificate Timestamps (SCTs) the server
+    /// sent with the certificate, if any.
     fn verify_server_cert(
         &self,
         end_entity: &Certificate,
         intermediates: &[Certificate],
-        roots: &RootCertStore,
         dns_name: webpki::DNSNameRef,
+        scts: &mut dyn Iterator<Item=&[u8]>,
```
briansmith, line 95: Using an iterator type here allows us to avoid allocating memory for the SCTs. In the future we can probably avoid even more allocation.

ctz:
This has been merged. Thanks!

PR #529, opened 2021-02-19
Store `SupportedCipherSuite` in `ClientSessionValue`.

briansmith:
Do the mapping to `SupportedCipherSuite` at the time `ClientSessionValue` is
constructed so that we don't have to look it up all the time. This eliminates
some unwrapping.

Simplify the implementation of `find_cipher_suite` to match what is done during
`ClientSessionValue` decoding.

### DIFF
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -433,13 +433,7 @@ impl ClientSessionImpl {
     }
 
     pub fn find_cipher_suite(&self, suite: CipherSuite) -> Option<&'static SupportedCipherSuite> {
-        for scs in &self.config.ciphersuites {
-            if scs.suite == suite {
-                return Some(scs);
-            }
-        }
-
-        None
+        self.config.ciphersuites.iter().copied().find(|&scs| scs.suite == suite)
     }
 
     pub fn wants_read(&self) -> bool {
--- a/rustls/src/msgs/persist.rs
+++ b/rustls/src/msgs/persist.rs
@@ -8,6 +8,7 @@ use webpki;
 
 use std::cmp;
 use std::mem;
+use crate::SupportedCipherSuite;
 
 // These are the keys and values we store in session storage.
 
@@ -53,7 +54,7 @@ impl ClientSessionKey {
 #[derive(Debug)]
 pub struct ClientSessionValue {
     pub version: ProtocolVersion,
-    pub cipher_suite: CipherSuite,
+    pub suite: &'static SupportedCipherSuite,
     pub session_id: SessionID,
     pub ticket: PayloadU16,
     pub master_secret: PayloadU8,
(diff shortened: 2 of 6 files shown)


rustls/src/msgs/persist.rs
```diff
@@ -80,9 +81,11 @@ impl Codec for ClientSessionValue {
         self.server_cert_chain.encode(bytes);
     }
 
-    fn read(r: &mut Reader) -> Option<ClientSessionValue> {
+    pub fn read(r: &mut Reader, enabled_cipher_suites: &[&'static SupportedCipherSuite]) -> Option<ClientSessionValue> {
```
briansmith, line 84: The prototype of `Code::read` is too restrictive. We should eventually split `Codec` into two traits: one for reading and one for encoding.

rustls/src/msgs/persist.rs
```diff
@@ -107,22 +110,28 @@ impl Codec for ClientSessionValue {
             server_cert_chain,
         })
     }
+
+    pub fn get_encoding(&self) -> Vec<u8> {
+        let mut result = Vec::new();
+        self.encode(&mut result);
+        result
+    }
```
djc, line 118: This appears to be unused?

rustls/src/client/mod.rs
```diff
@@ -433,13 +433,7 @@ impl ClientSessionImpl {
...
 
     pub fn find_cipher_suite(&self, suite: CipherSuite) -> Option<&'static SupportedCipherSuite> {
-        for scs in &self.config.ciphersuites {
-            if scs.suite == suite {
-                return Some(scs);
-            }
-        }
-
-        None
+        self.config.ciphersuites.iter().copied().find(|&scs| scs.suite == suite)
```
djc, line 436: This seems unrelated to the other changes. If the goal is to make this the same as some other code, maybe they could share an actual implementation?

rustls/src/msgs/persist.rs
briansmith: It is used, which is why I added it. It has a default implementation in the `Codec` trait.

rustls/src/client/mod.rs
briansmith: It wouldn't be right to add a dependency on `ClientSessionImpl` to the `ClientSessionValue`.

I don't think it's strictly unrelated because it helps us understand that the logic is the same. You are right that this change isn't strictly needed, so I could revert this change, if you'd prefer.

rustls/src/msgs/persist.rs
briansmith: If you have access to the codecov report, you should be able to see it is actually executed too.

djc: I'm very confused. Your custom implementation appears to be identical to the default implementation -- so why is it useful to add one here?

briansmith: > I'm very confused. Your custom implementation appears to be identical to the default implementation -- so why is it useful to add one here?

This type no longer implements the `Codec` trait because the `Codec` trait's `read` method doesn't allow passing in additional arguments. Since it no longer implements `Codec`, it can no longer use the default implementation of `Codec::get_encoding()`.

djc: Ah, sorry -- the change from `impl Codec for` to `impl ` was a bit subtle.

djc (approved):

PR #530, opened 2021-02-19
Client: Remove `sent_tls13_fake_ccs` from `HandshakeDetails`.

briansmith:
It is only used by the client, so it shouldn't be in `HandshakeDetails`, which
is shared between the client and the server. Further, it is only used for the
initial messages, so there's no need to store it for the whole handshake.

This refactoring enables further refactorings to clarify the state machine. In
particular, this allows us to clear up the unwrapping logic related to
`fill_in_psk_binder`, but this is also useful for other refactorings, including
the ECH implementation.

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -58,7 +58,6 @@ pub struct HandshakeDetails {
     pub randoms: SessionRandoms,
     pub using_ems: bool,
     pub session_id: SessionID,
-    pub sent_tls13_fake_ccs: bool,
     pub dns_name: webpki::DNSName,
 }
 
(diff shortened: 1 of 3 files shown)


djc (approved):

briansmith:
@djc please take another look at this, since I updated it. It's important to initialize `sent_tls13_fake_ccs` in `emit_initial_client_hello` and the test suite checks this.

ctz:
Thanks! Regarding the commit message "It is only used by the client, so it shouldn't be in `HandshakeDetails`" -- note that `client::common::HandshakeDetails` is a client-specific struct -- but it isn't specific to TLS1.3 which I think might be what you meant.

PR #531, opened 2021-02-21
Client: Implement slightly better eviction logic for session cache.

briansmith:
Never evict an unrelated entry from the cache when replacing an entry
with the same key.

Reduce the number of lock acquisitions per insertion from 2 to 1.

Improve the type of the capacity of the cache to guarantee avoidance of
underflows.


briansmith:
Closing this. I just posted this to share with @dhobsd.

PR #533, opened 2021-02-22
Client: Track the exact lifetime of the client/server randoms.

briansmith:
Thread the client and server randoms through the state machine explicitly
so that we can exactly track when each value is known.

### DIFF
--- a/rustls/src/session.rs
+++ b/rustls/src/session.rs
@@ -22,6 +22,7 @@ use std::io::{Read, Write};
 use std::collections::VecDeque;
 use std::io;
 use ring::digest::Digest;
+use crate::msgs::handshake::Random;
 
 /// Generalises `ClientSession` and `ServerSession`
 pub trait Session: quic::QuicExt + Read + Write + Send + Sync {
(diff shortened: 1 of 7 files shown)


djc:
I think this duplicates what I currently have in #443?

rustls/src/session.rs
```diff
@@ -231,9 +232,16 @@ pub struct SessionRandoms {
     pub server: [u8; 32],
 }
 
-static TLS12_DOWNGRADE_SENTINEL: &[u8] = &[0x44, 0x4f, 0x57, 0x4e, 0x47, 0x52, 0x44, 0x01];
+pub static TLS12_DOWNGRADE_SENTINEL: &[u8] = &[0x44, 0x4f, 0x57, 0x4e, 0x47, 0x52, 0x44, 0x01];
 
 impl SessionRandoms {
+    pub fn from_client_and_server(we_are_client: bool, client_random: &Random, server_random: &Random) -> Self {
```
djc, line 238: It looks like the only caller always passes `true` for `we_are_client`. Maybe restrict the API to make this more obvious?

Should the `SessionRandoms::client` and `server` fields just be `Random` rather than untyped arrays? It feelsincongruent to me that you pass in a `&Random` here only to deref it back into an array whereas elsewhere you have `SessionSecrets::for_resume()` start to require the caller to clone.

djc (commented):
Also would still prefer (for reviewing purposes) the state threading part to be in a separate commit from the other changes made here.

briansmith (commented):
> I think this duplicates what I currently have in #443?

Probably. it seems like #443 is stalled indefinitely.

rustls/src/session.rs
briansmith: > It looks like the only caller always passes `true` for `we_are_client`. Maybe restrict the API to make this more obvious?

I don't think that will continue to be the case if/when somebody does the server side of this.
 
> Should the `SessionRandoms::client` and `server` fields just be `Random` rather than untyped arrays? It feels incongruent to me that you pass in a `&Random` here only to deref it back into an array whereas elsewhere you have `SessionSecrets::for_resume()` start to require the caller to clone.

I agree, but that was an issue before as well.

If you'd like to make some PRs that refactor the random stuff, I am happy to base this PR on top of them, no matter how disruptive they are to this PR.

djc:
@briansmith would you like me to rebase this for you?

briansmith:
> @briansmith would you like me to rebase this for you?

I don't think it's worth rebasing at this time. Let's get the other changes done first.

PR #536, opened 2021-02-24
Refactor into_expect_tls12_certificate_verify to avoid unwrapping

ab-cgh:
Pass the certificate through explicitly instead of assuming a `.is_some()` call.

### DIFF
--- a/rustls/src/server/tls12.rs
+++ b/rustls/src/server/tls12.rs
@@ -102,11 +102,11 @@ pub struct ExpectClientKX {
 }
 
 impl ExpectClientKX {
-    fn into_expect_tls12_certificate_verify(self, secrets: SessionSecrets) -> hs::NextState {
+    fn into_expect_tls12_certificate_verify(self, secrets: SessionSecrets, client_cert: ClientCertDetails) -> hs::NextState {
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice!

rustls/src/server/tls12.rs
```diff
@@ -173,8 +173,8 @@ impl hs::State for ExpectClientKX {
         sess.common
             .start_encryption_tls12(&secrets);
 
-        if self.client_cert.is_some() {
-            Ok(self.into_expect_tls12_certificate_verify(secrets))
+        if let Some(client_cert) = self.client_cert.take() {
+            Ok(self.into_expect_tls12_certificate_verify(secrets, client_cert))
```
briansmith, line 177: This PR, and others, shows the downside of the `self.into_expect_...` pattern: We have to use `take()` to move a part of `self` out of `self` can call `self.into_expect_`. This means `self` only contains part of its full state in these `into_expect_...` functions, which is hard to reason about. It also means several fields in the `Expect` structures have to be `Option` for otherwise no reason.

In the (near) future we should inline all the self.into_expect_` functions into their callers so that we can avoid these bad patterns.

PR #537, opened 2021-02-24
Remove unwrap in unpad_tls13

ab-cgh:
Return an unknown content type instead..

### DIFF
--- a/rustls/src/cipher.rs
+++ b/rustls/src/cipher.rs
@@ -332,7 +332,8 @@ fn unpad_tls13(v: &mut Vec<u8>) -> ContentType {
         match v.pop() {
             Some(0) => {}
 
-            Some(content_type) => return ContentType::read_bytes(&[content_type]).unwrap(),
+            Some(content_type) => return ContentType::read_bytes(&[content_type])
+                .unwrap_or(ContentType::Unknown(0)),
 
             None => return ContentType::Unknown(0),
         }


rustls/src/cipher.rs
```diff
@@ -332,7 +332,8 @@ fn unpad_tls13(v: &mut Vec<u8>) -> ContentType {
         match v.pop() {
             Some(0) => {}
 
-            Some(content_type) => return ContentType::read_bytes(&[content_type]).unwrap(),
+            Some(content_type) => return ContentType::read_bytes(&[content_type])
+                .unwrap_or(ContentType::Unknown(0)),
```
briansmith, line 336: The caller is:
```
    fn decrypt(&self, mut msg: Message, seq: u64) -> Result<Message, TLSError> {
        ....
        let content_type = unpad_tls13(&mut buf);	        let content_type = unpad_tls13(&mut buf);
        if content_type == ContentType::Unknown(0) {
```

However, `msg.typ` is already known to be the content type of the message, IIUC. Thus it seems like this code isn't making sense as we could just pass `msg.typ` as an input into `unpad_tls13` and avoid this re-parsing and checking.

briansmith: I forgot, in TLS 1.3 we have a dummy content type and the real content type, so the above thinking may be wrong, depending on whether `Message.typ` is the inner or outer content type.

Assuming that we actually do need to do this checking, we should have a integration test that shows how a malicious peer could trigger this panic. This would also warrant a new release of Rustls.

briansmith: OK, I see now that the `unwrap()` is unreachable, so there's no bug here; it's just unclear. This is another case where the `Codec` interface has the wrong types. I don't think the answer is to use `unwrap_or` since that will be dead code. I think instead we should avoid the use of the `Codec` interface for `ContentType` here.

ab-cgh: Good catch. I didn't notice where this was coming from entirely. I will look at trying to remove `Codec` entirely here if possible.

ctz: I suggest that we add, in msgs::enums (and associated macros) a `from_u8(value: u8) -> ContentType`, ensure that is used in `read()` over there, and then we can use it here.

ctz:
Thanks to recent changes in enums.rs, this is now trivially `ContentType::from(u8)`

PR #538, opened 2021-02-24
Refactor ExpectClientKX to inline methods to avoid unwrapping state

ab-cgh:
Based on the idea presented by @briansmith [in this comment](https://github.com/ctz/rustls/pull/536#discussion_r581680900). This lets `self` be taken apart by value instead of relying on `Option::take`.

### DIFF
--- a/rustls/src/server/common.rs
+++ b/rustls/src/server/common.rs
@@ -30,16 +30,12 @@ impl HandshakeDetails {
 }
 
 pub struct ServerKXDetails {
-    pub kx: Option<kx::KeyExchange>,
+    pub kx: kx::KeyExchange,
 }
 
 impl ServerKXDetails {
     pub fn new(kx: kx::KeyExchange) -> ServerKXDetails {
-        ServerKXDetails { kx: Some(kx) }
-    }
-
-    pub fn take_kx(&mut self) -> kx::KeyExchange {
-        self.kx.take().unwrap()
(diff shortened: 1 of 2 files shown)


rustls/src/server/common.rs
```diff
@@ -30,16 +30,16 @@ impl HandshakeDetails {
...
 
 impl ServerKXDetails {
     pub fn new(kx: kx::KeyExchange) -> ServerKXDetails {
-        ServerKXDetails { kx: Some(kx) }
+        ServerKXDetails { kx }
     }
 
-    pub fn take_kx(&mut self) -> kx::KeyExchange {
-        self.kx.take().unwrap()
+    pub fn take_kx(self) -> kx::KeyExchange {
```
briansmith, line 41: Do we still need this function? Field `kx` is already pub so I think the callers of this function can just move it out of the structure. One of the goals of my idea is to remove the need for these `take_*` functions.

I.e., instead of `let kx = self.server_kx.take_kx()`, we can just write `let kx = self.server_kx.kx` or even maybe remove the local `kx` variable and use `self.server_kx` directly where it is used.

ab-cgh: The second idea sounds good. I had kept this in spirit of more minimal changes, but I've removed the method in favor of a direct field access now.

djc:
LGTM!

Issue #508, opened 2021-02-13
ECHConfig implementation


sayrer:
In #318, there's a test client that uses trust-dns to pull this data from DNS and then hand over the bytes for parsing by rustls. I think that's the right way to do things, so there's no new dependency introduced.

If there are no objections, I'd like to start work on this aspect of the ECH problem. I'll also add a command-line tool and API to generate ECHConfig records.

ECHConfig is defined here: https://tools.ietf.org/id/draft-ietf-tls-esni-09.html#name-encrypted-clienthello-confi

From the TLS WG list:

"Cloudflare is now running an ECH test server here:
https://crypto.cloudflare.com

We're running draft-ietf-tls-esni-09. The HTTPS resource record containing the current ECH config is available in DNS."

sayrer:
Here is a small Python program that will print the DNS HTTPS-RR data for Cloudflare's expermental -09 server.

```python3
#!/usr/bin/env python3                                                                                                                                                                                         

import dns.resolver

resolver = dns.resolver.Resolver(configure=False)
resolver.nameservers = ['1.1.1.1']

answer = resolver.resolve('crypto.cloudflare.com.', 'HTTPS')

print(answer.rrset)
```

It should print something like:

```
./testdns.py 
crypto.cloudflare.com. 265 IN HTTPS 1 . alpn="h2" ipv4hint="162.159.135.79,162.159.136.79" echconfig="AEf+CQBDABNjbG91ZGZsYXJlLWVzbmkuY29tACCD91Ovu3frIsjhFKo0I1fPd/a09nzKMrjC9GZV3NvrfQAgAAQAAQABAAAAAA==" ipv6hint="2606:4700:7::a29f:874f,2606:4700:7::a29f:884f"
```

The ECHConfig is encoded as base64.

djc:
Sounds good to me, but obviously I can't really speak for Joe.

sayrer:
An update here: Cloudflare has updated their server to draft-10, which looks good in most respects. The WG is still thrashing on ECH and HRR interactions, so I don't expect this feature will ship until summer at the earliest. I will update my PR for draft-10, and should be able to get a lot farther now that I am more familiar with the handshake code (thanks for all the reviews, even the negative ones).

cjpatton:
Just for everyone's awareness: Cloudflare's test endpoint now implements draft-ietf-tls-esni-14 and has been on this draft for some time.

sayrer:
[...]

PR #539, opened 2021-02-24
WIP: ECH Draft-09 serialization code

sayrer:
In the interest of sharing early and often, here's code that serializes and deserializes ECH draft-09 ECHConfig records. The test record is from a live server at crypto.cloudflare.com. Part of addressing #508.

The next step is to add a little demo Trust-DNS client that pulls this record and parses it. I'll add the SVCB/HTTPS-RR code to Trust-DNS if need be.

After that, I think I'll start prototyping with the rust-hpke crate, since it has the older HPKE code needed for draft-09. I don't really think ring should bother with it, and instead just focus on the newest HPKE draft.

cc @djc @briansmith @ctz

### DIFF
--- a/rustls/src/msgs/ech_test.rs
+++ b/rustls/src/msgs/ech_test.rs
@@ -1,0 +1,16 @@
+use base64;
+use crate::msgs::codec::{Codec, Reader};
+use crate::msgs::handshake::ECHConfigs;
+
+#[test]
+fn test_echconfig_serialization() {
+    // An ECHConfig record from Cloudflare for
+    let base64_esni = "AEf+CQBDABNjbG91ZGZsYXJlLWVzbmkuY29tACCD91Ovu3frIsjhFKo0I1fPd/a09nzKMrjC9GZV3NvrfQAgAAQAAQABAAAAAA==";
+    let bytes = base64::decode(&base64_esni).unwrap();
+    let records = ECHConfigs::read(&mut Reader::init(&bytes)).unwrap();
+    let name = String::from_utf8(records[0].contents.public_name.clone().into_inner()).unwrap();
+    assert_eq!("cloudflare-esni.com", name.as_str());
+    let mut output = Vec::new();
+    records.encode(&mut output);
+    assert_eq!(base64_esni, base64::encode(&output));
+}
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -80,6 +80,10 @@ impl PayloadU16 {
         (slice.len() as u16).encode(bytes);
         bytes.extend_from_slice(slice);
     }
+
+    pub fn into_inner(self) -> Vec<u8> {
+        self.0
+    }
 }
 
 impl Codec for PayloadU16 {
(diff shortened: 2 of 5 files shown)


rustls/src/msgs/ech_test.rs
```diff
@@ -0,0 +1,16 @@
+use base64;
+use crate::msgs::codec::{Codec, Reader};
+use crate::msgs::handshake::ECHConfigs;
+
+#[test]
+fn test_echconfig_serialization() {
+    // An ECHConfig record from Cloudflare for
+    let base64_esni = "AEf+CQBDABNjbG91ZGZsYXJlLWVzbmkuY29tACCD91Ovu3frIsjhFKo0I1fPd/a09nzKMrjC9GZV3NvrfQAgAAQAAQABAAAAAA==";
```
djc: Should we replace the `esni` with `ech` here?

sayrer: I think you're reading an old diff. Should be "let base64_echconfig = ..." now.

sayrer:
I'm happy to add some tests to keep the handshake.rs coverage up, but the CI job seems to be failing in bogo/

sayrer:
Looks like the min-version checks are failing due to code that rust-hpke is pulling in, so I'll ignore that for now.

Going to refactor HPKEKeyPair a bit to carry its algorithm with it.

sayrer:
Looks like there will be a renamed struct and DNS field, based on discussion and patches in

https://github.com/MikeBishop/dns-alt-svc/issues/298

and

https://github.com/tlswg/draft-ietf-tls-esni/pull/391

sayrer:
I don't really like Cloudflare's ECHKey serialization, but I'll keep the code around for now to facilitate testing.

sayrer:
When it came time to write the HPKE tests, I found myself wondering what data would be available. So I had to hack in a demonstration of the API @ctz suggested here: https://github.com/ctz/rustls/pull/318#issuecomment-547629147

I think it will be a little nicer once HelloData (or whatever we call it) is threaded through the handshake code. extra_exts isn't easy to use with ClientSession::new right now, and ECH adds a few more variations of that problem.

sayrer:
Oops, that broke quic. I will use `cargo test --all-features` from now on.

djc:
Probably useful to extract that part into a separate PR?

sayrer:
> Probably useful to extract that part into a separate PR?

I agree. I just needed to figure it out before I did the rest of the HPKE stuff. I think your other PRs on the handshake code will change it too.

sayrer:
I've gotten this to the point that I'm digging through the various Expect* structs after any HRR. At that point, only the dns_name is needed, so I'm looking for the boundary where I can stop using "HelloData" and just pass the dns_name on. I *think* that same boundary will be where ECH ends too.

I'll do a PR for just the handshake changes, which can simply combine dns_name + extra_exts into a struct initially, after #547 lands. Then, a new EncryptedHost struct can be added later. It will be useful through HRR as well.

sayrer:
Closing in favor of #663.

PR #544, opened 2021-03-04
Support stable QUIC transport parameters handshake extension

Ralith:
Preserves the prior extension for backwards-compatibility on the wire,
requiring an API tweak for the application to dictate which version to
be compatible with.

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -11,6 +11,7 @@ use sct;
 use webpki;
 
 use rustls::internal::msgs::enums::ProtocolVersion;
+use rustls::quic;
 use rustls::quic::ClientQuicExt;
 use rustls::quic::ServerQuicExt;
 use rustls::ClientHello;
@@ -1041,6 +1042,7 @@ fn main() {
             } else {
                 rustls::ServerSession::new_quic(
                     scfg.as_ref().unwrap(),
+                    quic::Version::V1,
                     opts.quic_transport_params.clone(),
                 )
             };
@@ -1052,6 +1054,7 @@ fn main() {
             } else {
                 rustls::ClientSession::new_quic(
                     ccfg.as_ref().unwrap(),
+                    quic::Version::V1,
                     dns_name,
                     opts.quic_transport_params.clone(),
                 )
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -332,6 +332,7 @@ pub trait ClientQuicExt {
     /// TLS-encoded transport parameters to send.
     fn new_quic(
         config: &Arc<ClientConfig>,
+        quic_version: Version,
         hostname: webpki::DNSNameRef,
         params: Vec<u8>,
     ) -> ClientSession {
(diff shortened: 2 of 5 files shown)


Ralith:
CI failures look unrelated.

rustls/src/quic.rs
```diff
@@ -342,12 +343,14 @@ pub trait ClientQuicExt {
                 .all(|x| x.get_u16() >= ProtocolVersion::TLSv1_3.get_u16()),
             "QUIC requires TLS version >= 1.3"
         );
+        let ext = if (0xff00_0000..=0xff00_0020).contains(&quic_version) {
```
djc, line 346: Do we have a sense of what QUIC v2 drafts might use? Should we just use `> 0xff00_0000` to try and future-proof it a little?

rustls/examples/internal/bogo_shim.rs
```diff
@@ -1041,6 +1041,7 @@ fn main() {
             } else {
                 rustls::ServerSession::new_quic(
                     scfg.as_ref().unwrap(),
+                    1,
```
djc, line 1044: The simplicity of the `1` here seems a little bit misleading. I wonder if we should make this a newtype `QuicVersion` with a `v1` method and maybe `v1_draft(u8)` methods, or something?

rustls/src/quic.rs
Ralith: Future draft versions (for e.g. QUIC 2) might use yet another TLS extension number. I think we should stick to known cases.

rustls/examples/internal/bogo_shim.rs
Ralith: How about an enum? It's not like we can promise to do the correct thing for future wire-incompatible versions regardless.

djc: Yeah, enum sounds good to me. Probably make it non-exhaustive then?

rustls/src/quic.rs
```diff
@@ -371,15 +373,26 @@ pub trait ServerQuicExt {
...
         ServerSession { imp }
     }
 }
 
 impl ServerQuicExt for ServerSession {}
 
+#[non_exhaustive]
+pub enum Version {
+    /// Draft versions prior to V1
+    Draft1,
```
djc, line 391: Sorry, final nit: `Draft1` reads a bit too much like draft-1, maybe make it `V1Draft` or something similar?

Ralith: Oh, yeah, that's way better!

rustls/src/quic.rs
```diff
@@ -332,6 +332,7 @@ pub trait ClientQuicExt {
     /// TLS-encoded transport parameters to send.
     fn new_quic(
         config: &Arc<ClientConfig>,
+        quic_version: u32,
```
djc, line 335: Missed one here, I think?

Ralith: Whoops. Flaky CI is a pain...

djc: Yeah, I'm hoping to work on that soon.

PR #545, opened 2021-03-07
Document state machines and simplify client authentication flow

djc:
Tests fixed, scaled back the ambition a bit.

I tried to simplify the flow of the client authentication logic, but it proved a bit harder than expected. Still, I think this makes the invariants a little bit clearer than before.


djc:
Let's scratch this in favor of #546.

PR #546, opened 2021-03-08
Transcript

djc:
* Documents state transitions
* Inline creation of next states everywhere
* Remove unwraps in `HandshakeHash` by splitting it into two types

(Reviewing per-commit is recommended.)

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -4,7 +4,7 @@ use crate::check::check_message;
 use crate::{cipher, SupportedCipherSuite};
 use crate::client::ClientSessionImpl;
 use crate::error::TLSError;
-use crate::hash_hs::HandshakeHash;
+use crate::hash_hs::HandshakeHashBuffer;
 use crate::key_schedule::KeyScheduleEarly;
 #[cfg(feature = "logging")]
 use crate::log::{debug, trace};
@@ -167,7 +167,7 @@ fn random_sessionid() -> SessionID {
 
 struct InitialState {
     handshake: HandshakeDetails,
-    transcript: HandshakeHash,
+    transcript_buffer: HandshakeHashBuffer,
     extra_exts: Vec<ClientExtension>,
 }
 
@@ -175,7 +175,7 @@ impl InitialState {
     fn new(host_name: webpki::DNSName, extra_exts: Vec<ClientExtension>) -> InitialState {
         InitialState {
             handshake: HandshakeDetails::new(host_name),
-            transcript: HandshakeHash::new(),
+            transcript_buffer: HandshakeHashBuffer::new(),
             extra_exts,
         }
     }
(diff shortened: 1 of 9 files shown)


rustls/src/client/hs.rs
```diff
@@ -38,6 +38,55 @@ use ring::digest::Digest;
...
+/// tls13::ExpectCertificateOrCertReq ->        tls13::ExpectCertificate
+///                                             | tls13::ExpectCertificateRequest
+///
+/// tls13::ExpectCertificateRequest ->          tls13::ExpectCertificate
+///
+/// tls13::ExpectCertificate ->                 tls13::ExpectCertificateVerify
+///
+/// tls13::ExpectCertificateVerify ->           tls13::ExpectFinished
+///
+/// tls13::ExpectFinished ->                    tls13::ExpectTraffic
```
briansmith, line 89: I don't think adding this comment is a good idea. This is just duplicating what's in the spec, it takes work to verify that the comment matches the spec, and it would take work to maintain the comment when the code changes (chances are very high the comment will get out of sync with the code). Instead we should just try to make the code match the diagrams in the spec more directly.

rustls/src/server/hs.rs
```diff
@@ -34,6 +34,34 @@ use crate::server::{tls12, tls13};
...
+/// tls13::CompleteClientHelloHandling ->   ExpectClientHello
+///                                         | tls13::ExpectCertificate
+///                                         | tls13::ExpectFinished
+///
+/// tls13::ExpectCertificate ->             tls13::ExpectFinished
+///                                         | tls13::CertificateVerify
+///
+/// tls13::CertificateVerify ->             tls13::ExpectFinished
+///
+/// tls13::ExpectFinished ->                tls13::ExpectTraffic
```
briansmith, line 65: Ditto.

briansmith:
There are too many changes in this PR for me to review it well. I think 

> Add some documentation about handshake state machines

Just drop this commit. The change is counterproductive for the reasons I mentioned above.

> Inline creation of next states

I agree we should do this. However, I disagree with doing it as one giant commit in one giant PR. I would much prefer one commit/PR per `into_` function removal. The reason is that often removing even one `into_` function allow us to also replace `Option<T>` with `T` and/or remove some `take_`-like functions.

> Eliminate unwrapping in HandshakeHash
> Clarify availability of client auth state in HandshakeHash

These should be their own PR.

I didn't look at the rest.

briansmith:
One additional request for refactoring changes like this: Besides in general making the PR much smaller, when practical let's avoid mixing client and server changes. Also, I think we should prioritize the changes that are most likely to impact the ECH implementation (changes to the early handshake) ahead of all other changes, so we can help Rob rebase his work onto the refactoring faster and otherwise help Rob.

djc:
Fair enough. I'll close this for now and come back with smaller PRs.

djc:
(FYI, the handshake hash changes do depend on the other changes done here, so it's not like they're grouped completely arbitrarily.)

PR #547, opened 2021-03-09
client: thread transcript through handshake states

djc:
This contains most of the client-relevant changes from #546, up to the point where the transcript can be threaded through the handshake states. This is useful because at a later point we'll want to split `HandshakeHash` up in two separate types to make the type's internal state more resilient and precise (no unwrapping required).

Most of these are highly mechanical commits (`inline .. -> ..`) that just inline state transition methods (and adjust some variable names to make things more natural). I think this is useful because the reduced indirection makes it easier to follow what is going on and it enables having partial states (this is necessary for the `HandshakeHash` goals).

A few other commits remove/reduce abstractions to take advantage of extracting owned data from the various messages, reusing that instead of cloning it, and flattening the logic required to extract the useful data.

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -1,4 +1,3 @@
-use crate::hash_hs;
 #[cfg(feature = "logging")]
 use crate::log::trace;
 use crate::msgs::enums::ExtensionType;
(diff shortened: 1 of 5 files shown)


djc:
I've scaled this back slightly, removing some refactoring commits that weren't really related to inlining the state transitions.

sayrer:
It would help me make progress on ECH to get this PR, #552, and #558 landed. That will let me split out the `dns_name` refactor in the client handshake.

djc:
I've audited the client states to make sure that there are no `Option` types that are unnecessary after these changes; the only `Option` values that are there actually get assigned `Some` or `None` in the same state.

Rebasing this is also fairly painful, although I've gotten pretty fast at it. Would be nice to get it reviewed and merged. If there is something else I can do to make it easier to review, please let me know.

PR #550, opened 2021-03-10
Rename public API types to conform to API guidelines

djc:
The following renames are applied here (these are the only names affected in the public API, as far as I can tell):

- `TLSError` -> `TlsError`
- `ResolvesServerCertUsingSNI` -> `ResolvesServerCertUsingSni`
- `WebPKIVerifier` -> `WebPkiVerifier`
- `sign::RSASigningKey` -> `sign::RsaSigningKey`


ctz:
Thanks!

PR #552, opened 2021-03-11
Remove duplicate code/state between client and server

djc:
Here's some relatively simple ones that should be relatively easy to review.

I feel there's a lot of noise with the `*SessionImpl` types implementing a number of methods seemingly only an additional privacy guard for the actual `Session` impl, very thin wrappers (like `get_protocol_version()`, which expands to `self.common.negotiated_version()` for both client and server). IMO most of the single-line few-use methods should just be inlined into their callers.


briansmith:
> I feel there's a lot of noise with the `*SessionImpl` types implementing a number of methods seemingly only an additional privacy guard for the actual `Session` impl,

I believe this is because Rustls predates `pub(crate)`. Probably some wrappers can be removed in favor of using `pub(crate)` now.

ctz:
Thanks!

ctz:
I'm not at my usual machine so I can't bisect this properly, but is it possible this PR broke the `SkipChangeCipherSpec-Client` bogo test? Will check it out either way tomorrow.

djc:
Well, it passed in PR CI for the coverage tests at least. I'll file an issue to make sure we track the failure and investigate either today or tomorrow.

PR #554, opened 2021-03-11
Use draft version of QUIC for bogo tests for now

djc:


djc:
I guess this isn't needed.

PR #560, opened 2021-03-12
Refactor client EMS state.

sayrer:
Just takes using_ems out of the HandshakeDetails. @djc @briansmith @ctz


rustls/src/client/hs.rs
```diff
@@ -420,13 +424,14 @@ fn emit_client_hello_for_retry(
...
         randoms,
+        using_ems,
         hello,
         early_key_schedule,
         sent_tls13_fake_ccs,
     };
 
     Ok(if support_tls13 && retryreq.is_none() {
-        Box::new(ExpectServerHelloOrHelloRetryRequest { next, extra_exts })
+        Box::new(ExpectServerHelloOrHelloRetryRequest { next, using_ems, extra_exts })
```
djc, line 434: Do we actually need to duplicate this across `ExpectServerHello` and `ExpectServerHelloOrHelloRetryRequest` (which contains `ExpectServerHello`)?

sayrer: Nope. Fixed.

sayrer:
I screwed up a rebase... will just make a new PR.

PR #561, opened 2021-03-12
Refactor server EMS state.

sayrer:
Removed `using_ems` from the server HandshakeDetails. I think I got this right wrt to TLS 1.3, but please check. @djc @briansmith @ctz.

### DIFF
--- a/rustls/src/server/common.rs
+++ b/rustls/src/server/common.rs
@@ -10,7 +10,6 @@ pub struct HandshakeDetails {
     pub transcript: hash_hs::HandshakeHash,
     pub hash_at_server_fin: Option<digest::Digest>,
     pub session_id: SessionID,
-    pub using_ems: bool,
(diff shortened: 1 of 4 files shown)


sayrer:
It looks like the server TLS 1.3 code depends on `using_ems` in a way that it shouldn't, which was masked by `HandshakeDetails` previously. I'll have to take a look at hs::ClientHello.

sayrer:
I think what I have written here has been correct the whole time, because none of the `into_tls13*` methods have `using_ems` in their parameter list (makes sense).

I think this patch is an improvement as-is, but points at a need to statically disentangle TLS 1.2 and TLS 1.3, in the `hs.rs` structs that have `into` methods for both.

sayrer:
The one failure here is in bogo/, but it passes here with the work in #558 and #565.

djc:
LGTM, thanks!

Issue #514, opened 2021-02-16
no server certificate chain resolved


andrewbanchich:
I'm trying to set up mTLS using rustls / tide / tide-rustls, but I'm kind of new to this. What I want to do is configure a server to only accept connections from clients which use a cert that was created by a root CA I trust.

To do this I'm using: https://docs.rs/rustls/0.19.0/rustls/struct.AllowAnyAuthenticatedClient.html

So far, what I have is this:

        // CA public key
        let ca_path = format!("{}/ca.pem", std::env::var("CERT_DIR").unwrap());
        let mut ca = BufReader::new(File::open(ca_path).unwrap());
    
        let mut cert_store = RootCertStore::empty();
        cert_store.add_pem_file(&mut ca).unwrap();
    
        let verifier = AllowAnyAuthenticatedClient::new(cert_store);
        let tls_config = ServerConfig::new(verifier);

This is compiling okay, but when I try making a request using this Reqwest Client:

        let mut pem = Vec::new();
    
        let key_path = format!(
            "{}/test-pkcs8-key.pem",
            std::env::var("CERT_DIR").unwrap()
        );
        std::fs::File::open(key_path)
            .unwrap()
            .read_to_end(&mut pem)
            .unwrap();
    
        let cert_path = format!("{}/test.pem", std::env::var("CERT_DIR").unwrap());
        std::fs::File::open(cert_path)
            .unwrap()
            .read_to_end(&mut pem)
            .unwrap();
    
        let identity = reqwest::Identity::from_pem(&pem).unwrap();
    
        let mut buf = Vec::new();
        let ca = format!("{}/ca.pem", std::env::var("CERT_DIR").unwrap());
        std::fs::File::open(ca)
            .unwrap()
            .read_to_end(&mut buf)
            .unwrap();
        let ca = reqwest::Certificate::from_pem(&buf).unwrap();
    
        let client = reqwest::Client::builder()
            .add_root_certificate(ca)
            .identity(identity)
            .build()
            .unwrap();

it produces an error: `error unexpected error: no server certificate chain resolved`.

I'm not sure what this means. How can I fix this?

Thanks

djc:
Is it possible you haven't configured the actual certificate for the server to use? I think you need to call `tls_config.set_single_cert()` on the server side.

andrewbanchich:
That would help, wouldn't it? 😄 That worked, thanks!

PR #564, opened 2021-03-14
Improved builder-type API for making ServerConfigs

ctz:
The purpose of this is:

- making it impossible for people to arrive at #514, which is a usability issue
- keep it possible to discard ciphersuites/kx groups at link time
- avoiding a huge growth in ServerConfig constructors

### DIFF
--- a/rustls-mio/examples/tlsserver.rs
+++ b/rustls-mio/examples/tlsserver.rs
@@ -579,6 +579,12 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
         rustls::ALL_CIPHERSUITES.to_vec()
     };
 
+    let versions = if !args.flag_protover.is_empty() {
+        lookup_versions(&args.flag_protover)
+    } else {
+        rustls::ALL_VERSIONS.to_vec()
+    };
+
     let certs = load_certs(
         args.flag_certs
             .as_ref()
@@ -591,18 +597,16 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
     );
     let ocsp = load_ocsp(&args.flag_ocsp);
 
-    let mut config = rustls::ServerConfigBuilder::new()
-        .with_cipher_suites(&suites)
+    let mut config = rustls::ConfigBuilder::with_cipher_suites(&suites)
         .with_safe_default_kx_groups()
+        .with_protocol_versions(&versions)
+        .for_server()
+        .expect("inconsistent cipher-suites/versions specified")
         .with_client_cert_verifier(client_auth)
         .with_single_cert_with_ocsp_and_sct(certs, privkey, ocsp, vec![])
         .expect("bad certificates/private key");
 
     config.key_log = Arc::new(rustls::KeyLogFile::new());
-
-    if !args.flag_protover.is_empty() {
-        config.versions.replace(&lookup_versions(&args.flag_protover));
-    }
 
     if args.flag_resumption {
         config.session_storage = rustls::ServerSessionMemoryCache::new(256);
--- a/rustls-mio/examples/tlsserver.rs
+++ b/rustls-mio/examples/tlsserver.rs
@@ -573,8 +573,11 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
         NoClientAuth::new()
     };
 
-    let mut config = rustls::ServerConfig::new(client_auth);
-    config.key_log = Arc::new(rustls::KeyLogFile::new());
+    let suites = if !args.flag_suite.is_empty() {
+        lookup_suites(&args.flag_suite)
+    } else {
+        rustls::ALL_CIPHERSUITES.to_vec()
+    };
 
     let certs = load_certs(
         args.flag_certs
@@ -587,33 +590,33 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
             .expect("--key option missing"),
     );
     let ocsp = load_ocsp(&args.flag_ocsp);
-    config
-        .set_single_cert_with_ocsp_and_sct(certs, privkey, ocsp, vec![])
+
+    let mut config = rustls::ServerConfigBuilder::new()
+        .with_cipher_suites(&suites)
+        .with_safe_default_kx_groups()
+        .with_client_cert_verifier(client_auth)
+        .with_single_cert_with_ocsp_and_sct(certs, privkey, ocsp, vec![])
         .expect("bad certificates/private key");
 
-    if !args.flag_suite.is_empty() {
-        config.cipher_suites = lookup_suites(&args.flag_suite);
-    }
+    config.key_log = Arc::new(rustls::KeyLogFile::new());
 
     if !args.flag_protover.is_empty() {
         config.versions = lookup_versions(&args.flag_protover);
     }
 
     if args.flag_resumption {
-        config.set_persistence(rustls::ServerSessionMemoryCache::new(256));
+        config.session_storage = rustls::ServerSessionMemoryCache::new(256);
     }
 
     if args.flag_tickets {
         config.ticketer = rustls::Ticketer::new().unwrap();
     }
 
-    config.set_protocols(
-        &args
-            .flag_proto
-            .iter()
-            .map(|proto| proto.as_bytes().to_vec())
-            .collect::<Vec<_>>()[..],
-    );
+    config.alpn_protocols = args
+        .flag_proto
+        .iter()
+        .map(|proto| proto.as_bytes().to_vec())
+        .collect::<Vec<_>>();
 
     Arc::new(config)
 }
--- a/rustls-mio/examples/tlsserver.rs
+++ b/rustls-mio/examples/tlsserver.rs
@@ -497,13 +497,13 @@ fn lookup_suites(suites: &[String]) -> Vec<&'static rustls::SupportedCipherSuite
 }
 
 /// Make a vector of protocol versions named in `versions`
-fn lookup_versions(versions: &[String]) -> Vec<rustls::ProtocolVersion> {
+fn lookup_versions(versions: &[String]) -> Vec<&'static rustls::SupportedProtocolVersion> {
     let mut out = Vec::new();
 
     for vname in versions {
         let version = match vname.as_ref() {
-            "1.2" => rustls::ProtocolVersion::TLSv1_2,
-            "1.3" => rustls::ProtocolVersion::TLSv1_3,
+            "1.2" => &rustls::version::TLS12,
+            "1.3" => &rustls::version::TLS13,
             _ => panic!(
                 "cannot look up version '{}', valid are '1.2' and '1.3'",
                 vname
@@ -601,7 +601,7 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
     config.key_log = Arc::new(rustls::KeyLogFile::new());
 
     if !args.flag_protover.is_empty() {
-        config.versions = lookup_versions(&args.flag_protover);
+        config.versions.replace(&lookup_versions(&args.flag_protover));
     }
 
     if args.flag_resumption {
--- a/rustls/src/server/mod.rs
+++ b/rustls/src/server/mod.rs
@@ -185,7 +185,7 @@ pub struct ServerConfig {
 
     /// Supported protocol versions, in no particular order.
     /// The default is all supported versions.
-    pub versions: Vec<ProtocolVersion>,
+    pub versions: crate::versions::EnabledVersions,
 
     /// How to verify client certificates.
     verifier: Arc<dyn verify::ClientCertVerifier>,
@@ -206,7 +206,7 @@ impl ServerConfig {
     /// versions *and* at least one ciphersuite for this version is
     /// also configured.
     pub fn supports_version(&self, v: ProtocolVersion) -> bool {
-        self.versions.contains(&v)
+        self.versions.contains(v)
             && self
                 .cipher_suites
                 .iter()
--- a/rustls/src/server/mod.rs
+++ b/rustls/src/server/mod.rs
@@ -212,11 +212,6 @@ impl ServerConfig {
                 .iter()
                 .any(|cs| cs.usable_for_version(v))
     }
-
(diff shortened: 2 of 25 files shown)


rustls-mio/examples/tlsserver.rs
```diff
@@ -568,8 +568,11 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
         NoClientAuth::new()
     };
 
-    let mut config = rustls::ServerConfig::new(client_auth);
-    config.key_log = Arc::new(rustls::KeyLogFile::new());
+    let suites = if !args.flag_suite.is_empty() {
+        lookup_suites(&args.flag_suite)
+    } else {
+        rustls::ALL_CIPHERSUITES.to_vec()
+    };
```
djc, line 575: Style nit: for short if-else conditionals like this, I find myself increasingly writing `match` expressions instead; these are a little bit more compact vertically and also usually nicely align the outcome values next to each other.

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
     }
 }
 
+/// Building a ServerConfig in a linker-friendly way.
```
djc, line 149: Since this is a decent amount of code, maybe it deserves its own `server::builder` module?

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+///     .with_default_kx_groups()
+///     .without_client_auth()
+///     .with_single_cert(certs, private_key)
+///     .build();
+/// ```
+///
+/// This may be shortened to:
+/// 
+/// ```ignore
+/// ServerConfigBuilder::defaults()
```
djc, line 169: What's the value of having `new()` and `defaults()` be different? I think I've builders usually set up sensible defaults "by default", and then methods are used to tweak the base configuration and configure those things where no meaningful default exists.

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+    /// Choose a specific set of key exchange groups.
+    pub fn with_kx_groups(
+        self,
+        kx_groups: &[&'static SupportedKxGroup],
+    ) -> ServerConfigBuilderWithSuitesAndKxGroups {
+        ServerConfigBuilderWithSuitesAndKxGroups {
+            ciphersuites: self.ciphersuites,
+            kx_groups: kx_groups.to_vec(),
+        }
+    }
```
djc, line 244: As far as I can tell, rustls did not even make key exchange groups configurable in the latest published release. Why is this necessary/useful?

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+    ///
+    /// This is a safe default: rustls doesn't implement any poor-quality groups.
+    pub fn with_default_kx_groups(self) -> ServerConfigBuilderWithSuitesAndKxGroups {
+        self.with_kx_groups(&ALL_KX_GROUPS)
+    }
+}
+
+/// A ServerConfigBuilder where we know the ciphersuites and key exchange
+/// groups.
+pub struct ServerConfigBuilderWithSuitesAndKxGroups {
```
djc, line 256: Did you consider giving this type the `ServerConfigBuilder` name, and using some other name for the completely empty `ServerConfigBuilder`? That would somewhat alleviate my concerns about the lack of defaults. (And I think naming that type would be an interesting exercise.)

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+    }
+
+    /// Disable client authentication.
+    pub fn without_client_auth(self) -> ServerConfigBuilderWithSuitesAndClientAuth {
+        ServerConfigBuilderWithSuitesAndClientAuth {
+            ciphersuites: self.ciphersuites,
+            kx_groups: self.kx_groups,
+            verifier: verify::NoClientAuth::new(),
+        }
+    }
```
djc, line 290: I'm not sure I would make `without_client_auth()` required. (If you only did that to make the type state work out, there are ways to work around that, although things get a little bit more verbose.)

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+            versions: self.optional.versions,
+            verifier: self.verifier,
+            key_log: self.optional.key_log,
+            #[cfg(feature = "quic")]
+            max_early_data_size: 0,
+        }
+    }
+
+    /// Set the MTU to `mtu`.
+    pub fn set_mtu(&mut self, mtu: Option<usize>) {
```
djc, line 417: I think these whould probably lose the `set_` prefix and either take `self` and return `Self` or return `&mut Self`, so that the more fluent combinator style can be used at this stage, too.

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+        self.optional.mtu = mtu;
+    }
+
+    /// Set the supported protocol versions to `versions`.
+    pub fn set_versions(&mut self, versions: &[ProtocolVersion]) {
```
djc, line 422: I understand if you want to leave that for a future PR, but this does seem to be related to the QUIC version restriction (and ECH) discussion. Do you want to scope that out for now?

djc (commented):
I really like the direction here, I think this will make the API much more misuse-resistance. As I feel it is an important topic, a number of questions/suggestions follow:

rustls/src/server/mod.rs
djc: Reflecting upon this, I think you set it up this way to prevent linking all the cipher suites even if not configured? I think we might achieve the same goal if we use something like the complete confit builder which has the cipher suites value as an Option internally which remains None until set and only gets filled in with the default (all suites) when build() is called.

ctz: The problem with this is that build() cannot have any link-time dependency on the defaults, even a conditional one.

djc: Do you agree that having all cipher suites enabled should be the default and the easiest to build, and the link-time restricted smaller subset should be possible but can be slightly less ergonomic?

Maybe we have a `ServerConfigBuilder` with `new()` starts with an empty `cipher_suites: Vec<_>` and `default()` that starts with a complete `cipher_suites`, with a method `cipher_suites(&[_])` that resets the internal cipher suites? That would (a) have a convenient path towards full cipher suites with `default()`, (b) straightforward path to custom cipher suite configurations with `new().cipher_suites()`. A potential downside is that doing `default().cipher_suites()` is possible and works but is not as a efficient as it could be, but it would reduce the complexity of the three-level builder stack quite a bit, and we could warn about it in the configuration. (I think this scheme would satisfy the lack of link-time dependency goal?)

rustls/src/server/mod.rs
briansmith: It is explained in the comment. Perhaps the comment needs to be clarified. Use `new()` and avoid `defaults()` when you want link-time dead code optimization to drop unused code. Use `default()` when you don't care about code size.

rustls/src/server/mod.rs
briansmith: I agree that the key exchange groups should be configurable. Why not? If you're using a profile of TLS that, say, only uses P-256, or mandates P-256 be supported as a key exchange algorithm, then you need to be able to switch the initial client hello to use P-256 instead of X25519 or else you're going to suffer unnecessarily terrible performance.

Although this functionality wasn't in the last release, it is already in main; there was an issue filed about X25519 being hard-coded and a PR was merged to address it.

rustls/src/server/mod.rs
briansmith: > Do you agree that having all cipher suites enabled should be the default and the easiest to build, and the link-time restricted smaller subset should be possible but can be slightly less ergonomic?

We should distinguish between "all cipher suites" and "default cipher suites." Currently Rustls enables all suites it supports by default because it doesn't implement any sub-optimal (in our estimation) cipher suites. However, this is unlikely to be the case in the very long term.

rustls/src/server/mod.rs
briansmith: I agree with the idea of making the user explicitly choose between client auth or no client auth. It's a security vulnerability for a server to be configured without client auth unless it is prepared to deal with unauthenticated (at the TLS level) clients.

rustls/src/server/mod.rs
briansmith: The naming convention without `set_` does seem to be what hyper and other libraries that I'm familiar with do.

rustls/src/server/mod.rs
djc: I guess I don't like the name `defaults()` because it's so close to and unobviously different from `Default::default()`. Also I think builders are often constructed through a `builder()` method on the type that they build. Maybe the way out here is to have a method on `ServerConfigBuilder` that directly returns `ServerConfigBuilderWithSuitesAndKxGroups`? Could be named `with_default_cipher_suites()` (or `with_default_cipher_suites_and_kx_groups()`, but that seems very wordy).

rustls/src/server/mod.rs
djc: Fair enough.

rustls/src/server/mod.rs
djc: Yeah, I think I now better understand why this makes sense.

briansmith:
I don't have strong feelings about the shape of the API here, but I will share my general advice against using the builder pattern: Minimize the use of the builder pattern when practical; avoid complicated, hard-to-understand, and boilerplate-heavy state machines to support the builder pattern.

Consider:

```
let mut cfg = ServerConfigBuilder::new()
        .with_default_ciphersuites()
        .with_default_kx_groups()
        .with_client_cert_verifier(client_auth)
        .with_single_cert(params.key_type.get_chain(), params.key_type.get_key())
```
vs:
```
let certified_key = FixedCertChain::new(params.key_type.get_chain(), params.key_type.get_key());
let mut cfg = ServerConfigBuilder::new(&[DEFAULT_CIPHERSUITES], &[DEFAULT_KX_GROUPS], client_auth, certified_key);
```

vs.:

```
let certified_key = FixedCertChain::new(params.key_type.get_chain(), params.key_type.get_key());
let settings = Settings {
    ciphersuites: &[DEFAULT_CIPHERSUITES],
    kx_groups: &[DEFAULT_KX_GROUPS],
    client_auth,
    certified_key,
};
let mut cfg = ServerConfigBuilder::new(settings);
```

The effect is the same in all cases but the latter two choices seem a lot simpler for people to learn and for us to maintain as it doesn't require creating (as big of a) state machine.

rustls/src/server/mod.rs
```diff
@@ -146,6 +146,311 @@ impl<'a> ClientHello<'a> {
...
+            ciphersuites: ciphersuites.to_vec(),
+        }
+    }
+
+    /// Choose the default set of ciphersuites.
+    ///
+    /// Note that this default provides only high-quality suites: there is no need
+    /// to filter out low-, export- or NULL-strength ciphersuites: rustls does not
+    /// implement these.
+    pub fn with_default_ciphersuites(self) -> ServerConfigBuilderWithSuites {
```
briansmith, line 224: I think it would be simpler to just have callers call `with_ciphersuites(DEFAULT_CIPHERSUITES)` directly, and recommend they pass `DEFAULT_CIPHERSUITES` unless they know what they're doing. Similarly for similar methods.

djc:
> I don't have strong feelings about the shape of the API here, but I will share my general advice against using the builder pattern: Minimize the use of the builder pattern when practical; avoid complicated, hard-to-understand, and boilerplate-heavy state machines to support the builder pattern.

> The effect is the same in all cases but the latter two choices seem a lot simpler for people to learn and for us to maintain as it doesn't require creating (as big of a) state machine.

As a user, I definitely prefer the first variant -- and given the idiom in the ecosystem, I think this is common. An additional advantage is that it can make evolution of a stable API easier. I agree that it requires the implementer to produce a bunch of boilerplate, although in my experience it's not particularly hard to maintain over time. IMO, the complexity lives mostly on the implementer's side and would be easily manageable, especially if we add a bunch of documentation/comments.

briansmith:
Alright. Is there anything to address in this PR then, other than the merge conflicts?

rustls/src/server/mod.rs
djc: It's a trade-off between downstream users having to reference/import the extra symbol vs maintaining an extra trivial method. I'd lean towards the latter option, but either seems okay.

djc:
> Alright. Is there anything to address in this PR then, other than the merge conflicts?

I think all the unresolved comments are still worthy of consideration.

djc:
(Also, once we reach agreement on what this looks like, I'd be happy to prepare a PR that makes the equivalent change on the client side.)

rustls/src/server/mod.rs
djc: I think maybe the simple improvement here is to just rename `defaults()` to `with_defaults()`.

rustls/src/server/mod.rs
djc: Fair enough.

ctz:
> (Also, once we reach agreement on what this looks like, I'd be happy to prepare a PR that makes the equivalent change on the client side.)

Thanks, I'd appreciate if you could take that on.

In this PR, I plan to:

- rename the defaults() to something like `with_safe_cryptography_defaults()` but maybe less wordy
- reduce the scope a bit -- and restrict the 'builder' shape API to decisions we need to link-time efficiency or security decision reasons

On balance that means I don't think this will address the QUIC+TLS1.2 problem

djc:
Sounds good -- I've tackled the TLS 1.2 + QUIC issue in #666 anyway.

djc:
Maybe we should also stop exposing versions as configurable explicitly and handle that by exposing disjoint TLS 1.2 and 1.3 supported cipher suite sets? Otherwise it's still possible to end up with invalid configuration.

Issue #566, opened 2021-03-15
Consider returning Arc from ResolvesServerCert


fd:
Currently `ResolvesServerCert` returns a `CertifiedKey` as is. This means that all implementations of `ResolvesServerCert` must make a full clone of the `CertifiedKey`. Returning a `Arc<CertifiedKey>` instead would prevent the full clone.

```diff
- pub fn resolve(&self, client_hello: ClientHello<'_>) -> Option<CertifiedKey>
+ pub fn resolve(&self, client_hello: ClientHello<'_>) -> Option<Arc<CertifiedKey>>
```

Are there any reasons for not doing this? I may be overlooking something.
I'd be willing to prepare a PR.

djc:
Yeah, I think this makes sense. Please submit a PR!

djc:
This is done.

PR #567, opened 2021-03-15
ResolvesServerCert::resolve now returns a Arc reference

fd:
This PR does two things:

* `ResolvesServerCert::resolve` was changed to return `Option<Arc<CertifiedKey>>` to prevent a full clone for each call to `resolve`. As a result the clone/to_owned calls are pushed down to the sites where the data is used. In the future the this can be further improved such that no clones are needed.
* `CertifiedKey` encoded some state of the handshake procedure. In particular `take_ocsp` and `take_sct_list` which lived on `CertifiedKey` controlled the encoding of the ocsp and sct data. This state was moved to a new type `ActiveCertifiedKey`.

Closes: #566

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
     pass the end-entity and intermediate certificates separately.  This means rustls deals with the case
     where the certificate chain is empty, rather than leaving that to ServerCertVerifier/ClientCertVerifier
     implementation.
+  - *Breaking API change*: `ResolvesServerCert::resolve` now returns `Option<Arc<CertifiedKey>>` instead of 
+    `Option<CertifiedKey>`.
 * 0.19.0 (2020-11-22):
   - Ensured that `get_peer_certificates` is both better documented, and works
     uniformly for both full-handshake and resumed sessions.
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -251,9 +251,9 @@ struct FixedSignatureSchemeServerCertResolver {
 }
 
 impl rustls::ResolvesServerCert for FixedSignatureSchemeServerCertResolver {
-    fn resolve(&self, client_hello: ClientHello) -> Option<rustls::sign::CertifiedKey> {
+    fn resolve(&self, client_hello: ClientHello) -> Option<Arc<rustls::sign::CertifiedKey>> {
         let mut certkey = self.resolver.resolve(client_hello)?;
-        certkey.key = Arc::new(Box::new(FixedSignatureSchemeSigningKey {
+        Arc::make_mut(&mut certkey).key = Arc::new(Box::new(FixedSignatureSchemeSigningKey {
             key: certkey.key.clone(),
             scheme: self.scheme,
         }));
--- a/rustls/src/server/handy.rs
+++ b/rustls/src/server/handy.rs
@@ -96,13 +96,13 @@ impl server::ProducesTickets for NeverProducesTickets {
 pub struct FailResolveChain {}
 
 impl server::ResolvesServerCert for FailResolveChain {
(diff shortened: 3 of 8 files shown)


ctz:
Thanks for working on this. What do you think about removing the Arc inside CertifiedKey -- with this change I think it is no longer useful?

fd:
> What do you think about removing the Arc inside CertifiedKey -- with this change I think it is no longer useful?

Yes I believe this is now redundant. I assume this was originally done to reuse the precomputed private keys.

fd:
Well, it seems that `CertifiedKey` needs to be Clone and `dyn SigningKey` doesn't like to `Clone`. I propose changing `Arc<Box<dyn SigningKey>>` to `Arc<dyn SigningKey>`. Like this there is no need for the double deference. And `Box<dyn SigningKey>` can easily be turned into `Arc<dyn SigningKey>` as `impl<T> From<Box<T>> for Arc<T>` is implemented.

djc:
I was looking at the `SigningKey` bounds recently and did some experimentation. Would this be helpful for you? https://github.com/ctz/rustls/pull/569/commits/16849e6e98f977478c83deee9b66b61f48160c71

djc:
See also #572 for changes I made around how `CertifiedKey` is used in the handshake. Do you have an opinion about how these should interact?

fd:
> I was looking at the SigningKey bounds recently and did some experimentation. Would this be helpful for you? #569

These trait bounds definitely make sense. SigningKey, after creation, should be immutable and shareable between tasks/threads. When a SigningKey is not Send+Sync it is probably an indication that something is very wrong.

> See also #572 for changes I made around how CertifiedKey is used in the handshake. Do you have an opinion about how these should interact?

#572 is incompatible with this PR as the members of CertifiedKey are moved out of CertifiedKey.

I think there are a couple of issues which should probably be addressed separately:

- CertifiedKey is, conceptually, a piece of immutable data used by many handshakes simultaneously. I believe the code should reflect that. `Arc<CertifiedKey>` probably works best to communicate this. Currently the handshake process owns the CertifiedKey (as opposed to a `Arc<CertifiedKey>` or other reference). This allows for hidden state to creep into CertifiedKey (ex. `CertifiedKey::take_ocsp`). #572 resolves the hidden state issue by destructuring the CertifiedKey but it doesn't communicate the shared nature of CertifiedKey.
- The other issue is that the Codec is currently only capable of working with owned data. This, I assume, is because most of it was written several years ago when rust's borrow checker and lifetime ellison were not as smart as they are now. I think it might be worthwhile to revisit this. Most uses of `Vec<u8>` could be replaced by `Cow<'a, [u8]>` which would reduce the clone/allocator load. This also teaches the Codec how to safely handle shared buffers like the certificate chain.

Does this sound right to you? I'm unfamiliar with the full history of the project so...

djc:
> These trait bounds definitely make sense. SigningKey, after creation, should be immutable and shareable between tasks/threads. When a SigningKey is not Send+Sync it is probably an indication that something is very wrong.

The point of #569 is that it shifts those bounds from the trait definition to usage sites. If you say "when as `SigningKey` is not `Send + Sync` .. is very wrong", that indicates that we should not make that change?

> #572 resolves the hidden state issue by destructuring the `CertifiedKey` but it doesn't communicate the shared nature of `CertifiedKey`.

Yup, I think you're right. Although in that case I feel we should not have a mutable `ActiveCertifiedKey`, but instead should just pass around a `&CertifiedKey`. Here's a take that builds upon your work: https://github.com/ctz/rustls/commit/60367cdc593263b4db9bed2ae4eb48f98e2f924f.

djc:
I've respun my take on your branch as #574. If we land that first, this PR should become much smaller, and we no longer need the separate `ActiveCertifiedKey` type.

djc:
Hey, would you be able to rebase this on top of current main (which now includes #574)? Would be nice to have.

ctz:
Have rebased and merged this, and applied the same change to the client. Thanks!

PR #569, opened 2021-03-15
Signing bounds

djc:
The change here moves the `Send + Sync` trait bounds from our own trait definitions to usage sites that need it. This is a little more verbose, but also more precise. I still think it's an improvement, but definitely a trade-off (also could consider making different trade-offs per trait, maybe `Signer` should be `Send + Sync` everywhere?).

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -225,7 +225,7 @@ impl rustls::ServerCertVerifier for DummyServerAuth {
 }
 
 struct FixedSignatureSchemeSigningKey {
-    key: Arc<Box<dyn rustls::sign::SigningKey>>,
+    key: Arc<Box<dyn rustls::sign::SigningKey + Send + Sync>>,
     scheme: rustls::SignatureScheme,
 }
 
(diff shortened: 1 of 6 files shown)


ctz:
I'm not sure I understand this, so you may have to explain it for me.

As I understood it, which may not be very well, the semantics here are:

```
pub trait StoresClientSessions: Send + Sync
```
"all implementors of StoresClientSessions must also be Send and Sync"

```
 storage: Arc<dyn rustls::StoresClientSessions + Send + Sync>,
```
"anything in this slot, must be an arc to something that implements StoresClientSessions, and Send, and Sync"

Aren't these two the ways of saying the same thing? As far as I can see the only effect will be on error messages: in the first you'll get an error when you `impl StoresClientSessions`, the second you'll get an error when you try to fill the `storage` slot.

djc:
Your summary is pretty much right on the money. It seemed nicer/more precise to take a `&dyn sign::SigningKey` as in #597, but it turns out these changes aren't actually necessary. I still believe the way this PR does things is maybe more precise (since it technically might be useful to be able to store a `dyn SigningKey` argument in a context where `Send + Sync` are not required), but the amount of boilerplate doesn't seem worth it especially since my initial use case (taking `dyn SigningKey` as a function argument) already works. I'll just close this for now.

PR #571, opened 2021-03-15
Refactor client EMS.

sayrer:
Addresses @djc's comment from #560.


djc (approved):

djc:
This looks good, I'm happy to merge after GH Actions finds its mojo back.

PR #572, opened 2021-03-16
Split CertifiedKey where this makes sense

djc:
(This is based on #547, only the last commit is relevant.)

Since the current handshake code is pretty cavalier about taking out parts of `CertifiedKey`, I wanted to see how this would look with the key actually being split up before different parts are used/consumed.


djc:
Superseded by #575.

PR #574, opened 2021-03-16
server: make CertifiedKey immutable during handshake

djc:
I think this would be a good precursor refactoring to make #567 easier.


djc:
I'm very confused about how "BoGo test suite" passes, but then the BoGo run in "Measure coverage" fails.

djc:
Also, "Measure coverage (push)" fails, but "Measure coverage (pull_request)" passes...

fd (approved):

PR #576, opened 2021-03-17
doc: add links to RFCs in README

Delapouite:
Hello

RFC 7905 was already linked just above and this PR adds links for other RFCs mentioned, to provide a handy quick access.

Thanks


djc:
Thanks!

Issue #541, opened 2021-02-25
Naming conventions


djc:
rustls uses uppercase acronyms in type names, contra the API guidelines:

> In `UpperCamelCase`, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`.

IMO it would be nice to align with the broader ecosystem here, particularly in something as visible as `TLSError`. On the other hand, changing it might not be worth the pain across the ecosystem. Even in that case, it might be good to change private names and/or stick to the API guidelines for new additions (such as the ECH stuff in #539.

sayrer:
Agreed. I tried to blend in with the rest of handshake.rs, but I didn't really know what to do.

ctz:
Maybe a better name for this type is just `rustls::Error`?

briansmith:
> Maybe a better name for this type is just `rustls::Error`?

SGTM.

Note you can also do `#[deprecated] pub type TLSError = Error;` to ease the transition, although I'd be fine if that was skipped.

djc:
> Maybe a better name for this type is just `rustls::Error`?

👍 So rename the error type and adjust other types to align with the guidelines? Feel like having a deprecated alias for one cycle?

ctz:
It seems pretty low cost to have the deprecated alias, so let's do that IMO.

djc:
What about enum variant names, including those for all those IANA constants? `CHACHA20_POLY1305` => `Chacha20Poly1305`? (I can focus on doing the public API types first, and leave the `internal` module alone.) Should we rename `ciphersuite` to `cipher_suite` everywhere?

ctz:
I think things which are named after external standards should use the naming in those standards, as far as possible. I think that means:

- `CHACHA20_POLY1305_SHA256` as a cipher suite name (but we should fix `BulkAlgorithm` item names)
- `cipher_suite` instead of `ciphersuite`

But, yes, let's concentrate on any public changes first.

djc:
More things to consider:

* Inconsistency like `cert`/`certificate`, `sig`/`signature` in names
* The trait name style of `ProducesTickets`, `ResolvesServerCert` seems fairly unidiomatic as well, I think either `TicketProducer` or `ProduceTicket` would be more common style (I think I prefer the former)

briansmith:
`AEADTicketer` -> `AeadTicketer`.

djc:
[...]

PR #581, opened 2021-03-17
Refactor ClientHello data.

sayrer:
This packages `dns_name` and `extra_exts` together, in anticipation of ECH. Also makes it easier to supply `extra_exts` with `ClientSession::new`.

The trait bounds are a bit repetitive, but I think that problem is waiting on https://github.com/rust-lang/rust/issues/41517. If someone knows of a nicer way to do this, that would be good.

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -10,7 +10,6 @@ use crate::msgs::handshake::SessionID;
 use crate::msgs::persist;
 use crate::sign;
 use crate::kx;
-use webpki;
 
 pub struct ServerCertDetails {
     pub cert_chain: CertificatePayload,
@@ -56,15 +55,13 @@ impl ServerKXDetails {
 pub struct HandshakeDetails {
     pub resuming_session: Option<persist::ClientSessionValue>,
     pub session_id: SessionID,
-    pub dns_name: webpki::DNSName,
 }
 
 impl HandshakeDetails {
-    pub fn new(host_name: webpki::DNSName) -> HandshakeDetails {
+    pub fn new() -> HandshakeDetails {
         HandshakeDetails {
             resuming_session: None,
             session_id: SessionID::empty(),
-            dns_name: host_name,
         }
     }
 }
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -428,7 +428,7 @@ fn emit_client_hello_for_retry(
 
     let next = ExpectServerHello {
         handshake,
-        dns_name: dns_name.clone(),
+        dns_name,
         randoms,
         using_ems,
         transcript,
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -320,7 +320,7 @@ fn emit_client_hello_for_retry<T: 'static + HelloData + Send + Sync>(
         hello_data
             .get_extra_exts()
             .iter()
-            .cloned()
+            .cloned(),
     );
 
     let fill_in_binder = if support_tls13
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -80,10 +80,14 @@ fn find_session(
     let key = persist::ClientSessionKey::session_for_dns_name(dns_name);
     let key_buf = key.get_encoding();
 
-    let value = sess.config.session_persistence.get(&key_buf).or_else(|| {
-        debug!("No cached session for {:?}", dns_name);
-        None
-    })?;
+    let value = sess
+        .config
+        .session_persistence
+        .get(&key_buf)
+        .or_else(|| {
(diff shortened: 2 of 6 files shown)


rustls/src/client/common.rs
```diff
@@ -58,15 +57,13 @@ impl ServerKXDetails {
 pub struct HandshakeDetails {
     pub resuming_session: Option<persist::ClientSessionValue>,
     pub session_id: SessionID,
-    pub dns_name: webpki::DNSName,
 }
 
 impl HandshakeDetails {
-    pub fn new(host_name: webpki::DNSName) -> HandshakeDetails {
+    pub fn new() -> HandshakeDetails {
```
djc, line 63: I think I'd prefer having a separate commit (or PR) for threading the `dns_name` through the handshake states (recommended process: manually edit the diff to ditch the `HelloData`-related chunks).

rustls/src/client/mod.rs
```diff
@@ -620,6 +617,55 @@ impl ClientSessionImpl {
     }
 }
 
+// Added in anticipation of another struct, EncryptedHost, that returns
+// Some() for the last two methods.
+pub trait HelloData {
```
djc, line 622: This should help reduce the trait bounds:

```suggestion
pub trait HelloData: Send + Sync + 'static {
```

rustls/src/client/mod.rs
```diff
@@ -620,6 +617,55 @@ impl ClientSessionImpl {
...
+    fn get_extra_exts(&self) -> &Vec<ClientExtension>;
+    fn push_extra_ext(&mut self, ext: ClientExtension);
+    fn get_hostname(&self) -> webpki::DNSNameRef;
+    fn get_ech_config(&self) -> Option<&ECHConfig>;
+    fn get_outer_exts(&self) -> Option<&Vec<ClientExtension>>;
+}
+
+// Nothing for now
+pub struct ECHConfig {}
```
djc, line 631: Let's just leave it out for now, then?

rustls/src/client/mod.rs
```diff
@@ -620,6 +617,55 @@ impl ClientSessionImpl {
     }
 }
 
+// Added in anticipation of another struct, EncryptedHost, that returns
+// Some() for the last two methods.
+pub trait HelloData {
+    fn get_extra_exts(&self) -> &Vec<ClientExtension>;
```
djc, line 623: If we're going to be sticking to the API guidelines more closely (#541), let's not add [new `get_` prefixed getters](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

djc (commented):
First round. Sorry for the formatting pain!

rustls/src/client/common.rs
sayrer: I'm not attached to the approach here, but do you want `dns_name` passed a separate parameter alongside `extra_exts` in the early parts of the handshake?

djc: Sorry -- no. Just, in terms of reviewing I'd prefer to separate the commit threading the `dns_name` all the way through the handshake states from the commit that introduces `HelloData` (including packaging up the `dns_name` in it for some of the handshake states).

sayrer: OK, but if I remove `dns_name` from `HandshakeDetails`, it has to go somewhere in the early parts of the handshake. Sorry if I'm being dense here. I see two options that might be easier to review than what's here: 

1) thread it through as a separate parameter alongside `extra_exts` early on, the later parts of the handshake would stay the same.

2) Cut this patch down a bit to use a private Host struct instead of a trait, leaving the public API the same, but bundling `extra_exts` and `dns_name` together.

djc: Yeah, I meant the first option. Maybe separate commits to thread dns_name and extra_exts?

sayrer: OK, done.

rustls/src/client/mod.rs
sayrer: I swear I tried that... but I'll check again. Not in the patch anymore, though.

rustls/src/client/mod.rs
sayrer: Got it. Also not in the patch anymore.

rustls/src/client/hs.rs
```diff
@@ -423,6 +428,7 @@ fn emit_client_hello_for_retry(
 
     let next = ExpectServerHello {
         handshake,
+        dns_name: dns_name.clone(),
```
djc, line 431: Do we need this `.clone()`?

sayrer: No, I meant to delete that. Fixed now.

djc (approved):
Thanks!

PR #583, opened 2021-03-18
Enable conversion from integer value to enum

djc:
As suggested by @icing to make conversions more robust for crustls.


ctz:
Looks good, can we use this in ::read() too?

djc:
Yeah, good call!

PR #584, opened 2021-03-19
Thread SessionID through the handshake.

sayrer:
Takes `session_id` out of `HandshakeDetails`

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -120,7 +120,6 @@ struct InitialState {
     dns_name: webpki::DNSName,
     transcript: HandshakeHash,
     extra_exts: Vec<ClientExtension>,
-    session_id: SessionID,
 }
 
 impl InitialState {
@@ -130,7 +129,6 @@ impl InitialState {
             dns_name,
             transcript: HandshakeHash::new(),
             extra_exts,
-            session_id: SessionID::empty(),
         }
     }
 
@@ -150,6 +148,7 @@ impl InitialState {
                 .set_client_auth_enabled();
         }
 
+        let mut session_id: Option<SessionID> = None;
         self.handshake.resuming_session = find_session(sess, self.dns_name.as_ref());
 
         if let Some(resuming) = &mut self.handshake.resuming_session {
@@ -160,22 +159,23 @@ impl InitialState {
                 if !resuming.ticket.0.is_empty() {
                     resuming.session_id = random_sessionid()?;
                 }
-                self.session_id = resuming.session_id;
+                session_id = Some(resuming.session_id);
             }
             debug!("Resuming session");
         } else {
             debug!("Not resuming any session");
         }
         // https://tools.ietf.org/html/rfc8446#appendix-D.4
-        // https://tools.ietf.org/html/draft-ietf-quic-tls-34#ref-TLS13
-        if self.session_id.is_empty() && !sess.common.is_quic() {
-            self.session_id = random_sessionid()?;
+        // https://tools.ietf.org/html/draft-ietf-quic-tls-34#section-8.4
+        if session_id.is_none() && !sess.common.is_quic() {
+            session_id = Some(random_sessionid()?);
         }
 
         let randoms = SessionRandoms::for_client()?;
         let hello_details = ClientHelloDetails::new();
         let sent_tls13_fake_ccs = false;
         let may_send_sct_list = sess.config.verifier.request_scts();
+
         emit_client_hello_for_retry(
             sess,
             self.handshake,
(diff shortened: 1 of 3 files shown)


rustls/src/client/hs.rs
```diff
@@ -120,6 +120,7 @@ struct InitialState {
     dns_name: webpki::DNSName,
     transcript: HandshakeHash,
     extra_exts: Vec<ClientExtension>,
+    session_id: SessionID,
```
djc, line 123: I feel like this shouldn't exist in InitialState, and we should use Option instead of empty() and is_empty().

rustls/src/client/tls12.rs
```diff
@@ -25,12 +25,14 @@ use crate::client::common::{ClientAuthDetails, ReceivedTicketDetails};
 use crate::client::common::{HandshakeDetails, ServerCertDetails, ServerKXDetails};
 use crate::client::hs;
 
+use crate::internal::msgs::handshake::SessionID;
```
djc, line 28: This should be sorted with the other handshake type imports, and not go through the internal reexport.

djc (commented):
Thanks for moving this forward!

rustls/src/client/tls12.rs
sayrer: Oh, sure. I think Intellij did this automatically. Oops.

rustls/src/client/hs.rs
sayrer: That makes sense, I'll get this in the morning. (it's late in Pacific time)

briansmith: > I feel like this shouldn't exist in InitialState, and we should use Option instead of empty() and is_empty().

I agree that it seems like it shouldn't exist in `InitialState`.

I also agree that eventually we should use `Option<SessionID>` instead, but I don't think it has to be done in this PR.

djc: I suppose we could leave that until we can do the same transformation on the server side.

sayrer: Well, I did this before I saw these comments... lmk if the new diff is ok. It wasn't that hard to thread it through.

rustls/src/client/hs.rs
```diff
@@ -341,12 +346,14 @@ fn emit_client_hello_for_retry(
         .map(ClientExtension::get_type)
         .collect();
 
+    let s_id = session_id.unwrap_or(SessionID::empty());
```
djc, line 349: Hmm, this defeats the purpose IMO. We should either go all the way and replace all uses of `empty()` and `is_empty()` with `None` and `is_none()` (in client) or keep it all as is.

(I would also prefer to avoid unrelated formatting changes.)

rustls/src/client/hs.rs
djc: To remove those methods I guess we'd have to do the server side, too? But I was definitely angling to have the use of those methods decimated from the client side handshake code.

rustls/src/client/hs.rs
sayrer: I don't think it does, since it gets the Option handling correct, up to this point. I'll look over the formatting changes.

rustls/src/client/hs.rs
sayrer: I'll resolve this conversation and move discussion down below, to the current patch.

rustls/src/client/hs.rs
djc: I guess that's true -- "defeats the purpose" is worded too strongly. Would you be up to change it throughout the rest of the client handshake?

sayrer: I think this is the best place to stop for one patch, but I am willing to keep going in another patch today. I'll have to change ClientHelloPayload to keep going, and I think that is best done separately.

djc: Fair enough.

djc: Can we just rename this `s_id` to `session_id` then? Suddenly giving it a different name doesn't seem helpful -- I think shadowing is more idiomatic in cases like this.

sayrer: Sure thing. Done.

djc:
Thanks for sticking with it through all the back and forth!

rustls/src/client/hs.rs
```diff
@@ -625,7 +632,7 @@ impl State for ExpectServerHello {
         server_hello
             .random
             .write_slice(&mut self.randoms.server);
-        self.handshake.session_id = server_hello.session_id;
+        self.session_id = server_hello.session_id;
```
briansmith, line 635: NIT: IMO this would have been better as `let session_id = `; there's no need to modify `self` in this function and the eventual goal is to make all these functions as close to purely functional as possible

djc: Fair enough; we'll pick it up in #596.

Issue #573, opened 2021-03-16
SkipChangeCipherSpec-Client BoGo test failing


djc:
This currently seems to be failing on main. Appears to have started after #552, even though BoGo there appeared to have passed for the PR.

djc:
The failure does appear to be intermittent. In a CI run for #574, "BoGo test suite" passed while "Measure coverage" failed [for push](https://github.com/ctz/rustls/actions/runs/658832356), while the [PR run](https://github.com/ctz/rustls/actions/runs/658834276) was all green. In a CI run for #468, both [push](https://github.com/ctz/rustls/actions/runs/658934960) and [PR](https://github.com/ctz/rustls/actions/runs/658935204) were all green.

sayrer:
I had another BoGo test intermittently fail today:

https://github.com/ctz/rustls/runs/2135785637#step:5:1043

Basic-Client-NoTicket-TLS-Sync-ImplicitHandshake

djc:
https://github.com/ctz/rustls/runs/2138795128?check_suite_focus=true

OCSPStapling-Client-TLS13-TLS-Async-PackHandshake

https://github.com/ctz/rustls/runs/2138795376?check_suite_focus=true

ClientAuth-NoFallback-TLS13
WrongMessageType-TLS13-ServerCertificateVerify-TLS

sayrer:
I tried the bogo tests under Valgrind's `DRD` and `Helgrind` tools, and it seemed ok.

https://github.com/grafica/boringssl/commits/drd

I did find errors by using these Rust sanitizer commands. This is my first time using them, although I have used Valgrind and the LLVM ASan/TSan tools a lot.

`RUSTFLAGS="-Z sanitizer=thread" cargo test --target x86_64-unknown-linux-gnu`
`RUSTFLAGS="-Z sanitizer=memory" cargo test --target x86_64-unknown-linux-gnu`

```
$ cargo version
cargo 1.52.0-nightly (90691f2bf 2021-03-16)
$ rustc --version
rustc 1.52.0-nightly (36f1f04f1 2021-03-17)
$ uname -a
Linux sayrer-ml 5.4.0-66-generic #74-Ubuntu SMP Wed Jan 27 22:54:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

sayrer:
Looks like these commands should be something like this to avoid the std library from giving bogus failures.

[...]

PR #587, opened 2021-03-19
[Proof of concept]: MSan branch.

sayrer:
In looking at the intermittent BoGo failures in #573, I began to suspect data races or uninitialized reads in either Rustls, BoGo, or BoringSSL. I tried Go's data race detection on BoGo and a bunch of different Valgrind tools on BoringSSL in #573, and didn't find anything.

However, I have been able to reliably reproduce two uninitialized reads in Rustls. I've gotten good stacks on linux-x64 now, so I'll post them here. These don't seem to be security problems, but they could be correctness problems if the errors are accurate.

So far, I've found two issues:

```
==973295==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x56274f1ff95e in memcmp /home/sayrer/.cargo/registry/src/github.com-1ecc6299db9ec823/regex-1.4.5/src/dfa.rs:878:19
...
```

and

```
test alpn ... ==973297==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55e17c1d2fe0 in rustls::cipher::unpad_tls13::h38be909cac4e072e /home/sayrer/github/grafica/rustls/rustls/src/cipher.rs:333:18
...
```

I'll follow up with steps to reproduce.

[msan.txt](https://github.com/ctz/rustls/files/6174086/msan.txt)

### DIFF
--- a/Xargo.toml
+++ b/Xargo.toml
@@ -1,0 +1,3 @@
+# if using `cargo test`
+[dependencies.test]
+stage = 1


djc:
That first stack doesn't look very credible, with code in core calling into regex. The second stack appears to be all in rustls code (the alpn test running a test handshake), which makes it seemingly extremely unlikely (what with rustls not having any unsafe code) that there's an uninitialized read in this path.

Do we have any data on how likely MSan is to generate false positives?

sayrer:
I was surprised to see anything come up, especially in Rustls itself.

The first stack is line ~692 in key_schedule.rs:

`assert_eq!(seal_output, expected_seal_output);`

Meaning it's likely complaining about the first parameter.

The second stack is here, cipher.rs:

```
fn unpad_tls13(v: &mut Vec<u8>) -> ContentType {
    loop {
        match v.pop() {
            Some(0) => {}

            Some(content_type) => return ContentType::read_bytes(&[content_type]).unwrap(),

            None => return ContentType::Unknown(0),
        }
    }
}
```

That Vec<u8> is read off the wire, I think. I could see a problem here being credible too.

As for how likely a false positive is, I'm not sure.

sayrer:
Steps to reproduce on Ubuntu 20.04 x64, roughly from https://github.com/japaric/rust-san#better-backtraces:

1) ```$ rustc --version
rustc 1.52.0-nightly (36f1f04f1 2021-03-17)
```
2) `$ rustup component add rust-src`
3) `$ cargo install xargo`
4) `sudo apt install llvm`
5) `sudo apt install llvm-dev`
6) `export LLVM_CONFIG=$(which llvm-config)`
7) `RUSTFLAGS="-Z sanitizer=memory" xargo test -Zbuild-std --target x86_64-unknown-linux-gnu --no-fail-fast -- --test-threads 1`

ctz:
I think it's notable that these issues are both complaining about the first operations on memory written by AES-GCM seal/unseal operations?

sayrer:
That makes sense. Same phenomenon as described here: https://github.com/openssl/openssl/issues/8772#issuecomment-484165974

Maybe I'll try to treat these sites with __msan_unpoison as described in that post, just to make it through once.

sayrer:
I got this to run with origin tracking, showing the location of the memory in question, and it is indeed inside ring's asm.

`RUSTFLAGS="-Z sanitizer=memory -Zsanitizer-memory-track-origins" ~/.cargo/bin/xargo test -Zbuild-std --target x86_64-unknown-linux-gnu --no-fail-fast -- --test-threads 1`

That means these two stacks are probably false positives, but I'd still like to get through the whole suite.

[msan_track_origins.txt](https://github.com/ctz/rustls/files/6176099/msan_track_origins.txt)

sayrer:
I also ran similar commands for `sanitizer=address` and `sanitizer=thread`. These finished cleanly.

sayrer:
Perhaps a better fix would be in core::mem::MaybeUninit.

If the sanitizer is on, maybe `assume_init()` should call `__msan_unpoison`

https://github.com/briansmith/ring/blob/74aa79b650f8a1f751993a56046ca43166ae9934/src/aead/aes.rs#L76

Ralith:
I feel like dynamically checking the correctness of `MaybeUninit::assume_init()` is a large part of what rust code wants from msan.

sayrer:
> I feel like dynamically checking the correctness of `MaybeUninit::assume_init()` is a large part of what rust code wants from msan.

Yeah, perhaps if it's C/C++, that's true.

sayrer:
Chatted with Brian about this, and he says he doesn't want to use `core::mem::MaybeUninit` and there are only a few instances left in ring. He'll remove them in the next few days, and I'll revisit this test run.

sayrer:
Well, maybe I got this wrong, but the issue is being tracked: https://github.com/briansmith/ring/issues/1227

briansmith:
Since the main branch of *ring* already has API-breaking changes, any release off of the main branch that would incorporate changes for enabling MSAN support will only be availabile in *ring* 0.17 or later. So, I'm going to concentrate on making the release of 0.17, and then updating Rustls to use 0.17. In order to make that easier, I'm going to be working on making it possible to link multiple versions of *ring*, 0.16 and 0.17, at the same time.

briansmith:
Looks like this is blocked on me making improvements to *ring*? I will prioritize them.

cpu:
I think since this branch is blocked on work happening in Ring, and the relevant issue (https://github.com/briansmith/ring/issues/12270) hasn't had updates in ~2 years we should close this branch for now as its unlikely to make progress in the short term and is accruing conflicts.

I've subscribed to the upstream issue. If it gets some traction we could revisit this PR. It would be really nice to get MSAN integrated :pray:

PR #589, opened 2021-03-20
Try bumping the Go version for bogo.

sayrer:
I read the changelogs for recent Go versions, and there were plenty of serious issues, so I thought this might be worth a try.


sayrer:
This passed on my machine, but failed with HelloRetryRequest-DuplicateCookie-TLS13 here.

sayrer:
This current revision got all the way to Client-AcceptJDK11DowngradeRandom (the second-to-last test) and then the CI server stopped responding.

sayrer:
I think looking at the tests as they run is causing a stall... the bogo suite should run in about 2 minutes and 20 seconds.

sayrer:
OK, got the typical style of failure to happen, so these variables can be eliminated. 🙃

PR #590, opened 2021-03-20
Print on failure.

sayrer:


sayrer:
Well, this is horrible. Back to #573, but at least it looks like the Go process exits, and rustls gives a uninformative error when this happens.

Issue #586, opened 2021-03-19
Refactor quic submodule into `client::quic` and `server::quic`


briansmith:
When reviewing PR #575 I noticed that lots of the implementation details of `ClientSession`/`ClientSessionImpl`, especially the fields, are `pub(crate)` or equivalent when they ideally wouldn't be. It seems like one of the biggest reasons for this is that the implementations of types in the `quic` module directly access those fields.

It seems to me like we should refactor the `quic` submodule so that there is a `client::quic` submodule for the client-specific parts and a `server::quic` submodule for the server-specific parts. This would allow us to encapsulate more of the client and server state better.

djc:
Makes sense. I started drafting this for the client, but will wait to submit it until after #575 lands, otherwise rebasing it will just cause a lot of pain.

PR #592, opened 2021-03-21
client: move QUIC client code into client module (see #586)

djc:


ctz:
Thanks!

Please, if you can, avoid mentioning PR/issue numbers in commit messages. Github doesn't handle it well with rebase workflows.

djc:
Not sure how it doesn't handle it well? Happy to avoid doing so in the future though.

PR #596, opened 2021-03-22
SessionID as Option

sayrer:
As discussed in #584, this patch makes SessionID an Option in most places. I had to touch parts of the codebase I'm not very familiar with, so I'm not that confident in this patch, but it does pass all of the tests.

I wasn't sure what to do when reading a zero-length SessionID off the wire. I left it as-is, but perhaps it should result in None.


rustls/src/client/hs.rs
```diff
@@ -683,71 +681,78 @@ impl State for ExpectServerHello {
...
-                if resuming.supported_cipher_suite() != suite {
-                    let error_msg = "abbreviated handshake offered, but with varied cs".to_string();
-                    return Err(TlsError::PeerMisbehavedError(error_msg));
-                }
-
-                // And about EMS support?
-                if resuming.extended_ms != self.using_ems {
-                    let error_msg = "server varied ems support over resume".to_string();
-                    return Err(TlsError::PeerMisbehavedError(error_msg));
+        match (&self.handshake.resuming_session, self.session_id) {
```
djc, line 684: Above, let's also fix the nit Brian mentioned here: https://github.com/ctz/rustls/pull/584/files/9d4761acc86de30e02f8460364ca49413d937b82#r599254767. (Line 634 in the current state of your branch.)

rustls/src/client/hs.rs
```diff
@@ -683,71 +681,78 @@ impl State for ExpectServerHello {
...
+                debug!("Did not resume session.");
             }
         }
 
         Ok(Box::new(tls12::ExpectCertificate {
             handshake: self.handshake,
-            session_id: self.session_id,
+            session_id: self
+                .session_id
+                .ok_or(TlsError::HandshakeNotComplete)?,
```
djc, line 755: I spent some time figuring out why this is here, but I can't quite figure it out. Why do we need this?

djc (commented):
Thanks, I think this is a useful change! Bunch of nits and one question left...

rustls/src/msgs/persist.rs
```diff
@@ -171,7 +171,7 @@ impl ClientSessionValueWithResolvedCipherSuite {
         self.supported_cipher_suite
     }
 
-    pub fn set_session_id(&mut self, id: SessionID) {
+    pub fn set_session_id(&mut self, id: Option<SessionID>) {
```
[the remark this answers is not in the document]
sayrer: I don't think I understand what you want here.

rustls/src/client/hs.rs
sayrer: I thought that must be true as a result of line 634, but I can make it an Option as well. It also relates to the question I asked above:

> I wasn't sure what to do when reading a zero-length SessionID off the wire. I left it as-is, but perhaps it should result in None.

I don't think this will hold if the server's session_id can be None.

rustls/src/msgs/persist.rs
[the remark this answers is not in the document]
sayrer: @djc need a little guidance here when you get some time, would be nice to land this.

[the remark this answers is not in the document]
djc: I meant to change the function signature to have an argument of type `SessionID`, leaving the `Some`-wrapping to this method.

rustls/src/client/hs.rs
djc: I do think reading a zero-length `SessionID` off the wire should result in `None` (IMO only non-empty `SessionID`s should be `Some`).

rustls/src/msgs/persist.rs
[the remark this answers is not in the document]
sayrer: Yes, I get it now. What you suggest would prevent setting this value to None, but I suppose that's what you're going for.

rustls/src/client/hs.rs
sayrer: OK, I will change adjust this to interpret zero-length `SessionID` messages as None.

rustls/src/msgs/persist.rs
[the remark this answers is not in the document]
djc: (I'll review the other parts, but this one is still open.)

rustls/src/msgs/handshake.rs
```diff
@@ -865,7 +861,7 @@ impl Codec for ClientHelloPayload {
         let mut ret = ClientHelloPayload {
             client_version: ProtocolVersion::read(r)?,
             random: Random::read(r)?,
-            session_id: SessionID::read(r)?,
+            session_id: Some(SessionID::read(r)?),
```
djc, line 864: All instances of `Some(SessionID::read(r)?)` can be replaced with just `SessionID::read(r)`.

rustls/src/server/tls12.rs
```diff
@@ -403,17 +403,19 @@ impl hs::State for ExpectFinished {
...
-            let worked = sess.config.session_storage.put(
-                self.handshake.session_id.get_encoding(),
-                value.get_encoding(),
-            );
-            if worked {
-                debug!("Session saved");
-            } else {
-                debug!("Session not saved");
+        if !self.resuming {
+            if let Some(session_id) = self.handshake.session_id {
```
djc, line 407: To prevent the increased indentation here, maybe try `if let (Some(session_id), false) = (self.handshake.session_id, self.resuming)`?

sayrer:
> I would expect that the SeessionID::new constructor will be replaced with a `try_from()`

I did this, just a few failing tests now.

sayrer:
I think the only thing left to do here is fix:

```
fn random_sessionid() -> Result<SessionID, rand::GetRandomFailed> {
    let mut random_id = [0u8; 32];
    rand::fill_random(&mut random_id)?;
    let id: &[u8] = &random_id;
    // TODO: fix this unwrap
    Ok(SessionID::try_from(id).unwrap())
}
```

The problem being that `rand::fill_random` and `SessionID::try_from` return different error types. I don't know the best way to deal with that.

djc:
I've submitted a solution for that a separate PR in #610. It turns out `SessionID` is really only used (except in test code) to create random `SessionID`s, so if we just flatten that abstraction, things work out for the better (and we don't even need the `TryFrom` impl, I think).

sayrer:
> I've submitted a solution for that a separate PR in #610. `SessionID` is only used (except in test code) to create random `SessionID`s, so if we just flatten that abstraction, things work out for the better (and we don't even need the `TryFrom` impl, I think).

That looks better than what I have. I'll wait until that lands and rebase.

sayrer:
Here is the fuzzer error: https://gist.github.com/sayrer/1c8bb7b1f1eb1559fb9d5ed93af842e4

I've never used `cargo fuzz` before, so I'm digging into how to reproduce it. It doesn't trigger easily locally.

djc:
Maybe throw both of  the byte sequence into `Message::read_with_detailed_error()` and see what comes out? It's unfortunate that your backtrace doesn't seem to find symbols.

sayrer:
These both failed in these fuzz cases on the 44th byte, I think. So it is reproducing reliably in CI. It does not happen on my machine.

sayrer:
Well, my copy of nightly is a week out-of-date. I will try updating it to the version used in CI. I am on x64 Linux.

```
sayrer@sayrer-ml:~/github/grafica/rustls$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/sayrer/snap/rustup/common/rustup

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)
sayrer@sayrer-ml:~/github/grafica/rustls$ rustc --version
rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)
```

sayrer:
I can reproduce this now with `cargo fuzz run message -- -max_total_time=10`.

sayrer:
It's this test that is failing:

```
fuzz_target!(|data: &[u8]| {
    let mut rdr = Reader::init(data);
    if let Some(mut m) = Message::read(&mut rdr) {
        m.decode_payload();
        //println!("msg = {:#?}", m);
        let enc = m.get_encoding();
        //println!("data = {:?}", &data[..rdr.used()]);
        assert_eq!(enc, data[..rdr.used()]);
    }
});
```
The 44th byte is different, but the output is otherwise identical.

sayrer:
```
m.typ: Handshake
m.version: TLSv1_2
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `[22, 3, 3, 0, 65, 2, 0, 0, 61, 3, 3, 150, 58, 219, 228, 186, 148, 105, 216, 198, 103, 26, 194, 118, 139, 92, 132, 183, 122, 108, 147, 213, 248, 17, 44, 104, 144, 251, 156, 124, 192, 203, 42, 0, 0, 116, 0, 0, 21, 0, 0, 0, 0, 255, 1, 0, 1, 0, 0, 11, 0, 4, 3, 0, 1, 255, 4, 35, 0, 0]`,
 right: `[22, 3, 3, 0, 65, 2, 0, 0, 61, 3, 3, 150, 58, 219, 228, 186, 148, 105, 216, 198, 103, 26, 194, 118, 139, 92, 132, 183, 122, 108, 147, 213, 248, 17, 44, 104, 144, 251, 156, 124, 192, 203, 42, 47, 0, 116, 0, 0, 21, 0, 0, 0, 0, 255, 1, 0, 1, 0, 0, 11, 0, 4, 3, 0, 1, 255, 4, 35, 0, 0]`', fuzzers/message.rs:17:9
```

sayrer:
It looks like ClientHelloPayload is the problem.

sayrer:
It seems like

```
    pub fn encode(session_id: Option<SessionID>, bytes: &mut Vec<u8>) {
        match session_id {
            Some(session_id) => session_id.encode(bytes),
            None => bytes.push(0u8),
        }
    }
```
is somehow incorrect in the fuzz tests.

sayrer:
OK, I think I found it. It looks like ServerHelloPayload and ClientHelloPayload were relying on SessionID deserialization in some some surprising ways. I got one of these fuzz tests to pass with this unacceptable diff. I think `ClientHelloPayload` has a similar issue. On the upside, I think this is why this patch is a good idea.

```
diff --git a/rustls/src/msgs/handshake.rs b/rustls/src/msgs/handshake.rs
index 8fbd7c2..921921c 100644
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1211,6 +1211,9 @@ impl Codec for ServerHelloPayload {
     // minus version and random, which have already been read.
     fn read(r: &mut Reader) -> Option<ServerHelloPayload> {
         let session_id = SessionID::read(r);
+        if session_id.is_none() {
+            return None;
+        }
         let suite = CipherSuite::read(r)?;
         let compression = Compression::read(r)?;
         let extensions = codec::read_vec_u16::<ServerExtension>(r)?;
```

sayrer:
I'll add a couple of fuzz tests to the corpus in this patch. Making them pass causes some unit tests to fail, so I think it's a good check.

sayrer:
Reproducible with

`cargo fuzz run message fuzz/corpus/message/Handshake.258f5d2d7a8677c9face5972a70ba9a6ef2b85c4.bin`

ctz:
> OK, I think I found it. 

I don't think ClientHelloPayload/ServerHelloPayload should have Option<SessionID> as a member, because by definition they always have a session ID inside (even if its empty). I don't think `Option<SessionID>::None` is a good representation for an empty session ID, since that's a valid session ID and, as you've found, it means we can't distinguish invalid and valid session ID encodings.

If you could rebase this work on main and squash some of the back-and-forth commits that'd be helpful.

sayrer:
> If you could rebase this work on main and squash some of the back-and-forth commits that'd be helpful.

Sure. Sorry this got a little messy, as it took me a few tries to reproduce the CI errors. I also figured you'd squash it all when merging, but I can be more careful about this kind of thing in the future.

sayrer:
I will get this working with revisions ClientHelloPayload/ServerHelloPayload here, and then do a new PR for a cleaner commit history on main. At least we should get an extra couple of fuzz tests out of this.

sayrer:
(This isn't ready to review yet, just made ctz's suggested change and got it compiling)

djc:
> I don't think ClientHelloPayload/ServerHelloPayload should have Option as a member, because by definition they always have a session ID inside (even if its empty). I don't think Option<SessionID>::None is a good representation for an empty session ID, since that's a valid session ID and, as you've found, it means we can't distinguish invalid and valid session ID encodings.

Not sure I'm quite convinced by this argument. It seems to me that the RFCs repeatedly call out empty session IDs as special, and encoding this in the type system seems quite useful, and less error prone than relying on calls to `is_empty()` to check for this special case. I'm not exactly sure how/why we can't distinguish invalid and valid session ID encodings if we use `None` to represent the empty session ID; we probably shouldn't be using `Option` to encode invalid vs valid anyway.

sayrer:
I'm going to close this in favor of #621, since this got a little messy. I'll copy over the last comment.

PR #597, opened 2021-03-27
server: simplify how SigningKey arguments are passed

djc:
Thinking about this some more, the whole `Box` part of `Arc<Box<T>>`  is completely superfluous, removed it (and carried forward the `any_supported_type()` flattening from the other PR).

(Or is keeping the `any_supported_type()` split separate useful for limited linking of algorithms? It doesn't seem so since they call each other unconditionally, but maybe I'm missing something.)

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -225,7 +225,7 @@ impl rustls::ServerCertVerifier for DummyServerAuth {
 }
 
 struct FixedSignatureSchemeSigningKey {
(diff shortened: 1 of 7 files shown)


ctz:
> (Or is keeping the any_supported_type() split separate useful for limited linking of algorithms? It doesn't seem so since they call each other unconditionally, but maybe I'm missing something.)

The whole of `mod sign` is public, so they're part of the public API: https://docs.rs/rustls/0.19.0/rustls/sign/index.html

djc:
Ah, sorry, I missed that. I'll scrap that commit then.

PR #598, opened 2021-03-27
Miscellaneous unwrap() removals

djc:
`copy_from_slice()` is arguably more idiomatic for this common case than using `Write`, and makes the code more compact.  By referring to array sizes explicitly where possible, I've tried to make it easy for the optimizer to reduce/remove any potential bounds checks in the generated code. `copy_from_slice()` itself can still panic if its arguments are not of equal length, but this still seems like an improvement.


ctz:
Thanks!

PR #599, opened 2021-03-27
admin/fmt script fixes

djc:


ctz:
I think admin/fmt should pass "$@" to cargo-fmt, and then the CI invocation can pass through `--check`

djc:
Fixed that up.

briansmith (approved):

PR #604, opened 2021-03-30
server: move handshake data into states

djc:
Detach all methods that take a state (to prevent issues with partial moves) and move most data from `HandshakeDetails`, `ServerKXDetails` and `ClientCertDetails` into the relevant handshake states. TLS 1.2-specific methods are moved from `server::hs` to `server::tls12`. Remove one unnecessary case of unwrapping and prevent cloning of extra (QUIC) extensions.

### DIFF
--- a/rustls/src/server/common.rs
+++ b/rustls/src/server/common.rs
@@ -3,8 +3,6 @@ use crate::key;
 use crate::kx;
 use crate::msgs::handshake::SessionID;
 
-use std::mem;
-
 pub struct HandshakeDetails {
     pub transcript: hash_hs::HandshakeHash,
     pub session_id: SessionID,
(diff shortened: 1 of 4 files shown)


ctz:
Concerned that in this PR we're lose type safety when messages are being sent. To take one commit as an example: ff5ee44f -- `emit_certificate` can now be called at any time during the handshake, but the _only_ right time is during `ExpectClientHello`'s lifetime

djc:
Fair enough. Alternative: move the TLS 1.2 handling to a separate type (similar to 1.3 CompleteClientHelloHandling), and put that type in an inline module that also contains all of these functions. Sound good?

djc:
Rebased and updated as suggested in my previous comment.

djc:
@ctz can you comment on whether you want to merge this in its current form or not? Rebasing it is a PITA. I still think the enabling of partial moves is useful (for the transcript hash type if nothing else), and I actually like the increased symmetry between TLS 1.2 and 1.3. I believe the module structure gives the same privacy guarantees as the previous design.

ctz:
Sorry, I hadn't got to rereviewing this and it was a bit of a slog with so much code movement. Merged as cf061493

djc:
Thanks!

Issue #563, opened 2021-03-13
`Read` impl on Session


jsha:
Related to #562

For `Session`, the `Read` trait is a somewhat awkward fit. https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read

> If n is 0, then it can indicate one of two scenarios:
>    This reader has reached its "end of file" and will likely no longer be able to produce bytes. Note that this does not mean that the reader will always no longer be able to produce bytes.

Currently, `Session::read()` takes advantage of the wiggle room in "likely." Returning `Ok(0)` from `Session::read()` means "no more plaintext bytes available in buffer, but that can change in the future." Returning `Err(io::ErrorKind::ConnectionAborted)` from `Session::read()` means "successful EOF for the overall connection."

That's a bit confusing to deal with because it inverts the usual way Read impls work:

> This function does not provide any guarantees about whether it blocks waiting for data, but if an object needs to block for a read and cannot, it will typically signal this via an Err return value.

And also it means you can't, for instance, stack a `Session` with something else that requires a `Read`, like a gzip decoder.

We could switch these: `Session::read` would return an error (e.g. `io::ErrorKind::WouldBlock`) when there are temporarily no bytes available, and return `Ok(0)` upon `close_notify`.

Another approach would be: remove the `Read` impl on `Session`. Instead, `process_new_packets` would return `Result<PartialReader, TLSError>`, where `PartialReader` would `impl Read`, and would read however many bytes are available in the buffer and then return `Ok(0)`.

djc:
It would be nice if we can align with the std semantics, but I'm a bit worried about the compatibility situation -- seems likely we would break a bunch of downstream users, and there's not really a way to make this obvious at compile time? I guess we could add a new method with the changed semantics.

jsha:
Agreed that changing the semantics without notice at compile time would be an issue. I think probably the best would be as you say, to deprecate the `read` method (and in general the `impl Read for Session`). The replacement could either be a differently-named method with new semantics, or a `PartialReader` returned from `process_new_packets`.

In the scenario above, `Stream` and `StreamOwned` would still `impl Read`. Worth noting, however, that they have similarly awkward semantics: they're return an error if they receive a `close_notify`, when we'd really like that to be treated as an EOF with no errors.And if they receive a TCP close without a `close_notify` first, that should be treated as an error.

briansmith:
It seems to me that the caller of `process_new_packets()` needs to be able to recognize many different states:

[...]

PR #605, opened 2021-03-31
Add is_close_notify helper.

jsha:
This helps users distinguish a CloseNotify from other types of I/O
errors.

### DIFF
--- a/rustls/src/session.rs
+++ b/rustls/src/session.rs
@@ -456,6 +456,19 @@ pub struct SessionCommon {
     pub(crate) quic: Quic,
 }
 
+/// Return true if the provided io::Error is the result of receiving a close_notify
+/// alert from the peer. Users should treat this as a clean shutdown of the connection.
+pub fn is_close_notify(err: &io::Error) -> bool {
+    let inner = match err.get_ref() {
+        Some(inner) => inner,
+        None => return false,
+    };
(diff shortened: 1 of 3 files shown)


jsha:
This is not just useful for crustls, but also for any Rust code that deals with close-delimited streams (i.e. any HTTP response without Content-Length or Transfer-Encoding). It's particularly useful when dealing with `StreamOwned`, where the returned `io::Error` could come from the underlying `TcpStream` _or_ from a `CloseNotify`.

rustls/src/session.rs
```diff
@@ -456,6 +456,19 @@ pub struct SessionCommon {
     pub(crate) quic: Quic,
 }
 
+/// Return true if the provided io::Error is the result of receiving a close_notify
+/// alert from the peer. Users should treat this as a clean shutdown of the connection.
```
briansmith, line 460: It seems wrong for us to treat `close_notify` as an error when really it's the least erroneous result we could get; only when we receive a `close_notify` has everything gone according to how the spec says it should. So, like I said in the related issue, it seems like we should redo the API to make receiving a close_notify an explicit non-error result.

briansmith:
Given the discussion in the issue, I think we should close this.

jsha:
Closing in favor of modifying the Session API to more clearly indicate close_notify as an "EOF with no errors" state: #563

PR #606, opened 2021-03-31
Thread cipher suite through handshake states

djc:
Avoids a lot of unwrapping via `SessionCommon::get_suite_assert()`.


ctz:
I'm afraid the 30 or so repetitions of this trivial `State::suite` function aren't really doing it for me :(

djc:
Fair enough. How about for now I keep the Option value in SessionCommon in addition to the values in the state types?

ctz:
I think that sounds like a reasonable compromise, yes.

PR #608, opened 2021-03-31
Use IPA for pronunciation guides

jglick:
As a native English speaker I can only interpret the current advice by analogy to https://en.wiktionary.org/wiki/rustle#Pronunciation but that does not really sound right either—silent /t/ really? I can only imagine how an ESL speaker (i.e., most programmers) would interpret this. Please edit or ask me to edit if my guess was not in fact correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 </p>
 
 <p align="center">
-Rustls is a modern TLS library written in Rust.  It's pronounced 'rustles'. It uses <a href = "https://github.com/briansmith/ring"><em>ring</em></a> for cryptography and <a href = "https://github.com/briansmith/webpki">libwebpki</a> for certificate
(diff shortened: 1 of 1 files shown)


briansmith:
Understandable. However, most people can't read IPA, and if the IPA is useful for them, they could just look up the IPA for "rustles" in the dictionary, right? So I'd suggest we don't do this.

jglick:
For a native English speaker it is helpful to say something like

> pronounced like _Russell’s_

(again, if that is in fact the pronunciation!) but for everyone else this just replaces one problem with another.

sayrer:
It's probably better to [delegate this problem to Google](https://user-images.githubusercontent.com/253473/113231236-a5671500-924f-11eb-935c-3e8eaa1cfb2f.png)). I always say "Rust Tee El Ess" in my mind, but the documented pronunciation doesn't bother me.

djc:
Yeah, I think linking to (something that can produce) actual audio is the best solution here.

ctz:
Maybe we should remove the "It's prounounced" sentence altogether? It's ultimately not very important information, and it's certainly not paying its rent for being the second thing someone reads in the readme!

sayrer:
That also sounds good. I remember thinking it was strange when I first read it.

ctz:
Have done that. Thanks for the impetus!

PR #609, opened 2021-03-31
Pin rustfmt and nightly revision used to build it

djc:


admin/fmt
```diff
@@ -4,7 +4,13 @@ set -xe
 ADMIN=$(dirname "$0")
 PREV=$(pwd)
 if [ ! -d $ADMIN/rustfmt ] ; then
-  git clone --depth 1 https://github.com/rust-lang/rustfmt $ADMIN/rustfmt
+  COMMIT=8829e455b722f1ba8ade915ad2687639a2abb2c3
```
briansmith, line 7: Please add a comment above this about why we're pinning the Rustfmt version: We want to ensure that people submitting PRs and CI agree on what version to use, and we want to limit our exposure to breaking changes in the rust-lang/rustfmt repo.

Further, it seems like we should use a fixed version of Rust Nightly, e.g. `+nightly-2021-03-31` to get a consistent build of rustfmt so the script doesn't break when the rust-lang/rust repo is updated.

Further, We need to ensure that that version of Rust Nightly is installed. Note that the runner of this script might not have any version of Rust Nightly installed already, especially if they typically use stable Rust.

admin/fmt
```diff
@@ -4,7 +4,13 @@ set -xe
 ADMIN=$(dirname "$0")
 PREV=$(pwd)
 if [ ! -d $ADMIN/rustfmt ] ; then
```
briansmith, line 6: It seems like we need to check the actual version of rustfmt stored in $ADMIN/rustfmt to see if it is the version that corresponds to the given commit. In particular, when we update $COMMIT below, this should trigger a re-download and rebuild of the new version of rustfmt.

briansmith (changes requested):

admin/fmt
sayrer: Please excuse this annoying comment: Bazel would do this as part of its approach. If we're going this route, it might be better to download Bazel and run this script through it, instead of depending on these shells scripts to get this right.

ctz (approved):

ctz:
Going to merge this. I don't think there's a lot of value in gold-plating this script, as hopefully it will go away in a future version of rustfmt available via rustup: see rust-lang/rustfmt/pull/4782 and rust-lang/rustfmt/issues/4769

djc:
Yeah, that's looking great!

PR #615, opened 2021-04-03
Stricter fuzz corpus test

sayrer:
I noticed a test I added to the corpus was failing if I pointed `cargo fuzz` at it, but passing here. This patch makes `message_test` just as tough as `cargo fuzz`, but leaves in what was there before, on the assumption that it's there to ease debugging.

Happy to squash these commits as suggested in other PRs, but I think GitHub's squash-on-merge feature would be fine here (I'll just do whatever is easiest to work with). I don't think the formatting test failure has anything to do with this patch.


sayrer:
I added this code to #596 as a demonstration. See https://github.com/ctz/rustls/pull/596/checks?check_run_id=2261336920.

djc:
Thanks!

PR #617, opened 2021-04-05
Ignore DS_Store

sayrer:
macOS leaves these annoying files around if you browse a directory in the Finder.


djc:
Makes sense, thanks!

PR #619, opened 2021-04-07
Set up clippy in CI

djc:
This also helped detect remaining naming issues.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -166,7 +166,7 @@ impl InitialState {
         let hello_details = ClientHelloDetails::new();
         let sent_tls13_fake_ccs = false;
         let may_send_sct_list = sess.config.verifier.request_scts();
-        emit_client_hello_for_retry(
+        Ok(emit_client_hello_for_retry(
             sess,
             self.resuming_session,
             randoms,
@@ -180,7 +180,7 @@ impl InitialState {
             self.extra_exts,
             may_send_sct_list,
             None,
-        )
+        ))
     }
 }
 
@@ -224,7 +224,7 @@ fn emit_client_hello_for_retry(
     extra_exts: Vec<ClientExtension>,
     may_send_sct_list: bool,
     suite: Option<&'static SupportedCipherSuite>,
-) -> NextStateOrError {
+) -> NextState {
     // Do we have a SessionID or ticket cached for this host?
     let (ticket, resume_version) = if let Some(resuming) = &resuming_session {
         (resuming.ticket.0.clone(), resuming.version)
@@ -341,14 +341,14 @@ fn emit_client_hello_for_retry(
         .map(ClientExtension::get_type)
         .collect();
 
-    let session_id = session_id.unwrap_or(SessionID::empty());
+    let session_id = session_id.unwrap_or_else(SessionID::empty);
 
     let mut chp = HandshakeMessagePayload {
         typ: HandshakeType::ClientHello,
         payload: HandshakePayload::ClientHello(ClientHelloPayload {
             client_version: ProtocolVersion::TLSv1_2,
             random: Random::from_slice(&randoms.client),
-            session_id: session_id,
+            session_id,
             cipher_suites: sess.get_cipher_suites(),
             compression_methods: vec![Compression::Null],
             extensions: exts,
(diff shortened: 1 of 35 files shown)


rustls/src/keylog.rs
```diff
@@ -160,6 +159,13 @@ impl KeyLog for KeyLogFile {
     }
 }
 
+impl Default for KeyLogFile {
```
briansmith, line 162: This is worse. Let's just disable this lint.

rustls/src/msgs/codec.rs
```diff
@@ -120,6 +120,7 @@ impl u24 {
     }
 }
 
+#[allow(clippy::from_over_into)]
```
briansmith, line 123: It seems like we should follow this PR up with a new one that does what the lint suggests. I think these are only written the way they are because of the old coherence rules.

rustls/src/server/handy.rs
```diff
@@ -135,16 +134,15 @@ impl server::ResolvesServerCert for AlwaysResolvesChain {
 
 /// Something that resolves do different cert chains/keys based
 /// on client-supplied server name (via SNI).
+#[derive(Default)]
```
briansmith, line 137: Again, we should undo this and disable the lint.

rustls/src/sign.rs
```diff
@@ -397,3 +399,7 @@ pub fn supported_sign_tls13() -> &'static [SignatureScheme] {
         SignatureScheme::ED25519,
     ]
 }
+
+/// Errors while signing
+#[derive(Debug)]
+pub struct SignError;
```
briansmith, line 405: Is this supposed to be publicly-constructible? I would expect it to be `pub struct SignError(());` or `pub struct SignError(pub(crate) ());`.

briansmith (changes requested):
I'm not convinced that the `SignError` stuff is the best API design but it's no worse than `()`.

rustls/src/msgs/codec.rs
djc: In my opinion, conceptually it does not make sense to have a conversion from `usize` -> `u24`.

rustls/src/sign.rs
djc: Fixed.

rustls/src/keylog.rs
djc: Fixed.

rustls/src/client/hs.rs
```diff
@@ -226,7 +224,7 @@ fn emit_client_hello_for_retry(
     extra_exts: Vec<ClientExtension>,
     may_send_sct_list: bool,
     suite: Option<&'static SupportedCipherSuite>,
-) -> NextStateOrError {
+) -> Box<dyn State + Send + Sync> {
```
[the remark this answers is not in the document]
djc: Changed it to `NextState` for consistency. Honestly, while I maybe see the point for the unwieldy `Result<Box<..>, Error>`, I sometimes find the aliases to be unhelpful in that they obscure what's actually there.

ctz:
Merged as d03bf27, thanks!

rustls/src/msgs/codec.rs
briansmith: I think the `from_over_into` lint is telling you to `impl From<u24> for usize` instead of of `impl Into<usize> for u24`; i.e. the same conversion you're currently doing, but in terms of `From`.

djc: Sure, but the reason given is that `From` impls have a blanket impl for the converse `Into` impl, which in this case seems undesirable to me.

briansmith: > Sure, but the reason given is that From impls have a blanket impl for the converse Into impl, which in this case seems undesirable to me.

No, it won't automatically generate the "converse." It will instead generate this exact `impl Into<usize> for u24`.

Basically, it's saying "Hey, it looks like you are implementing `Into<usize> for u24`  because the old coherence rules won't let you write `From<u24> for usize`, but the new coherence rules do let you do that, and that's more idiomatic."

Basically we should almost never implement `Into` directly in any case where it's possible to implement the equivalent `From` instead.

djc: Hmm, you're right, of course. Implemented this in #631.

PR #621, opened 2021-04-07
SessionID as an Option

sayrer:
Makes SessionID an Option in more places, but leaves it as-is in the Hello payloads. Also adds a fuzz test for an invalid SessionID.


sayrer:
One unsettled debate from the old PR:

>> I don't think ClientHelloPayload/ServerHelloPayload should have Option as a member, because by definition they always have a session ID inside (even if its empty). I don't think Option::None is a good representation for an empty session ID, since that's a valid session ID and, as you've found, it means we can't distinguish invalid and valid session ID encodings.

>Not sure I'm quite convinced by this argument. It seems to me that the RFCs repeatedly call out empty session IDs as special, and encoding this in the type system seems quite useful, and less error prone than relying on calls to is_empty() to check for this special case. I'm not exactly sure how/why we can't distinguish invalid and valid session ID encodings if we use None to represent the empty session ID; we probably shouldn't be using Option to encode invalid vs valid anyway.

I think this patch can go in, and adjusting the Hello payloads can be done in a follow-up if we decide to do that.

rustls/src/client/hs.rs
```diff
@@ -345,14 +345,12 @@ fn emit_client_hello_for_retry(
...
 
-    let session_id = session_id.unwrap_or(SessionID::empty());
-
     let mut chp = HandshakeMessagePayload {
         typ: HandshakeType::ClientHello,
         payload: HandshakePayload::ClientHello(ClientHelloPayload {
             client_version: ProtocolVersion::TLSv1_2,
             random: Random::from_slice(&randoms.client),
-            session_id: session_id,
+            session_id: session_id.unwrap_or(SessionID::empty()),
```
djc, line 353: This should use `unwrap_or_else(SessionID::empty)`.

rustls/src/msgs/handshake.rs
```diff
@@ -163,6 +163,14 @@ impl SessionID {
     pub fn is_empty(&self) -> bool {
         self.len == 0
     }
+
```
djc, line 166: Spurious empty line here.

rustls/src/server/hs.rs
```diff
@@ -369,7 +369,10 @@ impl ExpectClientHello {
                 payload: HandshakePayload::ServerHello(ServerHelloPayload {
                     legacy_version: ProtocolVersion::TLSv1_2,
                     random: Random::from_slice(&randoms.server),
-                    session_id: self.handshake.session_id,
+                    session_id: self
+                        .handshake
+                        .session_id
+                        .unwrap_or(SessionID::empty()),
```
djc, line 375: As above.

rustls/src/server/tls13.rs
```diff
@@ -116,7 +116,7 @@ impl CompleteClientHelloHandling {
                 payload: HandshakePayload::ServerHello(ServerHelloPayload {
                     legacy_version: ProtocolVersion::TLSv1_2,
                     random: Random::from_slice(&self.randoms.server),
-                    session_id: *session_id,
+                    session_id: session_id.unwrap_or(SessionID::empty()),
```
djc, line 119: And here.

djc (commented):
Some nits. I think this is looking good, but I think @ctz should be the one to merge it.

djc (approved):
Thanks for sticking with this!

ctz:
The changes here in client/ and server/ looks reasonable, but I think the changes in msgs/ are addidental?

sayrer:
So, if you undo the changes in msgs/ you end up unwinding a lot of the other things. I think either way is fine, and perhaps you were pointing out that there will be a SessionID in many parts of the handshake where I had listed as optional.

briansmith:
Here's what I think is confusing about this: In the TLS spec, SessionID allows an empty value, so if we're modeling the TLS type then SessionID should allow empty values. But elsewhere we want to have a strong, type-level distinction between the case where the SessionID is empty vs when it isn't.

First, it would be nice to clarify in the PR description and commit messages *why* we want this; i.e. what is the purpose/intent of this change? It seems to reference an old PR discussion but that's too much archeology, especially for people looking back in the commit history after this PR (presumably) gets merged.

The last time I looked at doing something similar to this, a few weeks ago, I had sketched something like this:

```
enum TicketOrLegacySessionID {
    Ticket(TicketWhatever),
    LegacySessionID(ResumableLegacySessionID),
}

/// A SessionID that is guaranteed to not be empty.
struct ResumableLegacySessionID(SessionID);
```

The problem I was working on at the time was that I wanted to eliminate the need to randomize the session ID field of a session to resume when using session tickets, when sending the initial client hello; `Option<TicketOrLegacySessionID>` models this precisely: Either resumption isn't relevant in this context, or we're resuming a ticket-based session (so generate a random throwaway session ID in the client hello) or we're doing legacy TLS 1.2 resumption.

Are you hoping to set up something similar with this change?

sayrer:
Yes, I was aiming for a type-level distinction where appropriate.

I did little bit of this work in the beginning of the client handshake already.

https://github.com/ctz/rustls/pull/584#discussion_r597858312

briansmith:
IIUC @ctz correctly, `SessionID` should admit an empty session ID, and the `SessionID` should be converted into an `Option<NonemptySessionID>` or whatever should be used whenever we want to have a strongly-typed distinction. That way there's no confusion. At least, that's what makes sense to me.

sayrer:
I thought a little bit more about this, and I don't think this patch is worth working on as a whole. I will rescue a few parts in smaller PRs. While I think @briansmith has made some good points here, I don't think I want to spend time working on code that bakes TLS 1.2 into the project's type-level code.

PR #622, opened 2021-04-07
Deduplicate code between client and server

djc:
Includes other changes that reduce data dependencies on the `Session` types, decrease coupling, and reduce the amount of code.


djc:
Note: I have followup changes that build on this which enable deduplicating the core state machine logic (process_main_protocol, process_handshake_messages and process_new_packets), as well as allowing the partial move of the handshake state (so that we no longer need to wrap it in Option).

ctz:
Nice, thanks!

PR #623, opened 2021-04-08
Remove HandshakeDetails from client code.

sayrer:
Perhaps `Option<persist::ClientSessionValueWithResolvedCipherSuite>` should have a type alias, not sure.


djc:
Nice, thanks!

PR #627, opened 2021-04-10
Refactor construction of AEAD keys for TLS 1.2 to clarify lack of panicking 

### LINKED ISSUE #626: Reduce explicit unwrapping in `new_tls12` and functions it calls
In private email, @djc mentioned an interest in making some of the code that `new_tls12` calls not panic. I looked into this.

Right now we have:

```
pub type BuildTls12Decrypter = fn(&[u8], &[u8]) -> Box<dyn MessageDecrypter>;
pub type BuildTls12Encrypter = fn(&[u8], &[u8], &[u8]) -> Box<dyn MessageEncrypter>;
```

Notice that these types are pretty bad because we can't tell what each argument is supposed to be.

If instead we had:
```
pub type BuildTls12Decrypter = fn(aead::LessSafeKey, &[u8]) -> Box<dyn MessageDecrypter>;
pub type BuildTls12Encrypter = fn(aead::LessSafeKey, &[u8], &[u8]) -> Box<dyn MessageEncrypter>;
```
then after the code is refactored we'd address @djc's original concern about unwraps within `BuildTls12Decrypter` and `BuildTls12Encrypter` implementations because the unwraps would be removed as part of this refactoring. This refactoring should be done regardless of anything below because it eliminates a bunch of duplicate code. Once that is done, those unwraps from the result of `UnboundKey::new()` would be moved into `new_tls12`. This will still be progress because there will be fewer unwraps due to the code de-duplication.

Once that refactoring is done, or instead of doing it, all the `BuildTls12Decrypter` implementations can probably be combined into a single one, and all the `BuildTls12Encrypter` implementations can be combined into a single one, I believe.

If you look at `new_tls12`, the way it is currently written is inherently panic prone because it has code like this:
```
    let key_block = secrets.make_key_block(scs.key_block_len());

    let mut offs = 0;

    let client_write_key = &key_block[offs..offs + scs.enc_key_len];
    offs += scs.enc_key_len;
    let server_write_key = &key_block[offs..offs + scs.enc_key_len];
    offs += scs.enc_key_len;
    let client_write_iv = &key_block[offs..offs + scs.fixed_iv_len];
    offs += scs.fixed_iv_len;
    let server_write_iv = &key_block[offs..offs + scs.fixed_iv_len];
    offs += scs.fixed_iv_len;
```

Notice that the array indexing will panic unless `secrets.make_key_block` creates a key block of the correct size (or larger). In turn this requires that it always return a key block of size `scs.key_block_len()`. There is at least one way to rewrite this code to completely avoid the possibility that the array indexing will fail. However, I don't see an easy way to avoid depending on `aead::Algorithm::key_len()` and `aead::UnbooundKey::new()` being consistent. If we rely on that consistency, i.e. accept that one call to `unwrap()` from the result of `UnboundKey::new()`, then we can avoid a lot of code to statically guarantee the key block splitting is correct.

Regardless, I wouldn't worry about dealing with that unless/until we refactor the key block generation/splitting code to statically avoid panicking.

briansmith:

Fixes #626.

### DIFF
--- a/rustls/src/cipher.rs
+++ b/rustls/src/cipher.rs
@@ -103,39 +103,45 @@ pub fn new_tls12(
     scs: &'static SupportedCipherSuite,
     secrets: &ConnectionSecrets,
 ) -> MessageCipherPair {
+    fn split_key<'a>(
+        key_block: &'a [u8],
+        alg: &'static aead::Algorithm,
+    ) -> (aead::LessSafeKey, &'a [u8]) {
+        // Might panic if the key block is too small.
+        let (key, rest) = key_block.split_at(alg.key_len());
+        // Won't panic because its only prerequisite is that `key` is `alg.key_len()` bytes long.
+        let key = aead::UnboundKey::new(alg, key).unwrap();
+        (aead::LessSafeKey::new(key), rest)
+    }
+
     // Make a key block, and chop it up.
     // nb. we don't implement any ciphersuites with nonzero mac_key_len.
-    let key_block = secrets.make_key_block(scs.key_block_len());
+    let key_block = secrets.make_key_block(scs);
 
-    let mut offs = 0;
-    let client_write_key = &key_block[offs..offs + scs.enc_key_len];
-    offs += scs.enc_key_len;
-    let server_write_key = &key_block[offs..offs + scs.enc_key_len];
-    offs += scs.enc_key_len;
-    let client_write_iv = &key_block[offs..offs + scs.fixed_iv_len];
-    offs += scs.fixed_iv_len;
-    let server_write_iv = &key_block[offs..offs + scs.fixed_iv_len];
-    offs += scs.fixed_iv_len;
+    let (client_write_key, key_block) = split_key(&key_block, scs.aead_algorithm);
+    let (server_write_key, key_block) = split_key(&key_block, scs.aead_algorithm);
+    let (client_write_iv, key_block) = key_block.split_at(scs.fixed_iv_len);
+    let (server_write_iv, extra) = key_block.split_at(scs.fixed_iv_len);
 
-    let (write_key, write_iv) = if secrets.randoms.we_are_client {
-        (client_write_key, client_write_iv)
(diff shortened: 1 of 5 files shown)


briansmith:
Fixes #626.

rustls/src/cipher.rs
```diff
@@ -117,35 +103,45 @@ pub fn new_tls12(
...
-    let (write_key, write_iv) = if secrets.randoms.we_are_client {
-        (client_write_key, client_write_iv)
+    let key_block = secrets.make_key_block(scs);
+
+    let (client_write_key, key_block) = split_key(&key_block, scs.aead_algorithm);
+    let (server_write_key, key_block) = split_key(&key_block, scs.aead_algorithm);
+    let (client_write_iv, key_block) = key_block.split_at(scs.fixed_iv_len);
+    let (server_write_iv, extra) = key_block.split_at(scs.fixed_iv_len);
+
+    let (write_key, write_iv, read_key, read_iv) = if secrets.randoms.we_are_client {
```
briansmith, line 126: I had to combine both of these `if secrets.randoms.we_are_client` blocks because of borrow checking reasons.

rustls/src/conn.rs
```diff
@@ -343,7 +343,9 @@ impl ConnectionSecrets {
         ret
     }
 
-    pub fn make_key_block(&self, len: usize) -> Vec<u8> {
+    pub fn make_key_block(&self, scs: &SupportedCipherSuite) -> Vec<u8> {
```
briansmith, line 346: As a follow-up to this PR, we should probably combine `make_key_block` and `new_tls12` to make it clearer than the key block is the right size.

rustls/tests/api.rs
```diff
@@ -611,7 +611,7 @@ impl ClientCheckCertResolve {
     fn new(expect_queries: usize) -> ClientCheckCertResolve {
         ClientCheckCertResolve {
             query_count: AtomicUsize::new(0),
-            expect_queries: expect_queries,
```
briansmith, line 614: rustfmt did this, I guess. I used the version `admin/fmt` built.

ctz:
Thanks!

PR #636, opened 2021-04-13
Fuzz test: Invalid SessionID in ServerHelloPayload.

sayrer:
Found this while working on a PR we decided against. Still good to add the test coverage.


djc:
Thanks!

PR #638, opened 2021-04-13
server: simplify SNI handling and make it more robust

djc:
Replaces an assertion with an `Error` and removes about 50 lines of code (net).

### DIFF
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -77,29 +77,12 @@ pub fn can_resume(
 
     if resumedata.cipher_suite == suite.suite
         && (resumedata.extended_ms == using_ems || (resumedata.extended_ms && !using_ems))
-        && same_dns_name_or_both_none(resumedata.sni.as_ref(), sni.as_ref())
+        && &resumedata.sni == sni
     {
         return Some(resumedata);
     }
 
     None
-}
-
-// Require an exact match for the purpose of comparing SNI DNS Names from two
-// client hellos, even though a case-insensitive comparison might also be OK.
-pub(super) fn same_dns_name_or_both_none(
-    a: Option<&webpki::DnsName>,
-    b: Option<&webpki::DnsName>,
-) -> bool {
-    match (a, b) {
-        (Some(a), Some(b)) => {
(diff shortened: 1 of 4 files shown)


ctz:
Thanks!

rustls/src/server/hs.rs
```diff
@@ -65,41 +65,18 @@ pub fn can_resume(
...
-        && same_dns_name_or_both_none(resumedata.sni.as_ref(), sni.as_ref())
-    {
-        return Some(resumedata);
-    }
-
-    None
-}
-
-// Require an exact match for the purpose of comparing SNI DNS Names from two
-// client hellos, even though a case-insensitive comparison might also be OK.
```
briansmith, line 89: To the extent webpki implements PartialEq for DNS names, it does so in a case-insensitive manner. Rustls was, prior this PR, using a case-sensitive comparison. If this is an intentional change then shouldn't we have a test that verifies that we're doing a case-insensitive comparison now? Otherwise, perhaps this aspect should be reverted? (If reverted, the implementation of this function could be made simpler.)

rustls/src/server/hs.rs
```diff
@@ -414,9 +384,16 @@ impl State for ExpectClientHello {
...
-        if !self.done_retry {
-            // save only the first SNI
-            save_sni(conn, sni.clone());
+        // save only the first SNI
+        if let (Some(sni), false) = (&sni, self.done_retry) {
+            // Save the SNI into the session.
+            // The SNI hostname is immutable once set.
+            assert!(conn.sni.is_none());
+            conn.sni = Some(sni.clone());
+        } else if conn.sni != sni {
```
briansmith, line 393: This is a case-insensitive comparison, right? Do you mean to check that the SNI didn't change at all?

rustls/src/server/hs.rs
djc: The change was not intentional. I guess we should maybe revert this part.

rustls/src/server/hs.rs
djc: I guess so.

Issue #640, opened 2021-04-15
KeyExchange::client_ecdhe does not check for unparsed trailing input


briansmith:
The analogous code for the server, in `decode_client_params` looks like:
```
        let mut rd = Reader::init(kx_params);
        let ecdh_params = ClientECDHParams::read(&mut rd).unwrap();
        if rd.any_left() {
            None
        } else {
            Some(ecdh_params)
        }
```

However, the client's code in `client_ecdhe` is missing the `any_left()` check.

briansmith:
Fixed in #644.

PR #644, opened 2021-04-15
kx: deduplicate ECDH params decoding

### LINKED ISSUE #641: Malformed ECDHE parameters from client may cause panic
I didn't verify this experimentally; I just read the code:

```
    fn decode_client_params(&self, kx_params: &[u8]) -> Option<ClientECDHParams> {
        let mut rd = Reader::init(kx_params);
        let ecdh_params = ClientECDHParams::read(&mut rd).unwrap();
        if rd.any_left() {
            None
        } else {
            Some(ecdh_params)
        }
    }
```
It seems like the `unwrap()` is reachable, if there's a syntax error in the ClientECDHParams. Either way, it seems like the `unwrap()` can easily be replaced with `?`.

djc:
Fixes #641 and #640.

### DIFF
--- a/rustls/src/kx.rs
+++ b/rustls/src/kx.rs
@@ -26,9 +26,7 @@ impl KeyExchange {
         kx_params: &[u8],
         supported: &[&'static SupportedKxGroup],
     ) -> Option<KeyExchangeResult> {
-        let mut rd = Reader::init(kx_params);
-        let ecdh_params = ServerECDHParams::read(&mut rd)?;
-
+        let ecdh_params = Self::decode_ecdh_params::<ServerECDHParams>(kx_params)?;
         KeyExchange::choose(ecdh_params.curve_params.named_group, supported)
             .and_then(KeyExchange::start)
             .and_then(|kx| kx.complete(&ecdh_params.public.0))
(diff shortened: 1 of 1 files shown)


briansmith:
Thanks for doing this so quickly. I agree this is a good approach. I would expect a PR that fixes #641 would contain a test that would fail before the fix and pass after the fix.

rustls/src/kx.rs
```diff
@@ -67,20 +65,19 @@ impl KeyExchange {
...
         let mut rd = Reader::init(kx_params);
-        let ecdh_params = ClientECDHParams::read(&mut rd).unwrap();
-        if rd.any_left() {
-            None
-        } else {
-            Some(ecdh_params)
+        let ecdh_params = T::read(&mut rd)?;
+        match rd.any_left() {
+            false => Some(ecdh_params),
+            true => None,
```
briansmith, line 73: I was expecting that codecov would report this line as uncovered by tests, but it actually reports it is covered already. I guess the test for this is somewhere in the server testing code and so maybe we should copy-pasta it, or otherwise refactor it, so it is testing the client too?

djc: I wonder, should we report these kinds of minor coverage gaps in bogo, too?

briansmith: IDK. The issue tracker for BoringSSL is https://bugs.chromium.org/p/boringssl/issues but I don't think filing an issue is very productive. Perhaps writing a PR for adding the test would be more effective, especially since Rustls is already using a fork (currently with no changes?) of BoGo.

djc: For now I just added some unit tests that cover #640 and #641.

briansmith (approved):

PR #646, opened 2021-04-15
Move some unit test modules into their own files to make static analysis easier

briansmith:
It's nice to exclude "*_test.rs" from static analysis and code search in general.

Split out many test module that contain `unwrap()`, but not all, to make it easier to prioritize work on removing panics in Rustls.


djc:
Hmm, I'm honestly not a fan of creating new 10-line modules just to make triage easier. In the triage file I kept I noted down the starting line of tests for a bunch of these modules so help make it easier between library and test code. Perhaps your automated triage thing could split the file on `#[cfg(test)]` before grepping through it, or something?

briansmith:
OK.

PR #647, opened 2021-04-15
Implement an infallible version of `Payload::read()` to reduce unwraps.

briansmith:


rustls/src/msgs/base.rs
```diff
@@ -23,6 +23,10 @@ impl Payload {
     pub fn empty() -> Payload {
         Payload::new(Vec::new())
     }
+
+    pub fn read(r: &mut Reader) -> Self {
```
djc, line 27: Maybe a different name would make the distinction between this and `Codec::read()` more obvious?

djc (commented):
Nice!

rustls/src/msgs/base.rs
briansmith: I think it's fine this way, but if you suggest a better name I'll change it. read_infallible?

djc: Maybe `Codec::read()` should become `try_read()` (though that should go in a different PR). What do you think?

PR #649, opened 2021-04-16
Client: Fix client key share selection during HelloRetryRequest.

briansmith:
During a HelloRetryRequest (HRR), if the HRR didn't request a different algorithm, then
we must send the same key share we sent before. During a retry we must never look up the
key share algorithm to use from the cache (`.or_else(|| find_kx_hint(conn, dns_name)`) in
the old code), because the cache contents may have changed since the initial HRR. Only
`emit_initial_client_hello` is allowed to read from the cache. (In a future refactoring, we
should enforce this with the type system.)

It turns out that the selection criteria for shares is completely disjoint between
initial and retry hellos. Instead of grouping the two code paths into one function, split
them into separate code paths.

When we are selecting the initial key share, and we find a the server's selected key share
grou pin the cache, and that group is not enabled for the current group, then fall back to
the default algorithm. Previously, we just kept going without any key share! Similarly, if
agreement key generation fails, stop with an error instead of siliently continuing without
sending a key share(!):

```diff
- else if let Some(key_share) =
-        kx::KeyExchange::choose(group, &conn.config.kx_groups).and_then(kx::KeyExchange::start)
```

We either send one key share (if TLS 1.3 is enabled) or zero (if it isn't). Encode this in
the types used, instead of using `Vec`s.

Move the offered key share out of `ClientHelloDetails` so that we can move it during HRR
processing. (In the future we should remove `ClientHelloDetails`.)

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -1,8 +1,6 @@
-use crate::kx;
 #[cfg(feature = "logging")]
 use crate::log::trace;
 use crate::msgs::enums::ExtensionType;
-use crate::msgs::enums::NamedGroup;
 use crate::msgs::handshake::CertificatePayload;
 use crate::msgs::handshake::DigitallySignedStruct;
 use crate::msgs::handshake::SCTList;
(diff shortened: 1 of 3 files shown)


briansmith:
This is still a WIP since I haven't written the tests for the cases that seem to have been wrongly handled.

ctz:
> Previously, we just kept going without any key share! Similarly, if
agreement key generation fails, stop with an error instead of siliently continuing without
sending a key share(!):

While this isn't ideal, it's in-spec and requires the server to send a HRR in response. I have a test that exercises this case now, so will merge this.

ctz:
Merged as 0a86815d1 et al.

PR #650, opened 2021-04-17
Stop panicking during key agreement on system RNG failure.

briansmith:
Report failure to generate a key pair as an RNG failure instead of as peer misbehavior.

Refactor the TLS 1.2 client code to inline the key agreement logic into the caller to facilitate this.

Improve related error reporting to fix misleading error codes, to be consistent with alerting, and to be consistent with error codes and messages. In the case of TLS 1.2, err on the side of reporting the same errors as TLS 1.3 would if the choice is unclear. Update the expected results for a few BoGo tests accordingly.

PR #649 separately deals with the case for TLS 1.3 clients, as that code is more problematic and requires a larger refactoring.


rustls/src/tls12.rs
```diff
@@ -0,0 +1,53 @@
+use crate::conn::ConnectionCommon;
+use crate::kx;
+use crate::msgs::codec::{Codec, Reader};
+use crate::msgs::enums::{AlertDescription, ContentType};
+use crate::Error;
+
+pub fn decode_ecdh_params<T: Codec>(
```
briansmith, line 7: I moved this logic to this new module because it seems inappropriate for the `kx` module to depend on `ConnectionCommon`.

ctz:
Thanks!

phbno1:
Can this issue be picked to other branches?@briansmith

ctz:
Which branches were you thinking about? This was first released in 0.20.0, and releases older than this won't receive security fixes (see https://github.com/rustls/rustls/blob/main/SECURITY.md).

phbno1:
> Which branches were you thinking about? This was first released in 0.20.0, and releases older than this won't receive security fixes (see https://github.com/rustls/rustls/blob/main/SECURITY.md).
We encountered this problem in versions 0.17.0 and 0.19.0, which caused program panics. The code of these versions is very different from that of 0.20.0.

phbno1:
> Why are you still using these old versions?

The upgrade brings great changes. We can only fix this problem by using patches.

djc:
Who's we? Would you be able to pay for extended support?

PR #652, opened 2021-04-17
Abort connection if no QUIC transport parameters are found

djc:
cc @Ralith

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -958,6 +958,12 @@ impl ConnectionCommon {
             .prepare_message_decrypter(dec);
     }
 
+    #[cfg(feature = "quic")]
+    pub fn missing_extension(&mut self, why: &str) -> Error {
+        self.send_fatal_alert(AlertDescription::MissingExtension);
+        Error::PeerMisbehavedError(why.to_string())
+    }
+
     pub fn send_warning_alert(&mut self, desc: AlertDescription) {
         warn!("Sending warning alert {:?}", desc);
         self.send_warning_alert_no_log(desc);
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -32,6 +32,12 @@ impl Secrets {
 /// Generic methods for QUIC sessions
 pub trait QuicExt {
     /// Return the TLS-encoded transport parameters for the session's peer.
+    ///
+    /// While the transport parameters are technically available prior to the
+    /// completion of the handshake, they cannot be fully trusted until the
+    /// handshake completes, and reliance on them should be minimized.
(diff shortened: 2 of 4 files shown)


rustls/src/conn.rs
```diff
@@ -958,6 +958,12 @@ impl ConnectionCommon {
             .prepare_message_decrypter(dec);
     }
 
+    #[cfg(feature = "quic")]
+    pub fn missing_extension(&mut self, why: &str) -> Error {
+        self.send_fatal_alert(AlertDescription::MissingExtension);
+        Error::PeerIncompatibleError(why.to_string())
```
briansmith, line 964: Seems like peer misbehavior, since the extensions are required for QUIC to work, right?

djc: Yup, copy/pasta error.

rustls/src/quic.rs
```diff
@@ -32,6 +32,12 @@ impl Secrets {
 /// Generic methods for QUIC sessions
 pub trait QuicExt {
     /// Return the TLS-encoded transport parameters for the session's peer.
+    ///
+    /// While the transport parameters are technically available prior to the
+    /// completion of the handshake, they cannot be fully trusted until the
+    /// handshake completes, and reliance on them should be minimized.
+    /// However, any tampering with the parameters will cause the handshake
+    /// to fail.
     fn get_quic_transport_parameters(&self) -> Option<&[u8]>;
```
briansmith, line 41: We should file a separate issue for this: We should refactor this API to minimize the danger here.

djc: Yeah, I wanted to get @Ralith's eyes on it before changing it. While reviewing code in Quinn, it did seem like it relies on being able to access this early. (The change I had in mind was just to return `None` while `self.is_handshaking()`.)

briansmith (approved):
Not sure what our QUIC testing situation is. Seems like we should improve that so we can write tests for things like this.

Ralith:
See `test_quic` in rustls/tests/api.rs for current testing. We indeed do not have coverage for this case.

rustls/src/quic.rs
Ralith: Responded in #655.

Ralith (approved):
This is consistent with the QUIC spec. We're currently handling this case explicitly in Quinn, but I think moving it into rustls is a bit nicer because this way quinn's ostensibly crypto-agnostic code doesn't need a hardcoded TLS alert number.

djc:
Going to close this in favor of a new PR that adds tests and address some other issues.

djc:
See #666.

PR #653, opened 2021-04-17
Avoid panicking when System PRNG fails during ticket encryption.

briansmith:


djc:
What is the reason `seal_in_place_separate_tag()` is fallible?

I really dislike the pervasive convention that `Option::None` can mean invalid.

briansmith:
> What is the reason `seal_in_place_separate_tag()` is fallible?

If you ask it to encrypt too large of a value, it will fail. Each AEAD has a limit on how much can be encrypted per nonce.

briansmith:
> I really dislike the pervasive convention that `Option::None` can mean invalid.

I mostly agree, and I hope we address that in the (near) future.

PR #660, opened 2021-04-19
Further deduplicate client/server session handling

djc:
We go through the following steps:

- Move side-specific data fields into a separate type (`*ConnectionData`)
- Introduce side-specific context types that are passed to `State::handle()`, to prevent borrowing all of the `*Connection`
- Introduce a new trait (currently called `HandleState`) that allows the `ConnectionCommon` to run through the state machine with minimal knowledge of the underlying types
- Move core state machine driver methods into `ConnectionCommon`

With the following advantages:

* Core state machine driving logic is no longer duplicated
* Privatize more of the `ConnectionCommon` interface
* ~90 lines less code
* Keep client and server state machine types separate

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -115,14 +115,14 @@ impl InitialState {
         }
     }
 
-    fn emit_initial_client_hello(mut self, conn: &mut ClientContext<'_>) -> NextStateOrError {
+    fn emit_initial_client_hello(mut self, cx: &mut ClientContext<'_>) -> NextStateOrError {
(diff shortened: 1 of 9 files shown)


djc:
@ctz thanks for cleaning up those binding names I missed. As a potential workflow optimization (being conscious of your time/effort), you should also feel free to approve my PRs conditional on my fixing up anything you noticed during review (whether general -- "fix up the comments to match" or specific -- "you missed this rename"). Since I can technically merge my own PRs, that way I can deal with rebasing and fixing up anything I missed and you don't have to bother with it.

PR #661, opened 2021-04-19
server: restore exact comparison of SNI values

djc:
As discussed in #638.

### DIFF
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -72,7 +72,22 @@ pub fn can_resume(
     // establish a new session."
     resumedata.cipher_suite == suite.suite
         && (resumedata.extended_ms == using_ems || (resumedata.extended_ms && !using_ems))
-        && &resumedata.sni == sni
+        && sni_exact_eq(&resumedata.sni, &sni)
+}
+
+/// Exactly compare `DnsName` values
+///
+/// webpki compares `DnsName`s case-insensitively, but for the purpose of comparing SNI values
+/// from resumption to current handshake or HelloRetryRequest to ServerHello, we need an exact
+/// comparison.
+pub(super) fn sni_exact_eq(a: &Option<webpki::DnsName>, b: &Option<webpki::DnsName>) -> bool {
+    let a = a
+        .as_ref()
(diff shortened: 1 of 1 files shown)


rustls/src/server/hs.rs
```diff
@@ -72,7 +72,22 @@ pub fn can_resume(
...
-        && &resumedata.sni == sni
+        && sni_exact_eq(&resumedata.sni, &sni)
+}
+
+/// Exactly compare `DnsName` values
+///
+/// webpki compares `DnsName`s case-insensitively, but for the purpose of comparing SNI values
+/// from resumption to current handshake or HelloRetryRequest to ServerHello, we need an exact
+/// comparison.
+pub(super) fn sni_exact_eq(a: &Option<webpki::DnsName>, b: &Option<webpki::DnsName>) -> bool {
```
briansmith, line 83: This function needs to take two `DnsNameRef`s. `DnsName`s are already case-normalized when constructed.

rustls/src/server/hs.rs
```diff
@@ -369,7 +384,7 @@ impl State for ExpectClientHello {
             // The SNI hostname is immutable once set.
             assert!(conn.sni.is_none());
             conn.sni = Some(sni.clone());
-        } else if conn.sni != sni {
+        } else if !sni_exact_eq(&conn.sni, &sni) {
```
briansmith, line 387: You need to do this comparison before you convert `sni` to `webpki::DnsName`. if it is important that we do this case-sensitively then we should have a test that would have found this issue.

djc: I don't know if it is important. What do you think?

ctz: I don't think this is important, and certainly not important enough to warrant completely fixing this. Ultimately `conn.sni` here is a `webpki:DnsName` and so has already been case-smashed.

Because we only use case-smashed names (eg, for certificate resolution, the public API, etc.), neither the ability to resume as session with a SNI that differs by case, nor varying the case in a HRR would lead to any real problem.

This also wasn't changed in #638.

djc:
I'll just close this, then.

PR #662, opened 2021-04-19
Use type system to ensure TLS 1.3 cipher suites aren't used for TLS 1.2

briansmith:

### DIFF
--- a/rustls/src/cipher.rs
+++ b/rustls/src/cipher.rs
@@ -99,10 +99,7 @@ pub fn build_tls12_chacha_encrypter(
 
 /// Make a `MessageCipherPair` based on the given supported ciphersuite `scs`,
 /// and the session's `secrets`.
-pub fn new_tls12(
-    scs: &'static SupportedCipherSuite,
-    secrets: &ConnectionSecrets,
-) -> MessageCipherPair {
+pub fn new_tls12(secrets: &ConnectionSecrets) -> MessageCipherPair {
     fn split_key<'a>(
         key_block: &'a [u8],
         alg: &'static aead::Algorithm,
@@ -116,7 +113,10 @@ pub fn new_tls12(
 
     // Make a key block, and chop it up.
     // nb. we don't implement any ciphersuites with nonzero mac_key_len.
-    let key_block = secrets.make_key_block(scs);
+    let key_block = secrets.make_key_block();
+
+    let suite = secrets.suite();
+    let scs = suite.scs();
 
     let (client_write_key, key_block) = split_key(&key_block, scs.aead_algorithm);
     let (server_write_key, key_block) = split_key(&key_block, scs.aead_algorithm);
@@ -140,8 +140,8 @@ pub fn new_tls12(
     };
 
     (
-        scs.build_tls12_decrypter.unwrap()(read_key, read_iv),
-        scs.build_tls12_encrypter.unwrap()(write_key, write_iv, extra),
+        (suite.tls12().build_tls12_decrypter)(read_key, read_iv),
+        (suite.tls12().build_tls12_encrypter)(write_key, write_iv, extra),
     )
 }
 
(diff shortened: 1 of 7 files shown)


rustls/src/client/hs.rs
```diff
@@ -576,6 +572,12 @@ impl State for ExpectServerHello {
         // For TLS1.3, start message encryption using
         // handshake_traffic_secret.
         if conn.common.is_tls13() {
+            if !suite.usable_for_version(TLSv1_3) {
+                return Err(conn
+                    .common
+                    .illegal_param("server chose unusable ciphersuite for version"));
+            }
+
```
briansmith, line 580: In the Future we should introduce a similar `Tls13CipherSuite` type that is used for in the TLS 1.3 state machine for similar reasons.

djc:
Instead of `Tls12CipherSuite` wrapping `SupportedCipherSuite`, did you consider doing it the other way around? If we're going down this road, I think we should end up like this:

```rust
enum SupportedCipherSuite {
    Tls12(Tls12CipherSuite),
    Tls13(Tls13CipherSuite),
}

struct Tls13CipherSuite {
    algs: CipherSuiteAlgorithms,
    hkdf_algorithm: ring::hkdf::Algorithm,
}

struct Tls12CipherSuite {
    algs: CipherSuiteAlgorithms,
    kx: KeyExchangeAlgorithm,
    .. etc
}

struct CipherSuiteAlgorithms {
    suite: CipherSuite,
    bulk: BulkAlgorithm,
    .. etc
}
```

I think this would make it easier to make TLS 1.2 support optional down the line. This also gets rid of the remaining option field in `SupportedCipherSuite`. It also looks like `hkdf_algorithm` is only used for TLS 1.3, so it would make sense to make that clear in the type system as well (and save the memory).

briansmith:
@djc I think that's probably a good idea, but I feel like that's expanding the scope too much beyond the initial intent. The goal here is just to make it clear that we never hit the `unwrap()`s that are removed. I think a change like you're describing could easily be built on top of this one, and more easily than starting from the previous code.

rustls/src/cipher.rs
```diff
@@ -140,8 +140,8 @@ pub fn new_tls12(
     };
 
     (
-        scs.build_tls12_decrypter.unwrap()(read_key, read_iv),
-        scs.build_tls12_encrypter.unwrap()(write_key, write_iv, extra),
```
briansmith, line 144: These are the two unwraps that I am seeking to remove with this PR.

rustls/src/server/hs.rs
```diff
@@ -476,6 +477,7 @@ impl State for ExpectClientHello {
             }
             .handle_client_hello(suite, conn, certkey, &m)
         } else {
+            let suite = suites::Tls12CipherSuite::try_from(suite).unwrap();
```
briansmith, line 480: In the future, the selection logic for `suitable_suites` should be refactored to eliminate this unwrap.

djc:
While I agree that this PR makes it easier to execute on those changes, I kind of feel like stopping at this point is similar to when you review my PRs asking me to provide tests -- in both cases, the change is an improvement, but also in both cases, there are obvious improvements that could be made with not, in my opinion, that much more effort. (I understand that there are differences to these two examples, but it still seems like a useful analogy.)

Anyway, I'm happy to merge this PR, except for one nit: let's rename the `scs()` method to something a little easier to understand. My proposal would be `supported()` or `supported_suite()`.

briansmith:
> While I agree that this PR makes it easier to execute on those changes, I kind of feel like stopping at this point is similar to when you review my PRs asking me to provide tests -- in both cases, the change is an improvement, but also in both cases, there are obvious improvements that could be made with not, in my opinion, that much more effort.

My preference is to state a goal and then get to that goal with a minimal amount of changes. In this case, the goal is to clarify the `unwrap()`s I removed will not be hit, because they were far from the code that guaranteed that. This way we should feel confident saying that the eliminating the remaining (new) `unwrap()` is a low priority. Going beyond that isn't a priority for me until after the next release.

> Anyway, I'm happy to merge this PR, except for one nit: let's rename the `scs()` method to something a little easier to understand. My proposal would be `supported()` or `supported_suite()`.

Done.

PR #663, opened 2021-04-19
WIP: ECH Draft-10

sayrer:


sayrer:
I removed the minimal-versions check temporarily, since that error is being caused by the "hpke" crate. I don't plan to leave that dependency in the end, but I want to be able to spot other errors easily.

briansmith:
@sayrer Are you wanting feedback on this? Not sure what your intent is RE the "WIP:" prefix.

sayrer:
> @sayrer Are you wanting feedback on this? Not sure what your intent is RE the "WIP:" prefix.

No, not yet. In particular, the `ech_key.rs` file is just for bootstrapping from other open-source implementations.

sayrer:
I switched this to hpke-rs for now, it's a bit easier to use since it uses the enum-heavy style that Rustls does. The hpke crate worked fine, but I would've had to implement some macros we won't end up using. Also using hpke-rs with the "hazmat" feature that lets you do things get private key bytes. This will be good for testing during development, but I don't expect ring will support anything like this.

sayrer:
I'm a little mystified by this Catalina test failure. It passes here on Big Sur, and also seems to pass on Linux (looking through the logs).

sayrer:
This latest patch is only failing on code coverage metrics. I am not sure what's going on with the macOS tests--I think I'll wait it out since they pass on my local Macs.

The next step is to work through the ClientHello transformations required by the spec, and then sign that plaintext and aad. This patch shows that HPKE itself works when instantiated from the TLS wire format.

sayrer:
Looks like the GitHub CI machines are pretty old--they don't have AES-NI and/or AVX instructions, which Evercrypt requires.

briansmith:
> Looks like the GitHub CI machines are pretty old--they don't have AES-NI and/or AVX instructions, which Evercrypt requires.

I believe M1 Macs don't have AVX either in their x86 emulation.

sayrer:
> > Looks like the GitHub CI machines are pretty old--they don't have AES-NI and/or AVX instructions, which Evercrypt requires.
> 
> I believe M1 Macs don't have AVX either in their x86 emulation.

Could be that, too. Turns out there are also some [AVX bugs](https://gitlab.stce.rwth-aachen.de/stce/eigen-ad/commit/71aa53dd6dfdc497324d9e87f59c4ba820191856#dbe24dc3c0e6300338d7276629e7e71aa09f2935_287_286) that I'm not sure are fixed when building on 10.15.7.

briansmith:
@sayrer How is this going? Is there any other prerequisite refactoring work that would be useful for this?

sayrer:
> @sayrer How is this going? Is there any other prerequisite refactoring work that would be useful for this?

It's going fine--I think the handshake is ready for it. I'm just going to work in ech.rs until I understand and test all of the mechanisms. Next step is to do ClientHelloOuterAAD, then actually seal the ClientHelloInner, which I'm currently working on.

Note that this work won't be possible to finish until the WG finalizes its approach to HRR.
https://mailarchive.ietf.org/arch/msg/tls/mRGvQ26RzRIvbw4Od2VENQBrorc/

briansmith:
OK, great to hear. I am not trying to be pushy about getting this done, BTW. I'm just interested to know if you've identified any other refactorings we should be doing.

sayrer:
Added a demo that connects to Cloudflare. It doesn't work yet, because I haven't used the ClienHellotInner/ClientHelloOuter code in the handshake yet.

```
Negotiated ciphersuite: TLS13_AES_256_GCM_SHA384
read bytes: 111
HTTP/1.1 421 Misdirected Request
Date: Tue, 18 May 2021 19:39:14 GMT
Content-Length: 0
Connection: close

```

sayrer:
```
% cargo run --example tls_ech_client
Error: Custom { kind: InvalidData, error: PeerMisbehavedError("server sent unsolicited encrypted extension") }
```

I can add EncryptedClientHello to be expected, and get past this. I think the next step is to operate on `tls13::handle_server_hello`.

sayrer:
It looks like the server is accepting the ECH now. There's some work to do in handle_server_hello to update the transcript upon acceptance. Getting this after that point:

```
Error: Custom { kind: InvalidData, error: DecryptError }
```

sayrer:
It looks like there are two issues now:

1) The spec isn't very precise about what goes in the transcript, so I am writing unit tests for this. I've tried to write it a few times, but the EncodedClientHelloInner and what's in the transcript are not the same. It seems to just vary by message envelopes etc.

2) In order to confirm ECH was accepted, the first 8 bytes of a ring::Prk need to be examined.  see: https://tlswg.org/draft-ietf-tls-esni/draft-ietf-tls-esni.html#section-7.2.

sayrer:
OK, I got this to work, at least for non-HRR requests (the code is still messy). The issue was that the ECH transcript is calculated from an unsent inner ClientHello message, which differs in a few ways from the EncodedClientHelloInner used in HPKE. Upon ECH acceptance, the transcript from the ClientHelloOuter is discarded, then substituted with one calculated from the unsent inner ClientHello message, and the client random is changed to inner random value. After that, the ServerHello message is added to the transcript as normal.

sayrer:
The draft-10 patch will have some refactors concerning ExpectServerHello::handle and tls13::handle_server_hello. I don't think they will be necessary in later drafts due to this change: https://github.com/tlswg/draft-ietf-tls-esni/pull/420/files

sayrer:
> * The spec isn't very precise about what goes in the transcript, so I am writing unit tests for this. I've tried to write it a few times, but the EncodedClientHelloInner and what's in the transcript are not the same. It seems to just vary by message envelopes etc.
> 
> * In order to confirm ECH was accepted, the first 8 bytes of a ring::Prk need to be examined.  see: https://tlswg.org/draft-ietf-tls-esni/draft-ietf-tls-esni.html#section-7.2.

I got these two things working, so I'm going to add HRR support, more tests, and clean this patch up enough to be ready for feedback. I don't think it will make sense to merge, because the draft-10 design is invasive to the client handshake in ways that draft-11 won't be. I'll add some notes on that via comments. After that, I'll work on the server.

rustls/src/key_schedule.rs
```diff
@@ -177,6 +180,14 @@ impl KeyScheduleHandshake {
         secret
     }
 
+    pub fn server_ech_confirmation_secret(&mut self, hs_hash: &Digest) -> PayloadU8 {
```
briansmith, line 183: The idea behind the the *ring* HKDF API is that you would do something like this:

```
pub struct Iv([u8; aead::NONCE_LEN]);
struct IvLen;

impl hkdf::KeyType for IvLen {
    fn len(&self) -> usize {
        aead::NONCE_LEN
    }
}
```

s/`Iv`/`EchConfirmationSecret`/.

sayrer: I don't think this code will live on in the final implementation, given https://github.com/tlswg/draft-ietf-tls-esni/pull/420/files, so I just did the easiest thing for interop testing in draft-10.

briansmith: Also, I would say all the existing uses of `PayloadU8Len` are unfortunate, so I wouldn't copy them. `derive_ticket_psk` in particular should not be using `Vec<u8>` as its return type, but instead it should use a stronger key type.

Ideally `export_keying_material` would have a more sophisticated API as well, more like *ring*'s HKDF API; I think the desire to avoid *ring* types in Rust's public API is what makes that difficult.

OTOH, it probably it makes more sense for `server_ech_confirmation_secret` to compute the return the entire `Random` including the 24 random bytes and the 8 derived bytes.

sayrer: To add to my above comment, I think the draft-11 ECH confirmation will be able to happen entirely in client/hs.rs. The draft-10 design requires changes to tls13.rs and key_schedule.rs, but I don't think those will be needed in the future.

cjpatton:
I'm curious if anyone plans to pick this up in the next few months. By the way, the current version is draft-13, which (we hope!) will be simpler to implement on the client side.

cpu:
I think we should close this PR to reflect that it is:

i. Not under active development by anyone at this time
ii. Has substantial conflicts with `main`
iii. Is out-of-date with the current ECH drafts

I think it would be great to revive the ECH effort, but leaving this draft PR open doesn't feel like it will help advance that goal. It also comes at the cost of distracting from other work closer to being mergeable in the PR queue.

djc:
> I think it would be great to revive the ECH effort, but leaving this draft PR open doesn't feel like it will help advance that goal.

Agreed.

> It also comes at the cost of distracting from other work closer to being mergeable in the PR queue.

This doesn't feel like a big win by itself...

cpu:
> This doesn't feel like a big win by itself...

For one PR in isolation it's not a huge win, but as more of them enter into this state the clutter gives the appearance of an unmaintained project. This branch in particular hasn't had a commit since 2021, it's very stale.

In either case, reopening closed PRs if someone feels differently is practically free ;-)

cjpatton:
I support closing. The protocol (not to mention rustls itself) has changed so much since draft-10 that I think it would be worth starting fresh.

PR #666, opened 2021-04-21
QUIC: improve error handling

### LINKED ISSUE #555: Panic when TLS 1.2 is enabled for a QUIC connection
In https://github.com/ctz/rustls/issues/549#issuecomment-796705388 @djc wrote:
> Starting new QUIC sessions currently panics if any version other than 1.3 is in config.versions.

There shouldn't be a panic here.

When starting a QUIC session, Rustls should just disable TLS 1.2 for the session and carry on.

djc:
Fixes #555.

This adds a bunch of tests for these new error paths. I didn't figure out a way to test that the client will error out when it receives an `EncryptedExtensions` message from the server that does not include the QUIC transport parameters.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,7 +210,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -p rustls -- -D warnings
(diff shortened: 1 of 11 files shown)


ctz:
Thanks, merged in f0b7bc58

PR #669, opened 2021-04-21
Eliminate one allocation and one copy in handshake message deframing.

briansmith:


briansmith:
I verified that the `swap` case is hit almost exclusively during the test suite.

djc:
Since payload is dead there, I think we can just assign payload.0 to self.buf rather than using swap?

briansmith:
> Since payload is dead there, I think we can just assign payload.0 to self.buf rather than using swap?

Thanks! Not sure what I was thinking. Maybe I thought I needed a swap because I originally wrote this on top of your `RawMessage` refactoring.

PR #672, opened 2021-04-24
Remove a few unwraps

ctz:


rustls/src/client/tls13.rs
```diff
@@ -1078,8 +1078,10 @@ impl ExpectTraffic {
 
         #[cfg(feature = "quic")]
         {
-            if cx.common.protocol == Protocol::Quic {
-                PayloadU16::encode_slice(cx.common.quic.params.as_ref().unwrap(), &mut ticket);
+            if let (Protocol::Quic, Some(ref quic_params)) =
+                (cx.common.protocol, &cx.common.quic.params)
```
ctz, line 1082: @djc this could be better if we thread the quic handshake parameters through the handshake state machine, and therefore ensure they're mandatory by the time we get here?

djc: We'd have to duplicate them since they're also API-accessible, but sounds good to me.

djc (approved):

djc:
I can take care of the QUIC improvement later if you like.

rustls/src/client/tls13.rs
djc: Actually, threading it through the state machine is tricky because we use the states for both TCP and QUIC, and of course there won't be transport parameters for TCP connections, so the state machine value will have to be `Option` anyway, in which case I think we're no better off than with this current solution.

PR #674, opened 2021-04-25
client: move more ClientHello handling into version-specific modules

djc:
Similar to the current state on the server side. This improves privacy and will help when we make TLS 1.2 optional.


ctz:
Thanks!

PR #676, opened 2021-04-26
CI/CD: Do the bare minimum to reduce GitHub token credential misuse potential

briansmith:


briansmith:
Pay attention to how this affects code coverage measurement.

djc:
It seemed to have no effect on code coverage measurement. As such, merging this seems sensible.

briansmith:
> It seemed to have no effect on code coverage measurement. As such, merging this seems sensible.

Agreed. I wasn't sure how codecov-action might be affected, but it apparently isn't affected.

Issue #665, opened 2021-04-19
`ClientConnection`/`ServerConnection` shouldn't hold onto its `config`


briansmith:
The `ClientConfig` structure is pretty large in memory. Except for `key_log`, AFAICT none of the fields are needed beyond the handshake phase of the connection. In an application switches configurations while some connections are still alive, then the old connection config will live until the last connection using it is closed, instead of only living until that connection's handshake is complete. This is a large difference in lifetime.

Also, holding this config during the entire connection lifetime means that our state machines aren't as effective at preventing errors as they should be. For example, in our state machine, we shouldn't have full access to `session_persistence` throughout the handshake. Indeed, `emit_initial_client_hello` should be able to read from the cache; after that, the only thing we should be able to do to the cache is update/insert entries in it; we shouldn't be able to query it. We could do that using an `Arc<dyn WriteOnlyCache>` type field if we refactor the code  to get rid of the full-featured `session_persistence` field. that If we were to make such a change, then we'd eliminate most or all potential for race conditions in Rustls; at least, that appears to be the case based on what we've seen so far.

The above is written for the client side, but the same argument applies to the server side.

briansmith:
More concretely, here's the refactoring I'm suggesting:
1. Remove the `config` fields from each `Connection` type.
2. Add the `config` to `InitialState`.
3. In `InitialState::emit_initial_client_hello`, copy whatever parts of the `config` that are still needed into the `ExpectServerHelloOrHelloRetryRequest` structure. In the case of the session cache, only a insert-or-update-only interface to it should be saved.
4. Throughout the handshake, continue forwarding the parts of the config that are still relevant to the next state, dropping stuff that's no longer needed along the way.

djc:
Oh, nice one. I'd be happy to take care of moving the config parts into the handshake states, though I'd prefer to do it after #660 lands.

djc:
Should be fine.

PR #681, opened 2021-04-29
Thread config through handshake states

djc:
As suggested in #665. This does not yet make the cache interface write only. It also makes obvious that we don't actually log keys after the initial handshake; ~~I'll file an issue for that~~ I think that's correct since the master secret is logged and other secrets are derived from it anyway?

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -113,7 +113,7 @@ fn find_session(
 pub(super) fn start_handshake(
     dns_name: webpki::DnsName,
     extra_exts: Vec<ClientExtension>,
-    config: &Arc<ClientConfig>,
+    config: Arc<ClientConfig>,
     cx: &mut ClientContext<'_>,
 ) -> NextStateOrError {
     let mut transcript = HandshakeHash::new();
@@ -129,13 +129,13 @@ pub(super) fn start_handshake(
     let mut session_id: Option<SessionID> = None;
     let mut resuming_session = find_session(
         dns_name.as_ref(),
-        config,
+        &config,
         #[cfg(feature = "quic")]
         cx,
     );
 
     let key_share = if support_tls13 {
-        Some(tls13::initial_key_share(config, dns_name.as_ref())?)
+        Some(tls13::initial_key_share(&config, dns_name.as_ref())?)
     } else {
         None
     };
@@ -167,7 +167,7 @@ pub(super) fn start_handshake(
     let sent_tls13_fake_ccs = false;
     let may_send_sct_list = config.verifier.request_scts();
     Ok(emit_client_hello_for_retry(
-        config.clone(),
+        config,
         cx,
         resuming_session,
         randoms,
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -30,6 +30,8 @@ use crate::SupportedCipherSuite;
 use crate::client::common::ClientHelloDetails;
 use crate::client::{tls12, tls13, ClientConfig, ClientConnectionData};
 
+use std::sync::Arc;
+
 pub(super) type NextState = Box<dyn State>;
 pub(super) type NextStateOrError = Result<NextState, Error>;
 
@@ -52,20 +54,14 @@ pub(super) trait State: Send + Sync {
 
 impl crate::conn::HandleState for Box<dyn State> {
     type Data = ClientConnectionData;
-    type Config = ClientConfig;
 
     fn handle(
         self,
         message: Message,
         data: &mut Self::Data,
         common: &mut ConnectionCommon,
-        config: &Self::Config,
     ) -> Result<Self, Error> {
-        let mut cx = ClientContext {
-            common,
-            data,
-            config,
-        };
+        let mut cx = ClientContext { common, data };
         self.handle(&mut cx, message)
     }
 }
@@ -73,18 +69,17 @@ impl crate::conn::HandleState for Box<dyn State> {
 pub(super) struct ClientContext<'a> {
     pub(super) common: &'a mut ConnectionCommon,
     pub(super) data: &'a mut ClientConnectionData,
-    pub(super) config: &'a ClientConfig,
 }
 
 fn find_session(
-    cx: &mut ClientContext<'_>,
     dns_name: webpki::DnsNameRef,
+    config: &ClientConfig,
+    #[cfg(feature = "quic")] cx: &mut ClientContext<'_>,
 ) -> Option<persist::ClientSessionValueWithResolvedCipherSuite> {
     let key = persist::ClientSessionKey::session_for_dns_name(dns_name);
     let key_buf = key.get_encoding();
 
-    let value = cx
-        .config
+    let value = config
         .session_storage
         .get(&key_buf)
         .or_else(|| {
@@ -95,7 +90,7 @@ fn find_session(
     let mut reader = Reader::init(&value[..]);
     let result = persist::ClientSessionValue::read(&mut reader).and_then(|csv| {
         let time = TimeBase::now().ok()?;
-        csv.resolve_cipher_suite(&cx.config.cipher_suites, time)
+        csv.resolve_cipher_suite(&config.cipher_suites, time)
     });
     if let Some(result) = result {
         if result.has_expired() {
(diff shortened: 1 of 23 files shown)


rustls/src/client/hs.rs
```diff
@@ -410,6 +412,7 @@ fn emit_client_hello_for_retry(
     });
 
     let next = ExpectServerHello {
+        config: config.clone(),
```
ctz, line 415: clone not required?

ctz (commented):
I'm concerned here that we're swapping one long-lived Arc::clone for many. My understanding is that these are relatively slow (like, hundreds of times slower than an Rc-flavour reference count)  because each one involves a full memory barrier which damages total memory bandwidth.

djc:
@briansmith any thoughts on that? I don't have a good intuition for the performance of Arc clones. Do we have any benchmarks that we could use to measure any performance impact?

briansmith:
> I'm concerned here that we're swapping one long-lived Arc::clone for many.

I think that's a valid thing to be concerned about.

AFAICT, this PR serves multiple purposes:
1. Ensure we throw away our reference to the Config when handshaking is done, or sooner, so that we don't cause it to live for the entire life of the connection. This could be done by threading the `Arc<Config>` through the handshake state machine, without breaking it up into multiple parts. This would be the same amount of `Arc`ing as previously, i.e. it is basically free. I think we should do at least this.

2. Clarify which parts of the Config are used by which parts of the handshake, and reduce the chance that some part of the handshake state machine is depending on part of the Config it shouldn't be. This is one way to do that. Another way to do that would be to have a bunch of wrapper types `Foo(Arc<Config>)`, `Bar(Arc<Config>)` that each expose a limited subset of the `Config` (an ever-shrinking subset), where we convert one type into another at each stage of the handshake. Then there would be no extra `Arc`ing.

IMO the first problem is more important to address than the second problem. I also don't think either is particularly urgent to address.

rustls/src/server/tls12.rs
```diff
@@ -208,11 +214,15 @@ mod client_hello {
...
-            let doing_client_auth = emit_certificate_req(&mut self.transcript, cx)?;
+            let doing_client_auth = emit_certificate_req(&self.config, &mut self.transcript, cx)?;
             emit_server_hello_done(&mut self.transcript, &mut cx.common);
 
             if doing_client_auth {
                 Ok(Box::new(ExpectCertificate {
+                    key_log: self.config.key_log.clone(),
+                    verifier: self.config.verifier.clone(),
+                    ticketer: self.config.ticketer.clone(),
+                    session_storage: self.config.session_storage.clone(),
```
briansmith, line 225: I think to make progress here, we should just keep threading the whole `config` instead of breaking it apart. If we do that then we'll be able to eliminate all the extra `clone()`s introduced and still be able to accomplish the main goal.

djc:
As suggested, changed this to just thread the `Arc<*Config>` through the states.

briansmith:
Commit `client: reorder states in chronological order` seems out of scope for this. I do think that's a good change but it's orthogonal to this one, and it's hard to review.

briansmith:
Commit "client: move TLS 1.2 save_session() into ExpectFinished impl block" also seems out of scope for this PR.

briansmith:
Is commit "server: fix incorrect rename artifacts" fixing a problem with the first commit? if so, let's merge those two commits together. Otherwise, is it a separate fix that should be in a separate PR?

djc:
Have scoped those other commits out for now, will bring them back as a separate PR. @ctz do you have a preference for one PR vs three?

rustls/src/client/hs.rs
```diff
@@ -118,27 +113,29 @@ fn find_session(
 pub(super) fn start_handshake(
     dns_name: webpki::DnsName,
     extra_exts: Vec<ClientExtension>,
+    config: &Arc<ClientConfig>,
```
briansmith, line 116: Wherever we're not going to clone the `Arc` then should't we use `&ClientConfig` instead of `&Arc<ClientConfig>` to be more general and to clarify the ownership/cloning?

I would expect everything to be either `config: Arc<ClientConfig>` if moving from one state to another, or `config: &ClientConfig` otherwise.

djc: `start_handshake()` is actually the entry point to the client handshake now, and it does clone the `Arc` when calling `emit_client_hello_for_retry`.

rustls/src/client/tls13.rs
```diff
@@ -642,9 +654,9 @@ impl hs::State for ExpectCertificateVerify {
             .split_first()
             .ok_or(Error::NoCertificatesPresented)?;
         let now = std::time::SystemTime::now();
-        let cert_verified = cx
+        let cert_verified = self
             .config
-            .get_verifier()
+            .verifier
```
briansmith, line 659: Not sure why we're removing `get_verifier()` here. Seems like an orthogonal change.

rustls/src/server/hs.rs
```diff
@@ -290,10 +286,11 @@ pub struct ExpectClientHello {
 
 impl ExpectClientHello {
     pub fn new(
-        server_config: &ServerConfig,
+        server_config: &Arc<ServerConfig>,
```
briansmith, line 289: I think at this point we should already be doing only moves; the cloning of `server_config` should be in the caller, so that this becomes `Arc<ServerConfig>`. That is, at the state machine layer, we're always moving, never cloning.

rustls/src/server/tls12.rs
```diff
@@ -422,10 +439,11 @@ mod client_hello {
     }
 
     fn emit_certificate_req(
+        config: &ServerConfig,
         transcript: &mut HandshakeHash,
         cx: &mut ServerContext<'_>,
     ) -> Result<bool, Error> {
-        let client_auth = cx.config.get_verifier();
+        let client_auth = config.get_verifier();
```
briansmith, line 446: Seems strange we got rid of the `get_verifier()` for the other side but not here. I think we should revert the other side to use `get_verifier()` like before.

rustls/src/server/tls12.rs
```diff
@@ -640,8 +664,8 @@ impl hs::State for ExpectCertificateVerify {
             let handshake_msgs = self.transcript.take_handshake_buf();
             let certs = &self.client_cert;
 
-            cx.config
-                .get_verifier()
+            self.config
+                .verifier
```
briansmith, line 668: Why not `get_verifier()`?

rustls/src/server/tls13.rs
```diff
@@ -296,6 +300,7 @@ mod client_hello {
             let (mut ocsp_response, mut sct_list) =
                 (server_key.get_ocsp(), server_key.get_sct_list());
             emit_encrypted_extensions(
+                &self.config,
```
briansmith, line 303: Should `config` parameters be at the beginning or end of the argument list? Let's be consistent.

rustls/src/client/hs.rs
briansmith: After reviewing the rest, I think this should be `Arc<ClientConfig>`, i.e. cloned by the caller.

briansmith (changes requested):

briansmith:
To the extent `.get_verifier()` vs. `.verifier` doesn't matter, let's revert all those changes. If we want to do it differently, let's do that in a separate PR.

briansmith:
Another than the nits above, I think this is a very good change. Besides the progress towards making it possible to drop the `Config` earlier, I like that you were able to eliminate some apparently-redundant confusing `config` parameters, and I like that you were able to narrow the input types of some functions so they don't even need to take a whole config.

briansmith:
> Have scoped those other commits out for now, will bring them back as a separate PR. @ctz do you have a preference for one PR vs three?

If you do the renaming/typo fixing one as a separate PR then I'll review it ASAP. The other changes that reorder the functions are better reviewed by @ctz.

rustls/src/client/hs.rs
djc: Not sure I agree. We don't need to clone it sooner, and there is value in not cloning it if we're going to error out of `start_handshake()`, for example.

rustls/src/client/hs.rs
djc: Sorry, I missed that. Fixed it up now.

rustls/src/client/tls13.rs
djc: Added it back.

rustls/src/server/tls12.rs
djc: Done.

rustls/src/server/tls12.rs
djc: Changed it back. Will add another changes to get rid of `get_verifier()` to another PR.

rustls/src/server/hs.rs
djc: Done.

rustls/src/client/hs.rs
djc: Changed it anyway for consistency.

rustls/src/client/hs.rs
briansmith: I still see the clone.

rustls/src/server/tls13.rs
djc: Argument order here is a bit of mess anyway, moved it to the end.

briansmith:
@djc Not sure what's going on with the review discussion. IIRC the only outstanding issue in this PR is this:

```
let next = ExpectServerHello {
        config: config.clone(),
```
IIUC, the `clone()` isn't needed and should be removed. I tried to fix this for you in https://github.com/djc/rustls/pull/1 but maybe you didn't noticed it, and now my PR is out of date w.r.t. your branch. Or maybe I'm misunderstanding this?

djc:
Huh, I was pretty sure I fixed that. Maybe I made a rebase error. Please check.

briansmith:
@djc When reviewing your PR again, I noticed that we can avoid the clones completely. I was able to modify your PR so that there are no clones of the configs at all. See https://github.com/djc/rustls/pull/2 for the server. I have an analogous change for the client.

The rule of thumb that I follow is to never clone an `Arc` if it is reasonable for the caller to clone the `Arc` themselves, if avoiding the cloning has the potential to eliminate cloning completely.

briansmith (approved):

PR #685, opened 2021-05-08
Borrow Message payloads from OpaqueMessage

djc:
The "Borrow Message payloads from OpaqueMessage" commit is very large, though most of it is very repetitive code in `msgs::handshake`, adding `to_owned()` methods to many of the payload types and their constituent types.

The idea here is to make `Payload`, `PayloadU8`, `PayloadU16` and `PayloadU24` contain a `Cow<'a, [u8]>` instead of a `Vec<u8>` and work from there to allow `Message` to borrow most of its data. There are some parts here that are not yet borrowed, but I think this proves that it can be made to work.

This does not yet touch folding the deframer and the handshake joiner, though I did explore that part as well. I think that can be made to work such that we can borrow `OpaqueMessage` and `Message` straight from the deframer buffer.

In the `read_tls with EWOULDBLOCK` benchmark, this shows up as a 5% performance improvement.

### DIFF
--- a/fuzz/fuzzers/deframer.rs
+++ b/fuzz/fuzzers/deframer.rs
@@ -16,6 +16,6 @@ fuzz_target!(|data: &[u8]| {
 
     while !dfm.frames.is_empty() {
         let msg = dfm.frames.pop_front().unwrap();
-        Message::try_from(msg).ok();
+        Message::try_from(&msg).ok();
     }
 });
--- a/fuzz/fuzzers/fragment.rs
+++ b/fuzz/fuzzers/fragment.rs
@@ -15,7 +15,7 @@ fuzz_target!(|data: &[u8]| {
         Err(_) => return,
     };
 
-    let msg = match message::Message::try_from(msg) {
+    let msg = match message::Message::try_from(&msg) {
         Ok(msg) => msg,
         Err(_) => return,
     };
@@ -25,6 +25,6 @@ fuzz_target!(|data: &[u8]| {
     frg.fragment(message::OpaqueMessage::from(msg), &mut out);
 
     for msg in out {
-        message::Message::try_from(msg).ok();
+        message::Message::try_from(&msg).ok();
     }
 });
--- a/fuzz/fuzzers/message.rs
+++ b/fuzz/fuzzers/message.rs
@@ -9,7 +9,7 @@ use std::convert::TryFrom;
 fuzz_target!(|data: &[u8]| {
     let mut rdr = Reader::init(data);
     if let Ok(m) = OpaqueMessage::read(&mut rdr) {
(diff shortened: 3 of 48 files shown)


briansmith:
Not sure what the deal is with this PR but it's "+1,247 −705" so basically unreviewable as it is now.

djc:
I already spent quit a bit of time splitting this up in smaller commits, but in order to actually have a meaningful impact, it's not easy to split up the largest commit ("Borrow Message payloads from OpaqueMessage") into smaller pieces. Essentially the +500 lines are `to_owned()` methods on ~all the message payload types (everything that's transitively part of `HandshakePayload`). These methods themselves are pretty trivial.

I'm open to splitting this up more if we (a) decide we want to make changes in this direction (a 5% perf improvement seems fairly compelling, but I don't know how representative the included benchmark is), and (b) there is a plausible route to doing so.

briansmith:
I'm not sure that changing these things to use `Cow` like this is a necessary intermediate step towards where we ultimately want to go. I do like the performance benefit, of course, but I think we should first figure out what the end-state we're hoping to achieve is, and then see how many intermediate states we can cut out to get there from here. I think in the very end, potentially the entire parsing strategy would change to be allocation-free (like webpki) and that would eliminate the need for a lot of what's being done here. I'd like to explore that.

djc:
I'm not sure how to square "it's "+1,247 −705" so basically unreviewable" with "see how many intermediate states we can cut". Given that some of the state machine states do hang on handshake message payloads and these states are currently boxed, it seems to me that using some single external allocation (or a small number of external allocations) is quite far off -- unless we get self-referential borrows somehow, perhaps. Meanwhile this seems to me real progress towards not allocating `Vec`s and copying data at every turn, presented in a way that I spent quite some effort to make reviewable.

briansmith:
I would like to have a plan with some prioritization to guide the development. I think more planning would reduce wasted work. Right now there's no plan and that makes me feel like we're not working efficiently. I'd like to fix that, instead of just reviewing whatever anybody happens to post.

djc:
As I already mentioned in #691, I'm open to more planning/prioritization, but given your and Joe's limited time, I'm not sure what the best way of planning is. Also not quite sure what you mean about "wasted work" (unless this is referring to cutting out intermediate states?) and "not working efficiently".

If my work seems to be heading off in several different directions, part of that is because due to the long review cycles, I'm trying to make reasonably self-contained PRs that don't interfere with each other too much, so I'm essentially juggling several different angles -- less unwrapping, less code duplication, fewer allocations/less copying, public API changes. I have a bunch more branches with improvements that I'm not even submitting because the things that I do submit have, IMO, higher priority.

PR #686, opened 2021-05-09
Include LICENSE files with published crates

decathorpe:
This PR adds symlinks for LICENSE-* files to the "rustls" workspace member, which will make them be included in crate tarballs published to - and available from - crates.io.

Most software licenses require that their license texts are included with redistributed sources - including the Apache 2.0 and standard / modern MIT licenses, so crate sources published and redistributed via crates.io should include the license texts.

(Note: I am assuming you're running "cargo publish" on Linux or mac OS. Git symlinks do not correctly work in Windows, AFAIK.)

### DIFF
--- a/rustls/LICENSE-APACHE
+++ b/rustls/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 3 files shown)


ctz:
Thanks!

Issue #522, opened 2021-02-17
RFC: Unify client and server APIs


briansmith:
* Whenever possible, the API for configuring a client and configuring a server should have one API, instead of two APIs that are the same. This way, we can write a single function that constructs the common parts of our TLS configuration.

```
#[derive(Clone)]
pub struct Config {
    /// List of ciphersuites, in preference order.
    pub ciphersuites: Vec<&'static SupportedCipherSuite>,

    /// Our MTU.  If None, we don't limit TLS message sizes.
    pub mtu: Option<usize>,

    /// Protocol names we support, most preferred first.
    /// If empty we don't do ALPN at all.
    pub alpn_protocols: Vec<Vec<u8>>,

    /// Supported protocol versions, in no particular order.
    /// The default is all supported versions.
    pub versions: Vec<ProtocolVersion>,

    /// How to output key material for debugging.  The default
    /// does nothing.
    pub key_log: Arc<dyn KeyLog>,
}

struct ClientConfig {
    common: Config,
    ...
}

struct ServerConfig {
    common: Config,
}
```

* Instead of having `ServerSession` and `ClientSession`, we should have one `Session` type with an API that abstracts away whether we're the client or the server after the handshake has started. Currently things that want to do this have to do something like this:
```
enum Session {
    Client(ClientSession),
    Server(ServerSession),
}
```
Then they create their own `Session` API that does this abstraction. I believe if we do this within Rustls instead, not only can we eliminate the need for such wrappers, but we can also eliminate significant duplicate logic within Rustls. For example, instead of having `server_cert_chain` in `ClientSession` and `client_cert_chain` in `ServerSession`, we should have `peer_cert_chain` in `SessionCommon`.

* We should merge some of the Server- and Client- specific plugin APIs. I think we should merge `ClientStoresSession` and `ServerStoresSession` into a single thing. Perhaps we should factor out the commonality of `ClientCertVerifier` and `ServerCertVerifier` APIs into a common super-trait; this might allow implementers of those APIs to reduce duplication, and may also allow Rustls internally to reduce some duplication. I am planning to experiment with changing the webpki API to facilitate this.

djc:
> an API that abstracts away whether we're the client or the server after the handshake

I'm not sure how this would look. Can you write some example code? I think it would cause ergonomic issues if the type of a session changes throughout the handshake.

Ralith:
[...]

PR #691, opened 2021-05-13
Further simplifications to the state machine

djc:
Reduce code duplication in client/server by flattening the state machine structure to a single `State` trait which is parametrized by the `Data` and `Config` types. This makes all of the `State` methods available to `ConnectionCommon`, which in turn allows moving more duplicate complexity from client/server into `ConnectionCommon` methods.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -745,6 +745,19 @@ impl ConnectionCommon {
         }
     }
 
+    pub(crate) fn export_keying_material<Data>(
+        &self,
+        output: &mut [u8],
+        label: &[u8],
+        context: Option<&[u8]>,
+        state: &Option<Box<dyn State<Data>>>,
+    ) -> Result<(), Error> {
+        state
+            .as_ref()
+            .ok_or(Error::HandshakeNotComplete)
+            .and_then(|st| st.export_keying_material(output, label, context))
+    }
+
     // Changing the keys must not span any fragmented handshake
     // messages.  Otherwise the defragmented messages will have
     // been protected with two different record layer protections,
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -661,10 +661,10 @@ impl ConnectionCommon {
         Ok(Some(MessageType::Data(msg)))
     }
 
-    pub(crate) fn process_new_packets<S: HandleState>(
+    pub(crate) fn process_new_packets<Data>(
         &mut self,
-        state: &mut Option<S>,
-        data: &mut S::Data,
+        state: &mut Option<Box<dyn State<Data>>>,
+        data: &mut Data,
     ) -> Result<IoState, Error> {
         if let Some(ref err) = self.error {
             return Err(err.clone());
@@ -694,10 +694,10 @@ impl ConnectionCommon {
         Ok(self.current_io_state())
     }
 
-    pub(crate) fn process_new_handshake_messages<S: HandleState>(
+    pub(crate) fn process_new_handshake_messages<Data>(
         &mut self,
-        state: &mut Option<S>,
-        data: &mut S::Data,
+        state: &mut Option<Box<dyn State<Data>>>,
+        data: &mut Data,
     ) -> Result<(), Error> {
         while let Some(msg) = self.handshake_joiner.frames.pop_front() {
             self.process_main_protocol(msg, state, data)?;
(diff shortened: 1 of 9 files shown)


rustls/src/conn.rs
```diff
@@ -1080,17 +1086,29 @@ impl ConnectionCommon {
     }
 }
 
-pub(crate) trait HandleState: Sized {
-    type Data;
-    type Config;
-
+pub(crate) trait State<Data, Config>: Send + Sync {
```
briansmith, line 1109: It doesn't make sense for this to be in conn.rs, does it? It isn't about connection state, but about handshake state.

I suggest we move this into a new `hs` top-level module.

Further, I suggest we start following the normal Rust convention where modules are used more frequently to control visibility, which results in types and traits usually being defined in many small submodules. In this case, I think this trait should be `hs::state::State` in src/hs/state.rs.

Also, it seems like this should be:
```
trait State<C> : Send + Sync {
  fn handle(
        self: Box<Self>,
        cx: &mut C,
        message: Message,
    ) -> Result<Box<dyn State<C>>, Error>;
  }

  fn export_keying_material(...) // same as you have it.
}
```

Then the implementations would be `impl hs::State<ClientContext>` and `impl hs::State<ServerContext>`.

rustls/src/conn.rs
```diff
@@ -1080,17 +1086,29 @@ impl ConnectionCommon {
...
+        Err(Error::HandshakeNotComplete)
+    }
+
+    fn perhaps_write_key_update(&mut self, _common: &mut ConnectionCommon) {}
+}
+
+pub(crate) struct Context<'a, Data, Config> {
+    pub(crate) common: &'a mut ConnectionCommon,
+    pub(crate) data: &'a mut Data,
+    pub(crate) config: &'a Config,
```
briansmith, line 1131: `config` will go away after #681, right?

Then couldn't we just add `data: Data` to `ConnectionCommon` and rename `ConnectionCommon` to `Connection`, and avoid having `Context` at all? In that case we'd have `trait State<Data>`?

Alternatively, if we merge this PR, then do we need #681 at all?

I'm guessing we need #681 because after the handshake we'll have a `State` to indicate we're done handshaking but we don't want to have a `Config` to pass to it.

briansmith:
In general, the idea here seems good.

Before we merge more of these refactorings, I'd like us to get together and agree on a rough plan and prioritization of the refactorings we're doing, so I can get more context in what we're doing here.

briansmith (changes requested):

rustls/src/conn.rs
djc: `Traffic` is also a state in the state machine. The `State` trait is about how the 1.2/1.3 client/server code interacts with the shared logic (buffers, encryption/decryption); so, in my view, `State` is an interface for `ConnectionCommon` to abstract over all those aspects of a particular connection. That said, I don't mind moving it somewhere else, but I think might be a little bit confusing to restrict to just `hs` since it's also used for post-handshake key updates.

In the projects I've worked on, modules are not used much more frequent than in current rustls to control visibility. Personally adding a module which just contains one item seems like unnecessary complexity to me.

briansmith: OK, thanks for explaining. Seems reasonable.

rustls/src/conn.rs
djc: I do definitely think #681 is still useful. In the future I think we don't want to pass all of `ConnectionCommon` into the states as part of the state machines, but only the more limited parts that are required (writing out some of the state for use by the API, and sending messages). In particular, if we want to do more borrowing of incoming message payloads, we cannot pass a mutable reference to the buffers that those messages reside in. In my current thinking, we might split `ConnectionCommon` into `Incoming`, `Outgoing`, and `CommonState` types, and then pass only `Data`, `Outgoing`, and `CommonState` into the handshake.

I did an earlier attempt at the state machine refactoring where `ConnectionCommon` grew type arguments for `Data` and `Config` directly, but this didn't work out well. I don't remember exactly why or how, though it might have been related to how the `ClientSession` gets initialized.

djc:
> Before we merge more of these refactorings, I'd like us to get together and agree on a rough plan and prioritization of the refactorings we're doing, so I can get more context in what we're doing here.

FWIW, I don't think this is particularly high priority, but I do think it helps with the goals outlined in #522 and in general prevents code duplication across server and client.

rustls/src/conn.rs
djc: I guess it could still make sense to have call `CommonState` `ConnectionState` instead and add the `Data` argument to it, and then it could live inside the `ConnectionCommon` which could be renamed to `Connection`. Still, I think the important part in this PR is to flatten the `HandleState` and `client::hs::State`/`server::hs::State` traits together, which we should still do anyway.

djc:
Closing in favor of #721.

PR #696, opened 2021-05-14
Follow clippy::clone_on_ref_ptr guidance.

briansmith:


briansmith:
The clippy job will fail until after #681 is merged. Then this will be ready to go.

This helps us see where exactly we are `Arc`ing.

djc:
Looks like there are a couple of instances that slipped through?

briansmith:
> Looks like there are a couple of instances that slipped through?

Sorry, I tried to clarify that in my comment above. Now that #681 is merged, I've rebased this on top of main and it seems it's good.

PR #697, opened 2021-05-14
Make sure set_mtu doesn't underflow.

### LINKED ISSUE #692: Integer underflow in `ClientConfig::set_mtu`
```
            debug_assert!(x > fragmenter::PACKET_OVERHEAD);
            self.mtu = Some(x - fragmenter::PACKET_OVERHEAD);
```

`debug_assert!` must not be used for this check.

sayrer:
An MTU that small is almost certainly a bug, so I wanted to leave the debug_assert in for non-tests. Fixes #692.

### DIFF
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -159,18 +159,20 @@ impl ClientConfig {
         //
         // Externally the MTU is the whole packet size.  The difference
         // is PACKET_OVERHEAD.
-        if let Some(x) = *mtu {
-            use crate::msgs::fragmenter;
-            if let Some(m) = x.checked_sub(fragmenter::PACKET_OVERHEAD + 1) {
-                self.mtu = Some(m + 1);
-                return Ok(());
+        match *mtu {
+            Some(x) => {
+                use crate::msgs::fragmenter;
+                if let Some(m) = x.checked_sub(fragmenter::PACKET_OVERHEAD + 1) {
+                    self.mtu = Some(m + 1);
+                    return Ok(());
+                }
+                Err(Error::MtuError)
             }
-
-            return Err(Error::MtuError);
+            None => {
+                self.mtu = None;
+                Ok(())
+            }
         }
-
-        self.mtu = None;
-        Ok(())
     }
 
     /// Access configuration options whose use is dangerous and requires
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -161,11 +161,11 @@ impl ClientConfig {
         // is PACKET_OVERHEAD.
         if let Some(x) = *mtu {
             use crate::msgs::fragmenter;
-            if !cfg!(test) {
-                debug_assert!(x > fragmenter::PACKET_OVERHEAD);
-            }
             self.mtu = x.checked_sub(fragmenter::PACKET_OVERHEAD);
             if self.mtu.is_none() {
+                if !cfg!(test) {
+                    debug_assert!(x > fragmenter::PACKET_OVERHEAD);
+                }
                 warn!(
                     "MTU must be greater than {} bytes. Setting to default.",
                     fragmenter::PACKET_OVERHEAD
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -4,7 +4,7 @@ use crate::key;
 use crate::keylog::KeyLog;
 use crate::kx::SupportedKxGroup;
 #[cfg(feature = "logging")]
-use crate::log::{trace, warn};
+use crate::log::trace;
 #[cfg(feature = "quic")]
 use crate::msgs::enums::AlertDescription;
 use crate::msgs::enums::CipherSuite;
@@ -153,7 +153,7 @@ impl ClientConfig {
 
     /// Sets MTU to `mtu`.  If None, the default is used.
     /// If Some(x) then x must be greater than 5 bytes.
-    pub fn set_mtu(&mut self, mtu: &Option<usize>) {
+    pub fn set_mtu(&mut self, mtu: &Option<usize>) -> Result<(), Error> {
         // Internally our MTU relates to fragment size, and does
         // not include the TLS header overhead.
         //
@@ -161,19 +161,18 @@ impl ClientConfig {
         // is PACKET_OVERHEAD.
         if let Some(x) = *mtu {
             use crate::msgs::fragmenter;
-            self.mtu = x.checked_sub(fragmenter::PACKET_OVERHEAD);
-            if self.mtu.is_none() {
-                if !cfg!(test) {
-                    debug_assert!(x > fragmenter::PACKET_OVERHEAD);
+            if let Some(m) = x.checked_sub(fragmenter::PACKET_OVERHEAD + 1) {
+                if m > 0 {
+                    self.mtu = Some(m);
+                    return Ok(());
                 }
-                warn!(
-                    "MTU must be greater than {} bytes. Setting to default.",
-                    fragmenter::PACKET_OVERHEAD
-                );
             }
-        } else {
-            self.mtu = None;
+
+            return Err(Error::MtuError);
         }
+
+        self.mtu = None;
+        Ok(())
     }
 
     /// Access configuration options whose use is dangerous and requires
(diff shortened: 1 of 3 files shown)


rustls/src/client/mod.rs
```diff
@@ -161,8 +161,16 @@ impl ClientConfig {
         // is PACKET_OVERHEAD.
         if let Some(x) = *mtu {
             use crate::msgs::fragmenter;
-            debug_assert!(x > fragmenter::PACKET_OVERHEAD);
-            self.mtu = Some(x - fragmenter::PACKET_OVERHEAD);
+            if !cfg!(test) {
+                debug_assert!(x > fragmenter::PACKET_OVERHEAD);
+            }
```
djc, line 166: Seems like `!cfg!(test)` is the wrong way around for this? We want to assert only when in `cfg!(test)`, right?

Also I'd maybe move that `if` into the `self.mtu.is_none()` branch? Makes it a little clearer IMO.

sayrer: > We want to assert only when in cfg!(test), right?

I don't think so. If done that way, the test will fail without being able to check the resulting value.

> Also I'd maybe move that if into the self.mtu.is_none() branch? Makes it a little clearer IMO.

Will do.

rustls/src/client/mod.rs
```diff
@@ -161,8 +161,16 @@ impl ClientConfig {
         // is PACKET_OVERHEAD.
         if let Some(x) = *mtu {
             use crate::msgs::fragmenter;
-            debug_assert!(x > fragmenter::PACKET_OVERHEAD);
-            self.mtu = Some(x - fragmenter::PACKET_OVERHEAD);
+            self.mtu = x.checked_sub(fragmenter::PACKET_OVERHEAD);
+            if self.mtu.is_none() {
+                if !cfg!(test) {
```
briansmith, line 166: First, the doc comment needs to be updated to describe the new behavior.

FWIW, I don't think we need to be this fancy. I suggest we stick to normal Rust patterns and just return a `Result<(), MtuTooSmallError>` where the caller can either choose to `unwrap()` it or deal with the error. Only advanced users of Rustls will be calling `set_mtu()` anyway.

rustls/src/client/mod.rs
```diff
@@ -603,3 +611,15 @@ pub trait ClientQuicExt {
...
+
+#[test]
+fn too_small_mtu() {
+    use crate::{ConfigBuilder, RootCertStore};
+    let mut client_config = ConfigBuilder::with_safe_defaults()
+        .for_client()
+        .unwrap()
+        .with_root_certificates(RootCertStore::empty(), &[])
+        .with_no_client_auth();
+    client_config.set_mtu(&Some(1));
```
briansmith, line 623: If I had to pick a single value to use, I would pick the maximum disallowed value. I'd also like to see a test with the minimum acceptable value.

rustls/src/client/mod.rs
sayrer: I did think about that, but the likelihood of doing anything productive with `MtuTooSmallError` seemed low. That seems fine by me, if others want to go that route. RFC 791 requires 68 byte MTUs not to be fragmented.

briansmith: > I did think about that, but the likelihood of doing anything productive with MtuTooSmallError seemed low.

Most likely they're `unwrap()` it or wrap it in their own error type and forward it back to their caller, and maybe log it if they feel like it. The callers already have to deal with a lot of "I can't do anything about this" errors during configuration in my experience, so they'd probably like to handle them all consistently. At least, as a user (but not of this function) that's what I prefer.

sayrer:
That last failure was the CI integration failing to run `rustup` on Windows. I've seen that happen a few times lately.

I'm still not really happy with this, because I don't think it makes sense to allow 1 byte of payload, but that was the old behavior. Will update the doc comments if people are happy with the approach.

briansmith:
> I'm still not really happy with this, because I don't think it makes sense to allow 1 byte of payload, but that was the old behavior. Will update the doc comments if people are happy with the approach.

If I understand you correctly, I think there are two issues:
1. The underflow.
2. The enforced limit is too small.

Is that right?

I suggest we finish this PR with it enforcing the current limit and then open a new PR on top of it to enforce a reasonable minimum.

briansmith:
> That last failure was the CI integration failing to run `rustup` on Windows. I've seen that happen a few times lately.

Yes, it sucks. It's been happening in GitHub Actions in my repos too. I heard of a workaround RUSTUP_UNPACK_RAM that can be used to limit the amount of RAM rustup uses which might help. I will try it out in my repos.

sayrer:
> > I'm still not really happy with this, because I don't think it makes sense to allow 1 byte of payload, but that was the old behavior. Will update the doc comments if people are happy with the approach.
> 
> If I understand you correctly, I think there are two issues:
> 
>     1. The underflow.
> 
>     2. The enforced limit is too small.
> 
> 
> Is that right?
> 
> I suggest we finish this PR with it enforcing the current limit and then open a new PR on top of it to enforce a reasonable minimum.

That's right. Sounds fine to me. I think RFC 791 (68 octet minimum) is the way to go, but maybe others no of non-Internet uses with smaller MTUs. The smallest I've encountered in the last decade is 1280, the IPv6 minimum (some wifi installations do this).

djc:
> The smallest I've encountered in the last decade is 1280, the IPv6 minimum (some wifi installations do this).

QUIC also requires this (but on UDP, of course). As an enforced limit, maybe 68 is better though.

sayrer:
It seems like the doc comments are still accurate. If this approach seems good, feel free to review the new patch, and we can work on a sensible minimum in another PR.

rustls/src/client/mod.rs
```diff
@@ -153,18 +153,25 @@ impl ClientConfig {
...
+                    self.mtu = Some(m + 1);
+                    return Ok(());
+                }
+                Err(Error::MtuError)
+            }
+            None => {
+                self.mtu = None;
+                Ok(())
+            }
         }
```
djc, line 175: I feel like this would be easier to follow, what do you think?

```rust
        self.mtu = (*mtu)
            .map(|mtu| {
                use crate::msgs::fragmenter;
                match mtu.checked_sub(fragmenter::PACKET_OVERHEAD + 1) {
                    Some(mtu) => Ok(mtu + 1),
                    None => Err(Error::MtuError),
                }
            })
            .transpose()?;
        Ok(())
```

rustls/src/error.rs
```diff
@@ -105,6 +105,9 @@ pub enum Error {
 
     /// An incoming connection did not support any known application protocol.
     NoApplicationProtocol,
+
+    /// The MTU value supplied to the client was too small.
+    MtuError,
```
djc, line 110: Maybe make the error name more specific, like `InvalidMtu`?

rustls/src/client/mod.rs
```diff
@@ -603,3 +610,33 @@ pub trait ClientQuicExt {
 
 #[cfg(feature = "quic")]
 impl ClientQuicExt for ClientConnection {}
+
+#[test]
```
djc, line 614: Since we don't have tests in this module and this is testing the public API, maybe just add this test in `tests/api.rs`, like most of the others? (And if it stays here it should go in a `#[cfg(test)] mod tests {}` block.)

djc (commented):
Some style/naming nits, thanks for working on this!

rustls/src/client/mod.rs
djc: Or alternatively, something like this if the combinator usage is a bit much:

```rust
        let mtu = match mtu {
            Some(mtu) => *mtu,
            None => {
                self.mtu = None;
                return Ok(());
            }
        };

        use crate::msgs::fragmenter;
        let mtu = mtu
            .checked_sub(fragmenter::PACKET_OVERHEAD + 1)
            .ok_or(Error::MtuError)?;
        self.mtu = Some(mtu + 1);
        Ok(())
```

sayrer:
I went with the non-combinator version you had. It's about the same length as what I had, but I'm fine with that style too. The map/transpose version didn't really seem to help readability.

Moved the test to api.rs, which made me think we should expose the fragmenter limits in the API. I suggest we can do that in the PR for the MTU limit itself.

djc:
Thanks!

djc:
Oops, I should have squashed this. Sorry about that.

PR #699, opened 2021-05-15
Leave allocation to caller for Session::peer_certificates()

djc:
This is a small change, but since it affects the public API, would be good to get it in.

### DIFF
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -463,12 +463,12 @@ impl Connection for ClientConnection {
         self.common.send_close_notify()
     }
 
-    fn peer_certificates(&self) -> Option<Vec<key::Certificate>> {
+    fn peer_certificates(&self) -> Option<&[key::Certificate]> {
         if self.data.server_cert_chain.is_empty() {
             return None;
         }
 
-        Some(self.data.server_cert_chain.to_vec())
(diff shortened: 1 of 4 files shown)


rustls/src/client/mod.rs
```diff
@@ -463,12 +463,12 @@ impl Connection for ClientConnection {
         self.common.send_close_notify()
     }
 
-    fn peer_certificates(&self) -> Option<Vec<key::Certificate>> {
+    fn peer_certificates(&self) -> Option<&[key::Certificate]> {
```
briansmith, line 466: If we're hoping to create a stable API then it would be better if the return type were `Option<PeerCertificatesIterator<'a>>` where `PeerCertificatesIterator` implements `Iterator<Item=&[u8]>` and where the initial implementation is `struct PeerCertificatesIterator<'a>(&'a [u8])`.

This way, in the future we can potentially eliminate more of the copying and allocating.

briansmith (approved):

ctz (approved):

rustls/src/client/mod.rs
djc: I'll change it after we decide what to do with #685; then, it could just be `impl Iterator<Item = &key::Certificate<'_>>`.

PR #700, opened 2021-05-15
Cleanups in state machine left over from config threading PR

djc:
These were scoped out of #681.


briansmith:
I reviewed the first commit "server: fix incorrect rename artifacts" and it's good to merge. I didn't review the other two.

ctz (approved):

PR #703, opened 2021-05-16
Fix up formatting on main.

sayrer:
Ran into this working on some patches.


ctz:
Thanks! The GH actions outage today really messed up my workflow.

Issue #713, opened 2021-05-25
The default send buffer limit should not be unlimited


briansmith:
The documentation for `set_buffer_limit()` says "By default, there is no limit. The limit can be set at any time, even if the current buffer use is higher."

This is not a safe default. It seems like it is pretty common (see #712) for users to get tripped up by this and then use massive amounts of memory. Given that we're changing the public API in this release, we should change it so that either the user must provide a buffer limit, or we should have a "safe" default.

ctz:
> we should change it so that either the user must provide a buffer limit, or we should have a "safe" default.

agreed. i'd suggest something the order of 64KB to match a typical tcp send window.

djc:
Currently we only set the limit on the `sendable_tls` and `sendable_plaintext` windows; should we also set a limit on the incoming application data buffer (`received_plaintext`)?0

ctz:
> Currently we only set the limit on the `sendable_tls` and `sendable_plaintext` windows; should we also set a limit on the incoming application data buffer (`received_plaintext`)?

See https://github.com/ctz/rustls/blob/beef0be1658cd457d4f47039abf4f4bb37d59a78/rustls/src/server/mod.rs#L336-L342

PR #714, opened 2021-05-25
Set buffer limits by default

### LINKED ISSUE #712: Significant heap growth with reqwest+rustls
I'm observing significant heap growth when using reqwest to upload a large file to a remote HTTPS server. (I'm using reqwest 0.11.3 with rustls 0.22.1 to stream the payload in the body of a PUT request, I can provide code snippets, if needed)

The funny thing is, everything is fine when there's a gigabit link between the client and the server. But if I throttle the network between the client and the server to 1 mbit/sec, I'm seeing huge growth in the client heap. I've confirmed this with valgrind's massif tool. According to valgrind massif, on the client 99% of the heap seems to be allocated here: https://github.com/ctz/rustls/blob/rel-19-backports/rustls/src/session.rs#L725

I can provide massif output for both throttled and unthrottled network runs...

I've found out that there's a rustls::Session::set_buffer_limit() API. I'm not sure if is (a) safe to use in a reqwest+rustls async app and (b) would actually reduce the per-session heap allocations in the case described above (uploading a large file over a slow network).

Any help on this would be greatly appreciated.

djc:
Fixes #712, #713.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 * Next release:
   - Planned: removal of unused signature verification schemes at link-time.
+  - *Breaking change*: internal buffers are now limited to 64 kB by default. Use
+    `Connection::set_buffer_limit` to change the buffer limits to suit your application.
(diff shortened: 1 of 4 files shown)


ctz:
Thanks!

PR #719, opened 2021-05-28
Make Message cloneable

maxammann:
Hello!

first thanks for the possibility to use rustls and the message definitions as a library. I have the need to clone a Message.
Sadly because Rust does not yet implement negative trait bounds it is not possible to do the following:

```rust

impl<T: 'static> SomeTrait for T
where
    T: Clone,
{
    fn ...
}

impl SomeTrait for Message {
    fn ...
}
```

This is because of `where T: Clone`. This causes an issue because rustls can always implement `Clone` and break the above code. Rust does not allow that to happen.

I think there is no problem with implementing Clone for more types.

### DIFF
--- a/rustls/src/msgs/alert.rs
+++ b/rustls/src/msgs/alert.rs
@@ -1,7 +1,7 @@
 use crate::msgs::codec::{Codec, Reader};
 use crate::msgs::enums::{AlertDescription, AlertLevel};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct AlertMessagePayload {
     pub level: AlertLevel,
     pub description: AlertDescription,
--- a/rustls/src/msgs/ccs.rs
+++ b/rustls/src/msgs/ccs.rs
@@ -1,6 +1,6 @@
 use crate::msgs::codec::{Codec, Reader};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ChangeCipherSpecPayload;
 
 impl Codec for ChangeCipherSpecPayload {
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -841,7 +841,7 @@ impl ServerExtension {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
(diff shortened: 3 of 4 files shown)


djc:
What's the use case for this? What are you using the `Message` type for? These types are not considered part of the public interface, so it's not clear that we should accept changes like these unless there's a clear need.

maxammann:
So even though it is not a stable API the comments in the code says that it might be oke to use the exposed types: https://github.com/ctz/rustls/blob/main/rustls/src/lib.rs#L286

I'm using the TLS message serialization and deserialization as part of a fuzzer which is written in Rust. rustls makes it very easy for me to parse the message, modify them and then pass them to OpenSSL for example.

I'm fine if this is not accepted, but it would also be very cool if rustls would allow this kind of usage. rustls is clear that what I'm doing is not stable and will break with each version.

As far as I know rustls is the only Rust library which allows parsing of TLS packets.

djc:
Sure, it says it is okay to use. Will these types still be useful for you if they are no longer `'static`?

maxammann:
> Will these types still be useful for you if they are no longer `'static`? (This might happen in the near future.)

Actually, I don't think so if understand you correctly. Static types are types which contain no references or only references with 'static, right?

That would actually cause major problems for my implementation because I rely on TypeIds and Anys in my fuzzer.

briansmith:
> So even though it is not a stable API the comments in the code says that it might be oke to use the exposed types: https://github.com/ctz/rustls/blob/main/rustls/src/lib.rs#L286

We should stop exposing these types in the public API.

I don't think it makes sense, logically, to make these type `Clone` in normal use.

If we need to have a special API that exposes more functionality for fuzzing, that's worth considering. But, it'd be better to find an alternate.

maxammann:
@briansmith I agree :) I'm kind of thinking about just stripping the parsing out. It could also be released as a separate library.

I think that the main goal of rusttls does not benefit from maintaining such a library. Also the other way around, just a library does not benefit from staying within rustls, as it might break like I noticed in the bast 2 months.

I'll close this now, if you have any thoughs feel free to reopen, discuss here or open a separate issue.

PR #721, opened 2021-05-29
Simplify state machine & split ConnectionCommon in two parts

djc:
This combines the main change from #691 with some of the changes from #710 to split up the `ConnectionCommon` type to separate out interfaces that the state machine states use to write into the `ConnectionCommon` type. This addresses the feedback from #691 that we should ditch the context types and move the data types into the `ConnectionCommon`.

Splitting the `ConnectionCommon` type also enables a bunch of borrowing because it separates the `handshake_joiner` and `message_deframer` from the state the state machine wants to write to.

I think this represents a good step towards the final situation while focusing on the larger changes. While this doesn't impact the public API directly, some of the borrowing changes might (for example due to the peer certificates return type changes).

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -265,6 +265,7 @@ pub(super) fn prepare_resumption(
     exts: &mut Vec<ClientExtension>,
     doing_retry: bool,
 ) {
+    cx.common.suite = Some(resuming_suite.into());
     cx.data.resumption_ciphersuite = Some(resuming_suite.into());
     // The EarlyData extension MUST be supplied together with the
     // PreSharedKey extension.
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -431,9 +431,11 @@ impl State<ClientConnectionData> for ExpectEncryptedExtensions {
                     );
             }
 
-            cx.data.server_cert_chain = resuming_session
-                .server_cert_chain
-                .clone();
+            cx.common.peer_certificates = Some(
+                resuming_session
+                    .server_cert_chain
+                    .clone(),
+            );
 
             // We *don't* reverify the certificate chain here: resumption is a
             // continuation of the previous session in terms of security policy.
@@ -729,7 +731,7 @@ impl State<ClientConnectionData> for ExpectCertificateVerify {
             )
             .map_err(|err| hs::send_cert_error_alert(cx.common, err))?;
 
-        cx.data.server_cert_chain = self.server_cert.cert_chain;
+        cx.common.peer_certificates = Some(self.server_cert.cert_chain);
         self.transcript.add_message(&m);
 
         Ok(Box::new(ExpectFinished {
@@ -987,7 +989,10 @@ impl ExpectTraffic {
             &SessionID::empty(),
             nst.ticket.0.clone(),
             secret,
-            &cx.data.server_cert_chain,
+            cx.common
+                .peer_certificates
+                .clone()
+                .unwrap_or_default(),
             time_now,
         );
         value.set_times(nst.lifetime, nst.age_add);
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -1,5 +1,5 @@
 use crate::check::{check_message, inappropriate_handshake_message, inappropriate_message};
-use crate::conn::{CommonState, ConnectionRandoms};
+use crate::conn::{CommonState, ConnectionRandoms, State};
 use crate::error::Error;
 use crate::hash_hs::{HandshakeHash, HandshakeHashBuffer};
 use crate::kx;
@@ -31,6 +31,7 @@ use crate::{conn::Protocol, msgs::base::PayloadU16, quic};
 use crate::{sign, KeyLog};
 
 use super::hs::ClientContext;
+use super::ClientConnectionData;
 use crate::client::common::ServerCertDetails;
 use crate::client::common::{ClientAuthDetails, ClientHelloDetails};
 use crate::client::{hs, ClientConfig, ServerName};
@@ -381,7 +382,7 @@ struct ExpectEncryptedExtensions {
     hello: ClientHelloDetails,
 }
 
-impl hs::State for ExpectEncryptedExtensions {
+impl State<ClientConnectionData> for ExpectEncryptedExtensions {
     fn handle(mut self: Box<Self>, cx: &mut ClientContext<'_>, m: Message) -> hs::NextStateOrError {
         let exts = require_handshake_msg!(
             m,
@@ -477,7 +478,7 @@ struct ExpectCertificateOrCertReq {
     may_send_sct_list: bool,
 }
 
-impl hs::State for ExpectCertificateOrCertReq {
+impl State<ClientConnectionData> for ExpectCertificateOrCertReq {
     fn handle(self: Box<Self>, cx: &mut ClientContext<'_>, m: Message) -> hs::NextStateOrError {
         check_message(
             &m,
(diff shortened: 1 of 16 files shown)


rustls/src/conn.rs
```diff
@@ -719,13 +700,177 @@ impl ConnectionCommon {
...
+            if !self.connection_at_eof() {
+                return Err(io::ErrorKind::WouldBlock.into());
+            }
+        }
+
+        Ok(len)
+    }
+}
+
+pub(crate) struct CommonState {
```
djc: I don't love this name but couldn't think of anything better. This is the connection state shared by client and server, but is different from the state machine types that implement the `State` trait.

rustls/src/conn.rs
```diff
@@ -719,13 +700,177 @@ impl ConnectionCommon {
         state: &mut Option<S>,
         data: &mut S::Data,
     ) -> Result<(), Error> {
+        self.common_state.aligned_handshake = self.handshake_joiner.is_empty();
```
djc, line 715: While almost all the changes in this commit (in this module) are moving code around, here we use one trick to allow state machine states to inspect the handshake joiner state by caching it as a `bool` in `CommonState`, which then gets checked in `CommonState::check_aligned_handshake()`.

rustls/src/conn.rs
```diff
@@ -62,7 +62,8 @@ impl IoState {
 
 /// A structure that implements `std::io::Read` for reading plaintext.
 pub struct Reader<'a> {
-    common: &'a mut ConnectionCommon,
+    received_plaintext: &'a mut ChunkVecBuffer,
```
djc, line 65: The immediate reason was this to prevent the type argument from `ConnectionCommon` having to leak into the `Connection` trait as an associated type, which would preclude turning `Connection` instances into trait objects, but I also like that this moves the logic out of the generic `ConnectionCommon` type and into the specific `Reader` where it is actually used.

rustls/src/conn.rs
```diff
@@ -587,7 +598,12 @@ impl ConnectionCommon {
     }
 
     pub fn reader(&mut self) -> Reader {
-        Reader { common: self }
+        Reader {
+            received_plaintext: &mut self.common_state.received_plaintext,
+            /// Are we done? i.e., have we processed all received messages, and received a
+            /// close_notify to indicate that no new messages will arrive?
+            connection_at_eof: !self.peer_eof || self.message_deframer.has_pending(),
```
samlh, line 605: The name of this value doesn't seem to match it's behavior (should it be connection_not_at_eof?)

rustls/src/conn.rs
```diff
@@ -64,7 +64,8 @@ impl IoState {
 
 /// A structure that implements [`std::io::Read`] for reading plaintext.
 pub struct Reader<'a> {
-    common: &'a mut ConnectionCommon,
+    received_plaintext: &'a mut ChunkVecBuffer,
+    connection_not_at_eof: bool,
```
ctz, line 68: I think I'd prefer for this to be inverted, ie `connection_at_eof`

rustls/src/conn.rs
```diff
@@ -621,7 +632,12 @@ impl ConnectionCommon {
     }
 
     pub(crate) fn reader(&mut self) -> Reader {
-        Reader { common: self }
+        Reader {
+            received_plaintext: &mut self.common_state.received_plaintext,
+            /// Are we done? i.e., have we processed all received messages, and received a
+            /// close_notify to indicate that no new messages will arrive?
```
ctz, line 638: This comment is not matching the code now

rustls/src/conn.rs
```diff
@@ -838,7 +834,8 @@ impl ConnectionCommon {
     }
 }
 
-pub(crate) struct CommonState {
+pub(crate) struct CommonState<Data> {
+    pub(crate) data: Data,
```
djc, line 838: I'm still not quite sure about this. If we kept an intermediate layer in `ConnectionCommon`, let's just call it `Context` again, with `ConnectionCommon` holding a `Context<Data>` and `Context` holding `data` and `common_state`, we would be able to mutably lend out the `CommonState` to something like the writer, so we could potentially do away with the somewhat awkward `PlaintextSink` API. I'm happy to revert that as a follow-up change, but that would maybe a bit silly. @ctz what do you think?

One reason that would be interesting is that we could implement something like `Deref<CommonState>` and `DerefMut<CommonState>` for `ClientConnection` and `ServerConnection`, which gives us a non-side-specific type which might be useful for downstream users as well as for internal things like reader or writer.

Further thinking on this in the context of #522, at this point my ideal API would probably be to have a public `Connection` type with a type param `Client` or `Server` (and this could also implement `Deref` and `DerefMut`). Those two types would be essentially `ClientConnectionData` and `ServerConnectionData` but we would have them implement some marker trait (`Side`) which has a private supertrait bound. This gives us:

* Precise typing, in that `Connection<T: Side>` can hold methods which apply to both sides, `Connection<Client>` and `Connection<Server>` can hold side-specific methods. Downstream crates could abstract over this as well.
* No more indirection through the `Connection` trait, which seems to cause a decent amount of confusion (in people not being able to find `Connection`'s methods).
* Less code from the `Connection` definition and the fairly trivial impl blocks for it.

I'd prefer to leave most of this stuff for a future PR, but it does inform the design choices here.

rustls/src/conn.rs
djc: This was fixed, thanks for the careful review!

briansmith:
Is this PR affecting the public API? (I'm trying to get a sense of what things need to be done ASAP.)

djc:
Nope.

rustls/src/client/tls13.rs
```diff
@@ -269,6 +269,7 @@ pub(super) fn prepare_resumption(
     exts: &mut Vec<ClientExtension>,
     doing_retry: bool,
 ) {
+    cx.common.suite = Some(resuming_suite.into());
```
djc, line 272: The only readers of the `CommonState::suite` field other than `Connection::negotiated_cipher_suite()` are `quic::write_hs()` and `quic::next_1rtt_keys()`, both of which will only read it after handshake secrets have been established (in other words, on the client side the "new" suite will be available), as well as `<ServerConnection as QuicExt>::zero_rtt_keys()`, which is irrelevant by virtue of being server-side only. (With #750, `write_hs()` and `next_1rtt_keys()` will no longer rely on `CommonState::suite` either, and will get it directly from the handshake state.)

rustls/src/conn.rs
```diff
@@ -731,6 +764,49 @@ impl ConnectionCommon {
...
+        context: Option<&[u8]>,
+    ) -> Result<(), Error> {
+        match self.state.as_ref() {
+            Ok(st) => st.export_keying_material(output, label, context),
+            Err(e) => Err(e.clone()),
+        }
+    }
+}
+
+#[cfg(feature = "quic")]
```
djc, line 672: Merges `quic::read_hs()` and its callers in the `QuicExt` implementations. Since this code is fairly tightly coupled to the `ConnectionCommon` and `process_new_handshake_messages()` in particular, this seems like a better interface, at the cost of having more QUIC-specific code in the shared `conn` module. (On the plus side, `ConnectionCommon::process_new_handshake_messages()` is now private.)

djc:
This now changes the public API somewhat in order to facilitate #787 and #789.

rustls/src/conn.rs
```diff
@@ -841,7 +841,7 @@ impl ConnectionCommon {
         // In the handshake case we don't have readable plaintext before the handshake has
         // completed, but also don't want to read if we still have sendable tls.
         self.received_plaintext.is_empty()
-            && !self.connection_was_cleanly_closed()
+            && !self.has_received_close_notify
```
djc: It doesn't seem like the existing of pending messages in the deframer should affect whether we want/need a read, after we've received a close notify from the peer (and this makes further changes easier because otherwise `wants_read()` depends on the deframer, which will stay in `ConnectionCommon`).

ctz (approved):
lgtm!

PR #722, opened 2021-05-29
Conceptually split PlainMessage out of OpaqueMessage

djc:
Following feedback in #667, further clean up the semantic message
types by distinguishing messages with opaque (potentially encrypted)
payloads and messages with plaintext payloads.

This will also help with borrowing, because OpaqueMessages will
want to mutably borrow their payloads in order to be able to
decrypt them, we can borrow a PlainMessage's payloads immutably.


ctz:
Thanks 👍🏽

PR #723, opened 2021-05-29
Eliminate unwrap and debug_assert guards from the handshake hash.

sayrer:
These were bugging me as I inspected `hash_hs.rs` during ECH work. They tend to roll up to a `NextStateOrError` result anyway, so the change was mostly mechanical.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -121,7 +121,7 @@ pub(super) fn start_handshake(
         .client_auth_cert_resolver
         .has_certs()
     {
-        transcript.set_client_auth_enabled();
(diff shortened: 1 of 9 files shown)


djc:
I actually already fixed this once, but never got around to finishing it up after the prerequisite refactoring landed. See here:

https://github.com/djc/rustls/commit/aa70cc04ebb4b2821e6028effaa9537c422b7b0c

I think my change is nicer since it removes the state confusion from `HandshakeHash` directly and doesn't need a new error variant. Do you want to take my change, or shall I rebase it?

sayrer:
Yours is nicer, I agree. You should rebase it. Only nit is that you left a debug_assert in take_handshake_buf.

PR #724, opened 2021-05-31
Use different types for TLS 1.2/1.3 cipher suites

djc:
Here's an attempt at splitting up the `SupportedCipherSuite` type to make it more obvious which version is being used. This eliminates the need for unwrapping to get at the TLS 1.2-specific type. I also changed the TLS 1.2 encrypter/decrypter setup because I ran into an issue with a bug in rustc (fixed on nightly) to use a trait object instead of function references. I think this maintains link-time dead code elimination, but could use some guidance on how to test for this.

If this is merged, we could consider configuring versions by splitting `cipher_suites` in `tls13_cipher_suites` and `tls12_cipher_suites` (and adding an appropriate builder stage). This seems like an improvement to me since it removes the redundant state between cipher suites and protocol versions.

### DIFF
--- a/rustls-mio/examples/tlsclient.rs
+++ b/rustls-mio/examples/tlsclient.rs
@@ -361,12 +361,12 @@ fn lookup_ipv4(host: &str, port: u16) -> SocketAddr {
 }
 
 /// Find a ciphersuite with the given name
-fn find_suite(name: &str) -> Option<&'static rustls::SupportedCipherSuite> {
+fn find_suite(name: &str) -> Option<rustls::SupportedCipherSuite> {
     for suite in rustls::ALL_CIPHERSUITES {
-        let sname = format!("{:?}", suite.suite).to_lowercase();
+        let sname = format!("{:?}", suite.suite()).to_lowercase();
 
         if sname == name.to_string().to_lowercase() {
-            return Some(suite);
+            return Some(*suite);
         }
     }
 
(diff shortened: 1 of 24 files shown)


ctz:
I think it would make more sense for `rustls::cipher_suite::TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256` and co to be of type `SupportedCipherSuite` rather than forcing every use of these values to re-state the version it applies to. I think that would also mean the Tls12CipherSuite type could be kept private?

djc:
That's what I had at first, but it makes things awkward [for QUIC](https://github.com/ctz/rustls/blob/main/rustls/src/quic.rs#L175). I guess we could special case that.

The real question is what your opinion is on configuring versions through distinct cipher suites. If you think that route makes sense, we should keep the cipher suites as different types here. I like that route because eliminating the redundancy between versions and cipher suites feels more robust to me. If you like, you can check out [my branch](https://github.com/ctz/rustls/compare/main...djc:versioned-suites?expand=1) to get a better feeling for how that would look.

ctz:
I think we should keep version and suite configuration separate for now, even though they're different ways to configure the same outcome. The reasons for that are:

1. only if you consider TLS1.2 and TLS1.3 do cipher suites uniquely identify a protocol version. While it's unlikely we ever support TLS1.1 here (but maybe a fork would?), TLS1.4 may become a thing and I'd really expect that to stick with TLS1.3-shaped suites.
2. I think controlling protocol versions directly is a more common and useful desire than customising ciphersuites. Forcing someone with that desire to express it through selecting ciphersuites is not ideal.

djc:
Okay, after making the changes I agree that this is an improvement. I kept a `pub(crate)` `_INTERNAL` suffixed item for the `TLS13_AES_128_GCM_SHA256` case that QUIC needs.

ctz:
Thanks!

PR #725, opened 2021-05-31
Clarify state of transcript hash through different types

djc:
Eliminates 2 unwrap calls in the `HandshakeHash` module.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -472,16 +472,19 @@ fn emit_certverify(
     client_auth: &mut ClientAuthDetails,
     common: &mut ConnectionCommon,
 ) -> Result<(), Error> {
-    let signer = match client_auth.signer.take() {
-        None => {
+    let (signer, message) = match (client_auth.signer.take(), transcript.take_handshake_buf()) {
+        (Some(signer), Some(msg)) => (signer, msg),
+        (None, _) => {
             trace!("Not sending CertificateVerify, no key");
             transcript.abandon_client_auth();
             return Ok(());
         }
-        Some(signer) => signer,
+        (_, None) => {
+            trace!("Not sending CertificateVerify, no transcript");
+            return Ok(());
+        }
     };
 
-    let message = transcript.take_handshake_buf();
     let scheme = signer.get_scheme();
     let sig = signer.sign(&message)?;
     let body = DigitallySignedStruct::new(scheme, sig);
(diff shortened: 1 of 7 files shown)


rustls/src/client/tls12.rs
```diff
@@ -490,7 +490,14 @@ fn emit_certverify(
         Some(signer) => signer,
     };
 
-    let message = transcript.take_handshake_buf();
+    let message = match transcript.take_handshake_buf() {
+        Some(msg) => msg,
+        None => {
+            trace!("Not sending CertificateVerify, no transcript");
+            return Ok(());
+        }
```
ctz, line 498: Can we match on both client_auth.signer.take() and transcript.take_handshake_buf() at once?

rustls/src/server/tls12.rs
```diff
@@ -661,12 +661,20 @@ impl hs::State for ExpectCertificateVerify {
...
+                Some(msgs) => {
+                    let certs = &self.client_cert;
+                    self.config
+                        .verifier
+                        .verify_tls12_signature(&msgs, &certs[0], sig)
+                }
+                None => {
+                    cx.common
+                        .send_fatal_alert(AlertDescription::AccessDenied);
+                    Err(Error::General("client authentication not set up".into()))
```
ctz, line 675: if this is reachable, can it have a test?

rustls/src/client/tls12.rs
ctz: (to achieve the aim of one piece of code that handles the not-sending CertificateVerify case)

djc: Yup, done.

djc: (This has two different takes on not-sending, let me know if you want to collapse those.)

rustls/src/server/tls12.rs
djc: I don't think this is actually reachable:

* We can only get to the `ExpectCertificateVerify` state if `ExpectClientKx::client_cert` is `Some`
* An `ExpectClientKx` state with `client_cert` set to `Some` is only created by `ExpectCertificate`
* In `ExpectCertificate`, we might abandon client auth, but we set `ExpectClientKx::client_cert` to `None` if so
* `ExpectCertificate` is set up only if we're doing client auth, meaning `config.verifier.offer_client_auth()` is `true`
* In `ExpectClientHello::new()`, we initialized the `HandshakeHashBuffer` with client auth enabled under the same condition

I've added a comment with a condensed version of  this reasoning.

ctz:
Thanks!

Issue #729, opened 2021-06-03
TLS-RFC Compliance


mmaehren:
Hi,
we (@jurajsomorovsky @ic0ns @mmaehren @XoMEX @Kavakuo) are performing an analysis of the RFC-compliance of open-source TLS implementations. Below we list our findings for this implementation. We admit that some are rather nit-picky, but we added them for the sake of completeness. We tried to keep the descriptions brief and didn’t want to spam the issues section so feel free to split up the list into individual issues as you see fit.
If you disagree with our interpretation of certain RFC statements, please leave feedback as we’re interested in your view.

Our results apply to the default configuration of version ***0.19.0***. We used the example implementations for client and server.

[S] = Applies to server
[C] = Applies to client
[C+S] = Applies to both

## Misc
- [S] RusTLS won't complete the handshake if a client offers only ECC cipher suites and groups but does not send an EcPointFormatsExtension
    - RFC 8422 - 5.1. Client Hello Extensions
        > A server that receives a ClientHello containing one or both of these [Supported Point Formats and Supported Elliptic Curves] extensions MUST use the client's enumerated capabilities to guide its selection of an appropriate cipher suite.
    - Note that the RFC defines that a client must send the extension but also specifically says that the server must not enforce this

- [C] RusTLS does not ignore the server's legacy version set in the ServerHello when a SupportedVersions extension was sent but terminates the handshake upon parsing an invalid version (0x0505)
    - Note that the RFC also demands that a server 'MUST' set the version to 0303 but a client is supposed to ignore the version
    - RFC 8446 - 4.2.1 Supported Versions
    > If this extension is present, clients MUST ignore the ServerHello.legacy_version value and MUST use only the "supported_versions" extension to determine the selected version.

- [C+S] RusTLS accepts zero-length handshakes fragments sent before the record containing the Finished message
    - RFC 8446 - 6.2.1 Fragmentation
        > Implementations MUST NOT send zero-length fragments of Handshake, Alert, or ChangeCipherSpec content types. Zero-length fragments of Application data MAY be sent as they are potentially useful as a traffic analysis countermeasure.

- [C] Upon receiving an encrypted Finished message that was not preceded by a Change Cipher Spec, RusTLS does not reject the record based on its content type but attempts to parse
the payload as an unencrypted message. For records with random IVs at the beginning, RusTLS assumes that the contained message is larger than the record (due to the seemingly high message length which, in fact, are just IV bytes). RusTLS therefore waits for more records.
[...]

PR #732, opened 2021-06-11
server: implement some TLS 1.3 abortions as required by the RFC

djc:
See #729.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1042,6 +1042,11 @@ impl ClientHelloPayload {
         self.find_extension(ExtensionType::EarlyData)
             .is_some()
     }
+
+    pub fn has_cookie_extension(&self) -> bool {
+        self.find_extension(ExtensionType::Cookie)
+            .is_some()
+    }
 }
 
 #[derive(Debug)]
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -362,6 +362,19 @@ impl State for ExpectClientHello {
             ProtocolVersion::TLSv1_2
         };
 
+        if version == ProtocolVersion::TLSv1_3
+            && client_hello.client_version != ProtocolVersion::TLSv1_2
+        {
+            // RFC 8446 - 4.1.2
+            // In TLS 1.3, the client indicates its version preferences in the
+            // "supported_versions" extension (Section 4.2.1) and the
+            // legacy_version field MUST be set to 0x0303, which is the version
+            // number for TLS 1.2.
+            return Err(Error::PeerMisbehavedError(
+                "TLS 1.3 ClientHello must set legacy_version to TLS 0x0303".to_string(),
+            ));
+        }
+
         cx.common.negotiated_version = Some(version);
 
         // --- Common to TLS1.2 and TLS1.3: ciphersuite and certificate selection.
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -403,6 +403,17 @@ impl State for ExpectClientHello {
             ));
         }
 
+        if version == ProtocolVersion::TLSv1_3
+            && !self.done_retry
+            && client_hello.has_cookie_extension()
+        {
+            // RFC 8446 - 4.2.2 Cookie
+            // Clients MUST NOT use cookies in their initial ClientHello in subsequent connections.
+            return Err(Error::PeerMisbehavedError(
+                "initial ClientHello contained cookie extension".to_string(),
+            ));
+        }
+
         // We communicate to the upper layer what kind of key they should choose
         // via the sigschemes value.  Clients tend to treat this extension
         // orthogonally to offered ciphersuites (even though, in TLS1.2 it is not).
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -3293,7 +3293,7 @@ mod test_quic {
             payload: MessagePayload::Handshake(HandshakeMessagePayload {
                 typ: HandshakeType::ClientHello,
                 payload: HandshakePayload::ClientHello(ClientHelloPayload {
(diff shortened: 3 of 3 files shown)


rustls/src/server/hs.rs
```diff
@@ -362,6 +362,19 @@ impl State for ExpectClientHello {
             ProtocolVersion::TLSv1_2
         };
 
+        if version == ProtocolVersion::TLSv1_3
```
djc, line 365: Alternatively, we could reject non-`TLSv1_2` legacy versions up front before doing any other processing (or here, but independent of whether we selected 1.3).

rustls/src/msgs/handshake.rs
```diff
@@ -1042,6 +1042,11 @@ impl ClientHelloPayload {
         self.find_extension(ExtensionType::EarlyData)
             .is_some()
     }
+
+    pub fn has_cookie_extension(&self) -> bool {
```
briansmith, line 1046: Not really a fan of having functions like this that only have one caller. Better to just inline it into the caller.

rustls/src/server/hs.rs
```diff
@@ -362,6 +362,19 @@ impl State for ExpectClientHello {
...
+            && client_hello.client_version != ProtocolVersion::TLSv1_2
+        {
+            // RFC 8446 - 4.1.2
+            // In TLS 1.3, the client indicates its version preferences in the
+            // "supported_versions" extension (Section 4.2.1) and the
+            // legacy_version field MUST be set to 0x0303, which is the version
+            // number for TLS 1.2.
```
briansmith, line 372: This is referencing/quoting the part of the spec that requires the client to set the value to 0x0303. Instead, we should be referencing/quoting the part of the spec that requires us to fail with an alert if the client put the wrong value in. I couldn't find that part of the spec. Is this check actually required by the spec? If not, I'm inclined to not do it.

rustls/src/server/hs.rs
```diff
@@ -403,6 +416,17 @@ impl State for ExpectClientHello {
             ));
         }
 
+        if version == ProtocolVersion::TLSv1_3
+            && !self.done_retry
+            && client_hello.has_cookie_extension()
+        {
+            // RFC 8446 - 4.2.2 Cookie
+            // Clients MUST NOT use cookies in their initial ClientHello in subsequent connections.
```
briansmith, line 424: Same. Instead of citing/quoting the part of the spec that talks about what the client does, we should cite the part of the spec that requires us to fail with an error here.

rustls/tests/api.rs
```diff
@@ -3293,7 +3293,7 @@ mod test_quic {
             payload: MessagePayload::Handshake(HandshakeMessagePayload {
                 typ: HandshakeType::ClientHello,
                 payload: HandshakePayload::ClientHello(ClientHelloPayload {
-                    client_version: ProtocolVersion::TLSv1_3,
+                    client_version: ProtocolVersion::TLSv1_2,
```
briansmith, line 3296: This issue would may have been avoided if `client_version` was instead named `legacy_version` as in the TLS 1.3 spec. We should (ideally in a separate PR) rename these fields to their TLS 1.3 names.

rustls/tests/api.rs
```diff
@@ -3412,6 +3412,118 @@ mod test_quic {
     }
 } // mod test_quic
 
+#[test]
```
briansmith, line 3415: It would be good, for these tests that test conformance with the spec, to also cite what part of the spec compliance is being tested for.

rustls/tests/api.rs
```diff
@@ -3412,6 +3412,118 @@ mod test_quic {
...
+        .unwrap();
+    assert_eq!(
+        server.process_new_packets().err(),
+        Some(Error::PeerMisbehavedError(
+            "initial ClientHello contained cookie extension".into(),
+        )),
+    );
+}
+
+#[test]
```
briansmith, line 3471: It would be good, for these tests that test conformance with the spec, to also cite what part of the spec compliance is being tested for.

briansmith (changes requested):
There are two cases of conformance checks on inputs:
1. The sender is required to do something, but the receiver isn't required to reject inputs where they don't do that.
2. The sender is required to do something, and the receiver is required to reject inputs where the sender didn't do it.

Are these instances of #1 or #2? I personally don't think it's a net positive to add extra checks beyond what the spec requires. In particular, if there's an input that we're supposed to ignore (e.g. legacy_version when the versoin negotiation extension is present) I would prefer we ignore the legacy value completely.

rustls/tests/api.rs
sayrer: fwiw, I got confused by this name while working on ECH, so I agree here.

rustls/src/msgs/handshake.rs
djc: I tend to agree, but this is prevailing style, and I'd prefer to address all of these together in a separate PR if we decide we should go that way.

ctz: I don't agree. For me, functions like this exist to make the caller more readable. It's not only about whether there are more than one caller.

djc:
@ctz do you want this or not? I'm inclined to leave it for now, given @briansmith's feedback.

briansmith:
This doesn't change the public API and isn't otherwise affecting backward compatibility so I think it can be deferred until later.

ctz:
Yes, let's leave this for later. I also started with some similar changes on 00526ea so we can coordinate what the right action is for each item in #729

briansmith:
I'd prefer to see this PR broken up and prioritized in order:
1. One or more PRs that add checks that we omit, which we are required to make according to the spec, or which we must make to be a correct and/or safe implementation. Please cite the specific requirement in the spec in a comment at the place the check is added.
2. Any other changes.

djc:
I'll just close it, I don't think I'm the best person to work on this.

PR #735, opened 2021-06-14
Make ClientConfigBuilder public

djc:
This seems to have been an oversight. Without it, no documentation is generated for `ClientConfigBuilder`.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -48,7 +48,7 @@ use std::sync::Arc;
 ///     .with_no_client_auth();
 /// ```
 ///
-/// # Resulting [`ConfigConfig`] defaults
+/// # Resulting [`ClientConfig`] defaults
 /// * [`ClientConfig::max_fragment_size`]: the default is `None`: TLS packets are not fragmented to a specific size.
 /// * [`ClientConfig::session_storage`]: the default stores 256 sessions in memory.
 /// * [`ClientConfig::alpn_protocols`]: the default is empty -- no ALPN protocol is negotiated.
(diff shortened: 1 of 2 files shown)


sayrer:
Although this patch passes the CI, there are a bunch of warnings of the form "this expression borrows a reference (...) that is immediately dereferenced by the compiler" if you read the diff.

djc:
Those are preexisting issues that are now surfacing because of new clippy lints on nightly.

sayrer:
It's not great that these things turn up on random PRs, but showing the warning is better than not. Should there be a separate PR to fix those clippy things, or should the clippy version be pinned?

briansmith:
IMO, we shouldn't be running nightly clippy, only stable clippy.

djc:
Well, the alternative is that PRs will suddenly start failing CI when those lints hit stable. I personally prefer having nightly clippy generate warnings so you can act before PRs will fail because of these.

djc:
Added a commit that fixes the clippy warnings.

sayrer:
> Well, the alternative is that PRs will suddenly start failing CI when those lints hit stable. I personally prefer having nightly clippy generate warnings so you can act before PRs will fail because of these.

I think the problem with the current approach is that it only covers files in the PR, so there's a steady stream of warnings as people work, rather than a single fix for the whole repo. For example, #741 has a bunch of similar warnings.

briansmith:
> Well, the alternative is that PRs will suddenly start failing CI when those lints hit stable. I personally prefer having nightly clippy generate warnings so you can act before PRs will fail because of these.

The problem I've run into in the past is that there are sometimes situations where you can't pass stable clippy and nightly clippy at the same time unless you (temorarily) disable some lints. 

That said, I don't have any feelings about this issue.

ctz (approved):

Issue #734, opened 2021-06-14
Session.is_handshaking() and buffers


icing:
`Session.is_handshaking()` returns `false` as soon as it has performed the TLS handshake in its internal buffers. The documentation should mention that the buffers need to be written out to the other side before any traffic can be received.

djc:
This was fixed in #737.

PR #737, opened 2021-06-15
Improve documentation and semantics around internal buffer handling

### LINKED ISSUE #733: wants_read() returns true on new ClientSession
In rustls/src/client/mod.rs line 445, the client session wants to read before the handshake has started. However, in the TLS protocol, there is never anything to read before the client sent the initial hello.

djc:
Fixes #733, #734.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -203,9 +203,11 @@ pub trait Connection: quic::QuicExt + Send + Sync {
     /// as possible.
     fn wants_write(&self) -> bool;
 
-    /// Returns true if the connection is currently performing the TLS
-    /// handshake.  During this time plaintext written to the
-    /// connection is buffered in memory.
+    /// Returns true if the connection is currently performing the TLS handshake.
+    ///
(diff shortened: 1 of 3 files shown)


rustls/src/conn.rs
```diff
@@ -835,8 +837,16 @@ impl ConnectionCommon {
         rc
     }
 
-    pub fn has_readable_plaintext(&self) -> bool {
-        !self.received_plaintext.is_empty()
+    pub fn wants_read(&self) -> bool {
+        // We want to read more data all the time, except when we have unprocessed plaintext.
+        // This provides back-pressure to the TCP buffers. We also don't want to read more after
+        // the peer has sent use a close notification.
```
jsha, line 843:
```suggestion
        // the peer has sent us a close notification.
```

jsha (commented):
One typo, otherwise looks good to me. Thanks!

rustls/src/conn.rs
djc: Good catch, fixed!

ctz:
👍🏽

Issue #736, opened 2021-06-14
Remove `webpki` types from the public API


briansmith:
There will be several backward-compatibility-breaking changes to the webpki API over the next year and beyond. Since the goal for the Rustls API is to be pretty stable and rarely have breaking changes, we need to insulate Rustls from those changes so that it can upgrade its version of webpki to the latest (and only supported) version without changing its public API.

In the case of `webpki::Error` in Rustls's `Error` type, I think we should change it to `Box<impl Debug>` or similar.

briansmith:
@djc Besides `webpki::Error`, are there any remaining `webpki` types in the public API?

djc:
* `RootCertStore::add_server_trust_anchors()` to deal with
* Methods and `From`/`Into` impls for `OwnedTrustAnchor`
* `ClientHello::server_name`
* The `DnsName` variant in `ServerName` and `<TryFrom<&str> for ServerName>::Error`

briansmith:
> * `RootCertStore::add_server_trust_anchors()`
> * [ ]  Methods and `From`/`Into` impls for `OwnedTrustAnchor`

These could be controlled with a `webpki-<version>` feature. Then we'd add new `webpki-<version+1>` feature for each webpki version.

>  ClientHello::server_name

It seems like this is a problem only because Rustls exposes a lot of internals to the outside that ideally shouldn't be exposed. I think we can/should hide these by default, and maybe provide an "unstable-api" feature that exposes them.

> The DnsName variant in ServerName and <TryFrom<&str> for ServerName>::Error

I think the goal here is to make Rustls public API operate directly on `&str` or `&[u8]` or so.

djc:
I think the only open item here is `impl From<webpki::Error> for WebPkiError`, which is fixed in #792.

djc:
This is done.

PR #738, opened 2021-06-15
Keep webpki dependency private where possible

djc:
See #736 for context.

### DIFF
--- a/rustls-mio/examples/tlsclient.rs
+++ b/rustls-mio/examples/tlsclient.rs
@@ -483,7 +483,7 @@ fn make_config(args: &Args) -> Arc<rustls::ClientConfig> {
         let mut reader = BufReader::new(certfile);
         root_store.add_parsable_certificates(&rustls_pemfile::certs(&mut reader).unwrap());
     } else {
-        root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+        root_store.add_server_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.0);
     }
 
     let suites = if !args.flag_suite.is_empty() {
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -185,14 +185,11 @@ impl rustls::ClientCertVerifier for DummyClientAuth {
         true
     }
 
-    fn client_auth_mandatory(&self, _sni: Option<&webpki::DnsName>) -> Option<bool> {
+    fn client_auth_mandatory(&self, _sni: Option<&str>) -> Option<bool> {
         Some(self.mandatory)
     }
 
-    fn client_auth_root_subjects(
-        &self,
-        _sni: Option<&webpki::DnsName>,
-    ) -> Option<rustls::DistinguishedNames> {
+    fn client_auth_root_subjects(&self, _sni: Option<&str>) -> Option<rustls::DistinguishedNames> {
         Some(rustls::DistinguishedNames::new())
     }
 
@@ -200,7 +197,7 @@ impl rustls::ClientCertVerifier for DummyClientAuth {
         &self,
         _end_entity: &rustls::Certificate,
         _intermediates: &[rustls::Certificate],
-        _sni: Option<&webpki::DnsName>,
+        _sni: Option<&str>,
         _now: SystemTime,
     ) -> Result<rustls::ClientCertVerified, rustls::Error> {
         Ok(rustls::ClientCertVerified::assertion())
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -83,14 +83,13 @@ impl RootCertStore {
 
     /// Adds all the given TrustAnchors `anchors`.  This does not
     /// fail.
-    pub fn add_server_trust_anchors(
-        &mut self,
-        &webpki::TlsServerTrustAnchors(anchors): &webpki::TlsServerTrustAnchors,
-    ) {
-        for ta in anchors {
-            self.roots
-                .push(OwnedTrustAnchor::from(ta));
-        }
+    pub fn add_server_trust_anchors<'a, I, T: 'a>(&mut self, iter: I)
+    where
+        I: IntoIterator<Item = &'a T> + 'a,
+        &'a T: Into<OwnedTrustAnchor>,
+    {
+        self.roots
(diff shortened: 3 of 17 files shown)


rustls/src/anchors.rs
```diff
@@ -34,15 +25,13 @@ impl OwnedTrustAnchor {
...
-    fn from(t: webpki::TrustAnchor) -> OwnedTrustAnchor {
-        Self::from_trust_anchor(&t)
-    }
-}
-
-impl<'a> From<&'a OwnedTrustAnchor> for webpki::TrustAnchor<'a> {
-    fn from(anchor: &'a OwnedTrustAnchor) -> Self {
-        anchor.to_trust_anchor()
+impl From<&webpki::TrustAnchor<'_>> for OwnedTrustAnchor {
+    fn from(t: &webpki::TrustAnchor) -> OwnedTrustAnchor {
```
briansmith, line 29: -> Self

rustls/src/anchors.rs
```diff
@@ -34,15 +25,13 @@ impl OwnedTrustAnchor {
...
+        OwnedTrustAnchor {
```
briansmith, line 30: Self.

I think there's a clippy lint that will do this style check for us. We should enable it.

briansmith (changes requested):

rustls-mio/examples/tlsclient.rs
```diff
@@ -483,7 +483,7 @@ fn make_config(args: &Args) -> Arc<rustls::ClientConfig> {
         let mut reader = BufReader::new(certfile);
         root_store.add_parsable_certificates(&rustls_pemfile::certs(&mut reader).unwrap());
     } else {
-        root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+        root_store.add_server_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.0);
```
briansmith, line 486: Shouldn't we instead just we implement `IntoIterator<Item = OwnedTrustAnchor>` for `TLSClientTrustAnchors` and `TLSServerTrustAnchors` and avoid the need to do the `.0` part?

rustls/src/client/mod.rs
```diff
@@ -201,7 +201,7 @@ pub enum ServerName {
     /// The server is identified by a DNS name.  The name
     /// is sent in the TLS Server Name Indication (SNI)
     /// extension.
-    DnsName(webpki::DnsName),
+    DnsName(DnsName),
```
briansmith, line 204: A `non_exhaustive` enum with public variants is already a bad and inconvenient API for callers. I think we should have `pub struct ServerName(ServerNameInner)` instead, where this enum is renamed to `ServerNameInner`.

Consider in particular that it is likely that `webpki` will have its own `IpAddress` type unless/until `std::net::IpAddr` can be moved to core.

briansmith: OK, having looked at the next commit, I'll backtrack.

> A non_exhaustive enum with public variants is already a bad and inconvenient API for callers. I think we should have pub struct ServerName(ServerNameInner) instead, where this enum is renamed to ServerNameInner.

I think we probably should do that at some point in the future, for the reason given, but elsewhere we already need to have a `DnsName` type in Rustls that we can expose in the public API.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -185,22 +185,19 @@ impl rustls::ClientCertVerifier for DummyClientAuth {
         true
     }
 
-    fn client_auth_mandatory(&self, _sni: Option<&webpki::DnsName>) -> Option<bool> {
+    fn client_auth_mandatory(&self, _sni: Option<&str>) -> Option<bool> {
```
briansmith, line 188: Why not use the new `DnsName` type that you created in an earlier commit here? That way, the fact that `sni` has been checked to be syntactically valid is encoded in the type.

briansmith: > Why not use the new `DnsName` type that you created in an earlier commit here? That way, the fact that `sni` has been checked to be syntactically valid is encoded in the type.

Not just here, but elsewhere.

Also, I think the way this API uses `sni` is not useful as explained in https://github.com/ctz/rustls/issues/473. So, maybe we should come up with a plan for 473 and see how we can remove all these `sni` uses.

rustls-mio/examples/tlsclient.rs
djc: I don't think coherence allows this, since `IntoIterator::Item` is an associated type (not a type param).

rustls/src/server/mod.rs
```diff
@@ -104,15 +104,15 @@ pub trait ResolvesServerCert: Send + Sync {
 
 /// A struct representing the received Client Hello
 pub struct ClientHello<'a> {
-    server_name: Option<webpki::DnsNameRef<'a>>,
```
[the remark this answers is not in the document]
djc: Fair enough; I kept `webpki::DnsNameRef<'a>` in the field and decoded it to `&str` in the getter. Do we expect DNS names to generally be valid UTF-8? (Was surprised to see that you store it as `&[u8]` internally.)

rustls-mio/examples/tlsclient.rs
djc: What's the reason for not exposing a slice directly anyway, but wrapping it in a custom type?

ctz:
I was thinking about what to do about Error::WebPkiError; the options seem to be:

- `Box<dyn std::error::Error>` and chuck the whole underlying error into the box. But that means consumers of the error can write `match` against them (is that correct?)
- have our own non-exhaustive 'certificate error' enum, and map the webpki error enum into it.

Both of these are improvements, since now non-webpki users can express errors with their certificate without using the webpki type. But the former is more flexible, yet much less usable for people wanting to handle specific errors. I slightly prefer the latter. Thoughts?

djc:
Mapping to our own error type seems like a decent solution. Bit boilerplatey but likely not too maintenance-intensive.

ctz:
I'll work on that and we can add that commit to this PR

ctz:
That's  62abbdb

djc:
Sorry, I ended up not having time for this today. Can continue on it on Monday if that's useful. One open question is where DnsName should live if the server wants to use it for ClientCertVerifier.

ctz:
No worries: I did the ClietnCertVerifier change and merged everything as 5cb5133e9

PR #739, opened 2021-06-15
Require more precise visibility by denying unreachable_pub

djc:
Makes it more locally obvious what's part of the public API. Also make a number of items private where possible.


ctz (approved):
👍🏽

PR #740, opened 2021-06-15
Make TLS 1.2 optional via a Cargo feature

### LINKED ISSUE #633: Make TLS 1.2 a crate feature
I think it's time to start separating TLS 1.2 as a crate feature (initially default-on).

djc:
Fixes #633.


rustls/src/cipher.rs
```diff
@@ -138,84 +52,16 @@ pub(crate) fn new_tls13_write(
     Box::new(Tls13MessageEncrypter::new(key, iv))
 }
 
-/// A `MessageEncrypter` for AES-GCM AEAD ciphersuites. TLS 1.2 only.
```
djc, line 182: Here we move all the TLS 1.2 stuff from `cipher` into `tls12::cipher`. This is nice because a lot of the interface here disappears behind `ConnectionSecrets::make_key_pair()`, allowing more privacy. On the other hand, it moves the TLS 1.2 cipher code further away from the TLS 1.3 cipher code. Alternatively, we could have an inline `tls12` module inside `crate::cipher`.

Similar trade-offs apply to the 1.2-related items in `suites` (including `Tls12CipherSuite` itself as well as the statics declaring the actual 1.2 cipher suites), where there could also be value in moving these into `crate::tls12` at the cost of moving keeping 1.2 and 1.3 suites in separate modules.

My intuition is that version-based coupling is stronger than functional coupling here, but would be interested to hear others' thoughts.

ctz:
This is looking good to me. Please add an entry to the "Crate features" section in src/lib.rs

rustls/src/cipher.rs
ctz: My opinion on this is: if we move TLS1.2-specific parts of (say) suites.rs into tls12/suites.rs then we should move the remainder to tls13/suites.rs to keep some symmetry.

djc: Yes, that was definitely part of my plan.

djc: Done.

djc:
Rebased this and added 5 commits.

briansmith:
This would require, at least, that CI/CD test both configurations with the complete test suite. It would be very easy for a `#[cfg(...)]` to disable something important that we didn't realize also affected TLS 1.3.

djc:
Added a commit to make sure we run the tests without default features. In the process, I split up the feature matrix from the platform matrix, which seems to reduce CI cycle time about 30%.

ctz (approved):

PR #741, opened 2021-06-15
Better represent random state in types

djc:
This is less code and makes it more obvious on the client that we don't have a server random yet.


sayrer:
This is a good change--noticed this working on ECH, where the client random needs to be changed upon ECH acceptance.

ctz:
Thanks!

Issue #711, opened 2021-05-25
Add ability to get Quic next_1rtt_keys without a Session


WesleyRosenblum:
Prior to #362, it was possible to derive the next 1rtt keys for QUIC without holding on to a client or server session by using the `Secrets` returned by `fn write_hs`. This allowed us to discard the TLS session immediately after the handshake and only hold on to a minimum amount of data in order to perform key updates.

Now that `Secrets` are no longer exposed to the user of Rustls, it is necessary to call `fn next_1rtt_keys` on the session every time we need to perform a key update. This means we need to maintain this TLS session for the entire lifetime of the QUIC connection, adding additional memory overhead and a major refactoring of our code.

Would it be possible for the `fn next_1rtt_keys` method to exist on the `Keys` type or `PacketKey` type itself? This would allow us to only hold onto the keys in order to perform a key update, without any additional exposure of the `Secrets`. Alternatively, allowing for the master `Secrets` to be exported would also be ok.

Thanks for considering this issue and let me know if you need any further information!

briansmith:
> This means we need to maintain this TLS session for the entire lifetime of the QUIC connection, adding additional memory overhead

I feel like it was a mistake to expose so much through the public API before. Recently work has been done to reduce the overhead of a session. I think a lot more work could be done to reduce that overhead further, perhaps on demand. We could expose a method that an application can use to free up and/or shrink buffers until the next I/O operation. We could add an API that drops all the stuff that is held indefinitely, but usually unnecessarily, like the peer's certificate chain. if this could be made to work, that seems more appealing to me. WDYT?

> and a major refactoring of our code.

It's understandably inconvenient to require a refactoring, but at the same time I think we should be designing the public API in a way where the security guarantees/invariants are maintained internally by Rustls and where keys are managed completely by Rustls and not "leaked" unnecessarily, to the greatest extent possible.

WesleyRosenblum:
It's not just the inconvenience of the refactor, having to keep the TLS session around just adds complication that we don't think is necessary. 

If `next_1rtt_keys` was available on the keys themselves, I think that would still keep the security guarantees/invariants maintained internally by Rustls, while being much more ergonomic to use.

WesleyRosenblum:
Or perhaps you could enable exporting the secrets with a feature flag so that it would only be used after deliberately turning it on.

briansmith:
[...]

PR #750, opened 2021-06-22
quic: internalize PacketKey API from Quinn

### LINKED ISSUE #748: QUIC API should encapsulate key material
Looking at the API, I see things like this:

```
pub struct PacketKey {
    /// Encrypts or decrypts a packet's payload
    pub key: aead::LessSafeKey,
    /// Computes unique nonces for each packet
    pub iv: Iv,
}
```

I don't think `key` and `iv` should be exposed publicly. Instead, I think we should have an API that looks like this:

```
    pub fn encrypt(&self, packet: u64, buf: &mut [u8], header_len: usize) {
        let (header, payload) = buf.split_at_mut(header_len);
        let (payload, tag_storage) =
            payload.split_at_mut(payload.len() - self.key.algorithm().tag_len());
        let aad = aead::Aad::from(header);
        let nonce = self.iv.nonce_for(packet);
        let tag = self
            .key
            .seal_in_place_separate_tag(nonce, aad, payload)
            .unwrap();
        tag_storage.copy_from_slice(tag.as_ref());
    }

    pub fn decrypt(
        &self,
        packet: u64,
        header: &[u8],
        payload: &mut BytesMut,
    ) -> Result<(), CryptoError> {
        if payload.len() < self.key.algorithm().tag_len() {
            return Err(CryptoError);
        }

        let payload_len = payload.len();
        let aad = aead::Aad::from(header);
        let nonce = self.iv.nonce_for(packet);
        self.key.open_in_place(nonce, aad, payload.as_mut())?;
        payload.truncate(payload_len - self.key.algorithm().tag_len());
        Ok(())
    }

    pub fn tag_len(&self) -> usize {
        self.key.algorithm().tag_len()
    }

    pub fn confidentiality_limit(&self) -> u64 {
        let cipher = self.key.algorithm();
        if cipher == &aead::AES_128_GCM || cipher == &aead::AES_256_GCM {
            2u64.pow(23)
        } else if cipher == &aead::CHACHA20_POLY1305 {
            u64::MAX
        } else {
            panic!("unknown cipher")
        }
    }

    pub fn integrity_limit(&self) -> u64 {
        let cipher = self.key.algorithm();
        if cipher == &aead::AES_128_GCM || cipher == &aead::AES_256_GCM {
            2u64.pow(52)
        } else if cipher == &aead::CHACHA20_POLY1305 {
            2u64.pow(36)
        } else {
            panic!("unknown cipher")
        }
    }
}
```

Note that I copy-pasted that directly out of quinn's rustls.rs. 

Basically, we should move the code from Quinn's rustls.rs that deals with *ring* types into Rustls's quic.rs, and remove all the APIs that expose *ring* types from Rustls's QUIC API.

This would ensure (1) We're minimizing how much key material is leaked outside of Rustls, improving security, and (2) we can upgrade *ring* to a new major version without it being a breaking API change.

djc:
Keep all *ring* types private in the QUIC API, addressing #748.

The last commit here rolls back some of the changes from #362. I believe the resulting API still satisfies the original concerns from #287, while also allowing the setup discussed in #711.

cc @Ralith @WesleyRosenblum

Fixes #748, #711.


rustls/src/quic.rs
```diff
@@ -84,12 +100,31 @@ impl DirectionalKeys {
...
 }
 
+/// A key for generating QUIC header protection masks
+pub struct HeaderProtectionKey(aead::quic::HeaderProtectionKey);
+
+impl HeaderProtectionKey {
+    /// Generate a new QUIC header protection mask
+    ///
+    /// `sample` must be exactly `sample_len()` bytes long.
+    pub fn new_mask(&self, sample: &[u8]) -> Result<[u8; 5], Error> {
```
briansmith, line 116: This should be `#[inline]`.

rustls/src/quic.rs
```diff
@@ -84,12 +100,31 @@ impl DirectionalKeys {
...
+        self.0
+            .new_mask(sample)
+            .map_err(|_| Error::General("sample of invalid length".into()))
+    }
+
+    /// Expected sample length for the key's algorithm
+    pub fn sample_len(&self) -> usize {
```
briansmith, line 123: Ditto, this should be `#[inline]`.

rustls/src/quic.rs
```diff
@@ -108,7 +145,62 @@ impl PacketKey {
...
+        self.key
+            .open_in_place(nonce, aad, payload)
+            .map_err(|_| Error::DecryptError)?;
+        Ok(payload_len - self.key.algorithm().tag_len())
+    }
+
+    /// Number of times the packet key can be used without sacrificing confidentiality
+    ///
+    /// See https://www.rfc-editor.org/rfc/rfc9001.html#name-confidentiality-limit.
+    pub fn confidentiality_limit(&self) -> u64 {
```
briansmith, line 190: 1. Why don't `encrypt()` and `decrypt()` enforce these limits?
2. If `encrypt()` and `decrypt()` were to enforce these limits then would we still need to expose them?
3. Should we be enforcing these limits within Rustls in TLS?

These functions are called frequently? Probably should be `#[inline]`.

rustls/src/quic.rs
```diff
@@ -108,7 +145,62 @@ impl PacketKey {
...
+
+    /// Number of times the packet key can be used without sacrificing integrity
+    ///
+    /// https://www.rfc-editor.org/rfc/rfc9001.html#name-integrity-limit.
+    pub fn integrity_limit(&self) -> u64 {
+        self.suite.integrity_limit
+    }
+
+    /// Tag length for the underlying AEAD algorithm
+    pub fn tag_len(&self) -> usize {
```
briansmith, line 202: First, if we need this it probably should be `#[inline]`.

But, do we really need this? I would hope that this kind of detail would be encapsulated by the `encrypt` and `decrypt` functions.

rustls/src/quic.rs
```diff
@@ -130,27 +222,25 @@ pub struct Keys {
...
+        let hs_secret =
+            hkdf::Salt::new(hkdf::HKDF_SHA256, initial_salt).extract(client_dst_connection_id);
 
         let secrets = Secrets {
             client: hkdf_expand(&hs_secret, hkdf::HKDF_SHA256, CLIENT_LABEL, &[]),
             server: hkdf_expand(&hs_secret, hkdf::HKDF_SHA256, SERVER_LABEL, &[]),
+            suite: TLS13_AES_128_GCM_SHA256_INTERNAL,
         };
-        Self::new(TLS13_AES_128_GCM_SHA256_INTERNAL, is_client, &secrets)
+        Self::new(&secrets, is_client)
```
briansmith, line 236: Shouldn't `is_client` be moved into a field of `Secrets` too? That would make `Secrets` harder to misuse because it wouldn't take `is_client` as a parameter.

rustls/src/quic.rs
```diff
@@ -108,7 +145,62 @@ impl PacketKey {
...
+    /// Encrypt a QUIC packet
+    ///
+    /// `buf` contains the header, payload and space for the AEAD tag, with the payload starting at
+    /// `header_len` and ending at `tag_len()` before the end of `buf`.
+    pub fn encrypt(&self, packet: u64, buf: &mut [u8], header_len: usize) -> Result<(), Error> {
+        let (header, payload) = buf.split_at_mut(header_len);
+        let (payload, tag_storage) =
+            payload.split_at_mut(payload.len() - self.key.algorithm().tag_len());
+        let aad = aead::Aad::from(header);
+        let nonce = self.iv.nonce_for(packet);
```
briansmith, line 161: Don't we need to also make `Iv` and maybe other parts of the public QUIC API like `nonce_for` private?

briansmith (changes requested):

rustls/src/quic.rs
Ralith: The QUIC implementation initiates a key update when the confidentiality limit is approaching, but not yet reached, so checking the limit only during `encrypt` would not be sufficient. Either it needs to be exposed, or rustls would have to make a call on what "approaching" means (which might be difficult to do in a way that satisfies every user) and that state would have to be polled.

The integrity limit could perhaps be checked during `decrypt`, but we'd need a separate error case for it so we can respond appropriately (by closing the connection) while still silently dropping packets that couldn't be authenticated otherwise.

rustls/src/quic.rs
Ralith: The QUIC implementation needs to know how much space to leave at the end of the packet, so that the tag can be written without exceeding the MTU.

rustls/src/quic.rs
```diff
@@ -130,27 +222,25 @@ pub struct Keys {
 
 impl Keys {
     /// Construct keys for use with initial packets
-    pub fn initial(
-        initial_salt: &hkdf::Salt,
-        client_dst_connection_id: &[u8],
-        is_client: bool,
-    ) -> Self {
+    pub fn initial(initial_salt: &[u8], client_dst_connection_id: &[u8], is_client: bool) -> Self {
```
Ralith, line 225: Since we've already added a `Version` enum, should we bake in the initial salt as well based on it?

Ralith:
I'd like to have a trial port of Quinn to this API before it's merged, just to ensure we don't miss anything.

djc:
I started on a trial already. It's a bit of a pain due to the builder API changes, but I'll continue on  it.

rustls/src/quic.rs
djc: The initial salt really requires more granularity than the versions, right?

rustls/src/quic.rs
djc: Alternatively we could return an `impl AsRef<[u8]>` here or something, akin to the separate tag approach that *ring* uses.

djc:
I think this now also address #751 and #752.

rustls/src/quic.rs
djc: Ah, good point, done.

rustls/src/quic.rs
djc: Makes sense, done.

rustls/src/quic.rs
djc: Done.

rustls/src/quic.rs
djc: Done.

rustls/src/quic.rs
djc: (It's `#[inline]` now.)

rustls/src/quic.rs
djc: > 3. Should we be enforcing these limits within Rustls in TLS?

Assuming you mean TCP-based TLS here as opposed to QUIC, RFC 8446 doesn't seem to mention anything like the stuff layed out in RFC 9001. In any case, IMO it would be out of scope for this PR.

rustls/src/quic.rs
```diff
@@ -108,8 +164,66 @@ impl PacketKey {
...
             iv: hkdf_expand(secret, IvLen, b"quic iv", &[]),
+            suite,
         }
     }
+
+    /// Encrypt a QUIC packet
+    ///
+    /// `buf` contains the header, payload and space for the AEAD tag, with the payload starting at
+    /// `header_len` and ending at `tag_len()` before the end of `buf`.
+    pub fn encrypt(&self, packet: u64, buf: &mut [u8], header_len: usize) -> Result<(), Error> {
```
briansmith, line 175: This API seems really error prone as designed as it requires the caller to do a bunch of slicing.

It also seems potentially suboptimal because it precludes the use of IO vectors for the header and tag to minimize the need to copy data.

AFAICT, it would be better to simply have `header: &[u8]` as a separate parameter; the caller can do `split_at_mut` themselves if they want the header and the rest to be in a single buffer.

Similarly, I think if we just returned the tag (in a newtype around `ring::aead::Tag`) then the caller could decide whether to use IOVecs or to append to the buffer.

This would eliminate all the slicing and dicing and potential for things to go wrong in the caller's invocation of `encrypt`.

I think having "_in_place" in the name would also make it clearer what's happening.

rustls/src/quic.rs
briansmith: OK. The same logic applies to both TLS and QUIC and we should factor this stuff out into a common place in the future, so that a TLS 1.3 application can also do a key update and so a TLS 1.2 application can stop if/when the keys are exhausted.

rustls/src/quic.rs
briansmith: I think if there's a way to eliminate all (any) potential for error in the choosing of the initial salt, we should do so now.

IIUC, the spec says the salt will be 0x38762cf7f55934b34d179ae6a4c80cadccbb7f0a so I'm not sure why we're asking the caller to pass it in, instead of just using that value as a constant.

rustls/src/quic.rs
```diff
@@ -238,6 +334,176 @@ pub enum Version {
 mod test {
     use super::*;
 
+    #[test]
```
briansmith, line 337: Tests of the public API should be integration tests (under `tests/`), not unit tests, whenever practical.

briansmith (changes requested):
Given the `Iv`/`nonce_for` stuff, we also need to do another sweep through the public API and see if there are old APIs that, even if they don't expose *ring* types, are now redundant and inappropriate given the current API.

rustls/src/quic.rs
briansmith: I filed issue #755 to follow up on this. From my perspective, this is resolved.

rustls/src/quic.rs
briansmith: This is resolved from my end.

rustls/src/client/tls13.rs
```diff
@@ -164,10 +164,7 @@ pub(super) fn handle_server_hello(
 
     #[cfg(feature = "quic")]
     {
-        cx.common.quic.hs_secrets = Some(quic::Secrets {
-            server: server_key,
-            client: client_key,
-        });
+        cx.common.quic.hs_secrets = Some(quic::Secrets::new(server_key, client_key, suite, true));
```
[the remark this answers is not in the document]
briansmith: Also, regardless, it seems like we need to ensure we have a test for this.

briansmith:
The codecov report says only  86.88% of the diff is tested. That seems low.

rustls/src/quic.rs
Ralith: The initial salt cannot be updated without a version change. There were multiple salts pre-1.0, but I don't think we need to support more than the latest draft salt, since it hadn't changed for some time at that point, and at least quinn isn't going to make any effort to support very old draft versions.

WesleyRosenblum:
Thanks for doing this! I tried integrating it on our side and got caught up with the builder changes, but I believe it has what we need :-)

rustls/src/quic.rs
djc: So do we decide to only support draft 29-32 and RFC 9001?

* Draft [23, 28]: 0xc3eef712c72ebb5a11a7d2432bb46365bef9f502
* Drafts [29, 32]: 0xafbfec289993d24c9e9786f19c6111e04390a899
* Drafts 33, 34, RFC 9001: 0x38762cf7f55934b34d179ae6a4c80cadccbb7f0a

rustls/src/error.rs
```diff
@@ -221,6 +221,9 @@ pub enum Error {
     /// We couldn't decrypt a message.  This is invariably fatal.
     DecryptError,
 
+    /// We couldn't encrypt a message.
+    EncryptError,
```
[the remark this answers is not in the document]
djc: Fixed.

rustls/src/error.rs
```diff
@@ -306,6 +309,7 @@ impl fmt::Display for Error {
             Error::CorruptMessage => write!(f, "received corrupt message"),
             Error::NoCertificatesPresented => write!(f, "peer sent no certificates"),
             Error::DecryptError => write!(f, "cannot decrypt peer's message"),
+            Error::EncryptError => write!(f, "cannot encrypt peer's message"),
```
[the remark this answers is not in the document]
djc: Yup, fixed.

rustls/src/client/tls13.rs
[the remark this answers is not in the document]
djc: Fixed it, will see about adding a test.

rustls/src/quic.rs
```diff
@@ -108,8 +164,66 @@ impl PacketKey {
...
+        Ok(())
+    }
+
+    /// Decrypt a QUIC packet
+    ///
+    /// Returns the length of the decrypted payload.
+    pub fn decrypt(&self, packet: u64, header: &[u8], payload: &mut [u8]) -> Result<usize, Error> {
+        if payload.len() < self.key.algorithm().tag_len() {
+            return Err(Error::DecryptError);
+        }
```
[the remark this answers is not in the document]
djc: Okay, removed.

rustls/src/quic.rs
djc: Fixed up as suggested.

rustls/src/quic.rs
djc: Moved them.

rustls/src/quic.rs
```diff
@@ -108,8 +164,66 @@ impl PacketKey {
...
+            .seal_in_place_separate_tag(nonce, aad, payload)
+            .map_err(|_| Error::EncryptError)?;
+        tag_storage.copy_from_slice(tag.as_ref());
+        Ok(())
+    }
+
+    /// Decrypt a QUIC packet
+    ///
+    /// Returns the length of the decrypted payload.
+    pub fn decrypt(&self, packet: u64, header: &[u8], payload: &mut [u8]) -> Result<usize, Error> {
```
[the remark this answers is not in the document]
djc: I guess I prefer not having the duplicate data between the `payload` argument and the return value.

> Alternatively, you could have separate `ciphertext_in_plaintext_out` and `tag` parameters, which would make the API very clear.

Not exactly sure what you mean here? *ring* does not seem to support passing in the tag separately for opening.

Renamed it and tried to improve the documentation along the lines you suggested.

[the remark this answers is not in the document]
briansmith: > Not exactly sure what you mean here? _ring_ does not seem to support passing in the tag separately for opening.

If we want to have a separate tag argument, then I'm happy to add that API to *ring*. It is needed for JWE anyway.

I'd rather design the API the way we want it, and then figure out how to implement it, rather than feel constrained by any limits of the current code.

rustls/src/quic.rs
briansmith: > So do we decide to only support draft 29-32 and RFC 9001?

Why even bother continuing to supporting older drafts. I suggest we just support the final version.

rustls/src/client/tls13.rs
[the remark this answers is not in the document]
briansmith: I definitely think we should have a test for this if at all practical since it would have found a bug before code review. It would have been easy for me to miss this in the review.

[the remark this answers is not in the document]
briansmith: Besides the test, I filed issue #765 regarding this.

rustls/src/quic.rs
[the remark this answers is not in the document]
briansmith: > I guess I prefer not having the duplicate data between the `payload` argument and the return value.

I'm not understanding this. There wouldn't be any duplication with what I'm proposing because the return slice would be borrowing the same (or a subslice) of the slice data of the input.

rustls/src/quic.rs
```diff
@@ -84,18 +117,41 @@ impl DirectionalKeys {
...
+            header: HeaderProtectionKey(hkdf_expand(secret, hp_alg, b"quic hp", &[])),
             packet: PacketKey::new(suite, secret),
         }
     }
 }
 
+/// A key for generating QUIC header protection masks
+pub struct HeaderProtectionKey(aead::quic::HeaderProtectionKey);
+
+impl HeaderProtectionKey {
```
briansmith, line 129: Looking at this again with fresh eyes, I wonder why we expose `new_mask` (and `sample_len`)? Why not just have the new encrypt/decrypt functions do the masking too?

I'm guessing we have to separate this because the QUIC implementation needs to unmask in order to keep track of packets, and then do decryption separately?

But, then, still I wonder if this is the right API. Maybe we should have `mask` and `unmask` functions that do the masking and slicing and dicing for a record?

In particular, the return value of `new_mask` is (somewhat) a secret and it would be nice to manage all the secret data within Rustls.

rustls/src/quic.rs
```diff
@@ -108,8 +164,78 @@ impl PacketKey {
...
             )),
             iv: hkdf_expand(secret, IvLen, b"quic iv", &[]),
+            suite,
         }
     }
+
+    /// Encrypt a QUIC packet
+    ///
+    /// Takes a header (which is used as the additional authenticated data) and payload. The
+    /// final `tag_len()` bytes of the payload will be overwritten with the authentication tag.
```
briansmith, line 174: The sentence "The final ..." is now inaccurate. It should state that the authentication tag is returned.

rustls/src/quic.rs
```diff
@@ -108,8 +164,78 @@ impl PacketKey {
...
+    pub fn tag_len(&self) -> usize {
+        self.key.algorithm().tag_len()
+    }
+}
+
+/// AEAD tag, must be appended to encrypted cipher text.
+pub struct Tag(aead::Tag);
+
+impl AsRef<[u8]> for Tag {
+    fn as_ref(&self) -> &[u8] {
```
briansmith, line 236: `#[inline]`

rustls/src/quic.rs
```diff
@@ -182,45 +307,50 @@ pub(crate) fn write_hs(this: &mut ConnectionCommon, buf: &mut Vec<u8>) -> Option
...
-    let keys = PacketKeySet {
-        local: PacketKey::new(suite, local),
-        remote: PacketKey::new(suite, remote),
-    };
-
-    this.quic.traffic_secrets = Some(next);
-    Some(keys)
+/// Key material for use in QUIC packet spaces
+#[allow(clippy::large_enum_variant)]
+pub enum KeyMaterial {
```
briansmith, line 331: Usually we use the term "key material" to refer to the raw bytes of the key, which I think we're trying not to expose outside of Rustls, so `KeyMaterial` seems like a bad name.

rustls/src/quic.rs
```diff
@@ -54,23 +92,18 @@ pub trait QuicExt {
     /// Emit unencrypted TLS handshake data.
     ///
     /// When this returns `Some(_)`, the new keys must be used for future handshake data.
```
briansmith, line 94: I have trouble understanding how `KeyMaterial` is to be used and I think it would be good to clarify how to use each variant. I think `KeyMaterial`'s and/or this function's documentation needs to be updated/expanded to explain how to use them, especially so people maintaining Rustls can understand it better.

rustls/src/quic.rs
```diff
@@ -171,7 +296,7 @@ pub(crate) fn read_hs(this: &mut ConnectionCommon, plaintext: &[u8]) -> Result<(
     Ok(())
 }
 
-pub(crate) fn write_hs(this: &mut ConnectionCommon, buf: &mut Vec<u8>) -> Option<Keys> {
+pub(crate) fn write_hs(this: &mut ConnectionCommon, buf: &mut Vec<u8>) -> Option<KeyMaterial> {
```
briansmith, line 299: Would it make sense to have two `write_hs` functions, the first which returns `Option<HandshakeKey>` and then second of which returns `Option<OneRttKey>`, where the QUIC implementation would call the first one until it returns Some, and then call the second one for the rest?

rustls/src/quic.rs
djc: Got any suggestions?

briansmith: IIUC, `KeyMaterial` represents a notification that the keys used by the QUIC implementation need to be replaced. So maybe `TrafficKeyChange` or `NewTrafficKey`. I was thinking also `KeyUpdate` but that term has specific meaning in TLS.

rustls/src/quic.rs
Ralith: I don't have a strong opinion either way, but if we drop draft version salt support we should also strip out the draft TLS extension number support. Maybe keep the version enum/interface around, since there's talk of an early, trivial QUIC 2 for anti-ossification purposes

rustls/src/quic.rs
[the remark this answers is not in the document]
djc: There would still be two separate ways of getting to the plain text data, which makes the API harder to understand IMO. That's the reason I didn't do it, anyway. If you feel strongly about it I can change it.

rustls/src/quic.rs
djc: I think we could do something like this:

```rust
fn xor(&self, sample: &[u8], first: &mut u8, length: &mut [u8]) -> Result<(), Error>
```

I don't love the name, but can't come up with something better for something that does both masking and unmasking.

rustls/src/quic.rs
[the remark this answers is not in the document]
briansmith: OK, I tried to implement my idea and I ran into an issue with my suggestion. Depending on how the QUIC implementation does buffer management, some of my ideas make less sense. If QUIC had defined its packet structure so that the header and tag were at the beginning then the IOVec stuff would work much better. If the QUIC implementation doesn't use IOVecs to read the header into a separate buffer from the data, then the API I've proposed is not great.

WDYT?

rustls/src/quic.rs
djc: How about just `KeyChange`? The first set of keys is for handshake, not traffic.

rustls/src/quic.rs
djc: That seems like a pretty awkward API. The current API seems pretty precisely typed -- I don't really understand why having to functions would be an improvement.

rustls/src/quic.rs
djc: So we have multiple things we want to return. When the handshake has started, we return handshake keys. These include header and packet keys, and we don't need to do key updates for these keys. When traffic can start, we get a new set of header and packet keys. Then, we also need to get a set of `Secrets` so we can derive future key updates when either the peer or us finds it necessary. Key updates only update the packet keys, not the header keys, so instead of just returning the `Secrets` from which the first set of 1-RTT keys can be derived, we return that full set (including the header protection keys), and yield the `Secrets` to be used for the first key update.

rustls/src/quic.rs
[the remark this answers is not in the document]
djc: I think we have one `IoSlice` per packet in Quinn, rather than having multiple slices per packet. Apart from the `IoSlice` stuff, I still kind of like separate tag opening because it puts all the slicing (header/payload/tag) in one place rather than having the header/payload split and payload/tag split handled differently.

(I was also wondering before if a similar sealing API that takes a `tag: &mut [u8]` argument rather than returning a `Tag` could avoid an additional copy.)

rustls/src/client/tls13.rs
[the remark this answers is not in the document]
djc: Looks like I removed tests that could have caught this in the first commit here. I think I can just fix those up.

rustls/src/quic.rs
djc: As I understand from the QUIC mailing list, draft 29 will be around for a bit at least in browsers.

rustls/src/quic.rs
djc: Fixed.

rustls/src/quic.rs
djc: Fixed.

rustls/src/quic.rs
djc: I added a longer comment for the type now named `KeyChange`.

rustls/src/quic.rs
djc: Renamed it as proposed.

rustls/src/client/tls13.rs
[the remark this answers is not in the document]
djc: Updated the tests that checked for this and verified that they would have caught this issue.

rustls/src/quic.rs
djc: Went ahead and added this, renaming it to `xor_in_place()`.

rustls/src/quic.rs
djc: I implemented this, keeping `V1Draft` and documenting it explicitly as supporting drafts 29 through 32.

djc:
Now hitting 96.29% diff coverage; refined the API a bunch more and add some changelog notes.

ctz:
Happy for this to be merged when you and other Quinn maintainers are happy with it.

ctz (approved):

rustls/src/quic.rs
```diff
@@ -94,23 +94,170 @@ pub trait QuicExt {
...
+    /// the correct length (see [Header Protection Sample] and [`Self::sample_len()`]),
+    /// or `packet_number` is longer than allowed (see
+    /// [Packet Number Encoding and Decoding]).
+    ///
+    /// Otherwise, `first` and `packet_number` will have the header protection removed.
+    ///
+    /// [Header Protection Application]: https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.1
+    /// [Header Protection Sample]: https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.2
+    /// [Packet Number Encoding and Decoding]: https://datatracker.ietf.org/doc/html/rfc9000#section-17.1
+    pub fn unmask_in_place(
```
djc, line 211: Changed: we now have separate `mask_in_place()` and `unmask_in_place()` methods. These turn out to be not quite symmetric because we need the unmasked packet number length to decide how many bytes of the packet number to mask/unmask.

rustls/src/suites.rs
```diff
@@ -328,6 +332,10 @@ pub static TLS13_CHACHA20_POLY1305_SHA256: SupportedCipherSuite =
             aead_algorithm: &ring::aead::CHACHA20_POLY1305,
         },
         hkdf_algorithm: ring::hkdf::HKDF_SHA256,
+        #[cfg(feature = "quic")]
+        confidentiality_limit: u64::MAX,
+        #[cfg(feature = "quic")]
+        integrity_limit: 68_719_476_736, // MSRV 1.50: 2u64.pow(36)
```
djc: Added the literals here to be somewhat more conservative about the MSRV.

djc (commented):
Made some more changes. This can now pass the Quinn tests.

rustls/src/suites.rs
WesleyRosenblum: You could use `integrity_limit: 1u64 << 36`

djc: Duh, I knew there was a better way I was missing.

briansmith:
> ```
>  integrity_limit: 68_719_476_736, // MSRV 1.50: 2u64.pow(36)
> ```
> Added the literals here to be somewhat more conservative about the MSRV.

MSRV is already 1.52.1: https://github.com/ctz/rustls/blob/main/.github/workflows/build.yml#L19.

briansmith:
> 96.05% of diff hit (target 96.52%)

What isn't being tested? @ctz is there a way to give us access to the codecov reports?

djc:
I do have access to the codecov reports, but the remaining uncovered things seem to be flukes in coverage support. The setting of the alert to `DecodeError` in `quic::read_hs()` is pretty much the only thing not covered in the quic module, but that wasn't covered before as well.

rustls/tests/api.rs
```diff
@@ -3653,22 +3668,24 @@ mod test_quic {
             0x18, 0xab, 0x08, 0x56, 0x97, 0x2e, 0x19, 0x4c, 0xd9, 0x34,
         ];
 
-        assert_eq!(buf.len(), PROTECTED.len() - PLAIN_HEADER.len() - tag_len);
-        assert_eq!(
-            &buf,
-            &PROTECTED[plain_header_len..PROTECTED.len() - tag_len]
-        );
-        assert_eq!(tag.as_ref(), &PROTECTED[PROTECTED.len() - tag_len..]);
+        assert_eq!(buf.len(), PROTECTED.len());
```
Ralith, line 3682: This assert is redundant.

rustls/tests/api.rs
```diff
@@ -3689,6 +3689,11 @@ mod test_quic {
             do_exporter_test(client_config, server_config);
         }
     }
+
+    #[test]
+    fn test_quic_version_can_clone() {
+        let _ = quic::Version::V1.clone();
+    }
```
Ralith, line 3718: Is this really a risk? `#[derive(Clone)]` seems unlikely to malfunction.

rustls/tests/api.rs
```diff
@@ -3067,36 +3067,39 @@ mod test_quic {
...
-            quic::KeyChange::Handshake { keys } => keys,
-            quic::KeyChange::OneRtt { keys, .. } => keys,
-        }))
+        Ok(change)
     }
 
     #[test]
     fn test_quic_handshake() {
-        fn equal_dir_keys(x: &quic::DirectionalKeys, y: &quic::DirectionalKeys) -> bool {
+        fn equal_dir_keys(x: &quic::PacketKey, y: &quic::PacketKey) -> bool {
```
Ralith, line 3075: s/dir/packet/ or similar

rustls/src/quic.rs
```diff
@@ -519,6 +519,56 @@ impl Version {
...
+        use crate::tls13::TLS13_CHACHA20_POLY1305_SHA256_INTERNAL;
+        let hpk = HeaderProtectionKey::new(TLS13_CHACHA20_POLY1305_SHA256_INTERNAL, &secret);
+        let packet = PacketKey::new(TLS13_CHACHA20_POLY1305_SHA256_INTERNAL, &secret);
+
+        const PLAIN: &[u8] = &[0x42, 0x00, 0xbf, 0xf4, 0x01];
+
+        let mut buf = PLAIN.to_vec();
+        let (header, payload_tag) = buf.split_at_mut(4);
+        let tag = packet
+            .encrypt_in_place(PN, &*header, payload_tag)
```
Ralith, line 543: s/payload_tag/payload/; the tag isn't there yet

rustls/src/quic.rs
```diff
@@ -519,6 +519,56 @@ impl Version {
...
+        let (header, payload_tag) = buf.split_at_mut(4);
+        let (first, rest) = header.split_at_mut(1);
+        let sample = &payload_tag[1..1 + hpk.sample_len()];
+        hpk.unmask_in_place(sample, &mut first[0], rest)
+            .unwrap();
+        let _ = packet
+            .decrypt_in_place(PN, &*header, payload_tag)
+            .unwrap();
+
+        assert_eq!(&buf[..PLAIN.len()], PLAIN);
```
Ralith: Should we instead compare the return value of `decrypt_in_place` to `PLAIN` directly here?

rustls/src/quic.rs
```diff
@@ -95,23 +95,152 @@ pub trait QuicExt {
...
+    ///
+    /// [Header Protection Application]: https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.1
+    /// [Header Protection Sample]: https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.2
+    /// [Packet Number Encoding and Decoding]: https://datatracker.ietf.org/doc/html/rfc9000#section-17.1
+    #[inline]
+    pub fn mask_in_place(
+        &self,
+        sample: &[u8],
+        first: &mut u8,
+        packet_number: &mut [u8],
```
Ralith, line 151: In Quinn we pass `pn_offset: usize, packet: &mut [u8]`. That's one less argument to worry about, and no confusion about whether `packet_number` needs to be precise or not. Is there a good reason not to do the same here? I guess we might not want to restrict users to having the header laid out contiguously with the rest of the packet...

Also applies to `unmask_in_place`.

rustls/src/quic.rs
```diff
@@ -95,23 +95,152 @@ pub trait QuicExt {
...
+    /// the correct length (see [Header Protection Sample] and [`Self::sample_len()`]),
+    /// or `packet_number` is longer than allowed (see [Packet Number Encoding and Decoding]).
+    ///
+    /// Otherwise, `first` and `packet_number` will have the header protection added.
+    ///
+    /// [Header Protection Application]: https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.1
+    /// [Header Protection Sample]: https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.2
+    /// [Packet Number Encoding and Decoding]: https://datatracker.ietf.org/doc/html/rfc9000#section-17.1
+    #[inline]
+    pub fn mask_in_place(
```
Ralith, line 147: Would it make sense to rename this `encrypt_in_place` (and similarly rename `unmask_...`) for consistency with `PacketKey`?

Ralith (changes requested):
Made a deep pass. Overall I'm happy with the direction of this. Aside from the packet number crypto bug (could we improve testing there?) Quinn's passing tests and looking reasonably tidy with these changes.

rustls/src/quic.rs
djc: Yeah, not having the requirement for contiguous layout seems like a decent argument for this. Also, no need to give the header protection key access to the entire packet? ([Previous discussion](https://github.com/ctz/rustls/pull/750#discussion_r658992013).)

rustls/src/quic.rs
djc: I don't have a strong opinion on this, so I renamed as suggested. @briansmith any thoughts on this?

rustls/tests/api.rs
djc: We have a few other tests like this, presumably to increase test coverage?

rustls/src/quic.rs
```diff
@@ -55,48 +84,174 @@ pub trait QuicExt {
...
+            .new_mask(sample)
+            .map_err(|_| Error::General("sample of invalid length".into()))?;
+
+        // The `unwrap()` will not panic because `new_mask` returns a
+        // non-empty result.
+        let (first_mask, pn_mask) = mask.split_first().unwrap();
+
+        // It is OK for the `mask_packet_number` to be longer than `packet_number`
+        // but a valid `packet_number` will never be longer than
+        // `mask_packet_number`.
```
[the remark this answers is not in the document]
djc: Fixed up the docs.

rustls/src/quic.rs
```diff
@@ -55,48 +84,174 @@ pub trait QuicExt {
...
+            true => 0x0f,  // Long header: 4 bits masked
+            false => 0x1f, // Short header: 5 bits masked
+        };
+
+        let pn_len = 1 + match masked {
+            // When unmasking, use the packet length bits after unmasking
+            true => (*first ^ (first_mask & bits)),
+            // When masking, use the packet length bits before masking
+            false => *first,
+        } as usize;
```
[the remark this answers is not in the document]
djc: Oh, good catch. I wrote the test for this with too much out-of-bounds knowledge. Rewrote the tests to catch this.

rustls/src/cipher.rs
```diff
@@ -29,17 +29,17 @@ impl dyn MessageDecrypter {
 /// A write or read IV.
 #[allow(unreachable_pub)] // Only exposed as part of `crate::quic`
```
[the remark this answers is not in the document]
djc: Nice, done.

rustls/tests/api.rs
djc: Done.

rustls/src/quic.rs
djc: Yup; fixed.

rustls/tests/api.rs
Ralith: Does the coverage checker actually count derived traits? Weird.

rustls/src/quic.rs
Ralith: Fair enough.

rustls/tests/api.rs
djc: I removed it and it didn't seem to change coverage, so that seems fine.

Ralith (approved):

PR #761, opened 2021-06-24
Enforce clippy's use_self lint

### LINKED ISSUE #753: Enforce clippy `use_self` rule in CI
When reviewing #750 I noticed at least two cases where we could be using `-> Self` and `Self { }` instead of the type name. Depending on who writes the code we have different styles. We should unify the style to the `use_self` style.

djc:
Fixes #753.


briansmith (approved):
Thanks for doing this!

Issue #758, opened 2021-06-24
ClientConfig and ServerConfig vs Builders


jsha:
In the latest master, ClientConfig and ServerConfig no longer have a `new()`. There are new Builder types to construct them, but someone attempting to port old code without many changes might be tempted to try and instantiated a ClientConfig or ServerConfig by setting all of their fields. That's currently impossible, because `verifier` on both types is private.

Is that an intentional choice, to prevent construction of ClientConfig and ServerConfig without going through the Builders? If so, we should document it on the respective structs. Also if that's the case: should all the other fields be made private, if the intent is only to support construction via Builder?

jsha:
FWIW, it would be quite convenient for crustls' purposes if `verifier` were public; that would allow it to maintain more or less the builder API it currently has, at least for now.

djc:
IMO we should keep the `verifier` field private, and probably also make the `cipher_suites`, `kx_groups` and `versions` fields private (since these are also set through the builder API). If we don't, the validation we do as part of the builder construction becomes somewhat useless. Also keeping the legacy config construction API working in rustls-ffi doesn't seem like great justification for this by itself?

ctz:
Yes, `verifier` should be kept private really. I wonder if the error would be better if we set these structs as `non_exhaustive`?

If you think it's useful, I wouldn't be strongly opposed to introducing `new()` on these types, with the following body (from the docs):

```
ConfigBuilder::with_safe_defaults()
    .for_server()
    .unwrap();
```

jsha:
I think mainly what's needed is a sentence in the documentation for ServerConfig and ClientConfig that says "these cannot be constructed manually; you must use ConfigBuilder to create one."

In terms of consistency of the API, it's odd that a user is allowed to modify post-hoc any field of a ServerConfig or a ClientConfig, _except_ verifier. The issue of gating behind dangerous_configation makes sense - but what about a `set_verifier` method that is only present under dangerous_configuration?

briansmith:
I agree with making all of the fields private in general. If there are any that need to be public, let's hear the reasoning for each one on a case-by-case basis and try to find alternate solutions.

I think making all the fields private will help make the API less confusing.

djc:
It looks like making all the (builder-specified) fields private will make things that `bogo_shim` and the tests need to do a bunch harder, though:

https://github.com/ctz/rustls/blob/main/rustls/examples/internal/bogo_shim.rs#L370

jsha:
[...]

PR #763, opened 2021-06-24
Make verifier field of configs public

jsha:
This matches the visibility of other fields on ServerConfig and ClientConfig.

Related to #758


briansmith:
I think it is intentional that `verifier` isn't public because this has to do with the "dangerous configuration" idea, so @ctz should review this.

jsha:
Closing based on further conversation in #758.

PR #764, opened 2021-06-24
CI: stop installing LLVM 7

### LINKED ISSUE #762: Use newer version of LLVM in code coverage CI jobs
In `.github/workflows/build.yml` we have:
```
      - name: Install dependencies
        run: sudo apt-get install llvm llvm-7 libclang-common-7-dev llvm-7-dev
```

llvm-7 is very old. Recent versions of Rust require clang-10 or higher to enable cross-language LTO and similar.

djc:
Fixes #762. Let's see if it works.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -159,9 +159,6 @@ jobs:
         with:
           go-version: "1.14"
 
-      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


ctz (approved):
Well that totally worked

PR #766, opened 2021-06-26
Rename builders from "With" to "Wants"

jsha:
This makes for a clearer connection between each type in the builder
state machine and the methods it provides.

This also removes ConfigBuilder as a type, since it's really just the
entry point into the state machine; it's now "ConfigWantsCipherSuites."

The entry point into the state machine is now one of two functions:

  config_builder() // full configuration
  config_builder_with_safe_defaults() // pick everything for me

This improves consistency: all settings are now added via methods.
Previously, the first-round settings (cipher suites) were set by
associated function, while everything else was set by method.

This is a followup to my suggestion at https://github.com/ctz/rustls/issues/758#issuecomment-868966850; there is some additional justification there.

You can see generated documentation for this branch at https://hoffman-andrews.com/rustls-with-to-wants/rustls/?search=config:

image

I still have a few touch-ups on the docs to make, but wanted to get feedback on the idea before proceeding.

### DIFF
--- a/fuzz/fuzzers/client.rs
+++ b/fuzz/fuzzers/client.rs
@@ -1,20 +1,23 @@
 #![no_main]
-#[macro_use] extern crate libfuzzer_sys;
+#[macro_use]
+extern crate libfuzzer_sys;
 extern crate rustls;
 extern crate webpki;
 
-use rustls::{ConfigBuilder, ClientConnection, Connection, RootCertStore};
+use rustls::{config_builder_with_safe_defaults, ClientConnection, Connection, RootCertStore};
+use std::convert::TryInto;
 use std::io;
 use std::sync::Arc;
-use std::convert::TryInto;
 
 fuzz_target!(|data: &[u8]| {
     let root_store = RootCertStore::empty();
-    let config = Arc::new(ConfigBuilder::with_safe_defaults()
-        .for_client()
-        .unwrap()
-        .with_root_certificates(root_store, &[])
-        .with_no_client_auth());
+    let config = Arc::new(
+        config_builder_with_safe_defaults()
+            .for_client()
+            .unwrap()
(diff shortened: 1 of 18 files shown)


ctz:
I think these changes are a good improvement 👍🏽

djc:
I like the naming improvements. Not a fan of functions as entry points, which seems pretty unidiomatic.

jsha:
Alright, I've touched up the docs (and updated them at https://hoffman-andrews.com/rustls-with-to-wants/rustls/?search=config). Ready for review.

> Not a fan of functions as entry points, which seems pretty unidiomatic.

The alternative seems to be creating a ConfigBuilder type whose only purpose is to hold associated methods `new()` and `with_safe_defaults()`. But idiomatically, if there is a `ConfigBuilder::new()`, it should return ConfigBuilder, not a ConfigWantsCipherSuites, which is the actual start of our type machine. I think using a function here is not so terribly unidiomatic (and not so different from associated methods).

However, there's one weird idea that might allow us to act more like ConfigBuilder is a real object. For the three states that have a safe default available (WantsCipherSuites, WantsKxGroups, WantsVersions), we could define a Deref impl that goes to the next state using the safe default. This would have the advantage that if you only want to specify one of the three, you can call that one method without having to write out `with_safe_x_default` for the two others.

Having introduced that pattern, it might also make sense to define a struct ConfigBuilder whose only purpose is to Deref into ConfigWantsCipherSuites. In that case we could make an associated function ConfigBuilder::new() that returns a ConfigBuilder.

I think this is probably too fancy, and maybe an abuse of Deref, but I thought I'd mention it in case it appeals.

ctz:
Merge as part of 5fedd1d

ctz:
Thanks!

PR #768, opened 2021-06-28
Add UnsupportedNameType error for certificate validation routines

complexspaces:
Based off [this comment](https://github.com/ctz/rustls/issues/758#issuecomment-869027675), this PR adds an error variant for custom certificate validation implementations to return when they can't find the correct `ServerName` variant when trying to validate a certificate. 

Intended use case:
```rust
impl ServerCertVerifier for Verifier {
    fn verify_server_cert(&self, 
        end_entity: &rustls::Certificate, 
        intermediates: &[rustls::Certificate], 
        server_name: &rustls::ServerName, 
        scts: &mut dyn Iterator<Item = &[u8]>, 
        ocsp_response: &[u8], 
        now: std::time::SystemTime
    ) -> Result<rustls::ServerCertVerified, Error> {
        ...
        let name = match server_name {
            rustls::ServerName::DnsName(n) => n.as_ref(),
            _ => return Err(Error::UnsupportedNameType)
        };
}
```

### DIFF
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -218,6 +218,9 @@ pub enum Error {
     /// The peer didn't give us any certificates.
     NoCertificatesPresented,
 
+    /// The certificate verifier doesn't support the given type of name.
+    UnsupportedNameType,
+
     /// We couldn't decrypt a message.  This is invariably fatal.
     DecryptError,
 
@@ -305,6 +308,7 @@ impl fmt::Display for Error {
             }
             Error::CorruptMessage => write!(f, "received corrupt message"),
             Error::NoCertificatesPresented => write!(f, "peer sent no certificates"),
+            Error::UnsupportedNameType => write!(f, "presented server name type wasn't supported"),
             Error::DecryptError => write!(f, "cannot decrypt peer's message"),
             Error::PeerSentOversizedRecord => write!(f, "peer sent excess record size"),
             Error::HandshakeNotComplete => write!(f, "handshake not complete"),


rustls/src/error.rs
```diff
@@ -218,6 +218,9 @@ pub enum Error {
     /// The peer didn't give us any certificates.
     NoCertificatesPresented,
 
+    /// The presented server name wasn't valid for certificate validation.
```
briansmith, line 221: I suggest "The certificate verifier doesn't support the given type of name."

rustls/src/error.rs
```diff
@@ -305,6 +308,7 @@ impl fmt::Display for Error {
             }
             Error::CorruptMessage => write!(f, "received corrupt message"),
             Error::NoCertificatesPresented => write!(f, "peer sent no certificates"),
+            Error::UnsupportedNameType => write!(f, "presented server name wasn't valid"),
```
briansmith, line 311: I suggest "wasn't supported".

briansmith (changes requested):

rustls/src/error.rs
complexspaces: 👍 Thank you.

rustls/src/error.rs
complexspaces: That's much better then what I had, thanks!

rustls/src/error.rs
```diff
@@ -305,6 +308,7 @@ impl fmt::Display for Error {
             }
             Error::CorruptMessage => write!(f, "received corrupt message"),
             Error::NoCertificatesPresented => write!(f, "peer sent no certificates"),
+            Error::UnsupportedNameType => write!(f, "presented server name was't supported"),
```
djc, line 311: I think you missed the `type` aspect?

complexspaces: I did, thank you for the catch 🤦

briansmith (approved):

djc:
Thanks!

PR #769, opened 2021-06-28
Extract TLS 1.3 AAD length into constant

complexspaces:
This makes the TLS 1.3 cipher implementation more consistent with the TLS 1.2 one, and also brings further clarity by removing a partially "magic number."


djc:
Thanks!

PR #771, opened 2021-06-29
builder: use type state to improve config builders

djc:
Inspired by https://matklad.github.io/2020/08/12/who-builds-the-builder.html.

(See discussion in #758.)

### DIFF
--- a/rustls/src/builder.rs
+++ b/rustls/src/builder.rs
@@ -70,17 +70,17 @@ use std::marker::PhantomData;
 ///     .with_no_client_auth();
 /// ```
 ///
-///
 /// The types used here fit together like this:
 ///
 /// 1. Call [`ClientConfig::builder()`] or [`ServerConfig::builder()`] to initialize a builder.
 /// 1. You must make a decision on which cipher suites to use, typically
-///    by calling [`ConfigWantsCipherSuites::with_safe_default_cipher_suites()`].
+///    by calling [`ConfigBuilder<S, WantsCipherSuites>::with_safe_default_cipher_suites()`].
 /// 2. Now you must make a decision
-///    on key exchange groups: typically by calling [`ConfigWantsKxGroups::with_safe_default_kx_groups()`].
+///    on key exchange groups: typically by calling
+///    [`ConfigBuilder<S, WantsKxGroups>::with_safe_default_kx_groups()`].
 /// 3. Now you must make
 ///    a decision on which protocol versions to support, typically by calling
-///    [`ConfigWantsVersions::with_safe_default_protocol_versions()`].
+///    [`ConfigBuilder<S, WantsVersions>::with_safe_default_protocol_versions()`].
 /// 5. Now see [`ConfigWantsServerVerifier`] or [`ConfigWantsClientVerifier`] for further steps.
 ///
 /// [`ServerConfig`]: crate::ServerConfig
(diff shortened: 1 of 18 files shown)


jsha:
I definitely like the approach of putting ::builder on ServerConfig and ClientConfig. One issue currently is that this results in the main "builder" documentation being on a page named "ConfigWantsCipherSuites," which feels somewhat arbitrary. That could be resolved with documentation, for instance the first sentence saying "This is the first in a sequence of types used as part of a [builder pattern] to construct a ServerConfig or ClientConfig."

image

Also, ConfigSide is not clickable in the documentation. I think it would be useful to export it so it's clickable and can hold a short piece of documentation saying "this indicates whether we are building a config for a client or a server."

djc:
Updated: this now rewrites the config builders to leverage type states as suggested in #758. All of the state's methods are now documented on the new `ConfigBuilder` type. I've also made the `ConfigSide` trait public.

rustls/src/builder.rs
```diff
@@ -62,81 +61,70 @@ use crate::versions;
...
 
-fn config_builder_with_safe_defaults() -> ConfigWantsPeerType {
-    config_builder()
-        .with_safe_default_cipher_suites()
-        .with_safe_default_kx_groups()
-        .with_safe_default_protocol_versions()
-}
+/// Config builder state where the caller must supply cipher suites.
+#[derive(Clone)]
+pub struct WantsCipherSuites;
```
hawkw, line 101: nit, take it or leave it: this will allow users to publicly construct this type, and means adding any fields in this state is a breaking change. although it won't really be useful to publicly construct this, i think it might be worth adding a private field so user code can't construct it, especially in case it becomes necessary to add fields later?
```suggestion
pub struct WantsCipherSuites {
   _private: (),
}
```

rustls/src/builder.rs
```diff
@@ -62,81 +61,70 @@ use crate::versions;
...
-        .for_server()
-        .unwrap()
-}
-
-/// A config builder where we want to know the cipher suites.
-pub struct ConfigWantsCipherSuites;
+impl<S: ConfigSide> ConfigBuilder<S, WantsCipherSuites> {
+    /// Start side-specific config with defaults for underlying cryptography.
+    ///
+    /// These are safe defaults, useful for 99% of applications.
```
hawkw, line 106: this is not directly related to the change made here, since the doc comment was already present in the previous code, but it seems like it might be nice to document what the defaults _are_?

rustls/src/builder.rs
```diff
@@ -62,81 +61,70 @@ use crate::versions;
...
 ///
 /// [`ServerConfig`]: crate::ServerConfig
 /// [`ClientConfig`]: crate::ClientConfig
-pub fn config_builder() -> ConfigWantsCipherSuites {
-    ConfigWantsCipherSuites {}
+/// [`ClientConfig::builder()`]: crate::ClientConfig::builder()
+/// [`ServerConfig::builder()`]: crate::ServerConfig::builder()
+/// [`ConfigWantsServerVerifier`]: crate::ConfigWantsServerVerifier
+/// [`ConfigWantsClientVerifier`]: crate::ConfigWantsClientVerifier
+#[derive(Clone)]
```
hawkw, line 93: should the builder derive `Debug`? if we don't want to implement `Debug` for the various states, it could just be something like this:
```rust
impl<Side: ConfigSide, State> fmt::Debug for ConfigBuilder<Side, State> {
    fn fmt(&self, f: &mut fmt::Formatter<'_> -> fmt::Result {
        f.debug_struct("ConfigBuilder")
            .field("side", format_args!("{}", &std::any::type_name::<Side>()))
            .field("state", format_args!("{}", &std::any::type_name::<State>()))
            .finish()
    }
}
```
or, we could derive `Debug` for the state types, if we want to format their data, and use `type_name` for the `Side` parameter.

hawkw (approved):
this looks great --- i think it should be much easier to navigate the docs when all the methods are present on one type.

it might be nice to add something in the docs for `ConfigBuilder` listing the states it transitions between, in order?

another thought is that I wonder if the various state types could derive `Default`, and there could be some kind of generic `builder.with_defaults()` that advances the builder to the next state. but, this would probably require some trait magic, and may not actually be useful --- it's probably nicer to explicitly say _which_ set of configurations you're accepting the defaults for...

djc:
Some follow-up in #776.

PR #774, opened 2021-07-04
Export IoState

eloff:
IoState is used in the public interface in the return value of process_new_packets, but isn't public (it's inside the private module conn)

This caused trouble for me when using rustls and defining a wrapper method for process_new_packets, I couldn't declare the return type.

I'm actually surprised Rust allows this.


djc:
Thanks!

PR #778, opened 2021-07-08
Reduce MSRV to 1.50.0

djc:
We previously raised the MSRV from 1.47 to 1.51 in #679 and to 1.52.1 in #728. Given that Quinn has a MSRV of 6 months ago (1.50 was released on Feb 11, so it will be 6 months in a month from now), which was [requested](https://github.com/quinn-rs/quinn/issues/979) by the hyper team because they want to keep a 6 month old MSRV (and of course hyper is also relatively large downstream user of rustls). My feeling is that raising the MSRV too aggressively will result in downstream users deferring rustls updates, which would be unfortunate IMO, or even using alternatives.

From #679:

> Rust 1.51.0 was released 2021-03-25 which is very recently. However, it is the first stable release that supports the `min_const_generics` feature. Many Rust-based projects have upgraded their MSRV to 1.51.0 for this reason. Cargo 1.51.0 now also supports the new feature resolver for the first time, which is another very important change that will likely be useful for us.

> I would like to use the `min_const_generics` feature in *ring*, and I expect the new feature resolver will be useful for webpki and/or ring soon, if not also by Rustls.

However, *ring* on current main still works with 1.50.0. I think we should postpone raising the MSRV to 1.51 until it's actually used on *ring*. Raising the MSRV in a  semver-compatible version should be fine (most of the ecosystem seems to have settled on doing so, provided that the new version is not too aggressive), so *ring* can raise it in a compatible rustls release (and with #750, we no longer have a public dependency on *ring* AFAIK, so we could also use a semver-incompatible version of *ring* within the same semver-compatible rustls release series).

From  #728:

> https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html recommends using 1.52.1 or later due to compiler bugs in all earlier versions. Stop trying to support older versions.

I don't think this is a good rationale for raising the rustls MSRV. Yes, downstream users should all pick up the latest stable. The MSRV mechanism as used in CI is mainly useful for awareness of contributors that they're using features that are unsupported in slightly older versions of the compiler. When we raise the MSRV proactively (as in #679 and #728), we don't even notice when we raise the actual MSRV. Downstream users might still use an older compiler and won't even notice that it's not supported because rustls will just compile, and might get surprised in a bugfix release when we do actually start to require a newer compiler/language (which we might not even notice because we already raised the MSRV).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.52.1 # MSRV as recommend by https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html
+          - 1.50.0
         os: [ubuntu-18.04]
         # but only stable on macos/windows (slower platforms)
         include:


briansmith:
I think it is fine to avoid gratuitously increasing the MSRV, but I'm not convinced that "6 months ago" (approximately 4 releases back) is the right policy.

> Given that Quinn has a MSRV of 6 months ago (1.50 was released on Feb 11, so it will be 6 months in a month from now), which was [requested](https://github.com/quinn-rs/quinn/issues/979) by the hyper team because they want to keep a 6 month old MSRV (and of course hyper is also relatively large downstream user of rustls).

Just to clarify, @seanmonstar requested "something like" Tokio's (not hyper) policy. I think we don't need to have exactly the same policy.

> hyper is also relatively large downstream user of rustls

https://github.com/hyperium/hyper and https://crates.io/crates/hyper/0.14.10/dependencies show that hyper isn't downstream of Rustls.

> My feeling is that raising the MSRV too aggressively will result in downstream users deferring rustls updates, which would be unfortunate IMO, or even using alternatives.

This is what people always say, but I've yet to meet anybody who has ever had this issue. There are a lot of bigger issues with Rustls and the things it depends on than the MSRV.

> However, ring on current main still works with 1.50.0. I think we should postpone raising the MSRV to 1.51 until it's actually used on ring.

No good deed goes unpunished. :) I deferred the work that relies on const generics in *ring* in part because I was helping this project. The reason I submitted that PR to Rustls is so that the Rustls team could provide feedback on my plans to increase the MSRV in *ring* so nobody is caught off guard.

In any case, the difference between 1.51.0 and 1.50.0 is 6 weeks. Maybe two weeks from now we might maybe have a Rustls release. Maybe a week or two later at best there would Reqwest release where the difference between 1.50.0 and 1.51.0 is an issue. So I think it makes sense to have the MSRV at least at 1.51.0 since, at worst, we can just wait 3 weeks, probably less, for the issue to resolve itself.

So, there are two issues:
1. Right now, I don't have any issue with dropping it to 1.51.0 but I don't see a reason to go below that.
2. In general, I'd rather not try to establish any official policy, especially not an official policy that matches what Tokio promises. We should give ourselves the flexibility to make choices when they come up. This has worked well for us for over half a decade (!) and I expect it will continue to work well, since you and everybody else would have a chance to participate in any future decisions to increase the MSRV in the future, due to the open nature of this project.

djc:
Thanks for the thoughtful reply!

> https://github.com/hyperium/hyper and https://crates.io/crates/hyper/0.14.10/dependencies show that hyper isn't downstream of Rustls.

Right, I got confused there. I guess this is more of an issue for reqwest, which currently tests for 1.46 in CI but doesn't seem to document this anywhere.

> No good deed goes unpunished. :) I deferred the work that relies on const generics in ring in part because I was helping this project. The reason I submitted that PR to Rustls is so that the Rustls team could provide feedback on my plans to increase the MSRV in ring so nobody is caught off guard.

Thanks for helping out with this project. So do you expect that you'll still want to land const generics support in the *ring* release that rustls 0.20 will depend on initially? If not, I still dislike the idea of restricting CI to 1.51 when we don't actually have any dependencies/code that depend on it.

I do appreciate the chicken/egg problem where you don't want to do work on const generics in *ring* unless you have some level of support of bumping the MSRV for important downstream dependencies, but I feel like, as long as we've established that rustls is fine with bumping the MSRV as soon as it's needed, it still doesn't quite make sense to bump it immediately.

> In any case, the difference between 1.51.0 and 1.50.0 is 6 weeks. Maybe two weeks from now we might maybe have a Rustls release. Maybe a week or two later at best there would Reqwest release where the difference between 1.50.0 and 1.51.0 is an issue.

I don't find this a particularly strong argument -- I could argue this the other way and say that waiting a few more weeks to land code that requires const generics in *ring* will also resolve the issue. Maybe it would be easier if you could more concretely talk about the benefits you expect from using const generics in *ring*?

> In general, I'd rather not try to establish any official policy, especially not an official policy that matches what Tokio promises. We should give ourselves the flexibility to make choices when they come up. This has worked well for us for over half a decade (!) and I expect it will continue to work well, since you and everybody else would have a chance to participate in any future decisions to increase the MSRV in the future, due to the open nature of this project.

I am not saying that we should exactly match Tokio's policy. I do think reasoning about a policy and communicating it to users are valuable, because downstream dependencies know better what to expect. Also, having a defined policy means not having to discuss each bump individually. To me, 4 releases doesn't seem too onerous, but I suppose 2 releases could work as well.

briansmith:
> Thanks for helping out with this project. So do you expect that you'll still want to land const generics support in the _ring_ release that rustls 0.20 will depend on initially? If not, I still dislike the idea of restricting CI to 1.51 when we don't actually have any dependencies/code that depend on it.

IDK. I do intend to release *ring* 0.17 soon and have webpki use it, and then update Rustls to also use the new version of webpkis and *ring*. Rustls will then ideally use it too, because otherwise it would end up depending on two different versions of *ring*, which will now be possible, but which would be wasteful. How far we get with the *ring* API improvements vs. how quickly we release a new version of Rustls is somewhat of a race.

djc:
Fair enough. I don't see a problem with postponing the 1.51 bump until such time as we land support for updated *ring* that needs it in rustls, given that everyone agrees updating it at that time will be acceptable.

briansmith:
Is there any motivation to do this any more?

djc:
No.

PR #779, opened 2021-07-08
Interpret end of incoming TLS stream as EOF

### LINKED ISSUE #777: potentially questionable overloading of ErrorKind::WouldBlock
PR #629 changed the behavior of the `std::io::Read` implementation for reading plaintext bytes from a TLS connection. `Ok(0)` (EOF) is only returned when a `CloseNotify` was received from the remote peer. When no bytes are available to read but the remote peer has *not* sent a `CloseNotify` message, an `Error` with [`ErrorKind::WouldBlock`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.WouldBlock) is returned instead.

What I'm wondering here is whether `WouldBlock` is, in fact, the correct `ErrorKind` to return in this case. Per the standard library documentation, `WouldBlock` indicates that
> The operation needs to block to complete, but the blocking operation was requested to not occur.

Generally, this is used when an IO operation is performed asynchronously: for example, on Linux, when a socket is opened with the `O_NONBLOCK` flag and a `read` or `write` call returned `EWOULDBLOCK`. Async IO libraries such as `tokio` and `mio` assign special meaning to this `ErrorKind`, as it is assumed to indicate that the operating system will somehow notify the program when that socket is once again readable (e.g. via an `epoll` event). 

Therefore, I wonder if it's really correct to overload the meaning of `WouldBlock` here. My understanding is that `rustls` will return it when the connection was closed abruptly by the peer (without sending a `CloseNotify`), *or* when the connection has not been closed, but additional TLS bytes must be read from IO before more plaintext bytes are available to read. Returning `WouldBlock` in both cases seems potentially confusing. 

I wonder if it would be more correct to return an error with [`ErrorKind::UnexpectedEof`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.UnexpectedEof) in the case where an EOF was encountered without receiving a `CloseNotify`?

djc:
The peer is not required to send us `CloseNotify`, but if the socket has
shutdown we should act accordingly.

Fixes #777.

Probably needs further comment updates and some tests.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -594,6 +594,7 @@ pub(crate) struct ConnectionCommon {
     pub(crate) record_layer: record_layer::RecordLayer,
     pub(crate) suite: Option<SupportedCipherSuite>,
     pub(crate) alpn_protocol: Option<Vec<u8>>,
+    /// If the peer has signaled end of stream.
     peer_eof: bool,
     pub(crate) traffic: bool,
     pub(crate) early_traffic: bool,
@@ -937,7 +938,7 @@ impl ConnectionCommon {
 
     /// Are we done? i.e., have we processed all received messages,
     /// and received a close_notify to indicate that no new messages
-    /// will arrive?
+    /// will arrive or ran out of bytes on the incoming TLS stream?
     fn connection_at_eof(&self) -> bool {
         self.peer_eof && !self.message_deframer.has_pending()
     }
@@ -946,7 +947,11 @@ impl ConnectionCommon {
     /// buffering, so `rd` can supply TLS messages in arbitrary-
     /// sized chunks (like a socket or pipe might).
     pub(crate) fn read_tls(&mut self, rd: &mut dyn io::Read) -> io::Result<usize> {
-        self.message_deframer.read(rd)
+        let res = self.message_deframer.read(rd);
+        if let Ok(0) = res {
+            self.peer_eof = true;
+        }
+        res
     }
 
     pub(crate) fn write_tls(&mut self, wr: &mut dyn io::Write) -> io::Result<usize> {
@@ -1067,9 +1072,8 @@ impl ConnectionCommon {
         let len = self.received_plaintext.read(buf)?;
 
         if len == 0 && !buf.is_empty() {
-            // no bytes available:
-            // - if we received a close_notify, this is a genuine permanent EOF
-            // - otherwise say EWOULDBLOCK
+            // No bytes available: if we think we ought to receive more data in the future,
+            // signal `WouldBlock` so that the caller knows it should supply us with more data.
             if !self.connection_at_eof() {
                 return Err(io::ErrorKind::WouldBlock.into());
             }


briansmith:
I think @hawkw is going to review this. Just want to add that we should add a test for this before merging.

rustls/src/conn.rs
```diff
@@ -937,7 +938,7 @@ impl ConnectionCommon {
 
     /// Are we done? i.e., have we processed all received messages,
     /// and received a close_notify to indicate that no new messages
-    /// will arrive?
+    /// will arrive or ran out of bytes on the incoming TLS stream?
```
hawkw, line 941: the comments https://github.com/ctz/rustls/issues/563#issuecomment-799827636, https://github.com/ctz/rustls/issues/563#issuecomment-811446138, https://github.com/ctz/rustls/issues/563#issuecomment-811562250 suggest that a TLS connection closing without a `CloseNotify` ought to be considered an error (although the user should be able to ignore that error if they choose to). it *looks* like this code will now always return a clean error (`Ok(0)`) regardless of whether a `CloseNotify` was received or the TCP connection just closed without one?

I think we should probably be distinguishing between `CloseNotify` and closing a connection without a `CloseNotify`, and return an `io::Error` in that case. 

IMO, [`ErrorKind::UnexpectedEof`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.UnexpectedEof) is probably the most correct error kind to use to indicate that a connection was closed without a `CloseNotify`, but I'm open to being convinced otherwise...i do think it's important to use an `ErrorKind` that rustls won't return in other cases, so user code can programmatically distinguish between the underlying connection EOF and other errors.

I wonder if it's also worth adding an option to consider TCP connection EOFs without `CloseNotify` a clean EOF, since some use cases may require talking to peers that don't support `CloseNotify`? Or is it sufficient to just allow the user to ignore that error kind themself?

hawkw (commented):
one major comment here --- what do you think?

djc:
@ctz I probably won't have much time to work on this over the weekend. If you want to take it over, please feel free. Otherwise I'll pick it back up on Monday.

rustls/src/conn.rs
ctz: I agree that this should be reported as an error rather than a 'clean' EOF. UnexpectedEof sounds like an excellent option, as does ConnectionAborted.

hawkw: IMO, I have a slight preference for `UnexpectedEof` over `ConnectionAborted`, as I think it's possible that the I/O reader passed to `read_tls` may return `ConnectionAborted`, but it probably won't return `UnexpectedEof`, so it's clearer where the error came from. But, either is probably fine here.

ctz:
I think I've addressed the above comments in 0b2b9bfdcf5eb81131ee7b6d91938b4ad507f24b

djc:
FWIW, your changes look good to me.

rustls/src/conn.rs
```diff
@@ -946,7 +947,11 @@ impl ConnectionCommon {
     /// buffering, so `rd` can supply TLS messages in arbitrary-
     /// sized chunks (like a socket or pipe might).
     pub(crate) fn read_tls(&mut self, rd: &mut dyn io::Read) -> io::Result<usize> {
-        self.message_deframer.read(rd)
+        let res = self.message_deframer.read(rd);
```
fantix, line 950: Probably the `read()` should not happen here if `self.peer_eof` is already `true`? The peer may not immediately terminate the TCP connection after sending a `CloseNotify`, and `read()` here will return `WantRead` error, leading to the caller waiting in `Pending` instead of getting a `ConnectionAborted` error with `CloseNotify alert received` as the message.

PR #780, opened 2021-07-12
Remove `webpki::TrustAnchor` from the public API.

briansmith:
The `webpki::TrustAnchor` type will change significantly in the near
future. Change Rustls's public APIs so that we can upgrade to the new
version of webpki without it being a breaking change. That is, allow
webpki and Rustls to evolve at different paces.

We expect that a future version of `webpki-roots` will also remove the
`webpki::TrustAnchor` from its public API, replacing it with its own
similar type.

The new `from_subject_spki_name_constraints` constructor will allow such
refactorings to be done without having to create a separate crate to
contain the types.

This makes the use of webpki-roots a bit messier than now, but most
applications probably shouldn't be using webpki-roots anyway.

### DIFF
--- a/rustls-mio/examples/tlsclient.rs
+++ b/rustls-mio/examples/tlsclient.rs
@@ -22,7 +22,7 @@ use docopt::Docopt;
 use rustls;
 use webpki_roots;
 
-use rustls::{Connection, RootCertStore};
+use rustls::{Connection, OwnedTrustAnchor, RootCertStore};
 
 const CLIENT: mio::Token = mio::Token(0);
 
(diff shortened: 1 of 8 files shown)


djc:
I think the alternative strategy you proposed earlier was that rustls could optionally depend on multiple versions of webpki to provide `From` implementations for each version of `TrustAnchor`, right? At this point I think just having webpki-roots expose DER directly for use with `RootCertStore::add_parsable_certificates()` would provide a more ergonomic interface.

briansmith:
> I think the alternative strategy you proposed earlier was that rustls could optionally depend on multiple versions of webpki to provide `From` implementations for each version of `TrustAnchor`, right?

There's really no reason to go through that pain. Keep in mind that as soon as webpki N+1 is released, version N becomes unsupported. So any design like that would freeze the use of an unsupported version of webpki into Rustls's public API, which is just bad.

> At this point I think just having webpki-roots expose DER directly for use with `RootCertStore::add_parsable_certificates()` would provide a more ergonomic interface.

Up to whoever maintains webpki-roots. Regardless, I think the API I'm suggesting here will be useful to applications that want to take advantage of one of the main design goals of the `TrustAnchor` type, which is to minimize wasted storage space for the trust anchor list, and to support adding name constraints that aren't present in the original certificate DER.

> It'd also be nice if you could add more thoughts on when/how/why `TrustAnchor` will change.

At a minimum, I will remove the public fields from `TrustAnchor` now that we can create `const fn` constructors. Also there over time, over several releases, I plan to add more per-root customization of validation behavior.

PR #781, opened 2021-07-13
Move most client/server-specific API into public client/server modules

djc:
The full list of items in the crate root was getting to be quite long, making it hard to find stuff even for experienced users (as in, myself). Splitting up client/server-specific items should make it easier to find what you need. In this PR we still reexport `ClientConfig`, `ClientConnection`, `ServerConfig` and `ServerConnection` in the crate root.

Additional benefit of the chosen approach is decreased/explicit visibility of items that were previously defined in the root of the private `client`/`server` modules.


ctz (approved):
I think this could benefit from a note in README.md

PR #787, opened 2021-07-23
Add Acceptor API for resolving server configuration based on ClientHello

### LINKED ISSUE #89: Asynchronous/tokio ResolvesServerCert?
I don't have a use case for this yet, but as more and more of the rust community moves towards async libraries like tokio, I wonder if there's a way `ResolvesServerCert` could support callback functions which do not return immediately/play nicely with the `tokio` `Core`.

Specific use case that might require this are servers which process many different domains and store certs/keys in a database, or servers which attempt to lazily obtain certificates from CAs like let's encrypt.

djc:
This builds on (the current version of) #721 (only the last three commits are new) and adds an API as proposed in #89. The current implementation requires a small change to the lifetimes exposed in the return type of `ClientHello::alpn()`; the final commit proceeds to remove the `sni` argument from `ClientCertVerifier` methods, which should no longer necessary.

Note that this is on my ISRG list, because it's something we'd like to use for the Apache httpd mod_tls project.

Fixes #89.

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -615,7 +615,8 @@ impl ResolvesServerCert for ServerCheckCertResolve {
         if let Some(expected_alpn) = &self.expected_alpn {
             let alpn = client_hello
                 .alpn()
-                .expect("alpn unexpectedly absent");
+                .expect("alpn unexpectedly absent")
+                .collect::<Vec<_>>();
             assert_eq!(alpn.len(), expected_alpn.len());
 
             for (got, wanted) in alpn.iter().zip(expected_alpn.iter()) {
@@ -4132,4 +4133,70 @@ fn test_client_tls12_no_resume_after_server_downgrade() {
     let mut server_2 = ServerConnection::new(Arc::new(server_config_2)).unwrap();
     common::do_handshake(&mut client_2, &mut server_2);
     assert_eq!(client_storage.puts(), 2);
+}
+
+#[test]
+fn test_acceptor() {
+    use rustls::server::Acceptor;
+
+    let client_config = Arc::new(make_client_config(KeyType::ED25519));
+    let mut client = ClientConnection::new(client_config, dns_name("localhost")).unwrap();
+    let mut buf = Vec::new();
+    client.write_tls(&mut buf).unwrap();
+
+    let server_config = Arc::new(make_server_config(KeyType::ED25519));
+    let mut acceptor = Acceptor::new().unwrap();
+    assert!(acceptor.wants_read());
+    acceptor
+        .read_tls(&mut buf.as_slice())
+        .unwrap();
+    let accepted = acceptor.accept().unwrap().unwrap();
+    let ch = accepted.client_hello();
+    assert_eq!(ch.server_name(), Some("localhost"));
+
+    let server = accepted
+        .into_connection(server_config)
+        .unwrap();
+    assert!(server.wants_write());
+
+    // Reusing an acceptor is not allowed
+    assert_eq!(
+        acceptor
+            .read_tls(&mut [0u8].as_ref())
+            .err()
+            .unwrap()
+            .kind(),
+        io::ErrorKind::Other,
+    );
+    assert_eq!(
+        acceptor.accept().err(),
+        Some(Error::General(
+            "cannot accept after successful acceptance".into()
+        ))
+    );
+
+    let mut acceptor = Acceptor::new().unwrap();
+    assert!(acceptor.accept().unwrap().is_none());
+    acceptor
+        .read_tls(&mut &buf[..3])
+        .unwrap(); // incomplete message
+    assert!(acceptor.accept().unwrap().is_none());
+    acceptor
+        .read_tls(&mut [0x80, 0x00].as_ref())
+        .unwrap(); // invalid message (len = 32k bytes)
+    assert!(acceptor.accept().is_err());
+
+    let mut acceptor = Acceptor::new().unwrap();
+    // Minimal valid 1-byte application data message is not a handshake message
+    acceptor
+        .read_tls(&mut [0x17, 0x03, 0x03, 0x00, 0x01, 0x00].as_ref())
+        .unwrap();
+    assert!(acceptor.accept().is_err());
+
+    let mut acceptor = Acceptor::new().unwrap();
+    // Minimal 1-byte ClientHello message is not a legal handshake message
+    acceptor
+        .read_tls(&mut [0x16, 0x03, 0x03, 0x00, 0x05, 0x01, 0x00, 0x00, 0x01, 0x00].as_ref())
+        .unwrap();
+    assert!(acceptor.accept().is_err());
 }
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -194,14 +194,11 @@ impl rustls::server::ClientCertVerifier for DummyClientAuth {
         true
     }
 
-    fn client_auth_mandatory(&self, _sni: Option<&rustls::server::DnsName>) -> Option<bool> {
+    fn client_auth_mandatory(&self) -> Option<bool> {
         Some(self.mandatory)
     }
 
-    fn client_auth_root_subjects(
-        &self,
-        _sni: Option<&rustls::server::DnsName>,
-    ) -> Option<rustls::DistinguishedNames> {
+    fn client_auth_root_subjects(&self) -> Option<rustls::DistinguishedNames> {
         Some(rustls::DistinguishedNames::new())
     }
 
@@ -209,7 +206,6 @@ impl rustls::server::ClientCertVerifier for DummyClientAuth {
         &self,
         _end_entity: &rustls::Certificate,
         _intermediates: &[rustls::Certificate],
-        _sni: Option<&rustls::server::DnsName>,
         _now: SystemTime,
     ) -> Result<rustls::server::ClientCertVerified, rustls::Error> {
         Ok(rustls::server::ClientCertVerified::assertion())
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -471,6 +471,43 @@ impl<Data> ConnectionCommon<Data> {
         }
     }
 
+    /// Extract the first handshake message.
+    ///
+    /// This is a shortcut to the `process_new_packets()` -> `process_msg()` ->
+    /// `process_handshake_messages()` path, specialized for the first handshake message.
+    pub(crate) fn first_handshake_message(&mut self) -> Result<Option<Message>, Error> {
+        if self.message_deframer.desynced {
+            return Err(Error::CorruptMessage);
+        }
+
+        let msg = match self.message_deframer.frames.pop_front() {
+            Some(msg) => msg,
+            None => return Ok(None),
+        };
+
+        let msg = msg.into_plain_message();
+        if !self.handshake_joiner.want_message(&msg) {
+            return Err(Error::CorruptMessagePayload(ContentType::Handshake));
+        }
+
+        if self
+            .handshake_joiner
+            .take_message(msg)
+            .is_none()
+        {
+            self.common_state
+                .send_fatal_alert(AlertDescription::DecodeError);
+            return Err(Error::CorruptMessagePayload(ContentType::Handshake));
+        }
+
+        self.common_state.aligned_handshake = self.handshake_joiner.is_empty();
+        Ok(self.handshake_joiner.frames.pop_front())
+    }
+
+    pub(crate) fn replace_state(&mut self, new: Box<dyn State<Data>>) {
+        self.state = Ok(new);
+    }
+
     fn process_msg(
         &mut self,
         msg: OpaqueMessage,
@@ -1175,6 +1212,11 @@ impl CommonState {
     fn send_warning_alert_no_log(&mut self, desc: AlertDescription) {
         let m = Message::build_alert(AlertLevel::Warning, desc);
         self.send_msg(m, self.record_layer.is_encrypting());
+    }
+
+    pub(crate) fn set_max_fragment_size(&mut self, new: Option<usize>) -> Result<(), Error> {
+        self.message_fragmenter
+            .set_max_fragment_size(new)
     }
 
     pub(crate) fn get_alpn_protocol(&self) -> Option<&[u8]> {
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -420,8 +420,10 @@ pub mod server {
     #[cfg_attr(docsrs, doc(cfg(feature = "quic")))]
     pub use server_conn::ServerQuicExt;
     pub use server_conn::StoresServerSessions;
+    pub use server_conn::{
+        Accepted, Acceptor, ServerConfig, ServerConnection, ServerConnectionData,
+    };
     pub use server_conn::{ClientHello, ProducesTickets, ResolvesServerCert};
-    pub use server_conn::{ServerConfig, ServerConnection, ServerConnectionData};
 
     #[cfg(feature = "dangerous_configuration")]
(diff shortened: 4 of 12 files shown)


jeromegn:
Playing with this branch and adapting `tokio-rustls` to be able to use it, I'm facing a few issues with reading from the `Reader`. I keep getting `WouldBlock`.

I'm assuming this is a problem with my logic in `tokio-rustls`'s `TlsStream`. Just in case, I'm writing here if this is a bug with how the `Reader` isn't returning the right thing? (since the tests pass, I feel like that's not the case). Or many somebody will see the obvious issue I'm running into 😄 

Here are my changes to tokio-rustls: https://github.com/tokio-rs/tls/compare/master...jeromegn:rustls-0.20

I've added a bunch of `println!` and here's the output I get running a single test from bad_ssl:

```
$ RUST_BACKTRACE=1 cargo test --package tokio-rustls --test badssl --all-features -- test_modern --exact --nocapture
warning: unused import: `Connection`
 --> tokio-rustls/src/client.rs:3:32
  |
3 | use rustls::{ClientConnection, Connection};
  |                                ^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `rustls::Connection`
 --> tokio-rustls/src/server.rs:3:5
  |
3 | use rustls::Connection;
  |     ^^^^^^^^^^^^^^^^^^

warning: 2 warnings emitted

   Compiling tokio-rustls v0.22.0 (/Users/jerome/src/github.com/jeromegn/tls/tokio-rustls)
    Finished test [unoptimized + debuginfo] target(s) in 0.77s
     Running tests/badssl.rs (target/debug/deps/badssl-88a8b09ec96a4b16)

running 1 test
write ready
read pending (would block)
read_tls err (would block)
read pending (would block)
read_tls err (would block)
read pending (would block)
read_tls err (would block)
read pending (would block)
read_tls err (would block)
read ready ok, size: 3319
read_tls size: 3319
write ready
read pending (would block)
read_tls err (would block)
read pending (would block)
read_tls err (would block)
read pending (would block)
read_tls err (would block)
read ready ok, size: 274
read_tls size: 274
read pending (would block)
read_tls err (would block)
write ready
still wants read
read pending (would block)
read_tls err (would block)
wants_read would block, breaking
prev: 32, buf remaining: 32
reading reader err: operation would block
Error: Kind(WouldBlock)
thread 'test_modern' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /Users/jerome/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/test/src/lib.rs:193:5
```

djc:
@jeromegn I think you'll want to look at #779 and incorporate those changes into your experiment.

jeromegn:
Thanks, I just gave at a try (https://github.com/ctz/rustls/commit/9482643affe2c48c298e6d971883bf58bac78538)

I'm still getting `WouldBlock` errors. To clarify: these errors happen on all tests, all the time. There's not a single test in tokio-rustls that passes, unfortunately. That's what led me to think it might be that I need to handle the reader / writer differently in tokio-rustls.

djc:
Sorry, you also need the follow-up change from @ctz mentioned in the comments.

rustls/src/server/mod.rs
```diff
@@ -143,8 +151,12 @@ impl<'a> ClientHello<'a> {
     /// Get the alpn.
     ///
     /// Returns `None` if the client did not include an ALPN extension
-    pub fn alpn(&self) -> Option<&'a [&'a [u8]]> {
-        self.alpn
+    pub fn alpn(&self) -> Option<impl Iterator<Item = &[u8]>> {
```
djc, line 154: This is a public API change; this avoids allocating a `Vec` for the protocols and makes it easier to abstract over the differences between the `Acceptor` path and the usual ClientHello handling path.

djc:
This has been rebased on newer main, which now includes the changes to fix read error handling. Thanks to a (private) review from @Ralith, I've changed the API to enable asynchronous initialization of the `ServerConfig`.

rustls/src/server/mod.rs
Ralith: Should this have `Item = &'a [u8]`?

rustls/tests/api.rs
```diff
@@ -3826,3 +3827,71 @@ fn test_server_rejects_clients_without_any_kx_group_overlap() {
...
+        .read_tls(&mut buf.as_slice())
+        .unwrap();
+    let accepted = acceptor.accept().unwrap().unwrap();
+    let ch = accepted.client_hello();
+    assert_eq!(ch.server_name(), Some("localhost"));
+
+    let key = server_config
+        .cert_resolver
+        .resolve(accepted.client_hello())
+        .unwrap();
```
Ralith, line 3851: Feels a bit odd that we're (indirectly) passing the `cert_resolver` in in this case at all since cert resolution happens entirely in user code on this path, but I respect that having multiple config types would be an unjustified maintenance burden.

Ralith (approved):
LGTM from a public-facing API standpoint. Releasing control flow to the user gives a lot more freedom than trait callbacks.

rustls/tests/api.rs
Matthias247: It also raises a question on what would happen if the client hello indicates that a different `server_config` should be used. Would it be save to use that? Probably yes, since the `cert_resolver` could definitely be standalone.

It might be less ambiguous if `cert_resolver` would not be part of `server_config` at all. Is it there for backward compatibility reasons?

rustls/src/server/mod.rs
```diff
@@ -324,113 +337,210 @@ impl ServerConnection {
...
 }
 
-impl PlaintextSink for ServerConnection {
-    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        Ok(self.send_some_plaintext(buf))
-    }
+/// Represents a `ClientHello` message received through the [`Acceptor`].
+///
+/// Contains the state required to resume the connection through [`Accepted::into_connection()`].
+pub struct Accepted<'a> {
```
Matthias247, line 487: That lifetime might make things a bit tricky for applications which really do an async lookup and have to pause the handshake for it. But that's just a feeling - I recommend testing it out in a more complete example example, e.g. one which uses `tokio::spawn_blocking` for reading certs.

rustls/tests/api.rs
djc: I suppose we could pass the `ResolvesServerCert` in separately for `ServerConnection::new()` -- I'm happy to make that change.

>  It also raises a question on what would happen if the client hello indicates that a different server_config should be used.

I'm not sure what you mean by that? There is only one call to the cert resolver; either it happens while handling the `ClientHello` message internally, or it happens on the "outside" for users of the `Acceptor` API.

rustls/src/server/mod.rs
```diff
@@ -324,113 +337,210 @@ impl ServerConnection {
...
     }
 
-    fn flush(&mut self) -> io::Result<()> {
-        Ok(())
+    /// Convert the [`Accepted`] into a [`ServerConnection`].
+    ///
+    /// Takes the state returned from [`Acceptor::accept()`] as well as the [`ServerConfig`] and
+    /// [`sign::CertifiedKey`] that should be used for the session. Returns an error if
+    /// configuration-dependent validation of the received `ClientHello` message fails.
+    pub fn into_connection(
```
Matthias247, line 509: While the type-state things are neat from the correctness point of view they also come with a cost:

Now the whole state has to be transferred into another location. If the state is bigger, it might be a more costly memcpy on the stack. Also for users, if they `Box`ed the `Connection`  before, they might now need a a separate `Box`  (or an enum) to do the transformation.

If the connection state is not that big it's likely not an issue. But I would recommend doing some benchmarking (e.g. by just doing handshakes in row) to determine the impact.

Also is the plan to go further and add a separate typestate to differentiate between the connection where the handshake is still in progress vs completed?

rustls/src/server/mod.rs
djc: I'm hoping @jeromegn will do that.

We could do without it, but that would involve reparsing the `ClientHello` message -- I thought it might be nice to avoid the overhead.

rustls/src/server/mod.rs
djc: Most of the stuff in `ConnectionCommon` is boxed up anyway (the state machine state, the buffers), so I don't think the type itself is particularly large. I don't know what you mean about boxing the `Connection` before -- there is no `Connection`, so you can't box it.

I don't currently have plans to differentiate between handshake in progress vs completed for the short term at least.

rustls/tests/api.rs
Matthias247: > I'm not sure what you mean by that?

I just meant it is not extremely obvious if people can use a `server_config.cert_resolver` first individually, and then pass a completely different `server_config` to `into_connection`. But given that this field isn't really connected to the remains of `server_config`, its also not super confusing.

rustls/src/server/mod.rs
Matthias247: Any kind of validation would be nice. If `Accepted` could not be hold due to whatever lifetime issues, the workaround with this kind of API is probably to drop `Accepted` and get it back later. But that might not work due to `accept()` taking state out of the connnection, and so dropping `Accepted` would not allow to resume the handshake?

rustls/src/server/mod.rs
Matthias247: > I don't know what you mean about boxing the Connection before 

I meant `Acceptor`. Good to know most things are boxed here

rustls/src/server/mod.rs
djc: Yeah, that wouldn't work. But as I said, we could just hold the `Message` in the `Accepted` while dropping the `client_hello` (instead cloning the ALPN protocols we need to build the `ClientHello`), then we'd reparse the `ClientHello` in `Accepted::into_connection()`.

rustls/tests/api.rs
djc: I guess we could just go the other way and only pass `ServerConfig` into `into_connection()`, letting it call `config.cert_resolver.resolve()` for you. That might be more orthogonal at the cost of potentially more indirection.

rustls/src/server/mod.rs
Matthias247: It could also hold the message and parsed clienthello as part of `Acceptor`, and either remove `Acceptor`  or make it just a tiny shim which retries that info? Then creating `Accepted` could be idempotent. As long as the ClientHello is there, you will get it again.

But that's just a solution before understanding whether its really a problem - integrating this plus async cert lookup in something like tokio-rustls should tell us.

djc: Ah yeah, I guess we wouldn't actually have to reparse the message, just extract the `ClientHelloPayload` again. Still, we need both the `Message` and the `ClientHelloPayload`, so that would just be more cloning. I think the borrowing solution is preferable as long as the lifetime doesn't get in the way too much.

rustls/tests/api.rs
Ralith: > I suppose we could pass the ResolvesServerCert in separately for ServerConnection::new() -- I'm happy to make that change.

That suits my tastes, at least.

> I guess we could just go the other way and only pass ServerConfig into into_connection(), letting it call config.cert_resolver.resolve() for you. That might be more orthogonal at the cost of potentially more indirection.

Then to e.g. do an async resolve, you'd have to do it separately and then pass in a trivial resolver, which is not the most discoverable pattern.

Matthias247: >>  I guess we could just go the other way and only pass ServerConfig into into_connection(), letting it call config.cert_resolver.resolve() for you. That might be more orthogonal at the cost of potentially more indirection.

> Then to e.g. do an async resolve, you'd have to do it separately and then pass in a trivial resolver, which is not the most discoverable pattern.

Yes, you couldn't even pass the same resolver you use outside, because it can introduce a race. Imagine the following: We have a list of SNIs which are mapped to server configurations. Something along

```rust
/// Maps from SNI to the configuration we use
struct Configs(Arc<RwLock<HashMap<String, EndpointConfig>>>)

struct EndpointConfig{
    rustls_config: <ServerConfig>,
    cert: Arc<CertifiedKey>,
    /* other stuff */
}
```

We can now definitely implement `ResolvesServerCert ` for `Configs` and use it both for looking up the to be used `ServerConfig`, as well as store the resolver based on that in the passed `ServerConfig`.

If we do that, there is however a chance that the rustls lookup up the cert again from that map, and gets a different config than the application previously monitored - since the map can change concurrently.

To avoid that, the resolver in the `ServerConfig` would need to be one that returns a single static cert.

That's doable, but a bit on the weird end. Moving `ResolvesServerCert` completely outside sounds more reasonable.

**What could however be potentially done is simply storing a `CertifiedKey` now as a plain field in `ServerConfig`**

Since applications can build or hold various ServerConfigs and pass them based on the observed `ClientHello` that seems totally reasonable. It also makes sense in a way that `ResolvesServerCert` is the only callback in `ServerConfig`. Everything else are plain values too (like `ciphersuites`).

rustls/src/server/mod.rs
djc: Fair enough, fixed that up.

rustls/tests/api.rs
djc: For now, I did go with the solution where the resolver inside the config is used. The reason is that I feel it's not reasonable to expose the extra complexity in what I believe will still be the common case of instantiating `ServerConnection`s directly. Having to instantiate a `ResolvesServerCert` separately for all users adds substantial complexity to the API.

> Then to e.g. do an async resolve, you'd have to do it separately and then pass in a trivial resolver, which is not the most discoverable pattern.

Actually, you can now only instantiate a `ServerConfig` through a builder API, and the easiest way to set up certificate resolution in the builder API is to pass a single cert, so this doesn't seem like a problem to me.

> If we do that, there is however a chance that the rustls lookup up the cert again from that map, and gets a different config than the application previously monitored - since the map can change concurrently.

There is no looking up again -- we only look up the certificate once either way. Plus, if the caller supplies a `ResolvesServerCert`, the responsibility to get that right is on them.

djc: Actually we can resolve the server certificate twice in the case of a `HelloRetryRequest`, but the same thing could happen today.

rustls/src/server/mod.rs
jeromegn: > I'm hoping @jeromegn will do that.

Thanks for including me here, but I'm barely following! I was hoping I could convert `tokio-rustls` to use this branch and get rid of our double-parsing of Client Hello for our async needs.

I can only spend time on this in short bursts. Somebody else is probably better suited than me to do the heavy-lifting if you want this to go forward faster 😄

rustls/tests/api.rs
Ralith: > For now, I did go with the solution where the resolver inside the config is used.

I'm not sure about this approach. It seems easy for a user to see that resolution still goes through a trait and assume that it therefore must be synchronous, without discovering the viable-but-elaborate workaround of implementing `ResolvesServerCert` for a newtype containing the e.g. asynchronously fetched results and ignoring the arguments to `resolve`.

rustls/src/server/mod.rs
djc: I did indeed end up getting right of the lifetime here, since it caused problems downstream.

ctz (approved):
I think this looks good. I read the comments about `ServerConfig` being required for `Accepted::into_connection` even though some parts of it won't be used. I think that's a cost we pay for having two ways of doing the same thing.

I'd consider in the future improving that in a later release with a further breaking change: perhaps removing synchronous cert resolution from `ServerConfig`, separating it as a separate requirement to construct a `ServerConnection`, or even requiring use of `Acceptor` for server connections. But we don't need to make that decision today.

PR #789, opened 2021-07-23
Change Connection to be an enum instead of a trait

### LINKED ISSUE #522: RFC: Unify client and server APIs

djc:
Fixes #522 by building on top of #721 (only the last commit is new).

### DIFF
--- a/rustls-mio/examples/tlsclient.rs
+++ b/rustls-mio/examples/tlsclient.rs
@@ -22,7 +22,7 @@ use docopt::Docopt;
 use rustls;
 use webpki_roots;
 
-use rustls::{Connection, OwnedTrustAnchor, RootCertStore};
+use rustls::{OwnedTrustAnchor, RootCertStore};
 
 const CLIENT: mio::Token = mio::Token(0);
 
--- a/rustls-mio/examples/tlsserver.rs
+++ b/rustls-mio/examples/tlsserver.rs
@@ -19,11 +19,10 @@ use docopt::Docopt;
 
 use env_logger;
 
-use rustls;
 use rustls::server::{
     AllowAnyAnonymousOrAuthenticatedClient, AllowAnyAuthenticatedClient, NoClientAuth,
 };
-use rustls::{Connection, RootCertStore};
+use rustls::{self, RootCertStore};
 use rustls_pemfile;
 
 // Token for our listening socket.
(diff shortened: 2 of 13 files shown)


ctz:
I think this needs a little attention where it touches the recent quic changes, but it looks quite reasonable to me.

djc:
Yup, I rebased a little hastily.

ctz (approved):

eloff:
Nice! I will definitely try this out soon, I think it will solve the problems I reported with using the Connection trait and having to abstract it behind an enum.

PR #793, opened 2021-08-04
remove extraneous detail from README

nilslice:
more clearly informs end-users that there are specific non-features which will not be supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,8 +147,7 @@ obsolete cryptography.
 
 ## Non-features
 
-The following things are broken, obsolete, badly designed, underspecified,
-dangerous and/or insane. Rustls does not support:
+Rustls does not support:
 
 * SSL1, SSL2, SSL3, TLS1 or TLS1.1.
 * RC4.


Darkspirit:
This would remove the warning and explanation why Rustls doesn't support them. With this patch, it would look like an arbitrary decision.

nilslice:
It kind of seems arbitrary to be there in the first place. The only thing the "explanation" does is call other things bad without explaining why. Not a great look, IMO.

Darkspirit:
Without the warning, inexperienced users might think Rustls is just feature-incomplete, might even ask again for implementation. To the contrary, they should be worried if they are still using these features somewhere. It's a quality aspect to be this conscientious and clear. This wonderful overview is hidden the docs: https://docs.rs/rustls/0.19.1/rustls/manual/_02_tls_vulnerabilities/index.html

nilslice:
Maybe more _tactful_ language can be used in the README. It comes off as a bit arrogant and overly broad. 

Something like...
> "To provide users with the most secure and high-quality implementation, Rustls intentionally does not support:"

(+ that manual page from the docs should be linked from the README, probably even in that sentence...)

The subsequent note after the list is informative enough so users know there are ways to accomplish those things outside of this crate:
> "There are plenty of other libraries that provide these features should you need them."

djc:
I agree with @Darkspirit this change as written is not an improvement, because it provides less information on (a) why these things aren't supported and (b) that these things are unlikely to be supported in the future. I do understand how you might feel these changes convey some kind of negative (arrogant/know-it-all/badmouthing) sense, and I personally would be accepting to changes that neutralize some of the negative "emotion" while retaining the information that's being provided here. I would probably also be a little bit sad because it feels like this gives the project more personality/flavor, but unfortunately those things don't always transfer well in text/across cultures.

Something like your last proposal seems okay to me, though it might be useful to be more explicit about point (b).

ctz:
> Maybe more _tactful_ language can be used in the README. It comes off as a bit arrogant and overly broad.

I agree, the current language comes off as heated or immature. Thanks for pointing that out. I think I've addressed that in 4bda1083aa5e778277f63c73612112d290aaaf95 based on the discussion here.

nilslice:
Thanks, @ctz - and apologies for the delayed response. I understand, it was written a long time ago, and the growth of the project just now puts this kind of perspective in front of a lot of people. I care about Rust & the community and am glad to see this project considering more of its audience through more thoughtful use of language. 

Thank you and all the contributors for all the work put into `rustls`!

PR #795, opened 2021-08-08
Fix error message when the client disabled the server's TLS version

blyxxyz:
This can for example come up if the client demands v1.3 and the server only has v1.2.


ctz (approved):
👍🏽

Issue #794, opened 2021-08-06
Improve the memory footprint of a large number of mostly idle connections


larsfillmore:
Our use case is a server that supports a large number of client connections that are usually idle or send/receive small messages (think of a chat server).

The C10k memory usage is about 300M, while switching to pure TCP reduces it to about 100M (BTW this corresponds well to https://jbp.io/2019/07/02/rustls-vs-openssl-memory-usage.html).

We're already using `Session::set_buffer_limit()`. But there are still internal buffers that are kept allocated even when there is no data to send/receive.

One possible solution might be to enable buffer reuse between connections. Something like `ServerConfig::buffer_provider:  Option<Arc<dyn BufferProvider>>`:
```rust
trait BufferProvider {
    fn acquire(&self) -> Buffer;
    fn release(&self, Buffer); 
}
```

The session would acquire a buffer once it sees activity on the underlying stream and release the buffer afterwards. The user may implement `BufferProvider` as a buffer pool.

That's rather a feature request than a design proposal as it doesn't touch related topics like reducing double-buffering (https://github.com/ctz/rustls/issues/716#issuecomment-851496486)

djc:
Short term, we could provide a method that deallocates buffers that are no longer needed after the handshake, only after the handshake is complete. I'd be interested in your assessment of the impact of something like #796. Warning: that's based on the main branch, which contains a substantial number of API changes. On the plus side, it's very possible that the main branch already improves memory usage for you.

For the longer term, I think the ideas discusses in #716 and #668 will likely yield an API that gives you enough control over memory allocation. If you'd be interested in helping fund that work, I'd be happy to chat about that.

briansmith:
> Short term, we could provide a method that deallocates buffers that are no longer needed after the handshake, only after the handshake is complete.

Why not just deallocate those buffers immediately when the handshake completes, automatically? It seems wrong to keep them around after the handshake.

larsfillmore:
> I'd be interested in your assessment of the impact of something like #796. Warning: that's based on the main branch, which contains a substantial number of API changes.

I didn't try applying #796 yet, sorry. It will be interesting to compare, but I need to first back-port it to 0.19.1 and implement an auto-minimize to make it useable with my Cargo.toml/[replace].

[...]

PR #796, opened 2021-08-09
Add ServerConnection::minimize() method

djc:


rustls/src/server/server_conn.rs
```diff
@@ -321,6 +321,26 @@ impl ServerConnection {
         );
         self.inner.data.reject_early_data = true;
     }
+
+    /// Minimize state kept around in memory.
+    ///
+    /// If called when [`CommonState::is_handshaking()`] is `false`, this will minimize the amount
+    /// of state kept for this connection (and return `true`). If the connection is still
+    /// handshaking, returns `true` and does nothing else.
```
thalesfragoso, line 329:
```suggestion
    /// handshaking, returns `false` and does nothing else.
```

djc:
As noted in #794, this change causes test failures and also doesn't seem to have much of an actual effect. Closing for now.

Issue #797, opened 2021-08-09
Session resumption in TLS 1.2 violates RFC 


tanyav2:
In the section of code for session resumption in TLS 1.2, the case where `session_id` has length zero for both `resuming` and `self` is not considered: https://github.com/rustls/rustls/blob/5e9ccd4b5b1154ac4b55316dfad178ba09957069/rustls/src/client/tls12.rs#L114-L115

According to the [TLS 1.2 RFC](https://datatracker.ietf.org/doc/html/rfc5246)

> The server may return an empty session_id to indicate that the session will not be cached and therefore cannot be resumed.

This can cause an error where a peer switches from using TLS 1.3 to TLS 1.2 (which causes a change in the negotiated ciphersuite) which results in a `PeerMisbehavedError` being thrown: https://github.com/rustls/rustls/blob/5e9ccd4b5b1154ac4b55316dfad178ba09957069/rustls/src/client/tls12.rs#L117-L122

I have reproduced the error here: https://github.com/tanyav2/replicate-rustls-bug

This doesn't occur when a TLS 1.2 server is restarted again using TLS 1.2 because when the `rustls-client` sends a certain `session_id` to the server, and the server responds with an empty `session_id`, which are different, and so the session is not resumed.

However, if the client first connects to a server running TLS 1.3, and then the server is restarted with TLS 1.2, the client sends an empty `session_id` because it is expecting to connect to a TLS 1.3 server. But because the server is now TLS 1.2, the empty `session_id` that the server sends in response matches the client's `session_id`, and the client ends up believing that the server is agreeing to resume the session. 

The fix can be something like https://github.com/tanyav2/rustls/blob/d70ee9240209d4acafdaf7cac1f911f9c681e5b5/rustls/src/client/tls12.rs#L113-L114:
```rust
    if (SessionID::empty() != resuming.session_id)
        && (resuming.session_id == self.session_id) 
    {
        debug!("Server agreed to resume");
```

The other thing worth noting is when TLS 1.2 server is restarted, it has an empty `session_id` but it sets the `SessionTicketAck` extension in response to the `SessionTicketOffer` sent by the client during `ClientHello`. However, when the server is downgraded from 1.3 -> 1.2, the client doesn't send a `SessionTicketOffer` or `SessionTicketRequest`, so the server doesn't respond with a `SessionTicketAck`.

briansmith:
[...]

PR #798, opened 2021-08-11
Minor refactors related to session resumption

djc:
Here are a number of smaller refactoring changes related to session resumption, made while investigating #797. Other than that, each commit stands on its own -- please refer to the specific commit messages for more information.


rustls/src/client/hs.rs
```diff
@@ -600,7 +600,6 @@ impl State<ClientConnectionData> for ExpectServerHello {
                 randoms,
                 using_ems: self.using_ems,
                 transcript,
-                session_id: server_hello.session_id,
```
briansmith, line 607: @djc for historical reference, please write a more complete commit message. I suggest something like:

> it isn't obvious whether `tls12::CompleteServerHelloHandling::session_id` is the ID from the ClientHello or the ServerHello message (it is the latter). Eliminate the possibility for confusion by eliminating the field, since it is redundant with the `handle_server_hello`'s `server_hello` parameter.

rustls/src/msgs/handshake.rs
```diff
@@ -104,11 +104,9 @@ pub struct SessionID {
 
 impl fmt::Debug for SessionID {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let mut t = f.debug_tuple("SessionID");
-        for i in 0..self.len() {
-            t.field(&self.data[i]);
-        }
-        t.finish()
+        f.debug_tuple("SessionID")
+            .field(&&self.data[..self.len])
```
briansmith, line 108: I am surprised at the `&&` instead of `&`. Not sure that's intentional.

briansmith:
@djc Could you please make the GitHub PR description more informative, for people who have to dig through these PRs later?

rustls/src/msgs/handshake.rs
djc: It is. `field()` takes a `&dyn Debug`, the slice is unsized so it cannot be passed as an argument directly. (Compilation fails without the extra &.)

djc: The exact reason here is that only `Sized` types can be cast to `dyn`. Added a comment to that effect.

rustls/src/client/hs.rs
djc: Added that. Sorry -- as mentioned in the other PR, I ran out of time yesterday. Also I didn't quite get to the point where I could articulate the benefit you're describing even though it was clearly what motivated me to make this change.

ctz (approved):

PR #799, opened 2021-08-11
Fixes for TLS 1.2 session resumption

### LINKED ISSUE #797: Session resumption in TLS 1.2 violates RFC

djc:
To make testing this easier (and because it seems useful to signal to the client), we add a method to `StoresServerSessions` that allows the implementation to signal whether it's accepting new sessions.

Fixes #797.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -111,7 +111,7 @@ mod server_hello {
 
             // See if we're successfully resuming.
             if let Some(ref resuming) = self.resuming_session {
-                if resuming.session_id == self.session_id {
+                if resuming.session_id() == Some(&self.session_id) {
                     debug!("Server agreed to resume");
 
                     // Is the server telling lies about the ciphersuite?
@@ -1008,9 +1008,8 @@ impl ExpectFinished {
 
         let master_secret = self.secrets.get_master_secret();
         let mut value = persist::ClientSessionValueWithResolvedCipherSuite::new(
-            ProtocolVersion::TLSv1_2,
+            persist::ResumeVersionWithSessionId::Tls12(self.session_id),
             self.secrets.suite().into(),
-            &self.session_id,
             ticket,
             master_secret,
             cx.common
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -29,9 +29,9 @@ use crate::SupportedCipherSuite;
 
 #[cfg(feature = "tls12")]
 use super::tls12;
+use crate::client::client_conn::ClientConnectionData;
 use crate::client::common::ClientHelloDetails;
 use crate::client::{tls13, ClientConfig, ServerName};
-use crate::client::client_conn::ClientConnectionData;
 
 use std::sync::Arc;
 
@@ -109,14 +109,15 @@ pub(super) fn start_handshake(
     };
 
     if let Some(resuming) = &mut resuming_session {
-        if resuming.version == ProtocolVersion::TLSv1_2 {
-            // If we have a ticket, we use the sessionid as a signal that
-            // we're  doing an abbreviated handshake.  See section 3.4 in
-            // RFC5077.
-            if !resuming.ticket.0.is_empty() {
-                resuming.set_session_id(SessionID::random()?);
+        let has_ticket = !resuming.ticket.0.is_empty();
+        if let Some(sid) = resuming.session_id_mut() {
+            // We only get a `SessionID` from the resuming session if it was on TLS 1.2. If we
+            // have a ticket, we use it as a signal that we're  doing an abbreviated handshake.
+            // See section 3.4 in RFC 5077.
+            if has_ticket {
+                *sid = SessionID::random()?;
             }
-            session_id = Some(resuming.session_id);
+            session_id = Some(*sid);
         }
 
         debug!("Resuming session");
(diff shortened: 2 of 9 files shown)


briansmith:
> To make testing this easier (and because it seems useful to signal to the client), we add a method to `StoresServerSessions` that allows the implementation whether it's accepting new sessions.

Seems some words are missing in "allows the implementation whether."

rustls/src/client/tls12.rs
```diff
@@ -111,7 +111,7 @@ mod server_hello {
 
             // See if we're successfully resuming.
             if let Some(ref resuming) = self.resuming_session {
-                if resuming.session_id == self.session_id {
+                if !resuming.session_id.is_empty() && resuming.session_id == self.session_id {
```
briansmith, line 114: I also think this all is hard to understand because there are implicit assumptions here that aren't documented and not documented in the type system.

Here we have `self.resuming_session` as a TLS 1.3 session where we don't store the session ID, so it is empty, but in this TLS 1.2 code we hadn't accounted for that possibility. Ideally, a TLS 1.3 session shouldn't even have a `session_id` that is comparable to a TLS 1.2 session_id. That is, this line above should have been a compile-time error. So I think a better solution would be to change the type of `resuming.session_id` to `Option<SessionID>` or even clearer define a separate type:
```
enum ResumptionSessionID {
    /// We don't store a session ID for TLS 1.3 sessions.
    Tls13,

    /// A non-empty session ID is always stored for TLS 1.2 sessions.
    ///
    /// The value of the session ID is always non-empty.
    Tls12(SessionID)
}
```

djc: Yes, I spent quite a bit of time yesterday trying to understand what was going on here and there are a number of things that could be improved. Unfortunately yesterday was my last day of work before my two-week vacation so I'm not sure how much time I'll be able to spend on this in the next week or so.

The improvement you suggested sounds good, but I think maybe we should go further? Some things I've come up with so far:

* We could start encoding the protocol version into the session cache key, making it an opaque `&[u8]`.
* We could change the type representation of the `SessionTicket` extension to use a nested enum rather than separate variants, where the nested enum's variants could be more explicit about the representative version.
* We should clarify what the intention is in preparing a `ClientHello` in terms of sending a 1.3 `PreSharedKeyOffer`, a 1.2 `SessionTicket` or both.
* In the client's 1.2 `ExpectFinished` state, it looks dubious to me that we can save the session if either the session ID or the ticket is empty (we only return early if both are).
* It looks like the client's 1.2 code might save an 1.2 session based on a session ID and a (presumably empty?) ticket from what was actually a cached 1.3 session.
* We might move any version-dependent fields from `ClientSessionValue`/`ServerSessionValue` into a version-variant enum, to make it more obvious what's going on there.

Yesterday I was still looking into the second problem @tanyav2 mentioned in #797, about which extensions the client sends in a 1.2-after-cached-1.3 scenario, so I just ran out of time. I dug through the RFCs to better understand the interactions but all I could find is that 1.3 obsoletes 5077, which doesn't quite make it clear what we should do here.

Nonetheless, I hope this is still a usable fix and a test case for the most egregious problem.

djc:
Apart from fixing the original issue, the third commit here fixes another issue where we don't filter the `resuming_session` based on whether it matches the presently negotiated version, which could lead to strange effects (added test to make sure we don't add a TLS 1.3 cache entry when resuming to 1.2). Will create another PR for deeper refactoring changes.

PR #802, opened 2021-08-18
Improve doctests in the documentation homepage

rukai:
Improve the doctests in the lib.rs file by:
*  fixing one case where syntax highlighting was missing
*  using no_run instead of ignore to ensure that the examples at least build.


ctz (approved):

ctz:
Thanks!

PR #811, opened 2021-09-07
Use stable rustfmt for formatting

### LINKED ISSUE #645: "Format" job in CI is too slow
According to https://github.com/ctz/rustls/actions/runs/753327095 this job takes 7 minutes. I'm guessing that's because building it from source is slow. We should consider using the GitHub Actions cache feature to cache the resultant binary.

It is very annoying to wait 7+ minutes and only then find out that I forgot to run admin/fmt, and then have to wait another 7+ minutes.

djc:
Now that https://github.com/rust-lang/rustfmt/pull/4782 is available in stable rustfmt, we can switch to it.

Fixes #645.


ctz (approved):
this is a big improvement!

Issue #786, opened 2021-07-20
How practical would it be to make structures related to traffic encryption/decryption and secrets handling public?


p1-mmr:
Hello,

In RusTLS 0.19.1, the `rutls::session::SessionSecrets` struct was module-public, and the corresponding module crate-public: https://github.com/ctz/rustls/blob/v/0.19.1/rustls/src/session.rs#L281

In the RustTLS 0.20.0 beta, the `rustls::conn::ConnectionSecrets` struct is crate-public: https://github.com/ctz/rustls/blob/v/0.20.0-beta1/rustls/src/conn.rs#L437

Similarly, in RustTLS 0.19.1, the `rustls::cipher::MessageDecrypter` trait was module-public, and the corresponding module crate-public: https://github.com/ctz/rustls/blob/v/0.19.1/rustls/src/cipher.rs#L14

In the RustTLS 0.20.0, it is now as well as associated stuff crate-public: https://github.com/ctz/rustls/blob/v/0.20.0-beta1/rustls/src/cipher.rs#L15

My use case of RustTLS is to attempt to statically decrypt TLS traffic from a given NSS-format `SSLKEYLOGFILE` file, containing connections secrets.

It is not practical to me to rely on the functions of the crate instancing this code because:

1. Obviously, I can't send back a Server Hello to a Client Hello as there is no interactivity in my context
2. And I can't neither perform any kind of Diffie-Hellman or curve-based exchange: I need to use the secrets that I was directly provided from an external source

However, RusTLS internally provides relevant packet decryption and TLS frame parsing routines that I am willing to use.

Hence, I would like to know how possible and conceivable it would be to make the parts of the corresponding module exposed to external users by changing attributes, without having to copy/fork the corresponding parts of the code?

Regards,

p1-mmr:
I think that [exposing under `::internal`](https://github.com/ctz/rustls/blob/v/0.20.0-beta2/rustls/src/lib.rs#L287) like you do for the `::msgs` module could be great.

(If useful to precise, the subject of this use case would be to provide PCAP-formatted and reprocessed excerpts of decrypted TLS traffics generated by our own application for debugging purposes.)

djc:
I suppose exposing it under `internal` would be okay. @ctz what is the purpose for `internal` that you had in mind? Is it mainly in order to enable testing from our side, or do you explicitly want to support the use cases from this issue and #784 as long as it doesn't cause extra maintenance burden in terms of semantic versioning?

[...]

PR #815, opened 2021-09-11
Pull request exposing strictly what is needed to provide the ability to statically decrypt TLS 1.3 messages from external users of the library

p1-mmr:
Hello,

This pull request is created following this issue: https://github.com/rustls/rustls/issues/786 

And has been submitted after it has been requested to do it in this comment: https://github.com/rustls/rustls/issues/786#issuecomment-914988853

It intends to make public what is needed to give to external users of the library the ability to statically decrypt TLS 1.3 messages. It does not include encrypting messages or supporting TLS 1.2 because it was not the original request, but can be added if needed.

It makes the `rustls::cipher::internal::MessageDecrypter` trait as well as the `rustls::tls13::Tls13CipherSuite::derive_decrypter` structure method public.

Regards,

### DIFF
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -340,7 +340,7 @@ pub mod internal {
     }
     /// Low-level TLS message decryption functions.
     pub mod cipher {
-        pub use crate::cipher::*;
+        pub use crate::cipher::MessageDecrypter;
     }
 }
 
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -338,6 +338,10 @@ pub mod internal {
     pub mod msgs {
         pub use crate::msgs::*;
     }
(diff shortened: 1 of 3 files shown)


rustls/src/lib.rs
```diff
@@ -338,6 +338,10 @@ pub mod internal {
     pub mod msgs {
         pub use crate::msgs::*;
     }
+    /// Low-level TLS message decryption functions.
+    pub mod cipher {
+        pub use crate::cipher::*;
```
djc, line 343: Please explicitly list the exported items here, rather than using a wildcart export.

p1-mmr: Done in commit 7e94bf6

djc (approved):

djc:

PR #820, opened 2021-09-17
Bump Golang to 1.17 in GHA

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

PR #823, opened 2021-09-18
Correct bogo checkout

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/rustls/rustls/pull/820#issuecomment-922354785


gliptak:
@ctz thank you for the pointer. the build is green

djc:
We should remove the command again that rebuilds `go.sum`.

gliptak:
@djc updated

ctz:
Thanks, I merged the original commit here along with a straight revert of 9407ada7

PR #826, opened 2021-09-20
Keep Certificate Transparency support implicitly off by default

djc:


rustls/src/client/builder.rs
```diff
@@ -82,6 +75,37 @@ impl ConfigBuilder<ClientConfig, WantsTransparencyPolicy> {
         )))
     }
 
+    /// Sets a single certificate chain and matching private key for use
```
djc, line 78: The duplication of these methods is a little ugly, but I think this does provide the most ergonomic API for callers.

ctz (approved):

PR #829, opened 2021-09-22
docs: Be explicit about how to not pass a list of DNs for client auth

### LINKED ISSUE #828: Add support for requesting a client certificate without giving the list of DNs
As far as I can tell, that is supported by TLS in general. 

It is obviously suboptimal in some situations (such as when used with interactive clients which have to choose from a set of certificates). However, in server-to-server situations where the standard server side certificates are used, having to send the DNs of the entire root store is really excessive (in addition, it causes interops with some Java SSL implementation which restricts handshake packets to 32 kiB – which is easily exceeded when sending the list of DNs from the entire root store).

So I would wish for [ClientCertVerifier::client_auth_root_subjects](https://docs.rs/rustls/0.20.0-beta2/rustls/trait.ClientCertVerifier.html#tymethod.client_auth_root_subjects) to have a way to say "No, don't send any root DNs, but also keep the connection going."

horazont:
It took me way longer than I care to admit to figure this out, so I
thought I'd spare others the trouble.

Fixes #828.

### DIFF
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -201,7 +201,8 @@ pub trait ClientCertVerifier: Send + Sync {
     /// Returns the subject names of the client authentication trust anchors to
     /// share with the client when requesting client authentication.
     ///
-    /// Return `None` to abort the connection.
+    /// Return `None` to abort the connection. Return an empty `Vec` to continue
+    /// the handshake without passing a list of CA DNs.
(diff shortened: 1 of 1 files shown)


djc (approved):
LGTM! If you're up for it, add backticks around `Vec`?

horazont:
Done :)

Issue #822, opened 2021-09-18
Get the negotiated CipherSuite as a string


kevinburke:
If I compile `psql` (the Postgres command line client) with OpenSSL, the default, and start a repl using TLS, I get the following output:

```
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

postgres=#
```

I'm trying to add a rustls backend, but as far as I can tell I can't retrieve the negotiated cipher - `TLS_AES_256_GCM_SHA384 ` - as a string. I tried for example:

```rust
use rustls::internal::msgs::enums::CipherSuite;

fn main() {
    let a = CipherSuite::TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384;
    println!("{}", a.to_string());
}
```

But I get:

```
5   |       println!("{}", a.to_string());
    |                        ^^^^^^^^^ method cannot be called on `CipherSuite` due to unsatisfied trait bounds
    |
   ::: /Users/kevin/.cargo/registry/src/github.com-1ecc6299db9ec823/rustls-0.19.1/src/msgs/enums.rs:316:1
    |
316 | / enum_builder! {
317 | |     /// The `CipherSuite` TLS protocol enum.  Values in this enum are taken
318 | |     /// from the various RFCs covering TLS, and are listed by IANA.
319 | |     /// The `Unknown` item is used when processing unrecognised ordinals.
...   |
695 | |     }
696 | | }
    | | -
    | | |
    | |_doesn't satisfy `CipherSuite: ToString`
    |   doesn't satisfy `CipherSuite: std::fmt::Display`
    |
    = note: the following trait bounds were not satisfied:
            `CipherSuite: std::fmt::Display`
            which is required by `CipherSuite: ToString`
```

It would be nice if it was possible to do this, without needing to manually match on the cipher suite name and then print the string. I'm not super great at rust programming, but it seems like here is a possible approach: https://stackoverflow.com/a/50334049/329700, I could try to implement that.

djc:
`CipherSuite` does implement `Debug`, so `format!("{:?}", a)` should work.

djc:
> My understanding is that you're not supposed to use debug formatting in production code?

That's the first I'm hearing about that, I don't think it's a thing in Rust.

https://doc.rust-lang.org/std/fmt/index.html#fmtdisplay-vs-fmtdebug

horazont:
[...]

PR #831, opened 2021-09-27
Expose enum variant names as str

djc:
As discussed in https://github.com/rustls/rustls-ffi/pull/147#pullrequestreview-763262982 and #822.

### DIFF
--- a/rustls/src/msgs/macros.rs
+++ b/rustls/src/msgs/macros.rs
@@ -59,6 +59,13 @@ macro_rules! enum_builder {
                     ,$enum_name::Unknown(x) => x
                 }
             }
+
+            pub fn as_str(&self) -> Option<&'static str> {
+                match self {
+                    $( $enum_name::$enum_var => Some(stringify!($enum_val))),*
+                    ,$enum_name::Unknown(_) => None,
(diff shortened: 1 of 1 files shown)


rustls/src/msgs/macros.rs
```diff
@@ -59,6 +59,13 @@ macro_rules! enum_builder {
                     ,$enum_name::Unknown(x) => x
                 }
             }
+
+            pub fn as_str(&self) -> Option<&'static str> {
```
djc, line 63: By adding this as another method generated by the `enum_builder!()` implementation for `U16` only we get this for 6 of the defined enums (including `CipherSuite`). This seems preferable to writing the code separately as an inherent impl for `CipherSuite` (since it would require repeating all the variants -> lots of boilerplate). Could additionally do this for `U8` enums, but it feels like we would just be generating more code that is probably not needed.

Chose to return `Option<u8>` so we can return `None` for the `Unknown` case, which seems most useful. Otherwise we could return `unknown` or change the return type to be `Cow<'static, str>` and format into `Unknown(13)` instead.

ctz (approved):

PR #833, opened 2021-09-27
Implement std::error::Error for InvalidDnsNameError

niklasf:
Noticed this when porting something like

```rust
fn main() -> Result<(), Box<dyn Error>> {
    let server_name = webpki::DNSNameRef::try_from_ascii_str(domain)?; // rustls 0.19
    // ...
}
```

to rustls 0.20.0:

```rust
fn main() -> Result<(), Box<dyn Error>> {
    let server_name = rustls::ServerName::try_from(domain)?;
    // ...
}
```

Additionally, the type is currently not exported from the crate, so it's not possible to explicitly refer to it. I did not address that, because I wasn't sure if it should be reexported from the root of the crate or some module.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -21,6 +21,7 @@ use super::hs;
 use crate::quic;
 
 use std::convert::TryFrom;
+use std::error::Error as StdError;
 use std::marker::PhantomData;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

zeroflaw:
Getting back to the the webpki::InvalidDnsNameError is incredible painful in rustls 0.20.
Can we not just keep the webpki error instead of wrapping it in a non-public! rustls error?
or make the rustls error public

I had to do this to keep the port change as minimal as possible (errors are handled upstream)
```
let server_name = match ServerName::try_from(websocket_options.domain) {
        Ok(server_name) => Ok(server_name),
        Err(_) => Err(webpki::InvalidDnsNameError),
    }?;
```

djc:
We definitely don't want to expose the webpki dependency publicly. If you submit a PR to make `InvalidDnsNameError` public I'd be happy to merge that.

PR #834, opened 2021-09-28
Stricter builder checks on cipher suite/version compatibility

djc:
We did not yet check that each configured version had matching cipher suites configured.

Make `Config::supports_version()` explicitly public. We would like to use this in Quinn to check that TLS configurations have 1.3 enabled.

### DIFF
--- a/rustls-mio/tests/client_suites.rs
+++ b/rustls-mio/tests/client_suites.rs
@@ -14,6 +14,7 @@ fn ecdhe_rsa_aes_128_gcm_sha256() {
     server
         .client()
         .verbose()
+        .version("1.2")
         .suite("TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256")
         .expect("Ciphers common between both SSL end points:\nECDHE-RSA-AES128-GCM-SHA256")
         .go();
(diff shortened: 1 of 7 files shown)


ctz:
I think it's an API usability problem if all users of rustls are required to configure _both_ cipher suites and protocol versions, to achieve the aim of configuring _either_. To illustrate, I'd expect these uses of `tlsclient` to continue to work:

```
$ ./target/debug/examples/tlsclient --http google.com --suite TLS13_AES_256_GCM_SHA384
thread 'main' panicked at 'inconsistent cipher-suite/versions selected: General("no cipher suite for version TLSv1_2 configured")', rustls-mio/examples/tlsclient.rs:516:10

$ ./target/debug/examples/tlsclient --http google.com --protover 1.2
thread 'main' panicked at 'inconsistent cipher-suite/versions selected: General("no version matching cipher suite TLS13_AES_256_GCM_SHA384")', rustls-mio/examples/tlsclient.rs:516:10
```
(nb. I'm not saying we should fix tlsclient, here. I think if we do that, we're saying that all consumers of rustls will have to do the same.)

djc:
Okay. Either way, let's not block 0.20.1 on this?

djc:
I guess this goes back to the discussion point in https://github.com/rustls/rustls/pull/724#issuecomment-852027936, in the sense that I still feel that configuring exclusively based on cipher suites would be superior. The case for configuring only a version would then be handled by providing constants with all cipher suites for a particular version, as in `with_cipher_suites(DEFAULT_TLS12_CIPHER_SUITES)`. Given that a potential TLS 1.4 is pretty far away at this point, that doesn't seem like a strong argument to me.

Absent, I would still argue that failing fast in the face of potentially incorrect configuration is more important than this particular kind of convenience. I expect that the number of users who tweak the cipher suites is relatively small anyway, though we could argue that this kind of change should be made in a semver-incompatible version.

djc:
I guess we'll just close this.

PR #836, opened 2021-09-28
Improve ConfigBuilder docs

djc:


g2p:
Thanks for the fixed links within the docs and for exposing WantsTransparencyPolicyOrClientCert!

+1 for me, because that state needs to be exposed so it can be returned.
Being able to name the type returned by with_root_certificates is necessary for adding methods like with_webpki_roots on a ConfigBuilderExt trait.

ctz (approved):

Issue #717, opened 2021-05-26
DoS prevention for zero-length or very small fragments


briansmith:
[I split this off from #698 so that #698 can be just about large messages.]

Zero-length fragments are not accepted in TLS 1.3, and zero-length plaintext handshake fragments are not accepted for TLS 1.2. Only TLS 1.2 encrypted application data records should be accepted if they have length 0. This should be enforced in the deframer, I think.

There are two issues with zero-length fragments: (1) We might be busy doing a lot of work, but not making any progress; this can be hard to detect by DoS countermeasures above and below Rustls in the network stack. (2) Some of our limits on memory usage are probably measuring memory usage without considering all the overheads; with many zero-length/tiny fragments, the overhead being ignored could end up being larger than what we measure to enforce the limits. For (2) one solution would be to always measure the full cost when computing a limit.

djc:
bogo does have a `TLS13-EmptyRecords` test, which we pass by failing with a `PeerMisbehavedError` as soon as we decrypt the message.

briansmith:
Issue #952 is an instance of this.

PR #848, opened 2021-11-12
Disallow zero length fragments

davidv1992:
This PR implements blocking of zero-length messages from being accepted on the TLS connection unless they are of type ApplicationData, thereby ensuring the implementation always makes progress on the other requests. ApplicationData of length 0 is allowed by the standard to ensure implementations can implement countermeasures to traffic flow analysis. This is a partial fix of #717, although short but non-zero length packages still might pose a DoS threat.


ctz (approved):

ctz:
Thanks!

PR #849, opened 2021-11-12
Ensure ticketswitcher mutex is not held when generator is called.

### LINKED ISSUE #656: Ticketer has inefficient Mutex usage patterns
When working on PR #654 I noticed that the Ticketer holds the mutex over potentially long-running operations. In particular, when rolling the key, a new random key is generated, and that process uses the system PRNG while holding the mutex. This is either a syscall or even File I/O (if falling back to /dev/urandom or equivalent). This is bad for two reasons:

1. The lock is held for potentially a long time.
2. It's hard for us to be confident that File I/O operations will never panic within libstd. Panicking would poison the lock.

In PR #654 I eliminated one lock/unlock pair per operation. Thus we could easily be able to refactor the code to avoid holding the lock during key generation. Even the most naïve way of doing this would likely have better performance than the pre-PR-654 code.

davidv1992:
This reduces possibility of panics during holding of the ticketswitcher mutex. Fixes #656

### DIFF
--- a/rustls/src/ticketer.rs
+++ b/rustls/src/ticketer.rs
@@ -148,17 +148,36 @@ impl TicketSwitcher {
     ///
     /// Calling this regularly will ensure timely key erasure.  Otherwise,
     /// key erasure will be delayed until the next encrypt/decrypt call.
-    fn maybe_roll(
-        &self,
-        now: TimeBase,
-        state: &mut MutexGuard<TicketSwitcherState>,
-    ) -> Result<(), rand::GetRandomFailed> {
+    ///
+    /// In order to reduce the ammount of lock/unlock operations in the
+    /// common case that no new ticketer is needed, we return the mutex-
+    /// guard taken here
+    fn maybe_roll(&self, now: TimeBase) -> Option<MutexGuard<TicketSwitcherState>> {
         let now = now.as_secs();
+
+        // First lock section, check whether we need a new ticketer
+        // We make this a block to ensure the take is dropped before
+        // generating the new ticketer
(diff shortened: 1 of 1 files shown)


rustls/src/ticketer.rs
```diff
@@ -148,17 +148,36 @@ impl TicketSwitcher {
...
     /// key erasure will be delayed until the next encrypt/decrypt call.
-    fn maybe_roll(
-        &self,
-        now: TimeBase,
-        state: &mut MutexGuard<TicketSwitcherState>,
-    ) -> Result<(), rand::GetRandomFailed> {
+    ///
+    /// In order to reduce the ammount of lock/unlock operations in the
+    /// common case that no new ticketer is needed, we return the mutex-
+    /// guard taken here
```
djc, line 154: It would be good to explain some of the stuff that Brian mentioned in his original ticket here, which seems like a fairly different/more specific motivation than what you mention here.

djc (commented):
Thanks for working on this! The code changes look good to me.

That said, the comments seem a little messy, with some typos and incorrect/unexpected words. Maybe spend a little time tightening that up and more closely aligning them with the concerns from the original ticket?

rustls/src/ticketer.rs
davidv1992: hmmm, after thinking and rereading brians original comment, i realized the solution here is not ideal either, I will spend some time soon to get to a good solution that covers all his goals

davidv1992:
Closed because #861 is a better option

PR #853, opened 2021-11-15
macros: point as_str at enum LHS, not RHS

kevinburke:
Previously we'd print out the integer constant for an enum, instead
of its name, which is more descriptive ("TLS13_AES_256_GCM_SHA384"
instead of "0x1302", for example). I believe that this was the
originally intended behavior in #831.

Add a test to expose the issue and verify the patch works as expected.


jsha:
Looks good to me.

djc (approved):
Thank you for fixing this up! Not sure why I didn't add a test in the first place (or why that didn't get flagged in review). 🤦

Issue #855, opened 2021-11-18
Passing multiple --protover flags to tlsclient


xvzcf:
The [tlsclient usage string](https://github.com/rustls/rustls/blob/main/rustls-mio/examples/tlsclient.rs#L313) says that the `--protover` flag `May be used multiple times.` 

How does one do so? I've tried passing in  `--protover 1.2 --protover 1.3`, `--protover 1.2 1.3` among other variations, but I just get an `Invalid arguments.` response.

tanyav2:
Created a PR that should fix the issue, it needs to be used as `--protover 1.2 --protover 1.3` because `docopt` doesn't allow multiple arguments for an option as it goes against standard POSIX cli: https://github.com/docopt/docopt/issues/484#issuecomment-680743971

PR #856, opened 2021-11-18
example client protover arg should allow multiple versions

tanyav2:
Fixes https://github.com/rustls/rustls/issues/855

### DIFF
--- a/rustls-mio/examples/tlsclient.rs
+++ b/rustls-mio/examples/tlsclient.rs
@@ -298,7 +298,7 @@ If --cafile is not supplied, a built-in set of CA certificates
 are used from the webpki-roots crate.
 
 Usage:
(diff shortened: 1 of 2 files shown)


thomwiggers (changes requested):
This looks like the right solution, evidenced by the fact that `--suite` and `--proto` already do the same.

This should also be patched on the server side in `rustls-mio/examples/tlsserver.rs` though.

djc:
Sounds good to me. @tanyav2 thanks for fixing this, are you up for fixing the server, too?

tanyav2:
@djc @thomwiggers Fixed

thomwiggers (approved):

ctz (approved):
Thanks!

Issue #839, opened 2021-10-22
`RootCertStore::add` returns `webpki::Error` as an error


paolobarbolini:
Looks like this was missed in #736

briansmith:
When is the "next major release?"

Because of this problem, it looks like I will need to release a new version of webpki that's semver-compatible with the latest webpki release, which depends on *ring* 0.17 instead of *ring* 0.16.x. This would cause Rustls to depend on both *ring* 0.16.x and 0.17 for most people. That's very undesirable, so then I'll submit a PR to Rustls that updates Rustls to *ring* 0.17 too. Then I hope we can have a new SemVer-compatible release of Rustls promptly after that.

Alternatively, we could just make the next release of Rustls be the one that fixes this bug, if we're willing for that to be a SemVer-breaking release, and if it will happen soon.

djc:
I think it would make sense to do a semver-incompatible release for this. We should probably also get the desired changes to `Error` type variants in before we release it, though.

cpu:
This looks like it was fixed with https://github.com/rustls/rustls/commit/a777077257b4898c996b5e8a641219330d8a7a7a

PR #858, opened 2021-11-19
removing webpki::Error as return type

Herbert-Karl:
hi,
following I propose a solution for issue #839. As I am new to Rust and the codebase of rustls, I don't have much confidence, that the change doesn't break API or introduces side effects.
I have run the existing tests and no errors were reported. I haven't fuzzed the changed codebase or did any manual testing.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -85,7 +85,7 @@ impl RootCertStore {
     /// Add a single DER-encoded certificate to the store.
     pub fn add(&mut self, der: &key::Certificate) -> Result<(), Error> {
         let ta = webpki::TrustAnchor::try_from_cert_der(&der.0)
-            .map_err(|e| Error::InvalidCertificateData(format!("{}", e)))?;
+            .map_err(|e| Error::FailedToDecodeRootCertData(format!("{}", e)))?;
         let ota = OwnedTrustAnchor::from_subject_spki_name_constraints(
             ta.subject,
             ta.spki,
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -84,7 +84,8 @@ impl RootCertStore {
 
     /// Add a single DER-encoded certificate to the store.
     pub fn add(&mut self, der: &key::Certificate) -> Result<(), Error> {
-        let ta = webpki::TrustAnchor::try_from_cert_der(&der.0).map_err(|e| Error::InvalidCertificateData(format!("{}", e)))?;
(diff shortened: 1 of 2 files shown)


ctz (dismissed):
Thanks! Going to hold this one for the next release, as it's a breaking API change.

rustls/src/anchors.rs
```diff
@@ -82,8 +83,9 @@ impl RootCertStore {
     }
 
     /// Add a single DER-encoded certificate to the store.
-    pub fn add(&mut self, der: &key::Certificate) -> Result<(), webpki::Error> {
-        let ta = webpki::TrustAnchor::try_from_cert_der(&der.0)?;
+    pub fn add(&mut self, der: &key::Certificate) -> Result<(), Error> {
+        let ta = webpki::TrustAnchor::try_from_cert_der(&der.0)
+            .map_err(|e| Error::InvalidCertificateData(format!("{}", e)))?;
```
briansmith, line 88: I suggest we use a different, new, error variant: `InvalidRootCertData`. `InvalidCertificateData` should only be returned during certificate validation.

briansmith (changes requested):
Thanks for working on this. I just have one suggestion.

rustls/src/anchors.rs
Herbert-Karl: i can understand that though
based on the description and error cases in the "try_from_cert_der" function, an error variant called "FailtedToDecodeDER" seems more appropriate, as the function doesn't actually validate the certificate, only decodes it

briansmith: > i can understand that though
> based on the description and error cases in the "try_from_cert_der" function, an error variant called "FailtedToDecodeDER" seems more appropriate, as the function doesn't actually validate the certificate, only decodes it

The problem I see with `FailedtoDecodeDER` is that it doesn't provide any context to the user about what was trying to be decoded. Also, I think `InvalidRootCertData` could be used in the future for any kind of problem with the root cert, not just problems with the ASN.1 encoding.

Herbert-Karl: You are right, `FailedToDecodeDER` by name lacks context, in which the error occurred. I don't like that `InvalidRootCertData` suggests something (a validation of the root certificate) which doesn't actually happen.
What do you think about naming the error `FailedToDecodeRootCertData`?

briansmith (approved):

ctz:
Thanks for the contribution here -- recently we've been trying to move away from errors that aren't machine readable because they squashed the cause into a string -- see eg. #1162. This has been fixed in the progress towards that -- see a777077.

PR #861, opened 2021-11-25
Reworked ticketswitcher to reduce time mutex is held.

davidv1992:
This PR aims to reduce the amount of time the ticketswitcher mutex is held, by excluding generation of new ticketers from the period that the mutex is held, solving #656. This also gives a secondary advantage that a panic during generation will not poison the mutex.

This implementation works better towards those goals then PR #849 because it avoids the potential of multiple threads generating a next ticketer in the common case where generation does not fail and is fast enough to complete within the switching period of the ticketer.

However, in order to ensure the next ticketer is available timely, it is generated ahead of time, just after the previous switch. This means that the ticketer keys are in memory for an extra period of 6 hours (though unused during that period), so we should evaluate whether that is acceptable from a security point of view.

### DIFF
--- a/rustls/src/ticketer.rs
+++ b/rustls/src/ticketer.rs
@@ -107,6 +107,7 @@ impl ProducesTickets for AeadTicketer {
 }
 
 struct TicketSwitcherState {
+    next: Option<Box<dyn ProducesTickets>>,
     current: Box<dyn ProducesTickets>,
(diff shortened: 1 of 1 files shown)


davidv1992:
Note: the pipeline failure seems unrelated to change, is due to rustup failing for some reason.

ctz (approved):

ctz:
This looks good to me. Thanks!

PR #862, opened 2021-11-26
CI: add CIFuzz integration

DavidKorczynski:
Add CIFuzz workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which Rustls already runs.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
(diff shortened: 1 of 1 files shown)


djc:
This makes sense to me. Many PRs already take about 10 minutes to run so this shouldn't slow us down too much more (assuming it mostly runs in parallel to our other stuff). @ctz what do you think?

DavidKorczynski:
We can always decrease the current 600 seconds of fuzzer runtime as a way of shorten the CI check

djc (approved):

PR #863, opened 2021-12-01
Fix deprecation notice

linkmauve:
The previous notice said to use `TlsError` instead of the deprecated `TLSError`, while the actual exported type is named `Error`.


djc (approved):
Makes sense, thanks!

PR #868, opened 2021-12-08
Respect the SSLKEYLOGFILE variable by default

jyn514:
Currently, to look at traffic in wireshark, you have to opt-in not just by setting the env variable,
but also opting-in in the application itself. This makes it possible to debug the traffic with just
the env variable, which is the same default behavior as curl and openssl.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -4,7 +4,7 @@ use crate::client::handy;
 use crate::client::{ClientConfig, ResolvesClientCert};
 use crate::error::Error;
 use crate::key;
-use crate::keylog::NoKeyLog;
+use crate::keylog::KeyLogFile;
 use crate::kx::SupportedKxGroup;
 use crate::suites::SupportedCipherSuite;
 use crate::verify::{self, CertificateTransparencyPolicy};
@@ -181,7 +181,7 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
             versions: self.state.versions,
             enable_sni: true,
             verifier: self.state.verifier,
-            key_log: Arc::new(NoKeyLog {}),
+            key_log: Arc::new(KeyLogFile::new()),
             enable_early_data: false,
         }
     }


djc:
FYI this was done very intentionally, to make it harder to accidentally leak secrets.

jyn514:
@djc what do you think about only disabling it when `with_safe_defaults()` is called? It was very surprising to me this didn't work at first, I had to dig around to figure out what was going on.

LPardue:
The confusing thing I found is that by default the indicated is created but not populated. If you don't want to support SSLKEYLOGFILE by default, at least not creating those files would be nice.

djc:
@LPardue that makes sense, see #871.

djc:
Original implementation in #171. It looks like the change to do this was made in https://github.com/rustls/rustls/commit/f52acb2348249b594c13b3df45831b26124759cc (as hinted at by the comment in #171, although that is less explicit about this issue than I remember).

jyn514:
> FYI this was done very intentionally, to make it harder to accidentally leak secrets.

You already have to opt-in at compile time with the `logging` feature, right? It seems weird to have 3 separate opt-ins and put two of them at compile time.

djc:
Personally I also find having to mess with the configuration code to enable this a bit annoying. Do curl and openssl support this at the library level or at the application/binary level?

jyn514:
> Do curl and openssl support this at the library level or at the application/binary level?

The curl binary certainly respects `SSLKEYLOGFILE` - looking at https://github.com/curl/curl/blob/5a3775fadf5d4f5ab1133d565c34e93a013791a2/lib/vtls/openssl.c#L1232 I think any application using curl + openssl as a library will respect it by default too.

ctz:
> You already have to opt-in at compile time with the `logging` feature, right? It seems weird to have 3 separate opt-ins and put two of them at compile time.

The logging feature controls general log output (via the `log` crate) and is unrelated to the `KeyLog` trait.

>  Do curl and openssl support this at the library level or at the application/binary level?

OpenSSL does not: instead the application must call `SSL_CTX_set_keylog_callback` to turn it on and get callbacks when the key material becomes available. Similar to the `KeyLog` trait we have here.

djc:
I think it makes a lot of sense for us to align with openssl on this and leave enabling this to applications and/or higher-level libraries.

jyn514:
That's a shame ... reqwest chose to keep the default the same as what rustls uses: https://github.com/seanmonstar/reqwest/issues/1016#issuecomment-708669293

cc @seanmonstar

briansmith:
> You already have to opt-in at compile time with the `logging` feature, right? It seems weird to have 3 separate opt-ins and put two of them at compile time.

Unfortunately, in a real application one might not have full control over what Rustls features third-party library crates enable/disable, especially since we don't have good auditing tools or ways to say "ensure no Rustls dependency has feature X enabled." Thus the enabling of a feature shouldn't change the (default) safety properties of Rustls (or really any crate).

briansmith:
> That's a shame ... reqwest chose to keep the default the same as what rustls uses: [seanmonstar/reqwest#1016 (comment)](https://github.com/seanmonstar/reqwest/issues/1016#issuecomment-708669293)

It shouldn't be the decision of any library to enable this functionality by default. The top-level application should choose to do so or not. (I wonder if they'd want to control the name of the environment variable as well, much like RUST_LOG?)

briansmith:
BTW, this would be a breaking API change for any application that's relying on the fact that the log isn't enabled by default, such as the one I maintain.

daladim:
From a security point of view, I like the fact that this is disabled by default, and that the application developer should opt-in to enable `SSLKEYLOGFILE` support.

However, I've just tried to opt-in recently in a project that uses reqwest (that's #875 that @djc has just linked to this PR), and that is not really easy to do, and poorly documented unfortunately.

Havig this feature disabled by default, but more easy to opt-in (e.g. with a mere Cargo feature that would enable this current MR) and more clearly documented would be very nice IMHO :-)

djc:
I don't think just a Cargo feature is right for this, as explained by Brian in https://github.com/rustls/rustls/pull/868#issuecomment-990097107. I'm going to close this now, as I think the consensus is clear that we should not enable writing key log files by default switching only on the environment variable. I'll also write up a PR with some documentation improvements.

briansmith:
> However, I've just tried to opt-in recently in a project that uses reqwest (that's #875 that @djc has just linked to this PR), and that is not really easy to do, and poorly documented unfortunately.

I feel like that's something that needs to be improved in reqwest. I know of another project that has issues with how to configure Rustls through reqwest. It seems like it can be done but it's a little error-prone.

LPardue:
I can live with the current default behaviour. Improving the ergonomics a bit won't hurt and #871 is one step to doing that.

bjorn3:
Would it be possible to use a cfg instead of a cargo feature for this. That way dependent crates can't opt in. Only the user who builds it using something like `RUSTFLAGS="--cfg rustls_respect_sslkeylogfile`.

PR #871, opened 2021-12-09
Defer opening key log file until writing to it

djc:
As suggested in #868.

### DIFF
--- a/rustls/src/keylog.rs
+++ b/rustls/src/keylog.rs
@@ -2,7 +2,7 @@ use std::env;
 use std::fs::{File, OpenOptions};
 use std::io;
 use std::io::Write;
-use std::path::Path;
+use std::path::PathBuf;
 use std::sync::Mutex;
 
 #[cfg(feature = "logging")]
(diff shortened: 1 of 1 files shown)


djc:
(clippy issues fixed in #867.)

LPardue:
thank you!

ctz (approved):

briansmith:
Technically, this is a technically a breaking change, for the following scenerios:
* The application implements sandboxing. They want to have the sandbox block opening files, but they want to support this feature, so they configure the key log before enabling the sandbox, which they do before they make any TLS connections. Before, this strategy worked, AFAICT. Now, I think it doesn't, because the file won't be opened until at least the first TLS connection, at which time the sandbox will block the opening.
* For similar reasons, if the current working directory changes between enabling the key logging and the first attempt to log, then a different file will be opened and written to.
* If the application tries to `unlink()` after configuring the logging but before the first thing is logged.

That's not to say the change shouldn't be made, but is good to note it.

briansmith:
This was reverted in PR #926.

PR #873, opened 2021-12-10
Strict ALPN checking for QUIC

Ralith:
Noticed some nonconformant corner cases in Quinn, plus an opportunity to move more TLS-specific logic into rustls, where it's much simpler to implement besides.

Behavior validated empirically under Quinn using https://github.com/quinn-rs/quinn/pull/1258.

### DIFF
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -120,6 +120,21 @@ impl ExtensionProcessing {
         #[cfg(feature = "quic")]
         {
             if cx.common.is_quic() {
+                // QUIC has strict ALPN, unlike TLS's more backwards-compatible behavior. RFC 9001
+                // says: "The server MUST treat the inability to select a compatible application
+                // protocol as a connection error of type 0x0178". We judge that ALPN was desired
+                // (rather than some out-of-band protocol negotiation mechanism) iff any ALPN
+                // protocols were configured locally or offered by the client. This helps prevent
+                // successful establishment of connections between peers that can't understand
+                // each other.
+                if cx.common.alpn_protocol.is_none()
+                    && (!our_protocols.is_empty() || maybe_their_protocols.is_some())
+                {
+                    cx.common
+                        .send_fatal_alert(AlertDescription::NoApplicationProtocol);
+                    return Err(Error::NoApplicationProtocol);
+                }
+
                 match hello.get_quic_params_extension() {
                     Some(params) => cx.common.quic.params = Some(params),
                     None => {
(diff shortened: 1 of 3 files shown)


Ralith:
CI failures look like pre-existing issues.

rustls/src/client/tls13.rs
```diff
@@ -396,8 +396,17 @@ impl State<ClientConnectionData> for ExpectEncryptedExtensions {
 
         #[cfg(feature = "quic")]
         {
-            // QUIC transport parameters
             if cx.common.is_quic() {
+                // RFC 9001 says: While ALPN only specifies that servers use this alert, QUIC
+                // clients MUST use error 0x0178 to terminate a connection when ALPN negotiation
+                // fails.
+                if cx.common.alpn_protocol.is_none() && !self.config.alpn_protocols.is_empty() {
```
djc, line 403: So what this leaves implicit, I think (here and on the server side), is the spec language around if/how ALPN is mandatory. In 8.1, it says "Unless another mechanism is used for agreeing on an application protocol, endpoints MUST use ALPN for this purpose."; I think you're translating that into the `!self.config.alpn_protocols.is_empty()` here, but it would be nice to make that more explicit.

Also, have you considered putting this in `process_alpn_protocol()`, separately guarded by `cx.common.is_quic()`? While that function is intended for both 1.2 and 1.3, it feels like this logic does fit in with what you've added here.

rustls/src/server/hs.rs
```diff
@@ -120,6 +120,17 @@ impl ExtensionProcessing {
         #[cfg(feature = "quic")]
         {
             if cx.common.is_quic() {
+                // QUIC has strict ALPN, unlike TLS's more backwards-compatible behavior. RFC 9001
+                // says: The server MUST treat the inability to select a compatible application
+                // protocol as a connection error of type 0x0178
+                if cx.common.alpn_protocol.is_none()
+                    && (!our_protocols.is_empty() || maybe_their_protocols.is_some())
```
djc, line 127: This could similarly stand some prose that elaborates on what it's trying to achieve exactly, or maybe more exactly, why this is a good way to achieve what we're trying to achieve.

djc (commented):
I think it makes sense to move this into rustls! Some comments below about the best way to integrate it.

djc:
(If you rebase, the CI issues should go away.)

rustls/src/client/tls13.rs
Ralith: Moved the logic into `process_alpn_protocol` (where it indeed feels natural) and expanded on the rationale a bit.

rustls/src/server/hs.rs
Ralith: Expanded on the rationale.

djc (approved):

Issue #40, opened 2016-12-08
DTLS support


Yamakaky:
Do you plan on supporting DTLS (TLS over UDP)?

ctz:
No plans at the minute. I know it's used for some VPNs, webrtc and some iot things. What is your use case?

Yamakaky:
VPN ^^

manuels:
Would you see DTLS within the scope of this library (i.e. would you accept pull request)?

manuels:
I am currently playing around implementing DTLS. Would it be the best to do this on the tls13 branch?

ctz:
> Would you see DTLS within the scope of this library

Perhaps. I'm not sure some of the external APIs are really suitable, to the extent that it might make sense as a separate crate built from the same codebase.

> Would it be the best to do this on the tls13 branch?

Yes. Though beware that this branch is still in major churn and is overdue a serious refactoring of `server_hs.rs`/`client_hs.rs`

manuels:
In case you are interested: https://github.com/manuels/rustls/tree/dtls12
(although I just implemented Payload data structures and parsers up until now)

ctz:
Looking good! 👍

manuels:
Ok, I figured out that the situation is a bit complex, and there are several ways to implement DTLS specific stuff. Since this decision can have quite an extensive impact on the codebase I'd like to discuss how to proceed.

# The Problem
DTLS requires some additional fields in `Message` and `HandshakeMessagePayload` and `ClientHelloPayload` as well as an additional handshake message `HelloVerifyRequest`. Additionally handshakes can be fragmented on the handshake protocol level instead of the record layer level.
See the sketch where modifications are highlighted.

![dtls](https://cloud.githubusercontent.com/assets/31315/22396479/0158c544-e55b-11e6-8f60-1102e9b1f40e.png)

# Possible Solutions

## `Option`s
The additional fields could be integrated as `Option`s into the regular TLS data types and would just be ignored in the TLS implementation. E.g.

    pub struct ClientHelloPayload {
        pub client_version: ProtocolVersion,
        pub random: Random,
        pub session_id: SessionID,
        pub cookie: Option<Cookie>, // <-- !!!
        pub cipher_suites: Vec<CipherSuite>,
        pub compression_methods: Vec<Compression>,
        pub extensions: Vec<ClientExtension>,
    }

## Traits
[...]

PR #874, opened 2021-12-10
Add some DTLS Enum values

ShadowJonathan:
As i'm working on #40, I'm realising its easier to interleave Rustls vetted types with some custom logic for DTLS modded on top of it, however, for that, I require that some core enums at least recognise DTLS versions and variants, such as HandshakeType and ProtocolVersion.

---

DTLS v1.0 version corresponds to [RFC 4347, section 4.1](https://datatracker.ietf.org/doc/html/rfc4347#section-4.1);

> version
>  The version of the protocol being employed.  This document scribes DTLS Version 1.0, which uses the version { 254, 255 }.
>  The version value of 254.255 is the 1's complement of DTLS Version 1.0.

DTLS v1.2 version corresponds to [RFC 6347, section 4.1](https://datatracker.ietf.org/doc/html/rfc6347#section-4.1);

> version
>  The version of the protocol being employed.  This document describes DTLS version 1.2, which uses the version { 254, 253 }.
>  The version value of 254.253 is the 1's complement of DTLS version 1.2.

The HelloVerifyRequest message type number corresponds to [RFC 6347, section 4.2.1](https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.1):

> The HelloVerifyRequest message type is hello_verify_request(3).

(And is also confirmed on the [TLS Parameters](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml) page)


ShadowJonathan:
CI is unhappy because of something on master :(

ctz (approved):

ctz:
Thanks! Merged as c8598c2ff

Issue #872, opened 2021-12-10
Support read_buf for efficient reading of plaintext into uninitialized memory


jsha:
`read_buf` (RFC 2930) has landed on nightly:

https://github.com/rust-lang/rfcs/blob/master/text/2930-read-buf.md#summary
https://doc.rust-lang.org/nightly/std/io/struct.ReadBuf.html
https://doc.rust-lang.org/nightly/std/io/trait.Read.html#method.read_buf

It would be useful to have support for this in rustls. In particular, rustls-ffi currently requires that its users fully initialize the buffer passed to `rustls_connection_read`, and face undefined behavior if they don't. This is an unusual requirement for C libraries, and creates a big risk of UB. If we can drop that requirement, we could improve safety for rustls-ffi users, and also potentially improve performance.

djc:
I think we'll wait until it's on stable, or do you think this is important enough that we should have a nightly feature?

jsha:
I think it's worth having a nightly feature, mainly because I'd like to be able to give feedback upstream about whether it meets our needs while the feature is still unstable. I'm assuming it will - lots of smart people have looked at the design - but it's still good to actually test.

divergentdave:
To provide some background on the motivation here, there are two ways `read_buf` can reduce memory initialization overhead, through amortization, or eliminating the need to initialize.

In the most common case, a caller will create a `ReadBuf` with a newly allocated buffer and use it multiple times in `Read::read_buf()` calls. The provided method from the trait will call `ReadBuf::initialize_unfilled()` which initializes the buffer on the first call, and then call `read()` (note that this preserves backwards compatibility). Subsequent calls with the same `ReadBuf` won't pay any costs for buffer initialization, since the `ReadBuf` struct tracks how much of the buffer has been initialized, and `ReadBuf::initialize_unfilled()` can rely on that. Thus, we initialize the buffer once, and amortize that across however many times we call `Read::read_buf() with it.

[...]

PR #877, opened 2021-12-12
Support for feature(read_buf)

divergentdave:
This implements what was proposed in #872. I created a new nightly-only Cargo feature to enable the new code. Note that I had to conditionally relax `forbid(unsafe_code)` and `forbid(unstable_features)`. I juggled CI build steps around, to ensure we don't try to enable the read_buf flag via --all-features with a stable toolchain. I also added a step to run a new `ChunkVecBuffer` test in Miri.

The core unsafe buffer writing code is in `ChunkVecBuffer::read_buf()`, and the rest of the changes in `<Reader as Read>::read_buf()`, `<Stream as Read>::read_buf()`, and `<StreamOwned as Read>::read_buf()` are all creating parallel functions to each `read()` that ultimately delegate to `ChunkVecBuffer::read_buf()`.

### DIFF
--- a/rustls/src/vecbuf.rs
+++ b/rustls/src/vecbuf.rs
@@ -100,50 +100,13 @@ impl ChunkVecBuffer {
     }
 
     #[cfg(feature = "read_buf")]
-    #[allow(unsafe_code)]
     /// Read data out of this object, writing it into `buf`.
     pub(crate) fn read_buf(&mut self, buf: &mut io::ReadBuf<'_>) -> io::Result<()> {
-        use std::mem::MaybeUninit;
-
-        while !self.is_empty() {
-            // There are three unsafe calls in this block to justify. First,
-            // `ReadBuf::unfilled_mut()` requires that we "not de-initialize portions of the buffer
-            // that have already been initialized." We write to the buffer using
-            // `std::ptr::copy_nonoverlapping`, with a `[u8]` slice from this `ChunkVecBuffer` as
-            // the source. All memory in the `[u8]` slice must be initialized, and the length of
-            // the copy is less than or equal to the length of the `[u8]` slice, so we will not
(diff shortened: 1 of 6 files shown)


rustls/src/vecbuf.rs
```diff
@@ -99,6 +99,56 @@ impl ChunkVecBuffer {
...
+            // bitwise copies is acceptible. The count argument is less than or equal to both the
+            // length of the slice we are copying from, and the length of the unfilled part of the
+            // buffer we are copying to. Thus, the source pointer is valid for reads of this
+            // length, and the destination pointer is valid for writes of this length.
+            //
+            // Lastly, `ReadBuf::assume_init` requires that "the first `n` unfilled bytes of the
+            // buffer have already been initialized". We write into the unfilled part of the buffer
+            // using `std::ptr::copy_nonoverlapping`, and, as argued above, this causes these bytes
+            // to become initialized, since we are copying from another buffer that is initialized.
+            unsafe {
```
ctz, line 131: So I haven't looked at this in detail, and I appreciate the thoroughness that is going into this and the commentary here, but my predominant question is why this doesn't call `ReadBuf::append()` with `&chunk[..used])`? It seems to me that is equivalently efficient and means we have a route to supporting this feature and maintaining `forbid(unsafe_code)`

ctz: To expand: I'd really, really, really like to keep `forbid(unsafe_code)` crate-wide.

divergentdave: Ah, thanks, I wasn't aware of that. Yeah, that's a safe abstraction of exactly what I was doing.

djc (approved):
Nice, thanks for doing the work on this!

ctz:
Thanks!

PR #878, opened 2021-12-13
tlsserver: fix new compile warning

ctz:
cmd_echo/cmd_http/cmd_forward are mutually exclusive, but
docopt's representation means the compiler doesn't know
this.  There is evidently a new/improved compiler warning
that catches that cmd_forward is written by never read.


djc:
Did you miss #867? 🙂

djc (approved):

djc:
(How would you feel about migrating from docopt to structopt? I much prefer the latter because it generates documentation from a type definition rather than you having to write both and then docopt generating the glue.)

ctz:
Doh! Yes I did. Merged that one.

Yes I just saw that docopt is unmaintained now, so it's probably a good idea to move to a different one.

Issue #808, opened 2021-09-02
Web support


autumnontape:
It seems ring has partial support for WebAssembly with web-sys (which is currently only used to access `crypto.getRandomValues()`). I'd like to use rustls on the Web, so I'm filing this issue to talk about what else would be necessary to make that possible. I might be willing to work on this.

I say "Web" specifically rather than just "WebAssembly" because I think it makes sense for a TLS implementation on the Web to depend on Web Crypto API, which would make support for other WebAssembly platforms, like WASI, a separate problem. I'm no crypto expert, though.

djc:
Currently it seems like the library builds with `--target wasm32-unknown-unknown`. I'm not sure what else would be needed or how to test it (`cargo test --target wasm32-unknown-unknown` fails with `cannot execute binary file`). In principle I would be happy to help make this happen, but I don't yet have much experience with WebAssembly and I don't know how invasive the changes would be.

Also, it would be interesting to know more about your use case.

autumnontape:
In terms of getting Web builds functioning, I assume most of the required changes would be in ring, with rustls maybe requiring some Cargo.toml changes. Running tests with wasm-bindgen would be more of a pain, since the built-in `#[test]` attribute doesn't work with it. It requires a special `#[wasm_bindgen_test]` attribute.

I recently discovered that the `SubtleCrypto` API on the Web only provides async versions of functions, even in workers, which seems like a problem because I expected to be able to depend on it for timing guarantees. Since plumbing async through every part of rustls that uses timing-sensitive algorithms seems like a non-starter, and Asyncify is pretty alien to Rust, this may have to wait for constant-time WebAssembly.

My use case is securing a WebSocket connection to a server without a TLS certificate. I'm working on a self-hosted service that wouldn't require a domain name to use, just a server with a public IP address and a trusted static host to serve the client code. It's a hobby project.

djc:
For the *ring* bits, it's probably good to open a separate issue in its repo. We would be happy to accept some Cargo manifest changes, though I think we'd need to somehow be able to run some tests in WebAssembly as well.

awakecoding:
Is there an update on this, and is there a corresponding issue in Ring? If the library can be compiled to WebAssembly, has anybody managed to package it in a web app to try and use it to see if it's in a functional state? I would be interested in using rustls in the browser for an RDP client (IronRDP). The web-based Microsoft RDP clients all use an in-browser TLS stack, and the protocol makes it very difficult to avoid this.

[...]

PR #879, opened 2021-12-13
Use clap macro for rustls-mio examples

djc:
docopt is no longer maintained. Meanwhile, structopt provides a nice way of declaratively setting up the CLI options and arguments.

```
tlsclient 0.0.1
The default `port` is 443. By default, this reads a request from stdin (to EOF) before making the connection.  --http
replaces this with a basic HTTP GET request for /.

If --cafile is not supplied, a built-in set of CA certificates are used from the webpki-roots crate.

USAGE:
    tlsclient [FLAGS] [OPTIONS] <hostname>

FLAGS:
    -h, --help          
            Prints help information

        --http          
            Send a basic HTTP GET request for /

        --insecure      
            Disable certificate verification

        --no-sni        
            Disable server name indication support

        --no-tickets    
            Disable session ticket support

    -V, --version       
            Prints version information

        --verbose       
            Emit log output


OPTIONS:
        --auth-certs <auth-certs>          
            Read client authentication certificates from `auth-certs`.
            
            `auth-certs` must match up with `auth-key`.
        --auth-key <auth-key>              
            Read client authentication key from `auth-key`

        --cache <cache>                    
            Save session cache to file `cache`

        --cafile <cafile>                  
            Read root certificates from `cafile`

        --max-frag-size <max-frag-size>    
            Limit outgoing messages to `max_frag_size` bytes

    -p, --port <port>                      
            Connect to `port` [default: 443]

        --proto <proto>...                 
            Send ALPN extension containing `proto`. May be used multiple times to offer several protocols

        --protover <protover>...           
            Disable default TLS version list, and use `protover` instead. May be used multiple times

        --suite <suite>...                 
            Disable default cipher suite list, and use `suite` instead. May be used multiple times


ARGS:
    <hostname>
```

```
rustls-mio 0.0.1
Runs a TLS server on :port.  The default port is 443.

`--certs' names the full certificate chain, `--key' provides the RSA private key.

USAGE:
    tlsserver [FLAGS] [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help            
            Prints help information

        --require-auth    
            Send a fatal alert if the client does not complete client authentication

        --resumption      
            Support session resumption

        --tickets         
            Support tickets

    -V, --version         
            Prints version information

        --verbose         
            Emit log output


OPTIONS:
        --auth <auth>               
            Enable client authentication, and accept certificates signed by those roots provided in `certs`

        --certs <certs>             
            Read server certificates from `certs`.
            
            This should contain PEM-format certificates in the right order (the first certificate should certify `key`,
            the last should be a root CA).
        --key <key>                 
            Read private key from `key`.
            
            This should be an RSA private key or PKCS8-encoded private key, in PEM format.
        --ocsp <ocsp>               
            Read DER-encoded OCSP response from OCSP and stable to certificate. Optional

    -p, --port <port>               
            Listen on `port` [default: 443]

        --proto <proto>...          
            Send ALPN extension containing `proto`. May be used multiple times to offer several protocols

        --protover <protover>...    
            Disable default TLS version list, and use `protover` instead. May be used multiple times

        --suite <suite>...          
            Disable default cipher suite list, and use `suite` instead. May be used multiple times


SUBCOMMANDS:
    echo       Write back received bytes
    forward    Forward traffic to/from given port on localhost
    help       Prints this message or the help of the given subcommand(s)
    http       Do one read, then write a bodged HTTP response and cleanly close the connection
```

### DIFF
--- a/rustls-mio/Cargo.toml
+++ b/rustls-mio/Cargo.toml
@@ -20,13 +20,11 @@ rustls-pemfile = "0.2.0"
 sct = "0.7"
 
 [dev-dependencies]
+clap = { version = "3", features = ["derive"] }
 ct-logs = "0.9"
-docopt = "~1.1"
 env_logger = "0.9.0"
 mio = { version = "0.8", features = ["net", "os-poll"] }
 regex = "1.0"
-serde = "1.0"
-serde_derive = "1.0"
 webpki-roots = "0.22"
 ring = "0.16.20"
 
--- a/rustls-mio/examples/tlsclient.rs
+++ b/rustls-mio/examples/tlsclient.rs
@@ -1,20 +1,15 @@
-use std::process;
-use std::sync::{Arc, Mutex};
-
-use mio::net::TcpStream;
-
 use std::collections;
 use std::convert::TryInto;
 use std::fs;
 use std::io;
 use std::io::{BufReader, Read, Write};
 use std::net::SocketAddr;
+use std::process;
 use std::str;
+use std::sync::{Arc, Mutex};
 
-#[macro_use]
-extern crate serde_derive;
-
-use docopt::Docopt;
+use clap::Parser;
+use mio::net::TcpStream;
 
 use rustls::{OwnedTrustAnchor, RootCertStore};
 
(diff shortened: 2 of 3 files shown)


djc:
`tlsserver` seems to trigger some edge cases in structopt here, will try to investigate soon.

djc:
Currently blocked on https://github.com/clap-rs/clap/issues/3267.

briansmith:
> Currently blocked on [clap-rs/clap#3267](https://github.com/clap-rs/clap/issues/3267).

* It seems a bit much to ask the clap developers to change their MSRV policy to accommodate our *examples*. I think the clap MSRV policy is good as it is.
* rustls-mio isn't even published to crates.io, so why do we need example programs for it?
* Why do we continue to use rustls-mio in the first place, instead of using tokio-rustls like everybody else?

djc:
> It seems a bit much to ask the clap developers to change their MSRV policy to accommodate our examples. I think the clap MSRV policy is good as it is.

It's also an issue for trust-dns (https://github.com/bluejekyll/trust-dns/pull/1616) and for Quinn.

As I understand it, rustls-mio contains a bunch of integration tests that depend on these example programs.

briansmith:
Even if we go by what Tokio does, which is 6 months, then 1.53 would be OK and we're very close to 1.54 being an acceptable MSRV for us now. IDK of anything in 1.53 or 1.54 that we need, but also I humbly suggest we don't bother people too much if they're just requiring 1.54.

> As I understand it, rustls-mio contains a bunch of integration tests that depend on these example programs.

Perhaps we should split tlsclient and tlsserver into two programs each: A demo/utility program and the test engine. It would be nice if the tests didn't have a 3rd-party command line parsing dependency at all, or ultimately even a mio dependency. That will be needed if/when we want to address the ability for Rustls to run on the web (https://github.com/rustls/rustls/issues/808), which multiple people have requested (publicly and privately).

ctz:
The reason rustls-mio exists as it does is to separate out and remove mio as dependency of the core crate (even as a dev-dependency). See #241 for background. It definitely does not stand in the way of running the core crate under wasm.

briansmith:
> It definitely does not stand in the way of running the core crate under wasm.

There are (at least) two different wasm32 environments that people are interested in: wasm32-wasi and wasm32 browser, tunneled through WebSockets. I expanded on this in https://github.com/rustls/rustls/issues/808.

cpu:
I'm going to close this draft PR since it's quite out of date with `main` and doesn't seem like a high priority to fix. I suspect some of the issues that made it difficult to merge back in ~2021 might be resolved but it's probably better to tackle in a fresh PR when someone has time.

PR #887, opened 2021-12-22
Correct docs for WebPkiVerifier server impl

### LINKED ISSUE #885: Inaccurate docs for ServerCertVerifier impl of WebPKIVerifier
The docs for the `ServerCertVerifier` impl of `WebPKIVerifier` claim that `verify_server_cert` verifies that OCSP data was present, but the function's result as implemented does not depend on the `ocsp_response` argument.

I don't think this is a particularly meaningful security vuln, given that the guarantee didn't even say anything about whether the OCSP data was valid, but the docs should probably be corrected. I'll send a PR in a few

teisenbe:
The ServerCertVerifier impl docs were claiming that `verify_server_cert` would verify that OCSP data was present,
but the function's result as implemented does not depend on the `ocsp_response` field.

Fixes #885

### DIFF
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -283,7 +283,6 @@ impl ServerCertVerifier for WebPkiVerifier {
     /// - Signed by a  trusted `RootCertStore` CA
     /// - Not Expired
     /// - Valid for DNS entry
-    /// - OCSP data is present
     fn verify_server_cert(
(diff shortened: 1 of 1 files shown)


djc:
This seems reasonable. I filed #888 about starting to actually verify OCSP responses.

djc (approved):

PR #891, opened 2021-12-23
Avoid copying key agreement key material

### LINKED ISSUE #890: Avoid copying key agreement input keying material onto the heap
We have this in `KeyExchangeResult`:

```rustls
    /// Completes the key exchange, given the peer's public key.  The shared
    /// secret is returned as a KeyExchangeResult.
    pub(crate) fn complete(self, peer: &[u8]) -> Option<KeyExchangeResult> {
        let peer_key = ring::agreement::UnparsedPublicKey::new(self.skxg.agreement_algorithm, peer);
        let pubkey = self.pubkey;
        ring::agreement::agree_ephemeral(self.privkey, &peer_key, (), move |v| {
            Ok(KeyExchangeResult {
                pubkey,
                shared_secret: Vec::from(v),
            })
        })
        .ok()
    }
```

If you look at the callers of `complete`, you will see they all immediately construct a `ring::hkdf::Prk` (for TLS 1.3) or the TLS 1.2 counterpart. 

Imagine instead the function were instead:

```rustls
    /// Completes the key exchange, given the peer's public key.  The shared
    /// secret is returned as a KeyExchangeResult.
    pub(crate) fn complete<T>(self, peer: &[u8], make_omk: impl FnOnce(&[u8]) -> T) -> Option<KeyExchangeResult<T>> {
        let peer_key = ring::agreement::UnparsedPublicKey::new(self.skxg.agreement_algorithm, peer);
        let pubkey = self.pubkey;
        ring::agreement::agree_ephemeral(self.privkey, &peer_key, (), move |v| {
            Ok(KeyExchangeResult {
                pubkey,
                shared_secret: make_okm(v),
            })
        })
        .ok()
    }
```

The *ring* API is designed the way it is in an attempt to steer applications towards doing the key derivation immediately in the function passed to `agree_ephemeral`, instead of making a copy. And more generally the *ring* API is designed to guide applications towards keeping short-lived secrets on the stack so they don't leak into the heap unnecessarily.

OTOH, I realize that some of the aspects of the protocol designs of TLS 1.2 (with EMS being optional) and TLS 1.3 (with early data) complicate this, so it isn't a trivial refactoring. Still, I think it is worth doing because it'd be better key management, it'd be more efficient, and it would reduce the tech debt that impedes implementing more flexible memory management (alternative allocators, no-abort-on-allocation-failure).

If we find that it is impossible or impractical to do this with the existing *ring* APIs, then I'll extend the *ring* API to support a version of `agree_ephemeral` that returns the key material in an `Ikm` structure that keeps the data on the stack. However, I'd like to avoid that if we can because I was hoping to avoid ever copying the key material, to minimize the leakage.

djc:
Fixes #890. Thanks for the nerdsnipe, @briansmith...

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -126,7 +126,7 @@ mod server_hello {
                     }
 
                     let secrets =
-                        ConnectionSecrets::new_resume(&self.randoms, suite, resuming.secret());
+                        ConnectionSecrets::new_resume(self.randoms, suite, resuming.secret());
                     self.config.key_log.log(
                         "CLIENT_RANDOM",
                         &secrets.randoms.client,
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -25,6 +25,7 @@ use crate::client::common::ClientAuthDetails;
 use crate::client::common::ServerCertDetails;
 use crate::client::{hs, ClientConfig, ServerName};
 
+use ring::agreement::PublicKey;
 use ring::constant_time;
 
 use std::sync::Arc;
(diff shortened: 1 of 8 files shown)


ctz (approved):

djc:
Going to just merge this for now, happy to address post-merge feedback though.

PR #895, opened 2021-12-29
Add abstraction for TLS vectors

djc:
And remove type aliases, which don't actually provide much abstraction.


ctz:
I think the value of type aliases like `pub type ECPointFormatList = TlsVec<u8, ECPointFormat>;` are to put the information that "a list of ECPointFormats has a u8 length" in once place, and not repeat that (rather trivial) information everywhere.

djc:
I guess that does convey some value. For me that value seems less than the opaqueness of the `ECPointFormatList` name which doesn't tell me much about what the type actually is, especially across module boundaries. I appreciate the preciseness and less indirection of using the type directly -- I've found myself using type aliases less and less, maybe  because I tend to just find it confusing that there are two (or more) ways to spell exactly the same type. This is maybe similar to how I often try to avoid having single-caller functions.

But, happy to put back some of the type aliases if you prefer it that way.

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
+#[derive(Clone, Debug)]
+pub struct TlsVec<L, T> {
+    items: Vec<T>,
+    len: PhantomData<L>,
+}
+
+impl<L: LengthPrefix, T: Codec> Codec for TlsVec<L, T>
+where
+    usize: From<L>,
+{
```
briansmith, line 193: In order to avoid code bloat due to monomorphizing, as much of the code as possible in these implementations should be factoed out into non-generic functions.

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
+        let len_offset = bytes.len();
+        bytes.extend(L::ZERO_LEN);
+
+        for i in &self.items {
+            i.encode(bytes);
+        }
 
-    for i in items {
-        i.encode(bytes);
+        let len = bytes.len() - len_offset - L::BYTES;
```
briansmith, line 202: This is assuming that the `encode` functions will never shrink the vector, which is true for a valid implementation of `encode`, but not guaranteed by the type system.

For this reason and the previous one, we should define a new trait, something like `trait Extend { type Error; fn extend(&mut self, bytes: &[Self::Item]) -> Result<(), Self::Error> }` and change the definition of the `encode` functions to `encode(&mut dyn Extend<Item=u8>)`.

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
+        L::write_into(len, len_offset, bytes);
```
briansmith, line 203: This might fail if `len` is larger than what can be encoded in `L::BYTES` bytes, right?

There is no need for `L::write_into` to be implemented for each type; it is easy to write a non-generic function that does it.

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
     }
 
-    let len = bytes.len() - len_offset - 1;
-    debug_assert!(len <= 0xff);
-    bytes[len_offset] = len as u8;
-}
+    fn read(r: &mut Reader) -> Option<Self> {
```
briansmith, line 206: Why not return a `Result` here, since any length that's too long to be represented in `L::BYTES` is a syntax error by the sender?

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
+#[allow(clippy::use_self)]
+impl LengthPrefix for u8 {
+    fn write_into(len: usize, offset: usize, bytes: &mut [u8]) {
+        debug_assert!(len <= 0xff);
+        bytes[offset] = len as u8;
     }
 
-    Some(ret)
+    const BYTES: usize = 1;
+    const ZERO_LEN: &'static [u8] = &[0];
```
briansmith, line 289: Given `ZERO_LEN`, `BYTES` is just `ZERO_LEN.len()`. Conversely, given `BYTES`, `ZERO_LEN` is just `[0u8; Self::BYTES]`.

Either way, `write_into` can be implemented once given either `BYTES` or `ZERO_LEN`. So, I think the `LengthPrefix` trait should just have a single member: `const BYTES: usize`.

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
+            .unwrap();
+        out.copy_from_slice(&len_bytes[1..]);
     }
 
-    Some(ret)
+    const BYTES: usize = 3;
+    const ZERO_LEN: &'static [u8] = &[0, 0, 0];
+}
+
+pub trait LengthPrefix: Codec + Copy + Debug {
```
briansmith, line 320: In general, traits should be defined ahead of their implementations.

Let's document the trait.

Given the code above, it is possible to construct a `TlsVec` that will fail to encode because the length of the vector is larger than can be encoded. This will be discovered only during encoding of the `TlsVec` and not before. This in particular should be documented because it is surprising.

rustls/src/msgs/handshake_test.rs
```diff
@@ -358,28 +358,34 @@ fn get_sample_clienthellopayload() -> ClientHelloPayload {
         client_version: ProtocolVersion::TLSv1_2,
         random: Random::from([0; 32]),
         session_id: SessionID::empty(),
-        cipher_suites: vec![CipherSuite::TLS_NULL_WITH_NULL_NULL],
-        compression_methods: vec![Compression::Null],
+        cipher_suites: vec![CipherSuite::TLS_NULL_WITH_NULL_NULL].into(),
```
briansmith, line 361: It seems wrong we're using `vec!` here and elsewhere. I would expect `TlsVec::from(&[CipherSuite::TLS_NULL_WITH_NULL_NULL])` or so. Or otherwise a `tls_vec!`.

rustls/src/msgs/codec.rs
```diff
@@ -174,88 +177,149 @@ impl Codec for u64 {
...
+impl<L: LengthPrefix, T: Codec> Codec for TlsVec<L, T>
+where
+    usize: From<L>,
+{
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        let len_offset = bytes.len();
+        bytes.extend(L::ZERO_LEN);
+
+        for i in &self.items {
+            i.encode(bytes);
```
briansmith, line 199: This can fail if the sum of the encoded bytes is larger than `isize::MAX` or `usize::MAX`. I guess it fails with a panic? I guess we need a follow-up bug to change the signature of `encode` to allow it to indicate failure to extend.

briansmith (changes requested):

briansmith:
> I think the value of type aliases like pub type ECPointFormatList = TlsVec<u8, ECPointFormat>; are to put the information that "a list of ECPointFormats has a u8 length" in once place, and not repeat that (rather trivial) information everywhere.

I agree that the aliases are useful for this reason and we should keep them.

rustls/src/msgs/codec.rs
briansmith: We should have an issue to eliminate all the resizing/reallocation that will occur here (and elsewhere). I believe that we'll end up doing that by eliminating all these `Vec`s and encoding directly into the output buffer in the future.

djc:
I think I'll just try a different approach.

Issue #884, opened 2021-12-15
Server-side support for TLS1.3 early data


ctz:
This would be nice, but the protocol design is pretty fraught and needs a reasonable amount of care.

This also unlocks improving testing of the client-side support.

Issue #232, opened 2019-02-03
0-RTT should be rejected for stale PSKs


Ralith:
This only affects QUIC, since regular TLS does not yet expose receiving 0-RTT

RFC 8446 §4.2.10:
> For PSKs provisioned via NewSessionTicket, a server MUST validate
   that the ticket age for the selected PSK identity (computed by
   subtracting ticket_age_add from PskIdentity.obfuscated_ticket_age
   modulo 2^32) is within a small tolerance of the time since the ticket
   was issued (see Section 8).  If it is not, the server SHOULD proceed
   with the handshake but reject 0-RTT, and SHOULD NOT take any other
   action that assumes that this ClientHello is fresh.

I'm happy to implement this, but it's not clear to me how `ticket_age_add` should be determined by the server. Does this need to be added to `ServerSessionValue`, perhaps?

briansmith:
Just to be clear, @ctz, was this closed becuase #897 fixes it?

ctz:
Correct yes

PR #897, opened 2021-12-30
Implement TLS1.3 "early data" for TCP-TLS servers

ctz:
This was previously implement for QUIC. This is intended to address #884.

- #232 is also addressed by 967c0cb


Ralith:
Should https://github.com/rustls/rustls/issues/232 be addressed?

rustls/src/server/tls13.rs
```diff
@@ -63,6 +63,13 @@ mod client_hello {
 
     use super::*;
 
+    #[derive(PartialEq)]
+    pub(super) enum EarlyData {
```
djc, line 67: Style nit: maybe name this `EarlyDataState`, since it doesn't contain any early data itself?

rustls/src/server/server_conn.rs
```diff
@@ -221,9 +222,17 @@ pub struct ServerConfig {
...
-    #[doc(hidden)]
+    /// Amount of early data to accept for sessions created by
+    /// this config.  Specify 0 to disable early data.  The
+    /// default is 0.
+    ///
+    /// Read the early data via [`ServerConnection::early_data`].
+    ///
+    /// The units for this are _both_ plaintext bytes, _and_ ciphertext
+    /// bytes, depending on whether the server accepts a client's early_data
+    /// or not.  It is recommended therefore to include some slop in
```
djc, line 233: Nit: recommended therefore -> therefore recommended?

rustls/src/server/server_conn.rs
```diff
@@ -251,6 +260,39 @@ impl ServerConfig {
     }
 }
 
+/// A structure which allows reading of early data (also known as
```
djc, line 263: For best results in rustdoc, the first sentence of a docstring should fit on one line.

rustls/src/server/tls13.rs
```diff
@@ -436,24 +510,32 @@ mod client_hello {
...
+        // If not doing early_data after all, this is corrected later to the handshake
+        // keys (now stored in key_schedule).
         cx.common
             .record_layer
             .set_message_encrypter(suite.derive_encrypter(&server_key));
-        cx.common
-            .record_layer
-            .set_message_decrypter(suite.derive_decrypter(&client_key));
+
+        if let Some(edck) = &early_data_client_key {
```
djc, line 526: Nit: I feel like renaming `edck` -> `key` would make this easier to follow.

rustls/src/server/tls13.rs
```diff
@@ -504,6 +586,57 @@ mod client_hello {
...
+        resumedata: Option<&persist::ServerSessionValue>,
+    ) -> EarlyData {
+        let early_data_requested = client_hello.early_data_extension_offered();
+        let rejected_or_disabled = if early_data_requested {
+            EarlyData::RequestedButRejected
+        } else {
+            EarlyData::Disabled
+        };
+
+        if let Some(resume) = resumedata {
```
djc, line 603: Maybe do a `let resume = match resumedata { };` here to reduce rightward drift, below?

rustls/src/server/tls13.rs
```diff
@@ -504,6 +586,57 @@ mod client_hello {
...
+        client_hello: &ClientHelloPayload,
+        suite: &'static Tls13CipherSuite,
+        resumedata: Option<&persist::ServerSessionValue>,
+    ) -> EarlyData {
+        let early_data_requested = client_hello.early_data_extension_offered();
+        let rejected_or_disabled = if early_data_requested {
+            EarlyData::RequestedButRejected
+        } else {
+            EarlyData::Disabled
+        };
```
djc, line 601: Style: writing this as a `match` expression would be more compact.

rustls/src/server/tls13.rs
```diff
@@ -504,6 +586,57 @@ mod client_hello {
         common.send_msg(m, false);
     }
 
+    fn decide_if_early_data_allowed(
```
djc, line 589: Style nit: note that the argument order here is sort of reversed from other nearby functions, which sort roughly from most changing/unique to least changing/unique.

ctz:
> Should #232 be addressed?

Ah yes, thanks for the reminder! I'll address that here.

rustls/src/server/tls13.rs
```diff
@@ -717,6 +854,28 @@ mod client_hello {
...
+    skip_data_left: usize,
+    next: Box<hs::ExpectClientHello>,
+}
+
+impl State<ServerConnectionData> for ExpectAndSkipRejectedEarlyData {
+    fn handle(mut self: Box<Self>, cx: &mut ServerContext<'_>, m: Message) -> hs::NextStateOrError {
+        /* "The server then ignores early data by skipping all records with an external
+         *  content type of "application_data" (indicating that they are encrypted),
+         *  up to the configured max_early_data_size."
+         * (RFC8442, 14.2.10) */
```
djc, line 867: I think this is referencing the wrong RFC?

rustls/src/server/tls13.rs
```diff
@@ -717,6 +854,28 @@ mod client_hello {
...
 }
 
+struct ExpectAndSkipRejectedEarlyData {
+    skip_data_left: usize,
+    next: Box<hs::ExpectClientHello>,
+}
+
+impl State<ServerConnectionData> for ExpectAndSkipRejectedEarlyData {
+    fn handle(mut self: Box<Self>, cx: &mut ServerContext<'_>, m: Message) -> hs::NextStateOrError {
+        /* "The server then ignores early data by skipping all records with an external
```
djc, line 864: After reading the commit message, comments and code, it's still a little unclear to me how we avoid acidentally dropping traffic-encrypted application data. Are early data records separated from traffic by other handshake messages? I feel like some context is missing and/or the exact mechanism is quite spread out over the `RecordLayer`/state machine that it's pretty hard to follow what's going on exactly.

djc (commented):
Awesome work on getting this feature in!

I've read through all the commits and left a bunch of review comments, most of them smaller remarks on style. I don't yet feel like I understand exactly how this design works in skipping over unwanted early data with the odd unencrypted/encrypted data size, so maybe there are still improvements to be made in that regard either in the comments or in how responsibility is distributed over different components, but given the test coverage and bogo tests it seems very likely to be correct.

rustls/src/server/tls13.rs
ctz: > Are early data records separated from traffic by other handshake messages?

Yes, the EndOfEarlyData handshake message.

The multiple mechanisms doing the same thing (which is: skipping over early data messages we can't or don't want to process) reflect that there are two cases that happens in:

1. we (the server) are happy to proceed with a full handshake - in this case we can't decrypt the early data messages and have to do trial decryption until we find a message encrypted with the handshake keys -- this is the case handled by the record_layer.rs and conn.rs changes
2. we not happy to proceed with the handshake, and send a HelloRetryRequest: this means the next client message (after the early data messages) will be a plaintext ClientHello: ExpectAndSkipRejectedEarlyData  deals with that case

Ideally we could move (1) into the state machine, but I can't see a good way of doing that without also moving decryption, handshake reassembly, alert handling, etc into each state. That itself could clear up the state held in record_layer.rs, and version-specific decisions made in common conn.rs code.

PR #901, opened 2022-01-07
Fix deny-by-default Clippy lints in examples and api.rs

zombiepigdragon:
`rust-analyzer` running Clippy on save was continuously lighting up some examples as errors, so I applied the appropriate lints. After this PR, `cargo clippy --all-targets` completes successfully, although the warning lints were left as-is.


djc (approved):

djc:
Thanks!

djc:
If you also want to propose fixes for the warn-by-default suggestions, that would be great!

PR #902, opened 2022-01-07
Derive debug for marker types

tsoutsman:
The marker structs (e.g. `CertVerified`) don't implement `Debug`, and so `Result<Marker, E>` don't either. This change also annotates the structs with the \`non_exhaustive\` attribute [preventing them from being constructed outside Rustls](https://doc.rust-lang.org/reference/attributes/type_system.html). Finally, the `HandshakeSignatureValid` docs contained implementation details about all the marker types, which I think was meant to be a regular comment above their definitions.

On a side note, could someone please explain what a 'goto fail'-style elision is and how the `assertion` constructors prevent them?

### DIFF
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -32,15 +32,18 @@ static SUPPORTED_SIG_ALGS: SignatureAlgorithms = &[
     &webpki::RSA_PKCS1_3072_8192_SHA384,
 ];
 
-/// Marker types.  These are used to bind the fact some verification
-/// (certificate chain or handshake signature) has taken place into
-/// protocol states.  We use this to have the compiler check that there
-/// are no 'goto fail'-style elisions of important checks before we
-/// reach the traffic stage.
-///
-/// These types are public, but cannot be directly constructed.  This
-/// means their origins can be precisely determined by looking
-/// for their `assertion` constructors.
+// Marker types.  These are used to bind the fact some verification
+// (certificate chain or handshake signature) has taken place into
+// protocol states.  We use this to have the compiler check that there
+// are no 'goto fail'-style elisions of important checks before we
+// reach the traffic stage.
+//
+// These types are public, but cannot be directly constructed.  This
+// means their origins can be precisely determined by looking
+// for their `assertion` constructors.
+
+/// Zero-sized marker type representing verification of a signature.
+#[derive(Debug)]
 pub struct HandshakeSignatureValid(());
 
 impl HandshakeSignatureValid {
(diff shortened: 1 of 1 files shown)


rustls/src/verify.rs
```diff
@@ -32,50 +32,87 @@ static SUPPORTED_SIG_ALGS: SignatureAlgorithms = &[
...
+/// ```compile_fail
+/// # use rustls::client::HandshakeSignatureValid;
+/// // Doesn't compile
+/// let _ = HandshakeSignatureValid;
+/// // Compiles
+/// let _ = HandshakeSignatureValid::assertion();
+/// ```
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct HandshakeSignatureValid;
```
briansmith, line 57: Normally `non_exhaustive` isn't used for a structure that has no public fields. Instead `HandshakeSignatureValid(())` would be more idiomatic, right? Or am I behind the times?

briansmith:
@tsoutsman I believe there are other cases where `#[non_exhaustive]` makes sense, especially:

```rust
#[derive(Debug, PartialEq)]
pub struct SupportedProtocolVersion {
    /// The TLS enumeration naming this version.
    pub version: ProtocolVersion,
    is_private: (),
}
```
The `is_private: ()` should be removed in favor of `#[non_exhaustive]`. But, I'm not sure what the recommended idiom is for empty structures.

rustls/src/verify.rs
djc: Yeah, that was also my hesitation. I think maybe this has the same effect, but I think it might be less idiomatic. So I  think we don't want to make changes that replace a single private field with a fieldless `#[non_exhaustive]` type.

(I'm guessing this PR may have been triggered by the unidiomatic way the assertion types instantiate their private fields with `0: ()`, which I think triggers some kind of warning? But this could just be fixed by initializing as `HandshakeSignatureValid(())`.)

tsoutsman: Yea fair enough. Now that I think about it `non_exhaustive` isn't as clear. I saw it in a book and thought it looked cleaner. Can I just remove all the changes other than deriving `Debug`?

djc: Sure, that should be fine.

briansmith:
@tsoutsman are you interested in rebasing this on the latest `main` and addressing the above feedback? I think it would be great to have this change made, as it makes implementing and testing custom certificate validators easier.

tsoutsman:
@briansmith Sorry for the delay. Should be good now.

briansmith (approved):
Thanks!

djc (approved):

PR #903, opened 2022-01-07
Stop if no compatible versions are selected.

briansmith:


rustls/src/client/hs.rs
```diff
@@ -219,10 +219,12 @@ fn emit_client_hello_for_retry(
         supported_versions.push(ProtocolVersion::TLSv1_2);
     }
 
-    let mut exts = Vec::new();
-    if !supported_versions.is_empty() {
-        exts.push(ClientExtension::SupportedVersions(supported_versions));
+    if supported_versions.is_empty() {
```
djc, line 222: I believe this is statically unreachable, https://github.com/rustls/rustls/blob/main/rustls/src/builder.rs#L224.

briansmith: In that case, I guess we should just `assert!(!supported_versoins.is_empty());` here.

djc: Or see #913.

PR #913, opened 2022-01-10
Better express version configuration in the type system

### LINKED ISSUE #904: Handle the case where no supported versions are enabled
Check out #903 and focus on this code when emitting ClientHello:

> ```rust
>     let mut exts = Vec::new();
>     if !supported_versions.is_empty() {
>         exts.push(ClientExtension::SupportedVersions(supported_versions));
>     }
> ```

My understanding of the TLS 1.3 spec is that a conformant client MUST always send the supported_versions extension, whether TLS 1.3 is enabled or not. Thus, we must not have any branch that skips adding the extension. It seems like we don't have tests for this?

djc:
`EnabledVersions` becomes an enum which guarantees that at least
one version is enabled. This was already enforced by the config
builder, which will error out if there are no usable cipher suites
(where one aspect of "usable" is whether the suite's version is
enabled).

At the expense of 19 extra lines of code, we can make it clear in the type system that the builder enforces that we always have at least one version enabled.

This is an alternative to #903. Fixes #904.

### DIFF
--- a/rustls/src/versions.rs
+++ b/rustls/src/versions.rs
@@ -26,6 +26,8 @@ pub static TLS13: SupportedProtocolVersion = SupportedProtocolVersion {
 };
 
 /// A list of all the protocol versions supported by rustls.
+///
+// This needs to be kept in sync with the `impl Default for EnabledVersions`.
 pub static ALL_VERSIONS: &[&SupportedProtocolVersion] = &[
     &TLS13,
     #[cfg(feature = "tls12")]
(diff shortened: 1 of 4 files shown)


rustls/src/versions.rs
```diff
@@ -37,41 +39,60 @@ pub static ALL_VERSIONS: &[&SupportedProtocolVersion] = &[
...
 }
 
 impl EnabledVersions {
-    pub(crate) fn new(versions: &[&'static SupportedProtocolVersion]) -> Self {
-        let mut ev = Self {
-            #[cfg(feature = "tls12")]
-            tls12: None,
-            tls13: None,
-        };
+    pub(crate) fn new(versions: &[&'static SupportedProtocolVersion]) -> Option<Self> {
```
briansmith, line 56: Let's remember the goal of the APIs that take `&[&'static Whatever]` is to make it so that link-time dead-code optimization can eliminate all the dead code. In this case, the intent is that `SupportedProtocolVersion` should have a (crate-private) API that allows us to move the bulk of TLS-1.2-only and TLS-1.3-only logic into methods of `SupportedProtocolVersion`, and eliminate the `if` and `match` on version numbers within the shared code. This way, there would be ~0 TLS-1.2-specific code in a project that doesn't enable TLS 1.2.

But the actual API of `EnabledVersions` and `SupportedProtocolVersion` is oriented towards doing the opposite of that.

We should refactor all of this to orient it with the dead code elimination goal. I bet this would also have the benefit of eliminating a lot of the `#[cfg]` logic based on whether the TLS 1.2 feature flag is enabled.

briansmith (changes requested):
> At the expense of 19 extra lines of code

This PR is +65 −32 as of right now, i.e. net +33 lines of code.

> we can make it clear in the type system that the builder enforces that we always have at least one version enabled.

This doesn't really accomplish as much as we'd like because it doesn't consider the "is quic" condition. So, we'd put a lot of effort into ensure we have a non-empty list of versions during configuration, but when we consider "is quic" we'd still have to consider the possibility that the list of was "TLS 1.2 only."

PR #914, opened 2022-01-10
Make side handling more obvious

djc:
Removes the `ConnectionRandoms::we_are_client` and `CommonState::is_client` fields in favor of getting this state from the context. The former was unused in case TLS 1.2 was disabled, leading to a warning. Saves a few lines of code, 1 bit of state in handshake states, 1 bit of state in `CommonState`, and a branch in `process_main_protocol()`. Also defines an `enum Side { Client, Server }` in favor of using `bool`, which makes the resulting code easier to follow.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -1240,8 +1240,8 @@ impl CommonState {
     }
 
     #[cfg(feature = "tls12")]
-    pub(crate) fn start_encryption_tls12(&mut self, secrets: &ConnectionSecrets) {
-        let (dec, enc) = secrets.make_cipher_pair();
+    pub(crate) fn start_encryption_tls12(&mut self, secrets: &ConnectionSecrets, side: Side) {
+        let (dec, enc) = secrets.make_cipher_pair(side);
         self.record_layer
             .prepare_message_encrypter(enc);
         self.record_layer
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -360,13 +360,9 @@ pub(crate) enum Protocol {
 
 #[derive(Debug)]
 pub(crate) struct ConnectionRandoms {
-    pub(crate) side: Side,
     pub(crate) client: [u8; 32],
     pub(crate) server: [u8; 32],
 }
-
-#[cfg(feature = "tls12")]
-static TLS12_DOWNGRADE_SENTINEL: [u8; 8] = [0x44, 0x4f, 0x57, 0x4e, 0x47, 0x52, 0x44, 0x01];
 
 /// How many ChangeCipherSpec messages we accept and drop in TLS1.3 handshakes.
 /// The spec says 1, but implementations (namely the boringssl test suite) get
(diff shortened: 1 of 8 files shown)


rustls/src/conn.rs
```diff
@@ -1356,7 +1331,22 @@ impl Quic {
...
+    Server,
+}
+
+pub(crate) use sealed::SealedSideData;
+
 /// Data specific to the peer's side (client or server).
-pub trait SideData {}
+pub trait SideData: SealedSideData {}
+
+mod sealed {
```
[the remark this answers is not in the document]
djc: The point of this was to keep the associated const crate private. We can't put it in `Sealed` because that's used for other traits as well. We could put it in a separate trait that's directly `pub(crate)` and then use both a private supertraits for `SideData`, but that doesn't seem like a significant improvement.

rustls/src/conn.rs
```diff
@@ -1356,7 +1331,22 @@ impl Quic {
...
+    use crate::msgs::enums::HandshakeType;
+    pub trait SealedSideData {
+        const REJECT_RENEGOTIATION_TYPE: HandshakeType;
```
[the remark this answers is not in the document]
djc: Since `CommonState::process_main_protocol()`, which uses this value, also needs to instantiate a `Context` with a concrete `Data` type, I think it doesn't really make sense to use a `&dyn` for this.

djc:
Failures seem unrelated, fixed in #917.

rustls/src/conn.rs
```diff
@@ -894,23 +872,20 @@ impl CommonState {
...
-        if self.traffic && !self.is_tls13() {
-            let reject_ty = match self.is_client {
-                true => HandshakeType::HelloRequest,
-                false => HandshakeType::ClientHello,
-            };
-            if msg.is_handshake_type(reject_ty) {
-                self.send_warning_alert(AlertDescription::NoRenegotiation);
-                return Ok(state);
-            }
+        if self.traffic
```
briansmith, line 883: @djc, I looked at this in more detail and I think there's a more general issue that, if resolved, would make this moot. Simply, this code shouldn't be here.

First of all, we're probably TLS 1.3, and we shouldn't have this TLS-1.2-specific code in the being executed, even just the "self.traffic && !is_tls13()" part, in this common, somewhat hot, code path.  In general, we should be getting rid of any "if this is TLS 1.2 { ... }` blocks in the common path, both to make the correctness analysis easier, and for whatever performance improvement and code bloat elimination we can eek out.

I suggest we move these checks to the `impl State` implementations. I believe if we do that, then the "is client" or "is server" determination is determined by the context (client/tls12.rs or server/tls12.rs) in which the check is being done.

Also, I wonder if either the old or new code is correct. The TLS 1.2 spec says this about the client receiving HelloRequest:

>      The HelloRequest message MAY be sent by the server at any time.

and

>  This message will be ignored by the client if the client is currently negotiating a session.  This message MAY be ignored by the client if it does not wish to renegotiate a session, or the client may, if it wishes, respond with a no_renegotiation alert.

What happens if we're a TLS 1.2 client and the server sends a `HelloRequest` during the handshake? It seems like we're supposed to ignore it and not send an alert. It doesn't look like that's what we do, as I searched the code for `HelloRequest` and didn't see any "ignore the HelloRequest if we're handshaking" logic. Maybe it's intentional that we don't ignore renegotiation requests that are made before the handshake finishes, but if so we should document that. Otherwise, if we are going to implement "ignore the HelloRequest if we're handshaking" logic then we could insert the "send the warning alert if the handshake is done" aspect at that point. I hope that would be out of the commonly-executed code path.

briansmith: Strawman: Continue to reject renegotiation requests that occur during the handshake, and then put this logic in the `ExpectTraffic` implementations in each of client/tls12.rs and server/tls12.rs.

briansmith: See PR #943.

djc:
Scaled this back to avoid changing the renegotiation logic. I still think this is an improvement in its current form.

rustls/src/tls12/mod.rs
```diff
@@ -440,6 +440,8 @@ fn decode_ecdh_params_<T: Codec>(kx_params: &[u8]) -> Option<T> {
     }
 }
 
+pub(crate) const DOWNGRADE_SENTINEL: [u8; 8] = [0x44, 0x4f, 0x57, 0x4e, 0x47, 0x52, 0x44, 0x01];
+
```
briansmith, line 444: I verified the value is the same.

briansmith (approved):
Looks good to me!

PR #917, opened 2022-01-11
Fix new unreachable_pub warnings

djc:
`CipherSuiteCommon` was always intended to be public (and was already exposed through the `common` field).


djc:
This should fix the OSS Fuzz issue at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43592.

briansmith (approved):

PR #918, opened 2022-01-11
Replace redundant field with a method

djc:


djc:
Failures likely unrelated, see #917.

rustls/src/suites.rs
```diff
@@ -35,15 +35,24 @@ pub enum BulkAlgorithm {
     Chacha20Poly1305,
 }
 
+impl BulkAlgorithm {
+    #[inline]
+    pub(crate) fn algorithm(&self) -> &'static ring::aead::Algorithm {
```
ctz, line 40: Is the inline here enough to ensure that the limited client test doesn't contain AES? (We ought to automate that test really.)

PR #925, opened 2022-01-13
Split `KeyLogFile` into its own submodule & make it optional.

briansmith:
Split `KeyLogFile` into its own module to make it clearer that
`KeyLog` doesn't depend on `std::{env,fs,io,path,sync}`.

Add a feature flag to allow disabling the `KeyLogFile` feature. Make it
a default feature for API backward compatibility. In the next breaking
release, we should consider making it a non-default feature.

Use `git diff main:rustls/src/keylog.rs rustls/src/keylogfile.rs` to
verify that none of the `KeyLogFile` code has changed at all.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -18,7 +18,8 @@ sct = "0.7.0"
 webpki = { version = "0.22.0", features = ["alloc", "std"] }
 
 [features]
-default = ["logging", "tls12"]
+default = ["keylogfile", "logging", "tls12"]
+keylogfile = []
 logging = ["log"]
 dangerous_configuration = []
 quic = []
--- a/rustls/src/keylogfile.rs
+++ b/rustls/src/keylogfile.rs
@@ -1,0 +1,151 @@
+use crate::KeyLog;
+use std::env;
+use std::fs::{File, OpenOptions};
+use std::io;
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::Mutex;
+
+#[cfg(feature = "logging")]
+use crate::log::warn;
+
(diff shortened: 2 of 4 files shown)


briansmith:
Blocked by #926.

briansmith:
Closing this. I'll open a new PR that splits the refactoring (splitting files) from the rest of the changes.

rustls/src/keylogfile.rs
```diff
@@ -0,0 +1,151 @@
+use crate::KeyLog;
```
djc, line 1: I think the file should probably be called `keylog_file` or (better, for consistency) `key_log_file`.

rustls/src/keylogfile.rs
```diff
@@ -0,0 +1,151 @@
...
+use std::env;
+use std::fs::{File, OpenOptions};
+use std::io;
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::Mutex;
+
+#[cfg(feature = "logging")]
+use crate::log::warn;
```
djc, line 10: Ideally group `crate` imports together, and separate them with an empty line from `std` imports (common style is to have `std` last rather than first).

rustls/Cargo.toml
```diff
@@ -18,7 +18,8 @@ sct = "0.7.0"
 webpki = { version = "0.22.0", features = ["alloc", "std"] }
 
 [features]
-default = ["logging", "tls12"]
+default = ["keylogfile", "logging", "tls12"]
+keylogfile = []
```
djc, line 22: Should probably also use underscore(s) in the feature name.

Issue #919, opened 2022-01-11
Make SSLKEYLOGFILE support (`KeyLogFile`) sandboxing-friendly again


briansmith:
In https://github.com/rustls/rustls/pull/871#issuecomment-1007775869, I wrote about a potential regression in PR #871 for sandboxed applications:
* The application implements sandboxing. They want to have the sandbox block opening files, but they want to support this feature, so they configure the key log before enabling the sandbox, which they do before they make any TLS connections. Before, this strategy worked, AFAICT. Now, I think it doesn't, because the file won't be opened until at least the first TLS connection, at which time the sandbox will block the opening.
* For similar reasons, if the current working directory changes between enabling the key logging and the first attempt to log, then a different file will be opened and written to.
* If the application tries to unlink() after configuring the logging but before the first thing is logged.

At least we should handle the sandboxing case, by providing an API to force the file to be opened eagerly instead of lazily. This would also be usable to handle the other scenerios. Perhaps instead of having a default for "eager or lazy," we could pass in the choice of the application as an `enum` and document the considerations above in the doc comments for this API.

briansmith:
The documentation for `KeyLogFile::new()` says:
```
    /// Makes a new `KeyLogFile`.  The environment variable is
    /// inspected and the named file is opened during this call.
```
I'm not sure exactly what the intent was for guaranteeing that the file is opened during the call, but it was probably for similar reasons to the ones I mentioned above.

In any case, at a minimum we should update the docs to describe the new behavior, or consider reverting to the promised behavior.

PR #926, opened 2022-01-13
Revert "Defer opening key log file until writing to it"

briansmith:
This reverts commit 5e3de58514dba261930f578db870e9ba99ebcbf4 because it
isn't a hackward compatible change. The `KeyLogFile::new()` constructor
promises that the file is opened within `new()`.

Fixes https://github.com/rustls/rustls/issues/919.

### DIFF
--- a/rustls/src/keylog.rs
+++ b/rustls/src/keylog.rs
@@ -2,7 +2,7 @@ use std::env;
 use std::fs::{File, OpenOptions};
 use std::io;
 use std::io::Write;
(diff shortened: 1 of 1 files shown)


briansmith:
@djc @ctz Please take a look. I think that if we want lazy behavior, we should add a new constructor that implements the lazy behavior. But also I think if/when we're adding constructors to this, we should consider all the other issues with the feature and ensure that the new API addresses them.

PR #929, opened 2022-01-13
Split `KeyLogFile` into its own module.

briansmith:
Make it clear that `KeyLog` doesn't depend on `std::{env,fs,io,path,sync}`.

Use this to verify that no code changes were made:
```sh
git diff main:rustls/src/keylog.rs rustls/src/key_log.rs
git diff main:rustls/src/keylog.rs rustls/src/key_log_file.rs
```


briansmith:
@djc In the predecessor to this PR, you suggested the filename `key_log_file.rs` or `keylog_file.rs`. I don't have strong feelings but if it's `key_log_file.rs` then should I rename `keylog.rs` to `key_log.rs`, and if `keylog_file.rs` then are we have many strings "key log" that treat them as two words. So I guess I'll rename keylog.rs to key_log.rs

djc (approved):

djc:
I prefer `key_log_file` as well because it has its word boundaries consistent with the existing `KeyLogFile`.

PR #935, opened 2022-01-15
Tests: Split out dangerous_configuration tests from rest, client/server cert verifier tests in particular.

briansmith:
Make it easier to find the {Client,Server}CertVerifier tests. Split
them into separate files so that we can use `diff` to compare how we're
testing each.

This removes all of the `dangerous_configuration` tests from tests/api.rs.

Each of the mock Client/Server verifiers were eached move into the single test
suite that uses it.

Use these commands with a whitespace-smart diff tool to see that the tests
were not modified except for whitespace and formatting:
```
git difftool HEAD^1:rustls/tests/api.rs rustls/tests/server_cert_verifier.rs
git difftool HEAD^1:rustls/tests/api.rs rustls/tests/client_cert_verifier.rs
git difftool HEAD^1:rustls/tests/common/mod.rs rustls/tests/client_cert_verifier.rs
git difftool HEAD^1:rustls/tests/common/mod.rs rustls/tests/server_cert_verifier.rs
```


briansmith:
Note that right now `MockClientVerifier` is only used by `client_cert_verifier.rs` and `MockServerVerifier` is only used by `server_cert_verifier.rs`, so maybe it makes sense to just inline them into each of those files?

ctz (dismissed):

djc:
> Note that right now `MockClientVerifier` is only used by `client_cert_verifier.rs` and `MockServerVerifier` is only used by `server_cert_verifier.rs`, so maybe it makes sense to just inline them into each of those files?

Yes, I do think that would be preferable over having them in separate module.

One thing that could be split out of `api.rs` is all the tests that are guarded with `#[cfg(feature = "quic)]`. I think that's a decent chunk -- IIRC they're already in an inline module today.

briansmith:
I rebased this on `main` to resolve the merge conflicts. I merged the mocks into the test suites that use them, and I updated the PR description. In particular, see the new `git difftool` commands that are useful for reviewing this PR.

briansmith:
> One thing that could be split out of `api.rs` is all the tests that are guarded with `#[cfg(feature = "quic)]`. I think that's a decent chunk -- IIRC they're already in an inline module today.

I do think that seems like a good idea. I do have a branch that attempts that. But, I don't want to include those changes in this PR because I think we should discuss splitting common/mod.rs differently as part of that.

djc (approved):

PR #937, opened 2022-01-17
TLS 1.3 client: Stop holding onto entire `ClientConfig` after handshake.

briansmith:
TLS 1.2 client and server `ExpectTraffic`, and TLS 1.3 server `ExpectTraffic` already
don't store `ClientConfig` any longer. The TLS 1.3 client `ExpectTraffic` seems to be
the last instance.


briansmith:
I think this closes #665.

djc (approved):
Huh, not sure how I missed that.

PR #939, opened 2022-01-17
Replace `doc(hidden)` with `pub(crate)` where those semantics are intended.

briansmith:
I think these uses of `doc(hidden)` are trying to implement `pub(crate)`
semantics. Perhaps it was done this way a long time ago when `pub(crate)`
wasn't a think.

This change causes an `unreachable_pub` warning for `PlaintextSink`, so
make that `pub(crate)` too.

Some external users might be using these even though they probably shouldn't
be, so this is technically a breaking change.

There is one `TryFrom` implementation marked `#[doc(hidden)]` that is used by the test suite. Besides that, all the remaining `#[doc(hidden)]` uses are on deprecated type aliases.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -159,11 +159,10 @@ impl ClientConfig {
         }
     }
 
-    #[doc(hidden)]
     /// We support a given TLS version if it's quoted in the configured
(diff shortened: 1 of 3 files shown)


djc (approved):
LGTM; we should pull this into the next semver-breaking release.

PR #941, opened 2022-01-17
Correct comment typos: `RSASigningKey` -> `RsaSigningKey`.

briansmith:
The type was renamed but the references in documentation weren't updated.

### DIFF
--- a/rustls/src/sign.rs
+++ b/rustls/src/sign.rs
@@ -179,7 +179,7 @@ static ALL_RSA_SCHEMES: &[SignatureScheme] = &[
 ];
 
 impl RsaSigningKey {
-    /// Make a new `RSASigningKey` from a DER encoding, in either
+    /// Make a new `RsaSigningKey` from a DER encoding, in either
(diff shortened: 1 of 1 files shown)


rustls/src/sign.rs
```diff
@@ -252,7 +252,7 @@ impl Signer for RsaSigner {
 /// A SigningKey that uses exactly one TLS-level SignatureScheme
 /// and one ring-level signature::SigningAlgorithm.
 ///
-/// Compare this to RSASigningKey, which for a particular key is
+/// Compare this to RsaSigningKey, which for a particular key is
```
djc, line 255: Maybe [`RsaSigningKey`]?

djc (approved):

rustls/src/sign.rs
briansmith: I'd rather just keep this PR scoped to fixing the typos. If we want to add intra-crate links to the docs, which I think is a fine idea, we should do that more thoroughly in another PR.

PR #942, opened 2022-01-17
Remove `RsaSigningKey` from the public API documentation.

briansmith:
The API isn't part of the public API. It is `pub` only because some tests use it.

### DIFF
--- a/rustls/src/sign.rs
+++ b/rustls/src/sign.rs
@@ -164,7 +164,11 @@ pub fn any_eddsa_type(der: &key::PrivateKey) -> Result<Arc<dyn SigningKey>, Sign
     Err(SignError(()))
 }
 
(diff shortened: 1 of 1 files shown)


rustls/src/sign.rs
```diff
@@ -164,7 +164,11 @@ pub fn any_eddsa_type(der: &key::PrivateKey) -> Result<Arc<dyn SigningKey>, Sign
     Err(SignError(()))
 }
 
-/// A `SigningKey` for RSA-PKCS1 or RSA-PSS
+/// A `SigningKey` for RSA-PKCS1 or RSA-PSS.
+///
+/// This is used by the test suite, so it must be `pub`, but it isn't part of
+/// the public, stable, API.
+#[doc(hidden)]
```
djc, line 171: This feels like a crutch. Maybe we should just move those tests to be unit tests inside `sign`?

PR #943, opened 2022-01-18
Move TLS-1.2-specific renegotiation request logic to TLS 1.2 state machine.

briansmith:


briansmith:
All the tests pass, and the code coverage for conn.rs increased from 94% to 100%, which means that the code for dealing with renegotiation requests isn't covered by the test suite. Probably we need to enable some Bogo tests for renegotiation requests.

briansmith:
I updated this PR with an implementation that's closer to being complete, and which documents the issues I found with the original implementation that are carried over into this implementation.

I still need to figure out which Bogo tests, if any, exercise this, and/or if we need to write additional tests.

rustls/src/client/tls12.rs
```diff
@@ -1072,6 +1072,14 @@ impl State<ClientConnectionData> for ExpectTraffic {
...
+            MessagePayload::Handshake(HandshakeMessagePayload {
+                typ: HandshakeType::HelloRequest,
+                payload: _,
+            }) => {
+                // XXX(https://github.com/rustls/rustls/issues/952): DoS potential.
+                cx.common
+                    .send_no_renegotiation_warning_alert();
+            }
             _ => {
                 return Err(inappropriate_message(&m, &[ContentType::ApplicationData]));
```
briansmith, line 1084: This needs to `Err(inappropriate_handshake_message(payload, &[ContentType::Application, ContentType::Handshake], &[HandshakeType::HelloRequest])` (if PR #955 and/or PR #957 are merged), or the more verbose equivalent, if they aren't.

rustls/src/server/tls12.rs
```diff
@@ -900,6 +900,14 @@ impl State<ServerConnectionData> for ExpectTraffic {
...
+            MessagePayload::Handshake(HandshakeMessagePayload {
+                typ: HandshakeType::ClientHello,
+                payload: _,
+            }) => {
+                // TODO(https://github.com/rustls/rustls/issues/952): DoS potential.
+                cx.common
+                    .send_no_renegotiation_warning_alert();
+            }
             _ => {
                 return Err(inappropriate_message(&m, &[ContentType::ApplicationData]));
```
briansmith, line 912: Ditto.

Issue #850, opened 2021-11-13
Adapt server certificate verification interface so it can be used by platform verifiers


jsha:
In https://github.com/rustls/rustls-native-certs/issues/25 we discussed the desirability of using the platform verifier on macOS and Windows instead of trying to extract a set of roots, because distrust of certain roots is implemented in the platform verifier.

[The platform verifiers perform blocking I/O]( https://github.com/rustls/rustls-native-certs/issues/25#issuecomment-943358491), so the current ServerCertVerifier interface is not appropriate, since it is likely to be called on a thread doing async I/O, where we don't want to block the whole thread.

#787 is an example of solving a similar problem for the server side, but that has the advantage that accepting a ClientHello can happen before any TLS work is done. Server certificate verifier happens in the middle of the handshake.

Ralith:
> Server certificate verifier happens in the middle of the handshake.

I think the same principle stands: create an object for an outgoing connection, drive I/O on it until it reports certificate availability, then perform certificate verification and pass the result into a constructor for the next state.

ctz:
I think maybe we'll want to solve a few issues with the same/similar mechanism:

- asynchronous server certificate validation by client (the primary problem)
- asynchronous client certificate validation by server (#468)
- asynchronous handshake authentication signature by server (mention in passing in #390)
- asynchronous handshake authentication signature by client (ditto)
- asynchronous session storage (#264 -- though i'm not totally convinced, i would expect any scalable distributed session storage implementation would serve from a complete store in memory, with a weakly consistent cache coherence protocol underneath?)

jsha:
What about a callback-based approach?

```rust
pub trait ServerCertVerifier  {
    fn verify_server_cert(
        &self,
        end_entity: &Certificate,
        intermediates: &[Certificate],
        server_name: &rustls::ServerName,
        _scts: &mut dyn Iterator<Item = &[u8]>,
        ocsp_response: &[u8],
        _now: SystemTime,
        callback: impl FnOnce(Result<ServerCertVerified, Error>),
    )
```

When this is called, the user code could choose to do one of:

 - do the validation synchronously, call the callback immediately and return, or
 - do the validation on a threadpool and arrange to call the callback when done.

The ClientConnection would keep an internal bool, `awaiting_verification`. It would be set when the verification was requested and cleared when the callback is called. While the bit was set:

 - `process_new_packets` would continue succeeding but would not actually process additional packets
 - `Reader::read` would return WouldBlock.
[...]

PR #955, opened 2022-01-22
Reduce the amount of code inlined into callers of `require_handshake_msg[_move]`

briansmith:
Refactor `inappropriate[_handshake]_message` to work better with the pattern matching patterns used in `require_handshake_msg[_move]` and elsewhere in the codebase (including upcoming changes I have queued).

Use the new version of `inappropriate_handshake_message` to reduce the amonut of code that will be inlined into the "caller" of the `require_handshake_msg[_move]` macros.

This is a step towards removing mostly-redundant calls to `check_message` and `is_handshake_type()` (I have a branch that removes both).

### DIFF
--- a/rustls/src/check.rs
+++ b/rustls/src/check.rs
@@ -15,7 +15,10 @@ macro_rules! require_handshake_msg(
             payload: $payload_type(hm),
             ..
         }) => Ok(hm),
-        payload => Err($crate::check::inappropriate_handshake_message(payload, &[$handshake_type]))
+        payload => Err($crate::check::inappropriate_handshake_message(
+            payload,
+            &[$crate::msgs::enums::ContentType::Handshake],
+            &[$handshake_type]))
     }
   )
 );
@@ -30,7 +33,10 @@ macro_rules! require_handshake_msg_move(
             ..
         }) => Ok(hm),
         ref payload =>
-            Err($crate::check::inappropriate_handshake_message(payload, &[$handshake_type]))
+            Err($crate::check::inappropriate_handshake_message(
+                payload,
+                &[$crate::msgs::enums::ContentType::Handshake],
+                &[$handshake_type]))
     }
   )
 );
@@ -51,7 +57,11 @@ pub(crate) fn check_message(
 
     if let MessagePayload::Handshake(hsp) = &m.payload {
         if !handshake_types.is_empty() && !handshake_types.contains(&hsp.typ) {
-            return Err(inappropriate_handshake_message(&m.payload, handshake_types));
+            return Err(inappropriate_handshake_message(
+                &m.payload,
+                content_types,
+                handshake_types,
+            ));
         }
     }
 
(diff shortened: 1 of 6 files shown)


briansmith:
> This is a step towards removing mostly-redundant calls to `check_message` and `is_handshake_type()` (I have a branch that removes both).

Another reason I want to remove `check_message()`: I find the `handshake_types.is_empty()` logic within `check_message()` to be confusing. AFAICT, we never call `check_message` with an empty `handshake_types` when `content_types` contents `ContentType::Handshake` so the `is_empty()` check is moot?. Regardless, I'd rather just clean this up and also make it more efficient.

djc:
I'd be interested in reviewing the entire approach that also removes `check_message()` and `is_handshake_type()`. This mostly feels like moving complexity around at little benefit.

Here's an alternative approach that I'm contemplating: maybe we should stop with implementing the `State` trait for individual state types, and instead implement it on an enum roughly like this:

```rust
pub enum ClientState {
    ServerHello(Box<ExpectServerHello>),
    ServerHelloOrHelloRetryRequest(Box<ExpectServerHelloOrHelloRetryRequest>),
    Certificate(Box<tls12::ExpectCertificate>),
    CertificateStatusOrServerKx(Box<tls12::ExpectCertificateStatusOrServerKx>),
    CertificateStatus(Box<tls12::ExpectCertificateStatus>),
    ServerKx(Box<tls12::ExpectServerKx>),
    ServerDoneOrCertReq(Box<tls12::ExpectServerDoneOrCertReq>),
    CertificateRequest(Box<tls12::ExpectCertificateRequest>),
    ServerDone(Box<tls12::ExpectServerDone>),
    NewTicket(Box<tls12::ExpectNewTicket>),
    Ccs(Box<tls12::ExpectCcs>),
    Finished(Box<tls12::ExpectFinished>),
    Traffic(Box<tls12::ExpectTraffic>),
}
```

(We could potentially pull out separate 1.2 vs 1.3 enums, which might be more robust?)

This way, we can easily `match` `Message` to state in a single place and yield good error values otherwise. I think this will make stuff like #850 easier and make the code actually easier to understand while the generated code is more or less equivalent.

briansmith:
> I'd be interested in reviewing the entire approach that also removes `check_message()` and `is_handshake_type()`. 

The removal of check_message(), which includes all of these changes, is #955. The removal of `is_handshake_type()` becomes trivial after #955 and #943.

> This mostly feels like moving complexity around at little benefit.

I tried to enumerate the benefits in the PR summary in #955. The patterns using `check_message()` are error-prone and don't match the safer way we do them elsewhere, and they're not efficient.

> Here's an alternative approach that I'm contemplating

IMO, how the state machine is encoded (the way it is now vs. enum) is a minor issue. Ideally, we'd separate out the logic for the transitions (the `match`es, including the ones within `require_handshake_msg[_move]!`) from the rest of the logic, which would then make it easy to change.

> I think this will make stuff like #850 easier and make the code actually easier to understand while the generated code is more or less equivalent.

FWIW, I'm planning to implement a solution for #850 that doesn't require such changes. If you're also working on #850 then let's discuss that elsewhere.

briansmith:
@djc See also https://github.com/rustls/rustls/pull/943#pullrequestreview-860256183, which is what motivated me to take a break and work on this. I don't want to copy/paste the big chunk of code that'd otherwise be needed.

briansmith:
This is subsumed by #957 now. Cloising.

PR #956, opened 2022-01-22
Remove unused variant HandshakeType::EarlyData

djc:
Collateral damage from some investigation.


briansmith (approved):

briansmith:
The Bogo failure is
> unexpected failure: local error 'tls: server did not accept early data when expected', child error 'none', stdout:

ctz (approved):
I believe the test failure is flickeryness from our failing to adequately control time in the tests.

PR #957, opened 2022-01-23
Remove `check_message` and match all incoming messages uniformly

briansmith:
* From inspection, `check_message(m, &[..., ContentType::Handshake], &[])` returns `Ok(())`, which doesn't make sense. It seems like we never do that, but that shouldn't even be a possibility. Remove the function.
* In some cases, `check_message()` is doing redundant work pessimistically with the `match` or `if ... else` immediately after it. Eliminate the redundant checks.
* The logic for returning `InappropriateHandshakeMessage` vs `InappropriateMessage` is duplicated in many different states. (Some of this duplication is hard to see because it's hidden in the `require_handshake_msg[_move]` macros. Minimize the amount of duplication.
* Make `require_handshake_msg[_move]` work like the rest of the code.

The removal of `check_message()` (and soon `is_handshake_type()`) is a step towards eliminating the need for `HandshakeMessagePayload::typ`; we may even be able to eliminate the need for `HandshakeMessagePayload` completely.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1072,8 +1072,11 @@ impl State<ClientConnectionData> for ExpectTraffic {
             MessagePayload::ApplicationData(payload) => cx
                 .common
                 .take_received_plaintext(payload),
-            _ => {
-                return Err(inappropriate_message(&m, &[ContentType::ApplicationData]));
+            payload => {
+                return Err(inappropriate_message(
+                    &payload,
+                    &[ContentType::ApplicationData],
+                ));
             }
         }
         Ok(self)
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1,4 +1,4 @@
-use crate::check::{check_message, inappropriate_handshake_message, inappropriate_message};
+use crate::check::{inappropriate_handshake_message, inappropriate_message};
 use crate::conn::{CommonState, ConnectionRandoms, Side, State};
 use crate::error::Error;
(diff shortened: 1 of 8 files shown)


djc:
> The removal of `check_message()` (and soon `is_handshake_type()`) is a step towards eliminating the need for `HandshakeMessagePayload::typ`; we may even be able to eliminate the need for `HandshakeMessagePayload` completely.

Hah, this is also what I was looking at yesterday.

rustls/src/client/tls12.rs
```diff
@@ -1072,8 +1072,11 @@ impl State<ClientConnectionData> for ExpectTraffic {
             MessagePayload::ApplicationData(payload) => cx
                 .common
                 .take_received_plaintext(payload),
-            _ => {
-                return Err(inappropriate_message(&m, &[ContentType::ApplicationData]));
+            payload => {
```
djc, line 1096: Here, you used `payload => .. &payload ..` whereas in other commits it seems you've mostly used `ref payload`. Why the distinction? IMO the former style is currently more idiomatic.

djc (commented):
Okay, now that I've reviewed this in more detail this looks like an improvement.

rustls/src/client/tls12.rs
briansmith: > Here, you used payload => .. &payload .. whereas in other commits it seems you've mostly used ref payload. Why the distinction? IMO the former style is currently more idiomatic.

I pushed a new commit on top of the existing ones that makes them all consistent.

djc (approved):

Issue #951, opened 2022-01-21
`cargo check --all-features` is broken on non-nightly builds


briansmith:
We now have the `read_buf` feature that only works on Nightly Rust. `cargo check --all-features` now only works on Nightly Rust. This is unfortunate as it has broken some workflows for people using Stable Rust, including in particular the way my IDE is set up to check/build Rustls. Also this has lead to a lot of ugliness in the CI/CD pipelines.

I suggest we change the semantics of the `read_buf` feature as follows:
* When `read_buf` is enabled, build.rs will check if Rust is currently a Nightly build. If so, then set a configuration flag "nightly" that will be tested in the source code. This might require adding a `rustc_version` or similar dependency if the `read_buf` feature is enabled.
* Change `cfg(feature = "read_buf")` to `cfg(all("nightly", feature = "read_buf"))` in the source code and tests.
* Once the `ReadBuf` API becomes stable, remove all of this.

djc:
Yup, I also dislike the current setup. On the other hand, not a big fan of build scripts either. Seems like [#[rustversion::nightly]](https://docs.rs/rustversion/latest/rustversion/) could do the trick nicely, though?

ctz:
Have we done something terribly wrong, or is it really this hard to have a crate feature that depends on an unstable std feature? As it is I agree it is pretty terrible.

The rustversion thing looks pretty reasonable; I guess this route we can just get rid of the feature altogether as AIUI the purpose of it was to avoid making rustls nightly-only.

djc:
I don't think we've done anything wrong, so I guess it's really this hard.

briansmith:
> On the other hand, not a big fan of build scripts either. Seems like [#[rustversion::nightly]](https://docs.rs/rustversion/latest/rustversion/) could do the trick nicely, though?

`rustversion` has its own build script. I agree it's better for it to have one and not us, though.

briansmith:
This was fixed in PR #960.

PR #960, opened 2022-01-23
Make `cargo {check,build,test} --all-features` work on Rust Stable.

briansmith:
Make the `read_buf` feature do nothing on non-Nightly Rust.

This makes it easier for people not depending on that feature to
build/test Rustls, as demonstrated by the CI/CD changes here.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -10,7 +10,11 @@ homepage = "https://github.com/rustls/rustls"
 repository = "https://github.com/rustls/rustls"
 categories = ["network-programming", "cryptography"]
 autobenches = false
+build = "build.rs"
 resolver = "2"
+
+[build-dependencies]
+rustc_version = { version = "0.4.0", optional = true }
 
 [dependencies]
 log = { version = "0.4.4", optional = true }
@@ -24,7 +28,7 @@ logging = ["log"]
 dangerous_configuration = []
 quic = []
 tls12 = []
-read_buf = []
+read_buf = ["rustc_version"]
 
 [dev-dependencies]
 env_logger = "0.9.0"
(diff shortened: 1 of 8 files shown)


briansmith:
To verify this change, noticed that
https://github.com/briansmith/rustls/runs/4914185840?check_suite_focus=true doesn't mention the `test_read_buf_in_new_state` test (because it isn't build/run for stable Rust) but https://github.com/briansmith/rustls/runs/4914185876?check_suite_focus=true#step:5:270 shows the test passes on Rust Nightly.

rustls/Cargo.toml
```diff
@@ -10,8 +10,12 @@ homepage = "https://github.com/rustls/rustls"
 repository = "https://github.com/rustls/rustls"
 categories = ["network-programming", "cryptography"]
 autobenches = false
+build = "build.rs"
```
djc, line 13: Why not use the rustversion attribute directly as discussed in #951?

briansmith: > Why not use the rustversion attribute directly as discussed in #951?

This is explained in build.rs.

rustls/build.rs
```diff
@@ -0,0 +1,16 @@
+/// This build script allows us to enable the `read_buf` language feature only
+/// for Rust Nightly.
+///
+/// We cannot use `rustversion` to avoid this build script because
+/// `rustversion` doesn't provide a mechanism for conditionally enabling
+/// language features like `read_buf`.
```
djc, line 6: We can just stick `#[rustversion(nightly)]` on each instance separately and dump the feature, as @ctz suggested? Seems simpler to me.

briansmith:
I'm going to close this in favor of #962, which switches to `rustversion` and also documents the ugliness that is (still) required with the build script.

PR #961, opened 2022-01-23
bogo: Remove redundant specification of features for bogo_shim.

briansmith:
The features will be enabled automatically since bogo_shim lists them as
required features in Cargo.toml.

### DIFF
--- a/bogo/runme
+++ b/bogo/runme
@@ -6,7 +6,7 @@
 set -xe
 
 if [ "x$USE_EXISTING_BOGO_SHIM" = "x" ] ; then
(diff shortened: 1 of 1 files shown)


djc:
I don't think that's how required features work. Required features make it so that the binary target will only be built if those features are enabled, but it doesn't mean that those features get enabled implicitly when building that target.

https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field

briansmith:
Yes, you are right., Let me try this a different way then. Sorry for the noise.

PR #962, opened 2022-01-23
Automatically enable `read_buf` functionality on Rust Nightly (only)

briansmith:
The functionality is now automatically enabled for Rust Nightly (only).

### DIFF
--- a/rustls/build.rs
+++ b/rustls/build.rs
@@ -3,10 +3,9 @@
 ///
 /// See the comment in lib.rs to understand why we need this.
 
-#[cfg_attr(feature = "read_buf", rustversion::not(nightly))]
+#[rustversion::not(nightly)]
 fn main() {}
 
-#[cfg(feature = "read_buf")]
 #[rustversion::nightly]
 fn main() {
     println!("cargo:rustc-cfg=read_buf");
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -14,7 +14,7 @@ build = "build.rs"
 resolver = "2"
 
 [build-dependencies]
-rustversion = { version = "1.0.6", optional = true }
+rustversion = { version = "1.0.6" }
 
 [dependencies]
 log = { version = "0.4.4", optional = true }
@@ -28,7 +28,6 @@ logging = ["log"]
 dangerous_configuration = []
 quic = []
 tls12 = []
-read_buf = ["rustversion"]
 
 [dev-dependencies]
 env_logger = "0.9.0"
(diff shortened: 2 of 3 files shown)


djc (approved):

djc:
Sucks about that inner attr thing.

briansmith:
@djc Please take another look at this as I had to update it since rustfmt complained and (depending which version you reviewed) to add more documentation.

rustls/build.rs
```diff
@@ -0,0 +1,12 @@
+/// This build script allows us to enable the `read_buf` language feature only
+/// for Rust Nightly.
+///
+/// See the comment in lib.rs to understand why we need this.
+
+#[rustversion::not(nightly)]
+fn main() {}
+
+#[rustversion::nightly]
```
djc, line 9: Would it work to stick the attribute on the `println!()` statement directly, avoiding the repetition here?

djc (dismissed):

rustls/build.rs
briansmith: > Would it work to stick the attribute on the `println!()` statement directly, avoiding the repetition here?

No, I tried:
```
error[E0658]: custom attributes cannot be applied to statements
  --> rustls\build.rs:10:5
   |
10 |     #[rustversion::nightly]
   |     ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #54727 <https://github.com/rust-lang/rust/issues/54727> for more information
   = help: add `#![feature(proc_macro_hygiene)]` to the crate attributes to enable

For more information about this error, try `rustc --explain E0658`.
```

briansmith:
Maybe we should hold off on doing this until we verify that we run the entire test suite (including Bogo) for the Rust Nightly build as well as the Rust Stable build. Otherwise, we're automatically enabling functionality that isn't fully tested in CI, I think.

briansmith:
> Maybe we should hold off on doing this until we verify that we run the entire test suite (including Bogo) for the Rust Nightly build as well as the Rust Stable build. Otherwise, we're automatically enabling functionality that isn't fully tested in CI, I think.

@djc, Sorry, but I think until the testing situation is clarified, we should split out "enable from default" like this PR does from the "get --all-features working on stable." I put up PR #965 to do the latter and I've made this (#962) a draft PR that I'll rebase on top of testing improvements.

briansmith:
Blocked by #966.

briansmith:
I'm going to close this unless/until I can resolve the testing stuff.

PR #967, opened 2022-01-23
TLS 1.3: Improve error reporting for handshake message received during early data.

### LINKED ISSUE #958: When a handshake message is received during early data, the error should mention `EndOfEarlyData`
See https://github.com/rustls/rustls/blob/8c203e150023b14b5ee8ee6a3d6ff99ae0b90940/rustls/src/server/tls13.rs#L1050-L1073

If the message we received was a Handshake message of type other than `EndOfEarlyData` then we don't return `InappropriateHandshakeMessage` and the error doesn't mention that the only handshake message allowed is `EndOfEarlyData`. When making a PR to fix this, I decided to first refactor the error reporting code, PR #955, to make the fix less messy.

briansmith:
Fixes #958.

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -1065,9 +1065,10 @@ impl State<ServerConnectionData> for ExpectEarlyData {
                     send_ticket: self.send_ticket,
                 }))
             }
-            payload => Err(inappropriate_message(
+            payload => Err(inappropriate_handshake_message(
                 &payload,
                 &[ContentType::ApplicationData, ContentType::Handshake],
+                &[HandshakeType::EndOfEarlyData],
             )),
         }
     }


rustls/src/server/tls13.rs
```diff
@@ -1065,9 +1065,10 @@ impl State<ServerConnectionData> for ExpectEarlyData {
                     send_ticket: self.send_ticket,
                 }))
             }
-            payload => Err(inappropriate_message(
+            payload => Err(inappropriate_handshake_message(
```
djc, line 1068: This would call `inappropriate_handshake_message()` even if the `MessagePayload` is an `Alert` or `ChangeCipherSpec`, which seems a little confusing.

briansmith: > This would call `inappropriate_handshake_message()` even if the `MessagePayload` is an `Alert` or `ChangeCipherSpec`, which seems confusing.

That's already the case throughout the codebase. The new `inappropriate_handshake_message` handles those kinds of messages appropriately.

djc (approved):

PR #968, opened 2022-01-24
DRY in TLS 1.2 and TLS 1.3 handshakes

briansmith:
In each place where we emit a particular kind of message (e.g. `Certificate`, `CertificateVerify`) for a protocol version, for both client and server, factor out that commonality into a shared function.

In each place where we verify a particular kind of message (e.g. `Finished`), factor out that common verification logic into a shared function.

This will make it easier to audit and test the code.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -14,6 +14,7 @@ use crate::msgs::handshake::{DigitallySignedStruct, ServerECDHParams};
 use crate::msgs::handshake::{HandshakeMessagePayload, HandshakePayload, NewSessionTicketPayload};
 use crate::msgs::message::{Message, MessagePayload};
 use crate::msgs::persist;
+use crate::sign::Signer;
 use crate::suites::SupportedCipherSuite;
 use crate::ticketer::TimeBase;
 use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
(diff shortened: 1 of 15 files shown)


rustls/src/conn.rs
```diff
@@ -899,19 +897,6 @@ impl CommonState {
...
-            let reject_ty = match self.side {
-                Side::Client => HandshakeType::HelloRequest,
-                Side::Server => HandshakeType::ClientHello,
-            };
-            if msg.is_handshake_type(reject_ty) {
-                self.send_warning_alert(AlertDescription::NoRenegotiation);
-                return Ok(state);
-            }
-        }
-
```
briansmith, line 914: These changes are part of #943. Ignore it.

rustls/src/lib.rs
```diff
@@ -312,37 +312,39 @@ mod log {
...
+mod key;
+mod key_log;
+mod key_log_file;
+mod kx;
 mod limited_cache;
 mod rand;
 mod record_layer;
 mod stream;
+mod suites;
+mod ticketer;
```
briansmith, line 338: Apparently my IDE decided to automatically alphabetize these. Will deal with that separately.

rustls/src/tls13/hs.rs
```diff
@@ -0,0 +1,125 @@
...
+    context: Option<Vec<u8>>,
+    cert_chain: Vec<Certificate>,
+    ocsp_response: Option<&[u8]>,
+    sct_list: Option<&[u8]>,
+    signer: Option<&dyn Signer>,
+    context_string_with_0: verify::ContextString,
+) -> Result<(), Error> {
+    let signer = signer.ok_or_else(|| {
+        common.send_fatal_alert(AlertDescription::HandshakeFailure);
+        Error::PeerIncompatibleError("no overlapping sigschemes".to_string())
```
briansmith, line 26: At least one test is failing because of this, as this is a change in behavior for the TLS 1.3 client, but I think that test may be relying on buggy behavior. I filed PR #969 about it.

briansmith (commented):
I'm planning to split this up into several PRs but I'm sharing the whole thing as one draft PR so we can see how much commonality can be factored out without much thinking/effort. There is probably more commonality that can be factored out but that would require more work.

rustls/src/client/tls12.rs
```diff
@@ -1036,36 +951,22 @@ impl ExpectFinished {
...
-                    Error::DecryptError
-                })
-                .map(|_| verify::FinishedMessageVerified::assertion())?;
+        let fin_verified = tls12::hs::handle_finished(
+            &st.secrets,
+            &st.transcript,
+            cx.common,
+            tls12::SERVER_FINISHED,
+            &m,
+        )?;
```
briansmith, line 961: I was surprised this PR is currently only net -132 lines of code (+416/−548). It seems like the styling of function calls like this makes measuring in LoC quite misleading.

gustav3d:
Hows it going ?.

djc:
> Hows it going ?.

There has been no progress (as you can see). Why do you ask?

PR #971, opened 2022-01-24
TLS client: Make the logic for sending Certificate/CertificateVerify clearer.

briansmith:
When the resolver returns a `CertifiedKey`, we give it the acceptable signature schemes. The expectation is that the resultant `CeritfiedKey` won't return `None` when asked for a signer given the same list of signature schemes. But, what if the CertifiedKey *does* return `None` when asked for the signer? (Maybe the resolver has some interior mutability of its state that would cause it to return inconsistent results.)

Handle that case by acting as if the resolver returned `None` when asked for the CertifiedKey.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -14,6 +14,7 @@ use crate::msgs::handshake::{DigitallySignedStruct, ServerECDHParams};
 use crate::msgs::handshake::{HandshakeMessagePayload, HandshakePayload, NewSessionTicketPayload};
 use crate::msgs::message::{Message, MessagePayload};
 use crate::msgs::persist;
+use crate::sign::Signer;
 use crate::suites::SupportedCipherSuite;
 use crate::ticketer::TimeBase;
 use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
@@ -462,21 +463,12 @@ fn emit_clientkx(transcript: &mut HandshakeHash, common: &mut CommonState, pubke
 
 fn emit_certverify(
     transcript: &mut HandshakeHash,
-    client_auth: &mut ClientAuthDetails,
+    signer: &dyn Signer,
     common: &mut CommonState,
 ) -> Result<(), Error> {
-    let (signer, message) = match (client_auth.signer.take(), transcript.take_handshake_buf()) {
-        (Some(signer), Some(msg)) => (signer, msg),
-        (None, _) => {
-            trace!("Not sending CertificateVerify, no key");
-            transcript.abandon_client_auth();
-            return Ok(());
-        }
-        (_, None) => {
-            trace!("Not sending CertificateVerify, no transcript");
-            return Ok(());
-        }
-    };
+    let message = transcript
+        .take_handshake_buf()
+        .ok_or_else(|| Error::General("Expected transcript".to_owned()))?;
 
     let scheme = signer.scheme();
     let sig = signer.sign(&message)?;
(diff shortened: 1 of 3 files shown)


rustls/src/client/tls12.rs
```diff
@@ -462,21 +463,12 @@ fn emit_clientkx(transcript: &mut HandshakeHash, common: &mut CommonState, pubke
...
-            return Ok(());
-        }
-        (_, None) => {
-            trace!("Not sending CertificateVerify, no transcript");
-            return Ok(());
-        }
-    };
+    let message = transcript
+        .take_handshake_buf()
+        .ok_or_else(|| Error::General("Expected transcript".to_owned()))?;
```
briansmith, line 471: This is a change in behavior in the case where we weren't able to get the transcript. I think that's an impossible case? Regardless, the old logic seems wrong.

rustls/src/client/tls12.rs
```diff
@@ -642,24 +632,31 @@ impl State<ClientConnectionData> for ExpectCertificateRequest {
...
+                        auth_context_tls13: None,
+                    }
+                })
+            })
+            .unwrap_or_else(|| {
+                debug!("Client auth requested but no cert/sigscheme available");
+                ClientAuthDetails::Empty {
+                    auth_context_tls13: None,
+                }
+            });
```
djc, line 659: How about moving this logic into a `ClientAuthDetails` constructor?

rustls/src/client/tls12.rs
```diff
@@ -780,11 +777,14 @@ impl State<ClientConnectionData> for ExpectServerDone {
         cx.common.peer_certificates = Some(st.server_cert.cert_chain);
 
         // 4.
-        if let Some(client_auth) = &mut st.client_auth {
-            if let Some(cert_key) = &client_auth.certkey {
-                emit_certificate(&mut st.transcript, cert_key.cert.clone(), cx.common);
-            } else {
-                emit_certificate(&mut st.transcript, Vec::new(), cx.common);
+        if let Some(client_auth) = &st.client_auth {
+            match client_auth {
```
djc, line 781: I'd extract the meat of this so it becomes 

```rust
let mut cert = match client_auth {
   ...
};
emit_certificate(&mut st.transcript, cert, cx.common);
```

rustls/src/client/tls13.rs
```diff
@@ -580,23 +581,30 @@ impl State<ClientConnectionData> for ExpectCertificateRequest {
...
+                            auth_context_tls13: Some(certreq.context.0.clone()),
+                        }
+                    })
+            })
+            .unwrap_or_else(|| {
+                debug!("Client auth requested but no cert selected");
+                ClientAuthDetails::Empty {
+                    auth_context_tls13: Some(certreq.context.0.clone()),
+                }
+            });
```
djc, line 607: So this could likely be shared with the 1.2 side if it's moved into a constructor?

rustls/src/client/tls13.rs
```diff
@@ -904,9 +902,27 @@ impl State<ClientConnectionData> for ExpectFinished {
...
+                } => {
+                    emit_certificate_tls13(
+                        &mut st.transcript,
+                        Some(&certkey),
+                        auth_context,
+                        cx.common,
+                    );
+                    emit_certverify_tls13(&mut st.transcript, signer.as_ref(), cx.common)?;
+                }
+            }
```
djc, line 925: I'd probably rewrite this so that there is a single `emit_certificate_tls13()` call.

djc (commented):
I like the direction here, but would be good to look at the deduplication of `ClientAuthDetails` construction (and assorted style suggestions).

rustls/src/client/tls13.rs
briansmith: I made the other changes you requested, but I don't think this change would make it clearer, so I didn't make this one.

djc: Fair enough, it's definitely less of a win here.

djc (approved):

PR #974, opened 2022-01-25
Server: Only pass compatible signature schemes to the cert resolver.

briansmith:
Prior to 0.20.0 we filtered the list of signature schemes passed to the
cert resolver to ones that are compatible with the cipher suites configured
in the `ServerConfig`. Restore that behavior so that the cert resolver get
misled about what signature schemes it has to choose from.

### DIFF
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -268,9 +268,12 @@ impl ExpectClientHello {
     }
 
     /// Continues handling of a `ClientHello` message once config and certificate are available.
+    ///
+    /// `client_sig_schemes` has been filtered at least according to the client's supported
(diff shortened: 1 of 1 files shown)


djc (approved):

cpu:
It looks to me like the goal of this draft PR was achieved in https://github.com/rustls/rustls/pull/1053/files

The `retain` filtering with `suites::compatible_sigscheme_for_suites` that is proposed in this PR is now present in `hs.rs` prior to choosing a certificate:
https://github.com/rustls/rustls/blob/5105cc29acd86ddcbcebda6d41e4b518e4e1dc2c/rustls/src/server/hs.rs#L337-L340

I'm going to close this draft, but if I'm missed some nuance between the two approaches I'm happy to revisit.

PR #976, opened 2022-01-25
Server: Change `Acceptor` to only allow one attempt per ClientHello.

briansmith:
When we encounter an error processing the client hello, don't try again.
Trying again generally won't do any good. This matches `Future::poll`
semantics (Futures shouldn't be polled after yielding `Ready`).

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -473,28 +473,26 @@ impl Acceptor {
 
     /// Check if a `ClientHello` message has been received.
     ///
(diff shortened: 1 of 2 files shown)


djc (approved):
Yup, this probably makes more sense.

djc:
@ctz any opinions on this one? I think we should just merge it.

gustav3d:
yes, it's the natural way to design it imo.

ctz (approved):
agree

PR #982, opened 2022-01-27
Remove all client auth tests that use OpenSSL

briansmith:
As far as I can tell, there is only one OpenSSL-based test that isn't redundant with the Bogo tests, which is the scenerio `client_auth_requested_but_unsupported` that handles the case where client auth is requested but not required. The existing tests were all redundant with Bogo's, as noted in the individual commit messages, except one. PR #983 adds a replacement for that one.


briansmith:
I forgot to mention: Each commit message mentions which BoGo tests cover the tests removed by that commit.

ctz (dismissed):

briansmith:
Note: the code coverage report above is stale and doesn't represent this change. Rather, it was measuring code coverage for an earlier version that had the changes in PR #983 in it. This didn't change the code coverage as the removed tests aren't measured by code coverage (and rightly so).

@ctz Your review got canceled when I rebased this to resolve the merge conflict so I need another approval.

djc (approved):

PR #994, opened 2022-01-30
Move release notes to seperate document.

TimonPost:
Release notes are getting large, perhaps it's better to store them in a separate file for those who want to check it out. Maybe it's less relevant for the average user? Also if this is done, it seems better to merge OLD_CHANGES into this document as well. 

Feel free to close this PR if you don't deem it necessary.


djc:
I think it could make sense to move the pre-0.20 changes over to OLDCHANGES. I don't think this is helpful in its current shape.

ctz:
Thanks! I merged this as ab060f8c with some minor changes: keeping the future plans and a handful of past release notes in README.md, plus a link to RELEASE_NOTES.md

PR #997, opened 2022-02-04
Add support for EC private keys

farcaller:
This adds support for [EC PRIVATE KEY](https://datatracker.ietf.org/doc/html/rfc5915).

Arguably this should be within ring but frankly I'm afraid to touch it as I couldn't even find the github tag for the current version. I suggest the support lands in rustls first so that its clients (e.g. kube-rs) can move on with the required support and then I'll see if this could go into ring.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -16,7 +16,6 @@ log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 sct = "0.7.0"
 webpki = { version = "0.22.0", features = ["alloc", "std"] }
-sec1 = "0.2.1"
 
 [features]
 default = ["logging", "tls12"]
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -16,6 +16,7 @@ log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 sct = "0.7.0"
 webpki = { version = "0.22.0", features = ["alloc", "std"] }
+sec1 = "0.2.1"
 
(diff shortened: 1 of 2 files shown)


farcaller:
The relevant pemfile change: https://github.com/rustls/pemfile/pull/5

Darkspirit:
A dependency to sec1 can't be added. Duplicate of https://github.com/rustls/rustls/issues/332#issuecomment-720708668.

farcaller:
Any idea where the source of ring 0.16.20 is? I'm not opposed to patching ring but I can't progress as ring from master won't work with rustls.

farcaller:
I've made a diff against ring's commit 9cc0d45f4d8521f467bb3a621e74b1535e118188, sent https://github.com/briansmith/ring/pull/1456 and updated this PR to match.

ctz:
Note #998 achieves this another (much less graceful) way. This PR can remain for doing it in the more reasonable way: so please add a `git revert 5288fcf` to the top.

farcaller:
Thanks for merging in a workaround!

cpu:
>  This PR can remain for doing it in the more reasonable way: so please add a git revert 5288fcf to the top.

I think we should close this PR for now since the feature was accomplished with a workaround. The upstream ring project hasn't been active in some time, making it seem unlikely that a better approach will be provided there anytime soon. Even if it were, this branch has a number of conflicts that would need to be addressed. I suspect it would be easier to start a fresh branch inspired by this one when the time comes.

Thanks!

Issue #970, opened 2022-01-24
Do we wrongly accept malformed/oversized ChangeCipherSpec messages in TLS 1.3


briansmith:
I was looking into refactoring this code and I noticed there might be a bug, if my understanding is correct:
https://github.com/rustls/rustls/blob/8d76aa67a9d1fef86d1acf413d1f78610ad49ba0/rustls/src/conn.rs#L540-L565

Before we do `return Ok(state);`, don't we need to make sure that the `ChangeCipherSpec` message is exactly one byte that has value 0? I think with this logic we'd accept a ChangeCipherSpec message of any non-zero length (zero-length records are rejected when a `OpaqueMessage` is constructed) and with any values in its payload.

We should make sure we have tests for these scenerios.

ctz:
Yeah that is a bug. We should ideally be looking for a `ChangeCipherSpecPayload` there, and not look at the `typ` field, ever.

PR #1000, opened 2022-02-06
Reject malformed change_cipher_specs in TLS1.3

ctz:
Aims to address #970

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -380,6 +380,13 @@ impl ConnectionRandoms {
 
 // --- Common (to client and server) connection functions ---
 
+fn is_valid_ccs(msg: &OpaqueMessage) -> bool {
+    // nb. this is prior to the record layer, so is unencrypted. see
+    // third paragraph of section 5 in RFC8446.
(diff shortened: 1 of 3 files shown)


rustls/src/conn.rs
```diff
@@ -537,23 +537,34 @@ impl<Data> ConnectionCommon<Data> {
         self.state = Ok(new);
     }
 
+    fn is_valid_ccs(msg: &OpaqueMessage) -> bool {
```
djc, line 540: If its not going to take any form of `self`, should it just be a free function?

rustls/src/conn.rs
```diff
@@ -537,23 +537,34 @@ impl<Data> ConnectionCommon<Data> {
...
+        // nb. this is prior to the record layer, so is unencrypted. see
+        // third paragraph of section 5 in RFC8446.
+        msg.typ == ContentType::ChangeCipherSpec
+            && msg.payload.0.len() == 1
+            && msg.payload.0[0] == 0x01
```
djc, line 545: Alternatively, `&msg.payload == &[0]` seems more concise?

djc (approved):
LGTM.

PR #1008, opened 2022-02-18
Only set QUIC state when the connection is QUIC.

briansmith:


Ralith (approved):

briansmith:
I rebased this on top of #1009 since it can't pass CI unless/until #1009 (or an alternative) is merged.

djc (approved):

ctz:
Merged as 5ed9ebcc

Issue #1012, opened 2022-02-19
Log spam of "Received a ServerHelloDone handshake message while expecting [CertificateRequest]" in 0.20.4


jbg:
[This log message](https://github.com/rustls/rustls/blob/main/rustls/src/check.rs#L66) is triggering for every HTTPS request we make to a certain upstream API, as of 0.20.4.

The message, logged at WARN level, is:

> Received a ServerHelloDone handshake message while expecting [CertificateRequest]

It seems erroneous given that it's valid to receive no `CertificateRequest` before `ServerHelloDone`.

ctz:
Thanks for the report, this is indeed spurious logging and a regression in 0.20.3 and 0.20.4.

TjeuKayim:
The URL https://httpbin.org/ is affected by this bug, while the URL https://example.com/ is not.

Minimal code to reproduce this with the reqwest crate:

```rs
reqwest::get("https://httpbin.org/").await?;
```

Logs the following warning:

```log
[WARN  rustls::check] Received a ServerHelloDone handshake message while expecting [CertificateRequest]
```

djc:
Should be fixed in #1013.

PR #1013, opened 2022-02-20
Correct spurious warn-level logging during successful TLS1.2 handshakes

ctz:
This addresses #1012. There's aa test which fails prior to the fix commit, and also verifies the effect of the logging crate feature.

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -1,4 +1,5 @@
 //! Assorted public API tests.
+use std::cell::RefCell;
 use std::convert::TryFrom;
 #[cfg(feature = "tls12")]
 use std::convert::TryInto;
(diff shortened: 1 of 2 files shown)


rustls/tests/api.rs
```diff
@@ -3986,3 +3989,100 @@ fn test_acceptor() {
...
+struct CountingLogger;
+
+static LOGGER: CountingLogger = CountingLogger;
+
+impl CountingLogger {
+    fn install() {
+        COUNTS.with(|c| {
+            c.borrow_mut().reset();
+        });
+        log::set_logger(&LOGGER).unwrap();
```
briansmith, line 4035: This is a bit confusing because the logger can only be set once per process. IMO it would be clearer to have `reset()` that resets the counts and `install()` that does the rest. It doesn't matter now but it will probably matter later.

briansmith (approved):

briansmith (approved):

djc:
I'll try to get a release out soon, I'll see if we can squeeze #1031 in.

lucacasonato:
That'd be great. Thanks @djc!

PR #1016, opened 2022-03-03
Attach location context to corrupt message errors

complexspaces:
This PR introduces source-line tracking to the `CorruptMessagePayload` variant of `Error`. I'm proposing these changes because, so far, its been impossible to track down the source of TLS handshake errors occurring when our client application tries to make a TLS connection with a middleware proxy in the path.

For example, this is all that's present in logs based off `rustls`'s error `Display`: `... received corrupt message of type Handshake`. There are ~7 places this could be returned on the `main` branch today, however. Most of the time there is no way to reproduce the exact environment a user is running the app in, so manually inspecting where it fails live is not an option.

To make the source problems easier to find, this PR moves the `CorruptMessagePayload` variant to contain an opaque struct that carries additional location information, and updates all the relevant error construction sites to use the new method instead.

### DIFF
--- a/rustls-mio/tests/badssl.rs
+++ b/rustls-mio/tests/badssl.rs
@@ -6,6 +6,7 @@ mod common;
 
 mod online {
     use super::common::TlsClient;
+    use super::common::DEBUG_LOCATION_REPLACEMENT;
 
     fn connect(hostname: &str) -> TlsClient {
         TlsClient::new(hostname)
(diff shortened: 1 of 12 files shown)


djc:
This seems like an interesting technique. I think the changes could be a bit cleaner by maybe adding a method like `Error::corrupt_message(kind: ContentType)` which could avoid the repetition of the `CorruptMessagePayload` token.

complexspaces:
Thanks for the feedback, @djc. That idea cleaned up the diff quite a lot.

djc (approved):

djc:
Thank you for working on this change!

Issue #1014, opened 2022-02-21
The sni must be case-insensitive.


silence-coding:
The sni method must be case-insensitive. The add method here is not converted to lowercase.

https://github.com/rustls/rustls/blob/b3e354b3dc0de23b13ec1979206066b4f2914395/rustls/src/server/handy.rs#L157

https://datatracker.ietf.org/doc/html/rfc3546#section-3.1

briansmith:
@silence-coding Are you saying that, when the user gives us a SNI value with uppercase letters, e.g. "EXAMPLE.org", that we don't normalize it to lowercase when we put it in the ClientHello message? I agree that we should, although the spec doesn't require or even recommend it.

Specifically, RFC6066 says about case and comparison only:
> DNS hostnames are case-insensitive.  The
 algorithm to compare hostnames is described in RFC5890 Section 2.3.2.4.

silence-coding:
@briansmith No, I'm talking about the problem that ctz has fixed
https://github.com/rustls/rustls/pull/1017

PR #1017, opened 2022-03-13
Fix server::ResolvesServerCertUsingSni case sensitivity

ctz:
Extend testing to show this working.

Addresses #1014

### DIFF
--- a/rustls/src/server/handy.rs
+++ b/rustls/src/server/handy.rs
@@ -156,11 +156,13 @@ impl ResolvesServerCertUsingSni {
     /// chain is syntactically faulty.
     pub fn add(&mut self, name: &str, ck: sign::CertifiedKey) -> Result<(), Error> {
         let checked_name = webpki::DnsNameRef::try_from_ascii_str(name)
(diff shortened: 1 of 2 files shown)


rustls/src/server/handy.rs
```diff
@@ -156,11 +156,13 @@ impl ResolvesServerCertUsingSni {
     /// chain is syntactically faulty.
     pub fn add(&mut self, name: &str, ck: sign::CertifiedKey) -> Result<(), Error> {
         let checked_name = webpki::DnsNameRef::try_from_ascii_str(name)
-            .map_err(|_| Error::General("Bad DNS name".into()))?;
+            .map_err(|_| Error::General("Bad DNS name".into()))?
+            .to_owned();
 
-        ck.cross_check_end_entity_cert(Some(checked_name))?;
+        ck.cross_check_end_entity_cert(Some(checked_name.as_ref()))?;
```
briansmith, line 162: I think this should be:
```
let lowercase_name = checked_name.as_ref();
ck.cross_check_end_entity_cert(lowercase_name);
let as_str: &str = lowercase_name.as_ref().into();
```

PR #1021, opened 2022-03-18
Expose cipher suites via `ClientHello`

### LINKED ISSUE #1019: Determining which certificates are compatible with a `ClientHello`
At Fly.io, we generate both a RSA and ECDSA certificate for every hostname we serve.

The way we did this was by using the `tls-parser` crate and peeking at the `TcpStream` to determine the list of supported client ciphers as well as the ALPN, servername and TLS protocol version.

With the ciphers, we knew which kinds of certificates we should fetch from our encrypted store.

```rust
static SUPPORTED_ANY_CIPHERSUITES: &[&str] = &[
     "TLS_CHACHA20_POLY1305_SHA256",
     "TLS_AES_128_GCM_SHA256",
     "TLS_AES_256_GCM_SHA38"
];

static SUPPORTED_ECDSA_CIPHERSUITES: &[&str] = &[
     "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
     "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
     "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
];

static SUPPORTED_RSA_CIPHERSUITES: &[&str] = &[
     "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
     "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
     "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
];
```

(Note: this was possibly flawed in its own way)

If we matched any of the ClientHello ciphers to the `SUPPORTED_ANY_CIPHERSUITES` list, we knew we could provide either RSA or ECDSA (we prioritize the latter). If not, we looked for ciphers in the other slices.

Now, with the new design, we're able to get more reliable information by starting the handshake and fetching the `ClientHello` (we're using tokio-rustls's `LazyAcceptor`).

Unfortunately, we don't have enough information to make a decision on which certificate we should fetch from our encrypted store.

The `ClientHello` doesn't provide a list of client ciphers we can match on. Instead, we need to rely on signature schemes.

As far as I can tell, this is not enough information for determining if we should fetch an RSA or ECDSA key / certificate. I suppose it might be possible by intersecting lists of known RSA, ECDSA, ANY signature schemes, but I don't know how much we can rely on this.

Would it be desirable to also expose the list of client ciphers in the `ClientHello`? I wouldn't mind knowing the protocol version as well.

Any other way to help determine with certainty which type of certificate key we should resolve in our `ResolvesServerCert` implementation would be helpful.

jeromegn:
Fixes #1019 

Exposes the client hello's payload's cipher suites via `ClientHello`. I've created functions to access both "raw" cipher suites and "parsed" `SupportedCipherSuite`s. I'm not overly attached to the "raw" cipher suites, but I figured they might be useful and it doesn't cost anything to expose them.

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -11,13 +11,13 @@ use crate::msgs::enums::ProtocolVersion;
 use crate::msgs::enums::SignatureScheme;
 use crate::msgs::handshake::{ClientHelloPayload, ServerExtension};
 use crate::msgs::message::Message;
-use crate::sign;
 use crate::suites::SupportedCipherSuite;
 use crate::vecbuf::ChunkVecBuffer;
 use crate::verify;
 use crate::KeyLog;
 #[cfg(feature = "quic")]
 use crate::{conn::Protocol, quic};
+use crate::{sign, CipherSuite};
 
 use super::hs;
 
@@ -112,6 +112,7 @@ pub struct ClientHello<'a> {
     server_name: &'a Option<webpki::DnsName>,
     signature_schemes: &'a [SignatureScheme],
     alpn: Option<&'a Vec<PayloadU8>>,
+    cipher_suites: &'a [CipherSuite],
 }
 
 impl<'a> ClientHello<'a> {
@@ -120,6 +121,7 @@ impl<'a> ClientHello<'a> {
         server_name: &'a Option<webpki::DnsName>,
         signature_schemes: &'a [SignatureScheme],
         alpn: Option<&'a Vec<PayloadU8>>,
+        cipher_suites: &'a [CipherSuite],
     ) -> Self {
         trace!("sni {:?}", server_name);
         trace!("sig schemes {:?}", signature_schemes);
@@ -129,6 +131,7 @@ impl<'a> ClientHello<'a> {
             server_name,
             signature_schemes,
             alpn,
+            cipher_suites,
         }
     }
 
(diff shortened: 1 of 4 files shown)


djc (commented):
The core change here looks good to me. Would you mind pulling out the clippy change into a separate PR?

rustls/src/server/server_conn.rs
```diff
@@ -158,6 +161,22 @@ impl<'a> ClientHello<'a> {
                 .map(|proto| proto.0.as_slice())
         })
     }
+
+    pub fn cipher_suites(&self) -> &[CipherSuite] {
+        self.cipher_suites
+    }
+
+    pub fn supported_cipher_suites(&self) -> Vec<SupportedCipherSuite> {
```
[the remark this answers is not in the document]
jeromegn: Yes, happy to do it that way if it helps move things forward and fits better how it's done in `rustls`. Going to make the change.

rustls/tests/api.rs
```diff
@@ -657,6 +662,16 @@ impl ResolvesServerCert for ServerCheckCertResolve {
             }
         }
 
+        if let Some(expected_cipher_suites) = &self.expected_cipher_suites {
+            if expected_cipher_suites != client_hello.cipher_suites() {
```
djc, line 666: Is there any reason not to use `assert_eq!()` here?

jeromegn: Not really, I copied what was already there. I can change it and update others that fit the pattern as well.

djc (approved):
This looks good to me.

ctz (approved):

PR #1022, opened 2022-03-18
Improve MessageDeframer mem usage

### LINKED ISSUE #794: Improve the memory footprint of a large number of mostly idle connections

larsfillmore:
Fixes #794

This PR removes `MessageDeframer::buf: Box<[u8; OpaqueMessage::MAX_WIRE_SIZE]>` and replaces it with:
* A thread-local shared buffer
* `MessageDeframer::partial_msg: Option<Box<[u8]>>` that keeps yet unparsed bytes between read attempts, if any

### CPU usage

~~According to `cargo run --release --example bench`, bulk transfer speed has decreased by a few percent.~~

~~On rustls/main:~~
<details>
  <summary><del>Click to expand</del></summary>
  
```
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	1827.21	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1904.42	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	2038.66	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1956.95	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	2028.71	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1978.17	MB/s
```
</details>

~~On this branch:~~
<details>
  <summary><del>Click to expand</del></summary>
  
```
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	1780.28	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1818.80	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	2014.43	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1851.29	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	1956.25	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1811.65	MB/s
```
</details>

**UPD**: The bulk transfer speed has been restored.

### MEM usage

Mem usage has massively improved for a chat-like server that handles a large number of long-running client connections that are mostly idle or send/receive short messages.

There have been three test runs: compiled with rustls/main, compiled with this branch, no TLS (just raw TCP). The C10k memory difference:
```
tls_main - no_tls = 262M
tls_this - no_tls = 45M
```

~~What bothers me is that the per-client difference of 21.7K (i.e. 26.2K - 4.5K) is larger than the optimised-out `OpaqueMessage::MAX_WIRE_SIZE` = 18K. I don't have an explanation for this yet.~~

**UPD**: The strangeness was due to jemalloc.

### DIFF
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -1,4 +1,3 @@
-use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::io;
 
@@ -17,9 +16,103 @@ pub struct MessageDeframer {
     /// the deframer cannot recover.
     pub desynced: bool,
 
-    /// Bytes left after the previous attempt to decode the read buffer.
-    /// A non-empty slice.
-    partial_msg: Option<Box<[u8]>>,
+    /// A fixed-size buffer containing the currently-accumulating TLS message.
+    buf: Option<Buf>,
+}
+
+impl Default for MessageDeframer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl MessageDeframer {
+    pub fn new() -> Self {
+        Self {
+            frames: VecDeque::new(),
+            desynced: false,
+            buf: None,
+        }
+    }
+
+    /// Read some bytes from `rd`, and add them to our internal
+    /// buffer.  If this means our internal buffer contains
+    /// full messages, decode them all.
+    pub fn read(&mut self, rd: &mut dyn io::Read) -> io::Result<usize> {
+        let buf = Self::get_or_init(&mut self.buf);
+
+        // Try to do the largest reads possible.  Note that if
+        // we get a message with a length field out of range here,
+        // we do a zero length read.  That looks like an EOF to
+        // the next layer up, which is fine.
+        debug_assert!(buf.used <= OpaqueMessage::MAX_WIRE_SIZE);
+        let frames = &mut self.frames;
+        let desynced = &mut self.desynced;
+        let res = rd
+            .read(&mut buf.buf[buf.used..])
+            .map(|new_bytes| {
+                buf.used += new_bytes;
+
+                loop {
+                    match buf.try_deframe_one() {
+                        BufferContents::Invalid => {
+                            *desynced = true;
+                            break;
+                        }
+                        BufferContents::Valid(m) => {
+                            frames.push_back(m);
+                            continue;
+                        }
+                        BufferContents::Partial => break,
+                    }
+                }
+
+                new_bytes
+            });
+
+        #[cfg(feature = "shared_bufs")]
+        if buf.used == 0 {
+            BUF_POOL.with(|v| *v.borrow_mut() = self.buf.take());
+        }
+
+        res
+    }
+
+    /// Returns true if we have messages for the caller
+    /// to process, either whole messages in our output
+    /// queue or partial messages in our buffer.
+    pub fn has_pending(&self) -> bool {
+        !self.frames.is_empty()
+            || self
+                .buf
+                .as_ref()
+                .map_or(false, |b| b.used != 0)
+    }
+
+    fn get_or_init(buf: &mut Option<Buf>) -> &mut Buf {
+        if let Some(b) = buf {
+            #[cfg(feature = "shared_bufs")]
+            debug_assert!(b.used > 0);
+            
+            return b;
+        }
+
+        #[cfg(feature = "shared_bufs")]
+        if let Some(mut b) = BUF_POOL.with(|v| v.borrow_mut().take()) {
+            debug_assert_eq!(b.used, 0);
+            b.used = 0;
+            return buf.insert(b);
+        }
+
+        buf.insert(Buf::new())
+    }
+}
+
+#[cfg(feature = "shared_bufs")]
+thread_local! {
+    /// A shared pool containing at most one item. A `MessageDeframer` takes the buffer from it
+    /// when it needs to read data and returns the buffer back afterwards.
+    static BUF_POOL: std::cell::RefCell<Option<Buf>> = std::cell::RefCell::new(None)
 }
 
 enum BufferContents {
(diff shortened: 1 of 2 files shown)


rustls/src/msgs/deframer.rs
```diff
@@ -42,62 +40,83 @@ impl Default for MessageDeframer {
     }
 }
 
+thread_local! {
+    static BUF: RefCell<Box<[u8; OpaqueMessage::MAX_WIRE_SIZE]>> =
```
SSebo, line 44: Suggest `SHARED_BUF` or add some comment. I read the first time did not realized that it is shard :)

larsfillmore: Done

paolobarbolini:
I've sent you a PR @larsfillmore (https://github.com/larsfillmore/rustls/pull/1) which tries to remove the extra memcpy and head allocation added by this PR.

<details>
<summary>Benchmarks</summary>

## main (0a5c9776345f29b79fc9fed2024b131fd70abfb3)

```
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	send	2314.77	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	1771.53	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	send	2238.39	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1730.98	MB/s
```

## unpatched (a121bfbf4efbc136edfd5f0eb66fad250edaacea)

```
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	send	2051.88	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	1680.92	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	send	2231.21	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1666.70	MB/s
```

## patched (a121bfbf4efbc136edfd5f0eb66fad250edaacea)

```
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	send	2298.95	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:default	recv	1783.68	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	send	2234.02	MB/s
bulk	SupportedProtocolVersion { version: TLSv1_2, is_private: () }	TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256	max_fragment_size:10000	recv	1731.13	MB/s
```
</details>

quininer:
Using `thread_local` affects the portability of rustls, causing rustls to not work on platforms that do not support `thread_local`. maybe it would be nice to allow customization of this behavior.

like

```rust
struct BufferPoolConfig {
    alloc: fn() -> Box<[u8; MAX_WIRE_SIZE]>,
    free: fn(Box<[u8; MAX_WIRE_SIZE]>)
}

thread_local!{
    static ref BUFFER_POOL: RefCell<Option<Box<[u8; MAX_WIRE_SIZE]>>> = RefCell::new(None);
}

static THREAD_LOCAL_BUFFER_POOL: BufferPoolConfig = BufferPoolConfig {
    alloc: || BUFFER_POOL.borrow_mut().take().unwrap_or_else(|| Box::new([0; MAX_WIRE_SIZE])),
    free: |buf| *BUFFER_POOL.borrow_mut() = Some(buf)
}

ServerConfig::builder()
    .with_buffer_pool(&THREAD_LOCAL_BUFFER_POOL)

// ...
```

djc:
I think the complexity this introduces is not really worth it. I think we should instead work on adding API that relies on caller-allocated memory only. I started out in this direction (again) and made some progress, but would be happy if anyone wants to help. My first step is 9f1503fd5944e3f62f58808ced057843f13ffc0e, which makes it so that the `ConnectionCommon` can statelessly deframe messages (this commit also shrinks the buffer size in the `ReceiveBuffer` but that is a tricky trade-off which we may not end up wanting).

Next, we'd want to pull in something like https://github.com/djc/rustls/commit/43f4a9d09c2c15024a1035ed85c627395bb54891, which would enable borrowing the `OpaqueMessage`s from the given buffer, decrypting them in place if necessary.

After that, the next step would be to handshake joining in place: if there are enough plain handshake messages in the incoming byte stream to gather one full handshake messages, we'll want to use `copy_within()` them so that they form a continuous slice, then build a `PlainMessage` from there that we can turn into a handshake message.

Once we no longer need a separate handshake joiner, we can move `ConnectionCommon::process_msg()` and some form of `process_new_handshake_messages()` into `CommonState`, which will allow us to pass a slice referencing a buffer held within `ConnectionCommon` (that is, from the `ReceiveBuffer`) into something much like `ConnectionCommon::process_new_packets()` which would yield a `usize` in addition to the `IoState` telling the caller how much of the passed in buffer has been processed.

I might keep chipping away at this, but any help would be appreciated. I'd be happy to answer further questions and/or review your code, in any state.

larsfillmore:
I've added a commit that fixes the performance loss in the way suggested by @paolobarbolini (by taking the buffer from the pool instead of allocating memory for a partial message).

It also addresses @quininer's concern by introducing a new non-default feature named "shared_bufs" (the name is discussable). When this feature is switched-off everything works almost exactly as in `rustls/main`, and hence existing code won't see any difference. Switching this feature on enables thread-local buffers.

@djc Your plan is longer-term and will affect API. In the meantime we could use these more local changes that might be OK even for a patch-release.

djc:
My plan can be implemented with additive API only. I'm personally not motivated to merge changes like these when we have a better plan in place.

PR #1025, opened 2022-03-27
quic: initial salt for version draft-23

catenacyber:
We would like to use rustls in Suricata to "decrypt" quic initial packets

And we would like to support more versions like 0xfaceb002

cf https://github.com/OISF/suricata/pull/7174

Is there another way to define ourselves the salt (than calling `rustls::quic::Keys::initial`) ?

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -603,6 +603,11 @@ pub trait ClientQuicExt {
         let ext = match quic_version {
             quic::Version::V1Draft => ClientExtension::TransportParametersDraft(params),
             quic::Version::V1 => ClientExtension::TransportParameters(params),
+            _ => {
+                return Err(Error::General(
+                    "QUIC unsupported version".into(),
+                ));
+            }
         };
 
         ClientConnection::new_inner(config, name, vec![ext], Protocol::Quic)
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -778,6 +778,11 @@ pub trait ServerQuicExt {
         let ext = match quic_version {
             quic::Version::V1Draft => ServerExtension::TransportParametersDraft(params),
             quic::Version::V1 => ServerExtension::TransportParameters(params),
+            _ => {
+                return Err(Error::General(
+                    "QUIC unsupported version".into(),
+                ));
+            }
         };
         let mut new = ServerConnection::from_config(config, vec![ext])?;
         new.inner.common_state.protocol = Protocol::Quic;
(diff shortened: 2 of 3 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -603,6 +603,11 @@ pub trait ClientQuicExt {
         let ext = match quic_version {
             quic::Version::V1Draft => ClientExtension::TransportParametersDraft(params),
             quic::Version::V1 => ClientExtension::TransportParameters(params),
+            _ => {
+                return Err(Error::General(
+                    "QUIC unsupported version".into(),
+                ));
+            }
```
Ralith, line 610: Should draft 23 not also use `TransportParametersDraft`?

rustls/src/server/server_conn.rs
```diff
@@ -778,6 +778,11 @@ pub trait ServerQuicExt {
         let ext = match quic_version {
             quic::Version::V1Draft => ServerExtension::TransportParametersDraft(params),
             quic::Version::V1 => ServerExtension::TransportParameters(params),
+            _ => {
+                return Err(Error::General(
+                    "QUIC unsupported version".into(),
+                ));
+            }
```
Ralith, line 785: As above

Ralith (commented):
What do you need support for long-obsolete drafts for? Is there software in the wild still using these?

rustls/src/client/client_conn.rs
catenacyber: I do not know : I just introduced the minimum set of changes to get my use case to work.
In facts, I would rather not need to change rustls and be able to use `rustls::quic::Keys::initial` with my own defined `Version`/original_salt

What is the best way to do so ?

catenacyber:
> What do you need support for long-obsolete drafts for?

For Suricata to use rustls to decrypt the first quic packet with SNI and that kind of stuff...
cf https://github.com/OISF/suricata/pull/7174

> Is there software in the wild still using these?

Looks like it... (2022 pcap)

victorjulien:
> > What do you need support for long-obsolete drafts for?
> 
> For Suricata to use rustls to decrypt the first quic packet with SNI and that kind of stuff... cf [OISF/suricata#7174](https://github.com/OISF/suricata/pull/7174)
> 
> > Is there software in the wild still using these?
> 
> Looks like it... (2022 pcap)

These versions are still in use today. Just confirmed by opening the instagram app on a fully updated iPhone. This is `faceb002`.

My chromecast devices use `ff00001d`.

djc:
Is this `0xfaceb002` version documented anywhere? I think we're usually pretty unenthusiastic about supporting anything that's not standardized. The experimental Facebook version is definitely not standardized. Older drafts are maybe slightly closer to the scope. What would an alternative API look like that would allow you to decrypt initial packets with deviant salts? Maybe it could be something that could live in the unstable `internal` API?

Ralith:
Perhaps a `Keys::initial_from_salt` interface could be added? That would be better than adding half-implemented cases to the `Version` enum.

djc:
`Keys::initial_from_salt()` seems like a more reasonable option to me, indeed.

Issue #1027, opened 2022-04-07
unable to do session resumption with api.devicecheck.apple.com


zh-jq:
from tokio-rs/tls/issues/103

there are two streams in the following pcap files, the first to appleid.apple.com, the second to api.devicecheck.apple.com, both use session_ticket extension.
[out.pcap.gz](https://github.com/rustls/rustls/files/8443223/out.pcap.gz)

The second failed while the first succeeded. The main difference seems to be, in stream one, packet 6 contains **Server Hello + Change Cipher Spec + Encrypted Handshake Message**, but in stream two, packet 23 only contains **Server Hello + Change Cipher Spec**, the missing **Encrypted Handshake Message** is later in packet 27, which seems to be ignored by rustls as it reply alert message and close the connection before packet 27.

tokio-rustls version 0.23.2, rustls version 0.20.4

The error returned by rustls is CorruptMessagePayload(Handshake), I'm trying to test it locally with master version of rustls to get more info.

djc:
@complexspaces is this the same thing you were seeing before? Did you make any progress since #1016?

zh-jq:
I can confirm that code in #1028 fixed this.

complexspaces:
@djc I believe that this is unrelated. We heard back from someone encountering the issue and the issue seems to be coming from [this section](https://github.com/rustls/rustls/blob/main/rustls/src/conn.rs#L603-L608).

djc:
@complexspaces probably a good idea to file a new issue about that?

complexspaces:
@djc Yep, that sounds good. I just wanted to confirm it was unrelated to this one. I've opened https://github.com/rustls/rustls/issues/1030 for everything else.

ctz:
Thanks for the good bug report. Fixed in a8d361c which will be released in 0.20.5 shortly.

Issue #603, opened 2021-03-30
Always compute transcripts from the bytes on the wire, not from re-encoded versions


briansmith:
In https://github.com/ctz/rustls/issues/601#issuecomment-810356284, @djc wrote:

> We update the transcript from a parsed and reencoded representation of incoming messages, and we normalize server names to lowercase as soon as they are parsed.

This seems *very* wrong to me. Any weakness in Rustls's decoding and/or encoding would cause a bug in our transcript hashing. Problems that would otherwise not be security-sensitive would be elevated to security-critical if they could be exploited. Issue #601 shows a benign example of this: The normalization that DNSName does absolutely shouldn't have affected transcript hashing.

Instead, hashing of the transcript should be done before parsing, and PR #602 should be reverted.

ctz:
> The normalization that DNSName does absolutely shouldn't have affected transcript hashing.

Disagree.

The architecture here is that the `msgs` module is capable of round-tripping any TLS message: that's why (for example) unknown enum values are preserved, unsupported extensions are parsed and kept, etc. Almost all other parts of the crate rely on this invariant -- not just the transcript.

The purpose of the transcript input being post-parsing, is *precisely because* we don't want to trust the encoding/decoding code to be correct. This is, really, just the ["Horton principle"](https://en.wikipedia.org/wiki/Horton_Principle).

Consider a decoding bug which changes the semantics of an incoming message.  If the input to the broken decoding code was immediately fed into the transcript, the handshake would complete successfully and this would likely cause a bug elsewhere, later, likely with security impact. If the broken decoding instead was fed into the transcript, this means the incorrect behaviour is immediately and safely detected, and the handshake fails.

TLS, as a protocol, _must_ be able to deal safely with the input the transcript not matching between two peers. This is the same mechanism that protects the handshake from adversarial editing.

briansmith:
> The architecture here is that the `msgs` module is capable of round-tripping any TLS message: that's why (for example) unknown enum values are preserved, unsupported extensions are parsed and kept, etc. Almost all other parts of the crate rely on this invariant -- not just the transcript.

OK, thanks for explaining that. Now everything is much clearer to me.

[...]

PR #1028, opened 2022-04-07
Deal with ServerHellos without any extensions

ctz:
This aims to address #1027, where the server gives us a ServerHello without any extensions when resuming. We already dealt with ClientHellos without any extensions in the correct way.


djc (approved):
LGTM, though I'm not sure what's up with the fuzzing job. Looks like more round-tripping issues?

ctz:
> Looks like more round-tripping issues?

Yep!

rustls/src/msgs/handshake.rs
```diff
@@ -890,12 +890,20 @@ impl Codec for ClientHelloPayload {
 }
 
 impl ClientHelloPayload {
+    pub fn get_extensions(&self) -> &[ClientExtension] {
+        if let Some(ref exts) = self.extensions {
```
djc, line 894: I'd write this as `self.extensions.as_deref().unwrap_or(&[])`.

(And `get_()` prefixes are discouraged in the API naming guidelines, which should still work since methods and fields don't share namespaces.)

rustls/src/msgs/handshake.rs
```diff
@@ -1029,9 +1037,11 @@ impl ClientHelloPayload {
     }
 
     pub fn set_psk_binder(&mut self, binder: impl Into<Vec<u8>>) {
-        let last_extension = self.extensions.last_mut();
-        if let Some(ClientExtension::PresharedKey(ref mut offer)) = last_extension {
-            offer.binders[0] = PresharedKeyBinder::new(binder.into());
+        if let Some(ref mut extensions) = &mut self.extensions {
```
djc, line 1040: I'd probably write these as `match` expressions with an early return clause in the `None arm.

rustls/src/msgs/handshake.rs
```diff
@@ -1232,9 +1242,9 @@ impl Codec for ServerHelloPayload {
          *  the ServerHello."
          */
         let extensions = if r.any_left() {
-            codec::read_vec_u16::<ServerExtension>(r)?
+            Some(codec::read_vec_u16::<ServerExtension>(r)?)
```
djc, line 1245: Consider writing this as a `match r.any_left() {}` expression?

djc (commented):
Some style suggestions.

rustls/src/msgs/handshake.rs
```diff
@@ -851,7 +851,7 @@ pub struct ClientHelloPayload {
     pub session_id: SessionID,
     pub cipher_suites: Vec<CipherSuite>,
     pub compression_methods: Vec<Compression>,
-    pub extensions: Vec<ClientExtension>,
+    pub extensions: Option<Vec<ClientExtension>>,
```
briansmith, line 854: This is a weird type to use because it allows encoding "No extensions" as `None` separately from "zero extensions" as `Some(vec![])`. Perhaps the invariant that the array is never empty is enforced somewhere when this is `Some` but IMO it would just be clearer to have one way of encoding "no extensions" as "zero extensions"; i.e. without the `Option` wrapper.

rustls/src/msgs/handshake.rs
```diff
@@ -890,12 +890,20 @@ impl Codec for ClientHelloPayload {
 }
 
 impl ClientHelloPayload {
+    pub fn get_extensions(&self) -> &[ClientExtension] {
```
briansmith, line 893: Notice how the return value encodes "no extensions" as an empty slice instead of as an `Option`.

rustls/src/msgs/handshake.rs
```diff
@@ -1029,9 +1037,11 @@ impl ClientHelloPayload {
     }
 
     pub fn set_psk_binder(&mut self, binder: impl Into<Vec<u8>>) {
-        let last_extension = self.extensions.last_mut();
-        if let Some(ClientExtension::PresharedKey(ref mut offer)) = last_extension {
-            offer.binders[0] = PresharedKeyBinder::new(binder.into());
+        if let Some(ref mut extensions) = &mut self.extensions {
+            let last_extension = extensions.last_mut();
```
briansmith, line 1041: Notice that the `Option` doesn't help this function; it makes it more complicated.

rustls/src/msgs/handshake.rs
```diff
@@ -1214,15 +1224,28 @@ impl Codec for ServerHelloPayload {
         self.session_id.encode(bytes);
         self.cipher_suite.encode(bytes);
         self.compression_method.encode(bytes);
-        codec::encode_vec_u16(bytes, &self.extensions);
+
+        if let Some(ref exts) = self.extensions {
+            codec::encode_vec_u16(bytes, exts);
```
briansmith, line 1229: Does this do the right thing when `exts` is empty? Or is this assuming that `exts` will never be empty if it is `None`?

briansmith (commented):
One way to DRY this code would be to define:

```
struct OptionalExtensions<T>(Vec<T>));
impl Codec for OptionalExtensions {
     // When decoding, accept no input as meaning "zero extensions"
     // When encoding, wrote no output when there are "zero extensions"
}
```
This would allow the ClientHello and ServerHello to share code.

rustls/src/msgs/handshake.rs
ctz: The problem here is arising from the roundtripping requirement discussed in #603 -- "no extensions" and "zero extensions" have different and valid wire encodings (b"" and b"\x00\x00" respectively), so this needs to be expressed somewhere in the in-memory representation of ClientHelloPayload and ServerHelloPayload

I think I'm talking my way towards fixing #603 rather than doing this.

PR #1031, opened 2022-04-24
[tls handshaked] fix:Interrupted system call (os error 4)

randsoy:
Dynamically load the FFI Library using flutter in Android system.  `Connection Failed: tls connection init failed: Interrupted system call (os error 4)`(ureq error);

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -488,14 +488,14 @@ impl<Data> ConnectionCommon<Data> {
                         rdlen += n;
                         Some(n)
                     }
-                    Err(ref err) if err.kind() == io::ErrorKind::Interrupted => None,// nothing to do
+                    Err(ref err) if err.kind() == io::ErrorKind::Interrupted => None, // nothing to do
                     Err(err) => return Err(err),
                 };
                 if read_size.is_some() {
                     break;
                 }
             }
-            
+
             match self.process_new_packets() {
                 Ok(_) => {}
                 Err(e) => {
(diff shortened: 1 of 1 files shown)


rustls/src/conn.rs
```diff
@@ -478,11 +478,27 @@ impl<Data> ConnectionCommon<Data> {
...
+                debug!("read_tls");
+                match self.read_tls(io) {
+                    Ok(0) => {
+                        eof = true;
+                        debug!("eof");
+                        break;
+                    }
+                    Ok(n) => {
+                        rdlen += n;
+                        debug!("read {} bytes", n);
```
djc, line 491: I think we should remove the `debug!()` call here, as that seems pretty noisy and not that useful?

rustls/src/conn.rs
```diff
@@ -478,11 +478,27 @@ impl<Data> ConnectionCommon<Data> {
...
+                        break;
+                    }
+                    Err(ref err) if err.kind() == io::ErrorKind::Interrupted => {
+                        // nothing to do
+                        debug!("complete_io: read interrupted");
+                    }
+                    Err(err) => {
```
djc, line 498: Style nit: could use `return` as an expression here (omitting brances and semi-colon here).

djc (commented):
Thanks for your contribution, I think this makes sense.

Could you change the first line of the commit message to describe the intended change, please? Also commented on some nits below.

djc:
We'd like to publish 0.20.5 soon. Would be great if you can fix up the formatting issues!

randsoy:
> We'd like to publish 0.20.5 soon. Would be great if you can fix up the formatting issues!

` cargo fmt`

djc (approved):

Issue #184, opened 2018-07-10
Can't connect to a server presenting a certificate with an IP address in its common name


jbg:
`ClientSession` requires a `DNSNameRef` for the `hostname` argument, but a `DNSNameRef` cannot be constructed from an IP address.

In some situations (in my case, connecting to a PostgreSQL server) the server presents a valid certificate, signed by a trusted CA, for a specific IP address. As far as I can tell this is a valid use of TLS certificates, but it doesn't seem possible to support it with rustls at present.

briansmith:
The underlying webpki library doesn't support IP addresses yet; see https://github.com/briansmith/webpki/issues/54.

qiujiangkun:
I have my own 'fork' for supporting ip address. However, it was a huge poor refactorization for both rustls and webpki. I even did not actually fork, but I simply copy the content of both the crates. I would not recommend anyone use my version.
Here's a simple example for you reference.
```rust
use crate::ChannelBuilder;
use crate::*;
use rustls::ClientConfig;
use rustls::ClientSession;
use serde::Deserialize;
use serde::Serialize;
use std::io::Read;
use std::io::Write;
use std::net::Shutdown;
use std::net::SocketAddr;
use std::net::TcpStream;
use std::net::ToSocketAddrs;
use std::sync::Arc;
use std::time::Duration;
use utils::now;
use utils::UnsafeArc;

#[derive(Serialize, Deserialize, Clone, Debug)]
pub struct TlsChannelBuilder {
    pub domain:       String,
    pub port:         u32,
    pub ip:           String,
    pub verification: bool,
}

impl ChannelBuilder for TlsChannelBuilder {
    type Channel = TlsChannel;

    fn connect_delayed(&self) -> Result<Self::Channel, Error> {
        Ok(TlsChannel {
            stream:    UnsafeArc::empty(),
            builder:   self.clone(),
            connected: false,
        })
    }
}
pub struct TlsChannel {
    stream:    UnsafeArc<rustls::StreamOwned<ClientSession, TcpStream>>,
    builder:   TlsChannelBuilder,
    connected: bool,
}

impl TlsChannel {
    pub fn connect(address: &str, port: u32) -> Result<TlsChannel, Error> {
        if utils::is_valid_ip(address) {
            Self::connect_no_verification(address, port)
        } else {
            Self::connect_domain(address, port)
        }
    }

    /// Connect to a domain but with pre-fetched ip address to work around DNS
[...]

PR #1032, opened 2022-05-03
Add support for IP addresses for dangerous clients only

jhartzell42:
This is an attempt to partially address #184 in some circumstances only: It would only apply to clients that also use the `dangerous_configuration` feature.

Currently, `rustls` doesn't support IP addresses because `webpki` does not support IP addresses. However, `webpki` is only relevant in the certificate verification phase, so if certificate verification can be suppressed or modified (as it can through the `dangerous_configuration` feature), it makes sense that IP addresses should be allowed if an alternative certificate verification system is used (including `NoVerifier`).

This code allows parsing IP addresses. For code that uses the default certificate verification, IP addresses simply fail further along in the pipeline in certificate validation. However, for other code, it simply works.

I have tested this with `reqwest` using its `danger_accept_invalid_certs` method, and connecting to a raw IP address now works in that context.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -23,6 +23,7 @@ use crate::quic;
 use std::convert::TryFrom;
 use std::error::Error as StdError;
 use std::marker::PhantomData;
+use std::net::IpAddr;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 use std::{fmt, io, mem};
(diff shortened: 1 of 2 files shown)


djc:
@ctz this seems reasonable to me. What do you think?

djc (approved):

jhartzell42:
@djc had to make a fix bc I accidentally left in a trace. Could you please re-run the workflow?

jhartzell42:
Alright, @djc I think it might be ready unless you have any comments :-)

ctz (approved):
lgtm!

PR #1042, opened 2022-05-18
Prepare 0.20.6

djc:
See discussion in #1041.

- [x] Changelog updated
- [x] Checked for outdated dependencies
- [x] Passed tests locally with `--all-features`
- [x] Bumped version number in `Cargo.toml`
- [x] Passed `cargo publish --dry-run`

Issue #1041, opened 2022-05-17
Breaking change from 0.20.4 to 0.20.5


jgallagher:
Hi! We ran into a failure when updating rustls from 0.20.4 to 0.20.5 in https://github.com/oxidecomputer/dropshot/pull/347. It looks like the commit that introduced the change is c42c53e13dfc54495cbb62577f6bb58eddf5ff8a; specifically [this bit](https://github.com/rustls/rustls/commit/c42c53e13dfc54495cbb62577f6bb58eddf5ff8a#diff-1a5dae8e5448dde6cce460d2ddb8eb3fd6dd66e917c755a078bf9c0175e4bd76R52) that changes the type of a case in the public `Error` enum.

We have some tests that construct specific error cases, and with 0.20.5 it doesn't look like it's possible for us to construct an instance of the new `CorruptMessagePayload` type. I have a couple questions:

* Should it be possible for external callers to construct errors of that case? (We currently do this for our unit tests, but maybe it's unnecessary; I'd need to read more to know for sure.)
* Should this change either be reverted or released as 0.21.0 to avoid a semver break?

Thanks!

djc:
Ah, sorry -- I missed that this was a breaking change.

Given the time it takes for the ecosystem to work through rustls upgrades, I think it would probably be better to revert this change for now, maybe figure out if we can rebuild it in a different way without breaking compatibility?

Or we can claim that only a small minority of downstream users are affected and it's not a big deal?

Maybe we should yank 0.20.5 while we decide?

No easy answers, I'm afraid.

jgallagher:
Agreed on no easy answers. Personally I'd vote yank for now - I agree with the belief that only a small number of users are affected, but it's hard to know. And since we know "small number" is strictly greater than zero in this case ;), it's possible there are others that might become broken when they try to upgrade (or worse, it breaks something that has a transitive dependency on rustls).

jgallagher:
FWIW we're going to remove our use of this error case, but that doesn't change my personal opinion on yanking.

complexspaces:
Ouch, sorry about this indirectly for breaking your code 😓.

I'm fine with yanking 0.20.5 personally. My only nit would be that it would be _nice_ to build our code with the changes in https://github.com/rustls/rustls/commit/c42c53e13dfc54495cbb62577f6bb58eddf5ff8a (without a fork) while not stalling the main branch from being able to do releases, but those two don't go together well and the wider ecosystem is a higher priority than my one project.

[...]

PR #1043, opened 2022-05-18
Attach location context to corrupt message errors

djc:
Originally from #1016, reverted in #1042 due to API compatibility
issues reported in #1041.


complexspaces:
Thanks for handling re-opening this!

PR #1046, opened 2022-05-19
impl Default for Acceptor

### LINKED ISSUE #1044: Infallible Acceptor::new
[Acceptor::new](https://docs.rs/rustls/latest/rustls/server/struct.Acceptor.html#method.new) returns a Result, which is somewhat surprising. Tracing through the calls, the [only error case appears to be](https://docs.rs/rustls/0.20.6/src/rustls/server/server_conn.rs.html#469) from calling the [private function CommonState::new](https://docs.rs/rustls/0.20.6/src/rustls/conn.rs.html#827).

`CommonState::new` returns an error [only when `max_fragment_size` is invalid](https://docs.rs/rustls/0.20.6/src/rustls/msgs/fragmenter.rs.html#66-73).

`Acceptor::new` passes a `max_fragment_size` of `None` (i.e. use the default), which is always valid. But the type system doesn't know that.

Can we make `Acceptor::new` infallible by calling an alternate internal `CommonState::new_with_default_fragment_size` that is infallible?

jsha:
`Acceptor::new()` returns an error only when `max_fragment_size` is invalid, which never happens. To avoid a breaking change, offer another way to construct an `Acceptor` via `Default`.

To make that possible, also implement `Default` for `MessageFragmenter`, replacing `MessageFragmenter::new()`. Previously there were two paths that would check that a max_fragment_size was in the acceptable range (new and set_max_fragment_size). Now there's just one path: calling set_max_fragment_size.

This provides a way to construct a MessageFragmenter with the default fragment size that is error-free.

Fixes #1044

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -824,8 +824,8 @@ pub struct CommonState {
 }
 
 impl CommonState {
-    pub(crate) fn new(max_fragment_size: Option<usize>, side: Side) -> Result<Self, Error> {
-        Ok(Self {
+    pub(crate) fn new(side: Side) -> Self {
+        Self {
             negotiated_version: None,
             side,
             record_layer: record_layer::RecordLayer::new(),
@@ -840,8 +840,7 @@ impl CommonState {
             has_seen_eof: false,
             received_middlebox_ccs: 0,
             peer_certificates: None,
-            message_fragmenter: MessageFragmenter::new(max_fragment_size)
-                .map_err(|_| Error::BadMaxFragmentSize)?,
+            message_fragmenter: Default::default(),
             received_plaintext: ChunkVecBuffer::new(Some(0)),
             sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
             sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
@@ -849,7 +848,7 @@ impl CommonState {
             protocol: Protocol::Tcp,
             #[cfg(feature = "quic")]
             quic: Quic::new(),
-        })
+        }
     }
 
     /// Returns true if the caller should call [`CommonState::write_tls`] as soon
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -840,7 +840,7 @@ impl CommonState {
             has_seen_eof: false,
             received_middlebox_ccs: 0,
             peer_certificates: None,
-            message_fragmenter: Default::default(),
+            message_fragmenter: MessageFragmenter::default(),
             received_plaintext: ChunkVecBuffer::new(Some(0)),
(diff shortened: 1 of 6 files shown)


djc:
I think the goals here are great, but I don't like the implementation much so far. In particular, I feel like the newtype doesn't really carry its weight. Can we instead have a setter on the `MessageFragmenter` that becomes the single way to set it? Also I like to review per commit, but you've left a lot of strange formatting in your earlier commits, so would be nice to fix it up.

jsha:
I've updated to make `set_max_fragment_size` the only way to set the size, and I've arranged this as two commits, each with proper formatting.

rustls/src/conn.rs
```diff
@@ -840,16 +840,15 @@ impl CommonState {
             has_seen_eof: false,
             received_middlebox_ccs: 0,
             peer_certificates: None,
-            message_fragmenter: MessageFragmenter::new(max_fragment_size)
-                .map_err(|_| Error::BadMaxFragmentSize)?,
+            message_fragmenter: Default::default(),
```
djc, line 843: Nit: would prefer to spell this as `MessageFragmenter::default()`.

rustls/src/msgs/fragmenter.rs
```diff
@@ -8,22 +8,12 @@ pub const MAX_FRAGMENT_LEN: usize = 16384;
 pub const PACKET_OVERHEAD: usize = 1 + 2 + 2;
 pub const MAX_FRAGMENT_SIZE: usize = MAX_FRAGMENT_LEN + PACKET_OVERHEAD;
 
+#[derive(Default)]
```
djc, line 11: I think this needs to be a custom `Default` implementation that sets `max_frag` to `MAX_FRAGMENT_LEN` rather than `0`?

rustls/src/conn.rs
jsha: Done.

rustls/src/msgs/fragmenter.rs
```diff
@@ -53,8 +60,9 @@ impl MessageFragmenter {
         }
     }
 
-    /// `max_fragment_size` is the maximum fragment size that will be produced --
-    /// this includes overhead. A `max_fragment_size` of 10 will produce TLS fragments
+    /// Set the maximum fragment size that will be produced.
+    ///
+    /// This includes overhead. A `max_fragment_size` of 10 will produce TLS fragments
```
djc, line 65: "fragments up to 10 bytes" -- (non-native speaker) does that need an "of" in there somewhere?

jsha: Good point. Change to "will produce TLS fragments up to 10 bytes long."

I also took the opportunity to name the parameter `max_fragment_size` and to additionally document the meaning of `None`.

jsha:
Pushed one more small edit adding to the documentation of set_max_fragment_size so it documents its error conditions.

djc (approved):

Issue #1036, opened 2022-05-11
documentation links to `internals` for enums


jsha:
At https://docs.rs/rustls/latest/rustls/server/struct.ClientHello.html#method.signature_schemes, the `SignatureScheme` type links to [rustls::internal::msgs::enums::SignatureScheme](https://docs.rs/rustls/0.20.4/rustls/internal/msgs/enums/enum.SignatureScheme.html). This is the same type as [rustls::SignatureScheme](https://docs.rs/rustls/0.20.4/rustls/enum.SignatureScheme.html), but exported under a different name.

The `internals` version of this name is exported here: https://github.com/rustls/rustls/blob/093e88e7e180d44eb94b6f2c30b92a89f9f0cdbe/rustls/src/lib.rs#L347-L358

In practice, even though the comment says the `internals` paths are not part of the public API, they become so by being linked from the public API.

One simple and effective fix would be: We could remove `pub mod internal` entirely. Then all links to `SignatureScheme` would go to the public API path, `rustls::SignatureScheme`.

djc:
I think we're exposing `internal` so it's available for use in tests -- I think it's also sometimes used by other projects to do low-level TLS stuff (even though the API is unstable). So taking that away seems like a heavy hammer.

In my mind we should instead be moving things that are getting exposed as part of the stable API out of `msgs` and into some other place, so that it's more obvious when working on rustls which APIs are (stably) public.

jsha:
Sounds like a good plan. Here's a quick-and-dirty search to find enums from internal that are part of public APIs:

```
$ cargo doc && find target/doc/rustls -name '*.html' | grep -v internal/ | xargs sed -n 's/.*enum rustls::internal::msgs::enums::\([A-Za-z]\+\).*/\1/p' | sort -u
warning: /home/jsha/rust/rustls/rustls/Cargo.toml: file found to be present in multiple build targets: /home/jsha/rust/rustls/rustls/tests/benchmarks.rs
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
AlertDescription
CipherSuite
ContentType
HandshakeType
NamedGroup
ProtocolVersion
SignatureAlgorithm
SignatureScheme
```

[...]

PR #1048, opened 2022-05-29
Move some enums that are part of the public API.

jsha:
Move ProtocolVersion, CipherSuite, and SignatureScheme from src/msgs/enums.rs to src/enums.rs. This removes them from the `internals` part of the public API and preserves their existing visibility as `rustls::ProtocolVersion`, `rustls::CipherSuite`, and `rustls::SignatureScheme`. Doing this ensures that they are visible in the docs under only one name: the stable API one.

Part of #1036


rustls/src/lib.rs
```diff
@@ -336,6 +336,9 @@ mod x509;
 mod check;
 mod bs_debug;
 mod builder;
+#[allow(non_camel_case_types)]
```
djc, line 339: If we need these, I think they should go in an inner attribute inside the module. Why do we need them?

rustls/src/lib.rs
```diff
@@ -383,6 +383,7 @@ pub use crate::ticketer::Ticketer;
 pub use crate::tls12::Tls12CipherSuite;
 pub use crate::tls13::Tls13CipherSuite;
 pub use crate::versions::{SupportedProtocolVersion, ALL_VERSIONS, DEFAULT_VERSIONS};
+pub use enums::*;
```
djc, line 386: I prefer to never use `*` imports, as they make it less obvious where names come from.

rustls/src/msgs/enums_test.rs
```diff
@@ -15,7 +15,7 @@ fn get16<T: Codec>(enum_value: &T) -> u16 {
     (enc[0] as u16 >> 8) | (enc[1] as u16)
 }
 
-fn test_enum16<T: Codec>(first: T, last: T) {
+pub(crate) fn test_enum16<T: Codec>(first: T, last: T) {
```
djc, line 18: Maybe this should also be `#[cfg(test)]`?

rustls/src/msgs/handshake.rs
```diff
@@ -1,15 +1,16 @@
-use crate::key;
+use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::msgs::base::{Payload, PayloadU16, PayloadU24, PayloadU8};
 use crate::msgs::codec;
 use crate::msgs::codec::{Codec, Reader};
 use crate::msgs::enums::ECCurveType;
+use crate::msgs::enums::HandshakeType;
```
djc, line 6: Style nit: I think it'd be better to concentrate `msgs::enums` imports in as few lines as possible.

rustls/src/msgs/message.rs
```diff
@@ -1,11 +1,12 @@
+use crate::enums::ProtocolVersion;
 use crate::error::Error;
 use crate::msgs::alert::AlertMessagePayload;
 use crate::msgs::base::Payload;
 use crate::msgs::ccs::ChangeCipherSpecPayload;
 use crate::msgs::codec::{Codec, Reader};
+use crate::msgs::enums::ContentType;
```
djc, line 7: Style nit: coalesce this with `HandshakeType` at least?

rustls/src/msgs/mod.rs
```diff
@@ -12,6 +12,7 @@ pub mod deframer;
 pub mod enums;
 pub mod fragmenter;
 #[allow(non_camel_case_types)]
+#[macro_use]
```
djc, line 15: I think it's better to specifically import the required macros by name, which I think is the idiomatic style starting with 2018 edition.

rustls/src/server/hs.rs
```diff
@@ -206,7 +205,7 @@ impl ExtensionProcessing {
             .is_some()
             || hello
                 .cipher_suites
-                .contains(&CipherSuite::TLS_EMPTY_RENEGOTIATION_INFO_SCSV);
+                .contains(&crate::enums::CipherSuite::TLS_EMPTY_RENEGOTIATION_INFO_SCSV);
```
djc, line 208: Style nit: this line is doing enough work as it is, probably better to import `CipherSuite` in some enclosing scope.

rustls/src/server/server_conn.rs
```diff
@@ -1,14 +1,14 @@
 use crate::builder::{ConfigBuilder, WantsCipherSuites};
 use crate::conn::{CommonState, ConnectionCommon, Side, State};
+use crate::enums::ProtocolVersion;
```
djc, line 3: Nit: use the opportunity to fold these into a single line?

rustls/src/enums.rs
```diff
@@ -0,0 +1,522 @@
+/// This file is autogenerated.  See https://github.com/ctz/tls-hacking/
```
djc, line 1: I think this is no longer correct. @ctz?

djc (commented):
I think this is a good direction.

I added some comments on import style. IMO the style is a bit messy and I've been trying to clean it up over time when making changes that required different imports. I've been working into a direction where we have fewer lines of imports with repeated prefixes.

rustls/src/lib.rs
jsha: I started down this path, and realized that the comments would be pointless repetition of the enum contents. For instance, commenting on `SSLv2` with "SSL Version 2" doesn't really help anybody. Similarly, commenting `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256` with some prose explanation of its components would be pointless. The RFCs don't attempt to do that when registering the parameters.

jsha:
Thanks to unpacking the `*` import, I realized I'd accidentally publicized the DecomposedSignatureScheme and SupportedSignatureSchemes types. I moved those back to their present position in `internals`.

I've kept the latest two commits separate for ease of review, but I'm happy to squash them into the first commit if you'd prefer.

rustls/src/lib.rs
djc: Fair enough, but maybe the enums themselves should get a short docstring pointing to the RFC/IANA registry or whatever?

rustls/src/client/client_conn.rs
```diff
@@ -1,14 +1,14 @@
 use crate::builder::{ConfigBuilder, WantsCipherSuites};
 use crate::conn::{CommonState, ConnectionCommon, Protocol, Side};
+use crate::enums::CipherSuite;
```
djc, line 3: So maybe fold these?

rustls/src/lib.rs
djc: Would still be good to move these to inner attributes IMO.

rustls/src/msgs/handshake_test.rs
```diff
@@ -2,6 +2,7 @@ use super::base::{Payload, PayloadU16, PayloadU24, PayloadU8};
 use super::codec::{put_u16, Codec, Reader};
 use super::enums::*;
 use super::handshake::*;
+use crate::enums::*;
```
djc, line 5: Boy scout rule explicit exports (for this new line only)?

djc (commented):
Almost there!

rustls/src/lib.rs
jsha: I tried putting `#[allow(non_camel_case_types)]` at the top of rustls/src/enums.rs, but it did not appear to take effect for the declarations within the file.

I tried putting that attribute directly above the `enum_builder!` calls, but it similarly did not appear to take effect.

I tried putting that attribute inside the definition of `enum_builder!`, under the logic that most of the enums are non-camel case. But some (like HandshakeType) _are_ camel case, so I got "warning: unused attribute `allow`".

I'm sure there's a way to do this and I'll keep poking at it, but if you have an idea of how it should work, let me know.

djc: It needs to become an **inner** attribute, I think? `#![allow(non_camel_case_types)]`.

jsha: Ah, thanks for that. I hadn't grokked the difference between inner and outer attributes.

PR #1049, opened 2022-05-29
Derive Eq wherever PartialEq is derived.

jsha:
Fixes a clippy lint on nightly.


djc:
I don't think we should do this for private enums:

https://github.com/rust-lang/rust-clippy/issues/8875

jsha:
Reverted the addition of `Eq` on private enums. Note that I kept it for semi-public items under `internals`, because they are actually used externally and because clippy won't be able to tell the difference (assuming your proposed change to clippy lands).

djc (approved):

Issue #670, opened 2021-04-21
`MessageEncrypter` should have a non-allocating interface for `encrypt`


briansmith:
Currently we have only:
```
    fn encrypt(&self, m: BorrowMessage, seq: u64) -> Result<Message, Error>;
```

We should have a variant that works like this:
```
    fn encrypt<P>(&self, m: Message, seq: u64) -> Result<Message, Error>;
```
This new variant would be able to avoid copying the message payload into a new `Vec` because it would be able to reuse the `Vec` in `m` and use the in-place *ring* API for encryption that as designed for this.

djc:
I've been looking at this, too. I think the issue here is the insertion of nonces. We can probably do it in a smarter way, though, and I plan to work on that.

briansmith:
> I think the issue here is the insertion of nonces. We can probably do it in a smarter way, though, and I plan to work on that.

First, the explicit part of the nonce is only inserted for TLS 1.2, not TLS 1.3. So, we need not worry to much about it in terms of performance since the TLS 1.3 path is the one that we should be focusing on optimizing. Regardless, I would expect a new API that takes Vec without doing anything smarter would still be at least as efficient even for the annoying TLS 1.2 case.

ctz:
I'm not sure this would be measurable in benchmarks, but it'd be very interesting to find out. Ultimately the most important benchmark involves taking non-owned bytes and encrypting them, which is why the interface is how it currently is. But perhaps this would affect the TLS1.3 full handshake measurements (as that involves a large encryption of owned bytes -- the server and client certificate chains).

alcore:
+1 on @briansmith's experience.

 > (...) Ultimately the most important benchmark involves taking non-owned bytes and encrypting them (...)
 
As it is currently, rustls **_owns_ the data _anyways_** due to at least 2 (IIRC) unnecessary allocating copies of the appdata passed in, with many more unnecessary copies for its own protocol messages during the handshake. The issue is further exacerbated by how scatter/gather IO is used in `VecChunkBuf`, where the current design entirely defeats the purpose of using vectored IO in the first place. As it is, it only adds additional unnecessary overhead and commits like https://github.com/ctz/rustls/commit/62d15df98e6830fee74d6155f28f14dfae1b0187 compound on that, where now it's stack allocating 1KB of empty slices for _each_ single write.

Since rustls _is buffering anyways_, the IMHO sane approach would be either to utilize this fact, or work towards a proper zero-copying implementation with vectored writes. 

[...]

PR #1050, opened 2022-05-29
MessageFragmenter: return an iterator

jsha:
This avoids a couple of unnecessary allocations: the individual Payloads of the fragmented messages were allocating and filling a Vec, and the call sites were allocating a VecDeque to be filled with messages.

Instead, always return an iterator of BorrowedPlainMessage, which the caller can iterate over.

This was mentioned in https://github.com/rustls/rustls/issues/670#issuecomment-826021362.


rustls/src/conn.rs
```diff
@@ -1378,7 +1372,7 @@ pub(crate) struct Quic {
     /// QUIC transport parameters received from the peer during the handshake
     pub(crate) params: Option<Vec<u8>>,
     pub(crate) alert: Option<AlertDescription>,
-    pub(crate) hs_queue: VecDeque<(bool, Vec<u8>)>,
+    pub(crate) hs_queue: std::collections::VecDeque<(bool, Vec<u8>)>,
```
djc, line 1375: Style nit: I'm not sure why you changed the import style here? This way seems unidiomatic.

rustls/src/conn.rs
```diff
@@ -1049,15 +1047,11 @@ impl CommonState {
             Limit::No => payload.len(),
         };
 
-        let mut plain_messages = VecDeque::new();
-        self.message_fragmenter.fragment_borrow(
+        for m in self.message_fragmenter.fragment_slice(
```
djc, line 1050: Style nit: I dislike having multi-line expressions in block headers like `for` or `if`. How about naming the return value of `fragment_slice()` (I'd suggest just `iter`)?

fuzz/fuzzers/fragment.rs
```diff
@@ -22,13 +22,12 @@ fuzz_target!(|data: &[u8]| {
...
     let frg = fragmenter::MessageFragmenter::new(Some(32)).unwrap();
-    let mut out = VecDeque::new();
-    frg.fragment(
-        message::PlainMessage::from(msg),
-        &mut out,
-    );
-
-    for msg in out {
-        message::Message::try_from(msg).ok();
+    for msg in frg.fragment_message(&message::PlainMessage::from(msg)) {
```
djc, line 25: Style nit: if this was new code, I think we'd outright import `PlainMessage` and `Message` at the top of the module. Since you're rewriting anyway, maybe do that now?

rustls/src/msgs/fragmenter.rs
```diff
@@ -26,41 +23,30 @@ impl MessageFragmenter {
...
-        for chunk in msg.payload.0.chunks(self.max_frag) {
-            out.push_back(PlainMessage {
-                typ: msg.typ,
-                version: msg.version,
-                payload: Payload(chunk.to_vec()),
-            });
-        }
+    /// Return an iterator across those messages.
+    /// Payloads are borrowed.
+    pub fn fragment_message<'a>(
```
djc, line 28: Since this is now a one-line wrapper of `fragment_slice()` used only in tests and for fuzzing, can we just get rid of this? We can potentially redo it as a helper function inside `tests`?

djc (commented):
Nice! I think I've tried similar things for the defragmenter, but this seems to be simpler.

rustls/src/conn.rs
jsha: This struct has `#[cfg(feature = "quic")]`, so it is only parsed when that feature is enabled. And this is now the only use of VecDeque in this file.

If I `use std::collections::VecDeque` at the top of the file, I get clippy errors (unused use) for builds that don't include quic. There's currently no smaller scope with `#[cfg(feature = "quic")]` where I could put that `use`. I could refactor the quic-related code here to put it inside its own feature-gated `mod`, but that seemed a little beyond the scope of the refactoring at hand.

rustls/src/msgs/fragmenter.rs
jsha: I'd be happy to consolidate on just one of these two, but FWIW it's not just tests and fuzzing.

fragment_message is used in send_msg and send_msg_encrypt in conn.rs.

fragment_slice is used in send_appdata_encrypt in conn.rs.

Since fragment_slice is more generally applicable (fragment_message requires the caller to have an owned PlainMessage), I'll consolidate on that one (and rename to `fragment` as you suggest).

rustls/src/conn.rs
djc: We usually just guard the import itself with `#[cfg(feature = "quic")]` IIRC.

rustls/src/msgs/fragmenter.rs
djc: Ah, sorry, I missed that. In that case I think your previous set up was better, let's ditch the commit you added for it?

djc (dismissed):

djc (approved):

PR #1053, opened 2022-05-31
server: move filtering of signature schemes after config selection

### LINKED ISSUE #973: `Acceptor` API should not force the use of `ALL_CIPHER_SUITES`
There's no way to use `Acceptor` without depending on `ALL_CIPHER_SUITES`. This means there's no way to ensure we avoid linking in the code for unused cipher suites.

djc:
This avoids the need for using `ALL_CIPHER_SUITES` in the `Acceptor`,
at the cost of exposing more signature schemes in the `ClientHello`
emitted by the `Acceptor`.

Fixes #973.


djc:
@ctz this seems pretty straightforward, want to take a look?

ctz (approved):

PR #1057, opened 2022-06-03
Implement QUIC V2

Ralith:
Draft until I've tested this end-to-end, but believed to be complete.

I considered making the version a field in the `Protocol::Quic` enum variant, but that was substantially more disruptive.

See https://www.ietf.org/archive/id/draft-ietf-quic-v2-04.html for details. Currently nearing stable ratification, so these values probably won't change, but we probably shouldn't publish them until it's ratified.

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -152,11 +152,13 @@ impl ClientConnection {
 
         let ext = match quic_version {
             Version::V1Draft => ClientExtension::TransportParametersDraft(params),
-            Version::V1 => ClientExtension::TransportParameters(params),
+            Version::V1 | Version::V2 => ClientExtension::TransportParameters(params),
         };
 
+        let mut inner = ConnectionCore::for_client(config, name, vec![ext], Protocol::Quic)?;
+        inner.common_state.quic.version = quic_version;
         Ok(Self {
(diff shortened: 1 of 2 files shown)


cpu:
@Ralith It looks like https://datatracker.ietf.org/doc/draft-ietf-quic-v2/ is in AUTH48 and will be upgraded to a proposed standard with an RFC number "Soon :tm:".

Do you think you'll be interested in reviving this branch? It's collected a number of conflicts with the updates that landed in 0.21.0

Ralith:
This fell off my radar for a while. I'm interested in seeing this happen, though I'm not sure how actively I'll be able to drive it in the immediate future.

Ralith:
Rewrote atop main and added better tests. Still haven't tested this with Quinn, but the tests matching the RFC test vectors are pretty good evidence of correctness.

djc (approved):

Ralith:
> Guess we'll want to hold off on merging this for a bit?

Provided it won't need a bunch of rebasing, I think there's no harm in waiting for an end-to-end (or better yet, interop) test. However, we shouldn't need to wait for a *major* release, should we?

djc:
Sorry, yes -- I was wrong, thought I'd spotted the addition of a `version` argument to a public method.

PR #1060, opened 2022-06-11
Expose some enums as part of the stable API

jsha:
Expose AlertDescription, ContentType, and HandshakeType, SignatureAlgorithm, and NamedGroup as part of the stable API. Previously they were part of the unstable internals API, but were referenced by parts of the stable API.

`AlertDescription`, `ContentType`, and `HandshakeType` are fields of [`Error`](https://docs.rs/rustls/latest/rustls/enum.Error.html) variants.

`SignatureAlgorithm` is returned by the [`SigningKey`](https://docs.rs/rustls/latest/rustls/sign/trait.SigningKey.html#tymethod.algorithm) trait.

`NamedGroup` is the type of a public field of [SupportedKxGroup](https://docs.rs/rustls/0.20.6/rustls/struct.SupportedKxGroup.html).

Related to #1036.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,10 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 * Next release:
   - Planned: removal of unused signature verification schemes at link-time.
+  - Expose AlertDescription, ContentType, and HandshakeType,
+    SignatureAlgorithm, and NamedGroup as part of the stable API. Previously they
+    were part of the unstable internals API, but were referenced by parts of the
+    stable API.
 * 0.20.6 (2022-05-18)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,6 +21,10 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 * Next release:
   - Planned: removal of unused signature verification schemes at link-time.
+  - Planned: expose AlertDescription, ContentType, and HandshakeType,
```
djc, line 24: I think `Planned` is used for features that are not yet implemented, so it doesn't apply here.

djc (approved):

djc (approved):

PR #1062, opened 2022-06-12
Export DigitallySignedStruct in stable API

### LINKED ISSUE #1061: Remove PayloadU16 from public API
As discussed in https://github.com/rustls/rustls/issues/1036, PayloadU16 shows up in the public API in two places:

[`ServerCertVerifier`](https://docs.rs/rustls/0.20.6/rustls/client/trait.ServerCertVerifier.html) and other traits take [DigitallySignedStruct](https://docs.rs/rustls/0.20.6/rustls/internal/msgs/handshake/struct.DigitallySignedStruct.html) as a parameter. DigitallySignedStruct contains `pub sig: `[`PayloadU16`](https://docs.rs/rustls/0.20.6/rustls/internal/msgs/base/struct.PayloadU16.html).

[`DistinguishedNames`](https://docs.rs/rustls/0.20.6/rustls/type.DistinguishedNames.html) is a type alias for `VecU16OfPayloadU16`.

These are wrapper types for `Vec<u8>` and `Vec<Vec<u8>>` respectively. The stable API should expose them as `Vec<u8>` and `Vec<Vec<u8>>`, since the fact that these fields are 16-bit length prefixed in TLS is not interesting to the user.

Changing these directly would require waiting for the next breaking change, but we can make progress before then.

For `DigitallySignedStruct`, we should deprecate the public fields and provide accessors; the accessor for `sig` should return `&[u8]`.

I'm not sure what a good path forward for DistinguishedNames would be.

jsha:
This was exposed as part of the public API via ServerCertVerifier,
ClientCertVerifier, and structs that implement those two traits.

Transitively, this should expose PayloadU16 as part of the stable API, since
DigitallySignedStruct's `sig` field has type PayloadU16. However, PayloadU16
exposes unnecessary implementation details (specifically, that a given field is
encoded with 16-bit length). Instead, deprecate the `sig` field for public
access and provide a `sig() -> &[u8]` accessor instead.

Fixes #1061

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1595,19 +1595,26 @@ impl Codec for ECParameters {
 #[derive(Debug, Clone)]
 pub struct DigitallySignedStruct {
     pub scheme: SignatureScheme,
+    #[deprecated(since = "0.20.7", note = "Use sig() accessor")]
     pub sig: PayloadU16,
 }
 
 impl DigitallySignedStruct {
+    #![allow(deprecated)]
     pub fn new(scheme: SignatureScheme, sig: Vec<u8>) -> Self {
         Self {
             scheme,
             sig: PayloadU16::new(sig),
         }
     }
+
+    pub fn signature(&self) -> &[u8] {
+        &self.sig.0
+    }
 }
 
 impl Codec for DigitallySignedStruct {
+    #![allow(deprecated)]
     fn encode(&self, bytes: &mut Vec<u8>) {
         self.scheme.encode(bytes);
         self.sig.encode(bytes);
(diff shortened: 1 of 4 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -1595,19 +1595,26 @@ impl Codec for ECParameters {
...
 impl DigitallySignedStruct {
+    #![allow(deprecated)]
     pub fn new(scheme: SignatureScheme, sig: Vec<u8>) -> Self {
         Self {
             scheme,
             sig: PayloadU16::new(sig),
         }
     }
+
+    pub fn sig(&self) -> &[u8] {
```
djc, line 1611: Let's call this `signature()`? As I understand it the API guidelines would usually avoid abbreviations.

djc (approved):

djc:
(Needs a rebase.)

PR #1063, opened 2022-06-12
Deprecate Acceptor::wants_read

### LINKED ISSUE #1045: When is Acceptor::wants_read needed?
The [`Acceptor::wants_read`](https://docs.rs/rustls/0.20.6/rustls/server/struct.Acceptor.html#method.wants_read) method seems like it should logically always return true. The point of an Acceptor is to read bytes from the network. If no more bytes are wanted (because of an error or because a full ClientHello has been read), I would expect that to show up in the return value for [Acceptor::accept](https://docs.rs/rustls/0.20.6/rustls/server/struct.Acceptor.html#method.accept).

Implementation-wise, Acceptor delegates to CommonState::wants_read:
https://github.com/rustls/rustls/blob/c6dcb4c0c63caf503bff80f07182ded0426107ed/rustls/src/conn.rs#L1316-L1332

I think all of these conditions will always be true for Acceptor:

```rust
        self.received_plaintext.is_empty()
            && !self.has_received_close_notify
            && (self.may_send_application_data || self.sendable_tls.is_empty())
```

jsha:
This method always returns true.

Fixes #1045

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -487,9 +487,10 @@ impl Acceptor {
 
     /// Returns true if the caller should call [`Connection::read_tls()`] as soon as possible.
     ///
-    /// For more details, refer to [`CommonState::wants_read()`].
+    /// Since the purpose of an Acceptor is to read and then parse TLS bytes, this always returns true.
     ///
     /// [`Connection::read_tls()`]: crate::Connection::read_tls
+    #[deprecated(since = "0.20.7", note = "Always returns true")]
     pub fn wants_read(&self) -> bool {
         self.inner
             .as_ref()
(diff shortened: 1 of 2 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -485,11 +485,12 @@ impl Acceptor {
         Ok(Self::default())
     }
 
-    /// Returns true if the caller should call [`Connection::read_tls()`] as soon as possible.
-    ///
-    /// For more details, refer to [`CommonState::wants_read()`].
+    /// By analogy with [`CommonState::wants_read()`], this returns true to indicate that
```
djc, line 488: Nit: would still like to keep the single self-contained line that summarizes the method.

djc (dismissed):
Mostly LGTM, one nit.

djc (approved):

PR #1064, opened 2022-06-12
Deprecate RootCertStore::subjects()

jsha:
In favor of a newly added `OwnedTrustAnchor::subject()` accessor. This eliminates one use of the type alias DistinguishedNames in the public, stable API, in favor of an accessor that returns a plain `&[u8]`.

Document DistinguishedNames and the new `OwnedTrustAnchor::subject()` with a description of how to decode them.

Add references to RFC 5280 in `OwnedTrustAnchor::new()`.

Part of #1061.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -47,6 +47,18 @@ impl OwnedTrustAnchor {
             name_constraints: name_constraints.map(|x| x.into()),
         }
     }
+
+    /// Return the subject field.
+    ///
+    /// This can be decoded using [x509-parser's FromDer trait](https://docs.rs/x509-parser/latest/x509_parser/traits/trait.FromDer.html).
+    ///
+    /// ```ignore
+    /// use x509_parser::traits::FromDer;
+    /// println!("{}", x509_parser::x509::X509Name::from_der(anchor.subject())?.1);
+    /// ```
+    pub fn subject(&self) -> &[u8] {
+        &self.subject
+    }
 }
 
 /// A container for root certificates able to provide a root-of-trust
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -86,6 +86,7 @@ impl RootCertStore {
     }
 
     /// Return the Subject Names for certificates in the container.
+    #[deprecated(since = "0.20.7", note = "Use OwnedTrustAnchor::subject() instead")]
     pub fn subjects(&self) -> DistinguishedNames {
         let mut r = DistinguishedNames::new();
 
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -24,12 +24,18 @@ impl OwnedTrustAnchor {
 
     /// Constructs an `OwnedTrustAnchor` from its components.
     ///
-    /// `subject` is the subject field of the trust anchor.
+    /// All inputs are DER-encoded.
     ///
-    /// `spki` is the `subjectPublicKeyInfo` field of the trust anchor.
+    /// `subject` is the [Subject] field of the trust anchor.
     ///
-    /// `name_constraints` is the value of a DER-encoded name constraints to
+    /// `spki` is the [SubjectPublicKeyInfo] field of the trust anchor.
+    ///
+    /// `name_constraints` is the [Name Constraints] to
     /// apply for this trust anchor, if any.
+    ///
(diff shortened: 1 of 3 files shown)


djc (approved):
This is mostly okay. I think I'd prefer not to reference the x509-parser crate directly to avoid making a public endorsement (even though I like and use it myself).

jsha:
I understand not wanting to make an endorsement, but without the example I think we are leaving users adrift. Figuring out what exactly a Subject (/ Name) is, which libraries might be capable of decoding one, and which functions to call is a challenging task for anyone without lots of experience in PKIX _and_ the Rust ecosystem. For instance, I first reached for the `asn1` crate and wasted some time before realizing that doesn't have the X.509-specific types I needed. Then I spent some time fiddling with the API to get it right: importing the FromDer trait; figuring out I needed to access the `.1` element of the result tuple; figuring out I want the Display output, not the Debug output.

Also, x509-parser is already by far the [most popular option](https://lib.rs/search?q=x509), and shares a lot of philosophy with rustls:

> It is written in pure Rust, fast, and makes extensive use of zero-copy. A lot of care is taken to ensure security and safety of this crate, including design (recursion limit, defensive programming), tests, and fuzzing. It also aims to be panic-free.

In practice, I think any of our users who actually processes these bytes will be doing so with x509-parser, and we can save them some time and make dealing with TLS less frustrating.

djc:
@ctz any opinions?

I think you haven't quite convinced me. I would argue that x509-parser being the most popular option means that it should already be fairly straightforward to find, and your new documentation will make life a lot easier to pull that code together. I would also argue that enabling parsing X.509 data structures "by hand" is kind of an anti-goal of this crate, which tries to provide all the functionality that most people need out of the boxes; we don't actually want to encourage low-level stuff, and people doing it anyway should know what they're doing.

jsha:
> I would argue that x509-parser being the most popular option means that it should already be fairly straightforward to find

I agree on this. The reason I included an example is not to help people find x509-parser, but to figure out what to do with it once they've found it. Parsing a standalone Name field is unusual, so it doesn't show up in any x509-parser examples, and I'm guessing the maintainers wouldn't welcome such an example (but who knows!).

> I would also argue that enabling parsing X.509 data structures "by hand" is kind of an anti-goal of this crate, which tries to provide all the functionality that most people need out of the boxes; we don't actually want to encourage low-level stuff, and people doing it anyway should know what they're doing.

We have two mechanisms to gate "allowed but discouraged" functionality: the `dangerous_configuration` feature and the `internals` modules. We shouldn't additionally have a "knowledge gate," since that devolves to "look it up on Stack Overflow." Also, one of the ways people come to know what they're doing is by reading documentation from high quality crates.

One other possibility would be to remove the subjects accessor entirely. It seems to me that the main use is to parse and print in a human readable form for debugging purposes. Are there other uses for it?

djc:
> One other possibility would be to remove the subjects accessor entirely. It seems to me that the main use is to parse and print in a human readable form for debugging purposes. Are there other uses for it?

It's used to gather authority names to send in the server's `CertificateRequest` message.

jsha:
Right, but that's internal. Is there a need for a public accessor for this?

ctz:
I think the fact this is pub predates `pub(crate)`

djc:
Do we want to make it easy for people to reuse `RootCertStore` for their custom `ClientCertVerifier` impls? If so, I think we'll want to keep `RootCertStore::subjects()` public.

jsha:
Currently `RootCertStore` doesn't have enough API surface to use in implementing `ClientCertVerifier`.

[ClientCertVerifier::verify_client_cert](https://docs.rs/rustls/latest/rustls/server/trait.ClientCertVerifier.html#tymethod.verify_client_cert) needs to be able to build paths. [RootCertStore](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html) exposes its internals as a `Vec<OwnedTrustAnchor>`, but [`OwnedTrustAnchor`](https://docs.rs/rustls/latest/rustls/struct.OwnedTrustAnchor.html) doesn't expose any useful accessors. Internally, AllowAnyAuthenticatedClient relies on a call to the private [OwnedTrustAnchor::to_trust_anchor](https://docs.rs/rustls/latest/src/rustls/verify.rs.html#469-488), which converts to a `webpki::TrustAnchor`.

djc:
This left a bunch of warnings in rustls/tests/client_cert_verifier.rs, would be nice to clean those up.

Issue #1055, opened 2022-05-31
Question: how to read ClientHello properly?


GlenDC:
Hi, I was trying to read the ALPN of a ClientHello for an accepted connection on the server side.

The server is written using `tokio` and as such I use `rustls` in combination with `tokio_rustls`.

Here is how I get the client hello:

```rust
let acceptor = rustls::server::Acceptor::new().unwrap(); // I do proper error handling in actual code
let lazy_acceptor = tokio_rustls::LazyConfigAcceptor::new(acceptor, socket);
let start = lazy_acceptor.await.unwrap(); // I do proper error handling in actual code
assert!(start.client_hello().alpn().is_some()); // <= this works
```

However if I try to read some data from the ALPN, I get nothing:

```rust
if let Some(alpn) = tls.alpn() {
            // this part is reached...
            for ext in alpn.filter_map(ClientExtension::read_bytes) {
                // this part never reaches... So somehow the conversion from the bytes to the ClientExtension
                // is going wrong
                match ext {
                   / ...
```

The ALPN is defined as I manage to get into that branch of the code. However nothing comes through the `alpn.filter_map`...
Anyone an idea what is going wrong here? How do I go from `::alpn().unwrap()` to something like an `Iter<ClientExtension`?

Eventually this is the kind of data I wish to extract from my ClientHello ALPN:

```rust
if let Some(alpn) = tls.alpn() {
    for ext in alpn.filter_map(ClientExtension::read_bytes) {
        match ext {
            ClientExtension::ECPointFormats(list) => {
                fp.supported_points = Some(list.iter().map(|v| v.get_u8()).collect());
            }
            ClientExtension::Protocols(list) => {
                fp.supported_protos = Some(
                    list.iter()
                        .map(|payload| {
                            let mut v = Vec::new();
                            payload.encode(&mut v);
                            String::from_utf8(v).unwrap_or_else(|_| {
                                let mut v = Vec::new();
                                payload.encode(&mut v);
                                base64::encode(v)
                            })
                        })
                        .collect(),
                );
            }
            ClientExtension::SupportedVersions(versions) => {
                fp.supported_versions =
                    Some(versions.iter().map(|v| v.get_u16()).collect());
            }
            // If I understand it correctly it seems that the client extension
            // provides no default support for EllipticCurves, so we'll get it ourselves...
            ClientExtension::Unknown(unknown_ext)
                if unknown_ext.typ == ExtensionType::EllipticCurves =>
            {
[...]

PR #1070, opened 2022-06-23
document the alpn method of ClientHello in more depth

GlenDC:
as discussed and agreed in
<https://github.com/rustls/rustls/issues/1055#issuecomment-1155655552>

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -159,15 +159,16 @@ impl<'a> ClientHello<'a> {
     /// submit a set of identifiers that each a represent an application-layer protocol.
     /// The server will then pick its preferred protocol from the set submitted by the client.
     /// Each identifier is represented as a byte array, although common values are often ASCII-encoded.
-    /// See the official specifications at <https://datatracker.ietf.org/doc/html/rfc7301> for more information on ALPN.
+    /// See the official RFC-7301 specifications at <https://datatracker.ietf.org/doc/html/rfc7301>
+    /// for more information on ALPN.
     ///
     /// Common examples of ALPN identifiers given by web clients are "http/1.1" and "h2".
     /// These indicate that the web client supports each of these HTTP protocols,
     /// and it is the server that will respond with one of these two as the chosen HTTP protocol.
     ///
-    /// On the server side you can configure the ALPN protocols supported by your server using `ServerConfig::alpn_protocols`,
-    /// it are the protocols that it has in common with the ones defined in the `ClientHello` message which will be sent
-    /// back to the client as part of the handshake phase.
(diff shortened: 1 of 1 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -154,6 +154,20 @@ impl<'a> ClientHello<'a> {
     /// Get the alpn.
     ///
     /// Returns `None` if the client did not include an ALPN extension
+    ///
+    /// ALPN (Application Layer Protocol Negotiation) contains a set of identifiers
```
djc, line 158: I'd rewrite this more like this: "Application Layer Protocol Negotiation (ALPN) is a TLS extension that lets a client submit a set of identifiers that each a represent an application-layer protocol. The server will then pick its preferred protocol from the set submitted by the client. Each identifier is represented as a byte array, although common values are often ASCII-encoded."

And probably link to the RFC: https://datatracker.ietf.org/doc/html/rfc7301.

rustls/src/server/server_conn.rs
```diff
@@ -154,6 +154,20 @@ impl<'a> ClientHello<'a> {
...
+    /// which are encoded as a set of octet values, which could be UTF-8 encoded.
+    /// It's a form of TLS extension allowing the client to provide an externally visible
+    /// marker for the application-layer protocol associated with the TLS connection.
+    ///
+    /// Common examples of ALPN identifiers given by web clients are "http/1.1", "h2" and "h3",
```
djc, line 163: I would probably not mention h3 here, since it has a more complicated situation (since H3 is built on QUIC, it doesn't use TCP).

rustls/src/server/server_conn.rs
```diff
@@ -154,6 +154,20 @@ impl<'a> ClientHello<'a> {
...
+    /// and are used to hint to the web server that these HTTP protocols are supported by the
+    /// web client can be established if desired.
```
djc, line 165: "can be established if desired" doesn't seem quite right?

rustls/src/server/server_conn.rs
```diff
@@ -154,6 +154,20 @@ impl<'a> ClientHello<'a> {
...
+    ///
+    /// You can find a list of many other possible identifier values at
+    /// <https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids>,
+    /// note however that this is not an exhaustive list, so please keep in mind
```
djc, line 169: I'd start a new sentence here and probably ditch the second part ("so please ..."), since it doesn't seem to add much value.

djc (approved):

djc:
Thanks for the suggestion, I'm proposing some further changes in #1078.

PR #1072, opened 2022-06-24
expose more extensions in client hello server struct

### LINKED ISSUE #1055: Question: how to read ClientHello properly?

GlenDC:
Closes #1055 and as discussed in that issue:

- careful with memory;
- added test cases;

Some of these extensions have enumerators defined,
but which aren't exposed as a public type. For now I exposed them as raw u8/u16 values.
Might perhaps be better to expose the types instead?

Something else I noticed is that the constructor to creating ClientHello is becoming lengthy, might be something we want to do something about as well.

### DIFF
--- a/rustls/src/server/handy.rs
+++ b/rustls/src/server/handy.rs
@@ -261,7 +261,16 @@ mod test {
     fn test_resolvesservercertusingsni_requires_sni() {
         let rscsni = ResolvesServerCertUsingSni::new();
         assert!(rscsni
-            .resolve(ClientHello::new(&None, &[], None, &[], None))
+            .resolve(ClientHello::new(
+                &None,
+                &[],
+                None,
+                &[],
+                None,
+                None,
+                None,
+                None
+            ))
             .is_none());
     }
 
@@ -272,7 +281,16 @@ mod test {
             .unwrap()
             .to_owned();
         assert!(rscsni
-            .resolve(ClientHello::new(&Some(name), &[], None, &[], None))
+            .resolve(ClientHello::new(
+                &Some(name),
+                &[],
+                None,
+                &[],
+                None,
+                None,
+                None,
+                None
+            ))
             .is_none());
     }
 }
--- a/rustls/src/server/handy.rs
+++ b/rustls/src/server/handy.rs
@@ -261,7 +261,7 @@ mod test {
     fn test_resolvesservercertusingsni_requires_sni() {
         let rscsni = ResolvesServerCertUsingSni::new();
         assert!(rscsni
-            .resolve(ClientHello::new(&None, &[], None, &[]))
+            .resolve(ClientHello::new(&None, &[], None, &[], None))
             .is_none());
     }
 
@@ -272,7 +272,7 @@ mod test {
             .unwrap()
             .to_owned();
         assert!(rscsni
-            .resolve(ClientHello::new(&Some(name), &[], None, &[]))
+            .resolve(ClientHello::new(&Some(name), &[], None, &[], None))
             .is_none());
     }
 }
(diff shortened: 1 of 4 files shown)


ctz:
I'm sorry, I don't think this is useful. Please write down, for example, how and why a server will choose one certificate over another if a client presents a different set of key exchange groups, or EC point formats, or PSK modes, or protocol versions (note that the latter could be useful if it weren't already implicit in the exposed `cipher_suites`). I don't think there are is a good reason to do that, because these are all orthogonal concerns when the server is selecting a certificate.

GlenDC:
This is mainly for fingerprinting purposes. E.g. detect if an incoming connection is coming from a legit client.
If this is out of the scope for this library than my apologies and I'll close it. Please do let me know.

GlenDC:
I'll close this MR. I seem to get the impression that anything in purpose of fingerprinting is outside the scope of this project. Which I respect. Thank you for taking the time to review though.

ctz:
For fingerprinting there are some sources of data that you'll really want but will be lost by a general purpose TLS library. Particularly: the order of extensions, and choices between multiple valid encodings for the same message, and how handshake messages map to TLS messages. My suggestion would be to do the fingerprinting separately from a TLS library, and then pass the messages through if your fingerprint heuristics are satisified.

But apologies for the wasted work here.

Issue #423, opened 2020-11-23
Support RawPublicKey (non-X509) certificates, e.g. for P2P


infinity0:
I am looking into using rustls for a p2p network, which authenticates peers (specifically their public keys) via custom means outside of the public X509 root certs.

Currently in rustls it is possible for a client to omit server certificate verification via [ClientConfig.dangerous()](https://docs.rs/rustls/0.19.0/rustls/struct.ClientConfig.html#method.dangerous), however the server via [ServerConfig.cert_resolver](https://docs.rs/rustls/0.19.0/rustls/struct.ServerConfig.html#structfield.cert_resolver) still has to provide a [CertifiedKey](https://docs.rs/rustls/0.19.0/rustls/sign/struct.CertifiedKey.html) which hard-codes a [X509 certificate](https://docs.rs/rustls/0.19.0/rustls/struct.Certificate.html).

TLS 1.3 mandates support for [different types of certificates](https://tools.ietf.org/html/rfc8446#section-4.4.2), specifically including RawPublicKey. This makes p2p usage more convenient as we don't have to faff around with creating a self-signed X509 certificate. Please consider supporting this in rustls. Of course the default behaviour for a client would be to fail to verify these "certificates".

(TLS 1.2 has optional support for the same functionality, but implemented differently as an extension, [RFC 7250](https://tools.ietf.org/html/rfc7250). The use-case I mention here is only about new protocols, so I don't consider this a necessary part of this issue.)

burdges:
A priori, one wants back certs however, which using a certificates provides.  

Story 1. We expand rustls by support for sr25519 and ecdsa secp256k1 (already support ed25519).  We then ask that controller keys certify nodes TLS keys, which nodes supply upon opening connections thereby proving their intent to use that certificate.  

Story 2. We expand rustls by RawPublicKey.  We then ask that controller keys certify nodes TLS keys on-chain like all other session keys now, although afaik our long-term transport key still never get handled this way even now.  

Story 3.  We start like story 2 but go further by asking that nodes TLS keys certify their controller key on-chain ala https://github.com/paritytech/substrate/issues/7398  

Is there any difference?  In story 2, Alice could register Bob's session key under her own controller key, perhaps creating strange effects, like Alice avoiding slashing while offline or whatever.  Afaik, there are only negligible differences between stories 1 and 3 so long as the chain acts as a source of truth, but these are transport keys so they get used before the chain gets established, so maybe some mild shenanigans exist. 

[...]

PR #1074, opened 2022-06-26
Allow ignoring hostnames in server cert validation

jeffparsons:
This may be useful for, e.g., peer-to-peer applications where peers
authenticate with each other using certificates but do not have stable
addresses or hostnames.

I'm guessing there may be objections to this API because it allows dangerous
configuration too easily. I'm mostly putting this up for discussion instead of
opening an issue, because at least this is something concrete to point at. :)

Perhaps a better API would be to simultaneously provide a single trusted
certificate _and_ ignore hostnames in one go — that way at least you _can't_
configure it to accept any certificate on the internet without meaning to.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -26,6 +26,7 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
                 kx_groups: self.state.kx_groups,
                 versions: self.state.versions,
                 root_store,
+                accept_invalid_hostnames: false,
             },
             side: PhantomData::default(),
         }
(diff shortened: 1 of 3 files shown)


Ralith:
If hostnames are not significant, rather than hackily ignoring them, I think the correct solution for P2P applications is `RawPublicKey` certificates as proposed in https://github.com/rustls/rustls/issues/423.

jeffparsons:
Thanks, @Ralith, I didn't even know that was a thing. It sounds a lot like what I actually want.

I probably should have also specified that the use case I have in mind involves using HTTP for request/response pattern. (I imagine eventually moving to HTTP/3 only, and having both HTTP and other streams all going over QUIC.) Any idea if raw public key certificates could be used with HTTP, or would the relevant HTTP spec need to have been written with it in mind? I guess even if it's not strictly kosher, as long as I'm handling the TCP+TLS or QUIC connection myself and then handing it off to an HTTP library, then I could probably work around whatever the rules for HTTP are supposed to be. 😅

I'll close this for now, because it sounds like there's a much better solution to my problem. Cheers!

PR #1082, opened 2022-07-12
Fix deprecation warnings for subjects() in tests

jsha:
Followup to https://github.com/rustls/rustls/pull/1064#issuecomment-1168462184

### DIFF
--- a/rustls/tests/client_cert_verifier.rs
+++ b/rustls/tests/client_cert_verifier.rs
@@ -10,6 +10,7 @@ use crate::common::{
     make_pair_for_arc_configs, ErrorFromPeer, KeyType, ALL_KEY_TYPES,
 };
 use rustls::client::WebPkiVerifier;
+use rustls::internal::msgs::base::PayloadU16;
 use rustls::internal::msgs::enums::AlertDescription;
 use rustls::internal::msgs::enums::ContentType;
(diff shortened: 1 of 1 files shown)


djc:
Looks like it's time to bump the MSRV, can you do that in this PR?

Thanks for fixing this up!

jsha:
The error appears to be in "plotters", which is a transitive dependency of "criterion", which is a dev-dependency. Rather than bumping the MSRV, perhaps we can exclude rustls/tests/benchmarks.rs when we run tests on the MSRV?

```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/plotters-0.3.2/Cargo.toml`
Caused by:
  feature `edition2021` is required
```

djc:
I guess, but as far as I've seen the cat is out of the bag on 1.56.

djc:
Issue in plotters: https://github.com/plotters-rs/plotters/issues/398.

djc:
@jsha did you want to finish this up by updating the MSRV?

jsha:
Yes. I thought I had already sent a PR updating the MSRV, but it looks like it never made it off my computer. Thanks for the ping.

jsha:
Actually, MSRV was updated in #1083. This just needed to be rebased. That's now done and this is passing tests.

djc (approved):

PR #1084, opened 2022-07-16
Fix clippy: used `assert_eq!` with a literal bool

TieWay59:
Change outdated use of `assert_eq!` in `handy.rs` and `api.rs`.

Because `#[warn(clippy::bool_assert_comparison)]` is on by default.

I've found some correct use of `assert!` in the codebase so I think it's preferable to include this Clippy fix.

For further information visit https://rust-lang.github.io/rust-clippy/master/index.html#bool_assert_comparison

### DIFF
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -107,7 +107,7 @@ mod test {
     #[test]
     fn test_noclientsessionstorage_drops_put() {
         let c = NoClientSessionStorage {};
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Thanks!

TieWay59:
Please review carefully with positive asserts and negative asserts. I hope I didn't break the original asserts intention accidentally in this PR. @djc ❤

Issue #1086, opened 2022-07-19
ClientCertVerifier not associate with sni_name


AlexZhao:
Hi,

For the client verifier trait ClientCertVerifier which can be used to verify client side certificate, for its interface function
fn verify_client_cert(&self, &Certificate, &[Certificate], SystemTime),  it doesn accept sni_name which buffered in the context of the session, is there any possibility to modify this interface can support sni_name as input parameter then, for the listening endpoint it can
achieve to verify the client side authentication certificate against multiple host name with different CAs?

Any idea is this doable?

Br
Alex

djc:
You should supply a different `ServerConfig` per SNI name if you want to customize how client certificates get verified, you can use the `Acceptor` for this.

AlexZhao:
> You should supply a different `ServerConfig` per SNI name if you want to customize how client certificates get verified, you can use the `Acceptor` for this.

I understand ServerConfig is associated with a listening endpoint, if single listening endpoint associate with multiple domain names, does it require to have distinguish the connection according to its domain name during TLS handshake phase just like the trait ResolvesServerCertName interface?

djc:
Your question doesn't make sense to me.

rustls itself does not deal with "listening endpoint"s as such. A `ServerConnection` is created per connection and `ServerConnection`s can share `ServerConfig`s. There is no `ResolvesServerCertName`, there is a `ResolvesServerCert` trait and a `ResolvesServerCertUsingSni` implementation that makes it possible to yield a different certificate for a different SNI. However, if you want to do more complex things like changing client certificate verification logic depending on the SNI, you'd need to have one `ServerConfig` per client certificate verification logic.

AlexZhao:
Ok, I mean for the ResolvesServerCertUsingSni, it has the possibility to pick certs according to SNI name, for the ClientCertVerifier is there any reason not able to support similiar functionalities which provide sni_name for input parameters, then it can work just like ResolvesServerCertUsingSni and support multiplexing resolve client cert according to the sni name.

I create a pull request to have sni name client cert verification can be supported, I'm not sure rustls plan on this, but if needed I can help on this part.

Thanks

djc:
There's just not as clear of a mapping from the SNI to the client certificate verification mechanism. We consider this a niche enough requirement that you'll need to do a little more work. But you still haven't responded to my suggestions about the `Acceptor` and why it would not satisfy your use case, so it feels like we keep talking past each other.

AlexZhao:
> 

[...]

PR #1087, opened 2022-07-20
To have ClientCertVerifier to support sni_name_str

AlexZhao:
Hi, experts

trying to modify ClientCertVerifier to support multiplexing certs according to DNSName, can you check is this doable in rustls project?

Br
Alex


djc:
We're not interested in supporting this directly. See more discussion in #1086.

PR #1095, opened 2022-08-15
Upgrade coverage LLVM to 15 to track nightly Rust

djc:


djc:
I'm not sure how lcov is failing here and I don't have a good setup for testing it locally. Maybe someone with a Debian machine can try to see what's going on here? It seems `lcov` just fails with a 255 exit code but nothing on stdout/stderr.

ctz:
It's odd, cos this worked for me locally after I'd updated my nightly and installed LLVM 15. Will try to get to the bottom of it.

djc:
Seems to be good now, thanks!

ctz (approved):

PR #1097, opened 2022-08-27
Reorganize example programs for clarity

bdaehlie:
The directory structure for the Rustls example code could be more clear. I think it should just be called examples at the top level, then have programs and tests under that. This will help point new users and contributors in the right direction for clear usage examples.

It doesn't seem like we need to keep calling out the mio dependency. At the top level it makes it sound like it's some kind of mio integration rather than just example code. Also, there are a bunch of dependencies, not just mio.


djc (commented):
I agree this might be clearer. I think it would be better to put `tlsserver` and `tlsclient` in `examples/src/bin`, and then you can get rid of the `[[example]]` bits from the `Cargo.toml` altogether.

bdaehlie:
Marking this as a draft because even after I address the comment from djc I'd like to look into further improvements before merging.

bdaehlie:
I moved the example programs from /rustls/examples to the top-level /examples directory.

There is still some stuff in /rustls/examples/internal but I'm not sure those are actually example programs. They seems like forms of tests. I'm hoping we can deal with those three after this PR. Once we put them somewhere better there will be a single top-level examples directory.

djc (approved):

PR #1098, opened 2022-09-01
Fix code depending on updated ReadBuf API

bdaehlie:


djc (dismissed):
LGTM modulo the format issue, thanks!

djc (approved):

PR #1102, opened 2022-09-13
Add derive Hash for ServerName and DnsName

wutchzone:
I found myself in need of adding the `Hash` derive to the `ServerName`.


djc (approved):
Makes sense, thanks!

PR #1104, opened 2022-09-18
Documented codec Reader and renamed variables

jacobtread:
Renamed offs to cursor for a more descriptive variable name as it acts like a cursor for traversing the underlying buffer

Renamed buf to buffer for the sake of doing so

Documented functions of Reader and also Codec trait

### DIFF
--- a/rustls/src/msgs/codec.rs
+++ b/rustls/src/msgs/codec.rs
@@ -1,78 +1,106 @@
 use std::convert::TryInto;
 use std::fmt::Debug;
 
-/// Read from a byte slice.
+/// Wrapper over a slice of bytes that allows reading chunks from
+/// with the current position state held using a cursor.
+///
+/// A new reader for a sub section of the the buffer can be created
+/// using the `sub` function or a section of a certain length can
+/// be obtained using the `take` function
 pub struct Reader<'a> {
-    buf: &'a [u8],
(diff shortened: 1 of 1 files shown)


djc:
What's your interest in this API? Are you aware that it is considered internal/unstable?

jacobtread:
I was working on a fork of this crate for a legacy project of mine and just saw that this portion of the code looked like it needed some love I made these changes in my personal fork so I though I might aswell make them for you guys aswell even if its not that big of a contribution :)

djc (approved):
Okay, thanks!

Would be curious to hear about what you forked the library for.

jacobtread:
> Okay, thanks!
> 
> Would be curious to hear about what you forked the library for.

I am working on a legacy game server for a game called Mass Effect 3 and it seems to be hard coded to only support SSLv3 and two very specific cipher suites. I was working on a fork of this library that implements those as they will not be ever supported in the official code base due to how insecure they are and such. 

The current work around uses schannel but requires registry tweaks and enabling the ciphers through PowerShell so I thought I might give implementing it myself a go. My very unfinished fork was basically me just stripping down this library for the sole purpose of trying to get SSLv3 working. 

However, I think I need to do some research on cryptography and such further to get a better understanding of how certain things work

PR #1108, opened 2022-09-22
Port to ring 0.17

djc:
Uses the latest commit from git for now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,8 +26,8 @@ jobs:
         include:
           - os: macos-latest
             rust: stable
-          - os: windows-latest
-            rust: stable
+          #- os: windows-latest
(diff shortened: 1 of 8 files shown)


cpu:
@djc Is this something you think should be revived? If there's still interest in moving to the unreleased code in Ring git I could take a look at adopting if you're short on time.

cpu:
I think we can close this out to keep the PR tracker tidy.

It doesn't seem like a Ring 0.17 release will be cut anytime soon, there are a number of conflicts to resolve, and the direction the project is heading towards will be de-emphasizing reliance on ring.

When Ring 0.17 is released we can rehydrate this branch as appropriate.

Thanks!

djc:
Arguably this could be recut as a separate CryptoProvider impl.

Issue #198, opened 2018-10-02
KTLS Support


quininer:
KTLS is tls support in linux kernel, but it only handles symmetric encryption and decryption, so using Rustls as a handshake is great.

We can simply public `SessionSecrets`, but this does not apply to TLS 1.3.

raphaelcohn:
This is a big win for servers that regularly serve large files, as it enables to use `sendfile()` and friends once again - and so transfer data from files to the client with zero copy - ie not having to pass copies of bytes through the kernel (and so also significantly reducing the number of syscalls required - another win post-Spectre).

thalesfragoso:
I would like to bring attention to this issue once more,  It seems that there was some hold back because it could make the library "less secure", but how about if we only expose it behind a feature flag like `dangerous_configuration` ?

djc:
Ideally we want some interface here that doesn't require leaking the secrets out of rustls. I have not worked with KTLS before, but would be open to reviewing a proposal on how to make this work.

Geal:
maybe this could be done by integrating https://github.com/quininer/ktls inside rustls? Then having `ServerSession` convert to it?

Geal:
so, after looking a bit more into it, I have a rough plan.
For some context: the way KTLS works is that the userland TLS implementation performs the handshake, then a KTLS file descriptor is created, then the IV, keys and sequence ids are extracted from the TLS session, and added to the KTLS fd (see https://github.com/quininer/ktls/blob/master/src/sys/mod.rs#L40-L62 ). Then read(), write() can be called directly on this fd.

We want to avoid making rustls session secrets easily accessible, so the idea would be to transform a `ClientSession` or `ServerSession` and the associated socket into a `KtlsStream`, which would be provided by rustls too, so the secrets can be transmitted and set up directly, with no easy access.

Possible issues:
- Linux only (I have not looked yet into FreeBSD's ktls implementation)
- limited cipher suites
- in async reactors like tokio or async-std, that rely on epoll, I think the new socket does not need to be registered to the epoll implementation, but I am not sure yet. If reading from the ktls fd consumes everything from the TCP socket's buffer, the kernel should understand that we want to read again? How do we tell the executor which task should be waked? The solution might be to register the ktls fd to epoll. This will require some tests
- who owns which file descriptors, should we forget about the socket's fd after transforming it to a ktls fd?

By reusing the previous ktls proof of concept, I should be able to have a test up and running quickly, if that scheme looks fine. WDYT @djc ?

[...]

PR #1110, opened 2022-10-03
extract_secrets feature for kTLS support

fasterthanlime:
As discussed in #198 + on Discord.

Security will be ensured by:

  * gating this behind a cargo feature flag
  * explicitly opting into this behavior via a ClientConfig + ServerConfig flag

I've chosen the callback approach (which is passed a struct that borrows from data) to 1) make sure the secrets live as shortly as possible, 2) avoid unnecessary allocations/copying.

This PR is not finished, I'm filling it up from my research branch at https://github.com/rustls/rustls/compare/main...hapsoc:rustls:export as I go, to keep the discussion open.

  * [x] Support TLS 1.2 and TLS 1.3, with AES128GCM, AES256GCM, and Chacha20Poly1305
  * [x] Gate behind `extract_secrets` cargo feature
  * [x] Gate behind explicitly `enable_secret_extraction` opt-in in `ServerConfig` and `ClientConfig`
  * [x] Add integration tests that extract secrets for all 6 supported combinations and encrypt/decrypt data with it
  * [x] Don't use any .unwrap() outside of tests


djc:
We'll need to figure out how to cover this code with tests.

fasterthanlime:
Okay, this is working as expected! All 12 tests of my ktls crate work with it, see https://github.com/hapsoc/ktls/pull/9 for example code.

This does minimal heap allocations (only one Vec<u8> in `make_key_block` afaict), it's all gated behind a feature flag as discussed, and it's very convenient to consume when setting up kTLS.

Testing this remains an open question, I'll research this tomorrow (but welcome any and all help!).

fasterthanlime:
There's a bunch of failing tests on there that seem unrelated to my changes:

```
error: unreachable `pub` item
   --> /src/rustls/rustls/src/lib.rs:365:32
    |
364 | pub use crate::builder::{
    | --- help: consider restricting its visibility: `pub(crate)`
365 |     ConfigBuilder, ConfigSide, WantsCipherSuites, WantsKxGroups, WantsVerifier, WantsVersions,
    |                                ^^^^^^^^^^^^^^^^^
    |
    = help: or consider exporting it for use by other crates
   --> /src/rustls/rustls/src/lib.rs:268:5
    |
268 |     unreachable_pub,
    |     ^^^^^^^^^^^^^^^
```

Not sure what to do about those.

fasterthanlime:
> Testing this remains an open question, I'll research this tomorrow (but welcome any and all help!).

I've looked into this a bit, but I'd honestly rather defer to someone more comfortable with the codebase. Integration tests at https://github.com/hapsoc/ktls all pass - I'm not sure exactly which parts matter when setting up unit tests for this. Strip too much and the test may as well not be there, strip too little and it's a whole integration test again.

ctz:
> Not sure what to do about those.

These are rustc nightly regression; see https://github.com/rust-lang/rust/issues/102352

rustls/src/tls13/key_schedule.rs
```diff
@@ -334,6 +334,92 @@ impl KeyScheduleTraffic {
...
+
+    #[cfg(feature = "extract_secrets")]
+    pub(crate) fn extract_secrets(
+        &self,
+        tx_seq: u64,
+        rx_seq: u64,
+        algo: &ring::aead::Algorithm,
+        side: crate::conn::Side,
+    ) -> Result<crate::ExtractedSecrets, Error> {
+        let expand = |secret: &hkdf::Prk, key: &mut [u8], iv: &mut [u8]| -> Result<(), Error> {
```
djc, line 346: It looks like this closure doesn't actually depend on any local state, so it can just be a function? It could even take two const generic parameters for the sizes of the key and IV arrays so they can be returned directly instead of being passed in?

And if that works, I'm not sure the `extract` closure still carries its weight.

rustls/src/client/tls12.rs
```diff
@@ -1107,4 +1107,10 @@ impl State<ClientConnectionData> for ExpectTraffic {
             .export_keying_material(output, label, context);
         Ok(())
     }
+
+    #[cfg(feature = "extract_secrets")]
+    fn extract_secrets(&self, tx_seq: u64, rx_seq: u64) -> Result<crate::ExtractedSecrets, Error> {
+        self.secrets
+            .extract_secrets(tx_seq, rx_seq, crate::conn::Side::Client)
```
djc, line 1114: Import `Side` here too, please.

rustls/src/server/tls12.rs
```diff
@@ -930,4 +930,10 @@ impl State<ServerConnectionData> for ExpectTraffic {
             .export_keying_material(output, label, context);
         Ok(())
     }
+
+    #[cfg(feature = "extract_secrets")]
+    fn extract_secrets(&self, tx_seq: u64, rx_seq: u64) -> Result<crate::ExtractedSecrets, Error> {
+        self.secrets
+            .extract_secrets(tx_seq, rx_seq, crate::conn::Side::Server)
```
djc, line 937: And here.

rustls/src/suites.rs
```diff
@@ -214,6 +214,51 @@ pub(crate) fn compatible_sigscheme_for_suites(
         .any(|&suite| suite.usable_for_signature_algorithm(sigalg))
 }
 
+/// Secrets for transmitting/receiving data over a TLS session. These can be
```
djc, line 217: For docstrings, the first line should be a self-contained short description. If you have more, add an empty line after that, then describe the additional details.

rustls/src/suites.rs
```diff
@@ -214,6 +214,51 @@ pub(crate) fn compatible_sigscheme_for_suites(
...
+#[cfg(feature = "extract_secrets")]
+pub struct ExtractedSecrets {
+    /// secrets for the "tx" (transmit) direction
+    pub tx: (u64, AlgorithmSecrets),
+
+    /// secrets for the "rx" (receive) direction
+    pub rx: (u64, AlgorithmSecrets),
+}
+
+/// Secrets specific to an AEAD algorithm. These are traffic secrets,
```
djc, line 228: Same as above. Maybe expand a little on how these might get use (for example, kTLS).

rustls/src/suites.rs
```diff
@@ -214,6 +214,51 @@ pub(crate) fn compatible_sigscheme_for_suites(
...
+/// post-handshake.
+#[cfg(feature = "extract_secrets")]
+#[non_exhaustive]
+pub enum AlgorithmSecrets {
```
djc, line 232: `Algorithm` is the variant axis, but I think really we represent something more like `ConnectionTrafficSecrets` maybe?

rustls/src/tls13/key_schedule.rs
fasterthanlime: I turned expand into a function with const generics, but I still think the `extract` closure is worth it due to the client/server duplication - especially when we have to slice iv into salt + iv.

fasterthanlime:
> Maybe we should have a new `Error::SecretExtraction` variant for these errors, with a proper inner enum that can just have variants, avoiding all the messy error string stuff?

Yeah I'm not fond of the messy error string stuff.. and _maybe_ it's a place where we should actually unwrap. It can't be user error, it's a bug in rustls if it panics (but different folks have different opinions about that).

fasterthanlime:
Alright, I think I've complied with all of @djc's asks (see checklist in the original issue). I'm fairly happy with the testing situation (there's nested structs/impls but it comes out pretty clean imho).

This is ready for the second stage of reviewing I think (cc @ctz)

rustls/tests/api.rs
```diff
@@ -4194,3 +4196,151 @@ fn test_no_warning_logging_during_successful_sessions() {
...
     }
 }
+
+/// Test that secrets can be extracted.
+#[cfg(feature = "extract_secrets")]
+#[test]
+fn test_extract_secrets() {
+    use ring::aead::LessSafeKey;
+
+    let kt = KeyType::Rsa;
```
djc, line 4206: Just a few comments in this long function would make it easier to understand how this is testing the secret extractions, I think.

rustls/src/tls13/key_schedule.rs
```diff
@@ -334,6 +336,86 @@ impl KeyScheduleTraffic {
...
+            server_secrets = extract(&self.current_server_traffic_secret)?;
+        } else {
+            return Err(Error::General(format!(
+                "exporting secrets for {:?}: unimplemented",
+                algo
+            )));
+        }
+
+        let (tx, rx) = match side {
+            crate::conn::Side::Client => (client_secrets, server_secrets),
```
djc, line 414: Please import `Side`, and also `ConnectionTrafficSecrets`.

rustls/Cargo.toml
```diff
@@ -26,6 +26,7 @@ webpki = { version = "0.22.0", features = ["alloc", "std"] }
 default = ["logging", "tls12"]
 logging = ["log"]
 dangerous_configuration = []
+extract_secrets = []
```
djc, line 29: Naming nit: I think I'd prefer `secret_extraction` because the noun form better fits in with something like `dangerous_configuration` (`read_buf` notwithstanding, given that it mirrors the rustc feature name).

djc (approved):
I think this is looking pretty good!

rustls/tests/api.rs
fasterthanlime: I did that, added another test, and clarified some of the code here as well.

djc: Great, thanks!

ctz:
One property I think we need here is an absolute, cast-iron guarantee that the same secrets aren't accidentally usable for two things. If a session were to be extracted for KTLS _and_ also used in the normal way (via the usual rustls API calling down to record_layer::message_encrypter) then we'd have a big confidentiality and integrity failure.

Could we, in `Connection::extract_secrets`, tell the record_layer that this session has been exported. This could be done atomically with getting the sequence numbers. At that point it can discard its state back to `RecordLayer::new()`

djc:
Another way of encoding that would be to have `extract_secrets()` take ownership of the `ConnectionCommon`, right?

fasterthanlime:
@ctz I've made an owned version of this PR at https://github.com/rustls/rustls/pull/1112 — I'm happy with that one being merged and this one being closed, it unblocks kTLS all the same for everyone.

djc:
> That said, I can see other ways to allow that in such a high-level crate

How would you do that? Run another handshake?

fasterthanlime:
Closing in favor of https://github.com/rustls/rustls/pull/1112, as per https://github.com/rustls/rustls/pull/1112#issuecomment-1275002650

Issue #1052, opened 2022-05-29
Adding more Debug implementations


jsha:
According to the Rust API Guidelines:

> [All public types implement Debug (C-DEBUG)](https://rust-lang.github.io/api-guidelines/debuggability.html#all-public-types-implement-debug-c-debug)
> If there are exceptions, they are rare.

(One notable exception here: we must make sure Debug impls do not emit secret data).

After some grepping and sedding, here's a list of concrete types that don't currently implement Debug:

ClientConfig
CipherSuiteCommon
ClientConnectionData
ClientSessionMemoryCache
NoClientSessionStorage
WriteEarlyData
Accepted
Acceptor
AllowAnyAnonymousOrAuthenticatedClient
AllowAnyAuthenticatedClient
ClientHello
NoClientAuth
NoServerSessionStorage
ReadEarlyData
ResolvesServerCertUsingSni
ServerConfig
ServerConnectionData
ServerSessionMemoryCache
CommonState
ConnectionCommon
KeyLogFile
NoKeyLogFile
Reader
ServerConfig
Ticketer
Writer
DistinguishedNames

Also, ServerConnection and ClientConnection have a stub Debug implementation that is not very useful: it prints just "ServerConnection" or "ClientConnection" when it should print the connection state, how many bytes it has buffered, whether it wants read and/or write, and whether it's received a close_notify.

I'm happy to contribute PRs adding Debug implementations to some of these. Would that be welcome?

djc:
> I'm happy to contribute PRs adding Debug implementations to some of these. Would that be welcome?

I think so!

TieWay59:
That will be a large work! Shell we do it one by one separately on different issues? 

And would you mind editing the types you listed above into a task list? So we can track the process and other people would know about it and reach out to help.

To learn more about task lists, see: https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

djc:
Don't need to do one by one, it probably makes sense to group them together into a least somewhat larger PRs (especially for the ones that just need to add a derive).

Xuanwo:
Is anyone working on this issue? I'm interested in taking this.

djc:
No, I don't think anyone is working on this issue, and your help would be much appreciated!

Xuanwo:
Great! I'm working on this now.

ctz:
Is there a lint we can activate to validate/defend this?

Xuanwo:
[...]

PR #1111, opened 2022-10-09
feat: impl Debug for all public structs

Xuanwo:
Signed-off-by: Xuanwo <github@xuanwo.io>

Fix https://github.com/rustls/rustls/issues/1052

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -147,6 +147,18 @@ pub struct ClientConfig {
     pub enable_early_data: bool,
 }
 
+impl fmt::Debug for ClientConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ClientConfig")
+            .field("alpn_protocols", &self.alpn_protocols)
+            .field("max_fragment_size", &self.max_fragment_size)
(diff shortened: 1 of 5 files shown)


Xuanwo:
Cc @djc for reviewing, thank!

djc:
I think this is adding too many `Debug` implementations. I think we should have `Debug` implementations only on public types and public fields. Any private types and private fields should not appear in a `Debug` impl.

Xuanwo:
Cc @djc, I have removed all Debug for internal types. PTAL, thanks.

djc (approved):

djc:
If you rebase this on top of main, we'll get a working CI run.

Xuanwo:
> If you rebase this on top of main, we'll get a working CI run.

Done!

rustls/src/lib.rs
```diff
@@ -266,6 +266,7 @@
     trivial_numeric_casts,
     missing_docs,
     //unreachable_pub, https://github.com/rust-lang/rust/issues/102352
+    unreachable_pub,
```
djc, line 269: You rebased incorrectly.

Xuanwo: Oh, sorry.

djc (approved):

PR #1112, opened 2022-10-10
extract_secrets feature for kTLS support (owned)

fasterthanlime:
This is a version of #1110, but where `extract_secrets` takes `self` rather than `&self`, as discussed in https://github.com/rustls/rustls/pull/1110#issuecomment-1273831870

I'm opening this a separate PR because I don't think a decision has been made yet, but I wanted to make either ready to merge, should a decision be made while I'm blissfully asleep.


djc:
Decision power is focused in Europe, so we're not as likely to make decisions while you sleep. :)

ctz:
I like this option!

fasterthanlime:
> I like this option!

Great! I've updated my `ktls` crate to use that option:

  * https://github.com/hapsoc/ktls/blob/rustls-pr/src/lib.rs

It now has a `CompatibleCiphers` type that's meant to be built once at startup, which detects if kTLS is supported at all, and if so, which ciphers are supported. It's quite hacky for now, but it proves that the rustls interface as proposed in this PR can be useful.

djc:
Can you add a changelog entry for this in the README?

djc (approved):

djc:
@jsha do you want to do a pass over this to check for FFI feasibility?

fasterthanlime:
Just rebased this on top of `main` so https://github.com/rustls/rustls/commit/4c9ac741b98f150fd7450a90352011502d8695f2 is included, at @djc's request.

jsha:
This looks fine from a rustls-ffi perspective. It does present the same hazards for FFI code that other "into" patterns do, but I think it's fine to solve them the same way described in https://github.com/rustls/rustls-ffi/issues/244, by adding an intermediate object that records the fact that a given connection has been consumed.

djc (approved):

PR #1116, opened 2022-10-23
More read_tls() backpressure

djc:


rustls/src/vecbuf.rs
```diff
@@ -36,6 +36,10 @@ impl ChunkVecBuffer {
         self.chunks.is_empty()
     }
 
+    pub(crate) fn is_full(&self) -> bool {
+        self.len() > Self::MAX_SIZE
```
ctz, line 40: suggest we make this ```
if let Some(limit) = self.limit {
  limit <= self.len()
} else {
  False
}```
And then construct `received_plaintext` with `Some(16384)` in conn.rs. That keeps all the mechanism in this file, and all the policy in conn.rs.

djc: Sounds good, will take care of it!

djc:
Added a test to cover this and adopted the suggestion to separate policy and mechanism.

ctz (approved):
Looks good. Needs an item in the changelog in README.md so people aren't surprised by the behaviour change (thought hopefully that is few people, if any)

Issue #1118, opened 2022-10-27
README example command lines seem to be outdated


FrancisRussell:
The README suggests:

```$ cargo run --example tlsclient -- --http mozilla-modern.badssl.com```

and elsewhere

```cargo run --example tlsserver -- --certs test-ca/rsa/end.fullchain --key test-ca/rsa/end.rsa -p 8443 echo```

but I needed to do 

```$ cargo run --bin tlsclient-mio -- --http mozilla-modern.badssl.com```

and 

```cargo run --bin tlsserver-mio -- --certs test-ca/rsa/end.fullchain --key test-ca/rsa/end.rsa -p 8443 echo```

djc:
Right. Would you be able to submit a PR?

FrancisRussell:
https://github.com/rustls/rustls/pull/1119

PR #1119, opened 2022-10-27
Fix issues due to `tlsserver-mio` and `tlsclient-mio` rename

FrancisRussell:
Fixes https://github.com/rustls/rustls/issues/1118.


djc (approved):
Thanks!

PR #1128, opened 2022-11-09
initial openssl vs rustls doc

### LINKED ISSUE #1124: Migration path for the openssl crate
Despite the significant benefits of using rustls, there is still 5.7 million recent [openssl](https://crates.io/crates/openssl) downloads vs 6.5 million of rustls. Migrating all those openssl lib users might be difficult because the API is significantly different.

I would like to propose two things:
* create a list of all common openssl APIs and the corresponding example rustls usages. It is OK for some API not to be implemented in rustls, but they should be clearly documented. Perhaps a wiki page?
* consider implementing/modifying new APIs to provide simpler migration path

nyurik:
This is an initial and possibly incorrect documentation of how to migrate from OpenSSL to Rustls crate. Please help expand it to many other use cases.

Fixes #1124

### DIFF
--- a/OPENSSL.md
+++ b/OPENSSL.md
@@ -1,0 +1,63 @@
+# Migrating from OpenSSL
+Here are some guidelines and hint how to migrate from the [OpenSSL](https://docs.rs/openssl/latest/openssl/).
+
+**This page is incomplete, and can use some help.**
+
+### Create a builder
+```rust
+// OpenSSL
+use openssl::ssl::{SslConnector, SslMethod};
+let mut builder = SslConnector::builder(SslMethod::tls());
+```
+```rust
+// Rustls
+let mut builder = rustls::ClientConfig::builder().with_safe_defaults();
+```
+
+### Add a root CA file
+```rust
+// OpenSSL
+builder.set_ca_file(ca_root_file)?;
+```
+```rust
+// Rustls
+let reader = File::open(ca_root_file)?;
+for cert in rustls_pemfile::certs(&mut BufReader::new(reader))? {
+    roots.add(&rustls::Certificate(cert))?;
+}
+```
+
+### Disable certificate validation (dangerous!)
+```rust
+// OpenSSL
+builder.set_verify(SslVerifyMode::NONE);
+```
+```rust
+// Rustls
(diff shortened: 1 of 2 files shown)


djc:
This seems kinda trivial except for the insecure verifier, which puts us back at square one of semi-recommending insecure workarounds. I'm not sure this adds much value in its current form.

nyurik:
> This seems kinda trivial except for the insecure verifier, which puts us back at square one of semi-recommending insecure workarounds. I'm not sure this adds much value in its current form.

@djc I think the triviality of it is only apparent to the frequent rustls developers. I have spent considerable time figuring it out during my (still incomplete) migration from openssl, and I think it would be really good to help the novice users to avoid all the pain I experienced.  Plus this offers a place for more migration recipes to be added by other contributors. Unless you want to make the migration a bit more difficult that is :)

With regards to the dangerous code: I understand you don't like dangerous and unsafe code. None of us do, and in an ideal world we won't have any of it. But there is not much we can when we port existing code and there is a requirement for such behavior.  We can pretend it doesn't exist, or we can at least provide documentation, and make it just a bit more safe, explaining the risks and how to avoid it.

In this specific case, a well known [Martin tile server](https://github.com/maplibre/martin#using-with-heroku-postgresql) needs to support PostgreSQL providers such as Heroku, which has self-signed certs that has caused a lot of issues for the Martin users. I would love to solve this in a better way of course, but the first step would be to migrate existing functionality without breaking too much.

nyurik:
I just realized I was referring to [Curse of knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge) above.  Hope we can move forward on this and collaboratively help all new devs migrate to Rustls as soon as they possibly can.

cpu:
> This seems kinda trivial except for the insecure verifier, which puts us back at square one of semi-recommending insecure workarounds. I'm not sure this adds much value in its current form.

I feel similar to Djc. I think if there are gaps in the existing documentation we should focus on improvements there. 

This PR has been open for a good while now without any additional community input in favour. I'm going to close this PR but would encourage you to consider improvements to the existing docs if you still believe there are gaps that make it difficult for new users to understand the Rustls API surface.

PR #1132, opened 2022-11-24
doc: add links to the examples in the documentation

usadson:
Hi,

I saw that the examples section of the [docs.rs](https://docs.rs/rustls/latest/rustls/) page were missing, so I added them in this commit.

Tristan


djc (approved):
Good call, thanks!

PR #1134, opened 2022-11-25
Expose `to_trust_anchor` on OwnedTrustAnchor

trevor-crypto:
This would greatly help when implementing a custom verifier that needs access to the root cert held in `RootCertStore`.  The workaround for this is to avoid using `RootCertStore` and just hold the root cert as DER bytes, which is not very clean, in my opinion.

Example:

Suppose we have a custom verify with functions:
```
  fn verify_end_entity_certificate(
        &self,
        end_entity_certificate: &EndEntityCert,
        intermediate_certs: &[Certificate],
        now: DateTime<Utc>,
    ) -> Result<(), webpki::Error> {
        let trust_anchors = self.root_cert_store
            .roots
            .iter()
            .map(|cert| cert.to_trust_anchor()) // this used to be available but was made crate-public
            .collect();
        let time = Time::from_seconds_since_unix_epoch(now.timestamp() as u64);
        let intermediate_certs: Vec<&[u8]> = intermediate_certs
            .iter()
            .map(|cert| cert.0.as_slice())
            .collect();

        end_entity_certificate.verify_is_valid_tls_server_cert(
            SUPPORTED_SIG_ALGS,
            &TlsServerTrustAnchors(&trust_anchors),
            &intermediate_certs,
            time,
        )
    }
```

Here we cannot call `verify_is_valid_tls_server_cert` since we cannot map the roots in `RootCertStore` to `TrustAnchor`s to wrap into `TlsServerTrustAnchors`.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -14,7 +14,7 @@ pub struct OwnedTrustAnchor {
 
 impl OwnedTrustAnchor {
     /// Get a `webpki::TrustAnchor` by borrowing the owned elements.
-    pub(crate) fn to_trust_anchor(&self) -> webpki::TrustAnchor {
+    pub fn to_trust_anchor(&self) -> webpki::TrustAnchor {
         webpki::TrustAnchor {
             subject: &self.subject,
             spki: &self.spki,


djc:
We don't want to do this because we don't want the webpki version we depend on to be part of the public API.

trevor-crypto:
Ok, fair enough

iamwwc:
Wow, same here. Does there have some workaround? I can't call `to_trust_anchor`

djc:
> Wow, same here. Does there have some workaround? I can't call `to_trust_anchor`

What's your use case? Do the [verify_server_cert_signed_by_trust_anchor](https://docs.rs/rustls/latest/rustls/client/fn.verify_server_cert_signed_by_trust_anchor.html) and [verify_server_name](https://docs.rs/rustls/latest/rustls/client/fn.verify_server_name.html) functions help?

iamwwc:
I want to implement custom `ClientCertVerifier` trait

`prepare` call crate only function `to_trust_anchor`, but I can't.

https://docs.rs/rustls/0.20.7/src/rustls/verify.rs.html#546

djc:
Why are you still using 0.20? Did you look at the functions I referenced in my previous comment?

djc:
Ahh, yes. Well, `verify_server_cert_signed_by_trust_anchor()` is actually limited to server certificates, since it passes `KeyUsage::server_auth()` to webpki. But, if you want to submit a PR that adds an analogous `verify_client_cert_signed_by_trust_anchor()` (and modifies the `WebPkiClientCertVerifier` to use it), we could release that in 0.21.x pretty soon to support your use case.

iamwwc:
will do

PR #1140, opened 2022-12-12
Update development dependencies, migrate to 2021 edition

djc:
Since our MSRV is 1.56+ anyway, might as well update to the new edition.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -1,12 +1,11 @@
 [package]
 name = "rustls-examples"
 version = "0.0.1"
-edition = "2018"
+edition = "2021"
 authors = ["Joseph Birr-Pixton <jpixton@gmail.com>"]
 license = "Apache-2.0/ISC/MIT"
 description = "Rustls example code and tests."
 publish = false
-resolver = "2"
 
 [features]
 default = ["logging"]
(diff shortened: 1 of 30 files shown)


ctz (approved):

djc:
@ctz are you okay with migrating away from criterion? (In a commit added after your review.)

ctz:
> @ctz are you okay with migrating away from criterion? (In a commit added after your review.)

Yeah I'm not wedded to criterion and it's quite heavy for what we're needing from it.

djc:
Any clue why this one CI job can't find env_logger 0.10? It doesn't make sense to me.

ctz:
> Any clue why this one CI job can't find env_logger 0.10? It doesn't make sense to me.

Had a quick look at this: env_logger 0.10.0 moved its MSRV to 1.60 in https://github.com/rust-cli/env_logger/commit/660cf7feb27e07bd308c3952637c5ae21c438846 and so our MSRV of 1.57 cannot see it. I guess our option is to bump our MSRV further or stick on env_logger on 0.9 for a while. I prefer probably the latter because env_logger is a pretty minor dev-dependency while increasing MSRV for relatively unimportant reasons has a real cost on downstream users.

djc:
Yup, will fix that up soon.

djc:
Fixed this up, and also bumped the base64 dependency.

PR #1141, opened 2022-12-22
Lazily parse messages from the handshake joiner

djc:
This will help borrowing data from the incoming buffers (and eventually a pass-through API). It removes more lines than it adds, while providing a more precisely typed API for the handshake joiner internally and externally.

### DIFF
--- a/rustls/src/msgs/hsjoiner.rs
+++ b/rustls/src/msgs/hsjoiner.rs
@@ -42,17 +42,18 @@ impl HandshakeJoiner {
         }
     }
 
-    /// Do we want to process this message?
-    pub fn want_message(&self, msg: &PlainMessage) -> bool {
-        msg.typ == ContentType::Handshake
-    }
-
     /// Take the message, and join/split it as needed.
     ///
-    /// Returns an `Err` if a received payload has an advertised size larger than we accept,
-    /// or a `bool` to indicate whether the handshake is "aligned": if the buffer currently
+    /// Returns `Err(JoinerError::Unwanted(msg))` if `msg`'s type is not `ContentType::Handshake` or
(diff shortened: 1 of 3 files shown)


djc:
This is passing all the tests now. Once this is merged I'll start looking into merging the `MessageDeframer` and `HandshakeJoiner`.

djc:
@ctz any blockers for reviewing this?

ctz:
Will look at this first thing tomorrow!

rustls/src/msgs/hsjoiner.rs
```diff
@@ -91,32 +91,26 @@ impl HandshakeJoiner {
             None => return Ok(None),
         };
 
-        let msg = parse_message(&self.buf[..len], self.version)?;
+        let buf = &self.buf[..len];
```
ctz, line 94: AIUI the compiler will do stuff like this for us (inline a single-caller, non-pub fn into its call site). I think having single-caller functions often serves a good documentary/readability purpose ("these lines are for parsing a message from a buffer"; a readability hint that was lost here) and ensures calling functions are a bit smaller and more readable.

ctz (approved):
lgtm!

rustls/src/msgs/hsjoiner.rs
djc: My usual preference is to inline single-caller functions because of the indirection, which I find makes it harder to understand code because you have to skip around to gain context -- as with so many things, it really depends. If the goal is to make it clear what the function of a block of code is, adding some comments can often serve that purpose quite well.

Would you prefer I drop this commit? It's definitely easy to do so? Could also just add a few lines of comments.

djc: Added some comments.

PR #1146, opened 2023-01-06
Move benchmarks to benches directory

### LINKED ISSUE #694: Cargo warns "file found to be present in multiple build targets: rustls/rustls/tests/benchmarks.rs"
Every time I build Rustls using Cargo I get this warning:

```
warning: /home/runner/work/rustls/rustls/rustls/Cargo.toml: file found to be present in multiple build targets: /home/runner/work/rustls/rustls/rustls/tests/benchmarks.rs
```

It would be nice to get a warning-free build.

jbr:
Hi! Here's a PR that fixes #694.

Previously, benchmarks were being built both as a test and as the benchmark target. By moving them out of the tests directory, we resolve this warning.

## Alternative approaches:
* split `rustls/rustls/tests/common/mod.rs` out into its own shared crate. This might be slightly preferred, but seems relatively high ceremony for a single test utility file.
* use a relative symlink to include that file in `rustls/rustls/benches`. I confirmed that this does resolve the warning, but it seems worse than the other two approaches

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -54,7 +54,7 @@ path = "examples/internal/bench.rs"
 
 [[bench]]
 name = "benchmarks"
-path = "tests/benchmarks.rs"
+path = "benches/benchmarks.rs"
 harness = false
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


djc (approved):
LGTM! @ctz was there any particular reason to put the benchmark in `tests`?

ctz:
> LGTM! @ctz was there any particular reason to put the benchmark in `tests`?

Solely the dependency on tests/common; and lack of knowledge that you could introduce a module with a specific path!

ctz (approved):

PR #1149, opened 2023-01-10
Fixing TLS version typo 1_2 -> 1_3

SimonHeinze:
I do not understand enough of the code to be absolutely sure, but this looks an awful lot like a typo to me. If it is, it would be nice to have some test that would have detected it earlier. If it aint it would be really nice to have a comment at this line explaining the choice.
Thanks a lot for your time, to whoever is reading this! :)

### DIFF
--- a/rustls/src/tls13/mod.rs
+++ b/rustls/src/tls13/mod.rs
@@ -181,7 +181,7 @@ impl MessageEncrypter for Tls13MessageEncrypter {
 
         Ok(OpaqueMessage {
             typ: ContentType::ApplicationData,
-            version: ProtocolVersion::TLSv1_2,
+            version: ProtocolVersion::TLSv1_3,
             payload: Payload::new(payload),
         })
     }


djc:
If you read the [RFC section](https://datatracker.ietf.org/doc/html/rfc8446#section-5.2) linked above, you'll see that this is actually correct. It's bit of an ugly workaround for the fact that middleboxes (proxies and intercepting devices) started relying on TLS 1.2 versioning, so it was impossible to change the version number in the unencrypted bits. Could definitely do with a comment that clarifies that this is correct, though!

ctz:
https://datatracker.ietf.org/doc/html/rfc8446#section-5.1: 
```
   legacy_record_version:  MUST be set to 0x0303 for all records
      generated by a TLS 1.3 implementation other than an initial
      ClientHello (i.e., one not generated after a HelloRetryRequest),
      where it MAY also be 0x0301 for compatibility purposes.  This
      field is deprecated and MUST be ignored for all purposes.
      Previous versions of TLS would use other values in this field
      under some circumstances.
```
(0x0303 is TLS1.2)

SimonHeinze:
I see, thank you so much for the explanation!

PR #1150, opened 2023-01-10
TLS 1.3: Immediately update keys when requested

### LINKED ISSUE #938: Key rotation on key update should be done immediately, not lazily
The `ExpectTraffic` state has a field `want_write_key_update` that remembers whether we received a key update message where the peer requested we update our write key. IIUC, the `want_write_key_update` field is not read until the next time we try to encrypt some data and send it. If it is a long time before we attempt to send data, then the old write key will live for a long time. This is counter to the goal of the key update feature.

Instead, as I mentioned in https://github.com/rustls/rustls/issues/288#issuecomment-1014093501, we should instead do something like this rough sketch:
* Replace `want_write_key_update: bool` with `enqueued_key_update_message: Vec<u8>`.
* When we receive a request to update our write key, and `enqueued_key_update_message` is non-empty, then do nothing.
* When we receive a request to update our write key, and `enqueued_key_update_message` is empty, immediately format and encrypt the key update message we'd need to send, and store it in `enqueued_key_update_message`. Then immediately update the write key and the key schedule.
* When sending data, where we check `want_write_key_update`, instead check if `enqueued_key_update_message` is non-empty. If it is non-empty, then send that enqueued message and reset it to empty, before trying to send any more data.
* When the application checks whether we have anything to send, consider `enqueued_key_update_message`. We need to send the key update message eagerly so that the peer can update its receive key in a timely fashion.

Additionally, when sending an alert (whether close_notify or an error), we should either update or throw away the write key and write part of the key schedule, as they are no longer needed, and having them around is bad. Similarly, when we receive a close_notify from the peer, we should throw away the receive key and read part of the key schedule. We might consider doing these in follow-ups.

jbr:
Hopefully will close #938 after expected review iterations.

Notes:

* Is the existing behavior tested anywhere? I couldn't figure out how to get a failing test
* I attempted to follow the suggested approach closely, but opted for an `Option<Vec<u8>>` instead of a possibly-zero-length `Vec<u8>`
* I was not at all sure if `common.sendable_tls.append(message)` is the appropriate way to send a pre-encrypted message, but it seemed like a reasonable choice.

Still outstanding, but possible candidates for subsequent PR:
* Dropping write key when a close_notify or error is sent
* Dropping read key when close_notify is received

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -889,7 +889,6 @@ impl State<ClientConnectionData> for ExpectFinished {
             suite: st.suite,
             transcript: st.transcript,
             key_schedule: key_schedule_traffic,
-            want_write_key_update: false,
             _cert_verified: st.cert_verified,
             _sig_verified: st.sig_verified,
             _fin_verified: fin,
@@ -913,7 +912,6 @@ struct ExpectTraffic {
     suite: &'static Tls13CipherSuite,
     transcript: HandshakeHash,
     key_schedule: KeyScheduleTraffic,
-    want_write_key_update: bool,
     _cert_verified: verify::ServerCertVerified,
     _sig_verified: verify::HandshakeSignatureValid,
     _fin_verified: verify::FinishedMessageVerified,
@@ -983,7 +981,7 @@ impl ExpectTraffic {
     fn handle_key_update(
         &mut self,
         common: &mut CommonState,
-        kur: &KeyUpdateRequest,
+        key_update_request: &KeyUpdateRequest,
     ) -> Result<(), Error> {
         #[cfg(feature = "quic")]
         {
@@ -997,15 +995,9 @@ impl ExpectTraffic {
         // Mustn't be interleaved with other handshake messages.
         common.check_aligned_handshake()?;
 
-        match kur {
-            KeyUpdateRequest::UpdateNotRequested => {}
-            KeyUpdateRequest::UpdateRequested => {
-                self.want_write_key_update = true;
-            }
-            _ => {
-                common.send_fatal_alert(AlertDescription::IllegalParameter);
-                return Err(Error::CorruptMessagePayload(ContentType::Handshake));
-            }
+        if common.should_update_key(key_update_request)? {
+            self.key_schedule
+                .update_encrypter_and_notify(common);
         }
 
         // Update our read-side keys.
@@ -1057,14 +1049,6 @@ impl State<ClientConnectionData> for ExpectTraffic {
     ) -> Result<(), Error> {
         self.key_schedule
             .export_keying_material(output, label, context)
-    }
-
-    fn perhaps_write_key_update(&mut self, common: &mut CommonState) {
-        if self.want_write_key_update {
-            self.want_write_key_update = false;
-            self.key_schedule
-                .update_encrypter_and_notify(common);
-        }
     }
 
     #[cfg(feature = "secret_extraction")]
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -6,8 +6,8 @@ use crate::log::{debug, error, trace, warn};
 use crate::msgs::alert::AlertMessagePayload;
 use crate::msgs::base::Payload;
 use crate::msgs::deframer::{Deframed, MessageDeframer};
-use crate::msgs::enums::HandshakeType;
 use crate::msgs::enums::{AlertDescription, AlertLevel, ContentType};
+use crate::msgs::enums::{HandshakeType, KeyUpdateRequest};
 use crate::msgs::fragmenter::MessageFragmenter;
 use crate::msgs::handshake::Random;
 use crate::msgs::message::{
@@ -694,9 +694,8 @@ impl<Data> ConnectionCommon<Data> {
     }
 
     pub(crate) fn send_some_plaintext(&mut self, buf: &[u8]) -> usize {
-        if let Ok(st) = &mut self.state {
-            st.perhaps_write_key_update(&mut self.common_state);
-        }
+        self.common_state
+            .perhaps_write_key_update();
         self.common_state
             .send_some_plaintext(buf)
     }
@@ -823,6 +822,8 @@ pub struct CommonState {
     received_plaintext: ChunkVecBuffer,
     sendable_plaintext: ChunkVecBuffer,
     pub(crate) sendable_tls: ChunkVecBuffer,
+    queued_key_update_message: Option<Vec<u8>>,
+
     #[allow(dead_code)] // only read for QUIC
     /// Protocol whose key schedule should be used. Unused for TLS < 1.3.
     pub(crate) protocol: Protocol,
@@ -853,6 +854,7 @@ impl CommonState {
             received_plaintext: ChunkVecBuffer::new(Some(DEFAULT_RECEIVED_PLAINTEXT_LIMIT)),
             sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
             sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
+            queued_key_update_message: None,
 
             protocol: Protocol::Tcp,
             #[cfg(feature = "quic")]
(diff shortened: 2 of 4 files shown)


jbr:
Is the `base64` failure unrelated?

djc (commented):
Looks like a good start!

This handles key updates on the server side, but not on the client side. I think we should do both? Might make sense to see if we can factor out common logic (in a separate commit).

I think we need to call `State::perhaps_write_key_update()` in `CommonState::wants_write()` before checking the contents of `self.sendable_tls`.

jbr:
@djc 
> I think we need to call State::perhaps_write_key_update() in CommonState::wants_write() before checking the contents of self.sendable_tls (let's do this in a separate commit as well).

When I tried this, KeyUpdate-RequestACK and KeyUpdate-RequestACK-UnfinishedWrite failed in the BoGo suite. But the suite passes without that change

djc:
I'd like a cleaner commit history:

* One logical change per commit (similar changes in client and server in the same commit), so maybe one for deduplicating client/server (first), and one for eager updates -- no intermediate commits for stylistic changes
* For terminology stick to existing names ("rotation" -> "update")
* Clean commit messages ("dry up" sounds like colloquialized jargon to me, I would write that more like "Deduplicate key update handling code")

`common.sendable_tls.append(message)` seems fine for sending pre-encrypted messages.

jbr:
It seems worth drawing attention to the fact that f33e06a moved the state into CommonState. There is a small memory cost (`Option<Vec<u8>>`, 24 bytes) for non-ExpectTraffic states. I don't know if that's acceptable for rustls users; figured I'd revert that commit if not. I do believe it improves the code.

jbr:
@djc Rebased on #1167, this ended up being a fairly small change. Squashed into a single commit. I still have the (original branch)https://github.com/jbr/rustls/tree/immediate-key-rotation-original) if #1167 doesn't end up getting merged.

djc (approved):
Nice!

djc:
@jbr want to rebase this on main?

jbr:
Was literally just looking at that 👍

rustls/src/client/tls13.rs
```diff
@@ -997,15 +995,9 @@ impl ExpectTraffic {
...
-        match kur {
-            KeyUpdateRequest::UpdateNotRequested => {}
-            KeyUpdateRequest::UpdateRequested => {
-                self.want_write_key_update = true;
-            }
-            _ => {
-                common.send_fatal_alert(AlertDescription::IllegalParameter);
-                return Err(Error::CorruptMessagePayload(ContentType::Handshake));
-            }
+        if common.should_update_key(key_update_request)? {
```
djc, line 998: Could we turn `common.should_update_key()` into `key_schedule.maybe_update_key()`, so this interface becomes simpler?

rustls/src/conn.rs
```diff
@@ -1316,6 +1318,35 @@ impl CommonState {
...
+            _ => {
+                self.send_fatal_alert(AlertDescription::IllegalParameter);
+                Err(Error::CorruptMessagePayload(ContentType::Handshake))
+            }
+        }
+    }
+
+    pub(crate) fn enqueue_key_update_notification(&mut self) {
+        let message = PlainMessage::from(Message::build_key_update_notify());
+        self.queued_key_update_message = Some(
```
djc, line 1338: Why do we have the indirection through `self.queued_key_update_message`? Why not `append()` to `self.sendable_tls` directly here?

ctz:
```
   If the request_update field is set to "update_requested", then the
   receiver MUST send a KeyUpdate of its own with request_update set to
   "update_not_requested" prior to sending its next Application Data
   record.  This mechanism allows either side to force an update to the
   entire connection, but causes an implementation which receives
   multiple KeyUpdates while it is silent to respond with a single
   update.  Note that implementations may receive an arbitrary number of
   messages between sending a KeyUpdate with request_update set to
   "update_requested" and receiving the peer's KeyUpdate, because those
   messages may already be in flight.  However, because send and receive
   keys are derived from independent traffic secrets, retaining the
   receive traffic secret does not threaten the forward secrecy of data
   sent before the sender changed keys.
```

I think it's the "causes an implementation which receives multiple KeyUpdates while it is silent to respond with a single update" clause that we're trying to do here.

djc: Ahh, so we only want to respond to `KeyUpdateRequest`s once and only send the `KeyUpdate` if we have something else to send, too. So would be good to add some comments to clarify this.

rustls/src/client/tls13.rs
jbr: I'm genuinely not sure if that's better, but it's slightly fewer lines and less duplication. It feels like there's something larger that isn't modeled quite right here, in that the _combination_ of common and key_schedule are needed for a bunch of behavior, and putting the logic on either of them ends up reaching into the other in a way that doesn't _feel_ quite right.

rustls/src/conn.rs
jbr: I think that's the intent of `// Mustn't be interleaved with other handshake messages.`, which was present in one call site but not the other, and I'll duplicate. There is, however, a bogo test that fails if we just directly enqueue the message without keeping track of that state

rustls/src/client/tls13.rs
djc: I think I understand what you're getting at... At the same time, the `key_schedule` in this case is holding the valuable cryptographic secrets, so IMO giving it control over other parts (in particular, the `CommonState`, which is like a fairly promiscuous part of the connection that is accessed from all over the place) doesn't feel too crazy to me. (On the other hand, making another thing that gets access to the `key_schedule` feels relatively scary.)

jbr: I'm curious what you think about the version I'm about to push. After a few shufflings in response to this feedback, I ended up moving it onto CommonState and feel relatively good about that.

jbr: changes in 975a8f8c

djc: Yeah, sorry -- not such a fan of giving `CommonState` `&mut` access to the `key_schedule`. I think the design is set up to keep the `key_schedule` private within a mostly opaque state for a reason.

jbr: Will happily invert the caller here. Is the underlying concern that malicious developers may make later changes misusing that `pub(crate)` `&mut` access? Or are there memory-safety concerns here that I should read up on?

jbr: Any caller of this code necessarily must have `&mut` for both CommonState and KeyScheduleTraffic, so it doesn't seem like it makes any safety difference where the associated function is defined?

djc: No, there are no memory-safety concerns, and I think all of this is in the end maybe more a matter of taste than strict rules. However I think one of the design guidelines in *ring* and rustls is to minimize how much code gets access to secret values, and so I feel like this is in line with that idea.

(We could keep this open for a bit to see if @ctz has an opinion.)

rustls/src/conn.rs
jbr: Added a comment, would appreciate feedback on accuracy and clarity

djc: I think deduplication is one thing but I would probably also mention that we're trying to avoiding sending a key update notification if there isn't anything else to send.

rustls/src/client/tls13.rs
ctz: On 975a8f8c3 I have some probably-unhelpful thoughts:

- `KeySchedule` is being passed `CommonState` so (mainly) it can manipulate the record layer. I don't think it should use its `CommonState` reference for sending messages or much else really. Ideally I'd hope that tls13/key_schedule.rs is limited to the computations in https://www.rfc-editor.org/rfc/rfc8446#section-7.1
- When we're hoisting up identical code from client/server files, I think ideally that should go in `CommonState` as a first step. But I think the readability/future-defect improvement that can be realised from this means deduplicating tens of lines of code. There's not likely to be a code size improvement because I think it's unlikely that consumers who care about code size compile in both the client and server code into the same application.
- It's not ideal to expose `KeySchedule*` outside the state machine objects that own it, but I think in this case we could expose a closure on `key_schedule.update_encrypter`?

djc: It's a little hard to get from those thoughts to a concrete design. I would propose we just go back to keeping `should_update_key()` and `update_encrypter_and_notify()` separate, and we can also move the call to `common.enqueue_key_update_notification()` back into the state machine states.

In this case I agree the readability/future-defect improvement isn't huge, although I think there are benefits downstream of deduplicating/sharing logic in a single place that you might be underestimating here (I already didn't think code size was a factor here).

I think a closure would be a little to express in the type system and I generally find that the indirection caused by closures usually makes code harder to understand.

ctz (approved):

djc:
(I've removed the later commits that revised the API, as discussed above.)

PR #1151, opened 2023-01-10
use saturating add in ticket expiration

### LINKED ISSUE #659: Potential unchecked integer overflows in ticket expiration logics
Consider:
```
            state.next_switch_time = now + u64::from(self.lifetime);
```

If `self.lifetime` is very large then the addition can overflow. If we intend for it to be wrapping then we should make that explicit. Otherwise, we should change this arithmetic to be checked.

Note that the above line is just an example. There is more unchecked arithmetic.

jbr:
closes #659


jbr:
build failure due to #1152

djc (approved):

djc:
LGTM, but let's hold off on merging until we can get a full CI run.

PR #1152, opened 2023-01-10
update github actions msrv to 1.57 so that base64 will build

jbr:
#1148 made CI red on main. This is the minimum change from #1140 that is required to make CI green

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.56 # MSRV
+          - 1.57 # MSRV
         os: [ubuntu-20.04]
         # but only stable on macos/windows (slower platforms)
         include:
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -2,7 +2,7 @@
 name = "rustls"
 version = "0.20.7"
 edition = "2018"
-rust-version = "1.56"
+rust-version = "1.57"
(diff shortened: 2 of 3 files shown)


djc:
#1148 is closed, so that was not the culprit -- rather, we're now pulling in base64 0.21 through rustls-pemfile.

.github/workflows/build.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.56 # MSRV
+          - 1.57 # MSRV
```
djc, line 23: Did you check for other mentions of the MSRV? We should at least update the `rust-version` in the Cargo metadata.

jbr: I wasn't sure how much of #1140 we wanted to duplicate. Will add rust-version and the readme comment as well.

rustls/Cargo.toml
```diff
@@ -21,6 +21,7 @@ log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 sct = "0.7.0"
 webpki = { version = "0.22.0", features = ["alloc", "std"] }
+zeroize = { version = "1.5.7", features = ["derive"] }
```
djc, line 24: This seems unrelated?

djc (approved):

rustls/Cargo.toml
jbr: entirely unrelated! 😬

djc (approved):

PR #1156, opened 2023-01-11
move `doc(cfg(feature = …))` to original definitions

### LINKED ISSUE #1155: Show feature requirements in documentation
Rustls has a number of items, like [ClientCertVerified](https://docs.rs/rustls/latest/rustls/server/struct.ClientCertVerified.html), that are available only with certain cargo features enabled (in this case, `dangerous_configuration`). These should get the "Available on crate feature `dangerous_configuration` only" annotation.

We do have some code that tries to get that annotation in place, specifically https://github.com/rustls/rustls/blob/be7b159ca66167fd9a0c20c3808c9a4d3a61ed31/rustls/src/lib.rs#L453-L455

And the corresponding code in Cargo.toml:

https://github.com/rustls/rustls/blob/be7b159ca66167fd9a0c20c3808c9a4d3a61ed31/rustls/Cargo.toml#L60-L62

However, in the end the docs that appear on docs.rs don't have the feature annotation. We should figure out why and fix it.

jbr:
it seems like reexports don't pick those annotations up. closes #1155

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -320,6 +320,7 @@ pub(super) mod danger {
 
     /// Accessor for dangerous configuration options.
     #[derive(Debug)]
+    #[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
     pub struct DangerousClientConfig<'a> {
         /// The underlying ClientConfig
         pub cfg: &'a mut ClientConfig,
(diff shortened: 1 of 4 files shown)


jbr:
I believe the macos build+test failure is spurious (network failure)

jbr:
cc @jsha for review (can't request for review)

jsha (approved):
Looks good! Thanks for making the change. For reference, I verified this with:

    cargo doc --open --no-deps --all-features --config "build.rustdocflags=\"--cfg docsrs\""

PR #1157, opened 2023-01-11
add a README for BoGo

jbr:
Is there anything else we'd want to include?


djc (approved):
Seems like a good start!

PR #1159, opened 2023-01-12
Reworked export_keying_material interface to be safer for the end user

davidv1992:
This changes the export_keying_material in such that ownership of the key buffer can pass through the export_keying_material function. Doing this ensures that the key buffer is dropped on error and cannot accidentally be used when export_keying_material fails.

The disadvantage is that this requires the two exported variants of export_keying_material to be generic, incurring a small compile duration penalty.

Also note that this is an incompatible change to the public api of rustls.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -79,15 +79,14 @@ impl Connection {
     /// Derives key material from the agreed connection secrets.
     ///
     /// See [`ConnectionCommon::export_keying_material()`] for more information.
-    pub fn export_keying_material(
+    pub fn export_keying_material<const N: usize>(
         &self,
-        output: &mut [u8],
         label: &[u8],
         context: Option<&[u8]>,
-    ) -> Result<(), Error> {
+    ) -> Result<[u8; N], Error> {
         match self {
-            Self::Client(conn) => conn.export_keying_material(output, label, context),
-            Self::Server(conn) => conn.export_keying_material(output, label, context),
+            Self::Client(conn) => conn.export_keying_material(label, context),
+            Self::Server(conn) => conn.export_keying_material(label, context),
         }
     }
 
@@ -748,14 +747,16 @@ impl<Data> ConnectionCommon<Data> {
     ///
     /// This function fails if called prior to the handshake completing;
     /// check with [`CommonState::is_handshaking`] first.
-    pub fn export_keying_material(
+    pub fn export_keying_material<const N: usize>(
         &self,
-        output: &mut [u8],
         label: &[u8],
         context: Option<&[u8]>,
-    ) -> Result<(), Error> {
+    ) -> Result<[u8; N], Error> {
+        let mut output = [0; N];
         match self.state.as_ref() {
-            Ok(st) => st.export_keying_material(output, label, context),
+            Ok(st) => st
+                .export_keying_material(&mut output, label, context)
+                .map(|_| output),
             Err(e) => Err(e.clone()),
         }
     }
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -730,22 +730,39 @@ fn exec(opts: &Options, mut sess: Connection, count: usize) {
         }
 
         if !sess.is_handshaking() && opts.export_keying_material > 0 && !sent_exporter {
-            let mut export = Vec::new();
-            export.resize(opts.export_keying_material, 0u8);
-            sess.export_keying_material(
-                &mut export,
-                opts.export_keying_material_label
-                    .as_bytes(),
-                if opts.export_keying_material_context_used {
-                    Some(
-                        opts.export_keying_material_context
+            let export = match opts.export_keying_material {
+                1 => sess
+                    .export_keying_material::<1>(
+                        opts.export_keying_material_label
                             .as_bytes(),
+                        if opts.export_keying_material_context_used {
+                            Some(
+                                opts.export_keying_material_context
+                                    .as_bytes(),
+                            )
+                        } else {
+                            None
+                        },
                     )
-                } else {
-                    None
-                },
-            )
-            .unwrap();
+                    .unwrap()
+                    .to_vec(),
+                1024 => sess
+                    .export_keying_material::<1024>(
+                        opts.export_keying_material_label
+                            .as_bytes(),
+                        if opts.export_keying_material_context_used {
+                            Some(
+                                opts.export_keying_material_context
+                                    .as_bytes(),
+                            )
+                        } else {
+                            None
+                        },
+                    )
+                    .unwrap()
+                    .to_vec(),
+                _ => panic!("Unsupported export length {}", opts.export_keying_material),
+            };
             sess.writer()
                 .write_all(&export)
                 .unwrap();
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -2016,40 +2016,34 @@ fn sni_resolver_rejects_bad_certs() {
 }
 
 fn do_exporter_test(client_config: ClientConfig, server_config: ServerConfig) {
-    let mut client_secret = [0u8; 64];
-    let mut server_secret = [0u8; 64];
-
     let (mut client, mut server) = make_pair_for_configs(client_config, server_config);
 
     assert_debug_eq(
-        client.export_keying_material(&mut client_secret, b"label", Some(b"context")),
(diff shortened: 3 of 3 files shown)


ctz:
Is it possible to make this work with the previous `&mut [u8]` type?

> cannot accidentally be used when export_keying_material fails.

I'd think that, thanks to `Result` being `#[must_use]`, ignoring errors from `export_keying_material` must be pretty explicit and deliberate?

davidv1992:
> Is it possible to make this work with the previous `&mut [u8]` type?

I think I might not fully understand what you mean here, but `&mut [u8]` satisfies the `AsMut<[u8]>` constraint and so can still be used as an argument

> 
> I'd think that, thanks to `Result` being `#[must_use]`, ignoring errors from `export_keying_material` must be pretty explicit and deliberate?

Not necessarily. As an example, consider the following code:
```
struct TlsUsingClient {
    tls_connection: rustls::ClientConnection,
    // Other data
}

type KeyType = [u8;32]; 

impl TlsUsingClient {
    fn get_negotiated_key(&self) -> Result<KeyType, rustls::Error> {
        let result = KeyType::default();
        if let Err(e) = self.tls_connection.export_keying_material(&mut result, b"label", None) {
            error!(error = ?e, "Key not yet negotiated")
        }
        Ok(result)
    }

    // other functionality using the tls connection
}
```
This is totally reasonable for someone to write right now, and there is no compiler warning or clippy message that tells them they have just placed a potential security landmine in their code.

With the patch suggested here, the above is still possible, but someone who wants to do better can use a non-copyable key type, and by passing that into export_keying_material they then get an error if the forget to properly do error checking, so something like this now gives compile errors:
```
struct TlsUsingClient {
    tls_connection: rustls::ClientConnection,
    // Other data
}

type KeyType = [u8;32]; 

impl TlsUsingClient {
    fn get_negotiated_key(&self) -> Result<KeyType, rustls::Error> {
        let result = KeyType::default();
        if let Err(e) = self.tls_connection.export_keying_material(result, b"label", None) {
            error!(error = ?e, "Key not yet negotiated")
        }
        Ok(result)
    }

    // other functionality using the tls connection
}
```

which will hopefully steer them to the better solution:
```
struct TlsUsingClient {
    tls_connection: rustls::ClientConnection,
    // Other data
}

type KeyType = [u8;32]; 

impl TlsUsingClient {
    fn get_negotiated_key(&self) -> Result<KeyType, rustls::Error> {
        match self.tls_connection.export_keying_material(KeyType::default()) {
            Err(e) => {
                error!(error = ?e, "key not yet negotiated";
                Err(e)
            }
            v => v,
    }

    // other functionality using the tls connection
}
```
which is not easily by accident changed into something that accidentally returns `Ok(KeyType::default())`

rustls/src/conn.rs
```diff
@@ -81,12 +81,12 @@ impl Connection {
     /// Derives key material from the agreed connection secrets.
     ///
     /// See [`ConnectionCommon::export_keying_material()`] for more information.
-    pub fn export_keying_material(
+    pub fn export_keying_material<T: AsMut<[u8]>>(
         &self,
-        output: &mut [u8],
+        output: T,
```
djc, line 86: Why can't we just stick to `&mut [u8]` as input and `&[u8]` as output?

davidv1992: We could, but then this won't fix the underlying risk factor in most usecases, particularly when follow on crypto libraries want types other than &[u8] as input (like for example RustCrypto, which wants GenericArrays everywhere).

djc: How about the caller is allowed to move in an `[u8; N]` with `const N: usize`, and we'll yield the same array?

ctz:
> I think I might not fully understand what you mean here, but &mut [u8] satisfies the AsMut<[u8]> constraint and so can still be used as an argument

I think this means we can drop the test changes from the PR?

> Not necessarily. As an example, consider the following code:

Thanks for outlining that, the hazard makes sense to me.

I guess I'd point out that this is common in rust code (most notably and commonly, [https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read](std::io::Read::read)!). And, as such, it's unidiomatic and a losing battle for rustls to be fixing issues like this piecemeal.

What would you think about `export_keying_material_buf(&mut self, mut output: std::io::BorrowedCursor)` under the existing `read_buf` feature? A user of such an API is invited, I think, to accurately account for which variables of theirs are fully initialised and avoid premature initialisation (in this case, to zeroes) which causes the hazard here.

rustls/src/conn.rs
djc: I guess this API doesn't see a lot of use, so breaking it (in a way that the compiler will detect) doesn't seem like a big deal to me.

djc:
@ctz `BorrowedCursor` would be great if it were stable, but I think that will still take a while. As it is, I think nightly-only APIs are fairly niche, so I feel like array moving might be a better option at this time. Do you see any downsides there?

rustls/src/conn.rs
davidv1992: @djc, thinking a bit more about it, we don't really need to move in the [u8;N]. The functions is a const generic in N either way. How do you feel about that.

djc: Not sure how you think that changes things?

davidv1992: For large returns, most calling conventions use a caller-allocated buffer, and I think (but not 100% sure) its the same for large parameters, so under the hood with two [u8;N] (one passed in, one passed out) there are potentially two buffers needed at the caller, because it might not be able to reason about whether both are live at the same time, which makes the memcpy reduction somewhat tricky. With just the return buffer, this becomes simpler for the compiler and therefore hopefully more likely to optimize any memcpys out. Plus as an input parameter the [u8;N] doesn't really do anything, in the sense that any value it contains is pretty much ignored (or at least, it should be).

djc: Ahh, so you're suggesting we pass in a `const N: usize` parameter but no buffer, and then the function can yield the `[u8; N]` as a return value? Makes sense to me!

davidv1992: Reworked it to that. Apologies this took a while, been a bit swamped with other work.

djc (approved):
Still some conflicts, mind rebasing on top of current main?

davidv1992:
> Still some conflicts, mind rebasing on top of current main?

done

djc:
Not passing CI yet. @ctz since you proposed a different design, would like your signoff (see [reasoning](https://github.com/rustls/rustls/pull/1159#issuecomment-1385344691) in earlier comment).

davidv1992:
@djc I fixed the bogo shim, it does show that this design doesn't really support exporting arbitrary number of bytes. You might want to take a look at that to check you are comfortable with this.

ctz:
Hm, yes, that's not ideal. We'd need the old API back to permit uses where the length is only known at runtime (example: implementing openssl `SSL_export_keying_material` FFI, or [implementing a common API over multiple TLS libraries](https://docs.rs/openssl/latest/openssl/ssl/struct.SslRef.html#method.export_keying_material)). As a result I feel like this isn't reaching a good bar for usefulness versus cost -- can we revisit this when the new APIs for dealing with uninitialised memory are stabilised?

djc:
@davidv1992 sorry for the churn on this.

I wonder if this there's still something to be done here... Something like this?

```rust
fn ekm<'a>(input: &'a mut impl AsMut<[u8]>) -> Result<impl AsRef<[u8]> + 'a, ()> {
    let buf = input.as_mut();
    for b in buf.iter_mut() {
        *b = 0;
    }
    
    Ok(buf)
}
```

cpu:
@davidv1992 gentle ping for this branch. Do you think you'll have a chance to revisit based on the feedback above? I think this PR is one selected for the next release.

cpu:
Since we haven't heard back from David in a little while I took a crack at reviving this branch in a separate PR based on @djc's most recent feedback. PTAL at https://github.com/rustls/rustls/pull/1211

As a non-maintainer I wasn't able to push commits to this branch, and I couldn't find a nice way to cherry pick https://github.com/rustls/rustls/commit/364b8945acad8907a0ebd7cdb5aa82caef322815 into a new branch from a fork. I hope I haven't stepped on anyone's toes. I'd be happy to close my PR if the original author has time to iterate on this one :bow: I think it's a nice change.

davidv1992:
Apologies for the late reaction, I have been (and to some extent still am) very busy at the moment, and this just isn't high enough on the priority list. Feel free to continue with your copy of the branch @cpu

PR #1167, opened 2023-01-13
TLS 1.3: update record layer directly from key schedule

djc:
Lets key schedule types provide a higher-level API that handles the updating of record layer encrypters/decrypters, moving complexity from the handshake states into the key schedule. Prevents secrets from being passed around outside of the `key_schedule` module, and avoids several secret clones.

cc @jbr

### DIFF
--- a/rustls/src/tls13/key_schedule.rs
+++ b/rustls/src/tls13/key_schedule.rs
@@ -2,6 +2,7 @@ use crate::cipher::{Iv, IvLen, MessageDecrypter};
 use crate::conn::{CommonState, Side};
 use crate::error::Error;
 use crate::msgs::base::PayloadU8;
+use crate::msgs::message::Message;
 #[cfg(feature = "quic")]
 use crate::quic;
 #[cfg(feature = "secret_extraction")]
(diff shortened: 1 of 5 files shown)


djc:
Rebased.

djc:
We might be able to put in place a similar API/strategy for TLS 1.2, too.

rustls/src/tls13/key_schedule.rs
```diff
@@ -463,6 +464,7 @@ impl KeyScheduleTraffic {
 
     pub(crate) fn update_encrypter(&mut self, common: &mut CommonState) {
```
ctz, line 465: Given this one now sends a message (which makes it uniquely specific to `KeyUpdate` usage) maybe we could vary its naming to break the implied symmetry with `update_decrypter`? Perhaps `update_encrypter_for_key_update`?

djc: Renamed it to `update_encrypter_and_notify()`, what do you think?

ctz: Sounds good!

ctz (approved):

PR #1171, opened 2023-01-14
In-place handshake joining

djc:
Attempt at fixing #794 by moving handshake joining into the deframer.

<details>
<summary>Hidden question about previously failing bogo tests.</summary>

This seems to pass the rustls test suite, but currently fails a bunch of bogo tests in a way that I can't figure out. For example:

```
djc-2021 in-place-joining bogo $ ./runme -test EmptyEncryptedExtensions-TLS13
+ '[' x = x ']'
+ cargo build --example bogo_shim --features dangerous_configuration,quic
   Compiling rustls v0.21.0-alpha.1 (/Users/djc/src/rustls/rustls)
    Finished dev [unoptimized + debuginfo] target(s) in 1.02s
+ '[' '!' -e bogo/ssl/test/runner/runner.test ']'
+ case $OSTYPE in
+ set +e
+ cd bogo/ssl/test/runner
+ ./runner.test -shim-path ../../../../../target/debug/examples/bogo_shim -shim-config ../../../../config.json -pipe -allow-unimplemented -test.timeout 60s -test EmptyEncryptedExtensions-TLS13
FAILED (EmptyEncryptedExtensions-TLS13)
bad error (wanted ':BAD_HANDSHAKE_MSG:' / 'remote error: error decoding message'): local error 'read tcp [::1]:50945->[::1]:50946: read: connection reset by peer', child error 'none', stdout:
options: ["-port", "50945", "-handshaker-path", "../../../build/ssl/test/handshaker"]
not checking -handshaker-path ../../../build/ssl/test/handshaker; NYI
opts Options { port: 50945, server: false, max_fragment: None, resumes: 0, verify_peer: false, require_any_client_cert: false, offer_no_client_cas: false, tickets: true, resume_with_tickets_disabled: false, queue_data: false, queue_data_on_resume: false, only_write_one_byte_after_handshake: false, only_write_one_byte_after_handshake_on_resume: false, shut_down_after_handshake: false, check_close_notify: false, host_name: "example.com", use_sni: false, send_sct: false, key_file: "", cert_file: "", protocols: [], support_tls13: true, support_tls12: true, min_version: None, max_version: None, server_ocsp_response: [], server_sct_list: [], use_signing_scheme: 0, curves: None, export_keying_material: 0, export_keying_material_label: "", export_keying_material_context: "", export_keying_material_context_used: false, read_size: 512, quic_transport_params: [], expect_quic_transport_params: [], enable_early_data: false, expect_ticket_supports_early_data: false, expect_accept_early_data: false, expect_reject_early_data: false, expect_version: 0, resumption_delay: 0 }
TLS error: CorruptMessagePayload(Handshake)

stderr:
:BAD_HANDSHAKE_MSG:
```

Note that it is `wanted ':BAD_HANDSHAKE_MSG:'` and stderr is `:BAD_HANDSHAKE_MSG:` but the test still fails. I can't figure out why this happens. @ctz got any clues?
</details>


rustls/src/msgs/deframer.rs
```diff
@@ -96,6 +98,8 @@ impl MessageDeframer {
         // we do a zero length read.  That looks like an EOF to
         // the next layer up, which is fine.
         debug_assert!(self.used <= OpaqueMessage::MAX_WIRE_SIZE);
+        let need_capacity = Ord::min(OpaqueMessage::MAX_WIRE_SIZE, self.used + READ_SIZE);
```
djc, line 260: With this change we no longer read the largest amount possible that fits in the `OpaqueMessage::MAX_WIRE_SIZE` buffer; instead we read at most 4096 bytes at a time. This is because we'd like to not pre-emptively allocate the large buffer if it ends up not being needed (see #794) and 4096 bytes seems to be enough for most common cases. We could read multiple times on a full read, but that would make the return type more complicated, since we'd have to yield both the amount of bytes read and a potential error.

rustls/src/msgs/deframer.rs
```diff
@@ -1,99 +1,248 @@
...
 pub struct MessageDeframer {
-    /// Completed frames for output.
-    frames: VecDeque<OpaqueMessage>,
-
     /// Set to true if the peer is not talking TLS, but some other
     /// protocol.  The caller should abort the connection, because
     /// the deframer cannot recover.
     desynced: bool,
 
     /// A fixed-size buffer containing the currently-accumulating
```
larsfillmore, line 23: The "fixed-size" part is now obsolete.

rustls/src/msgs/deframer.rs
djc: One future tweak I had in mind is to reset the buffer size after the handshake is done, which should be responsible for most of the larger messages. But even now I'm seeing that a real-world handshake probably doesn't need to grow the buffer.

We also technically should allow the buffer to grow larger during the handshake (previously I think the joiner allowed up to 64k bytes).

ctz:
> I can't figure out why this happens. @ctz got any clues?

looks like we've stopped sending alerts in quite a number of cases involving invalid handshake messages. to read that bogo error:

```
bad error (wanted ':BAD_HANDSHAKE_MSG:'
```

wanted to see `:BAD_HANDSHAKE_MSG:` on stdout from the stub (which happened)

```
/ 'remote error: error decoding message'):
```

and also wanted to receive an alert (`AlertDescription::DecodeError`)

```
local error 'read tcp [::1]:50945->[::1]:50946: read: connection reset by peer'
```

but actually the stub just finished without sending one.

the code deleted in this PR that was sending this alert is: https://github.com/rustls/rustls/blob/main/rustls/src/conn.rs#L628-L632 - i think the intended replacement in `deframe` is not having the desired effect because, at that point, nothing has called `Message::try_from` to parse the semantic message contents from the `PlainMessage` that `deframe` details with.

rustls/src/msgs/deframer.rs
```diff
@@ -96,6 +98,8 @@ impl MessageDeframer {
         // we do a zero length read.  That looks like an EOF to
         // the next layer up, which is fine.
         debug_assert!(self.used <= OpaqueMessage::MAX_WIRE_SIZE);
+        let need_capacity = Ord::min(OpaqueMessage::MAX_WIRE_SIZE, self.used + READ_SIZE);
+        self.buf.resize(need_capacity, 0);
```
ctz, line 261: Hm, this adds (potential) allocator use and guaranteed-never-used memory zeroing (given `MessageDeframer::used` tracks meaningful memory in `buf`).

Just off some quick measurements, I think this halves bulk receive performance?

main:
```
$ BENCH_MULTIPLIER=30 ./target/release/examples/bench bulk TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
bulk	TLSv1_2	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256	max_fragment_size:default	send	2643.45	MB/s
bulk	TLSv1_2	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256	max_fragment_size:default	recv	3729.10	MB/s
```

this pr:
```
$ BENCH_MULTIPLIER=30 ./target/release/examples/bench bulk TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
bulk	TLSv1_2	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256	max_fragment_size:default	send	2702.09	MB/s
bulk	TLSv1_2	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256	max_fragment_size:default	recv	1958.14	MB/s
```

djc: Right, I think this just needs an if-statement to skip resizing if the buffer is large enough already, which should be a large part of the time in most cases?

djc:
Ahh, that makes sense, thanks! Will fix that up.

rustls/src/msgs/deframer.rs
djc: This has been improved, I get this on main:

```
djc-2021 main rustls $ BENCH_MULTIPLIER=30 ./target/release/examples/bench bulk TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256   max_fragment_size:default     send     2611.72 MB/s
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256   max_fragment_size:default     recv     2414.83 MB/s
```

and on this branch:

```
djc-2021 in-place-joining rustls $ BENCH_MULTIPLIER=30 ./target/release/examples/bench bulk TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256   max_fragment_size:default     send     2598.67 MB/s
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256   max_fragment_size:default     recv     2455.62 MB/s
```

That is, on my machine my branch can now receive 2% faster than on main. (It seems a little weird that on your machine, sends are about as fast as on mine, while receiving is ~50% faster.)

djc:
Fuzzer fails due to https://github.com/google/oss-fuzz/pull/9425.

rustls/src/msgs/deframer.rs
djc: Thanks, updated the comment.

rustls/src/msgs/deframer.rs
```diff
@@ -117,15 +268,101 @@ impl MessageDeframer {
...
 }
 
+const HEADER_SIZE: usize = 1 + 3;
+
+/// TLS allows for handshake messages of up to 16MB.  We
+/// restrict that to 64KB to limit potential for denial-of-
+/// service.
+const MAX_HANDSHAKE_SIZE: u32 = 0xffff;
+
+const READ_SIZE: usize = 4096;
```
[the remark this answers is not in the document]
djc: This seems specific enough that I don't think I'll bother with it for this PR. This seems like it'd be very sensitive to particular platforms and/or allocators, which we don't control at all, so I'm not sure it makes sense to make this kind of assumption here. Feel free to follow-up with another PR if you think this is important.

rustls/src/msgs/deframer.rs
djc: I've added some logic for this, see https://github.com/rustls/rustls/pull/1171/files#diff-7444beb5839fece40e64a80048a8cc03ab7b9f2e799a22628f97d8c0b5cf064cR284. In short, now allow large handshake messages (necessary to pass bogo tests), but after a large handshake message (larger than our 18k limit) we will shrink the buffer to a much smaller size again. We can probably tweak this more but, again, I think we should probably leave that to future PRs.

djc:
Not sure why the one bogo test keeps failing during the coverage run. Maybe it's getting killed because CIFuzz fails? The dedicated bogo runs succeed, and the same bogo test passes on my machine.

rustls/src/msgs/deframer.rs
larsfillmore: The use case I have in mind is occasional large application-level messages on a long-running connection that mostly receives rare small messages.

As a simplified example, imagine a chat server that mostly deals with small messages from its clients but occasionally clients send pictures or videos. If there is no eventual auto-shrinking for  `self.buf`  then most of the connections will end up with `OpaqueMessage::MAX_WIRE_SIZE` buffers.

Nevertheless, I agree that finding/implementing the best strategy for auto-shrinking  is better left for a separate PR.

djc: My goal is still to aim for a zero-allocation API, so in that sense I don't want to spend a lot of time on optimizing the internal buffering strategy. But if you want to propose something after this is merged, we can definitely consider it.

rustls/src/msgs/deframer.rs
```diff
@@ -1,76 +1,308 @@
...
+        if self.used >= allow_max {
+            return Err(io::Error::new(io::ErrorKind::Other, "message buffer full"));
+        }
+
+        // If we can and need to increase the buffer size to allow a 4k read, do so. After
+        // dealing with a large handshake message (exceeding `OpaqueMessage::MAX_WIRE_SIZE`),
+        // make sure to reduce the buffer size again (large messages should be rare).
+        let need_capacity = Ord::min(allow_max, self.used + READ_SIZE);
+        if need_capacity > self.buf.len() || self.buf.len() > allow_max {
+            self.buf.resize(need_capacity, 0);
```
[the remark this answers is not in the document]
djc: Good call, will fix that up.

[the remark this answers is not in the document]
djc: I've added code to shrink the capacity, please review!

djc:
I think this is good to go now -- I figured out the cause of the intermittent bogo issue.

ctz (approved):

rustls/src/lib.rs
```diff
@@ -356,6 +356,10 @@ pub mod internal {
     pub mod cipher {
         pub use crate::cipher::MessageDecrypter;
     }
+    /// Low-level TLS record layer functions.
+    pub mod record_layer {
+        pub use crate::record_layer::{Decrypted, RecordLayer};
+    }
```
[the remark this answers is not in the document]
djc: It now appears in the MessageDeframer API, which is public as part of the internals API.

PR #1172, opened 2023-01-14
Improve error reporting for corrupt messages

complexspaces:
Re-introduction and rebasing of https://github.com/rustls/rustls/pull/1016 now that the API is open for breaking changes again.

Closes https://github.com/rustls/rustls/pull/1043

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -9,6 +9,7 @@ use env_logger;
 use rustls;
 
 use rustls::internal::msgs::codec::Codec;
+use rustls::internal::msgs::enums::KeyUpdateRequest;
 use rustls::internal::msgs::persist;
 use rustls::quic::{self, ClientQuicExt, QuicExt, ServerQuicExt};
 use rustls::server::ClientHello;
@@ -592,7 +593,7 @@ fn quit_err(why: &str) -> ! {
 
 fn handle_err(err: rustls::Error) -> ! {
     use rustls::{AlertDescription, ContentType};
-    use rustls::{Error, PeerMisbehaved};
+    use rustls::{Error, InvalidMessage, PeerMisbehaved};
     use std::{thread, time};
 
     println!("TLS error: {:?}", err);
@@ -610,13 +611,22 @@ fn handle_err(err: rustls::Error) -> ! {
         Error::AlertReceived(AlertDescription::InternalError) => {
             quit(":PEER_ALERT_INTERNAL_ERROR:")
         }
-        Error::CorruptMessagePayload(ContentType::Alert) => quit(":BAD_ALERT:"),
-        Error::CorruptMessagePayload(ContentType::ChangeCipherSpec) => {
+        Error::InvalidMessage(InvalidMessage::MissingPayload(ContentType::Alert)) => {
+            quit(":BAD_ALERT:")
+        }
+        Error::InvalidMessage(InvalidMessage::MissingPayload(ContentType::ChangeCipherSpec)) => {
             quit(":BAD_CHANGE_CIPHER_SPEC:")
         }
-        Error::CorruptMessagePayload(ContentType::Handshake) => quit(":BAD_HANDSHAKE_MSG:"),
-        Error::CorruptMessagePayload(ContentType::Unknown(42)) => quit(":GARBAGE:"),
-        Error::CorruptMessage => quit(":GARBAGE:"),
+        Error::InvalidMessage(InvalidMessage::MissingPayload(ContentType::Handshake)) => {
+            quit(":BAD_HANDSHAKE_MSG:")
+        }
+        Error::InvalidMessage(InvalidMessage::InvalidKeyUpdate(KeyUpdateRequest::Unknown(42))) => {
+            quit(":BAD_HANDSHAKE_MSG:")
+        }
+        Error::InvalidMessage(InvalidMessage::InvalidCertRequest)
+        | Error::InvalidMessage(InvalidMessage::InvalidDhParams)
+        | Error::InvalidMessage(InvalidMessage::MissingKeyExchange) => quit(":BAD_HANDSHAKE_MSG:"),
+        Error::InvalidMessage(_) => quit(":GARBAGE:"),
         Error::DecryptError => quit(":DECRYPTION_FAILED_OR_BAD_RECORD_MAC:"),
         Error::PeerIncompatible(_) => quit(":INCOMPATIBLE:"),
         Error::PeerMisbehaved(PeerMisbehaved::TooMuchEarlyDataReceived) => {
@@ -764,7 +774,8 @@ fn exec(opts: &Options, mut sess: Connection, count: usize) {
             let mut one_byte = [0u8];
             let mut cursor = io::Cursor::new(&mut one_byte[..]);
             sess.write_tls(&mut cursor).unwrap();
-            conn.write(&one_byte).expect("IO error");
+            conn.write_all(&one_byte)
+                .expect("IO error");
 
             quench_writes = true;
         }
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1,8 +1,9 @@
 use crate::check::{inappropriate_handshake_message, inappropriate_message};
 use crate::conn::{CommonState, ConnectionRandoms, Side, State};
 use crate::enums::ProtocolVersion;
-use crate::error::{Error, PeerMisbehaved};
+use crate::error::{Error, InvalidMessage, PeerMisbehaved};
 use crate::hash_hs::HandshakeHash;
+use crate::kx;
 #[cfg(feature = "logging")]
 use crate::log::{debug, trace};
 use crate::msgs::base::{Payload, PayloadU8};
@@ -22,7 +23,7 @@ use crate::suites::PartiallyExtractedSecrets;
 use crate::suites::SupportedCipherSuite;
 use crate::ticketer::TimeBase;
 use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
-use crate::{kx, verify};
+use crate::verify;
 
 use super::client_conn::ClientConnectionData;
 use super::hs::ClientContext;
@@ -405,7 +406,7 @@ impl State<ClientConnectionData> for ExpectServerKx {
             .ok_or_else(|| {
                 cx.common
                     .send_fatal_alert(AlertDescription::DecodeError);
-                Error::CorruptMessagePayload(ContentType::Handshake)
+                InvalidMessage::MissingKeyExchange
             })?;
 
         // Save the signature and signed parameters for later verification.
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -5,7 +5,7 @@ use crate::conn::Protocol;
 use crate::conn::Side;
 use crate::conn::{CommonState, ConnectionRandoms, State};
 use crate::enums::{ProtocolVersion, SignatureScheme};
-use crate::error::{Error, PeerIncompatible, PeerMisbehaved};
+use crate::error::{Error, InvalidMessage, PeerIncompatible, PeerMisbehaved};
 use crate::hash_hs::{HandshakeHash, HandshakeHashBuffer};
 use crate::kx;
 #[cfg(feature = "logging")]
@@ -522,7 +522,7 @@ impl State<ClientConnectionData> for ExpectCertificateRequest {
             warn!("Server sent non-empty certreq context");
             cx.common
                 .send_fatal_alert(AlertDescription::DecodeError);
-            return Err(Error::CorruptMessagePayload(ContentType::Handshake));
+            return Err(InvalidMessage::InvalidCertRequest.into());
         }
 
         let tls13_sign_schemes = sign::supported_sign_tls13();
@@ -588,7 +588,7 @@ impl State<ClientConnectionData> for ExpectCertificate {
             warn!("certificate with non-empty context during handshake");
             cx.common
                 .send_fatal_alert(AlertDescription::DecodeError);
-            return Err(Error::CorruptMessagePayload(ContentType::Handshake));
+            return Err(InvalidMessage::InvalidCertRequest.into());
         }
 
         if cert_chain.any_entry_has_duplicate_extension()
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -1,4 +1,4 @@
-use crate::msgs::enums::{AlertDescription, ContentType, HandshakeType};
+use crate::msgs::enums::{AlertDescription, ContentType, HandshakeType, KeyUpdateRequest};
 use crate::rand;
 
 use std::error::Error as StdError;
@@ -31,11 +31,8 @@ pub enum Error {
         got_type: HandshakeType,
     },
 
-    /// The peer sent us a syntactically incorrect TLS message.
-    CorruptMessage,
-
     /// The peer sent us a TLS message with invalid contents.
-    CorruptMessagePayload(ContentType),
+    InvalidMessage(InvalidMessage),
 
     /// The peer didn't give us any certificates.
     NoCertificatesPresented,
@@ -92,6 +89,42 @@ pub enum Error {
     /// The `max_fragment_size` value supplied in configuration was too small,
     /// or too large.
     BadMaxFragmentSize,
+}
+
+/// A corrupt TLS message payload that resulted in an error.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy)]
+pub enum InvalidMessage {
+    /// The peer sent us a syntactically incorrect TLS message.
+    IncorrectFrame,
+    /// An advertised message was larger then expected.
+    HandshakePayloadTooLarge,
+    /// A message was zero-length when its record kind forbids it.
+    InvalidEmptyPayload,
+    /// A TLS message payload was larger then allowed by the specification.
+    MessageTooLarge,
+    /// An unknown content type was encountered during message decoding.
+    InvalidContentType,
+    /// An unknown TLS protocol was encountered during message decoding.
+    UnknownProtocolVersion,
+    /// Decoding a message payload for a [ContentType] failed because a different
+    /// type was encountered.
+    MissingPayload(ContentType),
+    /// A peer did not advertise its supported key exchange groups.
+    MissingKeyExchange,
+    /// Context was incorrectly attached to a certificate request during a handshake.
+    InvalidCertRequest,
+    /// A peer sent an unexpected key update request.
+    InvalidKeyUpdate(KeyUpdateRequest),
+    /// A peer's DH params could not be decoded
+    InvalidDhParams,
+}
+
+impl From<InvalidMessage> for Error {
+    #[inline]
+    fn from(e: InvalidMessage) -> Self {
+        Self::InvalidMessage(e)
+    }
 }
 
 #[non_exhaustive]
@@ -298,7 +331,7 @@ impl fmt::Display for Error {
                 got_type,
                 join::<HandshakeType>(expect_types)
             ),
-            Self::CorruptMessagePayload(ref typ) => {
+            Self::InvalidMessage(ref typ) => {
                 write!(f, "received corrupt message of type {:?}", typ)
             }
             Self::PeerIncompatible(ref why) => write!(f, "peer is incompatible: {:?}", why),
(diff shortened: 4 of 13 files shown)


ctz:
Given there are ten different call sites for `corrupt_message`, maybe we could just describe them in a new `non_exhaustive` enum and put that in `Error::CorruptMessagePayload`?

I haven't seen people putting `core::panic::Location` in error variants in other crates, and I'm not sure what it means in terms of structure and program size?

djc:
Yeah, I think I'd prefer that approach as well, should match well with the Codec yielding Result.

complexspaces:
> maybe we could just describe them in a new non_exhaustive enum and put that in Error::CorruptMessagePayload

Just to double check, are either of you opposed to having ~10 different enum variants to start? One of the goals of this PR was to know exactly which failure branch was reached in the state machine and roughly that many (AFAICT) would be required.

djc:
No, I think Joe was describing exactly that.

complexspaces:
Quick status update: I am still working on this, but I've been short on time. Sorry for the delay.

complexspaces:
I've rebased these changes onto `main` and made the requested changes. As the end result, the existing corrupted message handling also got merged into this "new way" so that there was only one `Error` variant related to corrupt/invalid messages.

Please have a look and let me know if anything else needs fixed or cleaned up.

djc (commented):
This is looking great! Here are some further suggestions for improvements.

If you won't have time for this, I'd be happy to contribute further work along these lines.

rustls/src/msgs/deframer.rs
```diff
@@ -75,7 +77,9 @@ impl MessageDeframer {
                 }
                 Err(_) => {
                     self.desynced = true;
-                    return Err(Error::CorruptMessage);
+                    return Err(Error::CorruptMessagePayload(
```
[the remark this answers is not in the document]
complexspaces: Hm, I am not sure I follow yet, sorry. Do you mean that we'd add a new variant to `CorruptMessage` that contains a `MessageError` for the deframer to use and fill with any errors that come out of `OpaqueMessage::read()`? 

> We could potentially rename...

Following, we'd then just rename the variants on `MessageError`?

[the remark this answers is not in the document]
djc: Sorry, I'm proposing that we add the following variants to `CorruptMessage`:

* `HandshakePayloadTooLarge`
* `EmptyApplicationData`
* `MessageTooLarge`
* `InvalidContentType`
* `UnknownProtocolVersion`

And that we modify `MessageError` to have these same variants (from today's `IncorrectSize`, `IllegalLength`, `IllegalContentType` and `IllegalProtocolVersion`), such that we propagate these variants from `MessageError` to the application while keeping `TooShortForHeader` and `TooShortForLength` as is (that is, handled by the deframer).

Does that make sense?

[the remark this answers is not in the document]
complexspaces: I think so, yeah. I take it then that `MessageError` is just another case of [an enum that shouldn't really be exposed from the internals module](https://docs.rs/rustls/latest/rustls/internal/msgs/message/enum.MessageError.html)?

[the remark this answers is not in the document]
djc: Yes!

rustls/src/error.rs
```diff
@@ -1,11 +1,32 @@
...
 use std::error::Error as StdError;
 use std::fmt;
 use std::sync::Arc;
 use std::time::SystemTimeError;
 
+/// A corrupt TLS message payload that resulted in an error.
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub enum CorruptMessageError {
+    /// The peer sent us a syntatically incorrect TLS message.
```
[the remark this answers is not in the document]
complexspaces: Oops, thank you for catching that.

rustls/src/error.rs
```diff
@@ -31,11 +52,8 @@ pub enum Error {
         got_type: HandshakeType,
     },
 
-    /// The peer sent us a syntactically incorrect TLS message.
-    CorruptMessage,
-
     /// The peer sent us a TLS message with invalid contents.
-    CorruptMessagePayload(ContentType),
+    CorruptMessagePayload(CorruptMessageError),
```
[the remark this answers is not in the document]
complexspaces: I left this one as `CorruptMessage`, even after renaming the inner enum. IMO its really weird to have a variant name and inner type named the exact same thing.

rustls/src/msgs/deframer.rs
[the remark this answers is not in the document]
complexspaces: I believe that I've implemented this like discussed, please have a look when you can.

rustls/src/msgs/deframer.rs
```diff
@@ -41,7 +41,9 @@ impl MessageDeframer {
     /// `Ok(Some(_))` if a valid message was found and decrypted successfully.
     pub fn pop(&mut self, record_layer: &mut RecordLayer) -> Result<Option<Deframed>, Error> {
         if self.desynced {
-            return Err(Error::CorruptMessage);
+            return Err(Error::CorruptMessagePayload(
+                CorruptMessageError::IncorrectFrame,
+            ));
```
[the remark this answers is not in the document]
complexspaces: I took a stab at implementing this in a second commit since it looked fairly simple. Please have a look.

rustls/src/error.rs
```diff
@@ -1,11 +1,32 @@
...
 
 use std::error::Error as StdError;
 use std::fmt;
 use std::sync::Arc;
 use std::time::SystemTimeError;
 
+/// A corrupt TLS message payload that resulted in an error.
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+pub enum CorruptMessageError {
```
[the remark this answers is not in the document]
complexspaces: I switched this to `Invalid`, since I agree its more generic/future-proof.

rustls/src/error.rs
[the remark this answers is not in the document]
djc: I think that's actually fairly common, but I don't care a lot.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -9,6 +9,7 @@ use env_logger;
 use rustls;
 
 use rustls::internal::msgs::codec::Codec;
+use rustls::internal::msgs::enums::KeyUpdateRequest;
```
djc, line 12: This seems unrelated?

rustls/examples/internal/bogo_shim.rs
```diff
@@ -764,7 +781,8 @@ fn exec(opts: &Options, mut sess: Connection, count: usize) {
             let mut one_byte = [0u8];
             let mut cursor = io::Cursor::new(&mut one_byte[..]);
             sess.write_tls(&mut cursor).unwrap();
-            conn.write(&one_byte).expect("IO error");
+            conn.write_all(&one_byte)
```
djc, line 784: Unrelated?

rustls/examples/internal/bogo_shim.rs
```diff
@@ -610,13 +611,29 @@ fn handle_err(err: rustls::Error) -> ! {
...
+        | Error::CorruptMessage(InvalidMessage::InvalidDhParams)
+        | Error::CorruptMessage(InvalidMessage::MissingKeyExchange) => quit(":BAD_HANDSHAKE_MSG:"),
+        Error::CorruptMessage(InvalidMessage::InvalidContentType)
+        | Error::CorruptMessage(InvalidMessage::EmptyDataForDisallowedRecord)
+        | Error::CorruptMessage(InvalidMessage::UnknownProtocolVersion)
+        | Error::CorruptMessage(InvalidMessage::MessageTooLarge)
+        | Error::CorruptMessage(InvalidMessage::MissingPayload(ContentType::Unknown(42))) => {
+            quit(":GARBAGE:")
+        }
+        Error::CorruptMessage(InvalidMessage::IncorrectFrame) => quit(":GARBAGE:"),
```
djc, line 636: Can we do a catchall `Error::CorruptMessage(_)` here at some point so we don't have to exhaustively list all the cases manually?

rustls/src/client/tls12.rs
```diff
@@ -405,7 +405,7 @@ impl State<ClientConnectionData> for ExpectServerKx {
             .ok_or_else(|| {
                 cx.common
                     .send_fatal_alert(AlertDescription::DecodeError);
-                Error::CorruptMessagePayload(ContentType::Handshake)
+                Error::CorruptMessage(InvalidMessage::MissingKeyExchange)
```
djc, line 408: Similar to the earlier PRs, it would good to add some `From` impls to make this slightly less wordy. For example, if there was an `impl From<InvalidMessage> for Error` then we could probably just yield `InvalidMessage::MissingKeyExchange` here, and let the `?` operator take care of wrapping it in `Error`?

rustls/src/client/tls13.rs
```diff
@@ -522,7 +522,7 @@ impl State<ClientConnectionData> for ExpectCertificateRequest {
             warn!("Server sent non-empty certreq context");
             cx.common
                 .send_fatal_alert(AlertDescription::DecodeError);
-            return Err(Error::CorruptMessagePayload(ContentType::Handshake));
+            return Err(Error::CorruptMessage(InvalidMessage::InvalidCertRequest));
```
djc, line 525: Same here with the explicit wrapping.

rustls/src/error.rs
```diff
@@ -1,11 +1,40 @@
-use crate::msgs::enums::{AlertDescription, ContentType, HandshakeType};
+use crate::msgs::enums::{AlertDescription, ContentType, HandshakeType, KeyUpdateRequest};
 use crate::rand;
 
 use std::error::Error as StdError;
 use std::fmt;
 use std::sync::Arc;
 use std::time::SystemTimeError;
 
+/// A corrupt TLS message payload that resulted in an error.
```
djc, line 9: Style nit: probably better to move this below the `Error` definition, close to `PeerMisbehaved`.

rustls/src/msgs/deframer.rs
```diff
@@ -70,12 +70,22 @@ impl MessageDeframer {
             let mut rd = codec::Reader::init(&self.buf[start..self.used]);
             let m = match OpaqueMessage::read(&mut rd) {
                 Ok(m) => m,
-                Err(MessageError::TooShortForHeader | MessageError::TooShortForLength) => {
-                    return Ok(None)
-                }
-                Err(_) => {
+                Err(msg_err) => {
```
djc, line 73: Why change this to move the `TooShortForHeader | TooShortForLength` arms into a nested `match`? That doesn't seem like an improvement to me, since they're handled differently than all the the other cases.

Hmm, I guess it has to do with the exhaustiveness of matching here? I guess it's a little ugly but I can't quite think of a better solution.

rustls/src/msgs/message.rs
```diff
@@ -91,23 +92,23 @@ impl OpaqueMessage {
         //  implemented per section 5.1 of RFC8446 (TLSv1.3)
         //              per section 6.2.1 of RFC5246 (TLSv1.2)
         if typ != ContentType::ApplicationData && len == 0 {
-            return Err(MessageError::IllegalLength);
+            return Err(MessageError::EmptyDataForDisallowedRecord);
```
djc, line 95: This variant name doesn't really make sense to me, what are you trying to convey here? It seems to be much more abstract than the `EmptyApplicationData` I proposed.

rustls/src/tls12/mod.rs
```diff
@@ -1,14 +1,14 @@
...
-use crate::kx;
 use crate::msgs::codec::{Codec, Reader};
-use crate::msgs::enums::{AlertDescription, ContentType};
+use crate::msgs::enums::AlertDescription;
 use crate::msgs::handshake::KeyExchangeAlgorithm;
 use crate::suites::{BulkAlgorithm, CipherSuiteCommon, SupportedCipherSuite};
 #[cfg(feature = "secret_extraction")]
 use crate::suites::{ConnectionTrafficSecrets, PartiallyExtractedSecrets};
 use crate::Error;
+use crate::{kx, InvalidMessage};
```
djc, line 11: Style nit: I would prefer to live `crate::kx` as is and import `Error` and `InvalidMessage` both directly from `crate::error`.

rustls/src/error.rs
```diff
@@ -1,11 +1,40 @@
...
+#[non_exhaustive]
+#[derive(Debug, Clone)]
```
djc, line 11: I think we want to derive `Copy` here, too.

rustls/src/msgs/deframer.rs
```diff
@@ -84,8 +84,11 @@ impl MessageDeframer {
                             InvalidMessage::UnknownProtocolVersion
                         }
                     };
-                    self.desynced = true;
-                    return Err(Error::CorruptMessage(err_kind));
+
+                    return Err(self
+                        .last_error
+                        .insert(Error::CorruptMessage(err_kind))
+                        .clone());
```
djc, line 91: Hmm, I would maybe go with `self.last_error = Some(Error::from(err_kind)); return Err(Error::from(err_kind));`, preferring the slight duplication over the verbosity of this construction (which is might've been nice if it didn't get spread over 4 lines).

rustls/src/msgs/message.rs
complexspaces: I'm fine working on the errors name more. I had changed it because, AFAICT, its not actually empty application data that is being checked for here. According to the RFC section (and the PR that added this, https://github.com/rustls/rustls/pull/848), application data is the only content type that's allowed to be empty, but not the others. Given that, I tried to come up with a name inverse to `EmptyApplicationData`.

rustls/examples/internal/bogo_shim.rs
complexspaces: This is currently used in the error handling shim further down in the file:
```rust
Error::CorruptMessage(InvalidMessage::InvalidKeyUpdate(KeyUpdateRequest::Unknown(42))) => {
            quit(":BAD_HANDSHAKE_MSG:")
        }
```

Would you prefer that I removed it in favor of a less explicit `InvalidMessage::InvalidKeyUpdate(_)` pattern instead?

rustls/examples/internal/bogo_shim.rs
complexspaces: I didn't change this manually. For some reason, running `cargo fmt` keeps forcing it onto a second line on my branch.

rustls/src/error.rs
complexspaces: Sure, I can do that if you'd like. I originally withheld from doing so since I was matching the traits implemented by `PeerIncompatible`, which only has `Clone`

rustls/src/msgs/deframer.rs
complexspaces: Yes, that's the reason why 👍. There's no way to tell Rust you handled some cases in a different early return branch and I figured this was more preferable to using `unreachable!()`.

rustls/src/msgs/deframer.rs
complexspaces: Does this feedback apply to all callsites that are currently using the `.insert()` pattern? I had originally done it with an intermediate everywhere but it got kind of cluttered.

djc: Hmm, why was it cluttered? Maybe one we derive `Copy` we can away without the `.clone()` call (might need a deref operator) so things become a little cleaner?

rustls/src/msgs/message.rs
djc: Ah yes, sorry, I missed that. How about `InvalidEmptyPayload`?

rustls/src/msgs/deframer.rs
djc: Okay, let's leave it like this, I guess.

rustls/src/error.rs
```diff
@@ -1,11 +1,40 @@
...
+pub enum InvalidMessage {
+    /// The peer sent us a syntactically incorrect TLS message.
```
nitnelave, line 13: Just from a user's point of view, what would be the error message when sending a starttls message to a TLS server or vice-versa? Is there a way to make it painfully clear in the error message?

djc:
Maybe also add a commit somewhat like this:

```diff
diff --git a/rustls/src/msgs/message.rs b/rustls/src/msgs/message.rs
index 6d1b2e60..861b47e0 100644
--- a/rustls/src/msgs/message.rs
+++ b/rustls/src/msgs/message.rs
@@ -84,26 +84,12 @@ impl OpaqueMessage {
     /// become valid if we read more data) and invalid data.
     pub fn read(r: &mut Reader) -> Result<Self, MessageError> {
         let typ = ContentType::read(r).ok_or(MessageError::TooShortForHeader)?;
-        let version = ProtocolVersion::read(r).ok_or(MessageError::TooShortForHeader)?;
-        let len = u16::read(r).ok_or(MessageError::TooShortForHeader)?;
-
-        // Reject undersize messages
-        //  implemented per section 5.1 of RFC8446 (TLSv1.3)
-        //              per section 6.2.1 of RFC5246 (TLSv1.2)
-        if typ != ContentType::ApplicationData && len == 0 {
-            return Err(MessageError::IllegalLength);
-        }
-
-        // Reject oversize messages
-        if len >= Self::MAX_PAYLOAD {
-            return Err(MessageError::IllegalLength);
-        }
-
         // Don't accept any new content-types.
         if let ContentType::Unknown(_) = typ {
             return Err(MessageError::IllegalContentType);
         }
 
+        let version = ProtocolVersion::read(r).ok_or(MessageError::TooShortForHeader)?;
         // Accept only versions 0x03XX for any XX.
         match version {
             ProtocolVersion::Unknown(ref v) if (v & 0xff00) != 0x0300 => {
@@ -112,6 +98,16 @@ impl OpaqueMessage {
             _ => {}
         };
 
+        let len = u16::read(r).ok_or(MessageError::TooShortForHeader)?;
+        // Reject undersize messages
+        //  implemented per section 5.1 of RFC8446 (TLSv1.3)
+        //              per section 6.2.1 of RFC5246 (TLSv1.2)
+        if typ != ContentType::ApplicationData && len == 0 {
+            return Err(MessageError::IllegalLength);
+        } else if len >= Self::MAX_PAYLOAD { // Reject oversize messages
+            return Err(MessageError::IllegalLength);
+        }
+
         let mut sub = r
             .sub(len as usize)
             .ok_or(MessageError::TooShortForLength)?;
```

I think that would make things a bunch clearer.

rustls/src/error.rs
djc: If you send a message like `STARTTLS` to a TLS server, rustls on the server would yield an `InvaidMessage::InvalidContentType`, because it doesn't find a known content type in the range 0x14..=0x18 in the first byte. That seems pretty reasonable?

For "vice-versa", rustls wouldn't get invoked if you send TLS messages to a peer expecting a plaintext byte stream, but TLS messages are fairly recognizable from their 5 byte header, starting with the `ContentType` byte, then a two-byte version (rustls accepts 0x03XX at the message decoding level), then a two-byte length.

djc: I don't think we have a straightforward route to decorating these error variants with more detailed context.

djc:
@complexspaces would you have the chance to iterate on this soon? Would you like me to take over?

complexspaces:
Sorry for the delay, @djc. I'll push the next iteration tomorrow.

djc:
No need to apologize, I can see how this PR might not be your top priority in life. 😉

rustls/src/error.rs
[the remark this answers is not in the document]
complexspaces: I didn't really feel strongly on this, so I've renamed this to `InvalidMessage` to match.

rustls/examples/internal/bogo_shim.rs
complexspaces: This is now done 👍

rustls/src/client/tls12.rs
complexspaces: I modified all of the inline/chained occurrences of this pattern. I left the ones that had just error construction in an un-nested line alone for clarity.

rustls/src/error.rs
complexspaces: Good call, this has been moved down in the file.

rustls/src/msgs/message.rs
complexspaces: That works for me. I think its more clear when combined with the doc comments then what I had.

rustls/src/msgs/deframer.rs
complexspaces: I ended up trying out a very small helper function, `set_err`, that encloses this convert, assign, return pattern the deframer uses in multiple places. The shortness of the function allowed all of the callsites to be formatted pretty cleanly, IMO. Please take a look and see if that also works for you.

djc (approved):

complexspaces:
Sorry about the import mess, @djc. I missed those when going through this branch again.

djc:
No problem!

PR #1173, opened 2023-01-15
Improve clippy usability with Cargo alias

complexspaces:
This PR allows local developers and CI to more easily share testing configurations. Now, developers can double check their code will pass CI by easily keeping the "right command and flags" synced automatically. `cargo clipppy-rustls` and `cargo test` are now the two commands "most" contributors need to remember.

Both the [Neon project](https://github.com/neon-bindings/neon/blob/main/.cargo/config.toml#L1) and internally at 1P make use of this pattern to improve the developer experience. Its especially helpful if `rustls` ever decides to make some code stricter (banning `as` casting, for example).

Additionally, this moves Clippy to run on the whole workspace instead of just `rustls`. The one remaining Clippy lint in the workspace was fixed in order to pass with the new command.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+clippy-rustls = "clippy --all-features"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -266,8 +266,8 @@ jobs:
           components: clippy
       - uses: actions-rs/cargo@v1
         with:
-          command: clippy
-          args: -p rustls --all-features -- -D warnings
+          command: clippy-rustls
+          args: -- -D warnings
 
   clippy-nightly:
     name: Clippy (Nightly)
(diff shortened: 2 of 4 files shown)


djc:
I suppose I feel that `cargo clippy --all-features` is simple/straightforward enough that it shouldn't be surprising? If we wanted to rewire which clippy lints we want to run we would do that with crate-level inner attributes (we have two clippy denies and 5 clippy allows right now). As such, I feel this is pretty low value, and the additional indirection in the CI config is not quite free.

cpu:
I lean towards agreeing with @djc. I think `cargo clippy --all-features` is slightly more verbose, but also translates better to other crates. Remembering to invoke `cargo clippy-rustls` is project specific :thinking:

cpu:
Since there hasn't been any additional input from the community in favour of the alias I'm going to close this one out. We can revisit if there's more interest later on.

PR #1174, opened 2023-01-16
examples: use inline variable names in format strings

djc:
As suggested by nightly clippy.


ctz:
This isn't compatible with the MSRV, but I don't really understand why clippy is suggesting it at all in that case - given we set rust-version in the Cargo.toml?

djc:
It's only suggesting it for the examples crate, which doesn't have `rust-version` set. Will fix.

Issue #1162, opened 2023-01-12
Certificate errors being are not machine readable


mdecimus:
Hi,

It is currently not possible to access the certificate error details from ``rustls::Error``, they are returned as a ``InvalidCertificateData(String)`` rather than ``CertificateError(Box<dyn Error>)`` or similar. 

For example, a certificate not valid for name error is returned as follows:

```
 InvalidCertificateData("invalid peer certificate: CertNotValidForName")
```

Thanks.

Issue #932, opened 2022-01-14
Certificate error variants are too tied to BOGO's expectations


briansmith:
Currently we have these error variants intended to be used for certificate errors:
```
    /// We received an invalidly encoded certificate from the peer.
    InvalidCertificateEncoding,

    /// We received a certificate with invalid signature type.
    InvalidCertificateSignatureType,

    /// We received a certificate with invalid signature.
    InvalidCertificateSignature,

    /// We received a certificate which includes invalid data.
    InvalidCertificateData(String),
```
When implementing custom certificate verifiers we've found that we're often just returning `InvalidCertificateData` or else we're doing wasted work to return one of the other variants.

It seems like the distinction between these five variants is primarily driven by the desire of BOGO to differentiate between different types of errors. But, if that is the case then it would be good to find a way to limit this classification to just BogoShim and not require every implementation to do this classification. In some implementations it isn't clear that we even can accurately classify the errors returned by the underlying certificate verification library into the above.

If we implement my suggestion in #930, then we can have BogoShim do the mapping from webpki::Error to the five categories above, and save other implementations the trouble.

briansmith:
Just to clarify, my recommendation here is that we replace all of the aforementioned variants with a single variant, perhaps `PeerCertificateInvalid(Box<dyn std::error::Error>`.

jsha:
That would be a nice cleanup for rustls-ffi. Right now we offer a helper method to determine whether a returned error is related to certificate verification. It would be nice if there was a single error variant to check and we could do away with the helper method.

Issue #930, opened 2022-01-14
Use `Box<dyn std::error::Error>` as the error type for certificate validation errors


briansmith:
Right now certificate validation errors have to be formatted into `String` to fit into these variants:
```
    /// We received a certificate which includes invalid data.
    InvalidCertificateData(String),
```
When implementing validators, and especially when testing them, we've found being limited to `String` is unfortunate. For example, imagine a certificate verification library that abstracts over the native Windows, Android, and macOS/iOS certificate validation libraries. It would be nice to be able to retain some of the OS-specific error information, especially the specific error codes returned by the OS, in a way that doesn't require us to parse a `String`.

If we change `InvalidCertificateData(String)` to `Box<dyn std::error::Error>` we can achieve the same effects as we can currently since `std::error::Error: Debug + Display` to provide a validator-implementation-agnostic interface, but also provide the opportunity for callers (tests) to safely downcast (using `Any`) the value to obtain the OS-specific data.

Stargateur:
I would suggest to rework the general variant too based on this proposition.

Currently:
> ```rust
> /// A catch-all error for unlikely errors.
> General(String),
> ```

tonyfinn:
I had hoped to use rustls to check the time until certificate expiration. For a non-expired cert, this information is inspectable through `ClientConnection`, but with the current error layout it's not possible to distinguish an expired cert from other instances of server misbehaviour. For this use case, a expired cert in this case would be effectively another value, in the past, while a protocol error would be a case of not being able to receive a result.. Prior versions of rustls would have allowed detecting this case - the ability to detect this is still documented on the front page of the docs, even though it has been removed by the error refactor.

https://docs.rs/rustls/latest/rustls/

> Any error returned from process_new_packets is fatal to the connection, and will tell you why. For example, if the server’s certificate is expired process_new_packets will return Err(WebPkiError(CertExpired, ValidateServerCert)).

So that's another use case where it would be nice to see underlying error message details.

Issue #889, opened 2021-12-22
Replace `Error::*Error(String)` with non-heap-allocating localizable enum variants


briansmith:
There are many places where we do things like this:

```rust
return Err(Error::General("No end-entity certificate in certificate chain".to_string()));
```
and
```rust
return Err(Error::PeerMisbehavedError("client sent wrong binder".to_string()));
```

Each of these does a heap allocation. There are many reasons why we want to reduce heap allocations in Rustls.

The most straightforward way of eliminating these allocations would be to use `&'static str`; this would also eliminate any possibility of (secret) application data getting into the strings. If that weren't practical, `Cow<'static, str>` would be another alternative.

However, in both of those cases, we're stuck with English strings in the error messages. We might as well replace the strings with an `enum` like this:
```
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub enum PeerMisbehavedError {
     NoEndEntityCertificateInCertificateChain,
     ...
}
```
In the case of `General`, instead of creating a new enum, we might just add the enum variants to `Error` itself.

briansmith:
In particular, an application should be able to embed Rustls without having all of these error strings in the resulting binary, and an application should be able to localize the errors into whatever language the authors prefer.

ctz:
Note this was done for all variants apart from `Error::General`

PR #1176, opened 2023-01-19
Remove string content from `rustls::Error`

ctz:
- [x] `PeerMisbehavedError` -> enum
- [x] `PeerIncompatibleError` -> enum
- [ ] `General` -> `Box<dyn Error>`
  - and audit uses of this
- [x] re-generalise certificate errors to  `Box<dyn Error>`
- [x] fix the README/lib.rs documentation that tells you to expect `Err(WebPkiError(CertExpired, ValidateServerCert))`

This is in pursuit of:

- #1162 
- #932 
- #930
- #889


djc:
Nice! I would maybe drop the Reason suffix from the nested enum and impl From<PeerMisbehaved> for Error, so that you can call .into() and minimize the boilerplate.

ctz:
Good idea! Have done that.

djc:
If we're improving error stuff, might also drop the `Error` suffix from `Error` variant names.

djc:
Also maybe do this as one PR per bullet point? Bit easier to digest and not much point in heaping them all together anyway.

ctz:
> Also maybe do this as one PR per bullet point? Bit easier to digest and not much point in heaping them all together anyway.

can definitely do that when we come to merge this; at the moment the second commit depends on the first so that's going to have to happen in order

djc:
> > Also maybe do this as one PR per bullet point? Bit easier to digest and not much point in heaping them all together anyway.
> 
> 
> 
> can definitely do that when we come to merge this; at the moment the second commit depends on the first so that's going to have to happen in order

Sure that's fine, so is this Ready for review?

ctz:
> * [ ]  re-generalise certificate errors to  `Box<dyn Error>`

Unfortunately seems not to be possible because `std::error::Error` does not implement `Clone` or `PartialEq` which we require for it to be part of our error type.

So I think these are best dealt with as another sub-enum?

djc:
We probably want to wrap a struct type around webpki error types, since just an enum variant can't hide the inner types.

ctz:
> We probably want to wrap a struct type around webpki error types, since just an enum variant can't hide the inner types.

I think we probably don't want the webpki error type involved at all? Any involvement of it at all means someone bringing their own `ServerCertVerifier` requires it to be publicly constructible (so they can express errors with the certificate) which means webpki types leak out the crate anyway, and forces everyone into a webpki dependency. (Though I think DnsName already does that and is unconditional.) If the webpki type is involved, better to just stick it straight in the rustls Error enum.

At this point I'm minded to just have rustls define its own `CertificateError` enum and funnel everything through that.

> Or we could propagate the bounds? Presumably the nested error types may also implement Clone?

Ah, that would be good. Seems like doing:

```
pub trait CloneableError : StdError + Clone {}
```

gets:

```
error[E0038]: the trait `CloneableError` cannot be made into an object
  --> rustls/src/error.rs:78:28
   |
78 |     InvalidCertificate(Box<dyn CloneableError>),
   |                            ^^^^^^^^^^^^^^^^^^ `CloneableError` cannot be made into an object
   |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
```

Which is not too surprising, since asking a trait object to clone itself is tricky. Same goes for `PartialEq` (so even avoiding `Clone` through `Rc`/`Arc` won't work)

djc:
> I think we probably don't want the webpki error type involved at all? Any involvement of it at all means someone bringing their own `ServerCertVerifier` requires it to be publicly constructible (so they can express errors with the certificate) which means webpki types leak out the crate anyway, and forces everyone into a webpki dependency. (Though I think DnsName already does that and is unconditional.) If the webpki type is involved, better to just stick it straight in the rustls Error enum.

I meant wrapping webpki errors into a `Box<dyn Error>`. But deriving a local `CertificateError` from that seems like a decent option, too.

> At this point I'm minded to just have rustls define its own `CertificateError` enum and funnel everything through that.

Yeah, probably makes sense.

Maybe we could just drop the `Clone` and `PartialEq` derives? I don't think that they're that commonly needed anyway.

ctz:
> Maybe we could just drop the `Clone` and `PartialEq` derives? I don't think that they're that commonly needed anyway -- ah, I do think we have some internal usage of `clone()` in places where we want to fuse the connection.

Yeah I think we need clone. But I think we can get rid of `PartialEq` -- working on that now. `Clone` can be obviated by using `Arc` rather than `Box`.

rustls/src/error.rs
```diff
@@ -60,17 +60,20 @@ pub enum Error {
...
+    /// We saw an invalid certificate.
+    ///
+    /// This includes, but is not limited to:
+    /// - we received an invalidly encoded certificate,
+    /// - we received a certificate that was not signed correctly
+    ///   by its alleged issuer,
+    /// - we received a certificate that was not issued by any
+    ///   of the configured trusted root certificates,
+    /// - we received a certificate that was not valid for the expect
+    ///   server name.
```
djc, line 72: I probably wouldn't keep this list here, just refer to the docs for `CertificateError`?

rustls/src/verify.rs
```diff
@@ -617,12 +617,22 @@ impl ClientCertVerifier for AllowAnyAnonymousOrAuthenticatedClient {
...
+        BadDer | BadDerTime => CertificateError::BadEncoding.into(),
+        CertNotValidYet => CertificateError::NotValidYet.into(),
+        CertExpired | InvalidCertValidity => CertificateError::Expired.into(),
+        UnknownIssuer => CertificateError::UnknownIssuer.into(),
+        CertNotValidForName => CertificateError::NotValidForName.into(),
+
+        InvalidSignatureForPublicKey
+        | UnsupportedSignatureAlgorithm
+        | UnsupportedSignatureAlgorithmForPublicKey => CertificateError::BadSignature.into(),
+        e => {
```
djc, line 629: Add a comment about `webpki::Error` being `non_exhaustive` and that this should thus be very unlikely? (Assuming you handled all the known variants above.)

rustls/src/error.rs
ctz: yeah will do that!

rustls/src/verify.rs
ctz: i don't think this handles the known webpki errors, but just now i've pushed something that stops discarding the underlying error in the `Other` case

PR #1178, opened 2023-01-23
Remove string content from `PeerMisbehavedError` and `PeerIncompatibleError`

ctz:


rustls/src/error.rs
```diff
@@ -99,6 +99,92 @@ pub enum Error {
...
+/// versions.  We also don't document what they mean.  Generally a user of
+/// rustls shouldn't vary its behaviour on these error codes, and there is
+/// nothing it can do to improve matters.
+///
+/// Please file a bug against rustls if you see `Error::PeerMisbehaved` in
+/// the wild.
+pub enum PeerMisbehaved {
+    AttemptedDowngradeToTls12WhenTls13IsSupported,
+    BadCertChainExtensions,
+    ClientHelloSniMustContainOneHostName,
```
djc, line 118: Is the `ClientHello` prefix useful here? Should we use `ServerName` instead of `Sni`?

rustls/src/error.rs
```diff
@@ -99,6 +99,92 @@ pub enum Error {
...
+    MissingQuicTransportParameters,
+    OfferedDuplicateKeyShares,
+    OfferedEarlyDataWithOldProtocolVersion,
+    OfferedEmptyAlpnProtocol,
+    OfferedIncorrectCompressions,
+    PskExtensionMustBeLast,
+    PskExtensionWithMismatchedIdsAndBinders,
+    RefusedToFollowHelloRetryRequest,
+    RejectedEarlyDataInterleavedWithHandshakeMessage,
+    ResumptionAttemptedWithVariedEms,
```
djc, line 156: Should we expand `Ems` in these errors?

rustls/src/error.rs
```diff
@@ -99,6 +99,92 @@ pub enum Error {
...
+    KeyUpdateReceivedInQuicConnection,
+    MessageInterleavedWithHandshakeMessage,
+    MissingBinderInPskExtension,
+    MissingKeyShare,
+    MissingQuicTransportParameters,
+    OfferedDuplicateKeyShares,
+    OfferedEarlyDataWithOldProtocolVersion,
+    OfferedEmptyAlpnProtocol,
+    OfferedIncorrectCompressions,
+    PskExtensionMustBeLast,
```
djc, line 152: Should we expand `Psk` in the variant names?

rustls/src/error.rs
```diff
@@ -99,6 +99,92 @@ pub enum Error {
...
+    ResumptionOfferedWithVariedCipherSuite,
+    ResumptionOfferedWithVariedEms,
+    ResumptionOfferedWithIncompatibleCipherSuite,
+    SelectedDifferentCipherSuiteAfterRetry,
+    SelectedInvalidPsk,
+    SelectedTls12UsingTls13VersionExtension,
+    SelectedUnofferedAlpnProtocol,
+    SelectedUnofferedCipherSuite,
+    SelectedUnofferedCompression,
+    SelectedUnofferedKxGroup,
```
djc, line 166: Should we expand `Kx` in these variant names?

rustls/src/error.rs
```diff
@@ -99,6 +99,92 @@ pub enum Error {
...
+    InvalidSctList,
+    KeyEpochWithPendingFragment,
+    KeyUpdateReceivedInQuicConnection,
+    MessageInterleavedWithHandshakeMessage,
+    MissingBinderInPskExtension,
+    MissingKeyShare,
+    MissingQuicTransportParameters,
+    OfferedDuplicateKeyShares,
+    OfferedEarlyDataWithOldProtocolVersion,
+    OfferedEmptyAlpnProtocol,
```
djc, line 150: Maybe `ApplicationProtocol` instead of `AlpnProtocol`?

rustls/src/error.rs
```diff
@@ -99,6 +99,92 @@ pub enum Error {
...
+    SelectedUnofferedPsk,
+    SelectedUnusableCipherSuiteForVersion,
+    ServerHelloMustOfferUncompressedEcPoints,
+    SignedKxWithWrongAlgorithm,
+    SignedHandshakeWithUnadvertisedSigScheme,
+    TooMuchEarlyDataReceived,
+    UnexpectedCleartextExtension,
+    UnsolicitedCertExtension,
+    UnsolicitedEncryptedExtension,
+    UnsolicitedSctList,
```
djc, line 177: Should we expand `Sct`?

rustls/tests/api.rs
```diff
@@ -3321,9 +3321,9 @@ mod test_quic {
             .unwrap();
         assert_eq!(
             server.process_new_packets().err(),
-            Some(Error::PeerMisbehavedError(
-                "QUIC transport parameters not found".into(),
-            )),
+            Some(Error::PeerMisbehaved(
+                rustls::PeerMisbehaved::MissingQuicTransportParameters
```
djc, line 3325: Style nit: maybe just import `PeerMisbehaved` in this module, as we do elsewhere?

rustls/src/error.rs
```diff
@@ -167,6 +167,7 @@ pub enum PeerMisbehaved {
     SelectedUnofferedPsk,
     SelectedUnusableCipherSuiteForVersion,
     ServerHelloMustOfferUncompressedEcPoints,
+    ServerNameDifferedOnRetry,
```
djc, line 170: Using `ServerName` here vs `Sni` in `PeerMisbehaved`, should probably be consistent at least.

rustls/src/server/hs.rs
```diff
@@ -30,14 +30,14 @@ pub(super) type NextState = Box<dyn State<ServerConnectionData>>;
 pub(super) type NextStateOrError = Result<NextState, Error>;
 pub(super) type ServerContext<'a> = crate::conn::Context<'a, ServerConnectionData>;
 
-pub(super) fn incompatible(common: &mut CommonState, why: &str) -> Error {
+pub(super) fn incompatible(common: &mut CommonState, why: PeerIncompatible) -> Error {
     common.send_fatal_alert(AlertDescription::HandshakeFailure);
-    Error::PeerIncompatibleError(why.to_string())
+    Error::PeerIncompatible(why)
```
djc, line 35: Want tot use `why.into()` here instead (also in `bad_version()`)?

rustls/tests/api.rs
```diff
@@ -3387,8 +3387,8 @@ mod test_quic {
             .unwrap();
         assert_eq!(
             server.process_new_packets().err(),
-            Some(Error::PeerIncompatibleError(
-                "Server requires TLS1.3, but client omitted versions ext".into(),
+            Some(Error::PeerIncompatible(
+                rustls::PeerIncompatible::SupportedVersionsExtensionRequired
```
djc, line 3391: Import `PeerIncompatible`?

djc (approved):
This is great!

I guess my main point of contention is where we should use acronyms/abbreviations versus expanding to full/natural name. I think we're generally a little inconsistent about this. I think `cert` is fair game as something that ~anyone will recognize but I think `kx` or `ems` will be confusing for many of our users. `Sni` is a well-known extension name but in context expanding to `ServerName` seems like an improvement, similar for `Alpn` vs `ApplicationProtocol`.

(The goal here, IMO, is to reduce our support load from issues by confused users, and we seem to already be erring on the side of pretty verbose/expansive variant names anyway...)

rustls/src/client/hs.rs
```diff
@@ -564,20 +556,20 @@ impl State<ClientConnectionData> for ExpectServerHello {
             .ok_or_else(|| {
                 cx.common
                     .send_fatal_alert(AlertDescription::HandshakeFailure);
-                Error::PeerMisbehavedError("server chose non-offered ciphersuite".to_string())
+                Error::PeerMisbehaved(PeerMisbehaved::SelectedUnofferedCipherSuite)
```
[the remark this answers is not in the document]
ctz: Unfortunately doing that meant the compiler couldn't determine the type of the closure without some extremely ugly type annotations.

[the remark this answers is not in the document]
djc: Fair enough!

rustls/src/server/hs.rs
ctz: I find `thing.into()` is a clarity vs verboseness tradeoff. Here we have `incompatible`, `bad_version`, `decode_error` functions already providing the verboseness saving, so there's no need to make them less clear (and force the reader to go and find the various `From<PeerIncompatible>`, `Into<Error>` that could be in play).

djc: Fair enough.

rustls/src/error.rs
ctz: > I think `kx` or `ems` will be confusing for many of our users.

I don't think our users will ever see these errors, and if they do they should definitely be for our eyes as they're not really user-actionable -- and that's why the docs for `PeerMisbehaved` asks for a bug to be filed. `PeerIncompatible` is arguably more user-visible and might even be user-actionable, hence why `ServerTlsVersionIsDisabledByOurConfig` is a separate error.

Also, we use "kx group" elsewhere in the public API, like `SupportedKxGroup` etc.

>`Sni` is a well-known extension name but in context expanding to `ServerName` seems like an improvement, similar for `Alpn` vs `ApplicationProtocol`.

Agree on both these though.

djc: (I would probably be in favor of renaming `SupportedKxGroup` to `SupportedKeyExchangeGroup`, FWIW.)

Honestly, even for me `Ems` is a little out there still -- not sure how commonly that acronym is used.

djc (approved):

PR #1179, opened 2023-01-23
Borrow incoming message payloads from the deframer buffer

djc:


rustls/src/msgs/deframer.rs
```diff
@@ -42,188 +43,254 @@ impl MessageDeframer {
     /// Returns an `Error` if the deframer failed to parse some message contents or if decryption
     /// failed, `Ok(None)` if no full message is buffered or if trial decryption failed, and
     /// `Ok(Some(_))` if a valid message was found and decrypted successfully.
-    pub fn pop(&mut self, record_layer: &mut RecordLayer) -> Result<Option<Deframed>, Error> {
+    pub(crate) fn deframe<Data>(
```
djc, line 46: @ctz so this commit makes the deframer responsible for driving the state machine Pre-processing messages and calling into the actual states has been moved into a `CommonState` method in prior commits, so that we can push messages from here. This avoids some problems with the current borrow checker, as discussed [here](https://users.rust-lang.org/t/tricky-borrow-checker-problems-in-the-rustls-deframer/88052/7). However, it leaves us with a pretty high-level API to the deframer which, for one thing, is not easy to drive from a fuzzer (for now, the main issue is that `State` is not public).

What do you think, should just make `State` public in `internals`? That would technically allow downstream crates to implement it, which seems icky, but at the same time I don't think they can actually do anything with that. Alternatively, we can expose a fuzzer-specific method for the deframer which doesn't need all the state to be passed in.

Thoughts?

ctz: Can we make the deframer generic over a general _takes messages and returns a state_ trait? I think that would make `TestContext` and the fuzzer not need to know about the real `State` type.

I'd be concerned in the deframer fuzzing driver contained the actual TLS handshake states, cos that will mean the fuzzer will spend quite a lot of time exploring that code rather than the whole of the deframer (there are `client` and `server` fuzzer drivers that contain everything like that).

djc: I mean, that's almost what the `State` trait is? There are no state types involved here, only the `State` trait. It has some other methods which the deframer doesn't need to call, but I don't think the existence of extra methods would influence how the fuzzer works. I guess we could create another trait on top of that specializes only for what the deframer needs and then use an implementer of that trait to wrap up the `Box<dyn State<Data>>` and `&mut Data` and `&mut CommonState`, but that seems like quite a bit of extra abstraction just for this fuzzer thingy.

ctz: I guess my opinion is that something called `MessageDeframer` should be mainly deframing messages, and ideally should:

- be loosely coupled to the next layer down,
- contain minimal policy about what messages come before or after others, or what a valid sequence of messages is,
- have minimal visibility into the rest of the library,

Perhaps this is just a case of moving and renaming it, but right now it's ~500 lines (from ~130 in 0.20) and its actual responsibilities are pretty unclear to me (though I understand now it has subsumed the handshake joiner, so some growth was always going to happen).

djc: The main responsibility is that of turning a byte stream that contains encrypted messages into a stream of `PlainMessage`s. That includes decryption and joining of fragmented messages, which means it does indeed need some knowledge of when and where fragmenting is allowed. If you prefer, we could extract the `allow_max` value out into a parameter on `read()` and `push()` to be passed into `prepare_read()`, which would allow the `CommonState` to allow a larger buffer while the connection is handshaking. Renaming might make sense as well -- maybe it should get pulled out of `msgs` and live in a module together with `RecordLayer` and `CommonState`?

The thing that's changed here is that it pushes those messages into the state machine instead of the messages getting pulled out of the deframer and then immediately getting pushed into the state machine, because today's borrow checker isn't up to the job.

djc:
In my benchmarks, this set of changes is not improving performance (bulk or handshake), and even sometimes making it worse. Given these results, it probably doesn't make sense to merge these changes given the added complexity.

PR #1180, opened 2023-01-24
Auto-shrink `MessageDeframer::buf` while at rest

larsfillmore:
Relates #794

This PR addresses https://github.com/rustls/rustls/pull/1171#discussion_r1070984963 by reducing the `MessageDeframer::buf` while it does not contain partial or full messages, which usually means that we've read everything the sender has sent so far.

The change does not affect the result of `cargo run --release --example bench`, i.e. neither handshake no bulk transfer speed is affected.

### DIFF
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -293,10 +293,12 @@ impl MessageDeframer {
         // If we can and need to increase the buffer size to allow a 4k read, do so. After
         // dealing with a large handshake message (exceeding `OpaqueMessage::MAX_WIRE_SIZE`),
         // make sure to reduce the buffer size again (large messages should be rare).
+        // Also, reduce the buffer size if there are neither full nor partial messages in it,
+        // which usually means that the other side suspended sending data.
         let need_capacity = Ord::min(allow_max, self.used + READ_SIZE);
         if need_capacity > self.buf.len() {
             self.buf.resize(need_capacity, 0);
-        } else if self.buf.len() > allow_max {
+        } else if self.used == 0 || self.buf.len() > allow_max {
             self.buf.resize(need_capacity, 0);
             self.buf.shrink_to(need_capacity);
         }


djc:
So did you test how the benchmarks cover this case? This will obviously have no ill effects if messages are always smaller than `READ_SIZE` (currently 4k bytes) and I don't know exactly if the benchmarks cover that case. It would likely still slow us down for every message larger than 4k that we get, or if a server is busy enough that connections only get to process each connection with larger intervals.

As I mentioned already, I'm at a fairly advanced stage working towards an API that allows the caller to manage memory allocations, which would allow you to customize allocation strategies for your use case.

All in all, I'm a little hesitant to merge this.

larsfillmore:
> So did you test how the benchmarks cover this case?

Yes, I ran "`cargo run --release --example bench`" on both the unmodified main and with this PR.  To remind, the benchmark output is like the following:
```text
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     max_fragment_size:default       send    1945.86 MB/s
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     max_fragment_size:default       recv    1813.87 MB/s
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     max_fragment_size:10000 send    1881.67 MB/s
bulk    TLSv1_2 TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     max_fragment_size:10000 recv    1743.22 MB/s
handshakes      TLSv1_2 Rsa     TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     client  server-auth     no-resume       4122.75 handshake/s
handshakes      TLSv1_2 Rsa     TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     server  server-auth     no-resume       1718.33 handshake/s
handshakes      TLSv1_2 Rsa     TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     client  mutual  no-resume       1340.40 handshake/s
handshakes      TLSv1_2 Rsa     TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256     server  mutual  no-resume       1455.18 handshake/s
...
```
So it tests both the handshake speed and the bulk transfer speed (1M transfers if I'm not mistaken).

Neither is affected by this PR.

I also logged the buffer behaviour in my own echo-test where the client was sending a few small messages, then a large message, then small messages again. There were no unneeded reallocations.

> It would likely still slow us down for every message larger than 4k that we get, or if a server is busy enough that connections only get to process each connection with larger intervals.

There is no slowdown. This PR shrinks `MessageDeframer::buf` when there is nothing at all in it. When the client streams a lot of data it is very likely that the buffer will always contain some remaining bytes. There will always be something like [<full_message><full_message><partial_message>] or at least [<partial_message>], which will prevent shrinking.

> As I mentioned already, I'm at a fairly advanced stage working towards an API that allows the caller to manage memory allocations, which would allow you to customize allocation strategies for your use case.

I see. My point is that this PR naturally extends #1171.

djc (approved):

djc:
Alright, thanks for the explanation! I think this makes sense to have then.

Issue #521, opened 2021-02-17
RFE: Alternate cryptographic backends


DemiMarie:
rustls currently depends on *ring* as the only crypto backend.

While *ring* is certainly a good crypto library, it has a poor story regarding backwards compatibility.  Additionally, it is statically linked, which bloats executables.

Therefore, I am filing this RFE to request support for alternate cryptographic implementations, such as OpenSSL.  This would require figuring out a story for certificate verification, since we don’t want to use OpenSSL’s certificate APIs and WebPKI only supports *ring*.  That said, on Windows and macOS there are valid reasons for just using the platform’s native certificate verification APIs instead.

djc:
See also: #390.

DemiMarie:
Another advantage of this approach is that a FIPS 140-2 certified implementation could be used where needed.  I don’t have such a need, but others do.

djc:
What is your issue with *ring*'s approach to backwards compatibility? Have you restricted the number of available ciphersuites so that executable sizes can be reduced at link time?

DemiMarie:
> What is your issue with _ring_'s approach to backwards compatibility?

Updates to _ring_ require time before they can propagate through the ecosystem; during this time, people are running an unsupported version.  If _ring_ provided security support to major version N - 1, this would not be anywhere near as much of a problem.  Hopefully, _ring_ will reach 1.0 and its API will stabilize.

> Have you restricted the number of available ciphersuites so that executable sizes can be reduced at link time?

That isn’t always practical for compatibility reasons, and it still won’t be able to beat linking to a shared system library such as OpenSSL.  The vast majority of nasty OpenSSL CVEs have been in the TLS and ASN.1 code, not the cryptography, and I suspect many of the cryptographic ones affect _ring_ too.  Using a system crypto library means that I can rely on the OS to handle security updates for it, and that those who need FIPS 140-2 compliance can use rustls at all.

Darkspirit:
https://github.com/briansmith/ring/issues/535 seems to be planned for the upcoming 0.17 release: https://github.com/briansmith/ring/pull/849#issuecomment-742863466

Could https://github.com/abetterinternet/crustls become a sufficient system package?
Otherwise it would be great if https://github.com/mesalock-linux/mesalink could become one. An OpenSSL 3.0 API might make it even more attractive for the upcoming transition. 

[...]

PR #1181, opened 2023-01-25
Split ring into a single crate

sonder-joker:
https://github.com/rustls/rustls/issues/521
>There has not been any further discussion. If you or your community wants to work on this, I'd be open to reviewing a potential design. I think essentially it would have to be a high-level set of traits that essentially mimic the current ring API to cover the rustls crypto requirements that are not currently covered by traits.

Rustls and ring coupled together, and the workload of trying to separate it in a PR will be huge, I wish it can split into many steps.

1. Change the direct dependency on ring to indirect dependency (a new crate: rustls-backend in this pr). And then we can perform further abstraction in this new crate.
2. Design trait in rustls-backend. We can replace indirect dependence on ring to abstraction of trait in to many step(e.g. hash/hmac/...) since their underylying implementations is same. 
3. After the design `cover the rustls crypto requirements that are not currently covered by traits`, we can stable them and support users to register different implementation. By default the underlying implementation should be ring. The rustls-backend trait should only expose to those want to use custom backend

This is an attempt, please correct me if there is anything wrong.

### DIFF
--- a/rustls-backend/Cargo.toml
+++ b/rustls-backend/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "rustls-backend"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+ring = "0.16.20"
--- a/rustls-backend/src/lib.rs
+++ b/rustls-backend/src/lib.rs
@@ -1,0 +1,1 @@
+pub use ring::*;
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -17,7 +17,7 @@ rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
 log = { version = "0.4.4", optional = true }
-ring = "0.16.20"
+rustls-backend = { path = "../rustls-backend" }
 sct = "0.7.0"
 webpki = { package = "rustls-webpki", version = "0.100.0-alpha.2", features = ["alloc", "std"] }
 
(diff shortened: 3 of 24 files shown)


djc:
I don't think we should have the indirection through a separate crate. Just a trait should be enough, and the *ring*-backed implementation of the trait can be made to compile conditionally using a Cargo feature.

sehz:
Many large Rust projects tended to split into smaller crates to make them maintainable and easier to understand.  Cargo workspace management is stable and much better to use than previously.

sonder-joker:
> I don't think we should have the indirection through a separate crate. Just a trait should be enough, and the *ring*-backed implementation of the trait can be made to compile conditionally using a Cargo feature.

Something like [Global allocator](https://doc.rust-lang.org/std/alloc/index.html#the-global_allocator-attribute) or [panic hook](https://doc.rust-lang.org/src/std/panicking.rs.html#96)? 
```rust
pub static IMP : Trait = RING;
```
So how should user replace it? I don 't think cargo feature can simple solve this issue.

sonder-joker:
Beside, only one trait will be too heavy? Sometime may user only want to replace one of algorithm rather all. Split it into many subtraits should be more easy to maintain/use.

djc:
See discussion in #1184.

> Many large Rust projects tended to split into smaller crates to make them maintainable and easier to understand. Cargo workspace management is stable and much better to use than previously.

Splitting into multiple crates also makes a bunch of things harder, by introducing coherence constraints, visibility issues. I think we should re-evaluate whether we need a separate crate after we've actually crated a trait + initial *ring* backed implementation, not before we've done so -- splitting into different crates is not the hard part here.

PR #1184, opened 2023-01-26
Draft RFC001: Alternative crypto backends

ctz:
This is attempting to address #521 

This so far has been written without really trying it, so a bunch of the ideas here may not work in practice.

### DIFF
--- a/rfcs/001-alternate-crypto-backends.md
+++ b/rfcs/001-alternate-crypto-backends.md
@@ -1,0 +1,100 @@
+# RFC001: Alternative crypto backends
+
+Issue: rustls/rustls#521
+
+## Why
+
+There are a bunch of reasons someone might want use crypto implementations
+other than those from *ring*.  The issue contains a thorough discussion,
+but to summarise:
+
+- runtime platform support: especially platforms like wasm who have their own
+  cryptography APIs.
+- CPU architecture support: we should have a support route for all architectures
+  that rustc can target.
+- upstream packaging requirements: RHEL has a hard policy that all cryptography
+  must come from a limited set of packages.
+- hardware support: including HSMs and cryptographic acceleration IP on microcontrollers.
+- assorted "pure rust TLS stack" reasons: including build system, memory safety, portability, etc.
+
+## Study
+
+First, let's look at where we use crypto and for what.  This includes
+uses which _already support_ alternative crypto backends via existing
+extension points.
+
+| Purpose | Current status | Example algorithm | Example code location |
+| --- | --- | --- | --- |
+| Random material generation | Statically uses *ring* | System-provided RNG | `src/rand.rs` |
+| Bulk encryption/decryption | Statically uses *ring* | AES-128-GCM | `src/cipher.rs` |
+| Certificate and signature verification | Pluggable extension point, behind `dangerous_configuration`. Default uses webpki which statically uses *ring*. | ECDSA-P256 | `src/verify.rs` |
+| Authentication key loading and signature generation | Pluggable extension point.  Default uses *ring*. | ECDSA-P256 | `src/sign.rs` |
+| Key exchange | Statically uses *ring* | X25519 | `src/kx.rs` |
+| TLS1.2 PRF | Statically uses *ring* | HMAC-SHA256 | `src/tls12/kdf.rs` |
+| TLS1.3 key schedule | Statically uses *ring* | HKDF-HMAC-SHA256 | `src/tls13/key_schedule.rs` |
+| Server ticket production | Pluggable extension point. Default does not support ticket generation. Provided optional implementation uses *ring*. | CHACHA20-POLY1305 | `src/ticketer.rs` |
+
+For the items here that statically depend on *ring*, how do we determine what
+individual algorithm to use?
+
+- Random material generation: currently fixed.
+- Bulk encryption/decryption: from cipher suite (`SupportedCipherSuite`)
+- Key exchange: from key exchange group (`SupportedKxGroup`)
+- TLS1.2 PRF: from cipher suite (`SupportedCipherSuite::Tls12`)
+- TLS1.3 key schedule: from cipher suite (`SupportedCipherSuite::Tls13`)
+
+The principal observation here is: the same mechanism which means that AES does not appear
+present in an application binary which only mentioned `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256`
+should be the same one that lets an application avoid a *ring* dependency.
+
+This gives us five remaining issues to deal with:
+
+### Random generation
+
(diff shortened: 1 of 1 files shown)


rfcs/001-alternate-crypto-backends.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Study
+
+First, let's look at where we use crypto and for what.  This includes
+uses which _already support_ alternative crypto backends via existing
+extension points.
+
+| Purpose | Current status | Example algorithm | Example code location |
+| --- | --- | --- | --- |
+| Random material generation | Statically uses *ring* | System-provided RNG | `src/rand.rs` |
```
djc, line 28: I think this is missing, such as for the handshake hashes?

rfcs/001-alternate-crypto-backends.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Move most of the crate to a `rustls-core` crate.  This should not depend on webpki, *ring* or sct.
+   But, alone, it is not useful: it does not contain any `SupportedCipherSuite`s, `SupportedKxGroup`s, etc.
+   However, it does contain all the traits and types for implementing these.
+2. `rustls` reexports most of the public parts of `rustls-core`, plus provides the default
+   `SupportedCipherSuite`s, `SupportedKxGroup`s, webpki-backed certificate verifier, loading and using
+   signing keys, early steps of config builders (as these know the default cipher suites), etc.
+
+   Most importantly: everything that `rustls` does can equally be done by a downstream crate against
+   different dependencies.
```
djc, line 88: It's really not obvious to me what we win by separating a rustls-core crate out from the rustls crate. If you're worried about coherence or visibility traps that make it impossible to implement the trait outside of rustls, I think we need a secondary implementation anyway to really flush those out but I don't think it is a strong argument for moving the trait implementation out of the rustls crate at this stage.

djc (commented):
Good idea to start doing RFCs for stuff like this (although for now I struggle to think of anything else of a similar impact)!

For me the guiding plan for this comes from the Quinn [`crypto::Session`](https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/crypto.rs#L28) trait, which IMO is quite similar in scope and has been used successfully to implement alternative cryptographic algorithms for sessions. So at least in the early stages until we get a bit more experience, I think it would make sense to sort of stick to a similar implementation strategy, since it has been proven to work and I have a bunch of experience with the approach.

djc:
It might be good to describe an initial exploration of the configuration API. I'm thinking this could be a method on `ConfigBuilder<S, WantsCipherSuites>`, like `with_ring_crypto(self)` and `with_crypto_provider(self, provider: &'static dyn CryptoProvider)` -- this would be bypassed by `with_safe_defaults()` which implicitly uses the *ring* provider.

rfcs/001-alternate-crypto-backends.md
ctz: My thought here was that _something_ needs to take the dependency on *ring*. If we have that in one crate, then we must rely on cargo features to enable/disable that dependency and that comes with the usual issue that (for a large enough application with enough dependencies) it's actually really annoying to disable a default feature. It seems easier to me for big applications to just say `[patch] rustls = { package = "rustls-wasi-crypto" }` or whatever.

djc: Hmm, it's an interesting idea. So far I've always considered `patch` entries to be temporary, so (as the maintainer of a semi-large application in my day job) I would hesitate to use that as a long term strategy. I also don't know how well it deals with version drift over time, can you `patch` rustls 0.22 and 0.23 separately?

Anyway, I would still maintain that we should work on the trait(s) first and consider splitting up the crate as more of an endgame thing, it's mostly orthogonal to the design anyway.

Alternatives: make `ring` not a default feature, so downstream libraries have to make an explicit choice to enable it, to raise the salience of crypto provider variance. Or maybe more far out, split out the `Config` types and builder into a separate crate which can maintain a more stable API (that is, 1.0), making it more feasible for downstream crates to expose that as part of their own public API.

rfcs/001-alternate-crypto-backends.md
```diff
@@ -0,0 +1,100 @@
...
+| Purpose | Current status | Example algorithm | Example code location |
+| --- | --- | --- | --- |
+| Random material generation | Statically uses *ring* | System-provided RNG | `src/rand.rs` |
+| Bulk encryption/decryption | Statically uses *ring* | AES-128-GCM | `src/cipher.rs` |
+| Certificate and signature verification | Pluggable extension point, behind `dangerous_configuration`. Default uses webpki which statically uses *ring*. | ECDSA-P256 | `src/verify.rs` |
+| Authentication key loading and signature generation | Pluggable extension point.  Default uses *ring*. | ECDSA-P256 | `src/sign.rs` |
+| Key exchange | Statically uses *ring* | X25519 | `src/kx.rs` |
+| TLS1.2 PRF | Statically uses *ring* | HMAC-SHA256 | `src/tls12/kdf.rs` |
+| TLS1.3 key schedule | Statically uses *ring* | HKDF-HMAC-SHA256 | `src/tls13/key_schedule.rs` |
+| Server ticket production | Pluggable extension point. Default does not support ticket generation. Provided optional implementation uses *ring*. | CHACHA20-POLY1305 | `src/ticketer.rs` |
```
DemiMarie, line 35: Another often-forgotten fact is that most hardware-accelerated cryptography is asynchronous.  Hardware accelerators have very deep pipelines with substantial latency, so one must have a large number of requests in flight to get good performance.  In the case of rustls, this means continuing to receive data and enqueue decryption requests before knowing if the earlier decryption requests succeeded.  Hardware accelerators are often many times faster than the CPU clusters they are attached to and use much less power, so it is worthwhile to support them if possible.

rfcs/001-alternate-crypto-backends.md
DemiMarie: `patch` is a bad idea for production.

> make `ring` not a default feature, so downstream libraries have to make an explicit choice to enable it, to raise the salience of crypto provider variance.

What about requiring a `Crypto` trait to be passed to every builder?

dignifiedquire:
FYI I didn't write an RFC, but I started on isolation of the actual crypto primitives here: https://github.com/dignifiedquire/rustls/tree/feat-crypto-backend

rfcs/001-alternate-crypto-backends.md
dignifiedquire: One alternative would be to standardize over the RustCrypto/traits and use https://crates.io/crates/ring-compat to fill in the ring version.

ctz:
The ideas in this RFC were implemented in the recent 0.22 release, so I'm closing this. See the current documentation: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html

PR #1189, opened 2023-01-31
Harmonize certificate verifiers construction API

### LINKED ISSUE #1106: Change `AllowAnyAuthenticatedClient::new` to return `Self`
This is a nice building block to implement client certificate verification, it's even used internally for another verifier in rustls.

Currently, `new` returns `Arc<dyn ClientCertVerifier>` instead of `Self`, creating unneeded indirection and overhead for outside users. Returning `Self` makes the API more flexible and consistent with `WebPkiVerifier`.

We could also go ahead and modify `AllowAnyAnonymousOrAuthenticatedClient::new` too, for consistency.

thalesfragoso:
Closes #1106

This turned out less elegant than first expected. The issue is that `ClientCertVerifier` is only reachable with the `dangerous_configuration` feature enabled.

Usually, the compiler is smart enough to do the coercion automatically. However, there are a few cases where you might need to coerce it in advance, I actually hit one such case in `tlsserver-mio`, where the compiler would delay the coercion on a nested if/else and cause a compilation error.

For the motivation, please see the issue and/or commit message.

Should I also add the changes description directly to the README ?

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -400,7 +400,7 @@ fn make_server_cfg(opts: &Options) -> Arc<rustls::ServerConfig> {
                 mandatory: opts.require_any_client_cert,
             })
         } else {
-            rustls::server::NoClientAuth::new_coerced()
+            rustls::server::NoClientAuth::boxed()
         };
 
     let cert = load_cert(&opts.cert_file);
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -400,7 +400,7 @@ fn make_server_cfg(opts: &Options) -> Arc<rustls::ServerConfig> {
                 mandatory: opts.require_any_client_cert,
             })
         } else {
-            rustls::server::NoClientAuth::new()
+            rustls::server::NoClientAuth::new_coerced()
         };
 
     let cert = load_cert(&opts.cert_file);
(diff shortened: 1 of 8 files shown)


rustls/src/verify.rs
```diff
@@ -634,8 +650,12 @@ fn pki_error(error: webpki::Error) -> Error {
 pub struct NoClientAuth;
 
 impl NoClientAuth {
-    /// Constructs a `NoClientAuth` and wraps it in an `Arc`.
-    pub fn new() -> Arc<dyn ClientCertVerifier> {
+    /// Constructs a [`NoClientAuth`], wraps it in an [`Arc`] and coerces it to
```
thalesfragoso, line 653: This documentation was already using the third person while the other two were using the imperative mood. Should I change all to third person or the other way around ?

thalesfragoso:
This is a breaking change, I didn't manage to add the label myself though.

djc:
The `coerce()` stuff seems a little ugly. Could we add a `From` or `Into` impl instead?

thalesfragoso:
The quirk is that `ClientCertVerifier` is essentially sealed, so things that would work easily become tricky.

thalesfragoso:
Updated with the requested changes.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -400,7 +400,7 @@ fn make_server_cfg(opts: &Options) -> Arc<rustls::ServerConfig> {
                 mandatory: opts.require_any_client_cert,
             })
         } else {
-            rustls::server::NoClientAuth::new()
+            rustls::server::NoClientAuth::new_coerced()
```
djc, line 403: Looks like you missed this one?

rustls/src/verify.rs
djc: Let's bring this one in line with the others.

djc (approved):

djc:
Thanks!

PR #1190, opened 2023-02-02
client: add support for sending unsupported cipher suites

### LINKED ISSUE #1125: Allow offering ciphersuites in the Client Hello that are not yet supported.
This is very edge-case-y.  A strict set and ordering of ciphersuites is required by Google's server-side authentication in order to proceed with a login: see https://stackoverflow.com/questions/22832104/how-can-i-see-hidden-app-data-in-google-drive.

Attempting to migrate https://github.com/EFForg/rs-google-play/ from `openssl`, I encountered the limitation that only supported ciphersuites are offered in the Client Hello.  Of course, this is a very reasonable assumption.  Unfortunately, in my case Google's ciphersuite strictures will limit me from migrating, as a login fails when using the `ClientConfig` to offer the ciphers supported by `rustls`.

Ideally, I would be able to specify a list of ciphers I would like to proceed with, even if they aren't supported, knowing that Google will actually pick a supported cipher in the Server Hello: TLS_AES_256_GCM_SHA384 (0x1302)

Here is my attempt at migrating, valid but login fails: https://github.com/EFForg/rs-google-play/commit/4779fde5f38cec89112c9c68bd858883c0f0adab
Corresponding issue: https://github.com/EFForg/apkeep/issues/152

djc:
Fixes #1125. @ctz I know you said that you didn't want to support this, but I just wanted to sketch out for myself how much complexity it is to implement this given that there have been increasing requests for this (I don't like that this is necessary either, but I fear it might be a reality of competing with OpenSSL). IMO the resulting change is not too bad. If you're open to it, I'll add some tests.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -187,6 +187,7 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
             #[cfg(feature = "secret_extraction")]
             enable_secret_extraction: false,
             enable_early_data: false,
+            unsupported_suites: Vec::new(),
         }
     }
 }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -1,12 +1,10 @@
 use crate::builder::{ConfigBuilder, WantsCipherSuites};
 use crate::conn::{CommonState, ConnectionCommon, Protocol, Side};
 use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
-use crate::error::Error;
+use crate::error::{Error, PeerMisbehaved};
 use crate::kx::SupportedKxGroup;
 #[cfg(feature = "logging")]
 use crate::log::trace;
-
-#[cfg(feature = "quic")]
 use crate::msgs::enums::AlertDescription;
 use crate::msgs::enums::NamedGroup;
 use crate::msgs::handshake::ClientExtension;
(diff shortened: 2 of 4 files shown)


ctz:
Unfortunately #1125 is an [XY](https://en.wikipedia.org/wiki/XY_problem) issue. Actually what we need to do to get the same JA3 fingerprint as another client is:

- have the same order of cipher_suites (adding them on the end isn't enough)
- have the same order and presence of extensions (this might mean implementing whole new TLS features)
- have the same order and presence of named groups (we might have enough configurability here already)
- have the same order and presence of ec point formats (i think we're good here, uncompressed=0 on its own is ubiquitous)

djc:
So if we scope this to (a) adding unsupported cipher suites, (b) sorting cipher suites, (c) sorting extensions, it feels like we'd have made some progress, and this seems somewhat straightforward? We could have a

```rust
trait ClientHelloModifier {
    fn update_cipher_suites(&self, suites: &mut Vec<CipherSuite>);
    fn update_extensions(&self, extensions: &mut [ClientExtension]);
}
```

(Link to JA3: https://github.com/salesforce/ja3.)

ctz:
Since this was last discussed, Chrome shipped randomised extension ordering as a countermeasure to fingerprinting:

- https://chromestatus.com/feature/5124606246518784
- https://github.com/salesforce/ja3/issues/88

I think I'd prefer to expend effort doing the same (#1313) which is really the antithesis of this PR. Thoughts?

FireMasterK:
I think this pull request should be re-opened since the extensions could be sorted, thus allowing for fingerprintable `ClientHello`s again. Such techniques may already be implemented on some sites as https://browserleaks.com/tls, https://tlsfingerprint.io/norm_fp, etc suggest.

djc:
@FireMasterK please open a new issue with your concrete use cases.

PR #1192, opened 2023-02-05
improve & test ServerName debug output - workaround

brodybits:
**hopefully** resolves #1177

**before:**

result from `format!("{:?}", rustls::ServerName::try_from("a.com"))`:

    Ok(DnsName(DnsName(DnsName(\"a.com\"))))

**after:**

result from `format!("{:?}", rustls::ServerName::try_from("a.com"))`:

    Ok(DnsName(\"a.com\"))

**also tested:**

`rustls::ServerName::IpAddress("127.0.0.1".parse().unwrap())` formats as:

    IpAddress(127.0.0.1)

As I said in 3a3e203d51f08463719513f0bd6cb5f7716631cb, I think the root cause lies in the difference between the standard debug formatting of tuples vs enums in Rust.

Quick disclaimer that I just started learning Rust after many years of programming and this is actually my first ever contribution in this language. I made a simple workaround solution for now but think it would be ideal to consider further improvement in the future. Please feel free to leave **candid** feedback or even rework before merging if needed.

I would personally favor merging this one in a squash commit and would be happy to squash the changes myself if needed.

PR #1193, opened 2023-02-05
Remove DnsName wrapper & ServerName debug formatting

### LINKED ISSUE #1177: std::fmt::Debug for `ServerName` is extremely odd-looking
Highly ugly:

```
DnsName(DnsName(DnsName("google.com")))
```

however IP addresses look good:

```
IpAddress(1.1.1.1)
```

brodybits:
alternate solution to resolve #1177 - _potentially controversial_

removes an internal, private wrapper that was documented as "dangerous_configuration"

tests from PR #1192

result from `format!("{:?}", rustls::ServerName::try_from("a.com"))`

    ServerName::IpAddress(127.0.0.1)

result from `rustls::ServerName::IpAddress("127.0.0.1".parse().unwrap())`:

    Ok(ServerName::DnsName(\"a.com\"))

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -272,7 +272,7 @@ pub enum ServerName {
     /// The server is identified by a DNS name.  The name
     /// is sent in the TLS Server Name Indication (SNI)
     /// extension.
-    DnsName(verify::DnsName),
+    DnsName(webpki::DnsName),
 
     /// The server is identified by an IP address. SNI is not
     /// done.
@@ -285,7 +285,7 @@ impl ServerName {
     /// in the handshake.
     pub(crate) fn for_sni(&self) -> Option<webpki::DnsNameRef> {
         match self {
-            Self::DnsName(dns_name) => Some(dns_name.0.as_ref()),
+            Self::DnsName(dns_name) => Some(dns_name.as_ref()),
             Self::IpAddress(_) => None,
         }
     }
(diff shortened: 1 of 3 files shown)


rustls/tests/api.rs
```diff
@@ -4611,3 +4611,22 @@ fn test_received_plaintext_backpressure() {
...
 }
+
+#[test]
+fn test_debug_server_name_from_ip() {
+    assert_eq!(
+        format!(
+            "{:?}",
+            rustls::ServerName::IpAddress("127.0.0.1".parse().unwrap())
+        ),
+        "ServerName::IpAddress(127.0.0.1)"
```
brodybits, line 4622:
```suggestion
        "ServerName(IpAddress(127.0.0.1))"
```

rustls/tests/api.rs
```diff
@@ -4611,3 +4611,22 @@ fn test_received_plaintext_backpressure() {
...
+    )
+}
+
+#[test]
+fn test_debug_server_name_from_string() {
+    assert_eq!(
+        format!("{:?}", rustls::ServerName::try_from("a.com")),
+        "Ok(ServerName::DnsName(\"a.com\"))"
```
brodybits, line 4630:
```suggestion
        format!("{:?}", rustls::ServerName::try_from("a.com").unwrap()),
        "ServerName::DnsName(\"a.com\")"
```

brodybits (commented):
I think this should be updated to give some more idiomatic output ... still learning on my side!

rustls/src/client/client_conn.rs
```diff
@@ -267,12 +267,12 @@ impl ClientConfig {
...
 /// ```
 #[non_exhaustive]
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Eq, Hash, PartialEq)]
 pub enum ServerName {
     /// The server is identified by a DNS name.  The name
     /// is sent in the TLS Server Name Indication (SNI)
     /// extension.
-    DnsName(verify::DnsName),
+    DnsName(webpki::DnsName),
```
djc, line 275: The reason we don't want this is that this exposes the `webpki` crate as part of the public API, which means that the public API for webpki and rustls have to evolve in lockstep (in particular, rustls cannot start using a semver-incompatible new version like 0.101 without also bumping our own version in a semver-incompatible way). Therefore, I don't think we want to make this change.

brodybits:
> this exposes the `webpki` crate as part of the public API

agreed on my part thanks!

PR #1195, opened 2023-02-10
Lints: use Self and fix missing semicolons

nyurik:
* fix [semicolon_if_nothing_returned](https://rust-lang.github.io/rust-clippy/master/index.html#semicolon_if_nothing_returned) clippy lint
* fix [use_self](https://rust-lang.github.io/rust-clippy/master/index.html#use_self) clippy lint

This makes the code a bit cleaner, esp the self lint, when browsing the code in an IDE

Other minor fixes:
* remove non-needed `use log;`
* remove a few unneeded `&`
* copied comment explain why env_logger is pinned

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -15,7 +15,7 @@ quic = ["rustls/quic"]
 
 [dependencies]
 docopt = "~1.1"
-env_logger = "0.9.0"
+env_logger = "0.9.0" # 0.10 requires an MSRV bump to 1.60
 log = { version = "0.4.4", optional = true }
 mio = { version = "0.8", features = ["net", "os-poll"] }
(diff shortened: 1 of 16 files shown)


djc (approved):
Thanks! It would be cool if we can systematically get clippy applied to all targets in the repo. I think we run it in CI only for the lib so far (not the examples/tests and possibly not the examples crate).

nyurik:
@djc getting clippy to fail CI is easy (the last `-W ...` is optional :) ).  

```bash
cargo fmt --all -- --check
cargo clippy --all-targets --all-features -- -D warnings -W clippy::pedantic
```

The trickier part is not to get frustrated when CI fails because of a missing space :)

Issue #578, opened 2021-03-17
Allow disabling Hostname Verification


roblabla:
Hello,

I'm working on an application that uses `rustls` to secure its client/server communications. We have our own private Root CA we use to make sure our communications are secure. In our threat model, the hostname validation isn't particularly useful: If the remote server has a valid certificate, it doesn't really matter through which domain name we connected to it, we know the connection is legitimate. Furthermore, it's causing us a fair amount of pain: The certificates being bound to a particular domain name means moving services to new domain names requires reprovisioning a certificate.

We could implement this ourselves by providing a custom `ServerCertVerifier` that simply doesn't run the last step [here](https://github.com/ctz/rustls/blob/eacc251a27164d86071e74cfb3f45d8d7f940fff/rustls/src/verify.rs#L314). But copy pasting rustls' verifier seems to be fairly dangerous. Ideally, it'd be nice if `rustls` provided a `DangerousClientConfig::disable_hostname_verification` function that would gate the call to `cert.verify_is_valid_for_dns_name`.

djc:
I don't think rustls should be in the business of facilitating different kinds of dangerous configuration. If you want to do something dangerous, then the risk of implementing a custom `ServerCertVerifier` is on you, which seems reasonable to me.

Why do you think copy pasting rustls's verifier is fairly dangerous?

sayrer:
Could you make yourself a wildcard cert for a `.corp` domain? That seems like a better approach than changing rustls.

ctz:
I don't think we're going to do this, and the supported way to achieve this is to supply your own `ServerCertVerifier`.

lambdalisue:
I faced same request and solved with the following code. In case you are interested in.

```rust
struct NoCertVerifier {}

impl rustls::ServerCertVerifier for NoCertVerifier {
    fn verify_server_cert(
        &self,
        _: &rustls::RootCertStore,
        _: &[rustls::Certificate],
        _: webpki::DNSNameRef<'_>,
        _: &[u8],
    ) -> Result<rustls::ServerCertVerified, rustls::TLSError> {
        Ok(rustls::ServerCertVerified::assertion())
    }
}

// ...

let mut config = ClientConfig::new();
config.dangerous().set_certificate_verifier(Arc::new(NoCertVerifier {}));
```

Note that I needed to add `webpki` crate to access `DNSNameRef` as well. I wonder if `rustls` re-export that for user who don't need to access that directly.

roblabla:
[...]

PR #1197, opened 2023-02-14
Dangerous verifiers API proposal

nyurik:
This is a rough draft of the possible API implementation to handle common but not secure existing cases of bad SSL usage,
 e.g. by PostgreSQL. This way users can use dangerous verifiers that have well understood and documented risk factors, rather than let each user shoot themselves in the foot in the most creative way. Plus this helps with the OpenSSL -> Rustls migrations.

Per @djc https://github.com/rustls/rustls/issues/578#issuecomment-1429496850

This should give a good start to discuss the API, and once agree, we can actually implement the verifiers. I used the no-certs verifier as an example because of its simplicity.

The implementation for the no-hostname verificator is missing - I am not knowledgeable enough in the area to actually implement it. Probably something similar to https://github.com/rustls/rustls/issues/578#issuecomment-816712636 by @roblabla ?

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -327,6 +327,28 @@ pub(super) mod danger {
     use super::verify::ServerCertVerifier;
     use super::ClientConfig;
 
+    /// Private no-certificate verifier. Usage example:
+    /// ```ignore
+    /// client_config
+    ///     .dangerous()
+    ///     .set_certificate_verifier(rustls::client::DangerousClientConfig::new_allow_all_verifier());
+    /// ```
+    struct NoCertificateVerification {}
+
+    impl ServerCertVerifier for NoCertificateVerification {
+        fn verify_server_cert(
+            &self,
+            _end_entity: &crate::Certificate,
+            _intermediates: &[crate::Certificate],
+            _server_name: &crate::ServerName,
+            _scts: &mut dyn Iterator<Item = &[u8]>,
+            _ocsp: &[u8],
+            _now: std::time::SystemTime,
+        ) -> Result<crate::client::ServerCertVerified, crate::Error> {
+            Ok(crate::client::ServerCertVerified::assertion())
+        }
+    }
+
     /// Accessor for dangerous configuration options.
     #[derive(Debug)]
     #[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
@@ -339,6 +361,36 @@ pub(super) mod danger {
         /// Overrides the default `ServerCertVerifier` with something else.
         pub fn set_certificate_verifier(&mut self, verifier: Arc<dyn ServerCertVerifier>) {
             self.cfg.verifier = verifier;
+        }
+
+        /// Create a new dangerous verifier that skips all certificate verification.
+        /// This verifier would only encrypt connection to prevent eavesdropping,
+        /// but will still allow a man-in-the-middle attacks and connections to unverified
+        /// counterparties - i.e. you trust the network not to be compromised.
+        pub fn new_allow_all_verifier() -> Arc<dyn ServerCertVerifier> {
+            Arc::new(NoCertificateVerification {})
+        }
+
+        /// Create a new dangerous verifier that skips certificate verification.
+        /// Will verify the certificate is valid in the following ways:
+        /// - Signed by a valid root
+        /// - Not Expired
+        ///
+        /// TODO: describe dangers of this verification method, possibly adapting
+        /// from the PostgreSQL SSL info page <https://www.postgresql.org/docs/15/libpq-ssl.html>
+        /// > The difference between verify-ca and verify-full depends on the policy of the root CA.
+        /// > If a public CA is used, verify-ca allows connections to a server that somebody else
+        /// > may have registered with the CA.
+        ///
+        /// In PG, verify-full implies hostname is matched against SAN, or CN if no SAN of type dNSName is present,
+        /// but verify-full is less commonly used, so this verifier would skip this verification.
+        ///
+        /// TODO: The above description is only for the PR discussion, and should be fixed before merging.
+        ///
+        pub fn new_ignore_hostname_verifier() -> Arc<dyn ServerCertVerifier> {
+            // TODO: replace this with a different verifier, possibly based on
+            // https://github.com/rustls/rustls/issues/578#issuecomment-816712636
+            Arc::new(NoCertificateVerification {})
         }
     }
 }


rustls/src/client/client_conn.rs
```diff
@@ -340,6 +362,36 @@ pub(super) mod danger {
         pub fn set_certificate_verifier(&mut self, verifier: Arc<dyn ServerCertVerifier>) {
             self.cfg.verifier = verifier;
         }
+
+        /// Create a new dangerous verifier that skips all certificate verification.
+        /// This verifier would only encrypt connection to prevent eavesdropping,
+        /// but will still allow a man-in-the-middle attacks and connections to unverified
+        /// counterparties - i.e. you trust the network not to be compromised.
+        pub fn new_allow_all_verifier() -> Arc<dyn ServerCertVerifier> {
```
djc, line 370: Conceptually a `new_()` style API doesn't make sense. Like `set_certificate_verifier()`, `DangerousClientConfig` methods should set the internal verifier. So this would be more like `without_verification()`.

rustls/src/client/client_conn.rs
```diff
@@ -340,6 +362,36 @@ pub(super) mod danger {
...
+        /// > The difference between verify-ca and verify-full depends on the policy of the root CA.
+        /// > If a public CA is used, verify-ca allows connections to a server that somebody else
+        /// > may have registered with the CA.
+        ///
+        /// In PG, verify-full implies hostname is matched against SAN, or CN if no SAN of type dNSName is present,
+        /// but verify-full is less commonly used, so this verifier would skip this verification.
+        ///
+        /// TODO: The above description is only for the PR discussion, and should be fixed before merging.
+        ///
+        pub fn new_ignore_hostname_verifier() -> Arc<dyn ServerCertVerifier> {
```
djc, line 390: I guess I'm still not exactly clear on the use case here. Given that rustls on main has support for certificates with IP addresses instead of a host name, it should always be possible to cheapyly/simply verify the identity of the server somehow?

roblabla: Here's my use-case back when I used rustls (I don't anymore, for unrelated reasons): We have a software that has a client/server architecture and is often installed on-premise, inside a user's internal network.

To secure the client/server comms, we use TLS, but we don't use the public PKI because some users do not want to have "publicly known" domain names on their internal network. Instead, we use our own root certificate, and give our clients a leaf certificate that chains back to it.

We don't actually *care* about the hostname. We authenticate the server by having a sort of userId field in the leaf cert. In our model, how we connect doesn't matter - a cert that chains to our root with the right userId is the guarantee that we're talking to the right server. Internal networks don't always have domain names, and sometimes the IP ranges will change because of an internal reorg or something. Having our users re-issue TLS certificates when that happens isn't desirable for us.

I guess "technically", we could use the hostname verification and pass the userId as a hostname. That feels like a terrible hack though, and isn't compatible with how we generate our certs today. Plus, it's not exactly compatible with a lot of libraries that use rusTLS under the hood and don't allow setting a different hostname.

djc: Fair enough, thanks for the additional context!

rustls/src/client/client_conn.rs
nyurik: @djc so something like this?

```rust
// client code
client_config.dangerous().without_verification();

// implementation
    impl<'a> DangerousClientConfig<'a> {
        pub fn without_verification(&mut self) {
            self.set_certificate_verifier(Arc::new(NoCertificateVerification {}))
        }
```

djc: Yes, and then including a bunch of docstring that explains the risks and when this could possibly be okay.

rustls/src/client/client_conn.rs
djc: So for these use cases I think you should be able to copy the `ServerCertVerifier` impl for `WebPkiVerifier` and leave off the calls to `cert.verify_is_valid_for_subject_name()`. Might be good to factor out the remaining code into a separate method.

nyurik: would it make sense to add a non-accessible bool flag inside the `WebPkiVerifier`, which will only be modified by the `danger().without_hostname_verification()` method?  Or you think there should be two totally separate structs?

djc: Yeah, a private flag for this seems reasonable to me although maybe it should be a custom email to make it extra obvious what's going on (along the lines of `enum VerifySubjectName { Always, InsecureIgnore }` or something).

ctz:
I'm not really sure about this. A few points on my mind:

- please, let us not add escape hatches to security-critical code to address insecure use cases.
- i think the place for the proposed library of verifiers that satisfies postgres's bad certificate verification UX ("required" means "do nothing"? huh?) it seems especially likely that webpki is not the correct choice for verifying non-webpki certificates (eg, they are more likely to be illegal when viewed through the lens of webpki baseline requirements)
- there's not must justification i can see for this being in the core crate; i would accept them being in a crate inside the rustls/ organisation. happy to hear counterarguments here though.

djc:
@ctz I am also ambivalent about it, but here are a few considerations why it might still be a good idea.

1. At the macro level, are people more secure if they stick to using OpenSSL because rustls makes the configuration they need to hard to use? We can't help them be safer at all if they're prevented from using rustls in the first place.
2. The crate already supports all of this. Making a thing that, in the end, a sizable portion of the users need, harder to use feels spiteful. Yes, we should put up boundaries to make it clear that they're doing things that are unsafe, but I feel like this PR can satisfy that requirement.
3. Particularly the use case where a certificate is still verified but without verifying the host name, in a scenario with a private CA, seems like not such an insecure set up. I suppose we could set up a different verifier that doesn't take a `RootCertStore` but instead takes a single CA? But then key rotation becomes hard, you need more than one, and that's more or less just equivalent to having a `RootCertStore`.
4. As for the separate crate, I suppose we could do that. But to me, there's not a huge difference between `config.dangerous().without_verification()` and `builder.set_custom_verifier(Box::new(rustls_insecure::NoVerification))`.

If we want rustls to be more broadly used, including in non-web use cases, I think we should find some way of facilitating these use cases (without involving cargo culting around empty trait impls). To me, the more interesting path is to push folks towards the safest path while also providing modular access to as much security as they can, with nudges that push people to assess the risks.

nyurik:
> * please, let us not add escape hatches to security-critical code to address insecure use cases.
Agree, secure core could expose reusable components.

> * i think the place for the proposed library of verifiers that satisfies postgres's bad certificate verification UX ("required" means "do nothing"? huh?) it seems especially likely that webpki is not the correct choice for verifying non-webpki certificates (eg, they are more likely to be illegal when viewed through the lens of webpki baseline requirements)
Sure, we can have a dummy `NoHostnameVerification` obj that does the same steps as webpki up until the hostname check. And just for clarification - it is not "do nothing", it is "prevent network sniffing, but trust network to route correctly to a trusted party" - see [pg docs](https://www.postgresql.org/docs/current/libpq-ssl.html#:~:text=security%20and%20overhead.-,Table,-34.1.%C2%A0SSL%20Mode).

> * there's not must justification i can see for this being in the core crate; i would accept them being in a crate inside the rustls/ organisation. happy to hear counterarguments here though.
This I think is not a good solution. I don't believe there is a significant security gains from having two crates, but there is a high maintenance burden of maintaining separate code bases that may result in security lapses.  Even more importantly, a higher barrier of entry will cause much worse security overall -- more projects will continue using openssl because of the migration costs.  We do not exist in a perfect vacuum -- we must support existing code base that uses openssl and have a simple migration plan for those projects. Having two independent crates will over-complicate this even further. Plus I suspect you don't propose to move everything including the `dangerous_configuration` feature to a separate crate, and somehow have a backdoor into the crate to still disable some parts (clearly a weird solution)?

rustls/src/client/client_conn.rs
nyurik: @djc it seems if I create the `.danged().without_hostname_verification()`, I will not be able to tell what verifier is being used before that call - because by that time the `ConfigBuilder` instance is gone, and the verifier itself is unknown (I only have a dyn trait of unknown type).  What are your thoughts on how I can migrate the ClientConfig state at that point  to the new `NoHostnameVerification` object?

ctz:
> Agree, secure core could expose reusable components.

👍🏽 

> This I think is not a good solution. I don't believe there is a significant security gains from having two crates

I think the core crate should contain the protocol implementation with extension points for its configuration surface, and safe defaults for those extension points. The less safe implementations of those extension points should be elsewhere, because they "infect" the trusted computing base of the 99% of applications that do not need or want them.

nyurik:
@ctz the dangerous code will be feature-flagged - doesn't need to infect anyone who doesn't need them.  Exposing some internal dangerous interface to another crate on the other hand would somewhat compromise that. So we shouldn't conflate "clean secure interface" with the "internal code structure".  First is clearly needed, but it does not equate to "must have separate crates". It could be separate modules with conditional on-demand compilation.

djc:
After talking to a bunch of people about this, I no longer think we should merge something like this. Specifically for the Postgres case, I think it would be reasonable for tokio-postgres-rustls to carry a modular verifier implementation that can handle some of the less secure modes that Postgres specifies.

nyurik:
@djc could you elaborate on the reasoning?  I have stated repeatedly why I think it is critical for 1) adaption/migration from openssl and 2) to reduce potential for incorrect multiple identical implementations. What are the reasons not to do it as part of Rustls?

djc:
@nyurik I don't think we've seen that this is substantial barrier for adoption/migration from OpenSSL. Additionally, we provide a pluggable API so it's relatively straightforward for users to plugin a custom verifier impl that does what they need. As we've seen, there's a decent amount of diversity in exactly what different protocols/use cases want to verify vs what they are comfortable not verifying, and so providing implementations that cover all of these use cases is a distraction for the rustls project, which focuses on (a) providing default-secure implementations, and (b) covering most common use cases.

pielgrzym:
This is a prime example of this approach: https://github.com/rustls/rustls/issues/1291 - unforgiving, unfriendly API is pushing developers to introduce security vulnerabilities.

pielgrzym:
Additionally - even if developer is patient and will untagle WebPkiServerVerifier trait and try to implement his own verifier that supports partial verification there are major caveats. Most of logic used in this trait is relying on private API from rustls::webpki module. Reimplementing this on your own is not only more error prone, but makes maitenance more complicated be there any important changes made to upstream code on which such verifier relies.

ctz:
> While I admire the security-enforsing design, some valid usecases for lessening security (like cert validation without hostname/ip verification)

Could you be precise where this is a valid use case?

[Past research](https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf) concluded:

> "In particular, APIs should present high-level abstractions to developers, such as “confidential and authenticated tunnel,” as opposed to requiring them to explicitly deal with low-level details such as hostname verification."

> due to lack of proper examples

We have an example of this in `tlsclient-mio.rs`

pielgrzym:
Thanks for quick response.

> Could you be precise where this is a valid use case?

When you try to implement bilateral certificate based authentication when both sides have dynamic IP addresses and generating certificates with hardcoded IP in SAN is impractical. Each side shares just internal CA certificate and verifies certificate authenticity by checking is cert is signed by proper CA. Having a high level abstraction to handle such scenario is obviously great, but there is no such abstraction at the moment. It's much more likely to use a dummy verifier that to say, use a public CA in a ready astraction of such auth scheme.

PR #1198, opened 2023-02-15
msgs: yield Result instead of Option from Reader/Codec

djc:
Built on top of #1172. I think it's good that this produces more fine-grained errors.

cc @complexspaces


cpu (approved):
This is a nice change :rocket: 

I reviewed this diff primarily from the context of ensuring existing semantics didn't change, and that the returned error types/messages matched their context of use. I didn't catch anything out of the ordinary.

ctz (approved):
🥇

PR #1199, opened 2023-02-22
Allow user provides more specified certificate validation error as AlertDescription in verifier

zkonge:
As I said in https://discord.com/channels/976380008299917365/1015156984007381033/1040188576727572530
In this pull request, following change applied:
## Align server and client certificate verification mechanisms
Now, the server side and client side `verifier` will act same behavior when return `Erro`. See the new `conn::send_cert_error_alert`.

Before this change, the server side verifier only sends `BadCertificate` in certificate chain verification and `AccessDenied` in TLS message verification. The client side sends, well, just see old `client::hs::send_cert_error_alert`.
## Allow user provides more specific certificate validation error
After reading RFC, choice in [OpenSSL Implementation](https://github.com/openssl/openssl/blob/45bb98bfa223efd3258f445ad443f878011450f0/ssl/statem/statem_lib.c#L1434) and choice in [BoringSSL Implementation](https://github.com/google/boringssl/blob/583c60bd4bf76d61b2634a58bcda99a92de106cb/ssl/ssl_x509.cc#L1361), 
I choose the following new `Error` to `TLS Alert` mapping, see `rustls/src/error.rs`.

---
## Potential break change?
Of course, each peer may receive different `TLS Alert`, but after fault alert, connection will be closed, I think it doesn't matter.

In API level, there isn't any change.

---
By the way, I didn't implement but how about allow user send specified `TLS Alert` directly in their owned verifier instead of using provided `Error` mapping?

### DIFF
--- a/bogo/config.json
+++ b/bogo/config.json
@@ -380,8 +380,10 @@
   },
   "TestLocalErrorMap": {
     "SendServerHelloAsHelloRetryRequest": "remote error: error decoding message",
-    "GarbageCertificate-Server-TLS12": "remote error: access denied",
-    "GarbageCertificate-Server-TLS13": "remote error: access denied",
+    "GarbageCertificate-Server-TLS12": "remote error: bad certificate",
+    "GarbageCertificate-Server-TLS13": "remote error: bad certificate",
+    "GarbageCertificate-Client-TLS12": "remote error: bad certificate",
+    "GarbageCertificate-Client-TLS13": "remote error: bad certificate",
     "Client-VerifyDefault-RSA_PKCS1_SHA1-TLS12": "tls: no common signature algorithms",
     "Server-VerifyDefault-RSA_PKCS1_SHA1-TLS12": "tls: no common signature algorithms",
     "Downgrade-TLS10-Client": "tls: no cipher suite supported by both client and server",
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -325,9 +325,16 @@ impl From<CertificateError> for AlertDescription {
         use CertificateError::*;
         match e {
             BadEncoding | UnhandledCriticalExtension | NotValidForName => Self::BadCertificate,
+            // RFC 5246/RFC 8446
+            // certificate_expired
+            //  A certificate has expired or **is not currently valid**.
             Expired | NotValidYet => Self::CertificateExpired,
             UnknownIssuer => Self::UnknownCA,
             BadSignature => Self::DecryptError,
+            // RFC 5246/RFC 8446
+            // certificate_unknown
+            //  Some other (unspecified) issue arose in processing the
+            //  certificate, rendering it unacceptable.
             Other(_) => Self::CertificateUnknown,
         }
     }
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -317,6 +317,22 @@ pub enum CertificateError {
     Other(Arc<dyn StdError + Send + Sync>),
 }
 
+// The following mapping are heavily referenced in:
+// * [OpenSSL Implementation](https://github.com/openssl/openssl/blob/45bb98bfa223efd3258f445ad443f878011450f0/ssl/statem/statem_lib.c#L1434)
+// * [BoringSSL Implementation](https://github.com/google/boringssl/blob/583c60bd4bf76d61b2634a58bcda99a92de106cb/ssl/ssl_x509.cc#L1323)
+impl From<CertificateError> for AlertDescription {
+    fn from(e: CertificateError) -> Self {
+        use CertificateError::*;
+        match e {
+            BadEncoding | UnhandledCriticalExtension | NotValidForName => Self::BadCertificate,
+            Expired | NotValidYet => Self::CertificateExpired,
+            UnknownIssuer => Self::UnknownCA,
+            BadSignature => Self::DecryptError,
+            Other(_) => Self::CertificateUnknown,
+        }
+    }
+}
+
 impl From<CertificateError> for Error {
     #[inline]
     fn from(e: CertificateError) -> Self {
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -3,7 +3,7 @@ use crate::bs_debug;
 use crate::check::inappropriate_handshake_message;
 use crate::conn::{CommonState, ConnectionRandoms, State};
 use crate::enums::{CipherSuite, ProtocolVersion};
-use crate::error::{CertificateError, Error, PeerIncompatible, PeerMisbehaved};
+use crate::error::{Error, PeerIncompatible, PeerMisbehaved};
 use crate::hash_hs::HandshakeHashBuffer;
 use crate::kx;
 #[cfg(feature = "logging")]
@@ -810,20 +810,4 @@ impl State<ClientConnectionData> for ExpectServerHelloOrHelloRetryRequest {
             )),
         }
     }
-}
-
-pub(super) fn send_cert_error_alert(common: &mut CommonState, err: Error) -> Error {
-    match err {
-        Error::InvalidCertificate(CertificateError::BadEncoding) => {
-            common.send_fatal_alert(AlertDescription::DecodeError);
-        }
-        Error::PeerMisbehaved(_) => {
-            common.send_fatal_alert(AlertDescription::IllegalParameter);
-        }
-        _ => {
-            common.send_fatal_alert(AlertDescription::BadCertificate);
-        }
-    };
-
-    err
 }
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1,5 +1,5 @@
 use crate::check::{inappropriate_handshake_message, inappropriate_message};
-use crate::conn::{CommonState, ConnectionRandoms, Side, State};
+use crate::conn::{self, CommonState, ConnectionRandoms, Side, State};
 use crate::enums::ProtocolVersion;
 use crate::error::{Error, InvalidMessage, PeerMisbehaved};
 use crate::hash_hs::HandshakeHash;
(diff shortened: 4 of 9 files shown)


djc:
I think this is broadly a good direction, but I think it would be good to split this into different commits.

* One commit to move `send_cert_alert_error()` into its new location, without other changes
* One commit to change `send_cert_alert_error()` to send `HandshakeFailure` instead of `BadCertificate` in the fallthrough case, with justification why this is a good change to make
* One commit to use the `CertificateError` -> `AlertDescription` mapping in `send_cert_alert_error()`
* One commit to change the server to start using `send_cert_alert_error()`

Specifically for the case of server-side TLS 1.2, there should be extra explanation of how it makes sense to treat the `Error::General` case where `take_handshake_buf()` yields `None`, it seems like we're sending two alerts in that case?

zkonge:
> Thanks for working on this! I'd be curious to hear more about the exact use case that you're trying to improve the outcomes for here.

We use rustls as the TLS implementation for our Zero Trust Infrastructure, we received lots of cases about TLS handshake failure from out users, but the actual reason is the certificate expired. After this modification, user can self check their problem, it may reduce our unnecessary communication and help us quickly locate the problem instead of using Wireshark all the time.

> * One commit to move `send_cert_alert_error()` into its new location, without other changes
> * One commit to change `send_cert_alert_error()` to send `HandshakeFailure` instead of `BadCertificate` in the fallthrough case, with justification why this is a good change to make
> * One commit to use the `CertificateError` -> `AlertDescription` mapping in `send_cert_alert_error()`
> * One commit to change the server to start using `send_cert_alert_error()`

Thanks, Working.
> Specifically for the case of server-side TLS 1.2, there should be extra explanation of how it makes sense to treat the `Error::General` case where `take_handshake_buf()` yields `None`, it seems like we're sending two alerts in that case?

Actually, the old implementation did that, I just follow it. :(
I see the [comment](https://github.com/rustls/rustls/blob/1efeeaf5b85e318289a2576daf2213713dc705c7/rustls/src/server/tls12.rs#LL664C28-L664C28). If you think it's not on purpose, I will try to solve it.
> Also, probably a good idea to make the tests pass first!

Working too, but i'm surprised it passed in my env, I think I forgot enable some test feature.

djc:
> We use rustls as the TLS implementation for our Zero Trust Infrastructure, we received lots of cases about TLS handshake failure from out users, but the actual reason is the certificate expired. After this modification, user can self check their problem, it may reduce our unnecessary communication and help us quickly locate the problem instead of using Wireshark all the time.

Making our error handling more specific has definitely been a theme recently (see #1178, #1182), so I think this fits in well with that.

> Actually, the old implementation did that, I just follow it. :(
I see the [comment](https://github.com/rustls/rustls/blob/1efeeaf5b85e318289a2576daf2213713dc705c7/rustls/src/server/tls12.rs#LL664C28-L664C28). If you think it's not on purpose, I will try to solve it.

Yeah, I noticed that this is not new behavior. Still, I think it might make sense to just add a `return` keyword for the `Error::General` case (that should be unreachable) so we avoid sending two different alerts on the same connection.

> Working too, but i'm surprised it passed in my env, I think I forgot enable some test feature.

Yeah, you want to run with `--all-features`, and the bogo test suite is separate (see `bogo/README.md`, you might also be interested in `bogo/config.json` for this).

zkonge:
Any other advice about the mapping design?

rustls/src/error.rs
```diff
@@ -262,6 +262,19 @@ pub enum CertificateError {
...
 
+impl From<CertificateError> for AlertDescription {
+    fn from(e: CertificateError) -> Self {
+        use CertificateError::*;
+        match e {
+            BadEncoding | UnhandledCriticalExtension | NotValidForName => Self::BadCertificate,
+            Expired | NotValidYet => Self::CertificateExpired,
+            UnknownIssuer => Self::UnknownCA,
+            BadSignature => Self::DecryptError,
+            Other(_) => Self::CertificateUnknown,
```
djc, line 273: `CertificateUnknown` seems like a surprising choice for `Other`. Maybe this should also become `BadCertificate`? We're already in the context of `CertificateError`s so that seems like a better choice to me.

I guess this is in line with OpenSSL and BoringSSL.

rustls/src/error.rs
```diff
@@ -262,6 +262,19 @@ pub enum CertificateError {
     Other(Arc<dyn StdError + Send + Sync>),
 }
 
+impl From<CertificateError> for AlertDescription {
+    fn from(e: CertificateError) -> Self {
+        use CertificateError::*;
+        match e {
+            BadEncoding | UnhandledCriticalExtension | NotValidForName => Self::BadCertificate,
+            Expired | NotValidYet => Self::CertificateExpired,
```
djc, line 270: Do you think it's potentially confusing to turn `NotValidYet` into `CertificateExpired`? OpenSSL turns `NotValidYet` into `BadCertificate`, BoringSSL turns it into `CertificateExpired` like you have here.

rustls/src/error.rs
```diff
@@ -262,6 +262,19 @@ pub enum CertificateError {
     Other(Arc<dyn StdError + Send + Sync>),
 }
 
+impl From<CertificateError> for AlertDescription {
+    fn from(e: CertificateError) -> Self {
+        use CertificateError::*;
+        match e {
+            BadEncoding | UnhandledCriticalExtension | NotValidForName => Self::BadCertificate,
```
djc, line 269: Should we turn `UnhandledCriticalExtension` into `UnsupportedExtension`?

rustls/src/error.rs
```diff
@@ -262,6 +262,19 @@ pub enum CertificateError {
     Other(Arc<dyn StdError + Send + Sync>),
 }
 
+impl From<CertificateError> for AlertDescription {
```
djc, line 265: Would be good to add the OpenSSL and BoringSSL links from the commit message here as comments, probably!

bogo/config.json
```diff
@@ -380,8 +380,10 @@
   },
   "TestLocalErrorMap": {
     "SendServerHelloAsHelloRetryRequest": "remote error: error decoding message",
-    "GarbageCertificate-Server-TLS12": "remote error: access denied",
-    "GarbageCertificate-Server-TLS13": "remote error: access denied",
+    "GarbageCertificate-Server-TLS12": "remote error: bad certificate",
```
djc, line 383: Please splice the changes from this commit into the earlier changes that require these changes to make the tests pass, so that each commit can pass the tests.

djc (approved):
This is looking great! @ctz any opinions?

rustls/src/error.rs
zkonge: Following the [RFC](https://www.rfc-editor.org/rfc/rfc5246#section-7.2.2)

>    certificate_expired
      A certificate has expired or is not currently valid.

djc: Ah, right!

rustls/src/error.rs
zkonge: Following the [RFC](https://www.rfc-editor.org/rfc/rfc5246#section-7.2.2)

>    certificate_unknown
      Some other (unspecified) issue arose in processing the
      certificate, rendering it unacceptable.

But it's just my own understand.
Should we really following other implementation ?

djc: Okay, good point. Maybe add these RFC references to the code as well?

rustls/src/error.rs
zkonge: Following the TLS1.2 [RFC](https://www.rfc-editor.org/rfc/rfc5246#section-7.2.2)

>    unsupported_extension
      sent by clients that receive an extended server hello containing
      an extension that they did not put in the corresponding client
      hello.  This message is always fatal.

And TLS1.3 [RFC](https://www.rfc-editor.org/rfc/rfc8446#section-6.2)

>    unsupported_extension:  Sent by endpoints receiving any handshake
      message containing an extension known to be prohibited for
      inclusion in the given handshake message, or including any
      extensions in a ServerHello or Certificate not first offered in
      the corresponding ClientHello or CertificateRequest.

I think it means only the extension in TLS, not certificate extension.

djc: Great!

zkonge:
I think the current CertificateError contains too little choice to express all errors.

I suggest picking up all certificate related items from standard TLS alerts, and putting them into CertificateError.

The TLS alert picked:
+ bad_certificate
+ unsupported_certificate
+ certificate_revoked
+ certificate_expired
+ certificate_unknown
+ unknown_ca
+ access_denied
+ decrypt_error
+ internal_error

Then, we don't need to care about how to map CertificateError to TLS Alert, and users have more choices to decide errors by themselves.

We just need to modify the `pki_error`. And write documents carefully to tell users to read the RFC and select the suitable error.

How about @ctz ,I found you did a great job refining the error :).

ctz (approved):
this looks like a good set of improvements to me!

djc:
@zkonge would you mind rebasing this so we can potentially merge it?

I don't think we want `CertificateError` to exactly match the potential alerts you listed. In particular, where possible I think we'd prefer to yield more precise errors, which help our users determine what went wrong (which in turn will likely decrease our support load).

If you want to send another PR (ideally soon, since we're about to release 0.21 and this would be an API-breaking change) to add `CertificateError` variants where you feel the existing variants don't adequately cover your needs, I'd be happy to review that.

djc:
@zkonge this is the final change blocking the 0.21.0 release at this point. Please let me know if/when you're able to rebase/address comments, otherwise I'll just take care of that for you.

zkonge:
i will deal with it this night

djc:
Great, thank you!

rustls/src/error.rs
cpu: @zkonge It looks like 77199400810439fb1d99654de73654facbb2f457 added the OpenSSL/BoringSSL references in-code, but left the RFC references just to the commit message. Was that intentional?

zkonge: It's intentional. Because RFC doesn't tell us how to mapping X.509 certificate error into TLS alerts.
But thanks, you remind me that I forgot to add comments for puzzling mapping choices.

djc (approved):

djc:
This is great, thanks!

PR #1201, opened 2023-03-01
small script fixes, gitignore updates.

cpu:
## Description

This branch collects up some small improvements I've made locally while trying to familiarize myself with the project and associated tooling.

### chore: update .gitignore.
Running the `admin/coverage` script produces a `default.profraw` in the project root. Prior to this commit that file was not ignored by git. This commit adds it to the ignore list since we don't want this file to be accidentally submitted upstream.

This commit also adds the JetBrains IDE state dir (`.idea`) to the gitignore. This is a small quality of life improvement for users of CLion.

### fix: avoid hardcoded (ba)sh path in helper scripts.
Prior to this commit some helper scripts used hardcoded paths to `/bin/sh` and `/bin/bash` in script shebangs. This will error on systems that don't place `bash` in `/bin/` (e.g. NixOS).

This commit updates the scripts to use `/usr/bin/env` to find `bash` based on the user's `$PATH`. This has better portability and allows the scripts to run without err (or specifying an interpreter explicitly) on systems with atypical `bash` installs.

### fix: avoid hardcoded python3 path, fix prints.
Prior to this commit the admin `bench-range`, `capture-certdata` and `format-bench` Python helpers used a hardcoded path to Python3 in their shebang. This commit updates each to use `/usr/bin/env` to find `python3` in the `$PATH`.

These helpers also used Python2 style `print` statements instead of the Python3 style `print` function. This commit adds the required braces to fix Python3 usage.

**Note to reviewers:** I think these Python scripts may have bitrot. The `bench-range` and `format-bench` programs have additional issues (see https://github.com/rustls/rustls/pull/1203), and the `trytls/runme` script assumes Python 2.7 in a directory path (see https://github.com/rustls/rustls/pull/1202). ~I'll will follow up on these separately from the hardcoded path issues fixed here.~.

### chore: set initial branch name for bogo checkout.
Prior to this commit the `fetch-and-build` script with newer `git` versions would spit out an ANSI coloured warning about choosing an initial branch name for the bogo test-suite checkout.

This commit simply specifies the `--initial-branch` to be `main` to silence the unnecessary output.

### DIFF
--- a/admin/format-bench
+++ b/admin/format-bench
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 
 import sys
 
@@ -24,9 +24,9 @@ for line in sys.stdin:
         handshakes.append((suite, role, auth, resumed, float(rate), unit))
 
 for suite, direction, rate, unit in sorted(bulks):
-    print '`%s` | %s | %g %s' % (suite, direction, rate, unit)
+    print('`%s` | %s | %g %s' % (suite, direction, rate, unit))
 
 for suite, role, auth, resumed, rate, unit in sorted(handshakes):
-    print '`%s` <br> %s, %s, %s | %g %s' % (suite, role,
+    print('`%s` <br> %s, %s, %s | %g %s' % (suite, role,
             auth if auth == 'server-auth' else 'mutual-auth',
-            translate_resume(resumed), rate, unit)
+            translate_resume(resumed), rate, unit))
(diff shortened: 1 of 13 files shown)


admin/format-bench
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!//usr/bin/env python3
```
djc, line 1: Extra slash?

djc (approved):

admin/format-bench
cpu: Oops, good catch. Thanks! Force pushed w/ a fix.

djc (approved):

Issue #1205, opened 2023-03-02
bench.rs improvements


cpu:
Let's chat about the future of `bench.rs` and with the goal of splitting out smaller tasks.

### Background

In https://github.com/rustls/rustls/pull/1203 we discussed the [admin/bench-range](https://github.com/rustls/rustls/blob/1ee4a2ba92befd02386341c9036d5e872a47a3c4/admin/bench-range) tool (written in bash) and concluded its functionality could be folded into the [rustls/examples/internal/bench.rs](https://github.com/rustls/rustls/blob/main/rustls/examples/internal/bench.rs) tool.

### Discussion

Adding support for this feature to `bench.rs` as written is certainly possible, but the complexity of the [`selected_tests`](https://github.com/rustls/rustls/blob/1ee4a2ba92befd02386341c9036d5e872a47a3c4/rustls/examples/internal/bench.rs#L571-L640) function is getting larger. Concurrently, the Prossimo initiative updated [the Rustls work plan](https://www.memorysafety.org/initiative/rustls/rustls-work-plan/) to include "Comprehensive performance benchmarking", a goal that is likely to include more development of `bench.rs`.

It may be worth evaluating separating the benchmark example into its own crate, and having that new create depend on an external 3rd party crate (maybe [clap](https://github.com/clap-rs/clap)?) that can offer richer command line argument semantics. This would let us clean up the `selected_tests` function, making it easier to express submodes and optional parameters specific to certain modes and focusing our efforts on the benchmarking specific code. 

Is that something the maintainers would appreciate or should we keep things simple and lift `bench-range`'s functionality into the existing code without larger rework? I'm newer to the broader Rust ecosystem and might be underestimating the challenge of introduce a benchmark specific crate, or the cost of a dependency like clap.

djc:
Using clap is great, but it pushes the MSRV (minimum supported Rust version) fairly aggressively so keeping it in the same crate/repo makes doing CI that checks the (more conservative) MSRV for the library a little harder.

I think we should discuss what the goals are for "Comprehensive performance benchmarking"? I think there are two:

* Assessing how rustls performance is changing over time, to make sure rustls doesn't get substantially slower over time (and ideally becomes faster)
* Assessing how rustls performance compares to other implementations, probably specifically OpenSSL (see [openssl-bench](https://github.com/ctz/openssl-bench))

Presumably for some (potential) users, lower latencies (especially on handshakes) and lower memory usage are potentially interesting selling points.

[...]

PR #1206, opened 2023-03-02
chore: delete admin/format-bench.

cpu:
The `admin/format-bench` script has bitrot so that it no longer functions with the current output produced by the benchmark tooling.

It's not clear that this tool is still needed, so rather than fix it (see https://github.com/rustls/rustls/pull/1203) or port it to Rust (ala https://github.com/rustls/rustls/issues/1205), we choose to remove it outright.

If it turns out there is a need for formatting benchmarking output we should consider building that into the benchmark tool itself to avoid needing to deal with a Python tool-chain and associated versioning challenges.

**Note to reviewers: We should block merging this on @ctz having a chance to confirm this tool isn't important to any external processes still in use.**


djc (approved):

djc:
@ctz thoughts?

ctz:
The goal of this really was to take the output of the `bench` tool and massage it into github markup for the purposes of https://github.com/rustls/rustls/wiki/Performance-measurements -- agree on the points made here though and probably deleting it is the best idea.

PR #1207, opened 2023-03-02
docs: add rustdoc for DigitallySignedStruct.

cpu:
The front-page index for the Rustls lib includes a list of re-exported structs comprising the public API. I noticed that of all of the structs included only `DigitallySignedStruct` was missing commentary about its purpose.

This tiny commit adds a simple rustdoc comment for this type. Now all of the structs mentioned in the front-page index have some documentation.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1592,6 +1592,7 @@ impl Codec for ECParameters {
     }
 }
 
+/// This type combines a [`SignatureScheme`] and a signature payload produced with that scheme.
 #[derive(Debug, Clone)]
 pub struct DigitallySignedStruct {
     pub scheme: SignatureScheme,


rustls/src/msgs/handshake.rs
```diff
@@ -1592,6 +1592,9 @@ impl Codec for ECParameters {
     }
 }
 
+/// This type combines a [`SignatureScheme`] and a signature payload produced with that scheme.
+///
+/// [`SignatureScheme`]: SignatureScheme
```
djc, line 1597: I think you don't need the explicit link target in this case? IIRC it's only necessary when the target is not available from the enclosing scope.

djc (approved):

rustls/src/msgs/handshake.rs
cpu: Oh neat. That's handy. I tested it out locally and it worked the way you described. Force pushed a fix.

PR #1208, opened 2023-03-02
docs: fix broken x509-parser FromDer trait links.

cpu:
The upstream `x509-parser` crate [moved the `FromDer` trait](https://github.com/rusticata/x509-parser/commit/ac4620696d8c4031f0d4bc01334ded8f76a5bf10) from a `traits` submodule to a `prelude` submodule, breaking the docs link/example in the Rustls `DistinguishedName` and `OwnedTrustAnchor.subject` doc strings.

This commit fixes the links/examples to use the current location of the `FromDer` trait.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -51,10 +51,10 @@ impl OwnedTrustAnchor {
 
     /// Return the subject field.
     ///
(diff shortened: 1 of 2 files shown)


djc (approved):
Nice!

Issue #995, opened 2022-01-31
Allow extracting information for `tls-unique` Channel Binding


dequbed:
Hello!

Authentication frameworks such as GSSAPI and SASL implement a concept called ["channel binding"](https://www.rfc-editor.org/rfc/rfc5056) in which they cryptographically verify in the authentication exchange that both parties are seeing the same cryptographic channel, thus binding the authentication to the underlying transport security layer and preventing a few attacks.

To be able to do this the authentication frameworks include data from the underlying encryption layer in the secrets used for the authentication itself.

In the case of TLS there are two common kinds of channel bindings, [standardized in RFC5929](https://www.rfc-editor.org/rfc/rfc5929.html): `tls-server-end-point` which uses the first Certificate provided by the server, and `tls-unique` which uses the first "Finished" Message sent in the connection.
Getting to the first certificate sent by the server is already possible via `ClientConnection::peer_certificates` as the client or … well … *keeping it around* as the server. However, I don't see any way to get to the (contents of the) Finished message as either server or client.

Am I missing something or is there no way currently? And if the latter, is this something that the rustls team would be willing to add or to merge a PR adding it?

ctz:
`tls-unique` is not defined for TLS1.3, and [its replacement](https://www.ietf.org/id/draft-ietf-kitten-tls-channel-bindings-for-tls13-12.html) can be implemented using existing APIs in this crate.

As for implementing `tls-unique` for _just_ TLS1.2 here, I'm not sure. Really I think it's a shame and a mistake that RFC5929 wasn't implemented in terms of RFC5705. If we were to implement it I think we'd have to restrict it to connections that successfully negotiate the RFC7627 extended master secret extension (due to the triple handshake vuln). That might make it unusable for your use case. Could you describe your use case?

dequbed:
Ah thanks, I didn't know TLS1.3 replaced `tls-unique`! 

> That might make it unusable for your use case. Could you describe your use case?

"I'm currently writing a SASL framework and would like to be able to test its channel binding support" ^^

Slightly more seriously though, the only real usecase I have for channel bindings is one where we control both sides of the connection so we can just have a hard requirement on TLS 1.3 if CB are to be used.

The biggest reason why I was looking at `tls-unique` is that a) I'm coming to the topic from the SASL/GSSAPI side and don't know enough about TLS to realize that TLS 1.3 is incompatible there, b) the library I am using to check compatability against only implements tls-unique.

dequbed:
[...]

PR #1211, opened 2023-03-07
feat: impl safer export_keyring_material interface.

cpu:
## Description

Prior to this branch the `export_keyring_material` function used a mutable out buffer for writing exported key material, and returned an empty Ok result when there was no error doing so.

This commit updates the function such that the ownership of the output buffer passes through the export function and is returned as the Ok result when there is no error.

Doing this makes for a safer interface for end users: the output buffer will be dropped if `export_keyring_material` errors. Callers can only access the buffer again using the Ok result.

## Kudos

All credit due to @davidv1992 for the implementation idea and initial code from https://github.com/rustls/rustls/pull/1159. I simply adopted the branch, fixed a compile error,  updated the affected unit tests and added a small rustdoc update.

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -2029,25 +2029,21 @@ fn do_exporter_test(client_config: ClientConfig, server_config: ServerConfig) {
     );
     do_handshake(&mut client, &mut server);
 
-    assert_debug_eq(
-        client.export_keying_material(&mut client_secret, b"label", Some(b"context")),
-        Ok(()),
-    );
-    assert_debug_eq(
-        server.export_keying_material(&mut server_secret, b"label", Some(b"context")),
-        Ok(()),
-    );
+    assert!(client
+        .export_keying_material(&mut client_secret, b"label", Some(b"context"))
+        .is_ok());
+    assert!(server
+        .export_keying_material(&mut server_secret, b"label", Some(b"context"))
+        .is_ok());
     assert_eq!(client_secret.to_vec(), server_secret.to_vec());
 
-    assert_debug_eq(
-        client.export_keying_material(&mut client_secret, b"label", None),
-        Ok(()),
-    );
+    assert!(client
+        .export_keying_material(&mut client_secret, b"label", None)
+        .is_ok());
     assert_ne!(client_secret.to_vec(), server_secret.to_vec());
-    assert_debug_eq(
-        server.export_keying_material(&mut server_secret, b"label", None),
-        Ok(()),
-    );
+    assert!(server
+        .export_keying_material(&mut server_secret, b"label", None)
+        .is_ok(),);
     assert_eq!(client_secret.to_vec(), server_secret.to_vec());
 }
 
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -79,12 +79,12 @@ impl Connection {
     /// Derives key material from the agreed connection secrets.
     ///
     /// See [`ConnectionCommon::export_keying_material()`] for more information.
-    pub fn export_keying_material(
+    pub fn export_keying_material<T: AsMut<[u8]>>(
         &self,
-        output: &mut [u8],
+        output: T,
         label: &[u8],
         context: Option<&[u8]>,
-    ) -> Result<(), Error> {
+    ) -> Result<T, Error> {
         match self {
             Self::Client(conn) => conn.export_keying_material(output, label, context),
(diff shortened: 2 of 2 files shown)


cpu:
> rustls / Build+test (1.57, ubuntu-20.04) (pull_request) Failing after 4m

Looks like [this failure](https://github.com/rustls/rustls/actions/runs/4358734354/jobs/7619684388) was another bit of bad internet weather:
> ---- online::microsoft stdout ----
> We expected to find 'HTTP/1.[01] ' in the following output:
> error: test failed, to rerun pass '-p rustls-examples --test topsites'
> Output { status: ExitStatus(unix_wait_status(256)), stdout: "TLS read error: Os { code: 104, kind: ConnectionReset, message: \"Connection reset by peer\" }\nConnection closed\n", stderr: "" }

rustls/tests/api.rs
```diff
@@ -2029,26 +2029,31 @@ fn do_exporter_test(client_config: ClientConfig, server_config: ServerConfig) {
...
     do_handshake(&mut client, &mut server);
 
-    assert_debug_eq(
-        client.export_keying_material(&mut client_secret, b"label", Some(b"context")),
-        Ok(()),
-    );
-    assert_debug_eq(
-        server.export_keying_material(&mut server_secret, b"label", Some(b"context")),
-        Ok(()),
+    let exported_client_secret_with_ctx = client
```
djc, line 2032: It doesn't look like these changes are actually necessary? Without them, do you get an error? What does that look like?

djc (approved):

rustls/tests/api.rs
cpu: If I back out all the changes to `rustls/test/api.rs` then the assertions fail to compile w/ a type mismatch because the return type isn't `Ok(())` in the success case any longer, e.g.:
```
error[E0308]: mismatched types
    --> rustls/tests/api.rs:2034:12
     |
2034 |         Ok(()),
     |         -- ^^ expected `&mut [u8; 64]`, found `()`
     |         |
     |         arguments to this enum variant are incorrect
     |
note: tuple variant defined here`
```

I think the test asserts will have to change somehow, but my first pass might be too invasive.

<details>
<summary>Here's one potential alternative. WDYT of this diff? :</summary>
```diff
diff --git a/rustls/tests/api.rs b/rustls/tests/api.rs
index 341fe068..6d6edf85 100644
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -2029,25 +2029,21 @@ fn do_exporter_test(client_config: ClientConfig, server_config: ServerConfig) {
     );
     do_handshake(&mut client, &mut server);

-    assert_debug_eq(
-        client.export_keying_material(&mut client_secret, b"label", Some(b"context")),
-        Ok(()),
-    );
-    assert_debug_eq(
-        server.export_keying_material(&mut server_secret, b"label", Some(b"context")),
-        Ok(()),
-    );
+    assert!(client
+        .export_keying_material(&mut client_secret, b"label", Some(b"context"))
+        .is_ok());
+    assert!(server
+        .export_keying_material(&mut server_secret, b"label", Some(b"context"))
+        .is_ok());
     assert_eq!(client_secret.to_vec(), server_secret.to_vec());

-    assert_debug_eq(
-        client.export_keying_material(&mut client_secret, b"label", None),
-        Ok(()),
-    );
+    assert!(client
+        .export_keying_material(&mut client_secret, b"label", None)
+        .is_ok());
     assert_ne!(client_secret.to_vec(), server_secret.to_vec());
-    assert_debug_eq(
-        server.export_keying_material(&mut server_secret, b"label", None),
-        Ok(()),
-    );
+    assert!(server
+        .export_keying_material(&mut server_secret, b"label", None)
+        .is_ok(),);
     assert_eq!(client_secret.to_vec(), server_secret.to_vec());
 }
```
</details>

djc: Yup, that looks good to me.

cpu: Great :-) Applied in my force push of dc6bef766a0f017770f49aa4f37a9d2aebc14ebf.

djc (approved):

Neustradamus:
Linked to:
- Channel Binding: https://github.com/scram-sasl/info/issues/1
- https://github.com/rustls/rustls/issues/995
- https://github.com/rustls/rustls/issues/1624
- https://github.com/rustls/rustls/pull/1159
- https://github.com/rustls/rustls/pull/1211

PR #1213, opened 2023-03-09
fix: fail to compile with logging feature disabled

bebecue:


djc (approved):
Thanks! Would you be able to look into making sure this gets caught in CI? I'm guessing it is because this is in the tls12 module, which is disabled in our `--no-default-features` run. If you could add some code to the CI config to fix this that would be great!

bebecue:
I enabled `tls12` feature in the previous `(debug; no default features)` case, is that the right place?

.github/workflows/build.yml
```diff
@@ -72,12 +72,12 @@ jobs:
         env:
           RUST_BACKTRACE: 1
 
-      - name: cargo build (debug; no default features)
-        run: cargo build --no-default-features
+      - name: cargo build (debug; without logging feature)
```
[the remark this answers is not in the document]
bebecue: Thanks for the detailed guidance. Changed in https://github.com/rustls/rustls/pull/1213/commits/cb025508a23761bd6527177ddce413455c623152

cpu:
> @cpu this saw CI fail with https://github.com/rustls/rustls/actions/runs/4373903788/jobs/7653805791, does that look similar to the other issues you were investigating?

@djc The `apple.com` online test failures? I think so, but I was testing with `s_client` in a tight loop so the err doesn't map 1:1. As an experiment I just tried running the same loop from GHA against `aws.amazon.com` and wasn't able to reproduce after 4-5 runs of 100 serial connects w/o sleeps.

.github/workflows/build.yml
```diff
@@ -73,11 +73,11 @@ jobs:
           RUST_BACKTRACE: 1
 
       - name: cargo build (debug; no default features)
-        run: cargo build --no-default-features
+        run: cargo test --no-default-features
         working-directory: rustls
 
-      - name: cargo test (debug; no default features)
-        run: cargo test --no-default-features
+      - name: cargo test (debug; without logging feature)
```
cpu, line 79: small nit: Would this name be clearer as "(debug; no default features; tls1.2)"? It seems slightly more future proof :-)

```suggestion
      - name: cargo test (debug; no default features; tls1.2)
```

cpu (approved):

.github/workflows/build.yml
bebecue: Thanks, I think that's also clearer. I squashed all the ci commits in https://github.com/rustls/rustls/pull/1213/commits/8d79b1aa7365dce42e3c74eba3e0aead94925bce

djc (approved):

.github/workflows/build.yml
cpu: Looks great, thanks!

PR #1214, opened 2023-03-10
CI: update workflow actions, Go tooling.

cpu:
Inspired by https://github.com/rustls/webpki/pull/33 this branch updates the Rustls CI.

To address [a pending deprecation](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/), and to remove CI warnings (_e.g. see the 80 warning level Annotations on [this recent build](https://github.com/rustls/rustls/actions/runs/4387177035)_), this branch:

* Updates `actions/checkout` from v2 -> v3.
* Updates `codecov/codecov-action` from v1.0.10 -> v3.
* Updates `actions/upload-artifact` from v1 -> v3.
* Updates `actions/setup-go` from v2 -> v3.

For the `setup-go` workflows in particular the Go version is also updated 1.17.1 -> 1.20.

Additionally, since the `actions-rs` upstream workflows [are unmaintained](https://github.com/actions-rs/toolchain/issues/216) this commit replaces `actions-rs/toolchain` with `dtolnay/rust-toolchain`, a well maintained alternative. Similarly, usages of `actions-rs/cargo` are replaced with direct invocation of `cargo`.

If this work looks good to folks I will follow-up with additional PRs on the other org repos to follow suit.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: weekly
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,15 +30,14 @@ jobs:
             rust: stable
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           persist-credentials: false
 
       - name: Install ${{ matrix.rust }} toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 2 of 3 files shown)


djc (approved):
LGTM. I think there is a way to have Dependabot ping us for these, too, which might be nice?

(FWIW, this would have been a little easier to review if you made one commit per bullet point, IMO.)

cpu:
Thanks for the :mag:

> I think there is a way to have Dependabot ping us for these, too, which might be nice?

Added in https://github.com/rustls/rustls/pull/1214/commits/18eddf866d0231a2260e306fa9f046ffbd1277cf

> FWIW, this would have been a little easier to review if you made one commit per bullet point, IMO.

ACK. I'll make sure to do that going forward.

PR #1216, opened 2023-03-12
QUIC: introduce separate types for QUIC sessions

djc:
Change the QUIC API to avoid relying on extension traits, instead implementing a `Connection` type hierarchy mirroring the `ClientConnection -> ConnectionCommon -> CommonState` one on the TCP side, with different specific types for `ClientConnection` and `ConnectionCommon` that shed their TCP-specific API in favor of QUIC-only API.

Also splits the `conn` module by moving `CommonState` and related types into a `common_state` module, and renames `sni_hostname()` accessors to `server_name()`.

It's all @Ralith's fault for mentioning on Discord that the extension trait approach ended up not being so great after all!


Ralith:
Haven't reviewed in great depth, but I think this is a good direction. Seems a bit tidier, and some nice refactoring regardless of QUIC.

djc:
@ctz and @cpu what do you think of blocking the release on this? I feel like it's a much cleaner separation of the TCP and QUIC APIs, and it does make changes to the public QUIC API so it'd need a semver-incompatible release.

@cpu would you be able to help out getting this compiling? Need to move some methods from `ClientConnection` to `Core<ClientConnectionData>` so they can be referenced from both `ClientConnection` and `quic::ClientConnection` and figure out a solution for how bogoshim abstracts over QUIC vs TCP, which is slightly wonky. Could probably do an enum, maybe something like `enum ConnectionLike<S: SideData, T: Deref<Target = ConnectionCommon<S>> + DerefMut, Q: Deref<Target = QuicConnection<S>> + DerefMut> { Tcp(T), Quic(Q) }`? (We'll probably need to fiddle a bit with adapting `read_tls()` to `read_hs()`.)

cpu:
> @cpu would you be able to help out getting this compiling? 

Sure, I can take a look today 👍

djc:
I would do it myself but a bit short on time today. I can probably make some time tomorrow if you can't figure it out.

cpu:
No problem :-) I'll see where I can get by EOD and we can join forces tomorrow on whatever's left.

From my perspective waiting a couple days to tag the release seems reasonable if there's review capacity to get this across the line.

cpu:
> Need to move some methods from ClientConnection to Core<ClientConnectionData> so they can be referenced from both ClientConnection and quic::ClientConnection

I was able to quickly figure out this part and pushed a couple of commits that address those compile errors.

> figure out a solution for how bogoshim abstracts over QUIC vs TCP, which is slightly wonky

I battled this all afternoon and haven't cracked it yet. I'm finding it challenging to build a workable solution that can be generic across two transport types (TCP vs Quic) and two sides (client vs server). Based on discussions OOB I think we need a combination of a `trait` and an `enum`. I've started working that up in another branch but it doesn't compile and I'm out of time for today :'(

sehz:
Ran into a similar challenge in trying to abstract different transport mechanisms.  Suggest using `dyn` trait which would be the easiest implement with flexibility and a minor cost in performance.  However, in the case of I/O,  performance degradation will be noise compared with I/O cost in general.

cpu:
I replaced the `rustls` dependency in [Quinn](https://github.com/quinn-rs/quinn) with a patch to use this branch as a smoke test to make sure we're not missing any required functionality: https://github.com/quinn-rs/quinn/pull/1515

That uncovered two additional changes that were required:
* one commit to add `sni_hostname` on the `quic::ServerConnection` type -   https://github.com/cpu/rustls/commit/67ea7a2053ed14f7ddcd7627b9ea3317aabf6a82
* one commit to add `export_keying_material` to the `quic::Connection` type - https://github.com/cpu/rustls/commit/50c786c3603fb9dd4b9420f0dc94891daf8d5e5e

Other than fixing up some API drift that was all that was required to get Quinn passing its test suite with this branch :+1:

djc:
Would you mind splicing those changes into the existing commits here? Previously I squashed the non-QUIC changes into the "Introduce ConnectionCore" commit and the QUIC changes into the "quic: introduce separate QUIC .." commit (please check after squashing that the commit still builds and passes tests and docs).

I also wonder if we should take the change to rename `sni_hostname()` to `server_name()` at least in the public API (probably still mention SNI in the documentation if people are looking for that).

cpu:
> Would you mind splicing those changes into the existing commits here? Previously I squashed the non-QUIC changes into the "Introduce ConnectionCore" commit and the QUIC changes into the "quic: introduce separate QUIC .." commit (please check after squashing that the commit still builds and passes tests and docs).

:ballot_box_with_check: - Done.

> I also wonder if we should take the change to rename sni_hostname() to server_name() at least in the public API (probably still mention SNI in the documentation if people are looking for that).

This sounded good to me. I put that in a separate commit on the end: https://github.com/rustls/rustls/pull/1216/commits/465a715dffcfcc1824975bcdf92185345e928c59

rustls/examples/internal/bogo_shim.rs
```diff
@@ -805,17 +804,6 @@ fn exec(opts: &Options, mut sess: Connection, count: usize) {
...
-        if !sess.is_handshaking()
-            && !opts
-                .expect_quic_transport_params
-                .is_empty()
-        {
-            let their_transport_params = sess
-                .quic_transport_parameters()
-                .expect("missing peer quic transport params");
-            assert_eq!(opts.expect_quic_transport_params, their_transport_params);
-        }
```
ctz, line 817: do we understand when/why this became unused? because, at some point in the past, it was added for a reason to give QUIC test coverage against an independent implementation. seems like that is valuable to have?

cpu: @djc Can you share more about how you reached the conclusion the shim code for quic should be removed?

The upstream BoringSSL [shim for bogo](https://github.com/google/boringssl/blob/92de195169d26d9f5cec7ef34df9194e614e50f8/ssl/test/bssl_shim.cc#L522-L533), and [the runner](https://github.com/google/boringssl/blob/92de195169d26d9f5cec7ef34df9194e614e50f8/ssl/test/runner/runner.go#L1456-L1498) both seem to have quic specific test coverage.

In our `config.json` I do see that Quic related bits are in the `DisabledTests`:
https://github.com/rustls/rustls/blob/fb9e181a1a777bc2f42bd7c172be1ebddb3b68fa/bogo/config.json#L110-L111

https://github.com/rustls/rustls/blob/fb9e181a1a777bc2f42bd7c172be1ebddb3b68fa/bogo/config.json#L118-L120

Is it fair to summarize that the upstream does support Quic coverage, but that we've been ignoring it and so now it makes more sense to remove the unused shim support vs adapting it to the new API to remain unused, or investing in getting it running?

djc: Went crazy with git blame for a bit:

- 2021-02-06 eb191d0cc4dbc36b6bc679d66cd12788461545f8 (#475, BoringSSL update): no changes
- 2020-04-12 339923fc5314d154b0522eb2ba07fd20ca98626d (Get new bogo version working): disabled more QUIC tests
- 2019-01-13 181765111cff8fe0c99e7f53670cfd971b3eb14d (Disable QUIC bogo tests): disabled the tests
- 2019-01-13 d766c4d8b6c03a2651aa13a01367232d8e2b14a7 (#187, Raw handshake I/O for QUIC): initially enabled QUIC tests in the bogo shim
- 2018-05-14 87ab639360d72ef4b45da1f3898f6d60d4316d3f (#168, enable QUIC tests in bogo): add some QUIC support in the bogo shim

So, while there initially was some QUIC support added in #187, that doesn't seem to have lasted very long, and it has been disabled ever since. This PR then merely rips out some of the scaffolding that was in place.

ctz: hm we need to look into the commit privileges of these guys! 😨

ctz (approved):
needs a rebase

PR #1221, opened 2023-03-17
fuzz: cargo fmt updates and a clippy fix

cpu:
Three small cleanups I stumbled into while fixing some errors I caused in a WIP refactor elsewhere.

## fuzz: cargo fmt updates - 19c98b47da48af218ad11490f753ea21f8d2b77a

Applies the diff created by running `cargo fmt` across the `fuzz` subcrate.

## fuzz: fix unused Result clippy finding - faf25944e5016d672507ba946577074ab73a494b.

Tiny fix for a clippy error to throw away an otherwise unused`Result` in the persist fuzzer. We don't care if there's an `Error`, just that nothing panics or otherwise goes off the rails.

<details>
<summary>Clippy error:</summary>

```
warning: unused `std::result::Result` that must be used
  --> fuzzers/persist.rs:14:5
   |
14 |     T::read(&mut rdr);
   |     ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: this `Result` may be an `Err` variant, which should be handled
```
</details>

## CI: add fmt and clippy coverage for fuzz workspace. - 37c2275ffd2eb2b33f30d4c3b8179c1d3f870776
The `fuzz` subdirectory is set up as a separate workspace from the main workspace that contains `rustls` and `examples`. Because of this running `cargo fmt --all` and other similar tooling doesn't include the `fuzz` directory. This can lead to 
formatting/clippy drift over time. 

In this commit the `build.yml` config is extended to also run `clippy` and `fmt` on the `fuzz` subdirectory using `--manifest-path`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -217,6 +217,8 @@ jobs:
           components: rustfmt
       - name: Check formatting
         run: cargo fmt --all -- --check
+      - name: Check formatting (fuzz workspace)
+        run: cargo fmt --all --manifest-path=fuzz/Cargo.toml -- --check
 
   clippy:
     name: Clippy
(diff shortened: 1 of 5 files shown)


djc:
Maybe add some stuff to CI to enforce this stuff?

Issue #920, opened 2022-01-12
Types from `internal::*` should not appear in public types


briansmith:
In error.rs we have:
```rust
enum Error {
    ...
    /// We received a fatal alert.  This means the peer is unhappy.
    AlertReceived(AlertDescription),
}
```

`AlertDescription` seems to be only reachable by importing `rustls::internal::msgs::AlertDescription`.

The description for `rustls::internal` says:
```
/// Internal classes which may be useful outside the library.
/// The contents of this section DO NOT form part of the stable interface.
```

However, our intent is that the entirety of `Error` is part of the stable API.

I propose we re-export `pub use msgs::AlertDescription` within errors.rs so that stable users can avoid depending on the unstable `rustls::internal` API.

This comes up when one wants to write tests that verify that a certain `AlertReceived` error is returned by Rustls.

jbr:
`AlertDescription` was made pub from the root in 5dab6ed654a35d8c3af2468732cb3410ab1dc32b, and the crate currently builds without internal being pub. Should this issue become "conditionally expose `rustls::internal` behind a default-disabled feature flag?" in order to prevent this sort of thing in the future?

djc:
Yes, that sounds like a good improvement.

jbr:
are we ok adding a dependency on (visibility)[https://crates.io/crates/visibility] in order to make functions conditionally `pub`/`pub(crate)` to remove unreachable_pub lints?

djc:
One example here is the `Codec` trait. This is faux-public through internals, and as such shows up in the docs for any types that implement it -- many of which are public (`AlertDescription` is one example). Ultimately if someone wants to use the trait they'll have to import it and the only way of doing so is through `internals` so maybe it's not much worse than the other things, but probably still something to consider.

jsha:
We also have this problem with NamedGroup, Tls12ClientSessionValue, and Tls13ClientSessionValue in the new ClientSessionStore trait. In the documentation they link to rustls::internal::msgs::enums::NamedGroup, rustls::internal::msgs::persist::Tls12ClientSessionValue, and rustls::internal::msgs::persist::Tls13ClientSessionValue.

/cc #1224

jsha:
Here's the current list of `internals` types that are reachable in the public interface, borrowing a search command from https://github.com/rustls/rustls/issues/1036#issuecomment-1140389210:

```
base::PayloadU16
codec::Reader
handshake::DigitallySignedStruct
handshake::DistinguishedNames
[...]

PR #1225, opened 2023-03-20
Make SessionId part of the public API

jsha:
This is exposed as part of the ClientSessionStore trait, but was previously public under `internals`.

Also, rename from SessionID to SessionId to follow Rust idioms for naming of acronyms.

Follows up on #1048, #1036, #920.


ctz:
i don't think this type should be public; it's a mistake that it is leaking out. i think where it appears in `Tls12ClientSessionValue` should be `pub(crate)` rather than just pub? likely that code predates `pub(crate)` by a number of years

the rename to SessionId is a good call though

djc:
I think #1226 makes this unnecessary?

PR #1226, opened 2023-03-21
Improve caching API for TLS 1.2

### LINKED ISSUE #1224: feature flag additionality problem for ClientSessionStore
The ClientSessionStore introduced in #1145 [defines remove_tls12_session as optional](https://github.com/rustls/rustls/blob/fb9e181a1a777bc2f42bd7c172be1ebddb3b68fa/rustls/src/client/client_conn.rs#L65-L71) for `not(feature = "tls12")` and required for `feature = "tls12"`. Given the additivity of feature flags, that can result in this bad scenario:

 - a binary depends on rustls with `no-default-features` and provides an implementation of ClientSessionStore that doesn't define `remove_tls12_session`.
 - the binary comes to (transitively) depend on some other crate that depends on rustls with the default features.
 - the binary no longer compiles, because `remove_tls12_session` is required.

[The cargo book says](https://doc.rust-lang.org/cargo/reference/features.html?#feature-unification):

> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html?#semver-compatibility).

There's a similar (actually worse) problem with `set_tls12_session` and `tls12_session`. For non-tls12 builds, they are not part of the trait and so are forbidden. But if the tls12 feature is added into the build, they become required.

I think the most straightforward fix would be to make `set_tls12_session`, `tls12_session,` and `remove_tls12_session` required regardless of feature flags. An implementation that does not want to deal with TLS 1.2 can make it a no-op.

djc:
Fixes #1224.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -55,20 +55,13 @@ pub trait ClientSessionStore: Send + Sync {
     /// Remember a TLS1.2 session.
     ///
     /// At most one of these can be remembered at a time, per `server_name`.
-    #[cfg(feature = "tls12")]
     fn set_tls12_session(&self, server_name: &ServerName, value: persist::Tls12ClientSessionValue);
 
     /// Get the most recently saved TLS1.2 session for `server_name` provided to `set_tls12_session`.
-    #[cfg(feature = "tls12")]
     fn tls12_session(&self, server_name: &ServerName) -> Option<persist::Tls12ClientSessionValue>;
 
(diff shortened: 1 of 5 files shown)


ctz (approved):

cpu (approved):
You beat me to it :-) I had started a branch but hadn't made it to green CI yet.

This looks good to me. Thanks @djc

PR #1227, opened 2023-03-21
Make ClientSessionValue private

djc:
With the new `ClientSessionStore` design, I don't think we need this in the public API anymore.

See #920.


cpu (commented):
This is a nice clean up! It looks reasonable to me.

PR #1228, opened 2023-03-21
Bring back support for encoding/decoding client session values

djc:
With the current implementation, it seems impossible to support cache implementations that don't use an in-memory representation (that preserves the opaque structure of `Tls13ClientSessionValue`/`Tls12ClientSessionValue`).

In order to facilitate this use case, bring back some of the code that was removed in #1145 (in slightly different form).

### DIFF
--- a/rustls/src/msgs/persist.rs
+++ b/rustls/src/msgs/persist.rs
@@ -65,7 +65,7 @@ impl<T> std::ops::Deref for Retrieved<T> {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct Tls13ClientSessionValue {
     suite: &'static Tls13CipherSuite,
     age_add: u32,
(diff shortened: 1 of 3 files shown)


rustls/src/msgs/persist.rs
```diff
@@ -133,6 +133,41 @@ impl Tls13ClientSessionValue {
...
+            suite,
+            age_add: u32::read(r)?,
+            max_early_data_size: u32::read(r)?,
+            common: ClientSessionCommon::read(r)?,
+            #[cfg(feature = "quic")]
+            quic_params: PayloadU16::read(r)?,
+        })
+    }
+
+    /// Inherent implementation of the [`Codec::get_encoding()`] method.
```
cpu, line 159: tiny nit: it seems like this is more like an impl. of `Codec::encode` since it takes a `Vec<u8>` as an argument instead of constructing one on behalf of the caller ala `get_encoding`.

rustls/src/msgs/persist.rs
```diff
@@ -197,6 +232,37 @@ impl Tls12ClientSessionValue {
...
+
+        Ok(Self {
+            suite,
+            session_id: SessionID::read(r)?,
+            extended_ms: u8::read(r)? == 1,
+            common: ClientSessionCommon::read(r)?,
+        })
+    }
+
+    /// Inherent implementation of the [`Codec::get_encoding()`] method.
```
cpu, line 256: Ditto earlier nit.

rustls/src/msgs/persist.rs
```diff
@@ -245,6 +311,30 @@ impl ClientSessionCommon {
         }
     }
 
+    /// [`Codec::read()`] is inherent here to avoid leaking the [`Codec`]
+    /// implementation through [`Deref`] implementations on
+    /// [`Tls12ClientSessionValue`] and [`Tls13ClientSessionValue`].
```
cpu, line 316: Double checking my own understanding: it would be tempting to `impl Codec for ClientSessionCommon` but since `Tls13ClientSessionValue` and `Tls12ClientSessionValue` deref to `ClientSessionCommon` we'd be leaking the `Codec` trait. Instead similar methods are directly imp'd on the concrete types and in the future we may want to clean up the `ServerSessionValue` type similarly (since it presently `impl`'s `Codec` directly).

Is that a fair summary or am I off the mark?

cpu (commented):
I have less context on the work from https://github.com/rustls/rustls/pull/1145 than other reviewers might but I went through this diff alongside the related parts that were removed in 1145 and don't see any glaring issues.

From a broader perspective are the tests you added enough to make sure the capability to write a serializing `ClientSessionStore` won't regress accidentally in the future? I _think_ so, but wanted to ask explicitly.

djc:
@ctz after posting this I seemed to remember we did discuss this at some point and you made the argument that client users probably wouldn't want/need to serialize their cache entries anyway, which I think seemed like a fair argument at the time. So if you want to just keep this out, I'd be fine with that, too. (I looked at my review feedback in #1145 but didn't find this discussion there.)

rustls/src/msgs/persist.rs
djc: Yup, copy/pasta error. Thanks for catching it!

ctz:
yeah i wasn't really convinced back at the time that this was worth the extra code. it seemed like this serialisation is going to be in the effort of either keeping the cache on disk (that seems dangerous because we need durable deletion of items in the cache now) or distributing it between many clients on the same machine (eg, in shared memory -- but that seems hard too for the same reason)

djc:
Right, it's gone!

PR #1229, opened 2023-03-21
Move three internal enums to public API

jsha:
These enums (AlertDescription, ContentType, and HandshakeType) were previously only available as part of the private API.  Eight months ago we added a public reexport of their names, but did not remove the private version to avoid semver breakage.

Now that we have a semver-incompatible version coming up we can move these fully to the public API.

This follows up on https://github.com/rustls/rustls/issues/1036#issuecomment-1140389210, and follows the example of #1048 where we previously made enums formally part of the public API by moving them out of the `msgs` directory.


ctz (approved):

djc:
Nice!

PR #1230, opened 2023-03-21
Privatize ClientSessionValue APIs

djc:


ctz:
i think we should actually `pub(crate)` all the pub fns here -- the only thing reasonable to do with values of these types outside the crate is to store and return them.

djc:
@ctz good call, done.

ctz (approved):

cpu (approved):

jsha (approved):

PR #1231, opened 2023-03-21
Move DigitallySignedStruct out of msgs

djc:
See #920. This gets used in the verifier implementations in `verify`, so I think it makes sense to move it there.


cpu (commented):
> This gets used in the verifier implementations in verify, so I think it makes sense to move it there.

Makes sense to me :+1:

cpu (approved):
Upgrading my comment to a more formal approval. 

I think this change is sensible, self-contained, and doesn't offer too much space for alternatives. I feel comfortable +1ing this but if the more experienced reviewers would prefer we wait for their input on PRs like this I will adjust my threshold for stamping :-)

jsha (approved):
Looks good to me too. 👍🏻

PR #1232, opened 2023-03-21
Avoid ClientSessionCommon leaking into public API

ctz:
Hoist up rewind_epoch() (test-only code called solely by bogo_shim) so it can be called on Tls12ClientSessionValue/Tls13ClientSessionValue directly.

### DIFF
--- a/rustls/src/msgs/persist.rs
+++ b/rustls/src/msgs/persist.rs
@@ -70,7 +70,7 @@ pub struct Tls13ClientSessionValue {
     suite: &'static Tls13CipherSuite,
     age_add: u32,
     max_early_data_size: u32,
-    pub common: ClientSessionCommon,
+    pub(crate) common: ClientSessionCommon,
     #[cfg(feature = "quic")]
     quic_params: PayloadU16,
 }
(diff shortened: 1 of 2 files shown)


rustls/src/msgs/persist.rs
```diff
@@ -229,7 +241,7 @@ impl ClientSessionCommon {
 
     /// Test only: wind back epoch by delta seconds.
     #[doc(hidden)]
-    pub fn rewind_epoch(&mut self, delta: u32) {
+    fn rewind_epoch(&mut self, delta: u32) {
```
djc, line 244: I would be inclined to just inline this. If not, can lose the `#[doc(hidden)]` since it's now private anyway.

djc (approved):
This seems like an improvement!

cpu (approved):
Nice work spotting this avenue for solving the problem.

rustls/src/msgs/persist.rs
cpu: >  If not, can lose the #[doc(hidden)] (and comment) since it's now private anyway.

Nice catch +1

PR #1233, opened 2023-03-22
Privatize ClientSessionStore

jsha:
In favor of ClientConfig::set_session_support, which selects among built-in implementations to either enable or disable session caching.

This allows removing the API surface for ClientSessionStore itself, and also ClientSessionMemoryCache and NoClientSessionStorage. However, these stay as doc(hidden) rather than fully private, because they are used in tests/api.rs and bogo_shim.rs, which use the public API.

Part of https://github.com/rustls/rustls/issues/466#issuecomment-1478728279.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -26,6 +26,8 @@ use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 use std::{fmt, io, mem};
 
+/// Test-only. Used in bogo_shim.
+///
 /// A trait for the ability to store client session data, so that sessions
 /// can be resumed in future connections.
 ///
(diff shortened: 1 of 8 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -155,7 +160,7 @@ pub struct ClientConfig {
     /// effect.
     ///
     /// The default is true.
-    pub enable_tickets: bool,
+    pub(crate) enable_tickets: bool,
```
cpu, line 163: If it turns out we prefer this branch's approach opposed to #1235 I think it would be sensible to adopt the https://github.com/rustls/rustls/pull/1235/commits/b5ca60b708e54f69796011dc1f94cb373f43fe37 commit that removed this field in favour of optionality for the session cache. Reading the two side-by-side I agree with @djc that having a separate bool for tickets that relies on the consumer ensuring there's a cache when the bool is set true is a little bit clumsy.

PR #1234, opened 2023-03-22
Remove DecomposedSignatureScheme trait and some public enum cleanup

djc:
See #920. `DecomposedSignatureScheme` has a single impl, for `SignatureScheme`, with two methods one of which is only used by its tests. It seems it was defined in 2017 and not touched much since.

Move `SignatureAlgorithm`, which is referenced from the local `SignatureScheme` and public in the crate root, to `crate::enums` as well, and split the enum tests between `crate::msgs::enums` and `crate::enums` as appropriate.


cpu (approved):
This looks good to me. I did a pass through GitHub code search for my own curiosity to see if I could spot anyone using this trait and came up empty handed.

_Edit: Also a nice lucky PR number :four_leaf_clover: _

jsha (approved):
Thanks for the cleanup!

jsha:
Alas, merging #1231 caused conflicts in this.

PR #1235, opened 2023-03-22
Seal ClientSessionStore trait and move test impls into the library

djc:
I like the idea of making it impossible to allow downstream users to implement the `ClientSessionStore` trait, but the `ClientSessionSupport` enum from #1233 with a hidden enum variant seems a little ugly. As an alternative, we could properly seal the trait using a private-in-public trait, moving the test implementations for the bogo shim and integration tests into the library, exposed through the `internal` module (but still `#[doc(hidden)]`).

Pro: keep the interface in the `ClientConfig` simpler.
Con: test impls in the library proper.

The first commit here simplifies the interface, unifying the confusing separation of `enable_tickets` and `session_storage` into a single field.

(Interestingly, this also avoids exposing the `rewind_epoch()` method as public API.)

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -8,10 +8,11 @@ use rustls::client::{ClientConfig, ClientConnection};
 use rustls::internal::msgs::codec::Codec;
 use rustls::internal::msgs::enums::KeyUpdateRequest;
 use rustls::internal::msgs::persist;
+use rustls::internal::ClientCacheWithoutKxHints;
 use rustls::server::{ClientHello, ServerConfig, ServerConnection};
 use rustls::{
     self, client, kx_group, server, sign, version, AlertDescription, Certificate, CertificateError,
-    Connection, DistinguishedNames, Error, InvalidMessage, NamedGroup, PeerMisbehaved, PrivateKey,
+    Connection, DistinguishedNames, Error, InvalidMessage, PeerMisbehaved, PrivateKey,
     ProtocolVersion, ServerName, Side, SignatureAlgorithm, SignatureScheme, SupportedKxGroup,
     SupportedProtocolVersion, Ticketer, ALL_KX_GROUPS,
 };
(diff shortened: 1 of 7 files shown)


djc:
It feels a bit crappy to put stuff in the library that the library doesn't actually use. Also means the compiler has more code to sift through when compiling the library (even for downstream projects).

ctz:
Yeah I really don't like the test/mock implementations being in the library.

Also convolving tickets=resumption is not correct, this removes the ability for a client to express that it wants to do TLS1.2 resumption, but _not_ with tickets. I think that's a valid thing to want (because TLS1.2 tickets are misdesigned).

jsha:
I agree that in #1233, using `#[doc(hidden)]` instead of making things properly private is ugly. When I started out, I thought more things could be private.

One goal I have is to separate our _internal_ API for client session caches from the _external_ API. The current ClientSessionStore API is convenient for our internal use, but as pointed out in https://github.com/rustls/rustls/issues/466#issuecomment-782484266, external users probably only need a simpler put/get/delete interface.

By removing ClientSessionStore (and Tls12ClientSessionValue and Tls13ClientSessionValue) from the public interface, we leave space open to introduce a different extension trait for client session caching later. The sealed trait approach more-or-less commits us to `ClientSessionStore` being the API for both internal and external implementations. Removing ClientSessionStore from the public API also allows us flexibility to change the internal interface if we choose.

Also, with the current version of this PR, all the ClientSessionStore methods wind up public in the documentation, even though users can't implement the trait. It would be better to make the methods private, to hide implementation details. But I'm guessing they need to be public because some of the tests call them to manipulate the cache? Perhaps in this approach we should `#[doc(hidden)]` all the methods.

> The first commit here simplifies the interface, unifying the confusing separation of enable_tickets and session_storage into a single field.

Unifying enable_tickets and session_storage is great. I tried to do that in #1233 but now I'm pretty sure I misunderstood the role of enable_tickets.

However, the unified API here is still confusing to me. What's the difference between None and Some(NoClientSessionStorage)? From a skim of the code it seems `None` doesn't advertise any session/ticket support, while `Some(NoClientSessionStorage)` advertises support but drops the sessions/tickets on the floor? We should clarify that in the documentation.

djc:
> Also convolving tickets=resumption is not correct, this removes the ability for a client to express that it wants to do TLS1.2 resumption, but _not_ with tickets. I think that's a valid thing to want (because TLS1.2 tickets are misdesigned).

I'm having a hard time understanding how this is supposed to work. If you look at the refactoring in #1239, it becomes obvious that `enable_tickets` currently seems to prevent all resumption. Maybe that was broken at some earlier point?

ctz:
Starting from 0.19:

server:
```
$ ./target/debug/examples/tlsserver --protover 1.2 --resumption -p 8443 --certs test-ca/ecdsa/end.fullchain --key test-ca/ecdsa/end.key --verbose http
```

(fairly straightforward TLS1.2 server with resumption)

client (full handshake):
```
$ ./target/debug/examples/tlsclient  --no-tickets --verbose --http -p 8443 --cafile test-ca/ecdsa/ca.cert --cache cache.bin localhost
(...)
[2023-03-23T17:42:04Z DEBUG rustls::client::hs] Not resuming any session
(...)
[2023-03-23T17:42:04Z TRACE rustls::client::tls12] Server cert is (...)
(...)
[2023-03-23T17:42:04Z DEBUG rustls::client::tls12] Session saved
```

client (resumed, from session saved into cache.bin from above handshake):
```
$ ./target/debug/examples/tlsclient  --no-tickets --verbose --http -p 8443 --cafile test-ca/ecdsa/ca.cert --cache cache.bin localhost
(...)
[2023-03-23T17:43:34Z DEBUG rustls::client::hs] Using ciphersuite TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
[2023-03-23T17:43:34Z DEBUG rustls::client::hs] Server agreed to resume
[2023-03-23T17:43:34Z DEBUG rustls::client::tls12] Session saved
(...)
```

djc:
Ah right, so in TLS 1.2 we can resume from just a session ID, without the use of tickets.

> Also convolving tickets=resumption is not correct, this removes the ability for a client to express that it wants to do TLS1.2 resumption, but not with tickets. I think that's a valid thing to want (because TLS1.2 tickets are misdesigned).

However, when setting `enable_tickets: false` we currently also block TLS 1.3 resumption. When you say "TLS1.2 tickets are misdesigned", are you implying that TLS 1.3 tickets are not? Should we separate configuration of resumption by protocol version, with a simple `bool` option for 1.3 and an `Option<enum Tls12Resumption { SessionIdOnly, SessionIdOrTickets }>` for 1.2?

ctz:
> However, when setting `enable_tickets: false` we currently also block TLS 1.3 resumption. When you say "TLS1.2 tickets are misdesigned", are you implying that TLS 1.3 tickets are not?

Correct.

> Should we separate configuration of resumption by protocol version, with a simple `bool` option for 1.3

This is already achievable by not providing any session storage.

> and an `Option<enum Tls12Resumption { SessionIdOnly, SessionIdOrTickets }>` for 1.2?

That seems like an improvement to me.

PR #1239, opened 2023-03-23
client: centralize and linearize resumption preparation

djc:
Trying to make this logic more accessible/precise.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -287,7 +287,7 @@ fn emit_client_hello_for_retry(
     {
         resuming_session
             .as_ref()
-            .and_then(|resuming| match (suite, resuming.tls13()) {
+            .and_then(|resuming| match (suite, resuming.map(|csv| csv.tls13())) {
                 (Some(suite), Some(resuming)) => {
                     suite
                         .tls13()?
@@ -816,15 +816,13 @@ impl ClientSessionValue {
             Self::Tls12(inner) => &inner.common,
         }
     }
-}
 
-impl persist::Retrieved<ClientSessionValue> {
-    fn tls13(&self) -> Option<persist::Retrieved<&persist::Tls13ClientSessionValue>> {
-        self.map(|v| match v {
-            ClientSessionValue::Tls13(v) => Some(v),
+    fn tls13(&self) -> Option<&persist::Tls13ClientSessionValue> {
+        match self {
+            Self::Tls13(v) => Some(v),
             #[cfg(feature = "tls12")]
-            ClientSessionValue::Tls12(_) => None,
-        })
+            Self::Tls12(_) => None,
+        }
     }
 }
 
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -298,14 +298,7 @@ fn emit_client_hello_for_retry(
                 _ => None,
             })
             .map(|resuming| {
-                tls13::prepare_resumption(
-                    &config,
-                    cx,
-                    ticket,
-                    &resuming,
-                    &mut exts,
-                    retryreq.is_some(),
-                );
+                tls13::prepare_resumption(&config, cx, &resuming, &mut exts, retryreq.is_some());
                 resuming
             })
     } else if config.enable_tickets {
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -280,7 +280,7 @@ fn emit_client_hello_for_retry(
     // Extra extensions must be placed before the PSK extension
     exts.extend(extra_exts.iter().cloned());
 
-    let fill_in_binder = if support_tls13
+    let tls13_session = if support_tls13
         && config.enable_tickets
         && resume_version == ProtocolVersion::TLSv1_3
         && !ticket.is_empty()
@@ -342,7 +342,7 @@ fn emit_client_hello_for_retry(
         }),
     };
 
-    let early_key_schedule = if let Some(resuming) = fill_in_binder {
+    let early_key_schedule = if let Some(resuming) = tls13_session {
         let schedule = tls13::fill_in_psk_binder(&resuming, &transcript_buffer, &mut chp);
         Some((resuming.suite(), schedule))
     } else {
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -394,6 +394,10 @@ fn prepare_resumption<'a>(
     cx: &mut ClientContext<'_>,
     config: &ClientConfig,
 ) -> Option<persist::Retrieved<&'a persist::Tls13ClientSessionValue>> {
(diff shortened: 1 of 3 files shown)


rustls/src/client/hs.rs
```diff
@@ -419,6 +376,56 @@ fn emit_client_hello_for_retry(
     }
 }
 
+fn prepare_resumption<'a>(
```
cpu, line 379: It was the case before you refactored as well, but I think this function is a bit light on test coverage for some of the branching points.

My runs are showing no coverage for:

* The `return None` when `!config.enable_tickets`
* The `None` and `_` branches of the `let resuming = match resuming` statement
* The two `Some` branches of the `match suite` check
* The `Some` branch that calls `can_resume_from` on `suite`

Give some of the confusion around the `enable_tickets` bool it would be nice to shore  up coverage while refactoring if isn't too gnarly. WDYT?

djc: Did you look at bogo coverage as well? It sure seemed to be able to catch a number of minor variants of this code as I was refactoring.

cpu: Ah sorry! I didn't.

Is there a trick to view the generated coverage emitted by those runs? I've been using my IDE's ability to run the Cargo test suite with coverage collection and I'm not sure I can do the same through the Bogo runner :thinking: Do you have an existing workflow that's good for that?

djc: No, I usually look at the CodeCov stuff posted in PRs but apparently Codecov has been having issues lately.

rustls/src/client/hs.rs
```diff
@@ -419,6 +376,56 @@ fn emit_client_hello_for_retry(
...
+    let suite = match suite {
+        Some(SupportedCipherSuite::Tls13(suite)) => Some(suite),
+        #[cfg(feature = "tls12")]
+        Some(SupportedCipherSuite::Tls12(_)) => return None,
+        None => None,
+    };
+
+    // If the selected cipher suite can't select from the session's, we can't resume.
+    if let Some(suite) = suite {
+        suite.can_resume_from(tls13.suite())?;
```
jsha, line 422: It's a bit confusing to call a non-side-effecty function just to take advantage of the `?` (and without using the Ok / Some return value). How about:

```suggestion
        if suite.can_resume_from(tls13.suite()).is_none() {
            return None
        }
```

rustls/src/client/hs.rs
```diff
@@ -419,6 +376,56 @@ fn emit_client_hello_for_retry(
...
+    let tls13 = match resuming.map(|csv| csv.tls13()) {
+        Some(tls13) if config.supports_version(ProtocolVersion::TLSv1_3) => tls13,
+        None if config.supports_version(ProtocolVersion::TLSv1_2) => {
+            exts.push(ClientExtension::SessionTicket(ClientSessionTicket::Offer(
+                Payload::new(resuming.ticket()),
+            )));
+            return None;
+        }
+        _ => return None,
+    };
```
jsha, line 410: It's not immediately clear from reading this that `None` corresponds to the case where the retrieved session value is for TLS 1.2. In the interest of further linearizing this logic and making it more explicit, how about:

```rust
    if let ClientSessionValue::Tls12(_) = resuming.deref() {
        if config.supports_version(ProtocolVersion::TLSv1_2) {
            exts.push(ClientExtension::SessionTicket(ClientSessionTicket::Offer(
                Payload::new(resuming.ticket()),
            )));
        }
        return None;
    }

    if !config.supports_version(ProtocolVersion::TLSv1_3) {
        return None;
    }

    let Some(tls13) = resuming.map(|csv| csv.tls13()) else {
        return None;
    };
```

jsha (changes requested):
Looks nice! Thanks for refactoring this. A couple of suggestions below. Also, the new prepare_resumption could use some documentation for what it does in different cases. Something like:

If resuming is None, adds a ticket request to extensions and returns None.

If resuming is a TLS 1.2 ticket, adds a ticket offer to extensions and returns None.

If resuming is a TLS 1.3 ticket, adds a preshared key offer to extensions and returns Some with the ticket value.

Does nothing and returns None if tickets are disabled, or if the `resuming` ticket is not compatible with the config or cipher suite.

rustls/src/client/hs.rs
djc: I don't disagree, and this is what I had at first. However, two downsides:

* Our rustfmt settings don't allow your suggested formatting, which turns it into this:

```rust
    if let Some(suite) = suite {
        if suite
            .can_resume_from(tls13.suite())
            .is_none()
        {
            return None;
        }
    }
```

* clippy warns against this approach:

```rust
warning: this block may be rewritten with the `?` operator
   --> rustls/src/client/hs.rs:422:9
    |
422 | /         if suite
423 | |             .can_resume_from(tls13.suite())
424 | |             .is_none()
425 | |         {
426 | |             return None;
427 | |         }
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#question_mark
    = note: `#[warn(clippy::question_mark)]` on by default
help: replace it with
    |
422 ~         suite
423 +             .can_resume_from(tls13.suite())?;
    |
```

Either of these in isolation might have been fine, but given both of them I'd prefer to use the form I have here.

rustls/src/client/hs.rs
djc: IMO one of the core values of the linear approach is to check each condition only once. In your suggested code, this is not the case: even though we know after the first `if` that `resuming` is `Tls13`, you need a final `if` to extract the `Retrieved<Tls13ClientSessionValue>` out of the `Retrieved<ClientSessionValue>`, but that branch will always be taken. I tried to add more comments instead to make it clearer what's going on.

rustls/src/client/hs.rs
jsha: Ah yeah, both good reasons to do it this way, though I disagree with clippy in this instance.

jsha (approved):
Looks good! 👍🏻

rustls/src/client/hs.rs
cpu: I can't seem to get CodeCov to re-run the failed task, but did find [this output](https://github.com/rustls/rustls/pull/1239/checks?check_run_id=12249465545) that indicates the `if !config.enable_tickets` return is missing coverage, as well as the `!config.supports_version(ProtocolVersion::TLSv1_3)` return. 

Installing `cargo llvm-cov` I was able to get generated `profraw` files with `LLVM_COV=$(which llvm-cov) LLVM_PROFDATA=$(which llvm-profdata) ./admin/coverage` but there's a failure to merge the results ("Unsupported instrumentation profile format version"). I might tilt at that more if I have some time.

In either event, I don't think this coverage is important enough to hold up this work.It's clearly an improvement to me :+1:

cpu (approved):

PR #1240, opened 2023-03-23
Remove RootCertStore::subjects from public API

jsha:
This was previously deprecated in favor of OwnedTrustAnchor::subject(). ~~Removing this from the public API means that VecU16OfPayloadU16 is no longer part of the public API.~~ Edit: I wasn't building with `--all-features`, and so missed the fact that VecU16OfPayloadU16 / DistinguishedNames is part of the [ClientCertVerifier API](https://docs.rs/rustls/latest/rustls/server/trait.ClientCertVerifier.html#tymethod.client_auth_root_subjects), so it has to stay for now.

Part of #920

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -87,8 +87,7 @@ impl RootCertStore {
     }
 
     /// Return the Subject Names for certificates in the container.
-    #[deprecated(since = "0.20.7", note = "Use OwnedTrustAnchor::subject() instead")]
-    pub fn subjects(&self) -> DistinguishedNames {
+    pub(crate) fn subjects(&self) -> DistinguishedNames {
         let mut r = DistinguishedNames::new();
 
         for ota in &self.roots {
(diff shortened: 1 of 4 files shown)


rustls/src/anchors.rs
```diff
@@ -87,8 +87,7 @@ impl RootCertStore {
     }
 
     /// Return the Subject Names for certificates in the container.
-    #[deprecated(since = "0.20.7", note = "Use OwnedTrustAnchor::subject() instead")]
-    pub fn subjects(&self) -> DistinguishedNames {
+    pub(crate) fn subjects(&self) -> DistinguishedNames {
```
djc, line 90: I'm confused. What's the rationale for undeprecating this?

jsha: Yeah, I'm going to do that. 👍🏻

jsha:
This was subsumed by #1242, closing.

PR #1242, opened 2023-03-23
ClientCertVerifier: remove "abort connection" return values

jsha:
[Originally](https://github.com/rustls/rustls/commit/03b1ef03d), ClientCertVerifier took an `sni` parameter to `client_auth_mandatory` and `client_auth_root_subjects`. If, for a given SNI, the implementation could not decide whether client auth was mandatory or pick a set of root subjects (for instance, because the list of appropriate root subjects was a map from hostnames to subjects), the implementation could return None, which would abort the connection.

Later, in #787, we removed the `sni` parameter. With the new Acceptor API, implementations can condition client authentication on the SNI by reading the ClientHello into an Acceptor, and then constructing an appropriate ServerConfig with a ClientCertVerifier.

Since there's no `sni` parameter, there's no useful data for `client_auth_mandatory` and `client_auth_root_subjects` to decide to return `None`. So we should remove the `Option` in their return types. This happens to allow us to remove some branches during handshaking.

While we're at it, change client_auth_root_subjects return type from (`Vec<DistinguishedName>`) to `&[DistinguishedName]`. The old signature required N+1 allocations per handshake where N = the number of subjects. The new signature can reuse a stored list of subjects, which seems like a common use case. Note: we still do those allocations internally, when we clone the returned values into a Vec<DistinguishedName> to be part of a CertificateRequestPayload, which expects to own all its data. However, this method signature allows us to make future optimizations to reduce allocations.

Related: #920

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -197,8 +197,8 @@ impl server::ClientCertVerifier for DummyClientAuth {
         self.mandatory
     }
 
-    fn client_auth_root_subjects(&self) -> Vec<DistinguishedName> {
-        Vec::new()
+    fn client_auth_root_subjects(&self) -> &[DistinguishedName] {
+        &[]
     }
 
     fn verify_client_cert(
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -193,12 +193,12 @@ impl server::ClientCertVerifier for DummyClientAuth {
         true
     }
 
-    fn client_auth_mandatory(&self) -> Option<bool> {
-        Some(self.mandatory)
+    fn client_auth_mandatory(&self) -> bool {
+        self.mandatory
     }
 
-    fn client_auth_root_subjects(&self) -> Option<Vec<DistinguishedName>> {
-        Some(Vec::new())
+    fn client_auth_root_subjects(&self) -> Vec<DistinguishedName> {
+        Vec::new()
     }
 
     fn verify_client_cert(
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -41,7 +41,6 @@ pub(super) use client_hello::CompleteClientHelloHandling;
 
 mod client_hello {
     use crate::enums::SignatureScheme;
-    use crate::kx;
     use crate::msgs::base::{Payload, PayloadU8};
     use crate::msgs::ccs::ChangeCipherSpecPayload;
     use crate::msgs::enums::NamedGroup;
@@ -66,6 +65,7 @@ mod client_hello {
         KeyScheduleEarly, KeyScheduleHandshake, KeySchedulePreHandshake,
     };
     use crate::verify::DigitallySignedStruct;
+    use crate::{kx, DistinguishedName};
 
     use super::*;
 
@@ -698,7 +698,10 @@ mod client_hello {
 
         let names = config
             .verifier
-            .client_auth_root_subjects();
+            .client_auth_root_subjects()
+            .iter()
+            .map(|n| DistinguishedName::from(n.clone()))
+            .collect::<Vec<_>>();
 
         if !names.is_empty() {
             cr.extensions
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -698,13 +698,7 @@ mod client_hello {
 
         let names = config
             .verifier
-            .client_auth_root_subjects()
-            .ok_or_else(|| {
-                debug!("could not determine root subjects based on SNI");
-                cx.common
-                    .send_fatal_alert(AlertDescription::AccessDenied);
-                Error::General("client rejected by client_auth_root_subjects".into())
-            })?;
+            .client_auth_root_subjects();
 
         if !names.is_empty() {
             cr.extensions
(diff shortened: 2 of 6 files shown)


rustls/examples/internal/bogo_shim.rs
```diff
@@ -198,8 +198,8 @@ impl server::ClientCertVerifier for DummyClientAuth {
         self.mandatory
     }
 
-    fn client_auth_root_subjects(&self) -> DistinguishedNames {
-        DistinguishedNames::new()
+    fn client_auth_root_subjects(&self) -> &[Vec<u8>] {
```
djc, line 201: I don't think this is a good idea.

First, my experience from benchmarking #1179 showed that allocations don't matter all that much in the performance of our TLS stack, and this is just a single allocation. To me, it seems perfectly valid for implementers of this trait to want to yield a list of subjects that is not tied to the lifetime of the verifier itself.

Second, I think going from the semantically useful `DistinguishedNames` return type to the completely generic `&[Vec<u8>]` seems to be a step back.

rustls/src/server/tls13.rs
```diff
@@ -696,9 +696,12 @@ mod client_hello {
...
             .push(CertReqExtension::SignatureAlgorithms(schemes.to_vec()));
 
-        let names = config
+        let names: Vec<PayloadU16> = config
             .verifier
-            .client_auth_root_subjects();
+            .client_auth_root_subjects()
+            .iter()
+            .map(|n| PayloadU16::new(n.clone()))
+            .collect();
```
djc, line 704: I'd typically use `collect::<Vec<_>>()` which avoids naming the type of the element twice (and I have a somewhat irrational dislike of type annotations).

rustls/src/verify.rs
```diff
@@ -539,14 +538,20 @@ fn prepare<'a, 'b>(
...
+    subjects: Vec<Vec<u8>>,
 }
 
 impl AllowAnyAuthenticatedClient {
     /// Construct a new `AllowAnyAuthenticatedClient`.
     ///
     /// `roots` is the list of trust anchors to use for certificate validation.
     pub fn new(roots: RootCertStore) -> Self {
-        Self { roots }
+        let subjects = roots
```
djc, line 549: Style nit: I'd avoid binding this separately, inlining this into the initialization below.

rustls/examples/internal/bogo_shim.rs
jsha: Agreed on the semantically useful name. And I like #1245 a lot. I see that it makes `DistinguishedName` a struct with private fields, so avoids exposing the `PayloadU8` type _and_ the `VecU16OfPayloadU8` type. 👍🏻

I agree in general with not prematurely optimizing on allocations, but I think it's worth avoiding public interfaces that _prevent_ us from optimizing on allocations if that later becomes something we want.

> To me, it seems perfectly valid for implementers of this trait to want to yield a list of subjects that is not tied to the lifetime of the verifier itself.

In general, yes. For the specific case of returning a list of root subjects for a client cert verifier, I think it's extremely unlikely. Roots (whether for client certificates or for server certificates) change rather infrequently. Also, with the change that remove the `sni` parameter from this call, any moderately complex behavior from this trait involves creating (or selecting) a different ClientCertVerifier with each handshake. So it's not burdensome for the lifetime of a `&[DistinguishedName]` to be tied to a ClientCertVerifier.

djc: > I agree in general with not prematurely optimizing on allocations, but I think it's worth avoiding public interfaces that _prevent_ us from optimizing on allocations if that later becomes something we want.

While I think we're in agreement, I just want to highlight that this is not just "not prematurely optimizing": in the PR linked above I avoided a zillion alllocations on the handshake path and performance was not noticeably faster (and seemed to be worse at times). So it looks to me like, in the context of the TLS handshake, most (small) allocations largely do not matter, and simplifying along other axes is more important.

> Also, with the change that remove the `sni` parameter from this call, any moderately complex behavior from this trait involves creating (or selecting) a different ClientCertVerifier with each handshake. So it's not burdensome for the lifetime of a `&[DistinguishedName]` to be tied to a ClientCertVerifier.

This is fair. However, me might want to reconsider how we document this, since we still say "Making one of these can be expensive, and should be once per process rather than once per connection." for `ServerConfig` and `ClientConfig`, and we might want to nuance that going forward to suggest that, especially on the server side, doing so is probably not overly expensive (I think the main expensive part these days is populating the client's `RootCertStore`?). This comes up somewhat regularly in issues...

jsha:
Rebased, squashed, and updated the PR description.

cpu (commented):
Seems reasonable. I'm not confident enough to stamp but I reviewed the diff and didn't come away with any questions/concerns.

djc (approved):

ctz (approved):

djc:
I rebased this and made some very minor tweaks:

* Moved the change that introduced the `AllowAnyAuthenticatedClient::subjects` field into the second commit into the first commit, which seems like a better conceptual fit
* Called `AllowAnyAuthenticatedClient::new()` from `AllowAnyAnonymousOrAuthenticatedClient::new()` in order to prevent duplicating the combinator chain to create `subjects`
* Removed the turbofishing from the `subjects` combinator chain in `AllowAnyAuthenticatedClient::new()` since it was unnecessary

Once CI passes, going to merge this.

PR #1243, opened 2023-03-24
Error: hide subvariants of InvalidMessage

jsha:
These subvariants were incidentally making enums from `internals` part of the public API:

 - CertificateStatusType
 - KeyUpdateRequest
 - ECCurveType

But these enums don't add much information. For instance, CertificateStatusType only has one valid value (OCSP), and all other values would be Unknown(u8). Reporting that level of detail for the unexpected response probably belongs more in a byte-by-byte protocol debugger than the workaday Error type.

Also, rename UnsupportedCurve to UnsupportedCurveType to be slightly more accurate.

Part of #920


djc (approved):
Makes sense.

PR #1245, opened 2023-03-24
Create newtype wrappers and remove Vec aliases

djc:
Create newtype wrappers around different usages of payload types to avoid type confusion, and remove `Vec` aliases in favor of a trait-based abstraction as discussed in #895. Alternative to parts of #1242.


jsha (approved):
Ooh, I really like this approach a lot. It exposes the `Vec`-ness of these types while also keeping specific types for the items being contained.

The only real comment I have is that TlsListElement, ListLength, and wrapped_payload could use some comments.

rustls/src/msgs/codec.rs
```diff
@@ -205,91 +205,68 @@ impl Codec for u64 {
...
-    let mut ret: Vec<T> = Vec::new();
-    let len = usize::from(u16::read(r)?);
-    let mut sub = r.sub(len)?;
-
-    while sub.any_left() {
-        ret.push(T::read(&mut sub)?);
-    }
-
-    Ok(ret)
+pub trait TlsListElement {
```
jsha, line 264:
```suggestion
pub(crate) trait TlsListElement {
```

rustls/src/msgs/codec.rs
```diff
@@ -205,91 +205,68 @@ impl Codec for u64 {
...
-    }
-
-    let mut sub = r.sub(len)?;
-
-    while sub.any_left() {
-        ret.push(T::read(&mut sub)?);
-    }
-
-    Ok(ret)
+pub enum ListLength {
```
jsha, line 268:
```suggestion
pub(crate) enum ListLength {
```

jsha (changes requested):

cpu (commented):
This looks great, clever use of a macro :+1:

djc:
Add some comments and privatized the `TlsListElement` trait and `ListLength` type.

ctz (approved):

jsha (approved):

Issue #1246, opened 2023-03-24
Certificate parsing error in 0.21.0-alpha.1


mgdm:
Hi,
I've been using rustls with rustls-native-certs and loading the certificates from the system keychain on macOS Ventura. This works fine with v0.20.8, but when I moved to the alpha it fails to parse at least one of the certificates in the system keychain. Using the system store works fine on Linux.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidByte(64, 10)', src/main.rs:31:48
```

I've created an example that reproduces the issue either with the system keychain or with the specific certificate at https://github.com/mgdm/rustls-parsing-crash.

jsha:
Thanks for the excellent bug report, along with the useful example.

At a first glance, I suspect what's needed is [RootCertStore::add_parseable_certificates](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add_parsable_certificates):

> Parse the given DER-encoded certificates and add all that can be parsed in a best-effort fashion.
> This is because large collections of root certificates often include ancient or syntactically invalid certificates.
> Returns the number of certificates added, and the number that were ignored.

We should definitely do a better job of signposting that function from rustls-native-certs and in rustls, though.

cpu:
> I've created an example that reproduces the issue either with the system keychain or with the specific certificate at https://github.com/mgdm/rustls-parsing-crash.

I don't have a MacOS machine handy to try the `cargo run system` reproducer option, but I do notice when I run `cargo run` for the embedded variant that the panic happening from unwrapping an error result from the `b64.decode` call. E.g. it happens _before_ Rustls is in the mix.

Is it possible the embedded cert's b64 DER content was mis-pasted into the repro?

cpu:
> At a first glance, I suspect what's needed is [RootCertStore::add_parseable_certificates](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add_parsable_certificates):

That seems to do the trick for me :tada: 

https://github.com/cpu/rustls-parsing-crash/commit/6d55785e972a417c18acf12da53dfda530c5ff7c

I used `rustls_pemfile` for convenience but the crux of the fix is using `add_parseable_certificates`.

mgdm:
That's certainly neater and I'll switch to that, thanks!

[...]

PR #1248, opened 2023-03-24
anchors: add a pointer from add to add_parseable.

cpu:
Users that are iterating over a pile of root certificates (e.g. from their system truststore) may call `RootCertStore::add` without realizing that it will reject invalid certificates (_and unfortunately some root stores contain a non-zero number..._). These users should prefer the more permissive `add_parseable_certificates`.

This commit introduces a rustdoc pointer from `add` to `add_parseable_certificates` to hopefully make this more discoverable.

See https://github.com/rustls/rustls/issues/1246

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -102,6 +102,10 @@ impl RootCertStore {
     }
 
     /// Add a single DER-encoded certificate to the store.
+    ///
+    /// This is suitable for a small set of root certificates that are expected to parse
+    /// successfully. If you're adding a larger collection of certificates, for instance from
+    /// a system root store, prefer [`RootCertStore::add_parsable_certificates`].
     pub fn add(&mut self, der: &key::Certificate) -> Result<(), Error> {
         let ta = webpki::TrustAnchor::try_from_cert_der(&der.0)
             .map_err(|_| Error::InvalidCertificate(CertificateError::BadEncoding))?;


rustls/src/anchors.rs
```diff
@@ -102,6 +102,9 @@ impl RootCertStore {
     }
 
     /// Add a single DER-encoded certificate to the store.
+    ///
+    /// If you are processing a number of root certificates that may include ancient,
+    /// or syntactically invalid certificates, prefer [`RootCertStore::add_parsable_certificates`].
```
jsha, line 107: I think most users aren't equipped to evaluate whether their root store contains ancient or syntactically invalid certificates. Indeed I suspect most people would assume that on an up-to-date system it contains neither!

Instead let's say: This is suitable if you're bringing in a small set of root certificates that you know will parse successfully. If you're parsing a larger collection of certificates, for instance from a system root store, it's better to call [`RootCertStore::add_parsable_certificates`].

cpu: > I think most users aren't equipped to evaluate whether their root store contains ancient or syntactically invalid certificates.

You're totally right. Thanks for calling that out. I pushed a revised version closer to your wording.

> Indeed I suspect most people would assume that on an up-to-date system it contains neither!

Haha. True, not everyone is a PKI cynic (yet) :zipper_mouth_face: 

>  If you're parsing a larger collection of certificates, for instance from a system root store, it's better to call [RootCertStore::add_parsable_certificates].

We should also adjust the [usage example](https://docs.rs/rustls-native-certs/latest/rustls_native_certs/) from `rustls-native-certs`, as it contradicts this advice and runs the root store certs through an `add` loop. I can look at fixing that next (but might not get to it tonight).

jsha (approved):

rustls/src/anchors.rs
cpu: Put up a draft for `rustls-native-certs`: https://github.com/rustls/rustls-native-certs/pull/62

rustls/src/anchors.rs
```diff
@@ -102,6 +102,10 @@ impl RootCertStore {
     }
 
     /// Add a single DER-encoded certificate to the store.
+    ///
+    /// This is suitable for a small set of root certificates that are expected to parse
```
djc, line 106: I'm nitpicking, but I feel like this is pretty opaque language. I'd want to know how this is different in order to understand when/why I use one or the other.

cpu: What about if I tweaked the final sentence from:
> "If you're adding a larger collection of certificates, for instance from a system root store, prefer [`RootCertStore::add_parsable_certificates`]."

to something like:

> "If you're adding a larger collection of certificates, for instance from a system root store, prefer [`RootCertStore::add_parsable_certificates`] to allow ignoring malformed roots without surfacing 
an error."

I think that "ignore bad roots and add what you can" is the difference to call out.

djc: But how is that ultimately different? It's kind of trivial to ignore the error from this function, too, right?

cpu: That's a fair point. Do you think there's any utility in `add_parseable_certificates` from that perspective?

cpu: @djc Take a look at https://github.com/rustls/rustls/pull/1249. If that's still not to your liking I can revert the original commit and we can go back to the drawing board.

PR #1249, opened 2023-03-24
anchors: tweak rustdoc comment about malformed roots.

cpu:
This is a quick follow up to https://github.com/rustls/rustls/pull/1248, which I merged before all review input had arrived (_mea culpa_).

The previous comment added to `RootCertStore.add` pointing out the existence of `add_parsable_certificates` didn't offer enough clarity around when to prefer `add` vs `add_parsable_certificates`. This commit tweaks the language further based on review feedback.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -104,8 +104,12 @@ impl RootCertStore {
     /// Add a single DER-encoded certificate to the store.
     ///
     /// This is suitable for a small set of root certificates that are expected to parse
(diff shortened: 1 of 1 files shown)


cpu:
> I think this is the right message, but the language feels a bit clunky in places.

@jsha WDYT about this branch updating the wording we merged earlier?

jsha:
Looks good to me!

cpu:
@djc Is there a particular part you still find clunky? It doesn't read awkwardly to me but I have author's bias.

ctz (approved):

PR #1251, opened 2023-03-27
Further client resumption tweaks

djc:
Includes the resumption enum discussed in #1235 in the last commit, but I'm honestly not sure it's worth it? There seem to be a lot of little edge cases with 1.2 resumption that still remain unaddressed; for example, after resuming we store the same ticket again, but with a `lifetime` of 0, which seems kind of useless? The client should also be discarding the session ID if it received a ticket, but it's currently storing both. Not sure how much of this is intentional.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -155,7 +155,7 @@ pub struct ClientConfig {
     /// effect.
     ///
     /// The default is true.
-    pub enable_tickets: bool,
+    pub tls12_resumption: Option<Tls12Resumption>,
 
     /// Supported versions, in no particular order.  The default
     /// is all supported versions.
@@ -187,12 +187,26 @@ pub struct ClientConfig {
     pub enable_early_data: bool,
 }
 
+/// What mechanisms to support for resuming a TLS 1.2 session.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Tls12Resumption {
+    /// Support 1.2 resumption using session ids only.
+    SessionIdOnly,
+    /// Support 1.2 resumption using session ids or RFC 5077 tickets.
+    ///
+    /// See[^1] for why you might like to disable RFC 5077 by instead choosing the `SessionIdOnly`
+    /// option. Note that TLS 1.3 tickets do not have those issues.
+    ///
+    /// [^1]: <https://words.filippo.io/we-need-to-talk-about-session-tickets/>
+    SessionIdOrTickets,
+}
+
 impl fmt::Debug for ClientConfig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("ClientConfig")
             .field("alpn_protocols", &self.alpn_protocols)
             .field("max_fragment_size", &self.max_fragment_size)
-            .field("enable_tickets", &self.enable_tickets)
+            .field("tls12_resumption", &self.tls12_resumption)
             .field("enable_sni", &self.enable_sni)
             .field("enable_early_data", &self.enable_early_data)
             .finish_non_exhaustive()
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -17,6 +17,7 @@ use crate::versions;
 use crate::ExtractedSecrets;
 use crate::KeyLog;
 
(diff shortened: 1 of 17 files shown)


jsha (commented):
This looks good to me.

I'm also not totally enthusiastic about the enum for enable_tickets / tls12_resumption. I think what we want is to unify the enable_tickets / tls12_resumption setting with the ClientSessionStore trait, so they are set at the same time and form a coherent policy.

djc:
I started down that road but then wasn't sure it was worth it. I can resurrect that approach tomorrow.

rustls/src/client/client_conn.rs
```diff
@@ -186,12 +186,21 @@ pub struct ClientConfig {
     pub enable_early_data: bool,
 }
 
+/// What mechanisms to support for resuming a TLS 1.2 session.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Tls12Resumption {
+    /// Support 1.2 resumption using session ids only.
+    SessionIdOnly,
+    /// Support 1.2 resumption using RFC 5077 tickets.
```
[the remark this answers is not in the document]
djc: Yup, fixed that up.

djc:
Okay, I think this is in a pretty good shape. I tweaked the work from #1259 a bit:

* Changed the `Tls12Resumption` enum to contain a `Disabled` variant instead of wrapping it in `Option`
* I rolled back the changes to use custom caches in the bogo shim and tests, instead keeping `ClientSessionMemoryCache::new()` public -- I think chances are that other users rely on wrapping the cache, so they need to construct it
* Pulled the renaming of the `Resumption` type from `bench` out into a separate commit

I think this is a nicer, more misuse-resistant public API.

djc:
> I think changing the visibility of the `ClientSessionMemoryCache` constructor seems reasonable if you think there are external users taking this approach, but what about keeping the reworked test from [7cb3b0b](https://github.com/rustls/rustls/commit/7cb3b0b6c43e8eaf23bee2d5913289867aa60ccf) & [a94f4c6](https://github.com/rustls/rustls/commit/a94f4c635b31750f10ce4b7be3cea10a979a831f) ? I think I lightly preferred the way it approached getting the same coverage but relying on externally visible properties.

Hmm, it seems okay I guess? All things being equal, I'd rather avoid adding more dependencies from the integration tests on `internals` API.

cpu:
>  I'd rather avoid adding more dependencies from the integration tests on internals API.

That's fair, I think conceptually I wasn't thinking of these as integration tests but I can see how they do fill that role.

djc:
> > I'd rather avoid adding more dependencies from the integration tests on internals API.
> 
> That's fair, I think conceptually I wasn't thinking of these as integration tests but I can see how they do fill that role.

Sorry, yes -- mostly referring to the Cargo organization, they're in the `tests` dir and don't get privileged access to the `crate` API.

rustls/src/client/client_conn.rs
```diff
@@ -186,12 +181,23 @@ pub struct ClientConfig {
...
 }
 
+/// What mechanisms to support for resuming a TLS 1.2 session.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Tls12Resumption {
+    /// Disable 1.2 resumption.
+    Disabled,
+    /// Support 1.2 resumption using session ids only.
+    SessionIdOnly,
+    /// Support 1.2 resumption using session ids or RFC 5077 tickets.
```
ctz, line 191:
```suggestion
    /// Support 1.2 resumption using session ids or RFC 5077 tickets.
    ///
    /// See[1] for why you might like to disable RFC5077 by instead choosing the `SessionIdOnly` option.
    /// Note that TLS1.3 tickets do not have those issues.
    /// [^1]: https://words.filippo.io/we-need-to-talk-about-session-tickets/
```
(writing the markdown blind, please check)

ctz (approved):

rustls/src/client/client_conn.rs
cpu: Folded into 55e1c6b7f1d2727554c7ef7978934245ce9de604 with a rebase. I tweaked the markdown slightly and verified it rendered correctly in `cargo doc --all-features`.

ctz:
(will need a bit of remedial test/coverage work, will look at that this weekend)

PR #1252, opened 2023-03-27
error: restore PartialEq implementation

jsha:
Previously this was removed when adding `Arc<dyn StdError>` as an enum field for CertificateError, which made it impossible to automatically derive PartialEq for the whole enum. Trait objects (dyn Trait) don't implement PartialEq.

However, we can get back PartialEq on the whole Error struct by manually implementing it for CertificateError, considering `Other` values to never be equal.

Follow-up to https://github.com/rustls/rustls/pull/1182 and https://github.com/rustls/rustls/pull/1182/commits/00d7ac50f90dccc01bf69c3359680282d81a53c9.


complexspaces:
I think I agree with the new API design around `Other`, FWIW. Custom verifiers can use downcasting to extract error types that can implement `PartialEq` themselves without having to burden `rustls` with that ability (while still being better then `String`).

While I don't think there will be much support for it since it introduces panics, could potential test failure causes be made more obvious by panicking with a helpful message when `#[cfg(debug_assertions)]` are enabled and `Other` is compared with itself, indicating that the callers error handling should be updated?

jsha:
> could potential test failure causes be made more obvious by panicking with a helpful message when #[cfg(debug_assertions)] are enabled and Other is compared with itself, indicating that the callers error handling should be updated?

I think cases of comparing Other to Other are likely uncommon enough that it's not worth adding a `#[cfg(debug_assertions)]` just for that.

complexspaces:
> This seems pretty minor/diminishing returns?,  Other to Other are likely uncommon enough that...

I agree that it has pretty minor returns. I was mostly just tossing the idea out to see if anyone could think of a scenario where it may be more helpful.

> What scenario are you worried about?

I don't have an exact scenario, just a general worry about if someone writing tests that involve `rustls` doesn't carefully read the docs, they'd have a frustrating subtle test failure to debug. I don't think I've once suspected a misleading `PartialEq` implementation writing tests in any project.

jsha:
> I don't think I've once suspected a misleading PartialEq implementation writing tests in any project.

The most minor of quibbles: [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) is specifically intended to allow partial equalities, with some items that don't compare equal to each other. If we implemented [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html) that would be misleading and incorrect.

Still, I do see your point: Rust uses PartialEq for `==` and most of the time we tend to assume it's fully reflexive; particularly for an enum it's a bit unusual for it not to be.

djc:
I think my take is that there's never a test case you could write where you'd expect to compare two `Other` values while wanting the result to be `true`. So my take is that yes, this would make for a very crappy potentially long debugging session, but the chances of that session happening/being necessary seem quite slim.

complexspaces:
> I think my take is that there's never a test case you could write where you'd expect to compare two Other values while wanting the result to be true

I'm not sure that I entirely agree? If a verifier wants to test specific error cases that the predefined `rustls::Error` variants don't cover, that failure would need to be conveyed via `Other`. It would then follow that the test case expected result would contain an `Other` too. I digress though, since its a very small target group because those people would probably be aware of the exacts of `rustls::Error`'s and `InvalidCertificate`.

PR #1257, opened 2023-03-28
examples: fix building w/o logging feature.

cpu:
Similar to https://github.com/rustls/rustls/pull/1255, cleaning up some breakages I noticed trying to build/test things without default features.

## examples: fix no logging feature usage.
Previously the `tlsserver-mio.rs` example unconditionally used `extern crate log`, causing errors of the following form to be emitted when building examples without the "logging" feature enabled.

<details><summary>Example err:</summary>
<p>

```
error[E0658]: use of unstable library feature 'rustc_private': this
crate is being loaded from the sysroot, an unstable location; did you
mean to load this crate from crates.io via `Cargo.toml` instead?
 --> examples/src/bin/tlsserver-mio.rs:6:1
  |
6 | extern crate log;
  | ^^^^^^^^^^^^^^^^^
  |
```
</p>
</details> 

This commit resolves the breakage, allowing building the examples without logging. To do two of the the log macro shims from rustls/lib.rs are duplicated for use from the example code for this situation. Once fixed, a couple of new clippy findings are addressed for the case where no logging support is enabled and some variables thus appear "unused".

## CI: test examples with --no-default-features.
Previously the example directories weren't being tested with `--no-default-features`, letting bitrot affect those configurations.

This commit includes those directories in the `--no-default-features` task that run `cargo test`.

### DIFF
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -3,7 +3,15 @@ use std::sync::Arc;
 use mio::net::{TcpListener, TcpStream};
 
 #[macro_use]
+#[cfg(feature = "logging")]
 extern crate log;
+
+#[cfg(not(feature = "logging"))]
+#[macro_use]
+mod log {
+    macro_rules! debug    ( ($($tt:tt)*) => {{}} );
+    macro_rules! error    ( ($($tt:tt)*) => {{}} );
+}
 
 use std::collections::HashMap;
 use std::fs;
@@ -62,8 +70,8 @@ impl TlsServer {
     fn accept(&mut self, registry: &mio::Registry) -> Result<(), io::Error> {
         loop {
             match self.server.accept() {
-                Ok((socket, addr)) => {
-                    debug!("Accepting new connection from {:?}", addr);
+                Ok((socket, _addr)) => {
+                    debug!("Accepting new connection from {:?}", _addr);
 
                     let tls_conn =
                         rustls::ServerConnection::new(Arc::clone(&self.tls_config)).unwrap();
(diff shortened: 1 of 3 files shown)


rustls/src/lib.rs
```diff
@@ -306,10 +306,15 @@ use log;
...
+pub mod log {
+    #[macro_export]
     macro_rules! trace    ( ($($tt:tt)*) => {{}} );
+    #[macro_export]
     macro_rules! debug    ( ($($tt:tt)*) => {{}} );
+    #[macro_export]
     macro_rules! warn     ( ($($tt:tt)*) => {{}} );
+    #[macro_export]
     macro_rules! error    ( ($($tt:tt)*) => {{}} );
 }
```
cpu, line 319: As an alternative to exposing these macros from the rustls crate we could also duplicate the `debug` and `error` macro rules in the `tlsserver-mio.rs` example.

I went with this approach to start but don't feel strongly about it.

cpu: I changed my mind. It's silly to export this stuff for one example. Force pushed a changed.

examples/src/bin/tlsserver-mio.rs
```diff
@@ -3,8 +3,16 @@ use std::sync::Arc;
...
 #[macro_use]
+#[cfg(feature = "logging")]
 extern crate log;
 
+#[cfg(not(feature = "logging"))]
+#[macro_use]
+mod log {
+    macro_rules! debug    ( ($($tt:tt)*) => {{}} );
+    macro_rules! error    ( ($($tt:tt)*) => {{}} );
+}
```
cpu, line 14: If folks dislike duplicating these macro rules we could also export them from the rustls lib.rs - I had that in an earlier version but decided a little duplication was more appetizing than additional API surface.

djc:
IMO it would be reasonable to make logging non-optional in the examples crate, which I think would make this simpler?

cpu:
>  IMO it would be reasonable to make logging non-optional in the examples crate, which I think would make this simpler?

Sure, that would work too :+1: It's definitely a smaller diff :-) Here's an alternative PR taking that approach to consider as a replacement for this one: https://github.com/rustls/rustls/pull/1258

cpu:
Closing in favour of https://github.com/rustls/rustls/pull/1258

PR #1259, opened 2023-03-28
Use type + builder pattern for client resumption config API

djc:
@jsha been a busy day, I was thinking along the lines of something like this. Instead of having a setter on `ClientConfig`, use a field with a type that uses some kind of builder pattern to wrap up states that make sense. This doesn't quite build and the builder methods don't quite abstract over the right states, but would like your thoughts.

Let's "lock" this branch explicitly: anyone should feel free to hack on it to whip it into shape, I can pick it up in the morning and polish it up.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -242,17 +242,20 @@ impl ClientConfig {
 /// Configuration for how/when a client is allowed to resume a previous session.
 #[derive(Clone)]
 pub struct Resumption {
-    /// How we store session data or tickets.
+    /// How we store session data or tickets. The default is to use an in-memory
+    /// [ClientSessionMemoryCache].
     pub(super) store: Arc<dyn ClientSessionStore>,
 
-    /// The default is true.
+    /// What mechanism is used for resuming a TLS 1.2 session. None, or
+    /// one of the choices from [Tls12Resumption].
     pub(super) tls12_resumption: Option<Tls12Resumption>,
 }
 
 impl Resumption {
     /// Create a new `Resumption` that stores data for the given number of sessions in memory.
     ///
-    /// This is default. By default, use of tickets is enabled.
+    /// This is the default `Resumption` choice, and enables resuming a TLS 1.2 session with
+    /// a session id or RFC 5077 ticket.
     pub fn in_memory_sessions(num: usize) -> Self {
         Self {
             store: ClientSessionMemoryCache::new(num),
@@ -262,7 +265,7 @@ impl Resumption {
 
     /// Use a custom [`ClientSessionStore`] implementation to store sessions.
     ///
-    /// By default, use of tickets is enabled.
+    /// By default, enables resuming a TLS 1.2 session with a session id or RFC 5077 ticket.
     pub fn store(store: Arc<dyn ClientSessionStore>) -> Self {
         Self {
             store,
@@ -278,7 +281,7 @@ impl Resumption {
         }
     }
 
-    /// Disable the use of RFC 5077 tickets for resumption.
+    /// Configure whether TLS 1.2 sessions may be resumed, and by what mechanism.
     ///
     /// This is meaningless if you've disabled resumption entirely.
     pub fn tls12_resumption(mut self, tls12: Option<Tls12Resumption>) -> Self {
@@ -296,6 +299,8 @@ impl fmt::Debug for Resumption {
 }
 
 impl Default for Resumption {
+    /// Create an in-memory session store resumption with up to 256 server names, allowing
+    /// a TLS 1.2 session to resume with a session id or RFC 5077 ticket.
     fn default() -> Self {
         Self::in_memory_sessions(256)
     }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -117,7 +117,8 @@ pub trait ResolvesClientCert: Send + Sync {
 /// # Defaults
 ///
 /// * [`ClientConfig::max_fragment_size`]: the default is `None`: TLS packets are not fragmented to a specific size.
-/// * [`ClientConfig::session_storage`]: the default stores 256 sessions in memory.
+/// * [`ClientConfig::resumption`]: supports resumption with up to 256 server names, using session
+///    ids or tickets, with a max of eight tickets per server.
 /// * [`ClientConfig::alpn_protocols`]: the default is empty -- no ALPN protocol is negotiated.
 /// * [`ClientConfig::key_log`]: key material is not logged.
 #[derive(Clone)]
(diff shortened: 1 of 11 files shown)


cpu:
> This doesn't quite build 

Working on some of that. I'll push a commit with `bench.rs` fixes in a moment.

cpu:
> Let's "lock" this branch explicitly

Releasing my lock :-)

I fixed a few of the low hanging compilation errors. The remaining are:
* usage of a private `ClientSessionMemoryCache::new` constructor
* usage of the private `Resumption.store` field.

I need to review the diff more carefully to have ideas on how to proceed. I'll let someone else take the lock in the meantime.

cpu:
> usage of a private ClientSessionMemoryCache::new constructor

One of these is L553 of `bogo_shim.rs` and is an easy fix:
```diff
7c7
< use rustls::client::{ClientConfig, ClientConnection};
---
> use rustls::client::{ClientConfig, ClientConnection, Resumption};
553,554c553
<     let persist = ClientCacheWithoutKxHints::new(opts.resumption_delay);
<     cfg.resumption.store = persist;
---
>     cfg.resumption = Resumption::store(ClientCacheWithoutKxHints::new(opts.resumption_delay));
```

cpu:
> Releasing my lock :-)
..
> One of these is L553 of bogo_shim.rs and is an easy fix:

Taking my lock back to fix these "private Resumption.store field." instances.

cpu:
> Taking my lock back 

:unlock: 

Just two more compilation errors, one in `api.rs` and one in `bogo_shim.rs`, both trying to build `ClientSessionMemoryCache` w/ private associated `new`.

jsha:
I like this approach. Quoting the API here:

> impl Resumption
> pub fn in_memory_sessions(num: usize) -> Self
> Create a new Resumption that stores data for the given number of sessions in memory.
>
> This is default. By default, use of tickets is enabled.
> 
> pub fn store(store: Arc<dyn ClientSessionStore>) -> Self
> Use a custom ClientSessionStore implementation to store sessions.
> 
> By default, use of tickets is enabled.
>
> pub fn disabled() -> Self
> Disable all use of session resumption.
> 
> pub fn tls12_resumption(self, tls12: Option<Tls12Resumption>) -> Self
> Disable the use of RFC 5077 tickets for resumption.
> 
> This is meaningless if you’ve disabled resumption entirely.

It's a little confusing to have a builder-style API where all but one of the methods is a constructor, and there's only one mutator/consumer. But I think it's fine. And it does have nice properties: it lets us add additional constructors for other config types we might want in the future, and most users don't have to interact directly with the ClientSessionStore trait or its implementations.

Also, in place of `Option<Tls12Resumption>` for a tri-state setting, I think it's clearer to flatten things out into `enum Tls12Resumption { Disabled, SessionIdOnly, SessionIdOrTickets }` (or maybe `Enabled` for that last one?). But I recognize this is a matter of taste; I'm curious to hear others' taste on this topic.

I'll take the lock 🔒 and see what I can do about those other compile errors.

cpu:
> Just two more compilation errors, one in api.rs and one in bogo_shim.rs, both trying to build ClientSessionMemoryCache w/ private associated new.

I think we might need to back out the `pub` -> `pub(crate)` change in `src/client/handy.rs` to accommodate these two test users that want to wrap a `ClientSessionMemoryCache`. Or, we could lift the "wrapped" test types up into the namespace that has visibility through `pub(crate)` but that feels a little icky too.


 ```diff
 74c74
<     pub(super) fn new(size: usize) -> Arc<Self> {
---
>     pub fn new(size: usize) -> Arc<Self> {
 ```

If I make that change this branch builds and passes tests (bogo included). I haven't taken the lock and pushed a change because it feels like it might be working counter to the goals of this branch to expose the in-memory session cache impl. Maybe someone else has a better idea?

cpu:
> I think it's clearer to flatten things out into enum Tls12Resumption { Disabled, SessionIdOnly, SessionIdOrTickets } (or maybe Enabled for that last one?). But I recognize this is a matter of taste; I'm curious to hear others' taste on this topic.

That does sound clearer to me :+1:

cpu:
> I've gotten things building by moving bogo and tests/api.rs to their own implementation of ClientSessionStore.

That's a nice third option I didn't think of :brain: 

> Holding onto the lock.

ACK - I'm going to be offline for a bit but can poke in later to see if I can be any more help before tomorrow morning.

rustls/src/client/client_conn.rs
```diff
@@ -244,6 +239,68 @@ impl ClientConfig {
     }
 }
 
+/// Configuration for how/when a client is allowed to resume a previous session.
+#[derive(Clone)]
+pub struct Resumption {
+    /// How we store session data or tickets.
+    pub(super) store: Arc<dyn ClientSessionStore>,
+
+    /// The default is true.
```
cpu, line 248: I think we'll want to change the wording here. Default is `in_memory_sessions(256)` which implies `Some(Tls12Resumption::SessionIdOrTickets)` as the default for `tls12_resumption`

jsha: Makes sense.

jsha:
I've updated with the test fixes I wanted and am releasing the lock.🔓

I think this could still use a little documentation love, as @cpu pointed out in one spot so far.

cpu:
> I've updated with the test fixes I wanted and am releasing the lock.unlock

The new test looks great. Thanks!

> I think this could still use a little documentation love, as @cpu pointed out in one spot so far.

I'll take the :lock: and do a quick pass for that.

cpu:
I think we also have to rewrite some of the test syntax to be 1.57 friendly :disappointed:

rustls/src/client/client_conn.rs
cpu: Reworded in f906c29c399f3a5347ed6118607677929edf78b9

cpu:
>>I think this could still use a little documentation love, as @cpu pointed out in one spot so far.
>
>I'll take the lock and do a quick pass for that.

Done :ballot_box_with_check: : https://github.com/rustls/rustls/pull/1259/commits/32313df03f2bb2ebbe638cab1bfc462bc5ebf85e

> I think we also have to rewrite some of the test syntax to be 1.57 friendly disappointed

Done :ballot_box_with_check: : https://github.com/rustls/rustls/pull/1259/commits/a94f4c635b31750f10ce4b7be3cea10a979a831f

Unfortunately `let...else` statements are unstable so I had to make the test code uglier, but compatible with the MSRV.

I'm going to yield the lock and sign off for the evening :wave: :unlock:

jsha:
Looks good, thanks for that! I'm satisfied with the current state.

djc:
Merged this into #1251.

PR #1260, opened 2023-03-29
Reduce CIFuzz timeout to 300s

djc:
In #862, we added CIFuzz integration. This is great, but it runs until a
specific timeout (currently 600s), in addition to needing about 6 minutes
of setup time. This is more than 3 times as long as the next longest CI job
(coverage took about 4m in #1242 just now). Given that we already have
OSS-Fuzz spending quite a bit of time running fuzzing for us and that (as
far as I remember) we have yet to find any issues from the CIFuzz integration,
I feel safe reducing the runtime for CIFuzz to 150s. This should still cover
quite a bit of ground given that we're executing pretty fast Rust code.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -15,7 +15,7 @@ jobs:
       uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
       with:
         oss-fuzz-project-name: 'rustls'
-        fuzz-seconds: 600
+        fuzz-seconds: 150
         dry-run: false
         language: rust
     - name: Upload Crash


djc:
What I missed before is that, in addition to the 600s of run-time, CIFuzz needs about 6.5 minutes to prepare the environment, so before this PR it's taking closer to 16 minutes. This will shave that down to 11 minutes, but maybe we should go a bit lower still, like taking 150s of runtime instead?

cpu (approved):
This makes sense to me. I've also noticed the CIFuzz job being the one I'm waiting on the longest when everything else is ready to go.

Thanks for tweaking.

PR #1265, opened 2023-03-29
docs: remove future release notes.

cpu:
This commit removes two "Future release" items from the release history section of the README.

It seems clearer to have this section dedicated to the release history, not upcoming work. I also think the two described pieces of work might not be what the project is currently prioritizing.


djc (approved):

cpu:
Merging with one review since this seems relatively uncontroversial. Happy to revert/tweak if new opinions arrive.

PR #1267, opened 2023-03-30
handshake: show illegal SNI hostname as lossy str.

cpu:
Previously when an illegal SNI hostname was received rustls would `warn!` the received value as a raw `Vec<u8>`, making it hard for a human to read the value received. E.g.:
```
WARN connection{id=2 source="RedisSource"}: rustls::msgs::handshake: Illegal SNI hostname received [49, 50, 55, 46, 48, 46, 48, 46, 49] 
```

This commit changes to `warn!` the `from_utf8_lossy` string version of the hostname. This will make it easier for end users to diagnose the root cause.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -229,7 +229,10 @@ impl ServerNamePayload {
             match webpki::DnsNameRef::try_from_ascii(&raw.0) {
                 Ok(dns_name) => dns_name.into(),
                 Err(_) => {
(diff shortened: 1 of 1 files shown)


jsha (approved):

cpu:
> Merging #1267 (85f22fd) into main (a0501e4) will decrease coverage by 0.02%.

I looked at this and didn't think it was worth the trouble to try and unit test `warn!` logging output. Can revisit if folks prefer to test those codepaths as well.

djc:
Nice!

Issue #1272, opened 2023-04-05
Move docs from ConnectionCommon to Connection


jsha:
Most of the methods on Connection refer to docs on ConnectionCommon. For instance, [complete_io](https://docs.rs/rustls/0.21.0/rustls/enum.Connection.html#method.complete_io):

> This function uses io to complete any outstanding IO for this connection.
>
> See [ConnectionCommon::complete_io()](https://docs.rs/rustls/0.21.0/rustls/struct.ConnectionCommon.html#method.complete_io) for more information.

However, ConnectionCommon isn't used directly by the user- as far as I can tell, there's no way for a user to access ConnectionCommon. So it makes more sense to put the real documentation on Connection, and change ConnectionCommon's docs to they refer to Connection.

jsha:
@djc explained in https://github.com/rustls/rustls/pull/1274#issuecomment-1499509563 why this is the way it is. 👍🏻

PR #1274, opened 2023-04-06
docs: move details ConnectionCommon -> Connection.

cpu:
Previously the `Connection` rustdoc comments were short and directed the reader to `ConnectionCommon` for most of the detail.

Since `ConnectionCommon` is largely a detail internal to Rustls this commit flips this pattern. The detailed doc comments are moved to the `Connection` type and the `ConnectionCommon` doc comments refer back to those.

Resolves https://github.com/rustls/rustls/issues/1272

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -25,9 +25,25 @@ pub enum Connection {
 }
 
 impl Connection {
-    /// Read TLS content from `rd`.
+    /// Read TLS content from `rd` into the internal buffer.
     ///
-    /// See [`ConnectionCommon::read_tls()`] for more information.
+    /// Due to the internal buffering, `rd` can supply TLS messages in arbitrary-sized chunks (like
+    /// a socket or pipe might).
+    ///
+    /// You should call [`process_new_packets()`] each time a call to this function succeeds in order
+    /// to empty the incoming TLS data buffer.
+    ///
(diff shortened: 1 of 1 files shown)


jsha (approved):

djc:
So the reason I did it this way is actually that `ClientConnection` and `ServerConnection`, which I think are more likely to be used in most applications than the `Connection` wrapper, sort of expose these methods through a `Deref<Target = ConnectionCommon<Data>>`. Therefore I actually think the `ConnectionCommon` versions of these methods are more likely to be read/referenced than the `Connection` version. So this was entirely on purpose.

jsha:
Aha, now that you say that it totally makes sense, and I got mixed up about which type would more commonly be used by the user. So it seems Connection is meant to be used (via its `From<ClientConnection>` and `From<ServerConnection>` impls) by code that wishes to generalize across client vs server?

At any rate, @cpu sorry for the unnecessary work here.

djc:
> So it seems Connection is meant to be used (via its `From<ClientConnection>` and `From<ServerConnection>` impls) by code that wishes to generalize across client vs server?

Yup. (You can also abstract over it, with something like `impl Deref<Target = ConnectionCommon<impl SideData>>`.)

cpu:
> At any rate, @cpu sorry for the unnecessary work here.

Zero apologies needed :-) It was next to no work and I could/should have thought deeper about it myself. Glad we're all on the same page now :+1:

PR #1278, opened 2023-04-11
Drop log messages on error paths

### LINKED ISSUE #1277: Add a way to disable “send fatal alert” log
Hi!

Currently rustls logs sending fatal alert on the warn level. In particular, bad certificate when trying to connect results in `Sending fatal alert: BadCertificate` log.

In our usecase bad certificates when connecting are OK and do not deserve a warn log. Is there any way to disable these logs in rustls or at least lower the severity?

djc:
Given that we now have very fine-grained error types and make pretty sure that error types are propagated up to the caller, I think it might make sense to reduce our logging on paths that will yield an error anyway, on the premise that the application may chose to do what to do with the error (including logging).

Fixes #1277.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -458,8 +458,10 @@ pub(super) fn process_alpn_protocol(
         // servers which accept a connection that requires an application-layer protocol they do not
         // understand.
         if common.is_quic() && common.alpn_protocol.is_none() && !config.alpn_protocols.is_empty() {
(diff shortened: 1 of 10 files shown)


cpu (approved):
Thanks Djc, this looks good to me. I think it's a nice demonstration of the value of the more specific errors to be able to drop these log lines.

ctz:
yeah i agree on this approach.

Issue #1275, opened 2023-04-07
Broken link in repo README.


VorpalBlade:
The link to supported platforms point to https://github.com/briansmith/ring#online-automated-testing

There is no such section in the ring readme, the link is broken.

cpu:
Good catch, thanks for opening an issue :+1: 

It looks like the upstream README dropped this content in https://github.com/briansmith/ring/commit/2047b68fa2aab521a57a445b0398760032226362. I don't see it having been migrated anywhere we can link to. Perhaps we should enumerate the supported platforms ourselves and keep it in-sync with upstream?

djc:
> It looks like the upstream README dropped this content in [briansmith/ring@2047b68](https://github.com/briansmith/ring/commit/2047b68fa2aab521a57a445b0398760032226362). I don't see it having been migrated anywhere we can link to. Perhaps we should enumerate the supported platforms ourselves and keep it in-sync with upstream?

Hmm, seems a bit of a pain in the ass, especially as we bring pluggable crypto providers in.

cpu:
> Hmm, seems a bit of a pain in the ass, especially as we bring pluggable crypto providers in.

Fair point.

Looking at the README again I think we already have enough information there to drop the link without adding more detail. It says:
>  At the time of writing this means x86, x86-64, armv7, and aarch64.

I'll put up a PR to remove the dead Ring link and we can try to keep this bit of text up to date. I don't think it will be a big burden since Ring development is pretty slow paced, pluggable crypto providers don't exist yet, and new arch target support doesn't happen with much frequency.

VorpalBlade:
I don't think that quite answers the question though. When it says "ARMv7", does it mean with hard float or not? What about musl vs glibc? What about ARM on Windows? Does any of that even matter?

Specific target triplets (or target triplet globs (e.g. `armv7-unknown-linux-*`) would be more useful.

VorpalBlade:
> The whole mess of trying to explain ARMv7 is exactly the kind of thing that we should probably avoid, since this project is not the limiting factor...

As a potential user who have to decide between using rustls or OpenSSL bindings for my project this is however very useful information. Both have downsides: OpenSSL supports most platforms and operating systems, but involves relying on the platform to provide it (or vendoring it., either statically or dynamically linked). Rustls has less platform support, but everything is linked statically and standalone.

To make an informed decision I need those complicated details.

djc:
[...]

PR #1279, opened 2023-04-13
docs: replace dead Ring link from README.

cpu:
Previously we linked to the *ring* README to describe Ring's supported architectures in more detail. Unfortunately that section of the upstream README was removed without a replacement.

This commit emphasizes that while Rustls is platform independent, *ring* is not. To replace the detailed platform support information we now link directly to the relevant *ring* CI configuration for the version in use by Rustls. 

Unfortunately there's no tag for *ring* 0.16.20 (see https://github.com/briansmith/ring/issues/1593) so I've used a permalink to the revision where the version number in `Cargo.toml` was bumped (https://github.com/briansmith/ring/commit/9cc0d45f4d8521f467bb3a621e74b1535e118188).

Resolves https://github.com/rustls/rustls/issues/1275

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,10 +129,12 @@ need them.
 
 ### Platform support
 
-Rustls uses [`ring`](https://crates.io/crates/ring) for implementing the
-cryptography in TLS. As a result, rustls only runs on platforms
-[supported by `ring`](https://github.com/briansmith/ring#online-automated-testing).
-At the time of writing this means x86, x86-64, armv7, and aarch64.
+While Rustls itself is platform independent it uses
+[`ring`](https://crates.io/crates/ring) for implementing the cryptography in
+TLS. As a result, rustls only runs on platforms
+supported by `ring`. At the time of writing this means x86, x86-64, armv7, and
+aarch64. For more information see [the supported `ring` CI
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,8 +131,8 @@ need them.
 
 Rustls uses [`ring`](https://crates.io/crates/ring) for implementing the
 cryptography in TLS. As a result, rustls only runs on platforms
-[supported by `ring`](https://github.com/briansmith/ring#online-automated-testing).
-At the time of writing this means x86, x86-64, armv7, and aarch64.
+supported by `ring`. At the time of writing this means x86, x86-64, armv7, and
```
VorpalBlade, line 134: That is too vague, as I wrote in https://github.com/rustls/rustls/issues/1275#issuecomment-1507370554, ARMv7 for example is incredibly vague (hard float? Linux, Windows or FreeBSD? Bare metal?). And Aarch64 is available in a big endian variant (apparently, TIL). You need a list of specific triplets that are supported, or at least globs of triplets.

VorpalBlade (changes requested):

README.md
cpu: I've added a link to the [Ring CI configuration](https://github.com/briansmith/ring/blob/9cc0d45f4d8521f467bb3a621e74b1535e118188/.github/workflows/ci.yml#L151-L167) where the supported triplets are listed.

djc (approved):

Issue #1281, opened 2023-04-18
MacOS: tests failing


mouse07410:
### Environment

- MacOS 13.3.1
- Xcode-14.3
- Macports-installed clang-16.0.1.
- Macports-installed OpenSSL-3.1.0

```text
$ rustup show
Default host: x86_64-apple-darwin
rustup home:  /Users/ur20980/.rustup

stable-x86_64-apple-darwin (default)
rustc 1.68.2 (9eb3afe9e 2023-03-27)
$ 
```

### To reproduce
1. Clone this repo into `rustls` and `cd rustls`
2. `cargo build`
3. `cargo test`
4. Observe the output

```text
     Running unittests src/lib.rs (target/debug/deps/rustls-8fb0713ad5a8e86f)

running 165 tests
test error::tests::time_error_mapping ... ok
test error::tests::rand_error_mapping ... ok
test error::tests::certificate_error_equality ... ok
.  .  .
test verifybench::test_wikipedia_cert ... ok
test x509::test_ludicrous ... ok

test result: ok. 165 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.41s

     Running tests/api.rs (target/debug/deps/api-47c586b28d6906b7)

running 122 tests
test all_suites_covered ... ok
test client_config_is_clone ... ok
.  .  .
test test_no_warning_logging_during_successful_sessions ... ok
test versions ... ok

test result: ok. 122 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s

     Running tests/bogo.rs (target/debug/deps/bogo-2faec40ce011e91b)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/client_cert_verifier.rs (target/debug/deps/client_cert_verifier-dd6f8f4f3477a5f1)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/key_log_file_env.rs (target/debug/deps/key_log_file_env-6b79f77d6324580e)

running 2 tests
test exercise_key_log_file_for_server ... ok
test exercise_key_log_file_for_client ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

[...]

PR #1282, opened 2023-04-21
examples: separate connectivity tests into crate.

cpu:
The `badssl.rs` and `topsites.rs` unit tests reach out to remote servers and so can be sources of flakes, slow down local testing, and may not operate correctly in the presence of HTTP(S) proxies.

This commit pulls those tests out of the `examples` crate and into a `connect-tests` crate that *isn't* part of our default workspace. This means running `cargo test --all-targets` and similar won't run these tests. Instead they must be explicitly run from within the `connect-tests` directory, or by specifying the crate's `Cargo.toml` as a `--manifest-path`.

To ensure we don't have bitrot/formatting errors we continue to lint the connectivity examples during our regular CI, but don't actually run the tests. Instead a separate cron-based workflow runs the tests once a week. Hopefully this will be a good balance between maintaining the coverage and not slowing down local tests, bumping into user HTTP(S) proxies, or hitting flakes.

Resolves https://github.com/rustls/rustls/issues/1281


examples/Cargo.toml
```diff
@@ -10,6 +10,8 @@ publish = false
 [features]
 dangerous_configuration = ["rustls/dangerous_configuration"]
 quic = ["rustls/quic"]
+# Run connectivity tests (default off)
+connect-tests = []
```
cpu, line 14: Oops, leftover from a feature-flag based attempt. Will remove.

cpu:
This was timely, it looks like CI on main [is currently failing](https://github.com/rustls/rustls/actions/runs/4773799177) due to BadSSL certificate expiry:

> We expected to find 'TLS error: InvalidCertificate\(NotValidForName\)' in the following output:
> Output { status: ExitStatus(unix_wait_status(256)), stdout: "TLS error: InvalidCertificate(Expired)\nConnection closed\n", stderr: "" }

ctz (approved):

Issue #1188, opened 2023-01-31
`rustls::Stream` EOF handling may not be appropriate


quininer:
> rustls Stream cannot distinguish between a normal close notify and an unexpected tcp eof, which allows truncation attack, which I suspect is not expected.

see https://github.com/tokio-rs/tls/pull/128

djc:
I've been wondering whether we should just remove `Stream`? At least for me it sort of feels like an afterthought since it seems to me that the audience is pretty narrow -- most people will be using something like tokio-rustls and for the majority of the remaining users, I feel like the `ConnectionCommon` API (including `complete_io()`) is probably a decent option.

ctz:
I think `Stream` should be tracking the behaviour introduced in #790; but it's unclear why that didn't happen. Needs some additional testing.

djc:
@cpu maybe this is something you can look into? Would be good to get this fixed.

cpu:
> @cpu maybe this is something you can look into?

Sure, I'll take a look.

cpu:
I was able to use the reproducers from https://github.com/tokio-rs/tls/pull/128 to make some progress here. Many thanks for including those! For convenience I created a Cargo workspace with both and all the required deps/feature flags to easily run locally: https://github.com/cpu/stream-eof-mrp

> I think Stream should be tracking the behaviour introduced in https://github.com/rustls/rustls/pull/790; but it's unclear why that didn't happen.

I think the crux of the issue is in this logic in `Stream.read` where `complete_io` and `process_new_packets` are called, trying to prematurely signal EOF: https://github.com/rustls/rustls/blob/da6aeb76858699293361caa4093235753d895181/rustls/src/stream.rs#L55-L73

As written, when the peer closes the connection prematurely this logic ends up returning `Ok(0)` (on L66) _before_ calling `conn.reader().read(buf)` (on L73). The `Reader.read` func is where #790 introduced logic to change how the EOF is handled based on whether the peer sent `close_notify` and so the `Stream.read` effectively side-steps the fix from #790.

I was able to fix the reproducer by updating the `Stream.read` logic to check the `io_state.peer_has_closed()` result, using that to help inform whether to prematurely signal EOF with `Ok(0)` (iff the peer has closed correctly) or `Err(io:ErrorKind::UnexpectedEof)` (iff the peer hasn't closed correctly). PTAL at  https://github.com/cpu/rustls/commit/1acb165b718ce44fab9712173b4dfdbdeda116fc

[...]

PR #1293, opened 2023-05-03
stream: remove premature EOF signal.

cpu:
Opening this as a draft because:

1. I'm not 100% confident this is the right fix. I'm not sure I understand the intent of the "premature signalling" that is being removed. It was added in https://github.com/rustls/rustls/commit/c94c223e4cabe49eb42b6e91ad307ff0813722e3 by @ctz.
2. I think it needs test coverage, but I'm not sure the best place to implement that (_and, selfishly I don't want to spend too much time on testing if my assumptions about the premature signalling are wrong_). ~Unlike most of the existing unit tests to be able to use `Stream` we need to bind sockets and do network I/O. Any hints here would be useful.~

## Description

Previously the `Stream` `read` and `read_buf` would:

* Call `complete_io` once.
* If the conn still `wants_read`, it would call `complete_io` in a loop.
  * If `complete_io` returned EOF, it would `process_new_packets` 
  * If there were no `plaintext_bytes_to_read`, the `Stream` would return `Ok(0)` to signal premature EOF. 
  * Otherwise it would break the loop.

This seems unnecessary, and obscures whether the EOF was "clean" or unexpected based on the peer notifying intent to close pre-EOF.

Instead,

* Call `complete_io` once.
* If the conn still `wants_read`, call `complete_io` in a loop.
  * If `complete_io` returns EOF, break the loop.
* Call `reader().read()` to potentially return read plaintext, return `Ok(0)` for a clean EOF, or return an error for premature EOF.

There's no need to call `process_new_packets` ourselves because `complete_io` does this after reading TLS bytes.

I believe there's no need to signal EOF prematurely if there's no plaintext bytes to read, because `reader().read()` handles this for us (and handles turning an unsignalled EOF into an err).

Resolves https://github.com/rustls/rustls/issues/1188

### DIFF
--- a/rustls/src/stream.rs
+++ b/rustls/src/stream.rs
@@ -55,17 +55,9 @@ where
         // We call complete_io() in a loop since a single call may read only
         // a partial packet from the underlying transport. A full packet is
         // needed to get more plaintext, which we must do if EOF has not been
-        // hit. Otherwise, we will prematurely signal EOF by returning 0. We
-        // determine if EOF has actually been hit by checking if 0 bytes were
-        // read from the underlying transport.
+        // hit.
         while self.conn.wants_read() {
-            let at_eof = self.conn.complete_io(self.sock)?.0 == 0;
-            if at_eof {
-                if let Ok(io_state) = self.conn.process_new_packets() {
-                    if at_eof && io_state.plaintext_bytes_to_read() == 0 {
-                        return Ok(0);
-                    }
-                }
+            if self.conn.complete_io(self.sock)?.0 == 0 {
                 break;
             }
         }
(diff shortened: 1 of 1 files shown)


djc:
> Unlike most of the existing unit tests to be able to use `Stream` we need to bind sockets and do network I/O. Any hints here would be useful.

Why do you need sockets? AFAICT we can make do with any `T: Read`.

cpu:
> Why do you need sockets? AFAICT we can make do with any T: Read.

You're right, I'm not sure how I confused myself into thinking otherwise :facepalm: That makes things easier.

cpu:
Going to close this until I have a chance to work on the unit tests. Probably in the next few days after I've finished up some webpki work.

cpu:
I reworked this with test coverage as https://github.com/rustls/rustls/pull/1299

PR #1295, opened 2023-05-12
Understanding PlaintextSink and Writer

xuganyu96:
Apologies in advance as this is less of an actual proposal for code change and more of a question, but I want to illustrate my point with some non-trivial amount of code that I find hard to fit into a GitHub issue, so I chose to open a pull request for this discussion.

I am new to Rust and to TLS and was reading through `rustls` source code to learn about these two subjects. One thing that I could not quite grasp is why we need the `Writer` struct and the `PlaintextSink` trait in the `conn.rs` module, when all they do is wrap around straightforward function calls like `ConnectionCommon::send_some_plaintext`.

Instead, I find it more straightforward to directly implement the `Write` trait for `ConnectionCommon`, and replace all calls to `conn.writer().write_xxx()` with `conn.write_xxx()` (where `write_xxx()` is some arbitrary write method provided by the `Write` trait, such as `write_all()`). I did the implementation and function call replacements, after which the project still compiles (with `cargo build`) and `cargo test` still passes.

I have not investigated the `read` side of this question. Could it be because there is something on the `read` side that requires a `reader()`? 

I am also not familiar with how `rustls` is used in production system. Could it be because some usage patterns that require a separate `Writer` struct instead of having `ConnectionCommon` implement `Write` directly? In any case, I understand that all this code change breaks public APIs and I don't expect them to be merged for now.

It would be great if you can provide some clarification regarding why the `Writer` struct and the `PlaintextSink` trait are there. Thank you very much for taking your time and helping me with my learning.

### DIFF
--- a/examples/src/bin/tlsclient-mio.rs
+++ b/examples/src/bin/tlsclient-mio.rs
@@ -63,7 +63,6 @@ impl TlsClient {
         let mut buf = Vec::new();
         let len = rd.read_to_end(&mut buf)?;
         self.tls_conn
-            .writer()
             .write_all(&buf)
             .unwrap();
         Ok(len)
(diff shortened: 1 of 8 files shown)


djc:
Please have a look at #629 and follow some of the links to other discussions from there. If you still have questions after that or want to verify your understanding, please follow up.

(This kind of stuff is pretty easy to find using `git blame` functionality, for example, in this case I looked at the `git blame` for the `PlaintextSink` trait on GitHub, clicked on a seemingly relevant commit -- Git commit pages on GitHub have a link to the PR they were part of.)

Issue #1200, opened 2023-02-23
benchmark ServerSessionMemoryCache


jsha:
The built-in [ServerSessionMemoryCache](https://docs.rs/rustls/latest/rustls/server/struct.ServerSessionMemoryCache.html) uses LimitedCache, which [has an O(n) operation for remove()](https://github.com/rustls/rustls/blob/a2fd30c05d8e4d999331af5c7cc4c4600ea60ba9/rustls/src/limited_cache.rs#L102) (note: the remove() codepath is exercised for TLS 1.3, not TLS 1.2).

It also [holds a mutex](https://github.com/rustls/rustls/blob/a2fd30c05d8e4d999331af5c7cc4c4600ea60ba9/rustls/src/server/handy.rs#L33) for all operations. When n is large, it's possible that this mutex is held long enough to cause contention problems.

It would be useful to benchmark ServerSessionMemoryCache under highly concurrent workloads for various values of `size`. If there's a value of `size` at which performance drops off badly, we should update the documentation for ServerSessionMemoryCache to recommend a maximum size.

jsha:
There's discussion at https://github.com/rustls/rustls/issues/469#issuecomment-812601706 about what we would want from a dependency that implements an LRU cache, if we find that performance merits one.

dhobsd:
N.B. if it does get to the point of analyzing different caches, LRU may not be the one to implement for a server cache since it is not scan-resistant. That might provide malicious clients with an opportunity to fill the cache and causing every lookup to have worst case performance.

aochagavia:
I've benchmarked this using a custom microbenchmark ([repo](https://github.com/aochagavia/rustls-cache-bench)) set up as follows:

- Create an `Arc<ServerSessionMemoryCache>` and share it among 8 worker threads.
- Concurrently send "handshake" requests to the worker threads (not a real handshake, just a call to generate and store session ids, and an additional call to `ServerSessionMemoryCache::take` for resumed handshakes).
- For each request, measure the time spent waiting (on the client side) and the time spent doing the actual work (on the server side). This reveals how much time is spent on synchronization overhead. The result of `server_duration / client_duration` is called saturation in the graphs.

## Results

Below I have tested a few scenarios with different cache sizes.

### 20k full + 40k resumed handshakes

![Untitled](https://github.com/rustls/rustls/assets/5196584/e84a2fb8-dae0-40fb-9ccc-a0fdff5ebd67)

### 30k full + 30k resumed handshakes

[...]

PR #1300, opened 2023-05-20
perf: avoid clones

kamadorueda:
LimitedCache is used in the codebase with `Vec<u8>` and `ServerName` as a key, whose clone() and memory layout are less efficient than that of a u64. By dropping the requirements of traits like Clone and Default, it may also find utility in other places of the codebase down the road

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -16,7 +16,6 @@ use rustls::{
 };
 
 use base64::prelude::{Engine, BASE64_STANDARD};
-use env_logger;
 
 use std::io::{self, BufReader, Read, Write};
 use std::sync::Arc;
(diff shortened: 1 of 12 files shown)


djc:
In general, I don't think our current version of `LimitedCache` is intended to be very high-performance. See previous discussion in https://github.com/rustls/rustls/issues/1200 and the linked issues. As such, while I'm happy to take simple improvements like replacing `cloned()` with `copied()` where possible (although I think these changes are unlikely to have much of an actual performance impact), I don't think it's a good idea to dismantle the structure of `LimitedCache` without a better understanding of use cases where people are running into issues and relevant benchmarks. As far as I remember, `LimitedCache` is, in its current state, more of a proof of concept and a starting point than a mature, optimal cache implementation.

Do you have an actual use case where the rustls-provided `LimitedCache` implementation is a limiting factor? Would be good to hear more about what you're trying to do and what you would like to have.

kamadorueda:
> Do you have an actual use case where the rustls-provided LimitedCache implementation is a limiting factor?

Hey! First of all, thanks for the reply

In reality no, I don't have a use case for this. I watched this video recently: https://youtu.be/YvsGfGDGQMs?t=1540 and given that I like the philosophy of the library and I had a couple of hours of free time I tried to find a way to contribute while familiarizing myself with the codebase. 

I'll close this PR, but happy to hear if there is any beginner-friendly ticket that the maintainers would like me to help with. Some ideas that came to my mind were enabling more clippy lints and/or increasing coverage

PR #1301, opened 2023-05-22
expand documentation for Certificate and PrivateKey

folkertdev:
Some of the rustls documentation is... not that helpful to those unfamiliar with TLS and certificates. The `Certificate` and `PrivateKey` types had particularly little documentation. This is my attempt to improve the situation.

I've included code samples using `rustls_pemfile`. tried to give some info to quickly check "does this file I have match roughly what rustls would expect" and provide some links to more info and the `rcgen` crate, if a user wants to generate a new certificate or private key.

### DIFF
--- a/rustls/src/key.rs
+++ b/rustls/src/key.rs
@@ -62,14 +62,6 @@ pub struct PrivateKey(pub Vec<u8>);
 /// to parse PEM files. The [`rcgen`](https://docs.rs/rcgen/latest/rcgen/) crate can be used to
 /// generate certificates and private keys.
 ///
-/// ## Note
-///
-/// If you are receiving certificates from an untrusted client or server, it is essential
-/// to validate the contents of the certificate manually. This typically involves verifying
-/// the certificate's signature, checking its expiration date, verifying the certificate
-/// chain (including intermediate and root certificates), and performing additional checks
-/// based on your specific requirements.
-///
 /// ## Examples
 ///
 /// Parsing a PEM file to extract DER-encoded certificates:


rustls/src/key.rs
```diff
@@ -5,21 +5,88 @@ use std::fmt;
...
+///
+/// The [`rustls-pemfile`](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/) crate can be used
+/// to parse PEM files. The [`rcgen`](https://docs.rs/rcgen/latest/rcgen/) crate can be used to
+/// generate certificates and private keys.
 ///
 /// ## Note
 ///
-/// If you are receiving certificates from an untrusted client or server, the contents
-/// must be validated manually.
+/// If you are receiving certificates from an untrusted client or server, it is essential
```
djc, line 67: I'm not sure that this paragraph was trying to say exactly, but I'm not sure the expansion here makes all that much sense since rustls will typically check these for you?

djc (approved):
Thanks, these are great improvements!

rustls/src/key.rs
```diff
@@ -5,21 +5,88 @@ use std::fmt;
...
 ///
-/// The `rustls-pemfile` crate can be used to extract
-/// private keys from a PEM file in these formats.
+/// A common format for storing private keys is
+/// [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail).
+/// PEM private keys are commonly stored in files with a `.pem` or `.key` suffix, and look like this:
+///
+/// ```txt
+/// -----BEGIN PRIVATE KEY-----
+///
```
cpu, line 14: tiny nit: maybe there should be some indication there's content between the header/footer?
```suggestion
/// <encoded private key content>
```

rustls/src/key.rs
```diff
@@ -5,21 +5,88 @@ use std::fmt;
...
 ///
-/// The `rustls-pemfile` crate can be used to parse a PEM file.
+/// A common format for storing certificates is
+/// [PEM](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail).
+/// PEM certificates are commonly stored in files with a `.pem`, `.cer` or `.crt` suffix, and look
+/// like this:
+///
+/// ```txt
+/// -----BEGIN CERTIFICATE-----
+///
```
cpu, line 57: ditto here if you agree with my earlier nit:
```suggestion
/// <encoded certificate content>
```

cpu (approved):
Thank you! I think this will be helpful.

djc:
(Going to work around the MSRV failure with #1302, feel free to ignore.)

rustls/src/key.rs
djc: I think this is a good idea, might be even better to be explicit about it being base64-encoded.

rustls/src/key.rs
folkertdev: just remove the note completely then? if it doesn't add anything it's only confusing

djc: This was introduced in #1056, I've asked the original author to weigh in.

complexspaces: I don't have an issue with removing these lines. I think that I may have worded this with too much of a bias for people implementing their own certificate verification.

The original goal was to make it more obvious that you shouldn't `.unwrap()` on any parsing methods, immediately read anything out of the parsed certificate, etc when handling these yourselves due to mistaking thinking `rustls` had pre-validated them.

djc (approved):

PR #1302, opened 2023-05-22
Only check library against MSRV

djc:
To avoid unnecessary MSRV bumps that are required for dev-dependencies or the examples project, only check the MSRV for the rustls library target.

We've recently started using this in Quinn, and I think it's a pretty nice improvement to avoid unnecessary MSRV bumps.


cpu (approved):
This seems reasonable to me. Thanks @djc :+1:

PR #1303, opened 2023-05-22
Upgrade env_logger to 0.10

djc:
Just a dev-dependency, so it no longer affects our MSRV (since #1302).


djc:
Ahhh, yes, so this has the issue that it's not just a library MSRV issue, but the `Cargo.toml` contents (which get parsed at some point earlier in the dependency resolution process) being invalid before Cargo 1.60 (due to the use of weak or namespaced features).

Issue #1304, opened 2023-05-23
`Iterator` support for `RootCertStore::add_parsable_certificates()`


daxpedda:
I've noticed that a couple of implementations out there, when trying to use [`rustls_native_certs::load_native_certs()`](https://docs.rs/rustls-native-certs/0.6.2/rustls_native_certs/fn.load_native_certs.html), write their own [`RootCertStore::add_parsable_certificates()`](https://docs.rs/rustls/0.21.1/rustls/struct.RootCertStore.html#method.add_parsable_certificates) because `rustls_native_certs::load_native_certs()` doesn't return `[Vec<u8>]` but rather `[Certificate]`. See [`reqwest` for example](https://github.com/seanmonstar/reqwest/blob/06c8e5b0b008afee8114fb979b85cd8b73415391/src/async_impl/client.rs#L399-L424).

I think this could be easily addressed by improving the API to take an `Iterator` instead of a `&[Vec<u8>]`. Though this is a breaking change, so I will leave this here to be addressed in the next version.

djc:
Maybe we should support any `impl AsRef<[u8]>`?

daxpedda:
Indeed, I just realized that both [`RootCertStore::add()`](https://docs.rs/rustls/0.21.1/rustls/struct.RootCertStore.html#method.add) and [`RootCertStore::add_parsable_certificates()`](https://docs.rs/rustls/0.21.1/rustls/struct.RootCertStore.html#method.add_parsable_certificates) don't actually need ownership of the bytes.

It might also be good to adjust the API of `RootCertStore::add()` to take `AsRef<[u8]>` as well then.

djc:
We can probably change `add_parsable_certificates()` to take `&[impl AsRef<[u8]>]` instead of `&[Vec<u8>]` today, since `Vec<u8>` implements `AsRef<u8>`. Maybe restricting `add()` to the more strongly typed `key::Certificate` might still have some utility?

daxpedda:
> We can probably change `add_parsable_certificates()` to take `&[impl AsRef<[u8]>]` instead of `&[Vec<u8>]` today, since `Vec<u8>` implements `AsRef<u8>`.

Sure, I can make a PR for that.

> Maybe restricting `add()` to the more strongly typed `key::Certificate` might still have some utility?

[...]

PR #1262, opened 2023-03-29
Prepare 0.21.0 release

djc:
Documents visible changes since the previous changelog update in #1187. Does not yet include changes from open PRs.

PRs left to consider:

- [x] #1251 
- [x] #1263

Release process steps:

- [x] Update changelog to accomodate above PRs
- [x] Dependencies are up to date (except the env_logger dev-dependency, for MSRV reasons)
- [x] `cargo test --all-features` passes locally (with latest dependencies)
- [x] Version updated in `rustls/Cargo.toml`
- [x] `cargo publish --dry-run`
- [ ] Get CI passing on this after rebasing on top of merged PRs
- [ ] Tag && push tag
- [ ] `cargo publish`

PR #1305, opened 2023-05-23
Take `&[u8]` in `RootCertStore::add_parsable_certificates()`

daxpedda:
This changes [`RootCertStore::add_parsable_certificates()`](https://docs.rs/rustls/0.21.1/rustls/struct.RootCertStore.html#method.add_parsable_certificates) to take `&[impl AsRef<[u8]>]` instead of `&[Vec<u8>]`.

Partially addresses #1304.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -105,15 +105,7 @@ impl RootCertStore {
     /// in order to add as many valid roots as possible and to understand how many certificates
     /// have been diagnosed as malformed.
     pub fn add(&mut self, der: &key::Certificate) -> Result<(), Error> {
-        let ta = webpki::TrustAnchor::try_from_cert_der(&der.0)
-            .map_err(|_| Error::InvalidCertificate(CertificateError::BadEncoding))?;
-        let ota = OwnedTrustAnchor::from_subject_spki_name_constraints(
-            ta.subject,
-            ta.spki,
-            ta.name_constraints,
-        );
-        self.roots.push(ota);
-        Ok(())
+        self.add_internal(&der.0)
     }
 
     /// Adds all the given TrustAnchors `anchors`.  This does not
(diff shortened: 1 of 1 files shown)


rustls/src/anchors.rs
```diff
@@ -95,6 +95,18 @@ impl RootCertStore {
...
+        let ta = webpki::TrustAnchor::try_from_cert_der(der)
+            .map_err(|_| Error::InvalidCertificate(CertificateError::BadEncoding))?;
+        let ota = OwnedTrustAnchor::from_subject_spki_name_constraints(
+            ta.subject,
+            ta.spki,
+            ta.name_constraints,
+        );
+        self.roots.push(ota);
+        Ok(())
+    }
```
daxpedda, line 108: I extracted the code of `RootCertStore::add()` here to not have to duplicate it in `RootCertStore::add_parsable_certificates()`. I couldn't use `RootCertStore::add()` directly because it requires owned bytes.

Let me know if you want me to handle it differently.

djc: I think extracting this makes sense. I'd like to have this added below the other relevant methods rather than before them. Maybe inline the `ota` binding?

djc (approved):
Nice!

rustls/src/anchors.rs
daxpedda: Done.

daxpedda:
Uhm, I think the code coverage is fine here, right?

I can't really make sense of it:
- master: https://app.codecov.io/gh/rustls/rustls/blob/main/rustls/src/anchors.rs
- pr: https://app.codecov.io/gh/rustls/rustls/pull/1305/blob/rustls/src/anchors.rs

Like code coverage should have increased even.

cpu (approved):
Thanks! Looks good to me :-)

daxpedda:
@djc do you think we could get a new release with this?
(apologies for pinging you all over the place ...)

djc:
Want to send PRs that bump the version number and fill out changelog if necessary both here and in rustls-native-certs? (Bonus points if you put in a bit of a check list following the `RELEASING.mid`, as in https://github.com/rustls/rustls/pull/1262.)

daxpedda:
Sure thing.

PR #1306, opened 2023-05-24
tests: start on de-duplicating stream API tests.

cpu:
This branch is a follow-up from https://github.com/rustls/rustls/pull/1299 and tries to remove some duplication in the API tests related to the `Stream` and `StreamOwned` types. It doesn't get 100% of the way there, but I think represents an improvement over the status quo.

We want coverage for reading/writing for both the owned and reference types. We also want coverage for reading unexpected EOF errors, and for reading with and without the `read_buf` feature. This branch is able to remove most of the duplication for testing these scenarios, but still leaves some for the server and client contexts.

The most obvious next step is to try and combine `test_client_stream_read` and `test_server_stream_read` as well as `test_client_stream_write` and `test_server_stream_write` - in practice these sets of functions are identical except for which side of the connection is used where. I started down this road in https://github.com/rustls/rustls/commit/e24881615ca1cdc2723ec539c29b9d825eeca9dd but ran into challenges finding a way to represent the `pipe` in a way that would allow the usage scenarios we have in mind and wouldn't run into lifetime issues. I've spent an unfortunate amount of time trying to figure out a path forward but have to admit defeat to switch to other tasks. I figured the cleanup I have is worthwhile as-is, and hopefully by sharing my progress someone else might be able to spot an approach for the last bit that I missed :-)

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -1546,268 +1546,133 @@ fn server_complete_io_for_read() {
 
 #[test]
 fn client_stream_write() {
-    for kt in ALL_KEY_TYPES.iter() {
-        let (mut client, mut server) = make_pair(*kt);
-
-        {
-            let mut pipe = OtherSession::new(&mut server);
-            let mut stream = Stream::new(&mut client, &mut pipe);
-            assert_eq!(stream.write(b"hello").unwrap(), 5);
-        }
-        check_read(&mut server.reader(), b"hello");
-    }
+    test_client_stream_write(StreamKind::Ref);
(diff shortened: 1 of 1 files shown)


djc (approved):
Would prefer to standardize on moving helper functions below the relevant tests instead of above them, but this definitely looks like a nice improvement!

cpu:
> Would prefer to standardize on moving helper functions below the relevant tests instead of above them

Done! Thanks for the review. Sorry about missing that convention across a couple PRs. I'll make a note of it for the future.

PR #1309, opened 2023-06-04
Added a resolver fetch callback and wouldblock handling

john-sharratt:
This patch implements a feature needed for ACME - the problem being that some certificates need to be fetched from IO engines (e.g. S3 buckets) or generated via proof of ownership challenges, however `rustls` uses blocking logic on the certificate resolve method and hence would lock up the thread. This is especially apparent in `tokio-rustls` which such blocking code would also freeze up the `tokio` shared thread pool.

The motivations for the choices in this patch are:
- minimize the impact to the external API, with forwards compatibility.
- do not add an async runtime to `rustls` itself
- minimize the amount of code changes as much as possible but while still giving the needed functionality.

This patch needs to arrive along with this one
https://github.com/tokio-rs/tls/pull/147

Before I update the references this patch needs to make upstream

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -4,7 +4,7 @@ use crate::check::inappropriate_handshake_message;
 use crate::common_state::{CommonState, State};
 use crate::conn::ConnectionRandoms;
 use crate::enums::{AlertDescription, CipherSuite, ContentType, HandshakeType, ProtocolVersion};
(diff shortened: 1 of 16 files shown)


ctz:
Thanks for the patch! I think we already addressed this use-case with the `Acceptor` API: https://docs.rs/rustls/latest/rustls/server/struct.Acceptor.html

john-sharratt:
Excellent, thanks for the help. Closed

PR #1312, opened 2023-06-12
Move MSRV to 1.60

ctz:
Doing this to fix the build after `log` changed to 1.60 in a minor release. Have also incorporated #1303


djc (approved):
Nice. Are you going to take care of downstream dependencies, too?

ctz:
Yup!

PR #1314, opened 2023-06-12
Take latest bogo test suite

### LINKED ISSUE #1158: Add support for latest BoGo updates
current output, running against unpatched google master (396625d50212143bc03517c9fdd21b1c965b3f45): https://gist.github.com/jbr/8d2a0d96719132f06666ff47544d8c89

it looks like there are unsupported flags on the shim as well as reported failures

ctz:
resolves #1158


rustls/src/error.rs
```diff
@@ -225,6 +225,7 @@ pub enum PeerMisbehaved {
     UnsolicitedSctList,
     UnsolicitedServerHelloExtension,
     WrongGroupForKeyShare,
+    MissingPskModesExtension,
```
djc, line 228: Keep this in alphabetic order?

djc (approved):

bogo/config.json
```diff
@@ -33,6 +33,8 @@
     "DuplicateCertCompressionExt*": "",
     "TLS-ECH-*": "",
     "ALPS-*": "",
+    "Kyber*": "",
+    "*Kyber*": "",
```
cpu, line 37: Do we need both of these? Wouldn't `*Kyber*` match the same things as `Kyber*` or am I missing a detail?

rustls/src/error.rs
```diff
@@ -225,7 +226,6 @@ pub enum PeerMisbehaved {
     UnsolicitedSctList,
     UnsolicitedServerHelloExtension,
     WrongGroupForKeyShare,
-    MissingPskModesExtension,
```
cpu: Rebase artifact? The commit leads with `fixup!` and doesn't appear to be doing what it says on the tin.

cpu: Aha, already caught. I was reviewing the stale branch :-)

bogo/fetch-and-build
```diff
@@ -15,7 +15,7 @@ util/testresult
 EOF
 
 # fix on a tested point of rustls-testing branch
-COMMIT=3fea50a7cb55bb72e4a69d34015f1a6971e7934c
+COMMIT=b81c2d26009b6cc5159b7a9cbdddfa4ae79defea
```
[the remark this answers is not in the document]
ctz: This is now the tip of the `rustls-testing` branch, which has two commits additional to upstream.

bogo/config.json
ctz: You're completely right -- will clean this up.

bogo/fetch-and-build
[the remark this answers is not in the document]
cpu: Understood, thanks :)

PR #1315, opened 2023-06-13
Limit scope of security support based on time

djc:
So far, we've promised support for the most recent 3 minor release lines. I think we should scope support based on time, not just the number of releases, because the number of releases can change based on our development pace.

Add an example to demonstrate the intended mechanism. (Might be overkill?)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,7 +2,16 @@
 
 ## Supported Versions
 
-Security fixes will be backported to the most recent three minor version lines.
+Security fixes will be backported only to the rustls versions for which the
+original semver-compatible release was published less than 2 years ago.
+
+For example, as of 2023-06-13 the latest release is 0.21.1.
+
+* 0.21.0 was released in March of 2023
+* 0.20.0 was released in September of 2021
+* 0.19.0 was released in November of 2020
+
+Therefore 0.20.x and 0.21.x will be updated, while 0.19.x will not be.
 
 ## Reporting a Vulnerability
 


ctz (approved):
I think it's still very generous, but judging by graphs on crates.io and current debian packaging there is no use of 0.19 now so this policy covers ~all current usage but no more than that.

djc:
Thinking about this a bit more, there are some tricky edge cases to the formulation I'm proposing here. For example, if we restricted the scope to 18 months instead of 2 years, then the 0.20.0 release falls out -- in general, the proposed formulation doesn't really "care" when the last release branch was made, which can lead to strange results. The obvious alternative would be to base this on the last release made on a particular branch (along the lines of, 1 year after the last release made from a semver branch), but that would make it recursive.

I feel like most of the ecosystem can absorb our releases in 6-9 months, but some parts (like sqlx today, to pick just one example) can be slower.

cpu (approved):
This seems good to me, and better specified than the previous text. 

I think for corner cases it might be reasonable to consider them on a case-by-case basis? I think it's reasonable to consider these guidelines instead of hard rules and depending on the uptake of different versions, the severity of the security issue, and the bandwidth of the maintainers we could make exceptions (?).

djc:
> I think for corner cases it might be reasonable to consider them on a case-by-case basis? I think it's reasonable to consider these guidelines instead of hard rules and depending on the uptake of different versions, the severity of the security issue, and the bandwidth of the maintainers we could make exceptions (?).

I mean, we can definitely extend extra support on the basis of exceptions. I feel like we should probably provide some clarity on what we promise to provide support for in the obvious cases. So maybe a release branch will always be supported for at least 12 months after the release of the next branch?

djc:
> I suppose we should decide on this before merging [rustls/webpki#78](https://github.com/rustls/webpki/pull/78) so I can keep it in-sync?

Yeah, that might make sense? In the grand scheme of things, probably doesn't matter much, we can just get that merged while we figure this out.

PR #1317, opened 2023-06-14
Fix typo in `ConfigBuilder::with_root_certificates()`

daxpedda:
Just stumbled on that. Pretty sure root certificates for `ClientConfig` are used to verify server certificates.


djc (approved):

djc:
Thanks!

PR #1318, opened 2023-06-14
Prepare 0.21.2

daxpedda:
- [x] Update changelog.
- [x] `cargo update`.
- [x] `cargo outdated`.
- [x] `cargo test --all-features`.
- [x] Update `version` in `Cargo.toml`.
- [x] `cargo publish --dry-run`
 
Happy to take any advice on the changelog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,11 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ## Release history
 
-* Next release:
+* Release 0.21.2 (2023-06-14)
   - Bump MSRV to 1.60 to track similar change in dependencies.
+  - Differentiate between unexpected and expected EOF in `Stream` and `OwnedStream`.
+  - `RootCertStore::add_parsable_certificates` now takes a `&[impl AsRef<[u8]>]`.
+  - Add QUIC V2 support.
 * Release 0.21.1 (2023-05-01)
   - Remove `warn`-level logging from code paths that also return a `rustls::Error` with
     the same information.
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls"
-version = "0.21.1"
+version = "0.21.2"
 edition = "2021"
 rust-version = "1.60"
 license = "Apache-2.0 OR ISC OR MIT"


ctz (approved):
Thanks, this looks good to me.

Just for my understanding:

>  `RootCertStore::add_parsable_certificates` now takes a `&[impl AsRef<[u8]>]`.

This isn't a breaking change, because previously it took `&[Vec<u8>]` and [generalisation of function arguments is allowed](https://doc.rust-lang.org/cargo/reference/semver.html#minor-generalizing-a-function-to-use-generics-supporting-original-type).

djc:
> Just for my understanding:
> 
> > `RootCertStore::add_parsable_certificates` now takes a `&[impl AsRef<[u8]>]`.
> 
> This isn't a breaking change, because previously it took `&[Vec<u8>]` and [generalisation of function arguments is allowed](https://doc.rust-lang.org/cargo/reference/semver.html#minor-generalizing-a-function-to-use-generics-supporting-original-type).

Yup, exactly.

README.md
```diff
@@ -20,8 +20,14 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ## Release history
 
-* Next release:
+* Release 0.21.2 (2023-06-14)
   - Bump MSRV to 1.60 to track similar change in dependencies.
+  - Differentiate between unexpected and expected EOF in `Stream` and `OwnedStream`.
+  - `RootCertStore::add_parsable_certificates` now takes a `&[impl AsRef<[u8]>]`.
+  - Add QUIC V2 support.
+  - Upgrade `env_logger` to v0.10.
```
djc, line 28: IIRC env_logger is a dev-dep only, so probably doesn't need to go in the changelog.

README.md
```diff
@@ -20,8 +20,14 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
...
+  - Correctly provide an empty session ID if none is available for TLS 1.2.
+  - Bug fix: abort clients offering "pre_shared_key" without a "psk_key_exchange_modes" extension.
```
djc, line 30: Historically we haven't really made node of minor bugfixes like this in the changelog (I would probably only keep the QUIC V2 support and the `Stream` EOF handling), @ctz do you feel we should change that?

README.md
daxpedda: Ah, right, my bad.

README.md
ctz: Yeah agree that it's a minor bug fix (and never going to be seen in the wild). I probably wouldn't put it in a change log (cos it's not interesting or actionable for users).

djc (approved):

djc:
Published and pushed tag.

PR #1319, opened 2023-06-16
Extract traits under crypto use

ctz:


rustls/src/client/tls12.rs
```diff
@@ -31,7 +31,7 @@ use crate::client::common::ClientAuthDetails;
 use crate::client::common::ServerCertDetails;
 use crate::client::{hs, ClientConfig, ServerName};
 
-use ring::constant_time;
+use subtle;
```
djc, line 34: I'm on board with this, but for this PR I feel like we shouldn't keep the commit + revert in there.

rustls/Cargo.toml
```diff
@@ -20,6 +20,7 @@ log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 sct = "0.7.0"
 webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
+subtle = "2.5.0"
```
djc, line 23: Style nit: alphabetic order?

rustls/src/client/tls12.rs
```diff
@@ -1047,12 +1047,13 @@ impl<C: CryptoProvider> State<ClientConnectionData> for ExpectFinished<C> {
         // Constant-time verification of this is relatively unimportant: they only
         // get one chance.  But it can't hurt.
         let _fin_verified =
-            constant_time::verify_slices_are_equal(&expect_verify_data, &finished.0)
-                .map_err(|_| {
-                    cx.common
-                        .send_fatal_alert(AlertDescription::DecryptError, Error::DecryptError)
-                })
-                .map(|_| verify::FinishedMessageVerified::assertion())?;
+            if subtle::ConstantTimeEq::ct_eq(&expect_verify_data[..], &finished.0).into() {
```
djc, line 1050: I'd import `ConstantTimeEq` directly, and probably write this as a `match` expression. Or better, maybe this is a helper wrapper around `CommonState::send_fatal_alert()`?

djc (commented):
(Early review, only looked at the subtle stuff so far.)

ctz:
So this is a good mid-point: all webpki/ring dependencies now are restricted to `crypto::ring`. Everything outside that is -- should be -- completely agnostic of crypto library.

The next step is to see if this works in practice by extracting all the agnostic parts into their own crate. The remaining `rustls` crate would continue to take a webpki/ring dependency and have the same public API as it currently does, which would be a combination of the concrete parts it provides (cipher suites, key exchange groups, certificate verifier, etc.) and a reexport of the crypto-agnostic public API.

```


                        ┌─────────────────┐                  ┌────────────────┐
                        │                 │                  │                │
                        │   rustls-core   │                  │  ring          │
                        │                 │                  │                │
                        └─▲─────────────▲─┘                  └─▲─┬────────────┤
                          │             │                      │ │            │
    crypto-agnostic APIs  │             │  provider APIs       │ │  webpki    │
                          │             │                      │ │            │
                        ┌─┼─────────────┴─┐                    │ └─▲──────────┘
                        │ │               ├────────────────────┘   │
                        │ │ rustls        │                        │
                        │ │               ├────────────────────────┘
                        └─┼─────────────▲─┘
                          │             │
                          └──────▲──────┘
                                 │
                                 │  current rustls API


```

Following that, I would plan to make a small (one cipher suite, one kx, etc.) crate against some of the rustcrypto ecosystem to see how that goes and use as an exercise to write some documentation.

djc:
> The next step is to see if this works in practice by extracting all the agnostic parts into their own crate. The remaining `rustls` crate would continue to take a webpki/ring dependency and have the same public API as it currently does, which would be a combination of the concrete parts it provides (cipher suites, key exchange groups, certificate verifier, etc.) and a reexport of the crypto-agnostic public API.

I forget if we managed to drive this discussion to a conclusion before, but why is splitting into two crates better than using a Cargo feature?

djc:
Have you run any benchmarks to see if/how indirection influences the results?

djc:
> Using a crate feature means the single crate (and its maintainers) ends up knowing about and testing all the possible providers. That I think is OK if there are two or three.

In my mind we would keep the *ring* one in-tree for now, and others would live outside the crate.

djc:
> > That's aside from the general issue with using features to switch between several options: for a big enough application probably several providers would get compiled in.
> 
> With that said, that is also a property of splitting things over several crates.

Yes, I don't think this would be a difference between the crate and Cargo feature approach. I feel like a Cargo feature gets us most of the way there for now and is a smaller change. We can decide separately if the *ring* backend should be part of the default features (to make it easier to build dependency graphs that don't include it, I suppose, though I'm honestly not sure if it makes a big difference) or if that should be opt-in. Do we want to make it in an explicit choice in the `ConfigBuilder` API?

djc:
> Yes, though it feels pretty big and might be better landed in several pieces?

Might make sense. IIRC @cpu and I are already familiar with the prefix that we worked on, so unless you changed a bunch of stuff from that part maybe it could make sense to just start by landing that?

ctz:
> so unless you changed a bunch of stuff from that part maybe it could make sense to just start by landing that?

Apart from rebasing I didn't change any of that meaningfully. The exception is 
[Use subtle::ConstantTimeEq instead of ring::constant_time](https://github.com/rustls/rustls/pull/1319/commits/e27653a038a36234912db5caa89c195499afb383) which replaced one of those commits.

> Do we want to make it in an explicit choice in the `ConfigBuilder` API?

Well at the moment it is an explicit choice in the sense that each `rustls::ClientConfig::builder()` has to become `rustls::ClientConfig::<Ring>::builder()`. I am wondering if -- with a default feature enabling ring -- we could avoid that by defaulting the type parameter?

djc:
> > Do we want to make it in an explicit choice in the `ConfigBuilder` API?
> 
> Well at the moment it is an explicit choice in the sense that each `rustls::ClientConfig::builder()` has to become `rustls::ClientConfig::<Ring>::builder()`. I am wondering if -- with a default feature enabling ring -- we could avoid that by defaulting the type parameter?

I think that would be tricky in practice if you want to have the compilation of the *ring* backend to be conditional on a Cargo feature. That would make the type default conditional as well, which I don't think would be easy to express... and honestly, given the nature of Cargo features (which can be enabled by indirect dependencies) I think for a security-salient choice like this, keeping the choice explicit is probably the right call here.

djc:
> > so unless you changed a bunch of stuff from that part maybe it could make sense to just start by landing that?
> 
> Apart from rebasing I didn't change any of that meaningfully. The exception is [Use subtle::ConstantTimeEq instead of ring::constant_time](https://github.com/rustls/rustls/pull/1319/commits/e27653a038a36234912db5caa89c195499afb383) which replaced one of those commits.

Not sure what this is waiting for, but to be explicit: I think either including or excluding that change in part 1 is fine.

ctz:
> Not sure what this is waiting for,

I think we'll probably want to do a release for the CRL stuff before landing any of this?

Aside from that I've been trying to get a clear idea whether this regresses performance. Unfortunately even taken on a commit-by-commit basis there is significant noise (4-5%) in the figures produced by our benchmarks. No conclusions yet.

djc:
I haven't dug in in detail yet, but one high-level question: how hard would it be to avoid the temporary disabling of QUIC in the commit history, now that we know where things are going? IMO it feels kinda ugly letting that go into the commit history.

ctz:
> I haven't dug in in detail yet, but one high-level question: how hard would it be to avoid the temporary disabling of QUIC in the commit history, now that we know where things are going?

Yeah I think that is doable, by just some more careful commit ordering/separation.

> And another question: is there another useful/valuable split here, or do we want to get this in in one go? 

Will hold that in mind -- aside from the TODOs in the final commit message, I also want to get webpki involved in that example before thinking this branch is reaching its goals. Then we can split it up further (eg, perhaps everything up to introducing the new default feature? No new external functionality happens before that one.)

djc:
> > I haven't dug in in detail yet, but one high-level question: how hard would it be to avoid the temporary disabling of QUIC in the commit history, now that we know where things are going?
> 
> Yeah I think that is doable, by just some more careful commit ordering/separation.

Would be great!

> > And another question: is there another useful/valuable split here, or do we want to get this in in one go?
> 
> Will hold that in mind -- aside from the TODOs in the final commit message, I also want to get webpki involved in that example before thinking this branch is reaching its goals. Then we can split it up further (eg, perhaps everything up to introducing the new default feature? No new external functionality happens before that one.)

Going up to the default feature seems like a large amount of stuff to digest, at least for me personally it might be nice to split it up in smaller chunks than that. To me, the heuristic would be if you're pretty sure that the design is right and set in stone for some prefix of the commits, then we can peel that off and start reviewing.

If this is not ready yet, perhaps revert it to Draft status?

ctz:
OK, I think this is ready to go now. I think I will try to arrange the commits here it in three sequential PRs:

1. changes to bulk encryption, quic and secret_extraction interfaces
1. adding an interface for hashing and hmac
1. making ring and webpki optional, plus example code

(1) will be coming up shortly

ctz:
> (1) will be coming up shortly

Erm, ok, that is actually more annoying than expected, because I'd have to write more throw-away glue code for ring's hkdf. Let's see how (1) and (2) together looks.

djc:
@ctz I think this can be closed?

cpu:
>  @ctz I think this can be closed?

I think so too. I'm going to do it and we can reopen if the need arrives.

PR #1320, opened 2023-06-20
add option for skipping domain-name verification of servers

Fritiofhedstrom:
I am developing a communication network using [Zenoh](https://github.com/eclipse-zenoh/zenoh) with a mTLS feature. Each node specifies a trusted CA cert that is used to verify both connecting clients as well as servers it wants to connect to. The CA is considered the network certificate so any cert signed by it should be fully trusted to partake. This means that i don't care about verifying the domain-name for servers, they could be accepting connections on any interface and the IpAdress could change.

Zenoh depends on rustls for tls communication. It is possible to implement this functionality without updates to rustls by implementing the ServerCertVerifier trait, however I think that this would be a common use case that rustls could fulfill. Also there is a lot of private code for WebPkiVerifier that i would like to reuse that would be duplicated if this would be done in the Zenoh codebase.

Here is the issue @ Zenoh [https://github.com/eclipse-zenoh/zenoh/issues/513](https://github.com/eclipse-zenoh/zenoh/issues/513)

I am completely new to this repository so please help me pass all the required checks! 
I suppose you would want me to write tests for this?

### DIFF
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -416,7 +416,7 @@ pub mod client {
     #[cfg(feature = "dangerous_configuration")]
     pub use crate::verify::{
         CertificateTransparencyPolicy, HandshakeSignatureValid, ServerCertVerified,
-        ServerCertVerifier, WebPkiVerifier,
+        ServerCertVerifier, WebPkiVerifier, WebPkiVerifierAnyServerName,
     };
     #[cfg(feature = "dangerous_configuration")]
     pub use client_conn::danger::DangerousClientConfig;
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -433,6 +433,71 @@ impl WebPkiVerifier {
     }
 }
 
+
+
+impl ServerCertVerifier for WebPkiVerifierAnyServerName {
+    /// Will verify the certificate is valid in the following ways:
+    /// - Signed by a  trusted `RootCertStore` CA
+    /// - Not Expired
+    fn verify_server_cert(
+        &self,
+        end_entity: &Certificate,
+        intermediates: &[Certificate],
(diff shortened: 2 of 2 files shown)


ctz:
I'm not sure the small amount of code needed here to address turning off a common security control is that costly to have in the downstream crate.

The background to having the crate like this is to avoid [situations in the past](https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf) where disabling security controls was easy, and people shipped libraries with them turned off (because, it works, so ship it).  I think having every such user to maintain their own code that turns off a security control transfers the cost of that to the right place in the system.

>  the IpAdress could change.

The certificate could have all possible names, addresses and address ranges (presumably that is knowable information, or the peer would not be able to connect). Alternatively the certificate could get refreshed each time its address changes.

Fritiofhedstrom:
Thanks for your replies!
I will address your comments.

> I'm not sure the small amount of code needed here to address turning off a common security control is that costly to have in the downstream crate.

As it is now it is a very small amount of code, but that is because i am reusing almost everything from the default verifier. If i were to implement this in the downstream crate i would need to do some copying unless i make a different pr to make some verification helper functions public, would you prefer that solution?

> The certificate could have all possible names, addresses and address ranges (presumably that is knowable information, or the peer would not be able to connect). Alternatively the certificate could get refreshed each time its address changes.

Hmm, i have not thought about setting adress range, how do i even do that? ServerName would need to be an std::net::IpAdress and that can't represent a range right?

Fritiofhedstrom:
I made this pr: https://github.com/rustls/rustls/pull/1325
Should this one be closed?

cpu:
> Should this one be closed?

Sure, let's focus discussion on #1325 instead :+1:

PR #1321, opened 2023-06-21
Localise DnsName types

ctz:
The goal of this is getting closer to avoiding a dependency on the `webpki` crate if our user chooses to not validate certificates with it.

After this PR, `webpki` remains referenced from:

- `src/verify.rs` (home of the default verifier)
- `src/server/handy.rs` (checking certificates against SNI names during configuration)
- `src/anchors.rs` (producing `webpki::TrustAnchor`s)

### DIFF
--- a/rustls/src/dns_name.rs
+++ b/rustls/src/dns_name.rs
@@ -1,0 +1,219 @@
+/// DNS name validation according to RFC1035, but with underscores allowed.
+use std::error::Error as StdError;
+use std::fmt;
+
+/// A type which encapsulates an owned string that is a syntactically valid DNS name.
+#[derive(Clone, Eq, Hash, PartialEq, Debug)]
+pub struct DnsName(String);
+
+impl<'a> DnsName {
+    /// Produce a borrowed `DnsNameRef` from this owned `DnsName`.
+    pub fn borrow(&'a self) -> DnsNameRef<'a> {
+        DnsNameRef(self.as_ref())
+    }
+
+    /// Validate the given bytes are a DNS name if they are viewed as ASCII.
+    pub fn try_from_ascii(bytes: &[u8]) -> Result<Self, InvalidDnsNameError> {
+        // nb. a sequence of bytes that is accepted by `validate()` is both
+        // valid UTF-8, and valid ASCII.
+        String::from_utf8(bytes.to_vec())
+            .map_err(|_| InvalidDnsNameError)
+            .and_then(Self::try_from)
+    }
+}
+
+impl TryFrom<String> for DnsName {
+    type Error = InvalidDnsNameError;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        validate(value.as_bytes())?;
+        Ok(Self(value))
+    }
+}
+
+impl AsRef<str> for DnsName {
+    fn as_ref(&self) -> &str {
+        AsRef::<str>::as_ref(&self.0)
+    }
+}
+
+/// A type which encapsulates a borrowed string that is a syntactically valid DNS name.
+#[derive(Eq, Hash, PartialEq, Debug)]
+pub struct DnsNameRef<'a>(&'a str);
+
+impl<'a> DnsNameRef<'a> {
+    /// Copy this object to produce an owned `DnsName`.
+    pub fn to_owned(&'a self) -> DnsName {
+        DnsName(self.0.to_string())
+    }
+
+    /// Copy this object to produce an owned `DnsName`, smashing the case to lowercase
+    /// in one operation.
+    pub fn to_lowercase_owned(&'a self) -> DnsName {
+        DnsName(self.0.to_lowercase())
+    }
+}
+
+impl<'a> TryFrom<&'a str> for DnsNameRef<'a> {
+    type Error = InvalidDnsNameError;
+
+    fn try_from(value: &'a str) -> Result<DnsNameRef<'a>, Self::Error> {
+        validate(value.as_bytes())?;
+        Ok(DnsNameRef(value))
+    }
+}
+
+impl<'a> AsRef<str> for DnsNameRef<'a> {
+    fn as_ref(&self) -> &str {
+        self.0
+    }
+}
+
+/// The provided input could not be parsed because
+/// it is not a syntactically-valid DNS Name.
+#[derive(Debug)]
+pub struct InvalidDnsNameError;
+
+impl fmt::Display for InvalidDnsNameError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("invalid dns name")
+    }
+}
+
+impl StdError for InvalidDnsNameError {}
+
+fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
+    use State::*;
+    let mut state = Start;
+
+    /// "Labels must be 63 characters or less."
+    const MAX_LABEL_LENGTH: usize = 63;
+
+    for ch in input {
+        state = match (state, ch) {
+            (Start | Next | Hyphen { .. }, b'.') => return Err(InvalidDnsNameError),
+            (Subsequent { .. }, b'.') => Next,
+            (Subsequent { len }, _) if len >= MAX_LABEL_LENGTH => return Err(InvalidDnsNameError),
+            (Start | Next, b'a'..=b'z' | b'A'..=b'Z') => Subsequent { len: 1 },
+            (Subsequent { len }, b'-') => Hyphen { len: len + 1 },
+            (
+                Subsequent { len } | Hyphen { len },
+                b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'-' | b'0'..=b'9',
+            ) => Subsequent { len: len + 1 },
+            _ => return Err(InvalidDnsNameError),
+        };
+    }
+
+    if matches!(state, Start | Hyphen { .. }) {
+        return Err(InvalidDnsNameError);
+    }
+
+    Ok(())
+}
+
+enum State {
+    Start,
+    Next,
+    Subsequent { len: usize },
+    Hyphen { len: usize },
+}
+
+#[cfg(test)]
+mod test {
+    static TESTS: &[(&'static str, bool)] = &[
+        ("", false),
+        ("localhost", true),
+        ("LOCALHOST", true),
+        (".localhost", false),
+        ("..localhost", false),
+        ("1.2.3.4", false),
+        ("127.0.0.1", false),
+        ("absolute.", true),
+        ("absolute..", false),
+        ("multiple.labels.absolute.", true),
+        ("foo.bar.com", true),
+        ("infix-hyphen-allowed.com", true),
+        ("-prefixhypheninvalid.com", false),
+        ("suffixhypheninvalid-.com", false),
+        ("foo.lastlabelendswithhyphen-", false),
+        ("infix_underscore_allowed.com", true),
+        ("_prefixunderscoreinvalid.com", false),
+        ("labelendswithnumber1.bar.com", true),
+        ("xn--bcher-kva.example", true),
+        (
+            "sixtythreesixtythreesixtythreesixtythreesixtythreesixtythreesix.com",
+            true,
+        ),
+        (
+            "sixtyfoursixtyfoursixtyfoursixtyfoursixtyfoursixtyfoursixtyfours.com",
+            false,
+        ),
+    ];
+
+    #[test]
+    fn test_validation() {
+        for (input, expected) in TESTS {
+            println!("test: {:?} expected valid? {:?}", input, expected);
+            let name_ref = super::DnsNameRef::try_from(*input);
+            assert_eq!(*expected, name_ref.is_ok());
+            let name = super::DnsName::try_from(input.to_string());
+            assert_eq!(*expected, name.is_ok());
+        }
+    }
+
+    #[test]
+    fn error_is_debug() {
+        assert_eq!(
+            format!("{:?}", super::InvalidDnsNameError),
+            "InvalidDnsNameError"
+        );
+    }
+
+    #[test]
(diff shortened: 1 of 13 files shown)


rustls/src/server/handy.rs
```diff
@@ -160,7 +160,43 @@ impl ResolvesServerCertUsingSni {
...
+        let end_entity_cert = ck.end_entity_cert().map_err(|_| {
+            Error::General("No end-entity certificate in certificate chain".to_string())
+        })?;
+
+        // Reject syntactically-invalid end-entity certificates.
+        let end_entity_cert =
+            webpki::EndEntityCert::try_from(end_entity_cert.as_ref()).map_err(|_| {
+                Error::General(
+                    "End-entity certificate in certificate \
+                                  chain is syntactically invalid"
```
cpu, line 182: style nit: This is copied verbatim from the `sign.rs` version, but is there a reason to keep the `\` escaped newlines? It looks like both instances fit into a normal contiguous string literal without any `rustfmt` weirdness.

cpu (approved):
This looks good to me. I went through the diff pretty carefully while cross-checking against the `webpki` representation of the same validation and I didn't spot anything suspicious :+1:

rustls/src/dns_name.rs
```diff
@@ -0,0 +1,176 @@
+/// DNS name validation according to RFC1035, but with underscores allowed.
+use std::error::Error as StdError;
+use std::fmt;
+
+/// "Labels must be 63 characters or less."
+const MAX_LABEL_LENGTH: usize = 63;
+
+fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
```
djc, line 8: Style: I'd like to pitch you on the idea that we should order code the other way around, in a top-down way instead of bottom-up. That is, I think `MAX_LABEL_LENGTH` and `validate()` should go near the bottom of the module because we're more likely to want to refer to/change `DnsName` and `DnsNameRef`.

I do this in all my projects and most of the rustls code that I've had a substantial hand in is also like this. I think it's easier to navigate.

rustls/src/dns_name.rs
```diff
@@ -0,0 +1,176 @@
...
+    let mut label_length = 0;
+    let mut previous_label_ch = b'-';
```
djc, line 10: For both of these, I would consider making them `Option`, having them be reset to `None` once a `.` rolls around.

Should we also error on invalid label lengths (64 characters or more)?

rustls/src/dns_name.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl StdError for InvalidDnsNameError {}
+
+/// A type which encapsulates an owned string that is a syntactically valid DNS name.
+#[derive(Clone, Eq, Hash, PartialEq, Debug)]
+pub struct DnsName(String);
+
+impl<'a> DnsName {
+    /// Produce a borrowed `DnsNameRef` from this owned `DnsName`.
+    pub fn borrow(&'a self) -> DnsNameRef<'a> {
```
djc, line 74: Maybe this should implement the `Borrow` trait, and we could implement the `ToOwned` trait on `DnsNameRef`?

rustls/src/dns_name.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl<'a> DnsNameRef<'a> {
+    /// Copy this object to produce an owned `DnsName`.
+    pub fn to_owned(&'a self) -> DnsName {
+        DnsName(self.0.to_string())
+    }
+
+    /// Copy this object to produce an owned `DnsName`, smashing the case to lowercase
+    /// in one operation.
+    pub fn to_owned_normalised(&'a self) -> DnsName {
```
djc, line 115: I think we should use US spelling in API?

rustls/src/dns_name.rs
```diff
@@ -0,0 +1,176 @@
...
+        return Err(InvalidDnsNameError);
+    }
+
+    Ok(())
+}
+
+/// The provided input could not be parsed because
+/// it is not a syntactically-valid DNS Name.
+#[derive(Debug)]
+pub struct InvalidDnsNameError;
```
djc, line 58: Why not make this an enum with precise errors?

rustls/src/msgs/handshake.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(non_camel_case_types)]
+use crate::dns_name;
```
djc, line 2: I feel like we should just import `DnsName` and `DnsNameRef` directly -- it doesn't seem like there's much potential for confusion with another type, and it seems more in line with the way we usually do imports.

rustls/src/msgs/handshake_test.rs
```diff
@@ -1,3 +1,4 @@
+use crate::dns_name;
```
djc, line 1: Import directly?

rustls/src/dns_name.rs
djc: Also I like to write these kinds of little parsers more like this:

```rust
fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
    use State::*;
    let mut state = Start;

    for ch in input {
        state = match (state, ch) {
            (Start | Next | Hyphen { .. }, b'.') => return Err(InvalidDnsNameError),
            (Subsequent { .. }, b'.') => Next,
            (Subsequent { len }, _) if len >= MAX_LABEL_LENGTH => return Err(InvalidDnsNameError),
            (Start | Next, b'a'..=b'z' | b'A'..=b'Z') => Subsequent { len: 1 },
            (Subsequent { len }, b'-') => Hyphen { len: len + 1 },
            (
                Subsequent { len } | Hyphen { len },
                b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'0'..=b'9',
            ) => Subsequent { len: len + 1 },
            _ => return Err(InvalidDnsNameError),
        };
    }

    if matches!(state, Start | Hyphen { .. }) {
        return Err(InvalidDnsNameError);
    }

    Ok(())
}

#[derive(Debug)]
enum State {
    Start,
    Next,
    Subsequent { len: usize },
    Hyphen { len: usize },
}
```

What do you think? This makes it a bit more explicit what the state is (less potential for `label_length` and `previous_label_ch` to be in inconsistent or invalid states) and is a bit more compact.

ctz: Oh, that is a big improvement! Will integrate that.

rustls/src/dns_name.rs
ctz: I think my brain is still a bit C++-addled and I default from dependencies to dependants. Will try the "headlines then detail" style.

rustls/src/dns_name.rs
ctz: -> `to_lowercase_owned` to avoid the question and better express the purpose.

rustls/src/dns_name.rs
djc: I don't mind calling you out on that if you agree that it's worthwhile! Just wanted to make sure you were even interested in doing the headlines-first style.

rustls/src/server/handy.rs
ctz: have fixed that, not sure where they came from originally -- i'm not sure rustfmt splits strings like this unless the code is really wide.

rustls/src/dns_name.rs
ctz: Please do!

rustls/src/dns_name.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+impl<'a> AsRef<str> for DnsNameRef<'a> {
+    fn as_ref(&self) -> &str {
+        self.0
+    }
+}
+
+/// "Labels must be 63 characters or less."
+const MAX_LABEL_LENGTH: usize = 63;
```
djc, line 86: So I would probably move `MAX_LABEL_LENGTH` below `State`, with `InvalidDnsNameError` between the two.

ctz: thoughts on putting `MAX_LABEL_LENGTH` inside `fn validate`?

rustls/src/dns_name.rs
ctz: I think making detailed/useful errors here makes things more complex and the benefit is relatively limited.

rustls/src/dns_name.rs
djc: Makes sense!

rustls/src/dns_name.rs
ctz: I tried that initially and didn't get it to work. `Borrow` wants to return a `&DnsNameRef<'a>` which would be a temporary. That would be OK, except the callers want a `DnsNameRef<'a>` and making `DnsNameRef` `Copy + Clone` broke `ToOwned`:

```
error[E0119]: conflicting implementations of trait `ToOwned` for type `dns_name::DnsNameRef<'_>`
  --> rustls/src/dns_name.rs:55:1
   |
55 | impl<'a> ToOwned for DnsNameRef<'a> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `alloc`:
           - impl<T> ToOwned for T
             where T: Clone;
```

The full diff is:

```
diff --git a/rustls/src/client/client_conn.rs b/rustls/src/client/client_conn.rs
index f77b5508..f5f8a53a 100644
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -25,6 +25,7 @@ use std::marker::PhantomData;
 use std::net::IpAddr;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
+use std::borrow::Borrow;
 use std::{fmt, io, mem};
 
 /// A trait for the ability to store client session data, so that sessions
@@ -367,7 +368,7 @@ impl ServerName {
     /// in the handshake.
     pub(crate) fn for_sni(&self) -> Option<dns_name::DnsNameRef> {
         match self {
-            Self::DnsName(dns_name) => Some(dns_name.borrow()),
+            Self::DnsName(dns_name) => Some(*dns_name.borrow()),
             Self::IpAddress(_) => None,
         }
     }
diff --git a/rustls/src/dns_name.rs b/rustls/src/dns_name.rs
index dde14502..18b6d0e9 100644
--- a/rustls/src/dns_name.rs
+++ b/rustls/src/dns_name.rs
@@ -1,5 +1,6 @@
 /// DNS name validation according to RFC1035, but with underscores allowed.
 use std::error::Error as StdError;
+use std::borrow::Borrow;
 use std::fmt;
 
 /// A type which encapsulates an owned string that is a syntactically valid DNS name.
@@ -7,11 +8,6 @@ use std::fmt;
 pub struct DnsName(String);
 
 impl<'a> DnsName {
-    /// Produce a borrowed `DnsNameRef` from this owned `DnsName`.
-    pub fn borrow(&'a self) -> DnsNameRef<'a> {
-        DnsNameRef(self.as_ref())
-    }
-
     /// Validate the given bytes are a DNS name if they are viewed as ASCII.
     pub fn try_from_ascii(bytes: &[u8]) -> Result<Self, InvalidDnsNameError> {
         // nb. a sequence of bytes that is accepted by `validate()` is both
@@ -22,6 +18,13 @@ impl<'a> DnsName {
     }
 }
 
+impl<'a> Borrow<DnsNameRef<'a>> for DnsName {
+    /// Produce a borrowed `DnsNameRef` from this owned `DnsName`.
+    fn borrow(&self) -> &'a DnsNameRef<'a> {
+        &DnsNameRef(self.as_ref())
+    }
+}
+
 impl TryFrom<String> for DnsName {
     type Error = InvalidDnsNameError;
 
@@ -38,15 +41,10 @@ impl AsRef<str> for DnsName {
 }
 
 /// A type which encapsulates a borrowed string that is a syntactically valid DNS name.
-#[derive(Eq, Hash, PartialEq, Debug)]
+#[derive(Eq, Hash, PartialEq, Debug, Copy, Clone)]
 pub struct DnsNameRef<'a>(&'a str);
 
 impl<'a> DnsNameRef<'a> {
-    /// Copy this object to produce an owned `DnsName`.
-    pub fn to_owned(&'a self) -> DnsName {
-        DnsName(self.0.to_string())
-    }
-
     /// Copy this object to produce an owned `DnsName`, smashing the case to lowercase
     /// in one operation.
     pub fn to_lowercase_owned(&'a self) -> DnsName {
@@ -54,6 +52,15 @@ impl<'a> DnsNameRef<'a> {
     }
 }
 
+impl<'a> ToOwned for DnsNameRef<'a> {
+    type Owned = DnsName;
+
+    /// Copy this object to produce an owned `DnsName`.
+    fn to_owned(&self) -> Self::Owned {
+        DnsName(self.0.to_string())
+    }
+}
+
 impl<'a> TryFrom<&'a str> for DnsNameRef<'a> {
     type Error = InvalidDnsNameError;
 
diff --git a/rustls/src/msgs/handshake.rs b/rustls/src/msgs/handshake.rs
index f8be007e..07423c30 100644
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -16,6 +16,7 @@ use crate::verify::DigitallySignedStruct;
 
 use std::collections;
 use std::fmt;
+use std::borrow::Borrow;
 
 /// Create a newtype wrapper around a given type.
 ///
@@ -295,7 +296,7 @@ impl ConvertServerNameList for [ServerName] {
     fn get_single_hostname(&self) -> Option<DnsNameRef> {
         fn only_dns_hostnames(name: &ServerName) -> Option<DnsNameRef> {
             if let ServerNamePayload::HostName(ref dns) = name.payload {
-                Some(dns.borrow())
+                Some(*dns.borrow())
             } else {
                 None
             }

```

djc: Fair enough!

rustls/src/dns_name.rs
cpu: That's a nice refactor :+1:

PR #1322, opened 2023-06-21
WIP: client certificate revocation checking support.

cpu:
**Note: left as a draft since it relies on unreleased Webpki code, and because there are some open questions left to resolve**

This branch is an initial implementation of client certificate revocation checking using the recently added webpki CRL support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,6 @@ members = [
 ]
 exclude = ["admin/rustfmt"]
 resolver = "2"
+
+[patch.crates-io]
+webpki = { package='rustls-webpki', git = 'https://github.com/cpu/webpki.git', branch = 'cpu-crl-support-rustls-fixes' }
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -558,10 +558,16 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
         for root in roots {
             client_auth_roots.add(&root).unwrap();
         }
+        // TODO(@cpu): Parse CRLs from Args.
+        let crls = Vec::default();
         if args.flag_require_auth {
-            AllowAnyAuthenticatedClient::new(client_auth_roots).boxed()
+            AllowAnyAuthenticatedClient::new(client_auth_roots, crls)
+                .expect("invalid client auth config")
+                .boxed()
         } else {
-            AllowAnyAnonymousOrAuthenticatedClient::new(client_auth_roots).boxed()
+            AllowAnyAnonymousOrAuthenticatedClient::new(client_auth_roots, crls)
+                .expect("invalid client auth config")
+                .boxed()
         }
     } else {
         NoClientAuth::boxed()
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -17,7 +17,8 @@ extern crate serde_derive;
 use docopt::Docopt;
 
 use rustls::server::{
-    AllowAnyAnonymousOrAuthenticatedClient, AllowAnyAuthenticatedClient, NoClientAuth,
+    AllowAnyAnonymousOrAuthenticatedClient, AllowAnyAuthenticatedClient, CertRevocationList,
+    NoClientAuth,
 };
 use rustls::{self, RootCertStore};
 
@@ -429,6 +430,8 @@ Options:
                         to certificate.  Optional.
     --auth CERTFILE     Enable client authentication, and accept certificates
                         signed by those roots provided in CERTFILE.
+    --crl CRLFILE ...   Perform client certificate revocation checking using the DER-encoded
+                        CRLFILE. May be used multiple times.
     --require-auth      Send a fatal alert if the client does not complete client
                         authentication.
     --resumption        Support session resumption.
@@ -454,6 +457,7 @@ struct Args {
     flag_suite: Vec<String>,
     flag_proto: Vec<String>,
     flag_certs: Option<String>,
+    flag_crl: Vec<String>,
     flag_key: Option<String>,
     flag_ocsp: Option<String>,
     flag_auth: Option<String>,
@@ -551,6 +555,21 @@ fn load_ocsp(filename: &Option<String>) -> Vec<u8> {
     ret
 }
 
+fn load_crls(filenames: &Vec<String>) -> Vec<CertRevocationList> {
+    let mut ret = Vec::new();
+
+    for filename in filenames {
+        let mut der = Vec::new();
+        fs::File::open(filename)
+            .expect("cannot open CRL file")
+            .read_to_end(&mut der)
+            .unwrap();
+        ret.push(CertRevocationList(der));
+    }
+
+    ret
+}
+
 fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
     let client_auth = if args.flag_auth.is_some() {
         let roots = load_certs(args.flag_auth.as_ref().unwrap());
@@ -558,8 +577,7 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
         for root in roots {
             client_auth_roots.add(&root).unwrap();
         }
-        // TODO(@cpu): Parse CRLs from Args.
-        let crls = Vec::default();
+        let crls = load_crls(&args.flag_crl);
         if args.flag_require_auth {
             AllowAnyAuthenticatedClient::new(client_auth_roots, crls)
                 .expect("invalid client auth config")
@@ -638,6 +656,11 @@ fn main() {
         env_logger::Builder::new()
             .parse_filters("trace")
             .init();
+    }
+
+    if !args.flag_crl.is_empty() && args.flag_auth.is_none() {
+        println!("-crl can only be provided with -auth enabled");
+        return;
     }
 
     let mut addr: net::SocketAddr = "0.0.0.0:443".parse().unwrap();
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -66,6 +66,10 @@ pub enum Error {
     /// implementation.
     InvalidCertificate(CertificateError),
 
+    /// A provided certificate revocation list (CRL) was invalid.
+    // TODO(@cpu): provide more detail, maybe with a sub-enum, or by exposing webpki::Error?
+    InvalidCrl,
+
     /// The presented SCT(s) were invalid.
     InvalidSct(sct::Error),
 
@@ -418,6 +422,9 @@ impl fmt::Display for Error {
             Self::AlertReceived(ref alert) => write!(f, "received fatal alert: {:?}", alert),
             Self::InvalidCertificate(ref err) => {
                 write!(f, "invalid peer certificate: {:?}", err)
+            }
+            Self::InvalidCrl => {
+                write!(f, "invalid certificate revocation list")
             }
             Self::NoCertificatesPresented => write!(f, "peer sent no certificates"),
             Self::UnsupportedNameType => write!(f, "presented server name type wasn't supported"),
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -575,6 +575,7 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
             &webpki::TlsClientTrustAnchors(&trustroots),
             &chain,
             now,
+            &[], // TODO(@cpu): provide CRLs as appropriate.
         )
         .map_err(pki_error)
         .map(|_| ClientCertVerified::assertion())
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -522,26 +522,46 @@ fn prepare<'a, 'b>(
     Ok((cert, intermediates, trustroots))
 }
 
+/// A DER encoded Certificate Revocation List (CRL).
+pub struct CertRevocationList(pub Vec<u8>);
+
+impl CertRevocationList {
+    /// Parse the CRL DER, yielding a [`webpki::CertRevocationList`] or an error if the CRL
+    /// is malformed, or uses unsupported features.
+    pub fn parse(&self) -> Result<webpki::CertRevocationList, Error> {
+        // TODO(@cpu): Revisit map_err error specificity.
+        webpki::CertRevocationList::from_der(&self.0).map_err(|_| Error::InvalidCrl)
+    }
+}
+
 /// A `ClientCertVerifier` that will ensure that every client provides a trusted
-/// certificate, without any name checking.
+/// certificate, without any name checking. Optionally, client certificates will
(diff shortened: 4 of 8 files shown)


Cargo.toml
```diff
@@ -7,3 +7,6 @@ members = [
 ]
 exclude = ["admin/rustfmt"]
 resolver = "2"
+
+[patch.crates-io]
+webpki = { package='rustls-webpki', git = 'https://github.com/cpu/webpki.git', branch = 'cpu-crl-support-rustls-fixes' }
```
cpu, line 12: We can change this to use `main` from `rustls/webpki` once https://github.com/rustls/webpki/pull/92 lands.

examples/src/bin/tlsserver-mio.rs
```diff
@@ -429,6 +429,8 @@ Options:
                         to certificate.  Optional.
     --auth CERTFILE     Enable client authentication, and accept certificates
                         signed by those roots provided in CERTFILE.
+    --crl CRLFILE ...   Perform client certificate revocation checking using the DER-encoded
+                        CRLFILE. May be used multiple times.
```
cpu, line 433: It probably makes sense to error if `--crl` is provided and not `--auth` - I haven't done that in this branch to keep things simple while we figure out some other more interesting bits.

rustls/src/error.rs
```diff
@@ -66,6 +66,10 @@ pub enum Error {
     /// implementation.
     InvalidCertificate(CertificateError),
 
+    /// A provided certificate revocation list (CRL) was invalid.
+    // TODO(@cpu): provide more detail, maybe with a sub-enum, or by exposing webpki::Error?
+    InvalidCrl,
```
cpu, line 71: This is one point I'd appreciate some guidance on - How do we want to handle the `webpki::Error`s returned by `CertRevocationList::from_der`.

We could:

* Follow the lead of `CertificateError` and make a sub-enum that has Rustls-specified versions of the possible error returns, and then impl a  `TryFrom<webpki::Error>` or something like `verify.rs`'s [`pki_error`](https://github.com/rustls/rustls/blob/5053fde25bb0fde5715a95601a62c8fae626600e/rustls/src/verify.rs#L653-L667) fn.
* Change `InvalidCrl` to something like `InvalidCrl(Arc<dyn StdError + Send + Sync>)` and wrap the `webpki::Error` in an arc as a generic error
* Leave it as-is and not provide more specificity.
* Something else?

rustls/tests/common/mod.rs
```diff
@@ -284,7 +284,8 @@ pub fn get_client_root_store(kt: KeyType) -> RootCertStore {
 pub fn make_server_config_with_mandatory_client_auth(kt: KeyType) -> ServerConfig {
     let client_auth_roots = get_client_root_store(kt);
 
-    let client_auth = AllowAnyAuthenticatedClient::new(client_auth_roots);
+    // TODO(@cpu): allow building test configuration with CRLs.
+    let client_auth = AllowAnyAuthenticatedClient::new(client_auth_roots, Vec::default());
```
cpu, line 288: I'll commit to writing some actual tests for this once we have the webpki side figured out and more of the questions in this branch are resolved.

rustls/src/verify.rs
```diff
@@ -539,24 +540,28 @@ fn prepare<'a, 'b>(
 }
 
 /// A `ClientCertVerifier` that will ensure that every client provides a trusted
-/// certificate, without any name checking.
+/// certificate, without any name checking. Optionally, client certificates will
+/// have their revocation status checked using the DER encoded CRLs provided.
 pub struct AllowAnyAuthenticatedClient {
     roots: RootCertStore,
     subjects: Vec<DistinguishedName>,
+    crls: Vec<Vec<u8>>,
```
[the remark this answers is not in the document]
cpu: I'll probably put up a branch extending https://github.com/rustls/webpki/pull/92 with an owned representation for comparison purposes anyway. I think it will be required one way or the other for https://github.com/rustls/webpki/issues/80 and https://github.com/rustls/webpki/issues/81

rustls/src/verify.rs
```diff
@@ -591,7 +596,7 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
             &webpki::TlsClientTrustAnchors(&trustroots),
             &chain,
             now,
-            &[], // TODO(@cpu): provide CRLs as appropriate.
+            &[], // TODO(@cpu): parse CRL DER and provide.
```
djc: I don't think we should keep this as a separate commit.

rustls/src/verify.rs
```diff
@@ -15,6 +15,7 @@ use ring::digest::Digest;
 
 use std::sync::Arc;
 use std::time::SystemTime;
+use webpki::CertRevocationList;
```
djc, line 18: Style nit: should separate the webpki import from the std import block.

rustls/src/verify.rs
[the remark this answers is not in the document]
djc: I think even if we're going to keep the CRL data around as a `Vec<u8>` we should wrap that in a newtype wrapper (like `PrivateKey` and `Certificate`) to make usage somewhat more intentional.

rustls/src/verify.rs
```diff
@@ -590,13 +590,24 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
         now: SystemTime,
     ) -> Result<ClientCertVerified, Error> {
         let (cert, chain, trustroots) = prepare(end_entity, intermediates, &self.roots)?;
+        let crls = self
+            .crls
+            .iter()
+            .map(|crl_der| webpki::CertRevocationList::from_der(crl_der))
+            .collect::<Vec<Result<CertRevocationList, webpki::Error>>>()
+            .into_iter()
```
djc, line 598: I think you can collect into `Result<Vec<_>, _>` directly (without the intermediate collection step).

rustls/src/verify.rs
```diff
@@ -590,13 +590,24 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
...
+            .iter()
+            .collect::<Vec<&CertRevocationList>>();
         let now = webpki::Time::try_from(now).map_err(|_| Error::FailedToGetCurrentTime)?;
         cert.verify_is_valid_tls_client_cert(
             SUPPORTED_SIG_ALGS,
             &webpki::TlsClientTrustAnchors(&trustroots),
             &chain,
             now,
-            &[], // TODO(@cpu): parse CRL DER and provide.
+            crls.as_slice(),
```
djc, line 610: Maybe `verify_is_valid_tls_client_cert()` should take an `impl Iterator<Item = &CertRevocationList<'_>>` instead of a slice? Could avoid the intermediate allocations here.

rustls/src/error.rs
djc: To what extent are CRL errors from webpki disjoint from other webpki errors? I do feel like we would definitely want to separate "error while parsing CRL" from "client certificate was revoked according to CRL" in our rustls error messaging.

In terms of timing, maybe we should also try to parse the `CertRevocationList`s once at `AllowAuthenticatedClient` construction just to draw out any errors from parsing?

rustls/src/verify.rs
ctz: perhaps something stronger than just `impl Iterator` would be needed, as the CRLs need to be iterated a bunch of times per `verify_is_valid_tls_client_cert`?

djc: Ahh yes, we'll need to loop over it once per certificate considered for chain building. I suppose in that case a slice would be best after all.

rustls/src/verify.rs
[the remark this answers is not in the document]
cpu: Good idea.

rustls/src/verify.rs
cpu: For my own education what do you mean by "something stronger"? Is the appeal of maintaining the slice related to some overhead of an `impl Iterator` approach? I have no strong pref. here, just want to understand the trade-offs better.

ctz: I don't know how to write it precisely, but it'd be something like `impl Iterator<...> + Clone` (ie, it's an Iterator, but you can also clone the iterator to get more) or a function that returns a fresh `impl Iterator` each time.

rustls/src/error.rs
cpu: I need to dig in to where the "client certificate was revoked" error is getting bubbled up through Rustls. I have a feeling it's being converted to a Rustls internal error already. The error I added/flagged for discussion here is strictly from the "error while parsing CRL" codepath.

For the parsing errors, I think we can emit:
* UnsupportedCrlVersion
* UnsupportedDeltaCrl
* UnsupportedIndirectCrl
* UnsupportedRevocationReason
* InvalidCrlNumber
* InvalidSerialNumber
* SignatureAlgorithmMismatch
* UnknownIssuer
* MalformedExtensions
* BadDerTime
* BadDer (like almost all parts of webpki :sweat_smile:)

Of these I think probably the "Unsupported" ones are the most interesting to a user and most disjoint from existing errors. Most of the others (minus InvalidCrlNuber) are already used by webpki elsewhere.

On the path of checking a cert for revocation with a CRL since we front-load most of the error checking for the CRL itself I think we only emit:

* CertRevoked
* IssuerNotCrlSigner
* InvalidSignatureForPublicKey

Of these only CertRevoked and IssuerNotCrlSigner are new errors not previously possible for Rustls code to see.

cpu: > In terms of timing, maybe we should also try to parse the CertRevocationLists once at AllowAuthenticatedClient construction just to draw out any errors from parsing?

That's a good idea :+1:

examples/src/bin/tlsserver-mio.rs
cpu: I added this quickly since it was a small change ultimately.

rustls/src/error.rs
cpu: Implemented up-front parsing in my recent push.

rustls/src/verify.rs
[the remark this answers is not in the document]
cpu: Added the newtype wrapper in my recent push.

rustls/src/verify.rs
cpu: Collapsed into one commit.

rustls/src/verify.rs
cpu: This fell away when I introduced the local newtype to wrap `Vec<u8>` the ~two places we want to use the webpki type I changed to be explicit references to `webpki::CertRevocationList`.

rustls/src/verify.rs
cpu: Good call, thanks. Fixed up.

rustls/src/verify.rs
cpu: I haven't tackled this feedback yet (want to try some other stuff first) but it's on my radar.

rustls/src/verify.rs
```diff
@@ -569,12 +589,21 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
         now: SystemTime,
     ) -> Result<ClientCertVerified, Error> {
         let (cert, chain, trustroots) = prepare(end_entity, intermediates, &self.roots)?;
+        let crls = self
+            .crls
+            .iter()
+            .map(|crl| crl.parse())
+            .collect::<Result<Vec<_>, _>>()?;
```
cpu, line 596: leaving a note we could `unwrap()` here now that we're pre-validating the CRL DER in `new`.

rustls/src/error.rs
cpu: Started on benchmarking: https://github.com/rustls/webpki/pull/94

rustls/src/verify.rs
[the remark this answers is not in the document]
cpu: > The intuition here is that going from DER to the CRL view over the DER is cheap (especially since signature validation isn't done until a cert with an issuer matching the CRL is consulted for a serial), but that should be verified with benchmarking (I will pursue that separately).

Started on benchmarking: https://github.com/rustls/webpki/pull/94

Initial results show parsing the CRL and finding serials in the CRL are roughly equivalent in runtime, with both seeming fairly quick from my perspective.

rustls/src/error.rs
cpu: > I need to dig in to where the "client certificate was revoked" error is getting bubbled up through Rustls. I have a feeling it's being converted to/masked by a Rustls internal error already.

Traced this out. We were turning the `webpki::Error::CertRevoked` return from the verifier into a `InvalidCertificate(CertificateError::Other(...))` err. I had to add a match arm in `verify.rs`'s `pki_error` for `webpki::RevokedCert`. Conveniently there was already a `CertificateError::Revoked` on the Rustls side to map it to. I force pushed an update to this branch to include that change and in the "happy" revocation path we now get a sensible error bubbled up to the user:

Server side:
```
cargo run --package rustls-examples --bin tlsserver-mio -- 
  -p 4443 
  --certs ./pki/localhost/cert.pem 
  --key ./pki/localhost/key.pem 
  --auth ./pki/minica.pem
  --crl ./pki/demo.crl
  --crl ./pki/demo.2.crl
  http
```

Client side:
```
cargo run --bin tlsclient-mio -- 
  -p 4443 
  --http localhost 
  --cafile ./pki/minica.pem 
  --auth-key ./pki/localhost/key.pem 
  --auth-certs ./pki/localhost/cert.pem
```

```
TLS error: AlertReceived(CertificateRevoked)
Connection closed
```

rustls/src/verify.rs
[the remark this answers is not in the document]
djc: I guess the next question is, how does this stuff compare in performance to the other things a webpki client certificate verification has to run through? That is, what is the impact on this relative to the whole operation?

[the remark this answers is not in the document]
cpu: > I guess the next question is, how does this stuff compare in performance to the other things a webpki client certificate verification has to run through?

Probably the most substantial comparison point would be to path building in general, though my expectation there is that signature validation will be the dominant factor. I'll give this some thought.

> Initial results show parsing the CRL and finding serials in the CRL are roughly equivalent in runtime, with both seeming fairly quick from my perspective.

Just noting that this has an obvious explanation I overlooked :facepalm: - When we parse CRLs we're also walking the entire set of revoked certificates to detect errors early.

Before I start on a more time intensive side quest to benchmark path building I'm going to rework the code under benchmark to avoid doing that up-front check so we can have a better sense of the speed of just parsing the base CRL vs parsing/searching the revoked certificates.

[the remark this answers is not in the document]
cpu: >  I'm going to rework the code under benchmark to avoid doing that up-front check so we can have a better sense of the speed of just parsing the base CRL vs parsing/searching the revoked certificates.

Pushed that change to https://github.com/rustls/webpki/pull/94 and the results now clearly show that parsing the base CRL is very fast relative to parsing and searching its revoked cert contents.

PR #1323, opened 2023-06-22
Import DnsName types directly

djc:
Didn't quite make this explicit in #1321, but seems like an improvement?


cpu (commented):
Seems like an improvement to me :+1: I'll let Ctz give it a real +1 since it was his code.

ctz (approved):

PR #1324, opened 2023-06-23
Fix connect-tests for 1000-sans.badssl.com

ctz:
This regressed in ae4ca32.  DNS names starting with and wholly numerals are in wide use.

RFC1035 actually doesn't really reflect reality, and RFC1123 updates it to allow labels that start with numerals or are wholly numeric.  The only restriction is that the last label cannot be wholly numeric.

Also:

- Limit the entire length of a DNS name.
- Ensure a label that ends with hyphens respects the label length limit.

(Both of these were found by writing a fuzzer that compared behaviour between the old and new implementations.)

### DIFF
--- a/rustls/src/dns_name.rs
+++ b/rustls/src/dns_name.rs
@@ -89,22 +89,45 @@ fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
     /// "Labels must be 63 characters or less."
     const MAX_LABEL_LENGTH: usize = 63;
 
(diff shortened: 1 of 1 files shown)


cpu (approved):
:sweat: the real world is complicated.

> (Both of these were found by writing a fuzzer that compared behaviour between the old and new implementations.)

Really good idea :+1:

djc:
Should we also add a link to the relevant section of the RFC?

PR #1325, opened 2023-06-26
expose verify helper functions

Fritiofhedstrom:
Alternative to: https://github.com/rustls/rustls/pull/1320 that makes it less easy to create a ServerCertVerifier that skips domain name verification. 

I chose not to call the new public functions from the WebPkiVerifier impl block because they need to parse the Certificate to webpki::EndEntity. If I did call the new functions, then the cert would be parsed twice.

### DIFF
--- a/rustls/src/key.rs
+++ b/rustls/src/key.rs
@@ -100,7 +100,7 @@ impl fmt::Debug for Certificate {
 }
 
 /// wrapper around internal representation of a parsed certificate. This is used in order to avoid parsing twice when specifying custom verification
-#[allow(unreachable_pub)]
+#[cfg_attr(not(feature = "dangerous_configuration"), allow(unreachable_pub))]
 #[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
 pub struct ParsedCertificate<'a>(pub(crate) webpki::EndEntityCert<'a>);
 
--- a/rustls/src/key.rs
+++ b/rustls/src/key.rs
@@ -1,5 +1,7 @@
 use std::fmt;
 
+use crate::Error;
+
 /// This type contains a private key by value.
 ///
 /// The private key must be DER-encoded ASN.1 in either
@@ -94,6 +96,20 @@ impl fmt::Debug for Certificate {
         f.debug_tuple("Certificate")
             .field(&BsDebug(&self.0))
             .finish()
+    }
+}
+
+/// wrapper around internal representation of a parsed certificate. This is used in order to avoid parsing twice when specifying custom verification
+#[allow(unreachable_pub)]
+#[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
+pub struct ParsedCertificate<'a>(pub(crate) webpki::EndEntityCert<'a>);
+
+impl<'a> TryFrom<&'a Certificate> for ParsedCertificate<'a> {
+    type Error = Error;
+    fn try_from(value: &'a Certificate) -> Result<ParsedCertificate<'a>, Self::Error> {
+        webpki::EndEntityCert::try_from(value.0.as_ref())
+            .map_err(crate::verify::pki_error)
+            .map(ParsedCertificate)
     }
 }
 
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -313,7 +313,8 @@ impl fmt::Debug for dyn ClientCertVerifier {
 /// `intermediates` contains all certificates other than `end_entity` that
 /// were sent as part of the server's [Certificate] message. It is in the
 /// same order that the server sent them and may be empty.
-#[allow(unreachable_pub, dead_code)]
+#[allow(dead_code)]
+#[cfg_attr(not(feature = "dangerous_configuration"), allow(unreachable_pub))]
 #[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
 pub fn verify_server_cert_signed_by_trust_anchor(
     cert: &ParsedCertificate,
@@ -339,7 +340,7 @@ pub fn verify_server_cert_signed_by_trust_anchor(
 /// Verify that the `end_entity` has a name or alternative name matching the `server_name`
 /// note: this only verifies the name and should be used in conjuction with more verification
 /// like [verify_server_cert_signed_by_trust_anchor]
-#[allow(unreachable_pub, dead_code)]
+#[cfg_attr(not(feature = "dangerous_configuration"), allow(unreachable_pub))]
 #[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
 pub fn verify_server_name(cert: &ParsedCertificate, server_name: &ServerName) -> Result<(), Error> {
     match server_name {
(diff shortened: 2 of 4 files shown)


cpu:
@Fritiofhedstrom It looks like your branch doesn't build. Could you take a look at fixing that (maybe also rebase on main)?

Fritiofhedstrom:
rebased on main and adapted to new DnsName struct

rustls/tests/server_cert_verifier.rs
```diff
@@ -7,9 +7,8 @@ use crate::common::{
     do_handshake, do_handshake_until_both_error, make_client_config_with_versions,
     make_pair_for_arc_configs, make_server_config, ErrorFromPeer, ALL_KEY_TYPES,
 };
-use rustls::client::{
-    HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier, WebPkiVerifier,
-};
+use rustls::client::WebPkiVerifier;
```
djc, line 10: This seems unrelated? I'd rather keep all imports from the same module on the same line.

rustls/src/verify.rs
```diff
@@ -338,15 +391,13 @@ impl ServerCertVerifier for WebPkiVerifier {
...
-        let cert = cert
-            .verify_is_valid_tls_server_cert(
-                SUPPORTED_SIG_ALGS,
-                &webpki::TlsServerTrustAnchors(&trustroots),
-                &chain,
-                webpki_now,
-            )
-            .map_err(pki_error)
-            .map(|_| cert)?;
+        cert.verify_is_valid_tls_server_cert(
```
djc, line 394: I feel like both this and the server name check should use the newly created functions, to avoid us having to keep them in sync manually.

rustls/src/verify.rs
```diff
@@ -321,6 +321,59 @@ impl fmt::Debug for dyn ClientCertVerifier {
...
 }
 
+/// Verify that the end-entity certificate `end_entity` is valid and chains to at least one of the [OwnedTrustAnchor] in the `roots` [RootCertStore].
+///
+/// `intermediates` contains all certificates other than `end_entity` that
+/// were sent as part of the server's [Certificate] message. It is in the
+/// same order that the server sent them and may be empty.
+#[allow(unreachable_pub, dead_code)]
+#[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
+pub fn verify_signed_by_trust_anchor(
```
cpu, line 331: I think this name/comment might be masking one important gotcha: `webpki::EndEntity.verify_is_valid_tls_server_cert` also asserts the cert has the `id-kp-serverAuth` extended key usage. E.g. this function might not be what you want if you were trying to verify that a client certificate was signed by a particular trust anchor.

I think at minimum we should call that out in the docstring, but it might be prudent to name this something like `verify_server_cert_signed_by_trust_anchor`.

rustls/src/key.rs
```diff
@@ -97,6 +99,28 @@ impl fmt::Debug for Certificate {
...
+    type Error = Error;
+    fn try_from(value: &'a Certificate) -> Result<ParsedCertificate<'a>, Self::Error> {
+        webpki::EndEntityCert::try_from(value.0.as_ref())
+            .map_err(crate::verify::pki_error)
+            .map(ParsedCertificate)
+    }
+}
+
+impl<'a> Deref for ParsedCertificate<'a> {
+    type Target = webpki::EndEntityCert<'a>;
```
djc, line 117: So one thing we've been trying to avoid is having webpki types and traits exposed through rustls' public API -- so I don't think this is the right solution here.

Fritiofhedstrom: Ok, i removed this impl and made the inner struct pub(crate)

rustls/src/verify.rs
```diff
@@ -321,6 +321,70 @@ impl fmt::Debug for dyn ClientCertVerifier {
...
+/// same order that the server sent them and may be empty.
+#[allow(unreachable_pub, dead_code)]
+#[cfg_attr(docsrs, doc(cfg(feature = "dangerous_configuration")))]
+pub fn verify_server_cert_signed_by_trust_anchor(
+    cert: &ParsedCertificate,
+    roots: &RootCertStore,
+    intermediates: &[Certificate],
+    now: SystemTime,
+) -> Result<(), Error> {
+    let chain: Vec<&[u8]> = intermediates
```
djc, line 338: Could we reuse `prepare()` here, maybe by changing its API a bit to fit better?

Fritiofhedstrom: I split it to seperate functions for each argument

djc (approved):
This seems like it might make life a little easier for downstream verifier users while keeping things more or less the same for everyone else -- seems okay? @ctz thoughts?

rustls/src/key.rs
```diff
@@ -97,6 +99,20 @@ impl fmt::Debug for Certificate {
     }
 }
 
+/// wrapper around internal representation of a parsed certificate. This is used in order to avoid parsing twice when specifying custom verification
+#[allow(unreachable_pub)]
```
cpu, line 103: I think we can tighten up this `allow` to only be in-effect when not using `dangerous_configuration`: 
```suggestion
#[cfg_attr(not(feature = "dangerous_configuration"), allow(unreachable_pub))]
```

rustls/src/verify.rs
```diff
@@ -321,6 +321,64 @@ impl fmt::Debug for dyn ClientCertVerifier {
     }
 }
 
+/// Verify that the end-entity certificate `end_entity` is a valid server cert
+/// and chains to at least one of the [OwnedTrustAnchor] in the `roots` [RootCertStore].
+///
+/// `intermediates` contains all certificates other than `end_entity` that
+/// were sent as part of the server's [Certificate] message. It is in the
+/// same order that the server sent them and may be empty.
+#[allow(unreachable_pub, dead_code)]
```
cpu, line 330: similar suggestion here: I think the `unreachable_pub` allow exception should be specific to `not(feature = "dangerous_configuration")`:

```suggestion
#[allow(dead_code)]
#[cfg_attr(not(feature = "dangerous_configuration"), allow(unreachable_pub))]
```

rustls/src/verify.rs
```diff
@@ -321,6 +321,64 @@ impl fmt::Debug for dyn ClientCertVerifier {
...
+            webpki_now,
+        )
+        .map_err(pki_error)
+        .map(|_| ())
+}
+
+/// Verify that the `end_entity` has a name or alternative name matching the `server_name`
+/// note: this only verifies the name and should be used in conjuction with more verification
+/// like [verify_server_cert_signed_by_trust_anchor]
+#[allow(unreachable_pub, dead_code)]
```
cpu, line 356: ditto here RE: `unreachable_pub`.

cpu (approved):
This looks like a fair compromise to me. Thanks for putting in the work :-) I think we should hold merge on Ctz's opinion but from my perspective I'm +1.

I left a couple small nits to consider RE: tightening up the `unreachable_pub` allows but I don't think that needs to be blocking feedback.

Could you also rebase to resolve conflicts? Thanks!

jadamcrain:
I like this too. It allows us to use rustls library code for the chain validation, but custom code for the name verification.

This will allow us to shrink our custom verifiers a bit and depend on the default Rustls chain validation piece.

ctz (approved):
Thanks! I think this is a reasonable compromise.

Fritiofhedstrom:
Thank you!

PR #1326, opened 2023-06-26
client certificate revocation checking support.

cpu:
## Description

This branch builds on the webpki CRL support added in v0.101.0 to offer client certificate revocation checking with the Rustls `AllowAnyAuthenticatedClient` and `AllowAnyAnonymousOrAuthenticatedClient` implementations of the `ClientCertVerifier` trait.

This replaces https://github.com/rustls/rustls/pull/1322:

* We're no longer parsing DER -> CRLs for each client cert validation. In this iteration we parse CRLs only once, when added, holding owned representations of the parsed CRLs.
* We provide a builder-like `with_crls` fn on the verifiers to optionally provide CRLs.
* The error handling has been revisited to provide more specificity around CRL related errors, particularly when parsing CRLs.
* Test CRL generation and integration test coverage have been added.

### deps: update rustls-webpki.

This commit updates the webpki dependency of rustls to v0.101.0, the release that adds CRL support.

### verify: AllowAnyAuthenticatedClient CRLs.

This commit introduces a `crls: Vec<webpki::OwnedCertRevocationList>>` field to the `AllowAnyAuthenticatedClient` implementation of `ClientCertVerifier` in preparation of supporting revocation checking of client certs with CRLs.

Each impl gets a builder-style `with_crls` method that can be used to augment the verifier with CRLs to use for client cert revocation status checking.

The constructed cert revocation lists are provided to webpki at cert verification time to use for revocation lookup.

The `tlsserver-mio` example is updated with a `--crl` flag to the that can be used to optionally provide a series of DER encoded CRLs to use for client certificate revocation checking.

### test-ca: generate demo CRLs for client certs.
This commit updates the `build-a-pki.sh` script to generate example certificate revocation lists (CRLs) that mark each of the client certificates as revoked. These can be used by server tests to ensure CRL validation works as expected.

The process of generating CRLs using `openssl` is... well... not great...

It can't be done without using `openssl ca`, which in turn requires using an `openssl.cnf` with all the associated warts. I've done my best to create the absolute minimum configuration that can be used for our purposes.

Using `openssl ca` also requires writing some intermediate state. The script is updated to create/delete this state through the process of generating the CRLs. This should be sufficient for our needs. Blech.

### test-ca: regenerate certs, keys, crls.
Generated with:

```
cd test-ca
./build-a-pki.sh
git add .
```
### tests: integration test for revoked client certs.
This commit introduces `api.rs` integration tests that verify a server configured with client authentication (mandatory or optional), and a CRL that specifies a client cert as revoked, will correctly return a revoked error for that client cert.

In addition to the test CRL data generated in a previous commit this work requires `rustls-pemfile` >= 1.0.3 to use unreleased CRL support in the `rustls-pemfile` crate. The rustls dev-dependency and the rustls example dependency on this crate are adjusted accordingly.

### docs: update README changelog for 0.21.3.
Includes:

* CRL feature for the client verifiers.
* Certificate validation helpers when using `dangerous_configuration`
  feature.


rustls/src/verify.rs
```diff
@@ -642,6 +672,8 @@ fn pki_error(error: webpki::Error) -> Error {
         CertExpired | InvalidCertValidity => CertificateError::Expired.into(),
         UnknownIssuer => CertificateError::UnknownIssuer.into(),
         CertNotValidForName => CertificateError::NotValidForName.into(),
+        CertRevoked => CertificateError::Revoked.into(),
+        IssuerNotCrlSigner => CrlError::InvalidIssuer.into(),
 
         InvalidSignatureForPublicKey
         | UnsupportedSignatureAlgorithm
```
cpu, line 679: Notably we can emit these errors when consulting a CRL for revocation information for a cert that has an issuer matching the CRL's issuer.

This is a bit suboptimal. For example, it might be confusing to have a `CertificateError::InvalidSignatureForPublicKey` be produced if it was the **CRL signature** that was invalid.

Perhaps that means on the webpki side we should add `InvalidCrlSignatureForPublicKey` and `UnsupportedCrlSignatureAlgorithm` errors that could be caught here and turned into `CrlError::xxxx` variants? WDYT?

rustls/tests/api.rs
```diff
@@ -479,7 +479,9 @@ fn server_allow_any_anonymous_or_authenticated_client() {
     let kt = KeyType::Rsa;
     for client_cert_chain in [None, Some(kt.get_client_chain())].iter() {
         let client_auth_roots = get_client_root_store(kt);
-        let client_auth = AllowAnyAnonymousOrAuthenticatedClient::new(client_auth_roots);
+        // TODO(@cpu): allow building test config with CRLs.
+        let client_auth =
+            AllowAnyAnonymousOrAuthenticatedClient::new(client_auth_roots, Vec::default()).unwrap();
```
cpu, line 484: Looking ahead to test coverage, how would folks like to see test CRL generation handled?

I can think of a few options:
1. We follow the `webpki` lead and provide a Python script using `cryptography` to generate test CRLs the rust code can use and check those in.
2. We use `rcgen` to generate test CRLs with rust at runtime, taking a dev-dependency on `rcgen`.
3. We use `openssl` (maybe with a small helper script) to generate test CRLs and check those in.
4. ???

I think option 2 is the nicest, but requires a yet-to-be-merged or released branch of `rcgen` that might hold up this work :-/

Option 3 is probably my second favourite since there are already some test resources being generated this way.

Option 1 is crummy. We'd like to move away from this in `webpki` and so it doesn't seem forward looking to add more Python here.

rustls/src/verify.rs
```diff
@@ -570,11 +592,18 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
     ) -> Result<ClientCertVerified, Error> {
         let (cert, chain, trustroots) = prepare(end_entity, intermediates, &self.roots)?;
         let now = webpki::Time::try_from(now).map_err(|_| Error::FailedToGetCurrentTime)?;
+
+        let mut crls: Vec<&dyn webpki::CertRevocationList> = Vec::new();
+        for owned_crl in &self.crls {
+            crls.push(owned_crl)
+        }
```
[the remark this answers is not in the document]
ctz: `self.crls.iter().collect()` yields `Vec<&webpki::CertRevocationList>`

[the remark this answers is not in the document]
cpu: I don't think it's that simple (or I'm missing something). When I try:
```
        let mut crls: Vec<&dyn webpki::CertRevocationList> = self.crls.iter().collect();
```

<details>
<summary> I get this error:</summary>

```
error[E0277]: a value of type `Vec<&dyn webpki::CertRevocationList>` cannot be built from an iterator over elements of type `&OwnedCertRevocationList`
   --> rustls/src/verify.rs:596:79
    |
596 |         let mut crls: Vec<&dyn webpki::CertRevocationList> = self.crls.iter().collect();
    |                                                                               ^^^^^^^ value of type `Vec<&dyn webpki::CertRevocationList>` cannot be built from `std::iter::Iterator<Item=&OwnedCertRevocationList>`
    |
    = help: the trait `FromIterator<&OwnedCertRevocationList>` is not implemented for `Vec<&dyn webpki::CertRevocationList>`
    = help: the trait `FromIterator<T>` is implemented for `Vec<T>`
note: the method call chain might not have had the expected associated types
   --> rustls/src/verify.rs:596:72
    |
596 |         let mut crls: Vec<&dyn webpki::CertRevocationList> = self.crls.iter().collect();
    |                                                              --------- ^^^^^^ `Iterator::Item` is `&OwnedCertRevocationList` here
    |                                                              |
    |                                                              this expression has type `Vec<OwnedCertRevocationList>`
note: required by a bound in `collect`
   --> /build/rustc-1.69.0-src/library/core/src/iter/traits/iterator.rs:1887:5

```
</details>

rustls/tests/api.rs
djc: Assuming your rcgen branch will be merged eventually (which seems like a reasonable assumption, assuming you haven't gotten pushback on the idea of merging something like it) I'd be okay taking on a temporary dependency on a Git commit from your branch. Given this is just a dev-dep, I don't think it should hold up publishing to crates.io?

rustls/src/verify.rs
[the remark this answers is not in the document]
djc: We can just do the allocation for now, though I'd like to take a shot at weeding it out once everything is merged -- I feel like it's possible without too many gymnastics.

(If the combinators are more concise, I'd just silence the clippy lint.)

rustls/tests/api.rs
ctz: I think the goals of this should be:

# tests are generated by an independent, quality implementation
# there's some chance the test data reflect how we expect people to use the feature
# the tests are as deterministic and fast as reasonably possible

I'm not sure rcgen (being backed by ring apart from anything else) would be independent or common enough?

rustls/src/verify.rs
```diff
@@ -522,26 +522,48 @@ fn prepare<'a, 'b>(
     Ok((cert, intermediates, trustroots))
 }
 
+/// A DER encoded Certificate Revocation List (CRL).
+pub struct CertRevocationList(pub Vec<u8>);
```
ctz, line 526: Perhaps `UnparsedCertRevocationList`? Or something else to designate that nothing is checking the contents of this.

rustls/tests/api.rs
djc: It's fair that our own implementation (even though reviewed by a separate maintainer) wouldn't be very independent, and I can see how that might be a reason to go with Python or OpenSSL instead. I wonder if the Rust openssl crate would expose support for this kind of stuff.

rustls/src/verify.rs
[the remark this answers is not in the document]
cpu: Both sound good to me.

rustls/tests/api.rs
cpu: I was able to use openssl-rs to parse CRLs in [an rcgen test](https://github.com/est31/rcgen/blob/3a8535a0724963e62dc3828ad8c73c957396a883/tests/openssl.rs#L391-L429) but I don't see any API surface for *generating* CRLs. I think I would probably have to use the `openssl` CLI (:nauseated_face:), or get into the weeds with new code for `openssl-sys` and `openssl-rs`.

I think I'd lightly prefer a small standalone `.py` file over bash+openssl if `rcgen` isn't independent enough. While tempting to get into the `openssl-rs` stuff I think I'd prefer not to block this PR on making new contributions to two codebases I haven't worked in before. That seems like better for a rainy day follow-up.

djc: I'd be okay with that.

rustls/src/verify.rs
cpu: Renamed.

rustls/src/verify.rs
```diff
@@ -651,6 +682,26 @@ fn pki_error(error: webpki::Error) -> Error {
     }
 }
 
+// Map errors that may occur when _parsing_ a Certificate Revocation List (CRL)
+// Errors that may occur as a result of _using_ a CRL for making a trust decision are handled
+// by pki_error.
+fn crl_parse_error(error: webpki::Error) -> Error {
```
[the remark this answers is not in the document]
cpu: Done.

rustls/src/error.rs
```diff
@@ -381,6 +384,51 @@ impl From<CertificateError> for Error {
...
+
+    /// The CRL contained an invalid CRL number, or an invalid revoked certifiate
+    /// serial number.
+    InvalidCrl,
+
+    /// The CRL issuer does not specify the cRLSign key usage.
+    InvalidIssuer,
+
+    /// The CRL is not correctly encoded.
+    BadEncoding,
```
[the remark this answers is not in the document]
cpu: Renamed.

rustls/src/error.rs
```diff
@@ -381,6 +384,51 @@ impl From<CertificateError> for Error {
     }
 }
 
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+/// The ways in which a certificate revocation list (CRL) can be invalid.
+pub enum CrlError {
+    /// The CRL has an unsupported version, contains a certificate with an unsupported revocation
```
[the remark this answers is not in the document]
cpu: I added some more text and a 5280 reference.

rustls/src/verify.rs
cpu: Done.

rustls/src/verify.rs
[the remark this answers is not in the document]
cpu: Went back to the combinators w/ a clippy allow.

rustls/tests/api.rs
cpu: > I think I'd lightly prefer a small standalone .py file over bash+openssl if rcgen isn't independent enough.

Digging into the existing API tests more I've come around to updating the `build-a-pki.sh` script that's already tooled around `openssl`. I think it'll be a pretty minimal diff and will fit into the existing approach better than inventing something new and retrofitting.

cpu: `openssl` never makes life easy, but I managed to get something workable going without too much fuss: https://github.com/rustls/rustls/pull/1326/commits/590c9c29fa1af1b15b83b2bc820e1756f839a4d5

Bonus fact: the default `openssl.cnf` configuration still [carries a workaround for Netscape Communicator](https://github.com/openssl/openssl/blob/8b7d5ea7dd602eb7c2c4bc5ad45489dc5fc711f6/apps/openssl.cnf#L107-L108) (last released ~Aug 2002) :roll_eyes:

cpu:
> cpu added 3 commits 51 minutes ago

* test-ca generator script updates to also produce CRLs.
* re-generated test-ca files.
* integration test using the test CRLs.

PR description updated. Also required a 2nd Cargo patch for rustls-pemfile until we cut a release w/ CRL support.

test-ca/build-a-pki.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
cpu, line 1: I made this change because my IDE was warning me that apparently sh + POSIX doesn't standardize the -n flag for echo. I need echo -n because writing a \n to index.txt causes openssl ca to barf parsing the db state.

In general it seemed fine to assume bash here.

test-ca/build-a-pki.sh
```diff
@@ -174,3 +201,7 @@ for kt in rsa ecdsa eddsa ; do
 
   openssl asn1parse -in $kt/ca.cert -out $kt/ca.der > /dev/null
 done
+
+# Tidy up openssl CA state.
+rm index.txt* || true
+rm crlnumber* || true
```
cpu, line 207: using wildcards here because openssl ca writes a variety of files, e.g. crlnumber.old, index.txt.attr, etc.

cpu:
>  cpu force-pushed the cpu-crl-support branch from 3346e76 to 2d75883 

Just swapping the `Cargo.toml` patch from my pemfile fork/branch to `main` of `rustls/pemfile`.

ctz:
Please note the breaking change[^1] in the README.md changelog.

[^1]: which is limited to, I think, the additional parameter on `AllowAnyAnonymousOrAuthenticatedClient::new`

cpu:
> which is limited to, I think, the additional parameter on AllowAnyAnonymousOrAuthenticatedClient::new leftwards_arrow_with_hook

In addition to the `AllowAnyAnonymousOrAuthenticatedClient` and `AllowAnyAuthenticatedClient` constructors getting a new arg, they're also now fallible and return a `Result` since CRL parsing can error.

I will note both aspects in the changelog :+1:

cpu:
> cpu force-pushed the cpu-crl-support branch from 344d566 to 304c688 

Rebased to resolve conflicts.

cpu:
> cpu force-pushed the cpu-crl-support branch from 13cdbe9 to 14ed8be 

Rebased, updated the webpki patch to use `main` instead of my fork.

One other substantial change: based on experience integrating this work in `rustls-ffi`, and more thought on the API design in general, I've reworked how CRLs are provided to the `AllowAnyAnonymousOrAuthenticatedClient` and `AllowAnyAuthenticatedClient` impls of `ClientCertVerifier`.

Initially I provided `Vec<UnparsedCertRevocationList>` as a new constructor arg. That's yucky in practice for ~3 reasons (also outlined in the updated PR desc):

1. It makes the constructor fallible since we have to handle CRL parse errors. It's unfortunate to force that complexity onto users that are likely not going to provide any CRLs.
2. It's a breaking API change, where-as adding a new fn to do this isn't (since it's defined on the concrete types, not the trait).
3. It's a pain to work with a `Vec` of `UnparsedCertRevocationList` (e.g. a `Vec<Vec<u8>>` on the C-side, as opposed to being able to use `add_crl` with a single `Vec<u8>`. 

Let me know what you think. I can go back to the previous approach if folks dislike this variation.

rustls/src/error.rs
```diff
@@ -377,6 +380,82 @@ impl From<CertificateError> for Error {
...
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+/// The ways in which a certificate revocation list (CRL) can be invalid.
+pub enum CrlError {
+    /// The CRL had a bad, or unsupported signature from its issuer.
+    BadSignature,
+
+    /// The CRL contained an invalid CRL number, or an invalid revoked certificate
+    /// serial number.
+    InvalidCrl,
```
djc, line 392: It feels odd that the comment has more specifics than the variant name. In general I wonder if it's useful that we converge the very specific webpki errors to a smaller/more general set here. It certainly makes for less code in rustls and maybe avoids overwhelming users with overly specific errors, but on the other hand throwing away the detail from the webpki errors seems unfortunate, maybe?

rustls/src/error.rs
```diff
@@ -377,6 +380,82 @@ impl From<CertificateError> for Error {
...
+
+    /// The CRL issuer does not specify the cRLSign key usage.
+    InvalidIssuer,
```
djc, line 395: Maybe `IssuerInvalidForCrl`?

examples/src/bin/tlsserver-mio.rs
```diff
@@ -551,15 +555,29 @@ fn load_ocsp(filename: &Option<String>) -> Vec<u8> {
     ret
 }
 
+fn load_crls(filenames: &Vec<String>) -> Vec<UnparsedCertRevocationList> {
+    let mut ret = Vec::new();
```
djc, line 559: Naming nit: maybe name this `crls`?

examples/src/bin/tlsserver-mio.rs
```diff
@@ -551,15 +555,29 @@ fn load_ocsp(filename: &Option<String>) -> Vec<u8> {
...
 fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
     let client_auth = if args.flag_auth.is_some() {
         let roots = load_certs(args.flag_auth.as_ref().unwrap());
         let mut client_auth_roots = RootCertStore::empty();
         for root in roots {
             client_auth_roots.add(&root).unwrap();
         }
-        // TODO(@cpu): Parse CRLs from Args.
-        let crls = Vec::default();
+        let crls = load_crls(&args.flag_crl);
```
djc, line 580: Can we avoid this indirection in the commit history? It doesn't seem useful to change the examples twice like this.

test-ca/build-a-pki.sh
djc, line 207: Nit: add a trailing newline?

rustls/tests/common/mod.rs
```diff
@@ -218,6 +221,19 @@ impl KeyType {
             .collect()
     }
 
+    pub fn get_client_crl(&self) -> UnparsedCertRevocationList {
```
djc, line 224: Style nit: avoid `get_` prefixes.

rustls/tests/common/mod.rs
```diff
@@ -218,6 +221,19 @@ impl KeyType {
...
+        UnparsedCertRevocationList(
+            rustls_pemfile::crls(&mut io::BufReader::new(
+                self.bytes_for("client.revoked.crl.pem"),
+            ))
+            .unwrap()
+            .iter()
```
djc, line 230: If we `into_iter()` here instead, can we avoid the `clone()` at the end?

djc:
I commented on this before seeing this discussion:

> One other substantial change: based on experience integrating this work in `rustls-ffi`, and more thought on the API design in general, I've reworked how CRLs are provided to the `AllowAnyAnonymousOrAuthenticatedClient` and `AllowAnyAuthenticatedClient` impls of `ClientCertVerifier`.
> 
> Initially I provided `Vec<UnparsedCertRevocationList>` as a new constructor arg. That's yucky in practice for ~3 reasons (also outlined in the updated PR desc):
> 
>     1. It makes the constructor fallible since we have to handle CRL parse errors. It's unfortunate to force that complexity onto users that are likely not going to provide any CRLs.
> 
>     2. It's a breaking API change, where-as adding a new fn to do this isn't (since it's defined on the concrete types, not the trait).

Not sure this is a great argument? One would argue that it might be useful for downstream consumers to notice they now have the option to provide a CRL.

>     3. It's a pain to work with a `Vec` of `UnparsedCertRevocationList` (e.g. a `Vec<Vec<u8>>` on the C-side, as opposed to being able to use `add_crl` with a single `Vec<u8>`.

Why is it a pain?

Maybe the right path forward is to provide a `with_crls(self, impl Iterator<Item = UnparsedCertRevocationList>) -> Result<Self, CrlError>` method on the `ClientCertVerifier` impls?

rustls/src/error.rs
cpu: > In general I wonder if it's useful that we converge the very specific webpki errors to a smaller/more general set here.

That's what I was trying to do here: squashing the invalid CRL number and invalid serial number errors into a more general "InvalidCrl" error.

Would it be less confusing if I simplified the comment to not mention the specifics?

cpu:
> Maybe the right path forward is to provide a with_crls(self, impl Iterator<Item = UnparsedCertRevocationList>) -> Result<Self, CrlError> method on the ClientCertVerifier impls? That would avoid breaking the API and making the constructor fallible while still having a more ergonomic builder-like API. Not sure how it would impact the -ffi side.

I'll give that a shot. I think it's a nicer API.

rustls/src/error.rs
djc: That would be more consistent at least. Still not sure why we don't propagate the full level of detail that webpki gives us to the rustls consumers.

rustls/src/verify.rs
```diff
@@ -439,10 +454,18 @@ impl AllowAnyAuthenticatedClient {
...
                 .map(|r| r.subject().clone())
                 .collect(),
+            crls: Vec::new(),
             roots,
         }
     }
 
+    /// Add a DER encoded unparsed certificate revocation list (CRL) to the list of CRLs used
+    /// to validate client certificates.
+    pub fn add_crl(&mut self, crl: UnparsedCertRevocationList) -> Result<(), Error> {
```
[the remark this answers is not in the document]
cpu: Discussed elsewhere.

examples/src/bin/tlsserver-mio.rs
cpu: Sure, I'll fold the commit that adds the command line flag processing into this commit.

rustls/src/error.rs
```diff
@@ -66,6 +66,9 @@ pub enum Error {
     /// implementation.
     InvalidCertificate(CertificateError),
 
+    /// A provided certificate revocation list (CRL) was invalid.
+    InvalidCrl(CrlError),
```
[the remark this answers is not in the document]
cpu: Done.

rustls/src/error.rs
cpu: > Still not sure why we don't propagate the full level of detail that webpki gives us to the rustls consumers.

I switched to propagating the more specific errors.

examples/src/bin/tlsserver-mio.rs
cpu: Done.

test-ca/build-a-pki.sh
cpu: Done.

rustls/tests/common/mod.rs
cpu: Done, but there are other instances of this anti-pattern I was matching. I will tidy those separately.

rustls/tests/common/mod.rs
cpu: Good call, fixed.

rustls/src/error.rs
cpu: Done.

cpu:
> codecov/patch Failing after 1s — 35.82% of diff hit (target 96.32%)

I'll see about shoring up the test coverage. Otherwise I think this is ready for review with all existing review feedback addressed.

With the webpki 0.101.0 release being available I removed the cargo patch and moved this out of draft status.

ctz (approved):
lgtm! for the coverage gaps i'd consider some unit-test-style direct exercise of the error types.

cpu:
> codecov/patch — 100.00% of diff hit (target 96.32%)

Some more unit tests did the trick :+1:

examples/src/bin/tlsserver-mio.rs
djc: I meant the `ret`. Did you rename something else? Actually, I think you could now write this as an iterator chain? `filenames.into_iter().map(|name| UnparsedCertRevocationList(fs::read(filename).expect("")).collect()`?

ctz (approved):

rustls/src/verify.rs
```diff
@@ -559,10 +576,24 @@ impl AllowAnyAuthenticatedClient {
...
         }
     }
 
+    /// Update the verifier to validate client certificates against the provided DER format
+    /// unparsed certificate revocation lists (CRLs).
+    pub fn with_crls(
+        self,
+        crls: impl IntoIterator<Item = UnparsedCertRevocationList>,
+    ) -> Result<Self, CertRevocationListError> {
+        let crls = crls
```
djc, line 590: Style nit: not much point in binding this?

djc (approved):

examples/src/bin/tlsserver-mio.rs
cpu: > I meant the ret. Did you rename something else?

Yeah I misunderstood and renamed `load_crls` to `crls`. Sorry about that. I restored the original name to match the other `load_xx` fns nearby.

> I think you could now write this as an iterator chain?

Done, but it ended up a little bit messier than suggested because of needing a mutable buf for reading the DER into. Also clippy wanted `iter` instead of `into_iter` and the `filenames` arg to be `&[String]` instead of `Vec<String>`:

```rust
fn load_crls(filenames: &[String]) -> Vec<UnparsedCertRevocationList> {
    filenames
        .iter()
        .map(|filename| {
            let mut der = Vec::new();
            fs::File::open(filename)
                .expect("cannot open CRL file")
                .read_to_end(&mut der)
                .unwrap();
            UnparsedCertRevocationList(der)
        })
        .collect()
}
```

rustls/src/verify.rs
cpu: Fixed. Thanks

PR #1328, opened 2023-06-28
Moved to a working host

xuganyu96:
Both `simpleclient` and `limitedclient` have been moved to communicate with `www.rust-lang.org` instead of `google.com` due to Google's HTTP server not implementing `close_notify` properly

### DIFF
--- a/examples/src/bin/limitedclient.rs
+++ b/examples/src/bin/limitedclient.rs
@@ -31,14 +31,14 @@ fn main() {
         .with_root_certificates(root_store)
         .with_no_client_auth();
 
-    let server_name = "google.com".try_into().unwrap();
+    let server_name = "www.rust-lang.org".try_into().unwrap();
     let mut conn = rustls::ClientConnection::new(Arc::new(config), server_name).unwrap();
-    let mut sock = TcpStream::connect("google.com:443").unwrap();
+    let mut sock = TcpStream::connect("www.rust-lang.org:443").unwrap();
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks!

cpu:
The changes to `simpleclient.rs` and `limitedclient.rs` seem fine, but I see that `simple_0rtt_client.rs` is also broken in its own unique way. It would be nice to fix all of the examples at once so we could add CI coverage.

The 0rtt demo is using `mesalink.io` as the target domain and that resolves NXDOMAIN presently, breaking the example. The MesaLock Linux homepage seems to be some kind of cryptocurrency nonsense now so I guess the original project is dead?

I think we should switch the 0rtt demo to using `www.rust-lang.org` as well. That host isn't configured to accept 0-rtt data, but I can't find a reliable server to use in its place and the demo is coded to fall-back to normal requests so at least using a valid domain name results in a real request.

Cloudflare's [blog post](https://blog.cloudflare.com/introducing-0-rtt/) introducing 0-rtt used a `0rtt.tls13.com` demo server, making requests to `https://0rtt.tls13.com/img.png` and peeking at returned headers but the in-browser demo doesn't work for me and trying to use `curl` or `simple_0rtt_client` I can only provoke a 404 response, no 0-rtt.

ctz:
Have added #1330 which addresses the simple_0rtt_client issues

PR #1330, opened 2023-06-29
simple_0rtt_client: fix it and clarify what it is demonstrating

ctz:
- mesalink.io doesn't exist now.  My website has cloudflare in front of it with 0-RTT enabled.
- add printed output showing what is happening, rather than two walls of raw HTTP responses.

The expected output is now:

```
* Sending first request:
  * Normal request sent
  * Server response: "HTTP/1.1 200 OK\r\n"
* Sending second request:
  * 0-RTT request sent
  * 0-RTT data accepted
  * Server response: "HTTP/1.1 200 OK\r\n"
```

### DIFF
--- a/examples/src/bin/simple_0rtt_client.rs
+++ b/examples/src/bin/simple_0rtt_client.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use std::io::{stdout, Read, Write};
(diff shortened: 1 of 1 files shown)


cpu (approved):
Thanks!

PR #1331, opened 2023-06-29
ci: run client examples in CI connect tests.

cpu:
We want to catch breakages in these client examples when they occur, but also don't want to run them during normal CI since they connect to external hosts and may occasionally flake.

This commit adds `simpleclient`, `limitedclient`, and `simple_0rtt_client` test runs to the `connect-tests.yml` CI configuration we run on a weekly basis.


ctz (approved):

djc:
Nice!

Issue #1332, opened 2023-06-29
Cannot work with BrainpoolP256r1 Key and Certificate


vbrandl:
I just tried executing the `tlsserver-mio` example using a BrainpoolP256r1 private key and certificate. The curve is listed in the [NamedCurve](https://docs.rs/rustls/latest/rustls/internal/msgs/enums/enum.NamedCurve.html#variant.brainpoolp256r1) enum.

When starting, the server panics:

```
$ cargo run -- --certs cert.pem --key key.pem http
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/tlsproxy --certs cert.pem --key key.pem http`
thread 'main' panicked at 'bad certificates/private key: General("invalid private key")', src/main.rs:598:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Is this a bug or are there no plans to support brainpool keys? I could observe this in multiple languages: Java dropped support for brainpool and also golang cannot use the same key.

Attached you can find the test key.

[key_and_cert.zip](https://github.com/rustls/rustls/files/11906457/key_and_cert.zip)

ctz:
Yes, we don't support any brainpool curves, or explicitly-defined curves. Though I'm aware they are mandated by a bunch of German national security standards, they are -- as I understand it -- not cryptographically better than the NIST curves.

djc:
Might be useful to note that #1319 should eventually make it possible to support Brainpool curves with rustls (although someone would have to write an implementation for it).

cpu:
> The curve is listed in the [NamedCurve](https://docs.rs/rustls/latest/rustls/internal/msgs/enums/enum.NamedCurve.html#variant.brainpoolp256r1) enum.

We've updated the Rustdoc's to help make it clearer that `NamedCurve` isn't the list of supported curves (https://github.com/rustls/rustls/pull/1333). Thanks for filing this issue!

djc:
With that, I think we can close this?

vbrandl:
Yeah, thanks for the pointer to the PR, I'll watch the progress on this one.
Also you are right. We must use brainpool due to German regulations...

PR #1333, opened 2023-06-29
enums: clarify NamedCurve isn't supported curves.

cpu:
To avoid any confusion (e.g. https://github.com/rustls/rustls/issues/1332) this commit clarifies the rustdoc comment on the `NamedCurve` enum to indicate that it has entries we don't support.

### DIFF
--- a/rustls/src/msgs/enums.rs
+++ b/rustls/src/msgs/enums.rs
@@ -139,6 +139,11 @@ enum_builder! {
     /// The `NamedCurve` TLS protocol enum.  Values in this enum are taken
     /// from the various RFCs covering TLS, and are listed by IANA.
     /// The `Unknown` item is used when processing unrecognised ordinals.
+    ///
+    /// This enum is used for recognizing elliptic curve parameters advertised
+    /// by a peer during a TLS handshake. It is **not** a list of curves that
+    /// Rustls supports. See [`crate::kx_group`] for the list of supported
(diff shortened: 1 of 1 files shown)


cpu:
@ctz WDYT? Have I over-simplified?

rustls/src/msgs/enums.rs
```diff
@@ -139,6 +139,11 @@ enum_builder! {
     /// The `NamedCurve` TLS protocol enum.  Values in this enum are taken
     /// from the various RFCs covering TLS, and are listed by IANA.
     /// The `Unknown` item is used when processing unrecognised ordinals.
+    ///
+    /// This enum is used for recognizing elliptic curve parameters advertised
+    /// by a peer during a TLS handshake. It is **not** a list of curves that
+    /// Rustls supports. Rustls only supports ECDSA NISTP256, ECDSA NISTP384
+    /// and Ed25519.
```
ctz, line 146: Maybe link to https://docs.rs/rustls/latest/rustls/kx_group/index.html (however that is achieved, is it `[crate::kx_group]`?) to avoid this getting out of date?

djc (approved):

rustls/src/msgs/enums.rs
cpu: Good idea, thanks. Fixed up.

ctz (approved):

PR #1334, opened 2023-06-30
Enable unit-test-like benchmarks

ctz:
Run these locally using:

```
$ RUSTFLAGS='--cfg=bench' cargo +nightly bench
```

The method of introducing a new `bench` build configuration is explained in https://github.com/RCasatta/qr_code/commit/c037ab97bfc55dd74e6fbe86581472e1f28b7773 and the odd use of `extern crate` is explained in https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html#an-exception

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,7 +135,7 @@ jobs:
           done
 
   benchmarks:
-    name: Smoke-test benchmarks
+    name: Run benchmarks
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout sources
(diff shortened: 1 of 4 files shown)


cpu (approved):
Thanks for linking to those extra resources from the PR description :+1: 

It's really nice to be able to sprinkle in benchmarks alongside the code under bench.

djc:
Might be able to get this to work on stable with bencher?

ctz:
> Might be able to get this to work on stable with bencher?

https://github.com/bluss/bencher/issues/3 sounds like you have to go without unit tests to have it work.

Issue #1313, opened 2023-06-12
Randomise ClientHello extension ordering as client fingerprinting countermeasure


ctz:
https://chromestatus.com/feature/5124606246518784

GomesGoncalo:
Hi, I raised a PR for this, noticed there was a PR with some shortcomings and while this does not use rand as a dependency it doesn't fully address the other problems

PR #1336, opened 2023-07-01
feat: randomize ClientHello extensions

rushilmehra:
https://github.com/rustls/rustls/issues/1313

I doubt the solution will be this easy, but I'm opening this as a starting point for discussion. BoringSSL decided to implement this via a flag, and the randomization is applied per connection. This means that all ClientHellos sent on a connection will be consistent. Is this something we want to emulate? https://boringssl-review.googlesource.com/c/boringssl/+/48045 has more details

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -17,6 +17,7 @@ rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
 log = { version = "0.4.4", optional = true }
+rand = "0.8.5"
 ring = "0.16.20"
 webpki = { package = "rustls-webpki", version = "0.100.0", features = ["alloc", "std"] }
 
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -34,6 +34,9 @@ use crate::client::{tls13, ClientConfig, ServerName};
 use std::ops::Deref;
 use std::sync::Arc;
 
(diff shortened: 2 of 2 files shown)


ctz:
Thanks for this!

I think having each connection having a different extension order is great. But it needs to be consistent - at the moment I think the failing tests are due to the order differing between the initial `ClientHello` and subsequent retry (after a `HelloRetryRequest`). The extensions in the retry need to have the same value and order (with a small number of exceptions in 4.1.2 in RFC8446).

I also think we wouldn't want to take a dependency on `rand` for this.

rushilmehra:
Yeah in that case we'll want to emulate BoringSSL and pin some randomization scheme to the connection. I'll find a way to do this without the `rand` crate.

cpu:
@rushilmehra :wave: Is revisiting this PR based on Ctz's feedback still something you would be interested in tackling?

cpu:
I'm going to close this draft PR since it's not actively being worked on. It sounds like there are some larger design questions to figure out before diving into an implementation so I would propose creating an RFC issue to talk through those aspects as a good first step towards reviving this work when someone has the bandwidth to work on it.

Thanks for getting the ball rolling

rushilmehra:
Sounds good, thanks

ctz:
The rough route I see us going with this is via #1475

PR #1338, opened 2023-07-04
Improve coverage of ip address names

ctz:
Noticed in reviewing #1325 that we had some gaps in verify.rs for testing IP address support. This aims to address that.


cpu (approved):
Nice!

PR #1339, opened 2023-07-05
Take `Iterator` in `add_parsable_certificates()`

### LINKED ISSUE #1304: `Iterator` support for `RootCertStore::add_parsable_certificates()`

daxpedda:
I suggest we actually take `IntoIterator` here instead of `Iterator`, which on it's own probably doesn't do much, but combine it with `Item = impl AsRef<[u8]>`, which would require adding a generic but removes the lifetime, and it would unburden the user significantly, being able to pass in almost anything without any further modification.

WDYT?

Fixes #1304.

### DIFF
--- a/rustls/src/anchors.rs
+++ b/rustls/src/anchors.rs
@@ -124,7 +124,10 @@ impl RootCertStore {
     /// include ancient or syntactically invalid certificates.
     ///
     /// Returns the number of certificates added, and the number that were ignored.
-    pub fn add_parsable_certificates(&mut self, der_certs: &[impl AsRef<[u8]>]) -> (usize, usize) {
+    pub fn add_parsable_certificates<C: AsRef<[u8]>>(
+        &mut self,
+        der_certs: impl IntoIterator<Item = C>,
+    ) -> (usize, usize) {
         let mut valid_count = 0;
         let mut invalid_count = 0;
 
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ## Release history
 
+* Next release:
+  - `RootCertStore::add_parsable_certificates` now takes a
+    `impl IntoIterator<Item = impl AsRef<[u8]>>`.
 * Release 0.21.3 (2023-07-05)
   - Added `with_crls` function to `AllowAnyAuthenticatedClient` and
     `AllowAnyAnonymousOrAuthenticatedClient` client certificate verifiers to
--- a/examples/src/bin/tlsclient-mio.rs
+++ b/examples/src/bin/tlsclient-mio.rs
@@ -377,7 +377,7 @@ fn make_config(args: &Args) -> Arc<rustls::ClientConfig> {
 
         let certfile = fs::File::open(cafile).expect("Cannot open CA file");
         let mut reader = BufReader::new(certfile);
-        root_store.add_parsable_certificates(&rustls_pemfile::certs(&mut reader).unwrap());
+        root_store.add_parsable_certificates(rustls_pemfile::certs(&mut reader).unwrap());
     } else {
         root_store.add_server_trust_anchors(
             webpki_roots::TLS_SERVER_ROOTS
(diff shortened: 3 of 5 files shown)


daxpedda:
I showcased my proposal from OP here:
https://github.com/rustls/rustls/compare/main...daxpedda:rustls:add-certs-iter-2

I'm assuming I can ignore the code coverage failure here?

cpu:
> I showcased my proposal from OP here:

Using `IntoIterator` seems better to me but I think Ctz or Djc will be a better judge. Is the only hesitance you have with that approach the part about needing to introduce a generic parameter? I thought the `impl Iterator<...>` approach was also adding a generic parameter, just hidden behind syntactic sugar. (_Happy to be corrected if I'm misunderstanding something_)

> I'm assuming I can ignore the code coverage failure here?

I think so :+1:

daxpedda:
> I thought the `impl Iterator<...>` approach was also adding a generic parameter, just hidden behind syntactic sugar.

I'm not sure if that's also called a generic parameter, my only concern here is that `where C: ...` is harder to read then `impl ...`. This is a really minor concern, personally I wouldn't care, but I thought I would point it out.

cpu:
FWIW I think we're going to back out the breaking change queued in `main` (https://github.com/rustls/rustls/pull/1340) to make a semver compatible release first. sorry about the churn. I think we can still review this PR and have it land right after along with the restore of the SCT removal. There's other work in the pipeline that will merit a breaking release in the near-ish future.

djc:
I think it would make sense to do `IntoIterator` instead. I think that would also allow this change to be semver-compatible?

daxpedda:
> I think that would also allow this change to be semver-compatible?

Well that's a good point, I think that could work.

daxpedda:
Done!
I'm pretty sure this isn't a breaking change at this point.

rustls/src/anchors.rs
```diff
@@ -124,7 +124,13 @@ impl RootCertStore {
     /// include ancient or syntactically invalid certificates.
     ///
     /// Returns the number of certificates added, and the number that were ignored.
-    pub fn add_parsable_certificates(&mut self, der_certs: &[impl AsRef<[u8]>]) -> (usize, usize) {
+    pub fn add_parsable_certificates<C>(
+        &mut self,
+        der_certs: impl IntoIterator<Item = C>,
```
djc, line 129: Style: I think I would specify `C: AsRef<[u8]>` in the type argument declaration to avoid the added verbosity of the where clause.

djc (commented):
Did you verify that passing a slice still works?

djc (approved):

daxpedda:
> Did you verify that passing a slice still works?

Yes, I double-checked in the tests.

rustls/src/anchors.rs
daxpedda: Done.

ctz:
This looks good to me!

- please merge the two commits
- perhaps we could back out one or two of the test changes (ie, at least one call site uses a slice as previously, at least one uses a vec as this PR currently does) to show that both work. but please keep the change in `tlsclient-mio.rs` as it currently is, as i think the direct vec use is clearest in example code we expect people to read.

daxpedda:
Done.

ctz (approved):

cpu:
Nice!

cpu:
Just wanted to note we rolled this change into [0.21.4](https://github.com/rustls/rustls/releases/tag/v%2F0.21.4).

cpu:
> Just wanted to note we rolled this change into 0.21.4.

Unfortunately this was short lived :sweat: - Reverted in 0.21.5 to address https://github.com/rustls/rustls/issues/1352

PR #1345, opened 2023-07-06
client: `with_single_cert` -> `with_client_auth_cert`

cpu:
### client: with_single_cert -> with_client_auth_cert 

This commit renames the `ClientConfig` builder's `with_single_cert`  function to be called `with_client_auth_cert`. The old `with_single_cert` function is left as an alias for `with_client_auth_cert` and marked as deprecated to encourage users to switch to the new name.

I believe this offers better symmetry with the `with_no_client_auth` function that's used to disable client authentication, and more clearly conveys the purpose of this function is for providing a client authentication certificate.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -67,13 +67,29 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
     /// `key_der` is a DER-encoded RSA, ECDSA, or Ed25519 private key.
     ///
     /// This function fails if `key_der` is invalid.
-    pub fn with_single_cert(
+    pub fn with_client_auth_cert(
         self,
         cert_chain: Vec<key::Certificate>,
         key_der: key::PrivateKey,
     ) -> Result<ClientConfig, Error> {
         let resolver = handy::AlwaysResolvesClientCert::new(cert_chain, &key_der)?;
         Ok(self.with_client_cert_resolver(Arc::new(resolver)))
+    }
+
+    /// Sets a single certificate chain and matching private key for use
+    /// in client authentication.
+    ///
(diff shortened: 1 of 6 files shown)


ctz:
Like this change!

> I believe this offers better symmetry with the `with_no_client_auth` function that's used to disable client authentication

It's also closer to `with_client_cert_resolver` function

cpu:
> Like this change!

Cool :+1: I updated the PR desc, rebased on `main` and flipped the PR to ready for review.

cpu:
> Should we do this as a deprecation with alias in the current release range, or in the next, before removing it entirely?

Sure, that sounds friendlier. I will adjust.

rustls/src/client/builder.rs
```diff
@@ -76,6 +76,22 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
...
     }
 
+    /// Sets a single certificate chain and matching private key for use
+    /// in client authentication.
+    ///
+    /// `cert_chain` is a vector of DER-encoded certificates.
+    /// `key_der` is a DER-encoded RSA, ECDSA, or Ed25519 private key.
+    ///
+    /// This function fails if `key_der` is invalid.
+    #[deprecated(since = "0.21.3", note = "Use `with_client_auth_cert` instead")]
```
cpu, line 86: I marked this as "since 0.21.3" but perhaps it should be the _next_ release? 0.22.0?

djc: The first release in which it will be deprecated might be 0.21.4, then we can remove it in 0.22.

cpu: Switched to 0.21.4 and will pull this into https://github.com/rustls/rustls/pull/1348

djc (approved):

PR #1350, opened 2023-07-10
Extract traits under crypto use (part 1)

ctz:
This is a first third of #1319 that I think we can land now. There was some previous review of the final commit on #1319 -- this PR includes addressing those issues.


rustls/src/crypto.rs
```diff
@@ -1,9 +1,14 @@
 use crate::rand::GetRandomFailed;
+use crate::server::ProducesTickets;
 
+use ring::aead;
 use ring::rand::{SecureRandom, SystemRandom};
 
 /// Pluggable crypto galore.
 pub trait CryptoProvider: Send + Sync + 'static {
+    /// Build a ticket generator.
+    fn ticket_generator() -> Result<Box<dyn ProducesTickets>, GetRandomFailed>;
```
ctz: I think this is an extra indirection that is not needed, because it only serves to customise `rustls::Ticketer` . But that interface just indirects back through `CryptoProvider` having layered on a `TicketSwitcher` which should instead be a building block for concrete ticketers. 

Instead callers of `rustls::Ticketer::new` should just be calling `rustls::crypto::ring::Ticketer::new` or similar.

cpu (approved):
Thanks for pulling this out for review. I left a couple small nits but it looks good to me (_unsurprisingly, since most of the commits were things I authored or reviewed already!_ :sweat_smile: )

rustls/src/server/tls13.rs
```diff
@@ -487,10 +487,9 @@ mod client_hello {
     ) -> Result<KeyScheduleHandshake, Error> {
         let mut extensions = Vec::new();
 
-        // Prepare key exchange
+        // Prepare key exchange; the caller ascertained that the `share.group` is supported
         let kx = kx::KeyExchange::choose(share.group, &config.kx_groups)
-            .and_then(kx::KeyExchange::start)
```
ctz, line 491: I think this means the `start()` function can be removed from the trait? Since there are now no longer any calls to it.

ctz: (nb. this is mostly a note to myself, as I work through ensuring these APIs can be used outside the crate.)

rustls/src/crypto.rs
ctz: Have added a commit that does this.

rustls/src/server/tls13.rs
ctz: Actually I was wrong about this -- there is one call, but I think one that can be refactored away.

ctz: This is done; that commit could be squashed into earlier ones perhaps.

cpu: Nice!

cpu (approved):

rustls/src/server/tls13.rs
```diff
@@ -1157,13 +1157,15 @@ impl<C: CryptoProvider> State<ServerConnectionData> for ExpectFinished<C> {
             .key_schedule
             .sign_client_finish(&handshake_hash, cx.common);
 
-        let fin = constant_time::verify_slices_are_equal(expect_verify_data.as_ref(), &finished.0)
-            .map_err(|_| {
+        let fin = match ConstantTimeEq::ct_eq(expect_verify_data.as_ref(), &finished.0[..]).into() {
+            true => verify::FinishedMessageVerified::assertion(),
+            false => {
                 warn!("Finished wrong");
```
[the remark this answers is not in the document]
ctz: I think I'm minded to remove this logging. Like other uses of `DecryptError` it points to data corruption (malicious or otherwise) between us and the peer. That's not actionable for us or the peer, so I'm not sure the logging is useful?

rustls/src/client/tls13.rs
```diff
@@ -210,15 +210,21 @@ pub(super) fn initial_key_share(
...
+                .any(|supported_group| supported_group.name == *hint_group)
+        })
+        .unwrap_or(
             config
                 .kx_groups
                 .first()
                 .expect("No kx groups configured")
-        });
+                .name,
+        );
```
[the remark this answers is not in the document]
ctz: Yeah that is a bit clearer to me. Adding as `tls13: refactor choosing of first kx group`

rustls/src/client/hs.rs
```diff
@@ -791,13 +791,16 @@ impl<C: CryptoProvider> ExpectServerHelloOrHelloRetryRequest<C> {
...
         let key_share = match req_group {
             Some(group) if group != offered_key_share.group() => {
-                let group = kx::KeyExchange::choose(group, &config.kx_groups).ok_or_else(|| {
-                    cx.common.send_fatal_alert(
-                        AlertDescription::IllegalParameter,
-                        PeerMisbehaved::IllegalHelloRetryRequestWithUnofferedNamedGroup,
-                    )
-                })?;
-                kx::KeyExchange::start(group).ok_or(Error::FailedToGetRandomBytes)?
+                match kx::KeyExchange::choose(group, &self.next.input.config.kx_groups) {
```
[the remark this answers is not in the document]
ctz: Fixed and squashed into this commit.

rustls/src/server/tls13.rs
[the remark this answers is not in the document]
cpu: That makes sense to me :+1:

PR #1351, opened 2023-07-10
docs: update RELEASING w/ maintenance release steps.

cpu:
This branch adds some short guidance on performing maintenance point releases when `main` has breaking changes, preventing using the normal release process.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -40,3 +40,16 @@
 1. Create a new GitHub release for that tag.  Use "Generate release notes" (against the tag for the previous release)
    as a starting point for the release description.  Then add the "headlines" produced earlier at the top.
 2. Update dependent crates (eg, hyper-rustls, rustls-native-certs, etc.) if this was a semver-incompatible release.
+
+## Maintenance point releases
+
+When point releases for bug fixes and small backwards compatible changes, but `main` contains unreleased breaking 
+changes we follow a modified release process using a longer-lived maintenance branch.
+
+1. Check if there is an existing release branch, e.g. `rel-0.21` for point releases in the `0.21.x` series.
+   - If there is, use that branch.
+   - If there is not, create a new branch from the tag for the previous release, e.g. `git checkout -b rel-0.21 v/0.21.0`.
+     Remember to also create a branch protection rule for the release branch, matching the settings from `main`.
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -40,3 +40,17 @@
...
+
+When point releases for bug fixes and small backwards compatible changes, but `main` contains unreleased breaking 
+changes we follow a modified release process using a longer-lived maintenance branch.
+
+1. Check if there is an existing release branch, e.g. `rel-0.21` for point releases in the `0.21.x` series.
+   - If there is, use that branch.
+   - If there is not, create a new branch from the tag for the previous release, e.g. `git checkout -b rel-0.21 v/0.21.0`.
+2. Make pull-requests for any changes you want to include in the point release, targeted against the release branch.
+   - When possible, we prefer to land compatible changes on the release branch first, and then do a merge into `main` 
+     after-wards.
```
ctz, line 54: Note I think we want to avoid things like bug fixes ending up on release branches _first_, because:

- generally the set of CI gates is best and most maintained on main, so a change that lands there first will have passed all these.
- there's the real risk of post-completion errors -- ie. we make the panic fix-the-bug release, but then forget to get the fix and its regression test onto main. keeping always the latest branch avoids that.

cpu: Those are convincing reasons to me. I can back out this part of the recommendations.

> generally the set of CI gates is best and most maintained on main

As a concrete example it looks like `rel-0.21` had been missing the cross compile check :sweat:

cpu: Dropped the note about preferring merge to feature branch first. Also added a note when creating a feature branch to create a branch protection rule for it matching the settings from `main`.

cpu:
@ctz Is there anything further you think we should hash out here? Maybe wait for Djc's input?

ctz (approved):
i think we can land this, then see if @djc has any thoughts when he's back.

djc:
This looks good to me.

On balance I probably still prefer merging fixes forwards rather than cherry-picking backwards since it helps that you can actually rely on Git to help you track changes that still need to be merged. The fix for the issue flagged by @ctz could just be to proactively merge forward and/or to include merging open release branches in the release process checklist.

PR #1358, opened 2023-07-13
Remove MSRV variant of connect-tests

ctz:
MSRV is important (an tested separately) for the core crate (and its dependencies) but doesn't apply to test code.

Run these daily to notice any breakage earlier.

(Currently this is broken due to env_logger -> is-terminal -> rustix -> linux-raw-sys requiring 1.63 -- see https://github.com/rustls/rustls/actions/runs/5531529915/jobs/10092344273 )

### DIFF
--- a/.github/workflows/connect-tests.yml
+++ b/.github/workflows/connect-tests.yml
@@ -5,9 +5,9 @@ permissions:
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


cpu (approved):
Thanks!

cpu (approved):

PR #1359, opened 2023-07-13
client::builder: fix PhantomData clippy lint

ctz:
New lint in rust 1.71 released today. I think we've addressed some others of these when it landed in nightly, but this lint is only a warning there - https://github.com/rustls/rustls/actions/runs/5534907687/jobs/10100494233


cpu (approved):
Thanks for the quick fix :rocket:

PR #1363, opened 2023-07-18
Correct unnecessarily &mut function args

ctz:
As of just now, nightly clippy is pointing out these arguments don't need to be `&mut`:

```
warning: this argument is a mutable reference, but not used mutably
   --> rustls/src/client/tls12.rs:955:36
    |
955 |     fn save_session(&mut self, cx: &mut ClientContext<'_>) {
    |                                    ^^^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&ClientContext<'_>`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
    = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default

warning: this argument is a mutable reference, but not used mutably
   --> rustls/src/server/tls13.rs:602:13
    |
602 |         cx: &mut ServerContext<'_>,
    |             ^^^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&ServerContext<'_>`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut

warning: `rustls` (lib) generated 2 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s

```

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -598,6 +598,7 @@ mod client_hello {
         common.send_msg(m, false);
     }
 
+    #[allow(clippy::needless_pass_by_ref_mut)] // cx only mutated if cfg(feature = "quic")
     fn decide_if_early_data_allowed(
         cx: &mut ServerContext<'_>,
         client_hello: &ClientHelloPayload,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -277,10 +277,10 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      - run: cargo clippy --package rustls --all-features -- --deny warnings
-      - run: cargo clippy --package rustls --no-default-features -- --deny warnings
-      - run: cargo clippy --manifest-path=connect-tests/Cargo.toml --all-features -- --deny warnings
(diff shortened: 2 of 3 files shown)


ctz:
Oh! Well, that explains that.

rustls/src/server/tls13.rs
```diff
@@ -598,6 +598,7 @@ mod client_hello {
         common.send_msg(m, false);
     }
 
+    #[allow(clippy::needless_pass_by_ref_mut)] // cx only mutated if cfg(feature = "quic")
```
cpu, line 601: The non-nightly clippy bails on this `allow` because the lint only exists in the nightly clippy right now. Probably need to deploy the same strategy as we used in rustls-ffi (https://github.com/rustls/rustls-ffi/commit/4f977fb49c5ca1cda6fa63dfafb68b6b245d3eae) and update the clippy invocation w/ `-A unknown-lints` at least temporarily :thinking:

cpu (approved):
Thanks!

rustls/src/server/tls13.rs
ctz: Thanks! Doing that now.

Issue #1360, opened 2023-07-13
client certificate verification: builder style API


jsha:
Over in https://github.com/rustls/rustls-ffi/pull/324#pullrequestreview-1525777028 we talked a bit about the current API for the built-in implementations of ClientCertVerifier:

[AllowAnyAnonymousOrAuthenticatedClient](https://docs.rs/rustls/0.21.5/rustls/server/struct.AllowAnyAnonymousOrAuthenticatedClient.html)

[AllowAnyAuthenticatedClient](https://docs.rs/rustls/0.21.5/rustls/server/struct.AllowAnyAuthenticatedClient.html)

[NoClientAuth](https://docs.rs/rustls/0.21.5/rustls/server/struct.NoClientAuth.html)

The naming of these is a little verbose and confusing, and now that they take a CRLs argument, they are a bit redundant. Instead, we could adopt a builder-style API that could construct these, like ServerConfig. E.g.:

```
ClientVerifier::builder()
  .with_roots(root_store)
  .with_crls(crls)
  .allow_unauthenticated()
  .build()

ClientVerifier::builder()
  .with_roots(root_store)
  .without_crls()
  .build()

// equivalent to NoClientAuth; or we could keep NoClientAuth as a shortcut for this
ClientVerifier::builder()
  .allow_unauthenticated()
  .build()
```

cpu:
Happy to take a crack at this one.

jsha:
By the way, as mentioned on #1361, using CRLs for client cert verification in practice requires (a) using the Acceptor API, and (b) constructing a new verifier object on each handshake.

The current impls take ownership of a RootCertStore, which is a `Vec<OwnedTrustAnchor>`. That might get a little expensive to clone on each handshake, particularly if it's large. As long as we are making changes in this code it's probably a good idea to take an `Arc<RootCertStore>`.

cpu:
The webpki client cert verifiers are created w/ a builder-style API and take `Arc<RootCertStore>` as of https://github.com/rustls/rustls/pull/1368

PR #1365, opened 2023-07-19
verifier: use builder API for webpki client verifiers.

cpu:
## Description

Previously users configuring a `ServerConfig` that wanted to use a webpki backed client certificate verifier had to make a choice of which concrete implementation to construct, and how to configure it (e.g. with trust anchors and CRLs). This made for a somewhat cumbersome experience.

In its place, this branch:

* Adds a `WebpkiClientVerifier` type that replace both the `AllowAnyAuthenticatedClient` and `AllowAnyAnonymousOrAuthenticatedClient` verifiers. The name emphasizes that the implementation is backed by   `rustls/webpki` to help distinguish it from platform verifiers.

  The new type can only be constructed external to the crate using a `ClientCertVerifierBuilder` builder that walks the user through specifying roots, CRLs, and policy for anonymous clients.

* Turns the `NoClientAuth` verifier into a crate internal type that also only be constructed via the `ClientCertVerifierBuilder`.

* Removes the `boxed()` fn's of the above, since they won't be needed anymore - consumers will construct a `Arc<dyn ClientCertVerifier>` through the builder and don't need to have `ClientCertVerifier` in-scope via the dangerous config feature.

* Updates all existing usages in tests and examples to use the new builder API.

Resolves https://github.com/rustls/rustls/issues/1360

### DIFF
--- a/rustls/src/server/verifier_builder.rs
+++ b/rustls/src/server/verifier_builder.rs
@@ -1,0 +1,305 @@
+use std::{fmt, sync::Arc};
+
+use crate::server::UnparsedCertRevocationList;
+use crate::verify::{
+    AnonymousClientsPolicy, ClientCertVerifier, NoClientAuth, WebpkiClientVerifier,
+};
+use crate::{CertRevocationListError, RootCertStore};
+
+/// A builder for configuring a `webpki` client certificate verifier.
+///
+/// For more information, see the [`WebpkiClientVerifier`] documentation.
+#[derive(Clone)]
+pub struct ClientCertVerifierBuilder<State> {
+    pub(crate) state: State,
+}
+
+impl<State: fmt::Debug> fmt::Debug for ClientCertVerifierBuilder<State> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let state_name = std::any::type_name::<State>();
+        let (_, param) = state_name
+            .rsplit_once("::")
+            .unwrap_or(("", "Unknown"));
+        f.debug_struct(&format!("ClientCertVerifierBuilder<{}>", param))
+            .field("state", &self.state)
+            .finish()
+    }
+}
+
+/// A client certificate verifier builder state where the caller must decide whether to perform
+/// client certificate authentication using a [RootCertStore], or not.
+///
+/// For more information, see the [`WebpkiClientVerifier`] documentation.
+#[derive(Clone, Debug)]
+pub struct WantsClientAuthentication(pub(crate) ());
+
+impl ClientCertVerifierBuilder<WantsClientAuthentication> {
+    /// Disable client certificate authentication. No client certificate authentication will be
+    /// done, allowing unauthenticated client access to the server.
+    pub fn allow_unauthenticated(&self) -> Arc<dyn ClientCertVerifier> {
+        Arc::new(NoClientAuth {})
+    }
+
+    /// Verify that if a client offers a client certificate, that it is issued by a trust anchor
+    /// present in the provided [RootCertStore].
+    ///
+    /// All clients must provide a client certificate unless you further allow unauthenticated
+    /// clients by calling [allow_unauthenticated()][ClientCertVerifierBuilder<WantsCrls>::allow_unauthenticated]
+    /// on the returned builder.
+    pub fn with_roots(&self, roots: RootCertStore) -> ClientCertVerifierBuilder<WantsCrls> {
+        ClientCertVerifierBuilder {
+            state: WantsCrls { roots },
+        }
+    }
+}
+
+/// A client certificate verifier builder state where clients are verified using a [RootCertStore]
+/// and the caller must now decide whether to perform certificate revocation list (CRL) checking,
+/// and whether to allow unauthenticated clients in addition to those that supply a valid client
+/// certificate.
+///
+/// For more information, see the [`WebpkiClientVerifier`] documentation.
+#[derive(Clone, Debug)]
+pub struct WantsCrls {
+    roots: RootCertStore,
+}
+
+impl ClientCertVerifierBuilder<WantsCrls> {
+    /// Build a client certificate that will allow clients that present a client certificate
+    /// that is issued by a trusted CA, or anonymous clients that do not offer a client certificate
+    /// at all. No revocation checking with CRLS will be done. This is useful for servers that
+    /// expect both authenticated an unauthenticated clients and impose their own authentication
+    /// and authorization.
+    pub fn allow_unauthenticated(self) -> ClientCertVerifierBuilder<ClientAuthenticationOptional> {
+        ClientCertVerifierBuilder {
+            state: ClientAuthenticationOptional {
+                roots: self.state.roots,
+                crls: Vec::default(),
+            },
+        }
+    }
+
+    /// Build a client certificate verifier that will require all clients to present a client
+    /// certificate that is issued by a trusted CA. No revocation checking with CRLs will be done.
+    /// No anonymous clients will be allowed to access the server.
+    ///
+    /// Note that not providing CRLs means that clients accessing the server with revoked
+    /// credentials will not be denied access.
+    pub fn without_crls(self) -> ClientCertVerifierBuilder<WantsUnauthenticatedPolicy> {
+        ClientCertVerifierBuilder {
+            state: WantsUnauthenticatedPolicy {
+                roots: self.state.roots,
+                crls: Vec::default(),
+            },
+        }
+    }
+
+    /// Build a client certificate verifier that will require all clients to present a client
+    /// certificate that is issued by a trusted CA. The revocation state of the client
+    /// certificate will be checked against the provided certificate revocation lists (CRLs).
+    ///
+    /// All clients must provide a client certificate unless you further allow unauthenticated
+    /// clients by calling
+    /// [allow_unauthenticated()][ClientCertVerifierBuilder<WantsUnauthenticatedPolicy>::allow_unauthenticated]
+    /// on the returned builder.
+    pub fn with_crls(
+        self,
+        crls: impl IntoIterator<Item = UnparsedCertRevocationList>,
+    ) -> Result<ClientCertVerifierBuilder<WantsUnauthenticatedPolicy>, CertRevocationListError>
+    {
+        let crls = crls
+            .into_iter()
+            .map(|der_crl| der_crl.parse())
+            .collect::<Result<Vec<_>, CertRevocationListError>>()?;
+
+        Ok(ClientCertVerifierBuilder {
+            state: WantsUnauthenticatedPolicy {
+                roots: self.state.roots,
+                crls,
+            },
+        })
+    }
+}
+
+/// A client certificate verifier builder state where clients are verified using a [RootCertStore]
+/// and optionally may have their revocation status checked via CRLs. The caller must now decide
+/// whether or not to allow unauthenticated clients in addition to those that supply a valid client
+/// certificate.
+///
+/// For more information, see the [`WebpkiClientVerifier`] documentation.
+#[derive(Debug, Clone)]
+pub struct WantsUnauthenticatedPolicy {
+    roots: RootCertStore,
+    crls: Vec<webpki::OwnedCertRevocationList>,
+}
+
+impl ClientCertVerifierBuilder<WantsUnauthenticatedPolicy> {
+    /// Build a client certificate verifier that will require all clients to present a client
+    /// certificate that is issued by a trusted CA. If configured, revocation checking with CRLs
+    /// will be done. No anonymous clients will be allowed to access the server.
+    pub fn require_authentication(self) -> Arc<dyn ClientCertVerifier> {
+        Arc::new(WebpkiClientVerifier::new(
+            self.state.roots,
+            self.state.crls,
+            AnonymousClientsPolicy::Forbid,
+        ))
+    }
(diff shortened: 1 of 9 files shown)


rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,427 @@
+use std::{fmt, sync::Arc};
```
cpu, line 1: I made a new module (`src/server/verifier_builder.rs`) for this because `src/verify.rs` is looking a bit busy to me. Arguably this new code could also live there :shrug:

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+}
+
+/// A client certificate verifier builder state where clients are verified using a [RootCertStore]
+/// and optionally may have their revocation status checked via CRLs. The caller must now decide
+/// or not to allow unauthenticated clients in addition to those that supply a valid client
+/// certificate.
+///
+/// For more information, see the [`WebpkiClientVerifier`] documentation.
+// TODO(XXX): derive Clone once OwnedCertRevocationList does.
```
cpu, line 209: I will follow up on this in `webpki` but didn't want to block the builder API on it.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+            },
+        }
+    }
+}
+
+impl fmt::Debug for WantsUnauthenticatedPolicy {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("WantsUnauthenticatedPolicy")
+            .field("roots", &self.roots)
+            // TODO(XXX): replace when OwnedCertRevocationList implements Debug.
```
cpu, line 244: Ditto here: I will follow up on this in `webpki` but didn't want to block the builder API on it.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,427 @@
...
+        });
+        roots
+    }
+
+    #[test]
+    fn test_noauth() {
+        // We should be able to build a verifier that does no client authentication.
+        let builder = WebpkiClientVerifier::builder().allow_unauthenticated();
+        // The builder should be Debug.
+        println!("{:?}", builder);
```
cpu, line 340: Originally I was asserting that the debug string exactly matched expected. I went back to this style because:

a) it felt super fragile to match the debug string output exactly in the tests.
b) when providing real trust anchors (making for better unit tests), the debug output can be quite large. This will also be worse once `webpki::OwnedCertRevocationList` is `Debug`.

rustls/src/verify.rs
```diff
@@ -468,40 +468,22 @@ impl AllowAnyAuthenticatedClient {
     /// Construct a new `AllowAnyAuthenticatedClient`.
     ///
     /// `roots` is the list of trust anchors to use for certificate validation.
-    pub fn new(roots: RootCertStore) -> Self {
+    /// `crls` are an iterable of owned certificate revocation lists (CRLs) to use for
+    /// client certificate validation.
+    pub(crate) fn new(
+        roots: RootCertStore,
+        crls: impl IntoIterator<Item = webpki::OwnedCertRevocationList>,
```
cpu, line 475: I changed things so that the builder does the work of converting from `UnparsedCertRevocationList` to `webpki::OwnedCertRevocationList` so that errors can be surfaced at the time of calling `with_crls(...)`. As a result the verifier side is changed to work with the owned representation and is infallible to construct.

rustls/src/server/verifier_builder.rs
cpu: https://github.com/rustls/webpki/pull/129

rustls/src/server/verifier_builder.rs
cpu: https://github.com/rustls/webpki/pull/129

rustls/src/verify.rs
```diff
@@ -458,7 +458,7 @@ impl UnparsedCertRevocationList {
 /// A `ClientCertVerifier` that will ensure that every client provides a trusted
 /// certificate, without any name checking. Optionally, client certificates will
 /// have their revocation status checked using the DER encoded CRLs provided.
-pub struct AllowAnyAuthenticatedClient {
+pub(crate) struct AllowAnyAuthenticatedClient {
```
ctz, line 461: Now this is non-public, maybe we could rename it, get rid of `AllowAnyAnonymousOrAuthenticatedClient`, and just allow configuration of `client_auth_mandatory`'s return?

cpu: > Now this is non-public, maybe we could rename it

Maybe I should smoosh the `WebpkiClientVerifier` that I added in `verifier_builder.rs` as an empty type that's only good for producing a `ClientCertVerifierBuilder` w/ `builder()` into this type, keeping the new name and having it be the impl of the `ClientCertVerifier` trait. It felt a bit awkard having that `WebpkiClientVerifier` type that wasn't actually a verifier, just a gateway to a builder.

>  get rid of AllowAnyAnonymousOrAuthenticatedClient, and just allow configuration of client_auth_mandatory's return?

Makes sense to me :+1:

cpu: Done :stars:

jsha:
FWIW, I find the state-based builder used in [ConfigBuilder](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html) to be really hard to use and understand as a user.

There's an alternative design that's more common, where there are no [typestates](http://cliffle.com/blog/rust-typestate/) and each accessor / mutator can be called at any time. So for instance you could call setters in any order:

```
            WebpkiClientVerifier::builder()
                .allow_unauthenticated()
                .with_roots(client_auth_roots)
                .with_crls(crls)
                .build()
```

OR

```
            WebpkiClientVerifier::builder()
                .with_roots(client_auth_roots)
                .with_crls(crls)
                .allow_unauthenticated()
                .build()

```

The downside of this style is that the type system can't rule out invalid configs; for instance calling `with_crls()` but not `with_roots()`. But I think there are not so many invalid configs here that we need to lean on the type system. For instance, calling `with_crls()` but not `with_roots()` would simply result in a config that considers nothing authenticated, failing closed.

ctz:
> The downside of this style is that the type system can't rule out invalid configs

I think we'd also have to define the semantics of calling things several times, eg:

```
            WebpkiClientVerifier::builder()
                .with_roots(client_auth_roots1)
                .with_roots(client_auth_roots2)
                .with_crls(crls1)
                .with_crls(crls2)
                .build()
```

But merging them seems plausible?

cpu:
> FWIW, I find the state-based builder used in [ConfigBuilder](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html) to be really hard to use and understand as a user.

@jsha Can you expand on what you find hard to use/understand? I think it will help evaluate which approach makes more sense to know what specific challenges we're trying to address with a different builder style.

Separately, if there are usability issues with the state based approach I think we should put some effort into addressing them for the `ConfigBuilder`. Fixing the rough edges for the client verifier builder won't have as much impact as doing the same for the `ConfigBuilder` - that API surface is something I think almost every user of the library touches.

> There's an alternative design that's more common, where there are no [typestates](http://cliffle.com/blog/rust-typestate/) and each accessor / mutator can be called at any time.

:+1: I'll put something like this together in a separate branch to compare/contrast with this branch.

> But merging them seems plausible?

Agreed.

cpu:
> I'll put something like this together in a separate branch to compare/contrast with this branch.

Here you go: https://github.com/rustls/rustls/pull/1368

I'm still in favour of the implementation in this branch but it's a loosely held opinion.

djc:
> @jsha Can you expand on what you find hard to use/understand? I think it will help evaluate which approach makes more sense to know what specific challenges we're trying to address with a different builder style.

The main issue I've ran into is that the reference docs based on rustdoc get pretty hard to read. I wonder if this would be better if we used real separate types instead of a single type parametrized with a state type.

djc (commented):
Nice!

jsha:
> @jsha Can you expand on what you find hard to use/understand? I think it will help evaluate which approach makes more sense to know what specific challenges we're trying to address with a different builder style.

Good point, thanks. Done: https://github.com/rustls/rustls/issues/1372

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,362 @@
+use std::{fmt, sync::Arc};
+
+use crate::verify::AnonymousClientsPolicy;
```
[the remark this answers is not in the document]
cpu: Fixed.

rustls/src/verify.rs
```diff
@@ -544,68 +600,12 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
...
-        intermediates: &[Certificate],
-        now: SystemTime,
-    ) -> Result<ClientCertVerified, Error> {
-        self.inner
-            .verify_client_cert(end_entity, intermediates, now)
-    }
+/// Controls how the [WebpkiClientVerifier] handles anonymous clients.
+pub(crate) enum AnonymousClientsPolicy {
+    /// Clients that do not present a client certificate are allowed.
+    Allowed,
```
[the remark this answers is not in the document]
cpu: Sounds good. Renamed.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,362 @@
...
+
+/// A client certificate verifier builder state where the server will perform no client
+/// authentication.
+#[derive(Clone, Debug)]
+pub struct NoClientAuthentication(pub(crate) ());
+
+impl ClientCertVerifierBuilder<NoClientAuthentication> {
+    /// Build a "no operation" client certificate verifier that will allow all clients to connect,
+    /// performing no client certificate authentication or revocation checking.
+    pub fn build(self) -> Arc<dyn ClientCertVerifier> {
```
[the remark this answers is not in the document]
cpu: I think you're right, it wasn't necessary. Removed!

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,362 @@
...
+/// certificate.
+///
+/// For more information, see the [`WebpkiClientVerifier`] documentation.
+#[derive(Clone, Debug)]
+pub struct WantsCrls {
+    roots: RootCertStore,
+}
+
+impl ClientCertVerifierBuilder<WantsCrls> {
+    /// Build a client certificate verifier that will require all clients present a client
```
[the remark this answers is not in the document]
cpu: Reworded throughout.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// No anonymous clients will be allowed to access the server.
+    pub fn build(self) -> Arc<dyn ClientCertVerifier> {
+        Arc::new(WebpkiClientVerifier::new(
+            self.state.roots,
+            Vec::default(),
+            AnonymousClientsPolicy::Forbidden,
+        ))
+    }
+
+    /// Build a client certificate that will allow clients that present a client certificate
```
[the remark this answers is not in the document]
cpu: Added a note.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,362 @@
...
+                crls: Vec::default(),
+            },
+        }
+    }
+
+    /// Build a client certificate verifier that will require all clients present a client
+    /// certificate that is issued by a trusted CA. No revocation checking with CRLs will be done.
+    /// No anonymous clients will be allowed to access the server.
+    ///
+    /// Equivalent to calling [build()][ClientCertVerifierBuilder<WantsCrls>::build].
```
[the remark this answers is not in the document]
cpu: Dropped the duplicative `build` step and added a note about the risk of not using CRLs.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,362 @@
...
+pub struct WantsUnauthenticatedPolicy {
+    roots: RootCertStore,
+    crls: Vec<webpki::OwnedCertRevocationList>,
+}
+
+impl ClientCertVerifierBuilder<WantsUnauthenticatedPolicy> {
+    /// Build a client certificate verifier that will require all clients present a client
+    /// certificate that is issued by a trusted CA. If configured, revocation checking with CRLs
+    /// will be done. No anonymous clients will be allowed to access the server.
+    pub fn build(self) -> Arc<dyn ClientCertVerifier> {
```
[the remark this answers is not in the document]
cpu: Done.

rustls/src/verify.rs
cpu: Good point, fixed.

rustls/src/server/verifier_builder.rs
cpu: We're on the updated webpki in `main` so I was able to resolve this.

rustls/src/server/verifier_builder.rs
cpu: We're on the updated webpki in `main` so I was able to resolve this.

cpu:
@ctz @djc What do you think given the feedback in https://github.com/rustls/rustls/issues/1372? Should I close this branch and focus on https://github.com/rustls/rustls/pull/1368 ? I've rebased + addressed any outstanding feedback on both.

cpu:
Let's roll with #1368

PR #1368, opened 2023-07-21
verifier: use (stateless) builder API for client verifier.

cpu:
**Note: this is an alternative implementation to https://github.com/rustls/rustls/pull/1365, that avoids using the [Typestate pattern](http://cliffle.com/blog/rust-typestate/).**

### verifier: use builder API for client verifier.

Previously users configuring a `ServerConfig` that wanted to use a webpki backed client certificate verifier had to make a choice of which concrete implementation to construct, and how to configure it (e.g. with trust anchors and CRLs). This made for a somewhat cumbersome experience.

In its place, this commit:

* Adds a `WebPkiClientVerifier` type that replace both the `AllowAnyAuthenticatedClient` and `AllowAnyAnonymousOrAuthenticatedClient` verifiers. The name emphasizes that the implementation is backed by `rustls/webpki` to help distinguish it from platform verifiers.

  The new type can only be constructed external to the crate using a `ClientCertVerifierBuilder` builder that walks the user through specifying roots, CRLs, and policy for anonymous clients.

* Turns the `NoClientAuth` verifier into a crate internal type that also only be constructed via the `ClientCertVerifierBuilder`.

* Removes the `boxed()` fn's of the above, since they won't be needed anymore - consumers will construct a `Arc<dyn ClientCertVerifier>` through the builder and don't need to have `ClientCertVerifier` in-scope via the dangerous config feature.

* Updates all existing usages in tests and examples to use the new builder API.

### anchors: add_server_trust_anchors -> add_trust_anchors

The `RootCertStore` type is used for both client and server trust anchors. This commit renames the `add_server_trust_anchors` method to be `add_trust_anchors` to reflect its general purpose.

### verify: WebPkiVerifier -> WebPkiServerVerifier.
In previous commits we reworked the primary implementation of the `ClientCertVerifier` trait to be named `WebPkiClientVerifier`.

This commit updates the corresponding `ServerCertVerifier` implementation in `WebPkiVerifier` to be named `WebPkiServerVerifier` to match the client naming scheme and to better emphasize its role.

### verify: take Arc<RootCertStore>.

As [pointed out by Jsha](https://github.com/rustls/rustls/issues/1360#issuecomment-1634705941), an implementation of the `Acceptor` API may want to create a verifier on something approaching a per-handshake basis in order to provide up-to-date CRLs and client trust anchors.

We can improve on the cost of this operation by allowing shared use of a `RootCertStore` across verifiers by wrapping it in an `Arc`.

This commit updates the `WebPkiClientVerifier`, and `ClientCertVerifierBuilder` to take a `Arc<RootCertStore>` instead of
`RootCertStore`. 

One side-effect of this change is the removal of the `add_roots` fn of the `ClientCertVerifierBuilder` - once we take an `Arc` we can't modify the backing `RootCertStore` without introducing some form of locking. I think the use-case for adding additional `RootCertStore`'s after constructing the builder is weak enough that we should drop that feature rather than introduce locking.

### DIFF
--- a/examples/src/bin/limitedclient.rs
+++ b/examples/src/bin/limitedclient.rs
@@ -11,7 +11,7 @@ use rustls::OwnedTrustAnchor;
 
 fn main() {
     let mut root_store = rustls::RootCertStore::empty();
-    root_store.add_server_trust_anchors(
+    root_store.add_trust_anchors(
         webpki_roots::TLS_SERVER_ROOTS
             .iter()
             .map(|ta| {
--- a/rustls/src/server/verifier_builder.rs
+++ b/rustls/src/server/verifier_builder.rs
@@ -9,29 +9,18 @@ use crate::{CertRevocationListError, RootCertStore};
 /// For more information, see the [`WebPkiClientVerifier`] documentation.
 #[derive(Debug, Clone)]
 pub struct ClientCertVerifierBuilder {
-    roots: RootCertStore,
+    roots: Arc<RootCertStore>,
     crls: Vec<UnparsedCertRevocationList>,
     anon_policy: AnonymousClientPolicy,
 }
 
 impl ClientCertVerifierBuilder {
-    pub(crate) fn new(roots: RootCertStore) -> Self {
+    pub(crate) fn new(roots: Arc<RootCertStore>) -> Self {
         Self {
             roots,
             crls: Vec::new(),
             anon_policy: AnonymousClientPolicy::Deny,
         }
-    }
-
-    /// Add additional `roots` to use to verify client certificates.
-    ///
-    /// All clients must provide a client certificate unless you have allowed unauthenticated
-    /// clients by calling [allow_unauthenticated()][ClientCertVerifierBuilder::allow_unauthenticated]
-    /// on the builder.
-    pub fn with_roots(mut self, roots: RootCertStore) -> Self {
-        self.roots
-            .add_trust_anchors(roots.roots.into_iter());
-        self
     }
 
     /// Verify the revocation state of presented client certificates against the provided
@@ -103,6 +92,7 @@ mod tests {
     use crate::server::{ClientCertVerifierBuilderError, UnparsedCertRevocationList};
     use crate::verify::WebPkiClientVerifier;
     use crate::{Certificate, RootCertStore};
+    use std::sync::Arc;
 
     fn load_crls(crls_der: &[&[u8]]) -> Vec<UnparsedCertRevocationList> {
         crls_der
@@ -126,17 +116,17 @@ mod tests {
         ])
     }
 
-    fn load_roots(roots_der: &[&[u8]]) -> RootCertStore {
+    fn load_roots(roots_der: &[&[u8]]) -> Arc<RootCertStore> {
         let mut roots = RootCertStore::empty();
         roots_der.iter().for_each(|der| {
             roots
                 .add(&Certificate(der.to_vec()))
                 .unwrap()
         });
-        roots
+        roots.into()
     }
 
-    fn test_roots() -> RootCertStore {
+    fn test_roots() -> Arc<RootCertStore> {
         load_roots(&[
             include_bytes!("../../../test-ca/ecdsa/ca.der").as_slice(),
             include_bytes!("../../../test-ca/rsa/ca.der").as_slice(),
@@ -156,20 +146,6 @@ mod tests {
         let builder = WebPkiClientVerifier::builder(test_roots());
         // The builder should be Debug.
         println!("{:?}", builder);
-        builder.build().unwrap();
-    }
-
-    #[test]
-    fn test_required_auth_add_roots() {
-        // We should be able to call `with_roots` to add more roots.
-        let initial_roots = test_roots();
-        let extra_roots = load_roots(&[include_bytes!("../../../test-ca/eddsa/ca.der").as_slice()]);
-        let builder =
-            WebPkiClientVerifier::builder(initial_roots.clone()).with_roots(extra_roots.clone());
-        // The builder should be Debug.
-        println!("{:?}", builder);
-        // There should be the expected number of roots.
-        assert_eq!(builder.roots.len(), initial_roots.len() + extra_roots.len());
         builder.build().unwrap();
     }
 
@@ -260,7 +236,7 @@ mod tests {
     #[test]
     fn test_builder_no_roots() {
         // Trying to create a builder with no trust anchors should fail at build time
-        let result = WebPkiClientVerifier::builder(RootCertStore::empty()).build();
+        let result = WebPkiClientVerifier::builder(RootCertStore::empty().into()).build();
         assert!(matches!(
             result,
             Err(ClientCertVerifierBuilderError::NoRootAnchors)
--- a/rustls/src/server/verifier_builder.rs
+++ b/rustls/src/server/verifier_builder.rs
@@ -30,7 +30,7 @@ impl ClientCertVerifierBuilder {
     /// on the builder.
     pub fn with_roots(mut self, roots: RootCertStore) -> Self {
         self.roots
-            .add_server_trust_anchors(roots.roots.into_iter());
+            .add_trust_anchors(roots.roots.into_iter());
         self
     }
 
--- a/rustls/src/server/verifier_builder.rs
+++ b/rustls/src/server/verifier_builder.rs
@@ -1,0 +1,269 @@
+use std::sync::Arc;
+
+use crate::server::UnparsedCertRevocationList;
+use crate::verify::{AnonymousClientPolicy, ClientCertVerifier, WebPkiClientVerifier};
+use crate::{CertRevocationListError, RootCertStore};
+
+/// A builder for configuring a `webpki` client certificate verifier.
+///
+/// For more information, see the [`WebPkiClientVerifier`] documentation.
+#[derive(Debug, Clone)]
+pub struct ClientCertVerifierBuilder {
+    roots: RootCertStore,
+    crls: Vec<UnparsedCertRevocationList>,
+    anon_policy: AnonymousClientPolicy,
+}
+
+impl ClientCertVerifierBuilder {
+    pub(crate) fn new(roots: RootCertStore) -> Self {
+        Self {
+            roots,
+            crls: Vec::new(),
+            anon_policy: AnonymousClientPolicy::Deny,
+        }
+    }
+
+    /// Add additional `roots` to use to verify client certificates.
+    ///
+    /// All clients must provide a client certificate unless you have allowed unauthenticated
+    /// clients by calling [allow_unauthenticated()][ClientCertVerifierBuilder::allow_unauthenticated]
+    /// on the builder.
+    pub fn with_roots(mut self, roots: RootCertStore) -> Self {
+        self.roots
+            .add_server_trust_anchors(roots.roots.into_iter());
+        self
+    }
+
+    /// Verify the revocation state of presented client certificates against the provided
+    /// certificate revocation lists (CRLs). Calling `with_crls` multiple times appends the
+    /// given CRLs to the existing collection.
+    pub fn with_crls(mut self, crls: impl IntoIterator<Item = UnparsedCertRevocationList>) -> Self {
+        self.crls.extend(crls);
+        self
+    }
+
+    /// Allow unauthenticated clients to connect.
+    ///
+    /// Clients that offer a client certificate issued by a trusted root, and clients that offer no
+    /// client certificate will be allowed to connect.
+    pub fn allow_unauthenticated(mut self) -> Self {
+        self.anon_policy = AnonymousClientPolicy::Allow;
+        self
+    }
+
+    /// Build a client certificate verifier. The built verifier will be used for the server to offer
+    /// client certificate authentication, to control how offered client certificates are validated,
+    /// and to determine what to do with anonymous clients that do not respond to the client
+    /// certificate authentication offer with a client certificate.
+    ///
+    /// Once built, the provided `Arc<dyn ClientCertVerifier>` can be used with a Rustls
+    /// [crate::server::ServerConfig] to configure client certificate validation using
+    /// [`with_client_cert_verifier`][crate::ConfigBuilder<ClientConfig, WantsVerifier>::with_client_cert_verifier].
+    ///
+    /// # Errors
+    /// This function will return a `ClientCertVerifierBuilderError` if:
+    /// 1. No trust anchors have been provided.
+    /// 2. DER encoded CRLs have been provided that can not be parsed successfully.
+    pub fn build(self) -> Result<Arc<dyn ClientCertVerifier>, ClientCertVerifierBuilderError> {
+        if self.roots.is_empty() {
+            return Err(ClientCertVerifierBuilderError::NoRootAnchors);
+        }
+
+        Ok(Arc::new(WebPkiClientVerifier::new(
+            self.roots,
+            self.crls
+                .into_iter()
+                .map(|der_crl| der_crl.parse())
+                .collect::<Result<Vec<_>, CertRevocationListError>>()?,
+            self.anon_policy,
+        )))
+    }
+}
+
+/// One or more root trust anchors must be provided to create a [ClientCertVerifierBuilder].
+/// If you wish to disable client authentication, then use [WebPkiClientVerifier::no_client_auth]
+/// instead of constructing a builder.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum ClientCertVerifierBuilderError {
+    /// No root trust anchors were provided.
+    NoRootAnchors,
+    /// A provided CRL could not be parsed.
+    InvalidCrl(CertRevocationListError),
+}
+
+impl From<CertRevocationListError> for ClientCertVerifierBuilderError {
+    fn from(value: CertRevocationListError) -> Self {
+        Self::InvalidCrl(value)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::server::{ClientCertVerifierBuilderError, UnparsedCertRevocationList};
+    use crate::verify::WebPkiClientVerifier;
+    use crate::{Certificate, RootCertStore};
+
+    fn load_crls(crls_der: &[&[u8]]) -> Vec<UnparsedCertRevocationList> {
+        crls_der
+            .iter()
+            .map(|pem_bytes| {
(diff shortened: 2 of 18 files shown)


rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,284 @@
...
+            anon_policy: AnonymousClientsPolicy::Allowed,
+        }
+    }
+
+    /// Build a client certificate verifier.
+    ///
+    /// Once built, the provided `Arc<dyn ClientCertVerifier>` can be used with a Rustls
+    /// [crate::server::ServerConfig] to configure client certificate validation using
+    /// [`with_client_cert_verifier`][crate::ConfigBuilder<ClientConfig, WantsVerifier>::with_client_cert_verifier].
+    pub fn build(self) -> Arc<dyn ClientCertVerifier> {
```
cpu, line 90: We could make `build()` return `Result<Arc<dyn ClientCertVerifier>, Error>` to flag nonsense states like "crls provided, no roots", or "no roots, AnonymousClientsPolicy::Forbidden`.

That'll require adding some new `Error` variants. In general, I think this clumsyness is a point in favour of https://github.com/rustls/rustls/pull/1365

cpu: Added some error checking.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,318 @@
...
+    ///
+    /// All clients must provide a client certificate unless you have allowed unauthenticated
+    /// clients by calling [allow_unauthenticated()][ClientCertVerifierBuilder::allow_unauthenticated]
+    /// on the builder.
+    pub fn with_roots(self, roots: RootCertStore) -> Self {
+        Self {
+            roots: match self.roots {
+                None => Some(roots),
+                Some(mut existing) => {
+                    existing.add_server_trust_anchors(roots.roots.into_iter());
```
ctz, line 53: Given `RootCertStore` is actually agnostic of server/client, I think this function is now misnamed?

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    /// Build a client certificate verifier.
+    ///
+    /// Once built, the provided `Arc<dyn ClientCertVerifier>` can be used with a Rustls
+    /// [crate::server::ServerConfig] to configure client certificate validation using
+    /// [`with_client_cert_verifier`][crate::ConfigBuilder<ClientConfig, WantsVerifier>::with_client_cert_verifier].
+    pub fn build(self) -> Result<Arc<dyn ClientCertVerifier>, ClientCertVerifierBuilderError> {
+        if !self.crls.is_empty() && self.roots.is_none() {
```
ctz, line 103: `roots.is_none()` means `with_roots()` wasn't called, but I think we could usefully catch calling `with_roots` with an empty `RootCertStore`? maybe `self.roots` could stop being an `Option`, and this could become `self.roots.empty()`?

ctz (commented):
i think i weakly prefer this version, so this is the one i reviewed

rustls/src/server/verifier_builder.rs
ctz: (er, this isn't really a result of this PR. just a general observation)

rustls/src/server/verifier_builder.rs
cpu: Good idea, reworked.

rustls/src/server/verifier_builder.rs
cpu: That's a good point. I tacked on a second commit to rename this function: 41d25d4fb1e74104f6e0efa2cc4fd15ab499674e

rustls/src/verify.rs
```diff
@@ -637,18 +637,7 @@ pub(crate) fn pki_error(error: webpki::Error) -> Error {
...
-    /// Construct a [`NoClientAuth`], wrap it in an [`Arc`] and coerce it to
-    /// `dyn ClientCertVerifier`.
-    #[inline(always)]
-    pub fn boxed() -> Arc<dyn ClientCertVerifier> {
-        // This function is needed because `ClientCertVerifier` is only reachable if the
-        // `dangerous_configuration` feature is enabled, which makes coercing hard to outside users
-        Arc::new(Self)
-    }
-}
+pub(crate) struct NoClientAuth;
```
djc, line 640: So the reason we're keeping this around is a little subtle when reading the source, so maybe there should be some comments (here but maybe also around the builder and/or the `WebpkiClientVerifier` type?): with `WebPkiClientVerifier` in `anon_policy: Allow` mode the server still **offers** client authentication, which is not the case for `NoClientAuth`. I think we should probably also call out this difference in the docstring for the builder's `build()` method.

djc:
I think this approach is fine for this scenario.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,309 @@
...
+            anon_policy: AnonymousClientsPolicy::Allowed,
+        }
+    }
+
+    /// Build a client certificate verifier.
+    ///
+    /// Once built, the provided `Arc<dyn ClientCertVerifier>` can be used with a Rustls
+    /// [crate::server::ServerConfig] to configure client certificate validation using
+    /// [`with_client_cert_verifier`][crate::ConfigBuilder<ClientConfig, WantsVerifier>::with_client_cert_verifier].
+    pub fn build(self) -> Result<Arc<dyn ClientCertVerifier>, ClientCertVerifierBuilderError> {
```
djc, line 98: The docstring here should clearly enumerate the reasons it might fail. I think `# Errors` is a common way to introduce those?

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,309 @@
+use std::sync::Arc;
+
+use crate::server::UnparsedCertRevocationList;
+use crate::verify::{
+    AnonymousClientsPolicy, ClientCertVerifier, NoClientAuth, WebpkiClientVerifier,
+};
+use crate::{CertRevocationListError, RootCertStore};
+
+/// Possible error states produced by [`ClientCertVerifierBuilder::build`].
```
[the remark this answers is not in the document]
cpu: Moved below the builder.

rustls/src/verify.rs
```diff
@@ -457,60 +459,113 @@ impl UnparsedCertRevocationList {
...
+    anon_policy: AnonymousClientsPolicy,
 }
 
-impl AllowAnyAuthenticatedClient {
-    /// Construct a new `AllowAnyAuthenticatedClient`.
+impl WebpkiClientVerifier {
+    /// Create builder to build up the `webpki` client certificate verifier configuration.
+    ///
+    /// For more information, see the [`WebpkiClientVerifier`] documentation.
+    pub fn builder() -> ClientCertVerifierBuilder {
```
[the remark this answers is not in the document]
cpu: > I think we should also offer a no_client_auth() constructor that short-circuits the builder here, constructing the obvious WebPkiClientVerifier without the requirement to go through builder().build(), especially since it seems like this will be the most common scenario by some distance.

Good idea, I've added this method and updated some comments throughout to point to it.

> After considering the subtle differences resulting from empty vs non-empty roots, maybe we should just enforce non-empty roots for the builder().build() path (perhaps by having builder() take a roots argument)?

I was hesitant about this at first but I implemented it and agree it makes things clearer in conjunction with the `no_client_auth` method. It's nice that this moves the chance of error to the initial construction of the builder, and when providing CRLs, and not at the time of calling `build`. I think that will be a better experience.

rustls/src/verify.rs
```diff
@@ -547,68 +602,13 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
...
-        end_entity: &Certificate,
-        intermediates: &[Certificate],
-        now: SystemTime,
-    ) -> Result<ClientCertVerified, Error> {
-        self.inner
-            .verify_client_cert(end_entity, intermediates, now)
-    }
+/// Controls how the [WebpkiClientVerifier] handles anonymous clients.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AnonymousClientsPolicy {
```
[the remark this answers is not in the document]
cpu: > I think AnonymousClientPolicy (singular Client) would fit in more? Feels like the policy covers decisions on one client at a time

Makes sense. Dropped the plurality.

>  Also, should we make the variants active verbs as elsewhere (Allow, Forbid or maybe Deny)?

Sorry about that. I addressed this feedback in the other PR and think I forgot to carry it forward here. Fixed now.

rustls/src/verify.rs
```diff
@@ -457,60 +459,113 @@ impl UnparsedCertRevocationList {
...
+///   .expect("invalid crls")
+///   .build();
+/// ```
+///
+/// [^1]: <https://github.com/rustls/webpki>
+pub struct WebpkiClientVerifier {
     roots: RootCertStore,
     subjects: Vec<DistinguishedName>,
     crls: Vec<webpki::OwnedCertRevocationList>,
+    anon_policy: AnonymousClientsPolicy,
```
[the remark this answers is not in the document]
cpu: Done.

rustls/src/verify.rs
cpu: I've updated some comments to emphasize this aspect. Good call out :+1:

rustls/src/server/verifier_builder.rs
cpu: I've added `Errors` sections to the `builder` constructor and to `with_crls`. The `build` function no longer returns a `Result` and so doesn't need it.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+#[derive(Debug, Clone)]
+pub struct ClientCertVerifierBuilder {
+    roots: RootCertStore,
+    crls: Vec<webpki::OwnedCertRevocationList>,
+    anon_policy: AnonymousClientPolicy,
+}
+
+impl ClientCertVerifierBuilder {
+    pub(crate) fn new(roots: RootCertStore) -> Result<Self, RootsRequiredError> {
+        match roots.is_empty() {
```
djc, line 19: So I think we should actually defer this error to `build()` time, because one of the downsides of the current approach is that there are potentially two places to deal with errors, both `builder()` and `with_crls()` -- and ideally the caller would want to handle those at the same time.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    /// Add additional `roots` to use to verify client certificates.
+    ///
+    /// All clients must provide a client certificate unless you have allowed unauthenticated
+    /// clients by calling [allow_unauthenticated()][ClientCertVerifierBuilder::allow_unauthenticated]
+    /// on the builder.
+    pub fn with_roots(mut self, roots: RootCertStore) -> Self {
+        self.roots
+            .add_server_trust_anchors(roots.roots.into_iter());
+        Self {
```
djc, line 37: For these `with_()` methods, I think it'd be more obvious to modify `self` and then reutrn the existing instance `self` rather than moving all the fields.

In most cases this would be more concise, too.

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,274 @@
...
+            roots: self.roots,
+            crls: self.crls,
+            anon_policy: self.anon_policy,
+        }
+    }
+
+    /// Verify the revocation state of presented client certificates against the provided
+    /// certificate revocation lists (CRLs). Calling `with_crls` multiple time will merge the
```
djc, line 45: "merge the CRL[ist]s" is actually not what happens, right? I think a better way to describe this would be more like "appends the given further CRLs to the existing collection".

rustls/src/verify.rs
```diff
@@ -457,60 +459,126 @@ impl UnparsedCertRevocationList {
...
-    }
 }
 
-impl ClientCertVerifier for AllowAnyAuthenticatedClient {
+impl ClientCertVerifier for WebPkiClientVerifier {
     fn offer_client_auth(&self) -> bool {
         true
     }
 
+    fn client_auth_mandatory(&self) -> bool {
```
djc, line 575: Since the API is changing anyway, should we change the trait method signature to use the `enum` instead of a bool (and, if so, maybe change the name to match)?

rustls/src/verify.rs
```diff
@@ -636,19 +649,11 @@ pub(crate) fn pki_error(error: webpki::Error) -> Error {
...
-    /// `dyn ClientCertVerifier`.
-    #[inline(always)]
-    pub fn boxed() -> Arc<dyn ClientCertVerifier> {
-        // This function is needed because `ClientCertVerifier` is only reachable if the
-        // `dangerous_configuration` feature is enabled, which makes coercing hard to outside users
-        Arc::new(Self)
-    }
-}
+/// Turns off client authentication. In contrast to using
+/// `WebPkiClientVerifier::builder().allow_unauthenticated().build()`, the `NoClientAuth`
```
djc, line 653: `builder()` is actually not possible anymore.

rustls/src/server/verifier_builder.rs
cpu: I can see the advantage to only needing to handle error results once. Switched to this approach.

rustls/src/server/verifier_builder.rs
cpu: Changed to mutate self.

rustls/src/server/verifier_builder.rs
cpu: Good catch. Updated the description to say they're appended vs merged.

rustls/src/verify.rs
cpu: I lightly prefer the trait as-is. I think it's easier to implement defaults at the trait level as-is.

djc: Fair enough.

djc:
LGTM. I weakly prefer keeping it as a separate commit.

cpu:
> I weakly prefer keeping it as a separate commit.

Sounds good, I'll leave it tacked on the end. Thanks!

rustls/src/server/verifier_builder.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Once built, the provided `Arc<dyn ClientCertVerifier>` can be used with a Rustls
+    /// [crate::server::ServerConfig] to configure client certificate validation using
+    /// [`with_client_cert_verifier`][crate::ConfigBuilder<ClientConfig, WantsVerifier>::with_client_cert_verifier].
+    ///
+    /// # Errors
+    /// This function will return a `ClientCertVerifierBuilderError` if:
+    /// 1. No trust anchors have been provided.
+    /// 2. DER encoded CRLs have been provided that can not be parsed successfully.
+    pub fn build(self) -> Result<Arc<dyn ClientCertVerifier>, ClientCertVerifierBuilderError> {
+        match self.roots.is_empty() {
```
djc, line 57: Similar to the webpki thing from earlier, maybe just make this an early returning `if`?

rustls/src/verify.rs
```diff
@@ -547,68 +610,13 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
...
-        self.inner
-            .verify_client_cert(end_entity, intermediates, now)
-    }
+/// Controls how the [WebPkiClientVerifier] handles anonymous clients.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AnonymousClientPolicy {
+    /// Clients that do not present a client certificate are allowed.
+    Allow,
+    /// Clients that do not present a client certificate are forbidden.
+    Forbid,
```
djc, line 619: `Forbid` -> `Deny`?

examples/src/bin/limitedclient.rs
```diff
@@ -11,7 +11,7 @@ use rustls::OwnedTrustAnchor;
 
 fn main() {
     let mut root_store = rustls::RootCertStore::empty();
-    root_store.add_server_trust_anchors(
+    root_store.add_trust_anchors(
```
djc, line 14: Maybe we should do this as an alias on the 0.21.x branch with a `#[deprecated]` to help folks transition?

djc (approved):
Should we get a @jsha review for FFIability?

rustls/src/server/verifier_builder.rs
cpu: Done.

rustls/src/verify.rs
cpu: Done.

examples/src/bin/limitedclient.rs
cpu: PTAL: https://github.com/rustls/rustls/pull/1383

cpu: Done: https://github.com/rustls/rustls/releases/tag/v%2F0.21.6

jsha:
This looks good from an FFI-ability standpoint. We'll want two new types:

 - `*rustls_client_cert_verifier`, corresponding to `Arc<dyn ClientCertVerifier>`
 - `*rustls_web_pki_client_cert_verifier_builder`, corresponding to `Box<ClientCertVerifierBuilder>`. Note the slightly odd naming. I'm thinking of `web_pki_cilent_cert_verifier` as the name because the main entry point seems to be `WebPkiClientVerifier::builder`. But that function returns `ClientCertVerifierBuilder` so alternately we could name this type `*rustls_client_cert_verifier_builder`.
 
Also, since `build()` consumes `self`, we'll need to follow the "[safer into](https://github.com/rustls/rustls-ffi/issues/244)" pattern. So really that second type should correspond to `Box<Option<ClientCertVerifierBuilder>>`, so that we can `take()` the option on successful build and still have something to unconditionally free.

Also we'll have to make a decision on whether to have a new error return type in `rustls-ffi` corresponding to the new `ClientCertVerifierBuilderError`, or whether we'll roll it into additional values for the `rustls_error` enum.

cpu:
@jsha thanks for taking a look + providing some guidance. I can run with that and draft something that uses a Cargo patch to get ahead of the eventual tagged release w/ this API surface.

ctz (approved):

cpu:
>  thanks for taking a look + providing some guidance. I can run with that and draft something that uses a Cargo patch to get ahead of the eventual tagged release w/ this API surface.

I got a bit distracted this week :sweat_smile: but I've put together a branch that implements ffi wrapping over the new verifier builder API (as well as tracking other breaking 0.22.x changes). PTAL: https://github.com/rustls/rustls-ffi/pull/341

PR #1377, opened 2023-07-28
Check usage/intro docs in README.md are in sync

ctz:


djc (approved):
Nice to have this enforced! Ideally switch the order of commits around?

cpu (approved):
Great idea :scroll: :mag:

PR #1378, opened 2023-07-31
directly use core:: & alloc:: API instead of std:: re-exports

japaric:
This is a first step towards adding `no_std` support to RusTLS. It makes easier to spot what parts of the code-base depend on the standard library. The clippy lints should prevent std re-exports being added back (although the lints do have false positives: https://github.com/rust-lang/rust-clippy/issues/11159 ). When `no_std` support is implemented and tested in CI, it won't be necessary to rely on the clippy lints.

### DIFF
--- a/rustls/src/bs_debug.rs
+++ b/rustls/src/bs_debug.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use core::fmt;
 
 /// Alternative implementation of `fmt::Debug` for byte slice.
 ///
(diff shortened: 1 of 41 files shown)


cpu:
:wave: Could you add a PR description with the rationale for the change? 

I think we'd also prefer that the commit history be a little bit cleaner too (e.g. folding the `rust fmt` fix into the original commits).

Thanks!

japaric:
Thanks for the review. I believe I've addressed all your comments.

cpu (approved):
Thanks!

djc (approved):

PR #1382, opened 2023-08-01
docs: Add note on unsupported self-signed certificates

exdx:
This clarifies the rustls position on self-signed certificates.
Users writing tests using rustls should be aware that a
self-signed cert won't work as expected.

Signed-off-by: Dan Sover <dan.sover@avalabs.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ rustls does not and will not support:
 * Compression.
 * Discrete-log Diffie-Hellman.
 * Automatic protocol version downgrade.
+* Self-signed certificates.
 
 There are plenty of other libraries that provide these features should you
 need them.
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -48,6 +48,7 @@
 //! * Compression.
 //! * Discrete-log Diffie-Hellman.
 //! * Automatic protocol version downgrade.
+//! * Self-signed certificates.
 //!
 //! There are plenty of other libraries that provide these features should you
 //! need them.


djc:
@ctz now that we control webpki, do you want to maintain that self-signed certificates are non-feature (that is, are intentionally not supported)? Or should we add an opt-in `dangerous_config()` feature to support them somehow?

exdx:
Yes, I noticed there was a test https://github.com/rustls/rustls/blob/1776e0ba3b9df78134b2d04f2fb70496bf209595/connect-tests/tests/badssl.rs#L126 that supports self-signed certs but only in a dangerous configuration. I can update the PR to reflect that.

exdx:
My personal opinion is that this is helpful -- developers looking to use rustls might see this and know which configuration to use for self-signed certs, in tests especially.

README.md
```diff
@@ -75,6 +75,7 @@ rustls does not and will not support:
 * Compression.
 * Discrete-log Diffie-Hellman.
 * Automatic protocol version downgrade.
+* Self-signed certificates (except with dangerous configuration enabled)
```
cpu, line 78:
```suggestion
* Self-signed certificates
```

I don't think it's accurate to say that Rustls supports self-signed certificates with the dangerous config option. 

The BoGo test you pointed to is passing `--insecure` to `tlsclient-mio`, and that ultimately ends up **disabling** certificate validation entirely (:warning:!): https://github.com/rustls/rustls/blob/460934b1b06670e8e84c3e99beadb6dfd8eee21c/examples/src/bin/tlsclient-mio.rs#L357-L363

In particular this means we're not:
* Checking that the self-signed certificate's signature is valid
* Checking that the self-signed certificate isn't expired
* Checking that the self-signed certificate has a subject name matching the name that was connected to

The mechanism that `tlsclient-mio` uses for the above is also an implementation detail specific to the example (a custom `ServerCertVerifier` impl), and not something you can get out-of-box with the Rustls library.

That's a lot of words (sorry!) but I think the right change here is to list self signed certificates as unsupported without further clarification. Does that make sense?

cpu (changes requested):
Thanks for the PR! I have a request for a change w.r.t the wording being added.

exdx:
> Thanks for the PR! I have a request for a change w.r.t the wording being added.

Sure, that makes sense to me. I'll change it. I'm on vacation for a week or so but will update afterwards. Thanks.

exdx:
@djc @cpu I have went ahead and updated the README. The PR is ready to go now 👍

cpu:
@exdx Thanks, the update looks good. Could you squash the branch history down into just one commit updating `README.md`?

ctz:
> now that we control webpki, do you want to maintain that self-signed certificates are a non-feature

I think such a thing would have to be a separate, explicit API at the webpki-level. Which would call for a separate, explicit API at the rustls level to call it.

When people say "self-signed certificates", they are generally saying "turn off the security, I don't care" rather than asking caring much about the exact shape of the certificate chain -- we have an extension point and example code for doing that already.

I'm not sure just changing webpki to deal with self signed certs would suffice either -- the certificates tend to come from openssl one-liners -- far away from what is acceptable under the BRs. It seems like it wouldn't be useful to say "we support self-signed certs, but you have a to have `basicConstraints`, and `subjectAlternativeNames` extensions" etc.

I think that it's generally harder to generate a reasonable self-signed certificate and pin it than just using something like [`mkcert`](https://github.com/FiloSottile/mkcert)?

cpu:
> @exdx Thanks, the update looks good.

Also, [as pointed out by CI failure](https://github.com/rustls/rustls/actions/runs/5866749170/job/15906556978?pr=1382) I think the updated text needs to be in `lib.rs` and then updated in `README.md` with `admin/pull-readme`.

exdx:
@cpu Updated!

cpu (approved):
Looks good. Thanks for landing the revisions :-)

PR #1388, opened 2023-08-03
Move to using prerelease rustls-webpki 0.102.0-alpha.0

ctz:
This is intended to just maintain the status-quo, not take advantage of the new features in this release.

### DIFF
--- a/rustls/src/verify.rs
+++ b/rustls/src/verify.rs
@@ -18,23 +18,23 @@ use ring::digest::Digest;
 use alloc::sync::Arc;
 use std::time::SystemTime;
 
-type SignatureAlgorithms = &'static [&'static webpki::SignatureAlgorithm];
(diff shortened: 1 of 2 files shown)


rustls/src/verify.rs
```diff
@@ -533,14 +533,25 @@ impl ClientCertVerifier for AllowAnyAuthenticatedClient {
             .map(|crl| crl as &dyn webpki::CertRevocationList)
             .collect::<Vec<_>>();
 
+        let revocation = if crls.is_empty() {
```
djc, line 536: Maybe spell this as `(!crls.is_empty()).then(|| ..)` instead?

djc (approved):

rustls/src/verify.rs
ctz: going to leave this as-is, as i think @cpu will be deleting it soon to expose the more flexible CRL APIs

PR #1389, opened 2023-08-03
Split out webpki callers into separate module

ctz:
This is drawing a line around everything that directly calls into the `webpki` crate. Most of this is from the pluggable crypto branch, but I'd like to get it on main as it is the most annoying part of that to rebase :)

At the end of this, `src/verify.rs` is generic to any sort of certificate verifier. All direct calls to the webpki crate exist in `src/webpki/*`.


rustls/src/tls13/mod.rs
```diff
@@ -198,3 +198,20 @@ impl MessageDecrypter for Tls13MessageDecrypter {
         Ok(msg.into_plain_message())
     }
 }
+
+/// The set of schemes we support for signatures and
+/// that are allowed for TLS1.3.
+///
+/// This prevents (eg) RSA_PKCS1_SHA256 being offered or accepted
+/// in TLS1.3.
+pub(crate) fn is_sigscheme_supported_in_tls13(sigscheme: &SignatureScheme) -> bool {
```
cpu, line 208: small nit: WDYT about dropping the `_in_tls13` suffix. 

At the call-sites, rather than importing the `fn` directly I think it would be clearer to import `tls13` and then call it as `tls13::is_sigscheme_supported`. 

That would remove the redundancy in the fn name but still keeps it clear that the context is TLS 1.3.

rustls/src/verify.rs
```diff
@@ -405,16 +397,15 @@ pub struct WebPkiServerVerifier {
 
 #[allow(unreachable_pub)]
 impl WebPkiServerVerifier {
-    /// Constructs a new `WebPkiVerifier`.
```
cpu, line 408: oops, missed this one in my rename, thanks :)

rustls/src/verify.rs
```diff
@@ -427,6 +418,26 @@ impl WebPkiServerVerifier {
             SignatureScheme::RSA_PKCS1_SHA256,
         ]
     }
+
+    /// An full implementation of `ServerCertVerifier::verify_tls12_signature` or
```
cpu: nit: `An full` -> `A full`

rustls/src/verify.rs
```diff
@@ -427,6 +418,26 @@ impl WebPkiServerVerifier {
...
+    /// `ClientCertVerifier::verify_tls12_signature`.
+    pub fn default_verify_tls12_signature(
+        message: &[u8],
+        cert: &Certificate,
+        dss: &DigitallySignedStruct,
+    ) -> Result<HandshakeSignatureValid, Error> {
+        verify_signed_struct(message, cert, dss)
+    }
+
+    /// An full implementation of `ServerCertVerifier::verify_tls13_signature` or
```
cpu: ditto the nit above.

rustls/src/tls13/mod.rs
```diff
@@ -215,3 +216,24 @@ pub(crate) fn is_sigscheme_supported_in_tls13(sigscheme: &SignatureScheme) -> bo
             | SignatureScheme::ED25519
     )
 }
+
+/// Constructs the signature message specified in section 4.4.3 of RFC8446.
+pub(crate) fn construct_tls13_client_verify_message(handshake_hash: &Digest) -> Vec<u8> {
```
cpu, line 221: this is another place where I wonder whether it makes sense to drop the `tls13` portion of the name and let module-qualified references to the `fn` disambiguate from other protocol versions. 

I think the existing name made more sense when they lived in the general `verify.rs` module, but less so now.

WDYT?

cpu (commented):
Overall looks great! Just a few nits to raise for discussion from my side.

rustls/src/tls13/mod.rs
djc: Yes, I wanted to suggest this too when skimming earlier. Or, go the other way around and just attach this as a method to `SignatureScheme`.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -226,6 +251,28 @@ impl client::ServerCertVerifier for DummyServerAuth {
...
         Ok(client::ServerCertVerified::assertion())
     }
+
+    fn verify_tls12_signature(
+        &self,
+        message: &[u8],
+        cert: &rustls::Certificate,
+        dss: &DigitallySignedStruct,
+    ) -> Result<HandshakeSignatureValid, Error> {
+        WebPkiServerVerifier::default_verify_tls12_signature(message, cert, dss)
```
djc, line 261: I wonder if there's a way we can factor this further to avoid the duplication here? We have identical `verify_tls12_signature()` and `verify_tls13_signature()` methods for both `ClientCertVerifier` and `ServerCertVerifier` and we end up with the `ClientCertVerifier` methods calling into the `ServerCertVerifier` methods to work around that.

The signature verification is different from other parts of the certificate verification because (a) no verifier state is involved and (b) it's the same across client and server. Feels like it might make sense to extract a separate trait for this?

rustls/src/tls13/mod.rs
djc: I agree that would be better -- as is, the names feel like what is sometimes described as stuttering (also, these names are already pretty long).

rustls/src/server/handy.rs
```diff
@@ -166,32 +169,9 @@ impl ResolvesServerCertUsingSni {
...
-        // valid.
-        let general_error =
-            || Error::General("The server certificate is not valid for the given name".to_string());
-
-        let name = webpki::DnsNameRef::try_from_ascii(checked_name.as_ref().as_bytes())
-            .map_err(|_| general_error())?;
-
-        end_entity_cert
-            .verify_is_valid_for_subject_name(webpki::SubjectNameRef::DnsName(name))
-            .map_err(|_| general_error())?;
```
djc, line 194: Would be nice to replace the substitution of this with `ParsedCertificate()` and `verify_server_name()` a separate intermediate commit.

rustls/src/webpki/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub(crate) mod anchors;
```
djc, line 1: Consider keeping these modules private and present a single interface to the rest of the crate from this module through re-exports from the inner modules?

djc (commented):
This is a good step forward, like the improved modularity.

rustls/src/tls13/mod.rs
ctz: have done the latter.

rustls/src/server/handy.rs
ctz: good point - done

rustls/src/server/handy.rs
```diff
@@ -155,7 +158,8 @@ impl ResolvesServerCertUsingSni {
     pub fn add(&mut self, name: &str, ck: sign::CertifiedKey) -> Result<(), Error> {
         let checked_name = DnsNameRef::try_from(name)
             .map_err(|_| Error::General("Bad DNS name".into()))
-            .map(|dns| dns.to_lowercase_owned())?;
+            .map(|name| name.to_lowercase_owned())?;
+        let server_name = ServerName::DnsName(checked_name.clone());
```
djc, line 161: I think we should be able to avoid the `.clone()` here by reordering a bit?

rustls/examples/internal/bogo_shim.rs
ctz: Yeah this is a bit of a separate concern from certificate chain building, and it's mainly like this because it draws a line between things that require x509 certificate parsing and everything else, and I'm not sure it makes sense to have two separate configuration surfaces for "stuff that needs to deal with certificates"?

In future commits I'm planning to introduce some shared configuration between these which is the [signature verification schemes to be used by webpki](https://github.com/rustls/rustls/blob/4a4d61b5534328254f02110be6ef582a13551bbc/rustls/src/webpki/verify.rs#L480-L504) -- which would be something shared by a certificate verifier for chain building and signature verification (though, still, the configuration is separate for both, but I think it makes sense for that to be declared together).

rustls/src/server/handy.rs
ctz: have managed this by moving the `DnsName` into and then out of the `ServerName` around the name verification, but better ideas welcome!

rustls/src/server/handy.rs
```diff
@@ -166,36 +172,14 @@ impl ResolvesServerCertUsingSni {
...
-        end_entity_cert
-            .verify_is_valid_for_subject_name(webpki::SubjectNameRef::DnsName(name))
-            .map_err(|_| general_error())?;
-
-        let as_str: &str = checked_name.as_ref();
-        self.by_name
-            .insert(as_str.to_string(), Arc::new(ck));
+        ck.end_entity_cert()
+            .and_then(ParsedCertificate::try_from)
+            .and_then(|cert| verify_server_name(&cert, &server_name))?;
```
cpu, line 176: I think this refactoring has the effect of not rejecting an empty certificate chain, or a malformed end entity cert. Were you thinking the value of that early err return isn't worthwhile?

I know this isn't security sensitive, but in the absence of a counter argument it feels like we could keep the early misconfiguration detections while still relying on `verify_server_name` for the heavy lifting.

cpu (approved):

rustls/src/server/handy.rs
djc: Are you sure you understand `and_then()` correctly?

ctz: I think this is ok -- in this commit you can see the tests for those cases changing to track the different errors being produced -- so a difference in behaviour but the error checking is still happening.

- empty cert chain: `Error::General` -> `Error::NoCertificatesPresented`
- malformed cert: `Error::General` -> `Error::InvalidCertificate(CertificateError::BadEncoding)`
- name doesn't appear in cert: `Error::General` -> `Error::InvalidCertificate(CertificateError::NotValidForName)`

cpu: :sweat_smile: My bad.

I was confused and thinking about `Option.and_then` yielding `None`.

Nevermind me :bow:

djc (approved):

Issue #1372, opened 2023-07-25
Ergonomics of ConfigBuilder


jsha:
In https://github.com/rustls/rustls/pull/1365#issuecomment-1645569939, @cpu asked for more of my thoughts on the ergonomics of ConfigBuilder's typestate-based API. I figured I'd hoist them into their own issue.

## Docs

rustdoc [doesn't do a good job of rendering types that have multiple impl blocks](https://github.com/rust-lang/rust/issues/85492). For instance, in ConfigBuilder's docs, there's are two entries in the sidebar named `with_client_auth_cert`: [1](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html#method.with_client_auth_cert) [2](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html#method.with_client_auth_cert-1). When you click on them, there's no indication that they are on different types (`ConfigBuilder<ClientConfig, WantsTransparencyPolicyOrClientCert>` vs `ConfigBuilder<ClientConfig, WantsClientCert>`), unless you scroll up or collapse the doc sections. This is a problem for rustdoc to fix, but speaking as a rustdoc maintainer it's a bit challenging - I haven't thought of a great way to do it yet.

## All or Nothing

As a user I think of a config as having various defaults, any of which I can change independently. The default ClientConfig is built like:

```
ClientConfig::builder()
    .with_safe_defaults()
    ...
```

But if I want to set "only use TLS 1.3," I have to not only call the `with_protocol_versions` accessor, I have to expand the list to include two settings I _don't_ want to change:

```
ClientConfig::builder()
    .with_safe_default_cipher_suites()
    .with_safe_default_kx_groups()
    .with_protocol_versions(&[rustls::version::TLS13])
...
```

And the selection of those settings feels arbitrary. Why do those two settings become mandatory when I choose to override the protocol versions? Why not `alpn_protocols` or `resumption`? I happen to know that it's because they control code paths that the linker may optimize away if the user doesn't want them, but it feels like a lot of implementation details being exposed.

## Order dependency

The above example is valid, but this is not:

```
ClientConfig::builder()
    .with_protocol_versions(&[rustls::version::TLS13])
    .with_safe_default_cipher_suites()
    .with_safe_default_kx_groups()
...
```

[...]

PR #1390, opened 2023-08-03
ConfigBuilder: expand documentation

jsha:
De-emphasize the linker-friendliness in favor of explaining to the user what they need to do and in what order to do it.

Explicitly explain the type parameters.

Add separate intro sections for server and client.

Shorten examples by using with_safe_defaults() in examples that aren't talking about protocol primitives.

After filing #1372 I wanted to see how far we could get in the current paradigm with more documentation.

Rendered: https://rustdoc.crud.net/jsha/configbuilder-doc/rustls/struct.ConfigBuilder.html

### DIFF
--- a/rustls/src/builder.rs
+++ b/rustls/src/builder.rs
@@ -6,66 +6,77 @@ use crate::versions;
 use core::fmt;
 use core::marker::PhantomData;
 
-/// Building a [`ServerConfig`] or [`ClientConfig`] in a linker-friendly and
-/// complete way.
+/// A [builder] for [`ServerConfig`] or [`ClientConfig`] values.
 ///
-/// Linker-friendly: meaning unused cipher suites, protocol
-/// versions, key exchange mechanisms, etc. can be discarded
-/// by the linker as they'll be unreferenced.
+/// To get one of these, call [`ServerConfig::builder()`] or [`ClientConfig::builder()`].
 ///
-/// Complete: the type system ensures all decisions required to run a
-/// server or client have been made by the time the process finishes.
+/// To build a config, you must make at least three decisions (in order):
 ///
-/// Example, to make a [`ServerConfig`]:
+/// - Which protocol primitives should be supported (cipher suites, key exchange groups, protocol versions)?
+/// - How should this client or server verify certificates provided by its peer?
+/// - What certificates should this client or server present to its peer?
 ///
-/// ```no_run
-/// # use rustls::ServerConfig;
-/// # use rustls::crypto::ring::Ring;
-/// # let certs = vec![];
-/// # let private_key = rustls::PrivateKey(vec![]);
-/// ServerConfig::<Ring>::builder()
-///     .with_safe_default_cipher_suites()
-///     .with_safe_default_kx_groups()
-///     .with_safe_default_protocol_versions()
-///     .unwrap()
-///     .with_no_client_auth()
-///     .with_single_cert(certs, private_key)
-///     .expect("bad certificate/key");
+/// For settings besides these, see the fields of [`ServerConfig`] and [`ClientConfig`].
+///
+/// The usual choice for protocol primitives is to call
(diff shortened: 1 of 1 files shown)


rustls/src/builder.rs
```diff
@@ -6,34 +6,101 @@ use crate::versions;
...
+/// (if any) to send to the peer. The methods to do this are specific to whether you're building a ClientConfig
+/// or a ServerConfig, as tracked by the [`ConfigSide`] type parameter on the various impls of ConfigBuilder.
+///
+/// # ClientConfig certificate configuration
+///
+/// For a client, _certificate verification_ must be configured either by calling one of:
+///  - [`ConfigBuilder::with_root_certificates`] or
+///  - [`ConfigBuilder::with_custom_certificate_verifier`] - requires dangerous_configuration feature flag
+///
+/// Next, _certificate sending_ (also known as "client authentication", "mutual TLS", or "mTLS") must configured
```
djc, line 74: "must configured or disabled" -> "must be configured or disabled"?

djc (approved):
This is great!

rustls/src/builder.rs
```diff
@@ -47,56 +114,43 @@ use core::marker::PhantomData;
...
-///    on key exchange groups: typically by calling
-///    [`ConfigBuilder<S, WantsKxGroups>::with_safe_default_kx_groups()`].
-/// 3. Now you must make
-///    a decision on which protocol versions to support, typically by calling
-///    [`ConfigBuilder<S, WantsVersions>::with_safe_default_protocol_versions()`].
-/// 5. Now see [`ConfigBuilder<ClientConfig, WantsVerifier>`] or
-///    [`ConfigBuilder<ServerConfig, WantsVerifier>`] for further steps.
+/// Additionally, ServerConfig and ClientConfig are parameterized by `C`, a [`CryptoProvider`],
+/// which determines a cryptographic backend to use (for instance, `ring`). That type parameter
+/// used in several of the `State` types as well.
```
cpu, line 141: small typo:

```suggestion
/// which determines a cryptographic backend to use (for instance, `ring`). That type parameter
/// is used in several of the `State` types as well.
```

rustls/src/builder.rs
```diff
@@ -6,34 +6,101 @@ use crate::versions;
...
+/// ClientConfig::<Ring>::builder()
+///     .with_safe_defaults()
+/// //  ...
+/// # ;
+///
+/// ServerConfig::<Ring>::builder()
+///     .with_safe_defaults()
+/// //  ...
+/// # ;
+/// ```
```
cpu, line 38: How about moving this section explaining that the usual choice is `with_safe_defaults` to be immediately after the description of the three choices to be made, moving the "additional configuration options" to after the safe defaults discussion?

If I'm someone reading these docs looking for the fastest route from docs -> a working solution I think having the safe defaults path presented as soon as possible is helpful. WDYT?

cpu (approved):
Really nice cleanup! Thank you @jsha. I flagged one typo and made one suggestion to consider RE: ordering of text.

rustls/src/builder.rs
jsha: My goal with the "additional configuration options" verbiage is this: up front, we're telling people they "must" make 3 configuration decisions. I wanted to avoid seeming like those were the only pieces of configuration available, or leave them sifting for information about what else they can configure.

I wound up revising that paragraph to be shorter and say more of what I meant:

> For settings besides these, see the fields of [ServerConfig](https://github.com/rustls/rustls/pull/server/struct.ServerConfig.html) and [ClientConfig](https://github.com/rustls/rustls/pull/client/struct.ClientConfig.html).

Let me know what you think.

ctz (approved):

rustls/src/builder.rs
cpu: The updated version seems better, thanks!

cpu (approved):

PR #1391, opened 2023-08-04
examples: IPv6 for mio client/server examples.

cpu:
Just a couple small things I noticed while using the client/server to dig into other issues.

I verified the connect tests continue to work in addition to the normal unit tests.

## examples: IPv6 for mio client/server examples.

In the server, bind `::` instead of `0.0.0.0`.

In the client:
 * delete a comment about `openssl s_client/s_server` not supporting IPv6 - we don't test these examples against OpenSSL anymore.
 * rework `lookup_ipv4` as `lookup_ip`, returning the first socket address whether or not it's IPv4.
   
 ## examples: IPv6 for mio client/server examples.

As described.


examples/src/bin/tlsclient-mio.rs
```diff
@@ -243,20 +243,17 @@ struct Args {
     arg_hostname: String,
 }
 
-// TODO: um, well, it turns out that openssl s_client/s_server
-// that we use for testing doesn't do ipv6.  So we can't actually
-// test ipv6 and hence kill this.
-fn lookup_ipv4(host: &str, port: u16) -> SocketAddr {
+fn lookup_ip(host: &str, port: u16) -> SocketAddr {
```
ctz, line 246: i think we could just delete this, as `TcpStream::connect` can just be passed `(args.arg_hostname.as_str(), port)` instead, and will connect to all the possible addresses that resolve from that: https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html

ctz (approved):

cpu:
Pretty boring updates to example code, going to throw this in the merge queue without waiting for more reviews.

Issue #1369, opened 2023-07-23
`DecryptError` in case of unknown certificate


thegreathir:
The `OpenSSL` is sending unknown CA alert without encryption. It's different from browsers that are sending every alert after handshake encrypted. When you are trying to connect to a `rustls`  server with a `OpenSSL` client without knowing the server's CA, `rustls` will produce `DecryptError` instead of `AlertReceived`.

To reproduce run a `rustls` server:
```bash
cargo run --bin tlsserver-mio -- --certs ~/tmp/cert/server.crt --key ~/tmp/cert/server.key -p 8000 http
```
And an `OpenSSL` client (`pip` is used here):
```bash
pip3 install -i https://127.0.0.1:8000 --trusted-host sample.host.com package_name
```

djc:
I can test this later, but I think this might be due to refactoring I worked on in the MessageDeframer. It seems likely we need a similar path for Alert as we have for ChangeCipherSpec.

If you want to try your hand at a PR and/or turn this failure scenario into an integration test, that would be great!

djc:
What version are you using? I tried this:

```
cargo run --bin tlsserver-mio -- --certs test-ca/ecdsa/end.cert --key test-ca/ecdsa/end.key -p 8000 http
```

And got this:

```
djc-2021 ~ $ pip3 install -i https://127.0.0.1:8000 --trusted-host sample.host.com package_name
Looking in indexes: https://127.0.0.1:8000
WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1002)'))': /package-name/
WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1002)'))': /package-name/
WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1002)'))': /package-name/
[...]

PR #1392, opened 2023-08-04
relax handling of TLS 1.3 plaintext alerts pre-encrypted data exchange

cpu:
## Description

This branch addresses #1369 while still maintaining protection against unexpected plaintext alerts.

### record_layer: reorder members.

This change moves the members of the `RecordLayer` type to match preferred convention:

* Ordered by visibility, `pub(crate)` parts first, private parts last.
* Ordered by complexity, more involved functions first, trivial functions later.

### record_layer: track whether decryption has occurred.
This commit updates the `RecordLayer` to maintain state about whether a decryption has ever occurred. This is maintained separate from the `read_seq` state that can be overwritten when a rekey occurs.

The net effect is that once a `RecordLayer` has decrypted a message successfully it will remember this fact across any number of rekeys. This information is useful for the deframer layer where we want to make a determination about how to handle a received plaintext alert based on whether any decryptions have occurred prior to the alert.

### deframer: remove stale comment.
The described behaviour in the removed part of the comment isn't happening at the location of the comment, but later in the code. We're only considering `ChangeCipherSpec` for early return, not returning an err for non-handshake messages while joining.

### deframer: allow plaintext alerts in early 1.3 HS.
Some TLS 1.3 implementations send plaintext alerts (e.g. for an unknown certificate issuer) early in the handshake.

Trying to decrypt these messages will produce a decrypt error (because they're plaintext!). We also don't want to allow plaintext alerts to be received after encrypted records have been exchanged, since this could allow an active adversary to inject alerts.

As a compromise to support clients that send a plaintext alert before any encrypted data, we adjust the deframer in this commit to pass through plaintext alerts iff:

* The message type is alert, (e.g. not application data, etc)
* There have been no encrypted records received yet.
* The message payload is no more than 2 bytes in size (matching an expected plaintext alert payload).
* The negotiated protocol version is TLS 1.3 - in TLS 1.2 the CCS messages make whether to expect plaintext or not unambiguous. It's only for TLS 1.3 that we need the heuristics mentioned above.

This retains protection against plaintext alerts being sent after encrypted content while still allowing the server to log the correct
alert in the early-handshake condition, instead of a decrypt error.

## Manual testing

In addition to the unit test coverage, you can reproduce the situation from #1369 with this branch manually using `tlsserver-mio` and an `openssl s_client` client.

1. Start a server built from this branch:
```bash
  cargo run \
    --package rustls-examples \
    --bin tlsserver-mio -- \
        -p 8080 \
        --certs test-ca/ecdsa/end.cert \
        --key test-ca/ecdsa/end.key \
        --verbose \
        http
```
2. Connect to it using a TLS 1.3 `openssl s_client`:
```bash
openssl \
  s_client \
    -connect localhost:8080 \
    --verify_return_error \
    -tls1_3 </dev/null
```
3. In the server output, observe the correct alert received:
> [2023-08-04T19:10:09Z ERROR tlsserver_mio] cannot process packet: AlertReceived(UnknownCA)

4. Now repeat the process, but rebuild `tlsserver-mio` from the code in `main`. You should observe different server output, notably:
> [2023-08-04T19:11:58Z ERROR tlsserver_mio] cannot process packet: DecryptError

Resolves https://github.com/rustls/rustls/issues/1369

### DIFF
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -87,8 +87,7 @@ impl MessageDeframer {
                 }
             };
 
-            // If we're in the middle of joining a handshake payload and the next message is not of
-            // type handshake, yield an error. Return CCS messages immediately without decrypting.
+            // Return CCS messages immediately without decrypting.
             let end = start + rd.used();
             if m.typ == ContentType::ChangeCipherSpec && self.joining_hs.is_none() {
                 // This is unencrypted. We check the contents later.
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -39,7 +39,11 @@ impl MessageDeframer {
     /// Returns an `Error` if the deframer failed to parse some message contents or if decryption
     /// failed, `Ok(None)` if no full message is buffered or if trial decryption failed, and
     /// `Ok(Some(_))` if a valid message was found and decrypted successfully.
-    pub fn pop(&mut self, record_layer: &mut RecordLayer) -> Result<Option<Deframed>, Error> {
+    pub fn pop(
+        &mut self,
+        record_layer: &mut RecordLayer,
+        negotiated_version: Option<ProtocolVersion>,
+    ) -> Result<Option<Deframed>, Error> {
         if let Some(last_err) = self.last_error.clone() {
             return Err(last_err);
         } else if self.used == 0 {
@@ -87,9 +91,29 @@ impl MessageDeframer {
                 }
             };
 
-            // Return CCS messages immediately without decrypting.
+            // Return CCS messages and early plaintext alerts immediately without decrypting.
             let end = start + rd.used();
-            if m.typ == ContentType::ChangeCipherSpec && self.joining_hs.is_none() {
+            let version_is_tls13 = matches!(negotiated_version, Some(ProtocolVersion::TLSv1_3));
+            let allowed_plaintext = match m.typ {
+                // CCS messages are always plaintext.
+                ContentType::ChangeCipherSpec => true,
+                // Alerts are allowed to be plaintext if-and-only-if:
+                // * The negotiated protocol version is TLS 1.3. - In TLS 1.2 it is unambiguous when
+                //   keying changes based on the CCS message. Only TLS 1.3 requires these heuristics.
+                // * We have not yet decrypted any messages from the peer - if we have we don't
+                //   expect any plaintext.
+                // * The payload size is indicative of a plaintext alert message.
+                ContentType::Alert
+                    if version_is_tls13
+                        && !record_layer.has_decrypted()
+                        && m.payload.0.len() <= 2 =>
+                {
+                    true
+                }
+                // In other circumstances, we expect all messages to be encrypted.
+                _ => false,
+            };
+            if self.joining_hs.is_none() && allowed_plaintext {
                 // This is unencrypted. We check the contents later.
                 self.discard(end);
                 return Ok(Some(Deframed {
@@ -514,13 +538,21 @@ mod tests {
     }
 
     fn pop_first(d: &mut MessageDeframer, rl: &mut RecordLayer) {
-        let m = d.pop(rl).unwrap().unwrap().message;
+        let m = d
+            .pop(rl, None)
+            .unwrap()
+            .unwrap()
+            .message;
         assert_eq!(m.typ, ContentType::Handshake);
         Message::try_from(m).unwrap();
     }
 
     fn pop_second(d: &mut MessageDeframer, rl: &mut RecordLayer) {
-        let m = d.pop(rl).unwrap().unwrap().message;
+        let m = d
+            .pop(rl, None)
+            .unwrap()
+            .unwrap()
+            .message;
         assert_eq!(m.typ, ContentType::Alert);
         Message::try_from(m).unwrap();
     }
@@ -640,7 +672,7 @@ mod tests {
 
         let mut rl = RecordLayer::new();
         assert_eq!(
-            d.pop(&mut rl).unwrap_err(),
+            d.pop(&mut rl, None).unwrap_err(),
             Error::InvalidMessage(InvalidMessage::InvalidContentType)
         );
     }
@@ -655,7 +687,7 @@ mod tests {
 
         let mut rl = RecordLayer::new();
         assert_eq!(
-            d.pop(&mut rl).unwrap_err(),
+            d.pop(&mut rl, None).unwrap_err(),
             Error::InvalidMessage(InvalidMessage::UnknownProtocolVersion)
         );
     }
@@ -670,7 +702,7 @@ mod tests {
 
         let mut rl = RecordLayer::new();
         assert_eq!(
-            d.pop(&mut rl).unwrap_err(),
+            d.pop(&mut rl, None).unwrap_err(),
             Error::InvalidMessage(InvalidMessage::MessageTooLarge)
         );
     }
@@ -684,7 +716,11 @@ mod tests {
         );
 
         let mut rl = RecordLayer::new();
-        let m = d.pop(&mut rl).unwrap().unwrap().message;
+        let m = d
+            .pop(&mut rl, None)
+            .unwrap()
+            .unwrap()
+            .message;
         assert_eq!(m.typ, ContentType::ApplicationData);
         assert_eq!(m.payload.0.len(), 0);
         assert!(!d.has_pending());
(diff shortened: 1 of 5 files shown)


rustls/src/msgs/deframer.rs
```diff
@@ -732,4 +762,126 @@ mod tests {
         );
         assert!(input_bytes(&mut d, &message).is_err());
     }
+
+    #[test]
+    fn test_plaintext_alert_pre_encrypted() {
```
cpu, line 767: I flailed around for a while trying to figure out the best way to test this change and ultimately landed on this unit test approach with carefully mocked `MessageDecrypter`s. 

Initially I wanted to write an integration test in `tests/api.rs` - this ends up being really challenging because Rustls (sensibly) doesn't expose low level guts for writing arbitrary TLS alerts. We can send close notify alerts through the public API, but these are encrypted based on the connection's encryption state (the right behaviour!). 

To be able to tickle the code path in question we need a way to send a plaintext alert when the remote peer has its record layer set up in a state that expects to decrypt messages and I simply couldn't think of a nice way to do that from an integration test. Maybe someone else will have a clever idea?

ctz: Try something like:
```
    let (mut client, _) = make_pair(KeyType::Rsa);
    client
        .read_tls(&mut io::Cursor::new(&[
            0x15, 0x03, 0x03, 0x00, 0x02, 0x02, 0x6f,
        ]))
        .unwrap();
    assert_eq!(
        client.process_new_packets(),
        Err(Error::AlertReceived(
            rustls::AlertDescription::CertificateUnobtainable
        )),
    );
```

rustls/src/msgs/deframer.rs
```diff
@@ -87,9 +91,20 @@ impl MessageDeframer {
...
+            // * CCS messages are always plaintext.
+            // * Alerts in TLS 1.3 _may_ be received plaintext early in the handshake process,
+            //   but are not expected once encrypted data has been exchanged (e.g. read_seq >0).
+            //   We also expect all plaintext alerts to be at most 2 bytes in size.
             let end = start + rd.used();
-            if m.typ == ContentType::ChangeCipherSpec && self.joining_hs.is_none() {
+            let type_is_ccs = m.typ == ContentType::ChangeCipherSpec;
+            let early_plaintext_alert = m.typ == ContentType::Alert
+                && m.payload.0.len() <= 2
+                && record_layer.read_seq() == 0
```
ctz, line 105: I'm afraid I don't think this is right, because `read_seq` can be reset to 0 later on every receipt of a `KeyUpdate`. Instead, I think we should expose `pub(crate) fn is_decrypting(&self) -> bool` (like the existing `is_encrypting`)

ctz (changes requested):

rustls/src/msgs/deframer.rs
cpu: @ctz I don't think this will work. When reproducing the condition from #1369 the `RecordLayer` has `encrypt_state = DirectionState::Active` and `decrypt_state = DirectionState::Active`, even though no encrypted message has been received by the peer - both the `write_seq` and `read_seq` are 0.

cpu: > Did you figure out when/where the decrypt_state gets set to Active?

Yeah, the TLS 1.3 `ExpectClientHello` state `handle_client_hello` with early data disabled is setting the handshake decrypter, which moves the record layer from `DirectionState::Invalid` -> `DirectionState::Prepared` -> `DirectionState::Active` in one operation via the `KeySchedule` calling `RecordLayer.set_message_decrypter`.

We could consider another `DirectionState` variant between `Prepared` and `Active` but it feels a little clumsy :thinking:

cpu: What about having the `RecordLayer` keep a `bool` for whether it has decrypted a message separate from the `read_seq`?

The `read_seq` would get reset on when processing a `KeyUpdate` but not the proposed `bool`.

ctz: > What about having the `RecordLayer` keep a `bool` for whether it has decrypted a message separate from the `read_seq`?
> 
> The `read_seq` would get reset when processing a `KeyUpdate` but not the proposed `bool`.

Yes I think that ought to work and seems like the least bad option.

cpu: Implemented in https://github.com/rustls/rustls/pull/1392/commits/9627b16d00f57dd2e7d849ec57e5f2baa0bef9fc

cpu:
I've flipped this out of draft status. I think it's ready for review. 

The updated branch replaces the clumsy unit testing approach w/ a better integration testing approach. It also drops the usage of `read_seq`, replacing it with having the record layer track whether its ever performed a decryption.

rustls/src/record_layer.rs
```diff
@@ -150,6 +152,12 @@ impl RecordLayer {
         self.write_seq >= SEQ_HARD_LIMIT
     }
 
+    /// Return true if we have ever decrypted a message. This is used in place
```
djc, line 155: Ordering: I would typically move trivial getters like this to the bottom of the `impl`.

rustls/src/msgs/deframer.rs
```diff
@@ -87,9 +91,22 @@ impl MessageDeframer {
...
+            let type_is_alert = m.typ == ContentType::Alert;
+            let payload_is_alert = m.payload.0.len() <= 2;
+            let version_is_tls13 = matches!(negotiated_version, Some(ProtocolVersion::TLSv1_3));
+            // We consider a message an early plaintext alert if-and-only-if:
+            //  * We haven't decrypted any messages from the peer yet.
+            //  * The message content type is Alert.
+            //  * The payload size is indicative of a plaintext alert message.
+            //  * The negotiated protocol version is TLS 1.3.
+            // In all other circumstances a plaintext alert is not expected.
+            let early_plaintext_alert =
```
djc, line 107: Style nit: have you tried inlining the values from above in this statement? That would save a lot of bindings, and maybe there's room for comments on each sub-expression's line?

rustls/src/msgs/deframer.rs
```diff
@@ -87,9 +91,22 @@ impl MessageDeframer {
...
 
-            // Return CCS messages immediately without decrypting.
+            // Return CCS messages and early plaintext alerts immediately without decrypting.
             let end = start + rd.used();
-            if m.typ == ContentType::ChangeCipherSpec && self.joining_hs.is_none() {
+            let type_is_ccs = m.typ == ContentType::ChangeCipherSpec;
+            let has_decrypted = record_layer.has_decrypted();
+            let type_is_alert = m.typ == ContentType::Alert;
+            let payload_is_alert = m.payload.0.len() <= 2;
+            let version_is_tls13 = matches!(negotiated_version, Some(ProtocolVersion::TLSv1_3));
```
djc, line 100: From first principles, why are we limiting this to TLS 1.3 again?

djc (approved):

rustls/src/msgs/deframer.rs
ctz: TLS1.2 has ChangeCipherSpec messages which make it unambiguous when the keying changes (including when transitioning from plaintext to encrypted). TLS1.3 did away with those[^1] which means the ambiguity of whether an alert will be encrypted is specific to TLS1.3.

[^1]: they can still be sent for "middlebox compatibility mode", but they have no meaning and cannot be relied on

rustls/src/msgs/deframer.rs
```diff
@@ -87,10 +91,22 @@ impl MessageDeframer {
...
 
-            // If we're in the middle of joining a handshake payload and the next message is not of
-            // type handshake, yield an error. Return CCS messages immediately without decrypting.
+            // Return CCS messages and early plaintext alerts immediately without decrypting.
             let end = start + rd.used();
-            if m.typ == ContentType::ChangeCipherSpec && self.joining_hs.is_none() {
+            let type_is_ccs = m.typ == ContentType::ChangeCipherSpec;
+            let has_decrypted = record_layer.has_decrypted();
+            let type_is_alert = m.typ == ContentType::Alert;
+            let payload_is_alert = m.payload.0.len() <= 2;
```
ctz, line 99: perhaps this could be recast as something like:

```rust
let allowed_plaintext = match m.typ {
  ContentType::ChangeCipherSpec => true,
  ContentType::Alert if version_is_tls13 && !record_layer.has_decrypted && m.payload.0.len() <= 2 => true,
  _ => false,
};
(...)
if self.joining_hs.is_none() && allowed_plaintext {
  (...)
}
```

djc: Yup, I like that better than my previous suggestion.

rustls/src/msgs/deframer.rs
cpu: I will try to capture more of this in the PR description and comments.

rustls/src/msgs/deframer.rs
cpu: Sure, I'm open to that. 

I had left these individual bindings in the initial push because I found them helpful when verifying the test coverage. When it was all one big boolean expression it was more difficult to determine _why_ the decision to not try and treat the message as plaintext was reached - was it the payload size? The protocol version? Etc.

I think with the test coverage established that's not a super great reason to keep it as verbose as it is. I'll run with `match` suggestion from Ctz.

cpu: Implemented the `match` suggestion.

PR #1395, opened 2023-08-08
Enable feature(doc_auto_cfg)

### LINKED ISSUE #1393: See if `feature(doc_auto_cfg)` works to avoid repeating feature gates
At the moment we do a lot of:

```
#[cfg_attr(docsrs, doc(cfg(feature = "secret_extraction")))]
pub enum ...
```

The repetition of feature gates (once for real, once for docs) is a bit annoying, and I would guess some are missing/wrong. It seems `doc_auto_cfg` is meant to avoid this.

If it does work, then let's enable it and delete all the manual `doc(cfg())` business.

ctz:
This removes duplicated manual feature gates for documentation and leaves it to `cargo doc` to derive the same information from the actual feature gates.

I didn't find any gaps in the auto-generated features and what we had before, but now things like `rustls::cipher_suite::TLS_ECDHE_*` are correctly marked tls12-only.

fixes #1393

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -172,7 +172,6 @@ pub struct ClientConfig<C: CryptoProvider> {
     /// Allows traffic secrets to be extracted after the handshake,
     /// e.g. for kTLS setup.
     #[cfg(feature = "secret_extraction")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "secret_extraction")))]
     pub enable_secret_extraction: bool,
 
     /// Whether to send data on the first flight ("early data") in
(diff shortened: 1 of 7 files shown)


ctz:
(i think we should do this in the webpki crate too)

cpu (commented):
This isn't from your diff, but I noticed while validating the change locally that we have a rustdoc warning:

```
warning: unresolved link to `ConfigBuilder::with_custom_certificate_verifier`
  --> rustls/src/builder.rs:71:10
   |
71 | ///  - [`ConfigBuilder::with_custom_certificate_verifier`] - requires dangerous_configuration feature flag
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `ConfigBuilder` has no field or associated item named `with_custom_certificate_verifier`
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```

Would you be open to fixing that in this branch? (also might be worth a CI tweak to fail closed on warnings.)

djc (approved):

cpu:
> Ah, so we do have -Dwarnings (deny warnings) on the cargo doc job, but also that uses --all-features

Oooh, that makes sense.

ctz:
Have raised rust-lang/rust#114626 to discuss that

PR #1401, opened 2023-08-15
Generalise all crypto in `rustls::SupportedCipherSuite` (part deux)

ctz:
The end goal of these PRs is that `rustls::SupportedCipherSuite` (and everything it contains) no longer references any types provided by *ring*, and can therefore be an extension point for people to provide their own crypto without taking a dependency on *ring*.

That means that rustls starts to provide abstractions for its use of hashing (in `rustls::crypto::hash`) building on the similar HMAC abstractions recently added in `rustls::crypto::hmac`. It grows an implementation of HKDF (which is a straightforward use of HMAC) instead of using the one from *ring*.

A later PR will move every direct caller of ring into `rustls::crypto::ring`, so this module can be compiled out. This PR almost gets there, but there are some stragglers for later cleanup.


ctz:
Apologies, I have missed 3 comments on https://github.com/rustls/rustls/pull/1398 -- working on those and then I will re-mark as ready for review.

rustls/src/crypto/ring/tls12.rs
```diff
@@ -165,34 +165,20 @@ impl Tls12AeadAlgorithm for GcmAlgorithm {
 
     #[cfg(feature = "secret_extraction")]
     fn extract_keys(&self, key: &[u8], iv: &[u8], explicit: &[u8]) -> ConnectionTrafficSecrets {
-        let salt = {
```
ctz: if people agree this is a reasonable thing to do, i can squash this into the earlier commits

djc: Seems reasonable. How about making it a function that does all three, taking three `const` arguments?

ctz: Well, some need two and others need three. but then trying to take the three version and discard one of the arguments breaks type inference (and I think fixing that would mean writing out all the lengths, which I'd like to avoid).

This is what I ended up with:

```
fn slices_to_arrays<const NK: usize, const NS: usize, const NI: usize>(
    k: &[u8],
    s: &[u8],
    i: &[u8],
) -> ([u8; NK], [u8; NS], [u8; NI]) {
    // these are guaranteed true because `ConnectionTrafficSecrets` items and
    // `key_len()` are in agreement.
    debug_assert_eq!(NK, k.len());
    debug_assert_eq!(NS, s.len());
    debug_assert_eq!(NI, i.len());
    (
        k.try_into().unwrap(),
        s.try_into().unwrap(),
        i.try_into().unwrap(),
    )
}
```

which works nicely in the 3-parameter case:

```
    fn extract_keys(&self, key: AeadKey, iv: Iv) -> ConnectionTrafficSecrets {
        let (key, salt, iv) = slices_to_arrays(key.as_ref(), &iv.0[..4], &iv.0[4..]);
        ConnectionTrafficSecrets::Aes256Gcm { key, salt, iv }
    }
```

but not for two:

```
    fn extract_keys(&self, key: AeadKey, iv: Iv) -> ConnectionTrafficSecrets {
        let (key, _, iv) = slices_to_arrays(key.as_ref(), &[], &iv.0);
        ConnectionTrafficSecrets::Chacha20Poly1305 { key, iv }
    }
```

ctz: Squashed!

rustls/src/tls13/mod.rs
```diff
@@ -199,6 +209,76 @@ impl MessageDecrypter for Tls13MessageDecrypter {
     }
 }
 
+struct RingAeadFactory(&'static ring::aead::Algorithm);
```
djc: The `Factory` naming seems unidiomatic to me. Maybe `RingAeadBuilder`? Feels like these should just be moved into a `ring` module directly so we can then ditch the prefix.

Ordering: `AeadChacha20Poly1305` (and co), then `RingAeadFactory`, then `Tls13MessageEncrypter` (and `Decrypter`)?

rustls/src/tls13/mod.rs
```diff
@@ -199,6 +209,76 @@ impl MessageDecrypter for Tls13MessageDecrypter {
...
+            iv,
+        })
+    }
+
+    fn key_len(&self) -> usize {
+        self.0.key_len()
+    }
+}
+
+struct AeadChacha20Poly1305(RingAeadFactory);
```
djc: Would suggest making `Aead` a suffix rather than a prefix for these type names.

This abstraction layering feels a bit odd. I think `RingAeadFactory` actually wants to be a trait with one method (the ring `Algorithm`) and three provided methods?

rustls/src/tls13/mod.rs
```diff
@@ -277,6 +290,29 @@ impl Tls13AeadAlgorithm for AeadAes256Gcm {
...
+
+    #[cfg(feature = "secret_extraction")]
+    fn extract_keys(&self, key: AeadKey, iv: Iv) -> ConnectionTrafficSecrets {
+        let (key, salt, iv) = slices_to_arrays(key.as_ref(), &iv.0[..4], &iv.0[4..]);
+        ConnectionTrafficSecrets::Aes256Gcm { key, salt, iv }
+    }
+}
+
+#[cfg(feature = "secret_extraction")]
+fn slice_to_array<const N: usize>(slice: &[u8]) -> [u8; N] {
```
djc: Ordering?

rustls/src/crypto/cipher.rs
```diff
@@ -1,11 +1,49 @@
...
+use crate::msgs::fragmenter::MAX_FRAGMENT_LEN;
+pub use crate::msgs::message::{BorrowedPlainMessage, OpaqueMessage, PlainMessage};
 
 /// Objects with this trait can decrypt TLS messages.
 pub trait MessageDecrypter: Send + Sync {
     /// Perform the decryption over the concerned TLS message.
     fn decrypt(&self, m: OpaqueMessage, seq: u64) -> Result<PlainMessage, Error>;
+
+    /// For TLS1.3 (only), checks the length msg.payload is valid and removes the padding.
+    fn tls13_check_length_and_unpad(&self, mut msg: OpaqueMessage) -> Result<PlainMessage, Error> {
```
djc, line 87: If we're going to make this public I think we should also expose `OpaqueMessage` and `PlainMessage` as part of the public API (as in, other than via `internals`), as opaque types.

Also, I don't love how this is attached to the `MessageDecrypter` while it doesn't actually use `self` anywhere. If we're going to expose `OpaqueMessage`, maybe this becomes a method on it?

rustls/src/crypto/cipher.rs
```diff
@@ -55,7 +93,11 @@ impl Iv {
     }
 }
 
-pub(crate) fn make_nonce(iv: &Iv, seq: u64) -> [u8; NONCE_LEN] {
+/// Combine an `Iv` and sequence number to produce a unique nonce.
```
djc, line 171: It feels like maybe `Nonce` should be a newtype wrapper like `Iv`.

rustls/tests/api.rs
```diff
@@ -3319,7 +3319,7 @@ mod test_quic {
 
     #[test]
     fn test_quic_handshake() {
-        fn equal_packet_keys(x: &quic::PacketKey, y: &quic::PacketKey) -> bool {
+        fn equal_packet_keys(x: &Box<dyn quic::PacketKey>, y: &Box<dyn quic::PacketKey>) -> bool {
```
djc, line 3322: Should probably call these with `&dyn` directly rather than requiring the `Box` wrapping?

rustls/src/quic.rs
```diff
@@ -483,38 +501,51 @@ impl Secrets {
...
-pub struct HeaderProtectionKey(aead::quic::HeaderProtectionKey);
-
-impl HeaderProtectionKey {
-    fn new(suite: &'static Tls13CipherSuite, secret: &hkdf::Prk, version: Version) -> Self {
-        let alg = match suite.common.bulk {
-            BulkAlgorithm::Aes128Gcm => &aead::quic::AES_128,
-            BulkAlgorithm::Aes256Gcm => &aead::quic::AES_256,
-            BulkAlgorithm::Chacha20Poly1305 => &aead::quic::CHACHA20,
-        };
+/// All AEADs we support have 16-byte tags.
```
djc, line 527: This seems like a big commit, would be nice to split it up, in particular I'd probably introduce the trait(s) and new code first and then move code around in a separate commit.

rustls/src/crypto/ring/tls13.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut payload = Vec::with_capacity(total_len);
+        payload.extend_from_slice(msg.payload);
+        msg.typ.encode(&mut payload);
+
+        let nonce = aead::Nonce::assume_unique_for_key(make_nonce(&self.iv, seq));
+        let aad = aead::Aad::from(make_tls13_aad(total_len));
+
+        self.enc_key
+            .seal_in_place_append_tag(nonce, aad, &mut payload)
+            .map_err(|_| Error::General("encrypt failed".to_string()))?;
```
cpu, line 93: I think this should be mapping to [`Error::EncryptError`](https://github.com/rustls/rustls/blob/6bdaf04e7af97821485da115b5436913fedea874/rustls/src/error.rs#L48-L50)?

rustls/src/crypto/ring/tls13.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn encrypt(&self, msg: BorrowedPlainMessage, seq: u64) -> Result<OpaqueMessage, Error> {
+        let total_len = msg.payload.len() + 1 + self.enc_key.algorithm().tag_len();
+        let mut payload = Vec::with_capacity(total_len);
+        payload.extend_from_slice(msg.payload);
+        msg.typ.encode(&mut payload);
+
+        let nonce = aead::Nonce::assume_unique_for_key(make_nonce(&self.iv, seq));
+        let aad = aead::Aad::from(make_tls13_aad(total_len));
+
+        self.enc_key
```
cpu, line 91: editorial nit: match the whitespace in `decrypt`, e.g. no space between `nonce`/`aad` bindingsz and the `self` call.

Also seems like the `nonce` and `aad` bindings (both here and in `decrypt`) might be unnecessary and could be constructed inline? Do you think there's a readability advantage as-is?

rustls/src/crypto/ring/tls13.rs
```diff
@@ -0,0 +1,224 @@
...
+        self.tls13_check_length_and_unpad(msg)
+    }
+}
+
+struct RingAeadFactory(&'static ring::aead::Algorithm);
+
+impl RingAeadFactory {
+    fn encrypter(&self, key: AeadKey, iv: Iv) -> Box<dyn MessageEncrypter> {
+        Box::new(Tls13MessageEncrypter {
+            enc_key: aead::LessSafeKey::new(aead::UnboundKey::new(self.0, key.as_ref()).unwrap()),
```
cpu, line 128: same comment as in an earlier commit: perhaps worth a comment about the safety of `unwrap` w.r.t an externally enforced invariant?

rustls/src/tls12/mod.rs
```diff
@@ -169,6 +66,36 @@ impl fmt::Debug for Tls12CipherSuite {
...
+    fn extract_keys(&self, key: &[u8], iv: &[u8], explicit: &[u8]) -> ConnectionTrafficSecrets;
+}
+
+/// How a TLS1.2 `key_block` is partitioned.
+///
+/// nb. ciphersuites with non-zero `mac_key_length` not currently supported
+pub(crate) struct KeyBlockShape {
+    /// How long keys are.
+    ///
+    /// `enc_key_len` terminology is from the standard.
```
cpu: nit: maybe replace "the standard" with RFC 5246 A.6? 

Also being highly pedantic the RFC terminology uses `length` throughout instead of `len`, e.g. `enc_key_length` and `fixed_iv_length`.

rustls/src/crypto/ring/quic.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+impl quic::HeaderProtectionKey for HeaderProtectionKey {
+    fn encrypt_in_place(
+        &self,
+        sample: &[u8],
+        first: &mut u8,
+        packet_number: &mut [u8],
+    ) -> Result<(), Error> {
+        self.xor_in_place(sample, first, packet_number, false)
```
cpu, line 84: this isn't specific to the crypto provider work, so feel free to punt, but it feels like a place where an `enum` could replace a `bool` to increase specificity and make it easier to distinguish what the call-sites intend without referencing the `xor_in_place` impl.

rustls/src/crypto/cipher.rs
```diff
@@ -3,11 +3,69 @@ use crate::error::{Error, PeerMisbehaved};
...
+    fn extract_keys(&self, key: &[u8], iv: &[u8], explicit: &[u8]) -> ConnectionTrafficSecrets;
+}
+
+/// How a TLS1.2 `key_block` is partitioned.
+///
+/// nb. ciphersuites with non-zero `mac_key_length` not currently supported
+pub struct KeyBlockShape {
+    /// How long keys are.
+    ///
+    /// `enc_key_len` terminology is from the standard.
```
cpu, line 65: similar comment as earlier: replace "from the standard" with a more concrete pointer to rfc5246 A.6.

rustls/src/crypto/cipher.rs
```diff
@@ -47,8 +105,9 @@ fn unpad_tls13(v: &mut Vec<u8>) -> ContentType {
 }
 
 /// Objects with this trait can encrypt TLS messages.
-pub(crate) trait MessageEncrypter: Send + Sync {
-    fn encrypt(&self, m: BorrowedPlainMessage, seq: u64) -> Result<OpaqueMessage, Error>;
+pub trait MessageEncrypter: Send + Sync {
+    /// Encrypt the message `msg`.
```
cpu, line 125: Should this docstring mention the `seq` arg or do you think it's contextual enough to understand as-is?

rustls/src/tls12/mod.rs
```diff
@@ -29,7 +29,9 @@ pub struct Tls12CipherSuite {
     /// How to sign messages for authentication.
     pub sign: &'static [SignatureScheme],
 
-    pub(crate) aead_alg: &'static dyn Tls12AeadAlgorithm,
+    /// Producing a suitable `MessageDecrypter` or `MessageEncrypter`
+    /// from the raw keys.
```
cpu, line 33: nit: maybe match the pattern used by the other docstrings and say "How to produce a ... "

rustls/src/tls13/mod.rs
```diff
@@ -1,29 +1,23 @@
...
     /// Common cipher suite fields.
     pub common: CipherSuiteCommon,
-    pub(crate) hmac_provider: &'static dyn crypto::hmac::Hmac,
-    pub(crate) aead_alg: &'static dyn Tls13AeadAlgorithm,
+
+    /// How to compute HMAC with the suite's hash function.
+    pub hmac_provider: &'static dyn crypto::hmac::Hmac,
+
+    /// Producing a suitable `MessageDecrypter` or `MessageEncrypter`
+    /// from the raw keys.
```
cpu, line 18: ditto re: "Producing" -> "How to produce"

cpu (commented):
Really cool seeing this shape up, great work!

I think I have less constructive feedback than @djc but hope some of my editorial comments are helpful anyway :sweat_smile:

rustls/src/crypto/mod.rs
```diff
@@ -8,6 +8,9 @@ pub mod ring;
 /// TLS message encryption/decryption interfaces.
 pub mod cipher;
 
+/// Hashing interfaces.
+pub mod hash;
```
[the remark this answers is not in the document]
ctz: I think I like the idea that one module corresponds to one small, independent thing/feature. It means a downstream user can have a small-ish file just dealing with that feature: like https://github.com/rustls/rustls/blob/jbp-generalise-crypto-usage-pt3/provider-example/src/hash.rs which only imports that module, and can be understood in isolation against its 3-item `cargo doc` page.

rustls/src/hkdf.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+pub(crate) struct OutputLengthError;
+
+/// This is a PRK ready for use via `expand()` et al.
+pub(crate) struct Expander(Box<dyn hmac::Key>);
+
+/// This is a single "output keying material" (OKM) block output from HKDF-Expand.
+#[derive(Clone)]
+pub(crate) struct OkmOneBlock(hmac::Tag);
```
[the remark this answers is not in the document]
ctz: have done this, and also the several other places using `_one` in reference

rustls/src/hash_hs.rs
```diff
@@ -143,8 +145,8 @@ impl HandshakeHash {
     }
 
     /// Get the current hash value.
-    pub(crate) fn get_current_hash(&self) -> digest::Digest {
-        self.ctx.clone().finish()
+    pub(crate) fn get_current_hash(&self) -> hash::Output {
+        self.ctx.fork().finish()
```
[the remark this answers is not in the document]
ctz: oh, yes, that is what it's for. i was switching back and forth to try and measure the difference it made (evidently, not much)

rustls/src/tls13/key_schedule.rs
```diff
@@ -798,35 +778,68 @@ where
...
+fn hkdf_expand_aead_key(
+    expander: &hkdf::Expander,
+    aead_algorithm: &'static ring::aead::Algorithm,
+    label: &[u8],
+    context: &[u8],
+) -> AeadKey {
+    let key_len = aead_algorithm.key_len();
+    hkdf_expand_inner(expander, label, context, key_len, |e, info| {
+        let key: AeadKey = e.expand(info);
+        key.with_length(key_len)
```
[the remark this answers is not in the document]
ctz: it ends up having to be `e.expand::<AeadKey, { AeadKey::MAX_LEN }>(info)...` which is a bit wordy. i didn't realise rust doesn't infer `N` in this case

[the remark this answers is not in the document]
djc: Ah, ok.

rustls/src/tls13/key_schedule.rs
```diff
@@ -798,35 +778,68 @@ where
...
-    f(okm)
+    f(expander, info)
 }
 
-pub(crate) struct PayloadU8Len(pub(crate) usize);
-impl hkdf::KeyType for PayloadU8Len {
-    fn len(&self) -> usize {
-        self.0
-    }
+/// HKDF-Expand where the output length is a compile-time constant, and therefore
```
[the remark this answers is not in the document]
ctz: Have corrected the order of these

> Should (some of) these become helper methods on Expander instead?

Have renamed these and added commentary/references making it clear these are implementing `HKDF-Expand-Label` from the TLS1.3 RFC, not `HKDF-Expand` from the HKDF RFC.

rustls/src/crypto/cipher.rs
```diff
@@ -85,6 +73,43 @@ pub(crate) fn make_nonce(iv: &Iv, seq: u64) -> [u8; NONCE_LEN] {
...
+        Self {
+            buf: self.buf,
+            used: len,
+        }
+    }
+
+    pub(crate) const MAX_LEN: usize = 32;
+}
+
+/// Largest possible AEAD key in the ciphersuites we support.
```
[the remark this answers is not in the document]
ctz: Good spot!

rustls/src/tls13/mod.rs
ctz: i think this is in a much better place now: have deleted the three `AeadChacha20Poly1305` types which weren't adding much

ctz: oh, actually they were providing separate impls of `Tls13AeadAlgorithm` which differed for `extract_keys`. hmm

rustls/src/crypto/cipher.rs
ctz: have added a new commit that does this

rustls/src/crypto/ring/tls13.rs
ctz: good call! added a commit which does this

rustls/src/crypto/ring/tls13.rs
ctz: have sorted the inconsistent vertical whitespace here

i think the separate bindings are ok, since the expressions for constructing them are pretty long as-is

rustls/src/quic.rs
ctz: i have split this up into 3 commits:

- adding the new stuff
- removing the unused BulkAlgorithm
- moving code into crypto::ring

rustls/src/crypto/ring/quic.rs
ctz: i think i'll defer this to a later PR to avoid growing this one any further

ctz:
I think this is ready for another pass.

rustls/src/crypto/ring/tls13.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::suites::BulkAlgorithm;
+#[cfg(feature = "secret_extraction")]
+use crate::suites::ConnectionTrafficSecrets;
+use crate::suites::{CipherSuiteCommon, SupportedCipherSuite};
+use crate::tls13::{Tls13AeadAlgorithm, Tls13CipherSuite};
+
+use ring::aead;
+
+/// The TLS1.3 ciphersuite TLS_CHACHA20_POLY1305_SHA256
+pub static TLS13_CHACHA20_POLY1305_SHA256: SupportedCipherSuite =
```
cpu, line 17: Do these keep the `TLS13` prefix that stutters with the module name because of the re-export into `cipher_suites` in `lib.rs`? 

If so, should the `tls12` ciphersuites keep a similar prefix? I've been leaving other feedback that recommends dropping the stutter but I can see how the `cipher_suite` module might change the calculus for that recommendation.

rustls/src/quic.rs
```diff
@@ -2,17 +2,18 @@
...
 use crate::enums::{AlertDescription, ProtocolVersion};
 use crate::error::Error;
+use crate::hkdf;
 use crate::msgs::handshake::{ClientExtension, ServerExtension};
 use crate::server::{ServerConfig, ServerConnectionData};
-use crate::suites::BulkAlgorithm;
-use crate::tls13::key_schedule::hkdf_expand;
-use crate::tls13::{Tls13CipherSuite, TLS13_AES_128_GCM_SHA256_INTERNAL};
+use crate::tls13::key_schedule::hkdf_expand_label_block;
+use crate::tls13::key_schedule::{hkdf_expand_label, hkdf_expand_label_aead_key};
```
cpu: I think these imports should be merged:

```rust
use crate::tls13::key_schedule::{hkdf_expand_label, hkdf_expand_label_block, hkdf_expand_label_aead_key};
```

rustls/src/quic.rs
```diff
@@ -689,70 +634,49 @@ impl PacketKey {
...
     /// Decrypts incoming packets
-    pub remote: PacketKey,
+    pub remote: Box<dyn PacketKey>,
 }
 
 impl PacketKeySet {
     fn new(secrets: &Secrets) -> Self {
         let (local, remote) = secrets.local_remote();
+        let local_expander = hkdf::Expander::from_okm(local, secrets.suite.hmac_provider);
+        let remote_expander = hkdf::Expander::from_okm(remote, secrets.suite.hmac_provider);
```
cpu, line 666: nit: I think these bindings could be inlined into the `packet_key` calls.

rustls/tests/api.rs
```diff
@@ -3345,8 +3345,8 @@ mod test_quic {
             }
 
             let (x, y) = (keys(x), keys(y));
-            equal_packet_keys(&x.local.packet, &y.remote.packet)
-                && equal_packet_keys(&x.remote.packet, &y.local.packet)
+            equal_packet_keys(&*x.local.packet, &*y.remote.packet)
+                && equal_packet_keys(&*x.remote.packet, &*y.local.packet)
```
cpu, line 3349: I think this might be a style thing, but personally I think `as_ref()` reads better than `&*`:
```rust
            equal_packet_keys(x.local.packet.as_ref(), y.remote.packet.as_ref())
                && equal_packet_keys(x.remote.packet.as_ref(), y.local.packet.as_ref())
```
WDYT?

I think there are some other `&*` instances in this general area that don't appear to be required, e.g. L3328 & L3332 can just be `header` instead of `&*header`. Am I missing something deeper here?

rustls/tests/api.rs
```diff
@@ -3432,8 +3432,8 @@ mod test_quic {
             let client_early = client.zero_rtt_keys().unwrap();
             let server_early = server.zero_rtt_keys().unwrap();
             assert!(equal_packet_keys(
-                &client_early.packet,
-                &server_early.packet
+                &*client_early.packet,
```
cpu, line 3435: ditto RE: `as_ref` here and 3519/3520. I won't leave the same feedback in each place but it's something to consider.

rustls/src/crypto/ring/tls13.rs
```diff
@@ -101,18 +101,21 @@ impl Tls13AeadAlgorithm for AeadAlgorithm {
     }
 }
 
+#[used]
```
cpu, line 104: Can you leave a `TODO` or some kind of comment here linking to the rustc bug?

It will be hard to remember why these attributes are here in the future otherwise. One day our MSRV will have the fix and we can rip these out again :crossed_fingers:

cpu (approved):
Just some small nits, I'm comfortable +1ing now. Thanks!

rustls/src/tls13/mod.rs
```diff
@@ -1,21 +1,27 @@
...
-use crate::enums::{CipherSuite, ProtocolVersion};
+use crate::enums::{CipherSuite, ContentType, ProtocolVersion};
 use crate::error::{Error, PeerMisbehaved};
 use crate::msgs::base::Payload;
 use crate::msgs::codec::Codec;
 use crate::msgs::fragmenter::MAX_FRAGMENT_LEN;
 use crate::msgs::message::{BorrowedPlainMessage, OpaqueMessage, PlainMessage};
-use crate::suites::{BulkAlgorithm, CipherSuiteCommon, SupportedCipherSuite};
+use crate::suites::BulkAlgorithm;
+use crate::suites::{CipherSuiteCommon, SupportedCipherSuite};
```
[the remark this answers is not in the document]
cpu: This fell away in later commits.

rustls/src/crypto/ring/tls13.rs
```diff
@@ -0,0 +1,204 @@
+use crate::crypto::cipher::{
+    make_tls13_aad, AeadKey, Iv, MessageDecrypter, MessageEncrypter, Nonce,
+};
+use crate::enums::{CipherSuite, ContentType, ProtocolVersion};
+use crate::error::Error;
+use crate::msgs::base::Payload;
+use crate::msgs::codec::Codec;
+use crate::msgs::message::{BorrowedPlainMessage, OpaqueMessage, PlainMessage};
+use crate::suites::BulkAlgorithm;
```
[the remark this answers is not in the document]
cpu: Fell away. Sorry, didn't catch this until I submitted my review.

rustls/src/quic.rs
cpu: Fell away.

rustls/src/tls13/mod.rs
```diff
@@ -142,30 +143,57 @@ fn make_tls13_aad(len: usize) -> ring::aead::Aad<[u8; TLS13_AAD_SIZE]> {
 // https://datatracker.ietf.org/doc/html/rfc8446#section-5.2
 const TLS13_AAD_SIZE: usize = 1 + 2 + 2;
 
-struct AeadAlgorithm(&'static ring::aead::Algorithm);
+struct AeadAlgorithm {
```
djc: (More ordering...) The `struct` definition here should go belong the `static`s that use it.

rustls/src/tls13/mod.rs
```diff
@@ -233,6 +261,41 @@ impl MessageDecrypter for Tls13MessageDecrypter {
     }
 }
 
+#[cfg(feature = "secret_extraction")]
+fn chacha20_poly1305_extract_keys(key: AeadKey, iv: Iv) -> ConnectionTrafficSecrets {
```
djc: I forget whether this is the same argument I made in my previous round of feedback, but it still feels like this means `AeadAlgorithm` should be more like a trait with an `extract_keys()` method? Why doesn't that work here?

rustls/src/msgs/message.rs
```diff
@@ -146,6 +147,30 @@ impl OpaqueMessage {
         }
     }
 
+    /// For TLS1.3 (only), checks the length msg.payload is valid and removes the padding.
+    ///
+    /// Returns an error if the message (pre-unpadding) is too long, or the padding is invalid,
+    /// or the message (post-unpadding) is too long.
+    pub fn into_plain_message_with_tls13_unpadding(mut self) -> Result<PlainMessage, Error> {
```
djc, line 154: This name seems very verbose... Maybe it can be tightened up a bit?

rustls/src/tls13/mod.rs
```diff
@@ -120,29 +121,6 @@ impl fmt::Debug for Tls13CipherSuite {
     }
 }
 
-fn unpad_tls13(v: &mut Vec<u8>) -> ContentType {
```
djc, line 124: In the future, I would isolate moving code around into a separate commit.

rustls/src/crypto/ring/suites.rs
```diff
@@ -0,0 +1,37 @@
...
+    TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+    TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+    TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+};
+use super::tls13::{
+    TLS13_AES_128_GCM_SHA256, TLS13_AES_256_GCM_SHA384, TLS13_CHACHA20_POLY1305_SHA256,
+};
+use crate::suites::SupportedCipherSuite;
+
+/// A list of all the cipher suites supported by rustls.
```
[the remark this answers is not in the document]
djc: Since `DEFAULT_CIPHER_SUITES` is based on `ALL_CIPHER_SUITES`, order `DEFAULT` before `ALL`?

Personally, I find this module on the small side and I would just keep these in `ring/mod.rs`. (They're also highly relevant so it makes sense to me for them to live there.)

rustls/src/crypto/ring/mod.rs
```diff
@@ -14,6 +14,10 @@ use core::fmt;
 
 pub(crate) mod hash;
 pub(crate) mod hmac;
+pub(crate) mod suites;
```
djc, line 19: In the future, recommend splitting up large commits like these into smaller commits that move one colocated block of code somewhere else.

rustls/src/crypto/cipher.rs
```diff
@@ -2,16 +2,83 @@ use crate::enums::{ContentType, ProtocolVersion};
 use crate::error::Error;
 use crate::msgs::codec;
 pub use crate::msgs::message::{BorrowedPlainMessage, OpaqueMessage, PlainMessage};
+#[cfg(feature = "secret_extraction")]
+use crate::suites::ConnectionTrafficSecrets;
+
+/// Factory trait for building `MessageEncrypter` and `MessageDecrypter` for a TLS1.3 cipher suite.
```
djc, line 8: In the future, consider not moving code and changing it in the same code.

rustls/src/quic.rs
```diff
@@ -2,17 +2,18 @@
...
 use crate::crypto::CryptoProvider;
 use crate::enums::{AlertDescription, ProtocolVersion};
 use crate::error::Error;
+use crate::hkdf;
 use crate::msgs::handshake::{ClientExtension, ServerExtension};
 use crate::server::{ServerConfig, ServerConnectionData};
-use crate::suites::BulkAlgorithm;
-use crate::tls13::key_schedule::hkdf_expand;
-use crate::tls13::{Tls13CipherSuite, TLS13_AES_128_GCM_SHA256_INTERNAL};
+use crate::tls13::key_schedule::hkdf_expand_label_block;
```
djc, line 11: In the future, also recommend splitting this up into smaller commits that do one thing.

djc (approved):

rustls/src/crypto/hmac.rs
```diff
@@ -20,7 +20,7 @@ pub struct Tag {
 impl Tag {
     /// Build a tag by copying a byte slice.
     ///
-    /// The slice can be up to `Tag::MAX_LEN` bytes in length.
+    /// The slice can be up to [`Tag::MAX_LEN`] bytes in length.
```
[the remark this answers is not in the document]
ctz: Likely, yes, but that commit is already on main

rustls/src/tls13/mod.rs
```diff
@@ -33,12 +38,11 @@ pub(crate) static TLS13_CHACHA20_POLY1305_SHA256_INTERNAL: &Tls13CipherSuite = &
         hash_provider: &crypto::ring::hash::SHA256,
     },
     hmac_provider: &crypto::ring::hmac::HMAC_SHA256,
-    aead_alg: &AeadAlgorithm(&ring::aead::CHACHA20_POLY1305),
+    aead_alg: &CHACHA20_POLY1305_AEAD,
```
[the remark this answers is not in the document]
ctz: It became necessary only in this commit, because this is where I broke the compiler by adding one parameter to `AeadAlgorithm`, ie `&AeadAlgorithm(&ring::aead::CHACHA20_POLY1305, chacha20_poly1305_extract_keys)`

rustls/src/crypto/hmac.rs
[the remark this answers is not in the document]
djc: Ahh, sorry, got confused about `hash` vs `hmac`.

rustls/src/tls13/mod.rs
ctz: eventually we need 3 values that `impl Tls13AeadAlgorithm`, where `encrypter`, `decrypter` and `key_len` are the same code (based on a `ring::aead::Algorithm`) but `extract_keys` is different (because it produces different `ConnectionTrafficSecrets` variants). have replaced these free functions to different structs that each impl a new `KeyExtractor` trait that just does the `extract_keys()` part. that doesn't really make a lot of sense, though, because it implies any implementer of that trait is interchangeable with any other.

i think on balance i prefer how it was on the first round of this PR, where there were three structs that `impl Tls13AeadAlgorithm` and the shared parts were extracted out into a struct that owned the `ring::aead::Algorithm`.

djc: Okay, feel free to merge it like this or to revert it to your preferred approach.

rustls/src/crypto/ring/tls13.rs
ctz: have managed to drop this commit and workaround entirely

rustls/src/crypto/ring/tls13.rs
ctz: these names are intended to echo the standard name (in this case) `TLS_CHACHA20_POLY1305_SHA256`. i think we could make a follow-up to this that makes them (and the other constants around here) actually match the standard IANA names?

cpu: > i think we could make a follow-up to this that makes them (and the other constants around here) actually match the standard IANA names?

That sounds good to me :+1:

rustls/src/crypto/ring/suites.rs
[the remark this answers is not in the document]
ctz: fixed the comment and hoisted this into `ring/mod.rs`

rustls/tests/api.rs
ctz: have gone back to `.as_ref()` here and elsewhere

Issue #1361, opened 2023-07-13
doc: examples of how to use ClientCertVerifier impls with CRLs


jsha:
Over in https://github.com/rustls/rustls-ffi/pull/324#issuecomment-1634285194 we discussed how using CRLs with ClientCertVerifier comes with certain implicit requirements that we should make explicit in the docs. CRLs should be updated on some interval, which is likely to be shorter than the restart interval of the TLS server. Since with_crls is called on [AllowAnyAnonymousOrAuthenticatedClient](https://docs.rs/rustls/0.21.5/rustls/server/struct.AllowAnyAnonymousOrAuthenticatedClient.html) (and friend) and construction time, it means the old-school way of building a ServerConfig once per server doesn't work.

Instead, to be able to update CRLs periodically, you need to use the [Acceptor](https://docs.rs/rustls/0.21.5/rustls/server/struct.Acceptor.html#) API, and construct a new ClientCertVerifier on each handshake, using whichever set of CRLs has been fetched most recently.

There's also a subtlety that should be called out in the docs: because the client certificate is not available in the ClientHello, the server needs to know ahead of time what CRLs it cares about. For client revocation specifically, that's probably okay, because client authentication issuer hierarchies are (I think!) usually simple, and the server can usually be configured ahead of time to know what CRLs to care about. However, it might be a surprise to some users who are expecting to be able to fetch a CRL from the CRL Distribution Points extension of a client certificate at handshake time.

cpu:
I don't think there's an existing example showing off the `Acceptor` API so it seems doubly useful to have in-tree with that in mind.

PR #1402, opened 2023-08-16
examples: add server_acceptor example.

cpu:
## Description

This branch resolves https://github.com/rustls/rustls/issues/1361 by adding an example of using the `Acceptor` API to provide up-to-date CRL information per-connection.

### server: fix acceptor rustdoc typo.

Small typo fix.

### examples: add server_acceptor example.

This example has two main purposes:

1. It shows how to use the `Acceptor` API to customize a `ServerConfig` per-connection, possibly using information from the received `ClientHello`.
2. It shows how to load CRL information per-connection to ensure the freshest CRL content is used when validating client certificate revocation status.

Additionally this example uses `rcgen` to generate its own test PKI (and CRLs), potentially being a helpful reference for folks that want to do similar without needing to manually construct certs with `openssl`.

To simulate CRL updates this example program spawns a background thread that periodically replaces the CRL content, flipping back and forth between a CRL that lists the client certificate as revoked, and a CRL that has no revoked certificates.

Using `tlsclient-mio` (or another TLS client program) with the generated client certificate/key you can observe the CRL updates happening by connecting to the server, waiting a little bit, and then connecting again. The result will differ based on the CRL update:

```
$ cargo run --bin tlsclient-mio -- \
  --auth-certs ./client-cert.pem \
  --auth-key ./client-key.pem \
  --cafile ca-cert.pem \
  --port 4443 \
  --http \
  localhost
TLS error: AlertReceived(CertificateRevoked)
Connection closed

<waiting>

$ cargo run --bin tlsclient-mio -- \
   --auth-certs ./client-cert.pem \
   --auth-key ./client-key.pem \
   --cafile ca-cert.pem \
   --port 4443 \
   --http \
   localhost
EOF
Connection closed
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,7 @@ members = [
 ]
 exclude = ["admin/rustfmt"]
 resolver = "2"
+
+[patch.crates-io]
+# TODO(XXX): Remove this once rcgen has cut a release w/ CRL support included. Only used in examples.
+rcgen = { git = 'https://github.com/est31/rcgen.git' }
--- a/examples/src/bin/server_acceptor.rs
+++ b/examples/src/bin/server_acceptor.rs
@@ -1,0 +1,356 @@
+//! A TLS server that accepts connections using a custom `Acceptor`, demonstrating how fresh
+//! CRL information can be retrieved per-client connection to use for revocation checking of
+//! client certificates.
+//!
+//! For a more complete server demonstration, see `tlsserver-mio.rs`.
+use std::fs::File;
+use std::io::{Read, Write};
+use std::ops::Add;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+use std::{fs, thread};
+
+use docopt::Docopt;
+use serde_derive::Deserialize;
+
+use rustls::crypto::CryptoProvider;
+use rustls::server::{
+    Acceptor, ClientHello, ServerConfig, UnparsedCertRevocationList, WebPkiClientVerifier,
+};
+use rustls::{Certificate, PrivateKey, RootCertStore};
+
+fn main() {
+    let version = concat!(
+        env!("CARGO_PKG_NAME"),
+        ", version: ",
+        env!("CARGO_PKG_VERSION")
+    )
+    .to_string();
+
+    let args: Args = Docopt::new(USAGE)
+        .map(|d| d.help(true))
+        .map(|d| d.version(Some(version)))
+        .and_then(|d| d.deserialize())
+        .unwrap_or_else(|e| e.exit());
+
+    if args.flag_verbose {
+        env_logger::Builder::new()
+            .parse_filters("trace")
+            .init();
+    }
+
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -7,3 +7,7 @@ members = [
 ]
 exclude = ["admin/rustfmt"]
 resolver = "2"
+
+[patch.crates-io]
+# TODO(XXX): Remove this once rcgen has cut a release w/ CRL support included. Only used in examples.
+rcgen = { git = 'https://github.com/est31/rcgen.git' }
```
cpu, line 13: It might be worth pinging the upstream to cut a release to avoid this.

I've been holding off on that only because I'm hoping to add support for the cert CRL distribution point and CRL issuing distribution point extensions first. 

I might be able to find time to work on that ~today, but perhaps it shouldn't block a release either way?

ctz (approved):

Cargo.toml
djc: > I might be able to find time to work on that ~today, but perhaps it shouldn't block a release either way?

Depends on whether those changes would need to be semver-breaking.

cpu: I think it can be done w/o breaking semver for the cert CRL distrib point because the `CertificateParams` struct is non-exhaustive.

For the CRL issuing distribution point I think it will need to be breaking, I didn't mark the `CertificateRevocationListParams` struct as non-exhaustive (in retrospect, an oversight).

examples/src/bin/server_acceptor.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+}
+
+/// CRL updater function that runs in a separate thread. This periodically updates the CRL file
+/// on disk, flipping between writing a CRL that describes the client certificate as revoked, and
+/// a CRL that describes the client certificate as not revoked.
+///
+/// In a real use case, the CRL would be updated by fetching fresh CRL data from an authoritative
+/// distribution point.
+fn crl_updater(duration: Duration, crl_path: &Path, pki: Arc<TestPki>) {
```
djc, line 275: Style suggestion: for these kinds of little thread thingies, I like to write a `struct` definition with the input state it needs, with a `run(self)` method on it. I feel like that more clearly signals intent that this is not just a normal function.

djc (approved):
Nice!

examples/src/bin/server_acceptor.rs
```diff
@@ -0,0 +1,336 @@
...
+use serde_derive::Deserialize;
+
+use rustls::crypto::CryptoProvider;
+use rustls::server::{
+    Acceptor, ClientHello, ServerConfig, UnparsedCertRevocationList, WebPkiClientVerifier,
+};
+use rustls::{Certificate, PrivateKey, RootCertStore};
+
+fn main() {
+    let version = env!("CARGO_PKG_NAME").to_string() + ", version: " + env!("CARGO_PKG_VERSION");
```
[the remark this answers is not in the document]
cpu: Fixed.

Issue #1385, opened 2023-08-02
Performance measurements for day-to-day development of Rustls


aochagavia:
_See https://github.com/rustls/rustls/issues/1205#issuecomment-1661883148 for context_

It would be very useful to have _automated_ and _accurate_ feedback on a PR's performance impact compared to the main branch. It should be automated, to ensure it is always used, and it should be accurate, to ensure it is actionable (i.e. too much noise would train reviewers to ignore the information). The [approach followed by rustc](https://github.com/rust-lang/rust/pull/112849#issuecomment-1661062264) is a good example, though it cost them a daunting amount of work.

We might want to keep an up-to-date plot of performance over time, too, for two reasons:

1. As a coherence check (i.e. if the results are truly accurate, we should see little noise over time, and be able to explain any significant performance differences).
2. To reveal long-term performance trends after a significant series of changes (e.g. after someone spent 6 months on an effort to optimize the library, after a series of refactorings to address security audit findings, etc).

aochagavia:
After giving it some more thought, [gathering feedback](https://old.reddit.com/r/rust/comments/15h6rsv/crate_authors_how_do_you_prevent_performance/), reading and experimenting, I think we probably want to go with the following approach in our initial setup:

1. Measure CPU instructions using cachegrind. It works with the default GitHub actions runners and the benchmarking code requires minimum maintenance.
2. Create additional time-based benchmarks using [paired benchmarking](https://www.bazhenov.me/posts/paired-benchmarking/) for cases in which instruction counts alone do not suffice (e.g. when introducing more parallelism, like mentioned in [this article](https://tratt.net/laurie/blog/2022/what_metric_to_use_when_benchmarking.html)). These benchmarks are not part of the rustls repository and the CI, because they have a higher maintenance cost, are more noisy, and are only occasionally necessary (i.e. for most PRs the instruction check should be enough).

# Additional context

### Measuring CPU instructions

This technique has been successfully used in tracking the Rust compiler's performance, and is generally [recommended](https://blog.mozilla.org/nnethercote/2020/09/08/how-to-speed-up-the-rust-compiler-one-last-time/) when comparing two versions of the same code to identify performance differences. It has incredibly low noise, and therefore makes for a very good metric for automatic PR checking (i.e. the automatic check will reliably identify significant performance changes).

[...]

PR #1404, opened 2023-08-17
Implement CPU instruction benchmarks

aochagavia:
This PR is a first step towards closing #1385. It implements a basic benchmarking setup suitable for CI, but doesn't hook it up yet to GitHub actions. The code is contained in its own crate (`ci-bench`).

Relevant features (check out the [readme](https://github.com/aochagavia/rustls/blob/ci-bench/ci-bench/README.md) for details):

1. Count CPU instructions for ~20 scenarios.
2. Separately measure server and client side CPU instructions.
3. Compare results between a baseline and a candidate, and generate a user-friendly overview in markdown (see the first comment to this PR as an example).

Non-features:

1. Aggregate statistics to better identify where a change in CPU counts comes from ([suggested](https://github.com/rustls/rustls/issues/1385#issuecomment-1678684307) by @djc). I'd rather get this out of the door in its current form, so we have something to iterate upon, and tweak it later if necessary.
2. GitHub Actions integration. I'd rather postpone this to focus on the benchmarking setup itself.
4. Benchmarks for memory usage. I still need to think about a reasonable setup for this one.

### Open question: significance threshold

Which instruction count diffs should be marked as significant (i.e. requiring attention of a reviewer)? I'm leaning towards a threshold of 0.2% because __it is the lowest change that can be confirmed or disproved by time-based benchmarks__.

If the instruction counts for a PR are suspect, we will want to confirm there is truly a performance regression through time-based benchmarks (remember that the change in executed CPU instructions is not necessarily in the same proportion as the change in wall time). Assuming a worst-case scenario where the wall time is affected by 5x the change in CPU instructions, a 0.2% change in instructions would amount to a 1% change in runtime. Time-based benchmarks take around 800 µs on my machine, and a difference of 1% (~8 µs) is just high enough to be detected through paired benchmarking.

There is a chance that the threshold is too low, and might result in noise (i.e. changes in instruction counts that cannot be reliably confirmed by time-based benchmarks). Still, as @djc [suggested](https://github.com/rustls/rustls/issues/1385#issuecomment-1671549976), it makes sense to start with a low threshold in order to gain experience. We can always tweak it later if it gets too annoying.

### Sidenote on the benchmarking code

The benchmarks themselves were adapted from [here](https://github.com/rustls/rustls/blob/6bdaf04e7af97821485da115b5436913fedea874/rustls/examples/internal/bench.rs). It would be nice if you can double check they are correct, because I'm not familiar enough with rustls to do so with confidence.

### DIFF
--- a/ci-bench/Cargo.toml
+++ b/ci-bench/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "rustls-ci-bench"
+version = "0.0.1"
+edition = "2021"
+license = "Apache-2.0 OR ISC OR MIT"
+description = "Rustls CPU instruction counting benchmarks."
+publish = false
+
+[dependencies]
+anyhow = "1.0.73"
+byteorder = "1.4.3"
+clap = { version = "4.3.21", features = ["derive"] }
+itertools = "0.11.0"
+rayon = "1.7.0"
+rustls = { path = "../rustls" }
+rustls-pemfile = "1.0.3"
+
--- a/ci-bench/README.md
+++ b/ci-bench/README.md
@@ -1,0 +1,110 @@
+# CI Bench
+
+This crate is meant for CI benchmarking. It measures CPU instructions using `cachegrind`, outputs
+the results in CSV format and allows comparing results from multiple runs.
+
+## Usage
+
+You can get detailed usage information through `cargo run --release -- --help`. Below are the most
+important bits.
+
+### Running all benchmarks
+
+_Note: this step requires having `valgrind` in your path._
+
+Use `cargo run --release -- run-all > out.csv` to generate a CSV with the instruction counts for
+the different scenarios we support. The result should look like the following:
+
+```csv
+handshake_no_resume_1.2_rsa_aes_server,11327015
+handshake_no_resume_1.2_rsa_aes_client,4314952
+handshake_session_id_1.2_rsa_aes_server,11342136
+handshake_session_id_1.2_rsa_aes_client,4327564
+handshake_tickets_1.2_rsa_aes_server,11347746
+handshake_tickets_1.2_rsa_aes_client,4331424
+transfer_no_resume_1.2_rsa_aes_server,8775780
+transfer_no_resume_1.2_rsa_aes_client,8818847
+handshake_no_resume_1.3_rsa_aes_server,11517007
+handshake_no_resume_1.3_rsa_aes_client,4212770
+...
+... rest omitted for brevity
+...
+```
+
+### Comparing results
+
+Use `cargo run --release -- compare out1.csv out2.csv`. It will output a report using
+GitHub-flavored markdown (used by the CI itself to give feedback about PRs). We currently
+consider differences of 0.2% to be significant, but might tweak it in the future after we gain
+experience with the benchmarking setup.
+
+### Supported scenarios
+
+We benchmark the following scenarios:
+
+- Handshake without resumption (`handshake_no_resume`)
+- Handshake with ticket resumption (`handshake_tickets`)
+- Handshake with session id resumption (`handshake_session_id`)
+- Encrypt, transfer and decrypt 1MB of data sent by the server
+  (`transfer_no_resume`)
+
+The scenarios are benchmarked with different TLS versions, certificate key types and cipher suites.
+CPU counts are measured independently for the server side and for the client side. Hence, we end up
+with names like `transfer_no_resume_1.3_rsa_aes_client`.
+
+## Internals
+
+We have made an effort to heavily document the source code of the benchmarks. In addition to that,
+here are some high-level considerations that can help you hack on the crate.
+
+### Architecture
+
+An important goal of this benchmarking setup is that it should run with minimum noise on
+standard GitHub Actions runners. We achieve that by measuring CPU instructions using `cachegrind`,
+which runs fine on the cloud (contrary to hardware instruction counters). This is the same
+approach used by the [iai](https://crates.io/crates/iai) benchmarking crate, but we needed more
+flexibility and have therefore rolled our own setup.
+
+Using `cachegrind` has some architectural consequences because it operates at the process level
+(i.e. it can count CPU instructions for a whole process, but not for a single function). The
+most important consequences are:
+
+- Since we want to measure server and client instruction counts separately, the benchmark runner
+  spawns two child processes for each benchmark (one for the client, one for the server) and pipes
+  their stdio to each other for communication (i.e. stdio acts as the transport layer).
+- There is a no-op "benchmark" that measures the overhead of starting up the child process, so
+  we can subtract it from the instruction count of the real benchmarks and reduce noise.
+- Since we want to measure individual portions of code (e.g. data transfer after the handshake),
+  there is a mechanism to subtract the instructions that are part of a benchmark's setup.
+  Specifically, a benchmark can be configured to have another benchmark's instruction count
+  subtracted from it. We are currently using this to subtract the handshake instructions from the
+  data transfer benchmark.
+
+### Debugging
+
+If you need to debug the crate, here are a few tricks that might help:
+
+- For printf debugging, you should use `eprintln!`, because child processes use stdio as the
+  transport for the TLS connection (i.e. if you print something to stdout, you won't even see it
+  _and_ the other side of the connection will choke on it).
+- When using a proper debugger, remember that each side of the connection runs as a child process.
+  If necessary, you can tweak the code to ensure both sides of the connection run on the parent
+  process (e.g. by starting each side on its own thread and having them communicate through TCP).
+  This should require little effort, because the TLS transport layer is encapsulated and generic
+  over `Read` and `Write`.
+
+### Why measure CPU instructions
+
+This technique has been successfully used in tracking the Rust compiler's performance, and is
+known to work well when comparing two versions of the same code. It has incredibly low noise,
+and therefore makes for a very good metric for automatic PR checking (i.e. the automatic check
+will reliably identify significant performance changes).
+
+It is not possible to deduce the exact change in runtime based on the instruction count
+difference (e.g. a 5% increase in instructions does not necessarily result in a 5% increase in
+runtime). However, if there is a significant change in instruction count, you can be fairly
+confident there is a significant change in runtime too. This is very useful information to have
+when reviewing a PR.
+
+For more information, including the alternatives we considered, check out [this comment]
+(https://github.com/rustls/rustls/issues/1385#issuecomment-1668023152) in the issue tracker.
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -1,0 +1,632 @@
+use std::collections::HashMap;
+use std::fs;
+use std::hint::black_box;
+use std::io::{self, BufRead, BufReader, Read, Write};
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use anyhow::Context;
+use clap::{Parser, Subcommand, ValueEnum};
+use itertools::Itertools;
+use rayon::iter::Either;
+use rayon::prelude::*;
+use rustls::client::Resumption;
+use rustls::crypto::ring::Ring;
+use rustls::server::{NoServerSessionStorage, ServerSessionMemoryCache, WebPkiClientVerifier};
+use rustls::{
+    ClientConfig, ClientConnection, ProtocolVersion, RootCertStore, ServerConfig, ServerConnection,
+    Ticketer,
+};
+
+use crate::benchmark::{
+    get_reported_instr_count, validate_benchmarks, Benchmark, BenchmarkKind, BenchmarkParams,
+    ResumptionKind,
+};
+use crate::cachegrind::CachegrindRunner;
+use crate::util::transport::{
+    read_handshake_message, read_plaintext_to_end_bounded, send_handshake_message,
+    write_all_plaintext_bounded,
+};
+use crate::util::KeyType;
+
+mod benchmark;
+mod cachegrind;
+mod util;
+
+/// The size in bytes of the plaintext sent in the transfer benchmark
+const TRANSFER_PLAINTEXT_SIZE: usize = 1024 * 1024;
+
+/// The amount of times a resumed handshake should be executed during benchmarking.
+///
+/// Handshakes with session resumption execute a very small amount of instructions (less than 200_000
+/// for some parameters), so a small difference in instructions accounts for a high difference in
+/// percentage (making the benchmark more sensitive to noise, because differences as low as 500
+/// instructions already raise a flag). Running the handshake multiple times gives additional weight
+/// to the instructions involved in the handshake, and less weight to noisy one-time setup code.
+///
+/// More specifically, great part of the noise in resumed handshakes comes from the usage of
+/// [`rustls::client::ClientSessionMemoryCache`] and [`rustls::server::ServerSessionMemoryCache`],
+/// which rely on a randomized `HashMap` under the hood (you can check for yourself by that
+/// `HashMap` by a `FxHashMap`, which brings the noise down to acceptable levels in a single run).
+const RESUMED_HANDSHAKE_RUNS: usize = 3;
+
+/// The threshold at which instruction count changes are considered relevant
+const CHANGE_THRESHOLD: f64 = 0.002; // 0.2%
+
+#[derive(Parser)]
+#[command(about)]
+pub struct Cli {
+    #[command(subcommand)]
+    pub command: Command,
+}
+
+#[derive(Subcommand)]
+pub enum Command {
+    /// Run all benchmarks and prints the measured CPU instruction counts in CSV format
+    RunAll,
+    /// Run a single benchmark at the provided index (used by the bench runner to start each benchmark in its own process)
+    RunSingle { index: u32, side: Side },
+    /// Compare the results from two previous benchmark runs and print a user-friendly markdown overview
+    Compare {
+        /// Path to a CSV file obtained from a previous `run-all` execution
+        baseline_input: PathBuf,
+        /// Path to a CSV file obtained from a previous `run-all` execution
+        candidate_input: PathBuf,
+    },
+}
+
+#[derive(Copy, Clone, ValueEnum)]
+pub enum Side {
+    Server,
+    Client,
+}
+
+impl Side {
+    /// Returns the string representation of the side
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Side::Client => "client",
+            Side::Server => "server",
+        }
+    }
+}
+
+fn main() -> anyhow::Result<()> {
+    let benchmarks = all_benchmarks()?;
+
+    let cli = Cli::parse();
+    match cli.command {
+        Command::RunAll => {
+            let executable = std::env::args().next().unwrap();
+            let results = run_all(executable, &benchmarks)?;
+
+            // Output results in CSV (note: not using a library here to avoid extra dependencies)
+            for (name, instr_count) in results {
+                println!("{name},{instr_count}");
+            }
+        }
+        Command::RunSingle { index, side } => {
+            // `u32::MAX` is used as a signal to do nothing and return. By "running" an empty
+            // benchmark we can measure the startup overhead.
+            if index == u32::MAX {
+                return Ok(());
+            }
+
+            let bench = benchmarks
+                .get(index as usize)
+                .ok_or(anyhow::anyhow!("Benchmark not found: {index}"))?;
+
+            let mut stdin = io::stdin().lock();
+            let mut stdout = io::stdout().lock();
+
+            let handshake_buf = &mut [0u8; 262144];
+            let resumption_kind = black_box(bench.kind.resumption_kind());
+            let params = black_box(bench.params);
+            let io = StepperIO {
+                reader: &mut stdin,
+                writer: &mut stdout,
+                handshake_buf,
+            };
+            let result = match side {
+                Side::Server => run_bench(
+                    ServerSideStepper {
+                        io,
+                        config: ServerSideStepper::make_config(&params, resumption_kind),
+                    },
+                    bench.kind,
+                ),
+                Side::Client => run_bench(
+                    ClientSideStepper {
+                        io,
+                        resumption_kind,
+                        config: ClientSideStepper::make_config(&params, resumption_kind),
+                    },
+                    bench.kind,
+                ),
+            };
+
+            result
+                .with_context(|| format!("{} crashed for {} side", bench.name(), side.as_str()))?;
+        }
+        Command::Compare {
+            baseline_input,
+            candidate_input,
+        } => {
+            let baseline = read_results(baseline_input.as_ref())?;
+            let candidate = read_results(candidate_input.as_ref())?;
+            let result = compare_results(&baseline, &candidate);
+            print_report(result);
+        }
+    }
+
+    Ok(())
+}
+
+/// Returns all benchmarks
+fn all_benchmarks() -> anyhow::Result<Vec<Benchmark>> {
+    let mut benchmarks = Vec::new();
+    for &param in ALL_BENCHMARK_PARAMS {
+        add_benchmark_group(&mut benchmarks, param);
+    }
+
+    validate_benchmarks(&benchmarks)?;
+    Ok(benchmarks)
+}
+
+/// The benchmark params to use for each group of benchmarks
+static ALL_BENCHMARK_PARAMS: &[BenchmarkParams] = &[
+    BenchmarkParams::new(
+        KeyType::Rsa,
+        rustls::cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+        &rustls::version::TLS12,
+        "1.2_rsa_aes",
+    ),
+    BenchmarkParams::new(
+        KeyType::Rsa,
+        rustls::cipher_suite::TLS13_AES_128_GCM_SHA256,
+        &rustls::version::TLS13,
+        "1.3_rsa_aes",
+    ),
+    BenchmarkParams::new(
+        KeyType::Ecdsa,
+        rustls::cipher_suite::TLS13_AES_128_GCM_SHA256,
+        &rustls::version::TLS13,
+        "1.3_ecdsa_aes",
+    ),
+    BenchmarkParams::new(
+        KeyType::Rsa,
+        rustls::cipher_suite::TLS13_CHACHA20_POLY1305_SHA256,
+        &rustls::version::TLS13,
+        "1.3_rsa_chacha",
+    ),
+    BenchmarkParams::new(
+        KeyType::Ecdsa,
+        rustls::cipher_suite::TLS13_CHACHA20_POLY1305_SHA256,
+        &rustls::version::TLS13,
+        "1.3_ecdsa_chacha",
+    ),
+];
+
+/// Adds a group of benchmarks for the specified parameters
+///
+/// The benchmarks in the group are:
+///
+/// - Handshake without resumption
+/// - Handshake with session id resumption
+/// - Handshake with ticket resumption
+/// - Transfer a 1MB data stream from the server to the client
+fn add_benchmark_group(benchmarks: &mut Vec<Benchmark>, params: BenchmarkParams) {
+    let params_label = params.label;
+
+    // Create handshake benchmarks for all resumption kinds
+    for &resumption_param in ResumptionKind::ALL {
+        let handshake_bench = Benchmark::new(
+            format!("handshake_{}_{params_label}", resumption_param.label()),
+            BenchmarkKind::Handshake(resumption_param),
+            params,
+        );
+
+        let handshake_bench = if resumption_param != ResumptionKind::No {
+            // Since resumed handshakes include a first non-resumed handshake, we need to subtract
+            // the non-resumed handshake's instructions
+            handshake_bench
+                .exclude_setup_instructions(format!("handshake_no_resume_{params_label}"))
+        } else {
+            handshake_bench
+        };
+
+        benchmarks.push(handshake_bench);
+    }
+
+    // Benchmark data transfer
+    benchmarks.push(
+        Benchmark::new(
+            format!("transfer_no_resume_{params_label}"),
+            BenchmarkKind::Transfer,
+            params,
+        )
+        .exclude_setup_instructions(format!("handshake_no_resume_{params_label}")),
+    );
+}
+
+/// Run all the provided benches under cachegrind to retrieve their instruction count
+pub fn run_all(executable: String, benches: &[Benchmark]) -> anyhow::Result<Vec<(String, u64)>> {
+    // Run the benchmarks in parallel
+    let cachegrind = CachegrindRunner::new(executable)?;
+    let results: Vec<_> = benches
+        .par_iter()
+        .enumerate()
+        .map(|(i, bench)| (bench, cachegrind.run_bench(i as u32, bench)))
+        .collect();
+
+    // Report possible errors
+    let (errors, results): (Vec<_>, HashMap<_, _>) =
+        results
+            .into_iter()
+            .partition_map(|(bench, result)| match result {
+                Err(_) => Either::Left(()),
+                Ok(instr_counts) => Either::Right((bench.name(), instr_counts)),
+            });
+    if !errors.is_empty() {
+        // Note: there is no need to explicitly report the names of each crashed benchmark, because
+        // names and other details are automatically printed to stderr by the child process upon
+        // crashing
+        anyhow::bail!("One or more benchmarks crashed");
+    }
+
+    // Gather results keeping the original order of the benchmarks
+    let mut measurements = Vec::new();
+    for bench in benches {
+        let instr_counts = get_reported_instr_count(bench, &results);
+        measurements.push((bench.name_with_side(Side::Server), instr_counts.server));
+        measurements.push((bench.name_with_side(Side::Client), instr_counts.client));
+    }
+
+    Ok(measurements)
+}
+
+/// Drives the different steps in a benchmark.
+///
+/// See [`run_bench`] for specific details on how it is used.
+trait BenchStepper {
+    type Endpoint;
+
+    fn handshake(&mut self) -> anyhow::Result<Self::Endpoint>;
+    fn sync_before_resumed_handshake(&mut self) -> anyhow::Result<()>;
+    fn transmit_data(&mut self, endpoint: &mut Self::Endpoint) -> anyhow::Result<()>;
+}
+
+/// Stepper fields necessary for IO
+struct StepperIO<'a> {
+    reader: &'a mut dyn Read,
+    writer: &'a mut dyn Write,
+    handshake_buf: &'a mut [u8],
+}
+
+/// A benchmark stepper for the client-side of the connection
+struct ClientSideStepper<'a> {
+    io: StepperIO<'a>,
+    resumption_kind: ResumptionKind,
+    config: Arc<ClientConfig<Ring>>,
+}
+
+impl ClientSideStepper<'_> {
+    fn make_config(params: &BenchmarkParams, resume: ResumptionKind) -> Arc<ClientConfig<Ring>> {
+        assert_eq!(params.ciphersuite.version(), params.version);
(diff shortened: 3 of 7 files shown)


aochagavia:
Below you can see the generated markdown output after comparing two benchmark runs. The baseline numbers are from current main, the candidate numbers are from a second run, with fake changes to simulate a performance regression in data transfer for TLS 1.2

---

# Benchmark results
### Noteworthy instruction count differences
| Scenario | Baseline | Candidate | Diff 
| --- | --- | --- | --- |
| transfer_no_resume_1.2_rsa_aes_client | 7767624 | 7789541 | 21917 (0.28%) |
| transfer_no_resume_1.2_rsa_aes_server | 8775849 | 8798099 | 22250 (0.25%) |
### Other instruction count differences
<details>
<summary>Click to expand</summary>

| Scenario | Baseline | Candidate | Diff 
| --- | --- | --- | --- |
| handshake_no_resume_1.3_ecdsa_chacha_server | 1621211 | 1621487 | 276 (0.02%) |
| handshake_session_id_1.3_ecdsa_aes_server | 1646752 | 1646524 | -228 (-0.01%) |
| handshake_no_resume_1.3_ecdsa_aes_server | 1618360 | 1618562 | 202 (0.01%) |
| handshake_tickets_1.3_ecdsa_aes_server | 1644306 | 1644129 | -177 (-0.01%) |
| transfer_no_resume_1.3_rsa_aes_client | 7780252 | 7779666 | -586 (-0.01%) |
| handshake_tickets_1.3_ecdsa_chacha_server | 1648326 | 1648444 | 118 (0.01%) |
| handshake_tickets_1.2_rsa_aes_client | 4333191 | 4332957 | -234 (-0.01%) |
| handshake_session_id_1.3_ecdsa_chacha_server | 1650284 | 1650201 | -83 (-0.01%) |
| transfer_no_resume_1.3_ecdsa_aes_server | 8784596 | 8784197 | -399 (-0.00%) |
| transfer_no_resume_1.3_ecdsa_aes_client | 7781504 | 7781850 | 346 (0.00%) |
| transfer_no_resume_1.3_rsa_aes_server | 8783940 | 8784306 | 366 (0.00%) |
| handshake_session_id_1.2_rsa_aes_client | 4328918 | 4329096 | 178 (0.00%) |
| handshake_no_resume_1.2_rsa_aes_client | 4316722 | 4316886 | 164 (0.00%) |
| handshake_session_id_1.3_rsa_aes_server | 11545656 | 11546084 | 428 (0.00%) |
| handshake_no_resume_1.2_rsa_aes_server | 11328165 | 11328536 | 371 (0.00%) |
| handshake_tickets_1.3_rsa_chacha_client | 4227132 | 4227018 | -114 (-0.00%) |
| handshake_tickets_1.3_rsa_aes_server | 11545061 | 11545320 | 259 (0.00%) |
| handshake_tickets_1.3_rsa_aes_client | 4217162 | 4217254 | 92 (0.00%) |
| handshake_session_id_1.2_rsa_aes_server | 11343190 | 11342955 | -235 (-0.00%) |
| transfer_no_resume_1.3_ecdsa_chacha_client | 11245701 | 11245922 | 221 (0.00%) |
| handshake_session_id_1.3_rsa_aes_client | 4217144 | 4217225 | 81 (0.00%) |
| handshake_no_resume_1.3_rsa_chacha_server | 11527772 | 11527991 | 219 (0.00%) |
| transfer_no_resume_1.3_rsa_chacha_client | 11243310 | 11243488 | 178 (0.00%) |
| transfer_no_resume_1.3_ecdsa_chacha_server | 12218266 | 12218112 | -154 (-0.00%) |
| handshake_no_resume_1.3_ecdsa_aes_client | 14018509 | 14018358 | -151 (-0.00%) |
| handshake_session_id_1.3_rsa_chacha_server | 11556973 | 11556867 | -106 (-0.00%) |
| handshake_tickets_1.3_rsa_chacha_server | 11556880 | 11556979 | 99 (0.00%) |
| handshake_tickets_1.2_rsa_aes_server | 11348871 | 11348786 | -85 (-0.00%) |
| handshake_no_resume_1.3_rsa_aes_client | 4213207 | 4213238 | 31 (0.00%) |
| handshake_no_resume_1.3_ecdsa_chacha_client | 14021170 | 14021069 | -101 (-0.00%) |
| handshake_no_resume_1.3_rsa_aes_server | 11517858 | 11517928 | 70 (0.00%) |
| handshake_tickets_1.3_ecdsa_chacha_client | 14022476 | 14022391 | -85 (-0.00%) |
| handshake_tickets_1.3_ecdsa_aes_client | 14019868 | 14019948 | 80 (0.00%) |
| transfer_no_resume_1.3_rsa_chacha_server | 12218085 | 12218115 | 30 (0.00%) |
| handshake_session_id_1.3_ecdsa_chacha_client | 14022384 | 14022416 | 32 (0.00%) |
| handshake_session_id_1.3_rsa_chacha_client | 4227044 | 4227040 | -4 (-0.00%) |
| handshake_session_id_1.3_ecdsa_aes_client | 14019792 | 14019783 | -9 (-0.00%) |
| handshake_no_resume_1.3_rsa_chacha_client | 4223239 | 4223239 | 0 (0.00%) |
</details>

ci-bench/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "rustls-ci-bench"
+version = "0.0.1"
+edition = "2021"
+rust-version = "1.60"
+license = "Apache-2.0 OR ISC OR MIT"
+description = "Rustls CPU instruction counting benchmarks."
+publish = false
+
+# Prevent this from interfering with workspaces
```
djc, line 10: Why don't we want this in the workspace?

ci-bench/src/util/mod.rs
```diff
@@ -0,0 +1,12 @@
+use std::io::Read;
+
+pub mod config;
+pub mod key;
+
+pub fn wait_till_stdin_closed(reader: &mut dyn Read, buf: &mut [u8]) {
```
djc, line 6: Given that all its callers are in `main.rs`, I would just move this there.

ci-bench/src/util/config.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    cfg
+}
+
+pub fn new_client(
+    params: &BenchmarkParam,
+    clientauth: ClientAuth,
+    resume: ResumptionParam,
+) -> ClientConnection {
+    let client_config = Arc::new(make_client_config(params, clientauth, resume));
```
djc, line 146: I would prefer inlining `make_client_config()` and `make_server_config()` into these functions.

ci-bench/src/rustls_facade.rs
```diff
@@ -0,0 +1,60 @@
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+use rustls::{ConnectionCommon, SideData};
+use std::io::{Read, Write};
+
+/// Sends data to the other side of the connection through the provided writer
+///
+/// Prepends a header consisting of the length of the TLS data (encoded as a big-endian u32), so the
+/// other side can know how many bytes to expect.
+pub fn send<T: SideData>(
```
djc, line 9: Can we reuse library elements like `complete_io()` instead?

ci-bench/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+
+fn transfer_client_side(
+    client: &mut ClientConnection,
+    reader: &mut dyn Read,
+    writer: &mut dyn Write,
+    plaintext_size: u64,
+) -> anyhow::Result<()> {
+    handshake_client_side(client, reader, writer, false)?;
+
+    let mut recv_buf = Vec::new();
```
djc, line 241: Clippy should flag this, should be `vec![0, plaintext_size as usize * 2]`.

ci-bench/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    let param = Arc::new(param);
+
+    for resumption_param in all_resumption_params {
+        let captured_param1 = param.clone();
+        let captured_param2 = param.clone();
+        benchmarks.push(
+            Benchmark::new(
+                format!("new_conn_{}_{tag}", resumption_param.label()),
+                move |_reader, _writer| {
```
djc, line 118: The `SideFn` stuff seems pretty hard to understand to me. Could we unroll this into an actual trait with two implementing types?

ci-bench/src/bench_lib/compare.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+/// Renders the diffs as a markdown table
+fn table(diffs: &[Diff]) {
+    println!("| Scenario | Baseline | Candidate | Diff ");
+    println!("| --- | --- | --- | --- |");
+    for diff in diffs {
+        println!(
+            "| {} | {} | {} | {} ({:.2}%) |",
```
djc, line 173: Can we do right-aligned numbers in Markdown? What about table cell background colors (green for significantly faster, red for significantly slower)?

ci-bench/src/bench_lib/runner.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::cli::Side;
+use anyhow::Context;
+use itertools::Itertools;
+use rayon::iter::Either;
+use rayon::prelude::*;
+use std::collections::HashMap;
+use std::io;
+
+/// Run a single bench at the specified side
+pub fn run_single(
```
djc, line 13: It seems strange for this to take a `&BenchmarkCollection` + `bench_index` instead of just a single `&Benchmark`.

ci-bench/src/bench_lib/benchmark.rs
```diff
@@ -0,0 +1,158 @@
+use crate::cli::Side;
+use itertools::Itertools;
+use std::collections::HashSet;
+use std::io::{Read, Write};
+
+/// A collection of benchmark scenarios
+pub struct BenchmarkCollection {
```
djc, line 7: What is the value of this type? It seems we could just pass a `Vec<Benchmark>` around which would be more straightforward.

ci-bench/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+    let mut benchmarks = Vec::new();
+    for param in all_benchmark_params() {
+        add_benchmark_group(&mut benchmarks, param);
+    }
+
+    BenchmarkCollection::new(benchmarks)
+}
+
+/// Returns the benchmark params to use for each group of benchmarks
+fn all_benchmark_params() -> Vec<BenchmarkParam> {
```
djc, line 57: Looks like this could be a `const` slice `&[BenchmarkParam]` instead, and we could ditch the trivial `BenchmarkParam::new()` method?

ci-bench/src/bench_lib/benchmark.rs
```diff
@@ -0,0 +1,158 @@
...
+        self.benchmarks.get(index)
+    }
+
+    /// Iterate through benchmarks in the order they were defined
+    pub fn as_slice(&self) -> &[Benchmark] {
+        &self.benchmarks
+    }
+
+    /// Returns an error if the benchmarks are invalid (see [`Self::new`])
+    fn validate_benchmarks(benchmarks: &[Benchmark]) -> anyhow::Result<()> {
```
djc, line 35: As a single-caller function, I would just inline this into the caller.

djc (commented):
There seem to be a lot of extra abstractions that don't add much value. I would suggest simplifying this a bunch, see the specific suggestions.

ci-bench/src/bench_lib/benchmark.rs
aochagavia: The type is meant to encode the invariant that the collection of benchmarks has been successfully validated. I was hoping it would make the code easier to read, but I can remove it if people have a clear preference for a plain `Vec`.

ci-bench/Cargo.toml
aochagavia: Hmm I actually copied this from [connect-tests](https://github.com/rustls/rustls/blob/6bdaf04e7af97821485da115b5436913fedea874/connect-tests/Cargo.toml). I haven't worked much with workspaces in the past, and I assumed this was necessary for "infrastructure" code. Is it superfluous?

ci-bench/src/main.rs
aochagavia: Great idea! `const` doesn't work here, but `static ALL_BENCHMARK_PARAMS: &'static [BenchmarkParam]` does

ci-bench/src/main.rs
aochagavia: I'm not sure how a trait would fit here, could you expand on the aspects that are currently hard to understand?

For context, each side of a benchmark (client and server) runs in its own subprocess. To make this possible, each benchmark must specify how to run each side. A `SideFn` is a function that runs one side of the benchmark: it needs access to a reader and a writer, and it returns a result.

ci-bench/src/bench_lib/compare.rs
aochagavia: Aligning is possible, but colors aren't. The Rust CI uses emoji to work around this. We could do something like follows. What do you think?

### Regression

| Scenario | Baseline | Candidate | Diff 
| --- | ---: | ---: | ---: |
| transfer_no_resume_1.2_rsa_aes_client | 7767624 | 7789541 | ⚠️ 21917 (0.28%) |
| transfer_no_resume_1.2_rsa_aes_server | 8775849 | 8798099 | ⚠️ 22250 (0.25%) |

### Improvement

| Scenario | Baseline | Candidate | Diff 
| --- | ---: | ---: | ---: |
| transfer_no_resume_1.2_rsa_aes_client | 7767624 | 7789541 | ✅ -21917 (-0.28%) |
| transfer_no_resume_1.2_rsa_aes_server | 8775849 | 8798099 | ✅ -22250 (-0.25%) |

djc: Emoji seems good!

ci-bench/Cargo.toml
djc: I would propose we ditch it and make this crate part of the workspace, so that it's included in CI by default.

ci-bench/src/bench_lib/benchmark.rs
djc: Currently I think we have only one place that create the `Vec<Benchmark>`, so for now it feels like overkill to add a wrapper for this -- instead we can just do the validation in that function.

ci-bench/Cargo.toml
aochagavia: All right!

ci-bench/src/bench_lib/benchmark.rs
aochagavia: Done 😉

ci-bench/Cargo.toml
cpu: The connect tests are a special case because they're hitting 3rd party servers over the internet and so can be prone to flakes outside of our control due to bad internet weather, cert expiry, etc.

aochagavia: Hmmm... One drawback of this is that the CI now enforces that the benchmarking code compiles with the minimum supported version. Or is [this](https://github.com/rustls/rustls/actions/runs/5891298366/job/15979135819?pr=1404) an unrelated failure? It also seems to skew the coverage measurements (and I assume has an impact on the length of some CI checks).

What is the benefit of having the new crate be part of the workspace? We are going to add a dedicated CI job for it anyway, so right now I can only see the drawbacks.

ci-bench/src/bench_lib/cachegrind.rs
```diff
@@ -0,0 +1,237 @@
...
+            .output()
+            .context("Failed to run `uname` to determine CPU architecture.")?;
+
+        let arch = String::from_utf8(output.stdout)
+            .context("`-uname -m` returned invalid unicode.")?
+            .trim()
+            .to_owned();
+
+        Ok(arch)
+    }
```
cpu, line 109: Could this be replaced with https://doc.rust-lang.org/std/env/consts/constant.ARCH.html ?

ci-bench/src/bench_lib/cachegrind.rs
```diff
@@ -0,0 +1,237 @@
...
+            .context("Failed to create cachegrind output directory")?;
+
+        // Run under setarch to disable ASLR, to reduce noise
+        let mut cmd = Command::new("setarch");
+        let child = cmd
+            .arg(arch)
+            .arg("-R")
+            .arg("valgrind")
+            .arg("--tool=cachegrind")
+            .arg("--cache-sim=no")
```
cpu, line 153: Maybe worth a comment similar to the `setarch` comment that mentions this disables collecting cache miss data? I had to read the fine manual to figure out what this arg does.

It also looks like this defaults to `no` (similar to `--branch-sim`). Is `cache-sim` important enough to be explicitly disabled, and if so, should `branch-sim` get the same treatment?

ci-bench/src/bench_lib/cachegrind.rs
```diff
@@ -0,0 +1,237 @@
...
+            .context("Failed to run benchmark in cachegrind")?;
+        if !status.success() {
+            anyhow::bail!(
+                "Failed to run benchmark in cachegrind. Exit code: {:?}",
+                status.code()
+            );
+        }
+
+        let instruction_count = parse_cachegrind_output(&self.output_file)?;
+        std::fs::remove_file(&self.output_file).ok();
```
cpu, line 196: would there be any value in an option to leave this output file in place for manual inspection?

ci-bench/src/bench_lib/compare.rs
```diff
@@ -0,0 +1,188 @@
...
+struct Diff {
+    scenario: String,
+    baseline: u64,
+    candidate: u64,
+    diff: i64,
+    diff_ratio: f64,
+}
+
+/// Compares all benchmark results between a baseline and a candidate, printing an overview to
+/// stdout
```
cpu, line 27: tiny nit: missing trailing `.` :-)
```suggestion
/// stdout.
```

ci-bench/src/bench_lib/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+mod cachegrind;
+mod compare;
+mod runner;
+
+pub use benchmark::{validate_benchmarks, Benchmark, ReportingMode};
+pub use cachegrind::CachegrindRunner;
+pub use compare::compare;
+pub use runner::{run_all, run_single};
+
+pub fn black_box<T>(dummy: T) -> T {
```
cpu, line 11: What's the story with this bit? Could you add a comment motivating this fn? 

I have passing familiarity with `core::hint::black_box` and wonder why something custom is required here.

ci-bench/Cargo.toml
djc: That failure looks unrelated. And you actually set `rust-version` in the Cargo metadata for this new crate, is that a lie?

ci-bench/src/rustls_facade.rs
```diff
@@ -0,0 +1,60 @@
+use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
+use rustls::{ConnectionCommon, SideData};
+use std::io::{Read, Write};
+
+/// Sends data to the other side of the connection through the provided writer
```
cpu, line 5: editorial nit: missing `.`
```suggestion
/// Sends data to the other side of the connection through the provided writer.
```

cpu (commented):
This is really cool!

ci-bench/Cargo.toml
aochagavia: Hmm I should have looked more closely when I copy-pasted the original `Cargo.toml` file. My natural choice would be to have no minimum rust version for this kind of code, assuming that developers of the library will always be using an up-to-date compiler. Is that ok? Or are there reasons to have a minimum rust version even for code that is not part of the library?

ci-bench/src/rustls_facade.rs
aochagavia: Thanks for the tip, I didn't know about `complete_io`. I have pushed changes to start using it.

ci-bench/src/bench_lib/benchmark.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// measure overhead instructions and subtract them from other benchmarks through
+    /// [`Benchmark::exclude_setup_instructions`]
+    pub fn hidden(mut self) -> Self {
+        self.reporting_mode = ReportingMode::Hidden;
+        self
+    }
+
+    /// Configure this benchmark to subtract the instruction count of the referenced benchmark when
+    /// reporting results
+    pub fn exclude_setup_instructions(mut self, name: String) -> Self {
```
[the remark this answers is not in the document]
aochagavia: I'll go with `String` everywhere 👍

ci-bench/src/bench_lib/cachegrind.rs
aochagavia: I'm afraid this could break in unexpected ways, because `uname -m` does not always use the same string representation of an architecture as the `ARCH` constant (see [this list](https://stackoverflow.com/a/45125525/2110623) for details).

But there is an even better option! I just discovered that the `arch` argument to `setarch` is optional since the 2.33 release (in 2018), so we can remove the `get_arch` function in its entirety.

ci-bench/src/bench_lib/cachegrind.rs
aochagavia: Cache simulation is only disabled by default since the 3.21 release, which is pretty recent (e.g. Debian 12 ships valgrind 3.19). Branch prediction stats seem to be disabled by default already in 3.19, so probably don't need to be explicitly disabled. I'll add comments explaining the relevant arguments to valgrind.

ci-bench/src/bench_lib/cachegrind.rs
aochagavia: I think that would only be necessary for local debugging if things break in weird ways. That sounds improbable, so I'd rather have the developer comment out the line of code than adding an option to the CLI. Or do you envision another use case than debugging?

ci-bench/README.md
```diff
@@ -0,0 +1,110 @@
+# CI Bench
+
+This crate is meant for CI benchmarking. It measures CPU instructions using cachegrind, outputs
```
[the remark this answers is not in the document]
aochagavia: Will do!

ci-bench/src/bench_lib/cachegrind.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        Ok(CachegrindRunner {
+            arch,
+            executable,
+            overhead_instructions,
+        })
+    }
+
+    /// Runs the benchmark at the specified index and returns the instruction counts for each side
+    pub fn run_bench(
```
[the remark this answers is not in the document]
aochagavia: Did you end up discovering? The index is necessary because it is passed to the child process (so it knows which specific benchmark it should run). Please let me know if you have any ideas to make this clearer 😉

ci-bench/src/bench_lib/mod.rs
aochagavia: I didn't know `core::hint::black_box` was stabilized last December! I'd gladly use it instead of a custom implementation if we aren't required to support a lower Rust version in the benchmarking code (currently being discussed [here](https://github.com/rustls/rustls/pull/1404#discussion_r1297432171)).

aochagavia:
Thanks for the reviews! I've resolved everything that could be resolved and left the rest open so we can discuss.

@ctz Next week I'll have a look at the issue with the resumed handshakes, thanks for pointing it out

ci-bench/Cargo.toml
djc: Actually the MSRV workflow in CI is already targeted at `-p rustls` only, so there should be no problem avoiding specifying a `rust-version` here and just using the stable compiler as used for most of the CI jobs.

aochagavia: All right, I have removed the `rust-version` line from `Cargo.toml`

ci-bench/src/bench_lib/mod.rs
aochagavia: Update: I've updated the code to use the `black_box` from the standard library

ci-bench/src/bench_lib/cachegrind.rs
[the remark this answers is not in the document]
cpu: I did thanks, sorry I should have self-resolved this at the end of my review.

ci-bench/src/bench_lib/cachegrind.rs
cpu: > Cache simulation is only disabled by default since the 3.21 release, which is pretty recent (e.g. Debian 12 ships valgrind 3.19)

Ahh! good call out. Thanks. I was looking at the upstream docs without any sense of what's shipped.

>  I'll add comments explaining the relevant arguments to valgrind. Do you think that would be enough here?

Yes, that sounds great. Thank you.

ci-bench/src/bench_lib/cachegrind.rs
cpu: Nope, that makes sense to me. Just spitballing :-)

ci-bench/src/bench_lib/mod.rs
cpu: Awesome, thanks!

ci-bench/src/bench_lib/cachegrind.rs
cpu: > But there is an even better option! I just discovered that the arch argument to setarch is optional since the 2.33 release (in 2018), so we can remove the get_arch function in its entirety.

Ah perfect, that sounds even better. Nice catch.

ci-bench/src/main.rs
djc: In practice, the `SideFn` abstraction seems to be abstracting over (a) running `make_client_config()` vs `make_server_config()` and (b) `handshake_client_side()` vs `handshake_server_side()`. If that's the goal of the abstraction, I think it would be clearer to have a trait that has these two required methods. There's also a bunch of duplication between the two `SideFn` implementations which would be nice to reduce if possible.

aochagavia:
I just pushed changes to fix the resumed handshake benchmarks. See below a bunch of instruction counts as an example. Do they look sensible to you now, @ctz?

| Scenario | Count |
| --- | ---: |
| handshake_no_resume_1.2_rsa_aes_client | 4102933 |
| handshake_session_id_1.2_rsa_aes_client | 164924 |
| handshake_tickets_1.2_rsa_aes_client | 181010 |
| handshake_no_resume_1.2_rsa_aes_server | 11603329 |
| handshake_session_id_1.2_rsa_aes_server | 173655 |
| handshake_tickets_1.2_rsa_aes_server | 195085 |
| handshake_no_resume_1.3_rsa_aes_client | 4195395 |
| handshake_session_id_1.3_rsa_aes_client | 1148836 |
| handshake_tickets_1.3_rsa_aes_client | 1150432 |
| handshake_no_resume_1.3_rsa_aes_server | 11797230 |
| handshake_session_id_1.3_rsa_aes_server | 1222456 |
| handshake_tickets_1.3_rsa_aes_server | 1227708 |

There is, however, something weird going on with the instruction counts for the `transfer` scenario. Suddenly, they have become noisy to the point of being useless: there are false positives above 2% on every run, even without recompiling the binary! It happens consistently, but only on the client side.

Cachegrind has a `cg_diff` command to compare the instruction counts between two runs, and the variation seems to be caused by `memset` and `memcpy` calls. My research suggests there is something going on with internal buffering inside the rustls client (i.e. in some runs, the buffers grow very big, while in others they remain small, causing a big difference in `memset` and `memcpy` calls). I'll be digging into it, also by looking at previous versions of the benchmarks that generate results without the noise. If anyone has a hunch about why this is happening, please share!

aochagavia:
Good news, I have confirmed the instruction count noise is due to buffering, and have a local fix that works for the transfer benchmark. Now I'll see if I can also bring down the noise in the handshakes, and after some polishing the PR should be ready for a final review :)

ctz:
> Do they look sensible to you now, @ctz?

Yes! That looks a lot better.

ci-bench/src/main.rs
aochagavia: I see. I got a bit carried away by the desire to keep the benchmarking stuff generic, but now I realize it makes the code more difficult to read. I have replaced the usage of closures with an enum to specify which code should be run for the benchmark (i.e. only the handshake vs. handshake + transfer). Haven't pushed yet but will give you a head's up when I do.

aochagavia:
Just pushed a new version which addresses the remaining feedback from @ctz and @djc, as well as solving the noise problems. Would you mind to have another look?

In the meantime I'll be looking into the necessary tweaks to make the CI go green again ;)

ctz (approved):

ci-bench/README.md
```diff
@@ -0,0 +1,110 @@
+# CI Bench
```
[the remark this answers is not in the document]
aochagavia: Oops, thanks!

ci-bench/src/cli.rs
```diff
@@ -0,0 +1,38 @@
...
+    RunSingle { index: u32, side: Side },
+    /// Compare the results from two previous benchmark runs and prints a user-friendly markdown overview
+    Compare {
+        baseline_input: PathBuf,
+        candidate_input: PathBuf,
+    },
+}
+
+#[derive(Copy, Clone, ValueEnum)]
+pub enum Side {
```
djc, line 25: Maybe we can use `rustls::Side` instead?

ci-bench/src/bench_lib/benchmark.rs
```diff
@@ -0,0 +1,179 @@
+use crate::cli::Side;
+use crate::util::key::KeyType;
```
djc, line 2: Style nit: would be nice to split these import blocks with newlines, between `crate`, external crates and `std` imports.

ci-bench/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+/// `HashMap` by a `FxHashMap`, which brings the noise down to acceptable levels in a single run).
+const RESUMED_HANDSHAKE_RUNS: usize = 3;
+
+fn main() -> anyhow::Result<()> {
+    let benchmarks = get_benchmarks()?;
+
+    let cli = Cli::parse();
+    match cli.command {
+        Command::RunAll => {
+            let executable = std::env::args().next().unwrap();
```
djc, line 46: Is there some way to get the executable name as part of clap? The direct use of `std::env::args()` in the context of a clap parser feels bit odd.

ci-bench/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+        } => {
+            bench_lib::compare(baseline_input, candidate_input)?;
+        }
+    }
+
+    Ok(())
+}
+
+/// Returns all benchmarks
+fn get_benchmarks() -> anyhow::Result<Vec<Benchmark>> {
```
djc, line 88: Naming nit: per the API guidelines, I would avoid the `get_()` prefix here.

ci-bench/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+            format!("transfer_no_resume_{params_label}"),
+            BenchmarkKind::Transfer,
+            params,
+        )
+        .exclude_setup_instructions(format!("handshake_no_resume_{params_label}")),
+    );
+}
+
+/// Runs a benchmark for the client-side of the connection
+fn run_bench_for_client(
```
djc, line 175: I think I still maintain that this would be better if you had a `Handshake` trait or similar that takes the `new_client()`, `handshake_client_side()` (maybe these can be combined?) and `handshake_client()` side out, so there's no need to deduplicate `run_bench_for_client()` and `run_bench_for_server()`.

djc (commented):
I still have a few nits.

At the top level, for my taste this has too many small files that I have to keep skipping around between (but I know this is somewhat opinionated). I would probably concatenate all the `util` modules into one.

It also feels weird to me that there's a `_lib` module inside the crate. I would restructure this to have the `main.rs` contain the executable code, move the `cli` contents in that same module, and then move all of the code in `bench_lib` one level up so it can be a proper library crate (with a `lib.rs` module).

ci-bench/src/cli.rs
aochagavia: Unfortunately we need our own type, so we can derive `ValueEnum` which is required by `clap`

djc: I guess. Alternatively we could have rustls implement `FromStr` for `Side`, but that might not be much better.

ci-bench/src/main.rs
aochagavia: I agree it's odd, but clap doesn't seem to offer this through their API (the docs have no relevant references to the word `binary` or `executable` and googling did not reveal anything either)

aochagavia:
@djc regarding your suggestions around the structure of the code, I have one general consideration and a few specific remarks.

The general consideration is that I personally prefer grouping code in a separate file when there is strong conceptual cohesion, even if that results in small files (it makes it easier for me to navigate the code, contrary to big files with less conceptual cohesion). I assumed this made things easier on other people, too, but I guess it's highly dependent on your workflow. I'll try to restrain myself a bit more, then, hoping I'll strike the right balance and you won't have to constantly remember me of it 😉.

On the specific suggestions:

- Re merging all `util::*` modules into a single `util` module. I agree that `util::config` and `util::key` could be merged, but `util::transport` has important module-level comments (those starting with `//!`) that we don't want to lose. What would you suggest we do with them? One alternative would be to leave the file alone and move it to the top level (`crate::transport`), but maybe that's just because I'm inclined to keep the file. Another alternative would be to have an inline module inside `util` (i.e. `mod transport { ... }`). Otherwise we could move the code inside `util` without ceremony, turning the module comments into normal ones `//` and ensuring they precede the relevant code (hoping future readers will go top-to-bottom and find them so they have relevant context). Any preference?
- Re merging `cli` into main. That sounds totally sensible.
- Re `bench_lib`. I agree that it is a weird choice, and while it might have made sense at the beginning (when I thought part of the code would be extracted as a library), it clearly doesn't make sense now. For that reason, instead of extracting that as a library crate, I think we should do the following (please let me know whether you agree):
  - Merge `bench_lib::runner` and `bench_lib::compare` into `main`, to get rid of the small files.
  - Move `bench_lib::cachegrind` and `bench_lib::benchmark` to the top level.

ci-bench/src/main.rs
aochagavia: Will do

ci-bench/src/main.rs
aochagavia: I'll have a look into it

aochagavia: Just pushed a commit which maces `run_bench` generic based on a `BenchStepper` trait. Does that look like what you had in mind?

ci-bench/src/cli.rs
aochagavia: I can't imagine that `FromStr` implementation would be useful anywhere else, so in my eyes the current solution is preferrable

ci-bench/src/main.rs
djc: It does seem like an improvement. I'd probably factor out the 3 common fields of the `Stepper` types into a separate type and also attach the `make_new_config()` types as methods to the `{Client,Server}SideStepper` types.

djc:
> * Re merging all `util::*` modules into a single `util` module. I agree that `util::config` and `util::key` could be merged, but `util::transport` has important module-level comments (those starting with `//!`) that we don't want to lose. What would you suggest we do with them? One alternative would be to leave the file alone and move it to the top level (`crate::transport`), but maybe that's just because I'm inclined to keep the file. Another alternative would be to have an inline module inside `util` (i.e. `mod transport { ... }`). Otherwise we could move the code inside `util` without ceremony, turning the module comments into normal ones `//` and ensuring they precede the relevant code (hoping future readers will go top-to-bottom and find them so they have relevant context). Any preference?

Inline modules seem like a decent middle ground here.

> * Re `bench_lib`. I agree that it is a weird choice, and while it might have made sense at the beginning (when I thought part of the code would be extracted as a library), it clearly doesn't make sense now. For that reason, instead of extracting that as a library crate, I think we should do the following (please let me know whether you agree):
>   
>   * Merge `bench_lib::runner` and `bench_lib::compare` into `main`, to get rid of the small files.
>   * Move `bench_lib::cachegrind` and `bench_lib::benchmark` to the top level.

These sound like sensible improvements.

aochagavia:
All right, I think I have now addressed all your comments @djc. Thanks for taking the time to review this!

djc (approved):

cpu:
Nice!

PR #1405, opened 2023-08-17
Make *ring* optional, and demonstrate how alternatives would be integrated

ctz:
This builds on #1401 to make *ring* and webpki dependencies optional. Then there is a demonstration of custom crypto using various crypto crates from the ecosystem.

TODOs:

- [x] we should run `cargo run --example client` in the `connect-tests` job
- [x] ~~this PR seems like a good venue to introduce an optional aws-lc-rs dependency, and arrange to run our tests against it~~
- [x] run `cargo hack --feature-powerset --no-dev-deps build` and `cargo hack --feature-powerset test` to make sure all combinations of features at least compile and pass tests
- [x] seems like the is some utility in making the provider example introduced here more reusable: https://github.com/rustls/rustls-ffi/issues/342#issuecomment-1680959617
  - this will be follow-on work, see discussion on #1440


djc:
I think we should postpone aws-lc-rs until a future PR -- this one is already large enough as it is.

rustls/src/signer.rs
```diff
@@ -0,0 +1,62 @@
+use crate::enums::{SignatureAlgorithm, SignatureScheme};
```
djc, line 1: Should this move into `crypto`?

rustls/src/webpki/verify.rs
```diff
@@ -47,14 +29,15 @@ pub fn verify_server_cert_signed_by_trust_anchor(
     roots: &RootCertStore,
     intermediates: &[Certificate],
     now: SystemTime,
+    supported_algs: &[&dyn webpki::SignatureVerificationAlgorithm],
```
djc, line 32: I think we wanted to avoid having items from the webpki crate in the rustls public API?

rustls/src/webpki/verify.rs
```diff
@@ -464,39 +466,130 @@ impl From<webpki::Error> for CertRevocationListError {
     }
 }
 
-static ECDSA_SHA256: SignatureAlgorithms = &[webpki::ECDSA_P256_SHA256, webpki::ECDSA_P384_SHA256];
-
-static ECDSA_SHA384: SignatureAlgorithms = &[webpki::ECDSA_P256_SHA384, webpki::ECDSA_P384_SHA384];
-
-static ED25519: SignatureAlgorithms = &[webpki::ED25519];
+type SignatureAlgorithms = &'static [&'static dyn webpki::SignatureVerificationAlgorithm];
```
djc, line 469: I don't think we should have this, makes it harder to understand the provenance (for example, how the webpki dependency is leaking into our public API).

rustls/Cargo.toml
```diff
@@ -19,13 +19,14 @@ rustversion = { version = "1.0.6", optional = true }
 log = { version = "0.4.4", optional = true }
 ring = { version = "0.16.20", optional = true }
 subtle = "2.5.0"
-webpki = { package = "rustls-webpki", version = "0.102.0-alpha.0", features = ["alloc", "std"] }
+webpki = { package = "rustls-webpki", version = "0.102.0-alpha.0", features = ["alloc", "std"], default-features = false, optional = true }
```
djc, line 22: Feels like this commit consists of 2-4 separate logical changes.

provider-example/src/aead.rs
```diff
@@ -0,0 +1,158 @@
+use chacha20poly1305::{AeadInPlace, KeyInit, KeySizeUser};
+use rustls::crypto::cipher;
+use rustls::internal::msgs::base::Payload; // FIXME
+use rustls::{ContentType, ProtocolVersion};
+
+pub struct Chacha20Poly1305;
```
djc, line 6: Most of the modules here seem to be in bottom-up ordering.

djc (commented):
Early feedback, hope it's useful.

rustls/src/webpki/verify.rs
ctz: Urgh, gross. I don't see a way of doing that without wrapping and reexporting every single `webpki::SignatureVerificationAlgorithm` value.

ctz: Actually, such a wrapper would need to be constructable outside the crate (to address the webpki-without-ring use case) with a value that is `dyn webpki::SignatureVerificationAlgorithm` which would also count as a webpki type in our public API. I'm not sure this is actually possible, given the goal is to pass through an impl of a trait exported by the webpki crate?

rustls/src/webpki/verify.rs
ctz: removed

provider-example/src/aead.rs
ctz: i think this is uniformly resolved now

ctz:
Quick note for bisecting icount changes:

```shell
$ mkdir ci-bench/commits
$ git rebase -i main --exec 'cd ci-bench && cargo run --release run-all > commits/$(git rev-parse HEAD).csv'
(...)
$ grep handshake_tickets_1.2_rsa_aes_client ci-bench/commits/*.csv
ci-bench/commits/0438e888d1617a32d244eabdf2b8c7232ea469c9.csv:handshake_tickets_1.2_rsa_aes_client,4777225
ci-bench/commits/39ab4ea727a2b7b96cec021c6e8876a2c4695b5d.csv:handshake_tickets_1.2_rsa_aes_client,4774559
ci-bench/commits/4b9efe02185956838fcb8c3bc97324847d07543e.csv:handshake_tickets_1.2_rsa_aes_client,4771167
ci-bench/commits/6d0896cd768220c68eb5f36f640023636da994c8.csv:handshake_tickets_1.2_rsa_aes_client,4769570
ci-bench/commits/9383141f6d44cccfc65a35e4e0080d77873d4a9a.csv:handshake_tickets_1.2_rsa_aes_client,4775916
ci-bench/commits/ff29c8a92082faa99aaee375dde120e169deedf7.csv:handshake_tickets_1.2_rsa_aes_client,4774989
```

Which doesn't point to a specific commit. Hmm!

aochagavia:
There is a high chance the "noteworthy" diffs you are seeing are just noise (it involves only 2 resumed handshakes, which are known to be noisy, and the diffs are close to the noise threshold). I thought I'd managed to solve the noise problem, but it looks like I'll have to tweak things some more.

ctz:
> What doesn't point to a specific commit?

Sorry, I mean the performance regression doesn't seem to be attributable to any one commit in this PR. But now I think that's because I can't see the regression locally.

> There is a very high chance the "noteworthy" diffs you are seeing are just noise

Ah, understood. Is there much that can be done for that?

I saw your comment about HashMap usage which makes sense for servers, but not clients (because they're only talking to one server ever, they would end up with their cache HashMap only ever containing one item, indexed by "localhost"?)

aochagavia:
Hmmm I just checked out the code locally and maybe it's not noise after all. Do you mind if I have a look at this next week?

ctz:
Sure, yes!

aochagavia:
After some more investigation, I discovered the stdio IPC introduces a low amount of noise that affects the resumed handshakes. I'd assume that's the problem here too (I'll be figuring out how to get rid from it, and afterwards compare again between main and this PR).

ctz:
> and `supported_verify_schemes` requires less instructions too

that's very interesting, because on main `supported_verify_schemes` is simpler. on main it is:

```
162     pub fn default_supported_verify_schemes() -> Vec<SignatureScheme> {
163         vec![
164             SignatureScheme::ECDSA_NISTP384_SHA384,
165             SignatureScheme::ECDSA_NISTP256_SHA256,
166             SignatureScheme::ED25519,
167             SignatureScheme::RSA_PSS_SHA512,
168             SignatureScheme::RSA_PSS_SHA384,
169             SignatureScheme::RSA_PSS_SHA256,
170             SignatureScheme::RSA_PKCS1_SHA512,
171             SignatureScheme::RSA_PKCS1_SHA384,
172             SignatureScheme::RSA_PKCS1_SHA256,
173         ]
174     }
```

on this branch it is:

```
520     fn supported_schemes(&self) -> Vec<SignatureScheme> {
521         self.mapping
522             .iter()
523             .map(|item| item.0)
524             .collect()
525     }
```

which _feels_ like it should take more instructions. (there are some indirections, but ones i'm assuming will be inlined away in release mode.)

aochagavia:
Oops, it looks like I didn't look carefully enough at the `+` and `-` signs in the report. `supported_verify_schemes` has in fact 10% _more_ instructions.

ctz:
Aha! That makes a lot more sense!

djc:
Stopped reviewing for a bit to await the `SignatureVerificationAlgorithm` move to pki-types.

ctz:
Ready for re-review now I think.

rustls/src/server/builder.rs
```diff
@@ -5,9 +5,9 @@ use crate::error::Error;
 use crate::server::handy;
 use crate::server::{ResolvesServerCert, ServerConfig};
 use crate::suites::SupportedCipherSuite;
+use crate::verify;
```
djc, line 8: Merge this with the following line?

rustls/src/lib.rs
```diff
@@ -409,7 +409,7 @@ pub use crate::tls12::Tls12CipherSuite;
 pub use crate::tls13::Tls13CipherSuite;
 pub use crate::verify::DigitallySignedStruct;
 pub use crate::versions::{SupportedProtocolVersion, ALL_VERSIONS, DEFAULT_VERSIONS};
-pub use crate::webpki::{RootCertStore, TrustAnchorWithDn};
+pub use crate::webpki::{RootCertStore, TrustAnchorWithDn, WebPkiSupportedAlgorithms};
```
djc, line 418: This looks kinda confusing to me. As far as I can tell, all of these types are independent of crypto providers yet they live within the `webpki` module, and the `WebPkiSupportedAlgorithms` even carries the `WebPki` prefix. How does that make sense?

rustls/src/webpki/verify.rs
```diff
@@ -134,64 +123,73 @@ impl ServerCertVerifier for WebPkiServerVerifier {
...
 impl WebPkiServerVerifier {
     /// Constructs a new `WebPkiServerVerifier`.
     ///
     /// `roots` is the set of trust anchors to trust for issuing server certs.
+    #[cfg(feature = "ring")]
     pub fn new(roots: impl Into<Arc<RootCertStore>>) -> Self {
+        Self::new_with_algorithms(roots, SUPPORTED_SIG_ALGS)
+    }
+
+    /// Constructs a new `WebPkiVerifier`.
```
djc, line 151: Nit: the type is named `WebPkiServerVerifier` now.

rustls/src/lib.rs
```diff
@@ -409,6 +414,7 @@ pub use crate::tls12::Tls12CipherSuite;
 pub use crate::tls13::Tls13CipherSuite;
 pub use crate::verify::DigitallySignedStruct;
 pub use crate::versions::{SupportedProtocolVersion, ALL_VERSIONS, DEFAULT_VERSIONS};
+#[cfg(feature = "webpki")]
```
djc, line 417: It doesn't seem like `RootCertStore` and `TrustAnchorWithDn` depend on webpki?

djc (commented):
Some thoughts.

rustls/src/lib.rs
ctz: `TrustAnchorWithDn` could live outside of `crate::webpki` now, if it would be useful independent of webpki.

`RootCertStore::add` and `RootCertStore::add_parsable_certificates` both depend on `webpki::extract_trust_anchor`, and without those functions it seems like `RootCertStore` is a more directly written `Vec<TrustAnchorWithDn>`?

rustls/src/lib.rs
ctz: I think these are all implementation details of `WebPki{Server,Client}Verifier` and so aren't relevant unless webpki is being used:

- `TrustAnchorWithDn`: needed for `WebPkiClientVerifier` to implement `ClientCertVerifier::client_auth_root_subjects`. This is mainly a polyfill over internal details pki-types/webpki's `TrustAnchor` encoding. I wonder if we should change that to avoid the need for this type?
- `WebPkiSupportedAlgorithms` is an implementation detail for `{Client,Server}CertVerifier::supported_verify_schemes`, but mostly the impedence mismatch between TLS1.2 and pki-types/webpki `SignatureVerificationAlgorithm`. I think that mismatch is OK, and feel the change to TLS1.3 which removes it supports that viewpoint.
- `RootCertStore` is really hiding implementation details of webpki -- that it needs an extra step to deal with a trust anchor available as DER.

djc: See #1459 for some steps to simplify this stuff.

rustls/src/webpki/verify.rs
```diff
@@ -431,42 +436,136 @@ impl From<webpki::Error> for CertRevocationListError {
...
-        _ => Err(PeerMisbehaved::SignedHandshakeWithUnadvertisedSigScheme.into()),
+    /// Return the first item in `mapping` that matches `scheme`.
+    fn convert_scheme(
+        &self,
+        scheme: SignatureScheme,
+    ) -> Result<&[&'static dyn SignatureVerificationAlgorithm], Error> {
+        self.mapping
+            .iter()
+            .filter(|item| item.0 == scheme)
+            .map(|item| item.1)
```
cpu, line 504: This could be combined into one step w/ `filter_map`:
```suggestion
            .filter_map(|item| if item.0 == scheme { Some(item.1) } else { None })
```

rustls/src/x509.rs
```diff
@@ -1,5 +1,6 @@
 // Additional x509/asn1 functions to those provided in webpki/ring.
 
+#[allow(dead_code)] // if webpki is compiled out
```
cpu, line 3: I think rather than indicating this with a comment we could use `cfg_attr` here:

```suggestion
#[cfg_attr(not(feature = "webpki"), allow(dead_code))]
```

rustls/src/x509.rs
```diff
@@ -18,6 +19,7 @@ pub(crate) fn wrap_in_asn1_len(bytes: &mut Vec<u8>) {
 }
 
 /// Prepend stuff to `bytes` to put it in a DER SEQUENCE.
+#[allow(dead_code)]
```
cpu, line 22: ditto RE: `cfg_attr` instead of comment.

provider-example/examples/client.rs
```diff
@@ -0,0 +1,51 @@
+use std::sync::Arc;
+
+use std::io::{stdout, Read, Write};
```
cpu, line 3: nit: remove whitespace between `use std::*` imports.

provider-example/src/aead.rs
```diff
@@ -0,0 +1,158 @@
+use chacha20poly1305::{AeadInPlace, KeyInit, KeySizeUser};
+use rustls::crypto::cipher;
+use rustls::internal::msgs::base::Payload; // FIXME
```
cpu, line 3: What's the `// FIXME` in relation to? Should it be addressed pre-merge?

provider-example/src/hash.rs
```diff
@@ -0,0 +1,42 @@
+use rustls::crypto::hash;
+use sha2::Digest;
+
+pub struct SHA256;
```
cpu, line 4: nit: in `hmac.rs` you're using `Sha256Hmac` and here `SHA256`, maybe match case with the upstream `sha2` crate and prefer camelcase vs uppercase?

provider-example/src/hash.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn algorithm(&self) -> hash::HashAlgorithm {
+        hash::HashAlgorithm::SHA256
+    }
+
+    fn output_len(&self) -> usize {
+        32
+    }
+}
+
+struct SHA256Context(sha2::Sha256);
```
cpu, line 24: ditto: re casing.

rustls/src/crypto/cipher.rs
```diff
@@ -120,7 +120,8 @@ impl Iv {
     }
 
     #[cfg(feature = "tls12")]
-    pub(crate) fn copy(value: &[u8]) -> Self {
+    /// Create a new `Iv` from a byte slice, of precisely `NONCE_LEN` bytes.
```
cpu, line 123: Could this rustdoc comment be a separate commit, or folded into earlier work?

cpu (approved):
I found a few nits but don't have anything substantial to add above what djc has already added. LGTM!

rustls/src/x509.rs
ctz: actually this is no longer needed since djc's other PR

rustls/src/lib.rs
djc: Yup, fair point about the dependency on `webpki::extract_trust_anchor`! I guess one more general question I have is, why are you making webpki optional in the first place? It doesn't seem necessary for making *ring* optional...

provider-example/src/aead.rs
ctz: ah, i forgot about that. have done the requisite refactoring to remove the need for that type to be used outside the crate.

rustls/src/lib.rs
ctz: Mmm, it is in fact completely unnecessary now webpki no longer implies a fundamental dependency on *ring*. Going to drop this commit entirely.

ctz:
nb. since the above reviews the webpki-is-optional commit has been dropped, and there are two net-new refactoring commits to `OpaqueMessage` which would benefit from a look.

rustls/src/msgs/handshake_test.rs
```diff
@@ -1201,6 +1201,6 @@ fn can_decode_server_hello_from_api_devicecheck_apple_com() {
 fn wrapped_dn_encoding() {
     let subject = b"subject";
     let dn = DistinguishedName::in_sequence(&subject[..]);
-    let expected_prefix = vec![ring::io::der::Tag::Sequence as u8, subject.len() as u8];
+    let expected_prefix = vec![0x30u8, subject.len() as u8];
```
djc, line 1204: Ah, sorry for sneaking that in. Still, maybe this should be a `const DER_SEQUENCE` or whatever? IIRC you worked on the no magic constants lint that one time!

rustls/src/webpki/verify.rs
```diff
@@ -408,42 +414,135 @@ impl From<webpki::Error> for CertRevocationListError {
...
+/// how they map to TLS `SignatureScheme`s.
+#[derive(Clone, Copy)]
+#[allow(unreachable_pub)]
+pub struct WebPkiSupportedAlgorithms {
+    /// A list of all supported signature verification algorithms.
+    ///
+    /// Used for verifying certificate chains.
+    ///
+    /// The order of this list is not significant.
+    pub all: &'static [&'static dyn SignatureVerificationAlgorithm],
```
djc, line 427: Okay, but so `SignatureVerificationAlgorithm` is in pki-types now, so why do we need to wrap over this? Sorry to keep harping at this, something keeps nagging at me about this types which feels unidiomatic and I can't quite grasp why it's necessary.

rustls/src/webpki/verify.rs
```diff
@@ -408,42 +414,135 @@ impl From<webpki::Error> for CertRevocationListError {
...
+        write!(f, " }}")
+    }
+}
 
-        SignatureScheme::RSA_PKCS1_SHA256 => Ok(RSA_SHA256),
-        SignatureScheme::RSA_PKCS1_SHA384 => Ok(RSA_SHA384),
-        SignatureScheme::RSA_PKCS1_SHA512 => Ok(RSA_SHA512),
+#[cfg(feature = "ring")]
+#[test]
+fn test_webpkisupportedalgorithms_is_debug() {
```
djc, line 458: IMO tests should go to the bottom of the module, in a `tests` module. I also dislike your pattern of repeating type names without word separators in test names, I think we should use a proper `UpperCamelCase` to `snake_case` conversion here and introduce underscores between words.

rustls/src/webpki/verify.rs
```diff
@@ -408,42 +414,135 @@ impl From<webpki::Error> for CertRevocationListError {
...
+    /// This is one (`SignatureScheme`) to many ([`SignatureVerificationAlgorithm`]) because
+    /// (depending on the protocol version) there is not necessary a 1-to-1 mapping.
+    ///
+    /// For TLS1.2, all `SignatureVerificationAlgorithm`s are tried in sequence.
+    ///
+    /// For TLS1.3, only the first is tried.
+    ///
+    /// The supported schemes in this mapping is communicated to the peer and the order is significant.
+    /// The first mapping is our highest preference.
+    pub mapping: &'static [(
```
djc, line 440: Why don't we want to this to just be a top-level `static`?

rustls/src/msgs/message.rs
```diff
@@ -93,6 +93,21 @@ impl OpaqueMessage {
...
+    pub fn payload(&self) -> &[u8] {
+        &self.payload.0
+    }
+
+    /// Access the message payload as a mutable `Vec<u8>`.
+    ///
+    /// This is useful for decrypting a message in-place.  After the message
+    /// is decrypted, call [`OpaqueMessage::into_plain_message()`] or
+    /// [`OpaqueMessage::into_tls13_unpadded_message()`] (depending on the
+    /// protocol version).
```
djc, line 106: The context about `into_plain_message()` etc feels out of place here, since this is a fairly straightforward getter...

rustls/src/crypto/cipher.rs
djc, line 123: Nit: docstring before attributes?

provider-example/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+        ],
+        hmac_provider: &hmac::Sha256Hmac,
+        aead_alg: &aead::Chacha20Poly1305,
+    });
+
+mod aead;
+mod hash;
+mod hmac;
+mod kx;
+mod verify;
```
djc, line 65: Maybe counter-intuitively from my other style considerations, I usually put module declarations early in the module, between imports and other items.

cpu (approved):
> nb. since the above reviews the webpki-is-optional commit has been dropped, and there are two net-new refactoring commits to OpaqueMessage which would benefit from a look.

New changes look good to me :+1:

rustls/src/webpki/verify.rs
ctz: not quite sure what you mean. it needs to be configurable rather than static so the `WebPkiServerVerifier` (etc.) can be reused without ring. there's a separate constructor of this type in https://github.com/rustls/rustls/pull/1405/commits/ba26a6d7c5037fb30362640ed33b43cdff5ace22#diff-8749c6ccb1619fb20bb8abc0a3651cc58eaf05ff79a0c395c7b9ef76047cea05R11-R17 backed by https://crates.io/crates/rsa if that makes the goal any clearer?

djc: Okay, so this is really a way for the crypto provider to provide it's signature verification algorithms, and a mapping from `SignatureScheme` to them? And you used the `WebPki` prefix since this is specifically used for the webpki-backed verifiers?

djc: I don't know, it feels more like a crypto provider concern that happens to be used by the verifiers, not all that webpki-specific.

rustls/src/msgs/message.rs
ctz: moved up into the main docs for this type

rustls/src/webpki/verify.rs
djc: I think I'm convinced that this a decent design. Should we be worried about the link-time inclusion of all signature verification algorithms by default?

djc (approved):

tarcieri:
FYI, the @RustCrypto project is going to try to put together a provider based on the example.

You can follow along at https://github.com/RustCrypto/rustls-rustcrypto (currently an empty repo)

djc:
@tarcieri maybe collaborate with @stevefan1999-personal, they started on https://github.com/stevefan1999-personal/rustls-provider-rustcrypto already?

rustls/src/webpki/verify.rs
ctz: I think that is the status quo. Though I guess this does now give people the opportunity to avoid that if desired.

tarcieri:
Wasn't aware @stevefan1999-personal already started on one but I think it would make sense for this provider to live under the @RustCrypto organization

stevefan1999-personal:
@tarcieri Yep. This should be a community collaboration effort. And not only RustCrypto, we also have https://github.com/franziskuskiefer/evercrypt-rust/ too, which is a formally verified TLS library from F*, but ported to Rust.

tarcieri:
@stevefan1999-personal would you want to upstream your current work to our repo? We can give you code review in the process.

stevefan1999-personal:
> @stevefan1999-personal would you want to upstream your current work to our repo? We can give you code review in the process.

Sure thing! The server example should be good for review, but I will need to make some adjustment first.

PR #1412, opened 2023-08-22
SECURITY.md: use github vuln reporting tool

ctz:
We have a mailing list for this. But, the first time that was used for real, it didn't go very well:

- the report and a follow-up went into spam. A private google group delivering to gmail -- you'd think this would work well, but  it did not.
- there was only me in the group.

Github now has a "private vulnerability reporting" feature that should be better for getting reports to the right people quickly. Let's try that?


djc (approved):
Same for webpki?

PR #1414, opened 2023-08-24
Support aws-lc-rs as an optional dependency

ctz:
The end goal of this is to let people opt-in to using aws-lc-rs in FIPS mode. The next step in that direction is to support and test aws-lc-rs in non-FIPS mode, which is this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,10 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
 
+      - name: Install NASM for aws-lc-rs on Windows
+        if: runner.os == 'Windows'
+        uses: ilammy/setup-nasm@v1
+
       - name: cargo build (debug; default features)
         run: cargo build --locked
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,23 +88,29 @@ jobs:
 
       - name: cargo build (debug; default features)
         run: cargo build --locked
+        working-directory: rustls
 
       - name: cargo test (debug; default features)
         run: cargo test --locked
+        working-directory: rustls
         env:
           RUST_BACKTRACE: 1
 
       - name: cargo test (debug; no default features)
         run: cargo test --locked --no-default-features
+        working-directory: rustls
 
       - name: cargo test (debug; no default features; tls12)
         run: cargo test --locked --no-default-features --features tls12
+        working-directory: rustls
 
       - name: cargo test (debug; no default features; aws-lc-rs,tls12)
         run: cargo test --no-default-features --features aws_lc_rs,tls12
+        working-directory: rustls
 
       - name: cargo test (release; no run)
         run: cargo test --locked --release --no-run
+        working-directory: rustls
 
   bogo:
     name: BoGo test suite
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,11 @@ jobs:
         env:
           RUST_BACKTRACE: 1
 
+      - name: cargo test (debug; aws-lc-rs)
+        run: cargo test --no-default-features --features aws_lc_rs,tls12,quic,read_buf,logging
+        env:
+          RUST_BACKTRACE: 1
+
       - name: cargo build (debug; rustls-provider-example)
         run: cargo build --locked -p rustls-provider-example
 
@@ -95,6 +100,9 @@ jobs:
       - name: cargo test (debug; no default features; tls12)
         run: cargo test --locked --no-default-features --features tls12
 
+      - name: cargo test (debug; no default features; aws-lc-rs,tls12)
+        run: cargo test --no-default-features --features aws_lc_rs,tls12
+
       - name: cargo test (release; no run)
         run: cargo test --locked --release --no-run
 
@@ -116,9 +124,17 @@ jobs:
           go-version: "1.20"
           cache: false
 
-      - name: Run test suite
+      - name: Run test suite (ring)
         working-directory: bogo
         run: ./runme
+        env:
+          BOGO_SHIM_PROVIDER: ring
+
+      - name: Run test suite (aws-lc-rs)
+        working-directory: bogo
+        run: ./runme
+        env:
+          BOGO_SHIM_PROVIDER: aws-lc-rs
 
 
   fuzz:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,6 +108,31 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7fb76b0a64f839f9e2be9871ea670a197a3e1d4b9634d741ec1456102a4fbaba"
+dependencies = [
+ "aws-lc-sys",
+ "mirai-annotations",
+ "untrusted 0.7.1",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "de27c152edd365909b8fe952bce7617c910bd413b5b4bb9b0238d37e412f7e2a"
+dependencies = [
+ "bindgen",
+ "cmake",
+ "dunce",
+ "libc",
+ "paste",
+]
+
+[[package]]
 name = "backtrace"
 version = "0.3.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -139,6 +164,29 @@ name = "bencher"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7dfdb4953a096c551ce9ace855a604d702e6e62d77fac690575ae347571717f5"
+
+[[package]]
+name = "bindgen"
+version = "0.68.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "726e4313eb6ec35d2730258ad4e15b547ee75d6afaa1361a922e78e59b7d8078"
+dependencies = [
+ "bitflags 2.4.1",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "log",
+ "peeking_take_while",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn",
+ "which",
+]
 
 [[package]]
 name = "bitflags"
@@ -189,6 +237,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cexpr"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
+dependencies = [
+ "nom",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -230,6 +287,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "clang-sys"
+version = "1.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c688fc74432808e3eb684cae8830a86be1d66a2bd58e1f248ed0960a590baf6f"
+dependencies = [
+ "glob",
+ "libc",
+ "libloading",
+]
+
+[[package]]
 name = "clap"
 version = "4.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -268,6 +336,15 @@ name = "clap_lex"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "702fc72eb24e5a1e48ce58027a675bc24edd52096d5397d4aea7c6dd9eca0bd1"
+
+[[package]]
+name = "cmake"
+version = "0.1.50"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a31c789563b815f77f4250caee12365734369f942439b7defd71e18a48197130"
+dependencies = [
+ "cc",
+]
 
 [[package]]
 name = "colorchoice"
@@ -410,6 +487,12 @@ dependencies = [
  "serde",
  "strsim",
 ]
+
+[[package]]
+name = "dunce"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56ce8c6da7551ec6c462cbaf3bfbc75131ebbfa1c944aeaa9dab51ca1c5f0c3b"
 
 [[package]]
 name = "either"
@@ -556,6 +639,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6fb8d784f27acf97159b40fc4db5ecd8aa23b9ad5ef69cdd136d3bc80665f0c0"
 
 [[package]]
+name = "glob"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
+
+[[package]]
 name = "h2"
 version = "0.3.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -654,6 +743,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6c49c37c09c17a53d937dfbb742eb3a961d65a994e6bcdcf37e7399d0cc8ab5e"
 dependencies = [
  "digest",
+]
+
+[[package]]
+name = "home"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5444c27eef6923071f7ebcc33e3444508466a76f7a2b93da00ed6e19f30c1ddb"
+dependencies = [
+ "windows-sys",
 ]
 
 [[package]]
@@ -776,10 +874,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "lazycell"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "830d08ce1d1d941e6b30645f1a0eb5643013d835ce3779a5fc208261dbe10f55"
+
+[[package]]
 name = "libc"
 version = "0.2.150"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "89d92a4743f9a61002fae18374ed11e7973f530cb3a3255fb354818118b2203c"
+
+[[package]]
+name = "libloading"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b67380fd3b2fbe7527a606e18729d21c6f3951633d0500574c4dc22d2d638b9f"
+dependencies = [
+ "cfg-if",
+ "winapi",
+]
 
 [[package]]
 name = "libm"
@@ -846,6 +960,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "minimal-lexical"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
+
+[[package]]
 name = "miniz_oxide"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -864,6 +984,22 @@ dependencies = [
  "log",
  "wasi",
  "windows-sys",
+]
+
+[[package]]
+name = "mirai-annotations"
+version = "1.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c9be0862c1b3f26a88803c4a49de6889c10e608b3ee9344e6ef5b45fb37ad3d1"
+
+[[package]]
+name = "nom"
+version = "7.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d273983c5a657a70a3e8f2a01329822f3b8c8172b73826411a55751e404a0a4a"
+dependencies = [
+ "memchr",
+ "minimal-lexical",
 ]
 
 [[package]]
@@ -969,6 +1105,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "de3145af08024dea9fa9914f381a17b8fc6034dfb00f3a84013f7ff43f29ed4c"
+
+[[package]]
+name = "peeking_take_while"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
+
+[[package]]
 name = "pem"
 version = "3.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1054,6 +1202,16 @@ name = "ppv-lite86"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
+
+[[package]]
+name = "prettyplease"
+version = "0.2.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae005bd773ab59b4725093fd7df83fd7892f7d8eafb48dbd7de6e024e4215f9d"
+dependencies = [
+ "proc-macro2",
+ "syn",
+]
 
 [[package]]
 name = "proc-macro2"
@@ -1246,6 +1404,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d626bb9dae77e28219937af045c257c28bfd3f69333c512553507f5f9798cb76"
 
 [[package]]
+name = "rustc-hash"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
+
+[[package]]
 name = "rustc_version"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1283,6 +1447,7 @@ dependencies = [
 name = "rustls"
 version = "0.22.0-alpha.4"
 dependencies = [
+ "aws-lc-rs",
  "base64",
  "bencher",
  "env_logger",
@@ -1398,6 +1563,7 @@ version = "0.102.0-alpha.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34d9ed3a8267782ba32d257ff5b197b63eef19a467dbd1be011caaae35ee416e"
 dependencies = [
+ "aws-lc-rs",
  "ring 0.17.5",
  "rustls-pki-types",
  "untrusted 0.9.0",
@@ -1461,6 +1627,12 @@ dependencies = [
  "cpufeatures",
  "digest",
 ]
+
+[[package]]
+name = "shlex"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7cee0529a6d40f580e7a5e6c495c8fbfe21b7b52795ed4bb5e62cdf92bc6380"
 
 [[package]]
 name = "signature"
@@ -1830,6 +2002,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42157929d7ca9c353222a4d1763c52ef86d25d0fd2eca66076df5975fd4e25ed"
 dependencies = [
  "rustls-pki-types",
+]
+
+[[package]]
+name = "which"
+version = "4.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "87ba24419a2078cd2b0f2ede2691b6c66d8e47836da3b6db8265ebad47afbfc7"
+dependencies = [
+ "either",
+ "home",
+ "once_cell",
+ "rustix",
 ]
 
 [[package]]
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -14,10 +14,11 @@ use itertools::Itertools;
 use rayon::iter::Either;
 use rayon::prelude::*;
 use rustls::client::Resumption;
-use rustls::crypto::ring::{cipher_suite, Ticketer};
+use rustls::crypto::{aws_lc_rs, ring};
 use rustls::server::{NoServerSessionStorage, ServerSessionMemoryCache, WebPkiClientVerifier};
 use rustls::{
-    ClientConfig, ClientConnection, ProtocolVersion, RootCertStore, ServerConfig, ServerConnection,
+    CipherSuite, ClientConfig, ClientConnection, ProtocolVersion, RootCertStore, ServerConfig,
+    ServerConnection,
 };
 
 use crate::benchmark::{
@@ -143,7 +144,7 @@ fn main() -> anyhow::Result<()> {
 
             let handshake_buf = &mut [0u8; 262144];
             let resumption_kind = black_box(bench.kind.resumption_kind());
-            let params = black_box(bench.params);
+            let params = black_box(&bench.params);
             let io = StepperIo {
                 reader: &mut stdin,
                 writer: &mut stdout,
@@ -153,7 +154,7 @@ fn main() -> anyhow::Result<()> {
                 Side::Server => run_bench(
                     ServerSideStepper {
                         io,
-                        config: ServerSideStepper::make_config(&params, resumption_kind),
+                        config: ServerSideStepper::make_config(params, resumption_kind),
                     },
                     bench.kind,
                 ),
@@ -161,7 +162,7 @@ fn main() -> anyhow::Result<()> {
                     ClientSideStepper {
                         io,
                         resumption_kind,
-                        config: ClientSideStepper::make_config(&params, resumption_kind),
+                        config: ClientSideStepper::make_config(params, resumption_kind),
                     },
                     bench.kind,
                 ),
@@ -198,7 +199,7 @@ fn main() -> anyhow::Result<()> {
 /// Returns all benchmarks
 fn all_benchmarks() -> anyhow::Result<Vec<Benchmark>> {
     let mut benchmarks = Vec::new();
-    for &param in ALL_BENCHMARK_PARAMS {
+    for param in all_benchmarks_params() {
         add_benchmark_group(&mut benchmarks, param);
     }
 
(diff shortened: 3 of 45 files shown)


ctz:
Raised https://github.com/aws/aws-lc-rs/issues/225 upstream for windows build issue

ctz:
# aws-lc-rs vs *ring* performance comparison

Repro instructions (in ci-bench/):

```shell
$ cargo run --release --no-default-features --features aws_lc_rs run-all > aws.csv
$ cargo run --release --no-default-features --features ring run-all > ring.csv
$ cargo run --release compare ring.csv aws.csv
```

### Noteworthy instruction count differences
| Scenario | Baseline | Candidate | Diff |
| --- | ---: | ---: | ---: |
| handshake_no_resume_1.3_ecdsa_aes_client | 14254834 | 5186655 | ✅ -9068179 (-63.61%) |
| handshake_no_resume_1.3_ecdsa_chacha_client | 14258070 | 5190268 | ✅ -9067802 (-63.60%) |
| handshake_no_resume_1.2_rsa_aes_server | 11880207 | 15189727 | ⚠️ 3309520 (27.86%) |
| handshake_no_resume_1.3_rsa_aes_server | 12081538 | 15391438 | ⚠️ 3309900 (27.40%) |
| handshake_no_resume_1.3_rsa_chacha_server | 12091835 | 15386478 | ⚠️ 3294643 (27.25%) |
| handshake_no_resume_1.2_rsa_aes_client | 4377134 | 3287857 | ✅ -1089277 (-24.89%) |
| handshake_no_resume_1.3_rsa_aes_client | 4481555 | 3408666 | ✅ -1072889 (-23.94%) |
| handshake_no_resume_1.3_rsa_chacha_client | 4492213 | 3419482 | ✅ -1072731 (-23.88%) |
| transfer_no_resume_1.2_rsa_aes_client | 6179161 | 7251232 | ⚠️ 1072071 (17.35%) |
| handshake_session_id_1.3_ecdsa_chacha_client | 34458544 | 30403832 | ✅ -4054712 (-11.77%) |
| handshake_session_id_1.3_ecdsa_aes_client | 34474859 | 30429771 | ✅ -4045088 (-11.73%) |
| handshake_tickets_1.3_ecdsa_chacha_client | 34576145 | 30556522 | ✅ -4019623 (-11.63%) |
| handshake_session_id_1.3_rsa_chacha_client | 34677136 | 30655462 | ✅ -4021674 (-11.60%) |
| handshake_session_id_1.3_rsa_aes_client | 34703748 | 30680283 | ✅ -4023465 (-11.59%) |
| handshake_tickets_1.3_ecdsa_aes_client | 34591539 | 30584691 | ✅ -4006848 (-11.58%) |
| handshake_tickets_1.3_rsa_chacha_client | 34814268 | 30792435 | ✅ -4021833 (-11.55%) |
| handshake_tickets_1.3_rsa_aes_client | 34825783 | 30815182 | ✅ -4010601 (-11.52%) |
| handshake_no_resume_1.3_ecdsa_chacha_server | 1848377 | 2028487 | ⚠️ 180110 (9.74%) |
| handshake_no_resume_1.3_ecdsa_aes_server | 1845343 | 2024728 | ⚠️ 179385 (9.72%) |
| handshake_tickets_1.2_rsa_aes_server | 4923003 | 5386002 | ⚠️ 462999 (9.40%) |
| handshake_session_id_1.3_ecdsa_aes_server | 36257315 | 32968266 | ✅ -3289049 (-9.07%) |
| handshake_session_id_1.3_ecdsa_chacha_server | 36207523 | 32935489 | ✅ -3272034 (-9.04%) |
| handshake_session_id_1.3_rsa_aes_server | 36258480 | 33004693 | ✅ -3253787 (-8.97%) |
| handshake_tickets_1.3_rsa_aes_server | 36506062 | 33242247 | ✅ -3263815 (-8.94%) |
| handshake_tickets_1.3_ecdsa_chacha_server | 36481057 | 33220849 | ✅ -3260208 (-8.94%) |
| handshake_tickets_1.3_ecdsa_aes_server | 36504567 | 33244445 | ✅ -3260122 (-8.93%) |
| handshake_session_id_1.3_rsa_chacha_server | 36206195 | 32974775 | ✅ -3231420 (-8.93%) |
| handshake_tickets_1.3_rsa_chacha_server | 36480498 | 33250947 | ✅ -3229551 (-8.85%) |
| handshake_tickets_1.2_rsa_aes_client | 4773214 | 5194657 | ⚠️ 421443 (8.83%) |
| handshake_session_id_1.2_rsa_aes_client | 4458714 | 4818240 | ⚠️ 359526 (8.06%) |
| handshake_session_id_1.2_rsa_aes_server | 4467027 | 4801548 | ⚠️ 334521 (7.49%) |
| transfer_no_resume_1.3_rsa_aes_server | 4842773 | 4901545 | ⚠️ 58772 (1.21%) |
| transfer_no_resume_1.2_rsa_aes_server | 4835302 | 4872061 | ⚠️ 36759 (0.76%) |
| transfer_no_resume_1.3_rsa_chacha_server | 8285777 | 8346719 | ⚠️ 60942 (0.74%) |
| transfer_no_resume_1.3_ecdsa_aes_client | 6190907 | 6211318 | ⚠️ 20411 (0.33%) |
| transfer_no_resume_1.3_rsa_aes_client | 6190443 | 6207868 | ⚠️ 17425 (0.28%) |
| transfer_no_resume_1.3_ecdsa_aes_server | 4843713 | 4856989 | ⚠️ 13276 (0.27%) |
### Other instruction count differences
<details>
<summary>Click to expand</summary>

| Scenario | Baseline | Candidate | Diff |
| --- | ---: | ---: | ---: |
| transfer_no_resume_1.3_rsa_chacha_client | 9661170 | 9675706 | 14536 (0.15%) |
| transfer_no_resume_1.3_ecdsa_chacha_client | 9663380 | 9674329 | 10949 (0.11%) |
| transfer_no_resume_1.3_ecdsa_chacha_server | 8286803 | 8295555 | 8752 (0.11%) |
</details>

The large regressions in `handshake_no_resume_1.2_rsa_aes_server` et al are because aws-lc-rs is doing much more RSA key validation than *ring*. As an example, *ring* doesn't do primality testing on the private primes, and there is justification in the code for that (because: side channels, which is fair enough).

This PR also adds the micro benchmarks added to work that out:

```
$ RUSTFLAGS=--cfg=bench cargo bench --all-features load_and_validate_rsa
(...)
running 4 tests
test crypto::aws_lc_rs::sign::benchmarks::bench_load_and_validate_rsa2048 ... bench:     187,227 ns/iter (+/- 3,637)
test crypto::aws_lc_rs::sign::benchmarks::bench_load_and_validate_rsa4096 ... bench:     619,634 ns/iter (+/- 15,841)
test crypto::ring::sign::benchmarks::bench_load_and_validate_rsa2048      ... bench:      31,714 ns/iter (+/- 735)
test crypto::ring::sign::benchmarks::bench_load_and_validate_rsa4096      ... bench:     123,337 ns/iter (+/- 1,808)
```

rustls/src/client/client_conn.rs
```diff
@@ -246,6 +246,15 @@ impl ClientConfig {
         Self::builder_with_provider(crate::crypto::ring::RING)
     }
 
+    #[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
+    /// Create a builder to build up the client configuration with the default
+    /// [`CryptoProvider`].
+    ///
+    /// For more information, see the [`ConfigBuilder`] documentation.
+    pub fn builder() -> ConfigBuilder<Self, WantsCipherSuites> {
```
cpu, line 254: It's hard to articulate, but describing this as the "default CryptoProvider" feels a little bit confusing. I guess I'm assuming we'd only have one of the `ring` or `aws_lc_rs` features enabled by default right?

Since this is conditional on the *ring* feature being absent and the aws_lc_rs feature being present maybe it would be less confusing to explicitly say it will use aws-lc-rs.

rustls/src/lib.rs
```diff
@@ -379,12 +379,22 @@ pub mod internal {
...
+#[cfg(feature = "ring")]
+use crate::crypto::ring as primary_provider;
+
 // The public interface is:
 pub use crate::builder::{
     ConfigBuilder, ConfigSide, WantsCipherSuites, WantsKxGroups, WantsVerifier, WantsVersions,
 };
 pub use crate::common_state::{CommonState, IoState, Side};
 pub use crate::conn::{Connection, ConnectionCommon, Reader, SideData, Writer};
+
```
cpu, line 397: is this whitespace diff intentional?

rustls/src/server/server_conn.rs
```diff
@@ -360,6 +360,15 @@ impl ServerConfig {
         Self::builder_with_provider(crate::crypto::ring::RING)
     }
 
+    #[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
+    /// Create builder to build up the server configuration with the default
+    /// `CryptoProvider`.
+    ///
+    /// For more information, see the [`ConfigBuilder`] documentation.
+    pub fn builder() -> ConfigBuilder<Self, WantsCipherSuites> {
```
cpu, line 368: If my comment in `client_conn.rs` resonates this site will also need updating.

rustls/src/client/client_conn.rs
ctz: > I'm assuming we'd only have one of the ring or aws_lc_rs features enabled by default right?

Maybe a table would make the intention clearer:

| Features | What provider backs `ClientConfig::builder()` ||
|-|-|-|
| _none_ | Absent ||
| `ring` | `ring::RING` | <- crate default |
| `aws_lc_rs` | `aws_lc_rs::AWS_LC_RS` ||
| `aws_lc_rs`, `ring` | `ring::RING` |

ctz: > Since this is conditional on the ring feature being absent and the aws_lc_rs feature being present maybe it would be less confusing to explicitly say it will use aws-lc-rs.

Done that (for all four `builder()` functions)

rustls/src/crypto/ring/mod.rs
```diff
@@ -2,7 +2,8 @@ use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::rand::GetRandomFailed;
 use crate::suites::SupportedCipherSuite;
 
-use ring::rand::{SecureRandom, SystemRandom};
+pub use ring as ringlike;
```
djc: Nit: I feel like calling it `ring_like` would be more aligned with the naming guidelines.

Cargo.lock
```diff
@@ -23,9 +23,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.5.0"
+version = "0.6.4"
```
djc, line 26: This feels odd as a separate commit? Presumably some of these changes should be included in the previous commit that bumps the webpki alpha, and this PR otherwise doesn't really need to update unrelated dependencies?

.github/workflows/build.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: "1.60"
+          toolchain: "1.61"
```
djc, line 73: Would be nice to mention the rationale for doing this here/now in the commit message -- is it because aws-lc-rs needs this?

rustls/Cargo.toml
```diff
@@ -16,6 +16,7 @@ build = "build.rs"
 rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
+aws-lc-rs = { version = "1.0.0", optional = true }
```
djc, line 19: Nit: I would spell this as `1`, since the `.0.0` are meaningless anyway.

rustls/src/lib.rs
```diff
@@ -248,6 +248,20 @@
 //!   which is used for cryptography.
 //!   Without this feature, these items must be provided externally to the core
 //!   rustls crate.
+//!
+//! - `aws_lc_rs`: this makes the rustls crate depend on the aws-lc-rs crate,
```
djc, line 252: I feel like this is where it gets tricky with the builder setup? As long as we have an implicit default it is possible for other dependencies to change the crypto provider without making it obvious, which seems like a bad idea.

Cargo.lock
```diff
@@ -81,6 +81,31 @@ version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
+[[package]]
```
djc, line 84: IMO this should go in the previous commit that adds the dependency.

rustls/src/client/builder.rs
```diff
@@ -4,21 +4,21 @@ use crate::client::{ClientConfig, ResolvesClientCert};
 use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::key_log::NoKeyLog;
 use crate::suites::SupportedCipherSuite;
-#[cfg(feature = "ring")]
+#[cfg(any(feature = "ring", feature = "aws_lc_rs"))]
```
djc, line 7: Instead of this, could/should we refactor the module hierarchy so that these things no longer have to be source from the current provider module?

rustls/src/client/handy.rs
```diff
@@ -5,10 +5,10 @@ use crate::msgs::persist;
 use crate::sign;
 use crate::NamedGroup;
 use crate::ServerName;
-#[cfg(feature = "ring")]
-use crate::{crypto::ring, error::Error};
+#[cfg(any(feature = "ring", feature = "aws_lc_rs"))]
```
djc, line 8: I think this commit should probably be squashed into one of the earlier commits.

rustls/tests/api.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(feature = "ring")]
+#![cfg(any(feature = "ring", feature = "aws_lc_rs"))]
```
djc, line 1: And this one, too.

.github/workflows/build.yml
```diff
@@ -51,6 +51,11 @@ jobs:
         env:
           RUST_BACKTRACE: 1
 
+      - name: cargo test (debug; aws-lc-rs)
```
djc, line 54: Maybe squash this with the previous commit? These seem tightly coupled.

rustls/src/crypto/ring/hmac.rs
```diff
@@ -5,7 +5,7 @@ use crate::crypto;
 
 pub(crate) static HMAC_SHA256: Hmac = Hmac(&ringlike::hmac::HMAC_SHA256);
 pub(crate) static HMAC_SHA384: Hmac = Hmac(&ringlike::hmac::HMAC_SHA384);
-#[cfg(all(test, feature = "tls12"))]
+#[allow(dead_code)] // only for TLS1.2 prf test
```
djc, line 8: If this remains test-only can we at least still guard it with `cfg(test)`?

rustls/src/tls12/prf.rs
```diff
@@ -45,6 +45,19 @@ mod tests {
         assert_eq!(expect.len(), output.len());
         assert_eq!(expect.to_vec(), output.to_vec());
     }
+
+    #[test]
+    fn check_sha384() {
```
djc, line 50: Should we link to the source in the code instead of the commit message?

rustls/src/crypto/ring/mod.rs
```diff
@@ -44,6 +30,38 @@ impl CryptoProvider for Ring {
     }
 }
 
+// Modules below here should always use `super::ringlike` to access a
```
djc, line 33: Why is this order better? I feel like we usually keep module declarations and module-level imports before any other items.

djc (commented):
Cool!

.github/workflows/build.yml
ctz: rustls-webpki -> ring 0.17 needs it

djc: Ah, forgot about that.

Cargo.lock
ctz: ah, i did just plain `cargo update`. have reworked the two cargo.lock commits to make minimal changes, and squashed them into the previous commits (though i'm expecting that this will make rebasing very annoying)

rustls/src/crypto/ring/hmac.rs
ctz: have done that

rustls/src/client/builder.rs
ctz: I think this would mean losing APIs which I think usefully hide complexity from downstream users, for example:

`.with_client_auth_cert(cert_chain, key_der)` would become something like `.with_client_cert_resolver(Arc::new(rustls::client::AlwaysResolvesClientCert::new(rustls::sign::CertifiedKey::new(cert_chain, rustls::crypto::ring::sign::any_supported_type(priv_key).expect("cannot parse private key")))))`

rustls/src/crypto/ring/mod.rs
ctz: have dropped this commit, but please re-review the ordering in `src/crypto/{ring,aws-lc-rs}/mod.rs`

ctz:
This approach has got a bit more tricky as aws-lc-rs and *ring*'s APIs have diverged; I have shimmed this but its not ideal. Hopefully aws-lc-rs tracks *ring*'s API over time.

ctz:
Reordered things slightly, and marked which commits I plan to squash subsequently.

rustls/src/lib.rs
djc: Did you have any thoughts on this?

rustls/src/client/client_conn.rs
cpu: Thanks, I think this is clearer now.

cpu:
> This approach has got a bit more tricky as aws-lc-rs and ring's APIs have diverged; I have shimmed this but its not ideal. Hopefully aws-lc-rs tracks ring's API over time.

WDYT about opening an upstream issue for this? I think it would be helpful to know what `aws-lc-rs` intends to do when the upstream *ring* API changes. The upstream [dev-dependency](https://github.com/aws/aws-lc-rs/pull/235/files) switched to 0.17 but I don't think there's been any action to resolve the API changes.

rustls/src/client/builder.rs
ctz: Oh, I see. I'll look into that direction. Thanks!

rustls/src/lib.rs
ctz: I think there are two classes of downstream users:

- would like TLS to work, but have little interest or opinion in detailed decisions like which `CryptoProvider` they are using. I think this sort of user can accept the default crate features and call `ClientConfig::builder()`. This has stable ("additive") behaviour irrespective of the choices made by other crates in their dependency tree.
- would like to configure details, and would call `ClientConfig::builder_with_provider()` to name a specific provider.

That leaves the possibility (with the PR as it currently stands) that someone is in the second camp, but (having opted out of this crate's default features and enabled `aws_lc_rs`) still call `ClientConfig::builder()`. This user would be a risk of getting a *ring*-backed `CryptoProvider` via their dependencies. Maybe we shouldn't provide aws-lc-rs-backed `builder()`?

djc: Yes, I think that would be better.

ctz:
> WDYT about opening an upstream issue for this?

Opened https://github.com/aws/aws-lc-rs/issues/245 for this

rustls/src/lib.rs
cpu: I agree that I think it would be better as an explicit choice using `builder_with_provider`.

rustls/src/client/builder.rs
cpu: Can this thread be considered resolved by https://github.com/rustls/rustls/pull/1414/commits/68792f97ab20d12c38c465c2307aec144958934d?

cpu (approved):
Mod the discussion of the builder bits I think this looks great. Nice work :-)

rustls/src/client/builder.rs
cpu: Can this thread be considered resolved by https://github.com/rustls/rustls/commit/68792f97ab20d12c38c465c2307aec144958934d?

rustls/src/client/client_conn.rs
```diff
@@ -241,15 +241,24 @@ impl fmt::Debug for ClientConfig {
...
     }
 
+    #[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
+    /// Create a builder to build up the client configuration with the default
+    /// [`CryptoProvider`]: [`crate::crypto::aws_lc_rs::AWS_LC_RS`].
+    ///
+    /// For more information, see the [`ConfigBuilder`] documentation.
+    pub fn builder() -> ConfigBuilder<Self, WantsCipherSuites> {
+        Self::builder_with_provider(crate::crypto::aws_lc_rs::AWS_LC_RS)
+    }
```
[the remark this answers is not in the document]
ctz: Yes, precisely. Have just added this as 147beab (needs some splitting/squashing, but the substance is there)

djc:
What's the next step for this? I should probably do another round of review, but would prefer to do that after rebasing and squashing has happened.

ctz:
> What's the next step for this?

I plan to land the KDF changes first, then return to this.

ctz:
This is rebased and squashed now, but I'm dissatisfied with how the webpki algorithms get chosen currently. I plan to hoist that choice up into `CryptoProvider` too, which eliminate a chunk of new feature gates this PR is relying on.

cpu (approved):
The new builder adjustments to have the crypto provider supply the webpki signature algorithms looks good to me, nice work.

ci-bench/src/main.rs
```diff
@@ -210,31 +210,31 @@ fn all_benchmarks() -> anyhow::Result<Vec<Benchmark>> {
 static ALL_BENCHMARK_PARAMS: &[BenchmarkParams] = &[
     BenchmarkParams::new(
         KeyType::Rsa,
-        rustls::cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+        rustls::crypto::ring::cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
```
djc: I think the library changes for this make sense, but IMO all of the example code and tests should probably repeat the path to the suite constants and import them in a `use` statement?

rustls/src/crypto/mod.rs
```diff
@@ -43,6 +46,14 @@ pub trait CryptoProvider: Send + Sync + Debug + 'static {
 
     /// Return a safe set of supported key exchange groups to be used as the defaults.
     fn default_kx_groups(&self) -> &'static [&'static dyn SupportedKxGroup];
+
+    /// Decode and validate a private signing key from `key_der`.
+    ///
+    /// Return an error if the key type encoding is not supported, or if the key fails validation.
+    fn load_private_signing_key(
```
djc, line 58: Is the `_signing` part important here? Feels like we could leave that to documentation? (Alternatively, should we rename `pki_types::PrivateKeyDer<'_>` to `PrivateSigningKeyDer<'_>`?

rustls/src/crypto/ring/mod.rs
```diff
@@ -38,10 +45,17 @@ impl CryptoProvider for Ring {
...
-    /// Return all supported key exchange groups.
     fn default_kx_groups(&self) -> &'static [&'static dyn SupportedKxGroup] {
         ALL_KX_GROUPS
     }
+
+    fn load_private_signing_key(
+        &self,
+        key_der: PrivateKeyDer<'static>,
+    ) -> Result<Arc<dyn SigningKey>, Error> {
+        sign::any_supported_type(&key_der)
```
djc, line 60: Should we inline this? If we're exposing this, should we get rid of the `pub fn any_supported_type()` API?

(Should we hide `any_ecdsa_type()` and `any_eddsa_type()` functions from the public API? I forget why they are there in the first place.)

rustls/src/crypto/ring/mod.rs
```diff
@@ -38,10 +45,17 @@ impl CryptoProvider for Ring {
...
+            .map_err(|_| Error::General(String::from("invalid private key")))
```
djc, line 61: Maybe `.to_owned()` instead of `String::from()`?

rustls/src/client/handy.rs
```diff
@@ -182,14 +179,11 @@ impl client::ResolvesClientCert for FailResolveClientCert {
 pub(super) struct AlwaysResolvesClientCert(Arc<sign::CertifiedKey>);
 
 impl AlwaysResolvesClientCert {
-    #[cfg(feature = "ring")]
     pub(super) fn new(
+        priv_key: Arc<dyn sign::SigningKey>,
```
djc, line 183: Nit: I would probably spell out `priv_key` (into `private_key`), perhaps because I feel like the abbreviation isn't as widely used as it is for `pub_key`. Might just be a personal opinion, feel free to ignore.

rustls/src/server/builder.rs
```diff
@@ -15,6 +16,19 @@ use alloc::vec::Vec;
 use core::marker::PhantomData;
 
 impl ConfigBuilder<ServerConfig, WantsVerifier> {
+    /// Create builder to build up the `webpki` client certificate verifier configuration.
+    /// Client certificate authentication will be offered by the server, and client certificates
+    /// will be verified using the trust anchors found in the provided `roots`. If you
+    /// wish to disable client authentication use [ConfigBuilder<ServerConfig, WantsVerifier>::with_no_client_auth()] instead.
+    ///
+    /// For more information, see the [`webpki::ClientCertVerifierBuilder`] documentation.
+    pub fn build_webpki_verifier(
```
djc, line 25: This seems like a pretty idiosyncratic API. I think I'd prefer mirroring the `ServerConfig` and `ClientConfig` `builder()` and `builder_with_provider()` constructors.

rustls/src/crypto/mod.rs
```diff
@@ -54,6 +55,11 @@ pub trait CryptoProvider: Send + Sync + Debug + 'static {
         &self,
         key_der: PrivateKeyDer<'static>,
     ) -> Result<Arc<dyn SigningKey>, Error>;
+
+    /// Return the signature verification algorithms for use with webpki.
+    ///
+    /// These are used for certificate chain verification, and handshake signature verification.
+    fn certificate_signature_verification_algorithms(&self) -> WebPkiSupportedAlgorithms;
```
djc, line 66: That's one verbose-ass method name... How do you feel about ditching the `certificate_` prefix?

rustls/src/client/client_conn.rs
```diff
@@ -122,7 +122,8 @@ pub trait ResolvesClientCert: Send + Sync {
 /// (the rustls-native-certs crate is often used for this) may take on the order of a few hundred
 /// milliseconds.
 ///
-/// These must be created via the [`ClientConfig::builder()`] function.
+/// These must be created via the [`ClientConfig::builder()`] or [`ClientConfig::builder_with_provider()`]
```
djc, line 125: I think this doesn't need to be part of this commit?

rustls/src/server/builder.rs
ctz: Yes, it's not ideal. I'm trying to avoid a situation where one provider is used for the config, and another is used for the verifier. (As a concrete example, someone asks for a FIPS provider for the config, but neglects to do that for the verifier. The result would be something that isn't FIPS approved.)

ctz: I guess the `ClientCertVerifierBuilder::new` could take a `ConfigBuilder<...>`?

rustls/src/client/client_conn.rs
ctz: hoisted these drive-by docs changes into their own commit

rustls/src/crypto/mod.rs
ctz: i'll drop `_signing` but keep it in the docs. i don't foresee a need to handle non-signing private keys; that would imply we're doing some non-ephemeral DH or RSA encryption things, both of which I think are far outside what we should be concerned about.

rustls/src/crypto/ring/mod.rs
ctz: I'm minded to leave these as they are, though if we were to replace this bit of the API I think it would look a bit different.

`any_ecdsa_type()` and co. is there to allow people to avoid linking in the whole of EdDSA and RSA if they know at compile time they will only use ECDSA keys. The ECDSA version also plasters over *ring*'s API which requires a-priori knowledge of which curve a key is on.

rustls/src/server/builder.rs
djc: Hmm, by reference? Seems like it could be okay... Still pretty tangled but probably better than this.

skmcgrail:
>     2. aws-lc-rs is doing much more RSA key validation than _ring_. As an example, _ring_ doesn't do primality testing on the private primes, and there is justification in the code for that (because: side channels, which is fair enough).

Though we were doing a little to much extra work on one code-path 😬. The AWS-LC function we use for parsing the RSA keys was already calling the appropriate RSA validation function internally, and then our aws-lc-rs bidnings were subsequently then running the validation function again. This accounts a good portion of the performance differences, outside of additional checks that we are performing on the RSA CRT parameters. That is addressed by this PR in our repository: https://github.com/aws/aws-lc-rs/pull/268

For the `fips` feature code-path we still have to call a specific function that performs additional validation checks, unfortunately as part of this code-path that results in the same validation function being called yet again + the additional checks. I'm hoping we can eventually reorganize this API so we can better optimize around that, but that might be some ways off.

rustls/src/server/builder.rs
ctz: > As a concrete example, someone asks for a FIPS provider for the config, but neglects to do that for the verifier. The result would be something that isn't FIPS approved.

Having thought about this a bit more, I think this is insufficient. It makes less likely, but still possible, to have a FIPS-approved `CryptoProvider` under the `{Client,Server}Config`, but with a different one under the verifier. So I'm going to put this back how it was, and add a `CryptoProvider` parameter to the previous `WebPkiClientVerifier::builder` constructor.

I think I have some better ideas about how to detect that case, merged in with the need to have a "FIPS service indicator": basically that each `{Client,Server}Connection` can say "yes, I'm in FIPS mode" with a good degree of accuracy.

ctz: So this commit is dropped, and instead we have `WebPkiServerVerifier::builder_with_provider`

rustls/src/webpki/client_verifier.rs
```diff
@@ -247,13 +240,32 @@ pub struct WebPkiClientVerifier {
 
 impl WebPkiClientVerifier {
     /// Create builder to build up the `webpki` client certificate verifier configuration.
```
cpu, line 248: I think now that there are two `builder`-ish functions this headline could use a tweak to emphasize that this one gives you some defaults like the ring crypto provider.

rustls/src/webpki/client_verifier.rs
```diff
@@ -247,13 +240,32 @@ pub struct WebPkiClientVerifier {
...
+    /// Use [`Self::builder_with_provider`] if you wish to customise this.
+    ///
     /// For more information, see the [`ClientCertVerifierBuilder`] documentation.
+    #[cfg(feature = "ring")]
     pub fn builder(roots: Arc<RootCertStore>) -> ClientCertVerifierBuilder {
-        ClientCertVerifierBuilder::new(roots)
+        Self::builder_with_provider(roots, crate::crypto::ring::RING)
+    }
+
+    /// Create builder to build up the `webpki` client certificate verifier configuration.
```
cpu, line 263: If the other headline were changed this one could stay as it is and would probably be clear enough with the `_with_provider` name. Right now both fns have the same headline and I think that could be made nicer.

rustls/src/webpki/server_verifier.rs
```diff
@@ -143,23 +134,41 @@ pub struct WebPkiServerVerifier {
 #[allow(unreachable_pub)]
 impl WebPkiServerVerifier {
     /// Create builder to build up the `webpki` server certificate verifier configuration.
```
cpu, line 136: Same feedback here w.r.t. the two builder fns having matching headlines.

cpu (approved):

rustls/src/webpki/client_verifier.rs
ctz: Thanks, this is looking better I think -- 
image

cpu: Much better, thanks!

ctz:
(reworking the benchmarking support, as i've realised it no longer builds and isn't exercised in CI...)

ci-bench/src/main.rs
```diff
@@ -210,31 +210,31 @@ fn all_benchmarks() -> anyhow::Result<Vec<Benchmark>> {
 static ALL_BENCHMARK_PARAMS: &[BenchmarkParams] = &[
     BenchmarkParams::new(
         KeyType::Rsa,
-        rustls::cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+        cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
```
djc: For the purpose of this file and other test/example modules in our repository, why not import the cipher suite names directly at the top of the module? (As in `use rustls::crypto::ring::cipher_suite::{TLS13_AES_128_GCM_SHA256, ..};`.)

rustls/src/server/builder.rs
```diff
@@ -93,7 +95,11 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
         key_der: PrivateKeyDer<'static>,
         ocsp: Vec<u8>,
     ) -> Result<ServerConfig, Error> {
-        let resolver = handy::AlwaysResolvesChain::new_with_extras(cert_chain, &key_der, ocsp)?;
+        let private_key = self
+            .state
+            .provider
+            .load_private_key(key_der)?;
+        let resolver = handy::AlwaysResolvesChain::new_with_extras(private_key, cert_chain, ocsp)?;
```
djc, line 102: Should we change `AlwaysResolvesChain` to have a constructor that can take a `PrivateKeyDer` and `CryptoProvider` instead?

rustls/src/crypto/mod.rs
```diff
@@ -54,6 +55,11 @@ pub trait CryptoProvider: Send + Sync + Debug + 'static {
         &self,
         key_der: PrivateKeyDer<'static>,
     ) -> Result<Arc<dyn SigningKey>, Error>;
+
+    /// Return the signature verification algorithms for use with webpki.
+    ///
+    /// These are used for certificate chain verification, and handshake signature verification.
```
djc, line 65: I think we could ditch the "," in the second sentence here? Maybe introduce "both"?

rustls/src/webpki/client_verifier.rs
```diff
@@ -135,7 +139,7 @@ impl ClientCertVerifierBuilder {
         mut self,
         supported_algs: WebPkiSupportedAlgorithms,
     ) -> Self {
-        self.supported_algs = Some(supported_algs);
+        self.supported_algs = supported_algs;
```
djc, line 142: Should we just ditch this method and require the caller to build a custom `CryptoProvider`? Seems pretty niche.

rustls/src/webpki/client_verifier.rs
```diff
@@ -246,14 +239,37 @@ pub struct WebPkiClientVerifier {
...
-    /// Create builder to build up the `webpki` client certificate verifier configuration.
+    /// Create builder to build up the `webpki` client certificate verifier configuration using
+    /// the default [`CryptoProvider`].
+    ///
     /// Client certificate authentication will be offered by the server, and client certificates
     /// will be verified using the trust anchors found in the provided `roots`. If you
     /// wish to disable client authentication use [WebPkiClientVerifier::no_client_auth()] instead.
     ///
+    /// The cryptography used comes from the default [`CryptoProvider`]: [`crate::crypto::ring::RING`].
+    /// Use [`Self::builder_with_provider`] if you wish to customise this.
```
djc, line 256: "customise" -> "customize"?

rustls/src/webpki/client_verifier.rs
```diff
@@ -246,14 +239,37 @@ pub struct WebPkiClientVerifier {
 }
 
 impl WebPkiClientVerifier {
-    /// Create builder to build up the `webpki` client certificate verifier configuration.
+    /// Create builder to build up the `webpki` client certificate verifier configuration using
```
djc, line 248: "Create [a] builder"?

rustls/src/webpki/verify.rs
```diff
@@ -144,68 +142,6 @@ impl<'a> TryFrom<&'a CertificateDer<'a>> for ParsedCertificate<'a> {
     }
 }
 
-/// A `WebPkiSupportedAlgorithms` value that reflects webpki's capabilities when
```
djc, line 147: Would be nice to move this to its new place in a separate commit.

rustls/src/client/client_conn.rs
```diff
@@ -249,15 +250,15 @@ impl fmt::Debug for ClientConfig {
 impl ClientConfig {
     #[cfg(feature = "ring")]
     /// Create a builder to build up the client configuration with the default
-    /// [`CryptoProvider`].
+    /// [`CryptoProvider`]: [`crate::crypto::ring::RING`].
```
djc, line 253: Squash these into the originating commit(s)?

ci-bench/src/main.rs
```diff
@@ -210,31 +213,31 @@ fn all_benchmarks() -> anyhow::Result<Vec<Benchmark>> {
 static ALL_BENCHMARK_PARAMS: &[BenchmarkParams] = &[
     BenchmarkParams::new(
         KeyType::Rsa,
-        cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+        provider::cipher_suite::TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
```
djc, line 216: (This wouldn't be necessary if we had a module-level import?)

rustls/src/server/builder.rs
ctz: I think this is most general as-is (being a trivial implementation of the trait that just returns the constructor parameters). Though I think the constructor should become infallible now it does that. Will make that change.

rustls/src/client/client_conn.rs
ctz: The originating commit for these is already on main (048ff1074)

rustls/src/webpki/client_verifier.rs
ctz: Have done that.

rustls/src/webpki/client_verifier.rs
ctz: Reworded to "Create a builder for the `webpki` client ..."

ctz: Also fixing similar instances in other types of "Create builder to build up" (in a separate commit)

ci-bench/src/main.rs
ctz: There are instances of this elsewhere (arguably tests that are interested in multiple individual suites should just `use ...::cipher_suite::*;` and be done with it for brevity's sake.

But in this particular case, this has fallen away with refactoring the ci-bench code to benchmark ring and aws-lc-rs in one program.

rustls/src/server/builder.rs
ctz: > Will make that change.

Done and squashed into this commit.

rustls/src/webpki/verify.rs
ctz: Have done that.

ctz:
> (reworking the benchmarking support, as i've realised it no longer builds and isn't exercised in CI...)

This is done; I have maintained the old benchmark names for continuity. But now there are also names like `handshake_no_resume_aws_lc_rs_1.3_rsa_aes_client` which should be comparable to `handshake_no_resume_1.3_rsa_aes_client` (which uses *ring*).

aochagavia (approved):
The changes to `ci-bench` make sense to me. My only suggestion would be to include `ring` in the name of the benchmarks, for consistency (we don't keep a result history yet, so as far as I can see there are no downsides to renaming).

ctz:
Thanks -- have just done that. For posterity the results prior to that change vs main are https://github.com/rustls/rustls/actions/runs/6811702313?pr=1414

PR #1415, opened 2023-08-24
Apply clippy suggestions from Rust 1.72

djc:


rustls/src/msgs/message_test.rs
```diff
@@ -81,12 +81,6 @@ fn alert_is_not_handshake() {
     assert!(!m.is_handshake_type(HandshakeType::ClientHello));
 }
 
-#[test]
-fn alert_is_not_opaque() {
-    let m = Message::build_alert(AlertLevel::Fatal, AlertDescription::DecodeError);
-    assert!(Message::try_from(m).is_ok());
```
djc, line 87: `try_from()` does an identity conversion here. I don't think this is testing anything meaningful at this point?

rustls/src/server/tls13.rs
```diff
@@ -604,7 +604,8 @@ mod client_hello {
         common.send_msg(m, false);
     }
 
-    #[allow(clippy::needless_pass_by_ref_mut)] // cx only mutated if cfg(feature = "quic")
+    #[allow(unknown_lints)]
+    #[allow(clippy::needless_pass_by_ref_mut)]
```
cpu, line 608: Concretely, this seems to work for me:

```suggestion
    #[cfg_attr(not(feature = "quic"), allow(clippy::needless_pass_by_ref_mut))]
```

rustls/src/server/tls13.rs
```diff
@@ -604,7 +604,8 @@ mod client_hello {
         common.send_msg(m, false);
     }
 
-    #[allow(clippy::needless_pass_by_ref_mut)] // cx only mutated if cfg(feature = "quic")
+    #[allow(unknown_lints)]
```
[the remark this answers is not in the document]
djc: The problem with CI commands that I don't like having to remember the complete incantation when I'm running clippy locally. I'd prefer turning this into a `#![allow(unknown_lints]` in the crate root, similar to what we do for other lint tuning.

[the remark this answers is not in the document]
cpu: That makes sense. Running `cargo clippy` locally the warnings aren't denied but default but it is noise that could confuse someone new to the project, so I think you've convinced me it's a good idea to fix at the site of the issue.

cpu (approved):

rustls/src/server/tls13.rs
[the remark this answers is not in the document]
djc: I think it's there because it allows the unknown lints on stable, but not on nightly, which is a bit more robust in the sense that we'll notice when a lint is actually unknown. But on the other hand, if a lint is ever renamed on nightly so it doesn't exist there anymore but still exists on stable, it will also break... So yes, I think the local solution is probably better, because at least that way we get to avoid that we don't notice unknown lints at all.

.github/workflows/build.yml
```diff
@@ -297,7 +297,7 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           components: clippy
-      - run: cargo clippy --package rustls --all-features
-      - run: cargo clippy --package rustls --no-default-features -- --deny warnings
-      - run: cargo clippy --manifest-path=connect-tests/Cargo.toml --all-features -- --deny warnings
-      - run: cargo clippy --manifest-path=fuzz/Cargo.toml --all-features -- --deny warnings
+      - run: cargo clippy --package rustls --all-features --all-targets
```
ctz, line 300: hm, why is this first one missing `--deny warnings`? (it was before too, so not a change being made in this PR)

djc: That difference seems to have always been there:

https://github.com/rustls/rustls/commit/66abf71bdb4efcf395e27beb2282449a6895fe56

I think my rationale would have been that we only need to `--deny warnings` on stable, failing CI for clippy issues that are new on nightly would be too painful. But then that wasn't followed when @cpu added the other nightly clippy runs.

cpu: I think just simple oversight on my part, not anything intentional.

cpu (approved):

djc:
Going to just merge this, I don't feel strongly either way about whether we should deny warnings from nightly clippy although the reasoning (that nightly lints can change more often which can be annoying, especially for PRs from external contributors).

Issue #1362, opened 2023-07-18
Allow for more flexible buffer management


ctz:
# Background
Currently a single rustls connection owns several buffers:

- incoming TLS data, that has been read with `read_tls()` but not yet processed by `process_new_packets()`. This is embedded into `msgs::MessageDeframer` and is a contiguous buffer large enough to store the largest single TLS message that can appear on the wire (~18KB). At the moment we are following a strategy of keeping this buffer small, at the potential expense of extra allocator traffic and system calls. This buffer needs to be contiguous because we borrow it during message parsing.
- outgoing TLS data, that has been produced by `process_new_packets()`, but not yet written by `write_tls()`. This is `CommonState::sendable_tls`.
- incoming application data, that has been decrypted by `process_new_packets()`, but not yet read via `reader()`. This is `CommonState::received_plaintext`.
- outgoing application data, that has been input with `writer()` but is not yet sendable because the connection is not yet set up. This is `CommonState::sendable_plaintext`.  It  is automatically sent at completion of the TLS handshake, and not used thereafter -- use of `writer()` once the connection is set up encrypts the data immediately and puts the result in `CommonState::sendable_tls`.

This arrangement allows for the rustls IO API to be (relatively!) simple, and separates IO and processing concerns. But it is inefficient because:

- some of these buffers are short-lived: outgoing TLS is destined for the TCP send buffer, incoming application data is destined for the application.
  - if the TCP send buffer is full, it may be necessary to keep outgoing TLS longer term. But buffering it in that case risks doing an end-run around TCP flow control. There is an attempt in rustls to quench the read-side if there is pending outgoing data (see `wants_read()`).
  - for an application that deals with many connections simultaneously, these buffers could be shared (when not in use!) between many connections.
- outgoing application data buffering is just a convenience for applications, and one they can arrange themselves if we choose to drop it as a feature in rustls.

# Things to note
- The incoming TLS data buffer is longer-term, because TLS is a packetised protocol and we need a whole packet before we can process it. These packets are up to ~18KB, and their receipt can be split across any TCP packet boundaries. Any application that hopes to amortise buffer usage over many simultaneous connections needs to be sized such that all the connections can concurrently be in the state of receiving the largest possible message.
- Sending or receiving a TLS message can mean sending (a limited number of) additional control messages. Messages that are sent must retain their order (ie, it would not work to have a data path that deals only with application data and another that only deals with control messages, because the peer must see receive them in the original order).
[...]

PR #1420, opened 2023-08-25
[RFC] Caller-side managed buffers

japaric:
# Summary

This RFC proposes adding a new, lower-level `{Client,Server}Connection` API where TLS and plain-text buffers are managed by the caller / end-user.
The design goals of this new API are to reduce / minimize memcpys, be no-std compatible and allow `async` certificate verification.

[Rendered version](https://github.com/japaric/rustls/blob/rfc-caller-side-buffers/rfcs/000-caller-side-managed-buffers.md)

This RFC addresses issue #1362 

Also check out [the mock implementation of the proposed API](https://github.com/japaric/rustls/tree/rfc-caller-side-buffers/sketch#sketch) in the PR branch

> allow `async` certificate verification.

~~This aspect still needs more work but I think the main part that addresses #1362 is ready for comments!~~ [this section is now complete](https://github.com/rustls/rustls/pull/1420#issuecomment-1702957996)


japaric:
> allow async certificate verification.

I have started working on revising the RFC to meet this design goal. the work is visible in this [WIP branch](https://github.com/japaric/rustls/blob/rfc-caller-side-buffers-wip/sketch/src/lib.rs#L511). when I'm done there, I'll push the changes into this branch / PR

japaric:
The RFC has been updated:
- the main proposal now uses the enum alternative to a `Status` struct that only contains boolean flags
- the 'async certificate verification' section has been completed and now it's already for feedback / comments

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+
+impl LlConnectionCommon {
+    /// Handles TLS records
+    ///
+    /// This method must always be called and its returned `Status` checked prior to
+    /// calling either `encrypt_outgoing` or `decrypt_incoming`
+    pub fn handle_tls_records<B>(
+        &mut self,
+        incoming_tls: &mut IncomingTls<B>,
+        outgoing_tls: &mut Vec<u8>,
```
DemiMarie, line 71: `&mut Vec` is not great, as it relies on `alloc`.  Some embedded systems might not have `alloc`, or may only have a subset of `alloc` that only includes fallable APIs.

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+    ) -> Result<Status>
+    where
+        B: AsRef<[u8]> + AsMut<[u8]>,
+    {
+       // ..
+    }
+
+    /// Encrypts `app_data` into the `outgoing_tls` buffer
+    pub fn encrypt_outgoing(&self, app_data: &[u8], outgoing_tls: &mut Vec<u8>) { /* .. */ }
```
DemiMarie, line 80: Ditto.

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+    ///
+    /// The returned iterator yields decrypted application data, one item per TLS record
+    ///
+    /// The iterator will return an `Err`or in these cases
+    /// - a record that's not application data was found, e.g. a handshake record
+    /// - a fatal Alert record was found
+    pub fn decrypt_incoming<'a, B>(
+        &self,
+        incoming_tls: &'a mut IncomingTls<B>,
+    ) -> impl Iterator<Item = Result<&'a [u8]>> {
```
DemiMarie, line 92: Should this be a GAT-based `LendingIterator`?

rfcs/000-caller-side-managed-buffers.md
djc: Yeah, I think we should be relying on `&mut [u8]` for these APIs. If `Status::state` is going to describe which actions are allowed/possible, then it feels like `Status` should hold references to the `LlConnectionCommon`, the `incoming_tls` and `outgoing_tls` and expose `encrypt_outgoing()` and `decrypt_incoming()` on a type exposed via the `State` variant.

I think this would be simpler if we only offered a method like `pub fn process_new_packets(&mut self, incoming_tls: &mut [u8], send_plaintext: &[u8], outgoing_tls: &mut [u8])` which would then yield, among other things, an iterator over `&[u8]` blocks of plaintext data pointing into `incoming_tls`.

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+In the case of `outgoing_tls`, the buffer can be resized while it contains data but
+care must be taken not to discard existing data if truncating it.
+
+# Implementation details
+
+This section contains implementation details about the new API.
+
+## Partially discarding app data records
+
+To implement an equivalent of `rustls::Stream` API using the `LlConnection` types and, in particular, to implement the `io::Read` trait,
```
djc, line 373: Let's place an equivalent of `rustls::Stream` out of scope for now?

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+This section describes some scenarios where the API is misused and what the outcomes are.
+
+- `Status.state` is `NeedsMoreTlsData`, no data or an incomplete TLS record is appended to `incoming_tls` and `process_tls_records` is called
+  - the newly returned `Status.state` will also be in the `NeedsMoreTlsData` state. the handshake won't make progress until sufficient data is appended to `incoming_tls`
+
+Also see the 'unresolved questions' section.
+
+# Supporting async certificate verification
+
+In the API described thus far the certificate verification still happens implicitly in `process_tls_records`.
```
djc, line 468: The idea of exposing certificate verification input via `Status`/`State` makes sense to me.

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+
+```rust
+enum State<'a> {
+    ReceivedEarlyData(&'a [u8]),
+}
+```
+
+But this alternative has two issues:
+`State`, and thus `Status` as well, would have to gain a lifetime that ties it to the `incoming_tls` buffer.
+Given how Rust works, even if `State` was not in the `ReceivedEarlyData` variant, the `status.state` field would still "freeze" `incoming_tls` meaning that it can't be mutated until `status.state` goes out of scope / is dropped.
```
djc, line 758: Why is this a problem?

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+
+The other issues is that if `process_tls_records` automatically decrypts early data, the end-user no longer has the option to discard the early data without decrypting it first (see snippet below) to save some CPU work.
```
djc, line 760: It seems like this could be offered as a method.

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+    }
+}
+```
+
+Given that `certificates` freezes `incoming_tls`, the end-user will likely have to copy the data into an owned version (`Certificate`) so
+they can stash the certificates for use in a later iteration of the event loop or to send them to a separate async task.
+In the end, the allocation is not eliminated but simply made more explicit, as the end user will have to perform it.
+
+This main proposal's criteria for whether to include owned data into one of `State` variants is to re-use existing owned types and avoid creating new "ref" versions of them.
+For instance, the internal `Codec` trait which creates instances the `Certificate` type from a `Reader` slice can only create owned types, not slices.
```
djc, line 804: I think we can easily stuff a `Cow<'_, [u8]>` inside `Certificate` to enable both options.

rfcs/000-caller-side-managed-buffers.md
```diff
@@ -0,0 +1,833 @@
...
+
+# Unresolved questions
+
+what should the behavior be in these scenarios
+
+- `Status.state` is `State::MustTransmitTlsData`, `outgoing_tls` is not fully transmitted and `process_tls_records` is called
+  - should `wants_write` "latch" to the `true` value until `outgoing_tls` is observed as being empty?
+- `Capabilities.may_send_app_data` returns `false` and `encrypt_outgoing` is called
```
djc, line 812: As mentioned previously, can we statically make this impossible?

djc:
BTW, one goal to keep in mind is code reuse across the high-level and low-level API. So ideally the high-level API can be implemented on top of the low-level API to avoid duplicating logic.

rfcs/000-caller-side-managed-buffers.md
japaric: @djc thanks for the feedback!

> I think this would be simpler if we only offered a method like

I made a new sketch (`sketch2`) using this idea and I like the direction. I suggest you check the compile-pass test first: https://github.com/japaric/rustls/blob/d21cb4f7e2e79a64e447e0e147cce5cb13148347/sketch2/src/lib.rs#L174

I pushed the sketch into this branch / PR so you can also see it in this PR webpage.

what I ended up with is:
- `LlConnectionCommon` only has one method
- `process_new_packets` / `process_tls_records` only takes a single argument: `incoming_tls`
- `outgoing_tls` and `send_plaintext` only appear in the methods of some `State` variants. I personally like this better over storing `outgoing_tls` in `State` as that increases its lifetime parameters. I think it would also make harder to resize `outgoing_tls` while handling a variant that requires encrypting some data
- also changed `outgoing_tls` from a `Vec` into a `&mut [u8]`. this means that encrypting can now return an `Err`or due to insufficient buffer size. there's an example of resizing `outgoing_tls` buffer in the linked compile-pass test
- changed the `AppDataReceived` state to yield a single record / slice instead of an iterator. I'm still a bit undecided over which is better (slice vs iterator) but having the `process_` method handle one record on every invocation feels consistent. (I think one record per invocation fits better with the `discard` idea mentioned below)

in the signature of the `process_new_packets` method you posted, `incoming_tls` has type `&mut [u8]`. would you prefer that over providing the `IncomingTls` abstraction as per the current RFC text? that would push the task of removing TLS records from the front of the buffer to the user. I think that could be more flexible but also make the API a bit more annoying to use.

in that case, I think `process_` can return a `struct Status<_> { discard: usize, state: State<_> }` value where `discard` indicates how many bytes from the front of `incoming_tls` must be discarded before the next call to `process_`. I *think* `&mut [u8]` + `discard` would also work with libstd's `read_buf` API (`BorrowedBuf`) without needing to have those libstd types in the API which would keep the API the same in both std and no-std mode and that'd be a plus.

rfcs/000-caller-side-managed-buffers.md
japaric: to clarify this is not proposing neither adding a new `Stream` on top of `LlConnectionCommon` nor modifying the existing implementation. it was more along the lines of me attempting to implement `Stream` on top of `LlConnectionCommon` to see if the API worked well enough to implement `Stream`.

I now think it's not absolutely necessary to support partially discarding records at the `LlConnection` layer to implement `Stream`. any app-data record fragment that doesn't fit in `Read::read`'s buffer could be buffered into `Stream` until the next `Read::read`'s call. as `Stream` is more of a convenience API and not a "max performance" API I think that may be an ok trade-off to keep the `LlConnection` API simpler.

rfcs/000-caller-side-managed-buffers.md
japaric: I was a bit concerned with end users running into borrow checker problems in the case of `State` borrowing the TLS buffers but after doing `sketch2` I think my concerns have been somewhat alleviated. I'll do some further testing with a mocked async IO API to confirm that the borrow checker is still happy with borrows across suspension (`.await`) points.

rfcs/000-caller-side-managed-buffers.md
japaric: yes, `sketch2` achieves this. `may_send_app_data` has been turned into a `State` variant that yields a newtype over `LlConnectionCommon` and provides the `encrypt_outgoing` method

rfcs/000-caller-side-managed-buffers.md
japaric: that should work. then the certificate related variants in `sketch2`'s `State` can provide non-allocating decrypting methods

japaric:
the [sketch2](https://github.com/japaric/rustls/blob/rfc-caller-side-buffers/sketch2/src/lib.rs#L864) crate has been updated and now includes the certification validation API. two mocked examples / tests are included: one uses mocked async IO; the other one uses libstd's `read_buf` API, also mocked

the sketch2 implements [this idea](https://github.com/rustls/rustls/pull/1420/files#r1319629980) of making certain operations possible only when certain conditions are met thus making the API statically impossible to misuse. 

I also went ahead and experimented with turning `incoming_tls` into a regular slice type. this pushes the task of removing already processed records from said buffer to the end user, which makes the API a bit more cumbersome to use. the inconvenience is traded off with increased flexibility as the API can now be used with seemingly any buffer type: `Vec` or `BorrowedBuf` is what I tested but I expected that a custom buffer type like a ring / circular buffer should also work.

rfcs/000-caller-side-managed-buffers.md
djc: > * `outgoing_tls` and `send_plaintext` only appear in the methods of some `State` variants. I personally like this better over storing `outgoing_tls` in `State` as that increases its lifetime parameters. I think it would also make harder to resize `outgoing_tls` while handling a variant that requires encrypting some data

This seems potentially problematic: what if I haven't received any TLS but I would like to send some plaintext now?

> * also changed `outgoing_tls` from a `Vec` into a `&mut [u8]`. this means that encrypting can now return an `Err`or due to insufficient buffer size. there's an example of resizing `outgoing_tls` buffer in the linked compile-pass test

Note there is some tricky stuff here: handshake states need to write their outgoing TLS **before** manipulating the connection state so we can give the caller a chance to try again with a bigger buffer, without ending up with connection state that's inconsistent with the messages processed.

> * changed the `AppDataReceived` state to yield a single record / slice instead of an iterator. I'm still a bit undecided over which is better (slice vs iterator) but having the `process_` method handle one record on every invocation feels consistent. (I think one record per invocation fits better with the `discard` idea mentioned below)

It feels like performance would be better if we can handle multiple records per invocation? Not sure though, maybe one record per invocation would be fine.

djc:
> I also went ahead and experimented with turning `incoming_tls` into a regular slice type. this pushes the task of removing already processed records from said buffer to the end user, which makes the API a bit more cumbersome to use. the inconvenience is traded off with increased flexibility as the API can now be used with seemingly any buffer type: `Vec` or `BorrowedBuf` is what I tested but I expected that a custom buffer type like a ring / circular buffer should also work.

Another important point to me at least is that the library cannot issue allocations (via resizing the `&mut Vec` internally) on the data path behind the caller's back. I definitely think we'll want to use `BorrowedBuf` for this in the future, after it stabilizes.

rfcs/000-caller-side-managed-buffers.md
djc: > I think the scenario were that would happen is [early (0-RTT) data](https://datatracker.ietf.org/doc/html/rfc8446#section-2.3). the client can send early data encrypted with a shared key before receiving a record from the server. In other cases, both sides need to agree on a key before application data can be encrypted, as far as I've seen in TLS logs.

I meant just after the handshake.

djc: > I think this is covered. whenever a handshake record needs to be encrypted `process_tls_record` returns the `MustEncryptTlsData`. this variant has that fallible `encrypt` method; only when that method succeeds the state machine advances. if `encrypt` fails due to insufficient space, `encrypt` can be called again without issues. this is (mock) tested in `sketch2`: in one of the tests, `outgoing_tls` starts as a `Vec` with 0-capacity so the `encrypt` failure path is hit during the execution of the test.

Hmm, okay, I think that makes sense.

japaric:
I've updated `sketch2` and turned `AppDataAvailable` into an iterator. that seems to work fine. I also re-implemented `Stream` on top of `LlClientConnection` to see if that would work out. even without support for partially discarding TLS records; that's doable with some extra buffering.

what I realized while implementing `Stream` is that we may need another version of `AppDataAvailable`. the current version is meant to represent app-data records during the handshake; that variant does not provide an `encrypt` method because during handshake that's represented with a different enum: `MayEncryptAppData`. after handshake, however, it's possible to encrypt app-data into `outgoing_tls` while decrypting app-data records out of `incoming_tls` -- even while iterating; the buffers are separate so the borrow checker won't complain about encrypting in the middle of an iteration loop, though one may need to use `while let` instead of a `for` loop.

OR, if it's always possible to `encrypt` when encrypted app data has arrived then we can add `encrypt` to `AppDataAvailable` variant / type and no need to add another variant to `State` (and it may even be possible to remove `MayEncryptAppData` variant?). the two sequence diagrams I have, one of TLS 1.2 and one of TLS 1.3 (this one is in the README of the `sketch` directory), do indicate that encryption is possible at that point of the handshake / connection.

rfcs/000-caller-side-managed-buffers.md
japaric: > I meant just after the handshake.

once the handshake is done, `process_tls_records` returns the state variant `TrafficTransit` and that has an `encrypt` method so encryption is possible before post-handshake app-data records are received. this should be clearer once the RFC text is updated to use the `sketch2` API as the main proposal, which I'll get to shortly.

japaric:
the RFC text has been update to use "sketch2" as the main proposal. the `sketch` project has been removed and `sketch2` has been renamed to `sketch`. 

this is ready for a new review / round of comments!

ctz:
Going to close this. We have some follow-on work in other issues, but the initial version of this was delivered in 0.23.

c410-f3r:
Is there an issue to track the remaining work of this feature?

PR #1427, opened 2023-08-29
Correct `cargo test` for all feature combinations

ctz:
It seems ideal if `cargo build` and `cargo test` work without errors for any combination of crate features.

`cargo hack --feature-powerset --keep-going test` shows a few existing issues that this commit addresses:

- `bogo_shim`: requires the `tls12` feature
- `test_secret_extraction_enabled` and `test_secret_extraction_disabled_or_too_early` require the `tls12` feature.

nb. probably not plausible to check this in CI, because it takes 15 minutes of CPU time at the moment and increases superlinearly with number of features.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -41,7 +41,7 @@ base64 = "0.21"
 [[example]]
 name = "bogo_shim"
 path = "examples/internal/bogo_shim.rs"
-required-features = ["dangerous_configuration", "quic"]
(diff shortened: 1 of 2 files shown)


djc (approved):

cpu:
> cargo hack --feature-powerset --keep-going test shows a few existing issues that this commit addresses

Is this something we could run in CI directly?

PR #1429, opened 2023-08-29
v0.20.10 release prep

cpu:
## Description

This branch prepares a 0.20.10 release, switching to the Rustls project maintained webpki fork. This resolves RUSTSEC-2023-0052 by switching to a our fork, with the patch for RUSTSEC-2023-0053.

This may not be strictly semver compatible: Rustls 0.20.x exposes Webpki types in its API.

- [x] `cargo update`.
- [x] `cargo outdated`.
- [x] `cargo test --all-features`.
- [x] Update `version` in `Cargo.toml`.
- [x] `cargo publish --dry-run -p rustls`
- [x] PR desc updated with release headlines

### Proposed release notes headlines:

 **Note**: This update is not strictly semver compatible, since [`RootCertStore::add`](https://docs.rs/rustls/0.20.9/rustls/struct.RootCertStore.html#method.add) returns `webpki::Error`.

* Switches `webpki` to `rustls/webpki`. This resolves RUSTSEC-2023-0052: When the `webpki` crate is given a pathological certificate chain to validate, it will spend CPU time exponential with the number of candidate certificates at each step of path building. Both TLS clients and TLS servers that accept client certificate are affected. The updated `rustls/webpki` crate now gives each path building operation a budget of 100 signature verifications.

### deps: update webpki v0.22 -> rustls-webpk v0.100.2

This resolves RUSTSEC-2023-0052 by switching to a Rustls project maintained fork, with the patch for RUSTSEC-2023-0053.

### Cargo: bump version 0.20.9 -> 0.20.10.
Prepares for 0.20.10 release.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -10,7 +10,7 @@ resolver = "2"
 cargo-fuzz = true
 
 [dependencies]
-webpki = "0.22"
+webpki = { package = "rustls-webpki", version = "0.100.2", features = ["alloc", "std"] }
 
 [dependencies.rustls]
 path = "../rustls"
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -20,7 +20,7 @@ rustversion = { version = "1.0.6", optional = true }
 log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 sct = "0.7.0"
-webpki = { version = "0.22.0", features = ["alloc", "std"] }
+webpki = { package = "rustls-webpki", version = "0.100.2", features = ["alloc", "std"] }
 
 [features]
 default = ["logging", "tls12"]
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls"
-version = "0.20.9"
+version = "0.20.10"
 edition = "2018"
 rust-version = "1.60"
 license = "Apache-2.0/ISC/MIT"
--- a/rustls/src/sign.rs
+++ b/rustls/src/sign.rs
@@ -108,7 +108,7 @@ impl CertifiedKey {
             // that the certificate is valid for, if the certificate is
             // valid.
             if end_entity_cert
-                .verify_is_valid_for_dns_name(name)
+                .verify_is_valid_for_subject_name(webpki::SubjectNameRef::DnsName(name))
                 .is_err()
             {
                 return Err(Error::General(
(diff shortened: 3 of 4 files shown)


cpu:
Opened as a draft pending discussion about the semver impact.

djc:
Probably good to trace down exactly what we changed about the public API for the release notes? I think there was a PR that was tagged as semver-incompatible.

cpu:
Would it be sufficient to point to the upstream rustls-webpki release notes? :thinking:

djc:
I meant, where did the webpki API leak into the rustls API.

ctz:
AIUI the semver issue was that `rustls::RootCertStore::add` used `webpki::Error` for its error type.

cpu:
There are a few places `webpki` makes an appearance in the `internal` package, probably not very consequential? 

* [`ServerNamePayload`](https://docs.rs/rustls/0.20.9/rustls/internal/msgs/handshake/enum.ServerNamePayload.html) has a `webpki::DnsName` field in the `Hostname` variant and `new_hostname` constructor.
* [`ClientExtension`](https://docs.rs/rustls/0.20.9/rustls/internal/msgs/handshake/enum.ClientExtension.html)'s [`make_sni`](https://docs.rs/rustls/0.20.9/rustls/internal/msgs/handshake/enum.ClientExtension.html#method.make_sni) takes a `webpki::DnsNameRef`
* [`ServerSessionValue`](https://docs.rs/rustls/0.20.9/rustls/internal/msgs/persist/struct.ServerSessionValue.html)'s `new` fn takes a `Option<&webpki::DnsName>` arg.

> AIUI the semver issue was that rustls::RootCertStore::add used webpki::Error for its error type.

Yup, this seems like the most meaningful occurrence: [`RootCertStore::add`](https://docs.rs/rustls/0.20.9/rustls/struct.RootCertStore.html#method.add) returns `webpki::Error`.

I will update the proposed release notes to mention the `RootCertStore` instance.

djc:
Yeah, `internal` API doesn't count.

cpu:
I think this is ready for review. I've updated the proposed release notes.

cpu:
We're collaborating on a fix to the upstream `webpki` crate, so i'm going to go ahead and close this out.

cpu:
[`briansmith/webpki` v0.22.1](https://crates.io/crates/webpki/0.22.1) was released with the fix for RUSTSEC-2023-0052. 

Anyone that stumbles on this pull request should know that there's no need for a Rustls v0.20.10 release anymore. The webpki 0.22.1 dep. maintains semver compatibility with the version used by Rustls v0.20.x and can be picked up in your `Cargo.lock` without needing a Rustls release.

PR #1430, opened 2023-08-30
Add verify_client_cert_signed_by_trust_anchor

iamwwc:
fix https://github.com/rustls/rustls/pull/1134

### DIFF
--- a/rustls/src/webpki/verify.rs
+++ b/rustls/src/webpki/verify.rs
@@ -68,6 +68,8 @@ pub fn verify_server_cert_signed_by_trust_anchor(
 
 /// The wrapper of `webpki::RevocationOptions`.
 /// Call `Into::into` on `RevocationOptions` to get `RustlsRevocationOptions`
+#[allow(dead_code)]
+#[cfg_attr(not(feature = "dangerous_configuration"), allow(unreachable_pub))]
 pub struct RustlsRevocationOptions<'a>(RevocationOptions<'a>);
 
 impl<'a> From<RevocationOptions<'a>> for RustlsRevocationOptions<'a> {
--- a/rustls/src/webpki/verify.rs
+++ b/rustls/src/webpki/verify.rs
@@ -1,5 +1,6 @@
 use alloc::sync::Arc;
 use std::time::SystemTime;
+use webpki::RevocationOptions;
 
 use super::anchors::{OwnedTrustAnchor, RootCertStore};
 use super::client_verifier_builder::ClientCertVerifierBuilder;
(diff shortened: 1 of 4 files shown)


iamwwc:
If this PR get merged, please release a new patch version quickly, thank you

rustls/src/webpki/verify.rs
```diff
@@ -65,6 +66,38 @@ pub fn verify_server_cert_signed_by_trust_anchor(
...
+    let webpki_now = webpki::Time::try_from(now).map_err(|_| Error::FailedToGetCurrentTime)?;
+
+    cert.0
+        .verify_for_usage(
+            SUPPORTED_SIG_ALGS,
+            &trust_roots,
+            &chain,
+            webpki_now,
+            webpki::KeyUsage::client_auth(),
+            revocation, // no CRLs
```
djc, line 95: The `no CRLs` comment doesn't make sense in the client context, where a `Option<RevocationOptions<'_>>` is actually passed in.

(Is `RevocationOptions<'_>` part of the public API?

djc (commented):
LGTM. Once this is merged, can you do a backport to the rel-0.21 branch?

iamwwc:
> LGTM. Once this is merged, can you do a backport to the rel-0.21 branch?

I can use cargo `rustls={git=xxx}` temporarily, when the patch version will release?

rustls/src/webpki/verify.rs
ctz: `RevocationOptions` is part of the webpki public API

iamwwc: I prefer so, If user don't want to check CRLs, they can pass None

djc: We don't want to have webpki public API in the rustls public API, so we'll need a different solution here.

iamwwc: `WebPkiClientVerifier#new` already rely on `webpki::OwnedCertRevocationList`

cpu (commented):
> iamwwc added 3 commits 5 hours ago

Could you rebase these three commits into one? I don't think the comment update and chore commit are valuable on their own.

iamwwc:
I will rebase into one commit when other revision advice is done

iamwwc:
I haveadd the `RustlsRevocationOptions` to avoid leak the `webpki::RevocationOptions` to public api

djc:
Your `RustlsRevocationOptions` has a `From<webpki::RevocationOptions>` impl, which still means the `webpki::RevocationOptions` type becomes part of the public rustls API.

iamwwc:
any advice? I don't know how to solve it gracefully

djc:
There's no easy fix. One option is to push down the transitive set of required types down into the new rustls-pki-types crate, but we can't backport that to 0.21.

iamwwc:
> LGTM. Once this is merged, can you do a backport to the rel-0.21 branch?

I try use main branch code via `rustls={git=xxx}` in my private repo. I found latest type `ClientConfig` changed to `ClientConfig<C>`, but latest `tokio-rustls#TlsConnector#from` does't support `ClientConfig<C>`. How I can do that?

Yes. Once this PR get merged, I will cherry-pick commit and make a new PR to `rel-0.21` branch

djc:
Yeah, there is no tokio-rustls branch to match current rustls main. See #1435 for release planning.

ctz:
I think people wanting to reuse `WebPkiClientVerifier` should wrap it. eg,

```
struct MyVerifier {
  parent: rustls::server::WebPkiClientVerifier,
}

impl rustls::server::ClientCertVerifier for MyVerifier {
  fn verify_client_cert(&self, ...) -> ... {
    let assertion = self.parent.verify_client_cert(...)?;
    // extra validations as desired here
    Ok(assertion)
  }
}
```

The `verify_client_cert_signed_by_trust_anchor` introduced in this PR is the entirety of `WebPkiClientVerifier::verify_client_cert`, the only difference is avoiding parsing the certificate twice, but that is extremely fast and zero-copy.

iamwwc:
> I think people wanting to reuse `WebPkiClientVerifier` should wrap it. eg,
> 
> ```
> struct MyVerifier {
>   parent: rustls::server::WebPkiClientVerifier,
> }
> 
> impl rustls::server::ClientCertVerifier for MyVerifier {
>   fn verify_client_cert(&self, ...) -> ... {
>     let assertion = self.parent.verify_client_cert(...)?;
>     // extra validations as desired here
>     Ok(assertion)
>   }
> }
> ```
> 
> The `verify_client_cert_signed_by_trust_anchor` introduced in this PR is the entirety of `WebPkiClientVerifier::verify_client_cert`, the only difference is avoiding parsing the certificate twice, but that is extremely fast and zero-copy.

Yes, This is a way to solve it. but It's a hack

I currently use another hack to solve it.  But I think we should have a suitable way to solve it

djc:
Why is this a hack?

cpu:
Flipping this to draft status since it seems there's outstanding work to be done.

PR #1431, opened 2023-08-30
Hook up icount benchmarks to CI

aochagavia:
This PR adds a GitHub actions job that does the following:

- Benchmark the PR's target branch and store the results in a temp file.
- Benchmark the PR itself and store the results in a temp file.
- Compare the two runs and set the job summary to display the pretty-printed results.
- Let the job fail if there is a noteworthy difference in instruction counts, hinting you should look at the job summary to see the details.

You can already see it in action by looking at the new job triggered by this very PR 😉. It is failing, because the icounts have increased a lot (I discovered we need to run the resumed handshakes about 30x to filter out noise, instead of 3x).

Relevant context:

- It's not possible to post the benchmark results as a comment to the PR directly from the GitHub Actions job. This happens because the job runner has read-only access to the issues when the PR has been created from a fork (see the [docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)). Codecov manages to do this by running as a separate application, with its own permissions.
- I'm using the term "noteworthy" instead of "significant" to avoid giving the impression we are doing anything fancy with statistics here.

### DIFF
--- a/.github/workflows/icount-bench.yml
+++ b/.github/workflows/icount-bench.yml
@@ -1,0 +1,34 @@
+name: icount bench
+on: [pull_request]
+
+jobs:
+  icount-benchmarks:
+    name: Run icount benchmarks
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install valgrind
+        run: sudo apt install -y valgrind
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Checkout ${{ github.base_ref }}
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
+          persist-credentials: false
+
+      - name: Run icount benchmarks for ${{ github.base_ref }}
+        run: cd ci-bench && cargo run --release -- run-all > ${{ runner.temp }}/base.csv
+
+      - name: Checkout PR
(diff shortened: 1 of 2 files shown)


djc (approved):

.github/workflows/icount-bench.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  icount-benchmarks:
+    name: Run icount benchmarks
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install valgrind
+        run: sudo apt install -y valgrind
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@nightly
```
cpu, line 13: The comment here says stable, but it's using nightly. Could one of the two be updated so they're in-sync?

cpu (approved):
Looks good, just one nit to flag.

.github/workflows/icount-bench.yml
aochagavia: Thanks for that, let's use stable so we have one less variable to account for noise

cpu:
I think we can go ahead and merge this and iterate if there's any additional feedback.

aochagavia:
Thanks for the reviews!

PR #1433, opened 2023-08-30
ci: rework connect tests to daily tests, add cargo hack.

cpu:
This branch reworks the existing "connect tests" CI job that runs once a day to be "daily tests". This better matches how we've been using it for things beyond the connect test suite (e.g. smoke test builds of example binaries). 

Additionally it:

* Adds `name` fields to each workflow `step` for better readability.
* Adds a step for building the server acceptor example.
* Adds [cargo hack](https://github.com/taiki-e/cargo-hack) to test the crate feature powerset, to catch problems like https://github.com/rustls/rustls/pull/1427 earlier than we would otherwise, without compromising PR dev velocity.


.github/workflows/daily-tests.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Check simple 0rtt client
+        run: cargo run --bin simple_0rtt_client
+
+      # Test the server_acceptor binary builds - we invoke with --help since it
+      # will run a server process that doesn't exit when invoked with no args
+      - name: Check server acceptor
+        run: cargo run --bin server_acceptor -- --help
+
+      - name: Check feature powerset
+        run: cargo hack check --feature-powerset --no-dev-deps
```
cpu, line 68: It might make sense to pull this (and maybe the non connect test task) into their own `job`, but for now I've done it all in one for simplicity's sake.

cpu:
Example run where I modified the trigger to run for `push` to test it: https://github.com/cpu/rustls/actions/runs/6027227144

.github/workflows/daily-tests.yml
ctz: I think this does make sense, mainly to avoid running the feature-powerset thing on multiple platforms. Just ubuntu-latest should be enough?

djc (approved):

cpu:
[Test run](https://github.com/cpu/rustls/actions/runs/6038333062/job/16384514296) for the updated branch.

.github/workflows/daily-tests.yml
cpu: I split the workflow up into three jobs: one for the connect tests, one for the example binary smoke tests, and one for the feature powerset. The feature powerset job only runs on ubuntu with stable rust.

ctz (approved):

PR #1439, opened 2023-09-01
doc: link ResolvesServerCert to Acceptor

jsha:
Previously, looking at ResolvesServerCert, it was not clear that there was an async version available.

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -99,6 +99,13 @@ pub trait ProducesTickets: Send + Sync {
 
 /// How to choose a certificate chain and signing key for use
 /// in server authentication.
+///
+/// This is suitable when selecting a certificate does not require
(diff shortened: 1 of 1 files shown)


cpu (approved):
Thanks, this is a nice update :+1:

rustls/src/server/server_conn.rs
```diff
@@ -234,7 +241,9 @@ pub struct ServerConfig<C: CryptoProvider> {
     /// How to produce tickets.
     pub ticketer: Arc<dyn ProducesTickets>,
 
-    /// How to choose a server cert and key.
+    /// How to choose a server cert and key. This is usually set by
```
djc, line 244: Doesn't rustdoc end up looking nicer if you isolate one sentence on the first line?

djc (approved):

PR #1440, opened 2023-09-03
add an example tls server with provider

stevefan1999-personal:
This provides an example server using the crypto provider with some code referenced from internal examples. This is a very bare-bone example with some of the unnecessary features such as CRL removed. It is also assumed to use PKCS_ECDSA_P256_SHA256 for key signing at the moment (due to the config in rcgen uses this).

Use `curl https://<your IP address>:4443 --insecure -v` to view SSL status with curl.
Use `openssl s_client -connect <your IP address>:4443` to debug SSL.

Only ECDSA keys and certs is available since rcgen depends on ring, but ring cannot generate RSA keys and certs yet. There is also an ongoing problem with ring regarding ed25519 support, that the PKCS#8 DER it generated is not standard-compliant: https://github.com/briansmith/ring/pull/1518, so ed25519 test is not included,


stevefan1999-personal:
@ctz should I split the code for the example provider instead?

stevefan1999-personal:
@ctz Sure. I will likely fork the new provider code out, but may I know what software license of that provider code is? I would like to make a repo for those code myself as you suggested but some of the code is yours, so I need to ask for permission first.

stevefan1999-personal:
I force pushed the changes and only included the necessary change to add the example server 

The old provider code is still available at [stevefan1999-personal/rustls at patch-add-more-provider (github.com)](https://github.com/stevefan1999-personal/rustls/tree/patch-add-more-provider) but I will split them by making a new crate that points to the rustls git repo with the ring-optional branch.

cpu:
> Thanks for the server example, we can integrate this after the parent branch is merged.

I converted this to a draft while we're waiting on another branch to make it easier to scan the PR list and get a sense of which PRs should be prioritized for review.

Issue #469, opened 2021-02-04
Memory session cache eviction order is arbitrary


briansmith:
@dhobsd wrote:
> ClientSessionMemoryCache iterates keys of a std::collections::HashMap to evict. This process executes in an arbitrary order, and it's possible to evict the just-inserted key. 

I noticed this too. When the cache is full, we evict an arbitrary element, instead of the least-recently used one.

briansmith:
@dhobsd mentioned that Fuchsia has an implementation of an LRU memory cache already: https://fuchsia.googlesource.com/fuchsia/+/refs/changes/62/478362/24/src/lib/fuchsia-hyper/src/session_cache.rs.

Perhaps we can solve this by just having Google contribute their LRU implementation to replace the default memory cache implementation in Rustls?

dhobsd:
Yeah, I'm happy to contribute -- given that it's part of what's needed for C.4 (but is a separate issue), I wonder if the desire is to implement it separately? We're stuck maintaining what we have in-tree anyway, so I'm happy to contribute however makes the most sense for rustls.

briansmith:
The severity of this problem has decreased with PR #506, since even if the current session is removed from the cache during the handshake, it won't affect this handshake. Now this is merely a performance issue instead of a correctness issue, IIUC.

dhobsd:
Yeah, agree. I think it's also worth mentioning that the correctness aspect isn't particularly security-weakening. Quick fix for this might just be to evict when full and we're inserting using the HashMap `entry` API. WDYT about something like this:

```diff
diff --git a/rustls/src/client/handy.rs b/rustls/src/client/handy.rs
index 712d27c..01fa62a 100644
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -41,7 +41,7 @@ impl ClientSessionMemoryCache {
 
     fn limit_size(&self) {
         let mut cache = self.cache.lock().unwrap();
-        while cache.len() > self.max_entries {
+        while cache.len() >= self.max_entries {
             let k = cache.keys().next().unwrap().clone();
             cache.remove(&k);
         }
@@ -53,8 +53,12 @@ impl client::StoresClientSessions for ClientSessionMemoryCache {
         self.cache
             .lock()
             .unwrap()
-            .insert(key, value);
-        self.limit_size();
+            .entry(key)
+            .and_modify(|e| *e = value.clone())
+            .or_insert_with(|| {
+                self.limit_size();
+                value.clone()
[...]

PR #1442, opened 2023-09-03
use caches instead for session cache

stevefan1999-personal:
This PR almost removes the entirety of `limited_cache.rs`, because I spotted that it is just a LRU in a nutshell. This let us use more cache replacement policies, for example, the [Adaptive Replacement Cache](https://en.wikipedia.org/wiki/Adaptive_replacement_cache) as an experiment.  

The `caches` crate offers `no_std` mode and it is taken care of. But there are few problems:

1.  [Hidden KeyRef type complicates key lookup · Issue #2 · al8n/caches-rs (github.com)](https://github.com/al8n/caches-rs/issues/2)
2.  [Buggy behavior on ARC · Issue #9 · al8n/caches-rs (github.com)](https://github.com/al8n/caches-rs/issues/9)

I have worked around the issue and made subsequent PRs for this. It is now await feedback and will replace the current git crate repo it to point to the main crates.io once the PRs are accepted and published.

Also it seems like the initial cache size is off by one, so I "fixed" it in the tests. It turns out the original LRU cache implementation seems to have an open interval.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -16,6 +16,7 @@ build = "build.rs"
 rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
+caches = { version = "0.2.4", features = ["core"], default-features = false, git = "https://github.com/stevefan1999-personal/caches-rs" }
 log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 subtle = "2.5.0"
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -2,7 +2,7 @@ use crate::client;
 use crate::enums::SignatureScheme;
 use crate::error::Error;
 use crate::key;
-use crate::limited_cache;
+use crate::limited_cache::CacheExt;
 use crate::msgs::persist;
 use crate::sign;
 use crate::NamedGroup;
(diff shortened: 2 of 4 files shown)


ctz:
Here's a comment from last time we decided against taking a crate dependency for this: https://github.com/rustls/rustls/issues/469#issuecomment-812601706

I think that still stands -- and I'm not feeling the caches crate is meeting that bar.

stevefan1999-personal:
@al8n wonder if you would like to fix those so that we could include it in rustls?

al8n:
> @al8n wonder if you would like to fix those so that we could include it in rustls?

I would like to fix those, I will take a look this weekend.

cpu:
It sounds like at a minimum this work is blocked on upstream adjustments to the `caches` dependency. I'm a little bit hesitant to leave a PR open in the tracker that has that sort of outstanding work blocking it. It's already collected up a number of conflicts with `main`.

I propose we close this PR until it seems more viable to land in days as opposed to weeks+. WDYT?

djc:
I agree we should close this, I don't think the caches dependency is a good fit for rustls.

al8n:
`caches` was developed when I was starting to learn Rust, but at the moment, I do not have enough time to improve the code quality. So I agree with closing this for now.

PR #1444, opened 2023-09-04
Make ci-bench more deterministic

aochagavia:
* Switch to unbuffered stdio
* Use FxHasher where possible


aochagavia:
I have another improvement in mind, which is to automatically output the cachegrind diff for the benchmarks that have significant instruction differences. I'll leave that for tomorrow, though

ci-bench/src/main.rs
```diff
@@ -116,8 +117,8 @@ fn main() -> anyhow::Result<()> {
                 .get(index as usize)
                 .ok_or(anyhow::anyhow!("Benchmark not found: {index}"))?;
 
-            let mut stdin = io::stdin().lock();
-            let mut stdout = io::stdout().lock();
+            let mut stdin = util::UnbufferedStdinReader;
```
ctz, line 120: if i've understood this correctly, it would be equivalent to do `let mut stdin = unsafe { File::from_raw_fd(0) };`

aochagavia: I like that! The only drawback I see is that each `File` will be closed upon drop, meaning that `stdin` and `stdout` will get closed too. I think that's not a real problem, though, because if we ever want to do use stdio elsewhere, we'll want to do so in an unbuffered way (reusing the files).

(Just pushed the changes use `File`)

djc (approved):

ci-bench/src/main.rs
djc: Could do a `mem::forget()` for the files, I suppose?

aochagavia:
The failing tests seem transient

PR #1446, opened 2023-09-06
CI bench fix: create cg_diff tmp file at the cwd

aochagavia:
The CI benchmarks are currently failing, because a temp file was previously created under `target` (and the `target` dir doesn't exist under `ci-bench` when running in the CI).

I've validated that this is working in a PR to my own fork: https://github.com/aochagavia/rustls/pull/2 (the error only happened when significant differences were found between the baseline and the candidate).

I also added a commit to show the valgrind version after installing it, since it might come in handy later for debugging. It's unrelated to the fix, but it felt overkill to create a separate PR for it (I can still split it off, if you think it's important).

### DIFF
--- a/.github/workflows/icount-bench.yml
+++ b/.github/workflows/icount-bench.yml
@@ -7,7 +7,9 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Install valgrind
-        run: sudo apt install -y valgrind
+        run: |
+          sudo apt install -y valgrind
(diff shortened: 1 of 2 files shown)


aochagavia:
I've also validated that this is working in a PR to my own fork: https://github.com/aochagavia/rustls/pull/2

ctz (approved):

ctz (approved):

Issue #265, opened 2019-07-08
Use zeroize for best-effort secret zeroisation


ctz:
As a defence-in-depth measure, we should make a passing, best-effort attempt to clear secrets when we're done with them.

burdges:
Actually clear_on_drop is outdated and unmaintained  https://github.com/cesarb/clear_on_drop/issues/20  and may brake in future rustc versions.

You should use zeroize instead:  https://github.com/iqlusioninc/crates/tree/develop/zeroize

If your dependencies do not implement zeroize then you can `impl Zeroize` yourself by copying and calling this `zeroize_hack` fn:  https://github.com/w3f/schnorrkel/blob/master/src/lib.rs#L225

djc:
I can see how it makes sense to use zeroize for TLS 1.2. For TLS 1.3 the secrets are kept as `ring::hkdf::Prk`s which I don't think we can apply zeroize to on the rustls side. There's an [open issue](https://github.com/briansmith/ring/issues/15) in *ring*.

briansmith:
> I can see how it makes sense to use zeroize for TLS 1.2. For TLS 1.3 the secrets are kept as `ring::hkdf::Prk`s which I don't think we can apply zeroize to on the rustls side. There's an [open issue](https://github.com/briansmith/ring/issues/15) in *ring*.

One of my goals for *ring* is to make it so people can write something like Rustls such that there are no `&[u8]` of key material exposed to the application except for whatever persistent keys it (de)serializes. I don't think *ring* quite meets that goal but it's still a goal. I would be happy to address whatever shortcomings there are there, e.g. in the HKDF code.

Along those lines, if there are non-application-data buffers with secrets to zeroize in Rustls's TLS 1.3 code then we should see what we can do to eliminate those buffers completely.

As @djc said, the design of TLS 1.2 may force some of these buffers to be exposed to the TLS 1.2 code. However, we should still try to minimize the lifetime of those buffers.

Regarding the application data buffers, I think it would make sense to have a design for how an application would successfully zeroize those buffers, presumably on demand--much like the proposed API for shrinking these to their minimum size on demand. I don't think automatic zeroization is likely to be practical for performance for these large buffers.

[...]

PR #1448, opened 2023-09-06
Use dynamic dispatch for `CryptoProvider`

### LINKED ISSUE #1409: ConfigBuilder: default crypto provider
Right now on `main` the way to build a ClientConfig or ServerConfig is:

```
ClientConfig::<Ring>::builder()
    .with_safe_defaults()
    ...
```

This has two downsides:

 1. Users need to know about crypto provider backends and the possibility of having different ones.
 2. Changing the crypto provider requires an update to all user code.

Most users don't care about the crypto provider backend and are happy to use whatever good default rustls provides.

It would be better if the basic config builder boilerplate did not specify the crypto provider, leaving it up to rustls. That would minimize the mental overhead for users; and it would also allow us to change the default on the rustls side without requiring all users to update their code.

Because the crypto provider is part of the type parameters of ClientConfig, ServerConfig, and (transitively) ConfigBuilder, this might be a little challenging to achieve, but I think it's possible. At a minimum we would have to document that the type parameter could change across versions.

Here's a straw proposal:

```diff
diff --git a/rustls/src/client/client_conn.rs b/rustls/src/client/client_conn.rs
index 497d02a6..fb076e13 100644
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -232,6 +232,18 @@ impl<C: CryptoProvider> fmt::Debug for ClientConfig<C> {
     }
 }
 
+/// Start building a ClientConfig using the default crypto provider.
+///
+/// The [`CryptoProvider`] type parameter of `ClientConfig` may change without being considered
+/// a semver-incompatible change.
+pub fn default_client_config_builder(
+) -> ConfigBuilder<ClientConfig<crate::crypto::ring::Ring>, WantsCipherSuites> {
+    ConfigBuilder {
+        state: WantsCipherSuites(()),
+        side: PhantomData,
+    }
+}
+
 impl<C: CryptoProvider> ClientConfig<C> {
     /// Create a builder to build up the client configuration.
     ///
```

ctz:
The `ClientConfig::builder()` API is restored and suitable for most uses, but now is conditional on the `ring` crate feature. 

`ClientConfig::builder_with_provider(&'static dyn CryptoProvider)` is unconditionally provided for being explicit or use with custom providers. 

Consumers can make their choice explicit:

```rust
rustls::ClientConfig::builder_with_provider(rustls::crypto::ring::RING)
    .with_safe_defaults()
```
(this is durable to changes in our defaults changing the meaning of `ClientConfig::builder()`.)

fixes #1409

### DIFF
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -14,7 +14,6 @@ use itertools::Itertools;
 use rayon::iter::Either;
 use rayon::prelude::*;
 use rustls::client::Resumption;
-use rustls::crypto::ring::Ring;
 use rustls::server::{NoServerSessionStorage, ServerSessionMemoryCache, WebPkiClientVerifier};
 use rustls::{
     ClientConfig, ClientConnection, ProtocolVersion, RootCertStore, ServerConfig, ServerConnection,
@@ -345,11 +344,11 @@ struct StepperIO<'a> {
 struct ClientSideStepper<'a> {
     io: StepperIO<'a>,
     resumption_kind: ResumptionKind,
-    config: Arc<ClientConfig<Ring>>,
+    config: Arc<ClientConfig>,
 }
 
 impl ClientSideStepper<'_> {
-    fn make_config(params: &BenchmarkParams, resume: ResumptionKind) -> Arc<ClientConfig<Ring>> {
+    fn make_config(params: &BenchmarkParams, resume: ResumptionKind) -> Arc<ClientConfig> {
         assert_eq!(params.ciphersuite.version(), params.version);
         let mut root_store = RootCertStore::empty();
         let mut rootbuf =
@@ -422,11 +421,11 @@ impl BenchStepper for ClientSideStepper<'_> {
 /// A benchmark stepper for the server-side of the connection
 struct ServerSideStepper<'a> {
     io: StepperIO<'a>,
-    config: Arc<ServerConfig<Ring>>,
+    config: Arc<ServerConfig>,
 }
 
 impl ServerSideStepper<'_> {
-    fn make_config(params: &BenchmarkParams, resume: ResumptionKind) -> Arc<ServerConfig<Ring>> {
+    fn make_config(params: &BenchmarkParams, resume: ResumptionKind) -> Arc<ServerConfig> {
         assert_eq!(params.ciphersuite.version(), params.version);
 
         let mut cfg = ServerConfig::builder()
--- a/provider-example/src/kx.rs
+++ b/provider-example/src/kx.rs
@@ -1,4 +1,4 @@
-use crypto::SupportedGroup;
+use crypto::SupportedKxGroup;
 use rustls::crypto;
 
 pub struct KeyExchange {
@@ -6,34 +6,17 @@ pub struct KeyExchange {
     pub_key: x25519_dalek::PublicKey,
 }
 
-impl crypto::KeyExchange for KeyExchange {
-    type SupportedGroup = X25519;
-
-    fn start(
-        name: rustls::NamedGroup,
-        _: &[&'static Self::SupportedGroup],
-    ) -> Result<Self, crypto::KeyExchangeError> {
-        if name == rustls::NamedGroup::X25519 {
-            let priv_key = x25519_dalek::EphemeralSecret::random_from_rng(rand_core::OsRng);
-            let pub_key = (&priv_key).into();
-            Ok(KeyExchange { priv_key, pub_key })
-        } else {
-            Err(crypto::KeyExchangeError::UnsupportedGroup)
-        }
-    }
-
-    fn complete<T>(
-        self,
+impl crypto::ActiveKeyExchange for KeyExchange {
+    fn complete(
+        self: Box<KeyExchange>,
         peer: &[u8],
-        f: impl FnOnce(&[u8]) -> Result<T, ()>,
-    ) -> Result<T, rustls::Error> {
+    ) -> Result<crypto::SharedSecret, rustls::Error> {
         let peer_array: [u8; 32] = peer
             .try_into()
             .map_err(|_| rustls::Error::from(rustls::PeerMisbehaved::InvalidKeyShare))?;
         let their_pub = x25519_dalek::PublicKey::from(peer_array);
         let shared_secret = self.priv_key.diffie_hellman(&their_pub);
-        f(shared_secret.as_bytes())
-            .map_err(|_| rustls::Error::from(rustls::PeerMisbehaved::InvalidKeyShare))
+        Ok(crypto::SharedSecret::from(&shared_secret.as_bytes()[..]))
     }
 
     fn pub_key(&self) -> &[u8] {
@@ -43,19 +26,23 @@ impl crypto::KeyExchange for KeyExchange {
     fn group(&self) -> rustls::NamedGroup {
         X25519.name()
     }
-
-    fn all_kx_groups() -> &'static [&'static Self::SupportedGroup] {
-        ALL_KX_GROUPS
-    }
 }
+
+pub const ALL_KX_GROUPS: &[&dyn SupportedKxGroup] = &[&X25519 as &dyn SupportedKxGroup];
 
 #[derive(Debug)]
 pub struct X25519;
 
-impl crypto::SupportedGroup for X25519 {
+impl crypto::SupportedKxGroup for X25519 {
+    fn start(&self) -> Result<Box<dyn crypto::ActiveKeyExchange>, rustls::crypto::GetRandomFailed> {
+        let priv_key = x25519_dalek::EphemeralSecret::random_from_rng(rand_core::OsRng);
+        Ok(Box::new(KeyExchange {
+            pub_key: (&priv_key).into(),
+            priv_key,
+        }))
+    }
+
     fn name(&self) -> rustls::NamedGroup {
         rustls::NamedGroup::X25519
     }
 }
-
-const ALL_KX_GROUPS: &[&X25519] = &[&X25519];
--- a/provider-example/src/lib.rs
+++ b/provider-example/src/lib.rs
@@ -6,32 +6,24 @@ mod hmac;
 mod kx;
 mod verify;
 
-pub struct Provider;
+pub static PROVIDER: &'static dyn rustls::crypto::CryptoProvider = &Provider;
 
-impl Provider {
-    pub fn certificate_verifier(
-        roots: rustls::RootCertStore,
-    ) -> Arc<dyn rustls::client::ServerCertVerifier> {
-        Arc::new(rustls::client::WebPkiServerVerifier::new_with_algorithms(
-            roots,
-            verify::ALGORITHMS,
-        ))
-    }
-}
+#[derive(Debug)]
+struct Provider;
 
 impl rustls::crypto::CryptoProvider for Provider {
-    fn fill_random(bytes: &mut [u8]) -> Result<(), rustls::crypto::GetRandomFailed> {
+    fn fill_random(&self, bytes: &mut [u8]) -> Result<(), rustls::crypto::GetRandomFailed> {
         use rand_core::RngCore;
         rand_core::OsRng
             .try_fill_bytes(bytes)
             .map_err(|_| rustls::crypto::GetRandomFailed)
     }
 
-    fn default_cipher_suites() -> &'static [rustls::SupportedCipherSuite] {
+    fn default_cipher_suites(&self) -> &'static [rustls::SupportedCipherSuite] {
         ALL_CIPHER_SUITES
     }
 
-    fn default_kx_groups() -> &'static [&'static dyn rustls::SupportedKxGroup] {
+    fn default_kx_groups(&self) -> &'static [&'static dyn rustls::SupportedKxGroup] {
         kx::ALL_KX_GROUPS
     }
 }
@@ -65,3 +57,12 @@ pub static TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256: rustls::SupportedCipherS
         hmac_provider: &hmac::Sha256Hmac,
         aead_alg: &aead::Chacha20Poly1305,
     });
+
+pub fn certificate_verifier(
+    roots: rustls::RootCertStore,
+) -> Arc<dyn rustls::client::ServerCertVerifier> {
+    Arc::new(rustls::client::WebPkiServerVerifier::new_with_algorithms(
+        roots,
+        verify::ALGORITHMS,
+    ))
+}
--- a/provider-example/src/lib.rs
+++ b/provider-example/src/lib.rs
@@ -20,8 +20,6 @@ impl Provider {
 }
 
 impl rustls::crypto::CryptoProvider for Provider {
-    type KeyExchange = kx::KeyExchange;
-
     fn fill_random(bytes: &mut [u8]) -> Result<(), rustls::crypto::GetRandomFailed> {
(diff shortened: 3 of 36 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -25,69 +25,58 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
...
-pub trait KeyExchange: Sized + Send + Sync + 'static {
-    /// The supported group the key exchange is operating over.
-    type SupportedGroup: SupportedGroup;
+/// A supported key exchange group.
+///
+/// This has a TLS-level name expressed using the [`NamedGroup`] enum, and
+/// a function which produces a [`ActiveKeyExchange`].
+pub trait SupportedKxGroup: Send + Sync + Debug {
+    /// Named group the SupportedKxGroup operates in.
+    fn name(&self) -> NamedGroup;
```
djc, line 44: Nit: order `name()` after `start`?

rustls/src/crypto/mod.rs
```diff
@@ -25,69 +25,58 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
...
-
-    /// Return the group being used.
-    fn group(&self) -> NamedGroup;
+    /// The shared secret is passed into the [`SharedSecretSink`] for further processing.
+    ///
+    /// This consumes and so terminates the [`ActiveKeyExchange`].
+    fn complete(
+        self: Box<Self>,
+        peer_pub_key: &[u8],
+        sink: &mut dyn SharedSecretSink,
```
djc, line 68: Hmm, this trait interface feels a little overwrought. Maybe this should just be an `FnOnce` instead?

rustls/src/client/hs.rs
```diff
@@ -814,16 +813,22 @@ impl<C: CryptoProvider> ExpectServerHelloOrHelloRetryRequest<C> {
 
         let key_share = match req_group {
             Some(group) if group != offered_key_share.group() => {
-                match KeyExchange::start(group, &config.kx_groups) {
-                    Ok(kx) => kx,
-                    Err(KeyExchangeError::UnsupportedGroup) => {
+                let skxg = match config
```
djc, line 816: Maybe turn this into a method on `config` that both this and `ExpectServerDone` can use?

rustls/src/server/tls12.rs
```diff
@@ -404,25 +400,18 @@ mod client_hello {
...
-            Err(_) => {
-                return Err(GetRandomFailed.into());
-            }
-        };
-        let secdh = ServerECDHParams::new(selected_group, kx.pub_key());
+    ) -> Result<Box<dyn ActiveKeyExchange>, Error> {
+        let kx = selected_group
+            .start()
+            .map_err(|_| Error::FailedToGetRandomBytes)?;
+        let secdh = ServerECDHParams::new(kx.group(), kx.pub_key());
```
djc, line 414: Should this constructor be changed to take a `&dyn SupportedKxGroup` instead?

rustls/src/crypto/ring/mod.rs
```diff
@@ -25,25 +25,28 @@ pub mod sign;
...
 
     /// Return all supported key exchange groups.
-    fn default_kx_groups() -> &'static [&'static dyn SupportedKxGroup] {
+    fn default_kx_groups(&self) -> &'static [&'static dyn SupportedKxGroup] {
         ALL_KX_GROUPS
     }
 }
 
+/// A `CryptoProvider` backed by the *ring* crate.
+pub static RING: &dyn CryptoProvider = &Ring;
```
djc, line 48: Nit: move this up before the `Ring` definition?

rustls/src/server/server_conn.rs
```diff
@@ -207,7 +207,10 @@ impl<'a> ClientHello<'a> {
 /// * [`ServerConfig::send_tls13_tickets`]: 4 tickets are sent.
 ///
 /// [`RootCertStore`]: crate::RootCertStore
-pub struct ServerConfig<C: CryptoProvider> {
+pub struct ServerConfig {
+    /// Source of randomness and other crypto.
```
djc, line 211: Nit: I'd keep it at the bottom, seems pretty boring?

ci-bench/src/main.rs
```diff
@@ -345,11 +344,11 @@ struct StepperIO<'a> {
 struct ClientSideStepper<'a> {
     io: StepperIO<'a>,
     resumption_kind: ResumptionKind,
-    config: Arc<ClientConfig<Ring>>,
+    config: Arc<ClientConfig>,
```
djc, line 347: I'd probably still put these in the previous commit.

provider-example/src/lib.rs
```diff
@@ -6,36 +6,28 @@ mod hmac;
...
     }
 
-    fn default_kx_groups() -> &'static [&'static dyn rustls::SupportedKxGroup] {
-        &kx::ALL_KX_GROUPS
+    fn default_kx_groups(&self) -> &'static [&'static dyn rustls::SupportedKxGroup] {
+        kx::ALL_KX_GROUPS
     }
 }
 
+pub static PROVIDER: &'static dyn rustls::crypto::CryptoProvider = &Provider;
```
cpu, line 29: Feels like this should go to the top of the file.

provider-example/src/lib.rs
```diff
@@ -6,36 +6,28 @@ mod hmac;
...
+
 static ALL_CIPHER_SUITES: &[rustls::SupportedCipherSuite] = &[
```
cpu, line 31: I think this static should be below the `pub` members.

rustls/src/crypto/mod.rs
```diff
@@ -24,15 +24,15 @@ pub use crate::rand::GetRandomFailed;
 pub use crate::msgs::handshake::KeyExchangeAlgorithm;
 
 /// Pluggable crypto galore.
-pub trait CryptoProvider: Send + Sync + 'static {
+pub trait CryptoProvider: Send + Sync + Debug + 'static {
```
cpu, line 27: Unfortunately adding the `Debug` constraint in this commit without updating the provider example at the same time means this commit won't build+pass tests in isolation.

rustls/src/crypto/mod.rs
ctz: have replaced this with `&mut dyn FnMut(&[u8])` and that's a big improvement (couldn't figure out how to use `FnOnce` without an extra box in a trait method)

rustls/src/crypto/mod.rs
```diff
@@ -25,69 +25,58 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
...
 }
 
-/// A trait describing a supported key exchange group that can be identified by name.
-pub trait SupportedGroup: Debug + Send + Sync + 'static {
-    /// Named group the SupportedGroup operates in.
-    fn name(&self) -> NamedGroup;
+/// What to do with the shared secret that results from a successful key exchange.
+pub trait SharedSecretSink {
+    /// Consume (in some unspecified way) the supplied `secret`.
+    fn process_shared_secret(&mut self, secret: &[u8]);
```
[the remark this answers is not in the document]
ctz: has fallen away

rustls/src/client/builder.rs
```diff
@@ -14,23 +14,23 @@ use super::client_conn::Resumption;
...
         ConfigBuilder {
             state: WantsClientCert {
+                provider: self.state.provider,
                 cipher_suites: self.state.cipher_suites,
                 kx_groups: self.state.kx_groups,
                 versions: self.state.versions,
                 verifier: Arc::new(webpki::WebPkiServerVerifier::new(root_store)),
             },
-            side: PhantomData,
+            side: core::marker::PhantomData,
```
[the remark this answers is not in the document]
ctz: i will put this back, but i did it for (what I think is) a good reason: 

```rust
#[cfg(any(feature = "dangerous_configuration", feature = "ring"))]
use core::marker::PhantomData;
```

this feels pretty gross; i can't really express how little i like having to cfg-gate imports to avoid `unused_imports` noise.

rustls/src/crypto/ring/mod.rs
```diff
@@ -25,25 +25,28 @@ pub mod sign;
...
     }
 
     /// Return all supported key exchange groups.
-    fn default_kx_groups() -> &'static [&'static dyn SupportedKxGroup] {
+    fn default_kx_groups(&self) -> &'static [&'static dyn SupportedKxGroup] {
         ALL_KX_GROUPS
     }
 }
 
+/// A `CryptoProvider` backed by the *ring* crate.
```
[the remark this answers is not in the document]
ctz: have done that

rustls/src/crypto/mod.rs
ctz: squashed this and the next commit

ci-bench/src/main.rs
ctz: squashed

rustls/src/server/server_conn.rs
ctz: going to move this to third, after the configured ciphersuites and kx-groups. it seems logical that it belongs next to these in the docs?

djc: Sure, that seems alright, too.

rustls/src/client/builder.rs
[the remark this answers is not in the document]
cpu: Seems reasonable to compromise on this if it avoids having to gate the import. Maybe worth a note about that in the `CONTRIBUTING.md`?

provider-example/src/kx.rs
```diff
@@ -43,19 +28,21 @@ impl crypto::KeyExchange for KeyExchange {
...
 #[derive(Debug)]
 pub struct X25519;
 
-impl crypto::SupportedGroup for X25519 {
+impl crypto::SupportedKxGroup for X25519 {
     fn name(&self) -> rustls::NamedGroup {
         rustls::NamedGroup::X25519
     }
+
+    fn start(&self) -> Result<Box<dyn crypto::ActiveKeyExchange>, rustls::crypto::GetRandomFailed> {
```
djc, line 41: Nit: match the method order from the trait definition?

Also I'd probably inline the `priv_key` and `pub_key` bindings.

provider-example/src/kx.rs
```diff
@@ -43,19 +28,21 @@ impl crypto::KeyExchange for KeyExchange {
...
+        let priv_key = x25519_dalek::EphemeralSecret::random_from_rng(rand_core::OsRng);
+        let pub_key = (&priv_key).into();
+        Ok(Box::new(KeyExchange { priv_key, pub_key }))
+    }
 }
 
-const ALL_KX_GROUPS: &[&X25519] = &[&X25519];
+pub const ALL_KX_GROUPS: &[&dyn SupportedKxGroup] = &[&X25519 as &dyn SupportedKxGroup];
```
djc, line 48: Nit: so this should go before the definition of `X25519`?

provider-example/src/kx.rs
```diff
@@ -1,39 +1,24 @@
...
-
-    fn complete<T>(
-        self,
+impl crypto::ActiveKeyExchange for KeyExchange {
+    fn complete(
+        self: Box<KeyExchange>,
         peer: &[u8],
-        f: impl FnOnce(&[u8]) -> Result<T, ()>,
-    ) -> Result<T, rustls::Error> {
+        sink: &mut dyn FnMut(&[u8]),
```
djc, line 13: Why not `FnOnce`? The current version of the call sites looks pretty bad with all the unwrapping, I think's probably a way to make this cleaner?

provider-example/src/lib.rs
```diff
@@ -65,3 +52,17 @@ pub static TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256: rustls::SupportedCipherS
         hmac_provider: &hmac::Sha256Hmac,
         aead_alg: &aead::Chacha20Poly1305,
     });
+
+static ALL_CIPHER_SUITES: &[rustls::SupportedCipherSuite] = &[
```
djc, line 56: Nit: keep this above the cipher suite statics?

djc (approved):

provider-example/src/kx.rs
ctz: `FnOnce` consumes `self`, which means the caller of it needs the original closure rather than a reference to it. But the caller is the implementer of a trait, so can't be generic over the several possible `FnOnce` closures. The other alternative would be to have a `Box<dyn FnOnce>` which can be consumed, or have `complete` return the secret value (which would solve all the other shenanigans, but has other costs).

djc: Can you make a quick list of options you see and their pros/cons?

ctz: - result passed into `&mut dyn FnMut(&[u8])` (currently)
  - 🟢 pro: no extra allocations
  - 🔴 con: lots of shenanigans needed to integrate with TLS1.3 key schedule session types; this will panic if _actually_ called twice
- result passed into a `Box<FnOnce(&[u8])>`
  - 🟢 pro: `FnOnce` is the correct type for the meaning we want
  - 🔴 con: extra allocation
- result directly returned from `finish()`
  - 🟢 pro: way simpler, especially for the TLS1.3 key schedule session types.
  - 🔴 con: extra copy
  - 🔴 con: extra allocation
  - 🔴 con: doing an end run around carefully designed underlying type from crypto libaries (eg, x25519-dalek has https://docs.rs/x25519-dalek/latest/x25519_dalek/struct.SharedSecret.html which impls Zeroize, ring passes the secret into a closure)

ctz: > result passed into a `Box<FnOnce(&[u8])>`

having tried this i'm no longer sure it will work, because the closure has its lifetime (and of its captures)  extended by the box. that means it's no longer possible to use anything captured after the `complete` call.

going to go for the simple option

ctz: > going to go for the simple option

added this as a seperate commit for incremental review; then i will squash it into the first commit

rustls/src/crypto/mod.rs
```diff
@@ -59,18 +59,30 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
...
+/// The result from `ActiveKeyExchange::complete` as a value.
+pub struct SharedSecret(Vec<u8>);
+
+impl From<&[u8]> for SharedSecret {
+    fn from(source: &[u8]) -> Self {
+        Self(source.to_vec())
+    }
+}
+
+impl AsRef<[u8]> for SharedSecret {
```
djc, line 84: How do you feel about doing some kind of zeroization here?

For the pki-types private key types I avoided `AsRef<[u8]>` in favor of an inherent `secret_der()` method, do you think it's worth doing something similar here? In particular, we could make the inherent method `pub(crate)` so that callers outside this crate can't access it, which might be nice?

ctz: Yeah I still think #265 is worth doing for belt-and-braces, especially now we have more types that contain secrets (eg, `AeadKey`). This type is another candidate for that treatment.

>  which might be nice?

Yeah, agree -- have done that.

The other follow-on from this is to take the `SharedSecret` in the key schedule state transition that requires it; I think that clarifies the meaning of the code a bit.

provider-example/src/kx.rs
ctz: reordered

have inlined `pub_key`, but `priv_key` needs to be a separate binding so `pub_key` can be derived from it

djc (approved):

PR #1453, opened 2023-09-08
ci-bench: reduce noise in cg_diff output

ctz:
The diffs produced tend to be noisy here because two separate compilations have different per-type and per-compilation uniqueness. eg:

```
 29,792 (124.8%)  ???:_ZN5alloc7raw_vec11finish_grow17h463b2c6f0ba30854E.llvm.2614985587368234107
-29,792 (-124.8%)  ???:_ZN5alloc7raw_vec11finish_grow17h463b2c6f0ba30854E.llvm.3375118279659775674
```

This diff line is here because some per-compilation unique value (after the '.llvm.') changed, not because the instruction count changed.

We can chop these out by giving a regular expression to cg_diff.

### DIFF
--- a/ci-bench/src/cachegrind.rs
+++ b/ci-bench/src/cachegrind.rs
@@ -242,6 +242,10 @@ pub fn diff(baseline: &Path, candidate: &Path, scenario: &str) -> anyhow::Result
     // cg_diff generates a diff between two cachegrind output files in a custom format that is not
     // user-friendly
     let cg_diff = Command::new("cg_diff")
+        // remove per-compilation uniqueness in symbols, eg
+        // _ZN9hashbrown3raw21RawTable$LT$T$C$A$GT$14reserve_rehash17hc60392f3f3eac4b2E.llvm.9716880419886440089 ->
+        // _ZN9hashbrown3raw21RawTable$LT$T$C$A$GT$14reserve_rehashE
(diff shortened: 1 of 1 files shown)


cpu (approved):
Looks good to me!

To save others having to google the docs for `--mod-funcname`:

> In this case, use the --mod-filename option. Its argument is a search-and-replace expression that will be applied to all the filenames in both Cachegrind output files. It can be used to remove minor differences in filenames. For example, the option --mod-filename='s/version[0-9]/versionN/' will suffice for the above example.
>
> Similarly, sometimes compilers auto-generate certain functions and give them randomized names like T.1234 where the suffixes vary from build to build. You can use the --mod-funcname option to remove small differences like these; it works in the same way as --mod-filename.

djc:
Do we have any clue about how stable the `17h` prefix is/will be?

djc (approved):

ctz:
> Do we have any clue about how stable the `17h` prefix is/will be?

I found the docs for that hash -- https://rustc-dev-guide.rust-lang.org/backend/libs-and-metadata.html#stable-crate-id. The 17 means "17 characters follow" and I guess the "h" designates hex or "hash". I would guess rustc changes symbol mangling extremely infrequently.

djc:
Makes sense! Maybe adding a comment about this would be nice.

PR #1461, opened 2023-09-12
CONTRIBUTING.md: "What would djc do"

cpu:
**Note: using #1407 as the merge base**.

Extends the RFC style-guide with additional wisdom gained from PR reviews. 

For now I've chosen not to include the (potentially controversial) guidance on preferring turbofish over explicit type annotations (e.g. when `collect`ing). We should probably confirm consensus on that one first.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,6 +13,11 @@ a pcap or reproduction steps.
 
 Feel free to file github issues to get help, or ask a question.
 
+If you believe you've found a security bug please 
+[open a draft security advisory](https://github.com/rustls/rustls/security/advisories/new) 
+in GitHub, and not as a regular repository issue. See [SECURITY.md] for more
+information.
+
 ## Code changes
 
 Some ideas and guidelines for contributions:
@@ -30,13 +35,10 @@ Some ideas and guidelines for contributions:
 
 ## Security bugs
 
-Please report security bugs by filing a github issue, or by
-email to jbp@jbp.io if you want to disclose privately.  I'll then:
+Please report security bugs by [opening a draft security advisory](https://github.com/rustls/rustls/security/advisories/new)
+in GitHub, and not as a regular repository issue. 
 
-- Prepare a fix and regression tests.
-- Backport the fix and make a patch release for most recent release.
-- Submit an advisory to [rustsec/advisory-db](https://github.com/RustSec/advisory-db).
-- Refer to the advisory on the main README.md and release notes.
+See [SECURITY.md] for more information.
 
 If you're *looking* for security bugs, this crate is set up for
 `cargo fuzz` but would benefit from more runtime, targets and corpora.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,6 +127,12 @@ differentiating between a given function and other functions, so putting them
 first makes it easier to infer the context/meaning of the function (compared to
 starting with a number of generic context-like types).
 
+#### Use `impl` where possible
+
+We prefer to use `impl ...` for arguments and return types when there's a single
+use of the type. Generic type argument bounds add a level of indirection that's
+harder to read in one pass.
+
 #### Validation
 
 Where possible, avoid writing `validate` or `check` type functions that try to
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,14 +143,14 @@ be represented.
 
 #### Error handling
 
-We use `Result` types pervasively throughout the code to signal error cases. We
-prefer to avoid `unwrap()` and `expect()` calls unless there is a clear
-invariant which can be locally validated by the structure of the code. If
-there is such an invariant, we usually add a comment explaining how the
-invariant is upheld. In other cases (especially for error cases which can arise
-from network traffic, which could represent an attacker), we always prefer to
-handle errors and ultimately return an error to the network peer or close the
-connection.
+We use `Result` types pervasively throughout the code to signal error cases. 
+Outside of unit/integration tests we prefer to avoid `unwrap()` and `expect()`
+calls unless there is a clear invariant which can be locally validated by the
+structure of the code. If there is such an invariant, we usually add a comment
+explaining how the invariant is upheld. In other cases (especially for error
+cases which can arise from network traffic, which could represent an attacker),
+we always prefer to handle errors and ultimately return an error to the network
+peer or close the connection.
 
 ### Expressions
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -194,6 +194,14 @@ We use 3 blocks of imports in our Rust files:
 
 We believe that this makes it easier to see where a particular import comes from.
 
+We prefer to reference types and traits by an imported symbol name instead of
+using qualified references. Qualification paths generally add noise and are
+unnecessary. The one exception to this is when the symbol name is overly
+generic, or easily confused between different crates. In this case we prefer to
+import the symbol name under an alias, or if the parent module name is short,
+using a one-level qualified path. E.g. for a crate with a local `Error` type,
+prefer to `import std::error::Error as StdError`.
+
 ## Licensing
 
 Contributions are made under [rustls's licenses](LICENSE).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -183,6 +183,18 @@ Ok(match foo {
 })
 ```
 
+#### Avoid `ref` in match patterns
+
+When writing match expressions, try to avoid using `ref` in patterns. Prefer
+taking a reference on the
+[scrutinee](https://doc.rust-lang.org/reference/expressions/match-expr.html) 
+of the `match`.
+
+Since the addition of [binding
+modes](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html) for improved
+match ergonomics the `ref` keyword is unidiomatic and can be unfamiliar to
+readers.
+
 ### Naming
 
 #### Use concise names
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,14 @@ splitting it up into many small single-use functions. Single-caller functions
 can be useful, but consider whether the code would be easier to follow if the
 function was inlined (especially for short functions).
 
+#### Consider avoiding free-standing functions
+
+If a function's semantics or implementation are strongly dependent on one of its
+arguments, and the argument is defined in a type within the current crate,
+prefer using a method on the type. Similarly, if a function is taking multiple
+arguments that originate from the same common type in all call-sites it is
+a strong candidate for becoming a method on the type.
+
 #### Order arguments from most specific to least specific
 
 When writing a function, we prefer to order arguments from most specific to
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -224,6 +224,12 @@ Per the
 [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter),
 `get_()` prefixes are discouraged.
 
+#### Don't elide generic lifetimes
+
+We prefer not to elide lifetimes when naming types that are generic over
+lifetimes. Always include a lifetime placeholder (e.g. `<'_>`) to avoid
+confusion.
+
 ### Imports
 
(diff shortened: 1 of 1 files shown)


cpu:
(CI failures are from the stale base branch)

CONTRIBUTING.md
```diff
@@ -30,13 +35,10 @@ Some ideas and guidelines for contributions:
 
 ## Security bugs
 
-Please report security bugs by filing a github issue, or by
-email to jbp@jbp.io if you want to disclose privately.  I'll then:
+Please report security bug by [opening a draft security advisory](https://github.com/rustls/rustls/security/advisories/new)
```
djc, line 38: "Please report security bug" -> "a security bug" or "security bugs"?

CONTRIBUTING.md
```diff
@@ -110,6 +110,12 @@ splitting it up into many small single-use functions. Single-caller functions
 can be useful, but consider whether the code would be easier to follow if the
 function was inlined (especially for short functions).
 
+#### Consider avoiding free-standing functions
+
+Often free-standing functions take arguments that could be associated with
```
djc, line 115: Maybe elaborate on this a bit? For example, if a function's semantics or implementation are strongly dependent on one of its arguments and that argument is defined in the current crate, it should probably become a method. (Also consider the case where a function is getting multiple arguments that originate from the same common type in all callers.)

CONTRIBUTING.md
```diff
@@ -169,7 +169,9 @@ We prefer concise names, especially for local variables. Avoid adding a suffix
 for a variable that describes its type (provided that its type is hard to
 confuse with other types -- for example, we do still use `_id` suffixes because
 we usually use numeric IDs for database entities). The precision/conciseness
-trade-off for variable names also depends on the scope of the binding.
+trade-off for variable names also depends on the scope of the binding. In
+general we prefer to expand acronyms (e.g. using `keyusage` instead of `ku`,
```
djc, line 209: `keyusage` is a bad example because we should snake_case it: `key_usage`. This feels a little tacked on to this paragraph... maybe it should go up front, after the first sentence -- especially since it is kind of at odds with it. Might also worth adding a note about how well-known the acronym/abbreviation is. For example, I would not hesitate to call a variable `html` or `url` but I think `ku` or `eku` are indeed relatively obscure.

CONTRIBUTING.md
```diff
@@ -189,6 +189,9 @@ We use 3 blocks of imports in our Rust files:
 
 We believe that this makes it easier to see where a particular import comes from.
 
+We prefer to import core traits, e.g. using `TryFrom` instead of
```
djc, line 242: `TryFrom` is a bad example, because it has been a part of the prelude since the 2021 edition (that is, it doesn't need to be imported at all but can still be referenced directly without qualifying).

I think this isn't really particular to `core` and more broadly holds: I think our style is that we generally prefer referencing types and traits by name directly without qualifying because qualifying adds some noise that is generally unnecessary. Except if the symbol is overly generic/easily confused between different crates (for example, `Error` will generally refer to the local crate's `Error` type, so we alias `std::error::Error` to `StdError` and refer to `io::Error` using a one-level qualified path -- benefitting from the short module name to avoid too much verbosity).

CONTRIBUTING.md
```diff
@@ -161,6 +161,24 @@ levels, which we call "rightward drift". This makes lines shorter, making the
...
+    1..10 => Ok(do_one_thing()),
+    _ => Ok(do_another()),
+}
+// Better
+Ok(match foo {
+    1..10 => do_one_thing(),
+    _ => do_another(),
+})
+```
+
```
djc, line 194: Maybe add some explicit explanation here: "This helps separate the logic that is different per match arm from logic that is shared across arms, thus reducing code duplication."

Also I would argue this is not just for `match` but applies to `if` too -- in general, taking advantage of the idea that many syntax constructs in Rust yield an expression.

Nits for the example: add an empty line between the two sides? "Not to style" seems a slightly uncommon way of formulating this...

CONTRIBUTING.md
```diff
@@ -179,6 +179,11 @@ Ok(match foo {
 })
 ```
 
+#### Avoid `ref` in match patterns
+
+When writing match expressions, try to avoid using `ref` in patterns. Prefer
+taking a reference on the subject of the `match`.
```
djc, line 198: The "subject" is technically called the `scrutinee`. Maybe refer to the Rust reference for readers who are unfamiliar with that expression?

The feature was introduced relatively late and is known as "binding modes", maybe worth referencing https://rust-lang.github.io/rfcs/2005-match-ergonomics.html or do you think that's too detailed?

Also consider adding some reasoning, like that it is unidiomatic in today's Rust and that the meaning of the `ref` keywords is a little unfamiliar since it is used here to take a reference (unlike the use of `&` that's usually used in expression contexts).

CONTRIBUTING.md
```diff
@@ -210,6 +210,12 @@ Per the
 [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter),
 `get_()` prefixes are discouraged.
 
+#### Don't elide generic lifetimes
+
+We prefer not to elide lifetimes when naming types that are generic over
+liftimes. Always include a lifetime placeholder (e.g. `<'_>`) to avoid
```
djc, line 229: "liftimes" -> "lifetimes"

CONTRIBUTING.md
```diff
@@ -125,6 +125,11 @@ differentiating between a given function and other functions, so putting them
 first makes it easier to infer the context/meaning of the function (compared to
 starting with a number of generic context-like types).
 
+#### Use `impl` where possible
+
+We prefer to use `impl ...` for arguments where possible. Generic type argument
```
djc, line 130: The title is "Use `impl` where possible" but the contents then restrict this to "for arguments". Should this also say something about the use of return position impl types?

Maybe also explain "where possible" a bit, this only works when there's a single use of the type.

CONTRIBUTING.md
```diff
@@ -234,6 +234,13 @@ We believe that this makes it easier to see where a particular import comes from
 We prefer to import core traits, e.g. using `TryFrom` instead of
 `core::convert::TryFrom`.
 
+### Misc
+
+#### Avoid type aliases
+
+We prefer to avoid type aliases as they can obfuscate the underlying type and
+don't often provide significant additional type safety.
```
djc, line 255: "often" and "significant" feel like "weasel words" to me here: IMO they just don't. I would remove "can" for similar reasons. Maybe add a note about using a newtype wrapper as an alternative where an abstraction boundary is warranted/worth the complexity.

CONTRIBUTING.md
```diff
@@ -215,6 +215,11 @@ Per the
 [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter),
 `get_()` prefixes are discouraged.
 
+#### Alphabetize enum variants
+
+When implementing or modifying an `enum` type, list its variants in alphabetical
```
djc, line 220: Unless overriding reasons not to, for example a discriminant order imposed by a standard.

CONTRIBUTING.md
```diff
@@ -215,11 +215,14 @@ Per the
 [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter),
 `get_()` prefixes are discouraged.
 
-#### Alphabetize enum variants
+#### Enum variants
 
 When implementing or modifying an `enum` type, list its variants in alphabetical
 order.
 
+Prefer active verbse for variant names. E.g. `Allow` instead of `Allowed`,
```
djc, line 223: "verbse". The other thing I noticed recently which I consider unidiomatic is faux-bools like `::Yes` and `::No` -- instead, use variant names that are descriptive of the different states.

CONTRIBUTING.md
```diff
@@ -244,6 +244,11 @@ We prefer to import core traits, e.g. using `TryFrom` instead of
 
 ### Misc
 
+#### Numeric literals
+
+We generally prefer numeric literals to be written in hexadecimal instead of
```
djc, line 249: I would qualify this more? Consider picking a base to fit in the with the domain in which the value is used. For example, you might want to use hexadecimal for protocol message encoding and octal for Unix privileges.

Maybe also mention something about the digit grouping here, as in `1_000_000`?

djc (commented):
Good stuff!

cpu:
@djc Thanks for the feedback, very helpful. I will iterate :+1:

CONTRIBUTING.md
cpu: Fixed, thanks

CONTRIBUTING.md
cpu: I borrowed some of your wording for this. Thanks.

CONTRIBUTING.md
cpu: Fixed `keyusage` -> `key_usage`, moved to the front, and added a note about very common counterexamples.

CONTRIBUTING.md
cpu: Reworked to not be specific to `core`, borrowing your examples.

CONTRIBUTING.md
cpu: Added more justification, de-emphasized the `match` context, added a space between examples and changed the comments to be less uncommon formulations.

CONTRIBUTING.md
cpu: > The "subject" is technically called the scrutinee. Maybe refer to the Rust reference for readers who are unfamiliar with that expression?

Sounds good. I wasn't sure if `scrutinee` was well know but with a link to the reference I think it's fine :+1: 

> Also consider adding some reasoning

Added some more reasoning and a link to the match ergonomics RFC. I think it's fair to link to more detailed resources.

CONTRIBUTING.md
cpu: Fixed thanks. Turns out writing a docs update at the very end of the day isn't a great way to avoid typos :sweat:

CONTRIBUTING.md
cpu: I added a mention of return types and tried to expand on "where possible". I'm not sure I've done this one justice (probably owing in part to not being super confident in this area myself).

CONTRIBUTING.md
cpu: Struck the weasel words and made a note about the newtype idiom.

CONTRIBUTING.md
cpu: Added a caveat about this.

CONTRIBUTING.md
cpu: > "verbse".

Fixed.

> The other thing I noticed recently which I consider unidiomatic is faux-bools like ::Yes and ::No

Explicitly mentioned this as well.

CONTRIBUTING.md
cpu: > I would qualify this more? Consider picking a base to fit in the with the domain in which the value is used.

Borrowed your wording for this :+1: 

> Maybe also mention something about the digit grouping here, as in 1_000_000?

Done.

cpu:
Rolled in all of the feedback, thanks!

ctz (approved):

CONTRIBUTING.md
```diff
@@ -194,6 +194,15 @@ We use 3 blocks of imports in our Rust files:
 
 We believe that this makes it easier to see where a particular import comes from.
 
+We prefer to reference types and traits by an imported symbol name instead of
+using qualified references. Qualification paths generally add noise and are
+unnecessary. The one exception to this is when the symbol name is overly
+generic, or easily confused between different crates. In this case we prefer to
+import the symbol name under an alias, or if the parent module name is short,
+using a one-level qualified path. E.g. for a crate with a local `Error` type,
+prefer to `import std::error::Error as StdError` and refer to IO specific errors
```
djc, line 265: Nit: "IO specific errors", in my previous comment I specifically meant `std::io::Error` -- I feel like `io` should always be `std::io` in relevant code. I don't think we need to go into that deeply here, so maybe just narrow this suggestion again?

djc (approved):
Great!

CONTRIBUTING.md
djc: Seems pretty good to me.

PR #1462, opened 2023-09-13
Merge pki-types history into rustls repository

djc:
This contains a merge commit, so I'm not sure we can/want to rebase it.


ctz:
I suggest when transferring the tags after this merges, we rename them `(subdir)/(semver)`, eg, `pki-types/0.2.1`?

djc:
Renaming them makes sense to me, I would paint this bikeshed the color hypen (`pki-types-0.2.1`).

ctz:
And for the workflow yml, I suggest we name that by the subdirectory name too? If we keep a separate pki-types workflow, should we restrict it to running for changes that touch that subdirectory? I think I'd prefer to avoid having separate workflows for each sub-crate until and unless we find it too inflexible.

djc:
> And for the workflow yml, I suggest we name that by the subdirectory name too? If we keep a separate pki-types workflow, should we restrict it to running for changes that touch that subdirectory? I think I'd prefer to avoid having separate workflows for each sub-crate until and unless we find it too inflexible.

I think we can get away without separate workflows for now, especially since pki-types is very fast to build and we can relegate slower parts to the daily tests workflow.

djc:
> > I think we can get away without separate workflows for now, especially since pki-types is very fast to build and we can relegate slower parts to the daily tests workflow.
> 
> Agree. Do you want to have a commit in this PR that deletes `.github/workflows/ci.yml` and extends `build.yml` to format/clippy etc pki-types/?

Yes, I think most of that will happen "automatically" because the PR makes pki-types part of the workspace,  but I'll make sure to check.

stevefan1999-personal:
Did you use `git subtree` to merge it or cherry-picked?

cpu (approved):
I think this looks good, but admittedly I haven't reviewed many branches that perform a repo-wide git import.

cpu:
@ctz @djc What needs to happen to move this forward? Is there a reason to defer the merge until a later date?

ctz (approved):
AFAICT the extra `ci.yml` has fallen away now.

ctz:
Now pki-types has a slightly wider scope I suspect we probably don't want to do this? (I'm also repeatedly struggling these days with cargo workspaces containing dependencies that depend on something inside the workspace, which guarantees unhelpful duplication of the crate.)

djc:
Yup, it also hasn't been much of a pain point IME?

cpu:
> Yup, it also hasn't been much of a pain point IME?

Yeah I think the main cost is in the overhead of duplicating CI bits and we haven't been tweaking those much lately.

PR #1465, opened 2023-09-13
More efficient handshake encoding

ctz:
This is showing a 2-3% improvement on resumption benchmarks for me.

### DIFF
--- a/rustls/src/msgs/codec.rs
+++ b/rustls/src/msgs/codec.rs
@@ -210,40 +210,10 @@ impl Codec for u64 {
 /// `TlsListElement` provides the size of the length prefix for the list.
 impl<T: Codec + TlsListElement + Debug> Codec for Vec<T> {
     fn encode(&self, bytes: &mut Vec<u8>) {
-        let len_offset = bytes.len();
-        bytes.extend(match T::SIZE_LEN {
-            ListLength::U8 => &[0][..],
-            ListLength::U16 => &[0, 0],
-            ListLength::U24 { .. } => &[0, 0, 0],
-        });
+        let nest = LengthPrefixedBuffer::new(T::SIZE_LEN, bytes);
 
         for i in self {
-            i.encode(bytes);
-        }
-
-        match T::SIZE_LEN {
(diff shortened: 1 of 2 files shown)


djc (commented):
Awesome! How did you discover this?

I would suggest changing the `NestedLength` type to hold on to the `bytes` reference (maybe `buf`?), and renaming it accordingly (maybe `ListBuffer`?). Could expose the buffer via `pub` field or a `get_mut()` getter, and add a `Drop` impl that does the `finish()` work.

ctz:
> Awesome! How did you discover this?

I'm experimenting with improving how extensions are represented internally. Especially, to remove the need for a `HashMap` and manual checking to detect duplicate extensions.

> I would suggest changing the `NestedLength` type to hold on to the `bytes` reference (maybe `buf`?), and renaming it accordingly (maybe `ListBuffer`?). Could expose the buffer via `pub` field or a `get_mut()` getter, and add a `Drop` impl that does the `finish()` work.

Yes, this is a lot better. I was a bit nervous of using Drop, probably without good reason.

rustls/src/msgs/codec.rs
```diff
@@ -270,9 +270,9 @@ impl<'a> LengthPrefixedBuffer<'a> {
     pub(crate) fn new(size_len: ListLength, buf: &'a mut Vec<u8>) -> LengthPrefixedBuffer<'a> {
         let len_offset = buf.len();
         buf.extend(match size_len {
-            ListLength::U8 => &[0][..],
-            ListLength::U16 => &[0, 0],
-            ListLength::U24 { .. } => &[0, 0, 0],
+            ListLength::U8 => &[0xff][..],
```
djc, line 273: Commit message has some interesting spellings of "interpreted" like "intepretted" and "interpretted".

djc (approved):
Cool!

rustls/src/msgs/codec.rs
ctz: shameful 🙈

cpu (approved):
Cool!

Issue #908, opened 2022-01-07
Parse extensions TLS vectors into a `struct` instead of a Vec


briansmith:
Currently we parse ClientHello extensions into a `Vec<ServerExtension>`, where `ServerExtension` is an enum like this:
```
pub enum ServerExtension {
    ECPointFormats(ECPointFormatList),
    ServerNameAck,
    SessionTicketAck,
    RenegotiationInfo(PayloadU8),
    Protocols(ProtocolNameList),
    KeyShare(KeyShareEntry),
    PresharedKey(u16),
    ExtendedMasterSecretAck,
    CertificateStatusAck,
    SignedCertificateTimestamp(SCTList),
    SupportedVersions(ProtocolVersion),
    TransportParameters(Vec<u8>),
    TransportParametersDraft(Vec<u8>),
    EarlyData,
    Unknown(UnknownExtension),
}
```

 Then we have a bunch of logic for iterating through this vector and pulling out extensions by ID. Instead of constructing this `Vec<ServerExtension>`, we could instead define a struct like this:
```
struct ClientHelloExtensions {
    ec_point_formats: Option<ECPointFormatList>,
    server_name: Option<()>,
    session_ticket: Option<()>,
    renegotiation_info: Option<PayloadU8>,
    protocols: Option<ProtocolNameList>,
    key_share: Option<KeyShareEntry>,
    preshared_key: Option<u16>,
    extended_msater_secret: Option<()>,
    certificate_status: Option<()>,
    signed_certificate_timestamp: Option<SCTList>,
    supported_versions: Option<ProtocolVersion>,
    transport_parameters: Option<Vec<u8>>,
    transport_parameters_draft: Option<Vec<u8>>,
    early_data: Option<()>,
    unknown: HashSet<ExtensionID>,
}
```
There is an efficient way to parse the extension list into this structure and automatically detect duplicates; look at how the webpki crate parses X.509 extensions using that technique.

This eliminates that outer `Vec`. In conjunction with the suggestions in #906 and #907, most intermediate (small) allocations during parsing should be eliminated.

Note that the reason we'd have `unknown: HashSet<ExtensionID>` is to be able to detect duplicates of extensions we don't know about in an efficient way. AFAICT, we are not required to detect duplicates in extensions we don't look at, and it seems like an unnecessary waste to try to do so, so we could then eliminate the `HashSet`, though that would be an optional aspect of this proposal.

briansmith:
[...]

PR #1475, opened 2023-09-14
Improve TLS extension representation

ctz:
~~not worth reviewing yet:~~

- ~~I've broken ECH~~ (now fixed)
- ~~I've commented out a slab of tests~~
- ~~For other tests, I'm not super happy that they mean making the extension representation "public"~~

this is re #908


djc:
I skimmed this a few days ago, and feel obliged to ask: what's the value of the macro here?

ctz:
> I skimmed this a few days ago, and feel obliged to ask: what's the value of the macro here?

At the minute it is mostly saving on writing quite a large number of lines of repeated code, mapping between fields on the struct and `ExtensionType` values.

rustls/src/client/tls13.rs
```diff
@@ -324,7 +324,9 @@ pub(super) fn fill_in_psk_binder(
     let real_binder = key_schedule.resumption_psk_binder_key_and_sign_verify_data(&handshake_hash);
 
     if let HandshakePayload::ClientHello(ch) = &mut hmp.payload {
-        ch.set_psk_binder(real_binder.as_ref());
+        if let Some(psk_offer) = &mut ch.extensions.preshared_key_offer {
+            psk_offer.binders[0] = PresharedKeyBinder::from(real_binder.as_ref().to_vec());
```
djc, line 334: Nit: document why `[0]` never panics?

rustls/src/msgs/handshake.rs
```diff
@@ -548,6 +528,32 @@ impl TlsListElement for CertificateCompressionAlgorithm {
...
+/// This is smaller than `ClientExtensions`, as it only contains the extensions
+/// we need to vary between different protocols (eg, TCP-TLS versus QUIC).
+#[derive(Clone, Default)]
+pub(crate) struct ClientExtensionsTemplate<'a> {
+    /// QUIC transport parameters (RFC9001 prior to draft 33)
+    pub(crate) transport_parameters_draft: Option<Payload<'a>>,
+
+    /// QUIC transport parameters (RFC9001)
+    pub(crate) transport_parameters: Option<Payload<'a>>,
+}
```
djc: We never have both, right? `Template` seems like an overly generic name for this (at least at this commit).

rustls/src/msgs/handshake.rs
```diff
@@ -219,15 +219,114 @@ impl TlsListElement for SignatureScheme {
     const SIZE_LEN: ListLength = ListLength::U16;
 }
 
+/// Lossy decoding for a `ServerName` extension to a single `DnsName`
 #[derive(Clone, Debug)]
-pub(crate) enum ServerNamePayload {
+pub(crate) enum LossyDnsName<'a> {
```
djc, line 224: Don't love this name -- how about something like `ServerNameCandidate`?

rustls/src/msgs/handshake.rs
```diff
@@ -2288,58 +2288,55 @@ impl Codec<'_> for NewSessionTicketPayload {
...
+                    if !unknown_extensions.insert(u16::from(typ)) {
+                        return Err(InvalidMessage::DuplicateExtension);
+                    }
+                }
+            };
+        }
 
-        sub.expect_empty("NewSessionTicketExtension")
-            .map(|_| ext)
+        Ok(out)
```
djc, line 2288: It looks like there's a decent amount of repetition of this stuff, can we abstract this out over some generic trait?

rustls/src/msgs/handshake.rs
```diff
@@ -962,127 +962,157 @@ impl<'a> Codec<'a> for ClientSessionTicket {
...
+
+        /// Encrypted inner client hello response (draft-ietf-tls-esni)
+        ExtensionType::EncryptedClientHello =>
+            pub(crate) encrypted_client_hello_ack: Option<ServerEncryptedClientHello>,
+    } + {
+        pub(crate) unknown_extensions: BTreeSet<u16>,
+    }
+}
+
+impl ServerExtensions<'_> {
```
djc, line 1089: Couldn't the macro generate these `into_owned()` impls?

rustls/src/msgs/handshake.rs
```diff
@@ -388,22 +388,18 @@ impl AsRef<[u8]> for SingleProtocolName {
 
 // --- TLS 1.3 Key shares ---
 #[derive(Clone, Debug)]
-pub struct KeyShareEntry {
+pub(crate) struct KeyShareEntry {
```
djc, line 391: I think these privatizations could/should be part of the earlier commit that makes their being public unnecessary?

djc (commented):
Sorry for a much delayed first pass.

I'm left wondering here if there are more options to split this up: for example, can we have a separate PR that introduces `LossyDnsName`, or one reorganizing pre-shared key support, without the wholesale conversion for client/server extensions?

ctz:
> I'm left wondering here if there are more options to split this up: for example, could we have a separate PR that introduces `LossyDnsName`, or one reorganizing pre-shared key support, without the wholesale conversion for client/server extensions?

Good idea, I'm going to split out some of these commits into a separate PR.

rustls/src/msgs/handshake.rs
```diff
@@ -1145,7 +1019,7 @@ pub struct ClientHelloPayload {
     pub session_id: SessionId,
     pub cipher_suites: Vec<CipherSuite>,
     pub compression_methods: Vec<Compression>,
-    pub extensions: Vec<ClientExtension>,
+    pub extensions: Box<ClientExtensions<'static>>,
```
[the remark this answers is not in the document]
ctz: `ClientExtensions` is unfortunately quite large. 600 bytes currently, and that is after earlier representation improvements (eg, `Option<SupportedProtocolVersions>` introduced earlier in 2caa30488178ebd08a4cfd6763d659df34c81dd9 is 2 bytes, from 24 earlier).

djc:
@ctz let me know if/when this is ready for another round of review.

rustls/src/client/tls13.rs
ctz: added a comment, and some debug asserts.

rustls/src/msgs/handshake.rs
ctz: This fell away and remains `ServerNamePayload`

rustls/src/msgs/handshake.rs
ctz: landed in earlier PR.

rustls/src/msgs/handshake.rs
ctz: changing this to `ClientExtensionsInput` (with an enum for the quic parameters); grateful for any better name ideas

rustls/src/msgs/handshake.rs
ctz: Maybe theoretically? I don't seem to be smart enough to write such a macro though (the issue is replacing a `ty` matcher with something else which can match (eg) `Vec<T>` like normal, but also `T<'a>` and discern the two, such that the lifetime case can output `.map(|x| x.into_owned())`.)

rustls/src/msgs/handshake.rs
```diff
@@ -265,6 +262,15 @@ pub struct ServerName {
     pub(crate) payload: ServerNamePayload,
 }
 
+impl ServerName {
```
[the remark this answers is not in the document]
ctz: (landed in earlier PR)

rustls/src/msgs/handshake.rs
djc: Now that you've got the enum for the QUIC parameters variants, do you still need the intermediate `ClientExtensionsInput` type, or can you just use the enum directly for all the internal API?

ctz: i'm thinking that the `alpn_protocols: Vec<Vec<u8>>,` parameter (eg in `client::hs::start_handshake`) also belongs in this struct

ctz:
> @ctz let me know if/when this is ready for another round of review.

OK, this is ready for another round. Not 100% happy with how many copies of ALPN data are going on, but that is probably an issue for another PR.

rustls/src/error.rs
```diff
@@ -194,6 +194,8 @@ pub enum InvalidMessage {
     IllegalEmptyList(&'static str),
     /// A peer sent an empty value, but a non-empty value is required.
     IllegalEmptyValue,
+    /// A peer sent a message where a given extension type was repeated
+    DuplicateExtension,
```
djc: Encountering this in the wild seems pretty unlikely, right? Is it worth putting a `&'static str` in here that describes which extension is duplicated?

rustls/src/msgs/handshake.rs
```diff
@@ -686,6 +686,272 @@ impl TlsListElement for CertificateCompressionAlgorithm {
...
+    /// - First, extensions not otherwise dealt with by other cases.
+    ///   These are encoded in random order, controlled by `self.order_seed`,
+    ///   and this is the set of extensions returned by this function.
+    ///
+    /// - Second, extensions named in `self.contiguous_extensions`, in the order
+    ///   given by that field.
+    ///
+    /// - Lastly, any ECH and PSK extensions (in that order).  These
+    ///   are required to be last by the standard.
+    fn order_insensitive_extensions_in_random_order(&self) -> Vec<ExtensionType> {
```
djc, line 966: Nit: move this below `used_extensions_in_encoding_order()`?

rustls/src/msgs/handshake.rs
```diff
@@ -686,6 +686,272 @@ impl TlsListElement for CertificateCompressionAlgorithm {
...
+
+        order.sort_by_cached_key(|new_ext| {
+            let seed = ((self.order_seed as u32) << 16) | (u16::from(*new_ext) as u32);
+            low_quality_integer_hash(seed)
+        });
+
+        order
+    }
+
+    pub(crate) fn used_extensions_in_encoding_order(&self) -> Vec<ExtensionType> {
```
djc: Do macros give us the length of a repeated input? Could we define `const NUM_KNOWN_EXTENSIONS: usize = _;` and then have this return `[ExtensionType; NUM_KNOWN_EXTENSIONS]` instead of `Vec`?

rustls/src/msgs/handshake.rs
```diff
@@ -863,6 +1129,22 @@ pub(crate) enum ClientSessionTicket {
...
+        }
+    }
+
+    fn read(r: &mut Reader<'a>) -> Result<Self, InvalidMessage> {
+        Ok(match r.left() {
+            0 => Self::Request,
+            _ => Self::Offer(Payload::read(r).into_owned()),
+        })
+    }
+}
```
djc: Why didn't we have this previously? Might be nice to extract it into a separate commit.

rustls/src/msgs/handshake_test.rs
```diff
@@ -377,6 +378,163 @@ fn client_has_duplicate_extensions_works() {
     assert!(!chp.has_duplicate_extension());
 }
 
+#[test]
+fn clientextensions_basics() {
```
djc: Tiny nit: I continue to find these test names that mash together words distracting. IMO it would be more idiomatic to spell type names with word separators, as in `client_extensions_basics`.

rustls/src/msgs/macros.rs
```diff
@@ -83,3 +83,162 @@ macro_rules! enum_builder {
...
+            )+
+            $($(
+              $(#[$meta_attr])*
+              $meta_vis $meta_slot: $meta_ty,
+            )+)*
+        }
+
+        impl<'a> $struct_name<'a> {
+            /// Decode `r` as `T` into `out`, only if `out` is `None`.
+            fn read_once<T>(r: &mut Reader<'a>, out: &mut Option<T>) -> Result<(), InvalidMessage>
```
djc: Nit: looks like `read_once()` should go below `read_extension_body()`.

rustls/src/msgs/macros.rs
```diff
@@ -83,3 +83,162 @@ macro_rules! enum_builder {
...
+
+                Ok(true)
+            }
+
+            /// Encode one extension body for `typ` into `output`.
+            ///
+            /// Adds nothing to `output` if `typ` is absent from this
+            /// struct, either because it is `None` or unhandled by
+            /// this struct.
+            fn encode_one_extension(
```
djc: Nits: suggest tightening up these method names a bit, maybe `encode_one()` / `collect_used()` / `clone_one()`? Seems to encode as much meaning (that can't be inferred from context) with one less "token".

rustls/src/client/ech.rs
```diff
@@ -202,7 +202,7 @@ impl EchGreaseConfig {
         secure_random: &'static dyn SecureRandom,
         inner_name: ServerName<'static>,
         outer_hello: &ClientHelloPayload,
-    ) -> Result<ClientExtension, Error> {
+    ) -> Result<EncryptedClientHello, Error> {
```
djc: Nit: consider extracting these simple/straightforward changes into a separate commit.

rustls/src/msgs/handshake.rs
```diff
@@ -548,6 +548,272 @@ impl TlsListElement for CertificateCompressionAlgorithm {
     const SIZE_LEN: ListLength = ListLength::U8;
 }
 
+extension_struct! {
```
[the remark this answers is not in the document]
djc: I am still confused by the split between the "Introduce ClientExtensions type" and "Move to using ClientExtensions struct" commits. Are these actually independent? Why/how?

(See also the temporary `#[allow(dead_code)]` attributes.)

rustls/src/client/hs.rs
```diff
@@ -1274,19 +1262,3 @@ impl Deref for ClientSessionValue {
         self.common()
     }
 }
-
-fn low_quality_integer_hash(mut x: u32) -> u32 {
```
djc, line 1276: Nit: ideally this is moved in a separate commit instead of a copy appearing and then this instance disappearing.

rustls/src/server/hs.rs
```diff
@@ -708,7 +731,7 @@ pub(super) fn process_client_hello<'m>(
     // but then act like the client sent no `server_name` extension.
     //
     // [RFC6066]: https://datatracker.ietf.org/doc/html/rfc6066#section-3
-    let sni: Option<DnsName<'_>> = match client_hello.sni_extension() {
+    let sni: Option<DnsName<'_>> = match &client_hello.extensions.server_name {
```
djc: Nit: it looks like this type annotation is unnecessary, remove it while you're in here?

rustls/src/server/tls12.rs
```diff
@@ -74,7 +74,11 @@ mod client_hello {
             // -- TLS1.2 only from hereon in --
             self.transcript.add_message(chm);
 
-            if client_hello.ems_support_offered() {
+            if client_hello
+                .extensions
```
djc: Maybe we should add an `impl Deref for ClientHello { type Target = ClientExtensions; }`?

rustls/src/client/client_conn.rs
```diff
@@ -880,21 +878,36 @@ impl UnbufferedClientConnection {
     /// Make a new ClientConnection. `config` controls how we behave in the TLS protocol, `name` is
     /// the name of the server we want to talk to.
     pub fn new(config: Arc<ClientConfig>, name: ServerName<'static>) -> Result<Self, Error> {
-        Self::new_with_alpn(config.clone(), name, config.alpn_protocols.clone())
+        Self::new_with_extensions(
+            config.clone(),
+            name,
+            ClientExtensionsInput::from_alpn_protocols(config.alpn_protocols.clone()),
```
djc, line 884: Suggest splitting the `ProtocolName` vs ALPN into `ClientExtensionsInput` into 2 separate commits.

rustls/src/msgs/handshake.rs
```diff
@@ -1607,84 +1607,68 @@ impl TlsListElement for CertificateDer<'_> {
...
+            let typ = ExtensionType::read(&mut sub)?;
+            let len = usize::from(u16::read(&mut sub)?);
+            let mut ext_body = sub.sub(len)?;
+            match out.read_extension_body(typ, &mut ext_body)? {
+                true => ext_body.expect_empty("CertificateExtension")?,
+                false => {
+                    return Err(InvalidMessage::UnknownCertificateExtension);
+                }
+            };
+        }
```
djc: Feels like we could deduplicate this logic?

rustls/src/server/tls12.rs
ctz: I feel that would be less clear -- the relationship here is "has-a" rather than "is-a". Could standardise on calling the field `exts` if that helps -- or maybe that is addressing a different point to the one you're making here.

rustls/src/msgs/handshake.rs
ctz: -> #2482

rustls/src/client/client_conn.rs
ctz: -> #2482

rustls/src/error.rs
ctz: Have added a `u16` to this naming the extension.

rustls/src/client/ech.rs
ctz: -> #2482

rustls/src/msgs/handshake.rs
ctz: Extracted into a `fn read_one()` on the generated extensions structs.

rustls/src/msgs/handshake.rs
ctz: Not directly addressing this question, but I've added a temporary commit at the end that turns `fn collect_used() -> Vec<ExtensionType>` into `fn iter_used() -> impl Iterator<Item=ExtensionType>`

Probably best not to think too much about the type generated inside that function, which is -- for your amusement -- 

<details>
<summary>click me if you dare</summary>

```
core::iter::adapters::chain::Chain<
  core::iter::adapters::chain::Chain<
    core::iter::adapters::chain::Chain<
      core::iter::adapters::chain::Chain<
        core::iter::adapters::chain::Chain<
          core::iter::adapters::chain::Chain<
            core::iter::adapters::chain::Chain<
              core::iter::adapters::chain::Chain<
                core::iter::adapters::chain::Chain<
                  core::iter::adapters::chain::Chain<
                    core::iter::adapters::chain::Chain<
                      core::iter::adapters::chain::Chain<
                        core::iter::adapters::chain::Chain<
                          core::iter::adapters::chain::Chain<
                            core::iter::adapters::chain::Chain<
                              core::iter::adapters::chain::Chain<
                                core::iter::adapters::chain::Chain<
                                  core::iter::adapters::chain::Chain<
                                    core::iter::adapters::chain::Chain<
                                      core::iter::adapters::chain::Chain<
                                        core::iter::adapters::chain::Chain<
                                          core::iter::adapters::chain::Chain<
                                            core::iter::adapters::chain::Chain<
                                              core::iter::sources::empty::Empty<
                                                rustls::msgs::enums::ExtensionType>
                                              , core::iter::adapters::map::Map<
                                                core::option::IntoIter<
                                                  &rustls::msgs::handshake::ServerNamePayload>
                                                , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                              >
                                            , core::iter::adapters::map::Map<
                                              core::option::IntoIter<
                                                &rustls::msgs::handshake::CertificateStatusRequest>
                                              , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                            >
                                          , core::iter::adapters::map::Map<
                                            core::option::IntoIter<
                                              &alloc::vec::Vec<
                                                rustls::msgs::enums::NamedGroup>
                                              >
                                            , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                          >
                                        , core::iter::adapters::map::Map<
                                          core::option::IntoIter<
                                            &alloc::vec::Vec<
                                              rustls::msgs::enums::ECPointFormat>
                                            >
                                          , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                        >
                                      , core::iter::adapters::map::Map<
                                        core::option::IntoIter<
                                          &alloc::vec::Vec<
                                            rustls::enums::SignatureScheme>
                                          >
                                        , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                      >
                                    , core::iter::adapters::map::Map<
                                      core::option::IntoIter<
                                        &alloc::vec::Vec<
                                          rustls::msgs::handshake::ProtocolName>
                                        >
                                      , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                    >
                                  , core::iter::adapters::map::Map<
                                    core::option::IntoIter<
                                      &alloc::vec::Vec<
                                        rustls::enums::CertificateType>
                                      >
                                    , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                  >
                                , core::iter::adapters::map::Map<
                                  core::option::IntoIter<
                                    &alloc::vec::Vec<
                                      rustls::enums::CertificateType>
                                    >
                                  , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                                >
                              , core::iter::adapters::map::Map<
                                core::option::IntoIter<
                                  &()>
                                , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                              >
                            , core::iter::adapters::map::Map<
                              core::option::IntoIter<
                                &alloc::vec::Vec<
                                  rustls::enums::CertificateCompressionAlgorithm>
                                >
                              , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                            >
                          , core::iter::adapters::map::Map<
                            core::option::IntoIter<
                              &rustls::msgs::handshake::ClientSessionTicket>
                            , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                          >
                        , core::iter::adapters::map::Map<
                          core::option::IntoIter<
                            &rustls::msgs::handshake::PresharedKeyOffer>
                          , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                        >
                      , core::iter::adapters::map::Map<
                        core::option::IntoIter<
                          &()>
                        , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                      >
                    , core::iter::adapters::map::Map<
                      core::option::IntoIter<
                        &rustls::msgs::handshake::SupportedProtocolVersions>
                      , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                    >
                  , core::iter::adapters::map::Map<
                    core::option::IntoIter<
                      &rustls::msgs::base::PayloadU16<
                        rustls::msgs::base::NonEmpty>
                      >
                    , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                  >
                , core::iter::adapters::map::Map<
                  core::option::IntoIter<
                    &alloc::vec::Vec<
                      rustls::msgs::enums::PskKeyExchangeMode>
                    >
                  , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
                >
              , core::iter::adapters::map::Map<
                core::option::IntoIter<
                  &alloc::vec::Vec<
                    rustls::msgs::handshake::DistinguishedName>
                  >
                , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
              >
            , core::iter::adapters::map::Map<
              core::option::IntoIter<
                &alloc::vec::Vec<
                  rustls::msgs::handshake::KeyShareEntry>
                >
              , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
            >
          , core::iter::adapters::map::Map<
            core::option::IntoIter<
              &rustls::msgs::base::Payload>
            , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
          >
        , core::iter::adapters::map::Map<
          core::option::IntoIter<
            &rustls::msgs::base::PayloadU8>
          , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
        >
      , core::iter::adapters::map::Map<
        core::option::IntoIter<
          &rustls::msgs::base::Payload>
        , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
      >
    , core::iter::adapters::map::Map<
      core::option::IntoIter<
        &rustls::msgs::handshake::EncryptedClientHello>
      , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
    >
  , core::iter::adapters::map::Map<
    core::option::IntoIter<
      &alloc::vec::Vec<
        rustls::msgs::enums::ExtensionType>
      >
    , rustls::msgs::handshake::ClientExtensions::iter_used::{{closure}}>
  >
```

</details>

ctz: OK, the performance of the iterator version sucks. That's a shame.

rustls/src/client/hs.rs
ctz: I think eventually I will squash these three commits together.

rustls/src/msgs/handshake.rs
ctz: I've looked at this a lot today, using #2483, and found: surprisingly, every iterator-based or array-based formulation I could come up with was worse than making a vec.

rustls/src/server/tls12.rs
djc: Some of these access chains get pretty long, and the `.extensions` field seems low-value. I agree that `Deref` is not an awesome fit, but I think could make this a little more ergonomic at little cost (since the types involved are pretty private anyway).

ctz:
Merged in:

- #2441
- #2447
- #2457
- #2462 
- #2472 
- #2473
- #2478
- #2482
- #2502 
- #2508 
- #2509

s-arash:
Hi @ctz. With this work done, is it a good time to start working on #2403?

ctz:
Yes. If you haven't found it already here's where I was expecting this change to (mostly) go: https://github.com/rustls/rustls/blob/c84675e34b525b7597c5632202e483c0b68f7e1c/rustls/src/server/hs.rs#L725-L735

s-arash:
Great! Thanks @ctz.

Issue #1386, opened 2023-08-02
Redundant type names via re-exports


jsha:
With a breaking release coming up, it might be a good time to address this minor issue. For several types, there are multiple names in the public API they can be referred to. This is a problem because someone reading the code doesn't automatically know that `rustls::ServerConfig` is the same as `rustls::server::ServerConfig`. I think each of the items listed under [Re-exports](https://docs.rs/rustls/latest/rustls/#reexports) at the top level has this property:

```
pub use client::ClientConfig;
pub use client::ClientConnection;
pub use client::ServerName;
pub use server::ServerConfig;
pub use server::ServerConnection;
```

The unreleased version in `main` adds more types with this property:

```
pub use crate::crypto::ring::Ticketer;
pub use crate::crypto::ring::SupportedKxGroup;
pub use crate::crypto::ring::ALL_KX_GROUPS;
pub use crypto::ring::kx_group;
```

I think in each of these cases it makes sense to remove the single-type re-export in favor of having users use the fully-qualified name.

djc:
I think I set it up this way; I haven't ever noticed that this was a problem for people. Personally I think re-exports at the top level are useful (less typing for the most common use cases). I put out a poll that seems to mostly agree that this is problematic, though: https://hachyderm.io/@djc/110824267001415573.

> I think in each of these cases it makes sense to remove the single-type re-export in favor of having users use the fully-qualified name.

I guess it's fine, though I don't find the prospect of forcing everyone to write an extra module name enticing.

aochagavia:
I'm obviously biased because of having used Rust for a long time, but I consider re-exports a Rust idiom to indicate the most interesting types in a crate (i.e. the types that I'm likely to need as a user). Even the standard library uses re-exports:

![image](https://github.com/rustls/rustls/assets/5196584/77a6383c-a72a-4f29-be26-4329f13bfd8b)

From the types mentioned above, I'd think we can keep the original ones, but that the new ones shouldn't be re-exported at the top level (for instance, there might be multiple `ALL_KG_GROUPS` constants, one per crypto implementation, and you probably don't want to privilege one of them).

djc:
> From the types mentioned above, I'd think we can keep the original ones, but that the new ones shouldn't be re-exported at the top level (for instance, there might be multiple `ALL_KG_GROUPS` constants, one per crypto implementation, and you probably don't want to privilege one of them).

That would also be my preference.

jsha:
[...]

PR #1477, opened 2023-09-14
Remove top-level re-exports

cpu:
This branch removes all top-level re-exports except for the "paved path" re-exports we expect all-to-most users will be using. 

An [informal poll](https://hachyderm.io/@djc/110824267001415573) indicates users find it confusing when there are multiple names for the same things. See https://github.com/rustls/rustls/issues/1386 for more discussion.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -266,6 +266,15 @@ import the symbol name under an alias, or if the parent module name is short,
 using a one-level qualified path. E.g. for a crate with a local `Error` type,
 prefer to `import std::error::Error as StdError`.
 
+### Exports
+
+We prefer to export types under a single name, avoiding re-exporting types from
+the top-level `lib.rs`. The exception to this are "paved path" exports that we
+expect every user will need. The canonical example of such types are 
+`client::ClientConfig` and `server::ServerConfig`. In general this sort of type
+is rare and most new types should be exported only from the module in which they 
+are defined.
+
 ### Misc
 
 #### Numeric literals
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -14,10 +14,10 @@ use itertools::Itertools;
 use rayon::iter::Either;
 use rayon::prelude::*;
 use rustls::client::Resumption;
+use rustls::crypto::ring::Ticketer;
 use rustls::server::{NoServerSessionStorage, ServerSessionMemoryCache, WebPkiClientVerifier};
 use rustls::{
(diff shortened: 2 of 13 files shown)


djc:
It makes sense to me to relegate the `crypto` parts to a module, as I think these will be rarely used (assuming most users will pick the default crypto provider or select a crypto provider via a feature flag).

For the others, I really don't see removing the top-level re-exports as a good idea. These have been at the top-level forever, so we'd be putting every downstream user through a bunch of churn for the sole goal of removing the re-exports.

I suppose we could alternatively remove the "original" re-exports from the `client`/`server` modules instead, which I personally also find unattractive because the documentation for those modules provides a good place to find everything related to that particular context.

So, after considering all the context, I would prefer to keep the situation as-is for `ClientConnection`, `ClientConfig`, `ServerName`, `ServerConfig` and `ServerConnection`.

cpu:
>  I would prefer to keep the situation as-is for ClientConnection, ClientConfig, ServerName, ServerConfig and ServerConnection.

@ctz @jsha What do you think of this proposal?

ctz:
I think of the API as being in two groups -- let's call them "paved path" and "detail". "Paved path" would be types that _every_ user will need[^1], and I think it makes sense for these types to be front-and-centre at the top of the crate path hierarchy. "Detail" would be for complex uses, or types that need to be public but aren't directly referred to by name[^2], and everything else.

For these commits, I think the first 6 are "detail" that deserve to be lower down in the hierarchy rather than the top level. The rest of the commits I think are "paved path" types that deserve a place at the top of the hierarchy, but shouldn't be lost further down.

[^1]: I'm assuming every user is interested in being a TLS server and client -- probably not accurate.
[^2]: example: `rustls::client::WantsClientCert` -- needs to be public, but would never be named

cpu:
@ctz What you describe makes sense to me, and I would be OK with leaving the `server`/`client` re-exports if we additionally moved some of your comment to a `CONTRIBUTING.md` style guide update to help evaluate additional re-exports in the future (_happy to do that in this branch, or as a follow-up_).

I'd like Jsha to have a chance to weigh in before I make updates to the branch since he opened #1386. Let's table this until then?

jsha:
Yes this makes sense to me. As I understand it: we would keep the five
"paved path" types we're talking about exported in two places, because even
across a breaking release, it's not worth breaking someone's code over. The
types that aren't yet in a release would get a single name (under a
module). And for new "paved path" types, that's presumably a discussion for
#1386 / a contibuting.md update. 👍

djc (approved):
Thanks for getting this done!

PR #1479, opened 2023-09-15
Update the readme

bdaehlie:
- Don't list dependencies in the headline
- Remove relativistic language like "mature" and "widely"
- Remove possible future features as it is incomplete and thus misleading, should eventually replace with a roadmap
- Make it clear that Rustls provides no unsafe features *by default*
- remove self-signed certs and compression from non-features list because it's nuanced and we don't want to turn people away
- Add a list of project leadership

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,9 @@ Rustls is a modern TLS library written in Rust.
 </p>
 
 # Status
-Rustls is ready for production use. We aim to maintain reasonable API surface stability but the API may evolve as we make changes to accomodate new features or performance improvements.
+Rustls is ready for production use. We aim to maintain reasonable API surface stability
+but the API may evolve as we make changes to accomodate new features or performance
+improvements.
 
 If you'd like to help out, please see [CONTRIBUTING.md](CONTRIBUTING.md).
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Rustls is a modern TLS library written in Rust.
 </p>
 
 # Status
+
 Rustls is used in production at many organizations and projects. We aim to maintain
 reasonable API surface stability but the API may evolve as we make changes to accomodate
 new features or performance improvements.
@@ -27,6 +28,7 @@ Release history can be found [on GitHub](https://github.com/rustls/rustls/releas
 https://docs.rs/rustls/
 
 # Approach
+
 Rustls is a TLS library that aims to provide a good level of cryptographic security,
 requires no configuration to achieve that security, and provides no unsafe features or
 obsolete cryptography by default.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,13 +3,13 @@
...
 <p align="center">
-Rustls is a modern TLS library written in Rust.  It uses <a href = "https://github.com/briansmith/ring"><em>ring</em></a> for cryptography and <a href = "https://github.com/rustls/webpki">webpki</a> for certificate
-verification.
+Rustls is a modern TLS library written in Rust.
 </p>
 
 # Status
-Rustls is mature and widely used. While most of the API surface is stable, we expect the next
-few releases will make further changes as needed to accomodate new features or performance improvements.
+Rustls is ready for production use. We aim to maintain reasonable API surface stability
```
djc, line 10: Maybe strengthen this to say something more like "Rustls is used in production at many organizations and projects."?

djc (approved):

README.md
```diff
@@ -72,10 +67,8 @@ rustls does not and will not support:
 * Ciphersuites without forward secrecy.
 * Renegotiation.
 * Kerberos.
-* Compression.
 * Discrete-log Diffie-Hellman.
 * Automatic protocol version downgrade.
-* Self-signed certificates.
```
cpu, line 78: I think we should keep these non-features. Folks ask about support for these things otherwise. The self-signed certificate non-support item was added recently for this reason: https://github.com/rustls/rustls/pull/1382

I think any nuance that applies to whether compression or self-signed certificates are supported likely applies to the other non-features like discrete-log DH, and renegotiation, no?

README.md
ctz: I wonder if we could go further and give a few concrete examples? servo? @olix0r could we mention linkerd?

djc: Cloudflare? Google?

README.md
ctz: Like I said in that other PR, I think people will read "self-signed certificates" here and conclude that rustls cannot be configured to turn off certificate checking. And I think if we start qualifying some of these items the list will become a bit of an exposition which doesn't really fit in a README.

Maybe we could add a section to https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html which covers the detail.

README.md
bdaehlie: I will add something like what @djc suggested but I think we should avoid listing particular consumers here, at least for now.

README.md
cpu: > Maybe we could add a section to https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html which covers the detail.

That sounds reasonable to me.

README.md
```diff
@@ -222,6 +215,12 @@ hello world
 ^C
 ```
 
+# Project Leadership
```
cpu, line 218: Three letter usernames not required, but encouraged :laughing:

cpu (approved):

bdaehlie:
I think this is ready to merge.

I still think we should remove self signed certificates from the prohibited list since some people want it and Rustls does allow it and I don't want to turn people away on that point, but we can do that in another PR.

ctz (approved):

README.md
olix0r: @ctz Yes it's fine to mention that Linkerd uses rustls :)

PR #1480, opened 2023-09-16
docs: clarify self-signed certificate limitation. 

bdaehlie:
That is a limitation of webpki and should be noted there.

See this [PR for webpki](https://github.com/rustls/webpki/pull/180).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,11 @@ rustls does not and will not support:
 * TLS 1.2 protocol compression.
 * Discrete-log Diffie-Hellman.
 * Automatic protocol version downgrade.
-* Self-signed certificates.
+* Using CA certificates directly to authenticate a server/client (often called "self-signed
+certificates"). _Rustls' default certificate verifier does not support using a trust anchor as
+both a CA certificate and an end-entity certificate in order to limit complexity and risk in
+path building. While dangerous, all authentication can be turned off if required --
+see the [example code](https://github.com/rustls/rustls/blob/992e2364a006b2e84a8cf6a7c3eaf0bdb773c9de/examples/src/bin/tlsclient-mio.rs#L318)_.
 
 There are plenty of other libraries that provide these features should you
 need them.
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -43,7 +43,11 @@
 //! * TLS 1.2 protocol compression.
 //! * Discrete-log Diffie-Hellman.
 //! * Automatic protocol version downgrade.
-//! * Self-signed certificates.
+//! * Using CA certificates directly to authenticate a server/client (often called "self-signed
+//!   certificates"). _Rustls' default certificate verifier does not support using a trust anchor as
+//!   both a CA certificate and an end-entity certificate in order to limit complexity and risk in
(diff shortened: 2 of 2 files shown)


cpu (approved):
Thanks!

djc:
I don't think this is a good change.

For most users, webpki is an implementation detail of the rustls default verifier. Even if/when we make the webpki verifier more explicit in the API (for example, by making callers pass a `WebPkiVerifiers::from(roots)` to the config builder API), for a large majority of users rustls is intimately connected with webpki -- that's probably even more true of the relatively unsophisticated users who don't have a good idea about the trade-offs of self-signed certificates.

Instead, I think we should be upfront about this limitation and its nuances.

ctz:
How about this?

* Using CA certificates directly to authenticate a server/client. _CA certificates should
    only issue other certificates, and cannot be used alone for authentication. All authentication can be turned off -- see our [example code](https://github.com/rustls/rustls/blob/a519303f0c4a76cf3ea746e1bae0707b5001b8e0/examples/src/bin/tlsclient-mio.rs#L322).)_

Note I'm carefully avoiding saying "self-signed", which I think can be used to mean three different things:

1. a standalone CA certificate that has CN=localhost (or something) and therefore authenticates itself.
2. a locally-administered PKI (ie, "i signed this myself, it is self-signed")
3. turn-off-all-authentication-I-don't-care (ie, "don't bother me with this certificate bidness")

bdaehlie:
If you're able to take this over that would be great, thanks.

cpu:
@ctz @djc I updated this branch with the suggestions from above. PTAL

ctz (approved):
(fwiw)

README.md
```diff
@@ -72,7 +72,10 @@ rustls does not and will not support:
 * TLS 1.2 protocol compression.
 * Discrete-log Diffie-Hellman.
 * Automatic protocol version downgrade.
-* Self-signed certificates.
+* Using CA certificates directly to authenticate a server/client (often called "self-signed
+certificates"). _CA certificates should only issue other certificates, and cannot be used
```
djc, line 76: "cannot be used alone for" seems a little awkward? This is a good direction, though!

djc (approved):

README.md
cpu: How about: "CA certificates should only issue other certificates, and are inappropriate for authentication in isolation." ? Less awkward?

djc: I don't really grok what you mean by "are inappropriate for authentication in isolation". I guess the real point is that authenticating a server against a certificate signed by its own private key, that is also marked as a CA, would actually be an okay way to authenticate, but we don't want to bear the extra complexity in webpki for enabling path building (safely) scenarios where this is the case?

In which case maybe we should say: "To limit complexity and risk in our path building implementation, our default verifier implementations do not support using an end-entity certificate as a trust anchor."?

cpu: > In which case maybe we should say: "To limit complexity and risk in our path building implementation, our default verifier implementations do not support using an end-entity certificate as a trust anchor."?

Added text closer to this. WDYT?

djc (approved):

Issue #1437, opened 2023-09-01
reducing Cargo feature list


jsha:
Right now we have two features that control dangerous code: `dangerous_configuration` and `secret_extraction`. We also have started using the cool `--feature-powerset` flag to test various combinations of features:

```
cargo hack check --feature-powerset
```

The size of the powerset increases very fast with number of features, so it's useful to keep that number under control.

I think `dangerous_configuration` and `secret_extraction` could probably stop being features, and we could make those part of the main API, under a new `::dangerous::` module. Using features to gate dangerous things runs into the classic problem of Cargo features: they are additive, so one of your transitive dependencies could enable them for you, which would allow using the gated items without explicitly enabling them.

For someone auditing codebases there are pros and cons. Pro: putting things in a specific module means you can grep for imports of that module in the codebase (and its dependencies). Con: the feature approach in theory allows an auditor to query the evaluated feature set after dependency resolution and see if `dangerous_configuration` is enabled. I'm not sure which Cargo commands one would use but I'm sure it's possible.

There's also the `tls12` feature. In the config builder we try to make sure unused code (like TLS 1.2) can be eliminated by the linker, but we also have the `tls12` Cargo feature. I haven't dug into how painful it would be to fully remove the feature, but it would have some nice benefits. For instance, we have some code that is only active for `#[cfg(not(feature = "tls12"))]`. I believe that's not understood by our coverage measurements.

ctz:
I generally agree with this. I think we should have a rule-of-thumb, that cargo features are only to be used when they correspond to taking different crate dependencies.

> There's also the tls12 feature. In the config builder we try to make sure unused code (like TLS 1.2) can be eliminated by the linker, but we also have the tls12 Cargo feature

This is https://github.com/rustls/rustls/issues/224

djc:
Once we have pluggable crypto, doesn't that more or less solve the linking issues because everything is routed via the cipher suite types?

ctz:
Now we got rid of `secret_extraction` and `dangerous_configuration` -- and have a plan and separate issue for deleting `tls12` -- shall we close this?

I was also thinking about the `quic` feature, but haven't looked in detail about the implications of removing that.

cpu:
> shall we close this?

SGTM

ctz:
I have looked at removing the `quic` feature today -- I think it is certainly doable and will have a PR for that tomorrow.

PR #1481, opened 2023-09-18
proj: remove dangerous_configuration feature

cpu:
In an effort to reduce our feature list, this commit replaces the `dangerous_configuration` feature flag with separate `danger` modules.

Cargo features are additive, which means transitive dependencies could enable them for you without explicit opt-in. Using obviously named modules will maintain the property that it's easy to grep for imports, but avoids feature flag bloat and the additive downsides.

After discussion we've chosen to not include the webpki verifier and helper functions as part of the dangerous API surface. Functionality for setting a custom verifier, or implementing one to make assertions about verification status, remain marked as dangerous via their module name.

Updates https://github.com/rustls/rustls/issues/1437

### DIFF
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -229,11 +229,6 @@
 //!   messages do not contain secret key data, and so are safe to archive without
 //!   affecting session security.  This feature is in the default set.
 //!
-//! - `dangerous_configuration`: this feature enables a `dangerous()` method on
-//!   `ClientConfig` and `ServerConfig` that allows setting inadvisable options,
-//!   such as replacing the certificate verification process.  Applications
-//!   requesting this feature should be reviewed carefully.
-//!
 //! - `quic`: this feature exposes additional constructors and functions
 //!   for using rustls as a TLS library for QUIC.  See the `quic` module for
 //!   details of these.  You will only need this if you're writing a QUIC
@@ -419,14 +414,15 @@ pub mod client {
     };
     pub use handy::ClientSessionMemoryCache;
 
-    #[cfg(feature = "dangerous_configuration")]
-    pub use crate::verify::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
-    #[cfg(feature = "dangerous_configuration")]
+    /// Dangerous configuration that should be audited and used with extreme care.
+    pub mod danger {
+        pub use super::client_conn::danger::DangerousClientConfig;
+        pub use crate::verify::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
+    }
+
     pub use crate::webpki::{
         verify_server_cert_signed_by_trust_anchor, verify_server_name, WebPkiServerVerifier,
     };
-    #[cfg(feature = "dangerous_configuration")]
-    pub use client_conn::danger::DangerousClientConfig;
 
     pub use crate::msgs::persist::Tls12ClientSessionValue;
     pub use crate::msgs::persist::Tls13ClientSessionValue;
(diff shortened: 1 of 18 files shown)


rustls/src/lib.rs
```diff
@@ -463,12 +458,12 @@ pub mod server {
...
-    pub use crate::dns_name::DnsName;
-    #[cfg(feature = "dangerous_configuration")]
-    pub use crate::verify::{ClientCertVerified, ClientCertVerifier};
-    #[cfg(feature = "dangerous_configuration")]
-    pub use crate::webpki::ParsedCertificate;
+    /// Dangerous configuration that should be audited and used with extreme care.
+    pub mod danger {
+        pub use crate::dns_name::DnsName;
+        pub use crate::verify::{ClientCertVerified, ClientCertVerifier};
+        pub use crate::webpki::ParsedCertificate;
```
ctz, line 465: I think I agree with @djc that these types aren't not inherently dangerous.

rustls/src/lib.rs
```diff
@@ -425,14 +420,14 @@ pub mod client {
...
-    };
-    #[cfg(feature = "dangerous_configuration")]
-    pub use client_conn::danger::DangerousClientConfig;
+    /// Dangerous configuration that should be audited and used with extreme care.
+    pub mod danger {
+        pub use super::client_conn::danger::DangerousClientConfig;
+        pub use crate::verify::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
+        pub use crate::webpki::{
+            verify_server_cert_signed_by_trust_anchor, verify_server_name, WebPkiServerVerifier,
+        };
```
ctz, line 429: I think I agree with @djc that these types (and `server::danger`) are not inherently dangerous.

rustls/src/lib.rs
```diff
@@ -425,14 +420,14 @@ pub mod client {
...
-        verify_server_cert_signed_by_trust_anchor, verify_server_name, WebPkiServerVerifier,
-    };
-    #[cfg(feature = "dangerous_configuration")]
-    pub use client_conn::danger::DangerousClientConfig;
+    /// Dangerous configuration that should be audited and used with extreme care.
+    pub mod danger {
+        pub use super::client_conn::danger::DangerousClientConfig;
+        pub use crate::verify::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
+        pub use crate::webpki::{
+            verify_server_cert_signed_by_trust_anchor, verify_server_name, WebPkiServerVerifier,
```
ctz, line 428: Seems odd that `rustls::server::danger::ParsedCertificate` is required to use the `rustls::client::danger::verify_server_name` API. (This isn't an issue introduced in this PR, just making a note here.)

djc: This is presumably because we want to reuse the same `ParsedCertificate` across calls to `verify_server_cert_signed_by_trust_anchor()` and `verify_server_name()`, right? I think that makes sense?

djc:
> The helper functions give me some pause. Using those piecemeal without a strong understanding of what parts of the overall verification process they do/don't cover is, in my mind, an on-ramp on the [highway to the danger-zone](https://www.youtube.com/watch?v=siwpn14IE7E).

Yes, there is some nuance. On the one hand you can argue that these expose a safe interface for what they're trying to do, on the other hand you can argue that what they do is a limited part of what needs to be done, so using them doesn't promise that the caller will be safe from risks. With `ServerCertVerifier`, the implementer must ascertain that their implementation is safe, this is tricky, and thus we want to "hide" this interface. With `verify_server_cert_signed_by_trust_anchor()` we promise that the implementation correctly does what it is named for/documented to do, but the caller must ascertain that they additionally do other things.

On balance... if we agree that the trait itself must be in a `danger` module, and these functions are really only useful if you're implementing your own verifier anyway, then it probably doesn't matter a whole lot? The caller is importing from `danger` anyway. But, I still think it might be nice to keep these useful safe building blocks available outside the `danger` API.

rustls/src/lib.rs
cpu: I've removed these types from the `danger `mod.

rustls/src/lib.rs
cpu: I removed `ParsedCertificate` from the `danger` mod.

cpu:
> OK, I will rework to remove them from the danger API.

Done :ballot_box_with_check:

djc (approved):

ctz (approved):

PR #1482, opened 2023-09-19
Test different nightly version selection strategy

ctz:
Prior to this we used `rustup`'s definition of nightly, which is always the latest available nightly for the desired platform.

This moves to using https://github.com/ctz/rust-nightly-but-make-it-fashion/ to choose a nightly version based on known issues tracked over there.

This is a bit gnarly, but succeeds in the goal of letting us back off the nightly version as desired in a semi-automated way.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         rust:
           - stable
           - beta
-          - nightly
+          - ${{ vars.NIGHTLY_VERSION }}
         os: [ubuntu-20.04]
         # but only stable on macos/windows (slower platforms)
         include:
@@ -127,7 +127,9 @@ jobs:
           persist-credentials: false
 
       - name: Install nightly toolchain
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ vars.NIGHTLY_VERSION }}
 
       - name: Install cargo fuzz
         run: cargo install cargo-fuzz
@@ -149,7 +151,9 @@ jobs:
           persist-credentials: false
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ vars.NIGHTLY_VERSION }}
 
       - name: Smoke-test benchmark program
         run: cargo run --release --example bench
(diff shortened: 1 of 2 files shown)


cpu:
This seems like a pretty good workaround, thanks @ctz 

If this is the approach we go with I think the [rust-nightly-but-make-it-fashion](https://github.com/ctz/rust-nightly-but-make-it-fashion/) README could use some text about the expected format that issues on that repo need to take so that other contributors could tweak the version when required. Right now that repo only says:
> Issues in this repository of a certain format document which nightly versions we think currently work.

Separate from that, I'm wondering if the overall process could be simplified. Did you consider whether it might be workable to have the output `nightly-version.txt` maintained as a file checked in directly to the `make-it-fashion` repo? It feels like it might not be very onerous to maintain that manually and would remove the need for the GitHub API work. Dependent repos could fetch it via the raw URL for the resource in the convergence branch. Admittedly having to do some `git` work is more onerous than opening/closing GitHub issues, but still feels reasonable. 

Alternatively, perhaps the [generator script](https://github.com/ctz/rust-nightly-but-make-it-fashion/blob/main/src/main.rs) should be changed so that it only considers issues opened by a Rustls maintainer? It would be unfortunate if a passer by disregards the README warning and files an issue that would cause the Rustls CI to change nightly versions without anyone having the chance to review the change.

djc:
I'm wondering to what extent we really need nightly? We could just get rid of nightly clippy if it's causing too much pain, and use a `--lib` strategy (like we have for MSRV) for other things so we don't need to compile dev-deps with nightly.

What breakages have we seen recently? Obviously this current one is about the time issue.

ctz:
> I'm wondering to what extent we really need nightly?

Could've sworn coverage measurement only worked on nightly, but actually: #1486

ctz:
> Separate from that, I'm wondering if the overall process could be simplified. Did you consider whether it might be workable to have the output `nightly-version.txt` maintained as a file

Yes, in retrospect this is massively over-engineered for what it achieves. An even simpler idea is to have a organisation-level github actions variable and use it everywhere "nightly" appears. This PR is now doing that, but I'm going to close it as #1486 has solved the immediate issue for this repo.

> What breakages have we seen recently? Obviously this current one is about the time issue.

There's also https://github.com/ctz/rust-nightly-but-make-it-fashion/issues/3 which is affecting webpki.

cpu:
>  An even simpler idea is to have a organisation-level github actions variable and use it everywhere "nightly" appears.

Ah cool, I didn't realize actions had that feature :+1:

Issue #1472, opened 2023-09-14
Crypto provider feature guards are not additive


djc:
I have VS Code configured to compile code with `--all-features` by default. After landing #1405 this is showing compilation failures in the new provider-example crate:

* `Tls13CipherSuite` has `pub(crate)` fields that are only enabled for `#[cfg(feature = "quic")]`. This means that other crypto providers can't create `Tls13CipherSuite` values when the `quic` feature is enabled.
* `Tls12AeadAlgorithm` and `Tls13AeadAlgorithm` have `extract_keys()` methods that only appear for `#[cfg(feature = "secret_extraction")]`. This means that also downstream implementations of these traits will also need to guard these trait methods.

This happens because my Rust-Analyzer setup is compiling rustls with all features enabled, but the provider-example crate cannot deal with the additional features -- it can only be compiled when rustls doesn't have these features enabled, but that can never be guaranteed in Cargo's feature-additive dependency graphs.

ctz:
I suggest we address this by:

- making those fields always included; they are not specific to quic. `record_layer` should be enforcing them also; though I'd like to make it enforce them more stringently so it is possible run a test in CI that exercises that code.
- deleting the `secret_extraction` gate feature in accordance with #1437, and make it fallible so providers who don't implement aes-128-gcm, aes-256-gcm, or chacha20poly1305 can refuse.

djc:
> * making those fields always included; they are not specific to quic. `record_layer` should be enforcing them also; though I'd like to make it enforce them more stringently so it is possible run a test in CI that exercises that code.

I agree for the limits. For the `quic: crate::quic::Algorithm` field, that certainly seems specific to QUIC? Do we want to support cipher suite implementations that don't have QUIC support? I think allowing that is reasonable.

> * deleting the `secret_extraction` gate feature in accordance with #1437, and make it fallible so providers who don't implement aes-128-gcm, aes-256-gcm, or chacha20poly1305 can refuse.

Instead, could we replace `ConnectionTrafficSecrets` with a type that contains things more like `HashOutput`? Again the question is, do we want to support cipher suite implementations that don't support secret extraction?

PR #1484, opened 2023-09-19
proj: remove `secret_extraction` feature.

cpu:
Similar to https://github.com/rustls/rustls/pull/1481 this branch removes another feature flag, `secret_extraction`. This functionality is now moved into `danger` modules that emphasize that this feature should be used with care, while avoiding the downsides of feature flags (combinatoric bloat for the powerset, additive nature, etc).

Additionally this branch does a little bit of refactoring to support crate-external providers implementing new AEAD algorithms. In particular we need to make it easier to get at the underlying `Iv` array and to make the `extract_keys` trait fns fallible so that algorithms that don't have a matching `ConnectionTrafficSecrets` variant can return an error when asked to extract keys.

Updates https://github.com/rustls/rustls/issues/1437, https://github.com/rustls/rustls/issues/1472.


rustls/src/crypto/cipher.rs
```diff
@@ -16,9 +15,9 @@ pub trait Tls13AeadAlgorithm: Send + Sync {
     /// The length of key in bytes required by `encrypter()` and `decrypter()`.
     fn key_len(&self) -> usize;
 
-    #[cfg(feature = "secret_extraction")]
-    /// Convert the key material from `key`/`iv`, into a `ConnectionTrafficSecrets` item.
-    fn extract_keys(&self, key: AeadKey, iv: Iv) -> ConnectionTrafficSecrets;
+    /// Convert the key material from `key`/`iv`, into a `ConnectionTrafficSecrets` item. May
+    /// return an `Error` if the AEAD algorithm is not a supported variant of `ConnectionTrafficSecrets`.
+    fn extract_keys(&self, key: AeadKey, iv: Iv) -> Result<ConnectionTrafficSecrets, Error>;
```
cpu, line 20: It might make more sense to avoid using `rustls::Error` here in place of a more specific "Unsupported Operation" error. WDYT?

rustls/src/suites.rs
```diff
@@ -216,6 +213,31 @@ pub enum ConnectionTrafficSecrets {
     },
 }
 
+impl ConnectionTrafficSecrets {
```
cpu, line 216: This feels a little bit awkward, but I also thought having `slice_to_array` and `slices_to_arrays` defined in three places wasn't going to cut it. Having the duplication across two crate-internal places was already a bit crummy, adding a third for the provider example is one too many for sure.

rustls/src/suites.rs
```diff
@@ -216,6 +213,31 @@ pub enum ConnectionTrafficSecrets {
...
+            Self::slice_to_array(s),
+            Self::slice_to_array(i),
+        )
+    }
+
+    /// Convert a slice to a fixed sized array, panicking if the slice is not the correct
+    /// constant size length.
+    pub fn slice_to_array<const N: usize>(slice: &[u8]) -> [u8; N] {
+        // this is guaranteed true because `ConnectionTrafficSecrets` items and
+        // `key_block_shape()` are in agreement.
```
cpu, line 235: I think this comment needs to be adjusted, and perhaps indicates that these helpers aren't as generic as they might seem? Open to suggestions for reworking

provider-example/src/aead.rs
```diff
@@ -22,6 +22,10 @@ impl cipher::Tls13AeadAlgorithm for Chacha20Poly1305 {
     fn key_len(&self) -> usize {
         chacha20poly1305::ChaCha20Poly1305::key_size()
     }
+
+    fn extract_keys(&self, _key: AeadKey, _iv: Iv) -> ConnectionTrafficSecrets {
```
djc: How about having the fallibility commit first and then making it required, so that we can have these emit an error directly?

rustls/src/crypto/cipher.rs
djc: Yeah, I think having a specific error probably makes sense here.

rustls/src/crypto/cipher.rs
```diff
@@ -124,6 +124,11 @@ impl Iv {
         iv.0.copy_from_slice(value);
         iv
     }
+
+    /// Return the `Iv`'s underlying byte slice, of precisely `NONCE_LEN` bytes.
+    pub fn value(&self) -> [u8; NONCE_LEN] {
```
djc, line 137: Does it make sense to copy here, or are we generally going to be passing this as a slice to the next API? Maybe this should yield a reference to the `Iv`'s field instead? And if so, maybe using `AsRef` for this makes sense?

rustls/src/suites.rs
djc: I think this is a reasonable approach.

rustls/src/suites.rs
ctz: Thoughts on doing this instead?

```rust
pub enum ConnectionTrafficSecrets {
  Aes128Gcm { key: AeadKey, iv: Iv }
  Aes256Gcm { key: AeadKey, iv: Iv }
  Chacha20Poly1305 { key: AeadKey, iv: Iv }
}
```

It will require a bit of churn in downstream https://github.com/hapsoc/ktls/blob/main/src/ffi.rs#L131

rustls/src/crypto/cipher.rs
cpu: Switched to a specific error.

provider-example/src/aead.rs
cpu: Re-ordered to put the fallibility and `ConnectionTrafficSecrets` changes ahead of everything else.

> so that we can have these emit an error

Once re-ordered since `Chacha20Poly1305` is supported by `ConnectionTrafficSecrets` we don't need to emit an error, just implement the functionality using `ConnectionTrafficSecrets::slice_to_array`.

rustls/src/crypto/cipher.rs
cpu: I switched to implementing `AsRef`.

rustls/src/suites.rs
cpu: > Thoughts on doing this ~instead~ later?

I'm going to run with this but think it might make sense to do separately from removing the `secret_extraction` feature.

cpu: @ctz WDYT about this as a starting point: https://github.com/cpu/rustls/commit/4e09b6da8a460266de15a439729b43e980343b70 I think it's approaching what you had in mind, but there might be some ways to make it more ergonomic for crate external implementations.

rustls/src/crypto/cipher.rs
```diff
@@ -171,6 +171,12 @@ impl From<[u8; NONCE_LEN]> for Iv {
     }
 }
 
+impl AsRef<[u8]> for Iv {
```
djc, line 171: Can/should we remove the `pub(crate)` from `Iv.0` here?

rustls/src/client/client_conn.rs
```diff
@@ -635,6 +647,12 @@ impl ClientConnection {
         self.inner.core.is_early_data_accepted()
     }
 
+    /// Access connection options whose use is dangerous and requires
+    /// extra care.
+    pub fn dangerous(self) -> danger::DangerousClientConnection {
```
djc, line 652: I don't love the `.dangerous().extract_secrets()` API. Maybe this should instead be an extension trait that's only exposed via a `danger` module? Like, you have to import `use rustls::danger::SecretExtraction;` in order to call `extract_secrets()` on `ClientConnection` or `ServerConnection`? Alternatively, maybe it would be more general to have the trait be called `DangerousConnection`?

cpu: > I don't love the .dangerous().extract_secrets() API

Is there something in particular about it you don't like? I sort of like that it's very explicit. An extension trait seems less explicit at the site of the actually dangerous operation.

ctz, line 652: I think this wants to be called `into_dangerous` to reflect that it consumes the connection. That also means it isn't a more general venue for other dangerous APIs that don't require consuming the connection.

cpu: > That also means it isn't a more general venue for other dangerous APIs that don't require consuming the connection.

Good point, maybe that's an argument in favour of [djc's suggestion](https://github.com/rustls/rustls/pull/1484#discussion_r1332919178) to use an extension trait instead. WDYT about that proposal?

rustls/src/crypto/cipher.rs
cpu: Done.

rustls/src/client/client_conn.rs
cpu: I went with Ctz's suggestion to use `dangerous_extract_secrets`.

cpu: > Can we just do dangerous_extract_secrets(self)?

:+1:, implemented in-branch. I think with the separate `enable_secret_extraction` bool gating this from being useful until you've taken action to enable it having the name imply the danger seems sufficient.

rustls/src/crypto/cipher.rs
```diff
@@ -147,11 +147,11 @@ impl dyn MessageDecrypter {
 
 /// A write or read IV.
 #[derive(Default)]
-pub struct Iv(pub(crate) [u8; NONCE_LEN]);
+pub struct Iv([u8; NONCE_LEN]);
 
 impl Iv {
     #[cfg(feature = "tls12")]
-    fn new(value: [u8; NONCE_LEN]) -> Self {
+    pub(crate) fn new(value: [u8; NONCE_LEN]) -> Self {
```
djc, line 151: Wouldn't external cipher suite implementations want to use this?

rustls/src/crypto/cipher.rs
```diff
@@ -197,7 +203,7 @@ impl Nonce {
 
 /// Size of TLS nonces (incorrectly termed "IV" in standard) for all supported ciphersuites
 /// (AES-GCM, Chacha20Poly1305)
-const NONCE_LEN: usize = 12;
+pub(crate) const NONCE_LEN: usize = 12;
```
djc, line 203: And this?

djc (approved):

rustls/src/crypto/cipher.rs
cpu: Done.

rustls/src/suites.rs
cpu: Let's land this version and I'll pull out the rework mentioned above as a separate PR to iterate on.

cpu: https://github.com/rustls/rustls/pull/1497

fasterthanlime:
Just seeing this now - this is great, thanks for working on this 🙏

PR #1483, opened 2023-09-19
doc: fix reference to CryptoProvider parameter

jsha:
Also fix link to ConfigBuilder::dangerous

PR #1488, opened 2023-09-20
Export `DangerousClientConfigBuilder`, follow-up doc tweaks

cpu:
This branch lands some small follow-ups to #1483:

* It changes the link display name for the default crypto provider to `RING` to match the target, and the module name.
* It exports `DangerousClientConfigBuilder`.
* It links directly to `DangerousClientConfigBuilder::with_custom_certificate_verifier`.
* It fixes the server config docs broken link to the `with_client_cert_verifier` fn.


ctz (approved):
lgtm!

PR #1489, opened 2023-09-21
Update rustls version to 0.22.0-alpha.3

stevefan1999-personal:
Per the request of @djc at https://github.com/rustls/tokio-rustls/pull/21#discussion_r1333158599, we should raise the alpha version because it contains key breaking changes which let me resolute to use crate patches which is not really a proper way to do things.


djc (approved):

cpu (approved):
Thanks!

djc:
Published alpha.3.

PR #1492, opened 2023-09-23
Zeroize types/fields containing secrets

### LINKED ISSUE #265: Use zeroize for best-effort secret zeroisation

ctz:
This is for #265. After the [Microsoft incident publicised a couple of months ago](https://www.feistyduck.com/bulletproof-tls-newsletter/issue_105_microsofts_compromised_private_key), I think I changed my mind about the relative importance of this. It also goes hand-in-hand with having more types containing secrets, introduced in the recent crypto provider work.

closes #265


cpu:
>  just for CI results currently.

@ctz The CI results seem promising. Is this something that could be reviewed if it were rebased on main or are there still aspects in flux?

ctz:
Yes I plan to rebase & prepare this for review.

ctz:
#1544 should mean I can drop the last couple of commits on this PR. Then I think it will be ready.

djc (approved):

cpu (approved):
LGTM.

I notice the icount benchmarks show a small bump, but it doesn't seem very significant. Perhaps worth running the local benchmarks to see if the instruction delta affects wallclock perf?

ctz:
> I notice the icount benchmarks show a small bump, but it doesn't seem very significant. Perhaps worth running the local benchmarks to see if the instruction delta affects wallclock perf?

FYI @aochagavia this PR might be a good case study for the new method of performance measurement -- instruction counts show a small (and expected) bump, but might cause an outsized impact on real-world performance.

PR #1493, opened 2023-09-25
Run CI bench for current branch before main

aochagavia:


djc:
Aahhh, I think the problem here is that this is calling `--locked` even for the `main` branch. So I suppose it needs to be a two-step process, where we land this first while only calling `--locked` for PR, and then submitting another PR that also uses `--locked` for `main`.

Agreed?

djc:
I'm surprised this works with `--locked` against the `main` branch. Do you know why that works?

aochagavia:
Yes, it happens because the second checkout preserves build artifacts (I enabled it to shorten CI times)

djc:
Hmm, so does `Cargo.lock` get counted as a build artifact? For this PR the base ref is my branch that also contains the `Cargo.lock`, but it seems likely that this will fail once merged to my branch (for which the PR's base is `main`).

aochagavia:
Ah, I didn't think about that. Now you say it, it indeed works in this case because your branch already has a `Cargo.lock`. Also, now the Cargo.lock is part of the repository, it will probably be removed when checking out `main`, so the second run will fail (which IMO is not a problem, because after merging the PR into main things should work)

djc:
Okay, so I think this PR (that targets #1469) actually doesn't make a difference, and the icount bench failure on #1469 is to be expected and will resolve itself fater merging to main? In that case I think we should just close this?

aochagavia:
That's true, the PR was not meant to solve the failure in your branch, but to make sure things work before merging. I originally thought it might be useful to merge the PR anyway, so future "breaking changes" that crash on `main` can still be somewhat tested (because they will run first in the modified branch). Closing without merging is also ok.

djc:
Okay, I don't have a strong opinion on which one to run first so if you think this is better I'll just merge this.

PR #1497, opened 2023-09-27
suites: rework `ConnectionTrafficSecrets`

cpu:
This branch is a follow up to https://github.com/rustls/rustls/pull/1484. It updates `ConnectionTrafficSecrets` to hold `AeadKey` and `Iv` instances, instead of byte arrays, removing the need for the `slices_to_arrays` and `slice_to_array` helpers, as [suggested by ctz](https://github.com/rustls/rustls/pull/1484#discussion_r1331477766).

### DIFF
--- a/rustls/src/crypto/ring/tls12.rs
+++ b/rustls/src/crypto/ring/tls12.rs
@@ -132,24 +132,9 @@ impl Tls12AeadAlgorithm for GcmAlgorithm {
         write_iv: &[u8],
         explicit: &[u8],
     ) -> Box<dyn MessageEncrypter> {
-        debug_assert_eq!(write_iv.len(), 4);
-        debug_assert_eq!(explicit.len(), 8);
-
         let enc_key =
             aead::LessSafeKey::new(aead::UnboundKey::new(self.0, enc_key.as_ref()).unwrap());
-
-        // The GCM nonce is constructed from a 32-bit 'salt' derived
-        // from the master-secret, and a 64-bit explicit part,
-        // with no specified construction.  Thanks for that.
-        //
-        // We use the same construction as TLS1.3/ChaCha20Poly1305:
-        // a starting point extracted from the key block, xored with
-        // the sequence number.
-        let mut iv = [0; NONCE_LEN];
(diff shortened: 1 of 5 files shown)


cpu:
@fasterthanlime :wave: Hi hi. Tagging you here to see if you have any thoughts since I think this refactoring will require some downstream changes [in hapsoc/ktls](https://github.com/hapsoc/ktls/blob/7c377cac000f60a8e91d34ae1af1798bcf90f2f0/src/ffi.rs#L131).

fasterthanlime:
> @fasterthanlime 👋 Hi hi. Tagging you here to see if you have any thoughts since I think this refactoring will require some downstream changes [in hapsoc/ktls](https://github.com/hapsoc/ktls/blob/7c377cac000f60a8e91d34ae1af1798bcf90f2f0/src/ffi.rs#L131).

Hi there! As long as the exact same information is available and I just need to peel away the extra typed wrappers, any interface changes are fine!

rustls/src/crypto/ring/tls12.rs
```diff
@@ -367,3 +342,21 @@ impl MessageEncrypter for ChaCha20Poly1305MessageEncrypter {
         Ok(OpaqueMessage::new(msg.typ, msg.version, buf))
     }
 }
+
+fn gcm_nonce(write_iv: &[u8], explicit: &[u8]) -> Iv {
```
djc, line 346: Should this be called `gcm_iv()` instead of `Iv`? Alternatively, should the `ConnectionTrafficSecrets` type hold a `Nonce` instead of `Iv` for GCM variants?

djc (approved):

rustls/src/crypto/ring/tls12.rs
cpu: > Should this be called gcm_iv() instead of Iv?

SGTM. Updated.

cpu:
> cpu enabled auto-merge now

Going to put this in the queue on the assumption Ctz likes his own suggestion. If there's feedback from post-merge reviews I'm happy to iterate.

PR #1498, opened 2023-09-27
Cargo: remove rcgen git patch

cpu:
The `rcgen` crate has cut a 0.11.2 release that includes the CRL functionality we were using a Cargo patch to depend on previously. This commit removes the patch, fixes one breakage in the server acceptor example, and updates the `Cargo.toml` and `Cargo.lock` files.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -16,7 +16,7 @@ env_logger = "0.10"
 log = { version = "0.4.4" }
 mio = { version = "0.8", features = ["net", "os-poll"] }
 pki-types = { package = "rustls-pki-types", version = "0.2" }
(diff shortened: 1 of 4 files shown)


examples/Cargo.toml
```diff
@@ -16,7 +16,7 @@ env_logger = "0.10"
 log = { version = "0.4.4" }
 mio = { version = "0.8", features = ["net", "os-poll"] }
 pki-types = { package = "rustls-pki-types", version = "0.2" }
-rcgen = { version = "0.11.1", features = ["pem"], default-features = false }
```
cpu, line 19: This is necessary to appease the minimum versions check: the examples depend on the 0.11.2+ CRL types.

djc (approved):

PR #1502, opened 2023-09-29
[3/3] no-std support phase I

japaric:
this PR implements phase I of RFC #1399 . NOTE that this PR is built on top of PR #1583 

originally phase I aimed to only allowing client functionality in no-std context but turns out that the changes were sufficient to also enable server functionality as evidenced in #1534 

breaking changes: implementing the RFC requires no breaking changes

blockers:
- [x] #1583 

this PR is best reviewed on a commit by commit basis (do skip PR#1583 commits)

### DIFF
--- a/provider-example/Cargo.toml
+++ b/provider-example/Cargo.toml
@@ -18,7 +18,7 @@ p256 = { version = "0.13.2", default-features = false, features = ["alloc", "ecd
 pkcs8 = "0.10.2"
 pki-types = { package = "rustls-pki-types", version = "1" }
 rand_core = { version = "0.6", features = ["getrandom"] }
-rustls = { path = "../rustls", default-features = false, features = ["logging", "tls12"] }
+rustls = { path = "../rustls", default-features = false, features = ["logging", "std", "tls12"] }
 rsa = { version = "0.9", features = ["sha2"], default-features = false }
 sha2 = { version = "0.10", default-features = false }
 signature = "2"
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -19,20 +19,21 @@ rustversion = { version = "1.0.6", optional = true }
 aws-lc-rs = { version = "1.6", optional = true, default-features = false, features = ["aws-lc-sys"] }
 log = { version = "0.4.4", optional = true }
 # remove once our MSRV is >= 1.70
-once_cell = "1"
+once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
 ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }
-webpki = { package = "rustls-webpki", version = "0.102.2", features = ["std"], default-features = false }
-pki-types = { package = "rustls-pki-types", version = "1.2", features = ["std"] }
+webpki = { package = "rustls-webpki", version = "0.102.2", features = ["alloc"], default-features = false }
+pki-types = { package = "rustls-pki-types", version = "1.2", features = ["alloc"] }
 zeroize = "1.7"
 
 [features]
-default = ["aws_lc_rs", "logging", "tls12"]
+default = ["logging", "std", "tls12"]
+std = ["aws_lc_rs", "webpki/std", "pki-types/std", "once_cell/std"]
 logging = ["log"]
-aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs"]
+aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs", "std"]
 ring = ["dep:ring", "webpki/ring"]
 tls12 = []
-read_buf = ["rustversion"]
+read_buf = ["rustversion", "std"]
 fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
 
 [dev-dependencies]
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -5,6 +5,7 @@ use crate::crypto::CryptoProvider;
 use crate::error::Error;
 use crate::key_log::NoKeyLog;
 use crate::msgs::handshake::CertificateChain;
+use crate::time_provider::TimeProvider;
 use crate::webpki::{self, WebPkiServerVerifier};
 use crate::{verify, versions};
 
@@ -56,6 +57,7 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
                 provider: self.state.provider,
                 versions: self.state.versions,
                 verifier,
+                time_provider: self.state.time_provider,
             },
             side: PhantomData,
         }
@@ -94,6 +96,7 @@ pub(super) mod danger {
                     provider: self.cfg.state.provider,
                     versions: self.cfg.state.versions,
                     verifier,
+                    time_provider: self.cfg.state.time_provider,
                 },
                 side: PhantomData,
             }
@@ -110,6 +113,7 @@ pub struct WantsClientCert {
     provider: Arc<CryptoProvider>,
     versions: versions::EnabledVersions,
     verifier: Arc<dyn verify::ServerCertVerifier>,
+    time_provider: Arc<dyn TimeProvider>,
 }
 
 impl ConfigBuilder<ClientConfig, WantsClientCert> {
@@ -161,6 +165,7 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
             enable_early_data: false,
             #[cfg(feature = "tls12")]
             require_ems: cfg!(feature = "fips"),
+            time_provider: self.state.time_provider,
         }
     }
 }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -32,7 +32,6 @@ use core::fmt;
 use core::marker::PhantomData;
 use core::mem;
 use core::ops::{Deref, DerefMut};
-use std::error::Error as StdError;
 
 #[cfg(doc)]
 use crate::{crypto, DistinguishedName};
@@ -894,7 +893,8 @@ impl fmt::Display for EarlyDataError {
     }
 }
 
-impl StdError for EarlyDataError {}
+#[cfg(feature = "std")]
+impl std::error::Error for EarlyDataError {}
 
 /// State associated with a client connection.
 #[derive(Debug)]
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -714,6 +714,7 @@ impl DerefMut for ClientConnection {
     }
 }
 
+#[cfg(feature = "std")]
 #[doc(hidden)]
 impl<'a> TryFrom<&'a mut crate::Connection> for &'a mut ClientConnection {
     type Error = ();
@@ -727,6 +728,7 @@ impl<'a> TryFrom<&'a mut crate::Connection> for &'a mut ClientConnection {
     }
 }
 
+#[cfg(feature = "std")]
 impl From<ClientConnection> for crate::Connection {
     fn from(conn: ClientConnection) -> Self {
         Self::Client(conn)
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -11,15 +11,20 @@ use crate::msgs::handshake::ClientExtension;
 use crate::msgs::persist;
 use crate::sign;
 use crate::suites::{ExtractedSecrets, SupportedCipherSuite};
+#[cfg(feature = "std")]
+use crate::time_provider::DefaultTimeProvider;
+use crate::time_provider::TimeProvider;
 use crate::unbuffered::{EncryptError, TransmitTlsData};
 use crate::versions;
 use crate::KeyLog;
-use crate::{verify, WantsVerifier, WantsVersions};
+#[cfg(feature = "std")]
+use crate::WantsVerifier;
+use crate::{verify, WantsVersions};
 
 use super::handy::{ClientSessionMemoryCache, NoClientSessionStorage};
 use super::hs;
 
-use pki_types::ServerName;
+use pki_types::{ServerName, UnixTime};
 
 use alloc::sync::Arc;
 use alloc::vec::Vec;
@@ -206,6 +211,9 @@ pub struct ClientConfig {
     #[cfg(feature = "tls12")]
     pub require_ems: bool,
 
+    /// Provides the current system time
+    pub time_provider: Arc<dyn TimeProvider>,
+
     /// Source of randomness and other crypto.
     pub(super) provider: Arc<CryptoProvider>,
 
@@ -223,6 +231,7 @@ impl ClientConfig {
     /// and safe protocol version defaults.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
+    #[cfg(feature = "std")]
     pub fn builder() -> ConfigBuilder<Self, WantsVerifier> {
         Self::builder_with_protocol_versions(versions::DEFAULT_VERSIONS)
     }
@@ -239,6 +248,7 @@ impl ClientConfig {
     ///   the crate features and process default.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
+    #[cfg(feature = "std")]
     pub fn builder_with_protocol_versions(
         versions: &[&'static versions::SupportedProtocolVersion],
     ) -> ConfigBuilder<Self, WantsVerifier> {
@@ -260,11 +270,41 @@ impl ClientConfig {
     /// version is not supported by the provider's ciphersuites.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
+    #[cfg(feature = "std")]
     pub fn builder_with_provider(
         provider: Arc<CryptoProvider>,
     ) -> ConfigBuilder<Self, WantsVersions> {
         ConfigBuilder {
-            state: WantsVersions { provider },
+            state: WantsVersions {
+                provider,
+                time_provider: Arc::new(DefaultTimeProvider),
+            },
+            side: PhantomData,
+        }
+    }
+    /// Create a builder for a client configuration with no default implementation details.
+    ///
+    /// This API must be used by `no_std` users.
+    ///
+    /// You must provide a specific [`TimeProvider`].
+    ///
+    /// You must provide a specific [`CryptoProvider`].
+    ///
+    /// This will use the provider's configured ciphersuites. You must additionally choose
+    /// which protocol versions to enable, using `with_protocol_versions` or
+    /// `with_safe_default_protocol_versions` and handling the `Result` in case a protocol
+    /// version is not supported by the provider's ciphersuites.
+    ///
+    /// For more information, see the [`ConfigBuilder`] documentation.
+    pub fn builder_with_details(
+        provider: Arc<CryptoProvider>,
+        time_provider: Arc<dyn TimeProvider>,
+    ) -> ConfigBuilder<Self, WantsVersions> {
+        ConfigBuilder {
+            state: WantsVersions {
+                provider,
+                time_provider,
+            },
             side: PhantomData,
         }
     }
@@ -327,6 +367,12 @@ impl ClientConfig {
             .copied()
             .find(|skxg| skxg.name() == group)
     }
+
+    pub(super) fn current_time(&self) -> Result<UnixTime, Error> {
+        self.time_provider
+            .current_time()
+            .ok_or(Error::FailedToGetCurrentTime)
+    }
 }
 
 impl Clone for ClientConfig {
@@ -345,6 +391,7 @@ impl Clone for ClientConfig {
             enable_early_data: self.enable_early_data,
             #[cfg(feature = "tls12")]
             require_ems: self.require_ems,
+            time_provider: Arc::clone(&self.time_provider),
         }
     }
 }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -1,6 +1,6 @@
 use crate::builder::ConfigBuilder;
 use crate::common_state::{CommonState, Protocol, Side};
-use crate::conn::{ConnectionCommon, ConnectionCore, UnbufferedConnectionCommon};
+use crate::conn::{ConnectionCore, UnbufferedConnectionCommon};
 use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::error::Error;
@@ -10,7 +10,7 @@ use crate::msgs::enums::NamedGroup;
 use crate::msgs::handshake::ClientExtension;
 use crate::msgs::persist;
 use crate::sign;
-use crate::suites::{ExtractedSecrets, SupportedCipherSuite};
+use crate::suites::SupportedCipherSuite;
 #[cfg(feature = "std")]
 use crate::time_provider::DefaultTimeProvider;
 use crate::time_provider::TimeProvider;
@@ -33,7 +33,6 @@ use core::marker::PhantomData;
 use core::mem;
 use core::ops::{Deref, DerefMut};
 use std::error::Error as StdError;
-use std::io;
 
 #[cfg(doc)]
 use crate::{crypto, DistinguishedName};
@@ -559,11 +558,6 @@ impl EarlyData {
         }
     }
 
-    fn check_write(&mut self, sz: usize) -> io::Result<usize> {
-        self.check_write_opt(sz)
-            .ok_or_else(|| io::Error::from(io::ErrorKind::InvalidInput))
-    }
-
     fn check_write_opt(&mut self, sz: usize) -> Option<usize> {
         match self.state {
             EarlyDataState::Disabled => unreachable!(),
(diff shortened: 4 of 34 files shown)


ctz:
> * `pki_types::UnixTime` will need to gain a constructor so that `no_std` users can implement the `GetCurrentTime` trait

I think this exists and is `UnixTime::since_unix_epoch(core::time::Duration)`

djc:
I think this would be much easier to digest if it was split into smaller commits, for example:

* One commit for importing `Vec` explicitly
* One commit for `TimeProvider`
* One commit for dealing with the cache
* One commit for compiling out all the server stuff

japaric:
the PR has been split in commits to make it easier to review. all the "rm (..)" commits can be squashed at a later time. also, there are several warning that I haven't fixed and they involve cfg-ing away more code. for now, I have only removed what's need to make rustls compile with `#![no_std]`

djc:
I think the `rm *` commits would actually be nice to keep, assuming each can pass `cargo check` on its own? Would make it a lot easier to review!

japaric:
> I think the rm * commits would actually be nice to keep, assuming each can pass cargo check on its own?

I reordered the commits a bit and now only these contiguous commits:

- add std feature
- no-std: remove field from `*Error::Other` variants
- no-std: add TimeProvider to ClientConfig
- no-std: rm TicketSwitcher
- no-std: add TimeProvider to ServerConfig

make `cargo b --no-default-features` fail. `cargo b` works on all the commits of this PR.

---

The first commit of this PR basically adds this to `src/lib.rs`

``` rust
#![no_std]
extern crate std;
```

this switches the prelude from `std::prelude` to `core::prelude` and forces one to explicitly import anything that's not in `core`, that is mainly `alloc` API.

I think it would be worthwhile to land that change in a separate PR ahead of this one: it'll prevent relying on implicit `alloc` imports so all future PRs will have to include explicit `alloc` imports to avoid breaking the build. That would not only prevent this PR from bitrotting but I think it would help maintain no-std support long term. I'll submit that commit as a separate PR with a fleshed out rationale.

japaric:
quick update: I have rebased this PR now that #1524 has been merged. I have also pushed a commit that makes `cargo b --no-default-features --features ring` work on Linux. however, I tried to build rustls with that set of features (`-std +ring`) for the `x86_64-unknown-none` target and it failed to compile.

turns out that *ring* has this sealed `SecureRandom` trait (note that there are two of them in the *ring* codebase; one is public and the other is not) that's implemented for `SystemRandom` but only for a [known list of OSes](https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/src/rand.rs#L127-L149). as "OS agnostic" targets like `x86_64-unknown-none` have `os = "none"` in their target specification, `SystemRandom` lacks the `SecureRandom` trait and that causes problems in the rustls codebase: [`RsaSigner`](https://github.com/rustls/rustls/blob/ac10422a898335230f64104b169461a50e9d61db/rustls/src/crypto/ring/sign.rs#L148), [`KxGroup`](https://github.com/rustls/rustls/blob/ac10422a898335230f64104b169461a50e9d61db/rustls/src/crypto/ring/kx.rs#L27), etc.

for more context: I'm building a small rustls no-std demo that only depends on `libc` bindings and should work on any OS that has POSIX layer as way to exercise this PR. I'm running the demo on Linux so I do have a "secure random" `getrandom` implementation but I can't signal that to *ring*. I'll explore using an alternative crypto provider as showcased in #1405.

djc:
@japaric probably also good to open a *ring* issue to signal this there and see if there is a better solution there. (Please link the issue here if you do so.)

japaric:
@djc done. submitted briansmith/ring#1734

japaric:
rebased this on top of #1583 and marked it as ready for review

rustls/src/client/client_conn.rs
```diff
@@ -378,7 +419,11 @@ pub enum ServerName {
 
     /// The server is identified by an IP address. SNI is not
     /// done.
+    #[cfg(feature = "std")]
     IpAddress(IpAddr),
+
+    /// Same as the `IpAddress` variant, which is a wrapper around `std::net::IpAddr`, but no-std compatible
+    NoStdIpAddress(webpki::IpAddr),
```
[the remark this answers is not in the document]
ctz: I think I'd prefer option (2), awaiting https://doc.rust-lang.org/core/net/index.html to be stabilised.

japaric:
rebased to fix merge conflicts.

thanks djc for the switch to `pki_types::ServerName`. that simplified a few things in this PR.

pvdrz:
rebased on top of https://github.com/rustls/rustls/pull/1583

rustls/src/client/tls12.rs
```diff
@@ -974,6 +977,12 @@ impl ExpectFinished {
             return;
         }
 
+        let now = if let Ok(now) = self.config.get_current_time() {
+            now
+        } else {
+            return;
```
cpu, line 983: Does this early return merit a `debug!` like on L976? I think it would be helpful.

rustls/src/client/hs.rs
```diff
@@ -68,7 +68,9 @@ fn find_session(
             None
         })
         .and_then(|resuming| {
-            let retrieved = persist::Retrieved::new(resuming, UnixTime::now());
+            let now = config.get_current_time().ok()?;
```
cpu, line 71: I think matching the err result here and `debug!` logging would be helpful. If the time provider is broken session resumption would silently return `None` for all sessions.

rustls/src/time_provider.rs
```diff
@@ -0,0 +1,47 @@
+//! no-std time provider
+
+use alloc::sync::Arc;
+use core::fmt::Debug;
+
+use pki_types::UnixTime;
+
+/// An object that provides the current time
```
cpu, line 8: After the rustdoc headline I think we should mention some of the places where a time provider is used . For certificate validation in particular it's important that the time be accurate or an expired certificate could be accepted. Calling that out explicitly might not hurt.

rustls/src/error.rs
```diff
@@ -523,7 +524,10 @@ impl fmt::Display for Error {
                 write!(f, "the supplied max_fragment_size was too small or large")
             }
             Self::General(ref err) => write!(f, "unexpected error: {}", err),
+            #[cfg(feature = "std")]
             Self::Other(ref err) => write!(f, "other error: {}", err),
+            #[cfg(not(feature = "std"))]
+            Self::Other(..) => f.write_str("write error"),
```
cpu, line 532: Is this an appropriate description for all cases or should it say something like "unknown error" or "other unspecified error"?

rustls/src/client/client_conn.rs
```diff
@@ -401,7 +404,13 @@ impl Default for Resumption {
     /// Create an in-memory session store resumption with up to 256 server names, allowing
     /// a TLS 1.2 session to resume with a session id or RFC 5077 ticket.
```
cpu, line 412: I think this rustdoc comment should be updated to mention that for no-std it actually disables resumption.

cpu (commented):
This is great, thank you. I picked up review at 94e507357a31e12030ccc224c21696d910ae475e onward. I only have some small comments.

The number of new `#[cfg(feature = "std")]` flags is a bit dizzying but I trust your process/experience here.

rustls/src/client/client_conn.rs
cpu: (Though maybe of limited value since this comes back in phase ii)

pvdrz: I did it anyway :P

cpu: Thanks :-)

provider-example/Cargo.toml
```diff
@@ -18,7 +18,7 @@ p256 = { version = "0.13.2", default-features = false, features = ["alloc", "ecd
 pkcs8 = "0.10.2"
 pki-types = { package = "rustls-pki-types", version = "1" }
 rand_core = { version = "0.6", features = ["getrandom"] }
-rustls = { path = "../rustls", default-features = false, features = ["logging", "tls12"] }
+rustls = { path = "../rustls", default-features = false, features = ["logging", "tls12", "std"] }
```
djc, line 21: Let's keep features in alphabetic order?

rustls/Cargo.toml
```diff
@@ -20,17 +20,18 @@ aws-lc-rs = { version = "1.5", optional = true }
...
 subtle = { version = "2.5.0", default-features = false }
-webpki = { package = "rustls-webpki", version = "0.102", features = ["std"], default-features = false }
-pki-types = { package = "rustls-pki-types", version = "1", features = ["std"] }
+webpki = { package = "rustls-webpki", version = "0.102", features = ["alloc"], default-features = false }
+pki-types = { package = "rustls-pki-types", version = "1" }
 zeroize = "1.6.0"
 
 [features]
-default = ["logging", "ring", "tls12"]
+default = ["logging", "ring", "tls12", "std"]
```
djc, line 28: Also, here.

rustls/src/client/client_conn.rs
```diff
@@ -326,6 +332,19 @@ impl ClientConfig {
             .copied()
             .find(|skxg| skxg.name() == group)
     }
+
+    pub(super) fn get_current_time(&self) -> Result<UnixTime, Error> {
```
djc, line 345: Nit: let's kill the `get_()` prefix while we're moving this.

rustls/src/client/tls12.rs
```diff
@@ -974,6 +977,13 @@ impl ExpectFinished {
             return;
         }
 
+        let now = if let Ok(now) = self.config.get_current_time() {
```
djc, line 980: Style: IMO this would be more idiomatic as a `match` expression.

rustls/src/time_provider.rs
```diff
@@ -0,0 +1,50 @@
...
+impl TimeProvider {
+    /// Creates a new time provider
+    pub fn new(time_getter: impl GetCurrentTime + 'static) -> Self {
+        Self {
+            inner: Arc::new(time_getter),
+        }
+    }
+
+    /// A time provider that always fails
+    pub fn none() -> Self {
```
djc, line 26: How does it make sense to have this? Is it possible to do TLS without time?

rustls/src/time_provider.rs
```diff
@@ -0,0 +1,50 @@
...
+use core::fmt::Debug;
+
+use pki_types::UnixTime;
+
+/// An object that provides the current time.
+///
+/// This is used to, for example, check if a certificate has expired during certificate validation,
+/// or to check the age of a ticket.
+#[derive(Clone, Debug)]
+pub struct TimeProvider {
```
djc, line 13: What value does this wrapper provide? IMO it would be more clear to avoid this wrapper and use `Arc<dyn GetCurrentTime>` directly.

rustls/src/time_provider.rs
```diff
@@ -0,0 +1,50 @@
...
+        Self::new(NoTimeProvider)
+    }
+
+    pub(crate) fn get_current_time(&self) -> Option<UnixTime> {
+        self.inner.get_current_time()
+    }
+}
+
+/// Get current time
+pub trait GetCurrentTime: Debug + Send + Sync {
```
djc, line 36: Not a big fan of this trait name, I'd name this `TimeProvider`.

rustls/src/msgs/persist.rs
```diff
@@ -406,9 +406,11 @@ impl ServerSessionValue {
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[cfg(feature = "std")]
```
djc, line 409: Why is this necessary? Seems surprising.

rustls/src/server/server_conn.rs
```diff
@@ -405,6 +411,19 @@ impl ServerConfig {
             .iter()
             .any(|cs| cs.usable_for_protocol(proto))
     }
+
+    pub(super) fn get_current_time(&self) -> Result<UnixTime, Error> {
```
djc, line 422: Name, as for `ClientConfig`.

rustls/src/lib.rs
```diff
@@ -470,7 +470,14 @@ use crate::crypto::ring as test_provider;
 // The public interface is:
 pub use crate::builder::{ConfigBuilder, ConfigSide, WantsVerifier, WantsVersions};
 pub use crate::common_state::{CommonState, IoState, Side};
-pub use crate::conn::{Connection, ConnectionCommon, Reader, SideData, Writer};
+pub use crate::conn::unbuffered::{
```
djc, line 474: Not sure this change makes sense in this commit?

rustls/Cargo.toml
```diff
@@ -29,7 +29,7 @@ default = ["logging", "ring", "tls12", "std"]
 std = ["webpki/std", "pki-types/std"]
 logging = ["log"]
 aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs", "std"]
-ring = ["dep:ring", "webpki/ring", "std"]
+ring = ["dep:ring", "webpki/ring"]
```
djc, line 32: IMO this should be squashed into the change at the start, so we're not flipflopping within the PR.

djc (commented):
It's unclear to me how usable `no_std` rustls is after this PR.

Does this pass `cargo check` without warnings at every commit? What about `cargo check --all-targets`?

The proliferation of `std` guards seems pretty ugly, it might be better if there was some grouping with inline modules.

rustls/src/time_provider.rs
pvdrz: As mentioned in [the RFC](https://github.com/japaric/rustls/blob/ja-no-std-support-rfc/rfcs/000-no-std-support.md#systemtimenow) this is the default time provider and will always error as the user must provide their own time provider in no-std contexts.

djc: But why do we want to provide a default time provider that doesn't actually provide time? Is it load-bearing that time providers can yield `None`?

pvdrz:
> It's unclear to me how usable `no_std` rustls is after this PR.

As stated in the [RFC](https://github.com/japaric/rustls/blob/ja-no-std-support-rfc/rfcs/000-no-std-support.md#first-stage), the unbuffered API from the client side should be usable (without support sessions)
> Does this pass `cargo check` without warnings at every commit? What about `cargo check --all-targets`?

Yes, both `cargo check` and `cargo check --all-targets` pass.

> The proliferation of `std` guards seems pretty ugly, it might be better if there was some grouping with inline modules.

I'll see what I can do.

pvdrz:
@djc I moved most of the std only items to inline modules or to gated `impl` blocks. The ones remaining weren't worth the change imho. Some of these changes will be reverted by https://github.com/rustls/rustls/pull/1688 so I don't think they will be a big deal and such adjustments can be done on that PR later

rustls/src/error.rs
```diff
@@ -76,7 +76,7 @@ pub enum Error {
     General(String),
 
     /// We failed to figure out what time it currently is.
-    FailedToGetCurrentTime,
+    FailedToTimeProvider,
```
djc, line 78: Guessing this is some kind of mass-rename fallout? 🙂 I'd expect the error variants to stick to `FailedToGetCurrentTime`.

rustls/src/client/hs.rs
```diff
@@ -68,7 +68,12 @@ fn find_session(
             None
         })
         .and_then(|resuming| {
-            let retrieved = persist::Retrieved::new(resuming, UnixTime::now());
+            let now = config
+                .current_time()
+                .map_err(|_err| debug!("Could not get current time: {}", _err))
```
djc, line 73: Style nit: please inline format arguments.

rustls/src/lib.rs
djc: Okay, at least they're in a separate commit now, but I'm still confused. We already expose all of this stuff in the `unbuffered` module, around line 454?

rustls/src/server/server_conn.rs
```diff
@@ -418,160 +417,178 @@ impl ServerConfig {
...
-/// Allows reading of early data in resumed TLS1.3 connections.
-///
-/// "Early data" is also known as "0-RTT data".
-///
-/// This structure implements [`std::io::Read`].
-pub struct ReadEarlyData<'a> {
-    early_data: &'a mut EarlyDataState,
-}
+#[cfg(feature = "std")]
+mod server_connection {
```
djc, line 425: Let's just call this `connection`? (Same on the client side.)

rustls/src/common_state.rs
```diff
@@ -51,6 +51,66 @@ pub struct CommonState {
     pub(crate) enable_secret_extraction: bool,
 }
 
+#[cfg(feature = "std")]
+impl CommonState {
```
djc, line 56: Style nit: I like the idea of splitting the `impl`, but I think the guarded `impl` should go directly below the unguarded one.

(Same for other split `impl`s.)

rustls/src/server/handy.rs
```diff
@@ -141,71 +135,114 @@ impl server::ResolvesServerCert for AlwaysResolvesChain {
...
 }
 
-/// Something that resolves do different cert chains/keys based
-/// on client-supplied server name (via SNI).
-#[derive(Debug)]
-pub struct ResolvesServerCertUsingSni {
-    by_name: HashMap<String, Arc<sign::CertifiedKey>>,
-}
+#[cfg(feature = "std")]
+mod resolves_server_cert_using_sni {
```
djc, line 197: Nit: suggest naming this `sni_resolver`.

rustls/src/server/handy.rs
```diff
@@ -28,56 +26,116 @@ impl server::StoresServerSessions for NoServerSessionStorage {
...
 }
 
-/// An implementer of `StoresServerSessions` that stores everything
-/// in memory.  If enforces a limit on the number of stored sessions
-/// to bound memory usage.
-pub struct ServerSessionMemoryCache {
-    cache: Mutex<limited_cache::LimitedCache<Vec<u8>, Vec<u8>>>,
-}
+#[cfg(feature = "std")]
+mod server_session_memory_cache {
```
djc, line 30: Nit: suggest naming this `cache`.

djc (commented):
Okay, this is much better -- I think the main sticking point is the `TimeProvider` API.

djc:
> @djc I moved most of the std only items to inline modules or to gated `impl` blocks. The ones remaining weren't worth the change imho. Some of these changes will be reverted by #1688 so I don't think they will be a big deal and such adjustments can be done on that PR later

Yup, this is much better, thanks for working on that!

rustls/src/time_provider.rs
djc: So as far as I can surmise, the only reason to have `TimeProvider::current_time()` yield `Option` is that the config builder must create a `TImeProvider` to build a complete config, and in `no_std` situation that is platform dependent so we can't create it. I think we should probably force `no_std` configurations through a new separate `ClientConfig::builder_with_time_provider()` path instead and make `TimeProvider` yield just `UnixTime` or, more likely, `Result<UnixTime, Error>`.

pvdrz: some real-time clock drivers are fallible [ds1307](https://docs.rs/ds1307/latest/ds1307/#set-and-get-the-current-date-and-time-all-at-once) is an example of this. Returning `Option<UnixTime>` allows representing such errors, you could argue that we could return `Result<UnixTime, TimeProvider::Error>` or something like that but I consider that the error provides little to no use.

rustls/src/error.rs
pvdrz: Weird, I did that rename due to a rebase conflict. Maybe I'm confused but I'll revert it as the original PR used `FailedToGetCurrentTime` as well.

cpu:
@ctz @djc I've rebased this branch on top of `main` and I think it's ready for review. This branch is mostly feature-gating things that are `std`-only and then adding a `x86_64-unknown-none` build to keep us from backsliding so it might be nice to land this soon so we don't accidentally trip into choices that will make things harder in the future.

Most of my changes were pretty boring (_resolving conflicts, fixing some new `clippy` findings, etc_). The one place worth an explicit note is that I had to do some fiddling in the first commit (ae1c3a5744e57db54890220835e18d6491171e1c) in order to get the no-std build working with the `once_cell` dependency that landed in-tree in e7a1b418527eceba639457dc2ad2dcb29078f612 between the time of the original work by Japaric  et. al. and now.

The `once_cell` crate has a [FAQ section](https://docs.rs/once_cell/latest/once_cell/#faq) titled "Does this crate support no_std?" where it offers two potential solutions. I went with the `critical-section` feature but I'm out of my depth here. It seems like there's some nuance with respect to the right way to handle this and while my choice seems to appease the `x86_64-unknown-none` target build in CI and on my local machine I'm not sure if it's truly the best approach. WDYT?

cpu:
Rebased on `main`

nspin:
> The once_cell crate has a [FAQ section](https://docs.rs/once_cell/latest/once_cell/#faq) titled "Does this crate support no_std?" where it offers two potential solutions. I went with the critical-section feature but I'm out of my depth here. It seems like there's some nuance with respect to the right way to handle this and while my choice seems to appease the x86_64-unknown-none target build in CI and on my local machine I'm not sure if it's truly the best approach. WDYT?

In my opinion, the dependency on `critical-section` is unnecessary, and using [`once_cell::race::OnceBox`](https://docs.rs/once_cell/latest/once_cell/race/struct.OnceBox.html) would suffice.

The only cost of [`OnceBox`](https://docs.rs/once_cell/latest/once_cell/race/struct.OnceBox.html) that would be relevant to its usage in Rustls is that it introduces another level of pointer indirection, but `PROCESS_DEFAULT_PROVIDER` isn't used in any performance-critical paths, so I don't see this as a problem.

Blocking (via `std` or `critical-section`) provides better semantics for the [`get_or_init`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_init) family of functions, but Rustls does not use these.

cpu:
> In my opinion, the dependency on critical-section is unnecessary, and using [once_cell::race::OnceBox](https://docs.rs/once_cell/latest/once_cell/race/struct.OnceBox.html) would suffice.

@nspin Thanks for your input, much appreciated :bow: I've switched over to using `race` and `OnceBox` in my last push alongside rebasing on `main`.

quic.fail.txt
```diff
@@ -0,0 +1,124 @@
+an_send_3_inital_packets stdout ----
```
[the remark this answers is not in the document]
cpu: :sweat_smile: Yup, my bad. Thanks for noticing.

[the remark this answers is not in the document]
cpu: Fixed. I'll work on the other feedback shortly.

rustls/src/lib.rs
```diff
@@ -610,3 +610,8 @@ pub mod ticketer;
 
 /// This is the rustls manual.
 pub mod manual;
+
+#[cfg(feature = "std")]
+mod time_provider;
+#[cfg(not(feature = "std"))]
+pub mod time_provider;
```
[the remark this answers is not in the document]
cpu: Switched to have this module public both with and without `std`.

rustls/src/error.rs
```diff
@@ -544,43 +546,56 @@ impl From<rand::GetRandomFailed> for Error {
...
 
-/// Any other error that cannot be expressed by a more specific [`Error`] variant.
-///
-/// For example, an `OtherError` could be produced by a custom crypto provider
-/// exposing a provider specific error.
-///
-/// Enums holding this type will never compare equal to each other.
-#[derive(Debug, Clone)]
-pub struct OtherError(pub Arc<dyn StdError + Send + Sync>);
+#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
cpu: I've reworked the relevant commit. Good suggestion, it seems cleaner to me :+1:

rustls/src/conn.rs
```diff
@@ -372,9 +373,127 @@ pub struct ConnectionCommon<Data> {
...
+    /// Success from this function comes with some sundry state data
+    /// about the connection.
+    ///
+    /// [`read_tls`]: Connection::read_tls
+    /// [`process_new_packets`]: Connection::process_new_packets
+    #[inline]
+    pub fn process_new_packets(&mut self) -> Result<IoState, Error> {
+        self.core
+            .process_new_packets(&mut self.deframer_buffer, &mut self.sendable_plaintext)
+    }
```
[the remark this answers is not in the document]
cpu: Fixed. Thanks

rustls/src/server/builder.rs
```diff
@@ -116,7 +116,10 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
             cert_resolver,
             ignore_client_order: false,
             max_fragment_size: None,
+            #[cfg(feature = "std")]
             session_storage: handy::ServerSessionMemoryCache::new(256),
+            #[cfg(not(feature = "std"))]
+            session_storage: Arc::new(handy::NoServerSessionStorage {}),
```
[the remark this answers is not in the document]
cpu: I've updated the docs. We'll also need to remember to update them again when https://github.com/rustls/rustls/pull/1688 re-introduces session storage for no-std w/ hashbrown. :reminder_ribbon: :ballot_box_with_check:

rustls/src/error.rs
```diff
@@ -544,43 +546,59 @@ impl From<rand::GetRandomFailed> for Error {
...
     }
-}
 
-impl fmt::Display for OtherError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.0)
+    #[cfg(feature = "std")]
+    impl fmt::Display for OtherError {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            write!(f, "{}", self.0)
```
ctz, line 585:
```suggestion
    impl fmt::Display for OtherError {
        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
            #[cfg(feature = "std")]
            write!(f, "{}", self.0)
            #[cfg(not(feature = "std"))]
            f.write_str("no further information available")
```

I think that should address the remaining gating in `impl fmt::Display for Error` above.

ctz (approved):

rustls/src/time_provider.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) struct DefaultTimeProvider;
+
+impl TimeProvider for DefaultTimeProvider {
+    fn current_time(&self) -> Option<UnixTime> {
+        #[cfg(feature = "std")]
+        return Some(UnixTime::now());
+        #[cfg(not(feature = "std"))]
```
djc, line 26: As explained [previously](https://github.com/rustls/rustls/pull/1502#discussion_r1434942346), I'm not a fan of this setup. Can we instead feature-gate the `DefaultTimeProvider` on `#[cfg(feature = "std")]`, forcing `no_std` users to supply their own?

rustls/src/time_provider.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use core::fmt::Debug;
+
+use pki_types::UnixTime;
+
+/// An object that provides the current time.
+///
+/// This is used to, for example, check if a certificate has expired during certificate validation,
+/// or to check the age of a ticket.
+#[allow(unreachable_pub)]
```
djc, line 11: Not sure this `unreachable_pub` makes sense, we should allow access to this before this PR is merged, right -- so might as well do it in this commit?

djc: (Maybe squash this commit with the first one that actually introduces usage of the trait?)

rustls/src/time_provider.rs
djc: Also, if every caller of `current_time()` is going to convert the `Option` to `Result` anyway, should we just allow the implementer of this trait to supply an `Error` directly?

cpu: > Can we instead feature-gate the DefaultTimeProvider on #[cfg(feature = "std")], forcing no_std users to supply their own?

I can give this a shot. @ctz Do you prefer this approach as well?

> I suppose the no_std users who are the most likely implementers don't currently have a Box<dyn std::error::Error> to smuggle an alternative error through...

Yeah, that seems like a good reason to leave it as an `Option` IMO.

rustls/src/time_provider.rs
cpu: Fixed.

cpu:
> cpu force-pushed the no-std-support branch from 4251a12 to 231cfdf 

Rebased to address conflicts w/ `main`.

rustls/src/client/builder.rs
```diff
@@ -161,6 +166,97 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
...
+#[cfg(not(feature = "std"))]
+#[derive(Clone)]
+pub struct WantsTimeProvider {
+    provider: Arc<CryptoProvider>,
+    versions: versions::EnabledVersions,
+    verifier: Arc<dyn verify::ServerCertVerifier>,
+    client_auth_cert_resolver: Arc<dyn ResolvesClientCert>,
+}
+
+#[cfg(not(feature = "std"))]
```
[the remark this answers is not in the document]
cpu: Would the appeal be maintaining symmetry, or is there a deeper usecase you had in mind where you would want to provide a time provider w/ `std` enabled?

I don't have a strong preference here so would generally favour whatever Djc thinks is best since his feedback prompted the initial change from the approach used by the PR's first author.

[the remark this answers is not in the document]
cpu: > Sorry for the drive-by comment

Happy for input here btw :-)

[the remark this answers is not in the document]
ctz: It feels less hazardous (in the sense of crate features changing return types, etc.) to:

- std-gate all existing `ClientConfig::builder*`
- add ungated `ClientConfig::builder_with_details` (or some better name) that takes absolutely everything: a CryptoProvider, this time provider, (potentially a `Resumption` to put the no-std unhelpful and surprising `Resumption::default` front-and-center?)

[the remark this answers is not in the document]
djc: Yeah, hiding API when a feature is enabled is generally a no-no because Cargo features are supposed to be additive.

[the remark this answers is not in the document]
ctz: To clarify: do you mean adding a parameter to `builder_with_provider`? Or is a new `builder_with_*` function OK?

(I think my opinion is: 100% of `builder_with_provider` callers are std currently, and inflicting that change on them seems a bit gratuitous; so I prefer a new API surface.)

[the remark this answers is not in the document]
ctz: I have pushed a commit (that can be dissolved into earlier ones) reflecting my preferences on this.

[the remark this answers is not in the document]
djc: Yup, that's pretty much exactly what I had in mind. Tried to explain (badly) in https://github.com/rustls/rustls/pull/1502#discussion_r1434942346.

djc (approved):

rustls/src/client/builder.rs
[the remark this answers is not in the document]
ctz: I've squashed this into the commits that introduced it. I also took the liberty of improving some of the docs around `TimeProvider`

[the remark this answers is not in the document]
cpu: This direction makes sense to me :+1:

cpu (approved):
Thanks for landing those last tweaks @ctz

PR #1508, opened 2023-10-02
Upgrade to ring 0.17

djc:
Question: do we want to backport this to rustls 0.21? Would enable usage of rustls a on bunch more targets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,13 @@ need them.
 While Rustls itself is platform independent, it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing, this means x86, x86-64, armv7, and
-aarch64. For more information, see [the supported `ring` CI
-targets](https://github.com/briansmith/ring/blob/9cc0d45f4d8521f467bb3a621e74b1535e118188/.github/workflows/ci.yml#L151-L167).
+supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
+armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
+supported `ring` CI targets][ring-ci-targets].
 
 Rustls requires Rust 1.61 or later.
+
+[ring-ci-targets]: https://github.com/briansmith/ring/blob/d34858a918b04127d085cdbc20325263bf8fdd36/.github/workflows/ci.yml#L171-L190
 
 # Example code
 There are two example programs which use
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -496,7 +496,7 @@ version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 dependencies = [
- "spin",
+ "spin 0.5.2",
 ]
 
 [[package]]
@@ -747,7 +747,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "52c4f3084aa3bc7dfbba4eff4fab2a54db4324965d8872ab933565e6fbd83bc6"
 dependencies = [
  "pem",
- "ring",
+ "ring 0.16.20",
  "time",
  "yasna",
 ]
@@ -790,10 +790,24 @@ dependencies = [
  "cc",
  "libc",
  "once_cell",
- "spin",
- "untrusted",
+ "spin 0.5.2",
+ "untrusted 0.7.1",
  "web-sys",
  "winapi",
+]
+
+[[package]]
+name = "ring"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fb9d44f9bf6b635117787f72416783eb7e4227aaf255e5ce739563d817176a7e"
+dependencies = [
+ "cc",
+ "getrandom",
+ "libc",
+ "spin 0.9.8",
+ "untrusted 0.9.0",
+ "windows-sys",
 ]
 
 [[package]]
@@ -849,7 +863,7 @@ dependencies = [
  "bencher",
  "env_logger",
  "log",
- "ring",
+ "ring 0.17.0",
  "rustls-pemfile",
  "rustls-pki-types",
  "rustls-webpki",
@@ -878,7 +892,7 @@ name = "rustls-connect-tests"
 version = "0.0.1"
 dependencies = [
  "regex",
- "ring",
+ "ring 0.16.20",
  "rustls",
 ]
 
@@ -892,7 +906,7 @@ dependencies = [
  "mio",
  "rcgen",
  "regex",
- "ring",
+ "ring 0.16.20",
  "rustls",
  "rustls-pemfile",
  "rustls-pki-types",
@@ -941,9 +955,9 @@ version = "0.102.0-alpha.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77fecb40b15f8d5d22e08f89e0a0bfa2c17ddd5fc385700f96920bba2b99b680"
 dependencies = [
- "ring",
+ "ring 0.16.20",
  "rustls-pki-types",
- "untrusted",
+ "untrusted 0.7.1",
 ]
 
 [[package]]
@@ -1016,6 +1030,12 @@ name = "spin"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6e63cff320ae2c57904679ba7cb63280a3dc4613885beafb148ee7bf9aa9042d"
+
+[[package]]
+name = "spin"
+version = "0.9.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
 
 [[package]]
 name = "spki"
@@ -1103,6 +1123,12 @@ name = "untrusted"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a156c684c91ea7d62626509bce3cb4e1d9ed5c4d978f7b4352658f96a4c26b4a"
+
+[[package]]
+name = "untrusted"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8ecb6da28b8a351d773b68d5825ac39017e680750f980f3a1a85cd8dd28a47c1"
 
 [[package]]
 name = "utf8parse"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -951,13 +951,13 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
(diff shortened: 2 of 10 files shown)


djc:
@aochagavia the instruction count differences seem pretty worrying. Could you run time-based tests on your fancy new server for this PR?

connect-tests/Cargo.toml
```diff
@@ -2,7 +2,6 @@
 name = "rustls-connect-tests"
 version = "0.0.1"
 edition = "2021"
-rust-version = "1.60"
```
cpu, line 5: Why are these `rust-version` directives falling away here, maybe a candidate for a separate commit? The [workspace docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) read to me like we need a `[workspace.package]` in the top-level `Cargo.toml` for this to be inherited, is that incorrect?

fuzz/Cargo.toml
```diff
@@ -9,9 +9,6 @@ edition = "2021"
 [package.metadata]
 cargo-fuzz = true
 
-[dependencies]
-webpki = { package = "rustls-webpki", version = "0.101.0", features = ["alloc", "std"] }
```
cpu, line 13: This is another change I'm having trouble matching up to the *ring* update. Could you split it off?

cpu (commented):
Generally looks good, just a couple nits/comments

ctz:
> I could spend some time tomorrow to create a bunch of ad-hoc criterion benchmarks

I actually have some of these on the aws-lc-rs branch -- will extract them to merge separately.

ctz:
> > I could spend some time tomorrow to create a bunch of ad-hoc criterion benchmarks
> 
> I actually have some of these on the aws-lc-rs branch -- will extract them to merge separately.

That is #1511 -- local runs after merging that onto this gives:

```
 name                                                             main.txt ns/iter  ring17.txt ns/iter  diff ns/iter   diff %  speedup 
 bench_ewouldblock                                                11                11                             0    0.00%   x 1.00 
 crypto::ring::kx::benchmarks::bench_ecdh_p256                    52,784            51,970                      -814   -1.54%   x 1.02 
 crypto::ring::kx::benchmarks::bench_ecdh_p384                    769,208           765,313                   -3,895   -0.51%   x 1.01 
 crypto::ring::kx::benchmarks::bench_x25519                       54,136            75,106                    20,970   38.74%   x 0.72 ⬅
 crypto::ring::sign::benchmarks::bench_ecdsa_p256_sha256          16,739            16,540                      -199   -1.19%   x 1.01 
 crypto::ring::sign::benchmarks::bench_ecdsa_p384_sha384          415,218           411,350                   -3,868   -0.93%   x 1.01 
 crypto::ring::sign::benchmarks::bench_eddsa                      15,920            41,516                    25,596  160.78%   x 0.38 ⬅
 crypto::ring::sign::benchmarks::bench_load_and_validate_eddsa    15,033            40,857                    25,824  171.78%   x 0.37 ⬅
 crypto::ring::sign::benchmarks::bench_load_and_validate_p256     11,426            11,321                      -105   -0.92%   x 1.01 
 crypto::ring::sign::benchmarks::bench_load_and_validate_p384     385,406           384,189                   -1,217   -0.32%   x 1.00 
 crypto::ring::sign::benchmarks::bench_load_and_validate_rsa2048  21,941            21,666                      -275   -1.25%   x 1.01 
 crypto::ring::sign::benchmarks::bench_load_and_validate_rsa4096  80,350            77,442                    -2,908   -3.62%   x 1.04 
 crypto::ring::sign::benchmarks::bench_rsa2048_pkcs1_sha256       500,895           502,346                    1,451    0.29%   x 1.00 
 crypto::ring::sign::benchmarks::bench_rsa2048_pss_sha256         507,693           503,370                   -4,323   -0.85%   x 1.01 
 tls12::prf::benchmarks::bench_sha256                             2,436             2,370                        -66   -2.71%   x 1.03 
 tls13::key_schedule::benchmarks::bench_sha256                    10,269            10,030                      -239   -2.33%   x 1.02 
```

djc:
@briansmith are these performance changes to be expected?

briansmith:
It looks like it is just Curve25519? The faster Curve25519 will come in *ring* 0.17.1. I will ping you with the PR when it is ready.

briansmith:
@ctz what is the exact hardware you are running on?

FWIW, I do think the Curve25519 regressions would be expected although I thought they would be smaller than that. There is a better-optimized Curve25519 that I brought into *ring* 0.17.0 as part of the BoringSSL merge but didn't enable because I wanted to do some additional review/testing.

In the case of RSA, it looks like the benchmarks are noisy since we wouldn't expect PSS vs PKCS#1 to make a difference. If you could clean up the noise it would be great. BoringSSL did make a change that might have negatively affected performance slightly, and my analogous change might have affected performance slightly more than that. If so, it can be fixed.

Similarly, some other algorithms will also get faster because I brought in new faster code, but didn't hook it up. However, unlike Curve25519, there weren't any major changes that would have regressed performance, at least intentionally, IIRC.

briansmith:
I would suggest waiting a bit before doing this, until more of the perf work in 0.17.1 is done, so that we can understand how that work affects the build requirements. It might require a newer GCC and/or clang than 0.17.0 requires right now, and I'd like to get that all clarified.

briansmith:
@ctz https://github.com/briansmith/ring/pull/1688 is the PR for adding the new optimized Curve25519.

ctz:
> @ctz what is the exact hardware you are running on?

Intel Skylake -- specifically i7-7820X.

README.md
```diff
@@ -86,7 +86,7 @@ supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
 armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
 supported `ring` CI targets][ring-ci-targets].
 
-Rustls requires Rust 1.60 or later.
+Rustls requires Rust 1.61 or later.
```
ctz, line 89: edit this also in src/lib.rs

djc: (It was actually there, just in the other commit -- fixed that up now.)

connect-tests/Cargo.toml
djc: Pulled the MSRV changes out into #1516.

djc:
Should we hold off on merging this until the performance regresssion is dealt with? Should we hold off on releasing a 0.21.x release with this until the performance regression is dealt with? (I'm thinking maybe not and yes.)

fuzz/Cargo.toml
cpu: (I still think this cleanup is unrelated to the ring update and would be better in a separate commit)

briansmith:
using `TARGET_CC=clang cargo bench x25519` with clang-16 and the latest cargo/rust, I get significantly better performance with https://github.com/briansmith/ring/pull/1688 according to the benchmarks at https://github.com/briansmith/ring/pull/1694.

ctz:
> Should we hold off on merging this until the performance regresssion is dealt with? Should we hold off on releasing a 0.21.x release with this until the performance regression is dealt with? (I'm thinking maybe not and yes.)

I'd agree with that assessment.

This PR probably also depends on (or could incorporate) taking rustls-webpki==0.102.0-alpha.4 -- see eg fb8ee7578d2ad85d1c40f0bf07368efb0fb2fdb0

djc:
Hmm, I think I disagree. Clearly there's a lot of excitement about ring 0.17 (given the amount of issues opened and the number of issues/PRs linking to this PR), probably because of the massively increased portability? The point of not having ring appear in the public API means that we can bump ring without this being externally visible. rustls 0.22 is probably at least a week from being released and even when we release it, it will take the ecosystem a while to update dependencies -- that means that, for the time being, downstreams might be likely to pull two versions of ring into their dependency graphs when there's really no good reason to do that. So I think we *should* have ring 0.17 backports for webpki 0.101 and rustls 0.21 and release them as soon as we think it's ready, but maybe after the worst of the performance regressions has been addressed.

Or do you see serious risks for rustls 0.21 from our upgrade to ring 0.17 (I don't really). Plus, once we release rustls 0.22 we're on the hook to provide rustls 0.21 maintenance for a year longer -- that can be quite limited but it might still be nice if they are both on the latest ring release.

ctz:
> The point of not having ring appear in the public API means that we can bump ring without this being externally visible.

I'm not sure we have anything that validates we've achieved this though. I just went to check 0.21 and `rustls::cipher_suite::TLS13_AES_256_GCM_SHA384.hash_algorithm()` is pub and leaks a *ring* type. Are there downstream callers of that? Probably not. If there is maybe they will accept "we wrote that `pub` before `pub(crate)` existed, and it should've been `pub(crate)` all along".

I guess it would be interesting if `cargo-semver-checks` catches that detail if we do the backport.

> Or do you see serious risks for rustls 0.21 from our upgrade to ring 0.17 (I don't really).

I don't see this as about risk, more about the extent to which previous release branches are active vs just maintained.

djc:
> > The point of not having ring appear in the public API means that we can bump ring without this being externally visible.
> 
> I'm not sure we have anything that validates we've achieved this though. I just went to check 0.21 and `rustls::cipher_suite::TLS13_AES_256_GCM_SHA384.hash_algorithm()` is pub and leaks a _ring_ type. Are there downstream callers of that? Probably not. If there is maybe they will accept "we wrote that `pub` before `pub(crate)` existed, and it should've been `pub(crate)` all along".

Hmm, that's fair. Well, if we expose that probably one reason to not backport to 0.21.

> I guess it would be interesting if `cargo-semver-checks` catches that detail if we do the backport.

That would be a really interesting test -- I suppose it would?

> > Or do you see serious risks for rustls 0.21 from our upgrade to ring 0.17 (I don't really).
> 
> I don't see this as about risk, more about the extent to which previous release branches are active vs just maintained.

That's fair. I would be inclined to say that the 0.21 branch should definitely be perceived as "active" since it's what ~everyone is using right now. That doesn't mean that we backport every feature and refactoring, but for something like this which (a) should be (mostly) semver-compatible and (b) makes downstream users` life better with broader compatibility and (c) low risk, I like the cost-benefit of backporting.

It's also really nice to stick with up-to-date dependencies to avoid having duplicates in dependency graphs...

fuzz/Cargo.toml
ctz: taken the liberty of splitting this into #1523

djc: Thanks!

cpu:
>  I like the cost-benefit of backporting.

fwiw, I agree. I don't think it's something we should necessarily do for every dependency but I think for this case it makes sense.

cpu (approved):

ctz (approved):

cpu:
Started a 0.21.x backport in https://github.com/rustls/rustls/pull/1525

briansmith:
0.17.1 was released with the performance improvement to Ed25519.

briansmith:
More of the performance optimizations added in recent BoringSSL changes will appear in upcoming 0.17.x releases. We are trying a new QA process for enabling these.

djc:
I bumped the lockfile to 0.17.2 in #1528, but it didn't seem to help much with the performance regressions seen here.

briansmith:
Here are the latest benchmarks: https://github.com/briansmith/ring/pull/1688#issuecomment-1754061489. The fastest combination is *ring* 0.17 using Clang 17.

PR #1509, opened 2023-10-02
Make connect-tests part of the workspace

djc:
In #1282 we moved the networking-based tests into a separate crate to keep them from creating intermittent failures (and relative slowness) while running the default test suite. However, since #1469 we started also versioning `Cargo.lock` and this increases overhead from keeping crates as a separate workspace. Instead, the solution we've previously used for ci-bench is to make it part of the workspace but not part of the `default-members`. Follow that strategy here.


ctz (approved):

cpu (approved):
Makes sense to me, thanks.

PR #1511, opened 2023-10-02
Misc crypto micro-benchmarks

ctz:
Pulling these out of #1414 in case they are useful for measuring #1508 performance difference.


rustls/src/crypto/ring/sign.rs
```diff
@@ -421,3 +421,116 @@ mod tests {
         assert!(any_ecdsa_type(&key).is_err());
     }
 }
+
+#[cfg(bench)]
+mod benchmarks {
+    use super::{PrivateKeyDer, PrivatePkcs8KeyDer, SignatureScheme};
+
+    const SAMPLE_TLS13_MESSAGE: &[u8] = &[
```
djc, line 429: Nit: move the sample message down?

djc (approved):

cpu (approved):

PR #1516, opened 2023-10-04
Bump MSRV to 1.61

djc:
This is used in #1508 and #1414, so probably makes sense to pull this out separately and get it merged sooner.

Addressed comments from @cpu in #1508.

### DIFF
--- a/connect-tests/Cargo.toml
+++ b/connect-tests/Cargo.toml
@@ -2,7 +2,6 @@
 name = "rustls-connect-tests"
 version = "0.0.1"
 edition = "2021"
-rust-version = "1.60"
 license = "Apache-2.0 OR ISC OR MIT"
 description = "Rustls connectivity based integration tests."
 publish = false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: "1.60"
+          toolchain: "1.61"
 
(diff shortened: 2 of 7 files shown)


connect-tests/Cargo.toml
```diff
@@ -2,7 +2,6 @@
 name = "rustls-connect-tests"
 version = "0.0.1"
 edition = "2021"
-rust-version = "1.60"
```
cpu, line 5: Can you expand on the rationale for this for my own edification? I don't think it's the wrong change, I just want to understand it better.

My previous comment said:
> The [workspace docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) read to me like we need a [workspace.package] in the top-level Cargo.toml for this to be inherited, is that incorrect?

Was that incorrect? Does the `rust-version` get inherited even in the absence of the top level `workspace.package`? Is this somehow related to the MSRV change or just tidying?

cpu (approved):

connect-tests/Cargo.toml
djc: No, there's no implicit inheritance: only if we used `rust-version = { workspace: true }` or `rust-version.workspace = true` would we inherit it. It's really only cleaning up version tracking that we don't actually use before we bump it in the places we do use it.

cpu: :+1: That makes sense, thank you! I think the main intuition I was missing here is that there's no value in configuring an MSRV for crates internal to the project that aren't published externally.

ctz (approved):

PR #1517, opened 2023-10-04
Group Dependabot updates into a single PR

djc:
Follow-up from discussion in #1506.

I think ignores won't work here, since the list of crates we depend on is just too long.

Instead, wire up Dependabot to submit one PR per week instead of one per dependency.


cpu (approved):
Thanks

PR #1520, opened 2023-10-05
Take rustls-webpki 0.102.0-alpha.4

ctz:


djc (dismissed):

djc:
Actually, I think we should make this part of #1508 to convert to ring 0.17 atomically?

ctz:
That sounds logical, yes.

PR #1524, opened 2023-10-06
use `core::prelude` instead of `std::prelude`

japaric:
This change forces one to _explicitly_ import (`use`) everything that is in `std::prelude` but not in `core::prelude`. This helps maintain no-std support as even developers that are not interested in, or aware of, no-std support and / or that never run `cargo b --no-default-features` locally will get errors when they rely on `std::prelude` API.

This change is complementary to the `alloc_instead_of_core` and `std_instead_of_core` clippy lints (added in #1378 ) which don't emit warning when you rely on `std::prelude` imports.

> will get errors when they rely on `std::prelude` API.

as long as no-std support is not checked as part of CI, contributors may import `std::boxed::Box` instead of `alloc::boxed::Box` and the clippy lints mentioned above will fail to detect it (the lints have false negatives) so some vigilance from the reviewers will be needed for the time being. it's not a problem if some of those `std` imports make it into `main` because #1502 will catch and fix them.

cc #1399


rustls/src/key_log_file.rs
```diff
@@ -1,6 +1,7 @@
 #[cfg(feature = "logging")]
 use crate::log::warn;
 use crate::KeyLog;
+use alloc::{string::String, vec::Vec};
```
cpu, line 4: Could you maintain the whitespace between the `use crate::*` imports and the std /core imports here please :bow:

rustls/src/builder.rs
```diff
@@ -3,6 +3,7 @@ use crate::error::Error;
 use crate::suites::SupportedCipherSuite;
 use crate::versions;
 
+use alloc::{format, vec::Vec};
```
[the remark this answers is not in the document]
djc: Yup, I think there might be rustfmt config for this? (Though that might not be stable.)

rustls/src/key_log_file.rs
djc: Also I think in most modules we keep core/std imports below crate imports, which we should either globally fix or adhere to here.

rustls/src/lib.rs
```diff
@@ -298,8 +298,16 @@
...
 #![cfg_attr(read_buf, feature(read_buf))]
 #![cfg_attr(bench, feature(test))]
+#![cfg_attr(not(test), no_std)]
 
 extern crate alloc;
+// This `extern crate` plus the `#![no_std]` attribute changes the default prelude from
+// `std::prelude` to `core::prelude`. That forces one to _explicitly_ import (`use`) everything that
+// is in `std::prelude` but not in `core::prelude`. This helps maintain no-std support as even
+// developers that are not interested in, or aware of, no-std support and / or that never run
+// `cargo b --no-default-features` locally will get errors when they rely on `std::prelude` API.
```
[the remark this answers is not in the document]
djc: I use those all the time... but would advocate that we use the full name in comments/documentation like this.

rustls/src/builder.rs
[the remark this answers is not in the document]
cpu: https://github.com/rustls/rustls/pull/1526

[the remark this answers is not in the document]
japaric: > Yup, I think there might be rustfmt config for this?

https://github.com/rustls/rustls/pull/1529 documents how to get the desired behavior from rustfmt and rust-analyzer

rustls/src/key_log_file.rs
japaric: >  we keep core/std imports below crate imports

where do `self` and `super` imports fall in that picture?

AFAICT, neither rust-analyzer or rustfmt allows grouping imports in a way that `crate` imports precede imports from the `std` facade. [rust-analyzer uses this ordering](https://rust-analyzer.github.io/manual.html#auto-import): 

- `std` and `core`
- External Crates
- Current Crate, paths prefixed by `crate`
- Current Module, paths prefixed by `self`
- Super Module, paths prefixed by `super`

and you can get a similar ordering using this unstable rustfmt setting:

``` toml
group_imports = "StdExternalCrate"
```

except that `crate`, `self` and `super` imports are lumped together.

djc: I would be in favor of a commit or PR that reformats all modules in the repo according to the StdExternalCrate order.

japaric: there's a rust-analyzer feature request to allow custom ordering of those 5 import groups: https://github.com/rust-lang/rust-analyzer/issues/13960

japaric:
I have addressed the review comments in separate commits for ease of review. I can squash them later.

> I would be in favor of a commit or PR that reformats all modules in the repo according to the StdExternalCrate order.

I also included a test commit running this so you can observe the changes. the issue is that running `StdExternalCrate` requires nightly rustfmt and running stable rustfmt afterwards undoes some of the changes, unrelated to the imports, that nightly rustfmt did ... it seems some default around `max_width` and / or `chain_width` has changed in nightly that makes it apply different formatting decisions.

japaric:
> running stable rustfmt afterwards undoes some of the changes, unrelated to the imports, that nightly rustfmt 

no, wait. it seems that the mere presence of a `rustfmt.toml` file changes the default settings of rustfmt. that's very unexpected

``` console
$ cd rustls
$ cargo fmt -- --check && echo OK
OK
$ touch rustfmt.toml
$ cargo fmt -- --check || echo FAILED
(..)
FAILED

japaric:
>  the issue is that running StdExternalCrate requires nightly rustfmt and running stable rustfmt afterwards undoes some of the changes

I found a `rustfmt.toml` that applies `StdExternalCrate` on nightly and does not change the existing formatting on stable and included it as a "try" commit in this PR

cpu (approved):
Thanks!

PR #1525, opened 2023-10-06
v0.21.8 preparation

cpu:
# Description

This branch targets the `rel-0.21` release branch to make a point release to upgrade to *ring* 0.17.

This involves bringing the work from https://github.com/rustls/rustls/pull/1508 onto the rel-0.101 branch. I didn't cherry-pick the commits directly because the restructuring of the modules in `main` made it more challenging to resolve conflicts than to re-implement the changes wholesale.

We don't typically mention MSRV updates, so that change is omitted from the proposed release notes. Other commits are clippy fixes and crate internal tidying also of no interest to downstream consumers.

As discussed in the [main 0.17 update PR](https://github.com/rustls/rustls/pull/1508#issuecomment-1748963163) and in [comments below](https://github.com/rustls/rustls/pull/1525#issuecomment-1755951802) updating *ring* is a breaking change for the 0.21.x release stream because we were accidentally exposing `ring::digest::Algorithm` via the `Tls13CipherSuite::hash_algorithm` and `Tls12CipherSuite::hash_algorithm` fns. We believe there should be few-to-none consumers of these fns downstream, and so a small breaking change within an otherwise semver compatible patch release to make these fn's crate-internal is the best way to unblock an update to *ring* that would be breaking on its own based on this API exposure.

## Proposed Release Notes

* Fixes `ConnectionCommon::complete_io()` to flush writers before potentially expecting a response.
* Upgrades `*ring*` to 0.17 - **Note**: ring 0.17 when built with `gcc` will experience slower X25519 and Ed25519 operations compared to previous releases.
* Upgrades `rustls-webpki` to 0.101.7 to match `*ring*` 0.17 dependency
* `Tls12CipherSuite::hash_algorithm()` and `Tls13CipherSuite::hash_algorithm()` are now crate-internal. This is a small breaking change to remove unintended exposure of underlying `*ring*` types in the public API.

## TODO

- [x] decide how to handle the semver breakage from `Tls13CipherSuite::hash_algorithm()` and `Tls12CipherSuite::hash_algorithm()` leaking `ring::digest::Algorithm`.
- [x] decide if we're comfortable releasing w/ the `gcc` ring 0.17 X25519/Ed25519 performance regression unresolved.
- [x] wait for `rustls-webpki` 0.101.7 release - https://github.com/rustls/webpki/pull/199
- [x] remove `Cargo.toml` patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ supported by `ring`. At the time of writing this means x86, x86-64, armv7, and
 aarch64. For more information see [the supported `ring` CI
 targets](https://github.com/briansmith/ring/blob/9cc0d45f4d8521f467bb3a621e74b1535e118188/.github/workflows/ci.yml#L151-L167).
 
-Rustls requires Rust 1.60 or later.
+Rustls requires Rust 1.61 or later.
 
 # Example code
 There are two example programs which use
--- a/README.md
+++ b/README.md
@@ -86,13 +86,15 @@ need them.
 While Rustls itself is platform independent it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing this means x86, x86-64, aarch64,
-armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
-supported `ring` CI targets][ring-ci-targets].
+supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
+x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
+64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
+support WebAssembly.
+For more information, see [the supported `ring` target platforms][ring-target-platforms].
 
 Rustls requires Rust 1.61 or later.
 
-[ring-ci-targets]: https://github.com/briansmith/ring/blob/d34858a918b04127d085cdbc20325263bf8fdd36/.github/workflows/ci.yml#L171-L190
+[ring-target-platforms]: https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/include/ring-core/target.h#L18-L64
 
 # Example code
 There are two example programs which use
--- a/README.md
+++ b/README.md
@@ -86,11 +86,13 @@ need them.
 While Rustls itself is platform independent it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing this means x86, x86-64, armv7, and
-aarch64. For more information see [the supported `ring` CI
-targets](https://github.com/briansmith/ring/blob/9cc0d45f4d8521f467bb3a621e74b1535e118188/.github/workflows/ci.yml#L151-L167).
+supported by `ring`. At the time of writing this means x86, x86-64, aarch64,
+armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
+supported `ring` CI targets][ring-ci-targets].
 
 Rustls requires Rust 1.61 or later.
+
+[ring-ci-targets]: https://github.com/briansmith/ring/blob/d34858a918b04127d085cdbc20325263bf8fdd36/.github/workflows/ci.yml#L171-L190
 
 # Example code
 There are two example programs which use
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls"
-version = "0.21.7"
+version = "0.21.8"
 edition = "2021"
 rust-version = "1.61"
 license = "Apache-2.0 OR ISC OR MIT"
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -2,7 +2,7 @@
 name = "rustls"
 version = "0.21.7"
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.61"
 license = "Apache-2.0 OR ISC OR MIT"
 readme = "../README.md"
 description = "Rustls is a modern TLS library written in Rust."
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -19,7 +19,7 @@ rustversion = { version = "1.0.6", optional = true }
 log = { version = "0.4.4", optional = true }
 ring = "0.16.20"
 sct = "0.7.0"
-webpki = { package = "rustls-webpki", version = "0.101.2", features = ["alloc", "std"] }
+webpki = { package = "rustls-webpki", version = "0.101.7", features = ["alloc", "std"] }
 
 [features]
 default = ["logging", "tls12"]
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -17,7 +17,7 @@ rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
 log = { version = "0.4.4", optional = true }
-ring = "0.16.20"
+ring = "0.17"
 sct = "0.7.0"
 webpki = { package = "rustls-webpki", version = "0.101.7", features = ["alloc", "std"] }
 
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -59,11 +59,13 @@
 //! While Rustls itself is platform independent it uses
 //! [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 //! TLS. As a result, rustls only runs on platforms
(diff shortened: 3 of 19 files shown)


rustls/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "rustls"
 version = "0.21.7"
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.61"
```
djc, line 5: Should we also bring in the commit that dropped `rust-version` from the internal crates?

djc (approved):

rustls/Cargo.toml
cpu: Sure. Done :ballot_box_with_check:

cpu:
The plan is to hold https://github.com/rustls/webpki/pull/199 and this branch for a conclusive outcome to the perf. regression discussed in https://github.com/rustls/rustls/pull/1508

README.md
```diff
@@ -86,11 +86,13 @@ need them.
 While Rustls itself is platform independent it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing this means x86, x86-64, armv7, and
-aarch64. For more information see [the supported `ring` CI
-targets](https://github.com/briansmith/ring/blob/9cc0d45f4d8521f467bb3a621e74b1535e118188/.github/workflows/ci.yml#L151-L167).
+supported by `ring`. At the time of writing this means x86, x86-64, aarch64,
+armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
```
briansmith, line 90: The targets supported are documented in code in https://github.com/briansmith/ring/blob/main/include/ring-core/target.h#L25

32-bit ARM, Aarch64 (64-bit ARM), X86, x86-64, LOONGARCH64, 32-bit & 64-bit little-endian MIPS, 32-bit PowerPC (Big Endian), 64-bit PowerPC (Big and Little Endian),  64-bit RISC-V, s390x, and WebAssembly.

rustls/src/lib.rs
```diff
@@ -59,11 +59,13 @@
 //! While Rustls itself is platform independent it uses
 //! [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 //! TLS. As a result, rustls only runs on platforms
-//! supported by `ring`. At the time of writing this means x86, x86-64, armv7, and
-//! aarch64. For more information see [the supported `ring` CI
-//! targets](https://github.com/briansmith/ring/blob/9cc0d45f4d8521f467bb3a621e74b1535e118188/.github/workflows/ci.yml#L151-L167).
+//! supported by `ring`. At the time of writing this means x86, x86-64, aarch64,
+//! armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
+//! supported `ring` CI targets][ring-ci-targets].
```
briansmith, line 64: Ditto.

briansmith (changes requested):

briansmith:
Regarding the SemVer issue, my understanding is there was accidentally one API that exposed *ring* and that you are proposing to just go ahead and break SemVer for that one API. Is that correct? If so, maybe now would be a good time to guard that specific API with a "unstable-ring" or similarly-named feature as it is likely to come up again in the future (*ring* 0.18, etc.).

cpu:
> Regarding the SemVer issue, my understanding is there was accidentally one API that exposed ring and that you are proposing to just go ahead and break SemVer for that one API. Is that correct?

I don't think we've explicitly decided how to handle this. I put a more detailed TODO in the PR description for deciding on this item. Thanks for flagging.

>  If so, maybe now would be a good time to guard that specific API with a "unstable-ring" or similarly-named feature as it is likely to come up again in the future (ring 0.18, etc.).

If we're comfortable breaking semver (I don't see how we could do this update otherwise) it might be cleaner to switch the two affected fns to be `pub(crate)` instead of `pub`.

README.md
cpu: Thanks :+1: This should be fixed in `main`: https://github.com/rustls/rustls/pull/1532 I'll backport once that's done.

rustls/src/lib.rs
cpu: The README content is pulled from `src/lib.rs` w/ [`admin/pull-readme`](https://github.com/rustls/rustls/blob/main/admin/pull-readme) and [checked in CI](https://github.com/rustls/rustls/blob/5427a4d6e99ffcb67ffa1adcb9d690e320c8de57/.github/workflows/build.yml#L179-L184) so we can fix this in one place and resolve both. I'll mark this thread resolved since the substantive feedback is in the other comment thread on `README.md`.

ctz:
> If we're comfortable breaking semver (I don't see how we could do this update otherwise) it might be cleaner to switch the two affected fns to be `pub(crate)` instead of `pub`.

I'd be down with that.

djc:
Yeah, seems okay.

briansmith:
> > If so, maybe now would be a good time to guard that specific API with a "unstable-ring" or similarly-named feature as it is likely to come up again in the future (ring 0.18, etc.).

> If we're comfortable breaking semver (I don't see how we could do this update otherwise) it might be cleaner to switch the two affected fns to be pub(crate) instead of pub.

That sounds better to me, if we don't expect people to be consuming it intentionally.

djc:
Nit: in the release notes, I'd suffix method names with `()`, otherwise it looks to me as if you're referencing a field rather than a method.

cpu:
> Nit: in the release notes, I'd suffix method names with ()

Fixed. Also pulled in https://github.com/rustls/rustls/pull/1532

README.md
cpu: Update in both `main` and this branch.

cpu:
> cpu force-pushed the cpu-0.21.8-prep branch from fc6cf7c to 3bb3e1f 

Backported https://github.com/rustls/rustls/pull/1542 and updated the proposed release notes.

djc:
What is this currently blocked on? Is it the *ring* 0.17 performance regression? Maybe we should just take that as a given and move forward? It doesn't seem like any improvements are on the (near-term) horizon?

cpu:
> What is this currently blocked on? Is it the ring 0.17 performance regression? 

I think so. If we decide the performance regression is acceptable then we can finish up https://github.com/rustls/webpki/pull/199 and get this released soon after.

> It doesn't seem like any improvements are on the (near-term) horizon?

Agreed, the two upstream bugs I filed aren't giving any indication they'll be resolved soon. I think it's unfortunate to have a performance regression in a patch release but between holding the release indefinitely or releasing with the regression well documented I think the latter is probably the best trade-off.

@ctz What are your feelings on the matter?

briansmith:
It's a bit frustrating to call it a performance regression when it's actually the fastest it's ever been, if you use clang. I do think it's worth calling out in the release notes that building with clang 14+ will result in faster X25519 and Ed25519 operations while building with GCC will result in slower X25519 and Ed25519 operations, when compared with the previous release.

cpu:
> It's a bit frustrating to call it a performance regression when it's actually the fastest it's ever been, if you use clang. 

Apologies, it's not my intent to frustrate and I do appreciate that the update comes with a number of advantages. I will be more specific when describing the issue to emphasize it as specific to `gcc`.

> building with GCC will result in slower X25519 and Ed25519 operations, when compared with the previous release.

With the "building with GCC" caveat applied I think it's uncontroversial to describe this as a performance regression, yes?

ctz:
I don't think we should block this further on the performance when using GCC. But we should add the C compiler and versions thereof as additional dimensions of performance testing now we're aware it can have an unexpectedly large effect.

djc:
I think performance like TLS connection throughput and handshake latency is a lot more tangible for people than the number of cycles spent on key exchange (to name just one example). Also if our goal in this project is to promote adoption of rustls (for whatever reasons), and given that we're (currently) definitely not the default option for most environments, showing that our performance is competitive seems an important component of that.

djc:
I think this is in a good state -- let's publish this thing?

ctz:
I think the scope of our performance measurements should be focussed on the crate's defaults (that is, the *ring* provider), then any other providers we ship inside the crate (eg, aws-lc-rs when that lands) behind feature flags. I think that ideally covers factors we know alter the performance: C compiler and its version, and CPU architecture.

I think "down-stream" providers (like the impending rustls-rustcrypto crate) are out of scope, though naturally it would be great if those reuse our existing performance and protocol testing code.

The idea about a do-nothing crypto provider for testing maximum theoretical performance is definitely an interesting one!

ctz (approved):

cpu:
> rustls / Check semver compatibility (pull_request) Failing after 31s

This is expected based on the change in 4797228267ec18af52f659ef420413224b13d62b we call out in the release notes.

cpu:
* Pushed `v/0.21.8` tag.
* Published rustls v0.21.8 at registry `crates-io`
* Created [GitHub v0.21.8 release](https://github.com/rustls/rustls/releases/tag/v%2F0.21.8)

PR #1532, opened 2023-10-10
docs: adjust *ring* platform compatibility

cpu:
Updated the README/lib.rs notes about *ring* platform compatibility to fold in [suggestions from Brian Smith](https://github.com/rustls/rustls/pull/1525#discussion_r1353026518).

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,13 +86,15 @@ need them.
 While Rustls itself is platform independent, it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
-armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
-supported `ring` CI targets][ring-ci-targets].
+supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
+x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
+64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
+support WebAssembly.
+For more information, see [the supported `ring` target platforms][ring-target-platforms].
 
 Rustls requires Rust 1.61 or later.
 
-[ring-ci-targets]: https://github.com/briansmith/ring/blob/d34858a918b04127d085cdbc20325263bf8fdd36/.github/workflows/ci.yml#L171-L190
+[ring-target-platforms]: https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/include/ring-core/target.h#L18-L64
 
 # Example code
 There are two example programs which use
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -57,13 +57,15 @@
 //! While Rustls itself is platform independent, it uses
 //! [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 //! TLS. As a result, rustls only runs on platforms
-//! supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -86,13 +86,14 @@ need them.
 While Rustls itself is platform independent, it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
-armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
-supported `ring` CI targets][ring-ci-targets].
+supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
+x86, x86-64, LOONGARCH64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
```
briansmith, line 90: I think they want people to write it as "LoongArch" or "LoongArch64"

rustls/src/lib.rs
```diff
@@ -57,13 +57,14 @@
 //! While Rustls itself is platform independent, it uses
 //! [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 //! TLS. As a result, rustls only runs on platforms
-//! supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
-//! armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
-//! supported `ring` CI targets][ring-ci-targets].
+//! supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
+//! x86, x86-64, LOONGARCH64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
```
briansmith, line 61: Ditto.

briansmith (approved):

README.md
cpu: :+1: Fixed. Thanks.

cpu: Fixed, thanks.

djc (approved):

README.md
```diff
@@ -86,13 +86,14 @@ need them.
 While Rustls itself is platform independent, it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing, this means x86, x86-64, aarch64,
-armv7, powerpc64le, riscv64gc and s390x. For more information, see [the
-supported `ring` CI targets][ring-ci-targets].
+supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
+x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
+64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, s390x, and WebAssembly.
```
djc, line 91: Actually, I think rustls doesn't support WebAssembly for reasons unrelated to the crypto?

cpu: >  I think rustls doesn't support WebAssembly for reasons unrelated to the crypto?

Couldn't the same be true for other targets that aren't being exercised in our own CI? In general this makes me wary to use `*ring*'`s platform support as the basis of what we advertise Rustls supporting. 

In either case I've removed WebAssembly from the list (I can't build tip of main w/ `cargo build --target wasm32-unknown-unknown` despite https://github.com/rustls/rustls/issues/808 mentioning it worked at some point).

djc: > Couldn't the same be true for other targets that aren't being exercised in our own CI?

Maybe? It's also different with 0.17 because *ring* is now much less of a factor in restricting rustls platform support. I sort of expect WebAssembly to be the outlier case though.

briansmith: Thanks for pointing this out. *ring* 0.17.3 didn't enable Curve25519 for wasm32. *ring* 0.17.4-alpha.1 adds it and also expands the wasm32 testing to be the same as other architectures. *ring* 0.17.4 will be released in the next day or two.

PR #1533, opened 2023-10-11
Introduce the `PushBytes` trait to avoid using `Vec` in `Codec`

pvdrz:


rustls/src/client/tls12.rs
```diff
@@ -388,7 +388,10 @@ impl State<ClientConnectionData> for ExpectServerKx {
 
         // Save the signature and signed parameters for later verification.
         let mut kx_params = Vec::new();
-        ecdhe.params.encode(&mut kx_params);
+        ecdhe
```
djc, line 391: The proliferation of `unwrap()` calls seems like a substantial downside to this approach, I don't think that's the right way to do it.

rustls/src/client/tls12.rs
```diff
@@ -432,7 +435,7 @@ fn emit_certificate(
 fn emit_clientkx(transcript: &mut HandshakeHash, common: &mut CommonState, pub_key: &[u8]) {
     let mut buf = Vec::new();
     let ecpoint = PayloadU8::new(Vec::from(pub_key));
-    ecpoint.encode(&mut buf);
+    ecpoint.encode(&mut buf).unwrap();
```
ctz, line 438: Could use `get_encoding`?

rustls/src/common_state.rs
```diff
@@ -392,7 +392,7 @@ impl CommonState {
                         "QUIC uses TLS for the cryptographic handshake only"
                     );
                     let mut bytes = Vec::new();
-                    m.payload.encode(&mut bytes);
```
ctz, line 395: Ditto, `get_encoding` candidate.

rustls/src/msgs/message.rs
```diff
@@ -235,7 +239,7 @@ impl From<Message> for PlainMessage {
             MessagePayload::ApplicationData(payload) => payload,
             _ => {
                 let mut buf = Vec::new();
-                msg.payload.encode(&mut buf);
+                msg.payload.encode(&mut buf).unwrap();
```
ctz, line 242: Ditto `get_encoding`

rustls/src/tls12/mod.rs
```diff
@@ -327,7 +329,9 @@ mod tests {
         let key = X25519.start().unwrap();
         let server_params = ServerECDHParams::new(&*key);
         let mut server_buf = Vec::new();
-        server_params.encode(&mut server_buf);
+        server_params
+            .encode(&mut server_buf)
+            .unwrap();
         server_buf.push(34);
```
ctz, line 335: Ditto `get_encoding`

rustls/src/common_state.rs
pvdrz: can't do. `MessagePayload` has a custom `encode` method instead of implementing `Codec`.

rustls/src/client/tls12.rs
pvdrz: I pushed some changes to avoid adding those extra `unwrap()`s

rustls/src/client/tls12.rs
```diff
@@ -430,10 +428,7 @@ fn emit_certificate(
 }
 
 fn emit_clientkx(transcript: &mut HandshakeHash, common: &mut CommonState, pub_key: &[u8]) {
-    let mut buf = Vec::new();
-    let ecpoint = PayloadU8::new(Vec::from(pub_key));
-    ecpoint.encode(&mut buf);
-    let pubkey = Payload::new(buf);
+    let pubkey = Payload::new(PayloadU8::new(Vec::from(pub_key)).get_encoding());
```
djc, line 431: Please avoid making unrelated changes in the same PR or commit.

rustls/src/msgs/alert.rs
```diff
@@ -12,9 +12,9 @@ pub struct AlertMessagePayload {
 }
 
 impl Codec for AlertMessagePayload {
-    fn encode(&self, bytes: &mut Vec<u8>) {
-        self.level.encode(bytes);
-        self.description.encode(bytes);
+    fn try_encode<B: TryPushBytes>(&self, bytes: &mut B) -> Result<(), B::Error> {
```
djc, line 15: I think we should avoid renaming the method in the same commit (or, in the current state, PR) as change the type, since this creates a bunch of otherwise unrelated churn.

rustls/src/client/tls12.rs
```diff
@@ -387,9 +387,7 @@ impl State<ClientConnectionData> for ExpectServerKx {
             })?;
 
         // Save the signature and signed parameters for later verification.
-        let mut kx_params = Vec::new();
-        ecdhe.params.encode(&mut kx_params);
-        let server_kx = ServerKxDetails::new(kx_params, ecdhe.dss);
+        let server_kx = ServerKxDetails::new(ecdhe.params.get_encoding(), ecdhe.dss);
```
djc, line 390: Would be nice to land these changes separately (either in a separate commit if you're going to squash and clean up commits here or in a separate PR).

rustls/src/client/tls12.rs
pvdrz: Oh I wasn't to but @ctz suggested it on their review. I'll revert it then.

rustls/src/msgs/alert.rs
pvdrz: I think I don't follow. `Codec::encode` is still used as it was before, it's signature changed but it is backwards compatible with the old one.

rustls/src/client/tls12.rs
pvdrz: Same answer as https://github.com/rustls/rustls/pull/1533#discussion_r1357165250

rustls/src/common_state.rs
ctz: Fair enough!

rustls/src/msgs/alert.rs
djc: Ahh, sorry, yes, I see what you did now.

djc:
I think this is essentially the right direction but I also think it should get a bunch of polish.

* Naming: the trait and type names you're introducing have very low-level, mechanic names. I think it would be good to use names that evoke more concept-level meaning. For example, `ByteBuffer` is an extremely generic name, whereas the specific functionality that it is trying to facilitate is fairly specific (to allow single-pass encoding of length-prefixed data). Similarly, `PushBytes` could be more like `ResizableBuffer`. It would also be good to add comments to each of the traits and types introduced that explain what they're trying to achieve (for example, I'm still unclear on what `SliceBuffer` is for).
* Ordering: in the rustls project we try to maintain a [top-down ordering](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#top-down-ordering-within-modules), so dependent types/traits should come before lower-level types/traits in the source. This is not yet followed everywhere but would be good to apply for the new code here, especially since it's a pretty complicated hierarchy of stuff.
* Commit history: currently the commit history is pretty much a mess. It would help a lot if there are ways to split these changes into smaller chunks. For example, I think the `ByteBuffer` trait could be introduced before the other stuff.

cpu:
@pvdrz Was this closed because it's no longer necessary, or for some other reason?

pvdrz:
Yeah this was superseded by #1595 basically

PR #1543, opened 2023-10-17
Tidy up remaining non-standard names

cpu:
This branch tidies up some remaining type names/enum variants that don't meet the [Rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

Instances in `enums.rs` are ignored with the intent that we want to exactly match upstream documentation in these cases.

The [`clippy::upper_case_acroynyms`](https://rust-lang.github.io/rust-clippy/rust-1.72.0/index.html#/upper_case_acronyms) linter is added to the `lib.rs` deny list, and a `.clippy.toml` is added to set the `upper-case-acronyms-aggressive: true` config option to enable linting on all camel case names.

Resolves https://github.com/rustls/rustls/issues/541

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+upper-case-acronyms-aggressive = true
--- a/ci-bench/src/benchmark.rs
+++ b/ci-bench/src/benchmark.rs
@@ -101,19 +101,19 @@ pub enum ResumptionKind {
     /// No resumption
     No,
     /// Session ID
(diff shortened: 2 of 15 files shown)


ctz (approved):

ctz:
I think this is pretty uncontroversial and mechanical; going to merge with just my approval.

djc:
Nice!

PR #1544, opened 2023-10-18
Clean up `rustls::internal` exports

ctz:
This is a tentative attempt to clean this up, starting by being precise about what parts are needed to run the existing tests we have. That has uncovered a few instances of unused code that were previously `pub`.

### DIFF
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -135,10 +135,6 @@ impl PayloadU8 {
     pub(crate) fn empty() -> Self {
         Self(Vec::new())
     }
-
-    pub(crate) fn into_inner(self) -> Vec<u8> {
-        self.0
-    }
 }
 
 impl Codec for PayloadU8 {
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -57,10 +57,10 @@ impl fmt::Debug for Payload {
 
 /// An arbitrary, unknown-content, u24-length-prefixed payload
 #[derive(Clone, Eq, PartialEq)]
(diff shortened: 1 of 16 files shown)


rustls/tests/api.rs
```diff
@@ -17,6 +17,10 @@ use rustls::internal::msgs::base::Payload;
 use rustls::internal::msgs::codec::Codec;
 use rustls::internal::msgs::enums::AlertLevel;
 use rustls::internal::msgs::message::PlainMessage;
+use rustls::internal::msgs::{
+    handshake::ClientExtension, handshake::HandshakePayload, message::Message,
+    message::MessagePayload,
+};
```
cpu, line 23: nit: I think the two `rustls::internal::msgs::message::*` imports in this block, and the one on L19 should be combined into their own line, leaving the two `rustls::internal::msgs::handshake::*` imports on their own alongside.

rustls/src/msgs/base.rs
```diff
@@ -135,10 +135,6 @@ impl PayloadU8 {
     pub(crate) fn empty() -> Self {
         Self(Vec::new())
     }
-
-    pub(crate) fn into_inner(self) -> Vec<u8> {
-        self.0
-    }
```
cpu: It might be nice to squash the unused fixes into the commit that reduces the internal exports before merging but I don't feel super strongly about it.

cpu (approved):
Really nice clean-up!

djc (approved):

rustls/tests/api.rs
ctz: good spot - done

Issue #1541, opened 2023-10-17
Revocation checking by clients


ctz:
This appears to be a hard requirement in SP800-52.

We could reuse the CRL support we already have (to the extent that CRLs are useful or tractable in this context), or add support for verifying stapled OCSP responses (to the extent that is well deployed), or start fetching OCSP responses/CRLs from elsewhere. Crumbs!

cpu:
I agree CRLite is interesting :+1:

That said, I'm not enough of a FIPS-head to be able to guess at whether checking an aggregate view generated by a central entity is a fit to the SP800-52 requirement. Might be a tough sell? I believe in the CRLite technology and why it's trustworthy but in my experience with similar audits sometimes you want to be able to point to something that will already be familiar to the auditor, like boring direct CRL support. The language ctz linked in #1538 is fairly specific about the revocation information sources :-/ 

I think it would be helpful if I put together a branch that brings the existing CRL support used in the `WebPkiClientVerifier` to the `WebPkiServerVerifier`. The webpki API is already there, we just haven't [wired it up](https://github.com/rustls/rustls/blob/3b8a7c3afbc0a90a048316584a8b7f27bb87995e/rustls/src/webpki/verify.rs#L44) for the client-checking-server context. 

After that I'd be keen on trying to look at how feasible something like CRLite could be. I think its an interesting idea. Maybe it would be worth trying to sketch out an RFC for that. WDYT?

cpu:
Veering mildly off-topic for this ticket but I refreshed myself a bit on CRLite and Firefox's impl. builds on top of both SCT verification and their [intermediate preloading](https://blog.mozilla.org/security/2020/11/13/preloading-intermediate-ca-certificates-into-firefox/) technology. Both would be pretty decent sized chunks of work if we needed those as pre-reqs.

djc:
> That said, I'm not enough of a FIPS-head to be able to guess at whether checking an aggregate view generated by a central entity is a fit to the SP800-52 requirement. Might be a tough sell? I believe in the CRLite technology and why it's trustworthy but in my experience with similar audits sometimes you want to be able to point to something that will already be familiar to the auditor, like boring direct CRL support. The language ctz linked in #1538 is fairly specific about the revocation information sources :-/

I was envisioning there might be a complete Rust library toolchain to build the CRLite database/filter and consume it, so that people could self-host it maybe if they need the FIPS stuff? IIRC some parts of it are already provided in Rust.

gibbz00:
[...]

PR #1547, opened 2023-10-20
expand webpki verifier CRL support

cpu:
This branch achieves three top-level goals:

1. It expands the existing webpki client certificate verifier builder to allow configuring the revocation checking depth (by default, full chain when CRLs are provided, optionally only end-entity), and to control how unknown revocation status is handled (by default, treating unknown revocation status as an error when CRLs are provided, optionally allowing unknown).
2. It reworks the webpki server certificate verifier to be constructed using a builder, matching the client verifier.
3. It adds optional CRL support to the server certificate verifier, allowing users to opt in to providing CRLs and controlling how revocation will be handled (depth, unknown status, etc). This support will be helpful towards resolving #1541.


rustls/tests/common/mod.rs
```diff
@@ -284,8 +302,8 @@ pub fn make_server_config_with_kx_groups(
 
 pub fn get_client_root_store(kt: KeyType) -> Arc<RootCertStore> {
     let mut roots = kt.get_chain();
-    // drop server cert
-    roots.drain(0..1);
+    // drop server cert and intermediate
+    roots.drain(0..2);
```
cpu, line 306: This change is worth a bit of discussion. Previously we were constructing the client root store by reading the `client.fullchain` file (containing EE cert, intermediate and root) and then discarding just the EE cert. The net effect is that we build a root store here that contains _both_ the intermediate and the root.

We use this root store when constructing a client certificate verifier for two purposes:

1. For populating a list of subjects we send in the [certificate_authorities](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.4) extension of a `CertificateRequest` message (TLS 1.2 is handled similarly).
2. For populating a set of trust anchors we provide to webpki when verifying the client certificate.

For the first purpose it _might_ make sense to include the intermediate subjects. I'm not 100% sure. Reading the RFC text I think it could go either way.

For the second purpose it doesn't make sense to include the intermediate cert. We end up using it as a trust anchor which isn't appropriate: it's an intermediate certificate itself issued by a trust anchor. If we try path-building with the EE cert, the intermediates, and these trust anchors we never look at the intemediates: our built path is EE->TA (where the TA is the first intermediate). This means checking revocation status of the intermediate is never performed: we consider it a TA and those aren't covered by CRLs.

In this branch to make it feasible to test the intermediate revocation checking I've changed this function to only return the root TA, not the intermediate. As a result one other unit test (`client_cert_resolve`) that expected to see the intermediate subject in the `certificate_authorities` ext has to be updated to only expect the root TA subject.

rustls/src/error.rs
```diff
@@ -372,7 +375,7 @@ impl From<CertificateError> for AlertDescription {
             // certificate_unknown
             //  Some other (unspecified) issue arose in processing the
             //  certificate, rendering it unacceptable.
-            Other(_) => Self::CertificateUnknown,
+            Other(_) | UnknownRevocationStatus => Self::CertificateUnknown,
```
cpu, line 378: I'm not sure this mapping from the new unknown revocation status certificate error to alert makes sense. Suggestions welcome.

rustls/src/webpki/verify.rs
```diff
@@ -34,17 +39,14 @@ pub fn verify_server_cert_signed_by_trust_anchor(
...
-            supported_algs,
-            &roots.roots,
-            intermediates,
-            now,
-            webpki::KeyUsage::server_auth(),
-            None, // no CRLs
-        )
-        .map_err(pki_error)
-        .map(|_| ())
+    verify_server_cert_signed_by_trust_anchor_impl(
```
cpu, line 42: We want to be able to provide the `revocation` argument to Webpki's `verify_for_usage` fn when we're crate-internal, but can't have the webpki revocation types bleed into the public API.

This approach, having the public helper fn not support revocation checking, and deferring to an internal impl that does was the best I could come up with :-(

As discussed in https://github.com/rustls/webpki/issues/159#issuecomment-1739695300 moving the revocation types into `pki-types` isn't very appealing, but I'm open to other suggestions.

cpu:
> codecov/patch — 95.15% of diff hit (target 96.46%)

I'll see if I can address some of these coverage gaps (likely ~Mon).

rustls/src/webpki/verify.rs
```diff
@@ -437,6 +437,18 @@ impl fmt::Debug for WebPkiSupportedAlgorithms {
     }
 }
 
+/// wrapper around internal representation of a parsed certificate. This is used in order to avoid parsing twice when specifying custom verification
```
ctz, line 499: not changed by this commit, but should start with a capital letter, start with a "headline" sentence, and be manually wrapped

```suggestion
/// Wrapper around internal representation of a parsed certificate.
///
/// This is used in order to avoid parsing twice when specifying custom verification.
```

rustls/src/webpki/client_verifier_builder.rs
```diff
@@ -195,13 +195,13 @@ mod tests {
     }
 
     #[test]
-    fn test_noauth() {
+    fn test_client_verifier_noauth() {
```
djc: Pre-existing issue, but maybe split `noauth` into `no_auth` while you're in here?

rustls/src/webpki/client_verifier_builder.rs
```diff
@@ -43,6 +50,20 @@ impl ClientCertVerifierBuilder {
         self
     }
 
+    /// When CRLs are provided with [`with_crls`][Self::with_crls] only check the end entity
```
djc: Maybe tighten up the docstring headline a bit? (I'd start by removing ", not the entire chain", which feels like something that doesn't need to be in the headline.)

rustls/src/error.rs
djc: Just in terms of names, `BadCertificate` seems more appropriate than `CertificateUnknown`? Did you check what BoringSSL/OpenSSL do?

rustls/src/webpki/client_verifier_builder.rs
```diff
@@ -73,6 +80,19 @@ impl ClientCertVerifierBuilder {
         self
     }
 
+    /// When CRLs are provided with [`with_crls`][Self::with_crls] if it isn't possible to
```
djc: Nit: tighten up the headline?

rustls/src/webpki/verify.rs
```diff
@@ -278,13 +279,16 @@ impl WebPkiClientVerifier {
     /// client certificate validation.
     /// `revocation_check_depth` controls which certificates have their revocation status checked
     /// when `crls` are provided.
+    /// `unknown_revocation_policy` controls how certificates with an unknown revocation status
```
djc, line 333: How does this look in rustdoc? Should we use a bullet list here?

rustls/src/webpki/mod.rs
```diff
@@ -3,12 +3,12 @@ use alloc::sync::Arc;
 use crate::error::{CertRevocationListError, CertificateError, Error};
 
 mod anchors;
-mod client_verifier_builder;
+mod verifier_builders;
```
djc, line 6: Alternative proposal: go the other way and split `client_verifier_builder` and `verify` into `client_verifier` and `server_verifier` builder modules that contain the builders? In general I don't love it when our builders live in a different file from the type they build.

rustls/src/webpki/verifier_builders.rs
```diff
@@ -159,7 +159,7 @@ impl ClientCertVerifierBuilder {
 /// instead of constructing a builder.
 #[derive(Debug, Clone)]
 #[non_exhaustive]
-pub enum ClientCertVerifierBuilderError {
+pub enum CertVerifierBuilderError {
```
djc, line 283: Naming nit: how about tightening this up to `VerifierBuilderError`? I feel like the `Cert` prefix is fairly superfluous in this context?

rustls/src/webpki/verify.rs
```diff
@@ -275,7 +275,7 @@ impl WebPkiClientVerifier {
     /// Construct a new `WebpkiClientVerifier`.
     ///
     /// `roots` is the list of trust anchors to use for certificate validation.
-    /// `crls` are an iterable of owned certificate revocation lists (CRLs) to use for
+    /// `crls` are a vec of owned certificate revocation lists (CRLs) to use for
```
djc, line 329: I would capitalize `Vec`. (I would also write `is` rather than `are`, and maybe this deserves a bullet list, too?)

provider-example/src/lib.rs
```diff
@@ -61,8 +61,8 @@ pub static TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256: rustls::SupportedCipherS
 pub fn certificate_verifier(
     roots: rustls::RootCertStore,
 ) -> Arc<dyn rustls::client::danger::ServerCertVerifier> {
-    Arc::new(rustls::client::WebPkiServerVerifier::new_with_algorithms(
-        roots,
-        verify::ALGORITHMS,
-    ))
+    rustls::client::WebPkiServerVerifier::builder(roots.into())
```
djc, line 64: I think the commit message got truncated for this cmmit ("builder for server cert verifier, CRL support")

rustls/src/webpki/verify.rs
djc: Seems like an okay trade-off for now.

rustls/src/webpki/verifier_builders.rs
```diff
@@ -154,6 +156,125 @@ impl ClientCertVerifierBuilder {
...
+        Self {
+            roots,
+            crls: Vec::new(),
+            revocation_check_depth: RevocationCheckDepth::Chain,
+            unknown_revocation_policy: UnknownStatusPolicy::Deny,
+            supported_algs: None,
+        }
+    }
+
+    /// Verify the revocation state of presented client certificates against the provided
```
djc, line 182: Nit: tighter headlines?

rustls/src/webpki/verify.rs
```diff
@@ -457,6 +508,40 @@ impl<'a> TryFrom<&'a CertificateDer<'a>> for ParsedCertificate<'a> {
...
+    cert.0
+        .verify_for_usage(
+            supported_algs,
+            &roots.roots,
+            intermediates,
+            now,
+            webpki::KeyUsage::server_auth(),
+            revocation,
+        )
+        .map_err(pki_error)
```
djc, line 541: I tend to dislike chaining combinators that operate on different variants -- maybe try binding `result` and using an explicit `match` for that, see what you think?

rustls/src/webpki/verify.rs
```diff
@@ -151,10 +181,31 @@ impl ServerCertVerifier for WebPkiServerVerifier {
...
         let cert = ParsedCertificate::try_from(end_entity)?;
 
-        verify_server_cert_signed_by_trust_anchor(
+        let crls = borrow_crls(&self.crls);
+        let revocation = if crls.is_empty() {
+            None
+        } else {
+            // Note: unwrap here is safe because RevocationOptionsBuilder only errors when given
+            //       empty CRLs.
+            let mut builder = webpki::RevocationOptionsBuilder::new(&crls)
```
djc, line 190: The `borrow_crls(&self.crls)` -> `RevocationOptionsBuilder::new(&crls).unwrap()` dance looks pretty convoluted... is this what we ended up on for the client? Maybe webpki should expose some API to make this more streamlined?

djc (commented):
Haven't looked at the testing commits much yet, here's a first round of feedback.

rustls/src/error.rs
cpu: >  Did you check what BoringSSL/OpenSSL do?

I find the BoringSSL/OpenSSL code pretty hard to follow but I'll put some legwork in and see what I can figure out. 

> Just in terms of names, BadCertificate seems more appropriate than CertificateUnknown?

Switched to `BadCertificate` in the meantime.

rustls/src/webpki/verify.rs
cpu: Fixed up.

rustls/src/webpki/client_verifier_builder.rs
cpu: Done.

rustls/src/webpki/client_verifier_builder.rs
cpu: Done. Hopefully clearer now :crossed_fingers:

rustls/src/webpki/client_verifier_builder.rs
cpu: Done.

rustls/src/webpki/verify.rs
cpu: Switched this and the server equivalent to use bulleted lists. They're both `pub(crate)` but it doesn't hurt to make the Rustdoc more readable for `--document-private-items`.

rustls/src/webpki/mod.rs
cpu: Sure, that seems sensible. I've reworked accordingly.

rustls/src/webpki/verifier_builders.rs
cpu: SGTM. Done.

rustls/src/webpki/verify.rs
cpu: Fixed up.

provider-example/src/lib.rs
cpu: Ooops. Thanks for noticing. Fixed.

rustls/src/webpki/verifier_builders.rs
cpu: Done.

rustls/src/webpki/verify.rs
cpu: That's fair. I think the explicit `match` reads better :+1: Updated.

rustls/tests/common/mod.rs
cpu: > SGTM. I'll do some tidying.

Done.

>  I could take a shot at breaking this up in a follow-up PR. I think it's distinct enough from the work here that it'd be clearer to handle separately.

Filed a separate issue for this so I can resolve this feedback thread: https://github.com/rustls/rustls/issues/1549

rustls/src/error.rs
cpu: > I'll put some legwork in and see what I can figure out

I believe for both OpenSSL and BoringSSL in the absence of callbacks doing something differently, the failure to find a relevant CRL for a certificate in-scope for revocation checking will be turned into a `X509_V_ERR_UNABLE_TO_GET_CRL` result ([OpenSSL](https://github.com/openssl/openssl/blob/fac61ea4618c83826b51aebf03cbc2bc3ac7b8c8/crypto/x509/x509_vfy.c#L1015-L1024), [BoringSSL](https://github.com/google/boringssl/blob/bfa8369795b7533a222a72b7a1bc928941cd66bf/crypto/x509/x509_vfy.c#L832-L843)). In both cases that error is mapped to an Unknown CA alert ([OpenSSL](https://github.com/openssl/openssl/blob/fac61ea4618c83826b51aebf03cbc2bc3ac7b8c8/ssl/statem/statem_lib.c#L1763), [BoringSSL](https://github.com/google/boringssl/blob/bfa8369795b7533a222a72b7a1bc928941cd66bf/ssl/ssl_x509.cc#L1330C1-L1330C1)).

We could do the same, but at first pass it feels to me like `BadCertificate` might be a better fit? I haven't been able to find any standards text that would suggest which is more appropriate.

cpu: > We could do the same, but at first pass it feels to me like BadCertificate might be a better fit? I haven't been able to find any standards text that would suggest which is more appropriate.

Since the existing `From<CertificateError> for AlertDescription ` impl is commented as "heavily referenced" from the OpenSSL/BoringSSL implementations I decided to tack on a commit that updates our handling of `Error::UnknownRevocationStatus` to use `AlertDescription::UnknownCA` for this condition too: 155fbd6b0f4c51eb0f7b249c0ea90dba5d1d3ad4

It's probably the least surprising thing to do and I can't find strong justification for doing otherwise :shrug: We can always change it again if someone disagrees with my logic here.

rustls/src/webpki/verify.rs
cpu: I have a couple branches under way that should make this more ergonomic. Since it will require some song/dance to get webpki fixed up and then a new alpha used here maybe we could look at that separately from this PR? WDYT?

cpu: > I have a couple branches under way that should make this more ergonomic.

Webpki: https://github.com/rustls/webpki/pull/203
Rustls: https://github.com/rustls/rustls/pull/1552

rustls/tests/api.rs
```diff
@@ -1183,6 +1185,42 @@ fn client_check_server_certificate_intermediate_revoked() {
...
+#[test]
+fn client_check_server_certificate_helper_api() {
+    for kt in ALL_KEY_TYPES.iter() {
+        let chain = kt.get_chain();
+        let correct_roots = get_client_root_store(*kt);
+        let incorrect_roots = get_client_root_store(match kt {
+            KeyType::Rsa => KeyType::Ecdsa,
+            _ => KeyType::Rsa,
+        });
+        // Using the correct turst anchors, we should verify without error.
```
ctz, line 1200:
```suggestion
        // Using the correct trust anchors, we should verify without error.
```

ctz (approved):

rustls/tests/api.rs
cpu: Oops. Fixed up :-)

cpu:
> icount bench / Run icount benchmarks (pull_request) Failing after 2m

The diffs here are either very small, or speedups, so I think we're OK here.

PR #1552, opened 2023-10-24
Improving webpki verifier CRL support ergonomics

cpu:
### Description

This branch updates Rustls to use the rustls-webpki v0.102.0-alpha.6 crate. This version of webpki improves CRL ergonomics (see https://github.com/rustls/webpki/pull/203). 

Since the changes in this alpha are breaking, I've taken the update & resolved the breaking changes in one commit to keep everything building and testing cleanly.

#### use `with_status_policy builder` fn

The upstream crate added a more ergonomic interface we can use in place of having to keep around a mutable builder and doing our own matching.

#### avoid CRL dyn trait hurdles

The upstream crate made working with CRLs easier by replacing the `CertRevocationList` trait with an `enum` representation.

Notably this makes working with the `Vec<OwnedCertRevocationList>` that the webpki verifier builders and verifiers hold much easier: we no long have to do as many contortions to convert to a `&[&dyn CertRevocationList]`.

## TODO

* [x] Wait for https://github.com/rustls/webpki/pull/203 to be merged
* [x] Wait for rustls-webpki v0.102.0-alpha.6 to be published
* [x] Remove `Cargo.toml` patch

### DIFF
--- a/rustls/src/webpki/client_verifier.rs
+++ b/rustls/src/webpki/client_verifier.rs
@@ -2,20 +2,16 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 
 use pki_types::{CertificateDer, CertificateRevocationListDer, UnixTime};
-use webpki::{
-    BorrowedCertRevocationList, OwnedCertRevocationList, RevocationCheckDepth, UnknownStatusPolicy,
-};
+use webpki::{CertRevocationList, RevocationCheckDepth, UnknownStatusPolicy};
 
-use super::{borrow_crls, crl_error, pki_error, VerifierBuilderError};
+use super::{pki_error, VerifierBuilderError};
 use crate::verify::{
     ClientCertVerified, ClientCertVerifier, DigitallySignedStruct, HandshakeSignatureValid,
     NoClientAuth,
 };
+use crate::webpki::parse_crls;
 use crate::webpki::verify::{verify_signed_struct, verify_tls13, ParsedCertificate};
-use crate::{
-    CertRevocationListError, DistinguishedName, Error, RootCertStore, SignatureScheme,
-    WebPkiSupportedAlgorithms,
-};
+use crate::{DistinguishedName, Error, RootCertStore, SignatureScheme, WebPkiSupportedAlgorithms};
 
 /// A builder for configuring a `webpki` client certificate verifier.
 ///
@@ -178,14 +174,7 @@ impl ClientCertVerifierBuilder {
         Ok(Arc::new(WebPkiClientVerifier::new(
             self.roots,
             self.root_hint_subjects,
-            self.crls
-                .into_iter()
-                .map(|der_crl| {
-                    BorrowedCertRevocationList::from_der(der_crl.as_ref())
-                        .and_then(|crl| crl.to_owned())
-                        .map_err(crl_error)
-                })
-                .collect::<Result<Vec<_>, CertRevocationListError>>()?,
+            parse_crls(self.crls)?,
             self.revocation_check_depth,
             self.unknown_revocation_policy,
             self.anon_policy,
@@ -249,7 +238,7 @@ impl ClientCertVerifierBuilder {
 pub struct WebPkiClientVerifier {
     roots: Arc<RootCertStore>,
     root_hint_subjects: Vec<DistinguishedName>,
-    crls: Vec<OwnedCertRevocationList>,
+    crls: Vec<CertRevocationList<'static>>,
     revocation_check_depth: RevocationCheckDepth,
     unknown_revocation_policy: UnknownStatusPolicy,
     anonymous_policy: AnonymousClientPolicy,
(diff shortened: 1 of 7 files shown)


djc:
Would it be Hard to stack #1547 on top of this one instead of the other way around?

cpu:
> Would it be Hard to stack https://github.com/rustls/rustls/pull/1547 on top of this one instead of the other way around?

I don't think it would be too hard, but it does mean that it can't be merged until the webpki parts land and we cut a new alpha (neither very big blockers). Do you feel strongly? #1547 has a +1 from Ctz so it feels closer to being mergeable than this and the webpki PR it uses.

djc:
It seems like the proper order to me just because it means there will be a bunch of churn from porting the code added in #1547 to the new webpki API. But if you prefer to merge #1547 first I can definitely live with that.

cpu:
> It seems like the proper order to me just because it means there will be a bunch of churn from porting the code added in https://github.com/rustls/rustls/pull/1547 to the new webpki API

I think 264bae2173b7697bb85f5edcd8ac81cbb5c9586b is the sum of the churn and it seems pretty minimal to me.

> In that case I will probably skip doing another round on https://github.com/rustls/rustls/pull/1547 and check back in with the rebased version of this.

Ok, that sounds good to me. I'm happy to revisit parts from #1547 if you find anything outside the scope of the rebased content.

djc:
> I think [264bae2](https://github.com/rustls/rustls/commit/264bae2173b7697bb85f5edcd8ac81cbb5c9586b) is the sum of the churn and it seems pretty minimal to me.

Okay, that doesn't seem too bad!

rustls/src/webpki/mod.rs
```diff
@@ -104,13 +107,13 @@ fn crl_error(e: webpki::Error) -> CertRevocationListError {
     }
 }
 
-fn borrow_crls(
-    crls: &Vec<webpki::OwnedCertRevocationList>,
-) -> Vec<&dyn webpki::CertRevocationList> {
-    #[allow(trivial_casts)] // Cast to &dyn trait is required.
+fn parse_crls(
+    crls: Vec<CertificateRevocationListDer>,
+) -> Result<Vec<CertRevocationList>, CertRevocationListError> {
```
djc, line 112: Let's make the `CertRevocationList` lifetime explicit here? (And the one for `CertificateRevocationListDer`, I suppose.)

cpu: Fixed!

cpu:
Rebased on `main`, addressed feedback, and updated the PR desc. Once https://github.com/rustls/webpki/pull/203 lands I can flip this out of draft status and should be able to resolve the few failed CI tasks.

cpu:
Leaving a reminder here: it'd be nice to get a new Rustls alpha after this lands so I can rework the FFI side in https://github.com/rustls/rustls-ffi/pull/341.

cpu:
@djc This is ready for a once-over now if you're interested.

rustls/src/webpki/client_verifier.rs
```diff
@@ -301,21 +290,19 @@ impl ClientCertVerifier for WebPkiClientVerifier {
...
-            if matches!(
-                self.unknown_revocation_policy,
-                webpki::UnknownStatusPolicy::Allow
-            ) {
-                builder = builder.allow_unknown_status();
-            }
-            Some(builder.build())
+            Some(
+                webpki::RevocationOptionsBuilder::new(&crl_refs)
+                    .unwrap()
```
djc, line 301: Add a note about why this is okay?

rustls/src/webpki/server_verifier.rs
```diff
@@ -253,22 +241,20 @@ impl ServerCertVerifier for WebPkiServerVerifier {
...
-            if matches!(
-                self.unknown_revocation_policy,
-                webpki::UnknownStatusPolicy::Allow
-            ) {
-                builder = builder.allow_unknown_status();
-            }
-            Some(builder.build())
+            Some(
+                webpki::RevocationOptionsBuilder::new(crl_refs.as_slice())
+                    .unwrap()
```
djc, line 253: Here too.

djc (approved):
Did you want to bump the rustls alpha version so you can continue moving downstream into -ffi?

rustls/src/webpki/client_verifier.rs
cpu: Good call. I think I had something like that in an earlier pass but lost it along the way. Fixed up.

rustls/src/webpki/server_verifier.rs
cpu: Fixed too.

cpu:
@ctz This could use a :mag: from you when you have a chance. Thanks!

ctz (approved):

cpu:
> cpu force-pushed the cpu-1541-crl-for-server-cert-ergonomics branch from c12ce0f to 3c58a24 

Rebased to resolve conflicts after #1553 landed first.

Issue #1549, opened 2023-10-23
Separate server hints list from accepted trust anchors


cpu:
Presently the `WebPkiClientVerifier` is configured with an `Arc<RootCertStore>` that's used for two purposes:

1. For the set of trust anchors used when verifying a given client certificate.
2. For the slice of `DistinguishedName`'s sent to peers as the server hints list, for e.g. in a TLS 1.3 `certificate_authorities` extension in a `CertificateRequest` message.

In [NIST SP.800 Section 3.5.4](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-52r2.pdf) they emphasize the utility of having the set of accepted trust anchors be separately configurable from the `DistinguishedNames` hints:

> The hints list shall be distinct from the server’s trust anchor store. In other words, the server shall continue to only populate its trust anchor store with the trust anchor of the server’s PKI domain and the domains it needs to trust directly for client authentication. Note that the distinction between the server hints list and the server’s own trust store is as follows: 1) the hints list is the list of trust anchors that a potential client might trust, and 2) the server’s trust store is the list of trust anchors that the server explicitly trusts.

This issue is to track a change to the `WebPkiClientCertVerifier` to allow configuring these elements separately in order to facilitate this sort of use-case.

jsha:
I found this requirement pretty confusing until I read the rest of section 3.5.4, so pasting it here for everyone's benefit:

> 3.5.4 Server Hints List
> Clients may use the list of trust anchors sent by the server in the CertificateRequest message to
determine if the client’s certification path terminates at one of these trust anchors. The list sent
by the server is known as a “hints list.” When the server and client are in different PKI domains
and the trust is established via direct cross-certification between the two PKI domains (i.e., the
server PKI domain and the client PKI domain) or via transitive cross-certification (i.e., through
cross-certifications among multiple PKI domains), the client may erroneously decide that its
certificate will not be accepted by the server since the client’s trust anchor is not sent in the hints
list. To mitigate this failure, the server shall either: 1) maintain the trust anchors of the various
PKIs whose subscribers are the potential clients for the server and include them in the hints list
or 2) be configured to send an empty hints list so that the client can always provide a certificate it
possesses. The hints list shall be distinct from the server’s trust anchor store. In other words,
the server shall continue to only populate its trust anchor store with the trust anchor of the
server’s PKI domain and the domains it needs to trust directly for client authentication. Note that
[...]

PR #1553, opened 2023-10-26
Separate server hints list from accepted trust anchors

cpu:
This branch resolves https://github.com/rustls/rustls/issues/1549 by updating the default webpki client certificate verifier to allow configuring a server's client root certificate subject hints list separately from the subjects present in the root cert store used to verify client certificates. This is a useful capability in the presence of cross-signing when there are circumstances (described in #1549 and the updated rustdoc comments in this branch) where a server may want to hint to a client that it accepts a certificate with a given subject, without adding it to the server's trust store. 

See individual commit messages for more detail.

### DIFF
--- a/rustls/examples/internal/bogo_shim.rs
+++ b/rustls/examples/internal/bogo_shim.rs
@@ -200,7 +200,7 @@ impl server::danger::ClientCertVerifier for DummyClientAuth {
         self.mandatory
     }
 
-    fn client_auth_root_subjects(&self) -> &[DistinguishedName] {
+    fn root_hint_subjects(&self) -> &[DistinguishedName] {
         &[]
     }
 
@@ -315,7 +315,7 @@ struct FixedSignatureSchemeClientCertResolver {
 impl client::ResolvesClientCert for FixedSignatureSchemeClientCertResolver {
     fn resolve(
         &self,
-        acceptable_issuers: &[&[u8]],
+        root_hint_subjects: &[&[u8]],
         sigschemes: &[SignatureScheme],
     ) -> Option<Arc<sign::CertifiedKey>> {
         if !sigschemes.contains(&self.scheme) {
@@ -323,7 +323,7 @@ impl client::ResolvesClientCert for FixedSignatureSchemeClientCertResolver {
         }
         let mut certkey = self
             .resolver
-            .resolve(acceptable_issuers, sigschemes)?;
+            .resolve(root_hint_subjects, sigschemes)?;
         Arc::make_mut(&mut certkey).key = Arc::new(FixedSignatureSchemeSigningKey {
             key: certkey.key.clone(),
             scheme: self.scheme,
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -86,20 +86,28 @@ pub trait ClientSessionStore: Send + Sync {
 /// A trait for the ability to choose a certificate chain and
 /// private key for the purposes of client authentication.
 pub trait ResolvesClientCert: Send + Sync {
-    /// With the server-supplied acceptable issuers in `acceptable_issuers`,
-    /// the server's supported signature schemes in `sigschemes`,
-    /// return a certificate chain and signing key to authenticate.
+    /// Resolve a client certificate chain/private key to use as the client's
+    /// identity.
     ///
-    /// `acceptable_issuers` is undecoded and unverified by the rustls
-    /// library, but it should be expected to contain a DER encodings
-    /// of X501 NAMEs.
+    /// `root_hint_subjects` is an optional list of certificate authority
+    /// subject distinguished names that the client can use to help
+    /// decide on a client certificate the server is likely to accept. If
+    /// the list is empty, the client should send whatever certificate it
+    /// has. The hints are expected to be DER-encoded X.500 distinguished names,
+    /// per [RFC 5280 A.1]. See [`crate::DistinguishedName`] for more information
+    /// on decoding with external crates like `x509-parser`.
     ///
-    /// Return None to continue the handshake without any client
+    /// `sigschemes` is the list of the [`SignatureScheme`]s the server
+    /// supports.
+    ///
+    /// Return `None` to continue the handshake without any client
     /// authentication.  The server may reject the handshake later
     /// if it requires authentication.
+    ///
+    /// [RFC 5280 A.1]: https://www.rfc-editor.org/rfc/rfc5280#appendix-A.1
     fn resolve(
         &self,
-        acceptable_issuers: &[&[u8]],
+        root_hint_subjects: &[&[u8]],
         sigschemes: &[SignatureScheme],
     ) -> Option<Arc<sign::CertifiedKey>>;
 
--- a/rustls/src/webpki/client_verifier.rs
+++ b/rustls/src/webpki/client_verifier.rs
@@ -142,8 +142,16 @@ impl ClientCertVerifierBuilder {
             .supported_algs
             .ok_or(VerifierBuilderError::NoSupportedAlgorithms)?;
 
+        let subjects = self
+            .roots
+            .roots
+            .iter()
+            .map(|ta| DistinguishedName::in_sequence(ta.subject.as_ref()))
+            .collect();
+
         Ok(Arc::new(WebPkiClientVerifier::new(
             self.roots,
+            subjects,
             self.crls
                 .into_iter()
                 .map(|der_crl| {
@@ -256,6 +264,7 @@ impl WebPkiClientVerifier {
     /// * `supported_algs` specifies which signature verification algorithms should be used.
     pub(crate) fn new(
         roots: Arc<RootCertStore>,
+        subjects: Vec<DistinguishedName>,
         crls: Vec<OwnedCertRevocationList>,
         revocation_check_depth: RevocationCheckDepth,
         unknown_revocation_policy: UnknownStatusPolicy,
@@ -263,15 +272,11 @@ impl WebPkiClientVerifier {
         supported_algs: WebPkiSupportedAlgorithms,
     ) -> Self {
         Self {
-            subjects: roots
-                .roots
-                .iter()
-                .map(|ta| DistinguishedName::in_sequence(ta.subject.as_ref()))
-                .collect(),
+            roots,
+            subjects,
             crls,
             revocation_check_depth,
             unknown_revocation_policy,
-            roots,
             anonymous_policy,
             supported_algs,
         }
--- a/rustls/src/webpki/client_verifier.rs
+++ b/rustls/src/webpki/client_verifier.rs
@@ -23,6 +23,7 @@ use crate::{
 #[derive(Debug, Clone)]
 pub struct ClientCertVerifierBuilder {
     roots: Arc<RootCertStore>,
+    root_hint_subjects: Vec<DistinguishedName>,
     crls: Vec<CertificateRevocationListDer<'static>>,
     revocation_check_depth: RevocationCheckDepth,
     unknown_revocation_policy: UnknownStatusPolicy,
@@ -33,6 +34,7 @@ pub struct ClientCertVerifierBuilder {
 impl ClientCertVerifierBuilder {
     pub(crate) fn new(roots: Arc<RootCertStore>) -> Self {
         Self {
+            root_hint_subjects: roots.subjects(),
             roots,
             crls: Vec::new(),
             anon_policy: AnonymousClientPolicy::Deny,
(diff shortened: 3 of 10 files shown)


rustls/examples/internal/bogo_shim.rs
```diff
@@ -200,7 +200,7 @@ impl server::danger::ClientCertVerifier for DummyClientAuth {
         self.mandatory
     }
 
-    fn client_auth_root_subjects(&self) -> &[DistinguishedName] {
+    fn client_auth_hint_root_subjects(&self) -> &[DistinguishedName] {
```
djc, line 203: How about we get rid of the client_auth_ prefix for these methods? The context seems fairly obvious from the trait name.

I also don't love the rest of this name. Maybe we can skip subjects as well, given that the type is DistinguishedName? root_hints() seems like a pretty good name.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -315,15 +315,15 @@ struct FixedSignatureSchemeClientCertResolver {
 impl client::ResolvesClientCert for FixedSignatureSchemeClientCertResolver {
     fn resolve(
         &self,
-        acceptable_issuers: &[&[u8]],
+        hint_ca_subjects: &[&[u8]],
```
djc, line 318: I think we should be consistent about root vs ca. Maybe this could be root_subject_hints?

rustls/src/client/client_conn.rs
```diff
@@ -86,20 +86,28 @@ pub trait ClientSessionStore: Send + Sync {
 /// A trait for the ability to choose a certificate chain and
 /// private key for the purposes of client authentication.
 pub trait ResolvesClientCert: Send + Sync {
-    /// With the server-supplied acceptable issuers in `acceptable_issuers`,
-    /// the server's supported signature schemes in `sigschemes`,
-    /// return a certificate chain and signing key to authenticate.
+    /// Optionally resolve a client certificate chain/private key to use
```
djc, line 89: Tighter headline?

djc (commented):
Nice work. Early feedback...

rustls/src/webpki/client_verifier.rs
```diff
@@ -42,6 +44,35 @@ impl ClientCertVerifierBuilder {
...
+
+    /// Add an additional [`DistinguishedName`] to the list of trust anchor subjects sent to the
+    /// client as hints of acceptable trust anchors during the TLS handshake.
+    ///
+    /// By default, the client cert verifier will [`RootCertStore::subjects`] with the root cert
+    /// store configured for client certificate authentication for the hint subjects.
+    ///
+    /// See [`ClientCertVerifier::client_auth_hint_root_subjects`] for more information on
+    /// circumstances where you may want to override the default hint subjects.
+    pub fn add_hint_subject(mut self, subject: DistinguishedName) -> Self {
```
djc, line 71: Given that a one-element array implements IntoIter, not sure this method pulls its weight?

djc (commented):
Okay, one more!

rustls/src/webpki/client_verifier.rs
cpu: > Given that a one-element array implements IntoIter, not sure this method pulls its weight?

True, but `add_hint_subject` _adds_ to the existing `Vec` whereas the other method replaces it outright.

I think there's value in both: if we only have this method then it isn't possible to replace the defaults, or specify an empty list to send. If we only have the other method then you have to clone the `RootCertStore` from the one used for authentication, add your extra subject to it, and then use the other fn to replace the whole list.

For a use-case like the one described in #1549 where you're happy with the default but just want to add one or two other subjects from intermediate certs it seemed helpful to have this fn.

cpu: Maybe there's an argument to be made that this should be `add_hint_subjects(mut self, subjects: impl IntoIterator<Item = DistinguishedName>) ` for consistency and to make it easier to use with multiple names?

djc: Ah, sorry, missed the add vs replace difference. I guess it comes down to whether there's a use case for removing some of the hints that refer back to the set anchors...

cpu: > I guess it comes down to whether there's a use case for removing some of the hints that refer back to the set anchors...

I'm not sure about removing one or two, but replacing the whole list with the empty set should be supported based on the text Jsha cited in #1549:
>  To mitigate this failure, the server shall either: 1) maintain the trust anchors of the various PKIs whose subscribers are the potential clients for the server and include them in the hints list or 2) be configured to send an empty hints list so that the client can always provide a certificate it possesses.

djc: Okay, in that case I think your method set probably makes sense.

rustls/examples/internal/bogo_shim.rs
djc: Alternatively, `root_hint_subjects` might align better with the `root_hints` from elsewhere? Or we could just use `root_hint_subjects` across the board maybe.

rustls/examples/internal/bogo_shim.rs
cpu: > How about we get rid of the client_auth_ prefix for these methods?

Done :heavy_check_mark: 

> I also don't love the rest of this name. Maybe we can skip subjects as well, given that the type is DistinguishedName? root_hints() seems like a pretty good name.

I went with `root_hint_subjects` and standardized on that terminology throughout. I think keeping the "subjects" part is useful since `DistinguishedName`'s appear in other contexts as well (issuer, for example).

rustls/examples/internal/bogo_shim.rs
cpu: I standardized on `root_hint_subjects` throughout.

rustls/src/client/client_conn.rs
cpu: Tightened this up to "Resolve a client certificate chain/private key to use as the client's identity.". I think from context it's clear enough that you make the resolution decision informed by the arguments, and the return type being `Option` gives enough of a hint that you can return `None`. We also spell out more detail about all of the above in the full comment past the headline.

rustls/src/webpki/client_verifier.rs
cpu: > Maybe there's an argument to be made that this should be add_hint_subjects(mut self, subjects: impl IntoIterator<Item = DistinguishedName>)  for consistency and to make it easier to use with multiple names?

I ended up making this change. If you only have one `DistinguishedName` it's easy enough to make a slice to give, and in other cases the symmetry is helpful. 

> I wonder if we should use a different name for the replacing method? The with_() doesn't feel very explicit about appending vs replacing.

I left the name as-is (missed your edit), but I could change it to `replace_root_hint_subjects` if you think that would be more explicit.

ctz: i think realistically either people will want to add to the existing hints, or have them empty -- i can't see why people would want to send _some_ hints, but not the ones for actual issuer names they trust.

how about `clear_root_hint_subjects(mut self)` (a good venue for recording that this means "clients will make a free choice of certificate to send" in a comment) and `add_root_hint_subjects(mut self, subjects: impl IntoIterator<Item = DistinguishedName>)`?

of course, those can be combined to set whatever contents someone likes.

djc: Yeah, that's probably the more explicit option.

rustls/tests/client_cert_verifier.rs
```diff
@@ -150,11 +150,7 @@ impl MockClientVerifier {
     pub fn new(verified: fn() -> Result<ClientCertVerified, Error>, kt: KeyType) -> Self {
         Self {
             verified,
-            subjects: get_client_root_store(kt)
-                .roots
-                .iter()
-                .map(|ta| DistinguishedName::in_sequence(ta.subject.as_ref()))
-                .collect(),
+            subjects: get_client_root_store(kt).subjects(),
```
ctz, line 153: nb. tests in this file aren't actually running right now, fixing that in #1556

cpu: Ah! Good catch, thank you.

rustls/src/webpki/client_verifier.rs
cpu: SGTM

cpu: Updated to replace w/ `clear_root_hint_subjects`.

cpu:
I think this branch is ready for re-review when folks have a chance.

ctz (approved):
looking good!

rustls/src/webpki/client_verifier.rs
```diff
@@ -142,12 +142,7 @@ impl ClientCertVerifierBuilder {
             .supported_algs
             .ok_or(VerifierBuilderError::NoSupportedAlgorithms)?;
 
-        let root_hint_subjects = self
-            .roots
-            .roots
-            .iter()
-            .map(|ta| DistinguishedName::in_sequence(ta.subject.as_ref()))
-            .collect();
+        let root_hint_subjects = self.roots.subjects();
```
djc: Consider going back to the way it was, inlining the initialization of the `root_hint_subjects` into the `WebPkiClientVerifier::new()` call?

rustls/src/webpki/client_verifier.rs
```diff
@@ -42,6 +44,39 @@ impl ClientCertVerifierBuilder {
         }
     }
 
+    /// Clear the list of trust anchor subjects sent to the client as hints of acceptable trust
```
djc, line 47: Maybe "Clear the list of trust anchor subjects" is enough as a headline?

rustls/src/webpki/client_verifier.rs
```diff
@@ -42,6 +44,39 @@ impl ClientCertVerifierBuilder {
...
+    /// to send.
+    ///
+    /// See [`ClientCertVerifier::root_hint_subjects`] for more information on
+    /// circumstances where you may want to clear the default hint subjects.
+    pub fn clear_root_hint_subjects(mut self) -> Self {
+        self.root_hint_subjects = Vec::default();
+        self
+    }
+
+    /// Add additional [`DistinguishedName`]s to the list of trust anchor subjects sent to the
```
djc, line 62: Similar.

rustls/src/webpki/client_verifier.rs
```diff
@@ -281,6 +281,8 @@ impl WebPkiClientVerifier {
     /// Construct a new `WebpkiClientVerifier`.
     ///
     /// * `roots` is a list of trust anchors to use for certificate validation.
+    /// * `root_hint_subjects` is a list of distinguished names to use for hinting acceptable
```
djc, line 284: Squash this into the earlier commit that renamed these and added a bunch of docs?

rustls/tests/api.rs
```diff
@@ -1349,6 +1349,55 @@ fn client_cert_resolve_default() {
...
+
+#[test]
+fn client_cert_resolve_server_added_hint() {
+    // Test that a server can add an extra subject above/beyond those found in its trust store
+    // and the client cert resolver gets the expected arguments.
+    let extra_name = b"0\x1a1\x180\x16\x06\x03U\x04\x03\x0c\x0fponyland IDK CA".to_vec();
+    for key_type in ALL_KEY_TYPES.into_iter() {
+        let expected_hint_subjects = match key_type {
+            KeyType::Rsa => {
+                vec![
```
djc, line 1374: I think there's a fair bit of low-hanging deduplication fruit here?

djc (approved):
LGTM modulo nits!

rustls/src/client/client_conn.rs
```diff
@@ -86,20 +86,30 @@ pub trait ClientSessionStore: Send + Sync {
...
-    /// of X501 NAMEs.
+    /// `root_hint_subjects` is an optional list of certificate authority
+    /// subject distinguished names that the client can use to help
+    /// decide on a client certificate the server is likely to accept. If
+    /// the list is empty, the client should send whatever certificate it
+    /// has. The hint subjects are provided undecoded and unverified by
+    /// the rustls library. The hints are expected to be DER-encoded
+    /// X.500 distinguished names, per [RFC 5280 A.1]. See
+    /// [`crate::DistinguishedName`] for more information on decoding with
+    /// external crates like `x509-parser`.
```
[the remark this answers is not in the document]
cpu: I carried that language forward from the existing commit, but I agree that the next sentence does a better job of communicating the main point (that Rustls' hasn't parsed this data).

rustls/src/webpki/client_verifier.rs
cpu: I don't think that will work - the `self.roots` argument comes first in the `WebPkiClientVerifier::new` argument list, and using `self.roots.subjects()` as the arg afterwards is trying to use a moved value. We have to pull it out before moving `roots`.

rustls/src/webpki/client_verifier.rs
cpu: Fixed.

rustls/src/webpki/client_verifier.rs
cpu: Fixed.

rustls/src/webpki/client_verifier.rs
cpu: Done :ballot_box_with_check:

rustls/tests/api.rs
cpu: Tidied this up.

rustls/src/webpki/client_verifier.rs
cpu: Ultimately we end up in the state you want by the end of the commits once the `root_hint_subjects` are stored in the builder separately from `roots`: https://github.com/cpu/rustls/blob/3d0960b3a3f7077d9d6e23cc56d8825134b1ba93/rustls/src/webpki/client_verifier.rs#L178-L181

PR #1558, opened 2023-10-27
docs: create issue templates

cpu:
Add Bug report, Feature request, and Dependency update issue templates to help prompt users into providing the information that will get them the best help. When a user creates a new issue they'll be asked to choose from one of these three templates, and will get a pre-populated new issue description based on their choice.

This is primarily motivated by the number of issues we see where folks are asking for semver compatible dependencies be updated (e.g. because they're trying to avoid multiple versions of a transitive dependency) without understanding they can use `cargo update` to pick up the required updates. This is a common source of confusion for folks new to the Rust ecosystem (I've certainly been confused about it myself in the past!).

My hope is that by having an explicit option for dependency update requests that itself has a checklist for steps like running `cargo update` we can cut down on the number of issues that get filed for this category of problem.

### DIFF
--- a/.github/ISSUE_TEMPLATE/dependency-update.md
+++ b/.github/ISSUE_TEMPLATE/dependency-update.md
@@ -1,0 +1,33 @@
+---
+name: Dependency Update
+about: Request a dependency be updated
+title: Dependency update request
+labels: ''
+assignees: ''
+
+---
+
+<!--
+Please note that we are only interested in **semver-incompatible** update requests. Updates to dependencies that are 
+semver-compatible can be done in dependent projects without needing changes in this repository.
+
+For example, if you are here because you believe Rustls is bringing in dependency `foo` at version `0.2.1` and
+you wish Rustls used `0.2.2` instead, you should not file an issue and instead should run `cargo update` in your
+dependent project. It would only be appropriate to file an issue if you require Rustls use `foo` at version `0.3.0+`.
+-->
+
+**Checklist**
+* [ ] I've searched the issue tracker for similar requests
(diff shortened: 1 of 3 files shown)


ctz (approved):
Thanks, this is a good start!

.github/ISSUE_TEMPLATE/dependency-update.md
```diff
@@ -0,0 +1,25 @@
...
+assignees: ''
+
+---
+
+**Checklist**
+* [ ] I've searched the issue tracker for similar requests
+* [ ] I've confirmed I've run `cargo update`
+* [ ] I've read the Cargo docs on [specifying dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)
+
+**Is your feature request related to a problem? Please describe.**
```
djc, line 15: I feel like we should skip a bunch of these questions and maybe state explicitly that we're only interested in semver-incompatible bump requests (and maybe quickly explain what that is given that this is like a honeypot for unsophisticated users).

djc (approved):

.github/ISSUE_TEMPLATE/dependency-update.md
cpu: @djc WDYT about the revision I just pushed? It adds more explicit guidance/explanation in a comment section that won't be rendered if left-in place by the user before submitting.

I left some question/checkbox content because I think it's still helpful to have. If you feel strongly I can revisit.

djc: Seems okay for now.

Issue #1555, opened 2023-10-26
Debug bounds for customization traits


jsha:
A quick review of some of our customization traits:

```
pub trait CryptoProvider: Send + Sync + Debug + 'static {
pub trait KeyLog: Send + Sync {
pub trait ClientSessionStore: Send + Sync {
pub trait Signer: Send + Sync {
pub trait ResolvesServerCert: Send + Sync {
```

Our newest customization trait (CryptoProvider) has a Debug bound, which is great. Most of our others don't.

That means some of our internal structs, like ServerConfig, implement a custom Debug that doesn't mention these traits:

https://github.com/rustls/rustls/blob/829471689e6c414b1bc2d51aef541d2ed1a31335/rustls/src/server/server_conn.rs#L340

Implementing a custom Debug isn't by itself a problem; we also do that in places to hide data that shouldn't wind up in logs. But when print the Debug format of a server config, for instance, it would be useful to know the type of the `CryptoProvider`, `KeyLog`, etc.

Since we have a breaking change coming up, it might be a useful time to add Debug bounds on these. The idea being that most implementations would implement something very simple, like the name of the implementing type. For instance, right now `struct Ring;` has `#[derive(Debug)]` so it just prints `Ring`.

cpu:
Adding the `Debug` bound to the `ResolvesServerCert` seems like a good idea to me :+1:  For `KeyLog`, `ClientSessionStore`, and `Signer` I'm slightly hesitant.

As you mention, you need to take care implementing `Debug` for types that might contain sensitive data. I think most 
implementations of these traits are going to fall into that category. Adding a `Debug` bound will force downstream implementations to add code to satisfy the new requirement and I worry it will be tempting to `#[derive(Debug)]` as part of tidying up an upgrade's breaking changes without considering the potential for leaking something sensitive. We can call out the risk explicitly but I'm not sure I'm 100% convinced the reward is worth the risk. WDYT?

cpu:
While we're considering this, I think to drop the custom `Debug` impl on `ServerConfig` and `ClientConfig` we'd also need to add a `Debug` bound on `StoresServerSessions`, `ProducesTickets`, and `ResolvesClientCert`

jsha:
Yeah, I think your logic about not requiring Debug for types at high risk of containing secrets makes sense. `KeyLog` probably doesn't contain secrets in general, but also I suspect most users will use one of the built-ins (`KeyLogFile` or `NoKeyLog`).

[...]

PR #1559, opened 2023-10-27
Debug bounds for all non-internal traits

cpu:
Here's a branch adding `Debug` bounds to all of the traits that exist outside of the internal namespace. Where possible, hand written `Debug` impl's are replaced with derived impls.  This will help consumers of Rustls debug, and will help us avoid having to keep hand-crafted `Debug` impls in sync with changes to their underlying types over time.

A few of these traits (e.g. `ClientSessionStore`, `Signer`) are likely to contain sensitive data. In our `Debug` implementations for these types care has been taken to avoid leaking sensitive data. There might be an argument that we shouldn't impose a `Debug` bound on these traits at all. If that's the consensus we reach I'll drop the commits specific to the types we don't think should have bounds. In general I figured it was easier to start with everything and then whittle the branch down.

When I had to add/update imports I also added commits to tidy the imports based on [CONTRIBUTING.md](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#imports). If that's too noisey I can drop those commits.

Resolves https://github.com/rustls/rustls/issues/1555


cpu:
FWIW I'm going to let this branch collect some conflicts for a bit until its been reviewed by folks. I'll address the conflicts when we're closer to thinking this could be merged.

rustls/src/key_log_file.rs
```diff
@@ -106,6 +116,16 @@ impl KeyLog for KeyLogFile {
     }
 }
 
+impl Debug for KeyLogFile {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        if let Ok(key_log_file) = self.0.try_lock() {
+            write!(f, "{:?}", key_log_file)
+        } else {
+            write!(f, "KeyLogFile {{ <locked> }}")
+        }
```
djc, line 125: Nit: I'd write it as a match, since it's one less line.

rustls/tests/api.rs
```diff
@@ -2931,6 +2932,14 @@ impl KeyLog for KeyLogToVec {
     }
 }
 
+impl Debug for KeyLogToVec {
```
djc, line 2937: Why not derive it?

rustls/src/key_log.rs
```diff
@@ -10,7 +12,7 @@
 ///
 /// See [`KeyLogFile`](crate::KeyLogFile) that implements the standard
 /// `SSLKEYLOGFILE` environment variable behaviour.
-pub trait KeyLog: Send + Sync {
+pub trait KeyLog: Send + Sync + Debug {
```
djc, line 15: Suggestion: I like to order bounds in order of least common to most common, so I'd probably put `Debug` before `Send` and `Sync`.

rustls/src/client/client_conn.rs
```diff
@@ -1,3 +1,12 @@
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+use core::fmt;
```
djc, line 3: Group these into `core::{fmt, mem, io}` while you're at it?

rustls/src/server/handy.rs
```diff
@@ -1,3 +1,12 @@
+use alloc::string::{String, ToString};
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+use std::collections;
```
djc, line 5: Pre-existing issue, but it seems more consistent with our import style to import `HashMap` directly rather than importing `collections` here.

djc (approved):
Nice!

rustls/tests/api.rs
cpu: We don't want to include `client_random` and `secret`. I'll add a comment indicating that to `fmt`.

djc: It's just test code, though... so not that important?

cpu: Fair point. I'll switch to deriving the trait.

ctz:
I would really rather we do "tidy imports"-style commits in one dedicated PR, all at once, in an automated fashion.

rustls/src/key_log_file.rs
cpu: Fixed.

rustls/tests/api.rs
cpu: Fixed.

rustls/src/key_log.rs
cpu: Fixed throughout

rustls/src/client/client_conn.rs
cpu: I left the import tidying out at request of ctz.

rustls/src/server/handy.rs
cpu: Fixed in a separate commit at the end.

cpu:
> https://github.com/rustls/rustls/pull/1577 is a line in the sand for this.

Thanks! I dropped all of the import tidying commits from this branch.

rustls/src/client/handy.rs
```diff
@@ -163,6 +165,15 @@ impl client::ClientSessionStore for ClientSessionMemoryCache {
     }
 }
 
+impl Debug for ClientSessionMemoryCache {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
```
ctz, line 169: Extremely low-grade nit: as above, I'd import `fmt` and then qualify `fmt::Debug`, `fmt::Formatter`, `fmt::Result`

rustls/src/client/client_conn.rs
```diff
@@ -400,7 +382,7 @@ pub enum ServerName {
     IpAddress(IpAddr),
 }
 
-impl fmt::Debug for ServerName {
+impl Debug for ServerName {
```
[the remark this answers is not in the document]
cpu: SGTM. Fixed.

rustls/src/client/handy.rs
cpu: Fixed as well.

PR #1563, opened 2023-11-02
x509: reduce allocations for wrap_in_sequence

### LINKED ISSUE #1562: x509: heap allocation hotspot in ASN1-wrapping logic
**Checklist**

**Applicable Version(s)**
A list of versions and platforms you've tested with.

`rustls-0.21.7`

**Describe the bug**

I was analyzing the memory/heap profile of a (non-public) Rust application with https://github.com/KDE/heaptrack, and it highlighted an heap-allocation hotspot coming from the DER-wrapping logic in `rustls`:
https://github.com/rustls/rustls/blob/b776a5778ad333653670c34ff9125d8ae59b6047/rustls/src/x509.rs#L5-L26

Apparently this logic can be reached (in a loop) when building an HTTPS client with `hyper` plus `rustls`, when processing trust anchors, through https://docs.rs/rustls/latest/rustls/struct.OwnedTrustAnchor.html#method.from_subject_spki_name_constraints.

There are a number of byte-prepending calls in that logic which can negatively affect memory allocation patterns.
It may be useful to benchmark it (e.g. against https://docs.rs/webpki-roots/) and try to optimize that.

jsha:
Instead of taking a `Vec<u8>` and inserting bytes at the beginning, take a `&[u8]` and return a new vector containing those bytes plus a tag and a length.

This isn't the perfect approach for all situations, but for one of the main places we call wrap_in_sequence (DistinguishedName::in_sequence), it's optimal because the input is `&[u8]`, meaning we can't write to a previously existing `Vec<u8>` (which would potentially save allocations by using excess capacity at the end of the Vec).

In the process, change the one call site for `wrap_in_asn1_len` to call the new `asn1_wrap` function instead, which encodes a tag and length at the same time, reducing reallocations and copies.

This has a slight secondary benefit: the resulting Vec is exactly sized to what it holds, instead of following the doubling approach and possibly over-allocating. This saves a handful of bytes in a long-lived data structure.

Fixes #1562

### DIFF
--- a/rustls/src/crypto/ring/sign.rs
+++ b/rustls/src/crypto/ring/sign.rs
@@ -1,7 +1,7 @@
 use crate::enums::{SignatureAlgorithm, SignatureScheme};
 use crate::error::Error;
 use crate::sign::{Signer, SigningKey};
-use crate::x509::{wrap_in_asn1_len, wrap_in_sequence};
+use crate::x509::{asn1_wrap, wrap_in_sequence};
 
 use pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer};
 use ring::io::der;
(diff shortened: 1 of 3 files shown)


rustls/src/x509.rs
```diff
@@ -33,68 +46,65 @@ mod tests {
...
+            vec![0x30, 0x81, 0xff, 0x12, 0x12, 0x12],
+            wrap_in_sequence(&val)[..6]
+        );
     }
 
     #[test]
     fn test_large() {
         let mut val = Vec::new();
         val.resize(4660, 0x12);
         wrap_in_sequence(&mut val);
```
cpu, line 74: Clippy is flagging this call-site as using `&mut` without a need.

rustls/src/crypto/ring/sign.rs
```diff
@@ -212,16 +212,12 @@ impl EcdsaSigningKey {
...
+        let sec1_wrap = asn1_wrap(der::Tag::OctetString as u8, maybe_sec1_der);
 
-        let mut pkcs8 = Vec::with_capacity(pkcs8_prefix.len() + sec1_wrap.len() + 4);
-        pkcs8.extend_from_slice(pkcs8_prefix);
-        pkcs8.extend_from_slice(&sec1_wrap);
-        wrap_in_sequence(&mut pkcs8);
+        let mut pkcs8_inner = Vec::with_capacity(pkcs8_prefix.len() + sec1_wrap.len());
+        pkcs8_inner.extend_from_slice(pkcs8_prefix);
+        pkcs8_inner.extend_from_slice(&sec1_wrap);
+        let pkcs8 = wrap_in_sequence(&pkcs8_inner);
```
cpu, line 220: nit: it seems like this `pkcs8` binding doesn't add much. Maybe in-line on L222 in the `from_pkcs8` call?

cpu (approved):
Nice! Looks good to me.

Since this was a fix for a perf related issue it would be interesting to know if a tool like https://github.com/KDE/heaptrack shows the expected improvement under profiling.

jsha:
Pushed fixes.

> Since this was a fix for a perf related issue it would be interesting to know if a tool like https://github.com/KDE/heaptrack shows the expected improvement under profiling.

I can take a look at this. Might take me a minute to get heaptrack set up.

jsha:
On main:

```
heaptrack ./target/release/simpleclient
...
heaptrack stats:
        allocations:            380
        leaked allocations:     30
        temporary allocations:  102
```

On this branch:

```
heaptrack ./target/release/simpleclient
...
heaptrack stats:
        allocations:            372
        leaked allocations:     30
        temporary allocations:  98
```

djc (approved):
Nice!

cpu:
This feels small enough in scope that we can merge it without waiting on Ctz to be back.

Issue #1501, opened 2023-09-29
rustls client flagged as bot by Cloudflare


darind:
I am using rustls 0.21.7 on MacOS 13.5.2 to make a TLS request to `chat.openai.com:443`:

```
use std::io::{Read, Write};
use std::{sync::Arc, net::TcpStream};
use rustls::{Certificate, RootCertStore, ClientConfig, Stream, ClientConnection};

fn main() {
    let mut root_store = RootCertStore::empty();
    for cert in rustls_native_certs::load_native_certs().expect("could not load platform certs") {
        root_store
            .add(&Certificate(cert.0))
            .unwrap();
    }

    let mut config = ClientConfig::builder()
        .with_safe_defaults()
        .with_root_certificates(root_store)
        .with_no_client_auth();

    config.alpn_protocols.push("http/1.1".as_bytes().to_vec());

    let mut conn = ClientConnection::new(
        Arc::new(config),
        "chat.openai.com".try_into().unwrap()
    ).unwrap();

    let mut sock = TcpStream::connect("chat.openai.com:443").unwrap();
    let mut tls = Stream::new(&mut conn, & mut sock);

    tls.write("GET / HTTP/1.1\r\nHost: chat.openai.com\r\nUser-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36\r\nConnection: close\r\n\r\n".as_bytes()).unwrap();

    let mut buf = vec![0; 2048];
    let len = tls.read(&mut buf).unwrap();
    std::io::stdout().write_all(&buf[..len]).unwrap();
}
```

This returns `HTTP/1.1 403 Forbidden`.

I suspect that the JA3 fingerprint that rustls is generating is not matching with the one generated by popular web browsers and Cloudflare bot protection is detecting that the ClientHello is not coming from a whitelisted source.

Is it possible to control the list of tls extensions and cipher suites sent by a rustls client?

djc:
I think this is a duplicate of https://github.com/rustls/rustls/issues/1421. Have you done some testing to ascertain that the TLS-level fingerprinting is what is causing the issues here? (Like running the same minimal test with native-tls.)

darind:
Not sure about `native-tls`, but a simple curl on the command line works fine:

```
[...]

Issue #1421, opened 2023-08-28
Support modifying `ClientHello` to match JA3 fingerprinting


AceRogue:
I find `extra_exts` is set to be empty by default. So I will have to modify the rustls if I want to set different `extra_exts`? Is it reasonable to add  `extra_exts` in `ClientConfig`?

djc:
Why do you want to use it? It's only intended for QUIC.

AceRogue:
Because I may need to change JA3. Default tls fingerprint may be blocked by websites, such as Cloudflare. To change JA3 fingerprint will need to change CipherSuites, order of extra_exts and etc.

djc:
Do you have a concrete requirement/website that you need to pass JA3 for?

See some earlier discussion in #1125 and linked issues/PRs.

cpu:
Potentially also relevant; https://github.com/rustls/rustls/issues/1313

cpu:
I think we should focus on https://github.com/rustls/rustls/issues/1313 and not support open ended modification to the `ClientHello`. 

I'm going to close this issue, but if there's a strong case to be made for why Rustls should prioritize this general request for more control over the `ClientHello` we could revisit. I think it will be a challenging case to argue. In my experience most folks expressly interested in fine grained ability to evade fingerprinting will be better served by more specialized TLS libraries that have the resources to focus on that goal above others.

XOR-op:
@cpu Hi, I'm curious if Rustls would be open to contributions on this feature but might not give priority to it, or there is no intention to integrate it even with some contributions? If it's the former scenario, I'd like to offer some help.

The most important reason I want to integrate this feature into rustls is the ecosystem: there are so many downstream dependencies on rustls, and most of them don't provide a way to use some specialized TLS libraries (even if forked rustls). So it's way complicated for an end-user to have this feature if they have only a specialized TLS library, since they also need some higher-level libraries e.g. tokio-rustls, reqwest, trust-dns, etc. What's more, from the prototype I'm implementing, adding this feature doesn't need too many modifications, and won't affect users if they don't use this feature at all.

cpu:
:wave: @XOR-op 

> I'm curious if Rustls would be open to contributions on this feature but might not give priority to it, or there is no intention to integrate it even with some contributions? If it's the former scenario, I'd like to offer some help.

[...]

PR #1564, opened 2023-11-03
client: add `ClientHelloCamouflager` to reduce TLS fingerprinting risk

XOR-op:
There have been many discussions related to TLS fingerprinting, including #1421, #1501 and #1190. To get rid of this issue, a customizable modifier in `ClientConfig` is a viable approach. 

This design will replace the original cipher suites and extensions with user-defined values if `client_hello_camouflager` field is set in the config. This design will also use the original values constructed by rustls, in order to make users easily create their own masks.

This PR introduces little modification to the other part of rustls. If a user doesn't choose to use this feature, rustls should work the exactly same as the previous version. This PR also re-exports some structures, which are still public in stable 0.21 but removed in 0.22-beta.

The example included in this PR tries to pretend the traffic is sent by a Chrome 102 browser. The JA3 hash of the example is "cd08e31494f9531f560d64c695473da9", which can be found on many JA3 databases identified as Chrome.

### DIFF
--- a/examples/src/bin/fingerprint-resist.rs
+++ b/examples/src/bin/fingerprint-resist.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 use rustls::client::ClientHelloCamouflager;
 use rustls::internal::msgs::base::Payload;
-use rustls::internal::msgs::enums::{ECPointFormat, ExtensionType, PSKKeyExchangeMode};
+use rustls::internal::msgs::enums::{EcPointFormat, ExtensionType, PSKKeyExchangeMode};
 use rustls::internal::msgs::handshake::ClientExtension::*;
 use rustls::internal::msgs::handshake::{
     ClientExtension, ClientSessionTicket, ProtocolName, UnknownExtension,
@@ -85,7 +85,7 @@ fn main() {
                     NamedGroup::secp256r1,
                     NamedGroup::secp384r1,
                 ]),
-                EcPointFormats(vec![ECPointFormat::Uncompressed]),
+                EcPointFormats(vec![EcPointFormat::Uncompressed]),
                 SessionTicket(ClientSessionTicket::Request),
                 Protocols(vec![
                     // ProtocolName::from("h2".as_bytes().to_vec()),
--- a/examples/src/bin/fingerprint-resist.rs
+++ b/examples/src/bin/fingerprint-resist.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 use rustls::client::ClientHelloCamouflager;
 use rustls::internal::msgs::base::Payload;
-use rustls::internal::msgs::enums::{EcPointFormat, ExtensionType, PSKKeyExchangeMode};
+use rustls::internal::msgs::enums::{EcPointFormat, ExtensionType, PskKeyExchangeMode};
 use rustls::internal::msgs::handshake::ClientExtension::*;
 use rustls::internal::msgs::handshake::{
     ClientExtension, ClientSessionTicket, ProtocolName, UnknownExtension,
@@ -110,7 +110,7 @@ fn main() {
                     payload: Payload::new(vec![]),
                 }),
                 keyshare,
-                PresharedKeyModes(vec![PSKKeyExchangeMode::PSK_DHE_KE]),
+                PresharedKeyModes(vec![PskKeyExchangeMode::PSK_DHE_KE]),
                 SupportedVersions(vec![
                     ProtocolVersion::Unknown(0x5a5a),
                     ProtocolVersion::TLSv1_3,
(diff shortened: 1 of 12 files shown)


cpu:
Haven't reviewed this yet, but I have a few comments on things that would make that process easier:

1. A tidier commit history.
2. Passing CI - it looks like MSRV task is failing.
3. Unit test coverage. The example is nice, but test coverage is also important.
4. Finding and linking to the commit that removed the exports you're restoring. If there was a reason the exports were removed it would be helpful to have that context in hand when considering restoring them.

Thanks!

XOR-op:
Hi, a new version has been pushed.
> 1. A tidier commit history.
> 2. Passing CI - it looks like MSRV task is failing.
> 3. Unit test coverage. The example is nice, but test coverage is also important.

I have rebased to clean up commits. Now the MSRV requirement should be satisfied and a unit test has been added. 

> 4. Finding and linking to the commit that removed the exports you're restoring. If there was a reason the exports were removed it would be helpful to have that context in hand when considering restoring them.

The exports were removed in #1544, where `enum_builder` macro was refactored to hide unused API at that time. To generate a customized ClientHello, types used to construct ClientHello are needed to be public. Since ClientHello is standardized in TLS, exporting its components should not introduce many potential breaking changes in the future.

rustls/src/client/client_conn.rs
```diff
@@ -115,6 +115,17 @@ pub trait ResolvesClientCert: Send + Sync {
...
 
+/// A trait providing the ability to camouflage ClientHello to resist fingerprinting.
+/// Notice: wrongly configured cipher suites and extensions may result in failure in connection.
+pub trait ClientHelloCamouflager: Send + Sync {
+    /// Transform cipher suites created from rustls to camouflaged cipher suites.
+    fn transform_cipher_suites(&self, cipher_suites: Vec<CipherSuite>) -> Vec<CipherSuite>;
+
+    /// Transform extensions created from rustls to camouflaged extensions. Users may need to
+    /// reuse some provided extensions for a successful connection.
+    fn transform_extensions(&self, extensions: Vec<ClientExtension>) -> Vec<ClientExtension>;
```
djc, line 126: IIRC some extensions have to be a in a particular order to work. This interface would make it pretty easy to get that wrong.

examples/src/bin/fingerprint-resist.rs
```diff
@@ -0,0 +1,174 @@
...
+        .with_safe_defaults()
+        .with_root_certificates(root_store)
+        .with_no_client_auth();
+
+    config.alpn_protocols = vec!["http/1.1".as_bytes().to_vec()];
+
+    struct Chrome102;
+    impl ClientHelloCamouflager for Chrome102 {
+        fn transform_cipher_suites(&self, _cipher_suites: Vec<CipherSuite>) -> Vec<CipherSuite> {
+            vec![
```
djc, line 36: This example should probably at least take the input into account somehow.

djc (commented):
I'm not really convinced we should merge this, but here's some feedback.

rustls/src/client/client_conn.rs
XOR-op: According to RFC5246#7.4.1.4, extensions MAY appear in any order in TLS 1.2 , while TLS 1.3 has an exception of pre_shared_key which must be the last if exists. I am not convinced that we should perform this check, because I believe only users who know what they are doing should use this feature, and they have more ways to mess up anything if they want.

examples/src/bin/fingerprint-resist.rs
XOR-op: In this particular example, the goal is to masquerade as a Chrome browser, so it's unnecessary to use existing cipher suites. The interface is designed in this form because there are other legitimate use cases like inserting a GREASE cipher while keeping the others same for testing purposes. Nevertheless, a test has been added to actually use the interface.

ctz:
Thanks for working on this; I do appreciate the effort. I'm not super keen in general on any feature in the core rustls library to make it look more or less like any other TLS library for the purposes of evading fingerprinting. Though I think this PR is in the right direction, since it is putting some of that logic outside the core crate.

Unfortunately, over on #1475 I have a rough plan to delete the `ClientExtension` type (indeed, all types representing an individual TLS extension). That would break the new interface: because a) it removes the stored ordering of extensions, b) it removes in-memory representation of extensions that rustls doesn't understand.

idontneedgithub:
i would love for this to be accepted!

XOR-op:
Thanks for your comment. Recently I also experimented other approaches to achieve my goal, e.g. using external libraries together with original rustls. However, the integrity requirement of TLS made such trials way more difficult and fragile than modifying the internals of rustls. Nevertheless, I understand that this PR would not be compatible with further change and lack some sorts of polishing. Maintaining a dedicated fork could be a better idea and I hope #1475 could consider randomizing extension order as well.

PR #1566, opened 2023-11-03
disable subtle's default features

japaric:
suble depends on libstd by default and its default features are not being used


japaric:
this commit has been extracted from the no-std-support PR #1502 into its own PR as it can be landed independently ahead of that one

djc (approved):

cpu (approved):
Thanks!

PR #1567, opened 2023-11-03
replace HashSet with BTreeSet in msgs::handshake

japaric:
BTreeSet is in alloc but HashSet is not. all the affected sets contain integers

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -20,12 +20,12 @@ use crate::x509::wrap_in_sequence;
 
 use pki_types::CertificateDer;
 
+use alloc::collections::BTreeSet;
 #[cfg(feature = "logging")]
 use alloc::string::String;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt;
-use std::collections;
 
 /// Create a newtype wrapper around a given type.
 ///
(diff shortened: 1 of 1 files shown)


djc:
We should probably get some kind of perf benchmark for this?

japaric:
the icount bench job found -2% and -0.2% regressions: https://github.com/rustls/rustls/actions/runs/6746904928?pr=1567

should I try `IndexMap` + `fnv`? that should have the same `O(1)` perf as `HashMap`

djc:
I don't think we'd want to bring in an extra dependency for this. It seems weird that the two benchmarks hit ~2% vs all the other ones that are smaller -- I wonder if that's just noise or if there's a real difference there.

ctz:
This makes sense to me.

The performance improvements I think indicate a general change, but are masked by baseline handshake cost:

- the TLS1.2 server resumptions (-2%) takes the brunt (because it has no significant cryptography),
-  the improvement is drowned out in TLS1.3 server resumptions (-0.2%) by the extra cryptography happening there
- pulled below 0.2% threshold elsewhere by heaps of cryptography.

I am guessing the performance difference comes from avoiding initialising a `Hasher` and hashing some small integers?

Note that when we land #1475 there is much less reliance on this data structure -- the performance delta will be closed.

The other note is that we rely on adversarial worse-case complexity of whatever data structure we use here; so b-tree is slightly better than a plain hash table in that case, even if the usual-case is worse. That is worth keeping in mind when thinking about different data structures in the future.

ctz:
> the icount bench job found -2% and -0.2% regressions:

to clarify, these are -2% and -0.2% _improvements_

ctz (approved):
(obv. needs a rebase)

japaric:
rebased!

> to clarify, these are -2% and -0.2% improvements

ah, since the icount job failed I thought the minus meant regression :sweat_smile:

cpu (approved):

PR #1568, opened 2023-11-03
Encrypted client hello configuration messages and serialization

cpu:
This commit breaks out representation of Encrypted Client Hello (ECH) configuration from overall support for the feature. This code is relatively isolated and so can be added without much impact to the rest of the codebase. It does _not_ provide any actual ECH support but is a necessary prerequisite. 

The code is almost entirely derived from earlier WIP branches adding ECH support (notably, https://github.com/rustls/rustls/pull/663), but updated for the current Rustls codebase, and spot checked against the latest ECH draft at the time of writing ([draft-17](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17)). HPKE references are also updated to use the published RFC ([RFC 9180](https://datatracker.ietf.org/doc/html/rfc9180)).


rustls/src/msgs/handshake.rs
```diff
@@ -2252,3 +2253,125 @@ impl HandshakeMessagePayload {
         }
     }
 }
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct HpkeSymmetricCipherSuite {
+    pub hpke_kdf_id: HpkeKdf,
```
djc, line 2259: Can probably skip the `hpke_` prefixes on these fields?

rustls/src/msgs/handshake.rs
```diff
@@ -2252,3 +2253,125 @@ impl HandshakeMessagePayload {
...
+    pub hpke_aead_id: HpkeAead,
+}
+
+impl Default for HpkeSymmetricCipherSuite {
+    // TODO(XXX): revisit the default configuration. This is just what Cloudflare ships right now.
+    fn default() -> Self {
+        Self {
+            hpke_kdf_id: HpkeKdf::HKDF_SHA256,
```
djc, line 2267: Should we push this down into `Default` impls on `HpkeKdf` and `HpkeAead`?

rustls/src/msgs/handshake.rs
```diff
@@ -2252,3 +2253,125 @@ impl HandshakeMessagePayload {
...
     }
 }
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct HpkeSymmetricCipherSuite {
+    pub hpke_kdf_id: HpkeKdf,
+    pub hpke_aead_id: HpkeAead,
+}
+
+impl Default for HpkeSymmetricCipherSuite {
```
djc, line 2263: Probably sort this after the `Codec` impl?

rustls/src/msgs/handshake.rs
```diff
@@ -2252,3 +2253,125 @@ impl HandshakeMessagePayload {
...
+}
+
+impl TlsListElement for HpkeSymmetricCipherSuite {
+    const SIZE_LEN: ListLength = ListLength::U16;
+}
+
+#[derive(Clone, Debug)]
+pub struct HpkeKeyConfig {
+    pub config_id: u8,
+    pub hpke_kem_id: HpkeKem,
```
djc, line 2294: `hpke_` prefixes?

rustls/src/msgs/handshake.rs
```diff
@@ -2252,3 +2253,125 @@ impl HandshakeMessagePayload {
...
+            },
+            extensions: PayloadU16::read(r)?,
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct EchConfig {
+    pub version: EchVersion,
+    pub contents: EchConfigContents,
```
djc, line 2351: Unless the standards specifically define a separate `_Contents` container, maybe rewrite this to contain the contents directly?

rustls/tests/ech.rs
```diff
@@ -0,0 +1,100 @@
+mod ech_configs {
+    use base64::prelude::{Engine, BASE64_STANDARD};
+
+    use rustls::internal::msgs::codec::{Codec, Reader};
+    use rustls::internal::msgs::enums::{EchVersion, HpkeAead, HpkeKdf, HpkeKem};
+    use rustls::internal::msgs::handshake::{EchConfig, HpkeKeyConfig, HpkeSymmetricCipherSuite};
+    use rustls::server::danger::DnsName;
+
+    // One EchConfig, with server-name "localhost".
+    const BASE64_ECHCONFIG_LIST_LOCALHOST: &str =
```
djc, line 10: `const`s at the bottom?

rustls/tests/ech.rs
```diff
@@ -0,0 +1,100 @@
...
+        "AED+DQA8AAAgACAxoIJyV36iDlfFRmqE+ho2PxXE0EISPfUUJYKCy6T8VwAIAAEAAQABAAOACWxvY2FsaG9zdAAA";
+
+    // Two EchConfigs, both with server-name "cloudflare-esni.com".
+    const BASE64_ECHCONFIG_LIST_CF: &str =
+        "AK3+DQBCwwAgACAJ9T5U4FeM6631r2bvAuGtmEd8zQaoTkFAtArTcMl/XQAEAAEAASUTY2xvdWRmbGFyZS1lc25pLmNvbQAA/g0AYwMAEABBBGGbUlGLuGRorUeFwmrgHImkrh9uxoPrnFKpS5bQvnc5grfMS3PvymQ2FYL02WQi1ZzZJg5OsYYdzlaGYnEoJNsABAABAAEqE2Nsb3VkZmxhcmUtZXNuaS5jb20AAA==";
+
+    fn get_ech_config(s: &str) -> Vec<EchConfig> {
```
djc, line 17: Push the helper down?

djc (approved):

rustls/tests/ech.rs
```diff
@@ -0,0 +1,100 @@
+mod ech_configs {
```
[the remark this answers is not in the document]
cpu: I was thinking this file would have additional ECH related tests in the future and it might be nice to group the config related ones in their own module.

I don't feel strongly though, happy to flatten the namespace.

rustls/src/msgs/handshake.rs
cpu: The standard does define the separate container in https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#section-4 so I think it makes sense to keep it as-is.

rustls/tests/ech.rs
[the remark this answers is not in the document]
djc: I would flatten for now, can reintroduce later if we think it makes sense.

rustls/src/msgs/handshake.rs
cpu: Done.

ctz (approved):

imsys:
Just tagging for better tracking.
* #199
* #508

PR #1571, opened 2023-11-04
Reduce allocation counts in a couple of places

jsha:
Messing around with heaptrack in #1563, I identified a couple of places with relatively high allocation counts when running `simpleclient` (modified to do several connections in a row).

In CertificatePayloadTls13::convert, we were unnecessarily doing a clone. Fixed this by converting that function to take `self`, and the two callsites to use `require_handshake_msg_move!` instead of `require_handshake_msg!`.

In `ChunkVecBuffer::consume`, when a chunk is partially consumed, we were calling [Vec::split_off](https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.split_off), which always "Returns a newly allocated vector." Instead, use [Vec::drain](https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.drain), which reuses the current allocation.

In local testing, the change to `ChunkVecBuffer::consume` appeared to reduce allocations by almost 10%. I'll be curious to see if the CI run shows actual runtime performance.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1448,12 +1448,11 @@ impl CertificatePayloadTls13 {
             .unwrap_or_default()
     }
 
-    pub(crate) fn convert(&self) -> CertificatePayload {
-        let mut ret = Vec::new();
-        for entry in &self.entries {
-            ret.push(entry.cert.clone());
-        }
-        ret
+    pub(crate) fn convert(self) -> CertificatePayload {
+        self.entries
+            .into_iter()
+            .map(|e| e.cert)
+            .collect()
     }
 }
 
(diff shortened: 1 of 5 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -1448,10 +1448,10 @@ impl CertificatePayloadTls13 {
             .unwrap_or_default()
     }
 
-    pub(crate) fn convert(&self) -> CertificatePayload {
+    pub(crate) fn convert(self) -> CertificatePayload {
         let mut ret = Vec::new();
```
paolobarbolini, line 1452: Drive by review: this function could make an iterator out of `self.entries`, `map` `entry.cert` and `collect`?

djc (approved):
Nice!

cpu (approved):
Nice finds!

Issue #909, opened 2022-01-07
Clarify support for non-unicode paths in SSLKEYLOGFILE


briansmith:
I noted in #899 that it seems like we should consider whether we should use `std::env::var_os` instead of `std::env::var` when reading paths from the environment. That's a separate issue from whether we should use `std::env` at all, so splitting that out.

If we really don't want to support non-Unicode paths, then we should add some documentation to state that. But, my understanding is that `var_os` and `OsString` are the right things to use for paths. This would eliminate the need to handle the `NotUnicode` error. (Do we have a test for this?)

briansmith:
Here's the current logic:

```rust
        let var = env::var("SSLKEYLOGFILE");
        Self(Mutex::new(KeyLogFileInner::new(var)))
```

```rust
    fn new(var: Result<String, env::VarError>) -> Self {
        let path = match var {
            Ok(ref s) => Path::new(s),
            Err(env::VarError::NotUnicode(ref s)) => Path::new(s),
            Err(env::VarError::NotPresent) => {
                return Self {
                    file: None,
                    buf: Vec::new(),
                };
            }
        };
```

So, basically, if `SSLKEYLOGFILE` can be converted to UTF-8, then we use the UTF-8 version of the path. Otherwise, we treat it as an `OsStr` and use that as the path.

I'm not sure why we do this instead of just using `env::var_os` though.

djc:
If we're going to do a semver-incompatible release anyway, we should just do as you proposed and pass a `PathBuf` into `KeyLogFile` directly.

ctz:
Resolved in #1572

PR #1572, opened 2023-11-05
key_log_file: replace `env::var` with `env::var_os`

paolobarbolini:
Instead of matching the error from `env:var` in order to accept non-unicode file paths we switch to the more idiomatic `env::var_os`.


djc (approved):

cpu (approved):
Thanks! This looks good to me, and I believe resolves https://github.com/rustls/rustls/issues/909

PR #1574, opened 2023-11-06

djc:


djc:
A nice performance improvement there. @aochagavia how hard would it be to filter the detailed icount benchmark differences so that a version upgrade is shown in a nicer way (unlike the current `-long-path-to-ring-0.17.2 .. \n +long-path-to-ring-0.17.5`)?

cpu (approved):
Thanks!

aochagavia:
> @aochagavia how hard would it be to filter the detailed icount benchmark differences so that a version upgrade is shown in a nicer way (unlike the current `-long-path-to-ring-0.17.2 .. \n +long-path-to-ring-0.17.5`)?

A while ago @ctz [introduced](https://github.com/rustls/rustls/commit/237ba876bc5d56119772d703f9bc00a8f5b7a6ea) some regex magic to strip unwanted elements from paths a while ago, so we might be able to reuse that here. I'll add this to my list of things to look at after we are done adding timing benchmarks.

aochagavia:
Today I've been thinking about how to show icount differences of version upgrades in a nicer way. Cachegrind supports a `--mod-filename` flag to replace path elements using regular expressions. I actually found a bug in the bench runner diffs and created a [fix](https://github.com/rustls/rustls-bench-app/pull/31) that uses this feature. The same mechanism could be used to replace the paths of dependencies, but it will be ugly because you can only provide a single regular expression (I'm pretty sure it's _doable_ though). I won't get around to do it myself unfortunately.

PR #1577, opened 2023-11-08
Reorder & tidy imports with rustfmt

ctz:
We can choose carefully when to merge this. The final commit is 100% mechanical, but it will cause significant conflicts in other PRs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -347,6 +347,29 @@ jobs:
       - name: Check formatting (fuzz workspace)
         run: cargo fmt --all --manifest-path=fuzz/Cargo.toml -- --check
 
+  format-unstable:
+    name: Format (unstable)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install rust nightly toolchain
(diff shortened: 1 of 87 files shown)


ctz:
> 🟡 **Format** _Expected — Waiting for status to be reported_

We'd need to change the expected name of this job in the branch protection config.

ctz:
Still not an opportune time to land this, but I reworked it to avoid stable `cargo fmt` seeing unstable options. Instead there's a `cargo fmt-unstable` alias which references the (separate) configuration file with them in.

.rustfmt.unstable.toml
```diff
@@ -0,0 +1,3 @@
+chain_width=40
```
djc, line 1: Tiny nit: spaces around the `=` signs?

.github/workflows/build.yml
```diff
@@ -311,6 +311,29 @@ jobs:
...
 
+  format-unstable:
+    name: Format (unstable)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install rust nightly toolchain
```
djc, line 322: Maybe we should pin the nightly here?

djc (approved):

PR #1578, opened 2023-11-08
implement the caller-side managed buffers (AKA `LlConnection`) API

japaric:
this is a WIP implementation of RFC #1420 

currently only the `LlClientConnection` API has been minimally implemented but it's able to interact with, for example, `example.com` using either TLS 1.2 or TLS 1.3.

I'm putting this up now because I'd like feedback on the approach. I have aimed for maximal code reuse (though there a few places that could use a bit more code reuse) and my approach has been to reuse the existing `State::handle` machinery to implement the `LlConnection` API.

Arguably that's not the most performant option because, for example, the current implementation does memcpy and allocations when encoding / decoding TLS records (see `Codec`) so the `LlConnection` API ends up inheriting those.

However, it is, I believe, the fastest approach towards landing the public API. Further performance improvements can be done without breaking the public API (I believe the `#[non_exhaustive]` attribute on `LlState` should let us extend it without incurring in semver breaking changes and any required change in traits like `Codec` and `State` won't be public facing).

This PR is best reviewed on a commit-by-commit basis

TODO list
- [ ] add size checking to `encrypt` method (depends on PR #1579  )
- [x] complete `LlMessageDeframer::append_hs`
- [x] add and test `LlServerConnection`
- [x] add support for early data
- add support for caller-side certificate-verification (to be done in a separate PR)
- perform decryption in-place (to be done in a separate PR)

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -1,3 +1,5 @@
+use core::ops;
+
 use crate::enums::{AlertDescription, ContentType, HandshakeType, ProtocolVersion};
 use crate::error::{Error, InvalidMessage, PeerMisbehaved};
 #[cfg(feature = "logging")]
@@ -25,62 +27,35 @@ use pki_types::CertificateDer;
 
 /// Connection state common to both client and server connections.
 pub struct CommonState {
-    pub(crate) negotiated_version: Option<ProtocolVersion>,
-    pub(crate) side: Side,
-    pub(crate) record_layer: record_layer::RecordLayer,
-    pub(crate) suite: Option<SupportedCipherSuite>,
-    pub(crate) alpn_protocol: Option<Vec<u8>>,
-    pub(crate) aligned_handshake: bool,
-    pub(crate) may_send_application_data: bool,
-    pub(crate) may_receive_application_data: bool,
-    pub(crate) early_traffic: bool,
-    sent_fatal_alert: bool,
-    /// If the peer has signaled end of stream.
-    pub(crate) has_received_close_notify: bool,
-    pub(crate) has_seen_eof: bool,
-    pub(crate) received_middlebox_ccs: u8,
-    pub(crate) peer_certificates: Option<Vec<CertificateDer<'static>>>,
-    message_fragmenter: MessageFragmenter,
     pub(crate) received_plaintext: ChunkVecBuffer,
     sendable_plaintext: ChunkVecBuffer,
     pub(crate) sendable_tls: ChunkVecBuffer,
-    queued_key_update_message: Option<Vec<u8>>,
 
-    #[allow(dead_code)] // only read for QUIC
-    /// Protocol whose key schedule should be used. Unused for TLS < 1.3.
-    pub(crate) protocol: Protocol,
-    #[cfg(feature = "quic")]
-    pub(crate) quic: quic::Quic,
-    pub(crate) enable_secret_extraction: bool,
+    inner: CommonStateInner,
+}
+
+impl ops::DerefMut for CommonState {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
+
+impl ops::Deref for CommonState {
+    type Target = CommonStateInner;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
 }
 
 impl CommonState {
     pub(crate) fn new(side: Side) -> Self {
         Self {
-            negotiated_version: None,
-            side,
-            record_layer: record_layer::RecordLayer::new(),
-            suite: None,
-            alpn_protocol: None,
-            aligned_handshake: true,
-            may_send_application_data: false,
-            may_receive_application_data: false,
-            early_traffic: false,
-            sent_fatal_alert: false,
-            has_received_close_notify: false,
-            has_seen_eof: false,
-            received_middlebox_ccs: 0,
-            peer_certificates: None,
-            message_fragmenter: MessageFragmenter::default(),
             received_plaintext: ChunkVecBuffer::new(Some(DEFAULT_RECEIVED_PLAINTEXT_LIMIT)),
             sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
             sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
-            queued_key_update_message: None,
 
-            protocol: Protocol::Tcp,
-            #[cfg(feature = "quic")]
-            quic: quic::Quic::default(),
-            enable_secret_extraction: false,
+            inner: CommonStateInner::new(side),
         }
     }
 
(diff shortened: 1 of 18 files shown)


jsha:
For the convenience of reviewers I've generated docs from this branch and published them here:

https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/client/client_conn/struct.LlClientConnection.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/client/struct.LlClientConnection.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/enum.EncodeError.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/enum.EncryptError.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/enum.LlState.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/enum.EncodeError.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/enum.EncryptError.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/enum.LlState.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.AppDataAvailable.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.AppDataRecord.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.InsufficientSizeError.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.LlConnectionCommon.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.LlStatus.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.MayEncryptAppData.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.MustEncodeTlsData.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/ll/struct.MustTransmitTlsData.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.AppDataAvailable.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.AppDataRecord.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.InsufficientSizeError.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.LlConnectionCommon.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.LlStatus.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.MayEncryptAppData.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.MustEncodeTlsData.html
https://rustdoc.crud.net/jsha/rustls-llconnection/rustls/struct.MustTransmitTlsData.html

rustls/src/conn.rs
```diff
@@ -325,12 +325,12 @@ impl<Data> ConnectionCommon<Data> {
     pub fn reader(&mut self) -> Reader {
         let common = &mut self.core.common_state;
         Reader {
-            received_plaintext: &mut common.received_plaintext,
             // Are we done? i.e., have we processed all received messages, and received a
             // close_notify to indicate that no new messages will arrive?
             peer_cleanly_closed: common.has_received_close_notify
                 && !self.core.message_deframer.has_pending(),
             has_seen_eof: common.has_seen_eof,
+            received_plaintext: &mut common.received_plaintext,
```
djc, line 334: Would prefer not to have this unrelated change in this commit.

rustls/src/common_state.rs
```diff
@@ -25,62 +27,35 @@ use pki_types::CertificateDer;
...
-    #[allow(dead_code)] // only read for QUIC
-    /// Protocol whose key schedule should be used. Unused for TLS < 1.3.
-    pub(crate) protocol: Protocol,
-    #[cfg(feature = "quic")]
-    pub(crate) quic: quic::Quic,
-    pub(crate) enable_secret_extraction: bool,
+    inner: CommonStateInner,
+}
+
+impl ops::DerefMut for CommonState {
```
djc, line 326: Hmm, I don't love extending the deref tower with another layer. What do you think about moving in the other direction, wrapping up `received_plaintext`, `sendable_plaintext` and `sendable_tls` into a `CommonBuffers` type and sticking that in the current `Connection` types as a sibling of `common`?

rustls/src/lib.rs
```diff
@@ -359,6 +359,7 @@ mod builder;
 mod enums;
 mod key_log;
 mod key_log_file;
+mod ll;
```
djc, line 362: I think it is important that we come up with a better name for this stuff... Maybe something like `unbuffered`?

rustls/src/common_state.rs
```diff
@@ -25,6 +26,202 @@ use alloc::vec::Vec;
 
 use pki_types::CertificateDer;
 
+#[allow(dead_code)]
+pub(crate) enum CommonState2<'a> {
```
djc, line 30: `CommonState2` definitely needs a new name. Maybe it should not exist at all and we could pass around a `CommonState` and an `Option<CommonBuffers>`?

rustls/src/msgs/deframer.rs
```diff
@@ -379,6 +379,273 @@ impl MessageDeframer {
     }
 }
 
+/// A low-level version of `MessageDeframer` that does not maintains an internal buffer. instead, it
```
djc, line 382: I don't think we want to merge all this duplicate code.

rustls/src/common_state.rs
```diff
@@ -39,6 +39,55 @@ pub(crate) enum CommonState2<'a> {
 }
 
 impl CommonState2<'_> {
+    fn reborrow(&mut self) -> CommonState2 {
```
djc, line 42: This should go below `fn process_main_protocol()`.

rustls/src/common_state.rs
```diff
@@ -149,6 +200,42 @@ impl CommonState2<'_> {
         }
     }
 
+    fn send_warning_alert(&mut self, desc: AlertDescription) {
```
djc, line 208: Please don't move and change this in the same commit.

rustls/src/common_state.rs
```diff
@@ -621,13 +678,111 @@ pub(crate) struct LlCommonState {
     inner: CommonStateInner,
 }
 
+impl ops::DerefMut for LlCommonState {
```
djc, line 643: Let's consistently order `DerefMut` after `Deref`?

djc (commented):
Here's an initial round of feedback. Feels like this doesn't yet hit on the right abstractions.

rustls/src/common_state.rs
ctz: My interpretation of this is it's a incremental migration to a new replacement type? ie

- add a new type with a temporary name
- replace instances of an old type with the new one incrementally, in successive commits
- delete the old type now it is unused
- rename the new type to the old name

Once you're done, all that can get squashed together and the temporary name disappears.

ctz:
Is it still the plan to have the buffering `{Client,Server}Connection` be implemented atop the unbuffered one? The duality of the `CommonState2` type suggests to me that we'll actually have to add a new dimension to all tests to test this adequately.

japaric:
> Is it still the plan to have the buffering {Client,Server}Connection be implemented atop the unbuffered one? 

that's on my list of things to try once I have expanded the unbuffered API to support early data and have tested it doing the server side of things.

that'd be one way to go: have the current API be implemented as the unbuffered one plus some ChunkVecBuffers then the current tests would exercise the unbuffered API. the `CommonState2` enum would be unnecessary and `LlCommonState` would become the only common state, etc.

japaric:
added `LlServerConnection` API. while writing an example, based on the existing `tlsserver-mio` example, to test it out. I realized that there should be some API equivalent to `send_close_notify` so I added that to one of the type states of the API.

japaric:
I've pushed early-data support for both the client and the server and updated the examples to exercise that functionality.

---

> Is it still the plan to have the buffering {Client,Server}Connection be implemented atop the unbuffered one? 

I gave this some idle thought and I think the buffered version could call the unbuffered version's `encrypt` / `encode` methods with an *empty* slice to probe the buffer size it needs to allocate; on the retry it can pass a fresh `Vec` and push that in the `ChunkVecBuffer`. something like this

``` rust
pub struct BufferedClientConnection {
    sendable_plaintext: ChunkVecBuffer,
    sendable_tls: ChunkVecBuffer,
    unbuffered: UnbufferedClientConnection,
}

// actually this should be implemented for the `Writer` but to keep this short
impl io::Write for BufferedClientConnection {
    fn write(&mut self, plaintext: &[u8]) -> io::Result<usize> {
        // .. state machine stuff omitted ..

        if self.unbuffered.may_send_application_data {
            // probe phase
            let err = self.unbuffered.encrypt(plaintext, &mut []).unwrap_err();
            let mut buf =
                if let EncryptError::InsufficientSize(InsufficientSizeError { required_size }) =
                    &err
                {
                    vec![0; *required_size]
                } else {
                    return Err(map_err(err)); // legitimate error (e.g. encrypter exhausted)
                };

            // execute phase
            let written = self
                .unbuffered
                .encrypt(plaintext, &mut buf)
                .map_err(map_err)?;
            buf.truncate(written);

            self.sendable_tls.append(buf);
        } else {
            self.sendable_plaintext.append(plaintext.to_vec());
        }

        Ok(plaintext.len())
    }
// ...
```

I'll give that a closer look

japaric:
closing in favor of #1583

PR #1579, opened 2023-11-08
add `encrypted_payload_len` to `MessageEncrypter`

japaric:
The `LlConnection` API proposed in RFC #1420 returns an error when the user attempts to encrypt app-data that won't fit in the provided buffer. To perform this check is necessary to compute the size of the resulting TLS record(s) without performing any encryption or encoding as that would be expensive in the retry cases where the end-user will call the `encrypt` API twice. To compute the size, `MessageEncrypter` implementations need to provide the post-encryption size.

This has been send as a separate PR from #1578 because it's a semver breaking change and it would be good to get it into the 0.22 release.

I'm submitting @pvdrz 's implementation as it but would it make sense to have `MessageEncrypter` report the encryption overhead (e.g. `fn encryption_overhead(&self) -> usize`), which appears to be constant in all in-tree implementations, instead?

### DIFF
--- a/rustls/src/crypto/cipher.rs
+++ b/rustls/src/crypto/cipher.rs
@@ -132,6 +132,10 @@ pub trait MessageEncrypter: Send + Sync {
     /// Encrypt the given TLS message `msg`, using the sequence number
     /// `seq which can be used to derive a unique [`Nonce`].
     fn encrypt(&self, msg: BorrowedPlainMessage, seq: u64) -> Result<OpaqueMessage, Error>;
+
+    /// Return the length of the ciphertext that results from encrypting plaintext of
+    /// length `payload_len`
+    fn encrypted_payload_len(&self, payload_len: usize) -> usize;
 }
 
 impl dyn MessageEncrypter {
@@ -299,6 +303,10 @@ struct InvalidMessageEncrypter {}
 impl MessageEncrypter for InvalidMessageEncrypter {
     fn encrypt(&self, _m: BorrowedPlainMessage, _seq: u64) -> Result<OpaqueMessage, Error> {
         Err(Error::EncryptError)
+    }
+
+    fn encrypted_payload_len(&self, payload_len: usize) -> usize {
+        payload_len
     }
 }
 
--- a/rustls/src/crypto/ring/tls12.rs
+++ b/rustls/src/crypto/ring/tls12.rs
@@ -266,8 +266,8 @@ impl MessageEncrypter for GcmMessageEncrypter {
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).0);
         let aad = aead::Aad::from(make_tls12_aad(seq, msg.typ, msg.version, msg.payload.len()));
 
-        let total_len = msg.payload.len() + self.enc_key.algorithm().tag_len();
-        let mut payload = Vec::with_capacity(GCM_EXPLICIT_NONCE_LEN + total_len);
+        let total_len = self.encrypted_payload_len(msg.payload.len());
+        let mut payload = Vec::with_capacity(total_len);
         payload.extend_from_slice(&nonce.as_ref()[4..]);
         payload.extend_from_slice(msg.payload);
 
@@ -277,6 +277,10 @@ impl MessageEncrypter for GcmMessageEncrypter {
             .map_err(|_| Error::EncryptError)?;
 
         Ok(OpaqueMessage::new(msg.typ, msg.version, payload))
+    }
+
+    fn encrypted_payload_len(&self, payload_len: usize) -> usize {
+        payload_len + GCM_EXPLICIT_NONCE_LEN + self.enc_key.algorithm().tag_len()
     }
 }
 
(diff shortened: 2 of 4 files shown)


rustls/src/crypto/cipher.rs
```diff
@@ -132,6 +132,10 @@ pub trait MessageEncrypter: Send + Sync {
     /// Encrypt the given TLS message `msg`, using the sequence number
     /// `seq which can be used to derive a unique [`Nonce`].
     fn encrypt(&self, msg: BorrowedPlainMessage, seq: u64) -> Result<OpaqueMessage, Error>;
+
+    /// returns the length of the cryptext that results from encrypting plaintext of
```
djc, line 136: Maybe spend a bit more attention on this docstring? (Proper capitalization, ideally a first line that is more concise, and probably don't use "cryptext"?)

djc (commented):
Makes sense to have something for this!

I agree respinning this as something more like `fn encryption_overhead() -> usize` is preferable if it can be made to work.

ctz:
Hm. It's not necessarily the case that the ciphertext length is a pure function on the plaintext length. It is definitely not a constant.

- not a pure function: someone could decide to apply a random amount of padding (in TLS1.3)
- not a constant: padding with length related to plaintext length

It would be good if the value sampled each time from this function was passed back into the encryption function itself. I'd expect this happens eventually anyway, because presumably we'll change the APIs to take an mut output slice?

For the same reason, it is generally unknowable what the plaintext size for a given cipher text prior to decrypting it -- it definitely is not `ciphertext.len() - encryption_overhead()`.

japaric:
> not a pure function: someone could decide to apply a random amount of padding (in TLS1.3)

but can that be upper bounded? the `LlConnection::encrypt(plaintext, out_buffer)` implementation needs to know **a** buffer size where the encrypted contents will fit; it doesn't have to be the *exact* size

> It would be good if the value sampled each time from this function was passed back into the encryption function itself.

not sure I follow. do you mean that the user of `MessageEncrypter` should be using the API like this?

``` rust
let plaintext;
let encrypted_len = encrypter.encrypted_payload_len(plaintext.len());
let cryptext = encrypter.encrypt(plaintext, encrypted_len);
```

>  I'd expect this happens eventually anyway, because presumably we'll change the APIs to take an mut output slice?

the way `LlConnection::encrypt(plaintext, out_buffer)` works is that it'll encrypt `plaintext` IFF it fits in `out_buffer`. it returns an error with the required buffer size if the buffer is too small; `out_buffer` is not modified in this error case.

if we change `MessageEncrypter` to operate on slices it'd expect to behave the same: either it encrypts the whole thing or does nothing so it should *not* need to be stateful to support incrementally encrypting a record.

ctz:
> but can that be upper bounded? the `LlConnection::encrypt(plaintext, out_buffer)` implementation needs to know **a** buffer size where the encrypted contents will fit; it doesn't have to be the _exact_ size

Upper bound is 2**14 + 1 bytes.

> not sure I follow. do you mean that the user of MessageEncrypter should be using the API like this?

I was thinking more like:

```
let plaintext = ...;
let encrypted_len = encrypter.encrypted_payload_len(plaintext.len());
// (validate that the following slice is non-panicing)
encrypter.encrypt_into(plaintext, &mut output_buffer[..encrypted_len]);
```

So, yes, the same as what you're describing I think?

rustls/src/crypto/cipher.rs
ctz: Yes I'd prefer it if we stick to "ciphertext", as elsewhere in the codebase and common in the TLS RFCs. (Styled as shown: one word, not hyphenated.)

ctz (changes requested):
one docs nit, but otherwise i'm happy with this

japaric:
> I was thinking more like:

ah, I now see what you meant. yeah, using it like that makes sense.

I have tweaked the doc comment and squashed the commits

rustls/src/crypto/ring/tls12.rs
```diff
@@ -266,7 +266,7 @@ impl MessageEncrypter for GcmMessageEncrypter {
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).0);
         let aad = aead::Aad::from(make_tls12_aad(seq, msg.typ, msg.version, msg.payload.len()));
 
-        let total_len = msg.payload.len() + self.enc_key.algorithm().tag_len();
+        let total_len = self.encrypted_payload_len(msg.payload.len());
         let mut payload = Vec::with_capacity(GCM_EXPLICIT_NONCE_LEN + total_len);
```
cpu, line 270: Shouldn't this be changed to `let mut payload = Vec::with_capacity(total_len)`? The `GCM_EXPLICIT_NONCE_LEN` is now accounted for in `encrypted_payload_len` right? I think this version adds the explicit nonce len twice (or I'm missing something)

rustls/src/crypto/ring/tls12.rs
```diff
@@ -278,6 +278,10 @@ impl MessageEncrypter for GcmMessageEncrypter {
 
         Ok(OpaqueMessage::new(msg.typ, msg.version, payload))
     }
+
+    fn encrypted_payload_len(&self, payload_len: usize) -> usize {
+        GCM_EXPLICIT_NONCE_LEN + payload_len + self.enc_key.algorithm().tag_len()
```
cpu, line 283: very small nit: the other `encrypted_payload_len` impls are written with `payload_len` as the first term in the addition. It might be nice to match that here for consistency's sake.

rustls/src/crypto/ring/tls12.rs
japaric: good catch

japaric:
@cpu I've addressed your comments

cpu (approved):
Thanks, LGTM

djc (approved):

ctz (approved):

cpu:
> github-merge-queue bot removed this pull request from the merge queue due to failed status checks

I think `crates.io` is having a bad time. I'm seeing CI fail across multiple repos :sob:

PR #1582, opened 2023-11-10
Fix build: new feature needed for `std::io::BorrowedBuf` et al.

ctz:
Tracking issue: https://github.com/rust-lang/rust/issues/117693


cpu (approved):
Thanks for fixing this :+1:

Berrysoft:
Would you consider a new alpha release? We need this feature.

cpu:
> Would you consider a new alpha release? We need this feature.

@Berrysoft https://crates.io/crates/rustls/0.22.0-alpha.4 was published by djc today and it includes this fix.

PR #1583, opened 2023-11-10
[2/3] implement the caller-side managed buffers API (take 2)

japaric:
supersedes #1578 ~~but I'm not closing that one just yet because this one is still WIP~~

I took a different approach and tried to first move the buffers out from `CommonState` into `ConnectionCommon` to implement `ConnectionCommon` on top of `UnbufferedConnectionCommon` ; that turned out to be a non-trivial exercise. Then I tried the same "unbuffering" exercise with `MessageDeframer` and that turned out to be relatively straightforward. That's when I realized that as a more incremental step we can implement `UnbufferedConnectionCommon` on top of the current, buffered `CommonState` while still achieving the goal of not depending on IO traits or anything else from the `std` crate. So that's what this PR does (I haven't finished moving the server and early data stuff over from #1578 )

I did some renaming while copying over bits from #1578

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -87,6 +87,12 @@ jobs:
       - name: Check unbuffered client 
         run: cargo run --locked --bin unbuffered-client
 
+      - name: Check unbuffered tokio client 
+        run: cargo run --locked --bin unbuffered-async-client
+
+      - name: Check unbuffered async-std client 
+        run: cargo run --locked --bin unbuffered-async-client --features=async-std
+
       # Test the server_acceptor binary builds - we invoke with --help since it
       # will run a server process that doesn't exit when invoked with no args
       - name: Check server acceptor
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -84,6 +84,9 @@ jobs:
       - name: Check simple 0rtt client
         run: cargo run --locked --bin simple_0rtt_client
 
+      - name: Check unbuffered client 
+        run: cargo run --locked --bin unbuffered-client
+
       # Test the server_acceptor binary builds - we invoke with --help since it
       # will run a server process that doesn't exit when invoked with no args
       - name: Check server acceptor
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -7,6 +7,7 @@ description = "Rustls example code and tests."
 publish = false
 
 [dependencies]
+async-std = { version = "1.12.0", features = ["attributes"], optional = true }
 docopt = "~1.1"
 env_logger = "0.10"
 log = { version = "0.4.4" }
@@ -17,4 +18,5 @@ rustls = { path = "../rustls", features = [ "logging" ]}
 rustls-pemfile = "2"
 serde = "1.0"
 serde_derive = "1.0"
+tokio = { version = "1.34.0", features = ["io-util", "macros", "net", "rt"]}
 webpki-roots = "0.26"
--- a/examples/src/bin/unbuffered-async-client.rs
+++ b/examples/src/bin/unbuffered-async-client.rs
@@ -66,9 +66,9 @@ async fn converse(
     let mut iter_count = 0;
     while open_connection {
         let UnbufferedStatus { mut discard, state } =
-            conn.process_tls_records(&mut incoming_tls[..incoming_used])?;
+            conn.process_tls_records(&mut incoming_tls[..incoming_used]);
 
-        match dbg!(state) {
+        match dbg!(state.unwrap()) {
             ConnectionState::ReadTraffic(mut state) => {
                 while let Some(res) = state.next_record() {
                     let AppDataRecord {
--- a/examples/src/bin/unbuffered-async-client.rs
+++ b/examples/src/bin/unbuffered-async-client.rs
@@ -1,0 +1,263 @@
+use std::error::Error;
+use std::sync::Arc;
+
+#[cfg(feature = "async-std")]
+use async_std::io::{ReadExt, WriteExt};
+#[cfg(feature = "async-std")]
+use async_std::net::TcpStream;
+use rustls::client::{ClientConnectionData, UnbufferedClientConnection};
+use rustls::unbuffered::{
+    AppDataRecord, ConnectionState, EncodeError, EncryptError, InsufficientSizeError,
+    UnbufferedStatus, WriteTraffic,
+};
+#[allow(unused_imports)]
+use rustls::version::{TLS12, TLS13};
+use rustls::{ClientConfig, RootCertStore};
+#[cfg(not(feature = "async-std"))]
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+#[cfg(not(feature = "async-std"))]
+use tokio::net::TcpStream;
+
+const SERVER_NAME: &str = "example.com";
+const PORT: u16 = 443;
+
+const KB: usize = 1024;
+const INCOMING_TLS_BUFSIZE: usize = 16 * KB;
+const OUTGOING_TLS_INITIAL_BUFSIZE: usize = KB;
+
+const MAX_ITERATIONS: usize = 20;
+
+#[cfg_attr(not(feature = "async-std"), tokio::main(flavor = "current_thread"))]
+#[cfg_attr(feature = "async-std", async_std::main)]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let root_store = RootCertStore {
+        roots: webpki_roots::TLS_SERVER_ROOTS.into(),
+    };
+
+    let config = ClientConfig::builder_with_protocol_versions(&[&TLS13])
+        .with_root_certificates(root_store)
+        .with_no_client_auth();
+
+    let config = Arc::new(config);
+
+    let mut incoming_tls = [0; INCOMING_TLS_BUFSIZE];
+    let mut outgoing_tls = vec![0; OUTGOING_TLS_INITIAL_BUFSIZE];
+
+    converse(&config, &mut incoming_tls, &mut outgoing_tls).await?;
+
+    Ok(())
+}
+
+async fn converse(
+    config: &Arc<ClientConfig>,
+    incoming_tls: &mut [u8],
+    outgoing_tls: &mut Vec<u8>,
+) -> Result<(), Box<dyn Error>> {
+    let mut conn = UnbufferedClientConnection::new(Arc::clone(config), SERVER_NAME.try_into()?)?;
+    let mut sock = TcpStream::connect(format!("{SERVER_NAME}:{PORT}")).await?;
+
+    let mut incoming_used = 0;
+    let mut outgoing_used = 0;
+
+    let mut open_connection = true;
+    let mut sent_request = false;
+    let mut received_response = false;
+
+    let mut iter_count = 0;
+    while open_connection {
+        let UnbufferedStatus { mut discard, state } =
+            conn.process_tls_records(&mut incoming_tls[..incoming_used])?;
+
+        match dbg!(state) {
+            ConnectionState::ReadTraffic(mut state) => {
+                while let Some(res) = state.next_record() {
+                    let AppDataRecord {
+                        discard: new_discard,
+                        payload,
+                    } = res?;
+                    discard += new_discard;
+
+                    if payload.starts_with(b"HTTP") {
+                        let response = core::str::from_utf8(payload)?;
+                        let header = response
+                            .lines()
+                            .next()
+                            .unwrap_or(response);
+
+                        println!("{header}");
+                    } else {
+                        println!("(.. continued HTTP response ..)");
+                    }
+
+                    received_response = true;
+                }
+            }
+
+            ConnectionState::EncodeTlsData(mut state) => {
+                try_or_resize_and_retry(
+                    |out_buffer| state.encode(out_buffer),
+                    |e| {
+                        if let EncodeError::InsufficientSize(is) = &e {
+                            Ok(*is)
+                        } else {
+                            Err(e.into())
+                        }
+                    },
+                    outgoing_tls,
+                    &mut outgoing_used,
+                )?;
+            }
+
+            ConnectionState::TransmitTlsData(mut state) => {
+                if let Some(mut may_encrypt) = state.may_encrypt_app_data() {
+                    encrypt_http_request(
+                        &mut sent_request,
+                        &mut may_encrypt,
+                        outgoing_tls,
+                        &mut outgoing_used,
+                    );
+                }
+
+                send_tls(&mut sock, outgoing_tls, &mut outgoing_used).await?;
+                state.done();
+            }
+
+            ConnectionState::BlockedHandshake { .. } => {
+                recv_tls(&mut sock, incoming_tls, &mut incoming_used).await?;
+            }
+
+            ConnectionState::WriteTraffic(mut may_encrypt) => {
+                if encrypt_http_request(
+                    &mut sent_request,
+                    &mut may_encrypt,
+                    outgoing_tls,
+                    &mut outgoing_used,
+                ) {
+                    send_tls(&mut sock, outgoing_tls, &mut outgoing_used).await?;
+                    recv_tls(&mut sock, incoming_tls, &mut incoming_used).await?;
+                } else if !received_response {
+                    // this happens in the TLS 1.3 case. the app-data was sent in the preceding
+                    // `TransmitTlsData` state. the server should have already written a
+                    // response which we can read out from the socket
+                    recv_tls(&mut sock, incoming_tls, &mut incoming_used).await?;
+                } else {
+                    try_or_resize_and_retry(
+                        |out_buffer| may_encrypt.queue_close_notify(out_buffer),
+                        |e| {
+                            if let EncryptError::InsufficientSize(is) = &e {
+                                Ok(*is)
+                            } else {
+                                Err(e.into())
+                            }
+                        },
+                        outgoing_tls,
+                        &mut outgoing_used,
+                    )?;
+                    send_tls(&mut sock, outgoing_tls, &mut outgoing_used).await?;
+                    open_connection = false;
+                }
+            }
+
+            ConnectionState::Closed => {
+                open_connection = false;
+            }
+
+            // other states are not expected in this example
+            _ => unreachable!(),
+        }
+
+        if discard != 0 {
+            assert!(discard <= incoming_used);
+
+            incoming_tls.copy_within(discard..incoming_used, 0);
+            incoming_used -= discard;
+
+            eprintln!("discarded {discard}B from `incoming_tls`");
+        }
+
+        iter_count += 1;
+        assert!(
+            iter_count < MAX_ITERATIONS,
+            "did not get a HTTP response within {MAX_ITERATIONS} iterations"
+        );
+    }
+
+    assert!(sent_request);
+    assert!(received_response);
+    assert_eq!(0, incoming_used);
+    assert_eq!(0, outgoing_used);
+
+    Ok(())
+}
+
+fn try_or_resize_and_retry<E>(
+    mut f: impl FnMut(&mut [u8]) -> Result<usize, E>,
+    map_err: impl FnOnce(E) -> Result<InsufficientSizeError, Box<dyn Error>>,
+    outgoing_tls: &mut Vec<u8>,
+    outgoing_used: &mut usize,
+) -> Result<usize, Box<dyn Error>>
+where
+    E: Error + 'static,
+{
+    let written = match f(&mut outgoing_tls[*outgoing_used..]) {
+        Ok(written) => written,
+
+        Err(e) => {
+            let InsufficientSizeError { required_size } = map_err(e)?;
+            let new_len = *outgoing_used + required_size;
+            outgoing_tls.resize(new_len, 0);
+            eprintln!("resized `outgoing_tls` buffer to {new_len}B");
+
+            f(&mut outgoing_tls[*outgoing_used..])?
+        }
+    };
+
+    *outgoing_used += written;
+
+    Ok(written)
+}
+
+async fn recv_tls(
+    sock: &mut TcpStream,
+    incoming_tls: &mut [u8],
+    incoming_used: &mut usize,
+) -> Result<(), Box<dyn Error>> {
+    let read = sock
+        .read(&mut incoming_tls[*incoming_used..])
+        .await?;
+    eprintln!("received {read}B of data");
+    *incoming_used += read;
+    Ok(())
+}
+
+async fn send_tls(
+    sock: &mut TcpStream,
+    outgoing_tls: &[u8],
+    outgoing_used: &mut usize,
+) -> Result<(), Box<dyn Error>> {
+    sock.write_all(&outgoing_tls[..*outgoing_used])
+        .await?;
+    eprintln!("sent {outgoing_used}B of data");
+    *outgoing_used = 0;
+    Ok(())
+}
+
+fn encrypt_http_request(
+    sent_request: &mut bool,
+    may_encrypt: &mut WriteTraffic<'_, ClientConnectionData>,
+    outgoing_tls: &mut [u8],
+    outgoing_used: &mut usize,
+) -> bool {
+    if !*sent_request {
+        let request = format!("GET / HTTP/1.1\r\nHost: {SERVER_NAME}\r\nConnection: close\r\nAccept-Encoding: identity\r\n\r\n").into_bytes();
+        let written = may_encrypt
+            .encrypt(&request, &mut outgoing_tls[*outgoing_used..])
+            .expect("encrypted request does not fit in `outgoing_tls`");
+        *outgoing_used += written;
+        *sent_request = true;
+        eprintln!("queued HTTP request");
+        true
+    } else {
+        false
+    }
+}
--- a/examples/src/bin/unbuffered-client.rs
+++ b/examples/src/bin/unbuffered-client.rs
@@ -1,0 +1,255 @@
+//! This is a simple client using rustls' unbuffered API. Meaning that the application layer must
+//! handle the buffers required to receive, process and send TLS data.
+
+use std::error::Error;
+use std::io::{Read, Write};
+use std::net::TcpStream;
+use std::sync::Arc;
+
+use rustls::client::{ClientConnectionData, UnbufferedClientConnection};
+use rustls::unbuffered::{
+    AppDataRecord, ConnectionState, EncodeError, EncryptError, InsufficientSizeError,
+    UnbufferedStatus, WriteTraffic,
+};
+#[allow(unused_imports)]
+use rustls::version::{TLS12, TLS13};
+use rustls::{ClientConfig, RootCertStore};
+
+const SERVER_NAME: &str = "example.com";
+const PORT: u16 = 443;
+
+const KB: usize = 1024;
+const INCOMING_TLS_BUFSIZE: usize = 16 * KB;
+const OUTGOING_TLS_INITIAL_BUFSIZE: usize = KB;
+
+const MAX_ITERATIONS: usize = 20;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let root_store = RootCertStore {
+        roots: webpki_roots::TLS_SERVER_ROOTS.into(),
+    };
+
+    let config = ClientConfig::builder_with_protocol_versions(&[&TLS13])
+        .with_root_certificates(root_store)
+        .with_no_client_auth();
+
+    let config = Arc::new(config);
+
+    let mut incoming_tls = [0; INCOMING_TLS_BUFSIZE];
+    let mut outgoing_tls = vec![0; OUTGOING_TLS_INITIAL_BUFSIZE];
+
+    converse(&config, &mut incoming_tls, &mut outgoing_tls)?;
+
+    Ok(())
+}
+
+fn converse(
+    config: &Arc<ClientConfig>,
+    incoming_tls: &mut [u8],
+    outgoing_tls: &mut Vec<u8>,
+) -> Result<(), Box<dyn Error>> {
+    let mut conn = UnbufferedClientConnection::new(Arc::clone(config), SERVER_NAME.try_into()?)?;
+    let mut sock = TcpStream::connect(format!("{SERVER_NAME}:{PORT}"))?;
+
+    let mut incoming_used = 0;
+    let mut outgoing_used = 0;
+
+    let mut open_connection = true;
+    let mut sent_request = false;
+    let mut received_response = false;
+
+    let mut iter_count = 0;
+    while open_connection {
+        let UnbufferedStatus { mut discard, state } =
+            conn.process_tls_records(&mut incoming_tls[..incoming_used])?;
+
+        match dbg!(state) {
+            ConnectionState::ReadTraffic(mut state) => {
+                while let Some(res) = state.next_record() {
+                    let AppDataRecord {
+                        discard: new_discard,
+                        payload,
+                    } = res?;
+                    discard += new_discard;
+
+                    if payload.starts_with(b"HTTP") {
+                        let response = core::str::from_utf8(payload)?;
+                        let header = response
+                            .lines()
+                            .next()
+                            .unwrap_or(response);
+
+                        println!("{header}");
+                    } else {
+                        println!("(.. continued HTTP response ..)");
+                    }
+
+                    received_response = true;
+                }
+            }
+
+            ConnectionState::EncodeTlsData(mut state) => {
+                try_or_resize_and_retry(
+                    |out_buffer| state.encode(out_buffer),
+                    |e| {
+                        if let EncodeError::InsufficientSize(is) = &e {
+                            Ok(*is)
+                        } else {
+                            Err(e.into())
+                        }
+                    },
+                    outgoing_tls,
+                    &mut outgoing_used,
+                )?;
+            }
+
+            ConnectionState::TransmitTlsData(mut state) => {
+                if let Some(mut may_encrypt) = state.may_encrypt_app_data() {
+                    encrypt_http_request(
+                        &mut sent_request,
+                        &mut may_encrypt,
+                        outgoing_tls,
+                        &mut outgoing_used,
+                    );
+                }
+
+                send_tls(&mut sock, outgoing_tls, &mut outgoing_used)?;
+                state.done();
+            }
+
+            ConnectionState::BlockedHandshake { .. } => {
+                recv_tls(&mut sock, incoming_tls, &mut incoming_used)?;
+            }
+
+            ConnectionState::WriteTraffic(mut may_encrypt) => {
+                if encrypt_http_request(
+                    &mut sent_request,
+                    &mut may_encrypt,
+                    outgoing_tls,
+                    &mut outgoing_used,
+                ) {
+                    send_tls(&mut sock, outgoing_tls, &mut outgoing_used)?;
+                    recv_tls(&mut sock, incoming_tls, &mut incoming_used)?;
+                } else if !received_response {
+                    // this happens in the TLS 1.3 case. the app-data was sent in the preceding
+                    // `TransmitTlsData` state. the server should have already written a
+                    // response which we can read out from the socket
+                    recv_tls(&mut sock, incoming_tls, &mut incoming_used)?;
+                } else {
+                    try_or_resize_and_retry(
+                        |out_buffer| may_encrypt.queue_close_notify(out_buffer),
+                        |e| {
+                            if let EncryptError::InsufficientSize(is) = &e {
+                                Ok(*is)
+                            } else {
+                                Err(e.into())
+                            }
+                        },
+                        outgoing_tls,
+                        &mut outgoing_used,
+                    )?;
+                    send_tls(&mut sock, outgoing_tls, &mut outgoing_used)?;
+                    open_connection = false;
+                }
+            }
+
+            ConnectionState::Closed => {
+                open_connection = false;
+            }
+
+            // other states are not expected in this example
+            _ => unreachable!(),
+        }
+
+        if discard != 0 {
+            assert!(discard <= incoming_used);
+
+            incoming_tls.copy_within(discard..incoming_used, 0);
+            incoming_used -= discard;
+
+            eprintln!("discarded {discard}B from `incoming_tls`");
+        }
+
+        iter_count += 1;
+        assert!(
+            iter_count < MAX_ITERATIONS,
+            "did not get a HTTP response within {MAX_ITERATIONS} iterations"
+        );
+    }
+
+    assert!(sent_request);
+    assert!(received_response);
+    assert_eq!(0, incoming_used);
+    assert_eq!(0, outgoing_used);
+
+    Ok(())
+}
+
+fn try_or_resize_and_retry<E>(
+    mut f: impl FnMut(&mut [u8]) -> Result<usize, E>,
+    map_err: impl FnOnce(E) -> Result<InsufficientSizeError, Box<dyn Error>>,
+    outgoing_tls: &mut Vec<u8>,
+    outgoing_used: &mut usize,
+) -> Result<usize, Box<dyn Error>>
+where
+    E: Error + 'static,
+{
+    let written = match f(&mut outgoing_tls[*outgoing_used..]) {
+        Ok(written) => written,
+
+        Err(e) => {
+            let InsufficientSizeError { required_size } = map_err(e)?;
+            let new_len = *outgoing_used + required_size;
+            outgoing_tls.resize(new_len, 0);
+            eprintln!("resized `outgoing_tls` buffer to {new_len}B");
+
+            f(&mut outgoing_tls[*outgoing_used..])?
+        }
+    };
+
+    *outgoing_used += written;
+
+    Ok(written)
+}
+
+fn recv_tls(
+    sock: &mut TcpStream,
+    incoming_tls: &mut [u8],
+    incoming_used: &mut usize,
+) -> Result<(), Box<dyn Error>> {
+    let read = sock.read(&mut incoming_tls[*incoming_used..])?;
+    eprintln!("received {read}B of data");
+    *incoming_used += read;
+    Ok(())
+}
+
+fn send_tls(
+    sock: &mut TcpStream,
+    outgoing_tls: &[u8],
+    outgoing_used: &mut usize,
+) -> Result<(), Box<dyn Error>> {
+    sock.write_all(&outgoing_tls[..*outgoing_used])?;
+    eprintln!("sent {outgoing_used}B of data");
+    *outgoing_used = 0;
+    Ok(())
+}
+
+fn encrypt_http_request(
+    sent_request: &mut bool,
+    may_encrypt: &mut WriteTraffic<'_, ClientConnectionData>,
+    outgoing_tls: &mut [u8],
+    outgoing_used: &mut usize,
+) -> bool {
+    if !*sent_request {
+        let request = format!("GET / HTTP/1.1\r\nHost: {SERVER_NAME}\r\nConnection: close\r\nAccept-Encoding: identity\r\n\r\n").into_bytes();
+        let written = may_encrypt
+            .encrypt(&request, &mut outgoing_tls[*outgoing_used..])
+            .expect("encrypted request does not fit in `outgoing_tls`");
+        *outgoing_used += written;
+        *sent_request = true;
+        eprintln!("queued HTTP request");
+        true
+    } else {
+        false
+    }
+}
--- a/examples/src/bin/unbuffered-client.rs
+++ b/examples/src/bin/unbuffered-client.rs
@@ -70,9 +70,9 @@ fn converse(
     let mut iter_count = 0;
     while open_connection {
         let UnbufferedStatus { mut discard, state } =
-            conn.process_tls_records(&mut incoming_tls[..incoming_used])?;
+            conn.process_tls_records(&mut incoming_tls[..incoming_used]);
 
-        match dbg!(state) {
+        match dbg!(state.unwrap()) {
             ConnectionState::ReadTraffic(mut state) => {
                 while let Some(res) = state.next_record() {
                     let AppDataRecord {
--- a/examples/src/bin/unbuffered-client.rs
+++ b/examples/src/bin/unbuffered-client.rs
@@ -6,7 +6,7 @@ use std::io::{Read, Write};
 use std::net::TcpStream;
 use std::sync::Arc;
 
-use rustls::client::{ClientConnectionData, UnbufferedClientConnection};
+use rustls::client::{ClientConnectionData, EarlyDataError, UnbufferedClientConnection};
 use rustls::unbuffered::{
     AppDataRecord, ConnectionState, EncodeError, EncryptError, InsufficientSizeError,
     UnbufferedStatus, WriteTraffic,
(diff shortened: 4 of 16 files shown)


japaric:
I think the only thing missing here is implementing buffer size checking but that depends on #1579 . I'll implement that locally and push it when #1579 gets merged.

I'll resume looking into implementing `*Connection` on top of `Unbuffered*Connection`. One thing I noticed there is that it may make sense to let `UnbufferedConnection` deref to `CommonState` but some of its methods related to internal buffers (like `wants_read`, `wants_write`, `set_buffer_limit`) should move from `CommonState` to `ConnectionCommon`. Either that, or we add a subset of `CommonState` methods to `UnbufferedConnection` (internally they would delegate to `CommonState`); that would let us provide better API docs on methods like `is_handshaking` (which talks about `Connection::process_new_packets` and it's not applicable to `UnbufferedConnection`).

jsha:
Like with the previous revision of this PR, I generated some docs for ease of review of the public API. Hope this is helpful!

<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/client/client_conn/enum.EarlyDataError.html">enum EarlyDataError</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/client/client_conn/struct.UnbufferedClientConnection.html">struct UnbufferedClientConnection</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/client/enum.EarlyDataError.html">enum EarlyDataError</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/client/struct.UnbufferedClientConnection.html">struct UnbufferedClientConnection</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/conn/struct.UnbufferedConnectionCommon.html">struct UnbufferedConnectionCommon</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/enum.ConnectionState.html">enum ConnectionState</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/enum.EncodeError.html">enum EncodeError</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/enum.EncryptError.html">enum EncryptError</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/internal/msgs/deframer/struct.MessageDeframerVecBuffer.html">struct MessageDeframerVecBuffer</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/msgs/deframer/struct.MessageDeframerVecBuffer.html">struct MessageDeframerVecBuffer</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/server/server_conn/struct.UnbufferedServerConnection.html">struct UnbufferedServerConnection</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/server/struct.UnbufferedServerConnection.html">struct UnbufferedServerConnection</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.AppDataAvailable.html">struct AppDataAvailable</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.AppDataRecord.html">struct AppDataRecord</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.EarlyDataAvailable.html">struct EarlyDataAvailable</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.InsufficientSizeError.html">struct InsufficientSizeError</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.MayEncryptAppData.html">struct MayEncryptAppData</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.MustEncodeTlsData.html">struct MustEncodeTlsData</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.MustTransmitTlsData.html">struct MustTransmitTlsData</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.UnbufferedConnectionCommon.html">struct UnbufferedConnectionCommon</a>
<a href="https://rustdoc.crud.net/jsha/rustls-rfc1420-take2/rustls/struct.UnbufferedStatus.html">struct UnbufferedStatus</a>

rustls/src/msgs/deframer.rs
```diff
@@ -244,24 +243,32 @@ impl MessageDeframer {
         version: ProtocolVersion,
         payload: &[u8],
         end: usize,
-        quic: bool,
+        quic_push: bool,
```
djc, line 247: What's the rationale here? IMO this name isn't much more explanatory than the old one.

rustls/src/msgs/deframer.rs
```diff
@@ -271,8 +278,16 @@ impl MessageDeframer {
                 // Write it into the buffer and create the metadata.
 
                 let expected_len = payload_size(payload)?;
-                let dst = &mut self.buf[..payload.len()];
+                let buf = if quic_push {
```
djc, line 283: Feels like these changes should be distinct from the larger change of introducing the `Buffer` type -- it's not obvious why these are necessary. Also, for these little single-expression `if`/`else` blocks I think `match` expressions are more readable.

rustls/src/msgs/deframer.rs
```diff
@@ -299,31 +314,54 @@ impl MessageDeframer {
...
+    /// Returns true if we have messages for the caller
+    /// to process, either whole messages in our output
+    /// queue or partial messages in our buffer.
+    pub fn has_pending(&self) -> bool {
+        !self.buffer.is_empty()
+    }
+}
+
+#[derive(Default)]
+struct MessageDeframerBuffer {
```
djc: This feels a little verbose. Just `DeframerBuffer` should do?

djc (commented):
Just looked at the first commit for now.

rustls/src/msgs/deframer.rs
japaric: that boolean is only set to `true` when this method is called from `MessageDeframer::push` which sits behind the `quic` Cargo feature. the boolean is not necessarily set to `true` when the used protocol is `Quic` since `pop` will call this method with the boolean set to `false`. `quic_push` seemed a bit clearer than just `quic` which could be interpreted as "is the underlying protocol QUIC?"

rustls/src/conn/unbuffered.rs
```diff
@@ -18,6 +20,10 @@ impl<Data> UnbufferedConnectionCommon<Data> {
         let mut buffer = MessageDeframerSliceBuffer::new(incoming_tls, &mut discard);
 
         let state = 'outer: loop {
+            if let Some(chunk) = self.pop_early_data() {
```
ctz, line 23: Is the dependency on RTTI introduced in `pop_early_data` better handled by having a common (to clients and servers) `process_tls_records()` that is called by a client- and server-specific version? This clause could then exist only in the server version.

ctz (commented):
Up to and including "early data support".

rustls/src/msgs/deframer.rs
japaric: they were required to make QUIC tests pass. AFAICT, `append_hs(false)` merges records within the "filled" contents of the buffer (`self.buf[..self.used]`), whereas `append_hs(true)` (used by the QUIC-only `push` method) appends data to the buffer ("pushes" the data into `self.buf[self.used..]` -- that's why you also need to call `advance` in that case; in the original version that responsibility (of updating `self.user`) was on the caller. using either `unfilled` or `filled_mut` without the conditional here made either QUIC or TCP tests fail. 

I could add a method that grants mutable access to the entire `self.buf` regardless of the state of `self.used` and remove `unfilled` and `filled_mut` but I find the distinction between the filled and unfilled parts useful in detection misuse of the API. the `SliceBuffer` has virtually no "unfilled" section so you cannot `append_hs(true)` into it.

I can split this commit into two and add a `whole_mut_buffer` method in the first just to remove it in favor of `unfilled` and `filled_mut` in the second commit if you think that makes the git history clearer.

rustls/src/conn/unbuffered.rs
```diff
@@ -0,0 +1,392 @@
...
+
+                    Err(e) => {
+                        self.core.state = Err(e.clone());
+                        return Err(e);
+                    }
+                }
+
+                self.core.state = Ok(state);
+            } else if self.wants_write {
+                break MustTransmitTlsData { conn: self }.into();
```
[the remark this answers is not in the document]
japaric: My idea (hope?) was to make it easier to batch records into a single TCP write by having an explicit `MustTransmitTlsData` state. without it, the end user may be tempted to do a `write` on each `MustEncode` state. As implemented, the examples certainly batch records so that only 1.5 - 2 (depending on the TLS version) round-trips are necessary to complete the handshake

I think we can remove this `wants_write` field and clearly document that after `MustEncode` has been handled the TLS must *eventually* be transmitted but not necessarily be transmitted *each time* `MustEncode` is handled because that results in no batching (more round-trips) -- that feels a bit hard to wrap the head around. I guess we could document that `outgoing_tls` must be transmitted in the `NeedsMoreTlsData` and `TrafficTransit` states if it's not empty ... would that cover all the cases :thinking:  ?

examples/src/bin/unbuffered-client.rs
```diff
@@ -0,0 +1,307 @@
...
+                            Err(e.into())
+                        }
+                    },
+                    outgoing_tls,
+                    &mut outgoing_used,
+                )?;
+            }
+
+            ConnectionState::MustTransmitTlsData(mut state) => {
+                // if let Some(mut may_encrypt_early_data) = state.may_encrypt_early_data() {
```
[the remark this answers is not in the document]
japaric: this gets uncommented in a later commit and can't be commented out in this commit because it uses an API that does not yet exist. I'll remove the commented code when tidying up the commit history.

japaric:
docs and module org wise, what do you think about exporting / exposing all the `UnbufferedConnection`-related types under a public `unbuffered` module that includes module-level docs with an overview of the API and perhaps a doctest example?

rustls/src/conn/unbuffered.rs
japaric: that worked. however, now the tests won't compile because they were relying on the generic `UnbufferedConnectionCommon<Data>` having the `process_tls_records` method. I basically did this
``` rust
impl UnbufferedConnectionCommon<ClientConnectionData> {
    pub fn process_tls_records<>(/* .. */) -> /* .. */ {
        // call common code
    }
}

impl UnbufferedConnectionCommon<ServerConnectionData> {
    pub fn process_tls_records<>(/* .. */) -> /* .. */ {
        // perform server-only logic
        // call common code
    }
}
```
and now I cannot add this because of the name collision
``` rust
impl<Data> UnbufferedConnectionCommon<Data> {
    pub fn process_tls_records<>(/* .. */) -> /* .. */ {
        // call common code
    }
}
```
however, I could add that last method to the generic type under a different name ...

another option would be to add an `UnbufferedConnection` enum that runtime-represents both the client and server sides and have the `process_tls_records` method there. the tests could then use that `enum` but that would result in several changes in the test code

japaric: > now the tests won't compile because they were relying on the generic UnbufferedConnectionCommon<Data> having the process_tls_records method

fixing up the tests to not rely on the generic struct having that method turned out to be not too onerous. I didn't have to add the `enum`.

rustls/src/conn/unbuffered.rs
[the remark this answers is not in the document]
ctz: Right, I think I understand and agree it is a necessity to have this to signal to consumers when we've finished writing a flight of messages; so they can flush/uncork/etc. the socket.

djc:
I'll suggest we trim this PR down to only the deframer buffer commits in order to make it more digestible?

japaric:
> I'll suggest we trim this PR down to only the deframer buffer commits in order to make it more digestible?

separated the deframer commits into #1595

rustls/src/record_layer.rs
```diff
@@ -204,6 +206,13 @@ impl RecordLayer {
         self.write_seq
     }
 
+    /// Returns the number of remaining write sequences
```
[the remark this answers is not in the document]
japaric: opted for not doing this since adding it on its own would require putting an `#[allow(dead_code]` attribute on it that gets removed in the immediate commit

japaric:
rebased to fix a merge conflict. I also have fleshed out more the API docs

rustls/tests/unbuffered.rs
```diff
@@ -0,0 +1,204 @@
...
+                State::NeedsMoreTlsData => buffers.server_send(),
+                State::TrafficTransit => client_handshake_done = true,
+            }
+
+            match advance_server(&mut server, &mut buffers.server, NO_ACTIONS) {
+                State::EncodedTlsData => {}
+                State::MustTransmitTlsData => buffers.server_send(),
+                State::NeedsMoreTlsData => buffers.client_send(),
+                State::TrafficTransit => server_handshake_done = true,
+            }
```
ctz: At the moment this test is more like example code, and doesn't fail unless it can't be completed within 100 events.

It might be better for this test to observe the precise expected sequences of the underlying `ConnectionState` events (though not the values or lengths of data involved).

examples/src/lib.rs
```diff
@@ -1,8 +1,49 @@
 use std::error::Error;
-use std::io::{Read, Write};
+use std::fs::File;
+use std::io::{BufReader, Read, Write};
 use std::net::TcpStream;
 
-use rustls::unbuffered::InsufficientSizeError;
+use rustls::client::ClientConnectionData;
```
ctz, line 6: I would suggest we don't do this commit. Ideally, the examples here should be standalone, even if it means duplication between different examples. The idea should be that downstream users can take the entire file into their own repo, add a dependency on the `rustls` crate (and co.), and end up with something that builds.

rustls/tests/unbuffered.rs
ctz: An idea to do this:

- have `let mut client_transcript = Vec::<String>::new();`, ditto for server
- for every each `ConnectionState` seen by `advance_client`, push the output of impl Debug of  it (ditto server) -- this also will provide coverage of that code
- at the end of the test, `assert_eq!(client_transcript, vec![ "MustEncodeTlsData", "MustTransmitTlsData", ... ]);`

examples/src/lib.rs
pvdrz: fixed

rustls/tests/unbuffered.rs
```diff
@@ -15,17 +15,106 @@ mod common;
 
 const MAX_ITERATIONS: usize = 100;
 
+const CLIENT_TRANSCRIPT_TLS13: &[&str] = &[
+    "EncodedTlsData",
```
ctz, line 19: Really, this should be in terms of the public API (`ConnectionState`) rather than the test code (`State`). As it is, this tests the test code more than the library. This also needs to extend to every test in this file.

I would also separate the tests by version (eg, `fn tls12_handshake`, `fn tls13_handshake`), and put the expectations inline in the test rather than at the top level.

ctz: For the sake of clarity and speed I've push a couple of commits to https://github.com/rustls/rustls/tree/jbp-unbuffered-api-testing which do this -- feel free to review and squash those into your top commit?

pvdrz: thanks, I just squashed those into my last commit

pvdrz:
@ctz I rebased this PR on top of main given that https://github.com/rustls/rustls/pull/1595 was merged.

examples/src/lib.rs
ctz: I think we still have this in [50911be](https://github.com/rustls/rustls/pull/1583/commits/50911be8a80e2b6504b295516d86a9c3112e361f)

examples/src/bin/unbuffered-client.rs
```diff
@@ -0,0 +1,281 @@
...
+                .iter()
+                .cloned(),
+        );
+    }
+    Ok(root_store)
+}
+
+fn build_http_request() -> Vec<u8> {
+    format!("GET / HTTP/1.1\r\nHost: {SERVER_NAME}\r\nConnection: close\r\nAccept-Encoding: identity\r\n\r\n").into_bytes()
+}
```
cpu, line 259: I think this helper might be too small to justify being a separate fn. It's only invoked from `encrypt_http_response` and inlining it there feels like a clarity improvement.

examples/src/bin/unbuffered-client.rs
```diff
@@ -0,0 +1,281 @@
...
+            ConnectionState::MustEncodeTlsData(mut state) => {
+                try_or_resize_and_retry(
+                    |out_buffer| state.encode(out_buffer),
+                    |e| {
+                        if let EncodeError::InsufficientSize(is) = &e {
+                            Ok(*is)
+                        } else {
+                            Err(e.into())
+                        }
+                    },
```
cpu, line 112: The `map_err` closure provided to `try_or_resize_and_retry` here and on L143 looks identical. Does this need to be a param? If yes, maybe this common code could be lifted into a helper fn to provide in each case.

rustls/src/client/client_conn.rs
```diff
@@ -663,6 +663,37 @@ impl ConnectionCore<ClientConnectionData> {
...
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl DerefMut for UnbufferedClientConnection {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
```
cpu, line 690: These should be below the inherent impl of `UnbufferedClientConnection` (c.f. [ordering for a given type](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#ordering-for-a-given-type))

rustls/src/common_state.rs
```diff
@@ -186,6 +187,78 @@ impl CommonState {
         self.send_plain(data, Limit::Yes)
     }
 
+    pub(crate) fn eager_send_some_plaintext(
```
cpu, line 190: Could this fn be updated with some comments describing the process being undertaken? It's a little tricky to follow.

rustls/src/common_state.rs
```diff
@@ -493,6 +566,50 @@ impl CommonState {
         self.send_warning_alert_no_log(AlertDescription::CloseNotify);
     }
 
+    pub(crate) fn eager_send_close_notify(
```
cpu, line 569: This fn looks like it shares a decent amount of code with `eager_send_some_plaintext` - is there a way to reduce some of the duplication?

rustls/src/conn/unbuffered.rs
```diff
@@ -0,0 +1,449 @@
...
+            if let Some(chunk) = self
+                .core
+                .common_state
+                .received_plaintext
+                .pop()
+            {
+                break (
+                    buffer.pending_discard(),
+                    AppDataAvailable::new(self, incoming_tls, chunk).into(),
+                );
```
cpu, line 61: TIL you can break with a return value, cool :stars:

rustls/src/conn/unbuffered.rs
```diff
@@ -0,0 +1,449 @@
...
+use core::num::NonZeroUsize;
+use core::{fmt, mem};
+use std::error::Error as StdError;
+
+use super::UnbufferedConnectionCommon;
+use crate::msgs::deframer::DeframerSliceBuffer;
+use crate::Error;
+
+impl<Data> UnbufferedConnectionCommon<Data> {
+    /// Processes the TLS records in `incoming_tls` buffer
```
cpu, line 15: I think this doc string could be expanded. The way it's written now makes it sound like it would process all of the records in `incoming_tls` buf, but the impl breaks after specific records/states.

rustls/src/conn/unbuffered.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// This value MUST be handled prior to calling
+    /// [`UnbufferedConnectionCommon::process_tls_records`] again. See the documentation on the
+    /// variants of [`ConnectionState`] for more details.
+    pub state: ConnectionState<'c, 'i, Data>,
+}
+
+/// The state of the [`UnbufferedConnectionCommon`] object
+#[non_exhaustive] // for forwards compatibility; to support caller-side certificate verification
+pub enum ConnectionState<'c, 'i, Data> {
+    /// One, or more, application data record is available
```
cpu, line 152: small typo: "application data record is available" -> "application data records are available"

rustls/src/conn/unbuffered.rs
```diff
@@ -0,0 +1,449 @@
...
+                .debug_struct("NeedsMoreTlsData")
+                .field("num_bytes", num_bytes)
+                .finish(),
+
+            Self::TrafficTransit(..) => f.debug_tuple("TrafficTransit").finish(),
+        }
+    }
+}
+
+/// Application-data is available
```
cpu, line 268: nit: "Application-data is available" -> "Application data is available".

rustls/src/unbuffered.rs
```diff
@@ -0,0 +1,33 @@
+//! Unbuffered connection API
+//!
+//! This is an alternative to the [`crate::ConnectionCommon`] API that does not internally buffers
+//! TLS nor plaintext data. Instead those buffers are managed by the API user so they have
```
cpu, line 4: small typo: "API that does not internally buffers TLS nor plaintext data" -> "API that does not internally buffer TLS nor plaintext data"

rustls/src/unbuffered.rs
```diff
@@ -0,0 +1,33 @@
...
+//! [`UnbufferedConnectionCommon::process_tls_records`]. The state machine API is documented in
+//! [`ConnectionState`].
+//!
+//! # Examples
+//!
+//! [`unbuffered-client`] and [`unbuffered-server`] are examples that fully exercise the API in
+//! std, non-async context.
+//!
+//! [`unbuffered-client`]: https://github.com/rustls/rustls/blob/main/examples/src/bin/unbuffererd-client.rs
+//! [`unbuffered-server`]: https://github.com/rustls/rustls/blob/main/examples/src/bin/unbuffererd-server.rs
```
cpu, line 26: The `unbuffered-server` example isn't added in this commit, should this doc update happen afterwards, or only mention `unbuffered-client`?

examples/src/bin/unbuffered-server.rs
```diff
@@ -0,0 +1,255 @@
...
+use rustls::unbuffered::{
+    AppDataRecord, ConnectionState, EncodeError, EncryptError, InsufficientSizeError,
+    UnbufferedStatus,
+};
+use rustls::ServerConfig;
+use rustls_pemfile::Item;
+
+const KB: usize = 1024;
+const INCOMING_TLS_BUFSIZ: usize = 16 * KB;
+const OUTGOING_TLS_INITIAL_BUFSIZ: usize = 0;
```
cpu, line 18: same nit feedback here RE: SIZE vs SIZ.

examples/src/bin/unbuffered-server.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            f(&mut outgoing_tls[*outgoing_used..])?
+        }
+    };
+
+    *outgoing_used += written;
+
+    Ok(written)
+}
+fn recv_tls(
```
cpu, line 224: nit: whitespace between fns.

examples/src/bin/unbuffered-server.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    assert!(received_request);
+    assert!(sent_response);
+    assert_eq!(0, incoming_used);
+    assert_eq!(0, outgoing_used);
+
+    Ok(())
+}
+
+fn try_or_resize_and_retry<E>(
```
cpu, line 198: I think this example should be using the helpers pulled out in 50911be8a80e2b6504b295516d86a9c3112e361f right?

examples/src/bin/unbuffered-server.rs
```diff
@@ -0,0 +1,255 @@
...
+) -> Result<(), Box<dyn Error>> {
+    sock.write_all(&outgoing_tls[..*outgoing_used])?;
+    eprintln!("sent {outgoing_used}B of data");
+    *outgoing_used = 0;
+    Ok(())
+}
+
+fn build_http_response() -> &'static [u8] {
+    b"HTTP/1.0 200 OK\r\nConnection: close\r\n\r\nHello world from rustls unbuffered server\r\n"
+}
```
cpu, line 248: same feedback as in the client example that this fn might not be useful enough on its own to justify being separated from the call-site

examples/src/bin/unbuffered-server.rs
```diff
@@ -0,0 +1,255 @@
...
+const KB: usize = 1024;
+const INCOMING_TLS_BUFSIZ: usize = 16 * KB;
+const OUTGOING_TLS_INITIAL_BUFSIZ: usize = 0;
+const MAX_EARLY_DATA_SIZE: Option<u32> = Some(128);
+const MAX_FRAGMENT_SIZE: Option<usize> = None;
+
+const PORT: u16 = 1443;
+const MAX_ITERATIONS: usize = 30;
+const CERTFILE: &str = "localhost.pem";
+const PRIV_KEY_FILE: &str = "localhost-key.pem";
```
cpu, line 25: could these be optionally sourced from env vars ? Having them hardcoded will make using it a little bit awkward.

rustls/src/server/server_conn.rs
```diff
@@ -561,6 +561,43 @@ impl From<ServerConnection> for crate::Connection {
...
+    }
+}
+
+impl DerefMut for UnbufferedServerConnection {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
+
+impl UnbufferedServerConnection {
```
cpu, line 585: Same feedback as for the client conn: inherent impl before the `Deref`/`DerefMut` impls please

rustls/src/server/server_conn.rs
```diff
@@ -561,6 +561,43 @@ impl From<ServerConnection> for crate::Connection {
...
+    /// Make a new ClientConnection. `config` controls how we behave in the TLS protocol, `name` is
+    /// the name of the server we want to talk to.
```
cpu, line 587: I think this doc string isn't correct - there's no `name` argument and the connection _is_ the server.

examples/src/bin/unbuffered-client.rs
```diff
@@ -0,0 +1,281 @@
...
+    let mut outgoing_tls = vec![0; OUTGOING_TLS_INITIAL_BUFSIZ];
+
+    converse(&config, false, &mut incoming_tls, &mut outgoing_tls)?;
+
+    Ok(())
+}
+
+fn converse(
+    config: &Arc<ClientConfig>,
+    send_early_data: bool,
```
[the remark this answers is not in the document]
cpu: Ah, I see: 9d429462cc13fa573afa64cd6d35d2ee05caedf9 introduces meaningful usage.

I think the early data parts should be left out in this commit and only included in 9d429462cc13fa573afa64cd6d35d2ee05caedf9.

rustls/src/client/client_conn.rs
```diff
@@ -694,6 +699,86 @@ impl UnbufferedClientConnection {
...
+/// Allows encrypting early (RTT-0) data
+pub struct MayEncryptEarlyData<'c> {
+    conn: &'c mut UnbufferedConnectionCommon<ClientConnectionData>,
+}
+
+impl MayEncryptEarlyData<'_> {
+    /// Encrypts `application_data` into the `outgoing_tls` buffer
+    ///
+    /// returns the number of bytes that were written into `outgoing_tls`, or an error if
+    /// the provided buffer was too small. in the error case, `outgoing_tls` is not modified
```
cpu, line 731: nit: "in the error case" - "In the error case"

rustls/src/conn/unbuffered.rs
```diff
@@ -127,6 +158,12 @@ pub enum ConnectionState<'c, 'i, Data> {
     /// Connection has been closed by the peer
     ConnectionClosed,
 
+    /// One, or more, early (RTT-0) data record is available
```
cpu, line 161: nit: "data record is available" -> "data records are available"

rustls/src/conn/unbuffered.rs
```diff
@@ -267,6 +314,57 @@ impl<'c, 'i, Data> AppDataAvailable<'c, 'i, Data> {
     }
 }
 
+/// Application-data is available
```
cpu, line 317: I think this should say "Early application-data is available" to distinguish from `AppDataAvailable` that already says "Application-data is available"

cpu (commented):
Here's some initial feedback after reviewing up to "test handshake" (650e37). I'll continue from there tomorrow.

examples/src/bin/unbuffered-client.rs
```diff
@@ -0,0 +1,281 @@
+use std::error::Error;
```
ctz, line 1: This example should start with a comment which explains what it does.

Generally I think example code should be focused on demonstrating one feature or API, to the exclusion of others (even if it is convenient for one-off testing in other scenarios, ideally those tests should be automated). I think the focus of this example should be `converse()`, and we should minimise the distractions in the remainder as far as possible.

Below, I think:
- the commented-out code should be removed,
- `CERTFILE` and the related branch in `build_root_store` (at which point `build_root_store` can be inlined into its caller, and the start of `main()` will be identical to `simpleclient.rs`.)

rustls/src/common_state.rs
```diff
@@ -493,6 +566,50 @@ impl CommonState {
...
+        &mut self,
+        outgoing_tls: &mut [u8],
+    ) -> Result<usize, EncryptError> {
+        debug_assert!(self.record_layer.is_encrypting());
+
+        let desc = AlertDescription::CloseNotify;
```
ctz, line 575: Suggest inlining this single-use binding.

rustls/src/msgs/message.rs
```diff
@@ -349,6 +350,13 @@ impl<'a> BorrowedPlainMessage<'a> {
             payload: Payload(self.payload.to_vec()),
         }
     }
+
+    pub fn encoded_len(&self, record_layer: &RecordLayer) -> usize {
+        let typ = 1;
+        let version = 2;
+        let len = 2;
+        typ + version + len + record_layer.encrypted_len(self.payload.len())
```
ctz, line 358:
```suggestion
        OpaqueMessage::HEADER_SIZE + record_layer.encrypted_len(self.payload.len())
```

examples/src/bin/unbuffered-client.rs
pvdrz: It is almost identical but it isn't. One of them takes an `EncryptError` and the other one takes an `EncodeError`. I could factor the common bits and replace this `map_err` param with one that tries to transform the error into an `InsufficientSize` value.

pvdrz: It is almost identical but it isn't. One of them takes an `EncodeError` and the other takes an `EncryptError`.

pvdrz:
@ctz this is ready for another review

examples/src/bin/unbuffered-client.rs
cpu: Ah ok! My mistake.

rustls/tests/unbuffered.rs
```diff
@@ -0,0 +1,204 @@
...
+                State::EncodedTlsData => {}
+                State::MustTransmitTlsData => buffers.server_send(),
+                State::NeedsMoreTlsData => buffers.client_send(),
+                State::TrafficTransit => server_handshake_done = true,
+            }
+
+            count += 1;
+
+            assert!(count <= 100, "handshake {version:?} was not completed");
+        }
```
cpu, line 479: I think it might be worthwhile to assert `client_handshake_done` and `server_handshake_done` after the while loop here:

```
    assert!(client_handshake_done);
    assert!(server_handshake_done);
```

rustls/tests/unbuffered.rs
```diff
@@ -0,0 +1,204 @@
...
+    EncodedTlsData,
+    NeedsMoreTlsData,
+    TrafficTransit,
+    MustTransmitTlsData,
+}
+
+const NO_ACTIONS: Actions = Actions {};
+
+#[derive(Clone, Copy, Debug)]
+struct Actions {}
```
cpu: What's the purpose of the `Actions` struct? In this revision it's always `NO_ACTIONS` at the call-sites, and it's threaded through to `handle_state` where it's an ignored parameter. If a later commit turns this into something functional I think it would be less confusing to introduce it then as opposed to making it a no-op here.

examples/src/bin/unbuffered-async-client.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[cfg_attr(not(feature = "async-std"), tokio::main(flavor = "current_thread"))]
+#[cfg_attr(feature = "async-std", async_std::main)]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let mut root_store = RootCertStore::empty();
+    root_store.extend(
+        webpki_roots::TLS_SERVER_ROOTS
+            .iter()
+            .cloned(),
+    );
```
cpu, line 44: I think this can be simplified to:
```rust
let root_store = RootCertStore {
    roots: webpki_roots::TLS_SERVER_ROOTS.into(),
};
```

Ditto for `unbuffered-client.rs` and `unbuffered-server.rs` (sorry I missed that in the prev review).

examples/src/bin/unbuffered-async-client.rs
```diff
@@ -0,0 +1,245 @@
...
+                    &mut may_encrypt,
+                    outgoing_tls,
+                    &mut outgoing_used,
+                ) {
+                    send_tls(&mut sock, outgoing_tls, &mut outgoing_used).await?;
+                    recv_tls(&mut sock, incoming_tls, &mut incoming_used).await?;
+                } else if !received_response {
+                    // this happens in the TLS 1.3 case. the app-data was sent in the preceding
+                    // `MustTransmitTlsData` state. the server should have already a response which
+                    // we can read out from the socket
```
cpu, line 150: I think this is the same typo I flagged in an earlier commit; "the server should have already" -> "the server should have already written"

cpu (commented):
Only minor comments from me after picking up the review I half-completed yesterday. I've made it through the whole diff now.

pvdrz:
@cpu I addressed your latest review. @ctz I removed the `num_bytes` field. I'll think about the discard issue next week.

rustls/src/common_state.rs
```diff
@@ -186,6 +187,78 @@ impl CommonState {
...
+    ) -> Result<usize, EncryptError> {
+        if plaintext.is_empty() {
+            return Ok(0);
+        }
+
+        let fragments = self.message_fragmenter.fragment_slice(
+            ContentType::ApplicationData,
+            ProtocolVersion::TLSv1_2,
+            plaintext,
+        );
```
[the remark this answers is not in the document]
pvdrz: Ok, I already implemented it :)

ctz:
> * Maybe we could make `process_tls_records` return `UnbufferedStatus`, and either make `state` there a Result

I have experimented with this approach in https://github.com/rustls/rustls/commit/8c5d2a33c76f8d795d989695129818ceae793b81

rustls/src/common_state.rs
```diff
@@ -186,6 +187,49 @@ impl CommonState {
         self.send_plain(data, Limit::Yes)
     }
 
+    pub(crate) fn eager_send_some_plaintext(
```
djc, line 190: I didn't love the `send_some_plaintext()` name and `eager_` is even worse (since it's longer). What is eager about it? The name should clearly explain how it's different from `send_some_plaintext()`. Maybe it's `buffer_plaintext()` vs `write_plaintext()`?

rustls/src/common_state.rs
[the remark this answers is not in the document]
djc: Doesn't seem to be showing up in the PR yet, unless it's gone in another PR?

rustls/src/conn/unbuffered.rs
```diff
@@ -0,0 +1,442 @@
...
+    AppDataAvailable(AppDataAvailable<'c, 'i, Data>),
+
+    /// Connection has been cleanly closed by the peer
+    ConnectionClosed,
+
+    /// A Handshake record is ready for encoding
+    ///
+    /// Call [`MustEncodeTlsData::encode`] on the enclosed object, providing an `outgoing_tls`
+    /// buffer to store the encoding
+    MustEncodeTlsData(MustEncodeTlsData<'c, Data>),
```
djc, line 170: I don't love these variant (and state type) names. In particular:

* Different style (`AppDataAvailable` is a state, `MustTransmitTlsData` describes the next action)
* Repetition (`ConnectionState::ConnectionClosed`)
* Verbosity (`MustTransmitTlsData`)
* Unclearly linked to connection phases (handshaking vs traffic)

I'd propose something more like this:

* `EncodeHandshake`
* `TransmitHandshake`
* `HandshakeBlocked`
* `ReadTraffic`
* `WriteTraffic`
* `Closed`

(It's also weird to me that `TransmitHandshake` is a separate `ConnectionState` rather than a method on some result of a method on `EncodeHandshake`?)

rustls/src/unbuffered.rs
```diff
@@ -0,0 +1,32 @@
...
+//! [`ConnectionState`].
+//!
+//! # Examples
+//!
+//! [`unbuffered-client`] and [`unbuffered-server`] are examples that fully exercise the API in
+//! std, non-async context.
+//!
+//! [`unbuffered-client`]: https://github.com/rustls/rustls/blob/main/examples/src/bin/unbuffererd-client.rs
+
+pub use crate::conn::unbuffered::{
```
djc, line 28: IMO this should be an inline `mod` in `lib.rs`, probably with a shorter docstring (which just relies on pointers to the relevant types).

examples/src/bin/unbuffered-client.rs
```diff
@@ -100,6 +109,25 @@ fn converse(
             }
 
             ConnectionState::MustTransmitTlsData(mut state) => {
+                if let Some(mut may_encrypt_early_data) = state.may_encrypt_early_data() {
+                    let written = helpers::try_or_resize_and_retry(
+                        |out_buffer| may_encrypt_early_data.encrypt(EARLY_DATA, out_buffer),
+                        |e| {
+                            if let EarlyDataError::Encrypt(EncryptError::InsufficientSize(is)) = &e
```
djc, line 116: Style: I think this would look better as a `match`.

rustls/src/conn/unbuffered.rs
```diff
@@ -6,19 +6,51 @@ use core::{fmt, mem};
...
 
-impl<Data> UnbufferedConnectionCommon<Data> {
+impl UnbufferedConnectionCommon<ClientConnectionData> {
+    /// Processes the TLS records in `incoming_tls` buffer until a new [`UnbufferedStatus`] is
+    /// reached.
+    pub fn process_tls_records<'c, 'i>(
+        &'c mut self,
+        incoming_tls: &'i mut [u8],
+    ) -> Result<UnbufferedStatus<'c, 'i, ClientConnectionData>, Error> {
+        self.process_tls_records_common(incoming_tls, |_| None::<()>, |_, _, _| unreachable!())
```
djc, line 21: Does this really need the turbofish?

rustls/src/conn/unbuffered.rs
```diff
@@ -125,6 +157,12 @@ pub enum ConnectionState<'c, 'i, Data> {
     /// Connection has been cleanly closed by the peer
     ConnectionClosed,
 
+    /// One, or more, early (RTT-0) data records are available
+    ///
+    /// See [`EarlyDataAvailable`] for more details on how to use the enclosed object to access
+    /// the received data.
```
djc, line 163: I think we should skip notes like this -- seems fairly intuitive.

rustls/tests/unbuffered.rs
```diff
@@ -554,10 +563,17 @@ fn handle_state<Data>(
...
-                .unwrap();
-            outgoing.advance(written);
+            write_with_buffer_size_checks(
+                |out_buf| state.encode(out_buf),
+                |e| {
+                    if let EncodeError::InsufficientSize(ise) = e {
+                        ise
+                    } else {
+                        unreachable!()
+                    }
```
djc, line 689: Deduplicate this?

examples/Cargo.toml
```diff
@@ -18,3 +18,5 @@ rustls-pemfile = "2"
 serde = "1.0"
 serde_derive = "1.0"
 webpki-roots = "0.26"
+async-std = { version = "1.12.0", features = ["attributes"], optional = true }
```
djc, line 21: Nit: keep these in alphabetic order.

examples/src/bin/unbuffered-async-client.rs
```diff
@@ -0,0 +1,242 @@
...
+use rustls::client::{ClientConnectionData, UnbufferedClientConnection};
+use rustls::unbuffered::{
+    AppDataRecord, ConnectionState, EncodeError, EncryptError, MayEncryptAppData, UnbufferedStatus,
+};
+#[allow(unused_imports)]
+use rustls::version::{TLS12, TLS13};
+use rustls::{ClientConfig, RootCertStore};
+use rustls_examples as helpers;
+
+#[cfg(not(feature = "async-std"))]
```
djc, line 14: Import nits: one line per module, no empty lines before the tokio imports, avoid unused imports but added the appropriate feature guard instead.

rustls/src/conn/unbuffered.rs
ctz: I don't think `EncodeHandshake` and `TransmitHandshake` are specific to handshake messages, eg they are also required for non-handshake messages (both during the handshake, and after).

rustls/src/conn/unbuffered.rs
pvdrz: yes, but we can set the closure's type instead which is shorter

examples/src/bin/unbuffered-client.rs
pvdrz: inlining `converse` wouldn't be practical as it is called twice when introducing early data in the unbuffered client example. I'm happy to hear any suggestions about the name.

rustls/src/common_state.rs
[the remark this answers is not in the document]
pvdrz: no, it is on this PR: https://github.com/rustls/rustls/pull/1583/files/15f1b55b984a6dd10deb99173f00949cf9322c29#diff-a67cab0bf382f9ffb8c9d394a9134814c9def1d645772afebea091647b84b0e5R210

rustls/src/common_state.rs
```diff
@@ -181,11 +182,54 @@ impl CommonState {
...
+            ProtocolVersion::TLSv1_2,
+            plaintext,
+        );
+
+        let remaining_encryptions = self
+            .record_layer
+            .remaining_write_seq()
+            .ok_or(EncryptError::EncryptExhausted)?;
+
+        if fragments.len() as u64 > remaining_encryptions.get() {
```
pvdrz, line 210: > Doesn't seem to be showing up in the PR yet, unless it's gone in another PR?

here @djc

examples/src/lib.rs
pvdrz: fixed again :sweat_smile:

examples/Cargo.toml
```diff
@@ -18,3 +18,6 @@ rustls-pemfile = "2"
 serde = "1.0"
 serde_derive = "1.0"
 webpki-roots = "0.26"
+# async runtime
+# async-std = { version = "1.12.0", features = ["attributes"] }
+tokio = { version = "1.34.0", features = ["io-util", "macros", "net", "rt"] }
```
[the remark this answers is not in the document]
pvdrz: fixed

rustls/src/conn/unbuffered.rs
pvdrz: I renamed the variants.

ctz:
> I'll think about the discard issue next week.

Any thoughts on that?

pvdrz:
> > I'll think about the discard issue next week.
> 
> Any thoughts on that?

I think your approach of putting the result inside the state field works. I'll bring those changes locally and see what we can do.

ctz:
> I think your approach of putting the result inside the state field works. I'll bring those changes locally and see what we can do.

Great. I have just rebased https://github.com/rustls/rustls/tree/jbp-unbuffered-api-testing-2 against this PR.

pvdrz:
> > I think your approach of putting the result inside the state field works. I'll bring those changes locally and see what we can do.
> 
> Great. I have just rebased https://github.com/rustls/rustls/tree/jbp-unbuffered-api-testing-2 against this PR.
I just merged those changes into this branch :+1:

ctz (approved):

examples/src/bin/unbuffered-client.rs
cpu: > This example should start with a comment which explains what it does.

I think this feedback is still unaddressed.

rustls/src/common_state.rs
cpu: I think this feedback is still unaddressed.

examples/src/bin/unbuffered-server.rs
cpu: What happened to the commit that pulled out the helpers? It looks like it got dropped from this branch and so this feedback remains unresolved.

rustls/tests/unbuffered.rs
cpu: This was marked resolved, but I don't see the asserts. Did you disagree with the suggestion, or did it get overlooked by mistake?

.github/workflows/daily-tests.yml
```diff
@@ -84,6 +84,12 @@ jobs:
       - name: Check simple 0rtt client
         run: cargo run --locked --bin simple_0rtt_client
 
+      - name: Check unbuffered tokio client 
+        run: cargo run --locked --bin unbuffered-async-client
+
+      - name: Check unbuffered async-std client 
+        run: cargo run --locked --bin unbuffered-async-client --features=async-std
```
cpu, line 91: What about the `unbuffered-client.rs` and `unbuffered-server.rs` examples?

rustls/tests/unbuffered.rs
```diff
@@ -888,3 +888,41 @@ fn make_connection_pair(
     let server = UnbufferedServerConnection::new(Arc::new(server_config)).unwrap();
     (client, server)
 }
+
+#[test]
+fn server_receives_handshake_byte_by_byte() {
```
cpu, line 889: This commit has the message: "See comment about how NeedMoreTlsData's num_bytes field is currently
unused.", but I don't see a comment about this in the diff. Am I missing something or should the commit message be adjusted?

ctz:
> What happened to the commit that pulled out the helpers? It looks like it got dropped from this branch and so this feedback remains unresolved.

I asked for that commit to be dropped, on the basis that example code should be standalone as far as possible.

cpu:
> I asked for that commit to be dropped, on the basis that example code should be standalone as far as possible.

Ahhh! Ok, I will resolve. That rationale makes sense to me.

examples/src/bin/unbuffered-server.rs
cpu: Explained by Ctz: this was removed to make the examples more stand-alone, which makes sense to me!

.github/workflows/daily-tests.yml
pvdrz: I added a step for the unbuffered client. Testing the server would require generating a certificate and then spawning a client that talks to the server which seems like a lot. I don't see the `tlsserver-mio` example being tested here either.

rustls/tests/unbuffered.rs
pvdrz: cc @ctz

.github/workflows/daily-tests.yml
cpu: > I added a step for the unbuffered client. 

Thanks!

> Testing the server would require generating a certificate and then spawning a client that talks to the server which seems like a lot. I don't see the tlsserver-mio example being tested here either.

Fair points :+1: I'll resolve this comment.

pvdrz:
@cpu  not sure why I can't comment but https://github.com/rustls/rustls/pull/1583#discussion_r1428348467 these functions have been renamed already.

cpu:
> @cpu not sure why I can't comment but https://github.com/rustls/rustls/pull/1583#discussion_r1428348467 these functions have been renamed already.

Sorry GitHub has really mangled the interleaving of my review comments with the older ones :sob:

cpu (approved):

pvdrz:
@ctz, @cpu, @djc is there anything else to be done here?

rustls/tests/unbuffered.rs
ctz: Please could you remove that bit from the commit message?

pvdrz: done

tshepang:
epic!

PR #1586, opened 2023-11-13
aws-lc-rs: consume new TLS-specific APIs

ctz:
aws-lc-rs 1.4 adds some TLS-specific APIs which are implemented inside in the FIPS boundary:

- TLS1.2 PRF
- an AEAD API that notices if rustls messes up the GCM nonce sequence

This PR moves to using those.


ctz:
looking at why the coverage is so bad for this

admin/coverage
```diff
@@ -7,4 +7,9 @@ cargo llvm-cov clean --workspace
 
 cargo build --locked --all-targets --all-features
 cargo test --locked --all-features
+
```
djc, line 10: Bullet lists in a commit message are never a good sign... Maybe split this up?

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -14,7 +14,7 @@ use crate::tls12::Tls12CipherSuite;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 
-use super::ring_like::aead;
+use super::ring_like::{aead, tls_prf};
```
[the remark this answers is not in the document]
ctz: I think `use aws_lc_rs::{aead, tls_prf};` would be clearer now, given this file is specific to aws-lc-rs.

rustls/src/crypto/aws_lc_rs/tls13.rs
```diff
@@ -220,6 +248,57 @@ impl MessageDecrypter for AeadMessageDecrypter {
...
 }
 
+struct GcmMessageEncrypter {
+    enc_key: aead::TlsRecordSealingKey,
+    iv: Iv,
+}
+
+impl MessageEncrypter for GcmMessageEncrypter {
+    fn encrypt(&mut self, msg: BorrowedPlainMessage, seq: u64) -> Result<OpaqueMessage, Error> {
+        let total_len = msg.payload.len() + 1 + self.enc_key.algorithm().tag_len();
```
cpu, line 258: I think this will need to be caught up with https://github.com/rustls/rustls/pull/1579 once it lands.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -177,6 +178,19 @@ fn load_key(filename: &str) -> PrivateKeyDer<'static> {
...
 
+fn load_root_certs() -> Arc<RootCertStore> {
+    let mut roots = RootCertStore::empty();
+
+    // this is not actually used by the tests, but must be non-empty
+    for c in load_cert("cert.pem") {
+        roots
+            .add(c)
+            .expect("cannot convert cert.pem to trust anchor");
+    }
```
cpu, line 189: given the comment maybe this could just be:
```suggestion
    // this is not actually used by the tests, but must be non-empty
   roots.add_parsable_certificates(load_cert("cert.pem"));
```

cpu (commented):
Here's some initial comments, nothing major!

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -0,0 +1,413 @@
...
+}
+
+struct Tls12Prf(&'static tls_prf::Algorithm);
+
+impl Prf for Tls12Prf {
+    fn for_secret(&self, output: &mut [u8], secret: &[u8], label: &[u8], seed: &[u8]) {
+        let result = tls_prf::Secret::new(self.0, secret)
+            .unwrap()
+            .derive(label, seed, output.len())
+            .unwrap();
```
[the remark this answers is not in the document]
ctz: Well, hmm. First I think it's a mistake that we allow `export_keying_material()` to succeed with an empty output buffer; no good can come of that! I will add a commit which prevents that at a higher level and tests the behaviour.

I did do a cursory check that `CRYPTO_tls1_prf` (ie, `tls1_P_hash`) doesn't report failure on anything environment or otherwise in our control (eg, it doesn't use the allocator). It would be a bit of an imposition to make this fallible, I think, especially because it's such a small and well-defined deterministic function. I will comment the unwraps.

admin/coverage
ctz: Have split this in two.

rustls/src/crypto/aws_lc_rs/tls13.rs
ctz: rebased

ctz:
Indeed -- have dropped a note on https://github.com/aws/aws-lc-rs/commit/27d447f3b3f43f899424742bd6cb103d7773efcb#r132855645

djc:
Would be good to rebase this?

ctz:
> Would be good to rebase this?

Done!

> Hm, I think the aws-lc-sys bump in this branch broke the min versions check:

About to open a PR there for this, but in the meantime #1601 made this irrelevent.

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -384,3 +384,38 @@ fn gcm_iv(write_iv: &[u8], explicit: &[u8]) -> Iv {
...
+struct Tls12Prf(&'static tls_prf::Algorithm);
+
+impl Prf for Tls12Prf {
+    fn for_secret(&self, output: &mut [u8], secret: &[u8], label: &[u8], seed: &[u8]) {
+        // safety:
+        // - [1] is safe because our caller guarantees `secret` is non-empty; this is
+        //   the only documented error case.
+        // - [2] is safe in practice because the only failure from `derive()` is due
+        //   to zero `output.len()`; this is outlawed at higher levels
+        let result = tls_prf::Secret::new(self.0, secret)
```
djc, line 412: Nit: I don't like the name `result` for this because I feel like bindings called `result` are usually of type `Result`. Maybe `next_secret` or `new`?

djc (approved):

PR #1587, opened 2023-11-14
Bump rustls version to alpha.5

djc:
As requested in #1582.


cpu (approved):
FWIW I don't think we ended up publishing .4.

I bumped the version in anticipation of using the alpha in rustls-ffi, but then was waiting on the aws-lc-rs work to land before publishing. That's happened but I got distracted by other work in the meantime.

PR #1588, opened 2023-11-14
Update semver-compatible versions (again)

djc:
While publishing, I saw a warning:

```
warning: package `cc v1.0.84` in Cargo.lock is yanked in registry `crates-io`, consider updating to a version that is not yanked
```

So here's an update reverting the cc update from #1585 (and taking other updates).


cpu (approved):
GitHub actions seems jammed up (but the status page is all green :roll_eyes:). Stamping this on the assumption CI will build cleanly when the action queue gets rolling again.

PR #1589, opened 2023-11-14
crypto: add HPKE module and traits, provider-example impl

cpu:
This branch introduces a trait for a hybrid public key encryption (HPKE) provider. HPKE is specified in [RFC 9180](https://www.rfc-editor.org/rfc/rfc9180), and is a pre-requisite for implementing [encrypted client hello (ECH)](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17). Implementations of these traits can use the cryptographic provider of their choice to provide HPKE using existing primitives from the crypto provider.

We've tailored the HPKE trait in Rustls to just what is required for ECH, e.g. it doesn't support modes other than the unauthenticated 'base' mode, and it only offers the "single-shot" APIs.

Demonstration of implementing the traits is achieved using [hpke-rs](https://github.com/franziskuskiefer/hpke-rs) with the rust-crypto backend.  Separately we expect to implement a HPKE provider using the *ring* crypto provider for proper ECH support. 

Since the HPKE traits are not yet used in Rustls a unit test of the provider-example HPKE provider based on the RFC 9180 test vectors is added to prevent bitrot and ensure the traits are workable. Likely in the future we will want to move this test logic somewhere outside of the provider-example crate and use it to test the other HPKE provider implementations.

### DIFF
--- a/provider-example/Cargo.toml
+++ b/provider-example/Cargo.toml
@@ -11,6 +11,9 @@ chacha20poly1305 = "0.10.0"
 der = "0.7.0"
 env_logger = "0.10"
 hmac = "0.12.0"
+hpke-rs = "0.1.0"
+hpke-rs-crypto = "0.1.2"
+hpke-rs-rust-crypto = "0.1.2"
 pki-types = { package = "rustls-pki-types", version = "0.2.0" }
 rand_core = "0.6.0"
 rustls = { path = "../rustls", default-features = false, features = ["logging", "tls12"] }
@@ -19,3 +22,8 @@ sha2 = "0.10.0"
 webpki = { package = "rustls-webpki", version = "0.102.0-alpha.1", default-features = false, features = ["alloc", "std"] }
 webpki-roots = "0.26.0-alpha.1"
 x25519-dalek = "2"
+
+[dev-dependencies]
+hex = "0.4.3"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
--- a/provider-example/src/hpke.rs
+++ b/provider-example/src/hpke.rs
@@ -1,0 +1,97 @@
+use std::fmt::{Debug, Formatter};
+
+use hpke_rs_crypto::types::{AeadAlgorithm, KdfAlgorithm, KemAlgorithm};
+use hpke_rs_crypto::HpkeCrypto;
+use hpke_rs_rust_crypto::HpkeRustCrypto;
+
+use rustls::crypto::hpke::{
+    EncapsulatedSecret, Hpke, HpkePrivateKey, HpkeProvider, HpkePublicKey, HpkeSuite,
+};
+use rustls::Error;
+
+pub static HPKE_PROVIDER: &'static dyn HpkeProvider = &HpkeRsProvider {};
+
+/// A Rustls HPKE provider backed by hpke-rs.
+#[derive(Debug)]
+struct HpkeRsProvider {}
+
+impl HpkeProvider for HpkeRsProvider {
+    fn start(&self, suite: &HpkeSuite) -> Result<Box<dyn Hpke>, Error> {
+        Ok(Box::new(HpkeRs(hpke_rs::Hpke::new(
+            hpke_rs::Mode::Base,
+            KemAlgorithm::try_from(suite.kem.get_u16()).map_err(general_err)?,
+            KdfAlgorithm::try_from(suite.sym.kdf_id.get_u16()).map_err(general_err)?,
+            AeadAlgorithm::try_from(suite.sym.aead_id.get_u16()).map_err(general_err)?,
+        ))))
+    }
+
+    fn supports_suite(&self, suite: &HpkeSuite) -> bool {
+        let kem = KemAlgorithm::try_from(suite.kem.get_u16()).ok();
+        let kdf = KdfAlgorithm::try_from(suite.sym.kdf_id.get_u16()).ok();
+        let aead = AeadAlgorithm::try_from(suite.sym.aead_id.get_u16()).ok();
+        match (kem, kdf, aead) {
+            (Some(kem), Some(kdf), Some(aead)) => {
+                HpkeRustCrypto::supports_kem(kem).is_ok()
+                    && HpkeRustCrypto::supports_kdf(kdf).is_ok()
+                    && HpkeRustCrypto::supports_aead(aead).is_ok()
+            }
+            _ => false,
+        }
+    }
+}
+
+struct HpkeRs(hpke_rs::Hpke<HpkeRustCrypto>);
+
+impl Debug for HpkeRs {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("HpkeRsHpke").finish()
+    }
+}
+
+impl Hpke for HpkeRs {
+    fn seal(
+        &mut self,
+        pk_r: &HpkePublicKey,
+        info: &[u8],
+        aad: &[u8],
+        plaintext: &[u8],
+    ) -> Result<(EncapsulatedSecret, Vec<u8>), Error> {
+        let pk_r = hpke_rs::HpkePublicKey::new(pk_r.0.clone());
+        let (enc, ciphertext) = self
+            .0
+            .seal(&pk_r, info, aad, plaintext, None, None, None)
+            .map_err(general_err)?;
+        Ok((EncapsulatedSecret(enc.to_vec()), ciphertext))
+    }
+
+    fn open(
+        &mut self,
+        enc: &EncapsulatedSecret,
+        sk_r: &HpkePrivateKey,
+        info: &[u8],
+        aad: &[u8],
+        ciphertext: &[u8],
+    ) -> Result<Vec<u8>, Error> {
+        let sk_r = hpke_rs::HpkePrivateKey::new(sk_r.secret_bytes().to_vec());
+        self.0
+            .open(
+                enc.0.as_slice(),
+                &sk_r,
+                info,
+                aad,
+                ciphertext,
+                None,
+                None,
+                None,
+            )
+            .map_err(general_err)
+    }
+}
+
+// TODO(XXX): Switch to using `Error::Other(Error::OtherError(err))` once a hpke-rs release
+//   with https://github.com/franziskuskiefer/hpke-rs/pull/44 is available.
+fn general_err(err: impl Debug) -> Error {
+    // Presently hpke_rs::HpkeError does not implement std::error::Error, so we use Debug
+    // and create a general error.
+    Error::General(format!("{:?}", err))
+}
--- a/rustls/src/crypto/hpke.rs
+++ b/rustls/src/crypto/hpke.rs
@@ -1,0 +1,98 @@
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use core::fmt::Debug;
+
+use crate::msgs::enums::HpkeKem;
+use crate::msgs::handshake::HpkeSymmetricCipherSuite;
+use crate::Error;
+
+/// A provider for [RFC 9180] Hybrid Public Key Encryption (HPKE) in base mode.
+///
+/// At a minimum each provider must support the [HPKE ciphersuite profile] required for
+/// encrypted client hello (ECH):
+///  * KEM: DHKEM(X25519, HKDF-SHA256)
+///  * symmetric ciphersuite:  AES-128-GCM w/ HKDF-SHA256
+///
+/// [RFC 9180]: <https://www.rfc-editor.org/rfc/rfc9180.html>
+/// [HPKE ciphersuite profile]: <https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#section-9>
+pub trait HpkeProvider: Debug + Send + Sync + 'static {
+    /// Start setting up to use HPKE in base mode with the chosen suite.
+    ///
+    /// May return an error if the suite is unsupported by the provider.
+    fn start(&self, suite: &HpkeSuite) -> Result<Box<dyn Hpke>, Error>;
+
+    /// Does the provider support the given [HpkeSuite]?
+    fn supports_suite(&self, suite: &HpkeSuite) -> bool;
+}
+
+/// An HPKE suite, specifying a key encapsulation mechanism and a symmetric cipher suite.
+pub struct HpkeSuite {
+    /// The choice of HPKE key encapsulation mechanism.
+    pub kem: HpkeKem,
+
+    /// The choice of HPKE symmetric cipher suite.
+    ///
+    /// This combines a choice of authenticated encryption with additional data (AEAD) algorithm
+    /// and a key derivation function (KDF).
+    pub sym: HpkeSymmetricCipherSuite,
+}
+
+/// An HPKE instance that can be used for base-mode single-shot encryption and decryption.
+pub trait Hpke: Debug + Send + Sync {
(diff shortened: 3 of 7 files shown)


rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,131 @@
...
+pub trait Hpke: Debug + Send + Sync {
+    /// Create a sender context to encrypt messages to a receiver.
+    ///
+    /// This will return an encapsulated secret and a sender context. The sender context
+    /// can be used to seal messages to the receiver.
+    fn setup_sender(
+        &self,
+        pk_r: &HpkePublicKey,
+        info: &[u8],
+        psk: Option<&HpkePreSharedKey>,
```
djc, line 77: Should we wrap up the arguments shared by `setup_sender()` and `setup_receiver()` in a common type?

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn setup(
+        &self,
+        mode: Mode,
+        kem: HpkeKem,
+        kdf: HpkeKdf,
+        aead: HpkeAead,
+    ) -> Result<Box<dyn Hpke>, Error>;
+
+    /// Does the provider support the given mode?
+    fn supports_mode(&self, mode: Mode) -> bool;
```
djc, line 31: What is the utility of these `supports_*()` methods? It feels like either they should be more like getters yielding a slice of supported elements, or, if there's not much we can do if we don't support a particular element, we should leverage the fallibility of the `setup()` method.

djc (commented):
Feels like the trait is fairly specialized to the hpke-rs interface, might be good to think about the conceptually ideal interface for this?

cpu:
> Feels like the trait is fairly specialized to the hpke-rs interface, might be good to think about the conceptually ideal interface for this?

I think it _is_ pretty close to the ideal interface. The main change I can think to suggest is to tailor the traits to only supporting `Mode::Base` since that's all ECH will require and the rest is speculative. That would let us remove the `Option` args to `setup_sender` and `setup_receiver`. Otherwise I think it maps very closely to what ECH requires and I'm not sure I can think of conceptual improvements. I'm open to suggestions though!

rustls/src/crypto/hpke.rs
cpu: We could wrap the `Optional` ones into something describing authentication, but per [my other comment](https://github.com/rustls/rustls/pull/1589#issuecomment-1810932503) maybe we'd prefer to drop them entirely and focus only on base mode?

rustls/src/crypto/hpke.rs
cpu: > What is the utility of these supports_*() methods? 

For the ECH use-case we'll be working with config structures that end up describing [`HpkeKeyConfig`](https://github.com/rustls/rustls/blob/ccb79947a4811412ee7dcddcd0f51ea56bccf101/rustls/src/msgs/handshake.rs#L2279-L2284) instances, that each contain a `Vec<HpkeSymmetricCipherSuite>`. Each [`HpkeSymmetricCipherSuite`](https://github.com/rustls/rustls/blob/ccb79947a4811412ee7dcddcd0f51ea56bccf101/rustls/src/msgs/handshake.rs#L2255-L2258) has KEM, KDF, and AEAD IDs. We'll need a way to match what we support vs what the ECH config advertises to arrive at a compatible suite to use.

>  It feels like either they should be more like getters yielding a slice of supported elements

I had something like this initially, and this might be a place where I really am over-fitting hpke-rs since it doesn't support getting a list. I thought about hardcoding a list, but that feels too fragile.

I think we'd end up doing `contains` testing on a per-element basis against ECH config elements in either case, so making the interface express that "Do you support X" idea vs getting a list of all supported and iterating seemed OK. WDYT?

provider-example/src/hpke.rs
```diff
@@ -0,0 +1,148 @@
...
+            .open(aad, ciphertext)
+            .map_err(general_err)
+    }
+}
+
+fn general_err(err: impl Debug) -> Error {
+    // TODO(@cpu): It would be better to use `Error::Other` here, but it requires `Error` to be
+    //   implemented, and `HpkeError` doesn't.
+    // TODO(@cpu): It would be nicer to user `err.to_string()` here instead of `format!` but
+    //   it requires `Display` to be implemented, and HpkeError doesn't.
```
cpu, line 146: I think this is stale and leftover from when I was using an older version of hpke-rs. I'll adjust.

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,131 @@
...
+///  * KDF: HKDF-SHA256
+///  * AEAD: AES-128-GCM
+///
+/// [RFC 9180]: <https://www.rfc-editor.org/rfc/rfc9180.html>
+/// [HPKE ciphersuite profile]: <https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#section-9>
+pub trait HpkeProvider: Debug + Send + Sync + 'static {
+    /// Create an HPKE instance for the given mode, KEM, KDF and AEAD.
+    ///
+    /// May return an error if the mode, KEM, KDF or AEAD are unsupported by the HPKE provider.
+    fn setup(
```
ctz, line 22: A bit of feedback on this design:

- I suggest we reduce this down to precisely what ECH needs, and no more (that means no need to mention mode apart from commentary).
- I suggest avoiding `setup` here, because that has specific use in HPKE -- `open` / `create` / `new`?
- Maybe wrap the parameter choices in a struct? (`HpkeKem` has an implicit KDF, and `HpkeSymmetricCipherSuite` has an explicit one; I think this doesn't need a further KDF specification).

```rust
pub struct HpkeSuite {
  kem: HpkeKem,
  sym: HpkeSymmetricCipherSuite,
}

pub trait HpkeProvider (...) {
  fn open(&self, suite: HpkeSuite) -> Result<Box<dyn Hpke>, Error>;
  fn supports(&self, suite: HpkeSuite) - >bool;
  // or, alternatively,
  fn supported_suites(&self) -> Vec<HpkeSuite>;
}

pub trait Hpke {
    fn setup_sender(
        &self,
        pk_r: &HpkePublicKey,
        info: &[u8],
    ) -> Result<(EncapsulatedSecret, Box<dyn HpkeSender + '_>), Error>;
    // ...
```

I think it's very unlikely that ECH starts requiring Auth or PSK modes (not sure how that would work protocol-wise!), but if that does happen we could make this interface wider (`open_auth` / `open_psk`, etc.) with distinct return traits.

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,131 @@
...
+        psk_id: Option<&[u8]>,
+        pk_s: Option<&HpkePublicKey>,
+    ) -> Result<Box<dyn HpkeReceiver + '_>, Error>;
+}
+
+/// An HPKE Sender context.
+///
+/// This is a stateful object that can be used to seal messages for receipt by
+/// a receiver.
+pub trait HpkeSender: Debug + Send + Sync {
```
ctz, line 98: I would be minded to start with a one-shot interface for the sake of simplicity. Then move to a streaming interface once everything is up and running?

That would obviate `HpkeSender` and `HpkeReceiver`, and make the methods in `trait Hpke` something like:

```rust
fn open(&self, 
        pk_r: &HpkePublicKey,
        info: &[u8],
        aad; &[u8],
        pt: &[u8],
) -> Result<(EncapsulatedSecret, Vec<u8>), Error>;
```

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Seal the provided `plaintext` with additional data `aad`, returning
+    /// ciphertext.
+    fn seal(&mut self, aad: &[u8], plaintext: &[u8]) -> Result<Vec<u8>, Error>;
```
ctz, line 101: Note most AEADs cannot accept `aad` after the first `plaintext` is processed. It may be more natural to put the `aad` in the `setup_sender` call.

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/// An HPKE Receiver context.
+///
+/// This is a stateful object that can be used to open sealed messages sealed
+/// by a sender.
+pub trait HpkeReceiver: Debug + Send + Sync {
+    /// Open the provided `ciphertext` with additional data `aad`, returning plaintext.
+    fn open(&mut self, aad: &[u8], ciphertext: &[u8]) -> Result<Vec<u8>, Error>;
```
ctz, line 110: This introduces the usual issue with AEADs: "premature release of unauthenticated plaintext" (ie, the decryption cannot tell you whether it was tampered until the end, but by the time you have already received some of the corrupted plaintext). A single-shot API would avoid that.

rustls/src/crypto/hpke.rs
cpu: > I suggest we reduce this down to precisely what ECH needs, and no more (that means no need to mention mode apart from commentary).

SGTM.

> I suggest avoiding setup here, because that has specific use in HPKE -- open / create / new?

:+1: I avoided `new` because Clippy wants `new` to not take `&self`. Open seems reasonable.

> Maybe wrap the parameter choices in a struct? 

I'll give that a shot.

rustls/src/crypto/hpke.rs
cpu: This fell away with the change to focus on base mode.

rustls/src/crypto/hpke.rs
cpu: I've implemented the suggested changes in the latest push.

rustls/src/crypto/hpke.rs
cpu: This fell away with the change to using a one-shot API.

rustls/src/crypto/hpke.rs
cpu: The latest push switched to the single-shot API.

rustls/src/crypto/hpke.rs
cpu: Switched to one-shot.

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,85 @@
...
+///  * KEM: DHKEM(X25519, HKDF-SHA256)
+///  * symmetric ciphersuite:  AES-128-GCM w/ HKDF-SHA256
+///
+/// [RFC 9180]: <https://www.rfc-editor.org/rfc/rfc9180.html>
+/// [HPKE ciphersuite profile]: <https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#section-9>
+pub trait HpkeProvider: Debug + Send + Sync + 'static {
+    /// Create a base mode HPKE instance for the chosen suite.
+    ///
+    /// May return an error if the suite is unsupported by the provider.
+    fn create(&self, suite: &HpkeSuite) -> Result<Box<dyn Hpke>, Error>;
```
djc, line 22: At risk of bikeshedding even more, maybe `start()`? `create()` still feels to constructorish to me.

rustls/src/crypto/hpke.rs
```diff
@@ -0,0 +1,85 @@
...
+        aad: &[u8],
+        ciphertext: &[u8],
+    ) -> Result<Vec<u8>, Error>;
+}
+
+/// An HPKE public key.
+pub struct HpkePublicKey(pub Vec<u8>);
+
+/// An HPKE private key.
+pub struct HpkePrivateKey(pub Vec<u8>);
```
djc, line 74: The private key with public `Vec<u8>` field gives me a bit of pause. Should we wrap some API around this that requires being more intentional about how this gets accessed?

provider-example/src/hpke.rs
```diff
@@ -0,0 +1,97 @@
...
+use rustls::crypto::hpke::{
+    EncapsulatedSecret, Hpke, HpkePrivateKey, HpkeProvider, HpkePublicKey, HpkeSuite,
+};
+use rustls::Error;
+
+pub static HPKE_PROVIDER: &'static dyn HpkeProvider = &HpkeRsHpkeProvider {};
+
+/// A Rustls HPKE provider backed by hpke-rs.
+#[derive(Debug)]
+struct HpkeRsHpkeProvider {}
```
djc, line 16: `HpkeRsHpkeProvider` -- I feel like `HpkeRsProvider` would be `Hpke` enough? (Same for other types with similar naming...)

rustls/src/crypto/hpke.rs
cpu: > Should we wrap some API around this that requires being more intentional about how this gets accessed?

Do you mean something like a `secret_bytes(&self) -> &[u8]` accessor similar to how `pki_types::PrivatePkcs1KeyDer` has `secret_pkcs1_der`? Or did you have something more involved in mind?

provider-example/src/hpke.rs
cpu: Agreed, `HpkeRsHpkeProvider` looks a bit like keyboard barf :laughing: 

Fixed up.

rustls/src/crypto/hpke.rs
cpu: Switched to `start` and adapted the rustdoc comment.

rustls/src/crypto/hpke.rs
cpu: I've made some changes loosely modeled on `pki_types`. Hopefully it's closer to what you had in mind.

djc: Yup, I think this is good!

provider-example/Cargo.toml
```diff
@@ -19,3 +19,11 @@ sha2 = "0.10.0"
 webpki = { package = "rustls-webpki", version = "0.102.0-alpha.1", default-features = false, features = ["alloc", "std"] }
 webpki-roots = "0.26.0-alpha.1"
 x25519-dalek = "2"
+hpke-rs = "0.1.0"
```
djc, line 22: Alphabetize?

djc (approved):

provider-example/Cargo.toml
cpu: Done, thanks.

ctz (approved):

rustls/src/crypto/hpke.rs
cpu: > I would be minded to start with a one-shot interface for the sake of simplicity.

Leaving a note here: we _do_ need the non-one-shot interface for ECH: specifically for handling HelloRetryRequests.

[Section 6.1.5 "Handshaking with ClientHelloInner"](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#section-6.1.5) says:
> It encrypts EncodedClientHelloInner as described in [Section 6.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#encrypting-clienthello), using the second partial ClientHelloOuterAAD, to obtain a second ClientHelloOuter. It reuses the original HPKE encryption context computed in [Section 6.1](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#real-ech) and uses the empty string for enc.
> 
> The HPKE context maintains a sequence number, so this operation internally uses a fresh nonce for each AEAD operation. Reusing the HPKE context avoids an attack described in [Section 10.11.2](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#flow-hrr-hijack).

So we specifically need to maintain the HPKE context used for the original client hello so that we can re-use it when creating the updated extension for a HRR response.

I'm trying to bundle up some required changes and will make sure this is one of them.

PR #1593, opened 2023-11-16
Add wall-time mode for CI benchmarks

aochagavia:
_See updated readme for details (or follow commit by commit)_

I have tested that the wall-time mode results in low noise in the following way:

1. Do 30 benchmark runs, each with 100 iterations per scenario.
2. Use the results to calculate the noise thresholds (for each scenario, we take the _median_ runtime of the 100 iterations, and use that as the result for a given run).
3. Using the calculated noise thresholds, look at the diffs between the runs, and check whether any of them exceeds the noise threshold.
4. Results: no false positives across the 29 diffs 🎉; most noise thresholds below 2%; no thresholds above 5%.

Open questions:

- Should we attempt to distinguish the time spent in the server from the time spent in the client? I'm not sure it would be viable (because of timer resolution), so I haven't tried to implement it (and if we do, I'll need to double-check that it doesn't result in noisy measurements).
- Should we scrape the command to do comparisons once we switch to the bare-metal bench runner? I'm assuming we will want to do that (the current version of the bench runner contains an improved version of the comparison code from the `ci-bench` crate). For that reason, I didn't implement a comparison for two runs of the wall-time benchmarks.

### DIFF
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -62,6 +62,9 @@ const CHANGE_THRESHOLD: f64 = 0.002; // 0.2%
 /// The name of the file where the instruction counts are stored after a `run-all` run
 const ICOUNTS_FILENAME: &str = "icounts.csv";
 
+/// Default size in bytes for internal buffers (256 KB)
+const DEFAULT_BUFFER_SIZE: usize = 262144;
+
 #[derive(Parser)]
 #[command(about)]
 pub struct Cli {
@@ -149,7 +152,7 @@ fn main() -> anyhow::Result<()> {
             let mut stdin = unsafe { File::from_raw_fd(stdin_lock.as_raw_fd()) };
             let mut stdout = unsafe { File::from_raw_fd(stdout_lock.as_raw_fd()) };
 
-            let handshake_buf = &mut [0u8; 262144];
+            let handshake_buf = &mut [0u8; DEFAULT_BUFFER_SIZE];
             let resumption_kind = black_box(bench.kind.resumption_kind());
             let params = black_box(&bench.params);
             let io = StepperIo {
@@ -203,11 +206,13 @@ fn main() -> anyhow::Result<()> {
                     let resumption_kind = black_box(bench.kind.resumption_kind());
                     let params = black_box(&bench.params);
 
-                    let (mut client_writer, mut server_reader) = async_io::async_pipe(262144);
-                    let (mut server_writer, mut client_reader) = async_io::async_pipe(262144);
+                    let (mut client_writer, mut server_reader) =
+                        async_io::async_pipe(DEFAULT_BUFFER_SIZE);
+                    let (mut server_writer, mut client_reader) =
+                        async_io::async_pipe(DEFAULT_BUFFER_SIZE);
 
                     let server_side = async move {
-                        let handshake_buf = &mut [0u8; 262144];
+                        let handshake_buf = &mut [0u8; DEFAULT_BUFFER_SIZE];
                         run_bench(
                             ServerSideStepper {
                                 io: StepperIo {
@@ -223,7 +228,7 @@ fn main() -> anyhow::Result<()> {
                     };
 
                     let client_side = async move {
-                        let handshake_buf = &mut [0u8; 262144];
+                        let handshake_buf = &mut [0u8; DEFAULT_BUFFER_SIZE];
                         run_bench(
                             ClientSideStepper {
                                 io: StepperIo {
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -153,8 +153,7 @@ fn main() -> anyhow::Result<()> {
             let mut stdout = unsafe { File::from_raw_fd(stdout_lock.as_raw_fd()) };
 
             let handshake_buf = &mut [0u8; DEFAULT_BUFFER_SIZE];
-            let resumption_kind = black_box(bench.kind.resumption_kind());
-            let params = black_box(&bench.params);
+            let resumption_kind = bench.kind.resumption_kind();
             let io = StepperIo {
                 reader: &mut stdin,
                 writer: &mut stdout,
@@ -166,7 +165,10 @@ fn main() -> anyhow::Result<()> {
                         run_bench(
                             ServerSideStepper {
                                 io,
-                                config: ServerSideStepper::make_config(params, resumption_kind),
+                                config: ServerSideStepper::make_config(
+                                    &bench.params,
+                                    resumption_kind,
+                                ),
                             },
                             bench.kind,
                         )
@@ -177,7 +179,10 @@ fn main() -> anyhow::Result<()> {
                             ClientSideStepper {
                                 io,
                                 resumption_kind,
-                                config: ClientSideStepper::make_config(params, resumption_kind),
+                                config: ClientSideStepper::make_config(
+                                    &bench.params,
+                                    resumption_kind,
+                                ),
                             },
                             bench.kind,
                         )
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -40,7 +40,7 @@ mod cachegrind;
 mod util;
 
 /// The size in bytes of the plaintext sent in the transfer benchmark
-const TRANSFER_PLAINTEXT_SIZE: usize = 1024 * 1024;
+const TRANSFER_PLAINTEXT_SIZE: usize = 1024 * 1024 * 10; // 10 MB
 
 /// The amount of times a resumed handshake should be executed during benchmarking.
 ///
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -6,6 +6,7 @@ use std::mem;
 use std::os::fd::{AsRawFd, FromRawFd};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::time::Instant;
 
 use anyhow::Context;
 use async_trait::async_trait;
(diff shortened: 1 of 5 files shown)


ci-bench/src/util.rs
```diff
@@ -35,8 +35,358 @@ impl KeyType {
     }
 }
 
+pub mod async_io {
+    //! Async IO building blocks required for sharing code between the instruction count and
+    //! wall-time benchmarks
+
+    use async_trait::async_trait;
```
cpu, line 42: same nit here w.r.t putting the `async_trait` import below the `std` imports with some separation.

ci-bench/src/util.rs
```diff
@@ -35,8 +35,358 @@ impl KeyType {
...
+                Poll::Ready(Ok(()))
+            }
+        }
+    }
+
+    #[cfg(test)]
+    mod test {
+        use super::*;
+
+        fn test_block_on_concurrent(capacity: usize) {
```
cpu, line 325: This should go below the `#[test]` fns that use it to match the top-down ordering style used in the crate.

ci-bench/src/main.rs
```diff
@@ -77,6 +78,11 @@ pub enum Command {
     },
     /// Run a single benchmark at the provided index (used by the bench runner to start each benchmark in its own process)
     RunSingle { index: u32, side: Side },
+    /// Run all benchmarks in walltime mode, each one for 100 times, and print the measured timings in CSV format
```
cpu, line 81: "each one for 100 times" - I think this part of the comment might have fallen out of sync with the implementation, it's `iterations_per_scenario` times each right? Or have I misunderstood?

ci-bench/src/main.rs
```diff
@@ -183,6 +189,70 @@ fn main() -> anyhow::Result<()> {
...
+        Command::Walltime {
+            iterations_per_scenario,
+        } => {
+            let mut timings = vec![Vec::with_capacity(iterations_per_scenario); benchmarks.len()];
+            for _ in 0..iterations_per_scenario {
+                for (i, bench) in benchmarks.iter().enumerate() {
+                    let start = Instant::now();
+
+                    let resumption_kind = black_box(bench.kind.resumption_kind());
+                    let params = black_box(&bench.params);
```
cpu, line 200: How come these get a `black_box` hint? I would expect `run_bench` would benefit from that but I'm not sure I understand why these two invocations are used. 

Can you add a comment to help me understand?

ci-bench/src/main.rs
```diff
@@ -183,6 +189,70 @@ fn main() -> anyhow::Result<()> {
...
+
+                    let (mut client_writer, mut server_reader) = async_io::async_pipe(262144);
+                    let (mut server_writer, mut client_reader) = async_io::async_pipe(262144);
```
cpu, line 203: Is this constant value meaningful?

It gets duplicated in a few places (L203, L204, L206, L222). Maybe worth hoisting into a `const`?

ci-bench/src/main.rs
```diff
@@ -150,31 +157,101 @@ fn main() -> anyhow::Result<()> {
...
+                            bench.kind,
+                        )
+                        .await
+                    };
+
+                    async_io::block_on_concurrent(client_side, server_side)
+                        .with_context(|| format!("{} crashed", bench.name()))?;
+
+                    let duration = Instant::now() - start;
+                    timings[i].push(duration);
```
cpu, line 242: I think this can be simplified with `elapsed()`:
```suggestion
                    timings[i].push(start.elapsed());
```

ci-bench/src/util.rs
```diff
@@ -114,6 +172,217 @@ pub mod async_io {
...
```
cpu, line 325: Nit: move this below the tests that use it.

cpu (commented):
Very cool!

My familiarity with Async Rust is pretty low, so while I've reviewed this branch as best I can it might be valuable to get input from someone with a strong grasp of the fundamentals.

ci-bench/src/main.rs
```diff
@@ -26,11 +27,12 @@ use crate::benchmark::{
     ResumptionKind,
 };
 use crate::cachegrind::CachegrindRunner;
+use crate::util::async_io::{AsyncRead, AsyncWrite};
 use crate::util::transport::{
     read_handshake_message, read_plaintext_to_end_bounded, send_handshake_message,
     write_all_plaintext_bounded,
 };
-use crate::util::KeyType;
+use crate::util::{async_io, KeyType};
```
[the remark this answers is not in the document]
aochagavia: Thanks, I'll have to be more careful with letting my IDE auto-import stuff!

ci-bench/src/util.rs
```diff
@@ -35,8 +35,358 @@ impl KeyType {
...
+    ) -> anyhow::Result<()> {
+        // We don't need a waker, because the future will complete in one go
+        let waker = noop_waker();
+        let mut ctx = task::Context::from_waker(&waker);
+
+        let future = pin!(future);
+        match future.poll(&mut ctx) {
+            Poll::Ready(result) => result,
+            Poll::Pending => {
+                panic!("future unexpectedly pending!")
```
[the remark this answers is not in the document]
aochagavia: I'm a bit torn on this one... The panic here signals that the caller violated the function contract by providing a future that does not finish in one poll. However, since this is all internal code and we control the calling code, this match arm should indeed be unreachable. I'd rather keep the panic and use a more descriptive error message, like `the provided future did not finish after one poll`. What do you think?

ci-bench/src/util.rs
```diff
@@ -35,8 +35,358 @@ impl KeyType {
...
+            }
+        }
+    }
+
+    /// Block on two futures that are run concurrently.
```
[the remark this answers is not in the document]
aochagavia: Hmmm... That set me thinking. It probably makes more sense to return both errors so the caller knows which side of the connection went wrong. I'll push a change for that and document it.

ci-bench/src/util.rs
```diff
@@ -35,8 +35,358 @@ impl KeyType {
...
+
+        match (x_output, y_output) {
+            (Some(x_output), Some(y_output)) => x_output.or(y_output),
+            _ => panic!("at least one of the futures seems to be stuck"),
+        }
+    }
+
+    // Copied from Waker::noop, which we cannot use directly because it hasn't been stabilized
+    fn noop_waker() -> Waker {
+        const VTABLE: RawWakerVTable = RawWakerVTable::new(|_| RAW, |_| {}, |_| {}, |_| {});
```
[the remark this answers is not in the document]
aochagavia: We should send it as an entry for a code beauty contest 😉

ci-bench/src/main.rs
aochagavia: Sorry, the comment fell indeed out of sync! Thanks for pointing it out

ci-bench/src/main.rs
aochagavia: TIL, thanks!

ci-bench/src/main.rs
aochagavia: I must say I find it difficult to get my thoughts straight on this one, but I wrote a comment that hopefully clarifies things. Otherwise, let me know!

ci-bench/src/main.rs
aochagavia: I took the constant over from the original benchmarks (`rustls/examples/internal/bench.rs`) and never really thought too much about it. I don't think there is a specific meaning to it, but I'll pull it into a `DEFAULT_BUFFER_SIZE` const.

ci-bench/src/util.rs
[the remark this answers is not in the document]
cpu: >  I'd rather keep the panic and use a more descriptive error message, like the provided future did not finish after one poll. What do you think?
 
That sounds good to me! I don't feel strongly about the suggestion.

ci-bench/src/main.rs
cpu: I think I understand better now, thank you.

Apologies, but I didn't notice that there were equivalent `black_box!` invocations earlier in the code on L156 and L157. Maybe it would be better to put the comment there and then reference it here?

aochagavia: The funny thing is... I don't think we actually need the black box on L156 and L157, because in that case the server and the client run in different processes, so each side of the connection has no compile-time information about the other side. Just pushed a commit to remove them.

ctz (approved):
this has quite a bit more gnarly async IO than I was expecting, and i'm not entirely confident in reviewing that. but otherwise looks alright to me and I'm generally happy to say "the ends justify the means" for test code :)

aochagavia:
> this has quite a bit more gnarly async IO than I was expecting, and i'm not entirely confident in reviewing that. but otherwise looks alright to me and I'm generally happy to say "the ends justify the means" for test code :)

@cpu @ctz Is there anything I could do to help clarify the async part of this PR? For instance, I can write something down, telling how I arrived at the current approach and explaining how the async building blocks of this specific PR are tied together. It might end up becoming a small blog article.

Other than that, I agree with @ctz that (ab)using async here is a bit gnarly, but that it is acceptable for test code.

cpu:
>  I can write something down, telling how I arrived at the current approach and explaining how the async building blocks of this specific PR are tied together. It might end up becoming a small blog article.

I would be interested in reading that for my own edification but I don't think it needs to be a blocker for merging this :-)

I'm just generally very inexperienced with writing Async Rust and have mostly done it using a big batteries-included runtime that took care of the details. This work is closer to the underlying primitives than I'm used to and so my confidence in review was lower than it might have been otherwise. 

> Other than that, I agree with @ctz that (ab)using async here is a bit gnarly, but that it is acceptable for test code.

I concur. I think for the situation at hand it's reasonable and my confidence will improve over time. I vote we merge and iterate as needed.

djc:
I feel like there should be some high-level explanation of why we need async here and why it's designed the way it is, maybe in the README? I think this could be regurgitating some of the discussion from the issue (which doesn't appear to be linked from this PR yet).

aochagavia:
@djc just updated the readme to be a bit more detailed about async/await and the code sharing approach, let me know if additional clarification is necessary 😉

PR #1594, opened 2023-11-17
Remove `quic` crate feature

### LINKED ISSUE #1437: reducing Cargo feature list

### LINKED ISSUE #1472: Crypto provider feature guards are not additive

### LINKED ISSUE #1491: Expose more QUIC mods (and friends)
Although RustCrypto doesn't seems to have QUIC support yet, we should expose the relevant mods so that we at least have a chance to experiment for an implementation.

Currently this includes but not limited to: `rustls::quic`, `rustls::hkdf`.

ctz:
fixes #1491
fixes #1437 
fixes #1472


ctz:
I think the coverage results are valid; they are quic-specific branches that were previously compiled out of the coverage tests.

rustls/src/quic.rs
```diff
@@ -2,12 +2,16 @@
 use crate::client::{ClientConfig, ClientConnectionData, ServerName};
 use crate::common_state::{CommonState, Protocol, Side};
 use crate::conn::{ConnectionCore, SideData};
-use crate::crypto::tls13::{HkdfExpander, OkmBlock};
+use crate::crypto::cipher::{AeadKey, Iv};
+use crate::crypto::tls13::HkdfExpander;
+use crate::crypto::tls13::OkmBlock;
```
cpu, line 7:
```suggestion
use crate::crypto::tls13::{HkdfExpander, OkmBlock};
```

rustls/src/crypto/ring/quic.rs
```diff
@@ -252,12 +233,16 @@ mod tests {
         let expander = TLS13_CHACHA20_POLY1305_SHA256_INTERNAL
             .hkdf_provider
             .expander_for_okm(&OkmBlock::new(SECRET));
-        let hpk =
-            super::HeaderProtectionKey::new(expander.as_ref(), version, &aead::quic::CHACHA20);
+        let header_key =
+            hkdf_expand_label_aead_key(expander.as_ref(), 32, version.header_key_label(), &[]);
+        let packet_key =
+            hkdf_expand_label_aead_key(expander.as_ref(), 32, version.packet_key_label(), &[]);
```
cpu, line 239: I think you could knock out the magic 32 constant with a binding that takes advantage of the new `aead_key_len` fn on the trait like:
```
let key_len = TLS13_CHACHA20_POLY1305_SHA256_INTERNAL
            .quic
            .unwrap()
            .aead_key_len();
```

connect-tests/Cargo.toml
```diff
@@ -7,7 +7,6 @@ description = "Rustls connectivity based integration tests."
 publish = false
 
 [features]
```
cpu, line 9: nit: maybe drop this empty `[features]` section?

examples/Cargo.toml
```diff
@@ -7,7 +7,6 @@ description = "Rustls example code and tests."
 publish = false
 
 [features]
```
cpu, line 9: ditto RE: empty features.

rustls/examples/internal/bogo_shim.rs
```diff
@@ -2,1293 +2,11 @@
 // test suite. See bogo/ for this in action.
 //
 // https://boringssl.googlesource.com/boringssl/+/master/ssl/test
-//
```
cpu, line 5: Would it be feasible to separate the `bogo_shim.rs` -> `bogo_shim_impl.rs` move in its own commit ahead of this work, maintaining the feature gates, and then removing those gates in this commit?

rustls/examples/internal/bogo_shim.rs
```diff
@@ -6,6 +6,11 @@
 #[cfg(any(feature = "ring", feature = "aws_lc_rs"))]
 mod bogo_shim_impl;
 
+#[cfg(any(feature = "ring", feature = "aws_lc_rs"))]
+fn main() {
+    bogo_shim_impl::main();
+}
```
cpu: Should this have been in the previous commit where the move was done and the quic feature removed? :thinking:

rustls/examples/internal/bogo_shim_impl.rs
```diff
@@ -930,7 +930,7 @@ fn exec(opts: &Options, mut sess: Connection, count: usize) {
     }
 }
 
-fn main() {
+pub fn main() {
```
cpu, line 970: ditto RE: wondering if this would make sense in the previous commit

rustls/src/suites.rs
```diff
@@ -18,6 +19,22 @@ pub struct CipherSuiteCommon {
 
     /// Which hash function the suite uses.
     pub hash_provider: &'static dyn crypto::hash::Hash,
+
+    /// The number of messages that can be encrypted by a single
+    /// instance of `MessageEncrypter` produced for this suite before
+    /// an attacker gains an advantage in distinguishing it from an ideal PRP.
```
cpu, line 25:
```suggestion
    /// an attacker gains an advantage in distinguishing it from an ideal pseudorandom permutation (PRP).
```

cpu (commented):
cool!

rustls/examples/internal/bogo_shim.rs
ctz: Yes, will do that.

connect-tests/Cargo.toml
ctz: Deleted.

examples/Cargo.toml
ctz: Deleted.

rustls/src/suites.rs
ctz: Thanks -- again with the jargon!

cpu (approved):
:rocket:

rustls/src/quic.rs
```diff
@@ -561,16 +564,26 @@ impl AsRef<[u8]> for Tag {
 }
 
 /// How a `Tls13CipherSuite` generates `PacketKey`s and `HeaderProtectionKey`s.
-pub(crate) trait Algorithm: Send + Sync {
+pub trait Algorithm: Send + Sync {
+    /// Produce a `PacketKey` encrypter/decrypter for this suite.
+    ///
+    /// `suite` is the entire suite this `Algorithm` appeared in.
+    /// `key` and `iv` is the key material to use.
```
djc, line 583: Given that the fields of `AeadKey` are private, I'm not sure this trait is practically implementable for foreign types?

ctz: The expected things for people to do with an `AeadKey` is:

- view the raw bytes, via `impl AsRef<[u8]>`
- move or drop it (has a non-trivial zeroizing `impl Drop`)

I don't think there is any need for construction of these outside the crate.

djc: Ahh, yes -- sorry for my confusion.

PR #1595, opened 2023-11-17
`MessageDeframer`: batch discard operations

japaric:
at @djc 's suggestion I have taken the `MessageDeframer` commits in #1583 into their own PR

I have also sneaked in an extra commit that batches `discard` operations (which may produce slight perf improvements in the `Connection` API) that makes the `MessageDeframer` operate on slices and it's part of the ongoing work on making decoding / decryption non-allocating / "in-place".


fuzz/fuzzers/deframer.rs
```diff
@@ -9,17 +9,18 @@ use rustls::internal::record_layer::RecordLayer;
...
     let mut dfm = deframer::MessageDeframer::default();
     if dfm
-        .read(&mut io::Cursor::new(data))
+        .read(&mut io::Cursor::new(data), &mut buf)
         .is_err()
     {
         return;
     }
-    dfm.has_pending();
+    buf.has_pending();
```
cpu, line 20: This was in the original code too, but I'm not sure why? Doesn't this end up having no effect? Should there be an `assert!` on the result? :thinking:

cpu (approved):
Thanks for breaking this out :+1:

rustls/src/msgs/deframer.rs
```diff
@@ -299,31 +314,54 @@ impl MessageDeframer {
...
+    pub fn has_pending(&self) -> bool {
+        !self.buffer.is_empty()
+    }
+}
+
+#[derive(Default)]
+struct DeframerBuffer {
+    /// Buffer of data read from the socket, in the process of being parsed into messages.
+    ///
+    /// For buffer size management, checkout out the `read()` method.
```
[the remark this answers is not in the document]
japaric: updated this comment to point to `prepare_read` which seems like the most relevant with respect to resizing the buffer

rustls/src/msgs/deframer.rs
```diff
@@ -299,31 +314,54 @@ impl MessageDeframer {
...
 
+    /// Returns true if we have messages for the caller
+    /// to process, either whole messages in our output
+    /// queue or partial messages in our buffer.
+    pub fn has_pending(&self) -> bool {
+        !self.buffer.is_empty()
+    }
+}
+
+#[derive(Default)]
```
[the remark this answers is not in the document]
japaric: literally add `#[derive(Debug)]`? or is there some helper to make the `Debug` impl more human-friendly, like hexadecimal representation?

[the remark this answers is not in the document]
cpu: > literally add #[derive(Debug)]?

That's what I had in mind :+1:

rustls/src/msgs/deframer.rs
```diff
@@ -271,8 +278,16 @@ impl MessageDeframer {
...
+                let buf = if quic {
+                    self.buffer.unfilled()
+                } else {
+                    self.buffer.filled_mut()
+                };
+                let dst = &mut buf[..payload.len()];
                 dst.copy_from_slice(payload);
+                if quic {
+                    self.buffer.advance(payload.len());
+                }
```
djc: I'd suggest that all of this should probably become a single method on `DeframerBuffer`.

rustls/src/msgs/deframer.rs
```diff
@@ -253,15 +252,23 @@ impl MessageDeframer {
...
+                let buf = if quic {
+                    self.buffer.unfilled()
+                } else {
+                    self.buffer.filled_mut()
+                };
+                let dst = &mut buf[meta.payload.end..meta.payload.end + payload.len()];
                 dst.copy_from_slice(payload);
+                if quic {
+                    self.buffer.advance(payload.len());
+                }
```
djc: This should probably be a single `DeframerBuffer` method.

rustls/src/msgs/deframer.rs
```diff
@@ -184,7 +178,9 @@ impl MessageDeframer {
         let message = PlainMessage {
             typ: ContentType::Handshake,
             version: meta.version,
-            payload: Payload::new(&self.buf[meta.payload.start..meta.payload.start + expected_len]),
+            payload: Payload::new(
+                &self.buffer.filled()[meta.payload.start..meta.payload.start + expected_len],
```
djc: Maybe `filled()` should take a `Range` directly?

rustls/src/msgs/deframer.rs
```diff
@@ -524,6 +528,41 @@ mod tests {
         include_bytes!("../testdata/deframer-invalid-version.bin");
     const INVALID_LENGTH_MESSAGE: &[u8] = include_bytes!("../testdata/deframer-invalid-length.bin");
 
+    // buffered version to ease testing
+    #[derive(Default)]
+    struct MessageDeframer {
```
djc, line 585: Maybe give this a different name, like `BufferedDeframer`? Also, probably makes sense to move `input_bytes()` and `input_bytes_concat()` into to this type's `impl`.

rustls/src/msgs/deframer.rs
```diff
@@ -335,7 +335,7 @@ impl MessageDeframer {
 }
 
 #[derive(Default)]
-pub struct DeframerBuffer {
+pub struct DeframerVecBuffer {
```
djc, line 338: If we're going to rename this, let's just introduce it under this name in the first place.

japaric:
@djc I've addressed your comments in new commits. I'll squash when you are all happy with the final state.

japaric:
@djc anything else you would like to see addressed here or I should I go ahead and squash the commits?

djc:
I prefer reviewing the squashed changes anyway. I'm pretty busy this week and focusing my rustls time on things that are blocking the 0.22 release so might take me a bit to get to this.

japaric:
squashed

japaric:
updated to not use the `<_>::default()` syntax as per PR #1620

rustls/src/msgs/deframer.rs
```diff
@@ -533,21 +531,52 @@ mod tests {
         include_bytes!("../testdata/deframer-invalid-version.bin");
     const INVALID_LENGTH_MESSAGE: &[u8] = include_bytes!("../testdata/deframer-invalid-length.bin");
 
-    fn input_bytes(d: &mut MessageDeframer, bytes: &[u8]) -> io::Result<usize> {
-        let mut rd = io::Cursor::new(bytes);
-        d.read(&mut rd)
+    // buffered version to ease testing
+    #[derive(Default)]
+    struct BufferedDeframer {
+        inner: super::MessageDeframer,
```
djc, line 588: Let's just do `use super::*` at the top of the `tests` module.

rustls/src/msgs/deframer.rs
```diff
@@ -440,6 +399,54 @@ impl DeframerVecBuffer {
     }
 }
 
+trait DeframerBuffer {
+    fn advance(&mut self, num_bytes: usize);
```
djc, line 429: I think we should keep empty lines between these even if we're just declaring the interfaces. I think `copy()` should also probably get a docstring.

(Please maintain order as previously explained.)

rustls/src/conn.rs
```diff
@@ -442,11 +442,15 @@ impl<Data> ConnectionCommon<Data> {
     /// This is a shortcut to the `process_new_packets()` -> `process_msg()` ->
     /// `process_handshake_messages()` path, specialized for the first handshake message.
     pub(crate) fn first_handshake_message(&mut self) -> Result<Option<Message>, Error> {
-        match self
-            .core
-            .deframe(&mut self.deframer_buffer)?
-            .map(Message::try_from)
-        {
+        let mut to_discard = 0;
```
djc, line 445: It looks like this would be simpler if `to_discard` (potentially with a different name) lived within the `MessageDeframer` and was reset by a call to a `discarded(&mut self) -> usize` like method.

rustls/src/msgs/deframer.rs
```diff
@@ -338,6 +328,13 @@ struct DeframerVecBuffer {
 }
 
 impl DeframerVecBuffer {
+    /// Returns true if we have messages for the caller
+    /// to process, either whole messages in our output
+    /// queue or partial messages in our buffer.
+    pub fn has_pending(&self) -> bool {
```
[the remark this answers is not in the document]
japaric: > This should drift to the bottom of the impl block. 

shouldn't this stay at the top as per [the style guide](https://github.com/rustls/rustls/blob/078f03334b2de9d8e464755d8836318b9a221346/CONTRIBUTING.md#ordering-associated-functions-within-an-inherent-impl-block) because it's public API?

> Please use the full 100-char width for doc strings. Ideally the first line stands on it own as a concise description.

ok but note that this an existing doc string that I have simply moved around.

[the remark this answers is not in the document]
djc: > shouldn't this stay at the top as per [the style guide](https://github.com/rustls/rustls/blob/078f03334b2de9d8e464755d8836318b9a221346/CONTRIBUTING.md#ordering-associated-functions-within-an-inherent-impl-block) because it's public API?

Err, yes -- sorry for the noise.

> > Please use the full 100-char width for doc strings. Ideally the first line stands on it own as a concise description.
> 
> ok but note that this an existing doc string that I have simply moved around.

Okay, that would have been more obvious if it wasn't part of a large commit with a bunch of other stuff.

japaric:
addressed the review comments

ctz:
> my only remaining question is whether we want to land this prior to cutting 0.22? looks like it gives a small but positive performance improvement -- but is the risk/reward there?

0.22 is out, so this concern is gone. apologies for how long that took; i hope we can make quick progress on these outstanding PRs now

ctz:
Both complaints from cargo-semver-checks are somewhat irrelevent; they are changes to types only available under `rustls::internals`.

rustls/src/msgs/deframer.rs
```diff
@@ -455,6 +489,24 @@ impl DeframerBuffer for DeframerVecBuffer {
     }
 }
 
+impl DeframerBuffer for DeframerSliceBuffer<'_> {
```
djc, line 492: This impl should live below the `DeframerSliceBuffer`'s inherent `impl`, and above the `DeframerBuffer` defintion.

rustls/src/msgs/deframer.rs
```diff
@@ -397,6 +403,34 @@ impl DeframerVecBuffer {
     }
 }
 
+/// A borrowed version of [`DeframerVecBuffer`] that tracks discard operations
+pub struct DeframerSliceBuffer<'a> {
```
djc, line 407: I find it confusing that this gets introduced as part of a commit that describes itself as "batch discard operations", which seems not very related? Can we split these two things in separate commits?

rustls/src/msgs/deframer.rs
```diff
@@ -455,6 +489,24 @@ impl DeframerBuffer for DeframerVecBuffer {
     }
 }
 
+impl DeframerBuffer for DeframerSliceBuffer<'_> {
+    fn advance(&mut self, _num_bytes: usize) {
+        unimplemented!("operation not supported by this type of buffer")
```
djc, line 494: If we're going to have this, I'd at least like some comments here to describe how we can statically be certain that these won't be invoked.

fuzz/fuzzers/deframer.rs
ctz: Yes, this is attempting to verify that this function is panic-free, and expose any branches to `panic!` inside to the coverage-driven fuzzer.

rustls/src/msgs/deframer.rs
pvdrz: well... we could split `DeframerBuffer` into two traits: One with `filled` and `filled_mut` and another one with `advance` and `unfilled`. Then `DeframerVecBuffer` would implement both but `DeframerSliceBuffer` would only implement the first. Would that work better for you?

rustls/src/msgs/deframer.rs
```diff
@@ -751,15 +779,148 @@ mod tests {
...
+                *b = i as u8;
+            }
+
+            let error = self.error.take().unwrap();
+            Err(error)
+        }
+    }
+
+    fn assert_len(want: usize, got: io::Result<usize>) {
+        if let Ok(gotval) = got {
```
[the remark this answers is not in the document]
pvdrz: done

rustls/src/msgs/deframer.rs
pvdrz: fixed

fuzz/fuzzers/deframer.rs
cpu: Gotcha, :+1:

rustls/src/msgs/deframer.rs
pvdrz: You're right, that wouldn't be as straightforward.

Another idea I just came up with is to add a const generic parameter to `DeframerBuffer` so it becomes `DeframerBuffer<const QUIC: bool>`.

Then `DeframerSliceBuffer` only implements `DeframerBuffer<false>` and `DeframerVecBuffer` implements `DeframerBuffer<QUIC>` for any value of `QUIC`.

This `QUIC` parameters gets propagated into `MessageDeframer::append_hs` and then we can statically guarantee that `<DeframerSliceBuffer as DeframerBuffer<true>>::copy` is never used as it is not implemented.

ctz: I think that looks OK -- WDYT @djc?

Eventually I think it would be sensible to refactor the deframer to extract just the parts that QUIC needs to use. Right now a QUIC connection has a full deframer, but only uses the handshake-specific parts of it. That should also make this fall away for TCP-TLS uses.

rustls/src/msgs/deframer.rs
```diff
@@ -855,8 +855,13 @@ mod tests {
             record_layer: &mut RecordLayer,
             negotiated_version: Option<ProtocolVersion>,
         ) -> Result<Option<Deframed>, Error> {
-            self.inner
-                .pop(record_layer, negotiated_version, &mut self.buffer)
+            let mut deframer_buffer = self.buffer.borrow();
```
djc, line 860: Same here with the discard vs error returns.

rustls/src/msgs/deframer.rs
```diff
@@ -431,50 +461,40 @@ impl<'a> DeframerSliceBuffer<'a> {
...
-    }
+    /// If `QUIC` is false the data will be copied into the filled section of the buffer
+    fn copy(&mut self, src: &[u8], at: usize);
+}
 
-    fn advance(&mut self, num_bytes: usize);
+fn copy_into_buffer(buf: &mut [u8], src: &[u8], at: usize) {
+    let len = src.len();
+    let into = &mut buf[at..at + len];
+    into.copy_from_slice(src);
```
djc, line 492: Do we need the bindings for some borrowck edge cases here? If not, would be nice to just inline the oneliner.

rustls/src/msgs/deframer.rs
djc: This is a pretty neat solution, yes. Would definitely be interesting to see if we can use a substantially simplified deframer for the QUIC case.

djc (approved):
Would be good IMO to squash the final commit into the commit that originally introduced the `DeframerBuffer` trait.

rustls/src/msgs/deframer.rs
pvdrz: same answer as the previous one

rustls/src/msgs/deframer.rs
pvdrz: fixed

pvdrz:
> Would be good IMO to squash the final commit into the commit that originally introduced the `DeframerBuffer` trait.

done

cpu (approved):
Looks great, thanks for your attention to the commit history. The `DeframerBuffer` const generic approach was a cool idea :+1:

rustls/src/conn.rs
```diff
@@ -452,11 +452,12 @@ impl<Data> ConnectionCommon<Data> {
...
-            .core
-            .deframe(&mut self.deframer_buffer)?
-            .map(Message::try_from)
-        {
+        let mut deframer_buffer = self.deframer_buffer.borrow();
+        let res = self.core.deframe(&mut deframer_buffer);
+        let discard = deframer_buffer.pending_discard();
+        self.deframer_buffer.discard(discard);
+
+        match res?.map(Message::try_from) {
```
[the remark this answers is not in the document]
ctz: I think the best motivation for this sort of code are tests that demonstrate how things go wrong if one were to (for example) only do the discarding in the happy path. At a guess, the error fusing inside the deframer will make the distinction not observable at the public API level.

Tomorrow I think I'll write some extra tests for this; on a purely selfish level that is the way I find most effective to understand code -- reading doesn't do a lot for me.

PR #1474, opened 2023-09-14
add an example tls server with provider (again)

stevefan1999-personal:
Per https://github.com/rustls/rustls/pull/1405#issuecomment-1717905136, but this time I just added the minimal requirements. More specifically I removed the generic code and flattened it out to support P256 only. I have a more comprehensive provider at https://github.com/stevefan1999-personal/rustls-provider-rustcrypto that uses generics to eliminate boilerplate code.

PR #1596, opened 2023-11-17
add server example for example provider

ctz:
this is a rebase, updating for changes to the provider APIs, and a couple of clippy fixes for #1474


provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
+use pki_types::CertificateDer;
+use pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer};
+use rustls::server::Acceptor;
+use rustls::ServerConfig;
+use rustls_provider_example::PROVIDER;
+use std::io::Write;
+use std::sync::Arc;
```
cpu, line 7:
```suggestion
use std::io::Write;
use std::sync::Arc;

use pki_types::{CertificateDer, PrivateKeyDer, PrivatePkcs8KeyDer};
use rustls::server::Acceptor;
use rustls::ServerConfig;

use rustls_provider_example::PROVIDER;
```

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+    server_key_der: PrivateKeyDer<'static>,
+}
+
+impl TestPki {
+    fn new() -> Self {
+        let alg = &rcgen::PKCS_ECDSA_P256_SHA256;
+        let mut ca_params = rcgen::CertificateParams::new(Vec::new());
+        ca_params
+            .distinguished_name
+            .push(rcgen::DnType::OrganizationName, "Rustls Server Acceptor");
```
cpu, line 20:
```suggestion
            .push(rcgen::DnType::OrganizationName, "Provider Server Example");
```

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+        ca_params
+            .distinguished_name
+            .push(rcgen::DnType::CommonName, "Example CA");
+        ca_params.is_ca = rcgen::IsCa::Ca(rcgen::BasicConstraints::Unconstrained);
+        ca_params.key_usages = vec![
+            rcgen::KeyUsagePurpose::KeyCertSign,
+            rcgen::KeyUsagePurpose::DigitalSignature,
+            rcgen::KeyUsagePurpose::CrlSign,
```
cpu, line 28: Harmless to leave this, but if we're going to duplicate the code from the `server_acceptor.rs` example we might as well trim it down to what this example needs:
```suggestion
```

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+use rustls::server::Acceptor;
+use rustls::ServerConfig;
+use rustls_provider_example::PROVIDER;
+use std::io::Write;
+use std::sync::Arc;
+
+struct TestPki {
+    server_cert_der: CertificateDer<'static>,
+    server_key_der: PrivateKeyDer<'static>,
+}
```
cpu, line 12: As future work it might be nice to pull out the commonalities between this rcgen setup code and the code in `server_acceptor.rs`. They're quite similar.

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+            .with_single_cert(vec![self.server_cert_der], self.server_key_der)
+            .unwrap();
+
+        server_config.key_log = Arc::new(rustls::KeyLogFile::new());
+
+        Arc::new(server_config)
+    }
+}
+
+fn main() {
```
cpu, line 65: I think for top-down ordering `main` should be above the `TestPki` helper code.

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+    env_logger::init();
+
+    let pki = TestPki::new();
+    let server_config = pki.server_config();
+
+    let listener = std::net::TcpListener::bind(format!("0.0.0.0:{}", 4443)).unwrap();
```
cpu, line 71: Might as well make this workable w/ IPv6
```suggestion
    let listener = std::net::TcpListener::bind(format!("[::]:{}", 4443)).unwrap();
```

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+                    "Connection: Closed\r\n",
+                    "Content-Type: text/html\r\n",
+                    "\r\n",
+                    "<h1>Hello World!</h1>\r\n"
+                )
+                .as_bytes();
+
+                let _ = conn.writer().write(msg);
+                let _ = conn.write_tls(&mut stream);
+                _ = conn.complete_io(&mut stream);
```
cpu, line 96: since we're unwrapping elsewhere, maybe do the same here?
```suggestion
                conn.writer().write(msg).unwrap();
                conn.write_tls(&mut stream).unwrap();
                conn.complete_io(&mut stream).unwrap();
```

provider-example/examples/server.rs
```diff
@@ -0,0 +1,107 @@
...
+
+                conn.send_close_notify();
+                let _ = conn.write_tls(&mut stream);
+                _ = conn.complete_io(&mut stream);
```
cpu, line 100:
```suggestion
                conn.write_tls(&mut stream).unwrap();
                conn.complete_io(&mut stream).unwrap();
```

provider-example/src/lib.rs
```diff
@@ -6,6 +6,7 @@ mod hash;
 mod hmac;
 mod hpke;
 mod kx;
+pub mod sign;
```
cpu, line 9: I'm not sure I see the reason this needs to be `pub`.
```suggestion
mod sign;
```

provider-example/src/sign.rs
```diff
@@ -0,0 +1 @@
+pub mod ecdsa;
```
cpu, line 1: It seems a bit overkill to make a directory for the `sign` module. Maybe just inline the `ecdsa` mod here?

provider-example/src/sign/ecdsa.rs
```diff
@@ -0,0 +1,58 @@
+use pkcs8::DecodePrivateKey;
+use pki_types::PrivateKeyDer;
+use rustls::{
+    sign::{Signer, SigningKey},
+    SignatureAlgorithm, SignatureScheme,
+};
+use signature::RandomizedSigner;
+use signature::SignatureEncoding;
+use std::sync::Arc;
```
cpu, line 9:
```suggestion
use std::sync::Arc;

use pkcs8::DecodePrivateKey;
use pki_types::PrivateKeyDer;
use rustls::sign::{Signer, SigningKey};
use rustls::{SignatureAlgorithm, SignatureScheme};
use signature::RandomizedSigner;
use signature::SignatureEncoding;

```

provider-example/src/sign/ecdsa.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    fn try_from(value: PrivateKeyDer<'_>) -> Result<Self, Self::Error> {
+        match value {
+            PrivateKeyDer::Pkcs8(der) => {
+                p256::ecdsa::SigningKey::from_pkcs8_der(der.secret_pkcs8_der()).map(|kp| Self {
+                    key: Arc::new(kp),
+                    scheme: SignatureScheme::ECDSA_NISTP256_SHA256,
+                })
+            }
+            _ => todo!(),
```
cpu, line 28: Maybe something like:
```suggestion
            _ => panic!("unsupported private key format"),
```

provider-example/src/sign/ecdsa.rs
```diff
@@ -0,0 +1,58 @@
...
+    scheme: SignatureScheme,
+}
+
+impl TryFrom<PrivateKeyDer<'_>> for EcdsaSigningKeyP256 {
+    type Error = pkcs8::Error;
+
+    fn try_from(value: PrivateKeyDer<'_>) -> Result<Self, Self::Error> {
+        match value {
+            PrivateKeyDer::Pkcs8(der) => {
+                p256::ecdsa::SigningKey::from_pkcs8_der(der.secret_pkcs8_der()).map(|kp| Self {
```
cpu, line 23: unnecessary qualification:
```suggestion
                ecdsa::SigningKey::from_pkcs8_der(der.secret_pkcs8_der()).map(|kp| Self {
```

provider-example/Cargo.toml
```diff
@@ -9,21 +9,26 @@ publish = false
...
 hpke-rs-crypto = "0.1.2"
 hpke-rs-rust-crypto = "0.1.2"
+p256 = "0.13.2"
+pkcs8 = { version = "0.10.2", features = ["std"] }
 pki-types = { package = "rustls-pki-types", version = "0.2.0" }
 rand_core = "0.6.0"
 rustls = { path = "../rustls", default-features = false, features = ["logging", "tls12"] }
 rsa = { version = "0.9.0", features = ["sha2"] }
 sha2 = "0.10.0"
+signature = "2.1.0"
```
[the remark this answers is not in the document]
ctz: This should be the earliest version that provides the APIs we use; not the latest. So I think `"2"` is the right version.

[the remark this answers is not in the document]
ctz: Though it seems the minimal-versions check doesn't cover this example code, and doesn't work. Will follow up with a separate PR for that.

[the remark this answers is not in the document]
cpu: You're right, I was thinking in terms of a `0.2.0` vs `0.1.0`. Sorry for the noise :+1:

provider-example/src/sign/ecdsa.rs
ctz: I think this is necessary: `p256` is the name of a crate so this refers to: https://docs.rs/p256/latest/p256/ecdsa/type.SigningKey.html

provider-example/examples/server.rs
ctz: Will do. Also will add a comment suggesting this is a bad idea in "real" code.

provider-example/examples/server.rs
ctz: Indeed, ditto `simpleclient.rs`.

provider-example/examples/server.rs
ctz: Have had rustfmt do this.

provider-example/examples/server.rs
ctz: Though when I original did that duplication, I couldn't find a way to make it fly without making the most simple example look odd or magical.

cpu: Feels fair to handle separately on a rainy day.. I'll resolve this feedback.

provider-example/src/sign/ecdsa.rs
cpu: Hm, it seems to compile/run when I remove it but it's such a tiny nit pick I don't think it matters much either way :laughing:

ctz: Ah, yeah, same here. It seems `ecdsa` is the name of a crate full of generics, and `p256::ecdsa` is the same set of generics reified with the p256 curve.

cpu (approved):

PR #1597, opened 2023-11-17
make the decode / decrypt pipeline non-allocating (groundwork)

japaric:
this PR lays the groundwork for making the decoding - decryption - decoding pipeline "non-allocating" AKA removing as many allocations as possible

at a very high level that is enabled by these changes

- `MessageDeframer::pop` performs decoding - decryption in-place and yields borrowed data (slices) to the next layer
- `Codec` gains a lifetime parameter `'a` to allow its `read` method to return maximally borrowed data (`&'a`) from `Reader<'a>` -- this lets the _post_-decryption decoding to also be non-allocating
- `Payload` becomes a `std::Cow`-like enum to support carrying either borrowed or owned data -- `HandshakePayload` and other types indirectly become `Cow`-like as well
- `State::handle` is tweaked to allow its return type hold borrowed data contained in the `message` parameter

to showcase how this enables the removal of allocations from the pipeline, this PR makes decoding & decryption of application-data as well as certificates (`CertificateChain`) non-allocating

removing the rest of the allocations that exist in the pipeline will be done in a subsequent PR

breaking changes:
- this change the signature of `MessageDecrypter::decrypt`. actual implementations don't have to change much, as you can see in the diff, but a breaking change is still a breaking change.

this PR builds on top of #1595 

--- 

I think it's worthwhile to explain why `Payload` and many of the `State` implementers need to become `Cow`-like and cannot contain just references.

In very simplified terms, you can think of the process of advancing `Connection`'s internal state machine as the following function: (this applies to both the buffered `Connection` and `UnbufferedConnection` APIs)

```rust
struct Connection {
    state: Option<Box<dyn State +'static>>,
    deframer: MessageDeframer,
    // ..
}

// in impl Connection
fn process_tls_records<'a>(&mut self, buffer: &mut DeframerSliceBuffer<'a>) {
    let mut current = self.state.take().unwrap();

    while let Some(message) = deframer.pop(buffer);{
        // message has type BorrowedPlainMessage<'a>
        let next = current.handle(state, message);
        current = next;
    }

    self.state = Some(current.into_owned());
}
```

each time this function is called, _several_ records are decoded and decrypted from the buffered incoming TLS data (the `buffer` parameter in the above snippet). as mentioned above, `Deframer::pop` now yields slices with lifetime `'a`. each `message`s that's fed into the state machine through the `State::handle` method can change the concrete state type behind the `current` trait object; each of those intermediate states can hold the `'a` borrowed data in `message` thanks to the `State::handle` change done in this PR.

eventually, the `deframer` is exhausted (e.g. the user may need to read data from the socket and append it to `buffer`) and it's time to return from the function. however, just after exiting the `while` loop `current` may contain borrowed `'a` content. it's not possible to store that borrowed content in `Connection` because its `state` field has a `'static` constraint. that's where we are forced to erase the `'a` lifetime by switching the true state behind `current` into its owned variant; this heap allocates any borrowed content. it's worth noting that `into_owned` it's a no-operation in the case the state type does not contain any borrowed data.

so yes, there's still one potential allocation in this function call but all of the many decoding and decryption events that happen within this function call becoming non-allocating; it also becomes possible to move between states -- within a single function call -- without needing to allocate borrowed content.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -734,20 +734,38 @@ impl<Data> ConnectionCore<Data> {
             }
         };
 
-        let mut borrowed_buffer = deframer_buffer.borrow();
-        while let Some(msg) = self.deframe(Some(&*state), &mut borrowed_buffer)? {
+        let mut discard = 0;
+        loop {
+            let mut borrowed_buffer = deframer_buffer.borrow();
+            borrowed_buffer.queue_discard(discard);
+
+            let res = self.deframe(Some(&*state), &mut borrowed_buffer);
+            discard = borrowed_buffer.pending_discard();
+
+            let opt_msg = match res {
+                Ok(opt_msg) => opt_msg,
+                Err(e) => {
+                    self.state = Err(e.clone());
+                    deframer_buffer.discard(discard);
+                    return Err(e);
+                }
+            };
+
+            let msg = match opt_msg {
+                Some(msg) => msg,
+                None => break,
+            };
+
             match self.process_msg(msg, state, Some(sendable_plaintext)) {
                 Ok(new) => state = new,
                 Err(e) => {
                     self.state = Err(e.clone());
-                    let discard = borrowed_buffer.pending_discard();
                     deframer_buffer.discard(discard);
                     return Err(e);
                 }
             }
         }
 
-        let discard = borrowed_buffer.pending_discard();
         deframer_buffer.discard(discard);
         self.state = Ok(state);
         Ok(self.common_state.current_io_state())
@@ -781,7 +799,7 @@ impl<Data> ConnectionCore<Data> {
                 }
 
                 self.common_state.aligned_handshake = aligned;
-                Ok(Some(message))
+                Ok(Some(message.into_owned()))
             }
             Ok(None) => Ok(None),
             Err(err @ Error::InvalidMessage(_)) => {
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -1,11 +1,12 @@
 use crate::common_state::{CommonState, Context, IoState, State, DEFAULT_BUFFER_LIMIT};
+use crate::crypto::cipher::BorrowedPlainMessage;
 use crate::enums::{AlertDescription, ContentType};
 use crate::error::{Error, PeerMisbehaved};
 #[cfg(feature = "logging")]
 use crate::log::trace;
 use crate::msgs::deframer::{Deframed, DeframerSliceBuffer, DeframerVecBuffer, MessageDeframer};
 use crate::msgs::handshake::Random;
-use crate::msgs::message::{Message, MessagePayload, PlainMessage};
+use crate::msgs::message::{Message, MessagePayload};
 use crate::suites::{ExtractedSecrets, PartiallyExtractedSecrets};
 use crate::vecbuf::ChunkVecBuffer;
 
@@ -336,11 +337,11 @@ impl ConnectionRandoms {
 
 // --- Common (to client and server) connection functions ---
 
-fn is_valid_ccs(msg: &PlainMessage) -> bool {
+fn is_valid_ccs(msg: &BorrowedPlainMessage) -> bool {
     // We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
     // Note: this is prior to the record layer, so is unencrypted. See
     // third paragraph of section 5 in RFC8446.
-    msg.typ == ContentType::ChangeCipherSpec && msg.payload.bytes() == [0x01]
+    msg.typ == ContentType::ChangeCipherSpec && msg.payload == [0x01]
 }
 
 /// Interface shared by client and server connections.
@@ -474,7 +475,8 @@ impl<Data> ConnectionCommon<Data> {
         let mut deframer_buffer = self.deframer_buffer.borrow();
         let res = self
             .core
-            .deframe(None, &mut deframer_buffer);
+            .deframe(None, &mut deframer_buffer)
+            .map(|opt| opt.map(|m| m.into_owned()));
         let discard = deframer_buffer.pending_discard();
         self.deframer_buffer.discard(discard);
 
@@ -772,11 +774,11 @@ impl<Data> ConnectionCore<Data> {
     }
 
     /// Pull a message out of the deframer and send any messages that need to be sent as a result.
-    fn deframe(
+    fn deframe<'b>(
         &mut self,
         state: Option<&dyn State<Data>>,
-        deframer_buffer: &mut DeframerSliceBuffer,
-    ) -> Result<Option<PlainMessage>, Error> {
+        deframer_buffer: &mut DeframerSliceBuffer<'b>,
+    ) -> Result<Option<BorrowedPlainMessage<'b>>, Error> {
         match self.message_deframer.pop(
             &mut self.common_state.record_layer,
             self.common_state.negotiated_version,
@@ -799,7 +801,7 @@ impl<Data> ConnectionCore<Data> {
                 }
 
                 self.common_state.aligned_handshake = aligned;
-                Ok(Some(message.into_owned()))
+                Ok(Some(message))
             }
             Ok(None) => Ok(None),
             Err(err @ Error::InvalidMessage(_)) => {
@@ -831,7 +833,7 @@ impl<Data> ConnectionCore<Data> {
 
     fn process_msg(
         &mut self,
-        msg: PlainMessage,
+        msg: BorrowedPlainMessage,
         state: Box<dyn State<Data>>,
         sendable_plaintext: Option<&mut ChunkVecBuffer>,
     ) -> Result<Box<dyn State<Data>>, Error> {
@@ -860,7 +862,7 @@ impl<Data> ConnectionCore<Data> {
         }
 
         // Now we can fully parse the message payload.
-        let msg = match Message::try_from(msg) {
+        let msg = match Message::try_from(msg.into_owned()) {
             Ok(msg) => msg,
             Err(err) => {
                 return Err(self
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -340,7 +340,7 @@ fn is_valid_ccs(msg: &PlainMessage) -> bool {
     // We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
     // Note: this is prior to the record layer, so is unencrypted. See
     // third paragraph of section 5 in RFC8446.
-    msg.typ == ContentType::ChangeCipherSpec && msg.payload.0 == [0x01]
+    msg.typ == ContentType::ChangeCipherSpec && msg.payload.bytes() == [0x01]
 }
 
 /// Interface shared by client and server connections.
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -471,16 +471,16 @@ impl<Data> ConnectionCommon<Data> {
     ///
     /// This is a shortcut to the `process_new_packets()` -> `process_msg()` ->
     /// `process_handshake_messages()` path, specialized for the first handshake message.
-    pub(crate) fn first_handshake_message(&mut self) -> Result<Option<Message>, Error> {
+    pub(crate) fn first_handshake_message(&mut self) -> Result<Option<Message<'static>>, Error> {
         let mut deframer_buffer = self.deframer_buffer.borrow();
         let res = self
             .core
             .deframe(None, &mut deframer_buffer)
-            .map(|opt| opt.map(|m| m.into_owned()));
+            .map(|opt| opt.map(|pm| Message::try_from(pm).map(|m| m.into_owned())));
         let discard = deframer_buffer.pending_discard();
         self.deframer_buffer.discard(discard);
 
-        match res?.map(Message::try_from) {
+        match res? {
             Some(Ok(msg)) => Ok(Some(msg)),
             Some(Err(err)) => Err(self.send_fatal_alert(AlertDescription::DecodeError, err)),
             None => Ok(None),
@@ -862,7 +862,7 @@ impl<Data> ConnectionCore<Data> {
         }
 
         // Now we can fully parse the message payload.
-        let msg = match Message::try_from(msg.into_owned()) {
+        let msg = match Message::try_from(msg) {
             Ok(msg) => msg,
             Err(err) => {
                 return Err(self
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -57,8 +57,38 @@ impl<'a> DerefMut for BorrowedPayload<'a> {
 }
 
 impl<'a> BorrowedPayload<'a> {
+    #[cfg(test)]
+    pub(crate) fn new(bytes: &'a mut [u8]) -> Self {
+        Self(bytes)
+    }
+
+    pub fn truncate(&mut self, len: usize) {
+        if len >= self.len() {
+            return;
+        }
+
+        self.0 = core::mem::take(&mut self.0)
+            .split_at_mut(len)
+            .0;
+    }
+
     pub(crate) fn read(r: &mut ReaderMut<'a>) -> Self {
         Self(r.rest())
+    }
+
+    pub(crate) fn into_inner(self) -> &'a mut [u8] {
+        self.0
+    }
+
+    pub(crate) fn pop(&mut self) -> Option<u8> {
+        if self.is_empty() {
+            return None;
+        }
+
+        let len = self.len();
+        let last = self[len - 1];
+        self.truncate(len - 1);
+        Some(last)
     }
 }
 
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -57,7 +57,6 @@ impl<'a> DerefMut for BorrowedPayload<'a> {
 }
 
 impl<'a> BorrowedPayload<'a> {
-    #[allow(dead_code)] // TODO(@cpu): remove in "introduce and expose BorrowedOpaqueMessage"
     pub(crate) fn read(r: &mut ReaderMut<'a>) -> Self {
         Self(r.rest())
     }
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -18,7 +18,7 @@ pub enum Payload<'a> {
     Owned(Vec<u8>),
 }
 
-impl<'a> Codec<'a> for Payload<'static> {
+impl<'a> Codec<'a> for Payload<'a> {
     fn encode(&self, bytes: &mut Vec<u8>) {
         bytes.extend_from_slice(self.bytes());
     }
@@ -35,11 +35,20 @@ impl<'a> Payload<'a> {
             Self::Owned(bytes) => bytes,
         }
     }
+
+    pub fn into_owned(self) -> Payload<'static> {
+        Payload::Owned(self.into_vec())
+    }
+
     pub fn into_vec(self) -> Vec<u8> {
         match self {
             Self::Borrowed(bytes) => bytes.to_vec(),
             Self::Owned(bytes) => bytes,
         }
+    }
+
+    pub fn read(r: &mut Reader<'a>) -> Self {
+        Self::Borrowed(r.rest())
     }
 }
 
@@ -50,10 +59,6 @@ impl Payload<'static> {
 
     pub fn empty() -> Self {
         Self::Borrowed(&[])
-    }
-
-    pub fn read(r: &mut Reader) -> Self {
-        Self::Owned(r.rest().to_vec())
     }
 }
 
--- a/rustls/src/msgs/base.rs
+++ b/rustls/src/msgs/base.rs
@@ -13,29 +13,47 @@ use super::codec::ReaderMut;
 
 /// An externally length'd payload
 #[derive(Clone, Eq, PartialEq)]
-pub struct Payload(pub Vec<u8>);
+pub enum Payload<'a> {
+    Borrowed(&'a [u8]),
+    Owned(Vec<u8>),
+}
 
-impl Codec<'_> for Payload {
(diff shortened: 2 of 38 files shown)


japaric:
@cpu  @ctz @djc could I get a quick sanity check on the direction here? the main commits that affect the end-user and/or internal _interfaces_ are:

- make Reader work on mutable slices
- make MessageDecrypter::decrypt operate on slices (:warning: this is a semver breaking change)
- make Deframer::pop return a slice (+)
- prepare Codec for non-allocating decoding
- add a Borrowed variant to Payload
- prepare State trait to handle slice fields

TL;DR make things `std::Cow`-like and add a bunch of lifetimes everywhere

the bulk of commits that will reduce allocations will be fairly mechanical like the 'reduce allocs around ServerCertDetails ' one

thoroughly minimizing the internal API in an incremental fashion is going to take a lot of commits but I'll break out a smaller PR out of this one that only does the interface changes

now, I mainly did this to enable in-place (non-allocating) decoding / decryption in the `UnbufferedConnection` API but this PR also improves the performance of the existing `Connection` API.

so far, I have only removed allocations from -- what my uneducated guess of what the biggest contributors are like -- application-data and certificates and I'm seeing improvements of this magnitude (ran these locally; I've filtered out the noisy `aws_lc_rs` benchmarks)

| Scenario                                          | Baseline | Candidate |                Diff |
| ------------------------------------------------- | -------: | --------: | ------------------: |
| handshake_tickets_ring_1.2_rsa_aes_client         |  4615881 |   4548840 |  ✅ -67041 (-1.45%) |
| handshake_session_id_ring_1.2_rsa_aes_server      |  4327962 |   4267674 |  ✅ -60288 (-1.39%) |
| handshake_tickets_ring_1.2_rsa_aes_server         |  4776473 |   4715679 |  ✅ -60794 (-1.27%) |
| handshake_session_id_ring_1.3_ecdsa_chacha_client | 41598210 |  41489715 | ✅ -108495 (-0.26%) |
| handshake_tickets_ring_1.3_ecdsa_chacha_client    | 41796305 |  41688683 | ✅ -107622 (-0.26%) |
| handshake_tickets_ring_1.3_rsa_chacha_client      | 42040674 |  41932666 | ✅ -108008 (-0.26%) |
| handshake_tickets_ring_1.3_rsa_aes_client         | 42083707 |  41975702 | ✅ -108005 (-0.26%) |
| handshake_tickets_ring_1.3_ecdsa_aes_client       | 41837011 |  41731569 | ✅ -105442 (-0.25%) |
| handshake_session_id_ring_1.3_ecdsa_aes_client    | 41651062 |  41546453 | ✅ -104609 (-0.25%) |
| handshake_session_id_ring_1.3_rsa_aes_client      | 41899822 |  41798231 | ✅ -101591 (-0.24%) |
| handshake_session_id_ring_1.3_rsa_chacha_client   | 41845117 |  41743759 | ✅ -101358 (-0.24%) |

---

(+) I intend to use this `RawSlice` type in `Context` to make the `UnbufferedConnection` truly not-allocating when it comes to decrypting app-data and it'll probably also become instrumental to allowing caller-side certificate verification with the `UnbufferedConnection` API for future compatibility with async cert verifiers

cpu:
> @cpu @ctz @djc could I get a quick sanity check on the direction here?

ACK'ing this but wanted to mention it might be a little bit. We're focused on 0.22 release prep and Ctz is AFK for a little bit. The high-level vision sounds good to me, but I think input from Djc and Ctz will be most valuable here. My experience in this domain is more limited.

japaric:
> but I'll break out a smaller PR out of this one that only does the interface changes

I, instead, trimmed down this PR and it's ready for review. I have added a high level description of the changes in the PR description. removing the bulk of the allocations will be need in a separate / subsequent PR.

(the MSRV seems to not like the function/struct update syntax used in this PR :cry: )

rustls/src/conn.rs
```diff
@@ -649,31 +651,40 @@ impl<Data> ConnectionCore<Data> {
...
         };
 
-        let mut borrowed_buffer = deframer_buffer.borrow();
-        while let Some(msg) = self.deframe(Some(&*state), &mut borrowed_buffer)? {
+        let mut discard = 0;
+        loop {
+            let mut borrowed_buffer = deframer_buffer.borrow();
+            borrowed_buffer.queue_discard(discard);
+
+            let opt_msg = self.deframe(Some(&*state), &mut borrowed_buffer)?;
```
ctz, line 659: I think this continues to have the same issue as the other PR? `borrowed_buffer.pending_discard()` is crucial data that must not be lost on error.

ctz:
@rustls-benchmarking bench

rustls/src/msgs/message.rs
```diff
@@ -387,6 +387,7 @@ impl TryFrom<PlainMessage> for Message {
 ///
 /// This type also cannot decode its internals and
 /// cannot be read/encoded; only `OpaqueMessage` can do that.
+#[derive(Debug)]
 pub struct BorrowedPlainMessage<'a> {
```
cpu, line 429: Let's roll the `Debug` derive into b99e527499e96ada274bf4a29383b88b8b27fa22

rustls/src/msgs/deframer.rs
```diff
@@ -155,19 +155,21 @@ impl MessageDeframer {
 
             // If it's not a handshake message, just return it -- no joining necessary.
             if msg.typ != ContentType::Handshake {
+                let message = msg.into_owned();
```
cpu: I'm not sure I spot how these `MessageDeframer` changes map to the `Decrypted` update. Possibly a candidate to move into a different commit?

rustls/src/msgs/message.rs
```diff
@@ -211,19 +228,6 @@ impl<'a> BorrowedOpaqueMessage<'a> {
...
-    pub(crate) fn into_owned(self) -> OpaqueMessage {
-        let Self {
-            typ,
-            version,
-            payload,
-        } = self;
-        OpaqueMessage {
-            typ,
-            version,
-            payload: Payload::new(&*payload),
```
cpu: It looks like while this was moved the `&*` was dropped, can you fix that in the commit that added `into_owned` on `BorrowedOpaqueMessage` instead?

rustls/src/msgs/handshake.rs
```diff
@@ -1352,12 +1363,17 @@ impl Codec<'_> for CertificateEntry {
...
+            cert: certificate_der_into_owned(CertificateDer::read(r)?),
             exts: Vec::read(r)?,
         })
     }
 }
 
+// FIXME(perf) sub-optimal in the owned case; switch to `CertificateDer::into_owned` when available
+fn certificate_der_into_owned(cert: CertificateDer<'_>) -> CertificateDer<'static> {
+    CertificateDer::from(cert.to_vec())
+}
```
cpu, line 1375: I think this FIXME can be resolved, I see [CertificateDer::into_owned](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.CertificateDer.html#method.into_owned) landed.

cpu (commented):
This is really cool.

rustls/src/msgs/deframer.rs
pvdrz: That's because this `msg` value comes from a `Decrypted` value, which has a `BorrowedPlainMessage` now instead of a `PlainMessage`

rustls/src/conn.rs
pvdrz: maybe here we can just match on the output of deframe and discard on error. What do you think?

ctz: I think it would be more robush if `deframer_buffer.borrow()` returned an object that automatically took care of this on drop. However, I think achieving that is not currently possible because the deframer buffer is mutable (despite deframing being a strictly read-only operation) and fixing that is a larger refactor.

So I guess ensuring every caller of `deframer_buffer.borrow()` does the right thing is the second best option now.

ctz:
> to showcase how this enables the removal of allocations from the pipeline, this PR makes decoding & decryption of application-data as well as certificates (`CertificateChain`) non-allocating

I think we're just moving the allocations around rather than eliminating them?

eg, for `CertificateChain`, in TLS1.3 client:

- the deframer message decoding is now allocation-free,
- in `ExpectCertificate`: from `Message` -> `ServerCertDetails` is allocation-free,
- (if this is the end of the flight, the contents of `ServerCertDetails` are copied, but that is infrequent -- however we do need test coverage in both cases)
- in `ExpectCertificateVerify`, we copy (both shallow and deep if needed) the `CertificateChain` into `cx.common.peer_certificates` to appear in the public API.

Total copies of the certificate data: 1, plus one extra small vec

Prior to this PR:

- the deframer message decoding allocates for the `CertificateChain` contents
- in `ExpectCertificate`: this is moved from `Message` into `ServerCertDetails` which is allocation-free,
- in `ExpectCertificateVerify` the `CertificateChain` is moved into `cx.common.peer_certificates` to appear in the public API.

Total copies of the certificate data: 1

rustls/src/conn.rs
pvdrz: Ok I did the second best thing :P

rustls/src/msgs/message.rs
ctz: yes, usually a "fixup" commit is for squashing into a previous commit

cpu: I don't think it was described as a fixup prefixed commit when I left this feedback, but if it was I apologize for missing it!

pvdrz: I moved this hunk to the commit you mentioned. Is that what you wanted?

rustls/src/msgs/message.rs
```diff
@@ -368,16 +381,38 @@ impl Message {
...
+        let Self { version, payload } = self;
+        Message {
+            version,
+            payload: payload.into_owned(),
+        }
+    }
+}
+
+// NOTE does not avoid allocations like the `BorrowedPlainMessage -> Message` conversion. this is
+// only used in (integration) tests
```
ctz, line 395: Seems like `Message::try_from(PlainMessage)` is used in `first_handshake_message`, which is not just an integration test?

That seems OK (avoiding it seems to require quite a bit of change on the `Acceptor` API side) but this comment is potentially misleading.

ctz (commented):
A few more comments, mainly about commit structuring and clarity. But I think this is getting closer.

rustls/src/msgs/deframer.rs
```diff
@@ -438,22 +439,26 @@ pub struct DeframerSliceBuffer<'a> {
     // a fully initialized buffer that will be deframed
     buf: &'a mut [u8],
     // number of bytes to discard from the front of `buf` at a later time
-    discard: usize,
```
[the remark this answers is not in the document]
pvdrz: I think I'm not following you. The `discard` field type is changed to `Cell<usize>` on this commit basically. this is done to avoid borrowing issues later but I think adding that to a later PR would make it a little bit confusing. I'll add an explicit `groundwork:` to it.

ctz (approved):

cpu (approved):
Thanks! All of my earlier feedback looks resolved and I didn't have anything new to add.

rustls/src/msgs/codec.rs
```diff
@@ -85,6 +85,68 @@ impl<'a> Reader<'a> {
...
+
+    fn take(&mut self, length: usize) -> Option<&'a mut [u8]> {
+        if self.left() < length {
+            return None;
+        }
+        let (taken, rest) = mem::take(&mut self.buffer).split_at_mut(length);
+        self.used += taken.len();
+        self.buffer = rest;
+        Some(taken)
+    }
```
djc, line 146: Nit: I feel like these should remain in the same order as for `Reader`?

rustls/src/msgs/message.rs
```diff
@@ -214,6 +190,66 @@ impl OpaqueMessage {
...
+
+        Ok(Self {
+            typ,
+            version,
+            payload,
+        })
+    }
+}
+
+fn read_opaque_message_minus_payload(
```
djc, line 252: Naming nit: I'd probably call this `read_opaque_message_header()`.

rustls/src/msgs/deframer.rs
```diff
@@ -491,6 +491,15 @@ fn copy_into_buffer(buf: &mut [u8], src: &[u8], at: usize) {
 }
 
 trait FilledDeframerBuffer {
+    fn filled_get_mut<I>(&mut self, index: I) -> &mut I::Output
```
djc, line 602: Style nit: `impl SliceIndex<[u8]>`?

rustls/src/msgs/base.rs
```diff
@@ -57,6 +57,36 @@ impl<'a> DerefMut for BorrowedPayload<'a> {
...
+    pub(crate) fn pop(&mut self) -> Option<u8> {
+        if self.is_empty() {
+            None
+        } else {
+            let len = self.len();
+            let last = self[len - 1];
+            self.truncate(len - 1);
+            Some(last)
+        }
+    }
```
djc: Style nit: this is a pretty surprising method order IMO.

rustls/src/msgs/base.rs
```diff
@@ -57,6 +57,36 @@ impl<'a> DerefMut for BorrowedPayload<'a> {
...
+    pub(crate) fn new(bytes: &'a mut [u8]) -> Self {
+        Self(bytes)
+    }
+
+    pub(crate) fn into_inner(self) -> &'a mut [u8] {
+        self.0
+    }
+
+    pub(crate) fn pop(&mut self) -> Option<u8> {
+        if self.is_empty() {
```
djc, line 103: Should do an early return here?

rustls/src/record_layer.rs
```diff
@@ -62,12 +62,12 @@ impl RecordLayer {
     /// an error is returned.
     pub(crate) fn decrypt_incoming(
         &mut self,
-        encr: OpaqueMessage,
+        encr: BorrowedOpaqueMessage,
```
djc: Nit: let's make the lifetime explicit?

rustls/src/msgs/deframer.rs
```diff
@@ -462,15 +491,28 @@ impl<'a> DeframerSliceBuffer<'a> {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    fn take(&mut self, raw: RawSlice) -> &'a mut [u8] {
+        let start = (raw.ptr as usize)
+            .checked_sub(self.buf.as_ptr() as usize)
+            .unwrap();
```
djc, line 543: So we're doing pointer arithmetic here -- how do we guarantee that `raw.ptr` and `self.buf` are from the same allocation? Even though there is no `unsafe` code here, this seems to be potentially brittle code which should come with more documentation.

rustls/src/msgs/deframer.rs
```diff
@@ -504,6 +549,13 @@ impl<'a> DeframerSliceBuffer<'a> {
 
         &mut taken[start..]
     }
+
+    fn raw_slice_to_filled_range(&self, raw: RawSlice) -> Range<usize> {
+        let adjust = self.discard - self.taken;
+        let start = ((raw.ptr as usize).checked_sub(self.buf.as_ptr() as usize)).unwrap() - adjust;
```
djc, line 555: Same here.

djc (commented):
Hmm, I'm not really comfortable with some of the things that are happening here. In particular the work around the `RawSlice` which has some invariants which are neither enforced by the type system nor clearly documented, and some complexity around the `AppendPayload` trait and impls that it's not obvious are needed.

(Informed by my experience working on #1179.)

rustls/src/msgs/deframer.rs
cpu: I think because of the `Output` associated type being referred to in the `return` using `impl` ends up being pretty ugly. The return needs to be qualified to avoid ambiguity:

```rust
fn filled_get_mut(
        &mut self,
        index: impl SliceIndex<[u8], Output = [u8]>,
    ) -> &mut <RangeFrom<usize> as SliceIndex<[u8]>>::Output 
```

I'm inclined to leave it as-is with the generic parameter. WDYT?

rustls/src/msgs/codec.rs
cpu: Fixed.

rustls/src/msgs/message.rs
cpu: Renamed

rustls/src/msgs/base.rs
cpu: Reordered

rustls/src/record_layer.rs
cpu: Done. A later commit further refines this to a named lifetime.

rustls/src/msgs/base.rs
cpu: Fixed.

rustls/src/msgs/deframer.rs
djc: Yeah, the impl argument doesn't make sense, sorry. I'd probably inline the I bound in the declaration instead of having a where clause, too.

cpu: Fixed.

rustls/src/msgs/deframer.rs
cpu: I'm struggling to understand the `RawSlice` abstraction and the commit that introduced it is missing any motivating context or guidance in the commit message (I guess that's part of what you're flagging in this review...). I gather we want to use pointers here instead of a normal slice into `self.buf` for reasons related to the lifetime of the slice and the `DeframerSliceBuffer` buffer, but I can't exactly put my finger on it.

>  how do we guarantee that raw.ptr and self.buf are from the same allocation?

I haven't been able to think of a way to tie the `RawSlice` back to the buffer it came from that won't trip into the lifetime issues that I think this design was built around avoiding. Do you have any suggestions?

>  this seems to be potentially brittle code which should come with more documentation.

Would additional documentation about the invariants be enough to satisfy your concerns? That's certainly doable :+1:

cpu: > how do we guarantee that raw.ptr and self.buf are from the same allocation?

We could drop the `RawSlice::from<&[u8]>` constructor and implement a method on `DeframerSliceBuffer` that takes a slice argument and asserts the base ptr and len of the slice falls within the `DeframerSliceBuffer`'s `buf` before giving back the `RawSlice`.

Writing that out, I'm not sure that it's a very useful change w.r.t safety or brittleness since it just changes where we panic: constructing the `RawSlice` or doing the `split_at_mut`/`checked_sub` in `take`/`raw_slice_to_filled_range` with the result of the pointer arithmetic.

@ctz Do you have any thoughts on this feedback or suggestions for refactoring?

ctz: I'd say we're doing a similar thing in a different way elsewhere inside msgs/deframer.rs: but with `Range` rather than pointer arithmetic.

That is, at least, more obvious -- though I think it still suffers from similar fragility problems if anyone changes the buffer while the `Range` is valid. That instance is especially egregious because the buffer is under control of the application while the `Range` is valid. I guess this is one of the negative results of pursuing a 100% (including handshaking) no-allocation solution here rather than just addressing the main data path.

ctz: I have an extremely broken and half-baked branch with some refactoring in this area, but haven't reached an ending on it (either achieving something good, or giving up!)

Some of the ideas I've so far failed to achieve:

- The deframer only ever reads from its buffer, and the `Deframed` object it returns doesn't keep a borrow of the message data; it just tells you where it is. That gives the deframer potential to become an iterator over `Result<Deframed, Error>` items in the future, maybe.
- Someone with a `Deframed` message can decrypt it in-place by providing the `record_layer` and the mutable slice containing it. That process should also not keep a mutable borrow of the buffer.
- The handshake part of the deframer is split out into a separate thing. At the moment the deframer is doing both slicing up TLS messages, and also gluing together TLS handshake protocol fragments. I would hope this smooths out the complexity hotspot in the deframer.
- The QUIC code can directly use the handshake-part-of-the-deframer, which should relieve us of a lot of the `if quic { }` decisions in the deframer.

djc: That sounds great. Maybe @cpu could take the half-baked branch and run with it?

ctz: Pushed https://github.com/rustls/rustls/tree/jbp-refactor-deframer

cpu: > That sounds great. Maybe @cpu could take the half-baked branch and run with it?

Sure, I can give that a go :+1:, but is it a blocker for landing this work? I'd rather not have to maintain two unfinished branches at the same time if possible. If it's OK with you folks my preference would be to land this branch and then rebase the refactor on top to continue.

> Pushed https://github.com/rustls/rustls/tree/jbp-refactor-deframer

@ctz Could you maybe summarize where you left off to help me get started?

cpu: > That seems reasonable.

Thanks :+1: I'll add some additional documentation around the `RawSlice` invariants and then merge this.

cpu: Updated with additional doc comments.

rustls/src/msgs/deframer.rs
cpu: Updated with additional doc comments.

cpu:
> Hmm, I'm not really comfortable with some of the things that are happening here

Pulling some discussion [out of a resolved review thread](https://github.com/rustls/rustls/pull/1597#discussion_r1446139734):

* I've added more documentation around the invariants required by the `RawSlice` portions of the work internal to the deframer.
* There's similar pre-existing "brittleness" w.r.t the deframer and `Range` operations.
* We're going to land this branch as-is, accepting that it introduces a bit more brittleness in the short term.
* I'm going to pick up [a refactor branch](https://github.com/rustls/rustls/tree/jbp-refactor-deframer) that ctz started to try and reign in some of the complexity as follow-up work.

PR #1599, opened 2023-11-20
QUIC changes follow-up

djc:
Feedback from #1594.

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -907,11 +907,10 @@ impl State<ClientConnectionData> for ExpectFinished {
             _fin_verified: fin,
         };
 
-        if cx.common.is_quic() {
-            Ok(Box::new(ExpectQuicTraffic(st)))
-        } else {
-            Ok(Box::new(st))
-        }
+        Ok(match cx.common.is_quic() {
+            true => Box::new(ExpectQuicTraffic(st)),
+            false => Box::new(st),
+        })
     }
 }
 
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -3,7 +3,7 @@ use crate::client::{ClientConfig, ClientConnectionData, ServerName};
 use crate::common_state::{CommonState, Protocol, Side};
 use crate::conn::{ConnectionCore, SideData};
 use crate::crypto::cipher::{AeadKey, Iv};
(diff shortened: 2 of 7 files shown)


rustls/src/client/tls13.rs
```diff
@@ -907,11 +907,10 @@ impl State<ClientConnectionData> for ExpectFinished {
             _fin_verified: fin,
         };
 
-        if cx.common.is_quic() {
-            Ok(Box::new(ExpectQuicTraffic(st)))
-        } else {
-            Ok(Box::new(st))
-        }
+        Ok(match cx.common.is_quic() {
```
ctz, line 910: if we use match here, i think doing it on `cx.common.protocol`, `Protocol::Quic`, etc. would be clearer than matching on a bool?

ctz (approved):

rustls/src/client/tls13.rs
ctz: (unless we expect to change the detail contained in `is_quic()` in the future, of course)

djc: Just because you dislike `match` on `bool`? That doesn't seem like a big deal to me.

IMO `is_quic()` is still a useful shorthand which avoids the need for a `Protocol` import.

ctz: i think i dislike bool in general and look for any and all opportunities to avoid it; but it's not a big deal :)

rustls/src/quic.rs
```diff
@@ -155,7 +155,7 @@ impl ClientConnection {
 
         if !config.supports_protocol(Protocol::Quic) {
             return Err(Error::General(
-                "At least one ciphersuite must support QUIC".into(),
+                "at least one ciphersuite must support QUIC".into(),
```
ctz, line 159: nit: "errror" in commit message

ctz (approved):

rustls/src/quic.rs
djc: Fixed that up, thanks for catching it!

cpu (approved):
Thanks :+1:

ctz:
Will need a rebase to pick up 076090d4bd2fa21c77497d12b3f7d19d4193d6e1 (forgot that names of GH actions steps are load-bearing!)

PR #1600, opened 2023-11-20
Update dependencies

djc:


djc:
Some nice perf improvements from aws-lc-rs in there...

ctz (approved):

ctz:
> Some nice perf improvements from aws-lc-rs in there...

Yep, I think this is the fix landing for the RSA issue mentioned here: https://github.com/rustls/rustls/pull/1414#issuecomment-1796515954

PR #1601, opened 2023-11-20
Use `-Z direct-minimal-versions` for minimal versions checks

ctz:
Following the advice on https://doc.rust-lang.org/cargo/reference/unstable.html#minimal-versions, apparently `-Z direct-minimal-versions` is closer to what we want (checking that _our_ `Cargo.toml` has correct versions in it, rather than every `Cargo.toml` in our dependencies).

Use `cargo-minimal-versions` which lets us care only about public crates.

This also side-steps the aws-lc-rs issue seen in #1586

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -237,7 +237,7 @@ jobs:
 
 
   minver:
-    name: Check minimum versions
+    name: Check minimum versions of direct dependencies
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -248,14 +248,15 @@ jobs:
...
-        # This has no effect if no `Cargo.lock` exists yet.
-        run: cargo update -Z direct-minimal-versions
+      - name: Install cargo-minimal-versions
+        uses: taiki-e/install-action@cargo-minimal-versions
 
-      - name: cargo test (debug; all features)
-        run: cargo test --locked --all-features
-        env:
-          RUST_BACKTRACE: 1
+      - name: Install cargo-hack
```
djc, line 254: Do we need both `cargo-hack` and `cargo-minimal-versions`? Why?

djc (approved):

.github/workflows/build.yml
ctz: `cargo-minimal-versions` shells out to `cargo-hack` -- see https://github.com/taiki-e/cargo-minimal-versions#prerequisites

djc: Okay, might want to add a comment in the workflow here?

ctz: will do

PR #1604, opened 2023-11-20
crypto: doc-hide the HPKE module

cpu:
The bits and pieces we're landing for HPKE support aren't ready for broad use yet. To avoid confusion before the 0.22 release this commit adds a `#[doc(hidden)]` attribute to the `crypto/hpke.rs` mod.

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -38,6 +38,7 @@ pub mod tls12;
 pub mod tls13;
 
 /// Hybrid public key encryption (RFC 9180).
+#[doc(hidden)]
 pub mod hpke;
 
 pub use crate::rand::GetRandomFailed;


rustls/src/crypto/hpke.rs
```diff
@@ -15,6 +15,7 @@ use crate::Error;
 ///
 /// [RFC 9180]: <https://www.rfc-editor.org/rfc/rfc9180.html>
 /// [HPKE ciphersuite profile]: <https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-17#section-9>
+#[doc(hidden)]
```
ctz, line 18: AIUI this could be put on `pub mod hpke` in `crypto/mod.rs` to apply to the whole module

cpu: I'll give that a shot, I wasn't sure if a top level attribute would work.

cpu: That worked great, thanks for suggesting :+1:

ctz (approved):

PR #1606, opened 2023-11-21
Migrate to pki-types ServerName

djc:
Prerequisites:

- [x] https://github.com/rustls/pki-types/pull/11
- [x] https://github.com/rustls/webpki/pull/207

Use the new `ServerName` type provided by pki-types.


cpu:
> Currently there's a surprising borrow checker issue in the ClientSessionMemoryCache implementation, would be happy if someone can take a look

I think the crux of the issue is that `ClientSessionMemoryCache` holds a `LimitedCache<ServerName<'static>, ServerData>`, matching the previous form where it used an owned `ServerName` as the key for the cache map. I think the `ClientSessionMemoryCache::remove_tls12_session` and `ClientSessionMemoryCache::take_tls13_ticket`'s `server_name` arguments need a `'static` lifetime instead of the elided lifetime in order to work with that arrangement.

After making that change, updating the impl, and adding a `'static` lifetime to the arg in `client/hs.rs`'s `find_session` things compile (mod the errors from the incompatible pemfile pki-types). PTAL at https://github.com/rustls/rustls/commit/0a92d84a20736bffa434f9734fb7d16ddd37a994 It's very possible I'm missing some deeper complication here!

djc:
So a tricky thing here is that the rustls `ServerName` implemented relies on `std::net::IpAddr` parsing to convert compressed IPv6 addresses into normal form, while pki-types tries to not rely on `std` unless specifically opted into. Should pki-types get more powerful `ServerName::try_from()` for IPv6 addresses depending on whether `std` is enabled? Or should I add a separate `std`-only method that uses `std` IP address conversions on `pki_types::ServerName` and `pki_types::IpAddr`?

cpu:
> Should pki-types get more powerful ServerName::try_from() for IPv6 addresses depending on whether std is enabled? 

That seems kind of reasonable to me but I'm trying to reason through the upsides/downsides.

Upside: feature unification could make code in one crate that's using `try_from` w/o `std` benefit from the expanded functionality when another crate activates `std` without a code change being required.
Downside: It might be surprising behavior. 

Are there other downsides I'm overlooking that gave you pause?

jsha:
It looks like https://github.com/rustls/pki-types/pull/11 has a copy of the IpAddr / Ipv4Addr / Ipv6Addr code from core, including the parsing code. Doesn't that suffice to allow the pki-types `ServerName` to be just as powerful as the rustls `ServerName`?

djc:
No, the parsing implementations are from what was in webpki before (which is relatively restrictive).

djc:
I've replaced the IP address parser implementation from webpki with the one from core in pki-types, see more in the pki-types PR: https://github.com/rustls/pki-types/pull/11#issuecomment-1824106793.

examples/src/bin/simple_0rtt_client.rs
```diff
@@ -2,12 +2,13 @@ use std::io::{BufRead, BufReader, Write};
 use std::net::TcpStream;
 use std::sync::Arc;
 
+use pki_types::ServerName;
 use rustls::RootCertStore;
```
cpu, line 6: nitpick: I think there's conventionally whitespace between the crate imports and dependency imports.

rustls/src/client/handy.rs
```diff
@@ -101,7 +102,7 @@ impl client::ClientSessionStore for ClientSessionMemoryCache {
 
     fn set_tls12_session(
         &self,
-        _server_name: &ServerName,
+        _server_name: ServerName<'static>,
```
cpu, line 105: nit: this is carried over from the old version, but I think `_server_name` should be `server_name` since it's used on L111 for the `get_or_insert_default_and_edit` call.

If you agree & don't mind, the same feedback applies to `tls_session`, & `remove_tls12_session`

rustls/src/client/hs.rs
```diff
@@ -232,8 +232,8 @@ fn emit_client_hello_for_retry(
         ClientExtension::CertificateStatusRequest(CertificateStatusRequest::build_ocsp()),
     ];
 
-    if let (ServerName::DnsName(sni_name), true) = (&input.server_name, config.enable_sni) {
-        exts.push(ClientExtension::make_sni(sni_name.borrow()));
+    if let (ServerName::DnsName(dns), true) = (&input.server_name, config.enable_sni) {
```
cpu, line 235: If my feedback about adding a comment about why we do this `let` match on the `DnsName` variant made sense to you it could also be duplicated here.

rustls/src/webpki/verify.rs
```diff
@@ -43,28 +42,13 @@ pub fn verify_server_cert_signed_by_trust_anchor(
...
-        }
-    }
-    Ok(())
+pub fn verify_server_name(
+    cert: &ParsedCertificate,
+    server_name: &ServerName<'_>,
+) -> Result<(), Error> {
+    cert.0
+        .verify_is_valid_for_subject_name(server_name)
+        .map_err(pki_error)
```
cpu, line 51: It's nice to see the "unlikely error" bits fall away here :stars:

cpu (approved):
Awesome :+1:

rustls/src/client/hs.rs
```diff
@@ -232,8 +232,8 @@ fn emit_client_hello_for_retry(
         ClientExtension::CertificateStatusRequest(CertificateStatusRequest::build_ocsp()),
     ];
 
-    if let (Some(sni_name), true) = (input.server_name.for_sni(), config.enable_sni) {
-        exts.push(ClientExtension::make_sni(sni_name));
+    if let (ServerName::DnsName(sni_name), true) = (&input.server_name, config.enable_sni) {
```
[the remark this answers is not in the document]
djc: Good point, added a comment.

examples/src/bin/simple_0rtt_client.rs
djc: Good catch, fixed.

rustls/src/client/handy.rs
djc: I think this is intentional -- you may have glossed over the `#[cfg(feature = "tls12")]` guarding the `get_or_insert_default_and_edit()` call, which means that this would only be conditionally used dependending on Cargo features, which in turn means it would generate warnings in the other case. So going to leave this as is.

cpu: Ahh! yes, my mistake.

cpu (approved):
Can you fix up the PR desc. to avoid future confusion? Otherwise LGTM

Issue #1603, opened 2023-11-20
doc: which traits must be implemented for a custom crypto provider


jsha:
I was looking up how to implement a custom crypto provider. It looks like it's not currently documented in a consolidated place. These are all the traits found in modules under [rustls::crypto](https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/index.html), plus some in [rustls::quic](https://rustdoc.crud.net/jsha/rustls-doc/rustls/quic/index.html). Could someone confirm whether this is the full set? Also, does each provider need to implement all of the ones not marked Optional, or is it expected that some providers would delegate to an existing default implementation for some of them?

 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/trait.CryptoProvider.html">CryptoProvider</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/trait.ActiveKeyExchange.html">ActiveKeyExchange</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/trait.SupportedKxGroup.html">SupportedKxGroup</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/cipher/trait.MessageDecrypter.html">cipher::MessageDecrypter</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/cipher/trait.MessageEncrypter.html">cipher::MessageEncrypter</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/cipher/trait.Tls12AeadAlgorithm.html">cipher::Tls12AeadAlgorithm</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/cipher/trait.Tls13AeadAlgorithm.html">cipher::Tls13AeadAlgorithm</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/signer/trait.Signer.html">signer::Signer</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/signer/trait.SigningKey.html">signer::SigningKey</a>
 - <a href="https://rustdoc.crud.net/jsha/rustls-doc/rustls/crypto/tls12/trait.Prf.html">tls12::Prf</a>
[...]

PR #1608, opened 2023-11-21
Improve crypto provider API documentation

ctz:
Some misc docs clean ups, but mainly trying to improve matters around `CryptoProvider`

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Rustls is a TLS library that aims to provide a good level of cryptographic secur
 requires no configuration to achieve that security, and provides no unsafe features or
 obsolete cryptography by default.
 
-## Current features
+## Current functionality (with default crate features)
 
 * TLS1.2 and TLS1.3.
 * ECDSA, Ed25519 or RSA server authentication by clients.
@@ -83,7 +83,7 @@ need them.
 
 ### Platform support
 
-While Rustls itself is platform independent, it uses
+While Rustls itself is platform independent, by default it uses
 [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
 TLS. As a result, rustls only runs on platforms
 supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
@@ -91,6 +91,12 @@ x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Bi
 64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
 support WebAssembly.
 For more information, see [the supported `ring` target platforms][ring-target-platforms].
+
+By providing a custom implementation of the [`crate::crypto::CryptoProvider`] trait, you
+can replace all cryptography dependencies of rustls.  This is a route to being portable
+to a wider set of architectures and environments, or compliance requirements.
+Specifying `default-features = false` when depending on rustls will remove the
+dependency on *ring*.
 
 Rustls requires Rust 1.61 or later.
 
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -18,6 +18,19 @@ use core::marker::PhantomData;
 
 impl ConfigBuilder<ClientConfig, WantsVerifier> {
     /// Choose how to verify server certificates.
+    ///
+    /// Using this function does not configure revocation.  If you wish to
+    /// configure revocation, instead use:
+    ///
+    /// ```diff
+    /// - .with_root_certificates(root_store)
+    /// + .dangerous()
+    /// + .with_custom_certificate_verifier(
+    /// +   WebPkiServerVerifier::builder_with_provider(root_store, crypto_provider)
+    /// +   .with_crls(...)
+    /// +   .build()?
+    /// + )
+    /// ```
     pub fn with_root_certificates(
         self,
         root_store: impl Into<Arc<webpki::RootCertStore>>,
@@ -99,8 +112,9 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
     /// in client authentication.
     ///
     /// `cert_chain` is a vector of DER-encoded certificates.
-    /// `key_der` is a DER-encoded RSA, ECDSA, or Ed25519 private key -- the precise
-    /// set of supported key types and parameters is defined by the selected `CryptoProvider`.
+    /// `key_der` is a DER-encoded private key -- the precise set of supported key
+    /// types and parameters is defined by the selected [`CryptoProvider`]'s `load_private_key()`
+    /// method.
     ///
     /// This function fails if `key_der` is invalid.
     pub fn with_client_auth_cert(
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -45,18 +45,142 @@ pub use crate::rand::GetRandomFailed;
 
 pub use crate::msgs::handshake::KeyExchangeAlgorithm;
 
-/// Pluggable crypto galore.
+/// Controls core cryptography used by rustls.
+///
+/// This crate comes with two built-in options, provided as
+/// `&dyn CryptoProvider` values:
+///
+/// - [`crate::crypto::ring::RING`]: (behind the `ring` crate feature, which
+///   is enabled by default).  This provider uses the [*ring*](https://github.com/briansmith/ring)
+///   crate.
+/// - [`crate::crypto::aws_lc_rs::AWS_LC_RS`]: (behind the `aws_lc_rs` feature,
+///   which is optional).  This provider uses the [aws-lc-rs](https://github.com/aws/aws-lc-rs)
+///   crate.
+///
+/// # Using a specific `CryptoProvider`
+///
+/// Supply the provider when constructing your [`crate::ClientConfig`] or [`crate::ServerConfig`]:
+///
+/// - [`crate::ClientConfig::builder_with_provider()`]
+/// - [`crate::ServerConfig::builder_with_provider()`]
+///
+/// When creating and configuring a webpki-backed client or server certificate verifier, a choice of
+/// provider is also needed to start the configuration process:
+///
+/// - [`crate::client::WebPkiServerVerifier::builder_with_provider()`]
+/// - [`crate::server::WebPkiClientVerifier::builder_with_provider()`]
+///
+/// # Making a custom `CryptoProvider`
+///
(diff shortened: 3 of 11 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -45,19 +45,137 @@ pub use crate::rand::GetRandomFailed;
...
+/// Content-Length: 19899
+/// ```
+///
+/// [provider-example/]: https://github.com/rustls/rustls/tree/main/provider-example/
+/// [rust-crypto]: https://github.com/rustcrypto
+/// [dalek-cryptography]: https://github.com/dalek-cryptography
 pub trait CryptoProvider: Send + Sync + Debug + 'static {
     /// Fill the given buffer with random bytes.
+    ///
+    /// The bytes must be cryptographically strong.
```
djc, line 146: Bytes by themselves are never cryptographically strong, right?

rustls/src/crypto/mod.rs
```diff
@@ -45,19 +45,137 @@ pub use crate::rand::GetRandomFailed;
...
+    /// This is used by [`crate::ConfigBuilder::with_safe_defaults()`] and
+    /// [`crate::ConfigBuilder::with_safe_default_kx_groups()`].
+    ///
+    /// Other (non-default) key exchange groups can be provided separately and configured
+    /// by passing them to [`crate::ConfigBuilder::with_kx_groups()`].
     fn default_kx_groups(&self) -> &'static [&'static dyn SupportedKxGroup];
 
     /// Decode and validate a private signing key from `key_der`.
     ///
+    /// This is used by [`crate::ConfigBuilder::with_client_auth_cert()`], [`crate::ConfigBuilder::with_single_cert()`],
```
djc, line 175: Should we also explain in docs for `with_client_auth_cert()`, `with_single_cert()` and `with_single_cert_with_ocsp()` that the key type must be supported by the `CryptoProvider`?

rustls/src/crypto/mod.rs
```diff
@@ -45,19 +45,137 @@ pub use crate::rand::GetRandomFailed;
...
+/// `&dyn CryptoProvider` values:
+///
+/// - [`crate::crypto::ring::RING`]: (behind the `ring` crate feature, which
+///   is enabled by default).  This provider uses the [*ring*](https://github.com/briansmith/ring)
+///   crate.
+/// - [`crate::crypto::aws_lc_rs::AWS_LC_RS`]: (behind the `aws_lc_rs` feature,
+///   which is optional).  This provider uses the [aws-lc-rs](https://github.com/aws/aws-lc-rs)
+///   crate.
+///
+/// # Using a specific `CryptoProvider`
```
djc, line 60: Style nit: I feel like these headers stand out more (and thus help reading/skimming) if there's an empty line before and after, instead of only before.

djc (approved):
Nice!

rustls/src/client/builder.rs
```diff
@@ -18,6 +18,19 @@ use core::marker::PhantomData;
...
 impl ConfigBuilder<ClientConfig, WantsVerifier> {
     /// Choose how to verify server certificates.
+    ///
+    /// Using this function does not configure revocation.  If you wish to
+    /// configure revocation, instead use:
+    ///
+    /// ```diff
+    /// - .with_root_certificates(root_store)
+    /// + .dangerous()
+    /// + .with_custom_certificate_verifier(
```
cpu, line 28: Not a problem for this diff, but I wonder if we should add a `with_certificate_verifier_builder` option to `ConfigBuilder<ClientConfig, WantsVerifier>` that takes `ServerCertVerifierBuilder` and avoids needing `dangerous`. 

My thinking is that it's a bit unfortunate to have to go through `dangerous` if you're using the `WebPkiServerVerifier` - we've arranged its APIs to be safe. It's only providing a from-scratch `dyn ServerCertVerifier` that's dangerous.

It's a little bit unusual/awkward to accept a builder instead of a finalized struct, but I think we want to leave `ServerCertVerifierBuilder` producing `Arc<dyn ServerCertVerifier>` when it builds, so we can't take a concrete `ServerCertVerifierBuilder` type as an arg.

WDYT?

README.md
```diff
@@ -33,7 +33,7 @@ Rustls is a TLS library that aims to provide a good level of cryptographic secur
 requires no configuration to achieve that security, and provides no unsafe features or
 obsolete cryptography by default.
 
-## Current features
+## Current features (with default crate features)
```
cpu, line 36: I'm wondering if we can avoid saying "features" twice here, where one time we mean Cargo features and another more generically the functionality Rustls has:

```suggestion
## Current functionality (with default crate features)
```

README.md
```diff
@@ -92,6 +92,11 @@ x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Bi
 support WebAssembly.
 For more information, see [the supported `ring` target platforms][ring-target-platforms].
 
+By providing a custom implementation of the [`crate::crypto::CryptoProvider`] trait, you
+can replace all cryptography dependencies of rustls.  This is a route to being portable
+to a wider set of architectures and environments.  Specifying `default-features = false`
```
cpu, line 97: Maybe worth hinting at the FIPS type use case here too?

```suggestion
can replace all cryptography dependencies of rustls.  This is a route to being portable
to a wider set of architectures and environments, or compliance requirements.  Specifying `default-features = false`
```

rustls/src/crypto/mod.rs
```diff
@@ -45,19 +45,137 @@ pub use crate::rand::GetRandomFailed;
...
+/// Supply the provider when constructing your [`crate::ClientConfig`] or [`crate::ServerConfig`]:
+///
+/// - [`crate::ClientConfig::builder_with_provider()`]
+/// - [`crate::ServerConfig::builder_with_provider()`]
+///
+/// When creating and configuring a webpki-backed client or server certificate verifier, the provider
+/// is also needed to start:
```
cpu, line 67: editorial tweaking, take it or leave it :-)
```suggestion
/// When creating and configuring a webpki-backed client or server certificate verifier, a choice of provider
/// is also needed to start the configuration process:
```

rustls/src/crypto/mod.rs
```diff
@@ -45,19 +45,137 @@ pub use crate::rand::GetRandomFailed;
...
+///
+/// - [`crate::client::WebPkiServerVerifier::builder_with_provider()`]
+/// - [`crate::server::WebPkiClientVerifier::builder_with_provider()`]
+///
+/// # Making a custom `CryptoProvider`
+/// Naturally start with a type that `impl rustls::crypto::CryptoProvider`.
```
cpu, line 73: I think this reads a little awkwardly and the `impl` bit of syntax is probably well understood:
```suggestion
/// Naturally start with a type that implements `rustls::crypto::CryptoProvider`.
```

rustls/src/crypto/mod.rs
cpu: Agreed I think this might be clearer if worded to describe that the bytes should be sourced from a CSRNG

rustls/src/crypto/mod.rs
cpu: Thinking about Jsha's feedback from #1603, should we also mention what the crate has typically provided to help guide what a provider should offer in a minimal support set to avoid confusion?

> While you're working on documentation in this area, [load_private_key](https://docs.rs/rustls/0.22.0-alpha.4/rustls/crypto/trait.CryptoProvider.html#tymethod.load_private_key) should probably specify which private key formats an implementation should minimally accept.

rustls/src/server/common.rs
```diff
@@ -2,8 +2,7 @@ use crate::sign;
 
 use pki_types::CertificateDer;
 
-/// ActiveCertifiedKey wraps CertifiedKey and tracks OSCP and SCT state
-/// in a single handshake.
+/// ActiveCertifiedKey wraps CertifiedKey and tracks OSCP state in a single handshake.
```
cpu, line 5:
```suggestion
/// ActiveCertifiedKey wraps [sign::CertifiedKey] and tracks OSCP state in a single handshake.
```

cpu (approved):

rustls/src/client/builder.rs
cpu: >  I think we want to leave ServerCertVerifierBuilder producing Arc<dyn ServerCertVerifier> when it builds

Maybe that's not true. I'll fiddle a bit.

djc: I started on a commit today to rip out the with_roots() methods from the config builders in favor of a with_verifier(). That seems conceptually nicer (more composable) but it's also a bit of an ergonomics hit for the common case.

cpu: Here's what I was thinking: https://github.com/rustls/rustls/pull/1609

I left the `with_root_certificates` option but implemented it in terms of `with_webpki_verifeir`.

rustls/src/crypto/mod.rs
ctz: in those places it does say "the precise set of supported key types and parameters is defined by the selected `CryptoProvider`", but will reword this a bit

rustls/src/crypto/mod.rs
ctz: reworded

ctz:
> Thinking about Jsha's feedback from https://github.com/rustls/rustls/issues/1603, should we also mention what the crate has typically provided to help guide what a provider should offer in a minimal support set to avoid confusion?

For whatever reason I can't reply to this in-thread.

I'm not sure. I think I probably want to avoid prescribing some mandatory-to-implement key types / parameters here, and just concentrate on making the link 100% crystal clear and leave people to it.

If we were to write some MTI suggestions, I guess most people would put RSA in there, and probably ECDSA. But I think that disavows a completely reasonable `CryptoProvider` which deliberately don't have or want to support those.

rustls/src/client/builder.rs
ctz: I am just about to hit merge on this and then go on holiday -- do you want to update this new comment in #1609?

cpu:
> If we were to write some MTI suggestions, I guess most people would put RSA in there, and probably ECDSA. But I think that disavows a completely reasonable CryptoProvider which deliberately don't have or want to support those.

That seems like a reasonable justification to avoid it :+1:

rustls/src/client/builder.rs
cpu: Sure, sounds good. We can resolve this thread.

cpu (approved):

PR #1609, opened 2023-11-21
client: allow providing webpki cert verifier w/o dangerous

cpu:
Previously to supply a custom webpki-based server certificate verifier when building a client configuration the caller had to invoke `dangerous` to get access to a fn that can accept an `Arc<dyn verify::ServerCertVerifier>`. We did this because implementing a `ServerCertVerifier` from scratch leaves a lot of room for dangerous errors.

However, when providing a `WebPkiServerVerifier` constructed with `webpki::WebPkiServerVerifier::builder`, there is much less danger. We've arranged the builder and concrete type to be safe for general usage.

This commit changes the builder to return the concrete verifier type, and then adds a new `with_webpki_verifier` fn to the client config builder that accepts a `Arc<WebPkiServerVerifier>` without needing to go through `dangerous`. This will make the standard case of customizing the built-in webpki verifier not appear dangerous, while still requiring fully customized verifiers be provided through the dangerous API.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -5,7 +5,7 @@ use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::error::Error;
 use crate::key_log::NoKeyLog;
 use crate::suites::SupportedCipherSuite;
-use crate::webpki;
+use crate::webpki::{self, WebPkiServerVerifier};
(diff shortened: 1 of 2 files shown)


ctz (approved):
good idea!

djc (approved):

PR #1610, opened 2023-11-21
turn CertificatePayload type alias into a newtype

japaric:
being careful not to use it in user-facing / non-internal API like `{Client,Server}Config`

extracted out of #1597 as it can land ahead of time. I found this useful in that PR as `CertificateChain` becomes non-owning, gains a lifetime parameter and there's a need to go from the borrowed version to the owned version which fits nicely in a method on the newtype

### DIFF
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -2,16 +2,14 @@ use crate::client;
 use crate::enums::SignatureScheme;
 use crate::error::Error;
 use crate::limited_cache;
+use crate::msgs::handshake::CertificateChain;
 use crate::msgs::persist;
 use crate::sign;
 use crate::NamedGroup;
 use crate::ServerName;
 
-use pki_types::CertificateDer;
-
 use alloc::collections::VecDeque;
 use alloc::sync::Arc;
-use alloc::vec::Vec;
 use core::fmt;
 use std::sync::Mutex;
 
@@ -193,9 +191,12 @@ pub(super) struct AlwaysResolvesClientCert(Arc<sign::CertifiedKey>);
 impl AlwaysResolvesClientCert {
     pub(super) fn new(
         private_key: Arc<dyn sign::SigningKey>,
-        chain: Vec<CertificateDer<'static>>,
+        chain: CertificateChain,
     ) -> Result<Self, Error> {
-        Ok(Self(Arc::new(sign::CertifiedKey::new(chain, private_key))))
+        Ok(Self(Arc::new(sign::CertifiedKey::new(
+            chain.0,
+            private_key,
+        ))))
     }
 }
 
@@ -218,6 +219,7 @@ mod tests {
     use super::NoClientSessionStorage;
     use crate::client::ClientSessionStore;
     use crate::msgs::enums::NamedGroup;
+    use crate::msgs::handshake::CertificateChain;
     #[cfg(feature = "tls12")]
     use crate::msgs::handshake::SessionId;
     use crate::msgs::persist::Tls13ClientSessionValue;
@@ -253,7 +255,7 @@ mod tests {
                     SessionId::empty(),
                     Vec::new(),
                     &[],
-                    Vec::new(),
+                    CertificateChain::default(),
                     now,
                     0,
                     true,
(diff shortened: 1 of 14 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -1256,7 +1256,26 @@ impl ServerHelloPayload {
     }
 }
 
-pub(crate) type CertificatePayload = Vec<CertificateDer<'static>>;
+#[derive(Clone, Default, Debug)]
+pub struct CertificateChain(pub Vec<CertificateDer<'static>>);
+
+impl core::ops::Deref for CertificateChain {
```
djc, line 1262: Please import `Deref` at the top of the module and move this `impl` below the `Codec` impl.

rustls/src/client/tls12.rs
```diff
@@ -755,8 +755,8 @@ impl State<ClientConnectionData> for ExpectServerDone {
         // 4.
         if let Some(client_auth) = &st.client_auth {
             let certs = match client_auth {
-                ClientAuthDetails::Empty { .. } => Vec::new(),
-                ClientAuthDetails::Verify { certkey, .. } => certkey.cert.clone(),
+                ClientAuthDetails::Empty { .. } => <_>::default(),
```
djc, line 758: Let's just write `CertificateChain::default()` here?

rustls/src/client/handy.rs
```diff
@@ -253,7 +254,7 @@ mod tests {
                     SessionId::empty(),
                     Vec::new(),
                     &[],
-                    Vec::new(),
+                    <_>::default(),
```
djc, line 257: `CertificateChain::default()`, please (also below).

djc (approved):
This seems reasonable to me.

japaric:
addressed review comments

cpu (approved):
Nice!

rustls/src/client/tls12.rs
japaric: for what it's worth, rust-analyzer's autocompletion suggests `Default::default()` in scenarios like "functional update syntax" (shown below)
``` rust
let new = Struct {
    some_field: new_value,
    .. Default::default() // <- suggestion upon typing `.`
};
```

would that be acceptable? PR #1597 is going to use that syntax a lot so being able to use auto-completion (without having to manually patch it up to adhere to style conventions) would certainly help

cpu: That form seems reasonable to me 👍

japaric: > I usually associate "standalone" as a free-standing fn outside of an impl block, but that's the terminology the [Rust language docs for the impl keyword](https://doc.rust-lang.org/std/keyword.impl.html) use:

IME, *traits aside*, `Foo::constructor()` that returns `Foo` (or a type that contain `Foo` like `Option<Foo>` or `Result<Foo, _>`) is called a "constructor". `Foo::function()` that does not meet the preceding criteria is called "an associated function" (just like `Foo::CONSTANT` is "associated constant"; everything inside an `impl` block is an associated item). `default` is a member of the `Default` trait so I would probably call it a "trait function" because there's an orthogonal axis: an "inherent method" is a method in `impl Foo`; a "trait method" is a method in  `impl Trait for Foo`. 

I guess a catch all could be to say "do not elide names when calling a function using [fully-qualified syntax](https://doc.rust-lang.org/beta/reference/expressions/call-expr.html#disambiguating-function-calls)" (when I learned that syntax it was still called UFCS)

cpu: Thanks for the clarification, that's helpful. I'll let djc weigh in and then consider a CONTRIB update PR.

djc: I also prefer to avoid `Default::default()` for the simple reason that for `Type::default()` IMO the benefits (local clarity, especially when outside an IDE context, for example during review) outweigh the costs (potentially a type name that's slightly more verbose or maybe an extra import).

I agree that it would make for a good style guide addition to say that fully-qualified function calls should not elide the type.

cpu: > I agree that it would make for a good style guide addition to say that fully-qualified function calls should not elide the type.

https://github.com/rustls/rustls/pull/1620

Issue #1612, opened 2023-11-21
private key types and CryptoProvider


jsha:
#1608 added some great documentation for CryptoProvider. As part of that, `ConfigBuilder::with_client_auth_cert`, `with_single_cert`, and `with_single_cert_with_ocsp` say:

>     /// `key_der` is a DER-encoded private key -- the precise set of supported key
>    /// types and parameters is defined by the selected [`CryptoProvider`]'s `load_private_key()`
>    /// method.

The parameter is of type [`rustls_pki_types::PrivateKeyDer`](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/enum.PrivateKeyDer.html), which specifies three variants:

```
pub enum PrivateKeyDer<'a> {
    Pkcs1([PrivatePkcs1KeyDer](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.PrivatePkcs1KeyDer.html)<'a>),
    Sec1([PrivateSec1KeyDer](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.PrivateSec1KeyDer.html)<'a>),
    Pkcs8([PrivatePkcs8KeyDer](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.PrivatePkcs8KeyDer.html)<'a>),
}
```

So I think it's more accurate to say that the contents can be at most one of these three encodings, but a given crypto provider may choose to support only a subset. There's no type-friendly way to represent some unknown fourth encoding.

As an example: for an HSM provider, the private key details might not be encodable as DER at all. Instead, the input might be a config file listing parameters for a PKCS#11 library to communicate with its backend. Or those settings might come from environment variables instead of a config file.

I think `with_client_auth_cert`, `with_single_cert`, and `with_single_cert_with_ocsp` need some version that takes an `Arc<dyn SigningKey>` instead of a rustls_pki_types::PrivateKeyDer.

ctz:
> So I think it's more accurate to say that the contents can be at most one of these three encodings

Right, so there's (at least!) three dimensions here:

- the encoding, of PKCS#1, sec1, or PKCS#8
- the type of key (PKCS#1 is RSA-specific, sec1 is EC-specific, PKCS#8 is general)
- the space of supported key parameters (size of RSA modulus, number of RSA exponents, EC curve shape and encoding, etc.)

[...]

PR #1613, opened 2023-11-21
doc: update docs for SigningKey

jsha:
Explain why the implementors section for SigningKey is empty, where SigningKey comes from, and what it is consumed by.

Update the functions that document encodings for loading private keys so they are more specific and concrete.

Part of #1612.

### DIFF
--- a/rustls/src/crypto/signer.rs
+++ b/rustls/src/crypto/signer.rs
@@ -12,6 +12,46 @@ use core::fmt::Debug;
 ///
 /// This interface is used by rustls to use a private signing key
 /// for authentication.  This includes server and client authentication.
+///
+/// Objects of this type are always used within Rustls as
+/// `Arc<dyn SigningKey>`. There are no concrete public structs in Rustls
+/// that implement this trait.
+///
+/// There are two main ways to get a signing key:
+///
+///  - [`CryptoProvider::load_private_key()`], or
+///  - some other method outside of the `CryptoProvider` extension trait,
+///    for instance:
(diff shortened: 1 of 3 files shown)


rustls/src/crypto/signer.rs
```diff
@@ -12,6 +12,21 @@ use core::fmt::Debug;
 ///
 /// This interface is used by rustls to use a private signing key
 /// for authentication.  This includes server and client authentication.
+///
+/// Objects of this type are always and only provided by
```
ctz, line 16: You can actually just make these outside of `load_private_key()`, eg, see https://docs.rs/rustls/0.22.0-alpha.4/rustls/crypto/ring/sign/fn.any_ecdsa_type.html and co. You would do this if you wanted to minimise the amount of stuff linked into your binary, or wanted to have a single server with multiple sets of keys and certificates (via, https://docs.rs/rustls/0.22.0-alpha.4/rustls/server/struct.ResolvesServerCertUsingSni.html or some other implementation of https://docs.rs/rustls/0.22.0-alpha.4/rustls/server/trait.ResolvesServerCert.html)

ctz: Which leads to the question: what's the point of `load_private_key()`? It's to regularise the dependency from the `ConfigBuilder` states on the crypto provider.

jsha: Aha, thanks for the pointers. Now I see more of the picture. Will update.

djc (approved):

cpu (approved):
Thank you, this is a nice update :books:

PR #1615, opened 2023-11-21
doc: replace "nb." with "Note:"

jsha:
"nota bene" (mark well) is jargon that we don't need.

Note: I wouldn't bother with this but I noticed it in the public documentation for KeyBlockShape. I figure it's valuable to use less jargon in the public docs when possible, and I might as well tweak the internal-facing uses as well.


djc (approved):

cpu (approved):
> "nota bene" (mark well) is jargon that we don't need.

Agreed, thanks for fixing this up. I couldn't spot any other instances outside the ones fixed in this branch.

PR #1616, opened 2023-11-22
Rename SignError to InvalidKeyError

djc:


rustls/src/crypto/ring/sign.rs
```diff
@@ -392,15 +394,15 @@ impl Debug for Ed25519Signer {
 
 /// Errors while signing
 #[derive(Debug)]
-pub struct SignError(());
+pub struct InvalidKeyError(());
 
-impl fmt::Display for SignError {
+impl fmt::Display for InvalidKeyError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
         f.write_str("sign error")
```
cpu, line 401:
```suggestion
        f.write_str("error constructing key")
```

rustls/src/crypto/ring/sign.rs
```diff
@@ -392,15 +394,15 @@ impl Debug for Ed25519Signer {
 
 /// Errors while signing
```
cpu, line 395:
```suggestion
/// An error produced when constructing a key
```

cpu (approved):
This LGTM but I think there are ~2 places that could use a tweak before merging. Thank you

cpu (approved):
Thanks!

PR #1618, opened 2023-11-22
CI: run clippy on the entire workspace

japaric:
also fix clippy warnings in the `ci-bench` package

I noticed that `rust-examples` and `ci-bench` were not covered by the clippy CI jobs so I added them and then I figured running clippy on the entire workspace would be equivalent and shorter


ci-bench/src/main.rs
```diff
@@ -275,10 +275,9 @@ fn find_suite(
     all: &[rustls::SupportedCipherSuite],
     name: CipherSuite,
 ) -> rustls::SupportedCipherSuite {
-    all.iter()
+    *all.iter()
```
djc, line 278: Would be nice to re-order the commits so that the fix comes before the check.

djc (approved):
Thanks!

japaric:
reordered the commits as requested

.github/workflows/build.yml
```diff
@@ -320,11 +320,10 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      - run: cargo clippy --locked --package rustls --all-features --all-targets -- --deny warnings
+        # because examples enable rustls' features, `--workspace --no-default-features` is not
+        # the same as `--package rustls --no-default-features` so run it separately
       - run: cargo clippy --locked --package rustls --no-default-features --all-targets -- --deny warnings
-      - run: cargo clippy --locked --manifest-path=connect-tests/Cargo.toml --all-features --all-targets -- --deny warnings
-      - run: cargo clippy --locked --manifest-path=fuzz/Cargo.toml --all-features --all-targets -- --deny warnings
```
cpu, line 326: I think this invocation for `--manifest-path=fuzz/Cargo.toml` needs to stay. That crate is a little funky in that it's not included in the top-level `Cargo.toml` workspace members.

If i introduce an issue into one of the fuzzers running `cargo clippy --locked --workspace --all-features --all-targets -- --deny warnings` doesn't flag it, but running `cargo clippy --locked --manifest-path=fuzz/Cargo.toml --all-features --all-targets -- --deny warnings` does.

.github/workflows/build.yml
```diff
@@ -338,11 +337,8 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           components: clippy
-      - run: cargo clippy --locked --package rustls --all-features --all-targets
       - run: cargo clippy --locked --package rustls --no-default-features --all-targets
-      - run: cargo clippy --locked --manifest-path=connect-tests/Cargo.toml --all-features --all-targets
-      - run: cargo clippy --locked --manifest-path=fuzz/Cargo.toml --all-features --all-targets
```
cpu, line 344: ditto here for leaving the special-case for the fuzzer crate.

.github/workflows/build.yml
japaric: thanks for catching that. fixed

cpu (approved):
Thanks!

Issue #1614, opened 2023-11-21
api: two import paths for `rustls::sign` items


jsha:
As of the latest alpha, both of these modules exist:

 - [`rustls::sign`](https://docs.rs/rustls/0.22.0-alpha.4/rustls/sign/index.html)
 - [`rustls::crypto::signer`](https://docs.rs/rustls/0.22.0-alpha.4/rustls/crypto/signer/index.html)
 
 They each have the same contents: CertifiedKey, Signer, and SigningKey.
 
 Since `rustls::sign` exists in 0.21, and `rustls::crypto::signer` does not, I'm guessing we'd like to keep these items exported under `rustls::sign` for 0.22, and unexport the `rustls::crypto::signer` module?

cpu:
> Since rustls::sign exists in 0.21, and rustls::crypto::signer does not, I'm guessing we'd like to keep these items exported under rustls::sign for 0.22, and unexport the rustls::crypto::signer module?

I think it might make more sense to remove the `rustls::sign` export. In https://github.com/rustls/rustls/pull/1477 we decided that re-exports are for "paved path" exports that we expect every user will need. I don't think the types under `crypto::signer::*` necessarily meet that bar.

Additionally, if we keep `rustls::sign` and unexport `rustls::crypto::signer` then `signer` becomes an outlier compared to the other `crypto` submodules that are all `pub mod`'s.

djc:
I agree that `rustls::crypto::signer` makes more sense with the current setup.

cpu:
I put up https://github.com/rustls/rustls/pull/1619 to drop the re-export if folks agree with that direction.

jsha:
My takeaway from #1619 was not that re-exports are for "paved path" items, but that _top level_ re-exports are. And also that reducing churn when practical is good.

[Quoting @djc](https://github.com/rustls/rustls/pull/1477#issuecomment-1720979780):
> For the others, I really don't see removing the top-level re-exports as a good idea. These have been at the top-level forever, so we'd be putting every downstream user through a bunch of churn for the sole goal of removing the re-exports.

This case seems similar. These are three types that have been in the API for a while. Internally we moved them to be consistent with a refactoring, but that doesn't mean we need to move the externally visible location.

djc:
That's also fair. 🤷

cpu:
[...]

PR #1619, opened 2023-11-22
lib: remove `rustls::sign` re-export

cpu:
In general it's confusing if there are two import paths for types and we prefer to only offer re-exports for "paved path" types that are expected to be used by all consumers.

The `rustls::sign` types re-exported from `rustls::crypto::signer` don't meet this bar, so this commit removes the re-export, preferring to use the `rustls::crypto::signer` import path throughout.

Resolves https://github.com/rustls/rustls/issues/1614


cpu:
Consensus is leaning towards solving this in the opposite direction, keeping the `rustls::sign` re-export and removing the `pub signer` export. Closing this to rework.

PR #1622, opened 2023-11-23
RootCertStore: better Debug impl

jsha:
The derive(Debug) impl was printing the subject and subjectpublickeyinfo for every single trust anchor in the root store, which made it very difficult to read other Debug output that happened to contain a RootCertStore (or a WebPkiServer, which contains a RootCertStore). For instance this made the Debug output for ClientConfig extremely long.

In the custom Debug impl, abbreviate the list of roots to simply say how many of them there are.

### DIFF
--- a/rustls/src/webpki/anchors.rs
+++ b/rustls/src/webpki/anchors.rs
@@ -1,4 +1,5 @@
 use alloc::vec::Vec;
+use alloc::{fmt, format};
 
 use pki_types::{CertificateDer, TrustAnchor};
 use webpki::anchor_from_trusted_cert;
@@ -10,7 +11,7 @@ use crate::{DistinguishedName, Error};
 
 /// A container for root certificates able to provide a root-of-trust
 /// for connection authentication.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct RootCertStore {
     /// The list of roots.
     pub roots: Vec<TrustAnchor<'static>>,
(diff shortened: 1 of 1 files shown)


rustls/src/webpki/server_verifier.rs
```diff
@@ -1,5 +1,7 @@
 #[cfg(feature = "logging")]
 use crate::log::trace;
+use alloc::fmt;
+use alloc::format;
```
djc, line 4: Nit: merge these in a single `use` statement.

rustls/src/webpki/server_verifier.rs
```diff
@@ -122,6 +123,18 @@ pub struct WebPkiServerVerifier {
     supported: WebPkiSupportedAlgorithms,
 }
 
+impl fmt::Debug for WebPkiServerVerifier {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("WebPkiServerVerifier")
+            .field("roots", &format!("({} roots)", &self.roots.len()))
```
djc, line 129: Should we change this on the `RootCertStore` `Debug` impl instead of on the `WebPkiServerVerifier` impl? Seems more principled -- if the long list of roots is a problem for `WebPkiServerVerifier` it should equally be a problem for `WebPkiClientVerifier` and other types that contain a `RootCertStore`.

cpu: Agreed, I think that would make the most sense.

Jsha mentioned:
>  it seemed useful to have some impl that actually would print the full contents of the RootCertStore

I recently added the [RootCertStore::subjects](https://github.com/rustls/rustls/blob/de41c70959844b5411c5018147a8a48c7817e387/rustls/src/webpki/anchors.rs#L80-L92) fn that returns `Vec<DistinguishedName>`. Since `DistinguishedName` impl's `Debug` that seems like a pretty good way to get this kind of output if you needed it.

djc: Sorry, missed that note.

jsha: I'll move this to the `RootCertStore`. 👍🏻

cpu (approved):
LGTM. Would you mind updating the PR description to reflect that the `Debug` impl was moved to `RootCertStore`?

djc (approved):

cpu:
> Updated the PR description!

Thanks!

PR #1623, opened 2023-11-23
doc: CryptoProvider defaults can be overridden

jsha:
It was surprising to me that builder_with_provider could set a CryptoProvider, and then with_cipher_suites could choose implementations from a different CryptoProvider. I've tried to document things to make that a little less surprising.

Related to #1603


cpu (approved):
These doc changes make sense to me. It sounds like the necessity might fall away if the crypto provider and builders are reworked but for the current state in-branch it seems like an improvement :+1:

djc (approved):

PR #1627, opened 2023-11-24
fix warnings in -default +ring build

japaric:
I found these while running `cargo c -p rustls --no-default-features --features ring` to test some WIP work

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -110,3 +110,5 @@ jobs:
 
       - name: Check feature powerset
         run: cargo hack check --feature-powerset --no-dev-deps
+        env:
+          RUSTFLAGS: --deny warnings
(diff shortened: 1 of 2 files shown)


cpu (approved):
Nice catch. Thanks for the fix.

It looks like the feature powerset task in the `daily-tests.yml` has been seeing these warnings, but not treating them as errors.

Would you mind updating that job to use `--deny warnings` to prevent regression?

djc (approved):

djc:
We should also bump the noise threshold on the benchmark jobs, especially for aws-lc-rs.

japaric:
@cpu done. (it doesn't seem like `cargo hack` accepts `-- --deny warnings` like `cargo clippy` does so I used RUSTFLAGS)

cpu (approved):
Great, ty

Issue #224, opened 2019-01-25
Allow unused protocol versions to be discarded at link-time


ctz:
If you were only interested in using TLS1.3, it would be ideal if all the TLS1.2 code were discarded at link-time. For code size, complexity and very convincing attack surface reduction reasons.

Ralith:
This would be awesome for QUIC!

briansmith:
> If you were only interested in using TLS1.3, it would be ideal if all the TLS1.2 code were discarded at link-time. For code size, complexity and very convincing attack surface reduction reasons.

I don't think we need to worry too much about being able to discard TLS-1.3-specific code at link time. I suggest we optimize this work for the case where TLS 1.3 is enabled, and focus on making sure as little TLS-1.2-specific code gets linked in as practical. 

The "tls12" feature provides a way for people to statically avoid the vast majority of the TLS-1.2-specific code. It would be nice to have `SupportedProtocolVersion` become a trait, where the code protected by `#[cfg(feature = "tls12")]` would be moved to the TLS 1.2 implementations of the methods of that trait. But, maybe the "tls12" feature is good enough.

ctz:
> maybe the "tls12" feature is good enough.

Really, cargo features as a means for detailed conditional compilation doesn't work well. This is quite a mess:

```
$ ack 'feature = "tls12"' | wc -l
110
```

And it grows ever more complex.

djc:
I wonder how hard it would be to turn `SupportedCipherSuite` into a trait instead of an enum, I think that would help quite a bit with this.

PR #1628, opened 2023-11-24
crypto: rework CryptoProvider as struct

cpu:
This branch replaces the existing `CryptoProvider` trait with a `CryptoProvider` struct. This has several advantages:

* it consolidates all of the cryptography related settings into one API surface, the `CryptoProvider` struct members. Previously the provider had methods to suggest default ciphersuites, key exchanges etc, but the builder API methods could override them in confusing ways.
* it allows removing the `WantsCipherSuites` and `WantsKxGroups` builder states - the "safe defaults" are automatically supplied by the choice of a crypto provider. Customization is achieved by overriding the provider's struct fields. Having fewer builder states makes the API easier to understand and document.
* it makes customization easier: the end user can rely on ["struct update syntax"](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) to only specify fields values for the required
  customization, and defer the rest to an existing `CryptoProvider`. 
  
Achieving this requires a couple of additional changes:

* The `fill_random` and `load_private_key` fns of the `CryptoProvider` trait are broken out into separate traits, `SecureRandom` and `KeyProvider`, so that we can hold separate `&dyn` references to them in the new `CryptoProvider` struct.
* The cipher suite and key exchange groups are now expressed as `Vec` elements. This avoids imposing a `&'static` lifetime that would preclude runtime customization (e.g. the tls*-mio examples that build the list of ciphersuites at runtime based on command line flags).
* As a result of the `Vec` members we can no longer offer the concrete `CryptoProvider`s as `static` members of their respective modules. Instead we add `pub fn provider() -> CryptoProvider` methods to the `ring` and `aws-lc-rs` module that construct the `CryptoProvider` with the safe defaults, ready for further customization.

Finally, to avoid an unnecessary `Result` unwrap when using the default crypto provider and default protocol versions, or customized protocol versions, the `ClientConfig` and `ServerConfig` default `builder` fns transition directly to `WantsVerifier` using default safe protocol versions. Customization of only protocol versions can be done with new `builder_with_protocol_versions` fns that transition to `WantsVerifier` without needing an unwrap. Using a custom provider requires handling the `Result` that may occur if either the default safe protocol versions or provided custom versions can't be supported by the provider's ciphersuites.

Updates #1372, #1603

### DIFF
--- a/examples/src/bin/server_acceptor.rs
+++ b/examples/src/bin/server_acceptor.rs
@@ -225,8 +225,6 @@ impl TestPki {
         // based on the ClientHello (e.g. selecting a different certificate, or customizing
         // supported algorithms/protocol versions).
         let mut server_config = ServerConfig::builder()
-            .with_safe_default_protocol_versions()
-            .unwrap()
             .with_client_cert_verifier(verifier)
             .with_single_cert(
                 vec![self.server_cert_der.clone()],
--- a/examples/src/bin/server_acceptor.rs
+++ b/examples/src/bin/server_acceptor.rs
@@ -225,7 +225,8 @@ impl TestPki {
         // based on the ClientHello (e.g. selecting a different certificate, or customizing
         // supported algorithms/protocol versions).
         let mut server_config = ServerConfig::builder()
-            .with_safe_defaults()
+            .with_safe_default_protocol_versions()
+            .unwrap()
             .with_client_cert_verifier(verifier)
             .with_single_cert(
                 vec![self.server_cert_der.clone()],
--- a/rustls/examples/internal/bogo_shim_impl.rs
+++ b/rustls/examples/internal/bogo_shim_impl.rs
@@ -6,7 +6,7 @@
 
 use rustls::client::danger::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
 use rustls::client::{ClientConfig, ClientConnection, Resumption, WebPkiServerVerifier};
-use rustls::crypto::SupportedKxGroup;
+use rustls::crypto::{CryptoProvider, SupportedKxGroup};
 use rustls::internal::msgs::codec::Codec;
 use rustls::internal::msgs::persist::ServerSessionValue;
 use rustls::server::danger::{ClientCertVerified, ClientCertVerifier};
@@ -19,9 +19,9 @@ use rustls::{
 };
 
 #[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
-use rustls::crypto::{aws_lc_rs as provider, aws_lc_rs::AWS_LC_RS as PROVIDER};
+use rustls::crypto::aws_lc_rs as provider;
 #[cfg(feature = "ring")]
-use rustls::crypto::{ring as provider, ring::RING as PROVIDER};
+use rustls::crypto::ring as provider;
 
 use base64::prelude::{Engine, BASE64_STANDARD};
 use pki_types::{CertificateDer, PrivateKeyDer, ServerName, UnixTime};
@@ -212,9 +212,12 @@ impl DummyClientAuth {
     fn new(mandatory: bool) -> Self {
         Self {
             mandatory,
-            parent: WebPkiClientVerifier::builder_with_provider(load_root_certs(), PROVIDER)
-                .build()
-                .unwrap(),
+            parent: WebPkiClientVerifier::builder_with_provider(
+                load_root_certs(),
+                provider::default_provider().into(),
+            )
+            .build()
+            .unwrap(),
         }
     }
 }
@@ -274,9 +277,12 @@ struct DummyServerAuth {
 impl DummyServerAuth {
     fn new() -> Self {
         DummyServerAuth {
-            parent: WebPkiServerVerifier::builder_with_provider(load_root_certs(), PROVIDER)
-                .build()
-                .unwrap(),
+            parent: WebPkiServerVerifier::builder_with_provider(
+                load_root_certs(),
+                provider::default_provider().into(),
+            )
+            .build()
+            .unwrap(),
         }
     }
 }
@@ -495,14 +501,18 @@ fn make_server_cfg(opts: &Options) -> Arc<ServerConfig> {
         provider::ALL_KX_GROUPS.to_vec()
     };
 
-    let mut cfg = ServerConfig::builder_with_provider(PROVIDER)
-        .with_safe_default_cipher_suites()
-        .with_kx_groups(&kx_groups)
-        .with_protocol_versions(&opts.supported_versions())
-        .unwrap()
-        .with_client_cert_verifier(client_auth)
-        .with_single_cert_with_ocsp(cert.clone(), key, opts.server_ocsp_response.clone())
-        .unwrap();
+    let mut cfg = ServerConfig::builder_with_provider(
+        CryptoProvider {
+            kx_groups,
+            ..provider::default_provider()
+        }
+        .into(),
+    )
+    .with_protocol_versions(&opts.supported_versions())
+    .unwrap()
+    .with_client_cert_verifier(client_auth)
+    .with_single_cert_with_ocsp(cert.clone(), key, opts.server_ocsp_response.clone())
+    .unwrap();
 
     cfg.session_storage = ServerCacheWithResumptionDelay::new(opts.resumption_delay);
     cfg.max_fragment_size = opts.max_fragment;
@@ -623,13 +633,17 @@ fn make_client_cfg(opts: &Options) -> Arc<ClientConfig> {
         provider::ALL_KX_GROUPS.to_vec()
     };
 
-    let cfg = ClientConfig::builder_with_provider(PROVIDER)
-        .with_safe_default_cipher_suites()
-        .with_kx_groups(&kx_groups)
-        .with_protocol_versions(&opts.supported_versions())
-        .expect("inconsistent settings")
-        .dangerous()
-        .with_custom_certificate_verifier(Arc::new(DummyServerAuth::new()));
+    let cfg = ClientConfig::builder_with_provider(
+        CryptoProvider {
+            kx_groups,
+            ..provider::default_provider()
+        }
+        .into(),
+    )
+    .with_protocol_versions(&opts.supported_versions())
+    .expect("inconsistent settings")
+    .dangerous()
+    .with_custom_certificate_verifier(Arc::new(DummyServerAuth::new()));
 
     let mut cfg = if !opts.cert_file.is_empty() && !opts.key_file.is_empty() {
         let cert = load_cert(&opts.cert_file);
--- a/rustls/src/builder.rs
+++ b/rustls/src/builder.rs
@@ -19,19 +19,18 @@ use std::sync::Arc;
 /// For settings besides these, see the fields of [`ServerConfig`] and [`ClientConfig`].
 ///
 /// The usual choice for protocol primitives is to call
-/// [`crate::ClientConfig::builder`]/[`ServerConfig::builder`] and [`ConfigBuilder::with_safe_default_protocol_versions`],
-/// which will use rustls' default cryptographic provider and default protocol versions.
+/// [`crate::ClientConfig::builder`]/[`ServerConfig::builder`]
+/// which will use rustls' default cryptographic provider and safe defaults for ciphersuites and
+/// supported protocol versions.
 ///
 /// ```
 /// # #[cfg(feature = "ring")] {
 /// use rustls::{ClientConfig, ServerConfig};
 /// ClientConfig::builder()
-///     .with_safe_default_protocol_versions()
 /// //  ...
 /// # ;
 ///
 /// ServerConfig::builder()
-///     .with_safe_default_protocol_versions()
 /// //  ...
 /// # ;
 /// # }
@@ -42,15 +41,13 @@ use std::sync::Arc;
 /// ```no_run
 /// # #[cfg(feature = "ring")] {
 /// # use rustls::ServerConfig;
-/// ServerConfig::builder()
-///     .with_protocol_versions(&[&rustls::version::TLS13])
-///     .unwrap()
+/// ServerConfig::builder_with_protocol_versions(&[&rustls::version::TLS13])
 /// //  ...
 /// # ;
 /// # }
 /// ```
 ///
-/// Overriding a default introduces a `Result` that must be unwrapped,
+/// Overriding the default cryptographic provider introduces a `Result` that must be unwrapped,
 /// because the config builder checks for consistency of the choices made. For instance, it's an error to
 /// configure only TLS 1.2 cipher suites while specifying that TLS 1.3 should be the only supported protocol
 /// version.
@@ -81,8 +78,6 @@ use std::sync::Arc;
 /// # use rustls::ClientConfig;
 /// # let root_certs = rustls::RootCertStore::empty();
 /// ClientConfig::builder()
-///     .with_safe_default_protocol_versions()
-///     .unwrap()
 ///     .with_root_certificates(root_certs)
 ///     .with_no_client_auth();
 /// # }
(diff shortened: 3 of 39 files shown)


examples/src/bin/server_acceptor.rs
```diff
@@ -225,7 +225,8 @@ impl TestPki {
         // based on the ClientHello (e.g. selecting a different certificate, or customizing
         // supported algorithms/protocol versions).
         let mut server_config = ServerConfig::builder()
-            .with_safe_defaults()
+            .with_safe_default_protocol_versions()
+            .unwrap()
             .with_client_cert_verifier(verifier)
             .with_single_cert(
                 vec![self.server_cert_der.clone()],
```
cpu, line 232: I think one downside of this branch is that the paved-road where you create a builder with the default crypto provider, and then use `with_safe_default_protocol_versions()` to use default protocol versions requires an `unwrap` before continuing the build.

In the case where the default provider is used we have confidence there won't be an error result: the default ciphersuites from *ring* and aws-lc-rs are compatible with all protocol versions. The potential for error only exists when `builder_with_provider()` is used.

Perhaps a more ergonomic solution is to have:

* `ServerConfig::builder()` -> builds a server config with the default crypto provider and the safe protocol defaults, transitions to `WantsVerifier` state without needing an unwrap.
* `ServerConfig::builder_with_protocol_versions()` -> builds a server config with the default crypto provider, and a specified set of protocol versions, transitions to `WantsVerifier` without needing an unwrap.
* `ServerConfig::builder_with_provider()` -> builds a server config with a specified provider, transitions to `WantsVersions` where both `with_safe_default_protocol_versions()` and `with_protocol_versions()` allow customization of protocol versions, but requires handling a `Result` to get to `WantsVerifier`

In general this relates to the "Multiple Unwraps" problem called out in https://github.com/rustls/rustls/issues/1372

cpu: I implemented this approach in [572f22b](https://github.com/rustls/rustls/pull/1628/commits/572f22b7b433485eeb9cb0454857a709d7fa9bbb) and I think it's an improvement.

If preferred I could break that out into a separate PR. In the meantime I'll resolve this thread.

rustls/examples/internal/bogo_shim_impl.rs
```diff
@@ -19,9 +19,9 @@ use rustls::{
 };
 
 #[cfg(all(not(feature = "ring"), feature = "aws_lc_rs"))]
-use rustls::crypto::{aws_lc_rs as provider, aws_lc_rs::AWS_LC_RS as PROVIDER};
+use rustls::crypto::aws_lc_rs as primary_provider;
```
djc, line 22: I think I'd just name this `provider`? `primary_provider` seems like a weird notion since I don't think there is another provider in play in the bogo shim at the same time? I guess you're trying to disambiguate from the `provider()` function but functions and modules live in a different namespace so `provider::provider()` should Just Work IIRC.

rustls/src/client/client_conn.rs
```diff
@@ -226,9 +216,7 @@ pub enum Tls12Resumption {
 impl Clone for ClientConfig {
     fn clone(&self) -> Self {
         Self {
-            cipher_suites: self.cipher_suites.clone(),
-            kx_groups: self.kx_groups.clone(),
-            provider: self.provider,
+            provider: Arc::<CryptoProvider>::clone(&self.provider),
```
djc, line 221: Do we really need the turbofish here? That seems surprising to me. If so maybe just prefer `self.provider.clone()` instead?

rustls/src/crypto/mod.rs
```diff
@@ -149,39 +140,33 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
...
     /// These are used for both certificate chain verification and handshake signature verification.
-    ///
-    /// This is called by [`crate::ConfigBuilder::with_root_certificates()`],
-    /// [`crate::server::WebPkiClientVerifier::builder_with_provider()`] and
-    /// [`crate::client::WebPkiServerVerifier::builder_with_provider()`].
-    fn signature_verification_algorithms(&self) -> WebPkiSupportedAlgorithms;
+    pub signature_verification_algorithms: WebPkiSupportedAlgorithms,
+
+    /// Source of cryptographically secure random numbers.
+    pub secure_random: &'static dyn SecureRandom,
```
djc, line 166: Might be nice to experiment with changing this to be a separate commit? That is, I think the change from `CryptoProvider` going from is-a `SecureRandom` to `has-a` `SecureRandom` could I think be orthogonal to `CryptoProvider` going from `trait` to `struct` -- it could return the `&'static dyn SecureRandom` in a method. (Same for `KeyProvider`, of course.)

rustls/src/crypto/ring/mod.rs
```diff
@@ -29,26 +29,20 @@ pub(crate) mod tls13;
 /// A `CryptoProvider` backed by the [*ring*] crate.
 ///
 /// [*ring*]: https://github.com/briansmith/ring
-pub static RING: &dyn CryptoProvider = &Ring;
+pub fn provider() -> CryptoProvider {
```
djc, line 32: I wonder if we should give this a more opinionated name? Of course this is not the only possible `CryptoProvider` instantiation based on *ring* cryptography, merely the one with default cipher suites, key exchange groups and signature algorithms. Maybe `default_provider()`?

djc (approved):
I think this is looking pretty good.

rustls/examples/internal/bogo_shim_impl.rs
cpu: > I guess you're trying to disambiguate from the provider() function

I think I picked this up from another module that was using `primary_provider` and thought it did a good job distinguishing from the provider() function, but as you mention maybe that's not necessary. 

I'll adjust to just use `primary` throughout :+1:

rustls/src/client/client_conn.rs
cpu: IIRC `clippy` was flagging this without the turbofish but I will verify.

rustls/src/crypto/mod.rs
cpu: > That is, I think the change from CryptoProvider going from is-a SecureRandom to has-a SecureRandom could I think be orthogonal to CryptoProvider going from trait to struct

Sure, that makes sense :+1: I will split it out.

rustls/src/crypto/ring/mod.rs
cpu: `default_provider` sounds good to me. I agree it would be nice to emphasize the intent is that this is a starting point for sensible defaults that can be further customized.

rustls/src/client/client_conn.rs
cpu: Yeah:
```
error: using `.clone()` on a ref-counted pointer
   --> rustls/src/client/client_conn.rs:221:23
    |
221 |             provider: self.provider.clone(),
    |                       ^^^^^^^^^^^^^^^^^^^^^ help: try: `Arc::<crypto::CryptoProvider>::clone(&self.provider)`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_ref_ptr
   --> rustls/src/lib.rs:278:5
    |
278 |     clippy::clone_on_ref_ptr,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `rustls` (lib) due to previous error
```

rustls/src/crypto/ring/mod.rs
briansmith: > `default_provider` sounds good to me. I agree it would be nice to emphasize the intent is that this is a starting point for sensible defaults that can be further customized.

I don't object to that. I expect that some packagers of Rustls (e.g. Linux distros) will change the default provider to something that isn't *ring*, like OpenSSL Or Windows CNG. And then people who want/need to use *ring* (e.g. because of FIPS, in the future) will need a way to force the use of *ring*. So ideally there would be both a way to force the use of *ring* and a `default_provider()`.

For the same reason, enabling "FIPS mode" shouldn't affect which provider is used; the option should only change how the provider is configured. So, for example if one wants to use AWS-LC in FIPS mode, they should choose AWS-LC *and* also choose "FIPS mode."

rustls/examples/internal/bogo_shim_impl.rs
cpu: Fixed.

rustls/src/crypto/mod.rs
cpu: I split out two additional commits for this, one for each new trait.

rustls/src/crypto/ring/mod.rs
cpu: I've switched to `default_provider` for the constructors.

>  So ideally there would be both a way to force the use of ring and a default_provider().

That's certainly possible. Using `ClientConfig::builder` or `ServerConfig::builder` will default to *ring*, but if you wanted to make the choice explicit it's just sugar over `builder_with_provider(rustls::crypto::ring::default_provider().into())`.

rustls/src/client/client_conn.rs
djc: Yeah... I don't care much for that particular lint, but okay.

rustls/src/builder.rs
```diff
@@ -1,62 +1,55 @@
...
+use std::sync::Arc;
 
 /// A [builder] for [`ServerConfig`] or [`ClientConfig`] values.
 ///
 /// To get one of these, call [`ServerConfig::builder()`] or [`ClientConfig::builder()`].
 ///
 /// To build a config, you must make at least three decisions (in order):
 ///
-/// - Which protocol primitives should be supported (cipher suites, key exchange groups, protocol versions)?
+/// - Which cryptographic provider to use, such as [`ring::provider`]?
```
jsha, line 16: The new API reduces the number of "must decides." In particular the choice of protocol primitives went away. Now I think it's more accurate to say:

```
/// To build a config, you must make at least two decisions (in order):
///
/// - How should this client or server verify certificates provided by its peer?
/// - What certificates should this client or server present to its peer?
```

To some extent you're "making a decision" by choosing `ClientConfig::builder` vs `ClientConfig::builder_with_XXX`. But that doesn't have to be part of this list. The goal of this list is to orient people to "what are the must-call methods I use to get to the point where I can finish building a config."

jsha (commented):
Thanks for working on this! Looks great.

The `unwrap` for checking cipher suite / TLS version compatibility still feels awkward, even with introducing the `builder_with_protocol_versions`.

There was some prior discussion when the ConfigBuilder was first introduced, starting here: https://github.com/rustls/rustls/pull/564#issuecomment-826049679

The suggestion was to have different types for TLS 1.2 cipher suites vs TLS 1.3 cipher suites. In fact we have that already, we just wrap them in an enum so they can share a type:

```rust
pub enum SupportedCipherSuite {
    Tls12(&'static Tls12CipherSuite),
    Tls13(&'static Tls13CipherSuite),
}
```

What if `CryptoProvider` had two fields:

```rust
tls12_cipher_suites: Vec<&'static Tls12CipherSuite>,
tls13_cipher_suites: Vec<&'static Tls13CipherSuite>,
```

And the way to disable TLS 1.2 would be to provide an empty `Vec` for `tls12_cipher_suites`. That way we could do away with both the `unwrap()` and `builder_with_protocol_versions`.

Under the hood, the builder could then bundle those up into the enum `SupportedCipherSuite` to minimize unnecessary diffs.

djc:
> And the way to disable TLS 1.2 would be to provide an empty `Vec` for `tls12_cipher_suites`.

This seems like a good idea to me.

I think I advocated at the time that there was no need to specify protocol versions anyway because this was implicitly expressed by the supply of cipher suites. I think that was deemed not explicit enough but just wanted to mention it here for completeness.

But most of this discussion seems out of scope for this specific PR?

rustls/tests/api.rs
```diff
@@ -5563,30 +5589,23 @@ impl rustls::crypto::SecureRandom for FaultyRandom {
...
-    };
 
     let client_config = finish_client_config(
         KeyType::Rsa,
-        rustls::ClientConfig::builder_with_provider(&TEST_PROVIDER).with_safe_defaults(),
+        rustls::ClientConfig::builder_with_provider(
+            CryptoProvider {
+                secure_random: &FAULTY_RANDOM,
+                ..provider::default_provider()
+            }
```
ctz, line 5589: _so_ much nicer!

examples/src/bin/server_acceptor.rs
```diff
@@ -225,8 +225,6 @@ impl TestPki {
         // based on the ClientHello (e.g. selecting a different certificate, or customizing
         // supported algorithms/protocol versions).
         let mut server_config = ServerConfig::builder()
-            .with_safe_default_protocol_versions()
-            .unwrap()
```
ctz: non-actionable thinking aloud:

with this commit, i probably would have gone for a different route: making `with_safe_default_protocol_versions()` infallible (it returns `WantsVerifier`), but that bakes-in the current property of that function that it enables both protocol versions (and, therefore, there's no route to achieving the degenerate TLS1.2-version but TLS1.3-suites or TLS1.3-version but TLS1.2-suites combinations that are unusable)

rustls/src/client/client_conn.rs
```diff
@@ -232,17 +234,41 @@ impl Clone for ClientConfig {
...
+    /// [`CryptoProvider`]: [`crate::crypto::ring::default_provider`], safe ciphersuite defaults and
+    /// the provided protocol versions.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
-    pub fn builder() -> ConfigBuilder<Self, WantsVersions> {
+    #[cfg(feature = "ring")]
+    pub fn builder_with_protocol_versions(
+        versions: &[&'static versions::SupportedProtocolVersion],
+    ) -> ConfigBuilder<Self, WantsVerifier> {
+        // Safety: we know the *ring* provider's ciphersuites are compatible with all protocol version choices.
```
ctz, line 259: moreover: we know that the provider has at least one ciphersuite for each version

i think this will panic for `builder_with_protocol_versions(&[])`; i don't think that necessarily needs fixing.

ctz (approved):
IMO this is a big improvement

rustls/src/builder.rs
cpu: Fixed up. Thanks!

rustls/src/client/client_conn.rs
cpu: > i think this will panic for builder_with_protocol_versions(&[]); i don't think that necessarily needs fixing.

I added a note to each fn's docstring about this :+1:

ctz:
> What if `CryptoProvider` had two fields:
> 
> ```rust
> tls12_cipher_suites: Vec<&'static Tls12CipherSuite>,
> tls13_cipher_suites: Vec<&'static Tls13CipherSuite>,
> ```

I guess we would put `tls13_cipher_suites` before `tls12_cipher_suites` on the wire? That seems like a bit of a policy decision that is currently not made by the library. Though -- hopefully -- nobody does version negotiation between TLS1.2 and TLS1.3 incorrectly based on the offered cipher suites?

> And the way to disable TLS 1.2 would be to provide an empty `Vec` for `tls12_cipher_suites`. That way we could do away with both the `unwrap()` and `builder_with_protocol_versions`.

We'd also need to put a `&'static SupportedProtocolVersion` in each `Tls12CipherSuite` and `Tls13CipherSuite`, and then require that everyone fills in the right one? Otherwise #224 would become impossible.

jsha:
> I guess we would put tls13_cipher_suites before tls12_cipher_suites on the wire? That seems like a bit of a policy decision that is currently not made by the library.

This is a good point, though as you say:

> hopefully -- nobody does version negotiation between TLS1.2 and TLS1.3 incorrectly based on the offered cipher suites?

So I think it's an okay policy decision to make.

> We'd also need to put a &'static SupportedProtocolVersion in each Tls12CipherSuite and Tls13CipherSuite, and then require that everyone fills in the right one? Otherwise https://github.com/rustls/rustls/issues/224 would become impossible.

Also a good point, and very achievable I think. But this does raise my estimate of how much work is involved.

jsha:
Filed as a separate issue: #1659

Issue #1635, opened 2023-11-28
UnexpectedEof Error with hyper HTTP/2 Requests


cablehead:
**Checklist**

* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**

When curl-ing a hyper HTTP/2 service over a connection wrapped with rustls, hyper sees an UnexpectedEof Error

**To Reproduce**

1. Run: https://github.com/cablehead/hyper-with-tls/tree/main/hyper-1.0.1-rustls
2. In another window: `curl -v -k --http2 https://localhost:3000`
3. Server logs: `Error serving connection: hyper::Error(Io, Kind(UnexpectedEof))`

**Applicable Version(s)**

rustls "0.21.9", with hyper 1.0.1 and 0.14

**Expected behavior**

Connection closes cleanly (no error logged).

**Additional context**

I've explored the issue with the hyper maintainers here: https://github.com/hyperium/hyper/issues/3427 

The most relevant comment related to rustls role is here: https://github.com/hyperium/hyper/issues/3427#issuecomment-1821789635

I've added a sample to https://github.com/cablehead/hyper-with-tls using native-tls, which doesn't seem to exhibit this behavior (on macOS at least).

I haven't seen this issue with rustls + hyper HTTP/1.1

There's a lot of moving parts here though. It may also be an issue with curl.

jsha:
Some quick context: rustls [returns UnexpectedEof from read](https://docs.rs/rustls/latest/rustls/struct.Reader.html#method.read) if a connection is closed without the peer sending the `close_notify` TLS message. This is to prevent attackers from truncating connections in the middle of a message and changing their semantics.

Higher level protocols like HTTP/1 and HTTP/2 have their own framing that knows the length of a message (Content-Length, or HTTP/2 frames). So both servers and clients often don't worry so much about sending TLS `close_notify` before shutting down a TCP connection, which is fine. So long as the protocol is between messages there's no risk of a truncation attack.

But that does mean servers and clients need a little extra filtering on errors, particularly for idle connections.

I am _guessing_ what's happening is this:

 - Curl makes a request, gets a response, and keeps the connection alive.
 - Curl finishes its list of URLs, and exits the process.
 - This closes the TCP connection (without explicitly sending `close_notify`).
 - Hyper is continually reading from the idle connection to see if more requests come in.
 - Hyper gets a TCP connection close.
 - But that is wrapped in a rustls reader, which turns it into an `UnexpectedEof` because there was no `close_notify`.
[...]

PR #1637, opened 2023-11-28
doc: add detail on UnexpectedEof

jsha:
Add documentation at the top level, and link to that documentation in the error message.

Related: #1635

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -142,8 +142,8 @@ impl<'a> io::Read for Reader<'a> {
     /// connection, so the underlying TCP connection should be half-closed too.
     ///
     /// If the peer closes the TLS session uncleanly (a TCP EOF without sending a
-    /// `close_notify` alert) this function returns `Err(ErrorKind::UnexpectedEof.into())`
-    /// once any pending data has been read.
+    /// `close_notify` alert) this function returns a `std::io::Error` of type
+    /// `ErrorKind::UnexpectedEof` once any pending data has been read.
     ///
     /// Note that support for `close_notify` varies in peer TLS libraries: many do not
     /// support it and uncleanly close the TCP connection (this might be
@@ -164,8 +164,13 @@ impl<'a> io::Read for Reader<'a> {
                 // cleanly closed; don't care about TCP EOF: express this as Ok(0)
                 (true, _) => {}
                 // unclean closure
-                (false, true) => return Err(io::ErrorKind::UnexpectedEof.into()),
-                // connection still going, but need more data: signal `WouldBlock` so that
+                (false, true) => {
+                    return Err(io::Error::new(
+                        io::ErrorKind::UnexpectedEof,
+                        UNEXPECTED_EOF_MESSAGE,
+                    ))
+                }
+                // connection still going, but needs more data: signal `WouldBlock` so that
                 // the caller knows this
                 (false, false) => return Err(io::ErrorKind::WouldBlock.into()),
             }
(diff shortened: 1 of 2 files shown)


rustls/src/lib.rs
```diff
@@ -271,6 +271,19 @@
 //!
 //!   Note that aws-lc-rs has additional build-time dependencies like cmake.
 //!   See [the documentation](https://aws.github.io/aws-lc-rs/requirements/index.html) for details.
+//!
+//!
+//! # Unexpected EOF
+//!
+//! TLS has a [`close_notify` mechanism to prevent truncation attacks](https://datatracker.ietf.org/doc/html/rfc8446#section-6.1).
+//! According to the TLS RFCs, each party is required to send it before closing the write side of the connection.
```
cpu, line 279: Above you talk about the close notify mechanism generally, but I think it means the "it" in the subsequent sentences isn't very clear. WDYT about saying "an alert" instead of "it":

```suggestion
//! According to the TLS RFCs, each party is required to send an alert before closing the write side of the connection.
```

rustls/src/lib.rs
```diff
@@ -271,6 +271,19 @@
...
+//! However, some implementations don't send it. So long as the application layer protocol (for instance HTTP/2)
+//! has message length framing and can reject truncated messages, this is not a security problem.
+//!
+//! Because rustls doesn't know the framing of the application layer above it, it treats an EOF without close_notify as an error,
```
cpu, line 283: nit: consistent backticks for `close_notify`.
```suggestion
//! Because rustls doesn't know the framing of the application layer above it, it treats an EOF without `close_notify` as an error,
```

rustls/src/lib.rs
```diff
@@ -271,6 +271,19 @@
...
+//! of type std::io::Error with ErrorKind::UnexpectedEof. In some situations it's appropriate for the application to treat this
```
cpu, line 284: nit: backticks
```suggestion
//! of type `std::io::Error` with `ErrorKind::UnexpectedEof`. In some situations it's appropriate for the application to treat this
```

rustls/src/lib.rs
```diff
@@ -271,6 +271,19 @@
...
+//! error the same way it would treat a normal EOF (a read returning `Ok(0)`). In particular if UnexpectedEof occurs on an idle
```
cpu, line 285:
```suggestion
//! error the same way it would treat a normal EOF (a read returning `Ok(0)`). In particular if `UnexpectedEof` occurs on an idle
```

cpu (approved):
Seems like a helpful clarification for an issue folks are bumping into frequently. Thanks :+1:

djc:
(It looks like the backticks aren't in sync across README and `lib.rs`? Weird that the CI job for this isn't complaining.)

djc (approved):

cpu:
> Weird that the CI job for this isn't complaining

Ah! I think we accidentally broke `pull-readme` with 1db4506dcfeb4a08b5517eafba84f16084321786 - the marker it was looking for was changed. I'll put up a PR to fix.

djc:
Nice catch!

cpu:
> I think we accidentally broke pull-readme with https://github.com/rustls/rustls/commit/1db4506dcfeb4a08b5517eafba84f16084321786

Sorry for the noise - I believe this was a false conclusion - the relevant headers are unchanged and the script seems to be working at that commit. I'm looking more carefully for a root cause now.

cpu:
> (It looks like the backticks aren't in sync across README and lib.rs? Weird that the CI job for this isn't complaining.)

ok, the root cause is actually much more benign: the readme extract script only grabs the content of `lib.rs` between "# Rustls - a modern TLS library" and "## Design Overview". 

The new content is _after_ the design overview, and so isn't extracted from `lib.rs` at all.

If we want to maintain the text in only one place it needs to be put above the `## Design Overview` marker.

jsha:
Sounds good. In the meantime I realized we also have some dedicated documentation pages in the manual (https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html). Perhaps that's a better home for this than the README / top-level doc?

cpu:
> Perhaps that's a better home for this than the README / top-level doc?

That sounds good to me. It does feel a little bit too specific for the README in retrospect.

cpu (approved):
The new home for the docs looks good to me. Thanks!

rustls/src/conn.rs
```diff
@@ -134,6 +134,9 @@ pub struct Reader<'a> {
     has_seen_eof: bool,
 }
 
+const UNEXPECTED_EOF_MESSAGE: &str = "peer closed connection without sending TLS close_notify: \
```
djc, line 137: Nit: we put `const` values near the bottom usually.

djc (approved):

ctz (approved):

PR #1640, opened 2023-11-29
Limit fragmentation in `write_vectored`

Keksoj:
We, [Wonshtrum](https://github.com/Wonshtrum) and myself, work at Clever Cloud on a custom-made reverse proxy called [Sōzu](https://github.com/sozu-proxy/sozu/). We switched from using both OpenSSL and Rustls to being Rustls-only, one year ago, and we are very happy about it. We appreciate the security orientation of Rustls, as well as its native Rust nature.

## Our issue

In Sōzu, our reverse proxy, we strive to optimize system calls and throughput. Here is the syscall we found with strace  when transmitting a simple HTTPS response to a client:

```
writev(
    11,
    [
        {iov_base="\27\3\3\0\31\210H\371\252\25w\304\275\346u\27\371\334g\345\244\371\331\203c\302\356\324i\367", iov_len=30},
        {iov_base="\27\3\3\0\22\366IR\236\357\374\37\30\310E\330Xr\3249\24r0", iov_len=23},
        {iov_base="\27\3\3\0\24\326\337\223\356R\360\37\215\343a\2L\236\30\24D\31\10\363\16", iov_len=25},
        {iov_base="\27\3\3\0\22\347\305\255\371?\224\6\212\325\345\250*a\250\321i\366\266", iov_len=23},
        {iov_base="\27\3\3\0\23\37\24Ex\334\vHp\2758\207\27G\306\16\226\253\320\240", iov_len=24},
        {iov_base="\27\3\3\0\23\23\362q\352\307\303\364\312]\361(\227<\17\334Y\333g\310", iov_len=24},
        {iov_base="\27\3\3\0\37\216\332\374\376\303H\206\35f\334\310\252\375\341\366\224\261*\367\30mg\\\230dbh"..., iov_len=36},
        {iov_base="\27\3\3\0\23-\312\31\v\353\212\303\265\231H\371\356%\317JX\32Tg", iov_len=24},
        {iov_base="\27\3\3\0\23d\356\0276\3236\252\206^5\346=\234F\2\200\33\314>", iov_len=24},
        {iov_base="\27\3\3\0\23U\30\37x7\5d\304/ZY\274\25\17;\276t]\216", iov_len=24},
        {iov_base="\27\3\3\0\25\240\371q\2416Z\202\6\35\311|\203Bai-\20\217=\3516", iov_len=26},
        {iov_base="\27\3\3\0\23\220\350\323\321bx\321\21:\35\203?\257\313)\200\364E\377", iov_len=24},
        {iov_base="\27\3\3\0.\226\254\30\240\315\307o\250\243b`Q\n\17\226\333\347\256[\331\324#~/\240\206,"..., iov_len=51},
        {iov_base="\27\3\3\0\23\251\207:\350?/\252Z\317\322\0017\24E\256\5*\342\224", iov_len=24},
        {iov_base="\27\3\3\0\30$w0!\205\310\276\372k\204?\375k\334\334\363\314\\!\355\266\0\362\33", iov_len=29},
        {iov_base="\27\3\3\0\23\257\277rj;\3\376\251\10\310\37\265\365\303\216\260\345\361\366", iov_len=24},
        {iov_base="\27\3\3\0+\341\0313\277'\223\3\272W&\316\306\360BYI9\225Vh\347t\3}\303\321\343"..., iov_len=48},
        {iov_base="\27\3\3\0\23\247\33\26M\255%\351\214F\254\177\250+z\253]\17F\325", iov_len=24},
        {iov_base="\27\3\3\0\23fm\262t\334\271>\226\225ITia\250,\31^\37\f", iov_len=24},
        {iov_base="\27\3\3\0\33S\25/zM'\343\302L/l\\\1\22\250M\0AL\357\304\36\227\317\322Q\365", iov_len=32}
    ],
    20
) = 563
```

What we see here is a number of small pieces of data, headers mostly, being encoded into TLS frames, one at a time. The `\27\3\3\0` pattern is the beginning of a TLS frame.
This repetition is a waste of bandwitdth.

## Our investigation

The system call above is due to how the Writer's [`write_vectored` method](https://docs.rs/rustls/latest/rustls/struct.Writer.html#method.write_vectored) encodes the chunks one by one.

In our case, our reverse proxy creates a number of smaller chunks. Half of the throughput is monopolized by TLS headers.

## Our proposal

We could have solved our issue on the proxy side, by accumulating chunks and calling `write()` instead of calling `write_vectored`, but we seek to avoid dynamic memory allocation and copying.

In this pull request on Rustls, instead of treating the chunks individually in `write_vectored`, we suggest to pass them down together in a BorrowedPayload form, until they are encrypted in one single TLS frame (or several if the payload exceeds `MAX_FRAGMENT_SIZE`). This new type `BorrowedPayload` does not copy the data but manipulates slices of bytes.

## Syscall improvement

This is the same response as above: a simple "hi" and some headers.

```
writev(
    14,
    [
        {iov_base="\27\3\3\0\214\376h\202\316\333w\252\25\24\375\354\254\24p\343\320\320\311\21\340\315B\367\327\342\225\340"..., iov_len=145}
    ],
    1
) = 145
```

We can see that there is only one `iov_base` (corresponding to 1 TLS frame), instead of many prior.
The syscall has dimished in size, from 563 to 145. 

## Benchmarks

Our modification reduces bandwidth usage significantly when benchmarking our reverse proxy.

These benchmarks run on a simple desktop machine, with 4 backends to route HTTPS traffic to. The backends reply with a simple "hi" and some headers.

### Sōzu + Rustls 0.21.9

```
bombardier -c 450 -n 100000 https://localhost:8443/api -l

Statistics        Avg      Stdev        Max
  Reqs/sec      7973.06    4220.70   18732.68
  Latency       56.68ms    69.91ms      1.15s
  Latency Distribution
     50%    51.98ms
     75%    54.38ms
     90%    56.90ms
     95%    58.64ms
     99%    61.33ms
  HTTP codes:
    1xx - 0, 2xx - 100000, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     5.01MB/s
```

### Sōzu + Rustls (this branch)

```
bombardier -c 450 -n 100000 https://localhost:8443/api -l                                             
Bombarding https://localhost:8443/api with 100000 request(s) using 450 connection(s)

Statistics        Avg      Stdev        Max
  Reqs/sec      8616.52    5132.64   22453.38
  Latency       52.58ms    64.24ms      1.04s
  Latency Distribution
     50%    47.85ms
     75%    50.40ms
     90%    53.53ms
     95%    55.80ms
     99%    76.39ms
  HTTP codes:
    1xx - 0, 2xx - 100000, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     2.00MB/s
```

### Discussion

The strace and the benchmark suggest that our addition to the rustls codebase diminishes bandwidth use and improves overall performance in a meaningful way. Requests per second, and latency, are the key metrics for us.

Diving in Rustls, we found a number of memory allocation that could be avoided. If you would like, we would be more than happy to work with you in this direction.

(edited for clarity)

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -5,11 +5,12 @@ use crate::error::{Error, PeerMisbehaved};
 use crate::log::trace;
 use crate::msgs::deframer::{Deframed, DeframerSliceBuffer, DeframerVecBuffer, MessageDeframer};
 use crate::msgs::handshake::Random;
-use crate::msgs::message::{InboundMessage, Message, MessagePayload};
+use crate::msgs::message::{InboundMessage, Message, MessagePayload, OutboundChunks};
 use crate::suites::{ExtractedSecrets, PartiallyExtractedSecrets};
 use crate::vecbuf::ChunkVecBuffer;
 
 use alloc::boxed::Box;
+use alloc::vec::Vec;
 use core::fmt::Debug;
 use core::mem;
 use core::ops::{Deref, DerefMut};
@@ -257,18 +258,27 @@ impl<T> PlaintextSink for ConnectionCommon<T> {
         Ok(self
             .core
             .common_state
-            .buffer_plaintext(buf, &mut self.sendable_plaintext))
+            .buffer_plaintext(buf.into(), &mut self.sendable_plaintext))
     }
 
     fn write_vectored(&mut self, bufs: &[io::IoSlice<'_>]) -> io::Result<usize> {
-        let mut sz = 0;
-        for buf in bufs {
-            sz += self
-                .core
-                .common_state
-                .buffer_plaintext(buf, &mut self.sendable_plaintext);
-        }
-        Ok(sz)
+        let payload_owner: Vec<&[u8]>;
+        let payload = match bufs.len() {
+            0 => return Ok(0),
+            1 => OutboundChunks::Single(bufs[0].deref()),
+            _ => {
+                payload_owner = bufs
+                    .iter()
+                    .map(|io_slice| io_slice.deref())
+                    .collect();
+
+                OutboundChunks::new(&payload_owner)
+            }
+        };
+        Ok(self
+            .core
+            .common_state
+            .buffer_plaintext(payload, &mut self.sendable_plaintext))
     }
 
     fn flush(&mut self) -> io::Result<()> {
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -1,12 +1,11 @@
 use crate::common_state::{CommonState, Context, IoState, State, DEFAULT_BUFFER_LIMIT};
-use crate::crypto::cipher::BorrowedPlainMessage;
 use crate::enums::{AlertDescription, ContentType};
 use crate::error::{Error, PeerMisbehaved};
 #[cfg(feature = "logging")]
 use crate::log::trace;
 use crate::msgs::deframer::{Deframed, DeframerSliceBuffer, DeframerVecBuffer, MessageDeframer};
 use crate::msgs::handshake::Random;
-use crate::msgs::message::{Message, MessagePayload};
+use crate::msgs::message::{InboundMessage, Message, MessagePayload};
 use crate::suites::{ExtractedSecrets, PartiallyExtractedSecrets};
 use crate::vecbuf::ChunkVecBuffer;
 
@@ -337,7 +336,7 @@ impl ConnectionRandoms {
 
 // --- Common (to client and server) connection functions ---
 
-fn is_valid_ccs(msg: &BorrowedPlainMessage) -> bool {
+fn is_valid_ccs(msg: &InboundMessage) -> bool {
     // We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
     // Note: this is prior to the record layer, so is unencrypted. See
     // third paragraph of section 5 in RFC8446.
@@ -778,7 +777,7 @@ impl<Data> ConnectionCore<Data> {
         &mut self,
         state: Option<&dyn State<Data>>,
         deframer_buffer: &mut DeframerSliceBuffer<'b>,
-    ) -> Result<Option<BorrowedPlainMessage<'b>>, Error> {
+    ) -> Result<Option<InboundMessage<'b>>, Error> {
         match self.message_deframer.pop(
             &mut self.common_state.record_layer,
             self.common_state.negotiated_version,
@@ -833,7 +832,7 @@ impl<Data> ConnectionCore<Data> {
 
     fn process_msg(
         &mut self,
-        msg: BorrowedPlainMessage,
+        msg: InboundMessage,
         state: Box<dyn State<Data>>,
         sendable_plaintext: Option<&mut ChunkVecBuffer>,
     ) -> Result<Box<dyn State<Data>>, Error> {
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -344,15 +344,6 @@ impl ConnectionRandoms {
     }
 }
 
-// --- Common (to client and server) connection functions ---
-
-fn is_valid_ccs(msg: &InboundMessage) -> bool {
-    // We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
-    // Note: this is prior to the record layer, so is unencrypted. See
-    // third paragraph of section 5 in RFC8446.
-    msg.typ == ContentType::ChangeCipherSpec && msg.payload == [0x01]
-}
-
 /// Interface shared by client and server connections.
 pub struct ConnectionCommon<Data> {
     pub(crate) core: ConnectionCore<Data>,
@@ -853,7 +844,7 @@ impl<Data> ConnectionCore<Data> {
                 .may_receive_application_data
             && self.common_state.is_tls13()
         {
-            if !is_valid_ccs(&msg)
+            if !msg.is_valid_ccs()
                 || self.common_state.received_middlebox_ccs > TLS13_MAX_DROPPED_CCS
             {
                 // "An implementation which receives any other change_cipher_spec value or
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -4,11 +4,10 @@ use core::slice::SliceIndex;
 use std::io;
 
 use super::codec::Codec;
-use super::message::{BorrowedOpaqueMessage, BorrowedPlainMessage};
 use crate::enums::{ContentType, ProtocolVersion};
 use crate::error::{Error, InvalidMessage, PeerMisbehaved};
 use crate::msgs::codec;
-use crate::msgs::message::{MessageError, OpaqueMessage};
+use crate::msgs::message::{BorrowedOpaqueMessage, InboundMessage, MessageError, OpaqueMessage};
 use crate::record_layer::{Decrypted, RecordLayer};
 
 /// This deframer works to reconstruct TLS messages from a stream of arbitrary-sized reads.
@@ -113,13 +112,13 @@ impl MessageDeframer {
                     version,
                     payload,
                 } = m;
-                let raw_payload = RawSlice::from(&*payload);
+                let raw_payload_slice = RawSlice::from(&*payload);
                 // This is unencrypted. We check the contents later.
                 buffer.queue_discard(end);
-                let message = BorrowedPlainMessage {
+                let message = InboundMessage {
                     typ,
                     version,
-                    payload: buffer.take(raw_payload),
+                    payload: buffer.take(raw_payload_slice),
                 };
                 return Ok(Some(Deframed {
                     want_close_before_decrypt: false,
@@ -130,14 +129,19 @@ impl MessageDeframer {
             }
 
             // Decrypt the encrypted message (if necessary).
-            let msg = match record_layer.decrypt_incoming(m) {
+            let (typ, version, plain_payload_slice) = match record_layer.decrypt_incoming(m) {
                 Ok(Some(decrypted)) => {
                     let Decrypted {
                         want_close_before_decrypt,
-                        plaintext,
+                        plaintext:
+                            InboundMessage {
+                                typ,
+                                version,
+                                payload,
+                            },
                     } = decrypted;
                     debug_assert!(!want_close_before_decrypt);
-                    plaintext
+                    (typ, version, RawSlice::from(payload))
                 }
                 // This was rejected early data, discard it. If we currently have a handshake
                 // payload in progress, this counts as interleaved, so we error out.
@@ -153,7 +157,7 @@ impl MessageDeframer {
                 Err(e) => return Err(e),
             };
 
-            if self.joining_hs.is_some() && msg.typ != ContentType::Handshake {
+            if self.joining_hs.is_some() && typ != ContentType::Handshake {
                 // "Handshake messages MUST NOT be interleaved with other record
                 // types.  That is, if a handshake message is split over two or more
                 // records, there MUST NOT be any other records between them."
@@ -162,19 +166,12 @@ impl MessageDeframer {
             }
 
             // If it's not a handshake message, just return it -- no joining necessary.
-            if msg.typ != ContentType::Handshake {
-                let BorrowedPlainMessage {
-                    typ,
-                    version,
-                    payload,
-                } = msg;
-                let raw_payload = RawSlice::from(payload);
-                let end = start + rd.used();
+            if typ != ContentType::Handshake {
                 buffer.queue_discard(end);
-                let message = BorrowedPlainMessage {
+                let message = InboundMessage {
                     typ,
                     version,
-                    payload: buffer.take(raw_payload),
+                    payload: buffer.take(plain_payload_slice),
                 };
                 return Ok(Some(Deframed {
                     want_close_before_decrypt: false,
@@ -186,9 +183,7 @@ impl MessageDeframer {
 
             // If we don't know the payload size yet or if the payload size is larger
             // than the currently buffered payload, we need to wait for more data.
-            let raw = RawSlice::from(msg.payload);
-            let version = msg.version;
-            let src = buffer.raw_slice_to_filled_range(raw);
+            let src = buffer.raw_slice_to_filled_range(plain_payload_slice);
             match self.append_hs(version, InternalPayload(src), end, buffer)? {
                 HandshakePayloadState::Blocked => return Ok(None),
                 HandshakePayloadState::Complete(len) => break len,
@@ -221,7 +216,7 @@ impl MessageDeframer {
             buffer.queue_discard(end);
         }
 
-        let message = BorrowedPlainMessage {
+        let message = InboundMessage {
             typ,
             version,
             payload: buffer.take(raw_payload),
@@ -686,7 +681,7 @@ pub struct Deframed<'a> {
     pub(crate) want_close_before_decrypt: bool,
     pub(crate) aligned: bool,
     pub(crate) trial_decryption_finished: bool,
-    pub message: BorrowedPlainMessage<'a>,
+    pub message: InboundMessage<'a>,
 }
 
 const HEADER_SIZE: usize = 1 + 3;
@@ -703,7 +698,7 @@ mod tests {
     use std::io;
 
     use crate::crypto::cipher::PlainMessage;
-    use crate::msgs::message::Message;
+    use crate::msgs::message::{BorrowedPlainMessage, Message};
 
     use super::*;
 
--- a/rustls/src/msgs/fragmenter.rs
+++ b/rustls/src/msgs/fragmenter.rs
@@ -1,6 +1,6 @@
 use crate::enums::ContentType;
 use crate::enums::ProtocolVersion;
-use crate::msgs::message::{OutboundMessage, PlainMessage};
+use crate::msgs::message::{OutboundChunks, OutboundMessage, PlainMessage};
 use crate::Error;
 pub(crate) const MAX_FRAGMENT_LEN: usize = 16384;
 pub(crate) const PACKET_OVERHEAD: usize = 1 + 2 + 2;
@@ -27,24 +27,22 @@ impl MessageFragmenter {
         &self,
         msg: &'a PlainMessage,
     ) -> impl Iterator<Item = OutboundMessage<'a>> + 'a {
-        self.fragment_slice(msg.typ, msg.version, msg.payload.bytes())
+        self.fragment_payload(msg.typ, msg.version, msg.payload.bytes().into())
     }
 
     /// Enqueue borrowed fragments of (version, typ, payload) which
     /// are no longer than max_frag onto the `out` deque.
-    pub(crate) fn fragment_slice<'a>(
+    pub(crate) fn fragment_payload<'a>(
         &self,
         typ: ContentType,
         version: ProtocolVersion,
-        payload: &'a [u8],
+        payload: OutboundChunks<'a>,
     ) -> impl ExactSizeIterator<Item = OutboundMessage<'a>> {
-        payload
-            .chunks(self.max_frag)
-            .map(move |payload| OutboundMessage {
-                typ,
-                version,
-                payload,
-            })
+        Chunker::new(payload, self.max_frag).map(move |payload| OutboundMessage {
+            typ,
+            version,
+            payload,
+        })
     }
 
     /// Set the maximum fragment size that will be produced.
@@ -65,13 +63,47 @@ impl MessageFragmenter {
     }
 }
 
+/// An iterator over borrowed fragments of a payload
+struct Chunker<'a> {
+    payload: OutboundChunks<'a>,
+    limit: usize,
+}
+
+impl<'a> Chunker<'a> {
+    fn new(payload: OutboundChunks<'a>, limit: usize) -> Self {
+        Self { payload, limit }
+    }
+}
+
+impl<'a> Iterator for Chunker<'a> {
+    type Item = OutboundChunks<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.payload.is_empty() {
+            return None;
+        }
+
+        let (before, after) = self.payload.split_at(self.limit);
+        self.payload = after;
+        Some(before)
+    }
+}
+
+impl<'a> ExactSizeIterator for Chunker<'a> {
+    fn len(&self) -> usize {
+        (self.payload.len() + self.limit - 1) / self.limit
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::{MessageFragmenter, PACKET_OVERHEAD};
     use crate::enums::ContentType;
     use crate::enums::ProtocolVersion;
     use crate::msgs::base::Payload;
-    use crate::msgs::message::{BorrowedPlainMessage, OutboundMessage, PlainMessage};
+    use crate::msgs::message::{
+        BorrowedPlainMessage, OutboundChunks, OutboundMessage, PlainMessage,
+    };
 
     fn msg_eq(
         m: &OutboundMessage,
@@ -82,7 +114,7 @@ mod tests {
     ) {
         assert_eq!(&m.typ, typ);
         assert_eq!(&m.version, version);
-        assert_eq!(m.payload, bytes);
+        assert_eq!(m.payload.to_vec(), bytes);
 
         let buf = m.to_unencrypted_opaque().encode();
 
@@ -158,5 +190,36 @@ mod tests {
             &ProtocolVersion::TLSv1_2,
             b"\x01\x02\x03\x04\x05\x06\x07\x08",
         );
+    }
+
+    #[test]
+    fn fragment_multiple_slices() {
+        let typ = ContentType::Handshake;
+        let version = ProtocolVersion::TLSv1_2;
+        let payload_owner: Vec<&[u8]> = vec![&[b'a'; 8], &[b'b'; 12], &[b'c'; 32], &[b'd'; 20]];
+        let borrowed_payload = OutboundChunks::new(&payload_owner);
+        let mut frag = MessageFragmenter::default();
+        frag.set_max_fragment_size(Some(37)) // 32 + packet overhead
+            .unwrap();
+
+        let fragments = frag
+            .fragment_payload(typ, version, borrowed_payload)
+            .collect::<Vec<_>>();
+        assert_eq!(fragments.len(), 3);
+        msg_eq(
+            &fragments[0],
+            37,
+            &typ,
+            &version,
+            b"aaaaaaaabbbbbbbbbbbbcccccccccccc",
+        );
+        msg_eq(
+            &fragments[1],
+            37,
+            &typ,
+            &version,
+            b"ccccccccccccccccccccdddddddddddd",
+        );
+        msg_eq(&fragments[2], 13, &typ, &version, b"dddddddd");
     }
 }
--- a/rustls/src/msgs/fragmenter.rs
+++ b/rustls/src/msgs/fragmenter.rs
@@ -1,6 +1,6 @@
 use crate::enums::ContentType;
 use crate::enums::ProtocolVersion;
-use crate::msgs::message::{BorrowedPlainMessage, PlainMessage};
+use crate::msgs::message::{OutboundMessage, PlainMessage};
 use crate::Error;
 pub(crate) const MAX_FRAGMENT_LEN: usize = 16384;
 pub(crate) const PACKET_OVERHEAD: usize = 1 + 2 + 2;
@@ -26,7 +26,7 @@ impl MessageFragmenter {
     pub fn fragment_message<'a>(
         &self,
         msg: &'a PlainMessage,
-    ) -> impl Iterator<Item = BorrowedPlainMessage<'a>> + 'a {
+    ) -> impl Iterator<Item = OutboundMessage<'a>> + 'a {
         self.fragment_slice(msg.typ, msg.version, msg.payload.bytes())
     }
 
@@ -37,13 +37,13 @@ impl MessageFragmenter {
         typ: ContentType,
         version: ProtocolVersion,
         payload: &'a [u8],
-    ) -> impl ExactSizeIterator<Item = BorrowedPlainMessage<'a>> {
+    ) -> impl ExactSizeIterator<Item = OutboundMessage<'a>> {
         payload
             .chunks(self.max_frag)
-            .map(move |c| BorrowedPlainMessage {
+            .map(move |payload| OutboundMessage {
                 typ,
                 version,
-                payload: c,
+                payload,
             })
     }
 
@@ -71,10 +71,10 @@ mod tests {
     use crate::enums::ContentType;
     use crate::enums::ProtocolVersion;
     use crate::msgs::base::Payload;
-    use crate::msgs::message::{BorrowedPlainMessage, PlainMessage};
+    use crate::msgs::message::{BorrowedPlainMessage, OutboundMessage, PlainMessage};
 
     fn msg_eq(
-        m: &BorrowedPlainMessage,
+        m: &OutboundMessage,
         total_len: usize,
         typ: &ContentType,
         version: &ProtocolVersion,
--- a/rustls/src/msgs/message.rs
+++ b/rustls/src/msgs/message.rs
@@ -437,6 +437,17 @@ pub struct InboundMessage<'a> {
     pub payload: &'a [u8],
 }
 
+impl InboundMessage<'_> {
+    /// Returns true if the payload is a CCS message.
+    ///
+    /// We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
+    /// Note: this is prior to the record layer, so is unencrypted. See
+    /// third paragraph of section 5 in RFC8446.
+    pub(crate) fn is_valid_ccs(&self) -> bool {
+        self.typ == ContentType::ChangeCipherSpec && self.payload == [0x01]
+    }
+}
+
 impl BorrowedPlainMessage for InboundMessage<'_> {
     fn payload_to_vec(&self) -> Vec<u8> {
         self.payload.to_vec()
--- a/rustls/src/msgs/message.rs
+++ b/rustls/src/msgs/message.rs
@@ -347,7 +347,7 @@ impl PlainMessage {
         OutboundMessage {
             version: self.version,
             typ: self.typ,
-            payload: self.payload.bytes(),
+            payload: self.payload.bytes().into(),
         }
     }
 }
@@ -468,7 +468,7 @@ impl BorrowedPlainMessage for InboundMessage<'_> {
 pub struct OutboundMessage<'a> {
     pub typ: ContentType,
     pub version: ProtocolVersion,
-    pub payload: &'a [u8],
+    pub payload: OutboundChunks<'a>,
 }
 
 impl BorrowedPlainMessage for OutboundMessage<'_> {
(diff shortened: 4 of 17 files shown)


Wonshtrum:
On the topic of allocation/copy, we put a lot of effort into Sozu to handle HTTP messages (parsing and proxying) in zero-copy in most cases. Very few dynamic allocations are also necessary, as Sozu preallocates and reuses a pool of buffers.

Analyzing the Rustls codebase, we found that in order to encrypt a payload, the entire plaintext message is copied by `MessageEncrypter::encrypt`, then the entire encrypted message is copied by `OpaqueMessage::encode`. Both copies require a memory allocation.

As explained in this PR description, aggregating the vector chunks on our side would mean a third allocation and copy. Instead, we placed the responsibility of aggregation on the copy of `MessageEncrypter::encrypt`. This copy could also be avoided if the ciphers accepted fragmented payloads, but it would be a lot more complicated.

On the other hand, `OpaqueMessage::encode` only appends the 5-bytes "TLS header". Allocating and copying an entire payload (up to 16Kb) seems wasteful and could be completely avoided if the payload built by `MessageEncrypted::encrypt` accounted for the 5-bytes prelude. `OpaqueMessage::encode` would only have to fill in the reserved space.

My remarks are not directly related to this PR, but to the overall talk on memory allocation. As @Keksoj said we are eager to contribute and would happily make this change as well.

djc:
> On the other hand, `OpaqueMessage::encode` only appends the 5-bytes "TLS header". Allocating and copying an entire payload (up to 16Kb) seems wasteful and could be completely avoided if the payload built by `MessageEncrypted::encrypt` accounted for the 5-bytes prelude. `OpaqueMessage::encode` would only have to fill in the reserved space.
> 
> My remarks are not directly related to this PR, but to the overall talk on memory allocation. As @Keksoj said we are eager to contribute and would happily make this change as well.

Thanks for the investigation! We will review your changes shortly -- meanwhile, I just wanted to acknowledge that we'd be happy to review a PR for this, too.

paolobarbolini (commented):
I'm too interested in this change. I've left a couple of comments (mostly nitpicks)

rustls/src/msgs/fragmenter.rs
```diff
@@ -2,6 +2,8 @@ use crate::enums::ContentType;
 use crate::enums::ProtocolVersion;
 use crate::msgs::message::{BorrowedPlainMessage, PlainMessage};
 use crate::Error;
+
+use super::message::BorrowedPayload;
```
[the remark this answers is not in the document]
Keksoj: sure!

Keksoj:
Thanks for the reviews! We appreciate your reception. 

We reproduced the icount test of the CI locally and found a lot of variance. Should we take them into account?

cpu:
@rustls-benchmarking bench

(_experimenting with our new benchmarking integration_)

cpu:
Have you folks taken a look at https://github.com/rustls/rustls/pull/1597 ? I haven't done a deep dive on your branch yet but I think it would be important to make sure what you're proposing and what's in the queue in #1597 can be harmonized. 

@pvdrz flagging this branch as one you might be interested in as well.

cpu:
Here's a branch I started to try and rebase this on main: https://github.com/cpu/rustls/tree/cpu-borrowed-payload

<details>
<summary>I made some minor changes along the way, none of which feel especially substantial:</summary>

* The `BorrowedPayload` type from this branch is now named `BorrowedPlainPayload` - there's already [a `BorrowedPayload` type from the ferrous work](https://github.com/rustls/rustls/blob/c1c917cfbecdf5c3d4035d852c198b0308b11899/rustls/src/msgs/base.rs#L65-L116) used for a non-owning payload in the context of opaque messages. It might make sense to name this `BorrowedOpaquePayload` to emphasize this, but I've avoided that change for now and renamed the new type.
* I reordered the members of `BorrowedPlainPayload` to match style guide conventions
* I employed more liberal use of `Self` within member `BorrowedPlainPayload` fns.
* Unit tests for `BorrowedPlainPayload` were moved to a tests mod at the end of `message.rs`.
* I implemented `Into<Vec<u8>>` for `BorrowedPlainPayload` - it already offered a `to_vec()` fn so this is just ergonomics.
* Throughout I tried to explicitly write lifetimes instead of eliding them, to match style guide convention.
* I inlined a `size` binding in the `write_vector` return
</details>

Most importantly, as it stands now my branch **doesn't build** because of the `deframer.rs` drift I mentioned in my previous comment w.r.t https://github.com/rustls/rustls/pull/1597. I haven't been able to square the `RawSlice` arrangement that was [already flagged as problematic](https://github.com/rustls/rustls/pull/1597#discussion_r1441599875) in previous reviews with the new borrowed plaintext payload representation where the data is spread across multiple independently borrowed slices. I've separately tried (and failed) to refactor the deframer code to try and wrangle out the `RawSlice` parts, so this feels like I'm hitting the same wall. 

I'm very open to the possibility I'm completely on the wrong track here but as it stands I'm stuck (again :sweat:) and could use guidance.

ctz:
Just a shower thought: should we change `pub fn encrypt` in `impl<Data> WriteTraffic<'_, Data> {` so `application_data` is `&[&[u8]]` instead?

At the current time, once we do #1723 we would have to implement `io::Write::write_vectored` by calling `encrypt` multiple times[^1] which would regress the behaviour desired in this PR.

[^1]: I am ignoring allocating and copying

Wonshtrum:
Hello @cpu, we finished your attempt at rebasing on main, all the while encountering the same problem that you have.
We managed to get it to compile, but while pattern matching , we had to leave some branches of `BorrowedPlainPayload` unimplemented:

https://github.com/Wonshtrum/rustls/blob/f31f35d910a40e36c65654dbf97eafa1829b82bb/rustls/src/msgs/message.rs#L412-L427
https://github.com/Wonshtrum/rustls/blob/f31f35d910a40e36c65654dbf97eafa1829b82bb/rustls/src/msgs/deframer.rs#L164-L169

In both cases, they should be truly unreachable. This stems from the fact that we designed `BorrowedPlainPayload` as an outbound type but it is currently used (through `BorrowedPlainMessage`) for inbound data. As such `BorrowedPlainPayload` can hold references to fragmented regions of memory (through its `Multiple` variant), which makes sense as the data comes from the user and may be spread out in memory (this was our original complaint about `write_vectored`). But it does not make sense for incoming frames which are guaranteed to hold a single consecutive slice of bytes (coming from the `Deframer` internal buffer).
So in both cases, there is no proper way to handle the `BorrowedPlainMessage::Multiple` variant and we are guaranteed we will never have to by design.

Coming to this realization I think we have 4 different options:
1. keep it like that, it's a bit awkward but we can document it
2. explicitly take into account this dichotomy between inbound and outbound in the types. We would have an `InboundBorrowedPayload` and an `OutboundBorrowedPayload` and potentially similar types for messages.
3. use a single "powerful" `Payload` type, merging our `BorrowedPlainMessage` with the current `Payload`, and explicitly allow the payload as potentially being fragmented everywhere, even in the code handling inbound data
4. give up :scream: 

In my opinion, options 2 and 3 are the most elegant ones, but it's a lot of work for a feature originally meant to be well delimited.
We would love to have a chat about this with you folks. Do you hang out on Discord?

cpu:
Hi @Wonshtrum, thanks for taking a look at this.

>  This stems from the fact that we designed BorrowedPlainPayload as an outbound type but it is currently used (through BorrowedPlainMessage) for inbound data.

:bulb: that does seem like the crux of the issues we were both running into. This is a good insight. 

> Coming to this realization I think we have 4 different options:

Of the four options presented (well, 3 since I don't think we should give up yet :laughing:) I think option 2 seems the most reasonable to me, but I'm curious what others think. I think option 1 should be a last resort, and option 3 seems like it might be more complexity than is merited. 

> We would love to have a chat about this with you folks. Do you hang out on Discord?

Absolutely, you can find us in [the "crates" discord](https://discord.gg/JZQRZkB8rZ) Djc created, in the #rustls channel.

djc:
I would also gravitate towards option 2 probably at least as an initial approach. If we find there's a lot of duplication, can always merge them again into something more like option 3? But I think it will be helpful in terms of understanding, too, to figure out what the needs on the receive side are vs the needs on the send side (not sure between "receive" vs "send" and "inbound" vs "outbound" which is the clearer framing? I have a slight preference to the former).

Wonshtrum:
I also think option 2 is the cleaner one. I quickly looked into the code, it seems there is only `BorrowedPlainMessage` that holds a `BorrowedPlainPayload`, and `BorrowedPlainMessage` is only held by `Decrypted` and `Deframed` which should both be inbound. So I would make the current `BorrowedPlainMessage` exclusively an inbound type holding a single slice of u8, and create a distinct outbound equivalent holding the current `BorrowedPlainPayload` and use it in all the send and encrypt code.
I may have overlooked something, but this seems like the only duplication needed.

Keksoj:
I agree with you all, and on top of that, option 2 seems relatively doable.

As for the naming, I would go for `Inbound-` and `Outbound` rather than `Send-` and `Receive-`, because we need to indicate the direction of the payload, rather than the action we perform with it.

But maybe there are other namings available. Maybe `Incoming-` and `Outgoing-` ?

cpu:
@rustls-benchmarking bench

cpu:
The last push looks good to me :+1: I think the branch needs a `rustfmt` pass to fix the one CI task that failed.

rustls/src/msgs/message.rs
```diff
@@ -459,6 +459,149 @@ impl<'a> BorrowedPlainMessage<'a> {
...
 
+#[derive(Debug, Clone)]
+/// A collection of borrowed plaintext slices
+pub enum BorrowedPlainPayload<'a> {
+    Empty,
+    /// A single byte slice. This allows only one redirection instead of several.
+    Single(&'a [u8]),
+    /// A collection of chunks (byte slices)
+    /// and inclusive-exclusive cursors to single out a slice of bytes.
+    /// [BorrowedPayload] assumes three invariants:
```
cpu, line 538: I think this should be `BorrowedPlainPayload` after the rename (_Sorry, I think I'm the one that missed that when doing the rename!_)

rustls/src/msgs/message.rs
```diff
@@ -459,6 +459,149 @@ impl<'a> BorrowedPlainMessage<'a> {
...
+        start: usize,
+        end: usize,
+    },
+}
+
+impl<'a> BorrowedPlainPayload<'a> {
+    /// Create a payload with a single byte slice
+    pub fn new_single(payload: &'a [u8]) -> Self {
+        Self::Single(payload)
+    }
```
cpu, line 553: This gets called fairly frequently through the change set. I wonder if it would make sense to impl `From<&'a [u8]>` for this purpose and leave just `new` for the from chunks constructor. WDYT?

rustls/src/msgs/message.rs
```diff
@@ -459,6 +459,149 @@ impl<'a> BorrowedPlainMessage<'a> {
...
+        match self {
+            Self::Empty => (Self::Empty, Self::Empty),
+            Self::Single(chunk) => {
+                if chunk.len() < mid {
+                    return (self.clone(), Self::Empty);
+                }
+                (Self::Single(&chunk[..mid]), Self::Single(&chunk[mid..]))
+            }
+            Self::Multiple { chunks, start, end } => {
+                let mut size = 0usize;
```
cpu, line 602: I think we can use just `0` here - the compiler seems able to infer the `usize` type from use.

rustls/src/msgs/deframer.rs
```diff
@@ -186,9 +187,9 @@ impl MessageDeframer {
 
             // If we don't know the payload size yet or if the payload size is larger
             // than the currently buffered payload, we need to wait for more data.
-            let raw = RawSlice::from(msg.payload);
+            // let raw = RawSlice::from(msg.payload);
```
cpu, line 186: let's delete this leftover.

rustls/src/msgs/message.rs
```diff
@@ -406,26 +414,48 @@ impl TryFrom<PlainMessage> for Message<'static> {
...
+pub trait BorrowedPlainMessage: Sized {
+    fn typ(&self) -> ContentType;
+    fn version(&self) -> ProtocolVersion;
+
+    fn payload_to_vec(&self) -> Vec<u8>;
+
+    fn payload_len(&self) -> usize;
+    fn encoded_len(&self, record_layer: &RecordLayer) -> usize {
+        OpaqueMessage::HEADER_SIZE as usize + record_layer.encrypted_len(self.payload_len())
+    }
```
cpu, line 437: nit: I think this would look better with consistent spacing between the trait fns at the top, 1 line of whitespace each.

rustls/src/msgs/message.rs
```diff
@@ -406,26 +414,48 @@ impl TryFrom<PlainMessage> for Message<'static> {
...
+    fn into_owned(self) -> PlainMessage {
+        PlainMessage {
+            version: self.version(),
+            typ: self.typ(),
+            payload: Payload::Owned(self.payload_to_vec()),
+        }
+    }
+
+    /// Returns true if the payload is a CCS message
+    fn is_ccs(&self) -> bool;
```
cpu, line 456: I think it might be worth duplicating some of the context for why we care about a message being plaintext CCS or not from the comment at the call-site to here.

rustls/src/msgs/message.rs
```diff
@@ -435,36 +465,65 @@ impl<'a> TryFrom<BorrowedPlainMessage<'a>> for Message<'a> {
...
-        OpaqueMessage {
-            version: self.version,
-            typ: self.typ,
-            payload: Payload::Owned(self.payload.to_vec()),
-        }
+impl BorrowedPlainMessage for BorrowedPlainMessageInbound<'_> {
+    fn typ(&self) -> ContentType {
+        self.typ
+    }
+    fn version(&self) -> ProtocolVersion {
```
cpu, line 478: nit: whitepsace between fns in the `impl`

rustls/src/msgs/message.rs
```diff
@@ -435,36 +465,65 @@ impl<'a> TryFrom<BorrowedPlainMessage<'a>> for Message<'a> {
...
+    pub typ: ContentType,
+    pub version: ProtocolVersion,
+    pub payload: BorrowedPlainPayload<'a>,
+}
+
+impl BorrowedPlainMessage for BorrowedPlainMessageOutbound<'_> {
+    fn typ(&self) -> ContentType {
+        self.typ
+    }
+    fn version(&self) -> ProtocolVersion {
```
cpu, line 512: same nit here w.r.t whitespace

cpu (commented):
Thanks again for picking this branch up and iterating with me. I think we're just about ready to follow up with some other folks for review.

Beyond the comments I left in my review I think we should clean up some of the intermediate work in the commit history. I think the right end state would be two commits:
  1. The first, doing the `BorrowedPlainMessage` split into `BorrowedPlainMessageInbound`/`BorrowedPlainMessageOutbound` leaving the existing code as-is as much as possible and just introducing the new trait/types. I think this commit message should touch on the rationale for the change, e.g. that we want to handle the payload differently for one of the two w.r.t chunks.
  2. The second, implementing the borrowed payload/chunker. I think this commit message should keep the existing comment from the original first commit in this branch, but also be augmented with a bit more of the context from the PR description. It doesn't need to have all the `strace` output but I think it's important to mention the performance motivation.
 
 Does that make sense to you?

Keksoj:
We certainly intend to edit the commit history. We thought of having two commits:

1. chunked payload
2. Inbound/outbound

But it is more elegant the other way around, as you suggest. Thanks for the review! We're pretty excited about this!

rustls/src/conn.rs
```diff
@@ -854,9 +849,10 @@ impl<Data> ConnectionCore<Data> {
                 .may_receive_application_data
             && self.common_state.is_tls13()
         {
-            if !is_valid_ccs(&msg)
-                || self.common_state.received_middlebox_ccs > TLS13_MAX_DROPPED_CCS
-            {
+            // We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
+            // Note: this is prior to the record layer, so is unencrypted. See
+            // third paragraph of section 5 in RFC8446.
+            if !msg.is_ccs() || self.common_state.received_middlebox_ccs > TLS13_MAX_DROPPED_CCS {
```
ctz, line 855: Here `msg` is statically of type `BorrowedPlainMessageInbound`, and this is the only call of `is_ccs`. That means, I think, that `is_ccs` can be removed from the `BorrowedPlainMessage` trait.

rustls/src/msgs/message.rs
```diff
@@ -3,17 +3,24 @@ use crate::enums::{AlertDescription, ContentType, HandshakeType};
...
-use super::codec::ReaderMut;
+#[derive(Debug)]
+pub enum MessageError {
+    TooShortForHeader,
+    TooShortForLength,
+    InvalidEmptyPayload,
+    MessageTooLarge,
+    InvalidContentType,
+    UnknownProtocolVersion,
+}
```
cpu, line 23: I think we want this `enum` to be located between `BorrowedOpaqueMessage` and `read_opaque_message_header` to match the [top-down ordering](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#top-down-ordering-within-modules) convention

It should be after `BorrowedOpaqueMessage` and `OpaqueMessage` because those types rely on it. It should be before `read_opaque_message_header` because the error type is `pub` and that helper isn't.

rustls/src/msgs/message.rs
cpu: (I think this nit is still unaddressed - not a blocker, just calling it out in case it was accidental)

cpu (approved):
This is looking good to me! @djc @ctz I think this branch is ready for you to review now.

ctz (approved):
Would be good to have one test that defends the desired result of this PR. Here's an example patch:

```diff
diff --git a/rustls/tests/api.rs b/rustls/tests/api.rs
index 2348e6d9..325bb176 100644
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -2277,6 +2277,25 @@ fn test_server_stream_read(stream_kind: StreamKind, read_kind: ReadKind) {
     }
 }
 
+#[test]
+fn test_client_vectored_write_yields_one_tls_record() {
+    let (mut client, mut server) = make_pair(KeyType::Rsa);
+    do_handshake(&mut client, &mut server);
+
+    const N: usize = 1000;
+    let data: Vec<IoSlice> = std::iter::repeat(IoSlice::new(b"A")).take(N).collect();
+
+    client.writer().write_vectored(&data).unwrap();
+    let byte_count = transfer(&mut client, &mut server);
+
+    println!("took {byte_count} for sending {N} bytes separately");
+    const MAX_REASONABLE_ENCRYPTION_OVERHEAD: usize = 128;
+
+    // should be N + 5 + (small encryption overhead) if this is working properly.
+    // if it is broken, it will be N * (5 + small encryption overhead).
+    assert!(byte_count < N + 5 + MAX_REASONABLE_ENCRYPTION_OVERHEAD);
+}
+
 struct FailsWrites {
     errkind: io::ErrorKind,
     after: usize,

```

rustls/src/msgs/message.rs
```diff
@@ -539,3 +539,257 @@ impl BorrowedPlainMessage for BorrowedPlainMessageOutbound<'_> {
...
+    }
+}
+
+impl<'a> From<BorrowedPlainPayload<'a>> for Vec<u8> {
+    fn from(payload: BorrowedPlainPayload<'a>) -> Self {
+        payload.to_vec()
+    }
+}
+
+impl<'a> From<BorrowedPlainPayload<'a>> for Payload<'a> {
```
djc, line 690: We generally keep `impl`s closer to their implementing type definitions, but I agree that this is more tightly coupled to `BorrowedPlainPayload` than to `Payload`. Maybe reverse this to be an `Into`-only impl (and suppress the lint if necessary)?

djc (commented):
This looks like a good change! I have a bunch of small suggestions -- hopefully not too hard to work through.

I don't feel like I've fully grokked the new buffering scheme yet, but I'm happy to trust other reviewers on that.

rustls/src/msgs/message.rs
```diff
@@ -406,65 +421,121 @@ impl TryFrom<PlainMessage> for Message<'static> {
...
-/// `BorrowMessage`, but this involves a copy.
+/// This type borrows its decrypted payload from a `MessageDeframer`.
+/// You can make a `OpaqueMessage` from an `BorrowedPlainMessageInbound`,
+/// but this involves a copy.
 ///
 /// This type also cannot decode its internals and
 /// cannot be read/encoded; only `OpaqueMessage` can do that.
 #[derive(Debug)]
-pub struct BorrowedPlainMessage<'a> {
+pub struct BorrowedPlainMessageInbound<'a> {
```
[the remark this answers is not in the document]
Wonshtrum: Sorry, I didn't see this discussion before posting https://github.com/rustls/rustls/pull/1774#issuecomment-1932096207.

I agree with ctz about `Borrowed` not telling much and I would argue the ownership of a type in general is not helpful. To me it feels like the ownership and mutability (and all other requirements) can be derived from the data format and "direction", but not the other way around.
If we have both `OpaqueMessage` and `OwnedOpaqueMessage` it's not clear why we don't only use the former.
But if we have `InboundOpaqueMessage` and `OutboundOpaqueMessage` it's immediately clear when to use one or the other. The fact that `OutboundOpaqueMessage` owns its data is an implementation detail, we do not use it because it owns its data, we use it to process outgoing messages and its data is owned because the encrypters allocate and copy plaintext before encrypting it.

rustls/src/msgs/message.rs
```diff
@@ -539,3 +539,257 @@ impl BorrowedPlainMessage for BorrowedPlainMessageOutbound<'_> {
         self.payload.len()
     }
 }
+
+#[derive(Debug, Clone)]
+/// A collection of borrowed plaintext slices
+pub enum BorrowedPlainPayload<'a> {
```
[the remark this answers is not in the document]
Wonshtrum: It is a field of `OutboundBorrowedPlainMessage` which is at the bottom, following the style guide `OutboundChunks` should be beneath, how do you suggest the file should be reordered?

Wonshtrum:
@ctz I like the idea of defending the desired result of this PR with a test. How strict should I make this test?
Potentially, we could test that `write_vectored` is equivalent to a single `write` with its data aggregated, as the `std::io::Write::write_vectored` documentation says:
> This method must behave as a call to [write] with the buffers concatenated would

```rs
fn test_client_write_and_vectored_write_equivalence() {
    let (mut client, mut server) = make_pair(KeyType::Rsa);

    const N: usize = 1000;

    let data_chunked: Vec<IoSlice> = std::iter::repeat(IoSlice::new(b"A"))
        .take(N)
        .collect();
    client
        .writer()
        .write_vectored(&data_chunked)
        .unwrap();
    let byte_count_chunked = transfer(&mut client, &mut server);
    println!("took {byte_count_chunked} for sending {N} bytes separately");

    let data_contiguous = &[b'A'; N];
    client
        .writer()
        .write(data_contiguous)
        .unwrap();
    let byte_count_contiguous = transfer(&mut client, &mut server);
    println!("took {data_contiguous} for sending {N} bytes contiguously");

    assert_eq!(byte_count_chunked, byte_count_contiguous);
}
```

rustls/src/msgs/message.rs
Wonshtrum: This `From<BorrowedPlainPayload>` implementation for `Payload` as well as `Vec<u8>` are unused, I propose to simply remove them.

rustls/src/msgs/message.rs
```diff
@@ -539,3 +539,257 @@ impl BorrowedPlainMessage for BorrowedPlainMessageOutbound<'_> {
...
+
+impl<'a> From<BorrowedPlainPayload<'a>> for Vec<u8> {
+    fn from(payload: BorrowedPlainPayload<'a>) -> Self {
+        payload.to_vec()
+    }
+}
+
+impl<'a> From<BorrowedPlainPayload<'a>> for Payload<'a> {
+    fn from(payload: BorrowedPlainPayload<'a>) -> Self {
+        match payload {
```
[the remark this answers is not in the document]
Wonshtrum: I use `Self`, but I personally don't like the `use enum::*`, as it was renamed `OutboundChunks` the match arms are quite shorter, should I still apply this change?

rustls/src/msgs/message.rs
```diff
@@ -332,21 +319,49 @@ pub struct PlainMessage {
...
-            version: self.version,
             typ: self.typ,
+            version: self.version,
             payload: self.payload,
         }
     }
 
-    pub fn borrow(&self) -> BorrowedPlainMessage<'_> {
-        BorrowedPlainMessage {
+    pub fn borrow_inbound(&self) -> BorrowedPlainMessageInbound<'_> {
```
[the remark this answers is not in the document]
Wonshtrum: I quite like how it's a clear split from the old `PlainMessage::borrow`, and I think it follows the pattern:
```
OpaqueMessage::borrow -> BorrowedOpaqueMessage
OpaqueMessage::into_plain_message -> PlainMessage

PlainMessage::borrow_inbound -> InboundBorrowedPlainMessage
PlainMessage::borrow_outbound -> OutboundBorrowedPlainMessage
PlainMessage::into_unencrypted_opaque -> OpaqueMessage
```
But if you really prefer, I can do `From` impls instead.

ctz:
> @ctz I like the idea of defending the desired result of this PR with a test. How strict should I make this test? Potentially, we could test that `write_vectored` is equivalent to a single `write` with its data aggregated, as the `std::io::Write::write_vectored` documentation says:

Aha, that is much better! 👍🏽

rustls/src/msgs/message.rs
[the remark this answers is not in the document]
djc: I can live with leaving this out.

rustls/src/msgs/message.rs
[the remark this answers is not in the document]
djc: Probably both should move up a bunch?

djc:
> For types names, I used `InboundBorrowedPlainMessage` and `OutboundBorrowedPlainMessage` (putting the "direction" first) for the split variants of `BorrowedPlainMessage` and `OutboundChunks` for `OutboundBorrowedPlainMessage::payload` (instead of `BorrowedPlainPayload`). Is that alright? I think most messages and payload types could be renamed, but as you said, it's better to keep the changes of this PR as atomic as possible.

For types that are introduced/renamed in this PR IMO we should pick the shorter names here instead of churning on them in another PR. So `InboundBorrowedPlainMessage` -> `InboundMessage`.

rustls/src/conn.rs
```diff
@@ -335,15 +334,6 @@ impl ConnectionRandoms {
     }
 }
 
-// --- Common (to client and server) connection functions ---
-
-fn is_valid_ccs(msg: &BorrowedPlainMessage) -> bool {
```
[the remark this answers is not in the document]
djc: This change seems to have disappeared entirely? I think we should still keep it in this PR, just as a separate commit.

rustls/src/msgs/message.rs
```diff
@@ -248,6 +245,16 @@ impl<'a> BorrowedOpaqueMessage<'a> {
     }
 }
 
+#[derive(Debug)]
+pub enum MessageError {
```
djc, line 249: Please make this move in a separate commit.

rustls/src/msgs/message.rs
```diff
@@ -406,65 +421,110 @@ impl TryFrom<PlainMessage> for Message<'static> {
...
-    fn try_from(plain: BorrowedPlainMessage<'a>) -> Result<Self, Self::Error> {
+    fn try_from(plain: InboundBorrowedPlainMessage<'a>) -> Result<Self, Self::Error> {
         Ok(Self {
             version: plain.version,
             payload: MessagePayload::new(plain.typ, plain.version, plain.payload)?,
         })
     }
 }
 
+pub trait BorrowedPlainMessage: Sized {
```
djc, line 435: This trait definition should go below the types implementing it.

rustls/src/msgs/message.rs
```diff
@@ -528,3 +528,226 @@ impl BorrowedPlainMessage for OutboundBorrowedPlainMessage<'_> {
...
+            Self::Multiple { start, end, .. } => end - start,
+        }
+    }
+
+    /// Returns true if the payload is empty
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    fn new_with_cursors(chunks: &'a [&'a [u8]], start: usize, end: usize) -> Self {
```
djc, line 635: As mentioned before, please move this up closer with the other constructor.

rustls/src/msgs/message.rs
```diff
@@ -528,3 +528,226 @@ impl BorrowedPlainMessage for OutboundBorrowedPlainMessage<'_> {
...
+    pub fn len(&self) -> usize {
+        match self {
+            Self::Empty => 0,
+            Self::Single(chunk) => chunk.len(),
+            Self::Multiple { start, end, .. } => end - start,
+        }
+    }
+
+    /// Returns true if the payload is empty
+    pub fn is_empty(&self) -> bool {
```
djc, line 631: This should go above `is_empty()`.

rustls/src/conn.rs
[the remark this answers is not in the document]
Wonshtrum: Ho ok, I missed that part. Should I put `is_valid_css` as a method of `InboundBorrowedPlainMessage` only? Or does it make sense for the outbound variant too?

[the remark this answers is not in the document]
ctz: I think it only makes sense for inbound messages, and there is only one call to it with that type.

rustls/src/msgs/message.rs
Wonshtrum: May I suggest it is done in a broader reorder of the file (as you suggested [here](https://github.com/rustls/rustls/pull/1640#discussion_r1484045051)), not in this PR?

Wonshtrum:
As requested I replaced:
- `InboundBorrowedPlainMessage` with `InboundMessage`
- `OuboundBorrowedPlainMessage` with `OutboundMessage`

I don't want the discussion on types to trail much longer, and as discussed multiple times this subject may be entirely revisited in a separate, dedicated PR.
But I would like to say I'm still not sure about stripping the "Plain" qualifier. For me, it seems to be important information. For instance, `OpaqueMessage` is used to hold messages that are outbound, but it is not equivalent to the newly created `OutboundMessage`. On top of that, this renaming clashes with other parts of the code. Three examples:
- `PlainMessage` still exists, and we can't strip its prefix, because `Message` also exists.
- The common trait for `InboundMessage` and `OutboundMessage` is still named `BorrowedPlainMessage`.
- `BorrowedOpaqueMessage` has an `into_inbound_plain_message` method, which we could rename `into_inbound_message`. But doing so might obscure the fact that `into_inbound_plain_message` performs a forced conversion between opaque and plain.

Wonshtrum:
I'm not sure but I might have caught a bug in `OutboundChunks` (which may cause an out-of-bound panic) even though all tests pass. Just in case, please do not merge this PR, I will investigate.

djc (approved):

rustls/src/msgs/message.rs
djc: That's fine too.

djc:
> I'm not sure but I might have caught a bug in `OutboundChunks` (which may cause an out-of-bound panic) even though all tests pass. Just in case, please do not merge this PR, I will investigate.

Converting to draft to make this more obvious.

Wonshtrum:
Regarding the suspected out-of-bound panic, I ultimately couldn't get any undefined behavior from the new `OutboundChunks`.
I will nonetheless point out what worried me:
- `split_at` can generate the `Multiple` variant with 2 slices even when equivalent to a `Single` (with the start index at the end of the first slice, or the end index at the start of the second).
- `split_at` can generate the `Multiple` variant with 2 slices even when equivalent to an `Empty`.
- `split_at` or the user can instantiate (through `write`) a `Single` variant with an empty slice equivalent to an `Empty`.
- users can instantiate (through `write_vectored`) a `Multiple` variant with one or more empty chunks, which wasn't tested.

Technically, none of those points break any invariants of `OutboundChunks`, I just don't particularly like when a state has multiple valid ways to be represented (because it usually means multiple code paths to handle them). We could constrain `OutboundChunks` to always instantiate the simplest variant, or we could reduce the number of possible representations (by removing `Empty` whose only purpose is to be simple to create, or even remove `Single`, whose purpose is to avoid a double pointer indirection).
But since it works as is, I didn't change anything. It is up to your preferences and style guidelines. The representation we chose and how it is instantiated has lots of geeky implications on the generated assembly (I had a lot of fun in godbolt), but I don't think this struct needs to be particularly optimized as is it pretty much passively passed around for all its lifetime.

Note: [this change](https://github.com/rustls/rustls/compare/30eb2960c151b24d7c63b3e46893d057bdbfef76..0566dc1e5925e11ebdb34bc7427a4afaaa87be9b#diff-c52965cd3966f64e12eef76ec5478491687f74bda15096241be6a80df0029875R622) has nothing to do with my initial worries, it's just an optimization to avoid 2 bound checks.

Wonshtrum:
I rebased on main.
For my part, I consider this PR done.

djc:
> Note: [this change](https://github.com/rustls/rustls/compare/30eb2960c151b24d7c63b3e46893d057bdbfef76..0566dc1e5925e11ebdb34bc7427a4afaaa87be9b#diff-c52965cd3966f64e12eef76ec5478491687f74bda15096241be6a80df0029875R622) has nothing to do with my initial worries, it's just an optimization to avoid 2 bound checks.

I think this diff link is showing the opposite change of what you actually meant. I like the actual change, though!

djc:
> Technically, none of those points break any invariants of `OutboundChunks`, I just don't particularly like when a state has multiple valid ways to be represented (because it usually means multiple code paths to handle them). We could constrain `OutboundChunks` to always instantiate the simplest variant, or we could reduce the number of possible representations (by removing `Empty` whose only purpose is to be simple to create, or even remove `Single`, whose purpose is to avoid a double pointer indirection).

I don't have a very strong opinion, but removing `Empty` in favor of an empty slice in `Single` sounds kinda appealing? Otherwise it looks like `OutboundChunks` provides a decent abstraction so it doesn't feel like also removing `Single` in favor of having a `Multi` with some otherwise enforced invariants would be a great improvement.

rustls/src/msgs/message.rs
```diff
@@ -419,46 +424,242 @@ impl<'a> TryFrom<BorrowedPlainMessage<'a>> for Message<'a> {
...
+    fn payload_len(&self) -> usize;
+
+    fn typ(&self) -> ContentType;
+
+    fn version(&self) -> ProtocolVersion;
+}
+
+#[derive(Debug, Clone)]
+/// A collection of borrowed plaintext slices.
+/// Warning: OutboundChunks does not garantee that the simplest variant is used.
```
cpu, line 536: small typo here:

```suggestion
/// Warning: OutboundChunks does not guarantee that the simplest variant is used.
```

cpu (approved):

Wonshtrum:
> I think this diff link is showing the opposite change of what you actually meant. I like the actual change, though!

Unfortunately, this is the right diff, I changed the `enumerate` for a plain range loop. This micro-optimization works because the plain loop couples strongly the index to the slice's length, whereas, due to how it is written in the standard lib, the enumerated index is (from the compiler's point of view) completely independent of the slice (so any indexing in the slice must be bound checked).

> I don't have a very strong opinion, but removing Empty in favor of an empty slice in Single sounds kinda appealing? Otherwise it looks like OutboundChunks provides a decent abstraction so it doesn't feel like also removing Single in favor of having a Multi with some otherwise enforced invariants would be a great improvement.

I removed `Empty` from `OutboundChunks`. The code can remain sensibly the same, just by removing the corresponding match arms. But just to see I tried to variations with less invariants.

First I removed the "at least two slices in Multiple" invariant. This simplifies the `split_at` which now avoids "downgrading" to `Single`. The private `with_cursors` constructor can be removed. The `copy_to_vec` has to be modified to handle the case where `Multiple` holds only one chunk with both cursors in it.
```rs
pub fn copy_to_vec(&self, vec: &mut Vec<u8>) {
    match self {
        Self::Single(chunk) => vec.extend_from_slice(chunk),
        Self::Multiple { chunks, start, end } => {
            let mut size = 0;
            let last = chunks.len() - 1;
            for (i, chunk) in chunks.iter().enumerate() {
                let start = if i == 0 { *start } else { 0 };
                let end = if i == last { end - size } else { chunk.len() };
                vec.extend_from_slice(&chunk[start..end]);
                size += chunk.len();
            }
        }
    }
}
pub fn split_at(&self, mid: usize) -> (Self, Self) {
    match self {
        Self::Single(chunk) => {
            if chunk.len() <= mid { return (self.clone(), Self::Single(&[]); }
            (Self::Single(&chunk[..mid]), Self::Single(&chunk[mid..]))
        }
        Self::Multiple { chunks, start, end } => {
            for i in 0..chunks.len() {
                if size + chunks[i].len() >= (mid + start) {
                    return (
                        Self::Multiple { chunks: chunks[..=i], start: *start, end: start + mid },
                        Self::Multiple { chunks: chunks[i..], start: start + mid, end: *end },
                    );
                }
                size += chunks[i].len();
            }
            (self.clone(), Self::new_empty())
        }
    }
}
```
Then I removed the "start must be in the first chunk" and "end must be in the last chunk" invariants. The only invariant this version assumes is that "start is smaller than end". This simplifies again `split_at` which is now O(1) as it only has to change the cursors of `Multiple`. Again `copy_to_vec` takes a bit more complexity as we don't know anymore in which chunks the cursors point.
```rs
pub fn copy_to_vec(&self, vec: &mut Vec<u8>) {
    match self {
        Self::Single(chunk) => vec.extend_from_slice(chunk),
        Self::Multiple { chunks, start, end } => {
            let mut size = 0;
            for chunk in *chunks {
                let psize = size;
                let len = chunk.len();
                size += len;
                if size <= *start || psize >= *end { continue; }
                let start = if psize < *start { start - psize } else { 0 };
                let end = if end - psize < len { end - psize } else { len };
                vec.extend_from_slice(&chunk[start..end]);
            }
        }
    }
}
pub fn split_at(&self, mid: usize) -> (Self, Self) {
    match self {
        Self::Single(chunk) => {
            if chunk.len() <= mid { return (self.clone(), Self::Single(&[]); }
            (Self::Single(&chunk[..mid]), Self::Single(&chunk[mid..]))
        }
        Self::Multiple { chunks, start, end } => (
            Self::Multiple { chunks, start: *start, end: start + mid },
            Self::Multiple { chunks, start: start + mid, end: *end },
        )
    }
}
```

It's hard to create a representative benchmark for `OutboundChunks`, but these two propositions seem a little faster than the current implementation. The second produces a smaller assembly, but they seem equivalent in performance.

djc:
> It's hard to create a representative benchmark for `OutboundChunks`, but these two propositions seem a little faster than the current implementation. The second produces a smaller assembly, but they seem equivalent in performance.

This sounds good to me! Thanks for diving in some more. Maybe `split_at()` for `Single` can be `(Self::Single(&chunk[..Ord::min(mid, chunk.len())]), Self::Single(&chunk[mid..]))`? Avoids a bit of repetition.

djc:
@Wonshtrum one more clippy issue to address!

Also looks like our external types workflow check broke, so we'll need to look into that (but not need to block this PR I think).

Wonshtrum:
Sorry about that.
Fixed and rebased on main.

cpu:
> Also looks like our external types workflow check broke, so we'll need to look into that (but not need to block this PR I think).

This was fixed in main w/ https://github.com/rustls/rustls/pull/1791

I think we probably just need Ctz to do another pass on this branch and then it's ready for merge? I suspect based on OOB conversations this will be doable early next week :crossed_fingers:

ctz (approved):

djc:
@Wonshtrum thanks for all your work on this!

PR #1645, opened 2023-11-30
ci-bench: increase threshold

jsha:
The old threshold was 0.2%, but it was triggering on doc-only changes, with diffs as high as 1.74%. I'm not sure why this measure is being noisier than it is supposed to be, but since it's causing most CI builds to get a red "x", it's worth bumping the threshold for now.

Example jobs:

https://github.com/rustls/rustls/actions/runs/7039851436?pr=1637
https://github.com/rustls/rustls/actions/runs/7040127211?pr=1643

Also, change the order of the CI job so the comparison is done using the code in the current PR. That means that changes to the ci-bench tooling can be reflected immediately in the CI runs for the current PR.

### DIFF
--- a/.github/workflows/icount-bench.yml
+++ b/.github/workflows/icount-bench.yml
@@ -15,14 +15,6 @@ jobs:
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Checkout PR
-        uses: actions/checkout@v4
-        with:
-          persist-credentials: false
-
-      - name: Run icount benchmarks for PR
-        run: cd ci-bench && cargo run --locked --release -- run-all --output-dir ${{ runner.temp }}/pr
-
       - name: Checkout ${{ github.base_ref }}
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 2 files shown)


djc (approved):
It seems like aws-lc-rs is substantially less deterministic than *ring* -- might be useful if we could set the threshold differently for those two. But failing that, the failure of the icount job had been pretty much noise at this point so this would be an improvement.

cpu (approved):
Thanks

ctz:
> It seems like aws-lc-rs is substantially less deterministic than _ring_

Quite. The extra RSA key validation that aws-lc does is non-deterministic, and expensive. That's why the false positives are always: on the server side tests, with RSA keys, for tests that are otherwise very fast (resumed handshakes) so the difference in key validation contributes a large proportional share of the runtime.

I think we should avoid throwing the baby out with the bathwater, and just ignore these specific test cases. Will push a commit that does this in a moment.

cpu:
> I think we should avoid throwing the baby out with the bathwater, and just ignore these specific test cases. Will push a commit that does this in a moment.

Sounds good to me. Thanks for taking a look

cpu:
Fixed w/ #1649

Issue #1435, opened 2023-08-31
0.22 release planning


djc:
Let's gather some context on what we should fix before pushing out the next semver-incompatible release.

- [x] https://github.com/rustls/rustls/issues/1437
- [x] https://github.com/rustls/rustls/issues/1472
- [x] https://github.com/rustls/webpki/issues/159
- [x] #1414
- [x] #1492 
- [x] https://github.com/rustls/rustls/issues/1386
- [x] https://github.com/rustls/rustls/issues/1409
- [x] https://github.com/rustls/rustls/pull/1432
- [x] https://github.com/rustls/rustls/pull/1405
- [x] https://github.com/rustls/rustls/pull/1448
- [x] https://github.com/rustls/rustls/pull/1430 (needs a solution for verifier composability that doesn't involve webpki leaking into rustls API)

Some potential validation we could do:

- [x] Port tokio-rustls to 0.22
- [ ] Port hyper-rustls to 0.22 - https://github.com/rustls/hyper-rustls/pull/222
- [ ] Port quinn to 0.22
- [ ] Port trust-dns to 0.22

djc:
@jsha anything else on your mind?

jsha:
Filed #1437; that's all I can think of at the moment. Thanks for organizing this list.

djc:
Some updates:

* Added #1437 to the list
* Replace tentative bullet for `CryptoProvider` dynamic dispatch with link to #1448 
* Added #1472 to the list

ctz:
I think I'd like to get #1414 in too.

ctz:
As mentioned on that ticket, I think we should drop #1372 for 0.22 and do it 0.23.

djc:
What's your rationale for wanting to push it out to the next release?

ctz:
I think there's some imperative to release 0.22 pretty soon, as PRs we don't really need or want in there are getting closer and/or landing already. I don't want to be keeping back PRs that are otherwise ready, or even worse backing them out, or having multiple trunks, etc.

cpu:
Selfishly I'd also like rustls-ffi to be ready for the 0.22 release as close to the time we cut the release as possible. Refactoring the builders will require another substantial-ish set of changes to https://github.com/rustls/rustls-ffi/pull/341 and getting those ready + reviewed will take more time.

djc:
Removed #1372 from the list.

jsha:
[...]

PR #1646, opened 2023-11-30
api: move CipherSuiteCommon into crypto

jsha:
The top level of the crate is meant for "paved path" exports.

In 0.21.x, this type was in `cipher_suites`, along with a few other types that got moved to specific crypto providers. Moving this to `crypto` instead of re-exporting under its old name in `cipher_suites` seems acceptable, because it will mainly be used in implementing crypto providers. Also, its internals have changed significantly so there is already churn for this type.

Related: #1435


djc (approved):

ctz (approved):
(needs a fixup in provider-example/)

cpu (approved):
I'm not sure if I've said this explicitly anywhere but I really appreciate how much attention you give to Rustls' public API and documentation. It's a lot of valuable polish and keeps the project user-focused in a meaningful way. Thank you!

PR #1648, opened 2023-11-30
[experiment / RFC] runtime-agnostic async API

japaric:
This PR contains an async adapter that works with both async-std and tokio but depends on neither crate. Instead it uses the `futures::Async{Read,Write}` interfaces to make the API, runtime generic.

- this is just a sketch (= I have not done extensive testing)
- the sketch only includes the client API
- a client example that can use either async-std or tokio is included
- there are no API docs but the API surface mimics the high level async API found in `rustls-tokio`
- the API is **not** no-std compatible because the `Async{Read,Write}` traits include the `std::io::Error` type in its interface
- the async API is implemented on top of the `UnbufferedConnection` API

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -689,12 +689,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
+name = "futures"
+version = "0.3.29"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


cpu:
@quininer If you had time this branch might be interesting to you as someone that has done a lot of work on rustls-tokio. Any feedback you could share would be helpful !

quininer:
This looks great, I have plans to use it in `tokio-rustls`, `futures-rustls` after `UnbufferedConnection` is released.

The `async` crate has a lot of TODOs, but i think direction is fine.

joshtriplett:
Suggestion: rather than using the full `futures`, consider using `futures-io` (https://crates.io/crates/futures-io), to reduce dependencies for people who don't need the full `futures`.

PR #1649, opened 2023-11-30
ci-bench: filter out noisy test cases from comparison

ctz:
See discussion on #1645 -- I also took [the second commit - 67b858c](https://github.com/rustls/rustls/pull/1645/commits/67b858ca22a61f8d05d329cf822e0a874185ef23) from there.


djc:
It's not completely obvious to me how we decide to pass/fail the icount-bench job, but should we institute another threshold for the known noisy jobs, instead of completely ignoring changes for them?

PR #1652, opened 2023-11-30
remove InvalidKeyError

jsha:
The per-provider key loading functions returned this singleton error, but it was usually then wrapped into Error::General("invalid private key"). That means the singleton error is unnecessary API surface, but also it means potentially valuable information is lost.

Move the wrapping into `Error::General` to a lower level, add detail about which specific parsing operation failed, and pass along error details from the lower-level library.

### DIFF
--- a/rustls/src/crypto/ring/sign.rs
+++ b/rustls/src/crypto/ring/sign.rs
@@ -11,16 +11,16 @@ use super::ring_like::signature::{self, EcdsaKeyPair, Ed25519KeyPair, RsaKeyPair
 use pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer};
 
 use alloc::boxed::Box;
+use alloc::format;
 use alloc::string::ToString;
 use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt::{self, Debug, Formatter};
-use std::error::Error as StdError;
 
 /// Parse `der` as any supported key encoding/type, returning
 /// the first which works.
-pub fn any_supported_type(der: &PrivateKeyDer<'_>) -> Result<Arc<dyn SigningKey>, InvalidKeyError> {
+pub fn any_supported_type(der: &PrivateKeyDer<'_>) -> Result<Arc<dyn SigningKey>, Error> {
     if let Ok(rsa) = RsaSigningKey::new(der) {
         Ok(Arc::new(rsa))
     } else if let Ok(ecdsa) = any_ecdsa_type(der) {
@@ -28,7 +28,9 @@ pub fn any_supported_type(der: &PrivateKeyDer<'_>) -> Result<Arc<dyn SigningKey>
     } else if let PrivateKeyDer::Pkcs8(pkcs8) = der {
         any_eddsa_type(pkcs8)
     } else {
-        Err(InvalidKeyError(()))
+        Err(Error::General(
+            "failed to parse private key as RSA, ECDSA, or EdDSA".into(),
+        ))
     }
 }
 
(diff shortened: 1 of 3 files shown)


rustls/src/crypto/ring/sign.rs
```diff
@@ -53,20 +55,16 @@ pub fn any_ecdsa_type(der: &PrivateKeyDer<'_>) -> Result<Arc<dyn SigningKey>, In
...
-) -> Result<Arc<dyn SigningKey>, InvalidKeyError> {
-    if let Ok(ed25519) = Ed25519SigningKey::new(der, SignatureScheme::ED25519) {
-        return Ok(Arc::new(ed25519));
-    }
-
+pub fn any_eddsa_type(der: &PrivatePkcs8KeyDer<'_>) -> Result<Arc<dyn SigningKey>, Error> {
     // TODO: Add support for Ed448
-
-    Err(InvalidKeyError(()))
+    let ed25519 = Ed25519SigningKey::new(der, SignatureScheme::ED25519)?;
```
djc, line 66: Nit: I think inlining the one-use trivial `ed25519` binding here would be more idiomatic.

djc (commented):
If we're going to do this, maybe these should just become `Error` variants? Would make it easier to do some programmatic fallback in case of key parsing failures.

ctz:
> If we're going to do this, maybe these should just become `Error` variants? Would make it easier to do some programmatic fallback in case of key parsing failures.

Not sure I agree. `Error::Other` is what I'd expect to be used for key parsing failures, really. I think `ring::error::KeyRejected` fulfils all the requirements for being used there? And it means we get more useful errors, though it is challenging for functions like `any_supported_type` that try a bunch of different things in succession.

djc:
Compared to stuffing pretty specific strings into `Error::General` (that don't do run-time formatting), it's not obvious to me why we wouldn't use new variants which are (a) cheaper to construct and (b) can be more easily handled by code. The cost of maintaining a few more variants seems neglible to me.

ctz:
In this case `badkey.pem` is a gratuitously-wrong PKCS#8 file:

```
-----BEGIN PRIVATE KEY-----
aGVsbG8gd29ybGQ=
-----END PRIVATE KEY-----
```

main:
```
$ cargo run --bin tlsserver-mio -- --certs ../test-ca/rsa/end.fullchain --key badkey.pem  echo
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `/home/jbp/rustls/target/debug/tlsserver-mio --certs ../test-ca/rsa/end.fullchain --key badkey.pem echo`
thread 'main' panicked at examples/src/bin/tlsserver-mio.rs:619:6:
bad certificates/private key: General("invalid private key")
```

this PR:
```
$ cargo run --bin tlsserver-mio -- --certs ../test-ca/rsa/end.fullchain --key badkey.pem  echo
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `/home/jbp/rustls/target/debug/tlsserver-mio --certs ../test-ca/rsa/end.fullchain --key badkey.pem echo`
thread 'main' panicked at examples/src/bin/tlsserver-mio.rs:619:6:
bad certificates/private key: General("failed to parse Ed25519 private key: InvalidEncoding")
```

while it's not incorrect, and there is more information, this error implies to me that this tool requires an ed25519 key. but, in reality, it swallows the errors for RSA and ECDSA keys but surfaces the EdDSA error.

ctz:
> The cost of maintaining a few more variants seems neglible to me.

Could you say what new `rustls::Error` variants you have in mind?

cpu:
Superceded by https://github.com/rustls/rustls/pull/1657 - thanks!

PR #1657, opened 2023-12-01
remove InvalidKeyError

ctz:
The per-provider key loading functions returned this singleton error, but it was usually then wrapped into Error::General("invalid private key"). That means the singleton error is unnecessary API surface, but also it means potentially valuable information is lost.

Move the wrapping into `Error::General` to a lower level, add detail about which specific parsing operation failed, and pass along error details from the lower-level library.

---

follow-up to #1652 -- this is that PR, but:

- elide an extra ed25519 binding: https://github.com/rustls/rustls/pull/1652#discussion_r1411795341
- early return, to avoid reporting ed25519 errors: https://github.com/rustls/rustls/pull/1652#issuecomment-1835866486
- add a test

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/mod.rs
+++ b/rustls/src/crypto/aws_lc_rs/mod.rs
@@ -9,7 +9,6 @@ use crate::Error;
 use pki_types::PrivateKeyDer;
 use webpki::aws_lc_rs as webpki_algs;
 
-use alloc::string::String;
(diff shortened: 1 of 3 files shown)


cpu (approved):
LGTM

djc (approved):

jsha:
Thanks for getting this across the finish line!

PR #1660, opened 2023-12-01
doc: remove crate:: prefix for links

jsha:
Instead, use `#[cfg(doc)]` to conditionally import names that we want to use in the docs. This provides a user-friendlier link name.

Also, remove rustdoc-style links from the section of src/lib.rs that gets extracted for the README, since they wind up as broken links in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,10 +93,10 @@ x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Bi
 support WebAssembly.
 For more information, see [the supported `ring` target platforms][ring-target-platforms].
 
-By providing a custom instance of the [`crate::crypto::CryptoProvider`] struct, you
+By providing a custom instance of the [`crypto::CryptoProvider`] struct, you
 can replace all cryptography dependencies of rustls.  This is a route to being portable
 to a wider set of architectures and environments, or compliance requirements.  See the
-[`crate::crypto::CryptoProvider`] documentation for more details.
+[`crypto::CryptoProvider`] documentation for more details.
 
 Specifying `default-features = false` when depending on rustls will remove the
 dependency on *ring*.
@@ -104,6 +104,7 @@ dependency on *ring*.
 Rustls requires Rust 1.61 or later.
 
 [ring-target-platforms]: https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/include/ring-core/target.h#L18-L64
+[crypto::CryptoProvider]: https://docs.rs/rustls/latest/rustls/crypto/trait.CryptoProvider.html
 
 # Example code
 There are two example programs which use
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -63,10 +63,10 @@
 //! support WebAssembly.
 //! For more information, see [the supported `ring` target platforms][ring-target-platforms].
 //!
-//! By providing a custom instance of the [`crate::crypto::CryptoProvider`] struct, you
+//! By providing a custom instance of the [`crypto::CryptoProvider`] struct, you
 //! can replace all cryptography dependencies of rustls.  This is a route to being portable
 //! to a wider set of architectures and environments, or compliance requirements.  See the
-//! [`crate::crypto::CryptoProvider`] documentation for more details.
+//! [`crypto::CryptoProvider`] documentation for more details.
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -93,10 +93,10 @@ x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Bi
 support WebAssembly.
 For more information, see [the supported `ring` target platforms][ring-target-platforms].
 
-By providing a custom instance of the [`crate::crypto::CryptoProvider`] struct, you
+By providing a custom instance of the crypto::CryptoProvider struct, you
```
cpu, line 96: Instead of removing the links what about specifying an explicit target? E.g. like the `[ring-target-platforms]` instance. That seems to work in both Rustdocs and the README.

rustls/src/lib.rs
```diff
@@ -63,10 +63,10 @@
 //! support WebAssembly.
 //! For more information, see [the supported `ring` target platforms][ring-target-platforms].
 //!
-//! By providing a custom instance of the [`crate::crypto::CryptoProvider`] struct, you
+//! By providing a custom instance of the crypto::CryptoProvider struct, you
```
djc, line 66: So I guess the syncing of crate root docs and README is a good reason to keep them there in the README, too?

jsha: I don't understand the question.

cpu (approved):

ctz:
It's a little disappointing that `[crate::Something]` doesn't produce reasonable output -- maybe I'm wrong, but I'd expect the `crate::` path to be dropped, as it doesn't make sense in the context of hyperlinked documentation.

ctz:
Maybe we could be doing:

```
diff --git a/rustls/src/builder.rs b/rustls/src/builder.rs
index 26de3adb..63f5a587 100644
--- a/rustls/src/builder.rs
+++ b/rustls/src/builder.rs
@@ -19,7 +19,7 @@ use std::sync::Arc;
 /// For settings besides these, see the fields of [`ServerConfig`] and [`ClientConfig`].
 ///
 /// The usual choice for protocol primitives is to call
-/// [`crate::ClientConfig::builder`]/[`ServerConfig::builder`]
+/// [`ClientConfig::builder`](crate::ClientConfig::builder)/[`ServerConfig::builder`]
 /// which will use rustls' default cryptographic provider and safe defaults for ciphersuites and
 /// supported protocol versions.
 ///
```

Instead of this? It's a bit more localised.

jsha:
```
[`ClientConfig::builder`](crate::ClientConfig::builder)
```

This works, and we've done in in one place (which I haven't update in this PR, yet). But it gets verbose, runs into line length issues, and makes the un-rendered docs hard to read.

There's another approach:

```
[`ClientConfig::builder`]

...

[`ClientConfig::builder`]: crate::ClientConfig::builder
```

This is more concise but you wind up having to repeat it for every doc section that mentions that item. And I've found the extra typing is pretty tedious.

One additional benefit of `use` statements is that they can take advantage of IDE autocomplete, while comments cannot. So typing out long item names twice in comments is riskier. Also there's the potential for undetected typos (in the link text, not the link target).

cpu (approved):

cpu:
@ctz Do you want to give this another pass?

rustls/src/webpki/server_verifier.rs
```diff
@@ -85,11 +88,11 @@ impl ServerCertVerifierBuilder {
     /// trust anchors, and to control how server certificate revocation checking is performed.
     ///
     /// If `with_signature_verification_algorithms` was not called on the builder, a default set of
-    /// signature verification algorithms is used, controlled by the selected [`crate::crypto::CryptoProvider`].
+    /// signature verification algorithms is used, controlled by the selected [`crypto::CryptoProvider`].
     ///
     /// Once built, the provided `Arc<dyn ServerCertVerifier>` can be used with a Rustls
-    /// [crate::server::ServerConfig] to configure client certificate validation using
-    /// [`with_client_cert_verifier`][crate::ConfigBuilder<ClientConfig, WantsVerifier>::with_client_cert_verifier].
+    /// [`ServerConfig``] to configure client certificate validation using
```
ctz, line 94: extraneous backtick here

jsha: Thanks, fixed!

ctz (approved):

PR #1664, opened 2023-12-04

djc:
This duplicates some windows libraries, but I think that only affects our internal crates/dev-dependencies.


djc:
Some nice performance improvements from *ring* 0.17.6!

PR #1665, opened 2023-12-04
quic: require that PacketKey and HeaderProtectionKey are Send + Sync

djc:
Note that this is technically semver-incompatible, but I think we should land it in 0.22.x anyway since expected breakage should be negligible.


cpu (approved):
Is this left as a draft because you anticipate needing to make more changes?

djc:
Yes, but let's just merge this.

djc:
I guess I should add some tests, but won't get to that until tomorrow.

ctz (approved):

PR #1668, opened 2023-12-05
remove the TLS 1.2 session ticket on DecryptError

### LINKED ISSUE #1617: Invalidating TLS 1.2 session tickets
apologies in advance because the issue is still unclear to me

we have some hyper-rustls clients that were connecting to a TLS 1.2 endpoint. At some point, those connections started failing in a loop, until they were restarted, and then it worked fine.

When looking at the traffic, a handshake was started by the client with TLS 1.2 session tickets, the server responded with its hello, change cipherspec and the handshake message. And right there the client sends a "Bad Record MAC" alert. The server answers with an encrypted alert, then the connection is closed by the client.

The session ticket was created by that server and should have been ok to reuse, but we don't control the server and don't have much info on its internals or what happened on the server side.

My wild guess is that there was something wrong with the ticket, maybe it was modified on the network and that server does not verify it, or maybe it's deserialized in a different way. But that would in the server using a different TLS state than the client, sending a message with it that the client cannot verify, so the client returns "Bad Record MAC".
And the issue I think is in rustls is that it does not remove tickets from the store, so on the next connection, it tries the same ticket and it fails reliably. It looks like the session ticket implementation was meant to be driven by the server, which would send a new session ticket to overwrite the old one.
But shouldn't the store delete the old ticket if there was an issue with the handshake?

Geal:
Fix #1617

if for some reason the recorded session ticket is invalid or decoded incorrectly by the server, we can get into a case where the resumption handshake happens, and right after the ChangeCipherSpec message, the server sends an encrypted handhsake message using the invalid ticket, and the client rejects it with the BadRecordMAC alert. Unfortunately, if the calling code retries the connection, if it will try again with the same ticket and obtain the same result. This commit makes sure that if we fail to decrypt the first message, we will remove the session ticket for this server, to start from cratch on the next connection.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -453,7 +453,7 @@ impl<Data> ConnectionCommon<Data> {
     pub(crate) fn first_handshake_message(&mut self) -> Result<Option<Message>, Error> {
         match self
             .core
-            .deframe()?
+            .deframe(None)?
             .map(Message::try_from)
         {
             Some(Ok(msg)) => Ok(Some(msg)),
@@ -635,7 +635,7 @@ impl<Data> ConnectionCore<Data> {
             }
         };
 
-        while let Some(msg) = self.deframe()? {
+        while let Some(msg) = self.deframe(Some(&*state))? {
             match self.process_msg(msg, state) {
                 Ok(new) => state = new,
                 Err(e) => {
@@ -650,7 +650,7 @@ impl<Data> ConnectionCore<Data> {
     }
 
     /// Pull a message out of the deframer and send any messages that need to be sent as a result.
-    fn deframe(&mut self) -> Result<Option<PlainMessage>, Error> {
+    fn deframe(&mut self, state: Option<&dyn State<Data>>) -> Result<Option<PlainMessage>, Error> {
         match self.message_deframer.pop(
             &mut self.common_state.record_layer,
             self.common_state.negotiated_version,
@@ -690,9 +690,14 @@ impl<Data> ConnectionCore<Data> {
             Err(err @ Error::PeerSentOversizedRecord) => Err(self
                 .common_state
                 .send_fatal_alert(AlertDescription::RecordOverflow, err)),
-            Err(err @ Error::DecryptError) => Err(self
-                .common_state
-                .send_fatal_alert(AlertDescription::BadRecordMac, err)),
+            Err(err @ Error::DecryptError) => {
+                if let Some(state) = state {
+                    state.handle_decrypt_error();
+                }
+                Err(self
+                    .common_state
+                    .send_fatal_alert(AlertDescription::BadRecordMac, err))
+            }
             Err(e) => Err(e),
         }
     }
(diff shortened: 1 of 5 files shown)


Geal:
I tried to keep the change minimal, but it results in choosing between a new specific method on the `State` trait, or a downcast from `State<Data>` to `ExpectFinished` in `ConnectionCore<Data>::process_new_packets`. If it is not satisfying, I'm open to suggestions to fix that

rustls/src/conn.rs
```diff
@@ -635,7 +635,12 @@ impl<Data> ConnectionCore<Data> {
             }
         };
 
-        while let Some(msg) = self.deframe()? {
```
djc, line 638: I think the formatting makes this pretty hard to follow. Maybe unroll the `while` loop into `loop { let msg = match self.deframe() { .. }; .. }` instead?

djc (commented):
I'm curious how the downcast thing looks? I think that might be better, since it makes it much clearer when this method is called.

rustls/src/conn.rs
ctz: I suggest we just put this in `fn deframe` a few lines below. It already has a specific match arm for `Error::DecryptError`.

Geal: right, that looks better like this: bc09dbf8

Geal:
>I'm curious how the downcast thing looks? I think that might be better, since it makes it much clearer when this method is called.

that would be a bit like this:
```rust
    if let Some(expect_finished) = state.downcast_ref::<ExpectFinished>() {
        expect_finished.remove_session_ticket();
    }
 ```

cpu:
Have you looked at adding a unit test covering this new behaviour?

Geal:
>Have you looked at adding a unit test covering this new behaviour?
that's what I'm trying to figure out right now. Are there tests for the state machine that I could get inspiration from?

cpu:
> Are there tests for the state machine that I could get inspiration from?

I'm not sure there's a great candidate to crib from. Maybe testing this at the integration test level in `api.rs` using a customized `ProducesTickets` impl that always fails `decrypt` would be sufficient to recreate the looping client situation you describe, and then demonstrate the loop no longer occurs?

ctz:
I have pushed a test for this in https://github.com/rustls/rustls/commit/682fa4f59d88766445e69a692bb526f17ccb8a02

djc:
@ctz your commit doesn't seem to appear in this branch, is it in a PR with this branch as a base?

ctz:
It's pushed as a separate branch based on this PR, so it could be cherry-picked/squashed into this PR.

Geal:
awesome, thank you for writing the test

ctz:
Yes - please do! Ideally the test could end up in the same commit that makes the change, if that doesn't make the commit too bloated. I think we're quite far from that on this PR!

Geal:
yes, I think we can squash it, it's small enough

ctz (approved):

cpu (approved):
Thanks again :-)

Geal:
alright, good to know!

PR #1674, opened 2023-12-07
Re-export the types crate contents

### LINKED ISSUE #1662: discussion: re-exporting rustls-pki-types
Pulling out a conversation from https://github.com/rustls/rustls/pull/1651#issuecomment-1836814664:

> One question I've had is if we should be re-exporting the pki-types, maybe as rustls::types? I think that would make life a bit easier? We could do the same across some of the other crates, I suppose.

My feeling is no. We can document that rustls depends on pki-types 1.0, but I think the pattern of re-exporting a crate in multiple places is confusing.

One of my very confusing experiences as a Rust beginner was noticing that [`hyper::Response`](https://docs.rs/hyper/latest/hyper/struct.Request.html) and [`http::Response`](https://docs.rs/http/latest/http/request/struct.Request.html) have the same documentation, the same methods, and so on, but are in different libraries. It's not documented that one is actually a re-export of the other, so I thought it might have been copy-pasted code or an older version of the same library. I wound up spending some time looking carefully for differences to see why these were different types, and of course found none.

djc:
Fixes #1662. (I think this should be on the list for backporting to 0.22.1.)

### DIFF
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -532,6 +532,11 @@ pub mod version {
     pub use crate::versions::TLS13;
 }
 
+/// Re-exports the contents of rustls-pki-types crate for easy access
+pub mod pki_types {
+    pub use pki_types::*;
+}
+
 /// Message signing interfaces.
 pub mod sign {
     pub use crate::crypto::signer::{CertifiedKey, Signer, SigningKey};
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -61,5 +61,6 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.cargo_check_external_types]
 allowed_external_types = [
+    "rustls_pki_types",
     "rustls_pki_types::*",
 ]


rustls/src/lib.rs
```diff
@@ -543,5 +543,10 @@ pub mod quic;
 /// APIs for implementing TLS tickets
 pub mod ticketer;
 
+/// Re-exports the contents of rustls-pki-types crate for easy access
+pub mod types {
```
djc, line 547: Alternatively, this could be named `pki_types` or `rustls_pki_types`.

djc:
The "Validate external types appearing in public API" jobs fail on this, even though we specifically have this exception in the crate metadata:

```
[package.metadata.cargo_check_external_types]
allowed_external_types = [
    "rustls_pki_types::*",
]
```

ctz:
> The "Validate external types appearing in public API" jobs fail on this, even though we specifically have this exception in the crate metadata:
> 
> ```
> [package.metadata.cargo_check_external_types]
> allowed_external_types = [
>     "rustls_pki_types::*",
> ]
> ```

I think this needs

```
diff --git a/rustls/Cargo.toml b/rustls/Cargo.toml
index bc8b5e77..46411aff 100644
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -62,4 +62,5 @@ rustdoc-args = ["--cfg", "docsrs"]
 [package.metadata.cargo_check_external_types]
 allowed_external_types = [
     "rustls_pki_types::*",
+    "rustls_pki_types",
 ]
```
(I don't particularly understand the distinction.)

rustls/src/lib.rs
djc: Changed it.

ctz (approved):

cpu (approved):
Thanks

cpu:
> I think this should be on the list for backporting to 0.22.1

I can take care of starting that up if nobody else is dying to do it :-)

cpu:
> I can take care of starting that up

https://github.com/rustls/rustls/pull/1677

PR #1676, opened 2023-12-07
Add a roadmap

ctz:
This is presently the https://www.memorysafety.org/initiative/rustls/rustls-work-plan/ -- more to come!

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -26,15 +26,15 @@ In rough order of priority:
 
 * **Support Encrypted Client Hello**.
   Encrypted Client Hello is an upcoming standard from the TLS WG providing better
-  production for some of the data sent by a client in the initial ClientHello
+  protection for some of the data sent by a client in the initial Client Hello
   message.
   rustls/rustls#508
 
 * **Improve OS Trust Verifier Support**.
   While we currently have a way to trust certificates stored in the platform trust
   store, platform trust stores can have other ways of restricting how/when roots
-  that they expose are trusted. In order to rely on these (on Darwin and Windows)
-  we should rely on the platform verifier directly.
+  that they expose are trusted. In order to rely on these (on Windows, Android,
+  and Apple platforms) we should rely on the platform verifier directly.
 
   Given that platform verifiers may require blocking I/O, some API changes are
   required.
@@ -69,7 +69,7 @@ Delivered in 0.21:
 * **Support IP Address Certificates**.
   There are some popular use cases where applications want TLS certificates for
   services that don’t have their own host name, relying on the IP address directly
-  instead. I've heard from one customer that this is common in Kubernetes deployments.
+  instead. This is common in Kubernetes deployments and service meshes.
   rustls/rustls#184
 
 * **Implement RFC 8446 Appendix C.4 in session cache**.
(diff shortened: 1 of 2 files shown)


cpu:
> Should we include https://github.com/rustls/rustls/issues/1541, given the FIPS connection?

Maybe we should split out a ticket specifically for OCSP? I think we can consider #1541 resolved w.r.t the FIPS requirements by way of CRL support for both client and server certs.

ctz:
> What is the goal of listing past priorities? If we keep them, I feel like they should be listed in reverse chronological order.

I think my goal is that someone external to the project can see this and have some confidence that we've delivered past roadmap items.

Have reordered these to be newest-first.

> Also, in the default GitHub Markdown rendering I feel like the headers in the current scheme don't stand out very much. I think formatting as a bullet list might work better?

Have reorganised it to be in a spaced bulleted list with bold headlines.

> (As a maintainer, I would also prefer it if the `Ticket` links were of the common #514 form.)

Using autolinking now.

ROADMAP.md
```diff
@@ -0,0 +1,85 @@
...
+* **Support no_std**.
+  Enables use of rustls in more memory-constrained environments.
+  RFC: rustls/rustls#1399
+
+* **OpenSSL API Compatibility Layer**.
+  Add an OpenSSL C API compatibility layer for adoption purposes.
+
+* **Support Encrypted Client Hello**.
+  Encrypted Client Hello is an upcoming standard from the TLS WG providing better
+  production for some of the data sent by a client in the initial ClientHello
```
cpu, line 29: one typo: "production" -> "protection"
one editorial change: we refer to "Client Hello" earlier, I think we should keep using that consistently (or put the `ClientHello` instance in backticks).
```suggestion
  protection for some of the data sent by a client in the initial Client Hello
```

ROADMAP.md
```diff
@@ -0,0 +1,85 @@
...
+  message.
+  rustls/rustls#508
+
+* **Improve OS Trust Verifier Support**.
+  While we currently have a way to trust certificates stored in the platform trust
+  store, platform trust stores can have other ways of restricting how/when roots
+  that they expose are trusted. In order to rely on these (on Darwin and Windows)
+  we should rely on the platform verifier directly.
```
cpu, line 37: Maybe mention Android/iOS in addition to Darwin + Windows? I think saying "MacOS" might also be more informative to a casual reader than Darwin.

ROADMAP.md
```diff
@@ -0,0 +1,85 @@
...
+  Allow plugging in different cryptographic back-ends.
+  rustls/rustls#1184
+
+Delivered in 0.21:
+
+* **Support IP Address Certificates**.
+  There are some popular use cases where applications want TLS certificates for
+  services that don’t have their own host name, relying on the IP address directly
+  instead. I've heard from one customer that this is common in Kubernetes deployments.
+  rustls/rustls#184
```
cpu, line 73: editorial: I think the first person narrative here is a bit odd compared to the rest of the document: 
```suggestion
  instead. This is common in Kubernetes deployments.
  rustls/rustls#184
```

ROADMAP.md
ctz: replaced with "on Windows, Android, and Apple platforms"

djc (approved):

PR #1673, opened 2023-12-06
Mark `SupportedProtocolVersion` as non-exhaustive

pvdrz:
This fixes a clippy warning that is showing up in other PRs:
```
 warning: this seems like a manual implementation of the non-exhaustive pattern
  --> rustls/src/versions.rs:11:1
   |
11 |   pub struct SupportedProtocolVersion {
   |   ^----------------------------------
   |   |
   |  _help: add the attribute: `#[non_exhaustive] pub struct SupportedProtocolVersion`
   | |
12 | |     /// The TLS enumeration naming this version.
13 | |     pub version: ProtocolVersion,
14 | |     is_private: (),
15 | | }
   | |_^
   |
help: remove this field
  --> rustls/src/versions.rs:14:5
   |
14 |     is_private: (),
   |     ^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive
   = note: `#[warn(clippy::manual_non_exhaustive)]` on by default
```

PR #1678, opened 2023-12-07
0.21.10 release preparation

cpu:
## Description

This branch targets the [rel-0.21](https://github.com/rustls/rustls/tree/rel-0.21) feature branch to prepare a point release bringing in https://github.com/rustls/rustls/pull/1668

## Proposed Release Notes

* TLS 1.2 servers now remove session tickets after observing a failure to decrypt, preventing future resumption with the same unusable ticket.


djc (approved):

cpu:
Nightly clippy was failing this branch with the finding addressed in https://github.com/rustls/rustls/pull/1673.

I tacked on a commit to add an ignore since I'm not sure I've convinced myself making the change to `non_exhaustive` in this branch won't have any unintended semver effects. In principle it seems like it could be safe since the "manual" implementation achieved the same effect? Given my lack of confidence here I went with the safest option.

ctz (approved):

djc (approved):

PR #1688, opened 2023-12-13
no-std support phase II

pvdrz:
This PR implements the second phase of no-std support described on [the RFC](https://github.com/rustls/rustls/pull/1399).

Part of the work here relies on using `hashbrown` as a no-std alternative to `std::collections::HashMap`. Which adds a new feature to `rustls` to avoid importing this dependency when std is available.

This continues from where https://github.com/rustls/rustls/pull/1502 left off.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -17,6 +17,7 @@ rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
 aws-lc-rs = { version = "1.6", optional = true, default-features = false, features = ["aws-lc-sys"] }
+hashbrown = { version = "0.13", optional = true } # 0.14+ requires 1.63 MSRV
 log = { version = "0.4.4", optional = true }
 # remove once our MSRV is >= 1.70
 once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -34,14 +34,14 @@ impl client::ClientSessionStore for NoClientSessionStorage {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
 mod cache {
     use alloc::collections::VecDeque;
     use core::fmt;
-    use std::sync::Mutex;
 
     use pki_types::ServerName;
 
+    use crate::lock::Mutex;
     use crate::msgs::persist;
     use crate::{limited_cache, NamedGroup};
 
@@ -80,11 +80,23 @@ mod cache {
     impl ClientSessionMemoryCache {
         /// Make a new ClientSessionMemoryCache.  `size` is the
         /// maximum number of stored sessions.
+        #[cfg(feature = "std")]
         pub fn new(size: usize) -> Self {
             let max_servers = size.saturating_add(MAX_TLS13_TICKETS_PER_SERVER - 1)
                 / MAX_TLS13_TICKETS_PER_SERVER;
             Self {
                 servers: Mutex::new(limited_cache::LimitedCache::new(max_servers)),
+            }
+        }
+
+        /// Make a new ClientSessionMemoryCache.  `size` is the
+        /// maximum number of stored sessions.
+        #[cfg(not(feature = "std"))]
+        pub fn new<M: crate::lock::MakeMutex>(size: usize) -> Self {
+            let max_servers = size.saturating_add(MAX_TLS13_TICKETS_PER_SERVER - 1)
+                / MAX_TLS13_TICKETS_PER_SERVER;
+            Self {
+                servers: Mutex::new::<M>(limited_cache::LimitedCache::new(max_servers)),
             }
         }
     }
@@ -180,7 +192,7 @@ mod cache {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
 pub use cache::ClientSessionMemoryCache;
 
 #[derive(Debug)]
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -389,7 +389,7 @@ mod conn;
 pub mod crypto;
 mod error;
 mod hash_hs;
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
 mod limited_cache;
 mod rand;
 mod record_layer;
@@ -582,7 +582,7 @@ pub mod server {
 
     pub use builder::WantsServerCert;
     pub use handy::NoServerSessionStorage;
-    #[cfg(feature = "std")]
+    #[cfg(any(feature = "std", feature = "hashbrown"))]
     pub use handy::ResolvesServerCertUsingSni;
     #[cfg(feature = "std")]
     pub use handy::ServerSessionMemoryCache;
@@ -638,3 +638,16 @@ pub mod ticketer;
 pub mod manual;
 
 pub mod time_provider;
+
+#[cfg(any(feature = "std", feature = "hashbrown"))]
+mod hash_map {
+    #[cfg(feature = "std")]
+    pub(crate) use std::collections::hash_map::Entry;
+    #[cfg(feature = "std")]
+    pub(crate) use std::collections::HashMap;
+
+    #[cfg(all(not(feature = "std"), feature = "hashbrown"))]
+    pub(crate) use hashbrown::hash_map::Entry;
+    #[cfg(all(not(feature = "std"), feature = "hashbrown"))]
+    pub(crate) use hashbrown::HashMap;
+}
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -521,7 +521,7 @@ pub use crate::stream::{Stream, StreamOwned};
 pub use crate::suites::{
     CipherSuiteCommon, ConnectionTrafficSecrets, ExtractedSecrets, SupportedCipherSuite,
 };
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
 pub use crate::ticketer::TicketSwitcher;
 #[cfg(feature = "tls12")]
 pub use crate::tls12::Tls12CipherSuite;
@@ -548,7 +548,7 @@ pub mod client {
     };
     #[cfg(feature = "std")]
     pub use client_conn::{ClientConnection, WriteEarlyData};
-    #[cfg(feature = "std")]
+    #[cfg(any(feature = "std", feature = "hashbrown"))]
     pub use handy::ClientSessionMemoryCache;
 
     /// Dangerous configuration that should be audited and used with extreme care.
@@ -584,7 +584,7 @@ pub mod server {
     pub use handy::NoServerSessionStorage;
     #[cfg(any(feature = "std", feature = "hashbrown"))]
     pub use handy::ResolvesServerCertUsingSni;
-    #[cfg(feature = "std")]
+    #[cfg(any(feature = "std", feature = "hashbrown"))]
     pub use handy::ServerSessionMemoryCache;
     pub use server_conn::{
         Accepted, ClientHello, ProducesTickets, ResolvesServerCert, ServerConfig,
@@ -630,7 +630,7 @@ pub mod sign {
 /// APIs for implementing QUIC TLS
 pub mod quic;
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
 /// APIs for implementing TLS tickets
 pub mod ticketer;
 
@@ -638,6 +638,9 @@ pub mod ticketer;
 pub mod manual;
 
 pub mod time_provider;
+
+/// APIs abstracting over locking primitives.
+pub mod lock;
 
 #[cfg(any(feature = "std", feature = "hashbrown"))]
 mod hash_map {
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -405,6 +405,7 @@ mod verifybench;
 mod x509;
 #[macro_use]
 mod check;
+#[cfg(feature = "logging")]
 mod bs_debug;
 mod builder;
 mod enums;
--- a/rustls/src/limited_cache.rs
+++ b/rustls/src/limited_cache.rs
@@ -1,8 +1,8 @@
 use alloc::collections::VecDeque;
 use core::borrow::Borrow;
 use core::hash::Hash;
-use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+
+use crate::hash_map::{Entry, HashMap};
 
 /// A HashMap-alike, which never gets larger than a specified
 /// capacity, and evicts the oldest insertion to maintain this.
@@ -19,7 +19,6 @@ pub(crate) struct LimitedCache<K: Clone + Hash + Eq, V> {
     oldest: VecDeque<K>,
 }
 
-#[cfg(feature = "std")]
 impl<K, V> LimitedCache<K, V>
 where
     K: Eq + Hash + Clone + core::fmt::Debug,
@@ -211,7 +210,6 @@ mod tests {
         }
     }
 
-    #[cfg(feature = "std")]
     #[test]
     fn test_get_or_insert_default_and_edit_evicts_old_items_to_meet_capacity() {
         let mut t = Test::new(3);
@@ -240,7 +238,6 @@ mod tests {
         assert_eq!(t.get("jkl"), None);
     }
 
-    #[cfg(feature = "std")]
     #[test]
     fn test_get_or_insert_default_and_edit_edits_existing_item() {
         let mut t = Test::new(3);
--- a/rustls/src/server/handy.rs
+++ b/rustls/src/server/handy.rs
@@ -25,13 +25,13 @@ impl server::StoresServerSessions for NoServerSessionStorage {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
 mod cache {
     use alloc::sync::Arc;
     use alloc::vec::Vec;
     use core::fmt::{Debug, Formatter};
-    use std::sync::Mutex;
 
+    use crate::lock::Mutex;
     use crate::{limited_cache, server};
 
     /// An implementer of `StoresServerSessions` that stores everything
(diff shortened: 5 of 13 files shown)


rustls/Cargo.toml
```diff
@@ -17,6 +17,7 @@ rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
 aws-lc-rs = { version = "1.5", optional = true }
+hashbrown = { version = "0.13", optional = true }
```
cpu, line 20: Did you consider using 0.14 here? Is there a reason to avoid that?

rustls/src/lib.rs
```diff
@@ -585,3 +585,17 @@ pub mod manual;
...
 pub mod time_provider;
+
+#[cfg(any(feature = "std", feature = "hashbrown"))]
+mod hash_map {
+    #[cfg(feature = "std")]
+    pub(crate) use std::collections::HashMap;
+    #[cfg(feature = "std")]
+    pub(crate) use std::collections::hash_map::Entry;
+
+
```
cpu: nit: extra superfluous whitespace here

rustls/src/ticketer.rs
```diff
@@ -62,7 +62,10 @@ impl TicketSwitcher {
     ///
     /// For efficiency, this is also responsible for locking the state mutex
     /// and returning the mutexguard.
-    pub(crate) fn maybe_roll(&self, now: UnixTime) -> Option<MutexGuard<TicketSwitcherState>> {
+    pub(crate) fn maybe_roll(
+        &self,
+        now: UnixTime,
+    ) -> Option<impl core::ops::Deref<Target = TicketSwitcherState> + '_> {
```
cpu: I think we'd want to `import core::ops::Deref` and use just `Deref` here.

rustls/src/server/handy.rs
```diff
@@ -46,51 +46,69 @@ impl server::StoresServerSessions for NoServerSessionStorage {
...
     /// Make a new ServerSessionMemoryCache.  `size` is the maximum
     /// number of stored sessions, and may be rounded-up for
     /// efficiency.
+    #[cfg(feature = "std")]
     pub fn new(size: usize) -> Arc<Self> {
         Arc::new(Self {
             cache: Mutex::new(limited_cache::LimitedCache::new(size)),
         })
     }
+    /// Make a new ServerSessionMemoryCache.  `size` is the maximum
```
cpu, line 65: nit: add separating whitespace

cpu (commented):
Nice :-) Phase II is a lot easier to digest than Phase I.

rustls/Cargo.toml
pvdrz: yes, 0.14 requires a higher msrv than what rustls currently has.

cpu: Ahh yup, that would be a good reason :-) Thanks

cpu:
Needs to be rebased on top of https://github.com/rustls/rustls/pull/1502 - I will look at that once 1502 has been reviewed.

I think the main divergences to address are:
* The time provider is a trait now, not a concrete struct
* Many of the fine-grain `#[cfg(feature = "std")]` conditions being replaced in this branch were replaced with fewer conditions for larger blocks of code, requiring adjustment here to update them.

cpu:
> cpu force-pushed the no-std-support-part-2 branch from c315801 to 1c892b1 

* Rebased on top of https://github.com/rustls/rustls/pull/1502
* Applied some minor review fixes (e.g. maintaining discrete commits vs changing the lock trait across multiple commits, import order, etc)
* Fixed up for the time provider changes in 1502
* Addressed conflicts from the no-std gating changes in 1502

cpu:
> cpu marked this pull request as ready for review now

Now that https://github.com/rustls/rustls/pull/1502 has landed this PR should be ready for review.

rustls/src/lib.rs
```diff
@@ -637,6 +637,41 @@ pub mod manual;
...
 pub mod time_provider;
 
+/// APIs for implementing locks on no-std environments.
+pub mod lock {
+    #[cfg(not(feature = "std"))]
+    use alloc::boxed::Box;
+
+    #[cfg(not(feature = "std"))]
+    /// A lock protecting shared data.
+    pub trait Lock: Send + Sync {
```
ctz, line 647: Like `TimeProvider` I feel like this is another candidate for living in another crate, like `abstractions-that-should-be-in-core`. In no sane world is a TLS library where abstractions over OS fundamentals should live, especially if they're gratuitously under-specified like these.

My other thought on this is -- can we instead use `spin` and avoid this entirely? That would be a similar solution to using hashbrown.

cpu: >  can we instead use spin and avoid this entirely? That would be a similar solution to using hashbrown.

That sounds appealing. I can look at that option.

ctz: > In no sane world is a TLS library where abstractions over OS fundamentals should live, especially if they're gratuitously under-specified like these.

Just found https://docs.rs/lock_api/0.4.11/lock_api/

cpu: :bulb:

cpu:

Putting this back to draft until I've had a chance to address the feedback RE: the lock trait.

cpu:
@ctz I've reworked this branch around the [`lock_api`](https://docs.rs/lock_api/0.4.11/lock_api/) dependency you found. I think it's workable, but the generics on the `Mutex` and `MutexGuard` types mean it's a more invasive diff to the parts of Rustls that embed a mutex into a struct since the struct now gains generic parameters, which in turn affects the impl blocks and so on. 

I thought we could perhaps use a type alias in the `crate::lock` module to fix some of the generic parameters but, [docs](https://doc.rust-lang.org/reference/items/type-aliases.html) and clippy told me:  "A type alias, when not used as an associated type, must include a Type and may not include TypeParamBounds." so that's a no go :-(

In the end I split most of the Mutex using implementations into two private modules (std vs no-std) and then conditionally re-export the correct one under the existing name based on feature flags. In most cases this felt OK because the impls being duplicated across the private modules were quite thin. For the `TicketSwitcher` the duplication of the `maybe_roll` fn feels less pleasant. I'm going to see if I can pull that out into something shared by both std and no-std constructions.

I'll leave this as a draft while I figure out that last bit, WDYT about the general shape?

rustls/src/client/handy.rs
```diff
@@ -69,119 +71,239 @@ mod cache {
         }
     }
 
-    /// An implementer of `ClientSessionStore` that stores everything
```
cpu, line 72: yuck. This diff needs splitting so the lift into a module is separate from the creation of the no-std version. I'll adjust.

cpu: Split up.

ctz:
> @ctz I've reworked this branch around the [`lock_api`](https://docs.rs/lock_api/0.4.11/lock_api/) dependency you found. I think it's workable, but the generics on the [`Mutex`](https://docs.rs/lock_api/0.4.11/lock_api/struct.Mutex.html) and [`MutexGuard`](https://docs.rs/lock_api/0.4.11/lock_api/struct.MutexGuard.html) types mean it's a more invasive diff to the parts of Rustls that embed a mutex into a struct since the struct now gains generic parameters, which in turn affects the impl blocks and so on.

Sorry, this might've been a wild goose chase. I had assumed that "it also allows users to write code which is generic with regards to different lock implementations" in the `lock_api` docs meant you could write code which is generic with regards to different lock implementations, but actually it doesn't include `std::sync::Mutex` in that.

I was assuming we could make the whole thing generic over `lock_api::RawMutex`, use `lock_api::Mutex` ourselves, and then type alias that away for the `std` feature.

This feels pretty dismal. At the risk of repeating the moan above, I think one of the things that `core` should contain (but doesn't) are traits for important types in `std`. That would make it easier for crates to be `no_std` without everyone inventing their own abstractions over them. Maybe I'll file an upstream issue about that.

cpu:
> Sorry, this might've been a wild goose chase.

No worries, it was an interesting exercise if nothing else.

> I was assuming we could make the whole thing generic over lock_api::RawMutex, use lock_api::Mutex ourselves, and then type alias that away for the std feature.

Yeah :-( I think that would be the ideal state but mechanically I'm not sure it's doable. Maybe if we created a type that wrapped `std::sync::Mutex` and implemented the `lock_api` traits and used that throughout for the `std` feature? 

> This feels pretty dismal. At the risk of repeating the moan above, I think one of the things that core should contain (but doesn't) are traits for important types in std. That would make it easier for crates to be no_std without everyone inventing their own abstractions over them. Maybe I'll file an upstream issue about that.

That does feel like the best direction long-term. Do you think this work should be shelved in the meantime?

cpu:
> That does feel like the best direction long-term. Do you think this work should be shelved in the meantime?

based on conversations out-of-band we agreed the original approach used by pvdrz et. al. (introducing a `Lock` trait of our own) was the least worst option. I've reworked the branch to go back to that idea and removed the `lock_api` feature/associated work.

I also noticed/fixed a warning w.r.t dead code and the `bs_debug` module that's fixed with an extra commit at the end."

With the above I believe this is ready for review again.

rustls/src/lib.rs
```diff
@@ -389,7 +389,7 @@ mod conn;
 pub mod crypto;
 mod error;
 mod hash_hs;
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "hashbrown"))]
```
djc, line 392: I think this change logically belongs with the previous one?

rustls/src/lib.rs
```diff
@@ -581,7 +581,7 @@ pub mod server {
 
     pub use builder::WantsServerCert;
     pub use handy::NoServerSessionStorage;
-    #[cfg(feature = "std")]
+    #[cfg(any(feature = "std", feature = "hashbrown"))]
```
djc, line 585: And this one too, IMO.

rustls/src/lib.rs
```diff
@@ -638,6 +638,41 @@ pub mod manual;
...
+        /// The state protected by the lock.
+        type Data: Send;
+
+        /// Acquire the lock.
+        fn lock(&self) -> MutexGuard<'_, Self::Data>;
+    }
+
+    #[cfg(not(feature = "std"))]
+    /// A lock builder.
+    pub trait MakeMutex {
```
djc, line 659: So the setup here seems to try to mimic the std API as much as possible to minimize the amount of code to change in the library. I think on balance I'd prefer to go the other way to make it more obvious that this isn't exactly the default stuff, and use `Arc<dyn Lock<T>>` explicitly at call sites.

It would be nice to fold the `MakeMutex` trait into a `Lock::new()` method if possible.

(I guess this assumes that the intermediate vtable does not cause substantial performance degradation for the case where `std` is available. I think that holds because we don't use `Mutex`es all that much and not never inside particularly hot code?)

rustls/src/server/handy.rs
```diff
@@ -46,32 +45,51 @@ mod cache {
...
         }
 
         fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
-            self.cache
-                .lock()
-                .unwrap()
-                .get(key)
-                .cloned()
+            #[cfg(feature = "std")]
+            let cache = self.cache.lock().unwrap();
```
djc, line 79: If possible, it would be nicer if the `Lock` trait impls can abstract over poisoning.

rustls/src/server/handy.rs
```diff
@@ -46,32 +45,51 @@ mod cache {
         /// Make a new ServerSessionMemoryCache.  `size` is the maximum
         /// number of stored sessions, and may be rounded-up for
         /// efficiency.
+        #[cfg(feature = "std")]
```
djc, line 48: I would probably squash these changes into the previous commit.

rustls/src/client/handy.rs
```diff
@@ -80,27 +80,44 @@ mod cache {
     impl ClientSessionMemoryCache {
         /// Make a new ClientSessionMemoryCache.  `size` is the
         /// maximum number of stored sessions.
+        #[cfg(feature = "std")]
```
djc, line 83: Would squash this, too.

djc (commented):
This mostly looks good. It feels to me like the commits are too small -- in my mind there is a `HashMap` commit and a `Mutex` commit, and splitting it up further doesn't help much.

cpu:
> This mostly looks good. It feels to me like the commits are too small -- in my mind there is a HashMap commit and a Mutex commit, and splitting it up further doesn't help much.

It's not often I get this feedback from you! :zany_face: I thought doing it piece-by-piece would make review easier (I even split it up a bit more than the original version), but I don't feel very strongly. I can collapse everything down to two commits.

rustls/src/lib.rs
cpu: > So the setup here seems to try to mimic the std API as much as possible to minimize the amount of code to change in the library. I think on balance I'd prefer to go the other way

@ctz I think both approaches here are reasonable, but I'm swayed by djc's position that it should be obvious when we're deviating from the standard path. Do you agree or would you prefer we keep the original approach that minimizes the lib changes?

ctz: I don't mind the call sites knowing about the difference. I think I would be concerned if `std` users are paying extra runtime cost (specifically in the lock/unlock path; not concerned about mutex construction or memory footprint) for the benefit of this extra abstraction?

rustls/src/server/handy.rs
cpu: > it would be nicer if the Lock trait impls can abstract over poisoning.

@djc I agree it would be nicer, but so far the ideas I've had to do it don't seem like a net positive. Did you have a specific approach in mind? 

My thoughts so far:

The `Lock` trait isn't in-play for the std-enabled case and I don't think we want to change that to avoid overhead in the lock/unlock path on std builds. When std is enabled the locks are `std::sync::Mutex` re-exports, not an arc over a `dyn Lock`. I think that rules out making the std case look like the no-std case for calling `lock()`.

That leaves making the no-std `lock()` look more like the std case by introducing the handling of a poisoned result. We could make the `Lock` trait's `lock()` fn return `Result<MutexGuard<'_, Self::Data>, Self::PoisonError>;` instead of `MutexGuard<'_, Self::Data>` but adding a second generic parameter to the `Lock` trait for the error will make the no-std usages even more divergent from the std-usages. Fixing the type to a concrete `Error` and not making it generic might be one option.

Am I missing/overthinking something here?

rustls/src/lib.rs
cpu: > (I guess this assumes that the intermediate vtable does not cause substantial performance degradation for the case where std is available. I think that holds because we don't use Mutexes all that much and not never inside particularly hot code?)

Just want to emphasize something I said in the other comment thread; there is no vtable in play for the case where std is available. The `crate::lock::Mutex` type in that case is just a re-export of `std::sync::Mutex`. We only use the `Lock` trait for no-std.

cpu: > I think on balance I'd prefer to go the other way to make it more obvious that this isn't exactly the default stuff, and use Arc<dyn Lock<T>> explicitly at call sites.

Do you mean dropping the `pub(crate) type Mutex<T> = alloc::sync::Arc<dyn Lock<Data = T>>;` reexport from `crate::lock`` and using `Arc<dyn Lock<Data = T>>` in all of the places that used `crate::lock::Mutex` , or something else?

> It would be nice to fold the MakeMutex trait into a Lock::new() method if possible.

I think that might be exclusive with keeping the `Lock` trait compatible a trait object no?

rustls/src/client/handy.rs
cpu: Done.

rustls/src/server/handy.rs
cpu: Done

rustls/src/lib.rs
cpu: Done.

rustls/src/lib.rs
cpu: Done.

cpu:
>  in my mind there is a HashMap commit and a Mutex commit, and splitting it up further doesn't help much.

Force pushed an update with just three commits: the clippy fix, the hashbrown feature and all things that can be enabled with just it, and the lock module and all things that can be enabled with it + hashbrown.

djc:
> Force pushed an update with just three commits: the clippy fix, the hashbrown feature and all things that can be enabled with just it, and the lock module and all things that can be enabled with it + hashbrown.

Yeah, might be a bit out of character, but I found this state a lot easier to review compared to the previous state!

rustls/src/limited_cache.rs
```diff
@@ -211,7 +210,7 @@ mod tests {
         }
     }
 
-    #[cfg(feature = "std")]
+    #[cfg(any(feature = "std", feature = "hashbrown"))]
```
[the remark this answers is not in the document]
cpu: Fixed.

rustls/src/server/handy.rs
cpu: > How about we just yield Option? Seems to be what we already use in the ticketer in practice?

Went with a wrapper type that has a lock fn yielding an `Option`.

rustls/src/lib.rs
cpu: > One thing I missed before is that the Lock here is taking an associated Data type, but I think that should be a type argument on the trait instead of an associated type 

Fixed.

> The other way to go here would be to provide our own Mutex type that wraps around an inner: std::sync::Mutex<T> in the std case or an inner: Arc<dyn Lock> in the no-stdcase.

I implemented this approach. I also lifted the lock module out of `lib.rs` so I could more easily create two side-by-side impls with minimal gating.

> if we use a local Mutex type to abstract over the options I think that type could also have a new() method and this issue would fall away.

The `MakeMutex` trait remains in my latest push. I personally don't see how it can be avoided but would be open to new ideas. 

The type being guarded by the mutex is crate internal in some instances (e.g. `limited_cache::LimitedCache`), so the caller can't construct a `Arc<dyn Lock<T>>` to give to us in `new()` because `T` is private. On the other hand, we can't construct the `Arc<dyn Lock<T>>` ourselves in `new()` without something like `MakeMutex` because we don't know the concrete choice of `dyn Lock` the caller wants to use.

We can probably fiddle with the shape of `MakeMutex` (maybe it could be a `Fn` provided by arg instead of a trait?) but in general I'm getting a little bit weary of refactoring this branch around through many different permutations. Is there a specific downside, or issue with, the `MakeMutex` approach you are trying to avoid?

cpu:
I think this branch is ready for another review pass now.

rustls/src/client/handy.rs
```diff
@@ -80,27 +80,36 @@ mod cache {
...
+        }
     }
 
     impl super::client::ClientSessionStore for ClientSessionMemoryCache {
         fn set_kx_hint(&self, server_name: ServerName<'static>, group: NamedGroup) {
-            self.servers
-                .lock()
-                .unwrap()
-                .get_or_insert_default_and_edit(server_name, |data| data.kx_hint = Some(group));
+            let mut servers = self.servers.lock().unwrap();
```
djc, line 106: I think these changes can be reverted? (Or split off into a separate commit -- it's just a style thing?)

djc (approved):

rustls/src/lib.rs
djc: Yup, agree that `MakeMutex` can't be avoided -- sorry for the noise.

rustls/src/client/handy.rs
cpu: Agreed. I will tidy up

cpu: Cleaned up!

cpu:
Since the impl has changed quite a bit since Ctz stamped I'm going to wait to merge this until he has a chance to do another pass (_avoiding a tag since it's the holiday weekend - I'll follow up ~Tues_).

ctz (approved):

PR #1693, opened 2023-12-15
examples: consistently document example program purpose

cpu:
Adding an introductory Rustdoc comment to new examples was review feedback delivered in https://github.com/rustls/rustls/pull/1583, but we're not doing this consistently for all existing examples (💔). 

This branch addresses this oversight that by adding short intro documentation to `simple_0rtt_client.rs`, and the two MIO examples (`tlsclient-mio.rs` and `tlsserver-mio.rs`).


ctz (approved):

djc (approved):
Tiny nit: I would probably prefer having an empty line between the module-level docstring and the first line of code.

cpu:
> I would probably prefer having an empty line between the module-level docstring and the first line of code.

Fixed and rebased on `main`.

PR #1697, opened 2023-12-18
bench: remove cachegrind diff post-processing

aochagavia:
This code was meant to strip unnecessary information from the start of a cachegrind diff. However, for some versions of cachegrind it results in a completely blank string. Instead of making it work for all cachegrind versions, it is probably better to get rid of it altogether (it does not make enough of a difference that the complexity of a proper solution would be worth it).


ctz (approved):

cpu (approved):
Thanks!

PR #1698, opened 2023-12-19
verify.rs: correct comment for `verify_server_name`

ctz:
This implies webpki checks the DN commonName value for DNS names. It does not.


djc (approved):

cpu:
Thanks!

PR #1702, opened 2023-12-20
Add support for FFDHE key exchange

### LINKED ISSUE #1576: DHE KeyExchangeAlgorithm Support
## Background

In my use case, I want to use `rustls` with my custom crypto provider to support DHE based TLS 1.2 cipher suites.

## Proposal

I noticed in `NamedGroup`, there are `FFDHE` groups.
DHE also has perfect forward secrecy but just slower.

So, I wondering if it possible to support DHE? I could help with adding this feature.

s-arash:
This PR adds support for FFDHE key exchange ([RFC 7919](https://www.rfc-editor.org/rfc/rfc7919.html)) to rustls, enabling custom `CryptoProvider`s to offer FFDHE key exchange. 

Note: as discused in #1576, this PR deliberately does not include FFDHE support in the default `CryptoProvider`s.

Closes #1576.


cpu:
Thanks for the PR :-) Just leaving a note that it might take a little bit to get reviews. We have some other work in the backlog and the number of reviewers is smaller than usual right now.

s-arash:
> Thanks for the PR :-) Just leaving a note that it might take a little bit to get reviews. We have some other work in the backlog and the number of reviewers is smaller than usual right now.

No worries. I wasn't expecting any reviews over the holidays 😅

ci-tests/src/validate_ffdhe_params.rs
```diff
@@ -0,0 +1,128 @@
...
+                "OpenSSL 3+ is required for the tests here. The installed version is {version:?}"
+            );
+        }
+        Err(e) => {
+            panic!("OpenSSL 3+ needs to be installed and in PATH.\nThe error encountered: {e}")
+        }
+    }
+}
+
+/// Get FFDHE parameters `(p, g)`` for the given `ffdhe_group` from OpenSSL
```
Taowyoo, line 46: Nit pick
```suggestion
/// Get FFDHE parameters `(p, g)` for the given `ffdhe_group` from OpenSSL
```

rustls/tests/api.rs
```diff
@@ -5034,6 +5097,259 @@ fn test_server_rejects_clients_without_any_kx_group_overlap() {
...
+
+    let (client, server) = make_pair_for_configs(client_config, server_config);
+    let (mut client, mut server) = (client.into(), server.into());
+    transfer_altered(&mut client, clear_named_groups_ext, &mut server);
+    assert!(server.process_new_packets().is_ok());
+}
+
+#[cfg(feature = "tls12")]
+#[test]
+fn test_server_picks_dhe_group_when_clienthello_has_no_groups_ext() {
```
Taowyoo, line 5138: Just want to check my understanding:
Does this case only happen when using DHE cipher suite ?

rustls/tests/api.rs
```diff
@@ -5034,6 +5097,259 @@ fn test_server_rejects_clients_without_any_kx_group_overlap() {
...
+            }
+            *encoded = Payload::new(parsed.get_encoding());
+        }
+        Altered::InPlace
+    }
+
+    let client_config = finish_client_config(
+        KeyType::Rsa,
+        rustls::ClientConfig::builder_with_provider(ffdhe::ffdhe_provider().into())
+            .with_protocol_versions(&[&rustls::version::TLS12])
```
Taowyoo, line 5153: Nitpick: it's better to use `.with_safe_default_protocol_versions()` to keep sync with other tests.

Please be free to resolve this comment, if this test case needs TLS 1.2.

rustls/tests/api.rs
s-arash: Yes, AFAIK, ECDHE requires the groups extension.

rustls/tests/api.rs
s-arash: I think the client needs to be TLS 1.2 only. I can configure server with `.with_safe_default_protocol_versions()` instead.

rustls/src/suites.rs
```diff
@@ -181,6 +195,72 @@ pub(crate) fn reduce_given_version_and_protocol(
...
+            ecdhe_kx_ok = true;
+        }
+        if ecdhe_kx_ok & ext_has_known_ffdhe_groups {
+            break;
+        }
+    }
+
+    #[cfg(feature = "tls12")]
+    let ffdhe_kx_ok_tls12 = ext_has_known_ffdhe_groups ||
+        // https://datatracker.ietf.org/doc/html/rfc7919#section-4 (paragraph 1)
```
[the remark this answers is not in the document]
s-arash: Good catch! Will fix.

rustls/tests/api.rs
Taowyoo: Got it

rustls/src/server/tls12.rs
```diff
@@ -83,27 +83,33 @@ mod client_hello {
...
+            let ecpoints_ext = client_hello.ecpoints_extension();
+
+            // TODO: according to the ECC RFC, sending this extension is not required
+            // https://datatracker.ietf.org/doc/html/rfc8422#section-5.1.2
+            if ecpoints_ext.is_none() && self.suite.kx == KeyExchangeAlgorithm::ECDHE {
+                return Err(cx.common.send_fatal_alert(
+                    AlertDescription::HandshakeFailure,
+                    PeerIncompatible::EcPointsExtensionRequired,
+                ));
+            }
```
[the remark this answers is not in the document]
s-arash: Done!

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +187,69 @@ pub struct CryptoProvider {
...
+        for cs in self.cipher_suites.iter() {
+            let cs_kx = cs.key_exchange_algorithms();
+            if !cs_kx
+                .iter()
+                .any(|kx| kx_algos.contains(kx))
+            {
+                let suite_name = cs.common().suite;
+                return Err(Error::General(alloc::format!(
+                    "Ciphersuite {suite_name:?} requires {cs_kx:?} key exchange, but no {cs_kx:?}-compatible \
+                    key exchange groups were present in `CryptoProvider`'s `kx_gropus` field",
```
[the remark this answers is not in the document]
s-arash: Fixed!

rustls/src/msgs/ffdhe_groups.rs
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+}
+
+/// FFDHE2048 group defined in [RFC 7919 Appendix A.1]
+///
+/// [RFC 7919 Appendix A.1]: https://datatracker.ietf.org/doc/html/rfc7919#appendix-A.1
+pub const FFDHE2048: FfdheGroup = FfdheGroup {
+    p: &[
+        0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xAD, 0xF8, 0x54, 0x58, 0xA2, 0xBB, 0x4A,
```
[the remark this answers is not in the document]
s-arash: Done!

rustls/src/server/tls12.rs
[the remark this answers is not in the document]
s-arash: And thanks for taking the time to review the PR!

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +187,69 @@ pub struct CryptoProvider {
...
+                    "Ciphersuite {suite_name:?} requires {cs_kx:?} key exchange, but no {cs_kx:?}-compatible \
+                    key exchange groups were present in `CryptoProvider`'s `kx_groups` field",
+                )));
+            }
+        }
+        Ok(())
+    }
+
+    /// Removes cipher suites from this `CryptoProvider` that have no matching `SupportedKxGroup`s in `kx_groups`
+    pub fn remove_cipher_suites_without_matching_kx(&mut self) {
```
djc, line 236: I would conceptually expect `CryptoProvider` values to be immutable. Why do we need this method (and its caller, below)?

rustls/src/lib.rs
```diff
@@ -547,6 +547,9 @@ pub use crate::verify::DigitallySignedStruct;
 pub use crate::versions::{SupportedProtocolVersion, ALL_VERSIONS, DEFAULT_VERSIONS};
 pub use crate::webpki::RootCertStore;
 
+#[cfg(feature = "tls12")]
```
djc: Nit: please avoid the empty line above.

rustls/src/msgs/enums.rs
```diff
@@ -1,7 +1,10 @@
 #![allow(clippy::upper_case_acronyms)]
 #![allow(non_camel_case_types)]
 /// This file is autogenerated.  See https://github.com/ctz/tls-hacking/
-use crate::msgs::codec::{Codec, Reader};
+use crate::{
```
djc, line 4: Style nit: we use a one line per module import style, so this should be two separate `use crate::..` lines.

rustls/src/msgs/ffdhe_groups.rs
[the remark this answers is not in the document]
djc: Please squash those changes into this commit.

rustls/src/msgs/handshake.rs
```diff
@@ -1492,8 +1494,15 @@ impl CertificatePayloadTls13 {
 pub enum KeyExchangeAlgorithm {
     /// Key exchange performed via elliptic curve Diffie-Hellman.
     ECDHE,
+    /// Diffie-Hellman Key exchange (with only known parameters as defined in [RFC 7919]).
+    ///
+    /// [RFC 7919]: https://datatracker.ietf.org/doc/html/rfc7919
+    DHE,
```
djc, line 1500: Nit: please move this above `ECDHE` to keep the variants in alphabetic order.

rustls/src/msgs/handshake.rs
```diff
@@ -1540,6 +1582,22 @@ impl Codec<'_> for ClientEcdhParams {
...
+}
+
+impl Codec<'_> for ClientDhParams {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.public.encode(bytes);
+    }
+
+    fn read(r: &mut Reader) -> Result<Self, InvalidMessage> {
+        let pb = PayloadU16::read(r)?;
+        Ok(Self { public: pb })
```
djc, line 1602: Style nit: please avoid the unnecessary `pb` binding by just inlining it.

rustls/src/msgs/handshake.rs
```diff
@@ -1524,6 +1533,39 @@ impl Codec<'_> for EcParameters {
...
+            Self::Ecdh(ecdh) => ecdh.encode(buf),
+            Self::Dh(dh) => dh.encode(buf),
+        }
+    }
+}
+
+impl KxDecode<'_> for ClientKeyExchangeParams {
+    fn decode(r: &mut Reader, algo: KeyExchangeAlgorithm) -> Result<Self, InvalidMessage> {
+        Ok(match algo {
+            KeyExchangeAlgorithm::ECDHE => Self::Ecdh(Codec::read(r)?),
```
djc, line 1568: Style nit: instead of calling `Codec::read()`, I think this would be easier to follow if you wrote them as `ClientEcdhParams::read()`/`ClientDhParams::read()`.

(Maybe consider writing `use KeyExchangeAlgorithm::*;` before the `match` to cut down on noise.)

rustls/src/msgs/handshake.rs
```diff
@@ -1577,7 +1635,120 @@ impl Codec<'_> for ServerEcdhParams {
...
+        match self {
+            Self::Ecdh(ecdh) => Some(ecdh.curve_params.named_group),
+            Self::Dh(dh) => dh.named_group(),
+        }
+    }
+}
+
+impl KxDecode<'_> for ServerKeyExchangeParams {
+    fn decode(r: &mut Reader, algo: KeyExchangeAlgorithm) -> Result<Self, InvalidMessage> {
+        match algo {
```
djc, line 1729: See other `KxDecode` impl.

rustls/src/msgs/handshake.rs
```diff
@@ -1596,16 +1767,55 @@ impl Codec<'_> for EcdheServerKeyExchange {
...
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.params.encode(bytes);
+        self.dss.encode(bytes);
+    }
+
+    fn read(r: &mut Reader) -> Result<Self, InvalidMessage> {
+        let params = ServerDhParams::read(r)?;
+        let dss = DigitallySignedStruct::read(r)?;
+
+        Ok(Self { params, dss })
```
djc: Nit: would be good to just inline these bindings.

rustls/src/msgs/handshake.rs
```diff
@@ -1619,19 +1829,25 @@ impl Codec<'_> for ServerKeyExchangePayload {
...
+                }
+            };
+            let result = ServerKeyExchange {
+                params: server_kx_params,
+                dss: DigitallySignedStruct::read(&mut rd).ok()?,
             };
 
             if !rd.any_left() {
-                return result.ok();
+                return Some(result);
```
djc, line 1788: Style nit: would suggest inlining the binding here.

rustls/src/server/tls12.rs
```diff
@@ -180,12 +180,35 @@ mod client_hello {
...
+                            .kx_groups
+                            .iter()
+                            .find(|skxg| {
+                                skxg.name().key_exchange_algorithm()
+                                    == Some(KeyExchangeAlgorithm::DHE)
+                            })
+                    } else {
+                        None
+                    }
+                })
```
djc, line 211: I think this combinator chain gets pretty unwieldy, I think it might be time to rewrite it into a `for`-loop?

rustls/src/suites.rs
```diff
@@ -181,6 +195,48 @@ pub(crate) fn reduce_given_version_and_protocol(
...
+        }
+    }
+    // https://datatracker.ietf.org/doc/html/rfc7919#section-4 (paragraph 1)
+    let no_known_ffdhe_groups = if let Some(groups_ext) = groups_ext {
+        let mut ext_has_ffdhe_groups = false;
+        let mut ext_has_known_ffdhe_groups = false;
+        for g in groups_ext
+            .iter()
+            .filter(|g| is_ffdhe(g))
+        {
```
djc: Would prefer to avoid the `filter()` combinator in the loop header in favor of inlining the `match` from `is_ffdhe()` in the loop body with an empty return.

rustls/src/suites.rs
```diff
@@ -181,6 +195,48 @@ pub(crate) fn reduce_given_version_and_protocol(
...
+    fn is_ffdhe(g: &NamedGroup) -> bool {
+        match g {
+            // Codepoints between 0x100 and 0x200 are reserved for ffdhe groups.
+            // https://datatracker.ietf.org/doc/html/rfc7919#section-2
+            NamedGroup::Unknown(x) => (0x100..0x200).contains(x),
+            _ => g.key_exchange_algorithm() == Some(KeyExchangeAlgorithm::DHE),
+        }
+    }
+    // https://datatracker.ietf.org/doc/html/rfc7919#section-4 (paragraph 1)
+    let no_known_ffdhe_groups = if let Some(groups_ext) = groups_ext {
```
djc: Style nit: I think a `match` would make this code easier to follow compared to the `if let` here.

rustls/src/tls12/mod.rs
```diff
@@ -313,14 +313,15 @@ fn join_randoms(first: &[u8; 32], second: &[u8; 32]) -> [u8; 64] {
 
 type MessageCipherPair = (Box<dyn MessageDecrypter>, Box<dyn MessageEncrypter>);
 
-pub(crate) fn decode_ecdh_params<'a, T: Codec<'a>>(
+pub(crate) fn decode_kx_params<'a, T: KxDecode<'a>>(
     common: &mut CommonState,
     kx_params: &'a [u8],
+    kx_algorithm: KeyExchangeAlgorithm,
```
djc, line 319: Style nit: please move this argument first.

rustls/tests/api.rs
```diff
@@ -272,6 +272,36 @@ fn config_builder_for_client_rejects_incompatible_cipher_suites() {
     );
 }
 
+#[cfg(feature = "tls12")]
```
djc, line 275: I think it would probably make sense to keep these tests in a separate file in `tests`, which can have an inner attribute for the `tls12` guard. That file can then inline the contents of `tests/common/ffdhe.rs` (at the bottom of the file).

rustls/tests/common/ffdhe.rs
```diff
@@ -0,0 +1,100 @@
...
+    crypto::{
+        self, ActiveKeyExchange, CipherSuiteCommon, CryptoProvider, KeyExchangeAlgorithm,
+        SupportedKxGroup,
+    },
+    CipherSuite, NamedGroup, SupportedCipherSuite, Tls12CipherSuite,
+};
+
+use super::*;
+
+static TLS12_DHE_RSA_WITH_AES_128_GCM_SHA256: Tls12CipherSuite =
```
djc, line 13: This file should be ordered top-down (see `CONTRIBUTING`).

rustls/tests/common/ffdhe.rs
```diff
@@ -0,0 +1,100 @@
+use rustls::{
```
djc, line 2: Style nit: one import per module, please.

ci-tests/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
djc, line 1: What's the rationale for adding a new crate for this? Why should we run these tests in CI only?

If we're going to put these tests into a separate crate, I think we should probably just rename `connect-tests` to `ci-tests`.

rustls/src/suites.rs
```diff
@@ -210,28 +210,52 @@ pub(crate) fn reduce_given_kx_groups(
...
-    let no_known_ffdhe_groups = if let Some(groups_ext) = groups_ext {
-        let mut ext_has_ffdhe_groups = false;
-        let mut ext_has_known_ffdhe_groups = false;
-        for g in groups_ext
-            .iter()
-            .filter(|g| is_ffdhe(g))
-        {
-            ext_has_ffdhe_groups = true;
+
+    let mut ecdhe_kx_ok = false;
```
djc, line 214: Please squash this commit into the earlier one that introduced these code, in order to make it easier to review.

ci-tests/src/validate_ffdhe_params.rs
```diff
@@ -43,7 +43,7 @@ fn verify_openssl3_available() {
     }
 }
 
-/// Get FFDHE parameters `(p, g)`` for the given `ffdhe_group` from OpenSSL
+/// Get FFDHE parameters `(p, g)` for the given `ffdhe_group` from OpenSSL
```
djc, line 6: Review feedback should be squashed into the originating commit(s).

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -10,6 +10,7 @@ use crate::msgs::fragmenter::MAX_FRAGMENT_LEN;
 use crate::msgs::message::{BorrowedPlainMessage, OpaqueMessage};
 use crate::suites::{CipherSuiteCommon, ConnectionTrafficSecrets, SupportedCipherSuite};
 use crate::tls12::Tls12CipherSuite;
+use crate::version::TLS12;
```
djc, line 13: Please squash this one.

djc (commented):
This is looking pretty good! I have a lot of small commits to get this more closely aligned to rustls code style, and suggest to clean up the commit history to make this easier to review.

rustls/src/msgs/handshake.rs
s-arash: It will be a net loss in line count (i.e., more lines)😅

rustls/src/suites.rs
s-arash: Done!

rustls/src/crypto/mod.rs
s-arash: With two different kx algorithms, we may get into a situation where the `CryptoProvider` offers a cipher suite but does not have a kx group corresponding to the cipher suite. 

For example, the provider could offer `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256` but have no FFDHE groups in its `kx_groups` field. If such a provider is configured for the client, and the server picks `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256`, the handshake cannot succeed. The purpose of this method is to prevent such cases (and be convenient for users). 

There is a related function, `verify_cipher_suites_have_matching_kx` that enforces this invariant when configuring a provider.

rustls/src/msgs/handshake.rs
djc: Right, but I think it will be easier to read/comprehend.

ci-tests/Cargo.toml
s-arash: The tests are not strictly ci-only 😅, we can run them on dev machines as well. The point was to not break rustls tests on dev machines in case OpenSSL 3 is not installed.

As for combining them with `connect-tests`, I thought since those tests require network connectivity and the tests in `ci-tests` don't, it may be good to keep them separate. I can combine them if that is preferred.

rustls/src/crypto/aws_lc_rs/tls12.rs
s-arash: Done!

rustls/src/msgs/handshake.rs
s-arash: Done!

rustls/src/msgs/handshake.rs
s-arash: Not sure how to do that here.

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +187,69 @@ pub struct CryptoProvider {
...
+        let mut res = Vec::with_capacity(ALL_KEY_EXCHANGE_ALGORITHMS.len());
+        for kx in self
+            .kx_groups
+            .iter()
+            .filter_map(|kx| kx.name().key_exchange_algorithm())
+        {
+            if !res.contains(&kx) {
+                res.push(kx);
+            }
+            if res.len() == ALL_KEY_EXCHANGE_ALGORITHMS.len() {
```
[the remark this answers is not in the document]
s-arash: I added a comment.

rustls/src/client/tls12.rs
```diff
@@ -436,7 +437,7 @@ impl State<ClientConnectionData> for ExpectServerKx<'_> {
         )?;
         self.transcript.add_message(&m);
 
-        let ecdhe = opaque_kx
+        let kx = opaque_kx
```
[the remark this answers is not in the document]
s-arash: Would you like me to make a separate commit for this change?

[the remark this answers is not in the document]
djc: Yes, it would be nice to keep renaming and other changes that are not strictly tied to the FFDHE changes in a separate commit that comes before the main FFDHE commit.

rustls/src/crypto/mod.rs
djc: I think we should just rely on verifying the invariant when configuring a provider and not have any methods that mutate an existing `CryptoProvider`. We should have documentation on the `CryptoProvider` fields that mention any invariants that must hold between separate fields.

rustls/src/msgs/handshake.rs
djc: Ahh, you're right. Never mind!

ci-tests/Cargo.toml
djc: I think I would prefer still collecting those in a single crate. @ctz / @cpu thoughts?

rustls/src/client/tls12.rs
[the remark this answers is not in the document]
s-arash: Will do.

ci-tests/Cargo.toml
cpu: I think I would also prefer a single crate.

s-arash: I will combine them in a single crate, `ci-tests`.

s-arash: So I tired doing this. A couple of issues:
- `connect-tests` run daily, which may or may not be an issue.
- `connect-tests` run on multiple platforms, including Windows and Mac, and my `ci-tests` required OpenSSL. By running on Ubuntu only, I can use OpenSSL that's bundled with Ubuntu. If we were to run these tests on other platforms as well, we'd need to install OpenSSL on those platforms before running the tests.

rustls/src/server/tls12.rs
s-arash: I'm not sure rewriting this into a for loop would improve it that much. Most of the complexity comes from `.or_else(..)` invocation, which I don't think we can get around easily.

s-arash:
Thanks @djc for the thorough review! I addressed most of your comments. There are a couple of comments remaining that I need further feedback for.

ci-tests/Cargo.toml
djc: Okay, that's good feedback. In light of this, maybe we should keep them as two crates after all. I'd suggest you rename the new crate to openssl-tests to be more explicit about what it provides.

rustls/src/server/tls12.rs
djc: Hmm, maybe this should be a method on `ServerConfig` that takes `&ClientHelloPayload` and `KeyExchangeAlgorithm` and yields `Result<&dyn SupportedKxGroup, Error>`? What do you think?

ci-tests/Cargo.toml
ctz: Yes, that sounds good to me too. #423 sounds like it could reuse that in the future.

s-arash: Done!

rustls/src/server/tls12.rs
s-arash: I tried simplifying it a bit first. Let me know what you think. If still too unwieldy, I can try your suggestion.

djc: I think I would still suggest changing this the way I proposed in my previous comment. It might even make sense to do that for the current setup in an early commit so that this commit can focus on the changes made for FFDHE.

Note in particular that I usually prefer not to have consecutive combinators like `.filter().find()`, since the closures can just be combined into one. I also expect that isolating this code in a separate function will bring together blocks of code that are currently separated (extracting `group_ext`, then finding the `group`) and will bring opportunities to "linearize" the code by using early returns to handle simple/exceptional cases.

s-arash: I added a `pick_kx_group()` function to `CompleteClientHelloHandling`. Let me know what you think. If it looks good, I'll squash dc5e74b15c1.

djc: I was suggesting that the method you've now called `pick_kx_group()` take `&ClientHelloPayload`, so we bring it together with the earlier block of code that defines `group_ext`. I think it would be good to update the `pick_kx_group()` commit to add that, and I'd actually prefer to keep that commit in the commit history, to keep the refactoring separate from the changes needed for FFDHE support.

rustls/src/server/tls12.rs
```diff
@@ -273,14 +273,35 @@ mod client_hello {
...
-            groups_ext: &[NamedGroup],
+            groups_ext: Option<&[NamedGroup]>,
         ) -> Option<&'static dyn SupportedKxGroup> {
-            self.config
-                .provider
-                .kx_groups
-                .iter()
-                .find(|skxg| groups_ext.contains(&skxg.name()))
-                .cloned()
+            let groups_ext_or_default = groups_ext.unwrap_or(&[]);
```
djc, line 278: Style suggestions:

* Rename `groups_ext_or_default` to `peer_kx_groups` -- as I understand it this represents the counterpart to `our_kx_groups` (and the `or_default` suffix doesn't really make sense in terms of a value IMO)
* IMO multi-line combinator expressions in `if` conditions make the code hard to follow -- probably better to name those values which should also make it clearer what's going on
* Would be good to invert the last branch so we can yield `None` early and reduce rightward drift in picking a group from `our_kx_groups`

s-arash: Suggestions applied!

rustls/src/server/tls12.rs
s-arash: To do that, I'd need to take `Context` as well, so I can send alerts from `pick_kx_group()`. Should I do that?
Also note that strictly speaking, this would change the semantics of `handle_client_hello()`. E.g., currently, it checks for existence of the groups extension before checking resumption tickets. With the suggested change, we don't do that.

djc: If we get no complaints from bogo I think it's probably okay.

rustls/src/crypto/mod.rs
```diff
@@ -299,41 +349,112 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    /// [RFC8446 section 4.2.8.1](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.8.1).
     fn pub_key(&self) -> &[u8];
 
     /// Return the group being used.
     fn group(&self) -> NamedGroup;
 }
 
 /// The result from [`ActiveKeyExchange::complete`].
-pub struct SharedSecret(Vec<u8>);
+pub struct SharedSecret {
```
djc, line 420: I would probably extract the change to name the field into a separate commit here, since it's logically separate. (I'd prefer a separate commit rather than adding this into your "refactoring" commit.)

rustls/src/msgs/handshake.rs
```diff
@@ -1577,35 +1642,142 @@ impl Codec<'_> for ServerEcdhParams {
...
+        use KeyExchangeAlgorithm::*;
+        Ok(match algo {
+            ECDHE => Self::Ecdh(ServerEcdhParams::read(r)?),
+            DHE => Self::Dh(ServerDhParams::read(r)?),
+        })
+    }
+}
+
+#[derive(Debug)]
+pub struct ServerKeyExchange {
```
djc, line 1740: Nit: it would be nice to extract the "boring" changes here (create a `ServerKeyExchange` type and a `ServerKeyExchangeParams` type, renaming the `ServerKeyExchangePayload` variant) into a separate commit, as much as is possible/makes sense.

djc (approved):
I think this is basically looking good. The commit history could/should be improved a bit, the "refactoring" commit looks a bit like a dumping ground now -- IMO it would be preferable to reorganize this to separate into commits that each only contain things that are related to each other (though renaming `ecdh_` prefixes to `kx_` is relation enough in this context). In any case, the goal is to make "the large commit" as small as possible because it doesn't need to do any refactoring, just slot in the new logic.

rustls/src/builder.rs
```diff
@@ -210,6 +211,38 @@ impl<S: ConfigSide> ConfigBuilder<S, WantsVersions> {
             return Err(Error::General("no kx groups configured".into()));
         }
 
+        // verifying cipher suites have matching kx groups
+        let mut supported_kx_algos = Vec::with_capacity(ALL_KEY_EXCHANGE_ALGORITHMS.len());
```
djc, line 215: Hmm, reviewing this here, IIRC we want to avoid referencing `ALL_` constants in this way, because we should avoid linking algorithms that aren't used. Maybe use `provider.kx_groups.len()` instead and drop the optimization?

rustls/src/crypto/mod.rs
```diff
@@ -304,16 +314,67 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    ) -> Result<SharedSecret, Error> {
+        if tls_version.version != ProtocolVersion::TLSv1_2 {
+            return self.complete(peer_pub_key);
+        }
+
+        let group = self.group();
+        let mut complete_res = self.complete(peer_pub_key)?;
+        let kx = group
+            .key_exchange_algorithm()
+            .expect("`ActiveKeyExchange` with invalid group");
```
djc, line 364: Add a comment on why this is statically impossible, or make it yield an `Error` instead?

rustls/src/msgs/handshake.rs
```diff
@@ -1635,14 +1795,8 @@ impl ServerKeyExchangePayload {
         if let Self::Unknown(ref unk) = *self {
             let mut rd = Reader::init(unk.bytes());
 
-            let params = match kxa {
```
djc: Nit: to minimize back and forth, I would modify the earlier commit to just inline `ServerKeyExchangeParams::Ecdh(ServerEcdhParams::read(&mut rd).ok()?)` into the `ServerKeyExchange` initialization expression.

rustls/src/server/hs.rs
```diff
@@ -351,6 +351,30 @@ impl ExpectClientHello {
...
+                .config
+                .provider
+                .supported_kx_group_names()
+                .collect::<Vec<_>>(),
+        );
+
+        if suitable_suites.is_empty() {
+            return Err(cx.common.send_fatal_alert(
+                AlertDescription::HandshakeFailure,
+                PeerIncompatible::NoKxGroupsInCommon,
```
djc, line 368: If `suitable_suites` goes empty, we don't really know if the lack of common key exchange groups was at fault, right? It might be due to signature algorithm, version or protocol?

rustls/src/suites.rs
```diff
@@ -176,10 +190,7 @@ pub(crate) fn choose_ciphersuite_preferring_server(
 
 /// Remove the suites in `all` that are
 /// incompatible with `SignatureAlgorithm` `sigalg`.
-pub(crate) fn reduce_given_sigalg(
-    all: &mut Vec<SupportedCipherSuite>,
-    sigalg: SignatureAlgorithm,
-) {
+pub(crate) fn reduce_given_sigalg(all: &mut Vec<SupportedCipherSuite>, sigalg: SignatureAlgorithm) {
```
djc, line 193: Nit: this belongs in the earlier commit.

djc (approved):

djc:
@ctz / @cpu I think this could benefit from review by someone other than me at this point! I think it's in pretty good shape.

rustls/src/builder.rs
ctz: I think `ALL_KEY_EXCHANGE_ALGORITHMS` doesn't link in the actual algorithms, it is just a static list of `KeyExchangeAlgorithm` enums.

rustls/src/server/hs.rs
ctz: I think you're right. Could use `NoCipherSuitesInCommon` instead?

rustls/src/msgs/enums.rs
```diff
@@ -193,6 +194,25 @@ enum_builder! {
     }
 }
 
+impl NamedGroup {
+    /// Return the **known** key exchange algorithm associated with the `NamedGroup`
+    pub fn key_exchange_algorithm(&self) -> Option<KeyExchangeAlgorithm> {
+        match self {
+            Self::secp256r1 | Self::secp384r1 | Self::secp521r1 | Self::X25519 | Self::X448 => {
```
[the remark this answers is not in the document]
ctz: Sorry -- to explain the rationale -- I hope that people can implement (say) `NamedGroup::Unknown(26)` to implement `brainpoolP256r1` and have this work as expected, without us having to keep in sync with IANA and all the current proposed drafts.

rustls/src/server/hs.rs
```diff
@@ -334,16 +334,19 @@ impl ExpectClientHello {
         };
         let certkey = ActiveCertifiedKey::from_certified_key(&certkey);
 
+        let mut suitable_suites = self
+            .config
+            .provider
+            .cipher_suites
+            .to_vec();
```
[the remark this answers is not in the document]
cpu: I think if you make this change `suites::reduce_given_sigalg` and `suites::reduce_given_version_and_protocol` can both be removed - it doesn't look like they have any other callers.

rustls/src/builder.rs
cpu: Can this feedback thread be resolved then? It seems like yes

openssl-tests/src/ffdhe.rs
```diff
@@ -0,0 +1,90 @@
+use num_bigint::BigUint;
+use rustls::{
+    crypto::{
+        ActiveKeyExchange, CipherSuiteCommon, KeyExchangeAlgorithm, SharedSecret, SupportedKxGroup,
+    },
+    ffdhe_groups::FfdheGroup,
+    CipherSuite, NamedGroup, SupportedCipherSuite, Tls12CipherSuite,
+};
```
cpu, line 8: I think these imports could use a tidy to match the [convention]https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#imports) where we separate imports that don't share a parent module.

openssl-tests/src/ffdhe_kx_with_openssl.rs
```diff
@@ -0,0 +1,239 @@
...
+    version::{TLS12, TLS13},
+    ClientConfig, RootCertStore, ServerConfig, SupportedProtocolVersion,
+};
+use rustls_pemfile::Item;
+use rustls_pki_types::{CertificateDer, PrivateKeyDer};
+
+use crate::{
+    ffdhe::{self, FfdheKxGroup},
+    utils::verify_openssl3_available,
+};
```
cpu, line 21: ditto here RE: import style nit

openssl-tests/src/ffdhe.rs
```diff
@@ -0,0 +1,90 @@
...
+
+static TLS12_DHE_RSA_WITH_AES_128_GCM_SHA256: Tls12CipherSuite =
```
cpu, line 10: I think this `static` should be below the `pub` members

openssl-tests/src/ffdhe_kx_with_openssl.rs
```diff
@@ -0,0 +1,239 @@
...
+
+use openssl::ssl::{SslAcceptor, SslFiletype, SslMethod};
+
+const CERT_CHAIN_FILE: &str = "../test-ca/rsa/end.fullchain";
+const PRIV_KEY_FILE: &str = "../test-ca/rsa/end.key";
+const CA_FILE: &str = "../test-ca/rsa/ca.der";
+const CA_PEM_FILE: &str = "../test-ca/rsa/ca.cert";
```
cpu, line 28: I think these constants should be at the bottom of the file

openssl-tests/src/ffdhe_kx_with_openssl.rs
```diff
@@ -0,0 +1,239 @@
...
+
+fn root_ca() -> RootCertStore {
```
cpu, line 30: I think these helpers should be below the `#[test]` instances that use them to match the [top-down ordering convention](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#top-down-ordering-within-modules)

openssl-tests/src/utils.rs
```diff
@@ -0,0 +1,48 @@
...
+        Err(e) => {
+            panic!("OpenSSL 3+ needs to be installed and in PATH.\nThe error encountered: {e}")
+        }
+    }
+}
+
+pub fn verify_openssl3_available() {
+    static VERIFIED: Lazy<()> = Lazy::new(verify_openssl3_available_internal);
+    *VERIFIED
+}
```
cpu, line 48: I think this should be above the non-exported verify fn

openssl-tests/src/utils.rs
```diff
@@ -0,0 +1,48 @@
...
+        .output();
+    match openssl_output {
+        Ok(output) => {
+            if !output.status.success() {
+                panic!(
+                    "OpenSSL exited with an error status: {}\n{}",
+                    output.status,
+                    std::str::from_utf8(&output.stderr).unwrap_or_default()
+                );
+            }
```
cpu, line 16: I think you could lift this conditional into its own `match` arm above the `Ok(output)` expressed as:

```rust
        Ok(output) if !output.status.success() {
            panic!(
                "OpenSSL exited with an error status: {}\n{}",
                output.status,
                std::str::from_utf8(&output.stderr).unwrap_or_default()
            );
        }
```

openssl-tests/src/validate_ffdhe_params.rs
```diff
@@ -0,0 +1,88 @@
...
+    let (p, g) = get_ffdhe_params_from_openssl(group);
+    let openssl_params = FfdheGroup::from_params_trimming_leading_zeros(&p, &g);
+    let rustls_params = FfdheGroup::from_named_group(group).unwrap();
+    assert_eq!(rustls_params.named_group(), Some(group));
+
+    assert_eq!(rustls_params, openssl_params);
+}
+
+#[test]
+fn ffdhe_params_correct() {
```
cpu, line 78: similar nit here RE: placing the test above the helpers it uses.

rustls/src/msgs/ffdhe_groups.rs
```diff
@@ -37,6 +37,23 @@ impl<'a> FfdheGroup<'a> {
             _ => None,
         }
     }
+
+    /// Construct an `FfdheGroup` from the given `p` and `g`, trimming any potential leading zeros.
+    pub fn from_params_trimming_leading_zeros(p: &'a [u8], g: &'a [u8]) -> Self {
```
cpu, line 42: Could this be lifted into the commit that introduces the `FfdheGroup`?

cpu (commented):
Really nice work! Thank you for the tidy commit history.

I don't have any significant feedback. I think most of my comments are related to style-guide nits in the test code.

rustls/src/suites.rs
s-arash: The function is gone now!

rustls/src/crypto/mod.rs
s-arash: It's not statically impossible. It can happen if a `CryptoProvider` has incorrectly behaving `ActiveKeyExchange`s in it. I made it panic because an `ActiveKeyExchange` should never behave like this.
Should I keep this behavior, or should I make it return an `Error` instead?

rustls/src/server/hs.rs
s-arash: You are absolutely right. I think it's still worth it to return the more accurate `NoKxGroupsInCommon`. So for now I fixed it by checking that `suitable_suites` was not empty before. Let me know if it's OK or if I should revert this change.

rustls/src/msgs/enums.rs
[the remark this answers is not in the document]
s-arash: Good observation! Will make this change.

rustls/src/client/tls12.rs
```diff
@@ -460,8 +461,14 @@ impl State<ClientConnectionData> for ExpectServerKx<'_> {
...
         {
-            let crate::msgs::handshake::ServerKeyExchangeParams::Ecdh(ecdh) = kx.params;
-            debug!("ECDHE curve is {:?}", ecdh.curve_params);
+            match &kx.params {
+                ServerKeyExchangeParams::Ecdh(ecdhe) => {
+                    debug!("ECDHE curve is {:?}", ecdhe.curve_params)
+                }
+                ServerKeyExchangeParams::Dh(dhe) => {
+                    debug!("DHE params are p = {:?}, g = {:?}", dhe.dh_p, dhe.dh_g)
+                }
```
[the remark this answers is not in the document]
s-arash: I wanted it to be in line with ECDHE logging, where we only log the curve params and not the public key.

rustls/src/client/tls12.rs
```diff
@@ -512,10 +519,22 @@ fn emit_certificate(
     common.send_msg(cert, false);
 }
 
-fn emit_clientkx(transcript: &mut HandshakeHash, common: &mut CommonState, pub_key: &[u8]) {
+fn emit_clientkx(
```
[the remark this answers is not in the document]
s-arash: Will update!

rustls/tests/api_ffdhe.rs
```diff
@@ -0,0 +1,424 @@
...
+        ..provider::default_provider()
+    };
+
+    assert_eq!(
+        ClientConfig::builder_with_provider(bad_crypto_provider.into())
+        .with_safe_default_protocol_versions()
+        .err(),
+        Some(Error::General("Ciphersuite TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 requires [ECDHE] key exchange, but no \
+                            [ECDHE]-compatible key exchange groups were present in `CryptoProvider`'s `kx_groups` field".into()))
+    );
```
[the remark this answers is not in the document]
s-arash: I'm all for it. I think I was following the example of another test.

rustls/src/client/tls12.rs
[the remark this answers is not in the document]
cpu: Ah! I missed that the existing log line for the ECDHE logging was using `{:?}` with the `ecdhe.curve_params` field and not the entire `ecdhe` struct.

Your reasoning makes sense to me. Thanks!

rustls/src/crypto/mod.rs
s-arash: Per @ctz's comment, I made `key_exchange_algorithm()` total, so `expect` is not needed here anymore.

rustls/tests/api_ffdhe.rs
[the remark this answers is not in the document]
s-arash: Done!

s-arash:
Thanks @ctz and @cpu for reviewing the PR! I addressed your comments. Please let me know what you think.

openssl-tests/src/ffdhe_kx_with_openssl.rs
```diff
@@ -0,0 +1,231 @@
+use std::fs::{self, File};
+use std::io::{BufReader, Read, Write};
+use std::net::{TcpListener, TcpStream};
+use std::{str, sync::Arc, thread};
```
cpu, line 4: nit: I think `sync::Arc` would be its own `use` line since it doesn't share a direct parent with `str` and `thread`.

openssl-tests/src/ffdhe_kx_with_openssl.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use rustls::crypto::{ring::default_provider, CryptoProvider};
```
cpu, line 6: same nit here for `rustls::crypto::ring::default_provider`.

openssl-tests/src/validate_ffdhe_params.rs
```diff
@@ -0,0 +1,88 @@
+use base64::prelude::*;
+use rustls::{ffdhe_groups::FfdheGroup, NamedGroup};
```
cpu, line 2: nit: separate `rustls::NamedGroup`  from `rustls::ffdhe_groups::FfdheGroup`.

cpu (approved):
Looks great, thank you!

I flagged a couple small tweaks to the update imports but they're very minor. Sorry to be pedantic :-) Hopefully we can automate this away soon.

ctz (approved):

s-arash:
> Appreciate the patience and persistence - thanks!

Thanks again for reviewing the PR!

s-arash:
Also thanks to @Taowyoo for reviewing the PR (here and in our fork).

PR #1704, opened 2023-12-21
no-std demo for stm32f429zi

Dajamante:
All details in README in no-std-embedded-demo/


djc:
I think we should start with a discussion about whether this should live in the rustls repository?

cpu:
In the effort of keeping our pull request tracker focused on work in progress I'm going to close this for now. Like https://github.com/rustls/rustls/pull/1534, were this to be revived I think we would want to see it done in a separate repository.

Thanks all,

ctz:
FWIW I am rescuing this into https://github.com/ctz/rustls-embedded-demo

PR #1706, opened 2023-12-22
Support P521-SHA512 signatures with aws-lc-rs

### LINKED ISSUE #1666: P-521 with aws-lc-rs
**Checklist**

https://github.com/rustls/rustls/issues/933 was closed with reference to the pluggable crypto providers. But aws-lc-rs is now supported but without P-521 support.

rustls should support certificates signed with P-521 when using the aws-lc-rs backend.

When using rustls with default options and the aws-lc-rs provider, P-521 signatures should get verified correctly.

I'm not aware of any alternatives. Support for P-521 could be opt in, but currently there is no way to do this either.

**Additional context**

AIUI aws-lc-rs has support for P-521 but the necessary constants are not defined in the webpki crate and the scheme is not part of the SUPPORTED_SIG_ALGS constant in the aws-lc-rs backend.

ctz:
This will likely have some benchmark result noise, because the curves used in ECDSA benchmarks have changed.

Work remaining:

- [x] reinstate coverage of P384 in benchmarks
- [x] attend to testing gaps in crypto/*/sign.rs

closes #1666


ctz:
I think this is ready for review, though the first commit would need to be dropped before it can be merged.

rustls/tests/common/mod.rs
```diff
@@ -323,6 +323,22 @@ impl KeyType {
         .next() // We only expect one CRL.
         .unwrap()
     }
+
+    pub fn ca_distinguished_name(&self) -> &'static [u8] {
```
cpu, line 327: Thanks for tidying this up :-)

djc:
Can/should we backport this to 0.22?

PR #1708, opened 2023-12-23
Implement FromIterator for RootCertStore

niklasf:
Natural addition for a type that already implements `Extend<TrustAnchor<'static>>`.

### DIFF
--- a/rustls/src/webpki/anchors.rs
+++ b/rustls/src/webpki/anchors.rs
@@ -103,6 +103,14 @@ impl RootCertStore {
     }
 }
 
+impl FromIterator<TrustAnchor<'static>> for RootCertStore {
+    fn from_iter<T: IntoIterator<Item = TrustAnchor<'static>>>(iter: T) -> Self {
+        let mut store = Self::empty();
+        store.extend(iter);
+        store
+    }
+}
+
 impl Extend<TrustAnchor<'static>> for RootCertStore {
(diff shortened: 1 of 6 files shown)


rustls/src/webpki/anchors.rs
```diff
@@ -103,6 +103,14 @@ impl RootCertStore {
     }
 }
 
+impl FromIterator<TrustAnchor<'static>> for RootCertStore {
+    fn from_iter<T: IntoIterator<Item = TrustAnchor<'static>>>(iter: T) -> Self {
+        let mut store = Self::empty();
+        store.extend(iter);
+        store
```
cpu, line 110: I think this might be more natural as:
```suggestion
        Self {
            roots: iter.into_iter().collect(),
        }
```

cpu (approved):
This seems reasonable. The `roots` field of `RootCertStore` is `pub` so there was already a fairly direct path from `Iterator<Item = TrustAnchor<'static>>>` -> `RootCertStore` via struct initialization + collect, but using `FromIterator` to express that path does seem like an improvement :+1:

rustls/src/webpki/anchors.rs
niklasf: This would reduce coverage by taking a separate path for `Extend`. To be fair, `Extend` is trivial anyway, so go for it regardless?

djc (approved):

PR #1713, opened 2024-01-02
WASM support

Endercass:
Ring already has support for WASM, so this PR just adds a feature that enables the wasm32_unknown_unknown_js flag for that crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -992,8 +992,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fe9006bed769170c11f845cf00c7c1e9092aeb3f268e007c3e760ac68008070f"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


ctz:
Appreciate the effort here, but I think wasm support needs a bit more thought:

- I'd prefer not to introduce a feature for this -- instead if we can make it work properly based on the target platform, that would be preferable (cargo bugs in this area notwithstanding).
- I think when we claim to support WASM, that should be justified by building and testing the library for it, and building and running sample code in representative ways.
- "wasm32_unknown_unknown_js" is WASM with a browser JS environment. That's a valid thing to want, but others will want to use WASM without that. I don't think we should convolve the two.

Remember that cargo features are additive across your end-application, so if your use requires enabling the "ring/wasm32_unknown_unknown_js" feature, you can just do that once in your application Cargo.toml. Intermediate crates that depend on ring (like this one) don't need to be involved in that.

Endercass:
I fully understand. To be completely honest, I forgot about that aspect of cargo features. I had just created the fork in order to add that feature, then I was suggested to submit a PR by a friend of mine. I should have put more thought and testing into this. Thank you for your feedback, however! It is very appreciated.

Issue #1711, opened 2023-12-29
Can someone explain what this library is for? 


fade2black:
Can someone explain what this library is for? 
I do understand how TLS/HTTPS/etc work, but I can't understand what this library provides? 
We usually establish a TCP connection, then TLS handshake, and then we start to exchange data. 
So, do I need to implement the handshake myself? Do I need to establish a TCP  connection myself? Where does this lib fit in this entire process?

The documentation is, as usually like many crates, is terrible, and I couldn't find anything meaningful to start with. PLEASE don't refer me to the 700-line code to "start" with.

jadamcrain:
@fade2black `rustls` is the "transport independent" block on which higher level libraries are written. Chances are you're looking for tokio-rustls which integrates with tokio's TcpSocket types:

https://crates.io/crates/tokio-rustls

fade2black:
@jadamcrain Thank you for a quick response.
Well, let me put it this way.
Both server and and client run locally on my Macbook.
A server runs and waits for incoming connections.  Assume a client initiates a TCP connection. Upon a successful connection I want them to create a TLS session (agree upon a asymmetric key). 
The server spawns a new thread for each established TLS session and they simply ping/pong some random strings encrypted with the session key.

I hope the problem is crystal clear. However, the documentation is toooooo terrible, it does not even include such a simple example and does not provide a simple reference.

jadamcrain:
I think the example client and server in the `tokio-rustls` crate do exactly that:

https://github.com/rustls/tokio-rustls/tree/main/examples

The server spawns a new  task onto the Tokio runtime to handle each connection concurrently.

djc:
When you declare our documentation to be terrible, please be very specific about both the problem you're trying to solve and which documentation you're reading, and which questions it is failing to answer.

fade2black:
I think either the API design is really bad and should be reviewed completely or I miss smth.
For example, suppose I have a piece of code written in Tokio or some other async lib. Doesn’t have to be Tokio. It is straightforward to write in Tokio to write a simple serv/client app where a server accepts client connections, spawns a thread and waits for another request, while the worker threads “talk” to the clients on the bkgd.

Now, I want to encrypt the communication. So, I thought I could establish a tcp connection and immediately after that using your library could write a single line or couple of code lines to perform a handshake and establish a session . Instead, I have to use weird connectors, etc.

djc:
The tokio-rustls README (though not the reference documentation) leads with this:

```rust
[...]

PR #1715, opened 2024-01-02
examples: add very simple server, README 

cpu:
In the process of reading through https://github.com/rustls/rustls/issues/1711 it occurred to me that while we have a very simple client example, we don't have a very simple server example.

This branch adapts the example written by @djc in that thread, and additionally adds a README describing the existing examples and providing guidance on which might be interesting to look at first.

### DIFF
--- a/examples/src/bin/simpleserver.rs
+++ b/examples/src/bin/simpleserver.rs
@@ -1,0 +1,51 @@
+//! This is the simplest possible server using rustls that does something useful:
+//! it accepts the default configuration, loads a server certificate and private key,
+//! and then accepts a single client connection.
+//!
+//! You must either set the CERTFILE and PRIV_KEY_FILE env vars to point to a server
+//! certificate and private key, or place 'localhost.pem' and 'localhost-key.pem' in
+//! the directory you run this example from.
+//!
+//! Note that `unwrap()` is used to deal with networking errors; this is not something
+//! that is sensible outside of example code.
+
+use std::error::Error as StdError;
+use std::fs::File;
+use std::io::{BufReader, Read, Write};
+use std::net::TcpListener;
+use std::sync::Arc;
+
+const CERTFILE: &str = match option_env!("CERTFILE") {
+    Some(certfile) => certfile,
+    None => "localhost.pem",
+};
+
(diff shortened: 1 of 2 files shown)


examples/src/bin/simpleserver.rs
```diff
@@ -0,0 +1,53 @@
...
+//! Note that `unwrap()` is used to deal with networking errors; this is not something
+//! that is sensible outside of example code.
+
+use std::error::Error as StdError;
+use std::fs::File;
+use std::io::{BufReader, Read, Write};
+use std::net::TcpListener;
+use std::sync::Arc;
+
+use rustls::ServerConnection;
```
ctz, line 18: I would drop this import and fully qualify the single use on line 42, for consistency with `rustls::ServerConfig` on line 35. Generally I like to write out full names in example code, where that doesn't cause immensely long lines or otherwise hide the purpose of the example.

cpu: Good point, fixed.

ctz (approved):

examples/src/bin/simpleserver.rs
```diff
@@ -0,0 +1,51 @@
...
+
+const CERTFILE: &str = match option_env!("CERTFILE") {
+    Some(certfile) => certfile,
+    None => "localhost.pem",
+};
+
+const PRIV_KEY_FILE: &str = match option_env!("PRIV_KEY_FILE") {
+    Some(priv_key_file) => priv_key_file,
+    None => "localhost-key.pem",
+};
```
djc, line 26: I was surprised by these, both that they precede `main()` (which seems counter to our usual item ordering) and also by the use of environment variables, whereas we use command-line arguments in what I thought were most of the other examples.

(Even more nitty, I also find the naming of `CERT_FILE` without underscore and the `PRIV` abbreviation instead of `PRIVATE_` too be not quite idiomatic.)

cpu: I lifted this from the unbuffered examples that were added recently, so they would need adjustment for this feedback as well

cpu: https://github.com/rustls/rustls/pull/1719

PR #1717, opened 2024-01-02
Misc testing improvements

ctz:
This is mostly chasing some low-hanging coverage gaps.

Along the way I did find that `SignatureScheme::ED25519.sign() == SignatureAlgorithm::Unknown(0)` -- still evaluating the impact of that.

### DIFF
--- a/rustls/src/crypto/ring/kx.rs
+++ b/rustls/src/crypto/ring/kx.rs
@@ -75,7 +75,6 @@ pub static ALL_KX_GROUPS: &[&dyn SupportedKxGroup] = &[X25519, SECP256R1, SECP38
 
 /// An in-progress key exchange.  This has the algorithm,
 /// our private key, and our public key.
-#[derive(Debug)]
 struct KeyExchange {
     name: NamedGroup,
     agreement_algorithm: &'static agreement::Algorithm,
@@ -99,6 +98,14 @@ impl ActiveKeyExchange for KeyExchange {
     /// Return the public key being used.
     fn pub_key(&self) -> &[u8] {
         self.pub_key.as_ref()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn kxgroup_fmt_yields_name() {
+        assert_eq!("X25519", format!("{:?}", super::X25519));
     }
 }
 
--- a/rustls/src/crypto/ring/ticketer.rs
+++ b/rustls/src/crypto/ring/ticketer.rs
@@ -194,4 +194,19 @@ mod tests {
         assert_eq!(t.decrypt(&cipher2).unwrap(), b"ticket 2");
         assert_eq!(t.decrypt(&cipher3).unwrap(), b"ticket 3");
     }
+
+    #[test]
+    fn aeadticketer_is_debug_and_producestickets() {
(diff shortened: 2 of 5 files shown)


rustls/src/crypto/ring/kx.rs
```diff
@@ -102,6 +101,14 @@ impl ActiveKeyExchange for KeyExchange {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn kxgroup_is_debug() {
+        assert_eq!("X25519", format!("{:?}", super::X25519));
+    }
```
cpu, line 109: I'm curios what coverage this is adding. The `Debug` bound is on `SupportedKxGroup` - `KxGroup` doesn't derive `Debug`. Should this be named `supported_kx_group_is_debug`?

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -194,4 +194,27 @@ mod tests {
...
+            lifetime: 1234,
+        };
+
+        assert_eq!(
+            format!("{:?}", t),
+            "AeadTicketer { alg: CHACHA20_POLY1305, lifetime: 1234 }"
+        );
+        assert!(t.enabled());
+        assert_eq!(t.lifetime(), 1234);
+    }
```
cpu, line 219: I think this test can be simplified if you use `make_ticket_generator()` - was there a reason you wanted to avoid that?

```rust
#[test]
    fn aeadticketer_is_debug_and_producestickets() {
        use super::*;
        use alloc::format;

        let t = make_ticket_generator().unwrap();

        assert_eq!(
            format!("{:?}", t),
            "AeadTicketer { alg: CHACHA20_POLY1305, lifetime: 43200 }"
        );
        assert!(t.enabled());
        assert_eq!(t.lifetime(), 43200);
    }
```

(_Note: had to change the expected lifetime - but the value used is static with the impl_)

rustls/tests/unbuffered.rs
```diff
@@ -17,9 +18,9 @@ const MAX_ITERATIONS: usize = 100;
 
 #[test]
 fn tls12_handshake() {
-    let (client_transcript, server_transcript) = handshake(&rustls::version::TLS12);
+    let outcome = handshake(&rustls::version::TLS12);
```
cpu, line 22: This is a nice refactoring, great stuff :+1:

cpu (approved):
Looks great, just a couple nits :+1: Thanks for chasing this coverage down.

rustls/src/crypto/ring/kx.rs
ctz: renamed to `kxgroup_fmt_yields_name`: this test is wanting to observe the behaviour of `KxGroup::fmt`, which is accessed via the Debug trait bound on `SupportedKxGroup`.

rustls/src/crypto/ring/ticketer.rs
ctz: oh, good idea!

rustls/src/crypto/ring/kx.rs
cpu: SGTM!

PR #1719, opened 2024-01-03
examples: fix style w.r.t env var usage, const placement

cpu:
Resolving [feedback from djc](https://github.com/rustls/rustls/pull/1715#discussion_r1440697185) on https://github.com/rustls/rustls/pull/1715

### DIFF
--- a/examples/src/bin/simpleserver.rs
+++ b/examples/src/bin/simpleserver.rs
@@ -9,27 +9,27 @@
 //! Note that `unwrap()` is used to deal with networking errors; this is not something
 //! that is sensible outside of example code.
 
+use std::env;
 use std::error::Error as StdError;
 use std::fs::File;
 use std::io::{BufReader, Read, Write};
 use std::net::TcpListener;
 use std::sync::Arc;
 
-const CERTFILE: &str = match option_env!("CERTFILE") {
-    Some(certfile) => certfile,
-    None => "localhost.pem",
-};
-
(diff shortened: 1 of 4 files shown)


examples/src/bin/simpleserver.rs
```diff
@@ -9,27 +9,30 @@
...
-
 fn main() -> Result<(), Box<dyn StdError>> {
-    let certs = rustls_pemfile::certs(&mut BufReader::new(&mut File::open(CERTFILE)?))
+    let mut args = env::args();
+    let cert_file = match args.next() {
+        Some(certfile) => certfile,
+        None => "localhost.pem".into(),
+    };
+
+    let private_key_file = match args.next() {
```
ctz, line 26: Not sure it really makes sense to allow one argument (which will, from my reading, be the certificate file name with the private key taken from the "localhost-key.pem" default). I'd be in favour of dropping the defaults altogether, and just requiring two arguments.

cpu: Good point. I've simplified.

ctz (approved):

djc (approved):

djc:
Eventually I think we should use clap in derive mode for all the examples (especially the ones currently using docopt), but that can probably wait.

cpu:
> Eventually I think we should use clap in derive mode for all the examples (especially the ones currently using docopt)

I agree for the ones using docopt, but I think the "simple" examples benefit from just being dead simple Rust with as few dependencies above/beyond Rustls as possible.

PR #1718, opened 2024-01-02
Encrypted Client Hello support (client only)

cpu:
## Description

This branch brings encrypted client hello (ECH) compatibility to Rustls **clients** based on [draft-ietf-tls-esni-18](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18). 

Closes https://github.com/rustls/rustls/issues/199

## Limitations

* It does _not_ add any support for writing Rustls' servers that accept ECH (in either shared or split mode). I think client support is the most important at this stage, and this branch is already very large/complex. We will track server-side support in https://github.com/rustls/rustls/issues/1980

* Extension compression is implemented separately in https://github.com/cpu/rustls/pull/6

<details>
<summary>Comparison to previous PRs:</summary>

## Server Name Enum

Initially we expected to update the `ServerName` enum to have a variant for ECH - I tried this initially and found that it didn't fit the design of ECH very well. Mainly the issue is that we need to carry a lot of additional state beyond the inner SNI name. Putting that state into the `ServerName` is challenging, and an impedance mismatch with the existing usage. Having separate state _and_ the ECH variant in `ServerName` meant having two pieces of information and having to contend with what to do if one is present but not the other. This is also made increasingly complex since we lifted the `ServerName` type into the pki-types crate. As a point of reference, you can compare this branch with the approach in https://github.com/rustls/rustls/pull/663 where a `ServerIdentity` enum stands in for `ServerName` and contains a `EncryptedClientHello(Box<EncryptedClientHello>)` variant that holds a `Box` of heap allocated state.

## Credit where credit is due

I started this branch by first adopting work from what @sayrer started in https://github.com/rustls/rustls/pull/663. Many thanks for kicking this effort off and giving me a solid foundation to start with!

Some major changes compared to that branch:

* This branch updates from draft-10 to draft-18, rolling in associated changes (mostly related to confirmation, extension types, etc).
* As discussed above, I opted to not use a `ServerName` variant for holding the inner SNI name and associated state.
* I've caught up the diff to the current Rustls `main`, adjusting for various `Codec` changes and associated code drift.
* HPKE usage is handled through the trait, abstracting away the choice of implementation as opposed to using hpke-rs directly.
* An HPKE provider backed by aws-lc-rs was implemented so we can have a crypto provider out-of-box that supports ECH.
* I've implemented support for ECH confirmation across Hello Retry Requests - this in turn required reworking how the "inner" transcript is maintained, calculating the HRR confirmation, and threading through more bits of state across the handshake.
[...]

PR #1725, opened 2024-01-04
macros: allow trailing enum_builder comma

cpu:
Small quality of life improvement pulled out of a suggestion [from djc](https://github.com/rustls/rustls/pull/1718#discussion_r1441536825) on #1718.

Without allowing a trailing comma for invocations of the `enum_builder!` macro we end up creating messy two line diffs for every addition to an existing built enum.

This branch updates the macro definition to allow an optional trailing comma, and then once the macro allows a trailing comma each usage is updated to include one. This will make any future diffs that add elements to these enums easier to review.


ctz (approved):
nice!

djc:
Woo, thanks for fixing that up!

PR #1726, opened 2024-01-04
msgs/handshake.rs: remove get_ fn prefixes

cpu:
This is a stylistic tidying pulled out of review feedback [from djc](https://github.com/rustls/rustls/pull/1718#discussion_r1441548330) on #1718 - we want to avoid proliferating new `get_xxx` fns in this file, but there are so many existing usages that it would be a jarring inconsistency to change style for new fns. 

Instead, this branch fixes all of the existing instances in one go so that moving forward we can avoid the `get_` prefix while maintaining consistency. We already call out avoiding `get_` prefixes in [our style guide](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#avoid-get_-prefixes).

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -42,7 +42,7 @@ impl ClientHelloDetails {
         allowed_unsolicited: &[ExtensionType],
     ) -> bool {
         for ext in received_exts {
-            let ext_type = ext.get_type();
+            let ext_type = ext.ext_type();
             if !self.sent_extensions.contains(&ext_type) && !allowed_unsolicited.contains(&ext_type)
             {
                 trace!("Unsolicited extension {:?}", ext_type);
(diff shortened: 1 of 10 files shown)


djc (approved):
Nice!

cpu:

A couple of the AWS LC RS benchmarks seem to be a bit noise-y again. Findings were raised here and on https://github.com/rustls/rustls/pull/1725 and both are no-op PRs.

ctz:
> A couple of the AWS LC RS benchmarks seem to be a bit noise-y again.

They have been disturbed because I changed some of the test names in #1706. AIUI it will eventually self-correct based on past results.

ctz (approved):
👍🏽

cpu:
> A couple of the AWS LC RS benchmarks seem to be a bit noise-y again.
> 
> They have been disturbed because I changed some of the test names in https://github.com/rustls/rustls/pull/1706. AIUI it will eventually self-correct based on past results.

Ahhh! yes, I remember reading your note about that now. Thanks for the reminder :+1:

PR #1729, opened 2024-01-05
Don't add empty `certificate_authorities` extension

### LINKED ISSUE #1727: About empty root hints list in "certificate_authorities" extension

This change (https://github.com/rustls/rustls/pull/1553/commits/3a4d9e289044c7d601799481ddad67b381ef5a08) in `0.22` cause one compatibility issue with OpenJdk.
The empty root hints list in `certificate_authorities` extension is not accepted by JDK.

It is not compatible with Java's TLS library (which is JSSE, all versions applys), see:
- https://bugs.openjdk.org/browse/JDK-8322509
- https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/security/ssl/CertificateAuthoritiesExtension.java#L66-L96

And Java's implementation is supported by following words in the RFC https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2.3:

>    -  If the "certificate_authorities" extension in the
>       CertificateRequest message was present, at least one of the
>       certificates in the certificate chain SHOULD be issued by one of
>       the listed CAs.

I am not mean the implementation in `rustls` is wrong, we just now have a compatibility issue because of the flexible part in the RFC.

I am wondering is it possible to have a new config field in server side to customize the behavior of sending `certificate_authorities` extension, such as:
- having a boolean to disable `certificate_authorities` extension
- having a boolean to avoid sending  `certificate_authorities` extension when root hints list returned by `ClientCertVerifier` is empty.

**Reproduce**

See my repo: https://github.com/Taowyoo/java-tls13-bug

**Note**

I am also trying to submit a patch on Java side to change their behavior. But at least one side needs change.

ctz:
There's already a test case for this and the API-level behaviour is unchanged.

fixes #1727

### DIFF
--- a/rustls/examples/internal/bogo_shim_impl.rs
+++ b/rustls/examples/internal/bogo_shim_impl.rs
@@ -48,6 +48,7 @@ struct Options {
     resumes: usize,
     verify_peer: bool,
     require_any_client_cert: bool,
+    root_hint_subjects: Vec<DistinguishedName>,
     offer_no_client_cas: bool,
     tickets: bool,
     resume_with_tickets_disabled: bool,
@@ -105,6 +106,7 @@ impl Options {
             shut_down_after_handshake: false,
             check_close_notify: false,
             require_any_client_cert: false,
+            root_hint_subjects: vec![],
             offer_no_client_cas: false,
             key_file: "".to_string(),
             cert_file: "".to_string(),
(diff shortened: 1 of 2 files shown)


djc:
If there is already a test case, how come it didn't fail? It must not be a good test case for this particular failure mode?

djc (approved):

ctz:
The existing test case is `client_cert_resolve_server_no_hints`, and checks that a client sees empty hints when a server is configured to provide that. The difference between an absent and empty extension is collapsed here in the client: https://github.com/rustls/rustls/blob/b3913a5ac2ecc4d3daea3c4259c84bfeceb749f6/rustls/src/client/common.rs#L75-L76 which is why the API-level behaviour is unchanged, even if our protocol behaviour is different.

---

With that said, I just took a second look, and:

- there's a bogo test for this (looks like boringssl also made the same mistake we did), though we don't have support for it in `bogo_shim`
- adding support (on main) gives:

```
FAILED (TLS13-Empty-Client-CA-List)
unexpected failure: local error 'tls: expected no certificate_authorities extension', child error 'none', stdout:
```
- it passes after this change

djc:
Okay, nice. Why don't we support that in the bogo shim? Would it be hard to add support?

cpu (approved):
Thank you. This looks good to me :+1: 

I feel like we should backport this to 0.22, do you agree? If so, maybe we should pick up https://github.com/rustls/rustls/pull/1706 while we're at it?

Issue #357, opened 2020-04-09
Implement RFC 8701 - GREASE Extensions


jabedude:
Hello all! RFC 8701 has been published, and I was wondering if this project would be okay with implementing it? I would be willing to work on this feature. 

I have a TLS-related crate that could benefit from rustls supporting GREASE, so this would be useful for me!

ctz:
I don't have any objection against this, but I'd prefer this not to add to the configuration/API surface of the library -- I think care is needed to balance the costs and benefits here. I think the benefit of GREASE in a library like this with (honestly) small deployment footprint is limited; while the cost of any additional API complexity is high.

I think GREASE being implemented in BoringSSL (which is, obviously, hugely well deployed) achieves 99% of its aims. It's not even implemented by OpenSSL...

jabedude:
I totally understand your point about the cost/benefits of adding unnecessary tuning knobs here. With that concern in mind, would this be appropriate behind a feature flag? Or could it simply be a part of the default rustls handshake? The second option seems best for not adding messy configuration.

The only downside would be the (slight, I'd imagine) overhead of the additional extension.

austinhartzheim:
@ctz BoringSSL has seen wide deployment thanks to its inclusion in Chrome. Comparatively, the footprint of Rustls is small, but may cover additional systems not normally accessed via web browser. If you are still open to adding RFC8701 support, I've been working on an implementation (#375 ) which keeps the configuration and API changes to a minimum.

I have placed the configuration behind a feature flag (as @jabedude suggested). The current implementation modular to allow swapping alternate GREASE implementations in (for example, if one was building a TLS fuzzer or scanner requiring GREASE support), but that interface need not be exposed if it is contrary to the goals of this crate. In general, no configuration changes should be required by the user beyond enabling the feature flag.

My implementation currently supports the "client ciphers list" and "client extensions" fields. I've managed to achieve that while maintaining the majority of the GREASE logic inside a single file with minimal changes elsewhere in the code. It seems that the internal and external APIs need not change (minus the addition of some GREASE functions to the internal API).

Let me know your thoughts. I can continue working on the PR if you would like to see a more complete implementation.

jabedude:
...I guess I'll close #358 then? ¯\_(ツ)_/¯

briansmith:
> I have a TLS-related crate that could benefit from rustls supporting GREASE, so this would be useful for me!

[...]

PR #1730, opened 2024-01-05
Randomize ClientHello extensions

### LINKED ISSUE #1313: Randomise ClientHello extension ordering as client fingerprinting countermeasure

GomesGoncalo:
Google Chrome project proposes Client Hello extensions should be
randomized in order to prevent fingerprinting [1]

This commit sorts all the extensions that have been sent in the same
order as before and randomizes all new extensions that have not been sent
in the position after the extensions already seen.
And keeps the PSK extension in the end.

[1] https://chromestatus.com/feature/5124606246518784

resolves https://github.com/rustls/rustls/issues/1313

### DIFF
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -38,12 +38,14 @@ impl<'a> ServerCertDetails<'a> {
 
 pub(super) struct ClientHelloDetails {
     pub(super) sent_extensions: Vec<ExtensionType>,
+    pub(super) extension_order_seed: u16,
 }
 
 impl ClientHelloDetails {
-    pub(super) fn new() -> Self {
+    pub(super) fn new(extension_order_seed: u16) -> Self {
         Self {
             sent_extensions: Vec::new(),
+            extension_order_seed,
         }
     }
 
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -141,6 +141,7 @@ pub(super) fn start_handshake(
     };
 
     let random = Random::new(config.provider.secure_random)?;
+    let extension_order_seed = crate::rand::random_u16(config.provider.secure_random)?;
 
     Ok(emit_client_hello_for_retry(
         transcript_buffer,
@@ -155,7 +156,7 @@ pub(super) fn start_handshake(
             #[cfg(feature = "tls12")]
             using_ems: false,
             sent_tls13_fake_ccs: false,
-            hello: ClientHelloDetails::new(),
+            hello: ClientHelloDetails::new(extension_order_seed),
             session_id,
             server_name,
         },
@@ -270,6 +271,22 @@ fn emit_client_hello_for_retry(
 
     // Do we have a SessionID or ticket cached for this host?
     let tls13_session = prepare_resumption(&input.resuming, &mut exts, suite, cx, config);
+
+    // Extensions MAY be randomized
+    // but they also need to keep the same order as the the previous ClientHello
+    exts.sort_by_cached_key(|new_ext| {
+        // PSK extension is always last
+        if let ClientExtension::PresharedKey(..) = new_ext {
+            return u32::MAX;
+        }
+
+        let seed =
+            (input.hello.extension_order_seed as u32) << 16 | (new_ext.ext_type().get_u16() as u32);
+        match low_quality_integer_hash(seed) {
+            u32::MAX => 0,
+            key => key,
+        }
+    });
 
     // Note what extensions we sent.
     input.hello.sent_extensions = exts
@@ -928,4 +945,20 @@ impl Deref for ClientSessionValue {
     fn deref(&self) -> &Self::Target {
         self.common()
     }
+}
+
+fn low_quality_integer_hash(mut x: u32) -> u32 {
+    x = x
+        .wrapping_add(0x7ed55d16)
+        .wrapping_add(x << 12);
+    x = (x ^ 0xc761c23c) ^ (x >> 19);
+    x = x
+        .wrapping_add(0x165667b1)
+        .wrapping_add(x << 5);
+    x = x.wrapping_add(0xd3a2646c) ^ (x << 9);
+    x = x
+        .wrapping_add(0xfd7046c5)
+        .wrapping_add(x << 3);
+    x = (x ^ 0xb55a4f09) ^ (x >> 16);
+    x
 }
--- a/rustls/src/rand.rs
+++ b/rustls/src/rand.rs
@@ -22,6 +22,13 @@ pub(crate) fn random_u32(secure_random: &dyn SecureRandom) -> Result<u32, GetRan
     Ok(u32::from_be_bytes(buf))
 }
 
+/// Return a uniformly random [`u16`].
+pub(crate) fn random_u16(secure_random: &dyn SecureRandom) -> Result<u16, GetRandomFailed> {
+    let mut buf = [0u8; 2];
+    secure_random.fill(&mut buf)?;
+    Ok(u16::from_be_bytes(buf))
+}
+
 /// Random material generation failed.
 #[derive(Debug)]
 pub struct GetRandomFailed;
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -5623,10 +5623,10 @@ fn test_client_construction_fails_if_random_source_fails_in_second_request() {
 }
 
 #[test]
-fn test_client_construction_requires_64_bytes_of_random_material() {
+fn test_client_construction_requires_66_bytes_of_random_material() {
     static FAULTY_RANDOM: FaultyRandom = FaultyRandom {
         rand_queue: Mutex::new(
-            b"nice random number generator !!!\
+            b"nice random number generator !!!!!\
                                  it's really not very good is it?",
         ),
     };


cpu:
> rustls / BoGo test suite (pull_request) Failing after 51s

This is the same issue that https://github.com/rustls/rustls/pull/1336 had w.r.t [Hello Retry Requests](https://github.com/rustls/rustls/pull/1336#issuecomment-1617601301).

GomesGoncalo:
> > rustls / BoGo test suite (pull_request) Failing after 51s
> 
> This is the same issue that #1336 had w.r.t [Hello Retry Requests](https://github.com/rustls/rustls/pull/1336#issuecomment-1617601301).

Yeah after raising this I found that PR and saw that comment.
So, I just read this:
```
   There MUST NOT be more than one extension of the
   same type in a given extension block.
```
This helps a bit I think, if in the 6 situations described in 4.1.2 the ClientHello must be exactly the same & we are storing the order of extension types in `input.hello.sent_extensions` & each extension type being unique, thatt means we can use the vector of `sent_extensions` to order it if it has been populated before

GomesGoncalo:
Kind of annoyingly I don't see the build failing in my dev setup, updating it with a fix

cpu:
> 293 |                 .map(|ext| (*ext).get_type())
>    |                                   ^^^^^^^^ help: there is a method with a similar name: `ext_type`

This looks like it would have been caused by the tidying refactor I did in https://github.com/rustls/rustls/pull/1726 Try rebasing on main and switching `get_type` for `ext_type` in your new work.

GomesGoncalo:
> > 293 |                 .map(|ext| (*ext).get_type())
> > |                                   ^^^^^^^^ help: there is a method with a similar name: `ext_type`
> 
> This looks like it would have been caused by the tidying refactor I did in #1726 Try rebasing on main and switching `get_type` for `ext_type` in your new work.

That worked, I ran one of the actions that was failing and ran BoGo with both the providers and it seemed to pass.

rustls/src/client/hs.rs
```diff
@@ -271,6 +271,47 @@ fn emit_client_hello_for_retry(
     // Do we have a SessionID or ticket cached for this host?
     let tls13_session = prepare_resumption(&input.resuming, &mut exts, suite, cx, config);
 
+    // Extensions MAY be randomised
+    // PSK message must be last, if it is we cannot move it
+    let randomisable_extensions = exts.len()
+        - if let Some(&ClientExtension::PresharedKey(..)) = exts.last() {
```
djc, line 277: Style nit: I think this would look be better if you use a `match` instead of `if let`.

Despite the parent issue, IIRC we do use US English spelling through the code base, so let's call it "randomized" and "randomizable"?

rustls/src/client/hs.rs
```diff
@@ -271,6 +271,47 @@ fn emit_client_hello_for_retry(
...
+
+            if let Some((pos, _)) = pos {
+                // By putting all extensions that have been seen in the start of the vector we
+                // make others bubble to the end
+                exts.swap(pos + acc, acc);
+                acc + 1
+            } else {
+                acc
+            }
+        });
```
djc, line 305: It's not entirely obvious to me what this is doing or why. I also think the `filter().fold()` (with another fairly long chained iterator expression inside) is a great way to express this (as opposed to a syntactical loop).

GomesGoncalo: This is to solve the issue where next Client Hellos need to include the extensions in the same order as the initial one. It was the part of the previous PRs that was failing in the BoGo test as all it had been done was naively shuffling everything.

I read in the TLS spec that each extension must only appear once, so with that assumption if we saved the extension type in `input.hello.sent_extensions` before, now we can generate it in the same order.

So for example: 
```
1st Client Hello sent this way (giving numbers to ext types): |2|1|4|3|PSK|
```
Now if we generate a ClientHello like this `|4|3|1|5|6|PSK|`
It iterates through the save ClientHello and finds position of the element in the newly generated extension list:
So for example 1st iteration, it'd see that ext n2 is missing, so `acc` keeps being 0, ignores it.
2nd iteration: ext n1 is in position 2, so swap 2 with 0, `acc` is now 1 and vector in this state:
```
`|1|3|4|5|6|PSK|`
```
3rd iteration: ext n4 is in position 1 now (we skipped the 1st element as it's sorted already), so swap 2 (`pos + acc`) with 1 (`acc`); `acc` is now 2 and result of swap is:
```
|1|4|3|5|6|PSK|
```
4rd iteration: ext 3 is in position 0 (as we skip the 2 first ones), so swap 2 with 2 (granted this one was unnecessary) and `acc` is now 3, resulting of swap is:
```
|1|4|3|5|6|PSK|
```
At the end of the loop `acc` would be the number of the extensions sorted so all other new extensions after those (for example a new Cookie extension) would be shuffled, except the PSK extension of course.
And it filters out the PSK one in the beginning cause there's no point processing that as it needs to be last.

This could be done with a for statement but I kinda needed the point where the extensions stopped being sorted as per the saved vector to know where to start shuffling which would require me to create a mutable variable, so this avoids that.

Then after this it starts randomising from `start` (3 in the previous example) to the index of the PSK extension (exclusive).

Not sure if there's an action to do with this comment, sorry couldn't understand, should I document this in the code? The `skip()` as well is not needed, it's just an optimisation in order to reduce the amount of processing to do on elements that were already processed, it could be removed if you think it makes it more confusing.

rustls/src/client/hs.rs
GomesGoncalo: Done!

rustls/src/client/hs.rs
djc: So this code is written to do nothing/unnecessary if `sent_extensions` is empty? IMO that should be made obvious by wrapping it in an `if !input.hello.sent_extensions.is_empty() { .. }`.

I would suggest that if this is doing some kind of sorting, it would be helpful to use [`sort_by_cached_key()`](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by_cached_key) if at all possible (seems like it should be?).

GomesGoncalo: Yeah was having a look at that and it's indeed possible, does it look a bit better now?

rustls/src/client/hs.rs
```diff
@@ -271,6 +271,47 @@ fn emit_client_hello_for_retry(
     // Do we have a SessionID or ticket cached for this host?
     let tls13_session = prepare_resumption(&input.resuming, &mut exts, suite, cx, config);
 
+    // Extensions MAY be randomized
+    // but they also need to keep the same order as the the previous ClientHello
+    exts.sort_by_cached_key(|new_ext| {
+        match new_ext {
+            // PresharedKey is always be last
+            ClientExtension::PresharedKey(..) => usize::max_value(),
```
djc, line 279: I'd propose the following structure:

```rust
if let ClientExtension::PresharedKey(..) = new_ext {
    return usize::MAX;
}

if input.hello.sent_extensions.is_empty() {
     ..
}

..
```

This helps reduce rightward drift. (And replaces `max_value()` with `MAX`, which is more idiomatic.)

djc:
Nice, this looks much better!

GomesGoncalo:
> Nice, this looks much better!
> 
> What exactly are the constraints in the case the set of extensions in the retry request is different? Do we have to retain the same ordering among the extensions already sent but can randomly intersperse others? Or do the extensions sent in the initial ClientHello need to be contiguous and we can append new ones at the end? (Do we ever send new ones?) Would be good to explicitly document these edge cases in comments.

IMO the RFC a bit confusing regarding this, it says `the client MUST send the same ClientHello without modification, except as follows` and then has a list of scenarios where things could be different. But I don't interpret it as them saying "just build a new ClientHello and shuffle it however you like", for example:

```
   -  If a "key_share" extension was supplied in the HelloRetryRequest,
      replacing the list of shares with a list containing a single
      KeyShareEntry from the indicated group.
   -  Updating the "pre_shared_key" extension if present by recomputing
      the "obfuscated_ticket_age" and binder values and (optionally)
      removing any PSKs which are incompatible with the server's
      indicated cipher suite.
   -  Optionally adding, removing, or changing the length of the
      "padding" extension [[RFC7685](https://www.rfc-editor.org/rfc/rfc7685)].
```

In these ones, are the extension allowed to appear in a different positions? Or must it be in the same position with a different content? Not very clear, my view is that since they don't mandate shuffling, it is safer to assume some implementations will want them in the same position (which is what this code does)

```
   -  Removing the "early_data" extension ([Section 4.2.10](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.10)) if one was
      present.  Early data is not permitted after a HelloRetryRequest.
```

For this one, I assumed that all the extensions would keep their relative position to each other and this space would be removed from the vector. I could have assumed that if we are changing something in the ClientHello, then we can shuffle everything, but there's no guarantee that all implementations make that assumption so felt safer to keep relative ordering (obviously if an extension is removed in the middle of the list we can't just leave a hole :smile:, so everything would be offset).

```
   -  Including a "cookie" extension if one was provided in the
      HelloRetryRequest.
```

This one I assumed that all the other already seen extensions would keep their ordering and any new extension would appear last before PSK (it should be fine as the RFC doesn't mandate any ordering other than PSK being last). In case there are multiple new extensions (not sure that's possible) they'd be in random order as well, didn't see any constraint in that.

```
   -  Other modifications that may be allowed by an extension defined in
      the future and present in the HelloRetryRequest.
```

This one is very vague! I assumed they mean that any RFC that describes a TLS extension is free to constrain what can be done with the extension. Again thought that since shuffling is not mandated, would be safer to keep it in the same place if it existed (whether the values inside were modified or not)
I wrote this in less words in the comment.

djc (approved):
This looks great!

My one remaining nit is that you have a bunch of instances of `max_value()`, all of which should become `MAX` -- sorry I missed that earlier.

GomesGoncalo:
> This looks great!
> 
> My one remaining nit is that you have a bunch of instances of `max_value()`, all of which should become `MAX` -- sorry I missed that earlier.

I didn't notice the docs saying that it's going to be deprecated ah! Changed it!

GomesGoncalo:
@djc regarding the benchmarks, I don't have enough context to say this is good or bad, I do expect some more processing done on subsequent ClientHellos because of the sorting and the loop to calculate position. I think I just thought of a possible optimisation (not sure it'll actually improve things - and for sure it'll make it look more complicated), thinking I could swap the element found in line 297 with the elements in the end of the vector (it's again more computation but might reduce the iteration, maybe we'll see better results)
Are the benchmarks significantly worse & should I attempt to get it a bit better?

djc:
@ctz do you want to pronounce on the performance impact here? I think you have a better sense on what's noise here or what kind of impact we'd allow for this.

rustls/src/client/hs.rs
```diff
@@ -271,6 +271,56 @@ fn emit_client_hello_for_retry(
...
+                _ => match crate::rand::random_u32(config.provider.secure_random) {
+                    Ok(value) => {
+                        // We do not know exactly how many extensions have been sorted as we
+                        // can remove extensions (e.g. early_data) but we know that if no extensions
+                        // have been removed we want to put the same ones in the same
+                        // positions so worst case we need to skip `sent_extensions.len()` elements
+                        // This will not create any issues as this number is not an actual index of
+                        // the vector but just set the numerical ordering used for the extensions
+                        (value as usize % (usize::MAX - input.hello.sent_extensions.len()))
+                            + input.hello.sent_extensions.len()
```
cpu, line 309: I don't think it's significant in this context, but it feels worth noting that this isn't uniformly random. As I understand things it will suffer a [modulo bias](https://romailler.ch/2020/07/28/crypto-modulo_bias_guide/).

Maybe worth mentioning that explicitly in the comment?

djc: I was looking for that too, but I thought it was mitigated? IIRC mitigating it is not too hard.

ctz:
I think we can actually do something quite a bit simpler here. Two observations:

- the order should be random, but the amount of entropy can be low and it doesn't need to be cryptographic quality. We immediately publish the order we come up with anyway, and it doesn't matter if any particular order is more likely than another (within reason).
- if we have a small, fixed-size random seed value, we can just keep that to produce the same ordering again -- that should eliminate the need for the O(N^2) search in `sent_extensions` (nb. N is very small in this case, so this wasn't a significant issue.)

As an example, we could pass the extension type and the random seed into a integer hash, to produce an extension's sort key. Very roughly, that looks like:

```diff
commit b7c067a9b46517302333b399ce66460513ece30c
Author: Joseph Birr-Pixton <jpixton@gmail.com>
Date:   Tue Jan 9 18:08:38 2024 +0000

    hax

diff --git a/rustls/src/client/common.rs b/rustls/src/client/common.rs
index 7c49e423..50cd53a0 100644
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -27,12 +27,16 @@ impl ServerCertDetails {
 
 pub(super) struct ClientHelloDetails {
     pub(super) sent_extensions: Vec<ExtensionType>,
+    pub(super) extension_order_seed: u32,
 }
 
 impl ClientHelloDetails {
-    pub(super) fn new() -> Self {
+    pub(super) fn new(mut extension_order_seed: u32) -> Self {
+        // leave bottom 16 bits clear for extension type.
+        extension_order_seed &= 0xffff0000;
         Self {
             sent_extensions: Vec::new(),
+            extension_order_seed,
         }
     }
 
diff --git a/rustls/src/client/hs.rs b/rustls/src/client/hs.rs
index 2a78002a..5c9ff614 100644
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -141,6 +141,7 @@ pub(super) fn start_handshake(
     };
 
     let random = Random::new(config.provider.secure_random)?;
+    let extension_order_seed = crate::rand::random_u32(config.provider.secure_random)?;
 
     Ok(emit_client_hello_for_retry(
         transcript_buffer,
@@ -155,7 +156,7 @@ pub(super) fn start_handshake(
             #[cfg(feature = "tls12")]
             using_ems: false,
             sent_tls13_fake_ccs: false,
-            hello: ClientHelloDetails::new(),
+            hello: ClientHelloDetails::new(extension_order_seed),
             session_id,
             server_name,
         },
@@ -273,54 +274,37 @@ fn emit_client_hello_for_retry(
 
     // Extensions MAY be randomized
     // but they also need to keep the same order as the the previous ClientHello
-    exts.sort_by_cached_key(|new_ext| {
+    exts.sort_unstable_by_key(|new_ext| {
         // PSK extension is always last
         if let ClientExtension::PresharedKey(..) = new_ext {
-            return usize::MAX;
+            return u32::MAX;
         }
 
-        if !input.hello.sent_extensions.is_empty() {
-            // There is an existing saved ClientHello, we make some assumptions here:
-            // - Extensions are unique (mandated by RFC 8446 section-4.2)
-            // - If extensions are removed, all ordering between the existing extension relative to
-            // the old ClientHello is kept, this is unclear from the RFC but it is the safer
-            // approach for compatibility with assumptions in existing TLS server implementations
-            // - If extensions are added they are randomized after the list of existing extensions
-            // and before PSK extension. The existing extensions are kept in the same order. Again
-            // this is unclear from the RFC but it is the cautious approach to maximize
-            // compatibility, since shuffling is not explicitly mandated.
-            match input
-                .hello
-                .sent_extensions
-                .iter()
-                .enumerate()
-                .find(|(_, &saved_ext)| saved_ext == new_ext.ext_type())
-            {
-                Some((pos, _)) => pos,
-                _ => match crate::rand::random_u32(config.provider.secure_random) {
-                    Ok(value) => {
-                        // We do not know exactly how many extensions have been sorted as we
-                        // can remove extensions (e.g. early_data) but we know that if no extensions
-                        // have been removed we want to put the same ones in the same
-                        // positions so worst case we need to skip `sent_extensions.len()` elements
-                        // This will not create any issues as this number is not an actual index of
-                        // the vector but just set the numerical ordering used for the extensions
-                        (value as usize % (usize::MAX - input.hello.sent_extensions.len()))
-                            + input.hello.sent_extensions.len()
-                    }
-                    _ => usize::MAX,
-                },
-            }
+        let seed = input.hello.extension_order_seed | (new_ext.ext_type().get_u16() as u32);
+        let key = low_quality_integer_hash(seed);
+        if key == u32::MAX {
+            0
         } else {
-            // As there was no saved ClientHello no considerations here regarding order,
-            // other than the PSK extension being last
-            match crate::rand::random_u32(config.provider.secure_random) {
-                Ok(value) => value as usize,
-                _ => usize::MAX,
-            }
+            key
         }
     });
 
+    fn low_quality_integer_hash(mut x: u32) -> u32 {
+        x = x
+            .wrapping_add(0x7ed55d16)
+            .wrapping_add(x << 12);
+        x = (x ^ 0xc761c23c) ^ (x >> 19);
+        x = x
+            .wrapping_add(0x165667b1)
+            .wrapping_add(x << 5);
+        x = x.wrapping_add(0xd3a2646c) ^ (x << 9);
+        x = x
+            .wrapping_add(0xfd7046c5)
+            .wrapping_add(x << 3);
+        x = (x ^ 0xb55a4f09) ^ (x >> 16);
+        x
+    }
+
     // Note what extensions we sent.
     input.hello.sent_extensions = exts
         .iter()
```

GomesGoncalo:
> ```diff
> sort_by_cached_key
> ```

This might not hold the same assumptions in case a Cookie extension is added (for example). In the previous code all the other extensions's order would be maintained, with this code the Cookie extension might be inserted in the middle of the others in a random position. Now I don't really know if that'll be a problem or not as the RFC to me is not very clear if that's a problem or not

ctz:
> Now I don't really know if that'll be a problem or not as the RFC to me is not very clear if that's a problem or not

I think strictly on the level of RFC compliance, it isn't a problem (because the RFC never talks about extension order outside of the PSK binder, and treats them as an unordered set). But we get a good amount of value from reusing BoringSSL's test suite, and that wants to see the same order in a given connection -- though I'm not sure if it covers the case when an extension is added in a HelloRetryRequest.

Anyway, I am happy with this either way. I think I would still like to pursue #1475 as a better implementation strategy for extensions. But it will have implications on ordering and ECH too.

GomesGoncalo:
>  never talks about extension order outside of the PSK binder, and treats them as an unordered set [...] BoringSSL's test suite, and that wants to see the same order in a given connection

The RFC actually mandates that:
```
   [...] The client will also send a
   ClientHello when the server has responded to its ClientHello with a
   HelloRetryRequest.  In that case, the client MUST send the same
   ClientHello without modification, [...]
```

I think it's the case where extensions are removed, added and/or modified that it's not very clear (at least to me) how much we can change.

djc:
> This might not hold the same assumptions in case a Cookie extension is added (for example, there might be other extensions that could be added that will have the same issue). In the previous code all the other extensions's order would be maintained, with this code the Cookie extension might be inserted in the middle of the others in a random position. Now I don't really know if that'll be a problem or not as the RFC to me is not very clear if that's a problem or not

Given that we don't really know if it's a problem, I think it'd be okay to assume for now that it isn't.

> The other issue with this code (not sure that was on purpose maybe I misunderstood it) is that hashes can collide and since this is unstable sort, the order of the extensions can actually change.

Actually, `sort_by_cached_key()` is documented to be stable, to this doesn't seem to be an issue?

GomesGoncalo:
> Actually, `sort_by_cached_key()` is documented to be stable, to this doesn't seem to be an issue?

Yes, I was commenting regarding the usage of `sort_unstable_by_key` in the diff posted in this comment: https://github.com/rustls/rustls/pull/1730#issuecomment-1883562739

> Given that we don't really know if it's a problem, I think it'd be okay to assume for now that it isn't.

Ok, that's fine by me :smile: Would it be okay to update the patch to the code in the comment and change the sorting function to the stable one? It should be much faster than my submission

djc:
> Ok, that's fine by me 😄 Would it be okay to update the patch to the code in the comment and change the sorting function to the stable one? It should be much faster than my submission

I think that sounds like the right trade-off, yes.

rustls/src/client/hs.rs
```diff
@@ -271,6 +272,39 @@ fn emit_client_hello_for_retry(
...
+        x = x
+            .wrapping_add(0x165667b1)
+            .wrapping_add(x << 5);
+        x = x.wrapping_add(0xd3a2646c) ^ (x << 9);
+        x = x
+            .wrapping_add(0xfd7046c5)
+            .wrapping_add(x << 3);
+        x = (x ^ 0xb55a4f09) ^ (x >> 16);
+        x
+    }
```
djc, line 306: I think I'd prefer to move this out of the way, to the bottom of the module.

Not sure if the other maintainers agree?

rustls/src/client/hs.rs
```diff
@@ -271,6 +272,39 @@ fn emit_client_hello_for_retry(
...
+    // Extensions MAY be randomized
+    // but they also need to keep the same order as the the previous ClientHello
+    exts.sort_by_cached_key(|new_ext| {
+        // PSK extension is always last
+        if let ClientExtension::PresharedKey(..) = new_ext {
+            return u32::MAX;
+        }
+
+        let seed = input.hello.extension_order_seed | (new_ext.ext_type().get_u16() as u32);
+        let key = low_quality_integer_hash(seed);
```
djc, line 284: I'd spell this as

```rust
match low_quality_integer_hash(seed) {
    u32::MAX => 0,
    key => key,
}
```

rustls/src/client/common.rs
```diff
@@ -27,12 +27,16 @@ impl ServerCertDetails {
...
 }
 
 impl ClientHelloDetails {
-    pub(super) fn new() -> Self {
+    pub(super) fn new(mut extension_order_seed: u32) -> Self {
+        // leave bottom 16 bits clear for extension type.
+        extension_order_seed &= 0xffff0000;
         Self {
             sent_extensions: Vec::new(),
+            extension_order_seed,
```
djc, line 39: Would prefer to instead write:

```
            // Leave bottom 16 bits clear for extension type.
            extension_order_seed: extension_order_seed & 0xffff_0000,
```

(That is, avoiding the extra binding.)

djc (approved):

GomesGoncalo:
Test is failing cause we are forcing  the initial connection to use a `FaultyRandom` generator, that was unexpected!
Is it okay to seed with 0 if the random gen fails?

djc:
> Test is failing cause we are forcing the initial connection to use a `FaultyRandom` generator, that was unexpected! Is it okay to seed with 0 if the random gen fails?

I think with no randomness available, the connection setup will fail either way. The tests that seem to fail are checking how much random material is required to start a connection, we should probably bump that number (and maybe add a `random_u16()` function so that we don't generate extra bytes of random which we're just going to mask out right after?).

rustls/src/client/common.rs
GomesGoncalo: Since I'm now generating u16 this is changed a bit

GomesGoncalo:
(rebased)

rustls/src/client/common.rs
```diff
@@ -38,12 +38,15 @@ impl<'a> ServerCertDetails<'a> {
...
+    pub(super) extension_order_seed: u32,
 }
 
 impl ClientHelloDetails {
-    pub(super) fn new() -> Self {
+    pub(super) fn new(extension_order_seed: u16) -> Self {
         Self {
             sent_extensions: Vec::new(),
+            // leave bottom 16 bits clear for extension type.
+            extension_order_seed: (extension_order_seed as u32) << 16,
```
djc, line 49: Sorry, looking at this I have one more nit: maybe the field here should remain an `u16` and we shift it at use time (as in `(input.hello.extension_order_seed as u32) << 16`). What do you think? Seems like that's clearer about what we're doing here?

djc (approved):

rustls/src/client/common.rs
GomesGoncalo: I don't have a strong opinion about this, on one hand I think what you're suggesting is clearer, on the other hand this is a bit less processing if done ahead of time (although I doubt it'd matter xD)

GomesGoncalo: Changed it!

GomesGoncalo:
No worries, got to learn a bit of a new codebase and read a bit of the TLS spec; it was pretty cool, thanks for all the reviews!

ctz (approved):

briansmith:
It is very likely that some and perhaps many applications will need to disable the randomization if/when they find compatibility problems with it. I think there really has to be an option to turn it off. I also think it would be good to have the option disabled by default at least for a while until people can experiment with it.

BTW, the use of the word "fingerprinting" in Chrome's description of what they're doing is easy to misinterpret as having some kind of privacy motivation. I don't think privacy is the motivation but rather they want to "prevent ossification", i.e. it is basically an extension of GREASE. See #357.

ctz:
My motivation for wanting to do this is less about ossification -- I don't think we're big enough to make a dent in anything, which is why we don't do anything like GREASE. And certainly don't have the telemetry, reach or resources to debug middlebox issues and get people like F5 to fix them.

Instead, it's more about removing a source of entropy in JA3 fingerprints. Google, Apple, Cloudflare and others use these to (in practice) deliberately break interop with non-OpenSSL connections. Once extension randomisation breaks that sufficiently, I guess fingerprinters will need to sort the extensions, and their fingerprints will have significantly less entropy.

briansmith:
Fair enough. I do think it is worth making it optional, so that the application can decide whether to do it.

Issue #1642, opened 2023-11-29
Process-level selection of CryptoProvider


jsha:
Right now, if a `CryptoProvider` is chosen, it's done on a per-config basis (more or less per-connection). That means, if different crates use rustls independently, and they explicitly configure `CryptoProvider`, they may choose different ones. That's non-ideal because it means multiple implementations linked in the final binary. And some of the use cases for `CryptoProvider` are policy ones: an organization may wish to only cryptography it has reviewed; or may be required to only use FIPS cryptography.

I think the status quo is that crates that use rustls should expose a way to pass in a configuration. For instance, HTTP clients and servers; SDK clients that build on top of HTTP clients.

You can imagine three levels at which to set program-level policy about which cryptographic providers to use:

 - Linker level: link only one cryptographic provider. This is how the C world does it, but the additive nature of Cargo features makes it hard to enforce in the Rust world.
 - Config / connection level: users have to know all the places their binary might use rustls, and make sure they pass a correct runtime configuration for them.
 - Process level: the binary is responsible for setting a global value at most once. [This is how the `log` package does it](https://docs.rs/log/latest/log/fn.set_logger.html).
 
Global state is bad, but I'm starting to think it may be a more effective way to set policy for an entire program.
 
Under the status quo, we probably need to document a strong requirement that non-binary crates using rustls must have a way for their callers to pass `ClientConfig` / `ServerConfig`.

djc:
I think it's good to have this discussion, but I still lean towards "global state is still bad and something to be avoided".

Assorted thoughts:

* How to configure the desired crypto provider is one aspect, but auditability of which crypto providers might be used in a given binary is another. I would argue that, as long as it is easy/straightforward to audit which crypto providers are used, it is somewhat less important that the configurability. IMO the cargo-audit tooling and `Cargo.lock` files give us pretty good tools to audit binaries built by Cargo, at least. I suppose it's harder to audit at the symbol level because there are many small functions involved in cipher suites, the API doesn't avoid mixing and matching, etc.

[...]

PR #1732, opened 2024-01-06
Add 'fips' cargo feature to enable to the aws-lc-rs crypto provider with the AWS-LC FIPS implementations.

skmcgrail:
Adopts the following commits:
* [aws-lc-rs: avoid chaha20poly1305 for ticketer algorithm](https://github.com/rustls/rustls/commit/327444fdb8d8bfd74fe88533ed2b0d7fa8cfcfcb)
* [Expose FIPS "service indicator" ](https://github.com/rustls/rustls/commit/afe43b0213879924670e2ddeaf5bd290a2409050)

I've updated the aws-lc-rs CryptoProvider's `DEFAULT_CIPHER_SUITES` to only include ciphers that are `fips_mode() == true` with the 'fips' cargo feature enabled. So this no longer is 1-1 the same list as the provider's `ALL_CIPHER_SUITES`. Applications would still have access to the complete set of cipher suites available in `ALL_CIPHER_SUITES`, just that the default provider would not contain the complete list when operating with the 'fips' feature enabled.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,16 +43,27 @@ jobs:
         if: runner.os == 'Windows'
         uses: ilammy/setup-nasm@v1
 
+      - name: Install ninja-build tool for aws-lc-fips-sys on Windows
+        if: runner.os == 'Windows'
+        uses: seanmiddleditch/gha-setup-ninja@v4
+
       - name: cargo build (debug; default features)
         run: cargo build --locked
 
-      - name: cargo test (debug; all features)
-        run: cargo test --locked --all-features
+      # nb. feature sets that include "fips" should be --release --
+      # this is required for fips on windows.
+      - name: cargo test (release; all features)
+        run: cargo test --release --locked --all-features
         env:
           RUST_BACKTRACE: 1
 
       - name: cargo test (debug; aws-lc-rs)
         run: cargo test --no-default-features --features aws_lc_rs,tls12,read_buf,logging
+        env:
+          RUST_BACKTRACE: 1
+
+      - name: cargo test (release; fips)
+        run: cargo test --release --no-default-features --features fips,tls12,read_buf,logging
         env:
           RUST_BACKTRACE: 1
 
@@ -109,6 +120,10 @@ jobs:
 
       - name: cargo test (debug; no default features; aws-lc-rs,tls12)
         run: cargo test --no-default-features --features aws_lc_rs,tls12
+        working-directory: rustls
+
+      - name: cargo test (debug; no default features; fips,tls12)
+        run: cargo test --no-default-features --features fips,tls12
         working-directory: rustls
 
       - name: cargo test (release; no run)
@@ -185,6 +200,9 @@ jobs:
 
       - name: Smoke-test benchmark program (aws-lc-rs)
         run: cargo run -p rustls --release --locked --example bench --no-default-features --features aws_lc_rs,tls12
+
+      - name: Smoke-test benchmark program (fips)
+        run: cargo run -p rustls --release --locked --example bench --no-default-features --features fips,tls12
 
       - name: Run micro-benchmarks
         run: cargo bench --locked --all-features
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,18 +52,27 @@ jobs:
 
       # nb. feature sets that include "fips" should be --release --
       # this is required for fips on windows.
+      # nb. "--all-targets" does not include doctests
       - name: cargo test (release; all features)
-        run: cargo test --release --locked --all-features
+        run: cargo test --release --locked --all-features --all-targets
+        env:
+          RUST_BACKTRACE: 1
+
+      # nb. this is separate so it can be skipped on macOS & windows, where
+      # doctests don't work: https://github.com/rust-lang/cargo/issues/8531
+      - name: cargo test --doc (release; all-features)
+        if: ${{ runner.os != 'macOS' && runner.os != 'Windows' }}
+        run: cargo test --release --locked --all-features --doc
         env:
           RUST_BACKTRACE: 1
 
       - name: cargo test (debug; aws-lc-rs)
-        run: cargo test --no-default-features --features aws_lc_rs,tls12,read_buf,logging
+        run: cargo test --no-default-features --features aws_lc_rs,tls12,read_buf,logging --all-targets
         env:
           RUST_BACKTRACE: 1
 
       - name: cargo test (release; fips)
-        run: cargo test --release --no-default-features --features fips,tls12,read_buf,logging
+        run: cargo test --release --no-default-features --features fips,tls12,read_buf,logging --all-targets
         env:
           RUST_BACKTRACE: 1
 
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -1,6 +1,8 @@
 use crate::builder::ConfigBuilder;
 use crate::common_state::{CommonState, Protocol, Side};
 use crate::conn::{ConnectionCommon, ConnectionCore, UnbufferedConnectionCommon};
+#[cfg(any(feature = "ring", feature = "fips"))]
+use crate::crypto::default_provider;
 use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::error::Error;
@@ -14,7 +16,7 @@ use crate::suites::{ExtractedSecrets, SupportedCipherSuite};
 use crate::unbuffered::{EncryptError, TransmitTlsData};
 use crate::versions;
 use crate::KeyLog;
-#[cfg(feature = "ring")]
+#[cfg(any(feature = "ring", feature = "fips"))]
 use crate::WantsVerifier;
 use crate::{verify, WantsVersions};
 
@@ -220,31 +222,40 @@ pub struct ClientConfig {
 
 impl ClientConfig {
     /// Create a builder for a client configuration with the default
-    /// [`CryptoProvider`]: [`crypto::ring::default_provider`] and safe ciphersuite and
-    /// protocol defaults.
+    /// [`CryptoProvider`].
+    ///
+    /// This is:
+    ///
+    /// - [`crypto::aws_lc_rs::default_provider`] if the `fips` crate feature is
+    ///   enabled.
+    /// - [`crypto::ring::default_provider`] if the `ring` crate feature is
+    ///   enabled and the `fips` crate feature is not enabled.
+    ///
+    /// If neither of these are true, this function is not available and you
+    /// must use [`ClientConfig::builder_with_provider()`] instead.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
-    #[cfg(feature = "ring")]
+    #[cfg(any(feature = "ring", feature = "fips"))]
     pub fn builder() -> ConfigBuilder<Self, WantsVerifier> {
-        // Safety: we know the *ring* provider's ciphersuites are compatible with the safe default protocol versions.
-        Self::builder_with_provider(crate::crypto::ring::default_provider().into())
+        // Safety: we know the *ring* and aws-lc-rs providers' ciphersuites are compatible with the safe default protocol versions.
+        Self::builder_with_provider(default_provider().into())
             .with_safe_default_protocol_versions()
             .unwrap()
     }
 
     /// Create a builder for a client configuration with the default
-    /// [`CryptoProvider`]: [`crypto::ring::default_provider`], safe ciphersuite defaults and
-    /// the provided protocol versions.
+    /// [`CryptoProvider`] (see [`ClientConfig::builder()`] for details), safe
+    /// ciphersuite defaults and the provided protocol versions.
     ///
     /// Panics if provided an empty slice of supported versions.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
-    #[cfg(feature = "ring")]
+    #[cfg(any(feature = "ring", feature = "fips"))]
     pub fn builder_with_protocol_versions(
         versions: &[&'static versions::SupportedProtocolVersion],
     ) -> ConfigBuilder<Self, WantsVerifier> {
-        // Safety: we know the *ring* provider's ciphersuites are compatible with all protocol version choices.
-        Self::builder_with_provider(crate::crypto::ring::default_provider().into())
+        // Safety: we know the *ring* and aws-lc-rs providers' ciphersuites are compatible with all protocol version choices.
+        Self::builder_with_provider(default_provider().into())
             .with_protocol_versions(versions)
             .unwrap()
     }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -266,6 +266,12 @@ impl ClientConfig {
         }
     }
 
+    /// Return true if connections made with this `ClientConfig` will
+    /// operate in FIPS mode.
+    pub fn fips(&self) -> bool {
+        self.provider.fips()
+    }
+
     /// We support a given TLS version if it's quoted in the configured
     /// versions *and* at least one ciphersuite for this version is
     /// also configured.
--- a/rustls/src/crypto/aws_lc_rs/mod.rs
+++ b/rustls/src/crypto/aws_lc_rs/mod.rs
@@ -197,3 +197,6 @@ mod ring_shim {
         })
     }
 }
+
+/// AEAD algorithm that is used by `mod ticketer`.
+pub(super) static TICKETER_AEAD: &ring_like::aead::Algorithm = &ring_like::aead::AES_256_GCM;
--- a/rustls/src/crypto/aws_lc_rs/mod.rs
+++ b/rustls/src/crypto/aws_lc_rs/mod.rs
@@ -35,8 +35,17 @@ pub(crate) mod tls13;
 /// A `CryptoProvider` backed by aws-lc-rs.
 pub fn default_provider() -> CryptoProvider {
     CryptoProvider {
-        cipher_suites: DEFAULT_CIPHER_SUITES.to_vec(),
-        kx_groups: ALL_KX_GROUPS.to_vec(),
+        // TODO: make this filtering conditional on fips feature
+        cipher_suites: DEFAULT_CIPHER_SUITES
+            .iter()
+            .filter(|cs| cs.fips())
+            .copied()
+            .collect(),
+        kx_groups: ALL_KX_GROUPS
+            .iter()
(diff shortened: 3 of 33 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -286,6 +286,12 @@ impl ClientConfig {
         }
     }
 
+    /// Return true if connections made with this `ClientConfig` will
+    /// operate in FIPS mode.
+    pub fn fips_mode(&self) -> bool {
```
djc, line 293: Naming nit: should we just call it `fips()` instead of `fips_mode()`? ISTM the `_mode` suffix doesn't really add value here.

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -63,6 +76,10 @@ impl KeyProvider for AwsLcRs {
     ) -> Result<Arc<dyn SigningKey>, Error> {
         sign::any_supported_type(&key_der)
     }
+
+    fn fips_mode(&self) -> bool {
+        aws_lc_rs::try_fips_mode().is_ok()
```
djc, line 88: If we have a local wrapper for this, should we just call that directly?

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +186,21 @@ pub struct CryptoProvider {
     pub key_provider: &'static dyn KeyProvider,
 }
 
+impl CryptoProvider {
+    /// Returns if this `CryptoProvider` is operating in FIPS mode.
+    pub fn fips_mode(&self) -> bool {
+        self.cipher_suites
```
djc, line 192: Should we do a `let Self { cipher_suites, .. }` thing here to make it more likely we'll catch field additions?

rustls/src/tls13/mod.rs
```diff
@@ -42,6 +42,19 @@ impl Tls13CipherSuite {
         (prev.common.hash_provider.algorithm() == self.common.hash_provider.algorithm())
             .then(|| prev)
     }
+
+    /// Return true if this is backed by a FIPS-approved implementation.
+    ///
+    /// This means all the constituent parts that do cryptography return true for `fips_mode()`.
+    pub fn fips_mode(&self) -> bool {
+        self.common.fips_mode()
```
djc, line 50: Same here with the field stuff.

rustls/tests/api.rs
```diff
@@ -5705,3 +5705,27 @@ fn test_client_removes_tls12_session_if_server_sends_undecryptable_first_message
         ClientStorageOp::RemoveTls12Session(_)
     ));
 }
+
+#[cfg(feature = "ring")]
+#[test]
+fn test_client_fips_service_indicator() {
+    assert_eq!(make_client_config(KeyType::Rsa).fips_mode(), false);
```
djc, line 5718: Style nit: I think we usually want to use `assert!()` directly for boolean values?

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -35,23 +36,29 @@ pub(crate) mod tls13;
...
-            .iter()
-            .filter(|cs| cs.fips_mode())
-            .copied()
-            .collect(),
-        kx_groups: ALL_KX_GROUPS
-            .iter()
-            .filter(|kx| kx.fips_mode())
-            .copied()
-            .collect(),
+        cipher_suites: DEFAULT_CIPHER_SUITES.to_vec(),
```
djc, line 39: Hmm, how about we add a `default_fips_provider()` which is only available on `feature("fips")`?

djc (approved):

djc:
Exciting to make this work end-to-end!

ctz:
@skmcgrail I'm happy to address the above comments and fix CI if that works for you?

skmcgrail:
> @skmcgrail I'm happy to address the above comments and fix CI if that works for you?

That is alright with me :)

rustls/src/crypto/mod.rs
ctz: Good idea!

ctz:
Just pushed fixes for the above comments.

Two things remain, I think:

- macos job isn't working
- we're currently ignoring `CryptoProvider::signature_verification_algorithms` for the purpose of working out fips status; probably need to chase some of this down into pki-types and webpki. Haven't quite figured out the nicest way to do that.

> I think `lib.rs`'s section on ["Crate Features"](https://github.com/rustls/rustls/blob/20ac87f64f28a7295e0a2085e0702461c3f2f513/rustls/src/lib.rs#L253) also needs an update for the `fips` feature.

I updated that, plus improved the docs around `ClientConfig::builder()` and co.

cpu:
Latest push looks good to me :+1:

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -197,3 +197,32 @@ mod ring_shim {
...
+                .count(),
+            0
+        );
+    }
+
+    #[cfg(not(feature = "fips"))]
+    #[test]
+    fn default_suites() {
+        assert_eq!(super::DEFAULT_CIPHER_SUITES, super::ALL_CIPHER_SUITES);
+    }
```
djc, line 272: I think these tests belong in the next commit? AIUI they'll fail in this commit.

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -197,3 +197,32 @@ mod ring_shim {
...
+    #[cfg(feature = "fips")]
+    #[test]
+    fn default_suites_are_fips() {
+        assert_eq!(
+            super::DEFAULT_CIPHER_SUITES
+                .iter()
+                .filter(|scs| !scs.fips())
+                .count(),
+            0
+        );
```
djc, line 265: `assert!(DCS.iter().all(scs.fips()))`?

rustls/src/crypto/mod.rs
```diff
@@ -329,3 +335,41 @@ impl From<&[u8]> for SharedSecret {
         Self(source.to_vec())
     }
 }
+
+#[cfg(any(feature = "ring", feature = "fips"))]
+pub(crate) fn default_provider() -> CryptoProvider {
```
djc, line 340: So this brings back the situation where a change in Cargo features can change the crypto provider. That still seems undesirable to me?

I guess it's less bad maybe because it fails closed in the sense that if either the application or a library enables `fips`, this will always default to `fips`, but if my application is set up to use `ring` and a library forces a change to `fips` that still seems suboptimal to me.

briansmith:
The FIPS feature shouldn't select a crypto provider, since there may be multiple FIPS-capable crypto providers. The FIPS feature should just make Rustls operate in FIPS mode, and enforce that the crypto provider is working in FIPS mode.

rustls/src/crypto/mod.rs
ctz: I think it _is_ desirable, in this case. Someone asking for the `fips` feature anywhere in the dependency tree should alter the default for every caller not being more specific, because I think the goal of such a user is to eradicate all other crypto from their program.

But I think it's clear we need to reach an acceptable conclusion to #1744 and #1642.

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -200,3 +217,25 @@ mod ring_shim {
...
+/// Are we in FIPS mode?
+pub(super) fn fips() -> bool {
+    aws_lc_rs::try_fips_mode().is_ok()
+}
+
+#[cfg(test)]
+mod tests {
+    #[cfg(feature = "fips")]
+    #[test]
+    fn default_suites_are_fips() {
```
djc, line 258: Looks like in this commit, these tests would still fail?

djc (approved):

cpu:
> Two things remain, I think:
>
> macos job isn't working

It looks like the Windows job is also failing with a linker error:
>  note: LINK : fatal error LNK1181: cannot open input file 'aws_lc_fips_0_11_2_crypto.lib'

rustls/src/crypto/aws_lc_rs/mod.rs
ctz: Double oops! Fixed that too.

cpu (approved):

rustls/src/crypto/mod.rs
```diff
@@ -69,7 +69,7 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
 ///   crate.
 /// - [`crypto::aws_lc_rs::default_provider`]: (behind the `aws_lc_rs` feature,
 ///   which is optional).  This provider uses the [aws-lc-rs](https://github.com/aws/aws-lc-rs)
-///   crate.
+///   crate.  The `fips` crate feature makes this option use FIPS140-3-approved cryptography.
```
briansmith, line 72: Does this mean that the build will fail if the version of the aws-lc library being built against isn't FIPS validated for the target platform?

What happens on a target for which the build succeeds, and the version of aws-lc is FIPS-validated for some targets, but not the current target?

And, what happens when the version of aws-lc used in the build isn't FIPS-validated?

What I would expect to happen in all these cases is that the build would fail, since the user is trying to ensure that they get a fips-validated configuration.

rustls/src/lib.rs
```diff
@@ -311,6 +311,10 @@
 //!   Note that aws-lc-rs has additional build-time dependencies like cmake.
 //!   See [the documentation](https://aws.github.io/aws-lc-rs/requirements/index.html) for details.
 //!
+//! - `fips`: enable support for FIPS140-3-approved cryptography, via the aws-lc-rs crate.
+//!   This feature enables the `aws_lc_rs` feature, which makes the rustls crate depend
+//!   on [aws-lc-rs](https://github.com/aws/aws-lc-rs).
```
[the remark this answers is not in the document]
ctz: I think we would do a breaking change where the `fips` feature activates `aws-lc-rs?/fips` and `ring?/fips` features (I am assuming that will be required for the additional build- and link-time complexity required), but does not activate the `dep:aws-lc-rs` feature and does not expose the implicit-provider builder functions (eg `ClientConfig::builder()`). That means users need to set features `[ "fips", "ring" | "aws_lc_rs" ]` and use the explicit builder functions (`ClientConfig::builder_with_provider()`).

"Global features" would make this easier, I think, but it's unclear if they will ever come to fruition.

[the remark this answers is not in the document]
djc: It's not clear that we would necessarily get to do a breaking change? At least it's possible in theory that FIPS support in *ring* does not need semver-breaking changes (the addition of a `fips` Cargo feature or a `fips()` canary function certainly wouldn't be semver-breaking). In that case, enabling `fips` and `ring` for rustls would still select `aws-lc-rs` even though that would be highly surprising behavior.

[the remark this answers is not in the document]
ctz: > At least it's possible in theory that FIPS support in _ring_ does not need semver-breaking changes (the addition of a `fips` Cargo feature or a `fips()` canary function certainly wouldn't be semver-breaking)

Yeah I accept that is theoretically possible, but equally that wouldn't be operating in FIPS mode. That means we're back to "I requested the `fips` feature but got defaults that aren't FIPS-approved" situation. The other theoretical possibility is that FIPS becomes the default for *ring*, in a minor release -- that would break existing applications at runtime when rustls tries to use chacha20 (for example).

> That's at odds with Cargo's additivity guidelines but IMO it might be a better solution than this.

For me that would be a worse solution (because it requires complete and instant cooperation from everyone to turn off rustls's default features), unless we also drop `ring` from the default feature set. And, aside from the cooperation requirement, the usability around turning off a default feature is highly neglected (rust-lang/cargo/issues/11467, rust-lang/cargo/issues/3126).

The other options I see are: splitting up the crates (mentioned in #1184) or removing APIs that re-state the crate default features (opposite of #1744, but maybe #1642 is relevant).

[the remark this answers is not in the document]
djc: > Yeah I accept that is theoretically possible, but equally that wouldn't be operating in FIPS mode. That means we're back to "I requested the `fips` feature but got defaults that aren't FIPS-approved" situation. The other theoretical possibility is that FIPS becomes the default for _ring_, in a minor release -- that would break existing applications at runtime when rustls tries to use chacha20 (for example).

Okay, so since we're fairly sure that *ring* must make breaking changes to enable FIPS mode (because this removes some cryptographic algorithms) we will therefore have to make explicit changes in rustls to enable support for that version of ring, and can change our support for the `fips` flag to forward to `ring/fips` instead of `aws-lc-rs/fips`.

I can live with that.

justsmth:
> > Edit: Ah, presumably blocked on [aws/aws-lc-rs#309](https://github.com/aws/aws-lc-rs/pull/309)
> 
> MacOS Build problem: I think this is related to a fix [we recently backported to our FIPS branch](https://github.com/aws/aws-lc/pull/1400). After [this PR is merged](https://github.com/aws/aws-lc-rs/pull/313), we plan to put out a release. Hopefully, tomorrow (Jan 19).
> 
> Windows Build Problem: Correct, blocked on [aws/aws-lc-rs#309](https://github.com/aws/aws-lc-rs/pull/309)

We have now released [aws-lc-rs v1.6.0](https://crates.io/crates/aws-lc-rs/1.6.0) which [addresses the build issues](https://github.com/aws/aws-lc-rs/releases/tag/v1.6.0) mentioned above.

cpu:
> rustls / Build+test (windows-latest, stable) (pull_request) Failing after 4m

Interesting: the Windows build is still failing w/ 1.6.0 Maybe we need to crib some missing CI setup from the upstream repo?

ctz:
> > rustls / Build+test (windows-latest, stable) (pull_request) Failing after 4m
> 
> Interesting: the Windows build is still failing w/ 1.6.0 Maybe we need to crib some missing CI setup from the upstream repo?

Yes, not sure what is going on there. I checked ours against https://github.com/aws/aws-lc-rs/blob/908f6f5df610f6c85264f2721d88cd78dc177bf5/.github/workflows/fips.yml#L53-L86 and didn't see anything glaringly different. We are using `windows-latest` but that is documented as being the same as `windows-2022`. Hmm.

justsmth:
> > > rustls / Build+test (windows-latest, stable) (pull_request) Failing after 4m
> > 
> > 
> > Interesting: the Windows build is still failing w/ 1.6.0 Maybe we need to crib some missing CI setup from the upstream repo?
> 
> Yes, not sure what is going on there. I checked ours against https://github.com/aws/aws-lc-rs/blob/908f6f5df610f6c85264f2721d88cd78dc177bf5/.github/workflows/fips.yml#L53-L86 and didn't see anything glaringly different. We are using `windows-latest` but that is documented as being the same as `windows-2022`. Hmm.

I put out a [PR related to this](https://github.com/aws/aws-lc-rs/pull/327). The additional logging should help with the diagnosis.

.github/workflows/build.yml
```diff
@@ -43,11 +43,17 @@ jobs:
...
+        if: runner.os == 'Windows'
+        uses: seanmiddleditch/gha-setup-ninja@v4
+
       - name: cargo build (debug; default features)
         run: cargo build --locked
 
-      - name: cargo test (debug; all features)
-        run: cargo test --locked --all-features
+      # nb. feature sets that include "fips" should be --release --
+      # this is required for fips on windows.
```
justsmth, line 54: It's ok to perform a dev/debug build on Windows with the FIPS feature enabled. The `aws-lc-fips-sys` build-script maps a dev/debug build (i.e. `OPT_LEVEL=0`) [to a "relwithdebinfo" build for AWS-LC](https://github.com/aws/aws-lc-rs/blob/f31bee9052926ed38ef6efc71ffaf54170c9d5fb/aws-lc-fips-sys/builder/cmake_builder.rs#L97).

justsmth:
> > Update: Unfortunately, I was unable to replicate a failure like this in the aws-lc-rs CI.
> 
> I had a look at this just now and came up with [aws/aws-lc-rs#332](https://github.com/aws/aws-lc-rs/pull/332)

Thanks so much for the fix! I put up a [PR to update our CI](https://github.com/aws/aws-lc-rs/pull/333) so we more easily detect issues like this for our crates.

justsmth:
We've published [aws-lc-fips-sys 0.12.1](https://crates.io/crates/aws-lc-fips-sys) with Windows FIPS build fix.

ctz:
Have been looking now at the macos failures, and I can only really conclude that this is a cargo bug. It seems not to set `DYLD_FALLBACK_LIBRARY_PATH` to an appropriate value when running doctests?

For a normal test binary, cargo sets `DYLD_FALLBACK_LIBRARY_PATH` to a value that includes `/Users/runner/work/rustls-chaos/rustls-chaos/target/release/build/aws-lc-fips-sys-b48a6fc8fabe5170/out/build/artifacts/` (which contains `libaws_lc_fips_0_12_1_crypto.dylib`)

For the Doc-tests run, it uses:
- `/Users/runner/work/rustls-chaos/rustls-chaos/target/release/deps`
- `/Users/runner/.rustup/toolchains/stable-x86_64-apple-darwin/lib`
- `/Users/runner/.rustup/toolchains/stable-x86_64-apple-darwin/lib` (sic)
- `/Users/runner/lib`
- `/usr/local/lib`
- `/usr/lib`

That notably doesn't include the place where `libaws_lc_fips_0_12_1_crypto.dylib` actually is.

I will file a cargo bug for this (which itself is challenging because I don't have a mac, and `cargo test -vv` truncates lines. Gah!)

justsmth:
> Have been looking now at the macos failures, and I can only really conclude that this is a cargo bug.

That's my understanding as well. I believe it's related to this issue: https://github.com/rust-lang/cargo/issues/8531

Our CI disables doctests when we build aws-lc-sys (AWS-LC) as a shared library.

djc:
There's even a proposed fix PR in Cargo: https://github.com/rust-lang/cargo/pull/12710.

ctz:
OK! I think this is looking good now.

The one remaining thing is:

> * we're currently ignoring `CryptoProvider::signature_verification_algorithms` for the purpose of working out fips status; probably need to chase some of this down into pki-types and webpki. Haven't quite figured out the nicest way to do that.

I propose to, in follow-up PRs:

- in the `pki-types` crate, add `fn fips(&self) -> bool { false }` to `trait SignatureVerificationAlgorithm`, and release that as 1.2.0.
- in the `webpki` crate, implement that in `aws_lc_rs_algs.rs` precisely as in this crate; there is no need to have any extra crate features passed through.
- do the obvious to `WebPkiSupportedAlgorithms`  and then include it in the fips calculation for a `CryptoProvider`

ctz:
> The one remaining thing is:

See [4e3a26c](https://github.com/rustls/rustls/pull/1732/commits/4e3a26c2ded040123a6aed8027d04bbc45e6469e) -- push separately for review clarity, but now squashed into the early service indicator commit.

cpu (approved):

PR #1733, opened 2024-01-06
examples: simpleserver: skip argv[0] & update doc comment

bnjmnjrk:
**Skip argv[0]**:
- argv[0] contains the program name, not the cert.pem

**update doc comment**:
- revised version neither reads the environment nor checks the local directory

### DIFF
--- a/examples/src/bin/simpleserver.rs
+++ b/examples/src/bin/simpleserver.rs
@@ -2,9 +2,7 @@
 //! it accepts the default configuration, loads a server certificate and private key,
 //! and then accepts a single client connection.
 //!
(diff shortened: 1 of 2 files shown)


cpu (commented):
Oops - thanks for catching/fixing that.

Would you mind also updating the `unbuffered-server.rs` example? It was changed at the same time as this example and has the same issue: https://github.com/rustls/rustls/blob/a3cc0bcb723adc368ff24f43dce466bbb3664107/examples/src/bin/unbuffered-server.rs#L22-L28

cpu (approved):
Thanks

djc (approved):

ctz (approved):

PR #1736, opened 2024-01-08
client_conn: style tweaks for member/impl ordering

cpu:
In #1718 we'll be adding new fields to `ClientConfig`, and new methods to the `impl ClientConfig` block. Before doing that let's tidy up the existing code to match the [style guide advice on ordering](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#ordering).


djc (approved):

cpu:
Going to merge this one w/ one review since it's just shuffling around existing code without any meaningful changes.

PR #1737, opened 2024-01-08
Small HPKE trait tweaks

cpu:
Pulling out more work from #1718 that can land ahead of the ECH-specific work.

* reorder seal/open args and use less terse arg names - addresses stylistic [feedback from djc](https://github.com/rustls/rustls/pull/1718#discussion_r1441532194).
* add stateful HPKE interface - this was part of the original work introducing this trait, but we pulled it thinking we could get away with just the simpler "one-shot" interface. In practice ECH requires the stateful interface so we can reuse the HPKE sealer state between initial client hello and hello retry request processing.
* derive Copy for HpkeSymmetricCipherSuite  - small quality of life improvement
* tighten up HPKE lifetime bounds - this better aligns with the existing crypto provider traits and makes maintaining the sealer/opener state much easier.
* make EncapsulatedSecret derive Debug - small quality of life improvement, required for holding an encapsulated secret in another type that is Debug.
* clarify the hpke-rs backend - small comment adjustment to emphasize the hpke-rs example in the larger provider-example uses the Rust Crypt backend in particular.

### DIFF
--- a/provider-example/src/hpke.rs
+++ b/provider-example/src/hpke.rs
@@ -20,7 +20,7 @@ pub static HPKE_PROVIDER: &'static dyn HpkeProvider = &HpkeRsProvider {};
 struct HpkeRsProvider {}
 
 impl HpkeProvider for HpkeRsProvider {
-    fn start(&self, suite: &HpkeSuite) -> Result<Box<dyn Hpke + '_>, Error> {
+    fn start(&self, suite: &HpkeSuite) -> Result<Box<dyn Hpke + 'static>, Error> {
         Ok(Box::new(HpkeRs(hpke_rs::Hpke::new(
             hpke_rs::Mode::Base,
             KemAlgorithm::try_from(suite.kem.get_u16()).map_err(other_err)?,
@@ -72,7 +72,7 @@ impl Hpke for HpkeRs {
         &mut self,
         info: &[u8],
         pub_key: &HpkePublicKey,
-    ) -> Result<(EncapsulatedSecret, Box<dyn HpkeSealer + '_>), Error> {
+    ) -> Result<(EncapsulatedSecret, Box<dyn HpkeSealer + 'static>), Error> {
         let pk_r = hpke_rs::HpkePublicKey::new(pub_key.0.clone());
         let (enc, context) = self
             .0
@@ -112,7 +112,7 @@ impl Hpke for HpkeRs {
         enc: &EncapsulatedSecret,
         info: &[u8],
         secret_key: &HpkePrivateKey,
-    ) -> Result<Box<dyn HpkeOpener + '_>, Error> {
+    ) -> Result<Box<dyn HpkeOpener + 'static>, Error> {
         let sk_r = hpke_rs::HpkePrivateKey::new(secret_key.secret_bytes().to_vec());
         Ok(Box::new(HpkeRsReceiver {
             context: self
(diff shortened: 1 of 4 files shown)


rustls/src/crypto/hpke.rs
```diff
@@ -39,31 +39,32 @@ pub struct HpkeSuite {
...
 /// An HPKE instance that can be used for base-mode single-shot encryption and decryption.
 pub trait Hpke: Debug + Send + Sync {
-    /// Seal the provided `plaintext` to the recipient public key `pk_r` with application supplied
+    /// Seal the provided `plaintext` to the recipient public key `pub_key` with application supplied
     /// `info`, and additional data `aad`.
     ///
     /// Returns ciphertext that can be used with [Self::open] by the recipient to recover plaintext
-    /// using the same `info` and `aad` and the private key corresponding to `pk_r`.
+    /// using the same `info` and `aad` and the private key corresponding to `pub_key`. RFC 9180
+    /// refers to `pub_key` as `pk_r`.
```
djc, line 47: I thought `pk_r` was the Rusty spelling of `pkR` in the RFC? For searchability, I'd try to retain the RFC spelling in doc comments.

provider-example/src/hpke.rs
```diff
@@ -89,6 +106,46 @@ impl Hpke for HpkeRs {
...
+        enc: &EncapsulatedSecret,
+        info: &[u8],
+        secret_key: &HpkePrivateKey,
+    ) -> Result<Box<dyn HpkeOpener + '_>, Error> {
+        let sk_r = hpke_rs::HpkePrivateKey::new(secret_key.secret_bytes().to_vec());
+        let context = self
+            .0
+            .setup_receiver(enc.0.as_slice(), &sk_r, info, None, None, None)
+            .map_err(other_err)?;
+        Ok(Box::new(HpkeRsReceiver { context }))
```
djc, line 121: Nit: I'd inline the `context` binding here.

djc (approved):

rustls/src/crypto/hpke.rs
cpu: Oops, you're right. Fixed up both `sk_r` and `pk_r` refs to be `skR` and `pkR` throughout.

provider-example/src/hpke.rs
cpu: Fixed.

rustls/src/crypto/hpke.rs
```diff
@@ -39,32 +39,73 @@ pub struct HpkeSuite {
...
+    /// Seal the provided `plaintext` to the recipient public key `pub_key` with application supplied
     /// `info`, and additional data `aad`.
     ///
     /// Returns ciphertext that can be used with [Self::open] by the recipient to recover plaintext
-    /// using the same `info` and `aad` and the private key corresponding to `pk_r`.
+    /// using the same `info` and `aad` and the private key corresponding to `pub_key`. RFC 9180
+    /// refers to `pub_key` as `pkR`.
     fn seal(
         &mut self,
-        pk_r: &HpkePublicKey,
```
ctz, line 49: It doesn't matter much, but personally I would be in favour of continuing with the argument order from the RFC.

ctz (approved):

rustls/src/crypto/hpke.rs
djc: It wouldn't be my preference but I also don't feel very strongly about it.

cpu: I guess I'm the tie breaker :laughing: Like you two my preference isn't strong but I think I lightly prefer arranging the API argument order to match Rustls style vs matching the RFC in this case.

PR #1738, opened 2024-01-08
handshake: derive a few more traits 

cpu:
Pulling out more work from https://github.com/rustls/rustls/pull/1718 that can land ahead of the ECH-specific work.

* derive PartialEq on more types - this will support embedding ECH configs into a `PeerIncompatible` error we'll return when the peer rejects ECH. The ECH configs included in the error can be used for retrying a client ECH offer.
* derive Clone on a few more messages - this will support cloning messages we need to mutate (e.g. to make an "inner" client hello) as part of constructing or verifying an ECH offer.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -793,7 +793,7 @@ impl ServerExtension {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct ClientHelloPayload {
     pub client_version: ProtocolVersion,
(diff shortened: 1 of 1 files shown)


djc:
> Should I add some coverage for these traits now, or wait for them to be covered by ECH work? I have no strong pref.

I think waiting for these to be covered by ECH work is probably okay.

djc (approved):

cpu:
I think this is small/uncontroversial enough to merge with one review. If there's any post-merge feedback I will iterate.

PR #1739, opened 2024-01-08
hash_hs: support cloning (forking) HandshakeHash[Buffer]

cpu:
Pulling out more work from https://github.com/rustls/rustls/pull/1718 that can land ahead of the ECH-specific work.

For client ECH support we'll need to be able to fork (e.g. clone) the `HandshakeHashBuffer` and `HandshakeHash` types used to maintain the client transcript.

For ECH confirmation we'll fork the existing hash(buffer), add some specially encoded messages, and then use the hash state to derive a shared secret. If the secret matches an expected value we'll use the original `HandshakeHash`/`HandshakeHashBuffer`'s state from before our twiddling to continue the handshake.

Along the way, fixup more unidiomatic `get_` fns in `hash_hs.rs`, like https://github.com/rustls/rustls/pull/1726 did for the `handshake` module.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -487,7 +487,7 @@ fn emit_finished(
     transcript: &mut HandshakeHash,
     common: &mut CommonState,
 ) {
-    let vh = transcript.get_current_hash();
+    let vh = transcript.current_hash();
     let verify_data = secrets.client_verify_data(&vh);
     let verify_data_payload = Payload::new(verify_data);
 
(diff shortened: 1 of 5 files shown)


djc (approved):
I'd probably squash the first two commits here? Feels like they belong together.

I wonder if we should have a `hash::Context: Clone` bound an replace `fork()` with that?

cpu:
> I wonder if we should have a hash::Context: Clone bound an replace fork() with that?

I think this ends up being complicated for the ring-like impls. The [`start` fn returns `Box<dyn crypto::hash::Context>`](https://github.com/rustls/rustls/blob/1a2405a26ec522668c755817bcf665d848876e38/rustls/src/crypto/ring/hash.rs#L15-L17) and adding a `Clone` bound on `Context` requires `Self: Sized`, and so the trait isn't object safe anymore. I'm going to punt on this unless you think it's worth more substantial rework.

djc:
Fair, that's fine.

cpu:
@ctz Do you want to give this a once-over?

ctz (approved):

PR #1742, opened 2024-01-12
docs: add more documentation on crypto providers

cpu:
* Describe why crypto providers exist, how they are chosen
* Enumerate the built-in providers and associated feature flags
* Link to third-party providers we're aware of
* Add some documentation about implementing a custom provider

I've left out the RustCrypto provider since [the official upstream repo](https://github.com/RustCrypto/rustls-rustcrypto) is empty at the time of writing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,9 +87,8 @@ need them.
 
 ### Platform support
 
-While Rustls itself is platform independent, by default it uses
-[`ring`](https://crates.io/crates/ring) for implementing the cryptography in
-TLS. As a result, rustls only runs on platforms
+While Rustls itself is platform independent, by default it uses [`ring`] for implementing
+the cryptography in TLS. As a result, rustls only runs on platforms
 supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
 x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
 64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -87,9 +87,8 @@ need them.
 
 ### Platform support
 
-While Rustls itself is platform independent, by default it uses
-[`ring`](https://crates.io/crates/ring) for implementing the cryptography in
-TLS. As a result, rustls only runs on platforms
+While Rustls itself is platform independent, by default it uses [`ring`] for implementing
+the cryptography in TLS. As a result, rustls only runs on platforms
```
briansmith, line 91: The " We do not presently support WebAssembly" isn't clear enough. It would be good to state that *ring* does support WebAssembly, and probably it would be good to mention why Rustls doesn't.

rustls/src/lib.rs
```diff
@@ -54,9 +54,8 @@
...
 //! ### Platform support
 //!
-//! While Rustls itself is platform independent, by default it uses
-//! [`ring`](https://crates.io/crates/ring) for implementing the cryptography in
-//! TLS. As a result, rustls only runs on platforms
+//! While Rustls itself is platform independent, by default it uses [`ring`] for implementing
+//! the cryptography in TLS. As a result, rustls only runs on platforms
 //! supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
 //! x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
 //! 64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
```
briansmith, line 61: Ditto here.

README.md
```diff
@@ -87,9 +87,8 @@ need them.
...
 supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
 x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
 64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
```
briansmith, line 94: It would be better if you didn't list all these platforms because they will get out of date, and it will also be extra work to maintain this list as potentially every time we add a target architecture, we have to update this list. There is such a list where BoringSSL or AWS-LC or medTLS are mentioned.

Also, "platform" is mostly a synonym for "target" and the OS (or lack of an OS) is another limiting factor.

djc: Especially now that rustls abstracts over crypto provider, I agree that it probably doesn't make sense to list platforms/targets here explicitly.

README.md
```diff
@@ -107,7 +106,54 @@ dependency on *ring*.
 Rustls requires Rust 1.61 or later.
 
 [ring-target-platforms]: https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/include/ring-core/target.h#L18-L64
-[crypto::CryptoProvider]: https://docs.rs/rustls/latest/rustls/crypto/trait.CryptoProvider.html
+[`crypto::CryptoProvider`]: https://docs.rs/rustls/latest/rustls/crypto/trait.CryptoProvider.html
+[`ring`]: https://crates.io/crates/ring
+
+### Cryptography Providers
```
djc, line 112: These new sections use capitals on every word, which existing sections do not. I (weakly) prefer the latter style.

README.md
cpu: Please open a new PR with the wording you'd prefer to see w.r.t web assembly compatibility. This is pre-existing text that isn't being adjusted in this branch.

README.md
cpu: There have been a few different iterations on the best way to describe platform support. @briansmith you approved the PR that last adjusted this text: https://github.com/rustls/rustls/pull/1532

Since this text is pre-existing and unmodified by my changes .I would prefer to limit the scope of this branch to how it's described: adding more documentation on the crypto providers that we know exist. 

It seems like there's a variety of opinions on how best to describe the platform support so I think it merits its own PR and discussion.

rustls/src/lib.rs
cpu: See my replies on your other comment threads.

README.md
cpu: Fixed the new headings (and the pre-existing  "Design Overview" heading)

README.md
```diff
@@ -107,7 +106,259 @@ dependency on *ring*.
...
+ecosystem.
+
+See the [Making a custom CryptoProvider] section of the documentation for more information
+on this topic.
+
+[`custom-provider`]: https://github.com/rustls/rustls/tree/main/provider-example/
+[`RustCrypto`]: https://github.com/RustCrypto
+[Making a custom CryptoProvider]: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#making-a-custom-cryptoprovider
+
+## Design overview
```
djc, line 158: Ah, I think you have to adjust the sync script because it was relying on the exact capitalization of "Design Overview" maybe? Otherwise this diff doesn't quite make sense.

cpu: Oops, good catch. Thank you.

Fixed up.

djc (approved):

PR #1756, opened 2024-01-22
Update README.md - server auth

elardus-erasmus:
Correct the text per the original intent. Servers (not clients) are authenticated by clients.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ obsolete cryptography by default.
 * TLS1.3 resumption via tickets or session storage.
 * TLS1.3 0-RTT data for clients.
 * TLS1.3 0-RTT data for servers.
-* Client authentication by clients.
+* Server authentication by clients.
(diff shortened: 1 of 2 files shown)


djc (approved):
That's silly -- thanks for fixing!

ctz (commented):
This was actually intentional -- client auth is done by clients (offering, selecting and using a key and corresponding certificate) and servers (requesting/requiring and verifying the client's signature and certificate), and those two are pretty disjoint features.

The reason for the odd wording is: the first release of rustls did no client auth, the second did client auth just on the client side, and the third added server support.

But I can accept this is pretty subtle and confusing wording. Maybe we could replace both items with "Server and optional Client authentication"?

djc (approved):

ctz (approved):

ctz:
(Made the same edit to `src/lib.rs`)

PR #1765, opened 2024-01-29
Move to aws-lc-rs 1.6

ctz:


rustls/Cargo.toml
```diff
@@ -16,7 +16,7 @@ build = "build.rs"
 rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
-aws-lc-rs = { version = "1.5", optional = true }
+aws-lc-rs = { version = "1.6", optional = true }
```
djc, line 19: 1.6 is semver-compatible with 1.5 and there are no other changes in this commit that seem to require a version bump -- so why do we want this?

rustls/Cargo.toml
```diff
@@ -22,7 +22,7 @@ ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }
 webpki = { package = "rustls-webpki", version = "0.102.1", features = ["std"], default-features = false }
 pki-types = { package = "rustls-pki-types", version = "1", features = ["std"] }
-zeroize = "1.6.0"
+zeroize = "1.7"
```
djc, line 25: Also not clear why this is needed?

ctz: Unclear to me too. https://github.com/rustls/rustls/actions/runs/7694066891/job/20964074480 seems to be complaining that our ^1.6.0 is incompatible with aws-lc-rs's ^1.7.0, but:

a) it isn't
b) we asked it to only check direct dependencies only, so aws-lc-rs's dependencies should be irreverent

djc: File an issue against `cargo-minimal-versions`?

ctz: https://github.com/rust-lang/cargo/issues/5657#issuecomment-1914446399

ctz:
Have inserted this under jbp-hybrid-kx and the fips branch.

PR #1766, opened 2024-01-29
Introduce concept of "process default" provider

### LINKED ISSUE #1744: Please provide a ClientConfig function that uses an enabled CryptoProvider automatically
**Checklist**

Right now, crates that want to support aws-lc-rs have to switch over to creating a ClientConfig using `builder_with_provider`, which is a more invasive change to ask crates in the ecosystem to make. Furthermore, those crates have to manually create a provider based on what's available, which may require additional conditionals.

In addition to the existing `builder_with_provider` mechanism, I'd love to have a `builder_with_default_provider` or similar, which automatically uses one of the available providers. (Ideally I'd love for that to be just spelled `builder` for compatibility with existing code, but if there's a reason to *not* do that, then having a new method would allow crates to opt in to being flexible about providers. Perhaps in rustls 0.23 that method could become the default, so that crates that move forward will automatically be flexible about providers.)

This would allow library crates to avoid adding any `cfg` logic in their codebase, and instead just add `default-features = false` to their `rustls` dependency, allowing their users to choose. (Some library crates *may* choose to add passthrough features for user convenience, though that has upsides and downsides.) Application crates could set the appropriate feature on rustls.

Applications would still need to audit their dependencies to make sure nothing depended on rustls with default features enabled, but that's easier than *also* porting those dependencies to call different functions.

ctz:
One can be installed with `CryptoProvider::install_as_process_default`. First call wins.

The current value can be retrieved with `CryptoProvider::process_default()`.

This can be set from the crate features, if any only if they are unambigious, using `CryptoProvider::install_process_default_from_crate_features`.

Use this for `ClientConfig::builder` and `ServerConfig::builder` et al. Naturally, `ClientConfig::builder_with_provider` and co. continue to exist.

--

~~This is for feedback mainly. Theoretically this could fix #1744 and #1642~~

~~Things that make this a draft:~~

- [x] give the same treatment to `WebPkiServerVerifier::builder()`, ditto `WebPkiClientVerifier`
- [x] make `cargo test --all-features` do something other than explode
- [x] `OnceLock` would require an MSRV of 1.70; alternatively we could depend on the `once_cell` crate (its precursor)
- [x] have some basic testing

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -18,6 +18,8 @@ rustversion = { version = "1.0.6", optional = true }
 [dependencies]
 aws-lc-rs = { version = "1.6", optional = true, default-features = false, features = ["aws-lc-sys"] }
 log = { version = "0.4.4", optional = true }
+# remove once our MSRV is >= 1.70
+once_cell = "1"
 ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }
 webpki = { package = "rustls-webpki", version = "0.102.2", features = ["std"], default-features = false }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -1,8 +1,6 @@
 use crate::builder::ConfigBuilder;
 use crate::common_state::{CommonState, Protocol, Side};
 use crate::conn::{ConnectionCommon, ConnectionCore, UnbufferedConnectionCommon};
-#[cfg(any(feature = "ring", feature = "fips"))]
-use crate::crypto::default_provider;
 use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::error::Error;
@@ -16,9 +14,7 @@ use crate::suites::{ExtractedSecrets, SupportedCipherSuite};
 use crate::unbuffered::{EncryptError, TransmitTlsData};
 use crate::versions;
 use crate::KeyLog;
-#[cfg(any(feature = "ring", feature = "fips"))]
-use crate::WantsVerifier;
-use crate::{verify, WantsVersions};
+use crate::{verify, WantsVerifier, WantsVersions};
 
 use super::handy::{ClientSessionMemoryCache, NoClientSessionStorage};
 use super::hs;
@@ -222,43 +218,38 @@ pub struct ClientConfig {
 }
 
 impl ClientConfig {
-    /// Create a builder for a client configuration with the default
-    /// [`CryptoProvider`].
-    ///
-    /// This is:
-    ///
-    /// - [`crypto::aws_lc_rs::default_provider`] if the `fips` crate feature is
-    ///   enabled.
-    /// - [`crypto::ring::default_provider`] if the `ring` crate feature is
-    ///   enabled and the `fips` crate feature is not enabled.
-    ///
-    /// If neither of these are true, this function is not available and you
-    /// must use [`ClientConfig::builder_with_provider()`] instead.
+    /// Create a builder for a client configuration with the process-default
+    /// [`CryptoProvider`]: [`CryptoProvider::get_default`] and safe
+    /// protocol version defaults.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
-    #[cfg(any(feature = "ring", feature = "fips"))]
     pub fn builder() -> ConfigBuilder<Self, WantsVerifier> {
-        // Safety: we know the *ring* and aws-lc-rs providers' ciphersuites are compatible with the safe default protocol versions.
-        Self::builder_with_provider(default_provider().into())
-            .with_safe_default_protocol_versions()
-            .unwrap()
+        Self::builder_with_protocol_versions(versions::DEFAULT_VERSIONS)
     }
 
-    /// Create a builder for a client configuration with the default
-    /// [`CryptoProvider`] (see [`ClientConfig::builder()`] for details), safe
-    /// ciphersuite defaults and the provided protocol versions.
+    /// Create a builder for a client configuration with the process-default
+    /// [`CryptoProvider`]: [`CryptoProvider::get_default`] and
+    /// the provided protocol versions.
     ///
-    /// Panics if provided an empty slice of supported versions.
+    /// Panics if
+    /// - the supported versions are not compatible with the provider (eg.
+    ///   the combination of ciphersuites supported by the provider and supported
+    ///   versions lead to zero cipher suites being usable),
+    /// - if a `CryptoProvider` cannot be resolved using a combination of
+    ///   the crate features and process default.
     ///
     /// For more information, see the [`ConfigBuilder`] documentation.
-    #[cfg(any(feature = "ring", feature = "fips"))]
     pub fn builder_with_protocol_versions(
         versions: &[&'static versions::SupportedProtocolVersion],
     ) -> ConfigBuilder<Self, WantsVerifier> {
-        // Safety: we know the *ring* and aws-lc-rs providers' ciphersuites are compatible with all protocol version choices.
-        Self::builder_with_provider(default_provider().into())
-            .with_protocol_versions(versions)
-            .unwrap()
+        // Safety assumptions:
+        // 1. that the provider has been installed (explicitly or implicitly)
+        // 2. that the process-level default provider is usable with the supplied protocol versions.
+        Self::builder_with_provider(Arc::clone(
+            CryptoProvider::get_default_or_install_from_crate_features(),
+        ))
+        .with_protocol_versions(versions)
+        .unwrap()
     }
 
     /// Create a builder for a client configuration with a specific [`CryptoProvider`].
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -7,6 +7,7 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 
+use once_cell::sync::OnceCell;
 use pki_types::PrivateKeyDer;
 use zeroize::Zeroize;
 
@@ -74,6 +75,25 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
 ///
 /// This structure provides defaults. Everything in it can be overridden at
 /// runtime by replacing field values as needed.
+///
+/// # Using the per-process default `CryptoProvider`
+///
+/// There is the concept of an implicit default provider, configured at run-time once in
+/// a given process.
+///
+/// It is used for functions like [`ClientConfig::builder()`] and [`ServerConfig::builder()`].
+///
+/// The intention is that an application can specify the [`CryptoProvider`] they wish to use
+/// once, and have that apply to the variety of places where their application does TLS
+/// (which may be wrapped inside other libraries).
+/// They should do this by calling [`CryptoProvider::install_default()`] early on.
+///
+/// To achieve this goal:
+///
+/// - _libraries_ should use [`ClientConfig::builder()`]/[`ServerConfig::builder()`]
+///   or otherwise rely on the [`CryptoProvider::get_default()`] provider.
+/// - _applications_ should call [`CryptoProvider::install_default()`] early
+///   in their `fn main()`.
 ///
 /// # Using a specific `CryptoProvider`
 ///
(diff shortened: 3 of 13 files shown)


djc:
I think an approach like this would be a good direction.

One thought (not sure how much value there is): we could yield an `Installed` newtype token from an install method, which libraries could then take?

joshtriplett:
This looks great.

With this changed, it'll be easy for application crates to set a default to be used, and libraries won't have to have any `cfg`s in their Rust code to use that default.

what would you recommend that library crates do in their Cargo manifests, to balance between "works without doing anything special" and "not forcing a dependency people don't want"? If library crates depend on rustls with `default-features = false`, then they won't work at *all* unless something in the tree makes this call or sets the feature flag in rustls. If library crates depend on rustls with `default-features = true`, they'll work out of the box, but force a dependency on one or the other crypto library even if that's not the one you want to use. And if library crates want to work out of the box *without* forcing such a dependency, they're still in the position of exposing feature flags, though at least those feature flags can just be passthroughs to rustls.

djc:
> what would you recommend that library crates do in their Cargo manifests, to balance between "works without doing anything special" and "not forcing a dependency people don't want"?

Good question? If we do this I feel like the "use existing process default provider" API should be enabled by default and none of the crypto providers should be provided by default. Maybe that's where it starts to make sense to move the crypto providers that we maintain to their own crate?

ctz:
> > what would you recommend that library crates do in their Cargo manifests, to balance between "works without doing anything special" and "not forcing a dependency people don't want"?
>
> Maybe that's where it starts to make sense to move the crypto providers that we maintain to their own crate?

Yeah I think this is the next step on. As things currently stand it is possible to get _three_ different underlying crypto libraries via cargo features: ring, aws-lc-sys, aws-lc-fips-sys. That is not a great situation for build-time efficiency, even if we can arrange that the linker throws some or all of those away.

The arrangement I had in mind was splitting all this into several crates (naming up for grabs):

- `rustls`: the crate as it currently stands, but without any `CryptoProvider`s or dependencies on any crypto crates. Intermediate libraries would depend on this.
- `rustls-ring`: the `CryptoProvider` backed by *ring*.
- `rustls-aws-lc-rs`: the `CryptoProvider` backed by aws-lc-rs.
- `rustls-default`: an "easy mode" anthology crate, which depends on and reexports `rustls` and exactly one provider crate that we recommend.

I'm not 100% sure about the `rustls-default` one in that list; I think I worry quite a bit about adding essential complexity to the first steps someone takes, and forcing them to make decisions on things which are somewhat inconsequential at that point.

rustls/src/client/client_conn.rs
```diff
@@ -219,34 +217,38 @@ pub struct ClientConfig {
...
+    /// [`CryptoProvider`]: [`CryptoProvider::process_default`] and
     /// the provided protocol versions.
     ///
-    /// Panics if provided an empty slice of supported versions.
+    /// Panics if
+    /// - the supported versions are not compatible with the provider (eg.
+    ///   the combination of ciphersuites supported by the provider and supported
+    ///   versions lead to zero cipher suites being usable),
+    /// - if a `CryptoProvider` cannot be resolved a combination of
+    ///   the crate features and process default.
```
cpu, line 238: I think a word might have been dropped here. Maybe:

```suggestion
    /// - if a `CryptoProvider` cannot be resolved using a combination of
    ///   the crate features and process default.
```

rustls/src/crypto/mod.rs
```diff
@@ -74,6 +75,25 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
...
+/// # Using the per-process default `CryptoProvider`
+///
+/// There is the concept of an implicit default provider, configured at run-time once in
+/// a given process.
+///
+/// It is used for functions like [`ClientConfig::builder()`] and [`ServerConfig::builder()`].
+///
+/// The intention is that an application can specify the [`CryptoProvider`] they wish to use
+/// once, and have that apply to the variety of places in their application does TLS
+/// (which may be wrapped inside other libraries).
```
cpu, line 87: another small readability tweak:
```suggestion
/// The intention is that an application can specify the [`CryptoProvider`] they wish to use
/// once, and have that apply to the variety of places where their application does TLS
/// (which may be wrapped inside other libraries).
```

rustls/src/server/server_conn.rs
```diff
@@ -356,34 +353,38 @@ impl Clone for ServerConfig {
...
+    /// [`CryptoProvider`]: [`CryptoProvider::process_default`] and
     /// the provided protocol versions.
     ///
-    /// Panics if provided an empty slice of supported versions.
+    /// Panics if
+    /// - the supported versions are not compatible with the provider (eg.
+    ///   the combination of ciphersuites supported by the provider and supported
+    ///   versions lead to zero cipher suites being usable),
+    /// - if a `CryptoProvider` cannot be resolved a combination of
+    ///   the crate features and process default.
```
cpu, line 374: same feedback here RE: dropped word.

cpu (commented):
This seems like the right direction given all of the options available. I found a couple small nits reviewing the doc strings in the diff:

joshtriplett:
> > > what would you recommend that library crates do in their Cargo manifests, to balance between "works without doing anything special" and "not forcing a dependency people don't want"?
> > 
> > 
> > Maybe that's where it starts to make sense to move the crypto providers that we maintain to their own crate?
> 
> Yeah I think this is the next step on. As things currently stand it is possible to get _three_ different underlying crypto libraries via cargo features: ring, aws-lc-sys, aws-lc-fips-sys. That is not a great situation for build-time efficiency, even if we can arrange that the linker throws some or all of those away.
> 
> The arrangement I had in mind was splitting all this into several crates (naming up for grabs):
> 
>     * `rustls`: the crate as it currently stands, but without any `CryptoProvider`s or dependencies on any crypto crates. Intermediate libraries would depend on this.
> 
>     * `rustls-ring`: the `CryptoProvider` backed by _ring_.
> 
>     * `rustls-aws-lc-rs`: the `CryptoProvider` backed by aws-lc-rs.
> 
>     * `rustls-default`: an "easy mode" anthology crate, which depends on and reexports `rustls` and exactly one provider crate that we recommend.
> 
> 
> I'm not 100% sure about the `rustls-default` one in that list; I think I worry quite a bit about adding essential complexity to the first steps someone takes, and forcing them to make decisions on things which are somewhat inconsequential at that point.

I think the best way to solve the essential complexity here might be to handle the last step with cargo key-value features that have defaults. If we could say "there's a global feature, the crypto provider, that has a default value if not set, but can be set to any of these values by anything in the stack, with higher points in the stack overriding lower ones", that *together* with this PR's mechanism should fully solve the problem.

djc:
Having a rustls-default crate doesn't seem like an attractive option to me. I think I agree that Cargo features might be a better alternative to solving that issue.

rustls/Cargo.toml
```diff
@@ -18,6 +18,7 @@ rustversion = { version = "1.0.6", optional = true }
 [dependencies]
 aws-lc-rs = { version = "1.6", optional = true, default-features = false, features = ["aws-lc-sys"] }
 log = { version = "0.4.4", optional = true }
+once_cell = { version = "1" }
```
djc, line 21: Style: just `once_cell = "1"` would suffice.

Maybe we should make a note somewhere that, once we bump the MSRV to 1.70, we could consider using `std::sync::OnceLock` instead of this extra dependency.

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +206,63 @@ pub struct CryptoProvider {
...
 
+impl CryptoProvider {
+    /// Sets the supplied `CryptoProvider` as the default for this process.
+    ///
+    /// This can be called successfully at most once in any process execution.
+    ///
+    /// Call this early in your process to configure which provider is used for
+    /// the provider.  The configuration should happen before any use of
+    /// [`ClientConfig::builder()`] or [`ServerConfig::builder()`].
+    pub fn install_as_process_default(provider: impl Into<Arc<Self>>) -> Result<(), Arc<Self>> {
```
djc, line 217: I wonder if tying the process default APIs into the `CryptoProvider` value type is confusing? I don't find it all that intuitive because `install_as_process_default()` might as well take `self` (rather than `provider`).

Alternatively, we could have a separate unit type `DefaultProvider` with `set()` and `get()` methods, which feels like it would be more obvious. Or make `DEFAULT_PROVIDER` a `pub static` (potentially leaking once_cell into the public API, which may or may not be a big deal).

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +206,63 @@ pub struct CryptoProvider {
...
+        #[cfg(all(feature = "aws_lc_rs", not(feature = "ring")))]
+        {
+            let _ = Self::install_as_process_default(aws_lc_rs::default_provider());
+        }
+    }
+
+    /// Returns the default `CryptoProvider` for this process.
+    ///
+    /// This will be `None` if no default has been set yet.
+    pub fn process_default() -> Option<Arc<Self>> {
```
djc, line 242: IMO it would be nicer to yield `&Arc<Self>` here. `cloned()` is cheap but not free.

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +206,63 @@ pub struct CryptoProvider {
...
+    pub fn install_as_process_default(provider: impl Into<Arc<Self>>) -> Result<(), Arc<Self>> {
+        PROCESS_DEFAULT_PROVIDER.set(provider.into())
+    }
+
+    /// Sets the default provider to the one named unambiguously by rustls crate features.
+    ///
+    /// This function does nothing if the crate features are ambiguous (ie, specify two
+    /// providers), or specify no providers.  In both cases the application should
+    /// explicitly specify the provider to use with [`CryptoProvider::install_as_process_default`].
+    pub fn install_process_default_from_crate_features() {
```
djc, line 226: Personally I find `install_as_process_default()` pretty wordy already, but this is really out there in terms of number of tokens. Is there no more compact version of this that would convey enough context? `set_default_from_feature()`? (Is it actually multiple features or rather a single feature?)

ctz: I've split this up into `CryptoProvider::from_crate_features()`, stopped it doing the installing too, and made it private.

rustls/src/crypto/mod.rs
ctz: Actually, just `self` works fine.

rustls/Cargo.toml
ctz: Added a comment next to this line about the MSRV.

rustls/src/crypto/mod.rs
```diff
@@ -186,6 +206,66 @@ pub struct CryptoProvider {
...
+        PROCESS_DEFAULT_PROVIDER.get()
+    }
+
+    /// An internal function that:
+    ///
+    /// - gets the process_default(), or
+    /// - installs one `from_crate_features()`, or else
+    /// - panics about the need to call [`CryptoProvider::install_as_process_default()`]
+    pub(crate) fn process_default_or_default_from_crate_features() -> &'static Arc<Self> {
+        match Self::process_default() {
```
djc, line 234: Given that the `None` branch is a little involved, maybe consider writing this as

```rust
if let Some(provider) = Self::process_default() {
    return provider;
}

..
```

?

rustls/src/crypto/mod.rs
djc: That makes the public interface much better! I'd probably still prefer a more concise name even for internal API, but obviously that's much less important.

rustls/src/crypto/mod.rs
djc: This is much more obvious! I still find `install_as_process_default()` to be pretty verbose and would prefer a more concise spelling -- I don't think all of these tokens are necessary to make the meaning obvious.

djc (approved):

rustls/src/crypto/mod.rs
ctz: Yeah that's nicer!

ctz:
> This is much more obvious! I still find install_as_process_default() to be pretty verbose and would prefer a more concise spelling -- I don't think all of these tokens are necessary to make the meaning obvious.

Any suggestions? I like the "process default" concept (though "process" also being an active verb is not ideal). `set_process_default()`? Not sure about that, because I think "install" is more evocative of how permanent it is.

djc:
I feel like the `process` part is kind of obvious if you think about it? So I would consider dropping that. That leaves `install_as_default()`, but I think `install_default()` is probably enough.

stormshield-gt:
The nightly job failure seems to come from `curve25519-dalek ` used in the `rustls-provider-example`  which has [a current PR to solve it ](https://github.com/dalek-cryptography/curve25519-dalek/pull/619)

ctz:
> I feel like the `process` part is kind of obvious if you think about it? So I would consider dropping that. That leaves `install_as_default()`, but I think `install_default()` is probably enough.

I think if we drop `process`, then that makes the getter (currently `CryptoProvider::process_default()`) just `CryptoProvider::default()`, which I think is confusingly similar to the `Default` trait fn of the same name?

djc:
Maybe `get_default()` for the getter?

ctz:
Will try that and see how it looks!

PR #1768, opened 2024-01-30
Depend on `aws-lc-rs` with `default-features = false`

joshtriplett:
rustls does not seem to need the ring-io or ring-sig-verify features,
and omitting them avoids a dependency on an old version of the
`untrusted` crate.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -16,7 +16,7 @@ build = "build.rs"
 rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
-aws-lc-rs = { version = "1.6", optional = true }
+aws-lc-rs = { version = "1.6", optional = true, default-features = false, features = ["aws-lc-sys"] }
 log = { version = "0.4.4", optional = true }
 ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }


djc (approved):
Nice!

cpu:
> rustls does not seem to need the ring-io or ring-cert-verify features

That's curious. I would have thought we did, so I'm trying to convince myself why this isn't the case :thinking: It seems like we might have been using `io::der` under the `aws-lc-rs` provider up until ~this morning when c1c917cfbecdf5c3d4035d852c198b0308b11899 landed, so perhaps this is just very well timed?

> the ring-io or ring-cert-verify features

Minor note: For folks trying to dig into the changeset, I think this latter feature is ["ring-sig-verify"](https://github.com/aws/aws-lc-rs/blob/067c9486eb117c67a9da977256f1a99bac11bfae/aws-lc-rs/README.md?plain=1#L44) not "ring-cert-verify"

ctz (approved):

cpu (approved):
Thanks! I've convinced myself :-) 

FWIW It looks like the same change is appropriate for webpki: https://github.com/rustls/webpki/pull/224

joshtriplett:
> Minor note: For folks trying to dig into the changeset, I think this latter feature is ["ring-sig-verify"](https://github.com/aws/aws-lc-rs/blob/067c9486eb117c67a9da977256f1a99bac11bfae/aws-lc-rs/README.md?plain=1#L44) not "ring-cert-verify"

Arg, good catch. Fixed and force-pushed.

PR #1772, opened 2024-02-05
Default to `require_ems` in FIPS mode

ctz:
Change default for `require_ems` based on `fips` crate feature, generalising the existing tests for `require_ems` to verify this too.

Include `require_ems` in `fips()` determination.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -198,7 +198,8 @@ pub struct ClientConfig {
     /// If set to `true`, requires the server to support the extended
     /// master secret extraction method defined in [RFC 7627].
     ///
-    /// The default is `false`.
+    /// The default is `true` if the `fips` crate feature is enabled,
+    /// `false` otherwise.
     ///
     /// It must be set to `true` to meet FIPS requirement mentioned in section
     /// **D.Q Transition of the TLS 1.2 KDF to Support the Extended Master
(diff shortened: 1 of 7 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -280,7 +281,15 @@ impl ClientConfig {
     /// Return true if connections made with this `ClientConfig` will
     /// operate in FIPS mode.
     pub fn fips(&self) -> bool {
-        self.provider.fips()
+        #[cfg(feature = "tls12")]
```
djc, line 284: If we're going to make the config-side `fips()` depend on the config, should we make the `CryptoProvider::fips()` `pub(crate)` to make it harder to accidentally rely on the `CryptoProvider::fips()` where that is not sufficient?

ctz: I think -- I hope -- that there is not much confusion in semantics between `CryptoProvider::fips()` and `Client/ServerConfig::fips()`, even if they were functionally the same up until this commit. Will extend the docs though.

cpu (approved):

djc (approved):

rustls/src/client/client_conn.rs
djc: Maybe the distinction needs to be explained in `CryptoProvider::fips()`? That's where it's risky, if you use that `fips()` but fail to check the config.

cpu: That seems like a good idea to me

ctz: I have added some clarifying text to `CryptoProvider::fips()` -- please check.

cpu: LGTM

djc (approved):

PR #1773, opened 2024-02-06
github actions: standardise on ubuntu-latest

ctz:
Prior to this we had a mix of ubuntu-20.04 and ubuntu-latest. I'd like to have just one, so I can use `act` locally to do test builds without having multiple docker images.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
           - stable
           - beta
           - nightly
-        os: [ubuntu-20.04]
+        os: [ubuntu-latest]
         # but only stable on macos/windows (slower platforms)
         include:
           - os: macos-latest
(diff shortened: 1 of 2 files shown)


djc (approved):

cpu (approved):
>  I'd like to have just one, so I can use act locally to do test builds without having multiple docker images.

Ah neat, I hadn't heard of this tool before: https://github.com/nektos/act

ctz:
> > I'd like to have just one, so I can use act locally to do test builds without having multiple docker images.
> 
> Ah neat, I hadn't heard of this tool before: https://github.com/nektos/act

Yeah this mostly works for me -- you need to opt into the "full" images (which are 40-50GB, because they contain every version of every language runtime available). Though our "Check cross compilation targets" job because it's trying to use docker nested inside a docker container. But the remainder work.

ctz:
Have updated the branch protection rules for this.

PR #1774, opened 2024-02-07
Avoid unnecessary allocation and copy in `OpaqueMessage::encode`

Wonshtrum:
In https://github.com/rustls/rustls/pull/1640#issuecomment-1832220732 I mentioned:
> `OpaqueMessage::encode` only appends the 5-byte "TLS header". Allocating and copying an entire payload (up to 16Kb) seems wasteful and could be completely avoided if the payload built by `MessageEncrypted::encrypt` accounted for the 5-byte prelude. `OpaqueMessage::encode` would only have to fill in the reserved space.

In this PR the type of `OpaqueMessage::payload` is changed from a `Vec<u8>` to a `PrefixedPayload`. This new type automatically accounts for the prelude, avoiding the unneeded allocation and copy. The only downside I found is that `OpaqueMessage::into_plain_message` which didn't copy now has to (to remove the prelude). But this function only seems to be used for tests.


ctz:
See also https://github.com/rustls/rustls/commit/22f083e404c769efd4536097c5c7933a7f2016bd and https://github.com/rustls/rustls/commit/0cca8222ed25d42c3af7f1c1d907428f8189d7f5 which is attacking the same problem, albeit quite a while ago now. I won't rebase that branch (I was mainly interested in how much performance gain we could expect), but maybe how I did it can be an inspiration.

Wonshtrum:
I like your approach better!

Talking about specialized types for a specific "direction", I think there are a lot of messages and payload-related types and it's not always clear what are their requirements and what they can be used for. Most of them seem to stem from the different requirements between outgoing and ingoing flow, but it is not clearly stated and some types are used for both. #1640 explicitly splits the `BorrowedPlainMessage` type to better suit some outbound requirements, this PR and your commits also specialize a type for outbound use. I wonder if it would not be worth recentering the data model on "flow direction" rather than ownership. Here are my observations and renaming/refactor proposition:

- `OpaqueMessage` is for outbound, it holds owned encrypted data. It owns it because encrypters take the borrowed payload of `BorrowedPlainMessage`, clone it in a vec (PR 1640 leverage this fact to merge the chunks of `write_vectored`) together with all necessary information (typ, nonce... depending on the `MessageEncrypter`) before encrypting it in place. May be renamed `OpaqueMessageOutbound`.
- `OpaqueMessage::payload` is a `Payload<'static>` which could be replaced by a `OpaquePayloadOutbound` (which would be a Vec<u8> under the hood like the `PrefixedPayload` of this PR). Alternatively `OpaqueMessage`, now `OpaqueMessageOutbound` could work by itself without payload like you did in your commits.
- `BorrowedOpaqueMessage` is for inbound, it holds encrypted data borrowed from a `Deframer`. May be renamed `OpaqueMessageInbound`. The `BorrowedOpaqueMessage::into_plain_message` would return unambiguously a `BorrowedPlainMessageInbound` (type introduced by PR 1640).
- `BorrowedOpaqueMessage::payload` is a `BorrowedPayload` which could be replaced by a `OpaquePayloadInbound` (which would be a `&mut [u8]` under the hood). The payload must be mutable as it may be decrypted in place.
- `BorrowedPlainMessage` is split by PR 1640 in `BorrowedPlainMessageInbound` and `BorrowedPlainMessageOutbound`.
- `BorrowedPlainMessageInbound::payload` is a `&[u8]` which could be replaced by a `PlainPayloadInbound`. It doesn't need to be mutable as it is an already decrypted inbound type.
- `BorrowedPlainMessageOutbound::playload` is a `BorrowedPlainPayload` which could be renamed `PlainPlayloadOutbound`. It doesn't have ownership over its data and can be fragmented to reflect user's needs. It doesn't need to be mutable as encrypters will clone its data as said before for `OpaqueMessage`.
- `PlainMessage` seems redundant. It is used both for inbound and outbound types and can hold owned and borrowed data. From what I see all its occurences can be replaced by a more explicit `BorrowedPlainMessageInbound` or `BorrowedPlainMessageOutbound`. An owned variant doesn't seem needed. If this type disappears, the Borrowed prefix may be dropped for its 2 "split types".
- `Message` is both for inbound and outbound and can hold owned as well as borrowed data. I don't think splitting it would be worth it, as it is a big enum containing itself types that would need to be split (mainly `MessagePayload` and `Payload`). The only additional benefit we could get is separate client-only and server-only messages (I don't know if that would be the case, I still don't know much about the TLS protocol). If we want to remain explicit in the usage of messages, we may create a `MessageInbound(Message)` and `MessageOutbound(Message)` which implement `Deref<Message>`. `MessageInbound` would additionally implement `TryFrom<BorrowedPlainMessageInbound>`. An equivalent for `MessageOutbound` would not exist as the data of `BorrowedPlainMessageOutbound` could be fragmented.

In summary, all messages and payload types would boil down to:
- `OpaqueMessageOutbound` with `OpaquePayloadOutbound`: owns encrypted outbound data (and TLS prelude if we want to avoid a copy)
- `(Borrowed)OpaqueMessageInbound` with `(Borrowed)OpaquePayloadInbound`: mutably borrows encrypted inbound data
- `(Borrowed)PlainMessageOutbound` with `(Borrowed)PlainPayloadOutbound`: borrows potentially fragmented outbound user data
- `(Borrowed)PlainMessageInbound` with `(Borrowed)PlainPayloadInbound`: borrows decrypted inbound data
- `MessageOutbound` with `Message/MessagePayload/Payload`: owns or borrows decoded outbound
- `MessageInbound` with `Message/MessagePayload/Payload`: owns or borrows decoded inbound

I quite like the "explicitness" of this system and the cartesian product of:
`(Opaque|Plain)(Message|Payload)(Inbound|Outbound)`
It also clearly shows the ownership and mutability requirements of each stages and highlights the (for now) needed copy between `PlainMessageOutbound` and `OpaquePayloadOutbound`.

djc:
Might make sense to keep the naming discussion in https://github.com/rustls/rustls/pull/1640#discussion_r1481197489 for now?

provider-example/src/aead.rs
```diff
@@ -4,21 +4,25 @@ use alloc::vec::Vec;
...
+    UnsupportedOperationError, NONCE_LEN,
 };
+use rustls::internal::msgs::codec::Codec;
 use rustls::{ConnectionTrafficSecrets, ContentType, ProtocolVersion};
 
 pub struct Chacha20Poly1305;
 
-impl cipher::Tls13AeadAlgorithm for Chacha20Poly1305 {
-    fn encrypter(&self, key: cipher::AeadKey, iv: cipher::Iv) -> Box<dyn cipher::MessageEncrypter> {
+impl Tls13AeadAlgorithm for Chacha20Poly1305 {
```
djc, line 17: I think it's a good idea to remove the type qualifiers here, but I would prefer to see that in a separate commit so that this one can focus on the actual logical changes.

provider-example/src/aead.rs
```diff
@@ -187,21 +175,45 @@ impl cipher::MessageDecrypter for Tls12Cipher {
...
+
+    fn truncate(&mut self, len: usize) {
+        (&mut self.0 as &mut Vec<u8> as &mut dyn Buffer).truncate(len)
+    }
+}
+
+struct DecryptBufferAdapter<'a, 'p>(&'a mut BorrowedPayload<'p>);
 
-impl AsRef<[u8]> for BufferAdapter<'_, '_> {
+impl AsRef<[u8]> for DecryptBufferAdapter<'_, '_> {
```
djc, line 204: Similarly, I think it would be better to leave `BufferAdapter` with the same name in the commit that adds `EncryptBufferAdapter`, and rename it in the commit after that.

rustls/src/msgs/macros.rs
```diff
@@ -1,73 +1,35 @@
...
-        }
-        impl From<u8> for $enum_name {
-            fn from(x: u8) -> Self {
-                match x {
-                    $($enum_val => $enum_name::$enum_var),*
-                    , x => $enum_name::Unknown(x),
-                }
-            }
-        }
+    ($(#[$comment:meta])* @U8 $($enum:tt)+) => {
```
djc, line 3: This also seems unrelated/should be in a separate commit?

rustls/src/msgs/message.rs
```diff
@@ -180,10 +176,66 @@ impl OpaqueMessage {
...
+    }
+}
+
+impl<'a> Extend<&'a u8> for PrefixedPayload {
+    fn extend<T: IntoIterator<Item = &'a u8>>(&mut self, iter: T) {
+        self.0.extend(iter)
+    }
+}
+
+impl Deref for PrefixedPayload {
```
djc, line 214: Seems surprising that `Deref` yields different data from `AsRef`. Why do we need this `impl`?

rustls/src/msgs/message.rs
```diff
@@ -180,10 +176,66 @@ impl OpaqueMessage {
...
+
+impl From<&[u8]> for PrefixedPayload {
+    fn from(content: &[u8]) -> Self {
+        Self([&[0u8; OpaqueMessage::HEADER_SIZE], content].concat())
+    }
+}
+
+impl<const N: usize> From<&[u8; N]> for PrefixedPayload {
+    fn from(content: &[u8; N]) -> Self {
+        let content: &[u8] = content;
```
djc, line 236: Nit: inline the binding?

rustls/src/msgs/message.rs
```diff
@@ -180,10 +176,66 @@ impl OpaqueMessage {
...
+
+impl DerefMut for PrefixedPayload {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl From<&[u8]> for PrefixedPayload {
+    fn from(content: &[u8]) -> Self {
+        Self([&[0u8; OpaqueMessage::HEADER_SIZE], content].concat())
```
djc, line 230: Is `concat()` optimal? I'd probably write this with an explicit `Vec::with_capacity()`.

rustls/src/msgs/message.rs
```diff
@@ -180,10 +176,66 @@ impl OpaqueMessage {
     const MAX_PAYLOAD: u16 = 16_384 + 2048;
 
     /// Content type, version and size.
-    const HEADER_SIZE: u16 = 1 + 2 + 2;
+    const HEADER_SIZE: usize = 1 + 2 + 2;
```
djc, line 179: I think this should be in a separate commit.

Wonshtrum:
I tried to apply some of my renaming propositions on a [branch](https://github.com/Wonshtrum/rustls/tree/message-types-rename) of my repo.
What do you think about it? I really like how ownership becomes an implementation detail of the payloads and how choosing a message variant becomes trivial as it boils down to: is it inbound? is it opaque?
As previously said, I think we can go even further and remove the `PlainMessage` struct and only keep `InboundPlainMessage`/`OutboundPlainMessage` as this intermediary owned type is not necessary, and I much prefer the explicitness of the two later types.
I don't want to force this change as ultimately it is only a style choice. But in case you find it interesting I could merge it here, or create a separate PR.

rustls/src/msgs/message.rs
Wonshtrum: Yes, it's a bit awkward. The reason is in most places I want to be able to access "after prefix buffer" (with `AsRef`), but in some cases, I have to access the container (to extend the vec for example).
I like the solution of @ctz which requires less impls, I will take some inspiration from it.

djc: Not entirely sure what solution you're referring to. I think if this potential for confusion exists (prefix-less vs full backing store access) I think we should avoid `AsRef` and `Deref` equally and instead use explicitly named methods.

Wonshtrum: `AsMut` and `Extend` have to be implemented because `PrefixedPayload` is passed to `seal_in_place_append_tag`.
I implemented `Deref` because we need to get the underlying container for those 3 operations used in multiple places by the `encrypt` methods:
```rs
payload.extend_from_slice(&nonce.as_ref()[4..]);
msg.payload.copy_to_vec(&mut payload);
msg.typ.encode(&mut payload);
```
It would be great to not expose the vector at all if possible. For the first operation, we can simply implement `extend_from_slice` on `PrefixedPayload`. For the second one, we can implement an `extend_from_chunks` method on `PrefixedPayload` which takes an `OutboundChunks` in. It avoids accessing the vector and I like the syntax better (as it resembles `copy_from_slice`). The last one is a bit more problematic. I would suggest modifying the `Codec` trait from:
```rs
pub trait Codec<'a>: Debug + Sized {
    fn encode(&self, bytes: &mut Vec<u8>);
    fn read(_: &mut Reader<'a>) -> Result<Self, InvalidMessage>;
    fn get_encoding(&self) -> Vec<u8> {
        let mut bytes = Vec::new();
        self.encode(&mut bytes);
        bytes
    }
    fn read_bytes(bytes: &'a [u8]) -> Result<Self, InvalidMessage> {
        let mut reader = Reader::init(bytes);
        Self::read(&mut reader)
    }
}
```
to:
```rs
trait Codec2<'a>: Debug + Sized {
    fn encode(&self, bytes: &mut Vec<u8>) {
        bytes.extend_from_slice(self.as_slice())
    }
    fn get_encoding(&self) -> Vec<u8> {
        self.as_slice().to_vec()
    }
    fn read_bytes(bytes: &'a [u8]) -> Result<Self, InvalidMessage> {
        let mut reader = Reader::init(bytes);
        Self::read(&mut reader)
    }
    fn as_slice(&self) -> &[u8];
    fn read(_: &mut Reader<'a>) -> Result<Self, InvalidMessage>;
}
```
In the end the 3 operations I listed would be rewritten:
```rs
payload.extend_from_slice(&nonce.as_ref()[4..]);
payload.extend_from_chunks(&msg.payload);
payload.extend_from_slice(msg.typ.as_slice());
```

Wonshtrum: I just realized `as_slice` wouldn't work as it would return a reference to temporary data.

Wonshtrum: Or instead of modifying `Codec` we could modify the macro `enum_builder` to generate a `get_slice` method (in addition of `get_u8`/`get_u16`).
So the code would look like:
```rs
payload.extend_from_slice(&msg.type.get_slice());
```
Another way would be to modify `Codec` to allow anything that implements `Extend<u8>` inside `encode` (not just `Vec<u8>`). But this is more involved and I don't know if the compiler is as efficient with `extend` as it can be with `extend_from_slice` or `push` and other vector-specific methods.
Or we could just add a `get_container` method on `PrefixedPayload` with a warning saying it contains the prefix as well as the body.

Wonshtrum: I implemented the modified `enum_builder` on this branch: https://github.com/Wonshtrum/rustls/tree/message-types-rename.

cpu: > we could modify the macro enum_builder to generate a get_slice method (in addition of get_u8/get_u16).

For what it's worth, I like that idea best of the set provided above. :+1:

cpu:
> I tried to apply some of my renaming propositions on a [branch](https://github.com/Wonshtrum/rustls/tree/message-types-rename) of my repo.
> ..
> I implemented the modified enum_builder on this branch: https://github.com/Wonshtrum/rustls/tree/message-types-rename.
> ..
> But in case you find it interesting I could merge it here, or create a separate PR.

Personally I think it would be easiest to roll this content into this PR and review it all at once. There's already review feedback left from Djc about breaking up the single commit in this branch. I think the changes to the enum builder, and the stylistic changes proposed to implement the renamings you mention could also be done step-wise in individual commits.

WDYT?

rustls/src/msgs/message.rs
djc: > I implemented `Deref` because we need to get the underlying container for those 3 operations used in multiple places by the `encrypt` methods:

Note to clarify, I'm not saying we should not allow access to the underlying container, merely that we shouldn't use the `Deref` trait implementation for it if it might be ambiguous if `Deref` yields the "full" container or just the payload part.

cpu:
> cpu closed this 1 minute ago

Sorry, hit a button by mistake :sweat:

ctz:
Hey @Wonshtrum, we'd like to release 0.23 next week, and have this PR in that release. Would you be happy for us (that is, @cpu and I) to take over the splitting up and landing this over the next couple of working days? We would do that in the context of this PR, based on your original commits, so you would continue to be recorded as the author etc.

cpu:
> We would do that in the context of this PR, based on your original commits, so you would continue to be recorded as the author.

Apologies for not waiting for a response. The timelines here are a little tight and I'm making the (_hopefully correct_) assumption you'll be ok with us running with this while you're AFK/timezone shifted.

I've pushed an updated copy of the branch with both ctz + djc's feedback applied, making sure to preserve authorship throughout:

* provider example AEAD cipher prefix tidying in its own commit
* `HEADER_SIZE` type change in its own commit
* enum generator macro improvements in its own commit (with the fix for `to_be_bytes`, and `get_slice` renamed to get_array from ctz's commit-level feedback)
* the change to avoid the confusing as_ref/Dref stuff merged into the change introducing the prefixed payload, while addressing djc nits (inlining a binding, avoiding `concat` for `with_capacity`)
* buffer adapter rename in its own commit
* big type rename in its own commit
* clippy, doc and formatting changes inlined with the relevant commits

I think the end state looks good. Thanks again for your hard work on this. The type renaming is a big win for clarity/understanding IMO and the performance improvements are the icing on the cake :-)

provider-example/src/aead.rs
cpu: Done.

provider-example/src/aead.rs
cpu: Done.

rustls/src/msgs/macros.rs
cpu: Done.

rustls/src/msgs/message.rs
cpu: Applied the `get_slice` approach from 5cfd6c49c92618a38cdc0004d40f43297da83d4e, but with the enum builder fn renamed to `get_array` to address Ctz's commit level feedback. I collapsed it into the commit introducing the relevant types instead of landing the deref version and updating it subsequently.

ctz (approved):

Wonshtrum:
Awesome work!
Sorry, I didn't see the message earlier.
I would like to propose a last change. It pushes the logic introduced by the renaming even further, I would understand if you don't like it. May I push it here when rebased (it should not take long)?

cpu:
> It pushes the logic introduced by the renaming even further, I would understand if you don't like it. May I push it here when rebased (it should not take long)?

If it can be done in a separate commit so it's easy to revert if it turns out to be controversial I don't have any opposition :+1:

Wonshtrum:
I think that’s the logical next step after the message types refactor. I quite like how it avoids conversions from and between `Message` and the specialized inbound/outbound types.
But maybe this is going too far. Some places feels awkward:
- `fragment_message` takes an `OutboundOpaqueMessage` and returns an iterator of `OutboundPlainMessage`
- `OutboundOpaqueMessage` has a method `as_plain_message` which is only used by `enqueue_key_update_notification` which has to create an `OutboundOpaqueMessage` as an intermediary data owner before converting it in `OutboundPlainMessage`

I think it stems from the fact that in those places (and some more) I use `OutboundOpaqueMessage` as a generic encoded message owner. There, calling it `EncodedMessage` would be clearer. The fact that an outbound opaque message has the same memory layout is just coincidental (there are reasons due to how encryption currently works, but if it changes in the future then those types may differ). Maybe it would be worth creating an alias or implementing `OutboundOpaqueMessage` as a wrapper around an `EncodedMessage`.

cpu:
Looks like the last push broke a few CI tasks (clippy, docs, fuzz)

Wonshtrum:
I'm fixing them

cpu (approved):

Wonshtrum:
My understanding is that fundamentally there is 2 types of messages: decoded messages and encoded messages. Decoded messages are represented by `Message` and encoded ones could all be represented with `PlainMessage`. Both owns their data and technically everything could be done with only those 2 types.
As we sought to avoid allocation and copies the need of borrowed versions of `PlainMessage` came and they were specialized as along inbound/outbound and plain/opaque. The exception is `OutboundOpaqueMessage` which currently can't borrows its data due to how encryption work, so it is at its core an non-specialized `PlainMessage`.
The need of an owned encoded message still exists in some place and I used `OutboundOpaqueMessage` there when semantically it's not very appropriate.

rustls/src/msgs/fragmenter.rs
```diff
@@ -25,9 +26,9 @@ impl MessageFragmenter {
     /// Payloads are borrowed.
     pub fn fragment_message<'a>(
         &self,
-        msg: &'a PlainMessage,
+        msg: &'a OutboundOpaqueMessage,
```
ctz, line 29: This is not really the correct type; fragmenting doesn't take a potentially encrypted outgoing message and make definitely-plaintext outgoing message from it. I think it should be `OutboundPlainMessage`?

rustls/src/msgs/message_test.rs
```diff
@@ -72,7 +68,7 @@ fn can_read_safari_client_hello() {
     let mut rd = Reader::init(bytes);
     let m = OutboundOpaqueMessage::read(&mut rd).unwrap();
     println!("m = {:?}", m);
-    assert!(Message::try_from(m.into_plain_message()).is_err());
+    assert!(Message::try_from(m).is_err());
```
ctz, line 71: This doesn't make precise sense either. When we're testing decoding a message, that should be using `InboundOpaqueMessage::read`. `OutboundOpaqueMessage::read` should not exist, because there is no functional reason for a TLS library to decode messages that is generating.

fuzz/fuzzers/message.rs
```diff
@@ -4,19 +4,17 @@ extern crate libfuzzer_sys;
...
 
 use rustls::internal::msgs::codec::Reader;
-use rustls::internal::msgs::message::{Message, OpaqueMessage, PlainMessage};
+use rustls::internal::msgs::message::{Message, OutboundOpaqueMessage};
 
 fuzz_target!(|data: &[u8]| {
     let mut rdr = Reader::init(data);
-    if let Ok(m) = OpaqueMessage::read(&mut rdr) {
-        let msg = match Message::try_from(m.into_plain_message()) {
+    if let Ok(m) = OutboundOpaqueMessage::read(&mut rdr) {
```
ctz, line 11: Ditto; this program should be fuzzing the Inbound types.

fuzz/fuzzers/fragment.rs
```diff
@@ -3,32 +3,26 @@
...
-use rustls::internal::msgs::base::Payload;
 use rustls::internal::msgs::codec::Reader;
 use rustls::internal::msgs::fragmenter::MessageFragmenter;
-use rustls::internal::msgs::message::{Message, OpaqueMessage, PlainMessage};
+use rustls::internal::msgs::message::{Message, OutboundOpaqueMessage};
 
 fuzz_target!(|data: &[u8]| {
     let mut rdr = Reader::init(data);
-    let msg = match OpaqueMessage::read(&mut rdr) {
+    let msg = match OutboundOpaqueMessage::read(&mut rdr) {
```
ctz, line 12: `OutboundOpaqueMessage::read` should not need to exist; see other comment.

rustls/src/common_state.rs
```diff
@@ -271,7 +271,7 @@ impl CommonState {
 
     /// Fragment `m`, encrypt the fragments, and then queue
     /// the encrypted fragments for sending.
-    pub(crate) fn send_msg_encrypt(&mut self, m: PlainMessage) {
+    pub(crate) fn send_msg_encrypt(&mut self, m: OutboundOpaqueMessage) {
```
ctz, line 274: Should be `OutboundPlainMessage` I think.

ctz:
> A lot of things going on in the final commit, but looks good in substance.

Sorry, going to change my mind on this after working through the issues. I think we need an owned, outgoing, plaintext type to accommodate messages that need to be encoded prior to fragmentation. Using the opaque type for that works, but is the wrong type.

PR #1780, opened 2024-02-09
Change crate default features to use aws-lc-rs

ctz:
The goal here is to provide the optimum performance by default. Naturally the `ring` feature remains available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,13 +86,12 @@ need them.
 
 ### Platform support
 
-While Rustls itself is platform independent, by default it uses [`ring`] for implementing
-the cryptography in TLS. As a result, rustls only runs on platforms
-supported by `ring`. At the time of writing, this means 32-bit ARM, Aarch64 (64-bit ARM),
-x86, x86-64, LoongArch64, 32-bit & 64-bit Little Endian MIPS, 32-bit PowerPC (Big Endian),
-64-bit PowerPC (Big and Little Endian), 64-bit RISC-V, and s390x. We do not presently
-support WebAssembly.
-For more information, see [the supported `ring` target platforms][ring-target-platforms].
+While Rustls itself is platform independent, by default it uses [`aws-lc-rs`] for implementing
+the cryptography in TLS.  See [the aws-lc-rs FAQ][aws-lc-rs-platforms-faq] for more details of the
+platform/architecture support constraints in aws-lc-rs.
+
+[`ring`] is also available via the `ring` crate feature: see
+[the supported `ring` target platforms][ring-target-platforms].
 
 By providing a custom instance of the [`crypto::CryptoProvider`] struct, you
 can replace all cryptography dependencies of rustls.  This is a route to being portable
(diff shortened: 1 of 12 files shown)


cpu (commented):
The code changes LGTM. 

I think there's some documentation to update in `lib.rs`. For e.g. it still describes ring as being used by default in a comple places:

https://github.com/rustls/rustls/blob/07747f6c5649804a18bac65a0180d37aaf293e83/rustls/src/lib.rs#L54-L70

https://github.com/rustls/rustls/blob/07747f6c5649804a18bac65a0180d37aaf293e83/rustls/src/lib.rs#L81-L82

https://github.com/rustls/rustls/blob/07747f6c5649804a18bac65a0180d37aaf293e83/rustls/src/lib.rs#L92-L93

https://github.com/rustls/rustls/blob/07747f6c5649804a18bac65a0180d37aaf293e83/rustls/src/lib.rs#L307-L309

ctz:
Erm, yes, I've missed quite a bit of the meat of this.

cpu:
Spotted two more stale doc refs (_really wish GitHub allowed leaving review feedback on arbitrary lines/files!_):

https://github.com/rustls/rustls/blob/07747f6c5649804a18bac65a0180d37aaf293e83/rustls/src/builder.rs#L143-L144

https://github.com/rustls/rustls/blob/07747f6c5649804a18bac65a0180d37aaf293e83/rustls/src/crypto/mod.rs#L69-L74

ctz:
> Spotted two more stale doc refs

Thanks -- addressed these in the latest push.

cpu (approved):

djc (approved):
We should probably do a documentation/example pass to advertise using the process default everywhere?

joshtriplett:
aws-lc-rs also introduces a new build-time dependency on `cmake`; you may want to document that requirement as well so that it doesn't surprise people.

sehz:
This will introduce quite bit of friction in my opinion.    Many rust developer wants to use rustls because it is purely rust (meaning it doesn't need any additional outside build dep).   That should be higher priority than optimal performance

joshtriplett:
@sehz Neither ring nor aws-lc-rs are pure Rust, they both have components that require a native toolchain to compile.

ctz:
> aws-lc-rs also introduces a new build-time dependency on `cmake`; you may want to document that requirement as well so that it doesn't surprise people.

We have https://github.com/rustls/rustls/blob/95067cb3338a3a71c0c178a05de7a9d7f1d0aa00/rustls/src/lib.rs#L316-L317 currently -- please shout if you think that is not visible enough.

I think we'll also reiterate this in release notes, and give advice on how to stick with *ring* if people so desire.

joshtriplett:
> I think we'll also reiterate this in release notes

That was the part I was suggesting, yes. You may also want to mention in the README that a default build requires `cmake`.

martinetd:
> @sehz Neither ring nor aws-lc-rs are pure Rust, they both have components that require a native toolchain to compile.

This is probably just a matter of adding cmake to whatever container the "cross" project is using, but aws-lc-rs doesn't build with cross while ring does, so it did introduce a bit of friction at least here :)

With that said, it's not overly complicated to keep using ring, but an alternative to default with ring might be appreciated so one doesn't have to keep track of "rustls default features, minus aws_lc_rs, plus ring" (e.g. being able to do `default-fatures = false, features = ["default_ring"]`)
It probably doesn't change all that much, but std also changed with the latest release so I'm not sure how stable I should expect this to be.

Thanks!

PR #1784, opened 2024-02-12
Correct kx group selection

### LINKED ISSUE #1512: Key share selection by server puts too much weight on key shares presented by client
For background, see https://mailarchive.ietf.org/arch/msg/tls/3SoegXZowJM567_O6Fh2_LCFkPU/

We choose a suitable key exchange group in the way described in https://www.ietf.org/archive/id/draft-davidben-tls-key-share-prediction-00.html#name-downgrades which puts emphasis on avoiding a HelloRetryRequest, at the expense of client or server preference.

This isn't a pressing issue because all the key exchange groups we support are broadly as good as each other. But once we support a PQ hybrid key exchange group, it becomes more important that the server can exercise its will that this is used if the client supports it, irrespective of the key shares sent by the client.

ctz:
See https://datatracker.ietf.org/doc/draft-davidben-tls-key-share-prediction/ for background.

fixes #1512


djc (approved):

rustls/src/server/tls13.rs
```diff
@@ -197,79 +197,74 @@ mod client_hello {
...
-                        .map(|share| (share, *group))
-                });
+                .find(|group| groups_ext.contains(&group.name()))
+            {
+                Some(group) => *group,
+                None => {
+                    return Err(cx.common.send_fatal_alert(
+                        AlertDescription::HandshakeFailure,
+                        PeerIncompatible::NoKxGroupsInCommon,
+                    ));
```
cpu, line 213: I was surprised there wasn't coverage here, but I think the reason is that we also detect this condition earlier, in `server/hs.rs` before calling tls13 `handle_client_hello`: https://github.com/rustls/rustls/blob/8e4afc6d141ba4e0df0f32e9079b2ac1c011d7b9/rustls/src/server/hs.rs#L364-L369

I'm not sure I spot a way we could get coverage here from an integration test as opposed to a more targeted unit test (and that doesn't seem worth the effort).

cpu (approved):
This looks good to me.

After skimming [draft-davidben-tls-key-share-prediction-00](https://www.ietf.org/archive/id/draft-davidben-tls-key-share-prediction-00.html) I expected there might be client changes needed here or in https://github.com/rustls/rustls/pull/1785 to track the notion of a prediction-safe supported group, but it looks like neither branch considers that. Am I overlooking something?

rustls/src/server/tls13.rs
ctz: Yeah I was looking at this, and it is unreachable now. There's a matching unreachable branch in tls12.rs. I am going to look into reorganising this so `NoKxGroupsInCommon` goes back to being raised when the kx group is chosen.

cpu: Thanks, 8ef4b8d LGTM

ctz:
> I expected there might be client changes needed here or in #1785 to track the notion of a prediction-safe supported group, but it looks like neither branch considers that. Am I overlooking something?

Sorry, yes, that isn't very clear. The only part I am fixing at this point is this paragraph:

> TLS 1.3 servers implementing this document MUST NOT assume the client's key_share extension reflects client preferences. Instead, servers SHOULD select the best common named group based on supported_groups, without reference to key_share. The server then looks for the selected named group in key_share to decide whether to send HelloRetryRequest or ServerHello.

I'm cherry-picking just that part, because our old way of selecting a group wasn't supported by [RFC8446](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8) (an RFC we definitely want to support as correctly as possible!), specifically this text:

> Each KeyShareEntry value MUST correspond to a group offered in the "supported_groups" extension and MUST appear in the same order.  However, the values MAY be a non-contiguous subset of the "supported_groups" extension **and MAY omit the most preferred groups.**

(emphasis mine)

I think if draft-davidben-tls-key-share-prediction-00 is adopted by the WG and becomes a real thing, then we should reflect the "prediction safe" concept (eg, `fn is_prediction_safe(&self) -> bool` in `impl NamedGroup`) and implement whatever logic goes with it. But, until then, I think that might be jumping the gun.

cpu:
>  But, until then, I think that might be jumping the gun.

Makes sense to me :+1:

rustls/src/suites.rs
```diff
@@ -189,7 +189,10 @@ pub(crate) fn choose_ciphersuite_preferring_server(
 }
 
 /// Return a list of the ciphersuites in `all` with the suites
-/// incompatible with the Groups extension removed.
+/// incompatible with the groups extension `groups_ext` removed.
+///
+/// Leaves `all` unchanged if there is no overlap between `groups_ext`
```
djc, line 194: This seems like an odd choice. Why not clear it proactively?

ctz: At the minute we do things in this order: 1) choose a cipher suite, 2) choose a key exchange group. The logic for (1) is the same for both TLS1.2 and 1.3 (happens in `hs.rs`), whereas (2) is version-specific (in `tls12.rs` or `tls13.rs`).

Prior to the FFDHE PR, and because we only ever supported ECDHE suites and key exchange, there was no need to filter cipher suites by key exchange type (recall that for TLS1.2 only, suites are specific to ECDHE or FFDHE, so if there is no prospect of ECDHE being used, those must be removed from the choice of cipher suite.)

So this change is restoring the old behaviour, allowing the version-specific key exchange choice code to do the `NoKxGroupsInCommon` error handling as before.

The alternatives would be:
- to hoist the key exchange choice up into `hs.rs` and fuse it with the suite selection, then pass the selected group into the version-specific code
- keep things as they are, and have some statically unreachable error handling code

ctz: > to hoist the key exchange choice up into hs.rs and fuse it with the suite selection, then pass the selected group into the version-specific code

Going to see how this one goes.

ctz: OK! I think this is not too bad. Please re-review.

rustls/src/server/hs.rs
```diff
@@ -444,11 +399,162 @@ impl ExpectClientHello {
...
+                // And version
+                && suite.version().version == selected_version
+                // And protocol
+                && suite.usable_for_protocol(protocol)
+                // And key exchange groups
+                && (!ecdhe_possible || suite.usable_for_kx_algorithm(KeyExchangeAlgorithm::ECDHE))
+                && (!ffdhe_possible || suite.usable_for_kx_algorithm(KeyExchangeAlgorithm::DHE))
+            })
+            .copied()
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
ctz: Have inlined these, but note that in the default client preference case the `suitable_suites` iterator needs to be iterated several times which is what the vec was here for.

djc (approved):

PR #1785, opened 2024-02-12
Support for X25519Kyber768Draft00 post-quantum key exchange

ctz:
Draft for the moment.  This is on top of #1784.

I probably won't do `rustls-postquantum` as a separate crate, but it made sense at the time.

Most likely this will end up inside rustls/crypto/aws_lc_rs/ somewhere. I don't think it should be used by default for now, given it is pre-standardisation.

### DIFF
--- a/rustls-postquantum/Cargo.toml
+++ b/rustls-postquantum/Cargo.toml
@@ -6,3 +6,4 @@ publish = false
 
 [dependencies]
 rustls = { path = "../rustls", features = [ "aws_lc_rs" ]}
+aws-lc-rs = { version = "1.6", features = [ "unstable" ], default-features = false }
--- a/rustls-postquantum/Cargo.toml
+++ b/rustls-postquantum/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "rustls-postquantum"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+rustls = { path = "../rustls", features = [ "aws_lc_rs" ]}
--- a/rustls-postquantum/Cargo.toml
+++ b/rustls-postquantum/Cargo.toml
@@ -7,3 +7,7 @@ publish = false
 [dependencies]
 rustls = { path = "../rustls", features = [ "aws_lc_rs" ]}
 aws-lc-rs = { version = "1.6", features = [ "unstable" ], default-features = false }
+
+[dev-dependencies]
+env_logger = "0.10"
+webpki-roots = "0.26"
--- a/rustls-postquantum/examples/client.rs
+++ b/rustls-postquantum/examples/client.rs
@@ -1,0 +1,51 @@
+use std::io::{stdout, Read, Write};
+use std::net::TcpStream;
+use std::sync::Arc;
+
+fn main() {
+    env_logger::init();
+
+    let mut root_store = rustls::RootCertStore::empty();
+    root_store.extend(
+        webpki_roots::TLS_SERVER_ROOTS
+            .iter()
+            .cloned(),
+    );
+
+    let config = rustls::ClientConfig::builder_with_provider(rustls_postquantum::provider().into())
+        .with_safe_default_protocol_versions()
+        .unwrap()
+        .with_root_certificates(root_store)
+        .with_no_client_auth();
+
+    let server_name = "pq.cloudflareresearch.com"
+        .try_into()
+        .unwrap();
+    let mut conn = rustls::ClientConnection::new(Arc::new(config), server_name).unwrap();
+    let mut sock = TcpStream::connect("pq.cloudflareresearch.com:443").unwrap();
+    let mut tls = rustls::Stream::new(&mut conn, &mut sock);
+    tls.write_all(
+        concat!(
+            "GET / HTTP/1.1\r\n",
+            "Host: pq.cloudflareresearch.com\r\n",
+            "Connection: close\r\n",
+            "Accept-Encoding: identity\r\n",
+            "\r\n"
+        )
+        .as_bytes(),
+    )
+    .unwrap();
+    let ciphersuite = tls
+        .conn
+        .negotiated_cipher_suite()
+        .unwrap();
+    writeln!(
+        &mut std::io::stderr(),
+        "Current ciphersuite: {:?}",
+        ciphersuite.suite()
+    )
+    .unwrap();
+    let mut plaintext = Vec::new();
+    tls.read_to_end(&mut plaintext).unwrap();
+    stdout().write_all(&plaintext).unwrap();
+}
--- a/rustls-postquantum/src/lib.rs
+++ b/rustls-postquantum/src/lib.rs
@@ -1,9 +1,15 @@
-use rustls::crypto::aws_lc_rs::default_provider;
-use rustls::crypto::CryptoProvider;
+use rustls::crypto::aws_lc_rs::{default_provider, kx_group};
+use rustls::crypto::{
+    ActiveKeyExchange, CompletedKeyExchange, CryptoProvider, SharedSecret, SupportedKxGroup,
+};
+use rustls::{Error, NamedGroup, PeerMisbehaved};
+
+use aws_lc_rs::kem;
+use aws_lc_rs::unstable::kem::{get_algorithm, AlgorithmId};
 
 pub fn provider() -> CryptoProvider {
     let parent = default_provider();
-    let mut kx_groups = vec![];
+    let mut kx_groups = vec![&X25519Kyber768Draft00 as &dyn SupportedKxGroup];
     kx_groups.extend(parent.kx_groups);
 
     CryptoProvider {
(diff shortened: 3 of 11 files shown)


rustls-postquantum/Cargo.toml
```diff
@@ -7,3 +7,7 @@ publish = false
 [dependencies]
 rustls = { path = "../rustls", features = [ "aws_lc_rs" ]}
 aws-lc-rs = { version = "1.6", features = [ "unstable" ], default-features = false }
+
+[dev-dependencies]
+env_logger = "0.10"
```
djc, line 12: Add the comment caveat about avoiding the upgrade?

djc (commented):
Cool!

rustls/src/crypto/mod.rs
```diff
@@ -371,6 +371,28 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
...
+        let kx = self.start()?;
+        let group = kx.group();
+        let pub_key = kx.pub_key().to_vec();
+        let secret = kx.complete(peer_pub_key)?;
+
+        Ok(CompletedKeyExchange {
+            group,
+            pub_key,
+            secret,
+        })
```
cpu, line 393: nit: since you have to create all of these bindings just to use them for the struct field initialization it feels more slightly more natural to do it in-line.
```suggestion
        Ok(CompletedKeyExchange {
            group: kx.group(),
            pub_key: kx.pub_key().to_vec(),
            secret: kx.complete(peer_pub_key)?,
        })
```

rustls/src/crypto/mod.rs
```diff
@@ -371,6 +371,28 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
...
+    /// Start and complete a key exchange, in one operation.
+    ///
+    /// The default implementation for this calls `start()` and then calls
+    /// `complete()` on the result.  This is suitable for Diffie-Hellman-like
+    /// key exchange algorithms, where there is not a data dependency between
+    /// our key share (named "pub_key" in this API) and the peer's (`peer_pub_key`).
+    ///
+    /// If there is such a data dependency (like key encapsulation mechanisms), this
+    /// function should be implemented.
+    fn start_and_complete(&self, peer_pub_key: &[u8]) -> Result<CompletedKeyExchange, Error> {
```
cpu, line 383: Revisiting [semver guidance](https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-default-item) this is "Possibly-breaking", but only if the same method exists in another trait that would be implemented on a type implementing `CryptoProvider`, and that feels unlikely :+1:

rustls/src/crypto/mod.rs
```diff
@@ -455,6 +477,18 @@ pub trait ActiveKeyExchange: Send + Sync {
     fn group(&self) -> NamedGroup;
 }
 
+/// The result from [`SupportedKxGroup::start_and_complete()`].
+pub struct CompletedKeyExchange {
```
cpu, line 481: I wondered if this should have a `Debug` impl, but it would need to be custom: `SharedSecret` doesn't impl `Debug` (and shouldn't) - so probably not worth it.

rustls/src/msgs/handshake.rs
```diff
@@ -358,10 +358,10 @@ pub struct KeyShareEntry {
 }
 
 impl KeyShareEntry {
-    pub fn new(group: NamedGroup, payload: &[u8]) -> Self {
+    pub fn new(group: NamedGroup, payload: impl Into<Vec<u8>>) -> Self {
```
cpu, line 361: I think we can ignore semver compat here because `KeyShareEntry` is only exposed via the `internal` namespace.

rustls-postquantum/src/lib.rs
```diff
@@ -1,13 +1,132 @@
...
+    let mut kx_groups = vec![&X25519Kyber768Draft00 as &dyn SupportedKxGroup];
     kx_groups.extend(parent.kx_groups);
 
     CryptoProvider {
         kx_groups,
         ..parent
     }
 }
+
+///
```
cpu, line 21: TODO?

rustls-postquantum/examples/client.rs
```diff
@@ -0,0 +1,51 @@
...
+
+fn main() {
+    env_logger::init();
+
+    let mut root_store = rustls::RootCertStore::empty();
+    root_store.extend(
+        webpki_roots::TLS_SERVER_ROOTS
+            .iter()
+            .cloned(),
+    );
```
cpu, line 13: nit:
```suggestion
    let root_store = RootCertStore {
        roots: webpki_roots::TLS_SERVER_ROOTS.into(),
    };
```

rustls-postquantum/examples/client.rs
```diff
@@ -0,0 +1,51 @@
+use std::io::{stdout, Read, Write};
```
cpu, line 1: Can you add a short rustdoc comment describing the purpose of the example here?

xuganyu96:
Hi all, I am working on a project experimenting with integrating post-quantum KEMs into `rustls`, and this PR's proposed addition of `start_and_complete` to `ActiveKeyExchange` seems to really help. Would it be possible to merge a subset of this PR, including `ActiveKeyExchange::start_and_complete` and the changes to `emit_server_hello`?

cpu:
>  Would it be possible to merge a subset of this PR

I think it would probably be easier for you to achieve your goal by taking a Cargo patch on this branch or a specific rev in it, no? You'll have to do the same thing to pick up changes in `main` that aren't in a released version.

rustls-postquantum/examples/client.rs
ctz: added a comment about the purpose and what to look for in its output

rustls-postquantum/examples/client.rs
ctz: also done this to simpleclient.rs

rustls-postquantum/src/lib.rs
ctz: added docs for this

ctz:
Have pushed a new revision of this:

- addressing the above review comments
- added quite a bit of commentary at the crate level
- made explicit ciphertext/share types in instead of lots of slicing

I think I like the idea of releasing this as a separate crate for now, because:

- it's pre-standardisation, so might have a short lifetime and different release cadence.
- it implies enabling the `unstable` feature on `aws-lc-rs`, I don't want to do this for the main crate (or add a new feature).

Once those two become untrue, I think we can migrate it into the main crate and sunset `rustls-postquantum`.

rustls-postquantum/src/lib.rs
```diff
@@ -0,0 +1,231 @@
+//! This crate provides a [`rustls::crypto::CryptoProvider`] that includes
+//! a hybrid[^1], post-quantum-secure[^2] key exchange algorithm --
+//! specifically [X25519Kyber768Draft00].
+//!
+//! X25519Kyber768Draft00 is pre-standardization, so you should treat
+//! this as experimental.  You may see unexpected interop failures, and
+//! the algorithm implemented here may not be the one eventually
+//! becomes widely deployed.
```
cpu, line 8: typo: missing a word here I think:

```suggestion
//! the algorithm implemented here may not be the one that eventually
//! becomes widely deployed.
```

rustls-postquantum/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+//!
+//! However, the two components of this key exchange are well regarded:
+//! X25519 alone is already used by default by rustls, and tends to have
+//! higher quality implementations that other elliptic curves.
```
cpu, line 12: typo:

```suggestion
//! higher quality implementations than other elliptic curves.
```

rustls-postquantum/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+    let mut kx_groups = vec![&X25519Kyber768Draft00 as &dyn SupportedKxGroup];
+    kx_groups.extend(parent.kx_groups);
+
+    CryptoProvider {
+        kx_groups,
+        ..parent
+    }
+}
+
+/// This is [X25519Kyber768Draft00] key exchange.
```
cpu, line 73: nit: feels like this reads better with "the" but I don't feel strongly:

```suggestion
/// This is the [X25519Kyber768Draft00] key exchange.
```

rustls-postquantum/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+impl CombinedShare {
+    fn combine(x25519: &[u8], kyber: kem::Ciphertext) -> Self {
+        let mut out = CombinedShare(vec![0u8; COMBINED_CIPHERTEXT_LEN]);
+        out.0[..X25519_LEN].copy_from_slice(x25519);
+        out.0[X25519_LEN..].copy_from_slice(kyber.as_ref());
+        out
+    }
+}
+
+const NAMED_GROUP: NamedGroup = NamedGroup::Unknown(0x6399);
```
cpu, line 223: I think it's unrelated to this branch, but I wonder if the `Unknown` variant could have a more descriptive name. In this context it _is_ known, just not to Rustls, but to the extension crypto provider :thinking: 

Maybe `NamedGroup::Extension`? I'll give it more thought. Just rubber ducking here.

rustls-postquantum/examples/client.rs
```diff
@@ -0,0 +1,58 @@
+//! This is the simplest possible client using rustls-postquantum, based on
```
cpu, line 1: Would you mind also adding a `cargo run --locked --bin rustls-postquantum` to the `.github/workflows/daily-tests.yml`'s `example-tests` section? It could go alongside the `simpleclient`, `limitedclient`, etc steps.

djc:
The preferred spelling for post-quantum appears to use a hyphen, so I'd expect the crate to be named rustls-post-quantum instead?

cpu (commented):
Nice! I just had some small nits to flag.

cpu (approved):

rustls-postquantum/src/lib.rs
```diff
@@ -1,14 +1,231 @@
...
+    fn pub_key(&self) -> &[u8] {
+        &self.combined_pub_key
+    }
+
+    fn group(&self) -> NamedGroup {
+        NAMED_GROUP
+    }
+}
+
+fn kyber768_r3() -> &'static kem::Algorithm<AlgorithmId> {
```
djc, line 165: Nit: move this down, to just above the consts?

rustls-postquantum/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "rustls-postquantum"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+rustls = { path = "../rustls", features = [ "aws_lc_rs" ]}
```
djc, line 8: Nit: I think convention is not to have spaces inside `[]` for features (also in later commits).

djc (approved):
Cool!

Some nits, plus the thing I mentioned about including a dash in the new crate's name.

rustls-postquantum/src/lib.rs
ctz: `Other` perhaps?

rustls-postquantum/examples/client.rs
ctz: Good point, have added this, with a `grep` at the end to check this basically works.

ctz:
> so I'd expect the crate to be named rustls-post-quantum instead?

Renamed.

rustls-postquantum/src/lib.rs
cpu: I like it, I'll put up a separate rename PR so we can resolve this thread.

ctz: I think this would be a breaking change -- maybe we can leave it for later?

ctz: (Or alternatively we could branch for 0.23 maintenance.)

rustls-postquantum/examples/client.rs
ctz: Hah, that was surprisingly annoying. It seems some cloudflare POPs serve different pages for pq.cloudflareresearch.com - in the UK I get a static HTML page, but CI in the US gets javascript 🤨 

So now this hits the API endpoint that the javascript targets.

cpu: Tricksy CDNs and their tricksy games :roll_eyes:

cpu (approved):

PR #1787, opened 2024-02-13
Make peer-certificates lifetime explicit

richardpringle:
I was having some issues with the inferred lifetimes here. I think `CertificateDer<'_>` is elided as `CertificateDer<'a>` because `'static: 'a`. If we explicitly set the lifetime as `'static`, this solves the issue.

With my change, this works:
```rust
fn get_first_cert_from_chain(conn: &rustls::ClientConnection) -> Option<rustls::pki_types::CertificateDer<'static>> {
    let certs = conn.peer_certificates();
    match certs {
        Some(certs) if certs.len() > 0 => Some(certs[0].clone()),
        _ => None,
    }
}
```
Without it, I get the following error:
```sh
error: lifetime may not live long enough
   --> src/rustls-test.rs:131:43
    |
127 |     conn: &rustls::ClientConnection,
    |           - let's call the lifetime of this reference `'1`
...
131 |         Some(certs) if certs.len() > 0 => Some(certs[0].clone()),
    |                                           ^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'static`
```

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -111,7 +111,7 @@ impl CommonState {
     /// if client authentication was completed.
     ///
     /// The return value is None until this value is available.
-    pub fn peer_certificates(&self) -> Option<&[CertificateDer<'_>]> {
+    pub fn peer_certificates(&self) -> Option<&[CertificateDer<'static>]> {
         self.peer_certificates.as_deref()
     }
 


richardpringle:
I've seen this before in a different library; I don't think it's a compiler bug. In general, `rustc` will elide a lifetime with a shorter bound.

cpu:
:wave: Hi, thanks for the PR 
> Without it, I get the following error:

I think you want [CertificateDer.into_owned](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.CertificateDer.html#method.into_owned) here:

```rust
fn get_first_cert_from_chain(
    conn: &rustls::ClientConnection,
) -> Option<rustls::pki_types::CertificateDer<'static>> {
    let certs = conn.peer_certificates();
    match certs {
        Some(certs) if certs.len() > 0 => Some(certs[0].clone().into_owned()),
        _ => None,
    }
}
```

The lifetime as written seems correct to me - it's content borrowed from the connection.

djc:
Yes, this shouldn't matter. The certificates are borrowed from the connection either way, so you have to allocate to hold on to them separately anyway. Whether the borrowed certificates have a static lifetime or not is irrelevant, so we might as well elide it.

richardpringle:
@cpu, thanks for the fix!

@djc 

>Yes, this shouldn't matter. The certificates are borrowed from the connection either way, so you have to allocate to hold on to them separately anyway. Whether the borrowed certificates have a static lifetime or not is irrelevant, so we might as well elide it.

🤔 for what I'm doing, you're right, but wouldn't you rather have the proper constraints in the signature? The lifetime can only be `'static` because of this:
https://github.com/rustls/rustls/blob/3179b925c2becc8d19eb83078590e412dff4ed28/rustls/src/common_state.rs#L42

And I would argue it's not _irrelevant_ since I have to call `into_owned` when I shouldn't need to. I get that this is kind of on the compiler eliding the wrong bound, but the error it produces is confusing since you expect the `'static` lifetime. You could save some poor future users from encountering the same issue I did and save your future self from having to respond to them. 

It's up to you; you can close the PR... but it's a better experience for this library's users if you land it (no matter how few people run into this issue).

richardpringle:
@cpu @djc thanks for the quick responses!

djc:
> > Yes, this shouldn't matter. The certificates are borrowed from the connection either way, so you have to allocate to hold on to them separately anyway. Whether the borrowed certificates have a static lifetime or not is irrelevant, so we might as well elide it.
> 
> 🤔 for what I'm doing, you're right, but wouldn't you rather have the proper constraints in the signature? The lifetime can only be `'static` because of this:
> 
> https://github.com/rustls/rustls/blob/3179b925c2becc8d19eb83078590e412dff4ed28/rustls/src/common_state.rs#L42
> 
> And I would argue it's not _irrelevant_ since I have to call `into_owned` when I shouldn't need to. I get that this is kind of on the compiler eliding the wrong bound, but the error it produces is confusing since you expect the `'static` lifetime. You could save some poor future users from encountering the same issue I did and save your future self from having to respond to them.

You do need to call `into_owned()`, because the `Connection` owns these `CertificateDer<'static>` -- it's giving you a reference to the slice containing them, but not ownership of them. Therefore, if you want to hold on to them outside of the `Connection`, you still have to allocate copies, and there's really no way around that -- and the lifetime internal to the `Connection` (which could only be `'static` given that we don't have self-referential lifetimes) is irrelevant.

ctz:
For the sake of argument, say we do this. Now a different user will come and write this code, which is equally as useful and valid as yours (and perhaps even objectively better, because it avoids copying the certificate data):

```rust
fn get_first_cert_from_chain(conn: &'a rustls::ClientConnection) -> Option<rustls::pki_types::CertificateDer<'a>> {
    let certs = conn.peer_certificates();
    match certs {
        Some(certs) if certs.len() > 0 => Some(certs[0]),
        _ => None,
    }
}
```

This would now no longer work, because the returned borrow of `certs[0]` would outlive the lifetime of the local `certs`.

> The lifetime can only be 'static because of this:

That is saying that this object keeps an owned copy of the certificate data (it does, and must). That doesn't imply that the `peer_certificates()` accessor function also needs to return an owned copy of the certificate data.

richardpringle:
If you don't want to accept the PR just close it. The fact remains that the cert is an owned type and the elided lifetime is incorrect. Let's say we didn't have lifetime elision, how would you write the return type for `peer_certificates`? Would you cast the lifetime on purpose?

I get that you have to allocate anyway, but I have to call `clone` then `into_owned` rather than just calling `clone`. It's an unnecessary lifetime cast.

djc (approved):

djc:
You can call into_owned() instead of clone(), there's no point in doing both. But I guess this might be nicer if you can also clone().

djc (approved):

ctz (approved):
Thanks! Sorry for the earlier confusion, this is the right change.

cpu (approved):

PR #1789, opened 2024-02-14
Improve provider and FIPS documentation

ctz:
This:

- builds docs off main nightly, and on changes. Making this work for multiple branches (release branches, etc) is possible, but _not_ for foreign forks or arbitrary branches. This appears at https://rustls.github.io/rustls/prerelease/
- Assorted docs bug fixes
- Beginnings of FIPS-specific docs. I suggest this will extend significantly when the aws-lc FIPS cert is issued, or there are further FIPS options to discuss.
- As suggested by djc in #1780, looking over every use of `CryptoProvider` and ensuring it is pointed towards the docs for that type.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,64 @@
+name: documentation
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: '0 18 * * *'
+
+jobs:
+  generate:
+    name: Generate pre-release documentation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
+
+      - name: Generate version information
+        run: |
+          echo >tag.html \
+            "<script>var version = document.querySelector(\"span.version\");" \
+            "version.innerHTML += \"<br>(pre-release docs from <tt>$GITHUB_REF</tt>)\";" \
+            "version.title = \"commit $GITHUB_SHA\";" \
+            "</script>"
+
+      - name: cargo doc
+        run: cargo doc --locked --all-features --no-deps --package rustls
+        env:
+          RUSTDOCFLAGS: -Dwarnings --cfg=docsrs --html-after-content tag.html
+
+      - name: Massage rustdoc output
+        run: |
+          # lockfile causes deployment step to go wrong, due to permissions
+          rm -f target/doc/.lock
+          # make resulting url be more sensible
+          mv target/doc/rustls target/doc/prerelease
+
+      - name: Package and upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: ./target/doc/
+
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    needs: generate
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}prerelease/
+    steps:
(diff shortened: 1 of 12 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,64 @@
...
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+on:
+  push:
+    #branches:
+    #  - main
```
ctz, line 11: will uncomment this before merge

rustls/src/manual/features.rs
```diff
@@ -10,25 +10,17 @@
...
 * TLS1.2 session resumption.
-* TLS1.2 resumption via tickets (RFC5077).
+* TLS1.2 resumption via tickets ([RFC5077](https://tools.ietf.org/html/rfc5077)).
 * TLS1.3 resumption via tickets or session storage.
 * TLS1.3 0-RTT data for clients.
-* Client authentication by clients.
-* Client authentication by servers.
-* Extended master secret support (RFC7627).
-* Exporters (RFC5705).
+* TLS1.3 0-RTT data for servers.
```
djc: Should we just remove the qualification from the previous bullet instead?

rustls/src/manual/features.rs
```diff
@@ -1,47 +1,53 @@
...
+* Kerberos
+* TLS 1.2 protocol compression
+* Discrete-log Diffie-Hellman `*`
+* Automatic protocol version downgrade
 * Using CA certificates directly to authenticate a server/client (often called "self-signed
   certificates"). _Rustls' default certificate verifier does not support using a trust anchor as
   both a CA certificate and an end-entity certificate in order to limit complexity and risk in
   path building. While dangerous, all authentication can be turned off if required --
-  see the [example code](https://github.com/rustls/rustls/blob/992e2364a006b2e84a8cf6a7c3eaf0bdb773c9de/examples/src/bin/tlsclient-mio.rs#L318)_.
+  see the [example code](https://github.com/rustls/rustls/blob/992e2364a006b2e84a8cf6a7c3eaf0bdb773c9de/examples/src/bin/tlsclient-mio.rs#L318)_ `*`
```
djc, line 51: This isn't really affected by crypto providers, right?

rustls/src/manual/features.rs
```diff
@@ -1,47 +1,53 @@
...
-* TLS1.2 session resumption.
-* TLS1.2 resumption via tickets ([RFC5077](https://tools.ietf.org/html/rfc5077)).
-* TLS1.3 resumption via tickets or session storage.
-* TLS1.3 0-RTT data for clients.
-* TLS1.3 0-RTT data for servers.
-* Server and optional client authentication.
-* Extended master secret support ([RFC7627](https://tools.ietf.org/html/rfc7627)).
-* Exporters ([RFC5705](https://tools.ietf.org/html/rfc5705)).
-* OCSP stapling by servers.
+* TLS1.2 and TLS1.3
```
djc, line 11: Tiny nit: it would be nice to split the removal of periods from the other changes here. Understand if you don't think that's worth it, though.

djc (approved):

rustls/src/manual/features.rs
cpu: Separate from this documentation spike it might be nice to land CLI flags in tlsserver-mio/tlsclient-mio for selecting a provider at runtime, or fips mode.

cpu (approved):
Looks like a nice set of improvements. Thanks!

rustls/src/manual/features.rs
ctz: have split that into its own commit

rustls/src/manual/features.rs
ctz: I have reworded this to avoid the implication that every customisation comes via `CryptoProvider`

PR #1790, opened 2024-02-14
examples: ignore interrupted syscalls for mio poll

cpu:
While in general these examples shouldn't be written to handle errors, the long-running mio poll operation is especially prone to returning interrupted syscall errors when a debugger is attached. 

I often use these examples for testing since they expose the most helpful CLI flags of all of the examples and are great for quick exploratory debugging sessions. This branch updates each mio example to ignore this class of error rather than panicing, improving the debugging experience.

### DIFF
--- a/examples/src/bin/tlsclient-mio.rs
+++ b/examples/src/bin/tlsclient-mio.rs
@@ -534,7 +534,14 @@ fn main() {
     tlsclient.register(poll.registry());
 
     loop {
-        poll.poll(&mut events, None).unwrap();
+        match poll.poll(&mut events, None) {
+            Ok(_) => {}
+            // Polling can be interrupted (e.g. by a debugger) - retry if so.
+            Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
+            Err(e) => {
+                panic!("poll failed: {:?}", e)
+            }
+        }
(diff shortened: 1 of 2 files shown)


examples/src/bin/tlsclient-mio.rs
```diff
@@ -534,7 +534,13 @@ fn main() {
     tlsclient.register(poll.registry());
 
     loop {
-        poll.poll(&mut events, None).unwrap();
+        match poll.poll(&mut events, None) {
+            Ok(_) => {}
+            Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
```
djc, line 539: Maybe add one line of comment on why this is here?

(If you ditch the `;` after the `panic!()` call you'll likely get more compact formatting.)

djc (approved):

examples/src/bin/tlsclient-mio.rs
cpu: > Maybe add one line of comment on why this is here?

Sure, done.

> (If you ditch the ; after the panic!() call you'll likely get more compact formatting.)

Done as well, but I don't think it nudged rustfmt into doing anything differently in this case.

PR #1794, opened 2024-02-19
Tidy away `BorrowedPlainMessage`

ctz:


rustls/tests/api.rs
```diff
@@ -798,21 +787,23 @@ fn test_tls13_late_plaintext_alert() {
...
-        ))
+        .read_tls(&mut io::Cursor::new(&build_alert(AlertLevel::Fatal, AlertDescription::UnknownCA, &[])))
         .unwrap();
 
     // The server should produce a decrypt error, trying to decrypt a plaintext alert.
     assert_eq!(server.process_new_packets(), Err(Error::DecryptError));
 }
 
+fn build_alert(level: AlertLevel, desc: AlertDescription, suffix: &[u8]) -> Vec<u8> {
+    let mut v = vec![ ContentType::Alert.get_u8() ];
```
djc, line 798: Not sure why were not using `Message::build_alert()` here anymore?

djc (approved):
Nice!

rustls/tests/api.rs
ctz: `Message::build_alert` can only make syntactically valid messages, but this function is less restrictive. An additional side-effect of this change is to remove the only use of `to_unencrypted_opaque()` on `InboundMessage`.

djc: Fair!

cpu (approved):
:broom: :+1:

PR #1797, opened 2024-02-20
Extract all straight lookups of suites and kx groups

ctz:
Move `find_cipher_suite` and `find_kx_group` from `ClientConfig` to `CryptoProvider` so they can be used in `crate::server` (though there's only one instance as it happens).


djc (approved):
Hmm, the increased verbosity (partly due to formatting) on the client side makes me doubt whether this ends up being an improvement?

ctz:
Yes, I agree.

I was thinking maybe of benchmarking an "index" type for these (so `find_cipher_suite` could be O(1) for a small set of common suites), but that wouldn't be inside `CryptoProvider` anyway. But these linear lookups don't appear in flamegraphs so it's probably not worth it either.

Issue #1777, opened 2024-02-08
0.23 release preparation


cpu:
Let's collect up a burndown list of tasks to be completed for a 0.23 release that includes FIPS support and the new choice for default crypto provider.

* [x] https://github.com/rustls/rustls/pull/1798
* [x] https://github.com/rustls/rustls/pull/1811
* [x] https://github.com/rustls/rustls/pull/1502 (maybe)
* [x] https://github.com/rustls/rustls/pull/1774
* [x] https://github.com/rustls/rustls/pull/1640
* [x] https://github.com/rustls/rustls/pull/1784
* [x] https://github.com/rustls/rustls/pull/1789
* [x] https://github.com/rustls/rustls/pull/1780

cpu:
I've proposed a handful of items to consider blocking on.

I think #1640 is very close and probably uncontroversial. I'm not sure if we want to roll #1774 in as well.

It feels important to tie a bow on the outstanding changes Quinn needs in #1741. 

I think we're missing an issue to link for switching the default provider.

djc:
This list looks good to me!

cpu:
Added https://github.com/rustls/rustls/pull/1784 to the list.

ctz:
Yes, unfortunately I think any change to the default feature list is a breaking change :(

cpu:
Thanks for confirming. I put it in the list as a nice-to-have. Looking forward I think pt2 _could_ be landed without breaking changes so it would be ideal if we could fit pt1 into this release, but if it doesn't happen it would be OK knowing there's further work to be done in either case.

djc:
Would be nice to get #1811 in.

cpu:
> Would be nice to get https://github.com/rustls/rustls/pull/1811 in.

Added to the list. It's in the merge queue now.

cpu:
[v0.23.0](https://github.com/rustls/rustls/releases/tag/v%2F0.23.0) was published this morning :rocket:

PR #1798, opened 2024-02-20
Next steps for Quinn support (cont'd)

cpu:
This is a continuation of https://github.com/rustls/rustls/pull/1741, with some adjustments/new work:

1. It's rebased on `main` instead of `rel-0.22` - I'm happy to backport this work if it would help with Quinn testing but in order to merge this to unblock #1777 we'll need a version targetting `main`.
2. [Ctz's feedback](https://github.com/rustls/rustls/pull/1741#discussion_r1486434929) w.r.t ciphersuite integrity and confidentiality limits has been integrated.
3. The change to start constructing quic keys from the crypto provider is separated from the change making the `CryptoProvider` accessible from a `[Client|Server]Config`.
4. Instead of implementing `AsRef<Arc<CryptoProvider>>` to expose the provider, simpler accessor methods are used.
5. The QUIC-specific deframer panic discovered [downstream in Quinn's testing](https://github.com/quinn-rs/quinn/pull/1715#issuecomment-1908972770) has been reproduced with a simple integration test, and fixed. This was a regression introduced in 0.22. The provided integration test passes when backported to 0.21, panics on 0.22, and panics on tip of `main`. With the fix proposed in this branch the test passes again.
6. The `quic::InitalSuite` API improvements from the original branch are reworked based on Djc's later feedback, adding a `quic::Suite` type and a `Tls13CipherSuite` helper.


rustls/src/msgs/deframer.rs
```diff
@@ -492,9 +492,9 @@ impl DeframerBuffer<'_, InternalPayload> for DeframerVecBuffer {
 }
 
 impl<'a> DeframerBuffer<'a, ExternalPayload<'a>> for DeframerVecBuffer {
-    fn copy(&mut self, payload: &ExternalPayload<'a>, at: usize) {
+    fn copy(&mut self, payload: &ExternalPayload<'a>, _at: usize) {
         let len = payload.len();
-        self.unfilled()[at..at + len].copy_from_slice(payload.0);
+        self.unfilled()[0..len].copy_from_slice(payload.0);
```
djc, line 497: Nit: would be idiomatic to just leave off the `0` in this range expression.

rustls/src/client/client_conn.rs
```diff
@@ -287,6 +287,11 @@ impl ClientConfig {
         }
     }
 
+    /// Return the crypto provider used to construct this client configuration.
+    pub fn crypto_provider(&self) -> Arc<CryptoProvider> {
```
djc, line 291: I think it would make sense to avoid the `Arc::clone()` inside the method, yielding `&Arc<_>` instead. This allows the caller to access the `CryptoProvider` without bumping the reference count. While bumping the reference count is cheap, it is not free, so I think we should allow the caller to decide whether they need it.

FWIW, I do still think `AsRef` is a nice fit for this. IMO `AsRef` (as opposed to `Deref`) conceptually implies more of a "has-a" relationship than an "is-a" relationship. But in the end this amounts to bikeshedding, so while I slightly prefer the `AsRef` solution I'm happy with either way.

rustls/src/quic.rs
```diff
@@ -779,6 +780,44 @@ impl<'a> KeyBuilder<'a> {
...
+/// Crypto provider context required for producing QUIC initial keys
+#[derive(Clone, Copy)]
+pub struct InitialSuite {
+    pub(crate) suite: &'static Tls13CipherSuite,
+    pub(crate) quic: &'static dyn Algorithm,
+}
+
+impl InitialSuite {
+    /// Find a suitable cipher suite from the given [`CryptoProvider`].
+    pub fn from_provider(provider: &CryptoProvider) -> Option<Self> {
```
djc, line 792: It occurred to me after writing this (which solves the main use case from Quinn) that this trivially generalizes to a more generic `QuicSuite` if we pass a `CipherSuite` parameter in addition to the `CryptoProvider`. In fact, given that we can now break compatibility, I think there are some benefits to be had if we make this a method on `CryptoProvider` that can yield this type given a `CipherSuite`.

I think my work on updating Quinn made it clear that the current `CryptoProvider` `Quic` abstraction is kind of painful because the `quic::Algorithm` doesn't get access to the full `Tls13CipherSuite`. The rustls 0.22 abstractions are kind of painful because a `Tls13CipherSuite` carries an `Option<quic::Algorithm>` which makes the only way of representing a cipher suite that **definitely** allows QUIC a pair of both where the `Option` has been peeled off or a type like this `InitialSuite` wrapping that.

As such, I think the better abstraction here (given that we can now break compatibility again) would be to have a trait implementation that crypto providers must implement that takes a `&'static Tls13CipherSuite` and yields `Option<quic::Suite>` (where `Suite` is a renamed version of this `InitialSuite`). This would be exposed as yet another `&'static dyn QuicSuiteProvider` field in `CryptoProvider`, I think.

djc (commented):
Thanks for picking this up, @cpu!

I think we should definitely get the [Quinn upgrade PR](https://github.com/quinn-rs/quinn/pull/1715) passing CI before we merge this, to avoid getting similar issues again.

rustls/src/quic.rs
ctz: I wonder if Quic is another ordinal in `SupportedCipherSuite`? Right now there is a lot of unused baggage that comes along if someone uses say `TLS13_AES_256_GCM_SHA384` with quic, such as `Tls13CipherSuite::aead_alg` (which AIUI gets instantiated but never used). I think that would be a fair significant refactoring, and implies that (eg) `QUIC_AES_256_GCM_SHA384` would need to be introduced.

djc: That might be practical but seems like it gets weird in terms of managing the server handshake for example, where the choice is really between 1.2 and 1.3 and not between 1.2/1.3/QUIC. I do feel like constructing a separate `QuicSuite` with the `Tls13CipherSuite` hits a decent sweet spot.

cpu: >  I do feel like constructing a separate QuicSuite with the Tls13CipherSuite as input hits a decent sweet spot.

I'll give this a shot and we can see how it looks :+1:

cpu:
> I think we should definitely get the https://github.com/quinn-rs/quinn/pull/1715 passing CI before we merge this

@djc I'm thinking the best route for that is to update quinn#1715 to target this branch/Rustls main. From my perspective it seems like it might make the most sense to have Quinn jump over 0.22 straight to the soon-to-be-cut 0.23 that would roll in the changes from this branch.

djc:
> > I think we should definitely get the [quinn-rs/quinn#1715](https://github.com/quinn-rs/quinn/pull/1715) passing CI before we merge this
> 
> @djc I'm thinking the best route for that is to update quinn#1715 to target this branch/Rustls main. From my perspective it seems like it might make the most sense to have Quinn jump over 0.22 straight to the soon-to-be-cut 0.23 that would roll in the changes from this branch.

Yup, I agree that that makes sense.

cpu:
I guess one downside is that it wouldn't help the folks still working through upgrading to the 0.22 release (e.g. h3 for the reqwest/hyper ecossystem). They'll still be blocked, but now on a 0.23 update instead of a Quinn update w/ 0.22.

rustls/src/quic.rs
cpu: > The rustls 0.22 abstractions are kind of painful because a Tls13CipherSuite carries an Option<quic::Algorithm> which makes the only way of representing a cipher suite that definitely allows QUIC a pair of both where the Option has been peeled off or a type like this InitialSuite wrapping that.

Or, could we address this awkwardness at the `Tls13CipherSuite` level?

```rust
impl Tls13CipherSuite {
    pub fn quic(&'static self) -> Option<crate::quic::Suite> {
        self.quic
            .map(|quic| crate::quic::Suite { quic, suite: self })
    }
}
```

cpu: > could we address this awkwardness at the Tls13CipherSuite level?

Here's some progress on this front for consideration:

* [`cpu-quic-support-review`](https://github.com/rustls/rustls/compare/main...cpu:rustls:cpu-quic-support-review) is this branch with review feedback applied and the last commit switched out for the `Tls13CipherSuite` update I described above.
* [`cpu-rustls-0.23`](https://github.com/quinn-rs/quinn/compare/main...cpu:quinn:cpu-rustls-0.23) is https://github.com/quinn-rs/quinn/pull/1715 updated to:
  * Track this branch, and rustls 0.23.0-alpha.0.
  * Use the new updated `Tls13CipherSuite` and `quic::Suite` types.
  
Like quinn#1715 the `quinn-proto` tests are passing, but there's more work to do in the other crates.

cpu: > there's more work to do in the other crates.

[`cpu-rustls-0.23-all-crates`](https://github.com/quinn-rs/quinn/compare/main...cpu:quinn:cpu-rustls-0.23-all-crates) updates the other crates so they'll build (note the `quinn` tests aren't passing yet).

For the `bench` and `perf` crates I switched what was the implied crypto provider of `ring` to `aws_lc_rs`.

I left the main `quinn` crate using the process-wide default provider, but it means the unit tests fail because there's no provider set. I think there are a few options to figure this out but I'm going to pause here for input on the Rustls branch changes and the direction you want to see Quinn go w.r.t default crypto.

ctz: > is this branch with review feedback applied and the last commit switched out for the Tls13CipherSuite update I described above.

I like this, FWIW

cpu: > I like this, FWIW

SG. I'll push the update and we can look at the Quinn side separately.

rustls/src/msgs/deframer.rs
cpu: Fixed.

rustls/src/client/client_conn.rs
cpu: > I think it would make sense to avoid the Arc::clone() inside the method, yielding &Arc<_> instead. 

Makes sense to me :+1: Changed.

> while I slightly prefer the AsRef solution I'm happy with either way.

Since I think Ctz and myself both lightly prefer the accessor method approach I've gone with that instead of `AsRef`.

ctz (approved):

rustls/src/quic.rs
cpu: > I'll push the update

(Done)

djc:
> I guess one downside is that it wouldn't help the folks still working through upgrading to the 0.22 release (e.g. h3 for the reqwest/hyper ecosystem). They'll still be blocked, but now on a 0.23 update instead of a Quinn update w/ 0.22.
> 
> There isn't really a fix for that other than backporting breaking changes to 0.22 and pretending they aren't :-/

Yeah, I think it's okay to expect that everyone downstream of Quinn will just skip 0.22 and move straight to 0.23.

rustls/src/quic.rs
cpu: Here's [a branch](https://github.com/cpu/quinn/tree/cpu-rustls-0.23-all-crates-ring-default) that passes green in CI using this branch of Rustls (minus a FreeBSD task complaining about build deps - haven't looked into this).

It's sufficient for showing Quinn can take this branch without error, but would need further cleaning up to upstream.

cpu:
> cpu force-pushed the cpu-quic-support branch from 9d547f7 to 9852c1d 

Just rebasing on `main` after all the merge activity this morning.

djc (approved):

djc:
Sorry for taking so long -- I think it looks good!

rustls/src/msgs/deframer.rs
```diff
@@ -504,9 +504,9 @@ impl DeframerBuffer<'_, InternalPayload> for DeframerVecBuffer {
 
 #[cfg(feature = "std")]
 impl<'a> DeframerBuffer<'a, ExternalPayload<'a>> for DeframerVecBuffer {
-    fn copy(&mut self, payload: &ExternalPayload<'a>, at: usize) {
+    fn copy(&mut self, payload: &ExternalPayload<'a>, _at: usize) {
         let len = payload.len();
-        self.unfilled()[at..at + len].copy_from_slice(payload.0);
+        self.unfilled()[..len].copy_from_slice(payload.0);
         self.advance(len);
     }
```
MOZGIII, line 511: Can someone explain this to me?

It looks like adding a `debug_assert!` would make sense here, instead of not using `at` completely. We'd compare the `at` with the `self.used` to see if they always match - and if they don't - oopsie we panic.

rustls/src/quic.rs
```diff
@@ -705,6 +705,28 @@ pub trait PacketKey: Send + Sync {
...
+    fn confidentiality_limit(&self) -> u64;
+
+    /// Number of QUIC messages that can be safely decrypted with a single key of this type
+    ///
+    /// Once a `MessageDecrypter` produced for this suite has failed to decrypt `integrity_limit`
+    /// messages, an attacker gains an advantage in forging messages.
+    ///
+    /// This is not relevant for TLS over TCP (which is implemented in this crate)
+    /// because a single failed decryption is fatal to the connection.  However,
+    /// this quantity is used by QUIC.
```
MOZGIII, line 728: This is a bit confusing - looks like a copy-paste from the older comment.
In particular - the `which is implemented in this crate` part. Would probably be better as `which is also implemented in this crate`.

rustls/src/msgs/deframer.rs
cpu: > Can someone explain this to me?

Can you be more specific about what part you're not understanding? The [commit message](https://github.com/rustls/rustls/pull/1798/commits/c3335ece382f467d30150dd9e53fcd019474afba) has some additional context in the event you haven't seen it yet.

> It looks like adding a debug_assert! would make sense here, instead of not using at completely. 

Sounds reasonable. Do you want to open a PR?

rustls/src/quic.rs
cpu: > Would probably be better as which is also implemented in this crate (added also).

:+1: Want to open a PR with the change?

rustls/src/msgs/deframer.rs
MOZGIII: > Can you be more specific about what part you're not understanding? The [commit message](https://github.com/rustls/rustls/pull/1798/commits/c3335ece382f467d30150dd9e53fcd019474afba) has some additional context in the event you haven't seen it yet.

The motivation behind this change, the commit message explain it. It would be better capture as a code comment though.

MOZGIII: This looks like an very odd implementation of a trait, should it be just remodelled in a different way - in other words, should we fix the trait itself maybe?

cpu: > in other words, should we fix the trait itself maybe?

I think we would be open to a refactoring if you see a path forward. Reviewing [CONTRIBUTING.md](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md might be helpful if so.

I understand there to be a general agreement on the team that the deframer and associated APIs have room for improvement. From my perspective it was challenging updating this area of the codebase to support the new unbuffered connection API. I've tried my hand at trying to clean up a bit and didn't arrive at anything that was a workable improvement in the time I had boxed for the task but that's likely more a product of my own relative inexperience.

PR #1799, opened 2024-02-21
Use `end_entity` variable when verifying CertificateVerify

thomwiggers:
The function very nicely constructs the `end_entity` variable, use it throughout instead of selecting it again.

This makes it so that we use the position of `end_entity` in the chain only once, and it makes it more clear that we're using the previously-verified certificate.


djc (approved):
Nice, thanks!

djc:
Also, a surprisingly large performance win!

thomwiggers:
Measuring performance is very hard and weird---maybe the small difference in codegen somehow results in other code getting aligned differently or something

PR #1810, opened 2024-02-27
Message type tweaks

djc:
More follow-up from #1774.


ctz (approved):
nice!

cpu (approved):
Thanks :+1: 

I think this will need a rebase to address conflicts.

PR #1811, opened 2024-02-27
server: allow acceptor to send alerts after error

### LINKED ISSUE #1792: When accepting new TLS connection with Accepted::into_connection(config) no TLS alert is sent
**Checklist**
* [X] I've searched the issue tracker for similar bugs.

**Describe the bug**
When using `Accepted::into_connection(config)`, no TLS alert is sent when an error occurs during the handshake.
E.g. TLS protocol version mismatch, a server is configured to support only TLS 1.3 and a client supports only TLS 1.2.
```rust
let mut acceptor = Acceptor::default();
let accepted = loop {
    acceptor.read_tls(stream)?;
    if let Some(accepted) = acceptor.accept()? {
        break accepted;
    }
};
// Use a fixed config
let mut conn = accepted.into_connection(config)?;
let (_bytes_read, _bytes_written) = conn.complete_io(&mut stream)?;
```

When using `ServerConnection::new(config)`, TLS alert is sent.
```rust
// Use a fixed config
let mut conn = ServerConnection::new(config)?;
let (_bytes_read, _bytes_written) = conn.complete_io(&mut stream)?;
```

**To Reproduce**
1. Accept new TLS connections with `Accepted::into_connection(config)` (e.g. with a config supporting only TLS 1.3)
2. Try to do a handshake supporting only TLS 1.2
`openssl s_client -connect 127.0.0.1:8443 -tls1_2`
3. No TLS alert is received
`CONNECTED(00000003)
C0C71E44F87F0000:error:0A000126:SSL routines:ssl3_read_n:unexpected eof while reading:ssl/record/rec_layer_s3.c:304:`

When accepting connections with `ServerConnection::new(config)`, TLS alert is sent
`CONNECTED(00000003)
C0C71E44F87F0000:error:0A00042E:SSL routines:ssl3_read_bytes:tlsv1 alert protocol version:ssl/record/rec_layer_s3.c:1586:SSL alert number 70`

**Applicable Version(s)**
0.22.2

**Expected behavior**
TLS alert is sent to a client.

djc:
Fixes #1792. 

* `Acceptor`'s `accept` and `into_connection` fns now return both an `Error` and an `AcceptedAlert` in the error case. The `AcceptedAlert` can be used to write any TLS alerts produced during processing.
* Test coverage is updated to include:
  * Tests for a junk handshake being rejected in `accept` with an alert produced
  * Tests for an incompatible handshake being rejected in `into_connection` with an alert produced
* The existing `server_acceptor.rs` example program is updated to demonstrate using the `AcceptedAlert` return to write alerts.

### DIFF
--- a/examples/src/bin/server_acceptor.rs
+++ b/examples/src/bin/server_acceptor.rs
@@ -117,17 +117,27 @@ fn main() {
         // connection.
         let accepted = loop {
             acceptor.read_tls(&mut stream).unwrap();
-            if let Some(accepted) = acceptor.accept().unwrap() {
-                break accepted;
+
+            match acceptor.accept() {
+                Ok(Some(accepted)) => break accepted,
+                Ok(None) => continue,
+                Err((e, mut alert)) => {
+                    alert.write(&mut stream).unwrap();
+                    panic!("error accepting connection: {e}");
+                }
             }
         };
 
         // Generate a server config for the accepted connection, optionally customizing the
         // configuration based on the client hello.
         let config = test_pki.server_config(&crl_path, accepted.client_hello());
-        let mut conn = accepted
-            .into_connection(config)
-            .unwrap();
+        let mut conn = match accepted.into_connection(config) {
+            Ok(conn) => conn,
+            Err((e, mut alert)) => {
+                alert.write(&mut stream).unwrap();
+                panic!("error completing accepting connection: {e}");
+            }
+        };
 
         // Proceed with handling the ServerConnection
         // Important: We do no error handling here, but you should!
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -503,6 +503,7 @@ mod connection {
     use crate::error::Error;
     use crate::server::hs;
     use crate::suites::ExtractedSecrets;
+    use crate::vecbuf::ChunkVecBuffer;
 
     use alloc::boxed::Box;
     use alloc::sync::Arc;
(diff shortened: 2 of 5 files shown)


djc:
I probably won't be able contribute the test/example work this week.

rustls/src/server/server_conn.rs
```diff
@@ -730,25 +730,36 @@ impl Acceptor {
     /// Returns `Ok(Some(accepted))` if the connection has been accepted. Call
     /// `accepted.into_connection()` to continue. Do not call this function again.
     ///
-    /// Returns `Err(err)` if an error occurred. Do not call this function again.
-    pub fn accept(&mut self) -> Result<Option<Accepted>, Error> {
+    /// Returns `Err((err, alert))` if an error occurred. If an alert is returned, the
+    /// application should call `alert.write()` to send the alert to the client. Do not call
+    /// this function again.
```
cpu, line 735: I think it might be worth clarifying here if "this function" is `accept` or `alert.write`.

rustls/src/server/server_conn.rs
```diff
@@ -758,6 +769,31 @@ impl Acceptor {
...
+    }
+}
+
+impl From<ConnectionCommon<ServerConnectionData>> for AcceptedAlert {
+    fn from(conn: ConnectionCommon<ServerConnectionData>) -> Self {
+        Self(conn.core.common_state.sendable_tls)
+    }
+}
+
+impl Debug for AcceptedAlert {
```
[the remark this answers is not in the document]
cpu: I think we might want `Debug` on `Accepted` as well. I was surprised that writing `let err_result = acceptor.accept().unwrap_err();` apparently requires it:

```
error[E0277]: `Accepted` doesn't implement `std::fmt::Debug`
    --> rustls/tests/api.rs:5408:40
     |
5408 |     let err_result = acceptor.accept().unwrap_err();
     |                                        ^^^^^^^^^^ `Accepted` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
     |
```

rustls/src/server/server_conn.rs
djc: Fair, clarified.

cpu:
>  I probably won't be able contribute the test/example work this week.

I'm looking at this now (but won't push to this branch myself since it seems like you're also rebasing ATM)

> would be good to have at least two test cases for this: one where the clienthello is just junk (so accept() fails & and sends eg decode_error)

I think https://github.com/rustls/rustls/commit/2a60b6373de636490f8e0cb12dec5bc38beb1464 would meet your requirement

> another where the handshake is impossible (so into_connection() fails and sends eg handshake_failure)

Working on this one now.

> They other thing we should do in this PR is make examples/src/bin/server_acceptor.rs send alerts; so it makes a good starting point for people to crib from.

Here's an example update commit: https://github.com/rustls/rustls/commit/0c8c0c3b900ce3c3929e6eb1c2ae7a239b32584e

> I think we might want Debug on Accepted as well. I was surprised that writing let err_result = acceptor.accept().unwrap_err(); apparently requires it:

Here's that fix: https://github.com/rustls/rustls/commit/0c8c0c3b900ce3c3929e6eb1c2ae7a239b32584e

djc:
Rebased this. Feel free to take the baton for this branch!

cpu:
> Feel free to take the baton for this branch!

:handshake: 

I completed the requested test coverage and rolled it and the one new `Debug` impl into the original commit. I tacked on a second commit that updates the acceptor example to write alerts.

I've updated the PR description and I think this is now ready for review.

examples/src/bin/server_acceptor.rs
```diff
@@ -117,8 +117,14 @@ fn main() {
...
-            if let Some(accepted) = acceptor.accept().unwrap() {
-                break accepted;
+
+            match acceptor.accept() {
+                Ok(Some(accepted)) => break accepted,
+                Ok(None) => continue,
+                Err((e, mut alert)) => {
+                    alert.write(&mut stream).unwrap();
+                    panic!("error accepting connection: {}", e);
+                }
```
ctz, line 127: I think we need the same again on line ~136

cpu: Good catch, thanks. Fixed

examples/src/bin/server_acceptor.rs
```diff
@@ -117,17 +117,27 @@ fn main() {
...
             acceptor.read_tls(&mut stream).unwrap();
-            if let Some(accepted) = acceptor.accept().unwrap() {
-                break accepted;
+
+            match acceptor.accept() {
+                Ok(Some(accepted)) => break accepted,
+                Ok(None) => continue,
+                Err((e, mut alert)) => {
+                    alert.write(&mut stream).unwrap();
+                    panic!("error accepting connection: {}", e);
```
djc, line 126: Nit: prefer inline format args? (Also in the other hunk, below.)

ctz (approved):

examples/src/bin/server_acceptor.rs
cpu: Good call, fixed in both spots.

PR #1813, opened 2024-02-27
Fix new nightly warnings

ctz:
This fixes one new nightly clippy lint, plus a heap of new nightly warnings to do with redundant imports.

https://github.com/rustls/webpki/pull/234 is the similar change in webpki.

### DIFF
--- a/rustls/src/limited_cache.rs
+++ b/rustls/src/limited_cache.rs
@@ -124,6 +124,8 @@ where
 
 #[cfg(test)]
 mod tests {
+    use std::prelude::v1::*;
+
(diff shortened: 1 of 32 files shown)


cpu (approved):

rustls/src/limited_cache.rs
```diff
@@ -124,6 +124,7 @@ where
 
 #[cfg(test)]
 mod tests {
+    use std::prelude::v1::*;
```
djc, line 127: Maybe keep an empty line between the import and the type alias?

djc (approved):

cpu (approved):

PR #1814, opened 2024-02-27
rel-0.22 CI fixes

cpu:
I started looking at backporting https://github.com/rustls/rustls/pull/1811 to rel-0.22, but it has breaking API changes (duh) so isn't appropriate. 

In the meantime, I noticed CI for `rel-0.22` needs some fixes from `main` cherry-picked to resolve some build failures. We might as well land these fixes now in case we do find something worth backporting.


ctz (approved):

cpu:
Merging with one review since this is already reviewed content

PR #1818, opened 2024-02-28
Cargo.toml: disentangle std/aws_lc_rs features

ctz:
(will see what this breaks, but I think this should be orthogonal? I know aws-lc-rs itself requires std, but that should be unrelated to this crate's features.)


cpu (approved):
This makes sense to me, thank you!

PR #1820, opened 2024-02-28
refactor: avoid pretty printing used in logging beyond debug level

Taowyoo:
# Problem

When using `rustls` in a web server, I monitored there is multiple line `warn` level logging, which is not ideal.

Logging should be one line so it could be easy processed later. (log process tools or other integration services)

# Solution

Just not use `#?` in any logging whose level >= `info`.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -434,7 +434,7 @@ impl CommonState {
             if self.is_tls13() && alert.description != AlertDescription::UserCanceled {
                 return Err(self.send_fatal_alert(AlertDescription::DecodeError, err));
             } else {
-                warn!("TLS alert warning received: {:#?}", alert);
+                warn!("TLS alert warning received: {:?}", alert);
                 return Ok(());
             }
         }


zh-jq-b:
This should be changed to a debug!().

Taowyoo:
> This should be changed to a debug!().

Yes, after another look. This line seems like for debugging.

Taowyoo:
Here is reference from TLS 1.2 which is helpful for further decision/discussion.

In [RFC 5246 section-7.2.2](https://www.rfc-editor.org/rfc/rfc5246#section-7.2.2):

>    If an alert with a level of warning is sent and received, generally
>    the connection can continue normally.  If the receiving party decides
>    not to proceed with the connection (e.g., after having received a
>    no_renegotiation alert that it is not willing to accept), it SHOULD
>    send a fatal alert to terminate the connection.  Given this, the
>    sending party cannot, in general, know how the receiving party will
>    behave.  Therefore, warning alerts are not very useful when the
>    sending party wants to continue the connection, and thus are
>    sometimes omitted.  For example, if a peer decides to accept an
>    expired certificate (perhaps after confirming this with the user) and
>    wants to continue the connection, it would not generally send a
>    certificate_expired alert.

>    user_canceled
>       This handshake is being canceled for some reason unrelated to a
>       protocol failure.  If the user cancels an operation after the
>       handshake is complete, just closing the connection by sending a
>       close_notify is more appropriate.  This alert should be followed
>       by a close_notify.  This message is generally a warning.

djc:
> This should be changed to a debug!().

Can you explain why you think so? It seems reasonable to me to log this at a warning level.

djc (approved):
I guess this makes sense. Maybe it makes sense to only log the `alert.description` -- the `AlertLevel` (the other field in `AlertMessagePayload`) is kind of implicit in the log level.

zh-jq-b:
> > This should be changed to a debug!().
> 
> Can you explain why you think so? It seems reasonable to me to log this at a warning level.

I just noticed that there is one logging feature that can be disabled. Then it's no problem to keep it warn.

ctz (approved):
Seems reasonable.

Taowyoo:
Thanks for the quick response & review!

PR #1824, opened 2024-02-29
crypto: gate ticketer module on std for aws-lc-rs

cpu:
Fixes a missing import error when building without `std` and with `aws_lc_rs` flagged by the [daily CI powerset tests](https://github.com/rustls/rustls/actions/runs/8100804284/job/22139591511):
```
$ cargo check -p rustls --no-default-features --features aws_lc_rs
   Compiling rustls v0.23.0 (/home/daniel/Code/Rust/rustls/rustls)
error[E0432]: unresolved import `ticketer`
   --> rustls/src/crypto/aws_lc_rs/mod.rs:228:9
    |
228 | pub use ticketer::Ticketer;
    |         ^^^^^^^^ use of undeclared crate or module `ticketer`
```

Adding a `std` gate on `TICKETER_AEAD` was also required to fix unused warnings for builds w/o `std` using either `ring` or `aws_lc_rs`:

```
$ cargo check -p rustls --no-default-features --features aws_lc_rs
   Compiling rustls v0.23.0 (/home/daniel/Code/Rust/rustls/rustls)
warning: static `TICKETER_AEAD` is never used
   --> rustls/src/crypto/aws_lc_rs/mod.rs:249:19
    |
249 | pub(super) static TICKETER_AEAD: &ring_like::aead::Algorithm = &ring_like::aead::AES_256_GCM;
    |                   ^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

```
$ cargo check -p rustls --no-default-features --features ring
warning: static `TICKETER_AEAD` is never used
   --> rustls/src/crypto/ring/mod.rs:196:19
    |
196 | pub(super) static TICKETER_AEAD: &ring_like::aead::Algorithm = &ring_like::aead::CHACHA20_POLY1305;
    |                   ^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

PR #1825, opened 2024-02-29
docs: update RELEASING to mention running daily-tests

cpu:
Our daily tests CI job runs some additional tests that are too slow or too flaky to be run for every merge requests. Before doing a release it's a good idea to run this workflow manually to make sure there aren't any lurking regressions (e.g. https://github.com/rustls/rustls/pull/1824) that `cargo hack` or another test from this workflow could catch pre-release.

This commit adds that guidance to `RELEASING.md` for future releases.


ctz (approved):

djc:
Ahh, didn't see this before my other comment!

PR #1827, opened 2024-03-01
Fix docs.rs build

### LINKED ISSUE #1826: 0.23 docs build is broken
Apologies if this was already reported somewhere. Seems the new build failed to build the docs: https://docs.rs/crate/rustls/0.23.0/builds/1143895

ctz:
See https://github.com/rust-lang/docs.rs/issues/1303: docs.rs's build environment has a golang package but doesn't set the environment variables needed to make it usable.

This PR avoids that by not generating docs with the `fips` feature, and then directly addressing the one API item we have under the `fips` feature.

fixes #1826

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,7 +34,8 @@ jobs:
             "</script>"
 
       - name: cargo doc
-        run: cargo doc --locked --all-features --no-deps --package rustls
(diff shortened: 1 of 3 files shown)


djc (approved):

cpu (approved):
Thanks for fixing :+1:

PR #1832, opened 2024-03-01
ci: fixup fmt unstable

cpu:
I noticed while trying to replicate what CI was doing on my local box that the fmt-unstable alias added in https://github.com/rustls/rustls/pull/1577 doesn't work like you'd hope w.r.t adding arguments before the `--`, like `--all` and `--manifest-path`. There were a couple of files in `fuzz` that weren't formatted correctly as a result. If you expand out the [format unstable job tasks](https://github.com/rustls/rustls/actions/runs/8115301189/job/22182886275) in a recent build you'll see they're erroring with:

```
> Run cargo fmt-unstable --all --manifest-path=connect-tests/Cargo.toml -- --check
Unrecognized option: 'all'
```

We have that job configured to `continue-on-error` so the errors weren't being surfaced.

To fix CI I've changed the tasks to not use the alias. For local development I've tweaked the alias slightly to include `--all`. The alias still isn't sufficient to format the two auxiliary crates (`fuzz` and `connect-tests`) but I'm not sure what the best fix is. We could add two more aliases? Or a bash helper script? Or leave it as-is since changes in those crates are rare?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,3 @@
 [alias]
 
-fmt-unstable = ["fmt", "--", "--config-path", ".rustfmt.unstable.toml"]
+fmt-unstable = ["fmt", "--all", "--", "--config-path", ".rustfmt.unstable.toml"]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -361,13 +361,13 @@ jobs:
           components: rustfmt
           toolchain: nightly-2024-02-21
       - name: Check formatting (unstable)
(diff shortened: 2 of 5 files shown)


djc (approved):

ctz:
> added in #1577 doesn't work like you'd hope w.r.t adding arguments before the `--`, like `--all` and `--manifest-path`.

Doh. That's disappointing, and kinda makes cargo aliases significantly less useful.

ctz (approved):

PR #1840, opened 2024-03-07
Attempt to improve `merge_group` job performance

ctz:


djc:
Interesting! Maybe there's something cleaner to achieve with this?

https://stackoverflow.com/questions/65384420/how-do-i-make-a-github-action-matrix-element-conditional

djc (approved):

ctz:
This worked -- https://github.com/rustls/rustls/actions/runs/8189236658 -- though fairly predictably I had to remove those platform's jobs from the branch protection rules.

PR #1842, opened 2024-03-11
Fix new nightly clippy lints

ctz:
One of these is a false positive that is being backed out upstream[^1].

[^1]: https://github.com/rust-lang/rust-clippy/issues/12438


cpu (approved):
LGTM, thanks!

djc (approved):

PR #1843, opened 2024-03-11
Add an `aws-lc-rs` feature as an alias for `aws_lc_rs`

joshtriplett:
The vast majority of Cargo features in the crates ecosystem use dashes
to separate words, rather than underscores. The fact that `aws_lc_rs`
uses underscores, and some crates depending on rustls naturally use the
same name for the feature that rustls does, has led some crates to end
up with inconsistent feature naming that throws people off (e.g. using
the wrong feature name and being surprised at the resulting compilation
failures), and has led other crates to use `aws-lc-rs` for consistency
with their other features which causes inconsistency with rustls.

Add an alias, so that it works either way, and people can reference
either one.


cpu (approved):
Seems like a reasonable solution to reduce downstream friction. Thanks :+1:

ctz (approved):

PR #1845, opened 2024-03-13
enums: rename Unknown variant to Other

cpu:
Previously the `enum_builder` macro created an `Unknown` variant as a catch all for unsupported values. Now that we've introduced new extension points for Rustls via the crypto provider interfaces some of these "Unknown" values are known and supported via extension.

This commit renames the `Unknown` variant to `Other` to better reflect that they may be known to code outside of the core of Rustls.

Opened as a draft for now becuase:

* The base branch is `jbp-hybrid-kx`.
* It's a breaking change (as called out by the semver check failing), and we might want to keep `main` for maintenance of 0.23 for now?
* It's scope is larger than just the `NamedGroup` enum that inspired the discussion in https://github.com/rustls/rustls/pull/1785, since the change happens at the level of the enum builder macro.

Maybe the juice isn't worth the squeeze, but it felt easier to decide with a diff to look at :shrug:

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -118,6 +118,10 @@ jobs:
       - name: Check provider-example client
         run: cargo run --locked -p rustls-provider-example --example client
 
+      - name: Check rustls-post-quantum client
+        run: cargo run --locked -p rustls-post-quantum --example client
+        #| grep 'You are using <em>X25519Kyber768Draft00</em>'
+
 
   feature-powerset:
     name: Feature Powerset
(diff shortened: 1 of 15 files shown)


djc:
I think it's a good change, but would probably avoid merging it until we make other (larger) semver-incompatible changes to main (that is, I don't think this by itself is a good enough reason to start maintaining a separate 0.23 branch at this point).

cpu:
> avoid merging it until we make other (larger) semver-incompatible changes to main

Yeah, I think that makes sense :+1:

PR #1846, opened 2024-03-13
Prepare 0.23.2

ctz:
# Release notes

* **Bug fix:** return correct `ConnectionTrafficSecrets` variant from `dangerous_extract_secrets()` when AES-256-GCM is negotiated.
* **New feature:** groundwork for supporting post-quantum key exchange. Experimental support for [X25519Kyber768Draft00](https://datatracker.ietf.org/doc/draft-tls-westerbaan-xyber768d00/03/) will be released as a separate crate: [rustls-post-quantum](https://crates.io/crates/rustls-post-quantum).
* Add `aws-lc-rs` crate feature as alias for `aws_lc_rs` crate feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ The community has also started developing third-party providers for Rustls:
 cryptography.
 * [`rustls-rustcrypto`] - an experimental provider that uses the crypto primitives
 from [`RustCrypto`] for cryptography.
+* [`rustls-post-quantum`]: an experimental provider that adds support for post-quantum
+key exchange to the default aws-lc-rs provider.
 
 [`rustls-mbedtls-provider`]: https://github.com/fortanix/rustls-mbedtls-provider
 [`mbedtls`]: https://github.com/Mbed-TLS/mbedtls
@@ -101,6 +103,7 @@ from [`RustCrypto`] for cryptography.
 [`boringssl`]: https://github.com/google/boringssl
 [`rustls-rustcrypto`]: https://github.com/RustCrypto/rustls-rustcrypto
 [`RustCrypto`]: https://github.com/RustCrypto
+[`rustls-post-quantum`]: https://crates.io/crates/rustls-post-quantum
 
 #### Custom provider
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2139,7 +2139,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.1"
+version = "0.23.2"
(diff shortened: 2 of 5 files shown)


cpu:
> New feature: groundwork for supporting post-quantum key exchange.

This feels a little bit terse. I was thinking it might be useful to mention:

* It's X25519Kyber768Draft00 based
* It's in a separate crate for now (and maybe a quick mention of why)
* There's an example program based on `simpleclient.rs` to crib from

WDYT?

djc (approved):

README.md
```diff
@@ -94,13 +94,16 @@ The community has also started developing third-party providers for Rustls:
...
+* [`rustls-post-quantum`]: an experimental provider that adds support for post-quantum
+key exchange to the default aws-lc-rs provider.
 
 [`rustls-mbedtls-provider`]: https://github.com/fortanix/rustls-mbedtls-provider
 [`mbedtls`]: https://github.com/Mbed-TLS/mbedtls
 [`boring-rustls-provider`]: https://github.com/janrueth/boring-rustls-provider
 [`boringssl`]: https://github.com/google/boringssl
 [`rustls-rustcrypto`]: https://github.com/RustCrypto/rustls-rustcrypto
 [`RustCrypto`]: https://github.com/RustCrypto
+[`rustls-post-quantum`]: https://crates.io/crates/rustls-post-quantum
```
cpu, line 106: If `rustls-post-quantum` is going to be published as a separate crate should its metadata have `publish = false`? (I haven't found an upstream docs page for what this option does yet)

https://github.com/rustls/rustls/blob/62e154cb997a81a1425ec99cf1b8da69acf49dd8/rustls-post-quantum/Cargo.toml#L5

cpu: Aha, [here are the docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field)

I think that would have to be removed.

ctz:
> This feels a little bit terse.

Ah, I was just really thinking of the changes to the core crate. Expanded.

cpu:
> Ah, I was just really thinking of the changes to the core crate. Expanded.

Ahhh ok, that makes sense :+1: The updated language looks good to me.

cpu (approved):

README.md
ctz: I will follow up with another PR that removes `publish = false`, and sets a version (rather than path) dependency on rustls.

cpu: SGTM

PR #1847, opened 2024-03-13
Prepare rustls-post-quantum 0.1.0

ctz:
Will make a github release for this like usual, but _not_ set it to the "latest release", and will call it "rustls-post-quantum 0.1.0". I'll copy the README text for the release notes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2159,6 +2159,21 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ctz:
> rustls / Check semver compatibility (push) Failing after 55s

`error: rustls-post-quantum not found in registry (crates.io)`

To be expected.

cpu (approved):

PR #1852, opened 2024-03-15
replace build-a-pki.sh with Rust+rcgen, rcgen 0.13

cpu:
Previously we used a bash script and the `openssl` command line tooling to generate a test PKI comprising client/end entity certificates, intermediate certificates, root certificates, and CRLs for a variety of algorithms and key sizes. This branch replaces that tooling with a Rust crate using [rcgen](https://github.com/rustls/rcgen). This will be easier to maintain over time (_who likes writing Bash....?_) and helps validate and test work underway in the rcgen crate.

As much as possible the generated certificates have been matched to the important properties of the original OpenSSL test certificates. The one significant divergence in this branch is that the RSA PKI has been spit. Previously it was one chain with a client and end-entity cert using RSA 2048, an intermediate using RSA 3072, and a root using RSA 4096. In the new iteration we match the ECDSA PKIs and make three chains: one using all RSA 2048, one using all 3072, and one using all 4096. This change requires two updates to unit tests: 

1. Because the all RSA 2048 chain is smaller than the mixed RSA chain we have to adjust an expected write byte count down slightly to account for the fact we send less data on the wire with the smaller chain.
2. The expected subject common name for the RSA CA is split into three, one per size, and the encoded bytes have changed due to the longer name that now includes the key size.


Cargo.toml
```diff
@@ -2,27 +2,35 @@
...
+  # Tests that require OpenSSL
+  "openssl-tests",
   # example of custom provider
   "provider-example",
+  # the main library and tests
+  "rustls",
   # experimental post-quantum algorithm support
   "rustls-post-quantum",
+  # test certificates and keys for rustls unit tests
+  "test-ca",
```
cpu, line 18: I started out by writing a single file Cargo script like the [`update_valid_ee_certs.rs`](https://github.com/rustls/rustls-platform-verifier/blob/main/rustls-platform-verifier/src/tests/verification_real_world/update_valid_ee_certs.rs) tool in the rustls-platform-verifier repo but ended up going back to a boring-old binary crate for a couple reasons:

1. There's no `cargo fmt` or `cargo clippy` support for these single file scripts at the current time, which is a huge downside IMO.
2. Running the script requires using nightly rust and that's a bit of a bummer.

cpu:
> rustls / MSRV (pull_request) Failing after 13s

Looks to be from the use of the Cargo patch:
> error: failed to load source for dependency `rcgen`

est31:
CI fails with:
> The package requires the Cargo feature called `workspace-inheritance`, but that feature is not stabilized in this version of Cargo (1.61.0 (a028ae4 2022-04-29)).

workspace inheritance was [stabilized in 1.64.0](https://github.com/rust-lang/cargo/pull/10859), but rustls MSRV is 1.61.0 as per https://github.com/rustls/rustls/pull/1516.

rgcen has a rolling MSRV model of 7 months support and currently the MSRV of rcgen is [1.71.1](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1711-2023-08-03).

I suppose that CI job is not attempting to build any of the dev-dependencies, but is instead to make sure that lib users can still use rustls. However, cargo still wants to build a resolved dependency graph (including Cargo.lock) and for that it touches rcgen.

The workspace inheritance might be an artifact of the PR using git dependencies: cargo does some normalization upon publication, and I *think* this also resolves all workspace inheritance and replaces it with concrete values. So maybe once we have an rcgen release and use it in this PR, this error will go away.

There might be future errors though, due to other cargo features that are not erased upon `cargo package`. Therefore, it might be a good idea to think about harmonization of rcgen and rustls MSRVs, but this is not a neccessity. We could also work on a by-case basis.

est31:
@cpu thanks for this amazing PR! It's a great demo of `rcgen` and also a good demo of rcgen's `aws_lc_rs` backend.

cpu:
>  So maybe once we have an rcgen release and use it in this PR, this error will go away.

That's my current suspicion, but if it turns out to be wrong we can reevaluate the right fix.

test-ca/src/main.rs
```diff
@@ -1,3 +1,361 @@
...
+use std::fs::{self, File};
+use std::io::Write;
+use std::net::IpAddr;
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::Duration;
+
+use rcgen::RsaKeySize;
+use rcgen::{
```
ctz, line 12: Merge these two imports?

ctz (approved):
Very nice!

Cargo.toml
djc: How about making this an `example` in one of the existing crates?

test-ca/Cargo.toml
```diff
@@ -5,3 +5,5 @@ edition = "2021"
 publish = false
 
 [dependencies]
+rcgen = { version = "0.13.0", default-features = false, features = ["aws_lc_rs", "pem"] }
```
djc, line 8: Nit: `0.13`.

rustls/tests/common/mod.rs
```diff
@@ -126,6 +126,40 @@ embed_files! {
...
+    (RSA_3072_CLIENT_CRL_PEM, "rsa-3072", "client.revoked.crl.pem");
+    (RSA_3072_INTERMEDIATE_CRL_PEM, "rsa-3072", "inter.revoked.crl.pem");
+    (RSA_3072_END_CERT, "rsa-3072", "end.cert");
+    (RSA_3072_END_CHAIN, "rsa-3072", "end.chain");
+    (RSA_3072_END_FULLCHAIN, "rsa-3072", "end.fullchain");
+    (RSA_3072_END_KEY, "rsa-3072", "end.key");
+    (RSA_3072_INTER_CERT, "rsa-3072", "inter.cert");
+    (RSA_3072_INTER_KEY, "rsa-3072", "inter.key");
+
+    (RSA_4096_CA_CERT, "rsa-4096", "ca.cert");
```
djc, line 147: Nit: your commit message references 4097-bit key sizes twice.

djc (commented):
Some minor nits, nice work!

cpu:
> How about making this an example in one of the existing crates?

I'm open to that. It seems reasonable to put it into the `examples` crate alongside `limitedclient.rs` and friends. We already take a `rcgen` dependency there for the `server_acceptor.rs` example, and I often use the test-ca certs when running the other examples.

WDYT @ctz ? Any pref?

cpu:
> Perhaps it could be in rustls/examples/internal?

That works for me :+1:

rustls/tests/common/mod.rs
cpu: Fixed.

test-ca/Cargo.toml
cpu: Fixed.

Cargo.toml
cpu: I've moved the tool to `rustls/examples/internal`.

cpu:
>> Perhaps it could be in rustls/examples/internal?
>
> That works for me 👍

I moved the tooling into `rustls/examples/internal/test_ca.rs`. I also added a short README in `test-ca` pointing to the tool. I left the commit that alphabetized the Cargo workspace members even though I'm not adding a new member anymore since we might as well do that tidying for the next time a new crate is needed.

cpu:
Starting to prep for the rcgen release required to unblock this in https://github.com/rustls/rcgen/pull/254

cpu:
Potential follow-up: RiiR [bogo regen-certs](https://github.com/rustls/rustls/blob/main/bogo/regen-certs) script.

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+            )
+            .unwrap();
+            crl_file
+                .write_all(crl.pem().unwrap().as_bytes())
+                .unwrap();
+        }
+    }
+}
+
+struct CertifiedKey {
```
djc, line 184: Why not use `rcgen::CertifiedKey` for this?

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+        &PKCS_RSA_SHA384,
+        &PKCS_RSA_SHA512,
+        &PKCS_ECDSA_P256_SHA256,
+        &PKCS_ECDSA_P384_SHA384,
+        &PKCS_ECDSA_P521_SHA512,
+        &PKCS_ED25519,
+    ];
+
+    let mut certified_keys = HashMap::with_capacity(roles.len() * sig_algs.len());
+    for role in roles {
```
djc, line 42: If each `(role, alg)` test is pretty much independent, I don't think we need three loops here? And if that's true, do we still need the hash map?

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+use rcgen::{
+    BasicConstraints, Certificate, CertificateParams, CertificateRevocationListParams,
+    DistinguishedName, DnType, ExtendedKeyUsagePurpose, Ia5String, IsCa, KeyIdMethod, KeyPair,
+    KeyUsagePurpose, RevocationReason, RevokedCertParams, RsaKeySize, SanType, SerialNumber,
+    SignatureAlgorithm, PKCS_ECDSA_P256_SHA256, PKCS_ECDSA_P384_SHA384, PKCS_ECDSA_P521_SHA512,
+    PKCS_ED25519, PKCS_RSA_SHA256, PKCS_RSA_SHA384, PKCS_RSA_SHA512,
+};
+use time::OffsetDateTime;
+
+fn main() {
```
djc, line 20: Maybe make this yield `Result<(), rcgen::Error>` so we can `?` all the errors? The try operator is quite a bit less noisy than `.unwrap()`...

(`Result<(), Box<dyn std::error::Error>>` also works if we need to unwrap other errors often.)

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+            certified_keys.insert((role, alg), CertifiedKey { cert, key_pair });
+        }
+    }
+
+    // Write the PEM chain and full chain files for the end entity and client certs.
+    // Chain files include the intermediate and root certs, while full chain files include
+    // the end entity or client cert as well.
+    for role in [Role::EndEntity, Role::Client] {
+        for alg in sig_algs {
+            let CertifiedKey { cert: root, .. } = certified_keys
```
djc, line 100: Is all the destructuring really worth it? I find these vertically squashed together three-line destructuring expressions kind of hard to read, and not sure they buy us a lot especially in the case where we're only taking a single field.

Compare:

```rust
            let CertifiedKey { cert: root, .. } = certified_keys
                .get(&(Role::TrustAnchor, alg))
                .unwrap();
            let CertifiedKey {
                cert: intermediate, ..
            } = certified_keys
                .get(&(Role::Intermediate, alg))
                .unwrap();
```

To

```rust
let root = certified_keys.get(&(Role::TrustAnchor, alg)).unwrap().cert;
let intermediate = certified_keys.get(&(Role::Intermediate, alg)).unwrap().cert;
```

(Don't know if rustfmt will actually allow these to be on one line...)

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+                        .unwrap();
+                for cert in chain_certs {
+                    chain_file
+                        .write_all(cert.pem().as_bytes())
+                        .unwrap();
+                }
+            }
+
+            write_chain_file(role, alg, "chain", &[intermediate, root]);
+            write_chain_file(role, alg, "fullchain", &[end_entity, intermediate, root]);
```
djc, line 130: Instead of a function, consider a loop over `let files = [("chain", &[intermediate, root]), ("fullchain", &[end_entity, intermediate, root])];"?

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+}
+
+struct CertifiedKey {
+    cert: Certificate,
+    key_pair: KeyPair,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
```
djc, line 189: Nit: maybe move this down.

djc (commented):
Turns out I did have some feedback... hope it's useful.

cpu:

Putting this back to draft until I have a chance to rebase + address feedback. It's a bit lower on my priority list ATM.

provider-example/examples/server.rs
```diff
@@ -70,25 +70,21 @@ impl TestPki {
...
-        let server_key_der =
-            PrivatePkcs8KeyDer::from(server_cert.serialize_private_key_der()).into();
+        let server_key = rcgen::KeyPair::generate_for(alg).unwrap();
+        let server_cert = server_ee_params
+            .signed_by(&server_key, &ca_cert, &ca_key)
+            .unwrap();
         Self {
-            server_cert_der,
-            server_key_der,
+            server_cert_der: server_cert.der().clone(),
```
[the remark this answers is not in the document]
cpu: Fixed

examples/src/bin/server_acceptor.rs
```diff
@@ -272,19 +276,19 @@ impl TestPki {
...
             key_identifier_method: rcgen::KeyIdMethod::Sha256,
-            alg: &rcgen::PKCS_ECDSA_P256_SHA256,
         };
-        rcgen::CertificateRevocationList::from_params(crl)
-            .unwrap()
-            .serialize_der_with_signer(&self.ca_cert)
+        crl_params
+            .signed_by(&self.ca_cert.cert, &self.ca_cert.key_pair)
             .unwrap()
+            .der()
```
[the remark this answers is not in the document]
cpu: Fixed.

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+    let key_pair = if *alg == PKCS_RSA_SHA256 {
+        KeyPair::generate_rsa_for(&PKCS_RSA_SHA256, RsaKeySize::_2048)
+    } else if *alg == PKCS_RSA_SHA384 {
+        KeyPair::generate_rsa_for(&PKCS_RSA_SHA384, RsaKeySize::_3072)
+    } else if *alg == PKCS_RSA_SHA512 {
+        KeyPair::generate_rsa_for(&PKCS_RSA_SHA512, RsaKeySize::_4096)
+    } else {
+        KeyPair::generate_for(alg)
+    };
+    key_pair.unwrap()
```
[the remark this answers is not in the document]
cpu: > Skip the key_pair binding and chain the unwrap()?

Done.

> Maybe use a for-loop over [(PKCS_RSA_SHA256, RsaKeySize::_2048), ..]?

This ended up being awkward so I skipped this suggestion.

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,365 @@
...
+                    ExtendedKeyUsagePurpose::ServerAuth,
+                    ExtendedKeyUsagePurpose::ClientAuth,
+                ];
+            }
+            Self::EndEntity | Self::Client => {
+                params.is_ca = IsCa::NoCa;
+                params.key_usages = vec![
+                    KeyUsagePurpose::DigitalSignature,
+                    KeyUsagePurpose::ContentCommitment,
+                ];
```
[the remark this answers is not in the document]
cpu: Done.

rustls/examples/internal/test_ca.rs
cpu: Done.

rustls/examples/internal/test_ca.rs
cpu: I pulled everything into one loop.

> And if that's true, do we still need the hash map?

It gets fairly awkward without it. We can't strictly look at the last iteration to get the issuer for the current iteration like we do in some other "build a chain" type loops because when we get to the client cert role the last iteration will have been for an end entity cert. We could save the last issuer in separate var only updated for certain roles, but then we also have to do the same for the trust anchor to be able to write a chain file for the EE or client cert role.

In the end since this is a utility only run to generate test data it felt like the approach using the hashmap had the best clarity overall.

rustls/examples/internal/test_ca.rs
cpu: Done, with the exception of `Option` `unwrap()`s where I didn't want to have to invent an error to map the `None` case to.

rustls/examples/internal/test_ca.rs
cpu: I removed the destructuring. You're right, it was noisey and wasn't providing much value.

rustls/examples/internal/test_ca.rs
cpu: Done.

rustls/examples/internal/test_ca.rs
cpu: Moved down.

cpu:
> Putting this back to draft until I have a chance to rebase + address feedback.

I've addressed the feedback and this branch is ready for another round of review.

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+}
+
+// Note: these are ordered such that the data dependencies for issuance are satisfied.
+const ROLES: [Role; 4] = [
+    Role::TrustAnchor,
+    Role::Intermediate,
+    Role::EndEntity,
+    Role::Client,
+];
```
djc, line 233: Nit: I guess this should go before the definition of `Role`.

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,321 @@
...
+
+            // Serialize the issued certificate to a PEM encoded file.
+            let mut cert_file = File::create(role.cert_pem_file_path(alg))?;
+            cert_file.write_all(cert.pem().as_bytes())?;
+            // And to a DER encoded file.
+            let mut cert_file = File::create(role.cert_der_file_path(alg))?;
+            cert_file.write_all(cert.der())?;
+
+            // If we're not a trust anchor, generate a CRL for the certificate we just issued.
+            if !matches!(role, Role::TrustAnchor) {
```
djc, line 60: Nit: since `Role` implements `PartialEq` anyway, just write `role != Role::TrustAnchor`?

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,321 @@
...
+            }
+
+            certified_keys.insert((role, alg.inner), CertifiedKey { cert, key_pair });
+        }
+    }
+
+    Ok(())
+}
+
+fn crl_for_serial(serial_number: &SerialNumber) -> CertificateRevocationListParams {
```
djc, line 124: Nit: maybe consider taking `SerialNumber` directly here, moving the `clone()` to the outside? The caller can just use `cloned()` instead of `as_ref()` so that's a small difference.

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,321 @@
...
+                    ),
+                    SanType::DnsName(Ia5String::try_from("localhost".to_string()).unwrap()),
+                    SanType::IpAddress(IpAddr::from_str("198.51.100.1").unwrap()),
+                    SanType::IpAddress(IpAddr::from_str("2001:db8::1").unwrap()),
+                ];
+            }
+        }
+
+        // Client certificates additionally get the client auth EKU.
+        if matches!(self, Self::Client) {
```
djc, line 203: Nit: rely on `PartialEq`?

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    fn label(&self) -> &'static str {
+        match self {
+            Self::Client => "client",
+            Self::EndEntity => "end",
+            Self::Intermediate => "inter",
+            Self::TrustAnchor => "ca",
+        }
+    }
```
djc, line 172: Nit: move all these simple getters below `params()` and `common_name()`?

rustls/examples/internal/test_ca.rs
```diff
@@ -0,0 +1,321 @@
...
+            Self::TrustAnchor | Self::Intermediate => {
+                params.is_ca = IsCa::Ca(BasicConstraints::Unconstrained);
+                params.key_usages = ISSUER_KEY_USAGES.to_vec();
+                params.extended_key_usages = ISSUER_EXTENDED_KEY_USAGES.to_vec();
+            }
+            Self::EndEntity | Self::Client => {
+                params.is_ca = IsCa::NoCa;
+                params.key_usages = EE_KEY_USAGES.to_vec();
+                params.subject_alt_names = vec![
+                    SanType::DnsName(Ia5String::try_from("testserver.com".to_string()).unwrap()),
```
djc, line 191: Nit: where building a `String` from a `&str`, prefer `to_owned()` over `to_string()` (which is more general and more expensive).

djc (approved):
Much better -- still some small suggestions.

rustls/examples/internal/test_ca.rs
cpu: Fixed.

rustls/examples/internal/test_ca.rs
cpu: Ah! Good point. Fixed.

rustls/examples/internal/test_ca.rs
cpu: Good idea, fixed.

rustls/examples/internal/test_ca.rs
cpu: Done.

rustls/examples/internal/test_ca.rs
cpu: Done.

rustls/examples/internal/test_ca.rs
cpu: Done.

PR #1856, opened 2024-03-18
Add auto build to bench-measure

SimonTate:
To allow for easy running and simplification of benchmarking, add cargo build to the bench measure.

Also add .PHONY for recipes with no output that are always expected to run.

### DIFF
--- a/admin/bench-measure.mk
+++ b/admin/bench-measure.mk
@@ -1,29 +1,31 @@
+.PHONY: perf perffull perf13 measure memory clean
+
 RECORD=perf record -F2000 --call-graph dwarf,16000 --
 FLAMEGRAPH=perf script | ~/FlameGraph/stackcollapse-perf.pl | ~/FlameGraph/flamegraph.pl >
 MEMUSAGE=/usr/bin/env time -f %M
+BENCH:=./target/release/examples/bench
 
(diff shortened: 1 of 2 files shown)


admin/bench-measure.mk
```diff
@@ -35,10 +37,19 @@ measure: ./target/release/examples/bench
...
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 100
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 1000
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 5000
+
+clean:
+	rm -f perf-*.svg
+	cargo clean
+
+$(BENCH):
+	cargo build --release -p rustls --example bench
```
ctz, line 53: This should be `--profile=bench` rather than `--release` (see 134e43f9a38f875af3c3d3afb12bd29a3bb5bdbf for background).

SimonTate: I've updated the makefile and the BENCHMARKING.md to reflect this comment

cpu (approved):
Thanks!

ctz (approved):

ctz:
Thanks!

Issue #1860, opened 2024-03-20
No common ciphersuit when FFDHE and ECDHE ciphersuites are available on server and client using TLS 1.2


Taowyoo:
# Bug description
When FFDHE and ECDHE ciphersuites are available on server and client when common version is TLS 1.2. Server will fail to choose common ciphersuite.

See:
https://github.com/rustls/rustls/blob/fb67f0145608c995241b852c68ed7ce28b1c241e/rustls/src/server/hs.rs#L479-L482

It is obvious that no suite cannot support DHE and ECDHE KeyExchangeAlgorithm at same time when `ecdhe_possible` and `ffdhe_possible` are true.

# Need backport

versions:
- 0.23

Taowyoo:
Created PR for fixing this: https://github.com/rustls/rustls/pull/1861

Notice, you could produce the error by revertting the fix code in `src` and run
```
cargo test --test api_ffdhe server_avoids_cipher_suite_with_no_common_kx_groups
```

djc:
Is this a regression from #1784, or has this been broken since the introduction of FFDHE support?

Taowyoo:
After testing, it's a regression from https://github.com/rustls/rustls/pull/1784.
My testing branch:
- https://github.com/fortanix/rustls/tree/yx/test-dhe
- https://github.com/fortanix/rustls/tree/yx/test-dhe-after-1784

ctz:
https://crates.io/crates/rustls/0.23.3 was released yesterday containing the fix for this.

PR #1861, opened 2024-03-20
Fix ffdhe no common cipher suite bug

Taowyoo:
# Background

See:

- #1860

# Solution

- Fix the bug that when both FFDHE and DHE ciphersuites are available on client and server, no ciphersuite is choose.
- Add missing test cases.

This PR fixs #1860


djc (approved):

cpu (approved):
Thanks for noticing/fixing. This seems sensible to me :+1:

ctz (approved):
Thanks! And sorry for the regression!

djc:
Maybe time to release 0.23.3?

PR #1863, opened 2024-03-20
Prepare 0.23.3

ctz:
Release notes:

* **Bug fix**: correct cipher suite selection when ECDHE and FFDHE suites are both offered.


djc (approved):

cpu (approved):
Thanks!

Taowyoo:
Thank you for the quick patch!

PR #1865, opened 2024-03-21
Add path for rustls-post-quantum -> rustls dependency

ctz:
This avoids our Cargo.lock containing a previous version of this crate, and means a local `cargo build` is sufficient to check rustls-post-quantum/ builds against the current rustls/.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2145,21 +2145,6 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


djc (approved):

cpu:
> rustls / Clippy (push) Failing after 1m

Looks like we have some more tidying to do on main

cpu (approved):

ctz:
> > rustls / Clippy (push) Failing after 1m
> 
> Looks like we have some more tidying to do on main

#1866

cpu:
> cpu force-pushed the jbp-rpq-path-dep branch from cab850d to 1aed96c 

Hit the UI button to rebase so the merge queue could take over.

Issue #1867, opened 2024-03-21
doc: AcceptedAlert::write doesn't necessarily write all bytes


jsha:
[The docs for AcceptedAlert::write()](https://docs.rs/rustls/latest/rustls/server/struct.AcceptedAlert.html#method.write) say "Send the alert to the client." Under the hood it calls ChunkVecBuffer::write_to, and returns the `usize`.

The problem is, `&mut dyn Write` is allowed to return short writes. So it's possible that `write_to` writes less than all of the ChunkVecBuffer's contents.

I think the correct user behavior is to call `AcceptedAlert::write()` in a loop until it returns `Ok(0)` or an error. Is that right? If so, examples/src/bin/server_acceptor.rs should be updated. And the docs should say "Send alert bytes to the client. Call this repeatedly until it returns `Ok(0)` or an error, to ensure all bytes are sent."

djc:
Fair point, we've been noticing this stuff in tokio-rustls: https://github.com/rustls/tokio-rustls/pull/48.

ctz:
Yeah we should fix the example code to deal with short writes for good form.

Though, in practice, there is really zero chance I think that that would ever happen with a real TCP sockets -- the server has written no data, so the write window is at least the MTU (the smallest of which is 68 bytes, and easily fits one alert). Though perhaps it's possible for protocols that upgrade to TLS from another protocol, like SMTP STARTTLS.

cpu:
> Yeah we should fix the example code to deal with short writes for good form.

Here's a quick patch for the doc update & example's usage of `AcceptedAlert::write`: https://github.com/rustls/rustls/pull/1868

It might make sense to do a more systematic evaluation of all the examples + writes as a follow up to make sure we address this holistically.

jbr:
Ran into exactly this in https://github.com/rustls/tokio-rustls/pull/52 by initially copying the example

jbr:
Maybe an `AcceptedAlert::write_all(&mut io) -> io::Result<()>` would improve ergonomics for blocking usage?

jsha:
Good point @jbr - or we could change the behavior of `write()` to do `write_all()`, since as @ctz says this would never happen with a real TCP socket. So we would pay the price of possibly blocking an async executor in a very unlikely situation for the convenience of users not having to loop over the result.

jbr:
[...]

PR #1868, opened 2024-03-21
Improve AcceptedAlert::write documentation, example usage

cpu:
Resolves https://github.com/rustls/rustls/issues/1867

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -810,6 +810,9 @@ mod connection {
         }
 
         /// Send the alert to the client.
+        ///
+        /// To account for short writes this function should be called repeatedly until it
+        /// returns `Ok(0)` or an error.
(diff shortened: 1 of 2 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -810,6 +810,9 @@ mod connection {
         }
 
         /// Send the alert to the client.
+        ///
+        /// To account for short-writes this function should be called repeatedly until it
```
djc, line 814: The use of a hyphen in "short-writes" seems surprising to me? I'd write it as two words (but you are the native speaker here, so happy to defer to your judgement).

djc (approved):

rustls/src/server/server_conn.rs
cpu: >  I'd write it as two words 

Yeah, agreed. Not sure why I had it with a hyphen :laughing: 

Fixed!

jsha (approved):

ctz (approved):

PR #1869, opened 2024-03-22
Correct ECDSA-SHA1 allergy

ctz:
This aims to fix https://github.com/seanmonstar/reqwest/issues/2191

Should be an easy backport to 0.22.


cpu (approved):

djc (approved):

djc:
(This is getting boring, sorry, but:) I think this probably merits a 0.23.4? (And 0.22.3?)

PR #1874, opened 2024-03-25
Cargo.lock: update semver compatible deps

cpu:
Replaces https://github.com/rustls/rustls/pull/1871

* async-trait 0.1.78 -> 0.1.79
* rayon 1.9.0 -> 1.10.0
* rustls-pki-types 1.10.3 -> 1.10.4
* regex 1.10.3 -> 1.10.4
* base64 0.21.7 -> 0.22

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2000,9 +2000,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.10.3"
+version = "1.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b62dbe01f0b06f9d8dc7d49e05a0785f153b00b2c227856282f671e0318c9b15"
+checksum = "c117dbdfde9c8308975b6a18d71f3f385c89461f7b3fb054288ecf2a2058ba4c"
 dependencies = [
  "aho-corasick",
  "memchr",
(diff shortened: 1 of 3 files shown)


djc:
So why aren't we updating to base64 0.22?

cpu:
> So why aren't we updating to base64 0.22?

`cargo update -p base64` gives me no `Cargo.lock` diff. Same for `env_logger`. 

You want to see the `Cargo.toml` versions manually incremented?

djc:
I would expect a change to update `Cargo.toml` in for the dev-dependency in rustls and in openssl-tests.

cpu:
Added a separate commit for the `base64` update. `env_logger` remains at 0.10 to avoid the 1.71 MSRV associated with 0.11.

I'll make sure to include these dev dep updates as part of the regular process for future PRs.

cpu:
>  as part of the regular process

Though, now that we know the upstream dependabot folks aren't likely to make their automation work the way we want I'm perhaps more open to the idea of using [renovatebot](https://docs.renovatebot.com/) like @ctz suggested at one point.

djc (approved):
FWIW, I also think having one commit per semver-compatible dep isn't really worth it.

For Renovate, the rustup repo has a decent setup which we could steal from.

Issue #1878, opened 2024-03-27
Option to Relax SNI Host Name Validation for IP Addresses


alexliesenfeld:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
Many HTTP clients seem to send the host IP in the [SNI host name extension](https://en.wikipedia.org/wiki/Server_Name_Indication) of the ClientHello message. This has been observed in cases where users only provide an IP address rather than a hostname for establishing a connection. At the moment, when clients send an IP address as the SNI, rustls refuses to process the ClientHello message with an `InvalidServerName` error (see [here](https://github.com/rustls/rustls/blob/074ced1fd5cbc06309d811c51c125707a7e17013/rustls/src/msgs/handshake.rs#L237)), preventing servers from serving requests for such clients. [This example project](https://gist.github.com/alexliesenfeld/1c366c8d30cff769a59cf74c6aaba9cc) demonstrates this issue with some Rust-based HTTP clients, but users observe this behavior with [other clients](https://github.com/rustls/rustls/issues/184#issuecomment-2022272735) as well. 

**Describe the solution you'd like**
Even though IP addresses are not allowed in the SNI, as per [RFC 6066](https://datatracker.ietf.org/doc/html/rfc6066#section-3) the SNI extension also seems to be optional for both sides, the client and server. Here is an excerpt highlighting this:

>   It is RECOMMENDED that clients include an extension of type
>
>    "server_name" in the client hello whenever they locate a server by a
>    supported name type.
> 
>    A server that receives a client hello containing the "server_name"
>    extension MAY use the information contained in the extension to guide
>    its selection of an appropriate certificate to return to the client,
>    and/or other aspects of security policy.  In this event, the server
>    SHALL include an extension of type "server_name" in the (extended)
>    server hello.  The "extension_data" field of this extension SHALL be
>    empty.

To be able to serve requests for as many clients as possible, it is desirable to have a possibility to relax the strictness in this regard.

At a first look, one option could be to let the [`cert_resolver`](https://docs.rs/rustls/latest/rustls/server/trait.ResolvesServerCert.html) handle this case (currently configured via [`ServerConfig`](https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html)), since the SNI extension seems to be primarily concerned with certificate resolution. This way, users could opt-out of the strict host name validation and use their own validation logic.

[...]

PR #1881, opened 2024-03-30
Relax `server_name` extension validation

ctz:
This PR makes us ignore the `server_name` ClientHello extension if it contains a literal IP address. We don't indicate support for the `server_name` extension if we ignored it, and it is not available via any API -- it is as if the client did not send the extension at all. Other illegal names continue to be rejected as before.

This is necessary to deal with non-compliant extension data sent by OpenSSL (https://github.com/openssl/openssl/issues/20041) and Apple SecureTransport (#1878).


rustls/src/msgs/message_test.rs
```diff
@@ -72,7 +72,7 @@ fn can_read_safari_client_hello() {
     let mut rd = Reader::init(bytes);
     let m = OutboundOpaqueMessage::read(&mut rd).unwrap();
     println!("m = {:?}", m);
-    assert!(Message::try_from(m.into_plain_message()).is_err());
+    Message::try_from(m.into_plain_message()).unwrap();
```
cpu, line 75: This is unchanged in your branch but I think this test case might benefit from having the fn renamed from `can_read_safari_client_hello` to something like `can_read_safari_client_hello_bad_sni`, or maybe there could be a comment on `bytes` that indicates the payload includes an invalid SNI of 127.0.0.1?

It's hard to divine that from the blob of hex otherwise :-)

cpu (approved):
I'm sad we have to do this. The standards are very clear on this behaviour being wrong and fixing it should be trivial for the affected projects.... It's not something deep and nuanced :face_exhaling: 

Thanks for working up a patch.

djc:
Why is the reformatting necessary/not causing issues in CI?

ctz:
> Why is the reformatting necessary/not causing issues in CI?

Unfortunately rustfmt doesn't format the inside of `{}`-braced macros, and that covers most of this file :(

rustls/src/msgs/handshake.rs
```diff
@@ -302,6 +307,12 @@ impl ConvertServerNameList for [ServerName] {
             .filter_map(only_dns_hostnames)
             .next()
     }
+
+    fn standards_complaint(&self) -> bool {
```
djc, line 311: I think you meant `_compliant`? Might be nice to have a docstring that explains the issue + maybe some links to the spec and/or issues we've seen.

I would probably prefer for this to be inlined at its single callsite.

rustls/src/msgs/handshake.rs
```diff
@@ -225,11 +226,13 @@ impl ServerNamePayload {
     }
 
     fn read_hostname(r: &mut Reader) -> Result<Self, InvalidMessage> {
+        use pki_types::ServerName;
```
djc, line 229: Nit: I think this should probably just be imported at the top of the file?

rustls/tests/api.rs
```diff
@@ -956,6 +958,86 @@ fn client_trims_terminating_dot() {
...
+                        snr.clear();
+                        snr.push(
+                            ServerNameExtensionItem::read_bytes(b"\x00\x00\x071.1.1.1").unwrap(),
+                        );
+                    }
+                }
+            }
+
+            *encoded = Payload::new(parsed.get_encoding());
+        }
```
djc, line 977: Consider linearizing this, maybe with `let .. else`?

rustls/tests/api.rs
```diff
@@ -956,6 +958,86 @@ fn client_trims_terminating_dot() {
...
+    check_sni_error(
+        insert_ip_address_server_name,
+        Error::General("no server certificate chain resolved".to_string()),
+    );
+}
+
+#[test]
+fn server_rejects_sni_with_illegal_dns_name() {
+    fn insert_illegal_server_name(msg: &mut Message) -> Altered {
+        if let MessagePayload::Handshake { parsed, encoded } = &mut msg.payload {
```
djc, line 991: Same here.

(Also looks like there's quite a bit of duplication between the two tests? Not sure if that can all be moved into `check_sni_error()`, but a helper function would probably be nicer?)

rustls/src/msgs/handshake.rs
ctz: Unfortunately there's already a `ServerName` type in this file (matching the structure defined in RFC6066). I think the more local `use` is preferable to a global one that needs to be renamed?

rustls/tests/api.rs
ctz: Have added a helper function, and then used it in the existing two other SNI editing tests.

djc (approved):

cpu (approved):

PR #1885, opened 2024-04-02
deps: update semver compatible deps

cpu:
Replaces https://github.com/rustls/rustls/pull/1882

* clap v4.5.3 -> v4.5.4 ([diff.rs](https://diff.rs/clap/4.5.3/4.5.4))
* rustls-pki-types v1.4.0 -> v1.4.1 ([diff.rs](https://diff.rs/rustls-pki-types/1.4.0/1.4.1))
* tokio v1.36.0 -> v1.37.0 ([diff.rs](https://diff.rs/tokio/1.36.0/1.37.0))
* serde_json v1.0.114 -> v1.0.115 ([diff.rs](https://diff.rs/serde_json/1.0.114/1.0.115))

Three updates from #1882 are deliberately omitted:

* hashbrown 0.13.2 -> 0.14.3 (0.14.x requires 1.63 MSRV)
* env_logger 0.10.2 -> 0.11.3 (0.11 requires 1.71 MSRV)
* rcgen 0.12.1 -> 0.13.0 (requires fixing breaking changes, in progress in https://github.com/rustls/rustls/pull/1852)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,9 +558,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.3"
+version = "4.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "949626d00e063efc93b6dca932419ceb5432f99769911c0b995f7e884c778813"
+checksum = "90bc066a67923782aa8515dbaea16946c5bcc5addbd668bb80af688e53e548a0"
 dependencies = [
  "clap_builder",
  "clap_derive",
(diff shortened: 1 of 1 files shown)


djc (approved):
Maybe we should just take the hashbrown upgrade and corresponding MSRV bump? Will still be good enough for Debian stable (and even things like Tokio are at 1.63 now IIRC).

cpu:
> Maybe we should just take the hashbrown upgrade and corresponding MSRV bump?

Seems reasonable to me, but I also don't spot anything super tempting from 1.62 or 1.63 to push me over the edge here.

cpu:
> Maybe we should just take the hashbrown upgrade and corresponding MSRV bump? 

@ctz Any opinions on this? It seems less controversial than the bump to 1.65 in https://github.com/rustls/webpki/pull/241 since 1.63 is in Debian stable.

djc:
It's interesting because the hashbrown dependency doesn't actually show up for me -- I guess it's only for non-`std` users? So might not be that big a deal anyway.

cpu:
> I guess it's only for non-std users? 

Correct ✅

djc:
I guess there's not as much value then, let's keep it like this for now?

cpu:
Going to merge this. We can always separately bump hashbrown if someone feels strongly.

PR #1886, opened 2024-04-02
connect-tests: ignore rsa8192.badssl.com

cpu:
This test server's certificate has expired, breaking the connect tests portion of the daily CI jobs since March 27th. The issue has been flagged with the upstream project (https://github.com/chromium/badssl.com/issues/530, https://github.com/chromium/badssl.com/issues/499#issuecomment-2027555009). Until resolved let's ignore this test.

### DIFF
--- a/connect-tests/tests/badssl.rs
+++ b/connect-tests/tests/badssl.rs
@@ -106,6 +106,7 @@ mod online {
             .unwrap();
     }
 
+    #[ignore] // https://github.com/chromium/badssl.com/issues/530
     #[test]
     fn rsa8192() {
         connect("rsa8192.badssl.com")


cpu:
Here's a [successful run](https://github.com/cpu/rustls/actions/runs/8527781058) that I invoked manually with the workflow dispatch.

ctz (approved):

djc:
I'm confused because I had missed these failures. Do you get notifications? I seem to recall we've encountered this (some people being notified while others aren't) but have forgotten the details.

cpu:
> I'm confused because I had missed these failures. Do you get notifications? I seem to recall we've encountered this (some people being notified while others aren't) but have forgotten the details.

I only get notifications by trying to remember to sync my forks periodically. I reliably seem to get failure emails that way, but it requires manual effort. Otherwise I think GitHub only sends notifications to whoever last edited the workflow.... It's very crummy. 

The best idea I've for a workaround is to configure our repo with a webhook that dumps failures into the Discord in a bot-specific channel. It seems like a big hammer to wield but I don't know if there are good alternatives :-(

cpu:
> Otherwise I think GitHub only sends notifications to whoever last edited the workflow.... 

I dug up [the reference](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs) I found before and it's more nuanced:

> Notifications for scheduled workflows are sent to the user who initially created the workflow. If a different user updates the cron syntax in the workflow file, subsequent notifications will be sent to that user instead. If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax.

cpu:
> Would be nice... but yeah, sounds hard?

I think ( :crossed_fingers: ) it would be pretty easy (a bit of clicking around for someone that has admin privs both here and for the Discord server). The main hesitation I have is that Discord's notification policy is very "on-by-default" and the many folks in the Discord would each have to mute the new bot channel to avoid seeing chatter they might not care about.

cpu:
That's a good idea. Maybe I can try and prototype something with a private repo/discord and then share the steps I used so that you could replicate it w/ the "production" discord/repo.

PR #1899, opened 2024-04-16
Expose connection resumption details

### LINKED ISSUE #705: Ability to tell whether an incoming connection is re-used/resumed
The ability to tell whether an incoming connection is a TLS resumption/re-use and get some details like the Session ID would be handy. We need this in [our FTPS implementation](https://crates.io/crates/libunftp) to match an incoming data channel connection to an active control channel connection to avoid [FTPS connection stealing](https://filezilla-project.org/misc/ftps_connection_stealing_exploit.cpp).

One can get a hold of the server session (ServerConnection):

```rust
let acceptor: TlsAcceptor = tls_config.into();
let stream: TlsStream<TcpStream> = acceptor.accept(socket).await.unwrap();
let server_session: &ServerSession = stream.get_ref().1;
```

and then do things with it like:

```rust
server_session.get_sni_hostname()
```
or 

```rust
server_session.export_keying_material(...)
```

What is needed is something like:

```rust
server_session.get_resumption_info()
```

In the returned result one then needs some identifying information to match the resumed connection to an original connection.

ctz:
My goal with this is to support implementation of [`SSL_session_reused`](https://www.openssl.org/docs/man1.1.1/man3/SSL_session_reused.html) but this should also fix #705.

I'm hoping what and how this works is relatively uncontroversial, but I'm not 100% pleased with the naming.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -26,6 +26,7 @@ use crate::{quic, record_layer};
 /// Connection state common to both client and server connections.
 pub struct CommonState {
     pub(crate) negotiated_version: Option<ProtocolVersion>,
+    pub(crate) handshake_kind: Option<HandshakeKind>,
     pub(crate) side: Side,
     pub(crate) record_layer: record_layer::RecordLayer,
     pub(crate) suite: Option<SupportedCipherSuite>,
@@ -56,6 +57,7 @@ impl CommonState {
     pub(crate) fn new(side: Side) -> Self {
         Self {
             negotiated_version: None,
+            handshake_kind: None,
             side,
             record_layer: record_layer::RecordLayer::new(),
             suite: None,
@@ -138,6 +140,17 @@ impl CommonState {
     /// This returns `None` until the version is agreed.
     pub fn protocol_version(&self) -> Option<ProtocolVersion> {
         self.negotiated_version
+    }
+
+    /// Which kind of handshake was performed.
+    ///
+    /// This tells you whether the handshake was a resumption or not.
+    ///
+    /// This will return `Err(Error::HandshakeNotComplete)` before it is
+    /// known which sort of handshake occurred.
+    pub fn handshake_kind(&self) -> Result<HandshakeKind, Error> {
+        self.handshake_kind
+            .ok_or(Error::HandshakeNotComplete)
     }
 
     pub(crate) fn is_tls13(&self) -> bool {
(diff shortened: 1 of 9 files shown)


djc:
Is there any vocabulary in the spec that we could reuse? I do feel like `HandshakeDetails` is kind of generic/vague.

rustls/src/common_state.rs
```diff
@@ -682,6 +695,32 @@ impl CommonState {
     }
 }
 
+/// Describes which sort of handshake happened.
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum HandshakeDetails {
+    /// A full handshake.
```
cpu, line 701: Maybe worth a little more detail here to match the other variants?

```
/// A full handshake.
///
/// This is the typical TLS connection initiation process when resumption is not yet unavailable, and the initial client hello was accepted by the server.
````

File: rustls/src/common_state.rs [hunk 2]
```diff
@@ -682,6 +695,32 @@ impl CommonState {
...
+    /// A full TLS1.3 handshake, with an extra round-trip for a `HelloRetryRequest`.
+    ///
+    /// The server can respond with a `HelloRetryRequest` if the initial `ClientHello`
+    /// is unacceptable for several reasons, the most likely if no supported key
+    /// shares were offered by the client.
+    FullWithHelloRetryRequest,
+
+    /// A resumed handshake.
+    ///
+    /// XXX: deleteme
```
cpu, line 713: leftover?

rustls/src/common_state.rs
```diff
@@ -682,6 +695,32 @@ impl CommonState {
...
+    /// In TLS1.2 there are two methods this can happen: via session IDs (with state
+    /// kept by the server) or tickets (with state kept by the client).  It is only
```
cpu, line 715: The "it is only known by the server" part of this seems interesting, but since the information isn't exposed by the variant maybe it shouldn't be mentioned here?

rustls/src/common_state.rs
```diff
@@ -682,6 +695,32 @@ impl CommonState {
     }
 }
 
+/// Describes which sort of handshake happened.
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum HandshakeDetails {
```
cpu, line 700: I don't feel strongly, but I wonder if given the current granularity of information if this would be better named `HandshakeKind`?

cpu (approved):
I'm bikeshedding the enum name and suggesting some doc changes but it looks great broadly.

rustls/src/common_state.rs
ctz: Yeah I like `HandshakeKind`, will go with that. The best other option I came up with was `HandshakeShape`, but it's a tongue-twister.

rustls/src/common_state.rs
ctz: Deleted this and replaced it with a more general explanation of what resumption means.

cpu (approved):

rustls/src/common_state.rs
```diff
@@ -140,6 +142,17 @@ impl CommonState {
         self.negotiated_version
     }
 
+    /// Which kind of handshake was performed.
+    ///
+    /// This tells you whether the handshake was a resumption or not.
+    ///
+    /// This will return `Err(Error::HandshakeNotComplete)` before it is
+    /// known which sort of handshake occurred.
+    pub fn handshake_kind(&self) -> Result<HandshakeKind, Error> {
```
djc, line 151: I'm a little surprised at the choice of `Result` here? IIRC most of the similar methods just return `Option`, which seems appropriate here, too?

PR #1902, opened 2024-04-18
MSRV 1.61 -> 1.63

cpu:
We're seeing more of our deps move to this MSRV or higher (e.g. `webpki`, `rustls-platform-verifier`) and it's shipped in Debian stable. Time to move our MSRV to 1.63.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,7 +102,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: "1.61"
+          toolchain: "1.63"
 
       - run: cargo check --locked --lib --all-features -p rustls
 
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ to a wider set of architectures and environments, or compliance requirements.  S
 Specifying `default-features = false` when depending on rustls will remove the
 dependency on aws-lc-rs.
 
-Rustls requires Rust 1.61 or later.
+Rustls requires Rust 1.63 or later.
 
 [ring-target-platforms]: https://github.com/briansmith/ring/blob/2e8363b433fa3b3962c877d9ed2e9145612f3160/include/ring-core/target.h#L18-L64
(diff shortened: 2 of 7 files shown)


djc:
I think we can/should postpone this until we make a rustls-webpki release?

cpu:
> I think we can/should postpone this until we make a rustls-webpki release?

I thought updating to aws-lc-rs 1.7.0 merited waiting, but bumping the MSRV ahead of that seems OK to me.

cpu:
(I'm in no rush either way, just explaining my thinking)

djc:
Ah yeah, we're going to take aws-lc-rs 1.7 here too so we'll need this? That makes sense.

djc (approved):

cpu:
> Ah yeah, we're going to take aws-lc-rs 1.7 here too

Yeah I think it's worthwhile for the `cmake`/platform support improvements.

ctz:
Related question: what do we think about taking _dev_ dependencies that have aggressively very recent MSRVs? I know that is not related to our crate's published MSRV -- it's more about how OK it is to require contributors take toolchain updates.

cpu:
> it's more about how OK it is to require contributors take toolchain updates.

I don't perceive a lot of friction using different toolchains when contributing to projects. It seems like there's an existing expectation developers have access to a nightly toolchain for various small tasks. E.g. in rustls-ffi it's required for `cbindgen` to handle our macro expansion.

ctz (approved):
Let's hold this for the pending 0.23.5 release.

djc:
> Related question: what do we think about taking _dev_ dependencies that have aggressively very recent MSRVs? I know that is not related to our crate's published MSRV -- it's more about how OK it is to require contributors take toolchain updates.

I pretty much exclusively use stable and usually update on the day of release. I don't think it's a big deal? Did you run into it anywhere? I know clap is pretty trigger-happy about MSRV bumps...

cpu:
> Let's hold this for the pending 0.23.5 release.

Going to merge this now that the 0.23.5 release is out.

PR #1903, opened 2024-04-18
docs: update SECURITY example

cpu:
The existing example should be easy enough to understand, but it's also easy enough to update for the current major releases for maximum clarity.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -5,13 +5,20 @@
 Security fixes will be backported only to the rustls versions for which the
 original semver-compatible release was published less than 2 years ago.
 
-For example, as of 2023-06-13 the latest release is 0.21.1.
+For example, as of 2024-04-18 the latest release is 0.23.4.
 
+* 0.23.0 was released in February of 2024
+* 0.22.0 was released in December of 2023
 * 0.21.0 was released in March of 2023
 * 0.20.0 was released in September of 2021
 * 0.19.0 was released in November of 2020
 
-Therefore 0.20.x and 0.21.x will be updated, while 0.19.x will not be.
+Therefore 0.23.x, 0.22.x and 0.21.x will be updated, while 0.20.x and 0.19.x
+will not be.
+
+_Note: We use the date of `crates.io` publication when evaluating the security
+policy. For example, while the Rustls 0.20.0 GitHub release note was created
+Jul, 2023 the actual release in `crates.io` was published in Sept. 2021._
 
 ## Reporting a Vulnerability
 


SECURITY.md
```diff
@@ -5,13 +5,16 @@
 Security fixes will be backported only to the rustls versions for which the
 original semver-compatible release was published less than 2 years ago.
 
-For example, as of 2023-06-13 the latest release is 0.21.1.
+For example, as of 2024-04-18 the latest release is 0.23.4.
 
+* 0.23.0 was released in February of 2024
+* 0.22.0 was released in December of 2023
 * 0.21.0 was released in March of 2023
 * 0.20.0 was released in September of 2021
```
cpu, line 13: In case it comes up: the [GitHub release](https://github.com/rustls/rustls/releases/tag/v%2F0.20.0) for 0.20.0 was created Jul 7, 2023 but the [original release](https://github.com/rustls/rustls/commit/bc754a4fbb586beb7b1dfce38ab880fd90c0e422) was made Sept 26th, 2021. The newer date on the release is an artifact of from back-filling GitHub releases when we switched to using that process. 0.20.x is uncontroversially outside the 2 year support window.

djc: Maybe we should explicitly mention here that we count the crates.io release event, not the GitHub release event?

djc (approved):

SECURITY.md
cpu: Seems worthwhile :+1: I added a bit of warning text.

djc: I feel like the load-bearing part I was hoping for is specifically calling out the crates.io version as acting as the definitive date, which this doesn't do -- on purpose?

cpu: I pushed a revision that mentions the crates.io publication date as the definitive source. Does this version read better to you?

SECURITY.md
```diff
@@ -5,13 +5,20 @@
...
 * 0.19.0 was released in November of 2020
 
-Therefore 0.20.x and 0.21.x will be updated, while 0.19.x will not be.
+Therefore 0.23.x, 0.22.x and 0.21.x will be updated, while 0.20.x and 0.19.x
+will not be.
+
+_Note: We use the date of `crates.io` publication when evaluating the security
+policy. E.g. while the Rustls 0.20.0 GitHub release note was created Jul, 2023
```
djc, line 20: (Sorry for more nits, I tend to avoid `e.g.` and `i.e.` since they're a little archaic and formal and usually be gainfully replaced with something like "For example".)

djc (approved):

SECURITY.md
cpu: Good call, we're already using `"For example" earlier in the text, so this should be consistent. Fixed up.

cpu:
Small docs only update. Going to merge with the one review.

PR #1905, opened 2024-04-19
Correct `complete_io` behaviour when `close_notify` alert is received

ctz:
Release notes:

- This release corrects a denial-of-service condition in `rustls::ConnectionCommon::complete_io`, reachable via network input. If a `close_notify` alert is received during a handshake, `complete_io` did not terminate. Callers which do not call `complete_io` are not affected.
- Add an API (`handshake_kind()`) for learning whether a handshake was resumed or not.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2175,7 +2175,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.4"
+version = "0.23.5"
 dependencies = [
  "aws-lc-rs",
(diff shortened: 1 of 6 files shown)


cpu (approved):

cpu:
> Add an API for learning whether a handshake was resumed or not.

Maybe worth mentioning the new API fn? "Add an API, `handshake_kind()`, for learning whether a handshake was resumed or not."

Issue #1915, opened 2024-04-22
Making impl ClientHelloPayload public ?


dawid-nowak:
In our application we would like to make a decision based on contents of SNI in ClientHelloPayload without necessary affecting the how rustls flow. So we peek some bytes from the stream and initialize Reader and then use Reader to produce ClientHelloPayload. Everything works as expected and it is great that these two are part of the public API. Unfortunately, getting the SNI extension is not as straightforward since those APIs are not part of public API. 

If would be also very helpful if other methods on ClientHelloPayload such as sni_extension, sni_extension, find_extension were added to public API. It seems that they were but now they visibility is only within the crate. 

So the question really is if it makes sense to make ```impl ClientHelloPublic``` part of the public API from rustls perspective?

cpu:
:wave: Hi there,

> In our application we would like to make a decision based on contents of SNI in ClientHelloPayload

Have you looked at using the `Acceptor` API? The [Accepted::client_hello()](https://docs.rs/rustls/latest/rustls/server/struct.Accepted.html#method.client_hello) fn's return type offers a [server_name](https://docs.rs/rustls/latest/rustls/server/struct.ClientHello.html#method.server_name) fn for accesing the SNI value to make a decision. 

If you look in the issue tracker I believe there have been previous discussions about exporting internal message types and this hasn't been something the project has been keen to support. In this case I'm not sure it's necessary at all given the `Acceptor` API.

dawid-nowak:
Thanks for a quick response. I think our use case is slightly different. 
I would like to check if the incoming TCP stream is TLS. If it is TLS then I can pass the stream to rustls for proper processing. If the stream is not then we would like to process it as TCP. 

The main problem with using the Acceptor (I am actually using [LazyConfigAcceptor](https://docs.rs/tokio-rustls/latest/tokio_rustls/struct.LazyConfigAcceptor.html) ) is that we can't get back the consumed bytes. 

That's why I resolved to [peeking](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.peek) bytes from TCP stream and then deserializing ```ClientHelloPayload``` myself. In this respect if I can successfully deserialize ClientHelloPayload and get SNI, I can pass the stream to rustls acceptor and if I can't then I can treat it as a TCP stream.

djc:
We've since added a trait for this in #269, so I'm going to consider this solved.

djc:
I also just submitted #1916, maybe it could help?

[...]

PR #1916, opened 2024-04-23
server: expose buffered bytes from Acceptor

djc:
As a potential help for #1915. Not sure if this is the right solution but seems like a low-cost API to offer.

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -799,6 +799,16 @@ mod connection {
                 sig_schemes,
             }))
         }
+
+        /// Yields an iterator over the buffered chunks of bytes
+        ///
+        /// Will yield an empty iterator after `accept()` has returned `Ok(Some(_))` or `Err(_)`.
+        pub fn buffered(&self) -> &[u8] {
+            match &self.inner {
+                Some(conn) => conn.deframer_buffer.as_slice(),
+                None => &[],
+            }
+        }
     }
 
     /// Represents a TLS alert resulting from handling the client's `ClientHello` message.
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -369,7 +369,7 @@ impl ConnectionRandoms {
 /// Interface shared by client and server connections.
 pub struct ConnectionCommon<Data> {
     pub(crate) core: ConnectionCore<Data>,
-    deframer_buffer: DeframerVecBuffer,
+    pub(crate) deframer_buffer: DeframerVecBuffer,
     sendable_plaintext: ChunkVecBuffer,
 }
 
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -482,6 +482,10 @@ impl DeframerVecBuffer {
     fn unfilled(&mut self) -> &mut [u8] {
         &mut self.buf[self.used..]
     }
+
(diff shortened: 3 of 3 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -799,6 +799,21 @@ mod connection {
...
+        /// Yields an iterator over the buffered chunks of bytes
+        ///
+        /// Will yield an empty iterator after `accept()` has returned `Ok(Some(_))` or `Err(_)`.
+        pub fn buffered(&self) -> impl Iterator<Item = &[u8]> {
+            self.inner
+                .iter()
+                .map(|conn| {
+                    conn.core
+                        .common_state
+                        .sendable_tls
```
ctz, line 812: Is this the right object? If I understood the purpose of this correctly, shouldn't this be `deframer_buffer`?

djc: 🤦 You're right, of course. Still pretty straightforward...

rustls/src/server/server_conn.rs
```diff
@@ -799,6 +799,16 @@ mod connection {
                 sig_schemes,
             }))
         }
+
+        /// Yields an iterator over the buffered chunks of bytes
+        ///
+        /// Will yield an empty iterator after `accept()` has returned `Ok(Some(_))` or `Err(_)`.
```
cpu, line 805: I think this is returning a slice or an empty slice in the current impl, not an iterator. Does this bit of docs need updating?

cpu:
If the goal is to make something like https://github.com/rustls/tokio-rustls/issues/54 easier I don't know that this meaningfully shifts the needle while also introducing more API surface on a core type. 

Maybe it should be something higher-level in the acceptor API specifically for the "is this actually a plaintext connection I'm about to accept?" case?

rustls/src/server/server_conn.rs
djc: Yup, but let's first decide if we think this API is useful.

djc:
> If the goal is to make something like [rustls/tokio-rustls#54](https://github.com/rustls/tokio-rustls/issues/54) easier I don't know that this meaningfully shifts the needle.
> 
> Maybe it should be something higher-level in the acceptor API specifically for the "is this actually a plaintext connection I'm about to accept?" case?

That seems reasonable. So what should that API look like? Maybe `take_plaintext() -> Option<Vec<u8>>`, which can be called before `accept()` and does minimal checking if the buffered bytes look like TLS?

cpu:
> Maybe take_plaintext() -> Option<Vec<u8>>, which can be called before accept() and does minimal checking if the buffered bytes look like TLS?

Yeah! Admittedly I haven't tried sketching anything out myself but this is more along the lines of what I was thinking.

cpu:
@djc WDYT about something [like this approach](https://github.com/rustls/rustls/compare/main...cpu:rustls:cpu-acceptor-plaintext?expand=1) as an alternative?

I tried to come up with an adjustment that wouldn't break semver, so it's a little bit more awkward than it could be if we're open to breaking changes in this area. For e.g. having to wrap the underlying `Error` in an `OtherError` and make callers downcast to interrogate could be avoided by making a new error return for the `accept()` fn, something like `AcceptError(Error, Vec<u8>)`.

cpu:
I think based on the conclusions in #1915 this won't be necessary. I'm going to close it for now.

PR #1921, opened 2024-04-25
Add feature to build for web

Mubelotix:
The build for wasm32-unknown-unknown with ring used to fail, but with this feature it works out of the box

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -34,6 +34,7 @@ logging = ["log"]
 aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs"]
 aws-lc-rs = ["aws_lc_rs"] # Alias because Cargo features commonly use `-`
 ring = ["dep:ring", "webpki/ring"]
+web = ["ring/wasm32_unknown_unknown_js"]
 tls12 = []
 read_buf = ["rustversion", "std"]
 fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1075,8 +1075,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "190092ea657667030ac6a35e305e62fc4dd69fd98ac98631e5d3a2b1575a12b5"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]


rustls/Cargo.toml
```diff
@@ -34,6 +34,7 @@ logging = ["log"]
 aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs"]
 aws-lc-rs = ["aws_lc_rs"] # Alias because Cargo features commonly use `-`
 ring = ["dep:ring", "webpki/ring"]
+web = ["ring/wasm32_unknown_unknown_js"]
```
cpu, line 37: Isn't this something that could be activated outside of rustls by a dependent crate that wanted to use *ring* by taking its own direct dependency?

If a feature for this were to be added here, should it also activate `pki-types/web`?

djc: Also should this be `ring?/wasm32_unknown_unknown_js` to avoid enabling the ring dependency in scenarios where the aws-lc-rs provider is actually selected?

ctz:
I think we are probably overdue for regularising  wasm support in this crate. What I mean by that is:

- a small example for one flavour of wasm or another (probably wasi is easiest?) and thereby document which features, providers, etc. are required.
- CI job that builds & runs that example to make sure we don't regress anything in the future.

cpu:
@Mubelotix Are you interested in trying to achieve the broader changeset Ctz described above?

rustls/Cargo.toml
Mubelotix: > Isn't this something that could be activated outside of rustls by a dependent crate that wanted to use `*ring*` by taking its own direct dependency w/ the feature enabled?
> 
> If a feature for this were to be added here, should it also activate `pki-types/web`?

Yes indeed, they don't have the feature yet but I made a PR, let's wait to see if they merge it

Mubelotix: > Also should this be `ring?/wasm32_unknown_unknown_js` to avoid enabling the ring dependency in scenarios where the aws-lc-rs provider is actually selected?

I thought it would only enable ring if I wrote it that way `web = ["dep:ring", "ring/wasm32_unknown_unknown_js"]` but I will check

Mubelotix:
> @Mubelotix Are you interested in trying to achieve the broader changeset Ctz described above?

Sure. I only have experience with wasm in the browser though, so that's what the example would be about

Mubelotix:
> > I only have experience with wasm in the browser though, so that's what the example would be about
> 
> Personally my understanding of the general ecosystem here is lacking, but I think we have a light preference for WASI over an example fixed to the browser context. I don't know how far the two goals are from one another to gauge if it would be helpful to consider a browser example as a stepping stone towards a more complete WASI example.

Ok I will look into WASI

cpu:
I think it's fair to close this for now while we wait on a reworked version closer to the goals Ctz outlined in [his comment](https://github.com/rustls/rustls/pull/1921#issuecomment-2078940118). Feel free to re-open once you have revisions ready!

PR #1922, opened 2024-04-25
Add support for enforcing CRL expiration

jasperpatterson:
This adds support for enforcing the CRL nextUpdate field (i.e. expiration). The corresponding changes in webpki are here: https://github.com/rustls/webpki/pull/227


cpu:
> rustls / Check semver compatibility (pull_request) Failing after 44s

Just wanted to note IME this job always fails when there's a cargo patch in play, so don't worry about debugging this one.

cpu (commented):
Thank you, this looks good :stars: 

Are you open to adding an integration test for this? There are some existing tests ([starting around here](https://github.com/rustls/rustls/blob/513e374b2e2ce9f1fb57ac78ab3ca053afc8f133/rustls/tests/api.rs#L1254-L1255)) in `api.rs`. 

I don't think it needs to be super extensive. Something like:

* updating [test_ca.rs](https://github.com/rustls/rustls/blob/main/rustls/examples/internal/test_ca.rs) to generate an expired CRL alongside the others
* adding one integration test that uses the default verification behaviour, allowing the expired CRL to be processed without err
* adding one integration test that uses the customized behaviour, verifying the CRL is rejected with the expected err.

jasperpatterson:
@cpu Thanks, I've added in this test! 👍

rustls/examples/internal/test_ca.rs
```diff
@@ -138,12 +141,25 @@ fn crl_for_serial(serial_number: SerialNumber) -> CertificateRevocationListParam
...
+}
+
 // Note: these are ordered such that the data dependencies for issuance are satisfied.
-const ROLES: [Role; 4] = [
+const ROLES: [Role; 5] = [
     Role::TrustAnchor,
     Role::Intermediate,
     Role::EndEntity,
     Role::Client,
+    Role::CrlExpired,
```
cpu, line 162: I think we can probably skip having a `Role` for this, and emit expired CRLs for the non-TrustAnchor roles unconditionally, alongside the exiting `$ROLE..revoked.crl.pem` CRL file as something like `$ROLE.revoked.expired.crl.pem`.

I'm imagining slotting in a new call to your `expired_crl()` fn here-ish: https://github.com/rustls/rustls/blob/ffcc718e23e8f2f8b9aef8035ceb0bd36a7f7f68/rustls/examples/internal/test_ca.rs#L73-L84

Does that make sense? I think it will simplify the `test_ca.rs` diff a good amount.

rustls/examples/internal/test_ca.rs
```diff
@@ -138,12 +141,25 @@ fn crl_for_serial(serial_number: SerialNumber) -> CertificateRevocationListParam
     }
 }
 
+fn expired_crl() -> CertificateRevocationListParams {
+    let now = OffsetDateTime::now_utc();
+    CertificateRevocationListParams {
+        this_update: now,
+        next_update: now + Duration::from_secs(60),
```
cpu, line 148: Ah, I guess I made rcgen smart enough to reject `next_update` values in the past eh? 

I think this is a reasonable workaround but might be worth a comment explaining it.

cpu (commented):
Thanks for adding the integration test. I had a bit of feedback on the details but the overall test logic looks spot on.

rustls/examples/internal/test_ca.rs
jasperpatterson: I like that idea 👍

rustls/examples/internal/test_ca.rs
jasperpatterson: Actually I didn't even try it. I've updated to use a `this_update` value in the past, and a `next_update` value with current time, which worked fine.

rustls/tests/common/mod.rs
```diff
@@ -337,9 +348,9 @@ impl KeyType {
         )
     }
 
-    fn get_crl(&self, role: &str) -> CertificateRevocationListDer<'static> {
+    fn get_crl(&self, role: &str, r#type: &str) -> CertificateRevocationListDer<'static> {
```
cpu, line 351: TIL this syntax works for the case where you want a binding name that overlaps a keyword. Cool :+1:

rustls/Cargo.toml
```diff
@@ -23,7 +23,7 @@ log = { version = "0.4.4", optional = true }
 once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
 ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }
-webpki = { package = "rustls-webpki", version = "0.102.2", features = ["alloc"], default-features = false }
+webpki = { package = "rustls-webpki", version = "0.102.4", features = ["alloc"], default-features = false }
```
cpu, line 26: I think to fix the ["Check minimum versions of direct dependencies"](https://github.com/rustls/rustls/actions/runs/9117850386/job/25070188927) CI failure you'll need a commit like https://github.com/cpu/rustls/commit/0eb3df6a37472559d74314143c08530ae40eebf1 to update the pki-types version :thinking:

cpu (approved):
Looks good to me. Thanks!

Would you mind tidying the commit history? Unless anyone else has an opinion I would vote to do this in two commits: one to land the feature, and one to land the integration test update. It feels worth keeping the two separate in this case since the latter requires a large test-ca diff.

ctz (commented):
LGTM! Could benefit from some squashes, especially the back-and-forth on the testing commits.

cpu (approved):
:rocket:

djc (approved):

ctz (approved):

cpu:
> cpu added this pull request to the merge queue 51 minutes ago

This seems to be jammed up. The [last job](https://github.com/rustls/rustls/actions/runs/9175714619/job/25229290743) (Build+test (stable, ubuntu-latest)) is completed, but spinning. 

I'm going to pull it out and requeue and see if that does the trick.

cpu:
> I'm going to pull it out and requeue and see if that does the trick.

Looks like there's a [broader outage](https://www.githubstatus.com/incidents/nj47vccwm2zj).

PR #1924, opened 2024-04-26
aws-lc-rs: reduce priority of `ECDSA_NISTP521_SHA512`

### LINKED ISSUE #1912: AWS LC fails against golang TLS server while ring works fine
**Checklist**

**Describe the bug**
I've stumbled over a very specific set-up where when using the AWS LC provider, a connection cannot be established while the ring implementation works fine. I'm not 100% sure this is an issue with this crate, but wan't able to trace it down any further.

The set-up: using a server written in golang enforcing TLS 1.2 (I couldn't reproduce the error with a server written in rust with rustls).

**To Reproduce**
The setup is quite large, even though I tried to reduce it as much as I could.

https://github.com/robsdedude/rusttls-ring-vs-aws-issue/tree/4f42dcf51beadce18dc2c44df610a5184bafa0f9

There's plenty more description about the output setup and the observations in the README of the linked repo.

The bottom line is that when using AWS LC, the client fails with
```
Custom { kind: InvalidData, error: InvalidCertificate(BadSignature) }
```
and the server complains
```
remote error: tls: error decrypting message
```

Whereas with `ring` everything works fine. Going through the trace logs, the only difference I could find is that LC offers `ECDSA_NISTP521_SHA512` as `SignatureAlgorithm` while ring doesn't. So I cannot rule out that go's TLS implementation might be at fault here. Sadly I don't know enough TLS details to debug this any further.

ctz:
In TLS1.2, this actually means ECDSA_SHA512.  If the peer selects that, we get caught out depending on the curve of the public key because we don't support (for example) `ECDSA_NISTP256_SHA512`.

Reducing the preference of this improves matters, because a peer that respects our priority will only select that if nothing else is possible (which includes the cases that SHA256 and SHA384 are not supported, in which case we are hosed, but also if the version is TLS1.3 and public key is on P521).

fixes #1912 though unsatisfyingly

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/mod.rs
+++ b/rustls/src/crypto/aws_lc_rs/mod.rs
@@ -169,10 +169,6 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
     mapping: &[
         // Note: for TLS1.2 the curve is not fixed by SignatureScheme. For TLS1.3 it is.
         (
-            SignatureScheme::ECDSA_NISTP521_SHA512,
-            &[webpki_algs::ECDSA_P521_SHA512],
-        ),
-        (
             SignatureScheme::ECDSA_NISTP384_SHA384,
             &[
                 webpki_algs::ECDSA_P384_SHA384,
@@ -185,6 +181,10 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
                 webpki_algs::ECDSA_P256_SHA256,
                 webpki_algs::ECDSA_P384_SHA256,
             ],
+        ),
+        (
+            SignatureScheme::ECDSA_NISTP521_SHA512,
+            &[webpki_algs::ECDSA_P521_SHA512],
         ),
         (SignatureScheme::ED25519, &[webpki_algs::ED25519]),
         (
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -1129,9 +1129,9 @@ fn server_cert_resolve_reduces_sigalgs_for_ecdsa_ciphersuite() {
         CipherSuite::TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
         if provider_is_aws_lc_rs() {
             vec![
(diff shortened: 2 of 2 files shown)


djc:
What is the right fix and why didn't you submit that? 😉

ctz:
> What is the right fix and why didn't you submit that? 😉

1) build a time machine, travel back and tell the tlswg that they shouldn't change the semantics of an existing bit of protocol. I tried this one earlier, but now my house is full of smoke and both my cats are missing?

2) in webpki & aws-lc-rs add support for: `ECDSA_P384_SHA512`, `ECDSA_P256_SHA512`, `ECDSA_P521_SHA256`, & `ECDSA_P521_SHA384`

djc:
> * in webpki & aws-lc-rs add support for: `ECDSA_P384_SHA512`, `ECDSA_P256_SHA512`, `ECDSA_P521_SHA256`, & `ECDSA_P521_SHA384`

Is that hard, or just copying a bunch of constants around?

cpu (approved):
This seems like a better state of affairs than `main`, and if the other realistic fix requires updating two separate crates (one outside of our org) I think it's fair to land this and pursue further improvements separately.

Thanks!

djc (approved):

djc:
So do we want to release a 0.23.6 for this?

cpu:
> So do we want to release a 0.23.6 for this?

I think :+1: but we also need a 0.22.x fix. Support for p521 w/ aws-lc-rs was added in [0.22.2](https://github.com/rustls/rustls/releases/tag/v%2F0.22.2).

samin-cf:
> > * in webpki & aws-lc-rs add support for: `ECDSA_P384_SHA512`, `ECDSA_P256_SHA512`, `ECDSA_P521_SHA256`, & `ECDSA_P521_SHA384`
> 
> Is that hard, or just copying a bunch of constants around?

Is anyone working on implementing the "right" fix?

cpu:
@samin-cf As the name hints the `webpki` crate is principally motivated by the webpki profile, largely defined by the CA/browser forum baseline requirements. Those requirements do not allow the algorithm combinations you're interested in (e.g. see [7.1.3.2.2](https://github.com/cabforum/servercert/blob/main/docs/BR.md#71322-ecdsa)). 

What is your use-case?

PR #1929, opened 2024-04-29
Debug assert to verify the correctness at DeframerVecBuffer

MOZGIII:
Follow up after https://github.com/rustls/rustls/pull/1798#discussion_r1582455822

### DIFF
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -504,7 +504,8 @@ impl DeframerBuffer<'_, InternalPayload> for DeframerVecBuffer {
 
 #[cfg(feature = "std")]
 impl<'a> DeframerBuffer<'a, ExternalPayload<'a>> for DeframerVecBuffer {
-    fn copy(&mut self, payload: &ExternalPayload<'a>, _at: usize) {
+    fn copy(&mut self, payload: &ExternalPayload<'a>, at: usize) {
+        debug_assert_eq!(at, self.used);
         let len = payload.len();
         self.unfilled()[..len].copy_from_slice(payload.0);
         self.advance(len);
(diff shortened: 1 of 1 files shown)


rustls/src/msgs/deframer.rs
```diff
@@ -504,7 +504,8 @@ impl DeframerBuffer<'_, InternalPayload> for DeframerVecBuffer {
 
 #[cfg(feature = "std")]
 impl<'a> DeframerBuffer<'a, ExternalPayload<'a>> for DeframerVecBuffer {
-    fn copy(&mut self, payload: &ExternalPayload<'a>, _at: usize) {
+    fn copy(&mut self, payload: &ExternalPayload<'a>, at: usize) {
+        debug_assert_eq!(at, self.used);
```
cpu, line 508: You [also indicated](https://github.com/rustls/rustls/pull/1798#discussion_r1583331582) you thought some of the context from the fix commit message deserved to be a comment. That sounds reasonable to me, could you roll that into this diff?

MOZGIII: Yes, I was thinking about that too. However it is not trivial how to formulate this context from a commit message explaining why this was changes into a more permanent why is it "ok to be like this" comment for the code.
While thinking about it - I figures that maybe this is actually not an ok thing to have - and a broader remodelling of how this trait and in general part of logic is in order.

If you can agree this is the case - that is what I'll capture and add as a comment.

MOZGIII:
Actually, is this even correct? it seems like the `at` position should just be computed relative to the current level of filling in the buffer - but that adjustment has to be done prior to calling of the `copy` fn...

cpu:
Thanks for the contribution. I chatted about this branch with the other maintainers and we've reached the conclusion we'd like to close this for now. As mentioned in [the other thread](https://github.com/rustls/rustls/pull/1798#discussion_r1583504485) we would be supportive of a broader refactor to address some of the awkward workarounds that remain in this area. Outside of that, if you have remaining concerns about the fix we landed it would be beneficial to have them articulated in a separate issue, ideally with a concrete case in which the existing fix can be demonstrated as insufficient.

MOZGIII:
A quick question here though: in your discussion, did you conclude this invariant check that I have added wouldn't hold? Cause if that's the case then there is likely a corruption possible here that might be exploited as a vulnerability.

cpu:
We believe the invariant holds.

PR #1930, opened 2024-04-29
Small correction to the quic::PacketKey::integrity_limit doc

MOZGIII:
Follow up after https://github.com/rustls/rustls/pull/1798#discussion_r1583104538


cpu (approved):
Thanks!

PR #1944, opened 2024-05-06
Smaller misc changes extracted from client-side ECH branch

cpu:
This branch pulls out a handful of commits I've been carrying at the head of the WIP client-side ECH branch (https://github.com/rustls/rustls/pull/1718). They're all small stand-alone changes that are easy to review/merge ahead of the bulk of the ECH implementation work.

### DIFF
--- a/rustls/src/crypto/hpke.rs
+++ b/rustls/src/crypto/hpke.rs
@@ -26,6 +26,7 @@ pub trait HpkeProvider: Debug + Send + Sync + 'static {
 }
 
 /// An HPKE suite, specifying a key encapsulation mechanism and a symmetric cipher suite.
+#[derive(Clone, Debug)]
 pub struct HpkeSuite {
     /// The choice of HPKE key encapsulation mechanism.
     pub kem: HpkeKem,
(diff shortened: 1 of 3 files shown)


rustls/src/client/builder.rs
```diff
@@ -68,40 +68,6 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
     }
 }
 
-/// Container for unsafe APIs
```
djc, line 71: I prefer the old order... the module itself is `pub(super)` but its contents are `pub` (so I don't find the argument in the commit message convincing), and it fits in the flow of the module in the sense of the method just above it returning the `DangerousClientBuilder` and `DangerousClientBuilder::with_custom_certificate_verifier()` yielding `ConfigBuilder<ClientConfig, WantsClientCert>` where `WantsClientCert` is the item defined immediately below.

rustls/src/crypto/hpke.rs
```diff
@@ -26,6 +26,7 @@ pub trait HpkeProvider: Debug + Send + Sync + 'static {
 }
 
 /// An HPKE suite, specifying a key encapsulation mechanism and a symmetric cipher suite.
+#[derive(Debug, Clone)]
```
djc: Tiny nit: maybe alphabetize these if you're newly adding them?

djc (approved):

rustls/src/crypto/hpke.rs
cpu: Alphabetized the new derives throughout.

ctz (approved):

PR #1948, opened 2024-05-15
bogo: fix config rewriting when cpp is clang

djc:
The previous setup failed for me on macOS, where `cpp` is "Apple clang version 15.0.0 (clang-1500.3.9.4)". I think this approach works with both gcc ("cpp (Gentoo 13.2.1_p20240210 p14) 13.2.1 20240210") and clang.


ctz (approved):
fwiw `clang-cpp-15` worked in both ways for me

djc:
> fwiw `clang-cpp-15` worked in both ways for me

Weird. On my Mac the command errors writes to a file called `-oconfig.json`.

PR #1949, opened 2024-05-15
Warn on lints, don't deny

djc:
We deny warnings in CI (during clippy runs), which seems sufficient. Denying lints is annoying during development especially when working on a release branches (after the lints have gotten more precise).


cpu:
Thanks, I've often resorted to commenting out some deny's during development and then restoring them after and this seems like a better approach :+1:

djc:
In the future we can have workspace-level lint configuration in Cargo.toml that is inherited by each crate, but for now that isn't supported by our MSRV.

PR #1950, opened 2024-05-15
Yield error after receiving CloseNotify with trailer

djc:
Intended to fix https://github.com/rustls/tokio-rustls/issues/72.

cc @quininer @VladimirBramstedt


ctz:
Hm, the duplication of this code for `read_buf` is pretty annoying. Maybe we could hoist that out?

djc:
> Hm, the duplication of this code for `read_buf` is pretty annoying. Maybe we could hoist that out?

Was just working on that. Probably need to add a test. Any suggestions on which test to crib off of/would you be able to take care of that?

VladimirBramstedt:
hmm, is error the correct thing to do? 
the specs for both tls 1.2 and 1.3 state that any data received after closure must be ignored. throwing an error is kind of ignoring it in the way that it isnt reading it, but the alternative is to pretend everything is fine and dandy and really treat is as if we havnt seen any data (i.e discard everything in the deframer and carry on)

ctz:
> would you be able to take care of that?

Sure!

djc:
> hmm, is error the correct thing to do? the specs for both tls 1.2 and 1.3 state that any data received after closure must be ignored. throwing an error is kind of ignoring it in the way that it isnt reading it, but the alternative is to pretend everything is fine and dandy and really treat is as if we havnt seen any data (i.e discard everything in the deframer and carry on)

Do you have references to the appropriate sections? (See #1951 which handles that cleanly for application data.)

VladimirBramstedt:
my interpretation is that data = any bytes after close_notify has been received, regardless of if its a valid ApplicationData frame, or random bytes, as you are not supposed to try and read it... i think.

djc:
Yeah, I think dropping any data that was read after a CloseNotify probably makes sense.

VladimirBramstedt:
we could take a peek at what other implementations (openssl?) deal with this, maybe theres prior art.

ctz:
I think we should ignore the data after the close notify, and return an EOF as if it didn't exist. I'll update the test to that effect in the morning. @VladimirBramstedt are you in a position to see how your application behaves with this branch as it currently stands?

VladimirBramstedt:
i think it should be possible, i'll have to look tomorrow morning (UTC+2) ish.

rustls/tests/api.rs
```diff
@@ -6367,6 +6367,48 @@ fn test_complete_io_with_no_io_needed() {
     );
 }
 
+#[test]
+fn test_junk_after_close_notify_received() {
```
djc, line 6371: Do we need/want a separate test for unbuffered connections?

rustls/src/conn.rs
```diff
@@ -791,6 +791,8 @@ impl<Data> ConnectionCore<Data> {
                 .common_state
                 .has_received_close_notify
             {
+                // "Any data received after a closure alert has been received MUST be ignored."
```
djc, line 794: Feel free to squash. 👍

ctz: Done -- also reworded the commit message.

rustls/tests/api.rs
ctz: Added one of these, the behaviour is different but better.

djc:
LGTM. I think we'll probably want a release for this? Maybe also for 0.21/0.22?

VladimirBramstedt:
tested this out, seems to help, no more excessive wakeups. 👍

ctz:
Great -- thank you for checking that!

ctz (approved):
(fwiw)

Issue #959, opened 2022-01-23
Do we do the right thing when there's data after a `close_notify` alert?


briansmith:
I was reading the code and was struck by this:

```rust
    pub fn reader(&mut self) -> Reader {
        Reader {
            received_plaintext: &mut self.common_state.received_plaintext,
            /// Are we done? i.e., have we processed all received messages, and received a
            /// close_notify to indicate that no new messages will arrive?
            peer_cleanly_closed: self
                .common_state
                .has_received_close_notify
                && !self.message_deframer.has_pending(),
            has_seen_eof: self.common_state.has_seen_eof,
        }
    }
```

Consider the unusual case where we receive in the same TCP packet a valid close_notify Alert followed by an ApplicationData record. Then `has_received_close_notify` will be true but `self.message_deframer.has_pending()` may be true. Then `peer_cleanly_closed` will be false, right?

Now consider this logic later:
```
            match (self.peer_cleanly_closed, self.has_seen_eof) {
                // cleanly closed; don't care about TCP EOF: express this as Ok(0)
                (true, _) => {}
                // unclean closure
                (false, true) => return Err(io::ErrorKind::UnexpectedEof.into()),
                // connection still going, but need more data: signal `WouldBlock` so that
                // the caller knows this
                (false, false) => return Err(io::ErrorKind::WouldBlock.into()),
            }
```

My thinking is that we may hit the `(false, false)` case and return `Err(io::ErrorKind::WouldBlock.into())`. Is that true?

If so, I am not sure that `Err(io::ErrorKind::WouldBlock.into())` would be the right thing to return. I think instead when we receive a `close_notify` alert then we should discard any remaining already-read data after the alert and ensure we never try to read any more (act like we've seen EOF). And, in particular, we should ensure that any subsequent attempt to read data will return EOF instead of returning any data after close_notify.

We should document that it is possible that there might have been data received that wasn't processed after the close_notify, and give advice on what the application should do.

Regardless, I think instead of combining `has_received_close_notify && !self.message_deframer.has_pending()` into a single `peer_cleanly_closed` variable, it would be clearer to instead have the match above handle the triple `(has_received_close_notify, deframer_has_pending, has_seen_eof)`.

If everything is the way we want it to be, then at the very least we should add some comments to explain this.

briansmith:
In particular, the public documentation has a different definition of a "clean close":

```
    /// If the peer closes the TLS session cleanly, this returns `Ok(0)`  once all
[...]

PR #1951, opened 2024-05-15
Drop application data received after `close_notify`

ctz:
This addresses #959 

Note this is different to #1950 -- this is ignoring data that was received & decrypted successfully, but came after a `close_notify` and so should not have been sent by the peer. We ignore that, as required by the RFC (weird choice, since it indicates a defect on the part of the peer).

#1950 is fixing a fragment of a message after a `close_notify`, but one that has not yet (or may never) be fully received & decrypted.

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -6367,6 +6367,31 @@ fn test_complete_io_with_no_io_needed() {
     );
 }
 
+#[test]
+fn test_data_after_close_notify_is_ignored() {
+    let (mut client, mut server) = make_pair(KeyType::Rsa2048);
+    do_handshake(&mut client, &mut server);
+
+    client
+        .writer()
+        .write_all(b"before")
+        .unwrap();
+    client.send_close_notify();
+    client
+        .writer()
+        .write_all(b"after")
+        .unwrap();
+    transfer(&mut client, &mut server);
+    server.process_new_packets().unwrap();
+    let mut output = [0u8; 10];
+    let count = server
+        .reader()
+        .read(&mut output)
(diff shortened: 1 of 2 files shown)


rustls/tests/api.rs
```diff
@@ -6367,6 +6367,31 @@ fn test_complete_io_with_no_io_needed() {
...
+        .unwrap();
+    transfer(&mut client, &mut server);
+    server.process_new_packets().unwrap();
+    let mut output = [0u8; 10];
+    let count = server
+        .reader()
+        .read(&mut output)
+        .unwrap();
+    assert_eq!(&output[..count], b"before");
+    assert_eq!(Some(0), server.reader().read(&mut output).ok());
```
djc, line 6392: I forget whether we've discussed this, but I prefer to order the `assert_eq!()` arguments the other way around (and I think that would be more idiomatic).

djc (approved):
This takes care of application data, but it doesn't address the case where an incomplete TLS record is stuck in the deframer buffer after the CloseNotify, right?

ctz:
> This takes care of application data, but it doesn't address the case where an incomplete TLS record is stuck in the deframer buffer after the CloseNotify, right?

Yes, exactly.

djc:
So do we still need this if we're going to merge (something like) #1950? I don't think we need it? Maybe we should only merge the test, or do you want to keep this as a kind of defense in depth?

cpu:
> This pull request is closed, but the jbp-ignore-data-after-close_notify branch has unmerged commits.

Could this branch be deleted or did you want to keep it around for a bit?

Issue #1918, opened 2024-04-24
Verify that SigningKey matches public key within certificate


lvkv:
👋 Hello!

**Checklist**
* [X] I've searched the issue tracker for similar requests

Apologies if this has already been asked and I didn't notice!

**Is your feature request related to a problem? Please describe.**

Right now, it looks like it's possible to create a `CertifiedKey` with inconsistent public and private keys, and there doesn't seem to be any functionality out of the box to check if this is the case.

**Describe the solution you'd like**

Ideally, I'm looking for a Rustls-compatible equivalent for OpenSSL's `X509_check_private_key`, which verifies that the given public and private keys are consistent [^cpk]:

> DESCRIPTION
X509_check_private_key() function checks the consistency of private key pkey with the public key in cert.

This would be great even it functions in the same way that `X509_check_private_key` does:

>BUGS
The X509_check_private_key() and X509_REQ_check_private_key() functions do not check if pkey itself is indeed a private key or not. They merely compare the public materials (e.g., exponent and modulus of an RSA key) and/or key parameters (e.g. EC params of an EC key) of a key pair. So they also return success if pkey is a matching public key.

**Describe alternatives you've considered**
I suppose I could write something myself that does some ASN.1/DER parsing. I feel like this is useful to have!
An external crate that has this functionality would also be great!

**Additional context**

[^cpk]: https://www.openssl.org/docs/manmaster/man3/X509_check_private_key.html

cpu:
:wave: Hi there, thanks for opening an issue

I think we're open to adding functionality for this. Previously (https://github.com/rustls/webpki/pull/67) I had started implementing something similar in `webpki` but the [conclusion](https://github.com/rustls/webpki/pull/67#discussion_r1226643795) at the time was that it'd be a better fit in rustls with a slightly different design. I haven't had time to come back around to that but it's been on my mind because we stubbed out `X509_check_private_key` with a TODO in some ongoing OpenSSL compatibility work.

Would you be interested in working on a PR?

lvkv:
> Would you be interested in working on a PR?

Yeah! I still have a lot to learn in this space, though. Are you guys are okay with nurturing a newbie?

Naively, the interface I would want is one of:

* A method on `CertifiedKey` that verifies consistency between the `CertifiedKey`'s underlying `end_entity_cert` (of type `CertificateDer`) and `key` (`Arc<dyn SigningKey>`):

    ```rust
    /// I think this might not work, though. See the paragraph below.
[...]

PR #1954, opened 2024-05-17
Add function to verify CertifiedKey consistency

lvkv:
Using the `subject_public_key_info()` function we added in https://github.com/rustls/webpki/pull/253, this PR:

* Adds a function to `trait SigningKey` that returns its `SubjectPublicKeyInfo`, but only if the implementer opts in
* Adds a function to `CertifiedKey` that verifies the consistency of its public and private keys by comparing the SPKI values of its end-entity cert and its key

The motivation behind this is to make it possible to verify consistency for public and private keys (https://github.com/rustls/rustls/issues/1918).

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -26,7 +26,7 @@ log = { version = "0.4.4", optional = true }
 once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
 ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }
-webpki = { package = "rustls-webpki", version = "0.102.4", features = ["alloc"], default-features = false }
+webpki = { package = "rustls-webpki", version = "0.102.5", features = ["alloc"], default-features = false }
 pki-types = { package = "rustls-pki-types", version = "1.7", features = ["alloc"] }
 zeroize = "1.7"
 zlib-rs = { version = "0.2", optional = true }
--- a/rustls/src/crypto/signer.rs
+++ b/rustls/src/crypto/signer.rs
@@ -3,10 +3,11 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 
-use pki_types::CertificateDer;
+use pki_types::{CertificateDer, SubjectPublicKeyInfoDer};
 
 use crate::enums::{SignatureAlgorithm, SignatureScheme};
-use crate::error::Error;
+use crate::error::{Error, InconsistentKeys};
+use crate::server::ParsedCertificate;
 
 /// An abstract signing key.
 ///
@@ -59,6 +60,12 @@ pub trait SigningKey: Debug + Send + Sync {
     /// using the chosen scheme.
     fn choose_scheme(&self, offered: &[SignatureScheme]) -> Option<Box<dyn Signer>>;
 
+    /// Get the RFC 5280-compliant SubjectPublicKeyInfo (SPKI) of this [`SigningKey`] if available.
+    fn public_key(&self) -> Option<SubjectPublicKeyInfoDer<'_>> {
+        // Opt-out by default
+        None
+    }
+
     /// What kind of key we have.
     fn algorithm(&self) -> SignatureAlgorithm;
 }
@@ -102,6 +109,21 @@ impl CertifiedKey {
             cert,
             key,
             ocsp: None,
+        }
+    }
+
+    /// Verify the consistency of this [`CertifiedKey`]'s public and private keys.
+    /// This is done by performing a comparison of SubjectPublicKeyInfo bytes.
+    pub fn keys_match(&self) -> Result<(), Error> {
+        let key_spki = match self.key.public_key() {
+            Some(key) => key,
+            None => return Err(InconsistentKeys::Unknown.into()),
+        };
+
+        let cert = ParsedCertificate::try_from(self.end_entity_cert()?)?;
+        match key_spki == cert.subject_public_key_info() {
+            true => Ok(()),
+            false => Err(InconsistentKeys::KeyMismatch.into()),
         }
     }
 
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -97,6 +97,11 @@ pub enum Error {
     /// or too large.
     BadMaxFragmentSize,
 
+    /// Specific failure cases from [`keys_match`].
+    ///
+    /// [`keys_match`]: crate::crypto::signer::CertifiedKey::keys_match
+    InconsistentKeys(InconsistentKeys),
+
     /// Any other error.
     ///
     /// This variant should only be used when the error is not better described by a more
@@ -105,6 +110,30 @@ pub enum Error {
     ///
     /// Enums holding this variant will never compare equal to each other.
     Other(OtherError),
+}
+
+/// Specific failure cases from [`keys_match`].
+///
+/// [`keys_match`]: crate::crypto::signer::CertifiedKey::keys_match
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum InconsistentKeys {
+    /// The public key returned by the [`SigningKey`] does not match the public key information in the certificate.
+    ///
+    /// [`SigningKey`]: crate::crypto::signer::SigningKey
+    KeyMismatch,
+
+    /// The [`SigningKey`] cannot produce its corresponding public key.
+    ///
+    /// [`SigningKey`]: crate::crypto::signer::SigningKey
+    Unknown,
+}
+
+impl From<InconsistentKeys> for Error {
+    #[inline]
+    fn from(e: InconsistentKeys) -> Self {
+        Self::InconsistentKeys(e)
+    }
 }
 
 /// A corrupt TLS message payload that resulted in an error.
@@ -563,6 +592,9 @@ impl fmt::Display for Error {
             Self::BadMaxFragmentSize => {
                 write!(f, "the supplied max_fragment_size was too small or large")
             }
+            Self::InconsistentKeys(ref why) => {
+                write!(f, "keys may not be consistent: {:?}", why)
+            }
             Self::General(ref err) => write!(f, "unexpected error: {}", err),
             Self::Other(ref err) => write!(f, "other error: {}", err),
         }
@@ -644,7 +676,7 @@ mod tests {
     use std::prelude::v1::*;
     use std::{println, vec};
 
-    use super::{Error, InvalidMessage};
+    use super::{Error, InconsistentKeys, InvalidMessage};
     use crate::error::{CertRevocationListError, OtherError};
 
     #[test]
@@ -731,6 +763,8 @@ mod tests {
             Error::PeerSentOversizedRecord,
             Error::NoApplicationProtocol,
             Error::BadMaxFragmentSize,
+            Error::InconsistentKeys(InconsistentKeys::KeyMismatch),
+            Error::InconsistentKeys(InconsistentKeys::Unknown),
             Error::InvalidCertRevocationList(CertRevocationListError::BadSignature),
             Error::Other(OtherError(
                 #[cfg(feature = "std")]
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -516,8 +516,8 @@ pub use crate::enums::{
     ProtocolVersion, SignatureAlgorithm, SignatureScheme,
 };
 pub use crate::error::{
-    CertRevocationListError, CertificateError, EncryptedClientHelloError, Error, InvalidMessage,
-    OtherError, PeerIncompatible, PeerMisbehaved,
+    CertRevocationListError, CertificateError, EncryptedClientHelloError, Error, InconsistentKeys,
+    InvalidMessage, OtherError, PeerIncompatible, PeerMisbehaved,
 };
 pub use crate::key_log::{KeyLog, NoKeyLog};
 #[cfg(feature = "std")]
--- a/rustls/src/webpki/verify.rs
+++ b/rustls/src/webpki/verify.rs
@@ -1,7 +1,9 @@
 use alloc::vec::Vec;
 use core::fmt;
 
-use pki_types::{CertificateDer, ServerName, SignatureVerificationAlgorithm, UnixTime};
+use pki_types::{
+    CertificateDer, ServerName, SignatureVerificationAlgorithm, SubjectPublicKeyInfoDer, UnixTime,
+};
 
 use super::anchors::RootCertStore;
 use super::pki_error;
@@ -124,6 +126,13 @@ impl fmt::Debug for WebPkiSupportedAlgorithms {
 ///
 /// This is used in order to avoid parsing twice when specifying custom verification
 pub struct ParsedCertificate<'a>(pub(crate) webpki::EndEntityCert<'a>);
+
+impl<'a> ParsedCertificate<'a> {
+    /// Get the parsed certificate's SubjectPublicKeyInfo (SPKI)
+    pub fn subject_public_key_info(&self) -> SubjectPublicKeyInfoDer<'static> {
+        self.0.subject_public_key_info()
+    }
+}
 
 impl<'a> TryFrom<&'a CertificateDer<'a>> for ParsedCertificate<'a> {
     type Error = Error;
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -34,9 +34,9 @@ use rustls::{
 use rustls::{
     sign, AlertDescription, CertificateError, CipherSuite, ClientConfig, ClientConnection,
     ConnectionCommon, ConnectionTrafficSecrets, ContentType, DistinguishedName, Error,
-    HandshakeKind, HandshakeType, InvalidMessage, KeyLog, NamedGroup, PeerIncompatible,
-    PeerMisbehaved, ProtocolVersion, ServerConfig, ServerConnection, SideData, SignatureScheme,
-    Stream, StreamOwned, SupportedCipherSuite,
+    HandshakeKind, HandshakeType, InconsistentKeys, InvalidMessage, KeyLog, NamedGroup,
+    PeerIncompatible, PeerMisbehaved, ProtocolVersion, ServerConfig, ServerConnection, SideData,
+    SignatureScheme, Stream, StreamOwned, SupportedCipherSuite,
 };
 
 use super::*;
@@ -2962,6 +2962,67 @@ fn sni_resolver_rejects_bad_certs() {
             sign::CertifiedKey::new(bad_chain, signing_key.clone())
         )
     );
+}
+
+#[test]
+fn test_keys_match() {
+    // Consistent: Both of these should have the same SPKI values
+    let expect_consistent =
+        sign::CertifiedKey::new(KeyType::Rsa2048.get_chain(), Arc::new(SigningKeySomeSpki));
+    assert!(matches!(expect_consistent.keys_match(), Ok(())));
+
+    // Inconsistent: These should not have the same SPKI values
+    let expect_inconsistent =
+        sign::CertifiedKey::new(KeyType::EcdsaP256.get_chain(), Arc::new(SigningKeySomeSpki));
+    assert!(matches!(
+        expect_inconsistent.keys_match(),
+        Err(Error::InconsistentKeys(InconsistentKeys::KeyMismatch))
+    ));
+
+    // Unknown: This signing key returns None for its SPKI, so we can't tell if the certified key is consistent
+    let expect_unknown =
+        sign::CertifiedKey::new(KeyType::Rsa2048.get_chain(), Arc::new(SigningKeyNoneSpki));
+    assert!(matches!(
+        expect_unknown.keys_match(),
+        Err(Error::InconsistentKeys(InconsistentKeys::Unknown))
+    ));
+}
+
+/// Represents a SigningKey that returns None for its SPKI via the default impl.
+#[derive(Debug)]
+struct SigningKeyNoneSpki;
+
+impl sign::SigningKey for SigningKeyNoneSpki {
+    fn choose_scheme(&self, _offered: &[SignatureScheme]) -> Option<Box<dyn sign::Signer>> {
+        unimplemented!("Not meant to be called during tests")
+    }
+
+    fn algorithm(&self) -> rustls::SignatureAlgorithm {
+        unimplemented!("Not meant to be called during tests")
+    }
+}
+
+/// Represents a SigningKey that returns Some for its SPKI.
+#[derive(Debug)]
+struct SigningKeySomeSpki;
+
+impl sign::SigningKey for SigningKeySomeSpki {
+    fn public_key(&self) -> Option<pki_types::SubjectPublicKeyInfoDer> {
+        let chain = KeyType::Rsa2048.get_chain();
+        let cert = ParsedCertificate::try_from(chain.first().unwrap()).unwrap();
+        Some(
+            cert.subject_public_key_info()
+                .into_owned(),
+        )
+    }
+
+    fn choose_scheme(&self, _offered: &[SignatureScheme]) -> Option<Box<dyn sign::Signer>> {
+        unimplemented!("Not meant to be called during tests")
+    }
+
+    fn algorithm(&self) -> rustls::SignatureAlgorithm {
+        unimplemented!("Not meant to be called during tests")
+    }
 }
 
 fn do_exporter_test(client_config: ClientConfig, server_config: ServerConfig) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2228,7 +2228,7 @@ dependencies = [
  "ring",
  "rustls-pemfile 2.1.2",
  "rustls-pki-types",
- "rustls-webpki 0.102.4",
+ "rustls-webpki 0.102.5",
  "rustversion",
  "serde",
  "serde_json",
@@ -2355,7 +2355,7 @@ dependencies = [
  "rsa",
  "rustls 0.23.10",
  "rustls-pki-types",
- "rustls-webpki 0.102.4",
+ "rustls-webpki 0.102.5",
  "sha2",
  "signature",
  "webpki-roots 0.26.3",
@@ -2385,9 +2385,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.102.4"
+version = "0.102.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ff448f7e92e913c4b7d4c6d8e4540a1724b319b4152b8aef6d4cf8339712b33e"
+checksum = "f9a6fccd794a42c2c105b513a2f62bc3fd8f3ba57a4593677ceb0bd035164d78"
 dependencies = [
  "aws-lc-rs",
  "ring",
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -401,9 +401,9 @@ checksum = "976295e77ce332211c0d24d92c0e83e50f5c5f046d11082cea19f3df13a3562d"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.102.4"
+version = "0.102.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ff448f7e92e913c4b7d4c6d8e4540a1724b319b4152b8aef6d4cf8339712b33e"
+checksum = "f9a6fccd794a42c2c105b513a2f62bc3fd8f3ba57a4593677ceb0bd035164d78"
 dependencies = [
  "aws-lc-rs",
  "ring",


rustls/src/error.rs
```diff
@@ -94,6 +94,15 @@ pub enum Error {
     /// or too large.
     BadMaxFragmentSize,
 
+    /// This [`rustls::crypto::signer::SigningKey`] has not implemented [`rustls::crypto::signer::SigningKey::public_key()`].
+    UnimplementedSubjectPublicKeyInfo,
+
+    /// We failed to parse DER into an end-entity certificate.
+    FailedToParseEndEntityCert(webpki::Error),
```
ctz, line 101: We avoid having `webpki` types in our public interface. Please reuse `Error::InvalidCertificate` here.

rustls/src/error.rs
```diff
@@ -94,6 +94,15 @@ pub enum Error {
...
+
+    /// We could not find a key's SubjectPublicKeyInfo.
+    KeyDoesNotContainSubjectPublicKeyInfo,
```
ctz, line 104: Not completely sure of the distinction between this and `UnimplementedSubjectPublicKeyInfo`?

rustls/src/crypto/signer.rs
```diff
@@ -61,6 +62,12 @@ pub trait SigningKey: Debug + Send + Sync {
 
     /// What kind of key we have.
     fn algorithm(&self) -> SignatureAlgorithm;
+
+    /// Get the RFC 5280-compliant SubjectPublicKeyInfo (SPKI) of this [`SigningKey`] if it exists.
+    fn public_key(&self) -> Result<Option<Vec<u8>>, Error> {
+        // Opt-out by default
+        Err(Error::UnimplementedSubjectPublicKeyInfo)
```
ctz, line 69: I would say `Ok(None)`?

(I was thinking we'd automatically do this checking, in places where we accept a `CertifiedKey` such as `ResolvesServerCertUsingSni::add` for example. Having the `Option` there codifies the opt-out behaviour, and allows higher levels to avoid raising and then quenching errors.)

cpu, line 69: I think the branch checked out at this commit won't build because the new `UnimplementedSubjectPublicKeyInfo` error isn't added until the subsequent commit.

Rather than fix that, WDYT about using `Error::General("public_key() unimplemented".into())`? I'm not sure this case merits a dedicated error type.

rustls/src/crypto/signer.rs
```diff
@@ -111,4 +118,16 @@ impl CertifiedKey {
             .first()
             .ok_or(Error::NoCertificatesPresented)
     }
+
+    /// Verify the consistency of this [`CertifiedKey`]'s public and private keys.
+    /// This is done by performing a comparison of SubjectPublicKeyInfo bytes.
+    pub fn certificate_and_key_are_consistent(&self) -> Result<bool, Error> {
+        Ok(*EndEntityCert::try_from(self.end_entity_cert()?)
+            .map_err(Error::FailedToParseEndEntityCert)?
```
ctz, line 126: Suggest the `ParsedCertificate` wrapper to avoid webpki dependencies leaking out of the `crate::webpki` module.

rustls/src/crypto/signer.rs
```diff
@@ -61,6 +62,12 @@ pub trait SigningKey: Debug + Send + Sync {
 
     /// What kind of key we have.
     fn algorithm(&self) -> SignatureAlgorithm;
+
+    /// Get the RFC 5280-compliant SubjectPublicKeyInfo (SPKI) of this [`SigningKey`] if it exists.
+    fn public_key(&self) -> Result<Option<Vec<u8>>, Error> {
```
cpu, line 67: Is there a reason to imagine that someone might implement this to return `Ok(None)`? It feels like `Result<Vec<u8>, Error>` might be sufficient and make this a little more ergonomic for a caller.

rustls/src/error.rs
cpu: It feels like `Error::InvalidCertificate` would be appropriate for this case, the SPKI is mandatory in the `TBSCertificate` per RFC 5280. It's absence indicates an invalid cert.

rustls/src/crypto/signer.rs
cpu: > Rather than fix that, WDYT about using Error::General("public_key() unimplemented".into())? I'm not sure this case merits a dedicated error type.

I like [Ctz's suggestion](https://github.com/rustls/rustls/pull/1954#discussion_r1605018797) to use `Ok(None)` here instead.

rustls/src/crypto/signer.rs
cpu: Can disregard this based on the idea of using `Ok(None)` for the unimplemented case.

cpu:
Realized I was reviewing at the same time as Ctz so I submitted my partial review :-) I agree with his feedback.

cpu:
Some (hopefully helpful) pointers for the failed CI tasks:

* [Check for documentation errors](https://github.com/rustls/rustls/actions/runs/9121777381/job/25081683478?pr=1954): I think you need to use `crate::` instead of `rustls::` in the rustdoc for `UnimplementedSubjectPublicKeyInfo` (assuming that error type doesn't fall away based on other feedback)
* [Check for minimum versions of direct dependencies](https://github.com/rustls/rustls/actions/runs/9121777381/job/25081684095?pr=1954) - I think you need to bump the pki-types version alongside the webpki update (something like https://github.com/cpu/rustls/commit/0eb3df6a37472559d74314143c08530ae40eebf1).
* [Clippy](https://github.com/rustls/rustls/actions/runs/9121777381/job/25081682592?pr=1954) - The fuzz crate has a separate `Cargo.lock` that needs updating. You should be able to cd into `fuzz`, run `cargo build` and then commit the updated lockfile.
* [Clippy (nightly)](https://github.com/rustls/rustls/actions/runs/9121777381/job/25081682973?pr=1954) - ditto above.
* [Validate external types appearing in public API](https://github.com/rustls/rustls/actions/runs/9121777381/job/25081682304?pr=1954) - this will be resolved with [ctz's suggestion](https://github.com/rustls/rustls/pull/1954#discussion_r1605021433).

rustls/src/crypto/signer.rs
djc: Shouldn't this be returning SubjectPublicKeyInfoDer?

cpu: Yes, agreed. Good call-out. Thanks!

rustls/src/error.rs
lvkv: Gotcha. I replaced the code that necessitated this error in the first place with `ParsedCertificate` in [38dd19e](https://github.com/rustls/rustls/pull/1954/commits/38dd19e0ea9471274a9cf67c06b0aad494966a49)

rustls/src/error.rs
lvkv: > It feels like Error::InvalidCertificate would be appropriate for this case, the SPKI is mandatory in the TBSCertificate per RFC 5280. It's absence indicates an invalid cert.

This makes sense. I removed the error and replaced the code that necessitated it with `Error::InvalidCertificate(CertificateError::BadEncoding)`, which aligns more with this perspective

rustls/src/crypto/signer.rs
lvkv: Makes sense. Changed the default return value to `Ok(None)` in https://github.com/rustls/rustls/pull/1954/commits/38dd19e0ea9471274a9cf67c06b0aad494966a49

rustls/src/crypto/signer.rs
lvkv: Thanks for the suggestion! I added this in https://github.com/rustls/rustls/pull/1954/commits/38dd19e0ea9471274a9cf67c06b0aad494966a49

rustls/src/crypto/signer.rs
lvkv: I updated the return type to `Result<Option<SubjectPublicKeyInfoDer>, Error>` in https://github.com/rustls/rustls/commit/38dd19e0ea9471274a9cf67c06b0aad494966a49. I assume a `pki-types` type is acceptable for a public API in rustls? `cargo check-external-types` isn't working locally for me yet

lvkv:
I'm converting this to a "ready for review" PR, as it's less of a draft now. Questions:

* It looks like we've agreed to return `Ok(None)` by default for `SigningKey::public_key`. This looks sensible, though it means that `certificate_and_key_are_consistent` will (as this PR is written) always return `Error::InvalidCertificate(CertificateError::BadEncoding)`, since 1. it interprets the absence of a SPKI as an invalid cert and 2. nobody has overridden the default behavior of `certificate_and_key_are_consistent` yet. I assume the solution is to update the `SigningKey` implementers in additional PR(s) before putting this current change in a release?
*  @djc / @cpu suggested that we updated the return type of `SigningKey::public_key` from `Result<Option<Vec<u8>>, Error>` to `Result<Option<SubjectPublicKeyInfoDer>, Error>` over at https://github.com/rustls/rustls/pull/1954#discussion_r1605019692. Does this count as an external type? I'm still poking `cargo check-external-types`, which I haven't gotten to work locally yet 🙂

rustls/src/crypto/signer.rs
djc: Yup, pki-types are fine in the public API.

rustls/src/crypto/signer.rs
```diff
@@ -61,6 +62,12 @@ pub trait SigningKey: Debug + Send + Sync {
 
     /// What kind of key we have.
     fn algorithm(&self) -> SignatureAlgorithm;
+
+    /// Get the RFC 5280-compliant SubjectPublicKeyInfo (SPKI) of this [`SigningKey`] if it exists.
+    fn public_key(&self) -> Result<Option<SubjectPublicKeyInfoDer>, Error> {
```
djc, line 67: Nit: per the ordering explained in `CONTRIBUTING.md`, I think this should go between `choose_scheme()` and `algorithm()` (here and in any impls).

rustls/src/crypto/signer.rs
```diff
@@ -111,4 +118,16 @@ impl CertifiedKey {
             .first()
             .ok_or(Error::NoCertificatesPresented)
     }
+
+    /// Verify the consistency of this [`CertifiedKey`]'s public and private keys.
+    /// This is done by performing a comparison of SubjectPublicKeyInfo bytes.
+    pub fn certificate_and_key_are_consistent(&self) -> Result<bool, Error> {
```
djc, line 124: Nit: this should be moved before `end_entity_cert()`.

This name also seems pretty verbose? Maybe `keys_match()` or `matching_keys()`?

rustls/src/webpki/verify.rs
```diff
@@ -134,6 +135,15 @@ impl<'a> TryFrom<&'a CertificateDer<'a>> for ParsedCertificate<'a> {
     }
 }
 
+/// So that [`ParsedCertificate`] may transparently behave like a [`webpki::EndEntityCert`].
+impl<'a> Deref for ParsedCertificate<'a> {
+    type Target = webpki::EndEntityCert<'a>;
```
djc, line 140: Unfortunately, this still counts as a webpi type appearing in the public API. If we're only relying on this to get access to `subject_public_key_info()` I'd recommend just adding an inherent impl which forwards that method to the webpki method of the same name.

rustls/src/crypto/signer.rs
```diff
@@ -111,4 +118,16 @@ impl CertifiedKey {
...
+        Ok(
```
djc, line 125: This formulation is pretty dense. Even though we often avoid single-use bindings, I think it would be nice to structure this like:

```rust
let public_spki = ParsedCertificate::try_from(self.end_entity_cert()?)?.subject_public_key_info();
let private_spki = self.key.public_key()?.ok_or(..)?;
Ok(public_spki == private_spki)
```

rustls/src/crypto/signer.rs
cpu: In case you're interested, the blessed types are specified in the rustls crate `Cargo.toml`: https://github.com/rustls/rustls/blob/c57079361b410bfbdae5f70913d8c89de2e71ed3/rustls/Cargo.toml#L79-L83

cpu:
> It looks like we've agreed to return Ok(None) by default for SigningKey::public_key. This looks sensible, though it means that certificate_and_key_are_consistent will (as this PR is written) always return Error::InvalidCertificate(CertificateError::BadEncoding), since 1. it interprets the absence of a SPKI as an invalid cert and 2. nobody has overridden the default behavior of certificate_and_key_are_consistent yet. I assume the solution is to update the SigningKey implementers in additional PR(s) before putting this current change in a release?

Hmm, yeah, it does seem a little off that if `SigningKey::public_key`'s default impl isn't overridden the new API produces a `BadEncoding` error. I could see that being confusing. I suggested the bad encoding error thinking it was used only when the SPKI was missing from a cert, but didn't think about he case where the `SigningKey` wasn't able to provide the information. 

Perhaps it would be better to formulate `certificate_and_key_are_consistent()` (or whatever its name ends up being) to be infallible. Instead of returning a `Result<bool>`, maybe we could return an enum with variants like `Consistent`, `Inconsistent`, `Unknown`. We'd return `Unknown` if the SPKI can't be produced for either the key or the cert, `Inconsistent` if we compared SPKIs and they weren't byte-for-byte identical, and `Consistent` if they were.

My thinking is that even when we implement `public_key()` for the aws-lc-rs and ring singing keys there is still the possibility of external implementations of the trait not having caught up yet. I think treating that as an error instead of a variant like `Unknown` might be too heavy-weight given I see this more of as a helper API than load-bearing. WDYT?

rustls/Cargo.toml
```diff
@@ -23,8 +23,8 @@ log = { version = "0.4.4", optional = true }
 once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
 ring = { version = "0.17", optional = true }
 subtle = { version = "2.5.0", default-features = false }
-webpki = { package = "rustls-webpki", version = "0.102.2", features = ["alloc"], default-features = false }
-pki-types = { package = "rustls-pki-types", version = "1.2", features = ["alloc"] }
+webpki = { package = "rustls-webpki", version = "0.102.4", features = ["alloc"], default-features = false }
+pki-types = { package = "rustls-pki-types", version = "1.7", features = ["alloc"] }
```
cpu, line 27: Since https://github.com/rustls/rustls/pull/1922 was merged first I think you can drop this part of your diff now :+1:

rustls/src/crypto/signer.rs
lvkv: Reordered this in [b3577ab](https://github.com/rustls/rustls/pull/1954/commits/b3577abc6b66c77c0e5024adcd8248f8f70115f7)!

rustls/src/crypto/signer.rs
lvkv: Moved and renamed to `keys_match()` in [b3577ab](https://github.com/rustls/rustls/pull/1954/commits/b3577abc6b66c77c0e5024adcd8248f8f70115f7)!

rustls/src/webpki/verify.rs
lvkv: That makes sense. I've opted to simply remove the impl in [b3577ab](https://github.com/rustls/rustls/pull/1954/commits/b3577abc6b66c77c0e5024adcd8248f8f70115f7) and access the inner cert via `.0` as done elsewhere in the repo

rustls/src/crypto/signer.rs
lvkv: I rewrote this taking into account @cpu's advice in [b3577ab](https://github.com/rustls/rustls/pull/1954/commits/b3577abc6b66c77c0e5024adcd8248f8f70115f7). The suggestion above doesn't exactly work because a temporary value is dropped whilst being borrowed, but it should look more simple nonetheless

lvkv:
> Perhaps it would be better to formulate certificate_and_key_are_consistent() (or whatever its name ends up being) to be infallible. Instead of returning a Result<bool>, maybe we could return an enum with variants like Consistent, Inconsistent, Unknown. We'd return Unknown if the SPKI can't be produced for either the key or the cert, Inconsistent if we compared SPKIs and they weren't byte-for-byte identical, and Consistent if they were.

@cpu, I took a stab at rewriting the `keys_match()` function with this in mind. The "unknown" in particular feels a bit goofy to me, like Rustls is a magic 8 ball that wants me to "ask again later" (i.e. when we've gone around and implemented `public_key` on all the `SigningKey`s) 🙂 

> My thinking is that even when we implement public_key() for the aws-lc-rs and ring signing keys there is still the possibility of external implementations of the trait not having caught up yet. I think treating that as an error instead of a variant like Unknown might be too heavy-weight given I see this more as a helper API than load-bearing. WDYT?

As a dev, I'm unsure if I'd ever handle `Unknown` differently from `Inconsistent` differently from `Error`. I basically only ever want `Consistent`, and anything else indicates misconfiguration—resulting in either a fast failure, alert or both for whatever system I'm developing.

I also think the design should reflect whether or not this check will happen by default during the creation of a `CertifiedKey` (going off of https://github.com/rustls/rustls/issues/1918#issuecomment-2097111280).

cpu:
> @cpu, I took a stab at rewriting the keys_match() function with this in mind

Thanks! That looks close to what I was thinking, but I was imagining it wouldn't be a `Result<KeyConsistency, Error>` return, just `KeyConsistency`.

>  The "unknown" in particular feels a bit goofy to me, like Rustls is a magic 8 ball that wants me to "ask again later" (i.e. when we've gone around and implemented public_key on all the SigningKeys)
> As a dev, I'm unsure if I'd ever handle Unknown differently from Inconsistent differently from Error. I basically only ever want Consistent, and anything else indicates misconfiguration—resulting in either a fast failure, alert or both for whatever system I'm developing.

Yeah :-/ It is awkward throwing a third variant in the mix.

If folks don't think my suggestion is a net improvement I'm not tied to it and open to alternatives. My main concern was avoiding a bad encoding error for the case where `SigningKey::public_key()` returns `Ok(None)`. If that's fixed another way I'm also satisfied. 

Maybe if the principle call-site we care about using that fn is going to turn it into an error anyway the default impl should be `Err(General("unimplemented".into()))` (bikeshedding the details as appropriate) instead of `Ok(None)`.

djc:
From #1918:

> Ideally Rustls would extract the SPKI from the EE certificate and then ask the crypto provider to do a pairwise consistency check as part of the construction of a CertifiedKey.

Is this actually what we're doing? Currently this PR actually just does a byte-for-byte comparison of the SPKI data, right?

cpu:
> Is this actually what we're doing?

Is that your preference? Without strong signal it seems better to follow the plan [outlined by Ctz](https://github.com/rustls/rustls/issues/1918#issuecomment-2093332074) unless there's consensus it isn't the right approach. When that comment was posted I didn't understand the advantages (I'm still not sure I do with crystal clarity). I asked [in the Discord thread]() and [Ctz said](https://discord.com/channels/976380008299917365/1236377676881596426/1237803871242293348):

> yes, i think he's saying that crypto libraries asked for the public half of a key pair could fulfil the request from a public key they read in (alongside the private key; for example RSA keys in PKCS#1 format) rather than one they computed. i think that seems like a harder situation for a user to get in though, and likely a different problem

djc:
> > Is this actually what we're doing?
> 
> Is that your preference? Without strong signal it seems better to follow the plan [outlined by Ctz](https://github.com/rustls/rustls/issues/1918#issuecomment-2093332074) unless there's consensus it isn't the right approach. When that comment was posted I didn't understand the advantages (I'm still not sure I do with crystal clarity).

Ahh, sorry for potentially derailing this discussion from an agreed upon plan. To be clear, doing this is definitely better than doing nothing. I just think "doing SPKI bytes comparison" is probably a weaker guarantee than if we could ask the crypto provider to do curve/group/whatever magic math thing to check that the keys are actually compatible.

lvkv:
>  My main concern was avoiding a bad encoding error for the case where SigningKey::public_key() returns Ok(None). If that's fixed another way I'm also satisfied.

Agree. @cpu How do you guys feel about the following options to avoid this? Ordered from least to most drastic:

1. Option 1: Change the default implementation of `SigningKey::public_key` to something along the lines of `err(Error::Unimplemented)`. This would allow us to disambiguate `None` from unimplemented when implementing `CertifiedKey::keys_match` and, in turn, avoid us mistaking "unimplemented" for "bad encoding".

2. Option 2: Remove the default implementation of `SigningKey::public_key` altogether and frontload the work of implementing it for types that `impl SigningKey`. This avoids the "unimplemented" case altogether.

3. Option 3: All of Option 2, plus removing the `Option` from the return type of `SigningKey::public_key`, taking it from `Result<Option<SubjectPublicKeyInfoDer>, Error>` to `Result<SubjectPublicKeyInfoDer, Error>`. This would eliminate both the "unimplemented" problem, but also the "unknown" case when one of the SPKIs is `None`.

Also open to suggestions!

cpu:
@lvkv From my perspective I think Option 1 is probably the most agreeable. Option 2 and 3 will be breaking API changes to the exported `SigningKey` trait and we're trying to avoid those while the ecosystem is catching up w/ our last release.

djc:
Option 1 sounds good, maybe we should file an issue with a label to remind ourselves to remove the default impl for the next semver-breaking release?

ctz:
I was thinking originally we could call the new consistency API `CertifiedKey::keys_match` in several places that accept `CertifiedKey`s and are already fallible. There are only a few such places, but doing that unprompted requires either a) we don't treat missing `SigningKey::public_key` as an error, or b) require  `SigningKey::public_key` as a fundamental part of that trait.

(a) means we can add those calls, and land this PR before writing the wedge of encoding conversion code that would be needed for the impls of SigningKey just in this crate.

(b) I think is a tough sell, needing a lot of research. For example, can you get a public key given a windows `NCRYPT_KEY_HANDLE`? I have no idea, but it seems not, and the suggestion is to look up the certificate and extract the public key from that (which would be quite circular...)

The other option, of course, is that we _don't_ insert those calls, and leave the API for others to call when they know or expect their `SigningKey` to support getting the public key.

djc:
Right, so I think this probably is the right direction still:

> I think the core question is the intended behavior for crypto providers that don't implement this. I think in rustls 0.24 we might want to tighten this up to force crypto provider implementers to support this? If that's the case, I think probably it'd make sense to just yield `Result<(), Error>` and have the default impl yield `Ok(())` for now.

cpu:
@lvkv Apologies, I think we lost some of the earlier momentum we had going here. Have you become busy (very reasonable!) or are there remaining points of uncertainty we should try and hash out to unblock progress?

lvkv:
Thanks for checking in. I'll be much more available after a few days! (Apartment hunting + moving in NYC)

lvkv:
Coming back to this now! It's _not_ clear to me, based on the discussion above, that there's a consensus yet on what the API for this should look like and why 🙂.

One idea to unstick the conversation: how do we feel about merging what we have right now (after cleanup)? I've also just added a commit to bring code coverage up to 100%. The plan would be:

* Clean up and this PR with its current API
* Begin working on `public_key` implementations
* (Optional) Revise the API as needed for the next minor release

lvkv:
Also—if there actually is a design consensus and I'm just oblivious to it, do let me know 🙂

cpu:
> One idea to unstick the conversation: how do we feel about merging what we have right now (after cleanup)?

That sounds OK to me :+1: 

> if there actually is a design consensus and I'm just oblivious to it, do let me know 

Let me try and summarize some things I think we have agreement on (_of course this is an invitation to please chime in if I'm misrepresenting anything_).

* It would be a stronger guarantee to make the `CryptoProvider` handle checking consistency, instead of looking at SPKI. However, doing that in a backwards compatible way is more challenging, and in practice SPKI matching will be fine in the majority of cases. **Consensus: continue with SPKI approach.**
* We don't want to require existing `SigningKey` impls update to provide a new mandatory fn for exposing the public key/SPKI - it would be backwards incompatible, and in some cases (e.g. [rustls-cng](https://github.com/rustls/rustls-cng)) it might not be possible to implement. **Consensus: we need a default implementation, and it should be possible for an impl to opt-out.**
* We want to check `CertifiedKey` consistency internally in the fallible places we accept one as input. Because of ^ we have to be prepared for a case where we _can't_ determine consistency. **Consensus: when we try to do consistency matching internally we should not hard-fail if the SPKI can't be determined.**

With the above in mind it feels like this branch is pretty close to matching the consensus. I think `SigningKey::public_key()` could be changed to return `Option<SubjectPublicKeyInfoDer>` instead of `Result<Option<SubjectPublicKeyInfoDer>>` - it doesn't seem like there's a use-case for surfacing a failure to produce the SPKI separate from just yielding `None`. 

When we call `CertifiedKey::keys_match` internally we'd treat `KeyConsistency::Inconsistent` as an error to surface immediately to the caller, and we'd ignore `KeyConsistency::Unknown` so that default/missing impls of `SigningKey::public_key()` are a non-failure.

I think it'd be best to land the internal calls to `CertifiedKeys::keys_match` alongside this work so that we don't end up with any dead code.

@ctz @djc How do you feel about the above?

ctz:
> Let me try and summarize some things I think we have agreement on (_of course this is an invitation to please chime in if I'm misrepresenting anything_).

Thanks for writing this down, and that does match where I think we are.

> With the above in mind it feels like this branch is pretty close to matching the consensus. I think `SigningKey::public_key()` could be changed to return `Option<SubjectPublicKeyInfoDer>` instead of `Result<Option<SubjectPublicKeyInfoDer>>` - it doesn't seem like there's a use-case for surfacing a failure to produce the SPKI separate from just yielding `None`.

Agree on this.

> I think it'd be best to land the internal calls to `CertifiedKeys::keys_match` alongside this work so that we don't end up with any dead code, but the updates toe implement `public_key()` for the crate internal certified keys could be follow-up.

I could take it or leave it. As this PR currently stands there is no dead code, since the new public API surface calls everything (and there are no uncovered lines because the tests exercise everything).

cpu:
> I mostly agree, the one thing I'm in doubt about is whether the KeyConsistency type pulls its weight in this setup compared to just using Result<(), Error> or maybe even just bool.

Using `CertifiedKey::keys_match(&self) -> bool` seems unappealing because if the fn returns `false` in one of the internal call-sites we would have to proceed without error not knowing whether it was `false` because the SPKI info was available and didn't match, or `false` because it wasn't available.

Using `CertifiedKey::keys_match(&self) -> Result<(), Error>` seems OK to me, but to avoid the situation above we'd need distinct `Error` results like `Error::KeysInconsistent` and `Error:SpkiUnavailable` (bikeshedding names tbd). We'd only be treating one of those two variants as an error internally, so it feels a little bit odd to express both as `Error` variants. 

I don't feel strongly. @ctz do you want to be a tie breaker?

ctz:
> I don't feel strongly. @ctz do you want to be a tie breaker?

Maybe I can get there by enumerating the things I don't think work, lol:

- `keys_match(&self) -> Result<KeyConsistency, Error>` (current PR state) begs me as a caller to say `keys_match().unwrap()` but that misses the `Ok(KeyConsistency::Inconsistent)` case I wanted to actually check for, oops.
- `keys_match(&self) -> bool` - agree that this collapses too many error cases to `false`

I think that leaves us with `keys_match(&self) -> Result<(), Error>`:

Generally I would prefer not to add very specific cases to the top of `rustls::Error`, or add further uses of `Error::General`. Reusing `CertificateError` comes with the baggage that we say it only comes from the certificate verification traits, and that each one has a TLS alert that it maps down to.

I suggest we add something like:

```rust
pub enum Error {
(...)
    /// Specific failure cases from [`CertifiedKey::keys_match`]
    InconsistentKeys(InconsistentKeys),
(...)
}

/// Specific failure cases from [`CertifiedKey::keys_match`]
#[non_exhaustive]
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
pub enum InconsistentKeys {
    /// The public key returned by the [`SigningKey`] does not match that in the certificate.
    SigningKeyAndCertificateMismatch,

    /// The [`SigningKey`] cannot produce its corresponding public key.
    SigningKeyCannotProducePublicKey,
}
```

That is very wordy, so please feel free to wordsmith as desired.

cpu:
> I suggest we add something like

That proposal sounds good to me. With reflection I think @djc's instincts are correct here & the consistent variant in the `KeyConsistency` enum doesn't hold its weight. When you remove that variant all that's left are the two error cases and I think you've made a good case for why they should be tucked behind na `Error::InconsistentKeys` variant as a distinct error enum.

So to summarize: 

* `SigningKey::public_key()` should be updated to return `Option<SubjectPublicKeyInfoDer>`, with the default impl being `None`
* `KeyConsistency` enum gets removed.
* We add a `InconsistentKeys` enum to `error.rs` with variants for "definitely mismatched" and "not sure"
* `Error` gets a new `Error::InconsistentKeys` variant holding a `InconsistentKeys`.
* `CertifiedKey::keys_match(&self)` should be updated to return `Result<(), Error>`, yielding `Ok(())` for the case where things check out, and `Err(Error::InconstentKeys(...))` otherwise. The specific `InconstentKeys` variant to use will depend on whether `public_key()` returned `None`, or a value that didn't match the cert's SPKI.
* In the Rustls-internal callsites where we take a `CertifiedKey` instance, we call `keys_match()`, matching on the `Err()` result to decide whether to eat the error (for the "not sure" case to preserve back-compat) or to pass it forward (for the "definitely mismatched" case).

lvkv:
> SigningKey::public_key() should be updated to return Option<SubjectPublicKeyInfoDer>, with the default impl being None

> KeyConsistency enum gets removed.

> We add a InconsistentKeys enum to error.rs with variants for "definitely mismatched" and "not sure"

> Error gets a new Error::InconsistentKeys variant holding a InconsistentKeys.

> CertifiedKey::keys_match(&self) should be updated to return Result<(), Error>, yielding Ok(()) for the case where things check out, and Err(Error::InconstentKeys(...)) otherwise. The specific InconstentKeys variant to use will depend on whether public_key() returned None, or a value that didn't match the cert's SPKI.

Done. Feel free to nit!

> In the Rustls-internal callsites where we take a CertifiedKey instance, we call keys_match()

I'll try adding this in too. In the meantime, are there any super-important call sites in particular you have in mind for this? We can also turn this bullet into a separate PR if you'd like.

rustls/src/crypto/signer.rs
```diff
@@ -105,6 +112,25 @@ impl CertifiedKey {
...
+                    .and_then(ParsedCertificate::try_from)
+                    .map(|ee| ee.subject_public_key_info() == spki)
+            })
+            // Error case: The signing key cannot produce a public key
+            // Error case: We've failed to parse this end-entity certificate
+            .ok_or(InconsistentKeys::Unknown)??
+            // Looking good: The SPKI values are equal
+            .then_some(())
+            // Error case: The SPKI values are not equal and these keys do not match
+            .ok_or(InconsistentKeys::KeyMismatch.into())
```
djc, line 131: I think chaining style is suboptimal for this case. Suggest more like this:

```rust
let key_spki = match self.key.public_key() {
    Some(key) => key,
    None => return Err(InconsistentKeys::Unknown.into()),
};

let cert = ParsedCertificate::try_from(self.end_entity_cert()?)?;
match key_spki == cert.subject_public_key_info() {
    true => Ok(())
    false => Err(InconsistentKeys::KeyMismatch.into()),
}
```

lvkv: Added nearly exactly this in [87bb093](https://github.com/rustls/rustls/pull/1954/commits/87bb09334b54d07e5f57b61ca3c9237af50849d0)! Nearly = missing a comma

lvkv:
Is there a meaningful way to add test coverage for the `Display` of this new error variant?

ctz:
> Is there a meaningful/not contrived way to add test coverage for the `Display` of this new error variant?

Add a line into this test: https://github.com/rustls/rustls/blob/3bb255cd58cda65f3e8a998f71df03429f03941c/rustls/src/error.rs#L724

rustls/src/crypto/signer.rs
```diff
@@ -59,6 +60,12 @@ pub trait SigningKey: Debug + Send + Sync {
     /// using the chosen scheme.
     fn choose_scheme(&self, offered: &[SignatureScheme]) -> Option<Box<dyn Signer>>;
 
+    /// Get the RFC 5280-compliant SubjectPublicKeyInfo (SPKI) of this [`SigningKey`] if it exists.
```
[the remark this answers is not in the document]
cpu: Agreed

rustls/src/error.rs
```diff
@@ -107,6 +110,24 @@ pub enum Error {
     Other(OtherError),
 }
 
+/// Specific failure cases from [`crate::crypto::signer::CertifiedKey::keys_match`].
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum InconsistentKeys {
+    /// The public key returned by the [`crate::crypto::signer::SigningKey`] does not match that in the certificate.
+    KeyMismatch,
```
cpu, line 118: minor editorial nit: I would be tempted to doc this a little bit more specifically:

```suggestion
    /// The public key returned by the [`crate::crypto::signer::SigningKey`] does not match the public key information in the certificate.
    KeyMismatch,
```

rustls/tests/api.rs
```diff
@@ -2936,6 +2936,64 @@ fn sni_resolver_rejects_bad_certs() {
     );
 }
 
+/// Represents a SigningKey that returns None for its SPKI.
```
cpu, line 2939: nit: maybe worth calling out the default impl? 

```suggestion
/// Represents a SigningKey that returns None for its SPKI via the default impl
```

rustls/tests/api.rs
```diff
@@ -2936,6 +2936,64 @@ fn sni_resolver_rejects_bad_certs() {
...
+#[derive(Debug)]
+struct SigningKeySomeSpki;
+
+#[cfg(test)]
+impl sign::SigningKey for SigningKeySomeSpki {
+    fn public_key(&self) -> Option<pki_types::SubjectPublicKeyInfoDer> {
+        let chain = KeyType::Rsa2048.get_chain();
+        let cert = ParsedCertificate::try_from(chain.first().unwrap()).unwrap();
+        let spki = cert.subject_public_key_info().into_owned();
+        Some(spki)
```
cpu, line 2966: nit: I think the `spki` binding doesn't hold much weight here:

```rust
        let chain = KeyType::Rsa2048.get_chain();
        let cert = ParsedCertificate::try_from(chain.first().unwrap()).unwrap();
        Some(cert.subject_public_key_info().into_owned())
```

rustls/tests/api.rs
```diff
@@ -2936,6 +2936,64 @@ fn sni_resolver_rejects_bad_certs() {
...
+    }
+
+    fn algorithm(&self) -> rustls::SignatureAlgorithm {
+        unimplemented!("Not meant to be called during tests")
+    }
+}
+
+#[test]
+#[cfg(test)]
+fn test_keys_match() {
```
cpu, line 2983: I think the style guide ordering would prescribe this test being _before_ the `SigningKeyNoneSpki` and `SigningKeySomeSpki` impls since it relies on them.

cpu (commented):
Thanks for rolling in the updated aspects of the design. Looks really good :+1: 

I think it's probably a good time to tidy up the commit history? WDYT?

lvkv:
Oh boy, I seem to have done the wrong thing with Git here 😓

rustls/src/crypto/signer.rs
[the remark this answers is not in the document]
lvkv: Done!

rustls/tests/api.rs
lvkv: Done!

rustls/tests/api.rs
lvkv: Done!

lvkv:
History rewritten! Feel free to nit further on the commit grouping, of course.

ctz (approved):

rustls/src/error.rs
```diff
@@ -97,6 +97,9 @@ pub enum Error {
     /// or too large.
     BadMaxFragmentSize,
 
+    /// Specific failure cases from [`crate::crypto::signer::CertifiedKey::keys_match`].
+    InconsistentKeys(InconsistentKeys),
```
cpu, line 101: I would be tempted to combine this commit with the one that follows that adds `keys_match`. I think the implementation + the error it can produce being together in one unit makes the most sense.

rustls/src/lib.rs
```diff
@@ -514,8 +514,8 @@ pub use crate::enums::{
     ProtocolVersion, SignatureAlgorithm, SignatureScheme,
 };
 pub use crate::error::{
-    CertRevocationListError, CertificateError, EncryptedClientHelloError, Error, InvalidMessage,
-    OtherError, PeerIncompatible, PeerMisbehaved,
+    CertRevocationListError, CertificateError, EncryptedClientHelloError, Error, InconsistentKeys,
+    InvalidMessage, OtherError, PeerIncompatible, PeerMisbehaved,
```
cpu, line 518: nit: I think adding `InconsistentKeys` to the `pub use` in `lib.rs` should be done in the commit that adds the error type.

cpu (approved):
Awesome! I had a couple suggestions for the commit order/content.

rustls/src/error.rs
```diff
@@ -107,6 +110,24 @@ pub enum Error {
     Other(OtherError),
 }
 
+/// Specific failure cases from [`crate::crypto::signer::CertifiedKey::keys_match`].
```
djc, line 113: For the docstring here and on its variants, might be nice to drop the whole path from the references, and hide the link target. (By doing like a `[keys_match]: crate::crypto::signer::CertifiedKey::keys_match` or whatever at the bottom.)

djc (approved):

lvkv:
Cool, I'll add these in today

lvkv:
Home stretch! This one felt a bit slow, so I'm down to increase the pace for whatever work follows this PR

rustls/src/webpki/verify.rs
```diff
@@ -125,6 +127,13 @@ impl fmt::Debug for WebPkiSupportedAlgorithms {
 /// This is used in order to avoid parsing twice when specifying custom verification
 pub struct ParsedCertificate<'a>(pub(crate) webpki::EndEntityCert<'a>);
 
+impl<'a> ParsedCertificate<'a> {
+    /// Get the parsed certificate's SubjectPublicKeyInfo (SPKI)
+    pub fn subject_public_key_info(&self) -> SubjectPublicKeyInfoDer {
```
djc, line 132: Nit: would be good to make the lifetime in the return value explicit here, `<'static>`.

rustls/src/crypto/signer.rs
```diff
@@ -59,6 +60,12 @@ pub trait SigningKey: Debug + Send + Sync {
     /// using the chosen scheme.
     fn choose_scheme(&self, offered: &[SignatureScheme]) -> Option<Box<dyn Signer>>;
 
+    /// Get the RFC 5280-compliant SubjectPublicKeyInfo (SPKI) of this [`SigningKey`] if available.
+    fn public_key(&self) -> Option<SubjectPublicKeyInfoDer> {
```
djc, line 64: Should have a lifetime here, too, I guess we can do `<'_>` so implementations can decide to borrow from `self`.

djc (approved):
Thanks for all the work on this!

cpu (approved):

rustls/src/error.rs
lvkv: Done!

rustls/src/crypto/signer.rs
lvkv: Done!

rustls/src/webpki/verify.rs
lvkv: Can you explain why we'd want to do this? Plus how to do this. I'm less solid on lifetimes.

* **Why** I'm not sure I understand:
  * The motivation for making the lifetime API explicit instead of elided. Is this because we expect (by inspection) the return value to be always `'static`, and so we might as well make it explicit?
  * Why the return value for this function would always be `'static`, and why we'd want that
* **How** This is how I'd naively do this. I'm unsure how to appease the borrow checker without tacking on an extra `into_owned` (which looks like it might incur an extra copy):

  ```diff
   impl<'a> ParsedCertificate<'a> {
       /// Get the parsed certificate's SubjectPublicKeyInfo (SPKI)
  -    pub fn subject_public_key_info(&self) -> SubjectPublicKeyInfoDer {
  -        self.0.subject_public_key_info()
  +    pub fn subject_public_key_info(self) -> SubjectPublicKeyInfoDer<'static> {
  +       self.0
  +            .subject_public_key_info()
  +            .into_owned()
      }
  }
  ```

djc: There's no need to do `into_owned()` because the webpki `subject_public_key_info()` method always yields a `SubjectPublicKeyInfoDer<'static>`.

cpu: > The motivation for making the lifetime API explicit instead of elided.

I've seen experienced rustaceans point to this practice as a debugging aid (e.g. https://quinedot.github.io/rust-learning/dont-hide.html).

lvkv: > (See https://github.com/rustls/webpki/commit/211550edfdb8fe2b0d03760acc30fede55874f28.)

Ahhh ok I see what we're missing. Your suggested change doesn't compile without your linked webpki commit that updates the `subject_public_key_info` signature to explicitly add that `'static` lifetime. I was able to confirm this by pointing Rustls' Cargo to webpki's main branch and compiling everything without a hitch. I believe this means we have to:

1. Put that commit into a webpki release
2. Update Rustls to use that release

Compiler error without the linked commit for convenience:
```rustc
error: lifetime may not live long enough
   --> rustls/src/webpki/verify.rs:133:9
    |
130 | impl<'a> ParsedCertificate<'a> {
    |      -- lifetime `'a` defined here
...
133 |         self.0.subject_public_key_info()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'a` must outlive `'static`

error: lifetime may not live long enough
   --> rustls/src/webpki/verify.rs:133:9
    |
132 |     pub fn subject_public_key_info(&self) -> SubjectPublicKeyInfoDer<'static> {
    |                                    - let's call the lifetime of this reference `'1`
133 |         self.0.subject_public_key_info()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'static`

help: replace `'a` with `'static`

error: could not compile `rustls` (lib) due to 2 previous errors
```

lvkv: ^ Bump. Am I missing something here?

cpu: You're correct :-) We'll get the webpki release published shortly and then this should be good to go.

djc: This has been released: https://github.com/rustls/webpki/releases/tag/v%2F0.102.5. Make sure to update the requirement to 0.102.5 in `Cargo.toml`.

lvkv: Updated!

rustls/tests/api.rs
```diff
@@ -2964,6 +2964,69 @@ fn sni_resolver_rejects_bad_certs() {
     );
 }
 
+#[test]
+#[cfg(test)]
```
cpu, line 2968: Is the `#[cfg(test)]` gate here, or on L2975 needed? It seems to me that they might be superfluous :thinking:

cpu (approved):
Nice!

rustls/tests/api.rs
lvkv: Removed these in [Add tests for CertifiedKey::keys_match](https://github.com/rustls/rustls/pull/1954/commits/c5590aea7b08c003e6acad3ed9f4c1bdd8675951). Also, I added a Cargo.lock update to [4446c9f](https://github.com/rustls/rustls/pull/1954/commits/4446c9f779e5ad1053b25b486204c19940af758e)

cpu: > Also, I added a Cargo.lock update to [4446c9f](https://github.com/rustls/rustls/pull/1954/commits/4446c9f779e5ad1053b25b486204c19940af758e)

The `fuzz` crate needs its `Cargo.lock` bumped too. I forget this roughly 50% of the time I do this dance :laughing:

cpu (approved):

PR #1955, opened 2024-05-17
Avoid sending multiple close notifies

ctz:
This PR also fixes #1950 which caused a desynchronisation after a close notify.

After this PR, tokio-rustls test suite breakage caused in 0.23.6 (see https://github.com/rustls/tokio-rustls/actions/runs/9123001981/job/25084713982) is fixed 

# 0.23.7 draft release notes

- `send_close_notify` is now idempotent, in case it is accidentally called more than once.
- `read_tls` now refuses to read further data after a `close_notify` is received, by returning `Ok(0)` (ie, an EOF).
- Correct fix in 0.23.6 to properly discard data after `close_notify` received, avoiding a spurious `DecryptError` on subsequent calls to `process_new_packets()`.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -475,14 +475,20 @@ impl CommonState {
         err.into()
     }
 
-    /// Queues a close_notify warning alert to be sent in the next
+    /// Queues a `close_notify` warning alert to be sent in the next
     /// [`Connection::write_tls`] call.  This informs the peer that the
     /// connection is being closed.
     ///
+    /// Does nothing if any `close_notify` or fatal alert was already sent.
+    ///
     /// [`Connection::write_tls`]: crate::Connection::write_tls
     pub fn send_close_notify(&mut self) {
+        if self.sent_fatal_alert {
+            return;
+        }
         debug!("Sending warning alert {:?}", AlertDescription::CloseNotify);
         self.send_warning_alert_no_log(AlertDescription::CloseNotify);
+        self.sent_fatal_alert = true;
     }
 
     pub(crate) fn eager_send_close_notify(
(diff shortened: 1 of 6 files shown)


rustls/src/common_state.rs
```diff
@@ -475,14 +475,19 @@ impl CommonState {
...
     /// [`Connection::write_tls`] call.  This informs the peer that the
     /// connection is being closed.
     ///
+    /// Does nothing if any `close_notify` or fatal alert was already sent.
+    ///
     /// [`Connection::write_tls`]: crate::Connection::write_tls
     pub fn send_close_notify(&mut self) {
-        debug!("Sending warning alert {:?}", AlertDescription::CloseNotify);
-        self.send_warning_alert_no_log(AlertDescription::CloseNotify);
+        if !self.sent_fatal_alert {
```
djc, line 486: Invert condition/early return?

djc (approved):

cpu (approved):
LGTM.

Can you add some proposed release notes to the PR desc?

ctz:
> Can you add some proposed release notes to the PR desc?

Added

cpu:
> Added

Thanks!

> Correct fix in 0.23.6 to discarding data after close_notify received, causing a spurious DecryptError on subsequent calls to process_new_packets().

This reads a little bit awkwardly to me. Maybe something like: "Correct fix in 0.23.6 to properly discard data after `close_notify` received, avoiding a spurious `DecryptError` on subsequent calls to `process_new_packets()`"?

ctz:
> This reads a little bit awkwardly to me.

Agreed, replaced with your wording. Thanks!

PR #1962, opened 2024-05-20
Misc changes around certificate encoding

ctz:
These are commits I am sloughing off of the cert compression branch.

### DIFF
--- a/admin/all-features-except
+++ b/admin/all-features-except
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+DESCRIPTION = """
+Usage: admin/all-features-except a,b,c crate
+
+Prints the set of all features declared by the given crate, minus
+a, b and c.
+
+The output is decorated with `--no-default-features --features` meaning
+it can be used directly with cargo, for example:
+
+$ cargo build $(admin/all-features-except std,logging rustls)
+
+It is assumed but not verified that the remaining features do not
+have dependencies on the disabled features (otherwise, cargo will
+re-enable them without notification).
+"""
+
+import subprocess
+import argparse
+import json
+
+
+def crate_features(crate_name):
+    js = json.loads(
+        subprocess.check_output(
+            ["cargo", "metadata", "--no-deps", "--format-version=1"]
+        )
+    )
+    package = [p for p in js["packages"] if p["name"] == crate_name]
+    if len(package) != 1:
+        raise ValueError(f"crate {crate_name} not found in {js}")
+    return set(package[0]["features"].keys())
+
+
+if __name__ == "__main__":
+    ap = argparse.ArgumentParser(description=DESCRIPTION)
+    ap.add_argument(
+        dest="except_features",
+        metavar="FEATURE,FEATURE,..",
+        type=lambda x: set(x.strip().split(",")),
+        help="Which features to remove from the complete set of features",
+    )
+    ap.add_argument("crate", help="Which crate to determine features for")
+    opts = ap.parse_args()
+
+    features = crate_features(opts.crate)
+
+    for f in opts.except_features:
+        try:
+            features.remove(f)
+        except KeyError as e:
+            ap.error(f"feature `{f}' not found (known are {features})")
+
+    # drop default, as it may re-enable other features
+    features.discard("default")
+
+    print("--no-default-features", "--features", ",".join(sorted(features)))
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1287,12 +1287,17 @@ impl<'a> Deref for CertificateChain<'a> {
 }
 
 impl TlsListElement for CertificateDer<'_> {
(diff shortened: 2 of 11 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -1287,9 +1287,18 @@ impl<'a> Deref for CertificateChain<'a> {
...
+/// TLS has a 16MB size limit on any handshake message,
+/// plus a 16MB limit on any given certificate.
+///
+/// We contract that to 64KB to limit the amount of memory allocation
+/// that is directly controllable by the peer.
+pub(crate) const CERTIFICATE_MAX_SIZE_LIMIT: usize = 0x1_0000;
+
 // TLS1.3 changes the Certificate payload encoding.
 // That's annoying. It means the parsing is not
 // context-free any more.
```
cpu, line 1304: This is a carry-over from `main` that isn't related to the PR but I'm curious if this comment should be moved somewhere else. It seems a bit oprhaned hanging out here between `CERTIFICATE_MAX_SIZE_LIMIT` and the `CertificateExtension` enum.

admin/all-features-except
```diff
@@ -0,0 +1,30 @@
...
+if [ $# -ne 2 ] ; then
+    echo "$0 <feature,feature,..> CRATE"
+    exit
+fi
+
+EXCEPT_FEATURES="$1"
+CRATE="$2"
+
+# gratuitously clone-and-hacked from
+# https://github.com/actix/actix-net/blob/master/justfile#L16-L22
```
cpu, line 24: :pray: Thank you for sparing us having to wield `jq` like this from first principles :laughing:

cpu (approved):
Nice :+1:

admin/all-features-except
djc: Personally I would feel more comfortable with a Python script over the use of shell script with `jq`, but I guess we have to make do?

rustls/src/msgs/handshake.rs
```diff
@@ -1295,28 +1295,35 @@ impl TlsListElement for CertificateDer<'_> {
...
+impl<'a> CertificateExtension<'a> {
     pub(crate) fn ext_type(&self) -> ExtensionType {
         match *self {
             Self::CertificateStatus(_) => ExtensionType::StatusRequest,
             Self::Unknown(ref r) => r.typ,
         }
     }
 
-    pub(crate) fn cert_status(&self) -> Option<&Vec<u8>> {
+    pub(crate) fn cert_status(&'a self) -> Option<&'a [u8]> {
```
djc, line 1320: My intuition is that we don't need the explicit lifetime for `&self`?

rustls/src/msgs/handshake.rs
```diff
@@ -1345,38 +1352,49 @@ impl Codec<'_> for CertificateExtension {
...
             exts: Vec::new(),
         }
     }
 
+    pub(crate) fn into_owned(mut self) -> CertificateEntry<'static> {
+        CertificateEntry {
+            cert: self.cert.into_owned(),
+            exts: self
+                .exts
+                .drain(..)
```
djc, line 1401: Nit: maybe `into_iter()` instead of `drain(..)`?

rustls/src/msgs/handshake.rs
```diff
@@ -1391,46 +1409,57 @@ impl CertificateEntry {
             .any(|ext| ext.ext_type() != ExtensionType::StatusRequest)
     }
 
-    pub(crate) fn ocsp_response(&self) -> Option<&Vec<u8>> {
+    pub(crate) fn ocsp_response(&'a self) -> Option<&'a [u8]> {
```
djc, line 1421: Same here.

rustls/src/msgs/handshake.rs
```diff
@@ -1391,46 +1409,57 @@ impl CertificateEntry {
...
             entries,
         }
     }
 
+    pub(crate) fn into_owned(mut self) -> CertificatePayloadTls13<'static> {
+        CertificatePayloadTls13 {
+            context: self.context,
+            entries: self
+                .entries
+                .drain(..)
```
djc, line 1468: As above.

djc (approved):

admin/all-features-except
cpu: I would be open to that alternative. It would be easier to debug if it breaks 👍

rustls/src/msgs/handshake.rs
ctz: Will remove this comment, not really sure what it's adding that is not now obvious from the code.

admin/all-features-except
ctz: Replaced with python; much more readable!

admin/all-features-except
```diff
@@ -0,0 +1,58 @@
...
+have dependencies on the disabled features (otherwise, cargo will
+re-enable them without notification).
+"""
+
+import subprocess
+import argparse
+import json
+
+
+def crate_features(crate_name):
```
cpu, line 24: much nicer :snake: :sparkles:

PR #1963, opened 2024-05-21
Implement RFC 9180 HPKE provider backed by aws-lc-rs

cpu:
In order to support Encrypted Client Hello (ECH) we need a Hybrid Public Key Encryption (HPKE) provider (see [RFC 9180](https://www.rfc-editor.org/rfc/rfc9180.html)). Previously we landed the traits required to abstract over an HPKE provider, and a proof-of-concept provider that used [hpke-rs](https://github.com/franziskuskiefer/hpke-rs) and Rust Crypto.  In order to support HPKE in the base Rustls crate we need an HPKE provider backed by either `aws-lc-rs`, or `*ring*`. 

Since static ECDH is not supported by `*ring*` (https://github.com/briansmith/ring/issues/331), but is supported by `aws-lc-rs` (https://github.com/aws/aws-lc-rs/pull/302) this branch implements an HPKE provider backed by `aws-lc-rs`. The proof-of-concept `hpke-rs` provider from the `provider-example` crate is removed. The existing test harness is moved to the rustls integration test package and lightly updated for `aws-lc-rs`.

Following the precedent set by the HPKE traits we landed, this is focused on only what is required for ECH. We don't implement any HPKE modes other than "base mode" (see [RFC 9180 §5](https://www.rfc-editor.org/rfc/rfc9180.html#section-5)) , and don't offer export secrets.

The HPKE RFC requires some HKDF operations that previously weren't needed for TLS 1.3. Most notably, we need to be able to extract pseudo-random keys (PRKs) without immediately expanding them. The existing `Hkdf` trait, and the upstream implementations from aws-lc-rs and `*ring*` both encapsulate the PRK to prevent misuse, and so can't be used for this purpose. We work around this by adding a new `HkdfPrkExtract` trait that is a superset of `Hkdf`, offering a new `extract_prk_from_secret()` fn. We then implement this for the HMAC based HKDF impl from the `tls13` package. This means the aws-lc-rs impl is tied to the HMAC based HKDF, but this doesn't seem to have any disadvantages in practice and the non-HMAC options can't support this operation without upstream changes.

Lastly, a new `rustls-provider-test` crate is added to allow writing integration tests that exercise different crypto providers and associated machinery. We use this new crate to adapt the existing HPKE integration test to test that our new aws-lc-rs HPKE suites, and the existing example hpke-rs based suites can interoperate.

Related to https://github.com/rustls/rustls/pull/1718


rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,602 @@
...
+    }
+
+    fn hkdf_algorithm(suite: &HpkeSuite) -> Option<&'static dyn Hkdf> {
+        Some(match suite.sym.kdf_id {
+            HpkeKdf::HKDF_SHA256 => HKDF_SHA256,
+            HpkeKdf::HKDF_SHA384 => HKDF_SHA384,
+            HpkeKdf::HKDF_SHA512 => HKDF_SHA512,
+            HpkeKdf::Unknown(_) => return None,
+        })
+    }
```
ctz, line 87: Great stuff on getting this working!

Generally I like to make things like this "linker-friendly" in the sense that the linker can throw away unused algorithms. These sort of functions that map an enum to an algorithm implementation would be in the way of that.

The way that could look here is:

- have static `HpkeContext` values, one for each supported `HpkeSuite`, referring to `&aead::AES_128_GCM` etc.
- `Provider` is initialised with a list of these, it `supports_suite` if one exists in its list, and for `start()` can look up the algorithm details similarly, then pass that into the object that is `dyn Hpke`

cpu: That makes sense 👍 I think this arrangement might also help me knock out a TODO I left RE: compile time constant output sizes in labelled expand.


I'll adjust.

Cargo.toml
```diff
@@ -1,28 +1,32 @@
...
-  "examples",
-  # Tests that require OpenSSL
-  "openssl-tests",
-  # example of custom provider
-  "provider-example",
-  # the main library and tests
-  "rustls",
-  # experimental post-quantum algorithm support
-  "rustls-post-quantum",
+    # CI benchmarks
```
cpu, line 3: IDE biffed this one. Will revert this formatting diff.

cpu: Fixed.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Reworked to be linker-friendly.

cpu:
> I'm reusing some general HKDF code from crate::crypto::tls13, but that means it's no longer TLS1.3 specific. Should these parts be lifted into a separate crate::crypto::hkdf module?

Thoughts on this? I'm leaning towards yes, as its own commit ahead of implementing HPKE.

provider-example/src/lib.rs
```diff
@@ -12,15 +12,10 @@ use rustls::pki_types::PrivateKeyDer;
...
 mod hash;
 mod hmac;
-#[cfg(feature = "std")]
-mod hpke;
 mod kx;
 mod sign;
 mod verify;
 
-#[cfg(feature = "std")]
-pub use hpke::HPKE_PROVIDER;
```
cpu, line 22: I'm also open to leaving this example in-tree if folks think that's useful. Or, we could open a PR against rustls-rustcrypto to see if they are interested in it.

The only hurdle I can think of for keeping it in-tree is finding a nice way to share the HPKE tests. I don't know that we have a great spot for integration tests that would depend on both the rustls + provider-example crates. Depending on provider-example (even as a dev-dep) in the main crate is annoying because of the large number of transitive deps (that also includes other versions of rustls).

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,770 @@
...
+    /// [0]: https://www.rfc-editor.org/rfc/rfc9180.html#section-5.1.1
+    fn new(
+        suite: &Suite,
+        info: &[u8],
+        pub_key: &HpkePublicKey,
+    ) -> Result<(EncapsulatedSecret, Self), Error> {
+        // def SetupBaseS(pkR, info):
+        //   shared_secret, enc = Encap(pkR)
+        //   return enc, KeyScheduleS(mode_base, shared_secret, info,
+        //                            default_psk, default_psk_id)
```
cpu, line 444: Throughout I've been putting the pseudo-code from the RFC up-front where applicable. I found this helpful while writing it, and thought it might be useful for reviewers.

It's also a bit verbose, and it isn't always 1:1 (e.g. variable names, AEAD Seal being in-lined, etc).

Should these comments be removed prior to merge or do we like them?

jplock:
I’ve been interested in a HPKE implementation that is backed by aws-lc-rs. Will these structs and implementation be publicly accessible and usable even if I’m not interested in any other functionality in rustls?

I’ve been using hpke-rs with their RustCrypto provider but would be interested in switching.

cpu:
> Will these structs and implementation be publicly accessible and usable even if I’m not interested in any other functionality in rustls?

Yes, they're in the public API and should work in isolation with the caveat that only base mode is offered. If you need more modes/functionality writing an aws-lc-rs backend for hpke-rs is probably a better option.

jplock:
Base mode is all I’m interested in so this sounds great. Thank you!

rustls/src/crypto/aws_lc_rs/hpke.rs
ctz: I like these comments. I would be tempted to intermingle with them with the code they apply to, eg:

```rust
        // def SetupBaseS(pkR, info):
        //   shared_secret, enc = Encap(pkR)
        let (shared_secret, enc) = suite.dh_kem.encap(pub_key)?;

        //   return enc, KeyScheduleS(mode_base, shared_secret, info,
        //                            default_psk, default_psk_id)
        let key_schedule = suite.key_schedule(shared_secret, info)?;
        Ok((enc, Self { key_schedule }))
```

Then, if there is a line in the standard pseudocode that is happening but non-obvious (like, below, `self.IncrementSeq()`) that should be clearer.

ctz:
> Thoughts on this? I'm leaning towards yes, as its own commit ahead of implementing HPKE.

TLS1.3 requires HKDF, and HPKE/ECH requires TLS1.3, so I think it's reasonable if HKDF lives in the `tls13` module. But that's not a strong objection if you prefer to move it.

cpu:
> TLS1.3 requires HKDF, and HPKE/ECH requires TLS1.3, so I think it's reasonable if HKDF lives in the tls13 module

I find this a reasonable line of thinking and don't feel strongly. Leaving things as-is works for me.

>  cpu force-pushed the cpu-aws-lc-rs-hpke branch from 2b1af46 to cde04d3 

The wonderful folks upstream cut a [aws-lc-rs 1.7.2](https://github.com/aws/aws-lc-rs/releases/tag/v1.7.2) release that pulls in a better fix for the HKDF info limit we were bumping into. I reworked this branch to take that update and simplified the testing workarounds I was using before.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+    // TODO(XXX): consider reworking to use `crypto::tls13::expand()` with a compile size constant
+    //   output to make this infallible.
+    let mut expanded = vec![0; len as usize];
+    expander.expand_slice(info, &mut expanded)?;
+
+    Ok(expanded)
+}
+
+/// A newtype wrapper for a DH KEM shared secret.
+struct KemSharedSecret(Vec<u8>);
```
ctz: Consider `drop()` for this type that calls `self.0.zeroize()`.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+        sym: HpkeSymmetricCipherSuite {
+            kdf_id: HpkeKdfId::HKDF_SHA256,
+            aead_id: HpkeAeadId::AES_128_GCM,
+        },
+    },
+    dh_kem: &DhKem {
+        id: HpkeKemId::DHKEM_P256_HKDF_SHA256,
+        agreement_algorithm: &agreement::ECDH_P256,
+        hkdf: &RingHkdf(hkdf::HKDF_SHA256, hmac::HMAC_SHA256),
+    },
```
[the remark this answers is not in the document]
ctz: In fact the `RingHkdf` instantiations could be hoisted out of the ciphersuites inside `tls13.rs`, then used directly here.

rustls/tests/hpke.rs
```diff
@@ -0,0 +1,92 @@
...
+            let plaintext = hpke
+                .open(&enc, &info, &aad, &ciphertext, &sk_r)
+                .unwrap();
+            assert_eq!(plaintext, pt);
+        }
+    }
+}
+
+#[derive(Deserialize, Debug)]
+struct TestVector {
```
ctz: Not for this PR, but at some point I think we could make a "rustls-provider-test" crate that runs these and other known-answer and functional tests at the level of a `CryptoProvider`.

ctz (approved):

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+
+impl Debug for Suite {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.suite.fmt(f)
+    }
+}
+
+/// Adapts a [KeySchedule] for the role of a [HpkeSealer].
+struct Sealer {
+    key_schedule: KeySchedule,
```
djc: Why don't we store the `SealingKey` in the `Sealer` instead of the `KeySchedule`?

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+
+impl Debug for Sealer {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Sealer").finish()
+    }
+}
+
+/// Adapts a [KeySchedule] for the role of a [HpkeOpener].
+struct Opener {
+    key_schedule: KeySchedule,
```
djc: Same here for the `OpeningKey`.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+}
+
+/// A newtype wrapper for a DH KEM shared secret.
+struct KemSharedSecret(Vec<u8>);
+
+fn expand_err(_: OutputLengthError) -> Error {
+    Error::General("HKDF expansion failed".into())
+}
+
+fn unspecified_err(_e: aws_lc_rs::error::Unspecified) -> Error {
```
djc, line 763: I think this could be a `feature`-guarded `From` impl for `Error`? Same for the other one below? That should work with `?` such that you don't need the `map_err()`s all over the place.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+        AwsNonce::try_assume_unique_for_key(&nonce)
+    }
+}
+
+/// See RFC 9180 §4 "Cryptographic Dependencies"[^0].
+///
+/// [0]: https://www.rfc-editor.org/rfc/rfc9180.html#section-4
+fn labeled_extract(
+    hkdf: &'static dyn Hkdf,
+    suite_id: &[u8],
```
djc, line 727: That's a lot of `&[u8]` arguments. Is there more type safety to be had? It looks like `label` is arguably a `&str` and `suite_id` is actually a `HpkeSuiteId`...

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+
+    let labeled_ikm = [&b"HPKE-v1"[..], suite_id, label, ikm].concat();
+    hkdf.extract_from_secret(salt, &labeled_ikm)
+}
+
+/// See RFC 9180 §4 "Cryptographic Dependencies"[^0].
+///
+/// [0]: https://www.rfc-editor.org/rfc/rfc9180.html#section-4
+fn labeled_expand(
+    suite_id: &[u8],
```
djc, line 744: As for `labeled_extract()`.

provider-example/tests/hpke.rs
```diff
@@ -1,87 +0,0 @@
-use std::fs::File;
```
djc, line 1: IMO it would be nice to move these tests in a single commit so that Git can see it as a rename, resulting in better provenance. IMO it would be okay to move the tests out of the example just prior to removing the actual implementation from the example crate?

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -418,23 +420,42 @@ impl Hpke for &Suite {
...
 
-impl Debug for Suite {
+impl<const KEY_SIZE: usize, const KDF_SIZE: usize> Debug for Suite<KEY_SIZE, KDF_SIZE> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         self.suite.fmt(f)
     }
 }
 
+/// A [Hpke] suite implementation backed by `aws-lc-rs` cryptography.
+pub trait AwsLcRsHpkeSuite: Hpke {
```
djc, line 430: Not sure I fully understand why this intermediate trait is needed?

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -30,7 +32,7 @@ pub static DEFAULT_PROVIDER: &'static dyn HpkeProvider = &Provider {
 };
 
 /// All supported HPKE suites.
-pub static ALL_SUPPORTED_SUITES: &[&Suite] = &[
+pub static ALL_SUPPORTED_SUITES: &[&dyn AwsLcRsHpkeSuite] = &[
```
djc, line 35: Couldn't/shouldn't these have been `&dyn Hpke` from the start? And if so, do we still need the intermediate trait?

djc (commented):
Looking good!

rustls/src/crypto/aws_lc_rs/hpke.rs
ctz: I'm not sure about label being a string. Perhaps 
```rust
enum Label { Key, Secret, ... };

impl AsRef<[u8]> for Label {
  fn as_ref(&self) -> &[u8] {
    match self { Label::Key => &b"key"[..], ... }
  }
}
```?

File: Cargo.lock [hunk 16]
```diff
@@ -341,9 +341,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.7.1"
+version = "1.7.2"
```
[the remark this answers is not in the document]
cpu: IMO it's only load bearing only in terms of unit tests and so I left it out. Additionally I think if we need to update this req we'd also have to do it in webpki.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+    fn start(&self, id: &HpkeSuiteId) -> Result<Box<dyn Hpke + 'static>, Error> {
+        Ok(Box::new(
+            self.suite(*id)
+                .ok_or(Error::General("unsupported HPKE suite".into()))?,
+        ))
+    }
+
+    fn supports_suite(&self, id: &HpkeSuiteId) -> bool {
+        self.suite(*id).is_some()
+    }
```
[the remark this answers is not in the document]
cpu: I think they're more useful for the entry-point to a user of HPKE, like in the ECH PR and so seem underutilized here. 

I think we do want to have some mechanism for selecting the `dyn Hpke` that doesn't need to be repeated by each use-case. Maybe it could be lifted onto the `CryptoProvider` if we end up moving the supported suites there?

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+};
+use aws_lc_rs::{aead, agreement, hkdf, hmac};
+
+use crate::crypto::aws_lc_rs::tls13::RingHkdf;
+use crate::crypto::hpke::{
+    EncapsulatedSecret, Hpke, HpkeOpener, HpkePrivateKey, HpkeProvider, HpkePublicKey, HpkeSealer,
+    HpkeSuite as HpkeSuiteId,
+};
+use crate::crypto::tls13::{Hkdf, HkdfExpander, OutputLengthError};
+use crate::msgs::enums::{HpkeAead as HpkeAeadId, HpkeKdf as HpkeKdfId, HpkeKem as HpkeKemId};
```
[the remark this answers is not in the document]
cpu: IMO I think it makes sense to have the enums named to match the specification they come from, especially if we'd have to do some mild semver breaking to change it.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+
+use aws_lc_rs::aead::{
+    Aad, BoundKey, Nonce as AwsNonce, NonceSequence, OpeningKey, SealingKey, UnboundKey,
+};
+use aws_lc_rs::{aead, agreement, hkdf, hmac};
+
+use crate::crypto::aws_lc_rs::tls13::RingHkdf;
+use crate::crypto::hpke::{
+    EncapsulatedSecret, Hpke, HpkeOpener, HpkePrivateKey, HpkeProvider, HpkePublicKey, HpkeSealer,
+    HpkeSuite as HpkeSuiteId,
```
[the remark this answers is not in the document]
cpu: It's true the suite ID isn't a numeric ID, but a collection of three other IDs. I'm not sure what the best name for that is. I can do some fiddling. I was running out of names to use :)

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Good idea.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Sounds reasonable to me. I was matching the RFC too closely here I think and could introduce some more type safety on top.

provider-example/tests/hpke.rs
cpu: > IMO it would be nice to move these tests in a single commit so that Git can see it as a rename, resulting in better provenance.

I'll take a look at this again. I tried this way originally but it can't be strictly a `mv` - the test won't build/run once moved outside of the provider crate because it has a dependency on the provider's rust-crypto HPKE provider. Perhaps moving it and updating it in one go would make an OK diff?

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: I tried to explain in the commit message but might have missed the mark. We can't collect up the different const-generic parameterized impls into one slice without a trait abstracting over them. The trait is also necessary to be able to match a suite ID to a `dyn Hpke` since that's not possible with just that interface. Lastly we need a way to convert from the `AwsLcRsHpkeSuite` to a `Box<dyn Hpke>` without using unstable trait upcasting.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: It might be possible if we change the `Hpke` trait to offer a way to get a corresponding suite ID. I'm not 100% sure. In general I was trying to avoid changing elements that already landed. I think we'd also have to break semver for that change since there isn't a sensible default impl to use.

cpu:
Thanks you both for the reviews! I will start working through the feedback.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+    DHKEM_X25519_HKDF_SHA256_AES_128,
+    DHKEM_X25519_HKDF_SHA256_AES_256,
+    DHKEM_X25519_HKDF_SHA256_CHACHA20_POLY1305,
+];
+
+/// A [RFC 9180] Hybrid Public Key Encryption (HPKE) provider backed by aws-lc-rs cryptography.
+///
+/// [RFC 9180]: https://www.rfc-editor.org/rfc/rfc9180.html
+#[derive(Debug)]
+pub struct Provider {
```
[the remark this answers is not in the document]
djc: Adding a field to `CryptoProvider` is an incompatible change, I guess I just don't see much value in the `Provider` abstraction here over referencing a slice of suites.

provider-example/tests/hpke.rs
djc: Yes, that's my hypothesis -- having some changes in the same commit as the move is fine.

rustls/src/crypto/aws_lc_rs/hpke.rs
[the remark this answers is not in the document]
cpu: > Adding a field to CryptoProvider is an incompatible change,

Hmm true. Are we going to need to land a breaking change for cert compression? If not then I'm hesitant to make this the reason for the first breakage in `main` but could be convinced. If yes, then we might as well roll this all in together and get cooking on the next major release.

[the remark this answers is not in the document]
djc: Given that the ecosystem is taking its time upgrading to 0.23 I think it'd be great to avoid breaking changes for a while longer.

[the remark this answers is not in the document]
cpu: OK - I'll look at removing the `hpke::Provider` trait in place of a slice of `dyn Hpke`, but will hold off on putting that slice into the `CryptoProvider`. The `hpke` module has been doc hidden and so I think making breaking changes there is fair game similar to the internal API space.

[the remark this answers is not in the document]
ctz: > Hmm true. Are we going to need to land a breaking change for cert compression?

I don't think that is essential, and having heard @djc's opinion I think I will avoid it.

My feeling is: to use ECH there already needs to be application changes in the form of passing-in the EchConfig so it makes sense to pass on the configuration cost to users using that bit of API (as currently happens in #1718)

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: > Do you mean the shared secret, KemSharedSecret

Ah, no, on second look I suspect you mean the AEAD key.

rustls/src/crypto/aws_lc_rs/hpke.rs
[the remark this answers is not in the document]
cpu: Done both. Good idea, thank you.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,774 @@
...
+        id: HpkeKemId::DHKEM_X25519_HKDF_SHA256,
+        agreement_algorithm: &agreement::X25519,
+        hkdf: &RingHkdf(hkdf::HKDF_SHA256, hmac::HMAC_SHA256),
+    },
+    hkdf: &RingHkdf(hkdf::HKDF_SHA256, hmac::HMAC_SHA256),
+    aead: &aead::CHACHA20_POLY1305,
+};
+
+/// Suite holds the concrete instantiations of the algorithms selected by the suite ID.
+pub struct Suite {
```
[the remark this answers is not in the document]
cpu: Reordered, hopefully correctly :crossed_fingers:

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Reworked.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Done.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Implemented the `Label` enum.

provider-example/tests/hpke.rs
cpu: Good suggestion thanks. It ended up working like you'd hope. The diff renders as just the small edits made for updating the test to use the aws-lc-rs impls.

cpu:
> cpu force-pushed the cpu-aws-lc-rs-hpke branch from 4bce613 to afd9a00 

I believe I've addressed all the feedback. Thank you, the current state is a nice improvement over my first draft.

@ctz I'm going to dismiss your review since there's been substantial change. If you could re- :mag: I'd appreciate it.

> The solution I arrived at for the above notches up the code complexity enough that I kept it in a separate commit at the end in case I'm on the wrong track.

Since it seems like everyone liked the const-ified version I squashed it down into the initial implementation commit.

> rustls / Build+test (nightly, ubuntu-latest) (pull_request) Failing after 3m

See https://github.com/rust-lang/rust/issues/125474

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,796 @@
...
+    DHKEM_P521_HKDF_SHA512_AES_128,
+    DHKEM_P521_HKDF_SHA512_AES_256,
+    DHKEM_P521_HKDF_SHA512_CHACHA20_POLY1305,
+    DHKEM_X25519_HKDF_SHA256_AES_128,
+    DHKEM_X25519_HKDF_SHA256_AES_256,
+    DHKEM_X25519_HKDF_SHA256_CHACHA20_POLY1305,
+]);
+
+/// Suite holds the concrete instantiations of the algorithms selected by the suite ID.
+#[derive(Clone)]
```
cpu, line 44: Note to self: this doesn't need to be `Clone` - I think this was a leftover from some earlier work that fell away.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,796 @@
...
+    }
+}
+
+impl<const KEY_SIZE: usize, const KDF_SIZE: usize> Debug for Opener<KEY_SIZE, KDF_SIZE> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Opener").finish()
+    }
+}
+
+static DHKEM_P256_HKDF_SHA256: &DhKem<SHA256_OUTPUT_LEN> = &DhKem {
```
cpu, line 496: note to self: I think these should be below the `DhKem` definition.

cpu:
> codecov/patch — 86.85% of diff hit (target 95.54%)

Looks like the main pieces of missing coverage are:

1. The `setup_xxx` APIs (the integration test uses the one-shot APIs)
2. The `Debug` impls
3. The error mapping from `aws-lc-rs`.

For 1, I could update the integration test to use both but it's already a slow(ish) test and doing 2x the work seems not worth it to me.
For 2, I could add some simple format tests but it seems maybe not super worthwhile. We'll also gain coverage here from the ECH work.
For 3, I might be able to provoke these with a couple tests feeding back key data in. I'll give it some thought.

Thoughts?

cpu:
Added some quick test coverage for the gaps coveralls was highlighting.

rustls/src/crypto/hpke.rs
```diff
@@ -8,23 +8,17 @@ use crate::msgs::enums::HpkeKem;
...
-pub trait HpkeProvider: Debug + Send + Sync + 'static {
-    /// Start setting up to use HPKE in base mode with the chosen suite.
-    ///
-    /// May return an error if the suite is unsupported by the provider.
-    fn start(&self, suite: &HpkeSuite) -> Result<Box<dyn Hpke + 'static>, Error>;
-
-    /// Does the provider support the given [HpkeSuite]?
-    fn supports_suite(&self, suite: &HpkeSuite) -> bool;
+/// A collection of supported [Hpke] suite instances.
+pub struct HpkeSuites(pub &'static [&'static dyn Hpke]);
```
djc, line 12: Still not sure if this type is worth it... How many callers does `supported()` end up having?

provider-example/src/hpke.rs
```diff
@@ -53,6 +59,10 @@ impl Debug for HpkeRs {
 }
 
 impl Hpke for HpkeRs {
+    fn suite(&self) -> HpkeSuite {
```
djc: Nit: `suite()` should go at the bottom (in the impls, too).

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,846 @@
...
+                &[],
+            ),
+            Label::Key,
+            &key_schedule_context,
+        ));
+
+        // Note: All supported AEADs use the same NONCE_LEN.
+        let base_nonce = labeled_expand::<NONCE_LEN>(
+            &suite_id,
+            labeled_extract(
```
djc, line 72: Looks like this is the same as the one for `key`. Extract?

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,846 @@
...
+        Ok((
+            key,
+            NonceSchedule {
+                base_nonce,
+                seq_num: 0,
+            },
+        ))
+    }
+
+    fn hpke_suite_id(&self) -> Vec<u8> {
```
djc, line 92: Why isn't this a method on `HpkeSuite`?

rustls/src/crypto/aws_lc_rs/hpke.rs
djc: What about `suite_id`? It looks like that could/should be a `&HpkeSuite`.

rustls/src/crypto/aws_lc_rs/hpke.rs
djc: This didn't look like what I had in mind. I thought we might be able to just store the `SealingKey` in the `Sealer`? Looking again, I guess that doesn't work because we have to create a new `SealingKey` based on the `nonce_schedule` for every `seal()` operatiion?

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: We'd need more than just the `&HpkeSuite` because the value used has a different prefix/encoding depending on context (quoting the bottom of [9180, §4](https://www.rfc-editor.org/rfc/rfc9180.html#section-4)):

> The value of suite_id depends on where the KDF is used; it is assumed implicit from the implementation and not passed as a parameter. If used inside a KEM algorithm, suite_id MUST start with "KEM" and identify this KEM algorithm; if used in the remainder of HPKE, it MUST start with "HPKE" and identify the entire ciphersuite in use. 

Right now the callers are constructing the correct form as passing `&[u8]`, but I think another enum like `Label` would be better. Something like:

```rust
enum LabeledSuiteId {
  Hpke(HpkeSuite),
  Kem(HpkeKem),
}
```

rustls/src/crypto/hpke.rs
cpu: Good point, it'll be ~1 so I think you're right. I'll remove.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: > Looking again, I guess that doesn't work because we have to create a new SealingKey based on the nonce_schedule for every seal() operatiion?

That's correct. Should I go back to the original form?

cpu: Reverted to keeping the AEAD bits in the key schedule.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Suite ID fixed up to use an enum.

rustls/src/crypto/hpke.rs
cpu: Removed.

provider-example/src/hpke.rs
cpu: Moved in the trait & impls.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: Extracted.

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: This ends up falling away with the idea to use an enum for the suite ID arg.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,846 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.suite.fmt(f)
+    }
+}
+
+/// Default [RFC 9180] Hybrid Public Key Encryption (HPKE) suites supported by aws-lc-rs cryptography.
+pub static ALL_SUPPORTED_SUITES: &HpkeSuites = &HpkeSuites(&[
+    DHKEM_P256_HKDF_SHA256_AES_128,
+    DHKEM_P256_HKDF_SHA256_AES_256,
+    DHKEM_P256_HKDF_SHA256_CHACHA20_POLY1305,
```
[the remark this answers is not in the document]
cpu: > Do we need to drop these (this and the subsequent CHACHA20_POLY1305 suites) based on the fips crate feature?

Done - these are omitted from `ALL_SUPPORTED_SUITES` based on the `fips` feature. I've left the actual implementations in place, and they will return `false` from `fips()`. I think that's what we want here, or should the `fips` crate feature make ref'ing the non-FIPS impls impossible by ensuring they aren't compiled at all?

> do we want to have a fips() function on HpkeSuites and Hpke?

The `HpkeSuites` type fell away, but I've added a `fips()` function on `dyn Hpke`.

> do we want to add a fips() function on ClientConnection/ServerConnection that delegates to the config version, but also takes into account of the crypto backing ECH?

Let's chat about this in the ECH PR.

cpu:
Ready for another round of review :-)

rustls/src/crypto/aws_lc_rs/hpke.rs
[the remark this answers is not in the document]
cpu: > Let's chat about this in the ECH PR.

Left a comment over there.

ctz (dismissed):

cpu:
> cpu force-pushed the cpu-aws-lc-rs-hpke branch from 5b444bf to 9c23e93 

Rebased to pick up the pinned nightly version to fix CI failures from the upstream nightly rustc bug.

cpu:

Need another rebase, and I want to get the ECH branch rebased and passing the bogo suite using this HPKE provider before we merge.

cpu:
>  cpu force-pushed the cpu-aws-lc-rs-hpke branch from 9c23e93 to 495a821 

Rebased, and reworked with some substantial changes that fell out of more testing.

1. I had made a mistake translating the labelled extract operations used to derive two of the inputs to the key schedule context based on a misunderstanding I had about the aws-lc-rs (and ring) HKDF APIs. To fix this mistake we need a capability the existing traits/upstream APIs can't provide: extracting a PRK from an HKDF without immediately expanding it. Fixing this ended up not being too difficult (details in commit message and PR desc, thanks to ctz for an assist with this). The changes I had previously needed to re-arrange/export the `RingHkdf` impls from the `SupportedCipherSuites` fell away as a result.

2. Based on the above I also wanted more testing. The pre-existing tests didn't catch this because we're not using the RFC test vectors as intended (more on that later), but were doing a simple round-trip within our own impl. For the short term I've reworked this branch to keep the hpke-rs based provider and added a new `rustls-provider-test` crate that can run integration tests that check the inter-operation of the two HPKE providers (e.g. sealing with one, opening with the other). 

3. I'd like to use more of the test vectors (e.g. matching our produced ciphertexts with the expected values). To do this I need to be able to re-jig the DH KEM so that it produces a fixed secret key from a test vector instead of randomly generating one. Doing that in a minimally invasive way so that tests can use this mechanism but nothing else will take another revision. I haven't finished that part so I've left this in draft.

TL;DR - re-reviews needed since some important implementation details changed. More test coverage coming soon.

provider-example/src/hpke.rs
```diff
@@ -5,77 +5,124 @@ use core::fmt::{Debug, Formatter};
...
         plaintext: &[u8],
         pub_key: &HpkePublicKey,
     ) -> Result<(EncapsulatedSecret, Vec<u8>), Error> {
         let pk_r = hpke_rs::HpkePublicKey::new(pub_key.0.clone());
         let (enc, ciphertext) = self
-            .0
+            .start()?
             .seal(&pk_r, info, aad, plaintext, None, None, None)
             .map_err(other_err)?;
         Ok((EncapsulatedSecret(enc.to_vec()), ciphertext))
```
jplock, line 115: When using AES GCM, can the `nonce` value be included in the `EncapsulatedSecret` (or separately returned) as the same value is required when opening

cpu:
> When using AES GCM, can the nonce value be included in the EncapsulatedSecret (or separately returned) as the same value is required when opening

GitHub ate my more detailed response when the comment was deleted :'( The summary: the AEAD nonce is computed from the key schedule context in HPKE, and not something that needs to be given to the caller. The AEAD tag is included in the returned ciphertext. There's tests in the rustls-provider-test crate that show the seal/open working without any additional context/data required in case that's helpful to verify what I'm saying tracks with the impl.

cpu:
> I'd like to use more of the test vectors (e.g. matching our produced ciphertexts with the expected values).

Done. I tried a few different approaches before settling on the one I pushed in 81ae6ae - I think it's the least invasive solution but would be open to alternatives if anyone has smarter ideas. 

Some notes about the new test:

* We're now matching the `Enc` returned from creating a `Sealer`, and the ciphertext returned from using it to seal the test vector plaintext with the test vector expected values.
* I didn't make the even more invasive changes that would be required to match intermediate values (`base_nonce`, `shared_secret`, `key_schedule_context`, etc) with the test vector values. I don't think there's much value here since we're now matching the user-facing values that are produced using those intermediates. If we messed something up with the intermediate values the things we _are_ asserting on would be wrong.
* We end up duplicating a little bit of test scaffolding between the aws-lc-rs unit test and the rustls-provider-test integration test but it felt reasonable and I don't know there's a better way - I really don't want to make the ability to fiddle with the DH KEM available outside of the core crate.

> I want to get the ECH branch rebased and passing the bogo suite using this HPKE provider before we merge.

Done: [cpu-aws-lc-rs-hpke-client-ech](https://github.com/cpu/rustls/tree/cpu-aws-lc-rs-hpke-client-ech) - I haven't pushed an update to the PR branch yet (need to sort out BOGO w/ *ring* skipping ECH tests) but it works in the aws-lc-rs case (meaning we have HPKE interop with a 3rd Go based impl) and against some of the test servers (themselves using a 4th HPKE impl).

Going to flip this out of draft status. I think it's ready to merge as soon as reviewers are happy.

provider-example/src/hpke.rs
```diff
@@ -54,7 +93,9 @@ impl HpkeRs {
 
 impl Debug for HpkeRs {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("HpkeRsHpke").finish()
+        f.debug_struct("HpkeRs")
```
djc, line 96: Nit: maybe we can just derive this now?

rustls/src/crypto/tls13.rs
```diff
@@ -94,6 +77,21 @@ impl<'a> Hkdf for HkdfUsingHmac<'a> {
...
+    fn extract_prk_from_secret(&self, salt: Option<&[u8]>, secret: &[u8]) -> Vec<u8> {
+        let zeroes = [0u8; hmac::Tag::MAX_LEN];
+        let salt = match salt {
+            Some(salt) => salt,
+            None => &zeroes[..self.0.hash_output_len()],
+        };
+        self.0
+            .with_key(salt)
+            .sign(&[secret])
+            .as_ref()
```
djc, line 90: Nit: `.as_ref().into()` is a little opaque, maybe there is a more direct way to express this? Could at least use `to_vec()` or `to_owned()` instead of `into()`?

djc (approved):

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,945 @@
...
+        labeled_expand(
+            suite_id,
+            labeled_extract_for_expand(self.hkdf, suite_id, None, Label::EaePrk, dh),
+            Label::SharedSecret,
+            kem_context,
+        )
+    }
+}
+
+static DHKEM_P256_HKDF_SHA256: &DhKem<SHA256_OUTPUT_LEN> = &DhKem {
```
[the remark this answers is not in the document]
cpu: Renamed.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,945 @@
...
+    DHKEM_X25519_HKDF_SHA256_AES_128,
+    #[cfg(not(feature = "fips"))]
+    DHKEM_X25519_HKDF_SHA256_AES_256,
+    #[cfg(not(feature = "fips"))]
+    DHKEM_X25519_HKDF_SHA256_CHACHA20_POLY1305,
+];
+
+/// HPKE suite using ECDH P-256 for agreement, HKDF SHA-256 for key derivation, and AEAD AES-128-GCM
+/// for symmetric encryption.
+pub static DHKEM_P256_HKDF_SHA256_AES_128: &HpkeAwsLcRs<AES_128_KEY_LEN, SHA256_OUTPUT_LEN> =
```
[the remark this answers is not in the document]
cpu: Renamed.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,945 @@
...
+        secret_key: &HpkePrivateKey,
+    ) -> Result<Box<dyn HpkeOpener + 'static>, Error> {
+        Ok(Box::new(Opener::new(self, enc, info, secret_key)?))
+    }
+
+    fn suite(&self) -> HpkeSuite {
+        self.suite
+    }
+
+    fn fips(&self) -> bool {
```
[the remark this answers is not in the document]
cpu: Agreed. Changed the order in the trait, and this impl.

rustls/src/crypto/ring/hmac.rs
```diff
@@ -5,12 +5,9 @@ use alloc::boxed::Box;
 use super::ring_like;
 use crate::crypto;
 
-#[cfg(feature = "tls12")]
 pub(crate) static HMAC_SHA256: Hmac = Hmac(&ring_like::hmac::HMAC_SHA256);
-#[cfg(feature = "tls12")]
 pub(crate) static HMAC_SHA384: Hmac = Hmac(&ring_like::hmac::HMAC_SHA384);
-#[cfg(test)]
-#[allow(dead_code)] // only for TLS1.2 prf test
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
cpu: I added a comment.

I also tried to use a `cfg_attr` to only allow the warning for `not(any(test, feature = "aws_lc_rs"))` - that doesn't work nicely because of the case where both `ring` and `aws-lc-rs` are enabled. 

In that situation we still get a warning because the `hmac.rs` file is used by both ring and aws-lc-rs via the `#[path = "../ring/hmac.rs"]` workaround, and in the ring case it ends up unused but in the aws-lc-rs it isn't. Gating by feature alone isn't sufficient and I can't think of another solution for that so an unconditional allow (as was already used) seems best.

rustls/src/crypto/tls13.rs
cpu: Changed to use `to_vec()`.

provider-example/src/hpke.rs
cpu: Switched to a derived `Debug` in the commit that reworked `HpkeRs` to only hold a `HpkeSuite`. That's the earliest point the change became possible.

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,945 @@
...
+use crate::crypto::hpke::{
+    EncapsulatedSecret, Hpke, HpkeOpener, HpkePrivateKey, HpkePublicKey, HpkeSealer, HpkeSuite,
+};
+use crate::crypto::tls13::{expand, HkdfExpander, HkdfPrkExtract, HkdfUsingHmac};
+use crate::msgs::enums::{HpkeAead, HpkeKdf, HpkeKem};
+use crate::msgs::handshake::HpkeSymmetricCipherSuite;
+use crate::{Error, OtherError};
+
+/// `HpkeAwsLcRs` holds the concrete instantiations of the algorithms specified by the [HpkeSuite].
+pub struct HpkeAwsLcRs<const KEY_SIZE: usize, const KDF_SIZE: usize> {
```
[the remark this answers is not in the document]
cpu: Re-reading the [style guide ordering section](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#ordering) it doesn't make any mention of `static`'s, just constants. 

I think the usual top-down ordering would apply in this case, so `ALL_SUPPORTED_SUITES` first (depends on the individual statics), then the individual statics (depends on the `HpkeAwsLcRs` type), then the `HpkeAwsLcRs` type and impl.

I'll switch to that. If that isn't the right interpretation we might need a CONTRIBUTING.md update to clarify so I'm not as confused in the future :laughing:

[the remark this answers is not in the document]
djc: Yup, that's what I meant!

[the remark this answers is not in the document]
cpu: Perfect, thanks!

rustls/src/crypto/aws_lc_rs/hpke.rs
```diff
@@ -0,0 +1,945 @@
...
+    #[cfg(not(feature = "fips"))]
+    DH_KEM_P384_HKDF_SHA384_CHACHA20_POLY1305,
+    DH_KEM_P521_HKDF_SHA512_AES_128,
+    DH_KEM_P521_HKDF_SHA512_AES_256,
+    #[cfg(not(feature = "fips"))]
+    DH_KEM_P521_HKDF_SHA512_CHACHA20_POLY1305,
+    #[cfg(not(feature = "fips"))]
+    DH_KEM_X25519_HKDF_SHA256_AES_128,
+    #[cfg(not(feature = "fips"))]
+    DH_KEM_X25519_HKDF_SHA256_AES_256,
```
ctz, line 40: Is this gated out because of X25519? That is inconsistent with what we're doing elsewhere. But having looked into it in detail this morning, I think what you're doing here is correct.

ctz (approved):

rustls/src/crypto/aws_lc_rs/hpke.rs
cpu: > Is this gated out because of X25519? 

Correct.

>  But having looked into it in detail this morning, I think what you're doing here is correct.

I did some research at the time and arrived at the conclusion X25519 for kex wasn't permitted. I think only ed25519 for signatures is permitted. My gut instinct was to mark it as not FIPS approved and wait for someone with a stronger stomach for wading through NIST documents to make a case for why it is :laughing: 

This also has some interplay with the ECH spec, where [section 9](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18#section-9) says:

> In the absence of an application profile standard specifying otherwise, a compliant ECH application MUST implement the following HPKE cipher suite:
>
> KEM: DHKEM(X25519, HKDF-SHA256) (see Section 7.1 of [HPKE])
> > KDF: HKDF-SHA256 (see Section 7.2 of [HPKE])
> AEAD: AES-128-GCM (see Section 7.3 of [HPKE])

I view "FIPS" as "an application profile standard specifying otherwise", and it felt like having ECH w/o X25519 for FIPS users was better than nothing. :man_shrugging:

PR #1964, opened 2024-05-22
Certificate compression preparation

ctz:
This is, like #1962, some commits that I think can be landed separate from certificate compression itself.

The last commit will upset the semver checks job. We could cut 0.23.8 shortly after this, or else drop that commit until later.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -143,7 +143,7 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
 /// * [`ClientConfig::key_log`]: key material is not logged.
 ///
 /// [`RootCertStore`]: crate::RootCertStore
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct ClientConfig {
     /// Which ALPN protocols we include in our client hello.
     /// If empty, no ALPN extension is sent.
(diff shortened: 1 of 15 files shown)


rustls/src/client/tls13.rs
```diff
@@ -786,20 +786,12 @@ fn emit_certificate_tls13(
     auth_context: Option<Vec<u8>>,
     common: &mut CommonState,
 ) {
-    let context = auth_context.unwrap_or_default();
-
-    let mut cert_payload = CertificatePayloadTls13 {
-        context: PayloadU8::new(context),
-        entries: Vec::new(),
+    let mut cert_payload = if let Some(certkey) = certkey {
```
djc, line 789: Nit: suggest using a `match` here instead of `if let`. I guess we can't hoist out the `new()` call because the iterators have different types?

rustls/src/msgs/handshake.rs
```diff
@@ -1449,10 +1449,32 @@ impl<'a> Codec<'a> for CertificatePayloadTls13<'a> {
...
 impl<'a> CertificatePayloadTls13<'a> {
-    pub(crate) fn new(entries: Vec<CertificateEntry<'a>>) -> Self {
+    pub(crate) fn new(
+        certs: impl Iterator<Item = &'a CertificateDer<'a>>,
+        ocsp_response: Option<&'a [u8]>,
+    ) -> Self {
         Self {
             context: PayloadU8::empty(),
-            entries,
+            entries: certs
```
djc, line 1458: Nit: IMO this commit would benefit from being split into smaller parts.

djc (approved):
Nice!

rustls/src/client/client_conn.rs
```diff
@@ -143,7 +143,7 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
 /// * [`ClientConfig::key_log`]: key material is not logged.
 ///
 /// [`RootCertStore`]: crate::RootCertStore
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
cpu, line 146: Nice :+1:

cpu (approved):
LGTM

rustls/src/client/tls13.rs
ctz: actually, i can hoist out the subject of the iterator, into `certkey.map(|ck| ck.cert.as_ref()).unwrap_or(&[][..]);`

Issue #436, opened 2021-01-05
Add support for RFC 8879 certificate compression


djc:
Might be interesting to have, since it supposedly improves performance.

https://tools.ietf.org/html/rfc8879

sharksforarms:
Submitted a WIP PR for this, let me know what you think! and I can make adjustments

briansmith:
I understand the desire to avoid the compression library dependencies but also nobody would have any clue about how to use this API and so the benefit of adding this as-is would be very limited. I think if we want to do this then we should have (perhaps in a different crate) a very easy-to-use mechanism to actually use this with just a couple of lines of code.

Strawman:
* Add an rustls-compres-crate to the workspace of this repo for each compression algorithm we want to support.
* Only bother to support what's currently considered the best compression algorithm that is commonly used.
* Add compression support to the client and server demos that demonstrates which 1-3 lines of code to add to an application to get the compression support working.
* Document why we chose the algorithm(s) we did.
* Use pure Rust compression libraries without `unsafe` code, or if those aren't competitive performance-wise, explain the situation.

briansmith:
Specifically, I suggest we add flags to `tlsclient` and `tlsserver` that enable/disable compression, with the default being "on" for all the algorithms we support.

PR #1966, opened 2024-05-23
Support RFC8879 certificate compression

ctz:
So far this supports compression and decompression of the server certificate chain, with zlib or brotli.

The protocol support is there unconditionally, the `zlib-rs` and `brotli` optional dependencies activate support for those algorithms. Alternatively people can avoid those dependencies, and insert their own implementations into (eg) `ServerConfig::cert_compressors`.

~~The "Store precompressed certificate chain" commit contains a breaking change, because it introduces a new private field on a struct that previously was entirely public. We could do this in a different way by caching some compressions somewhere, which would be more transparent.~~

TODO:

- [x] compression of the client certificate chain (by the client)
- [x] decompression of the client certificate chain (by the server)

Resolves https://github.com/rustls/rustls/issues/436


rustls/src/compress.rs
```diff
@@ -75,3 +81,67 @@ pub struct DecompressionFailed;
...
+
+        fn compress(
+            &self,
+            input: Vec<u8>,
+            level: CompressionLevel,
+        ) -> Result<Vec<u8>, CompressionFailed> {
+            let mut output = alloc::vec![0u8; deflate::compress_bound(input.len())];
+            let config = match level {
+                CompressionLevel::Interactive => deflate::DeflateConfig::default(),
+                CompressionLevel::Amortized => deflate::DeflateConfig::new(9),
```
cpu, line 136: Is there a citation to use for this magic number? The [zlib-rs docs](https://docs.rs/zlib-rs/latest/zlib_rs/deflate/struct.DeflateConfig.html#method.new) offer no commentary, it's just a bare `i32`. Some googling around leads me to believe the effort level is 0..9 with 9 being "the most effort". Maybe sufficient to leave a comment here saying as much?

rustls/Cargo.toml
```diff
@@ -17,6 +17,7 @@ rustversion = { version = "1.0.6", optional = true }
 
 [dependencies]
 aws-lc-rs = { version = "1.6", optional = true, default-features = false, features = ["aws-lc-sys"] }
+brotli = { version = "6", optional = true, default-features = false, features = ["std"] }
```
cpu, line 20: Seems reasonable to not worry about a no-std compatible impl of this at this stage :+1:

rustls/src/compress.rs
```diff
@@ -228,3 +228,85 @@ mod feat_brotli {
 
 #[cfg(feature = "brotli")]
 pub use feat_brotli::{BrotliCompressor, BrotliDecompressor};
+
+#[cfg(all(test, any(feature = "brotli", feature = "zlib-rs")))]
+pub mod tests {
```
cpu, line 233: The tests look good, but I think the ordering is inverted. I'd expect to see the `#[test]` fns up front and the helpers afterwards.

rustls/src/msgs/handshake.rs
```diff
@@ -968,18 +968,14 @@ impl ClientHelloPayload {
...
-                    return true;
-                }
-            }
+            has_duplicates::<_, _, u16>(
+                entries
+                    .iter()
+                    .map(|kse| u16::from(kse.group)),
+            )
+        } else {
+            false
```
cpu, line 977: nit: I think this could be a separate commit since it's unrelated to cert compression.

rustls/src/server/tls13.rs
```diff
@@ -158,6 +159,26 @@ mod client_hello {
...
+                    // prefer server order when choosing a compression: the client's
+                    // extension here does not denote any preference.
+                    self.config
+                        .cert_compressors
+                        .iter()
+                        .find(|compressor| offered.contains(&compressor.algorithm()))
+                        .cloned()
+                } else {
+                    None
+                };
```
cpu, line 180: I think you can drop the `else` and simplify this a bit chaining an `and_then()` off of the extension optional:

```rust
 let cert_compressor = client_hello.certificate_compression_extension().and_then(|offered|
                // prefer server order when choosing a compression: the client's
                // extension here does not denote any preference.
                self.config
                    .cert_compressors
                    .iter()
                    .find(|compressor| offered.contains(&compressor.algorithm()))
                    .cloned()
            );

```

rustls/src/server/tls13.rs
```diff
@@ -715,6 +747,43 @@ mod client_hello {
...
+    ) {
+        let encoding =
+            CertificatePayloadTls13::new(cert_chain.iter(), ocsp_response).get_encoding();
+        let uncompressed_len = encoding.len() as u32;
+
+        let compressed = match cert_compressor.compress(encoding, CompressionLevel::Interactive) {
+            Ok(compressed) => PayloadU24(Payload::new(compressed)),
+            Err(_) => {
+                return emit_certificate_tls13(transcript, common, cert_chain, ocsp_response);
+            }
```
cpu: This fall-back to uncompressed certs seems interesting. I think it's a valid strategy based on 8879 indicating that peers _MAY_ compress when support is negotiated :+1: I don't see any text in the RFC that discusses this case concretely to offer any counter advice.

rustls/src/server/tls13.rs
```diff
@@ -715,6 +747,43 @@ mod client_hello {
...
+        };
+
+        let body = CompressedCertificatePayload {
+            alg: cert_compressor.algorithm(),
+            uncompressed_len,
+            compressed,
+        };
```
cpu, line 794: nit: I would be tempted to inline the `body` binding into the `Message` init on L778. It isn't ref'd anywhere.

rustls/tests/api.rs
```diff
@@ -3455,7 +3455,7 @@ fn vectored_write_for_server_handshake_with_half_rtt_data() {
         let mut pipe = OtherSession::new(&mut client);
         let wrlen = server.write_tls(&mut pipe).unwrap();
         // don't assert exact sizes here, to avoid a brittle test
-        assert!(wrlen > 3000); // its pretty big (contains cert chain)
+        assert!(wrlen > 2500); // its pretty big (contains cert chain)
```
cpu, line 3458: Nice :laughing:

rustls/src/crypto/signer.rs
```diff
@@ -102,7 +109,41 @@ impl CertifiedKey {
             cert,
             key,
             ocsp: None,
+            precompressed_zlib: None,
+            precompressed_brotli: None,
+        }
+    }
+
+    /// Construct, compress and save the TLS1.3 Certificate message.
+    pub fn precompress(
```
cpu, line 118: This is clever :+1:

cpu (commented):
Nice! Not a bad diff at all. Great work.

I'd characterize my review pass as pretty light but I'll do a second more in-depth pass with the RFC in hand in the morning when my brain has more juice. Here's some initial comments:

rustls/src/msgs/handshake_test.rs
```diff
@@ -811,6 +819,14 @@ fn get_sample_certificatepayloadtls13() -> CertificatePayloadTls13<'static> {
     }
 }
 
+fn get_sample_compressedcertificate() -> CompressedCertificatePayload<'static> {
```
cpu, line 822: ordering nit: I think this should be below `get_sample_serverkeyexchangepayload_ecdhe()`.

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+pub trait CertDecompressor: Debug + Send + Sync {
+    /// Which algorithm this decompressor handles.
+    fn algorithm(&self) -> CertificateCompressionAlgorithm;
+
+    /// Decompress `input`, writing the result to `output`.
+    ///
+    /// `output` matches the declared length of the decompressed data.  The
+    /// decompression process must precisely fill `output`.  If it produces
+    /// more or fewer bytes than fit in `output`, `Err(DecompressionFailed)` must
+    /// be returned.
```
cpu, line 18: Perhaps clearer as:
"`output` is sized to match the declared length of the decompressed data."?

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// `Err(DecompressionFailed)` should be returned if the `input` is in any
+    /// way malformed.
```
cpu, line 21: Maybe the sentence about returning this error for the case where the # of bytes produced is a mismatch to `output` size should be rolled into this sentence? Something like:

> `Err(DecompressionFailed)` should be returned if decompression produces more, or fewer bytes than fit in `output`, or if the `input` is in any way malformed.

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Returns the supported `CertDecompressor` implementations enabled
+/// by crate features.
+pub fn default_cert_decompressors() -> &'static [&'static dyn CertDecompressor] {
+    &[]
+}
+
+/// An available certificate compression algorithm.
+pub trait CertCompressor: Debug + Send + Sync {
+    /// Which algorithm this compressor handles.
+    fn algorithm(&self) -> CertificateCompressionAlgorithm;
```
cpu, line 39: ditto RE: ordering nit.

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// `Err(CompressionFailed)` may be returned for any reason.
+    fn compress(
+        &self,
+        input: Vec<u8>,
+        level: CompressionLevel,
+    ) -> Result<Vec<u8>, CompressionFailed>;
+}
+
+/// A hint for how many resources to dedicate to a compression.
+#[derive(Debug, Copy, Clone, PartialEq)]
```
cpu, line 57: This feels like a case where deriving `Eq` is also appropriate. Am I missing a reason to avoid that?

rustls/src/compress.rs
```diff
@@ -75,3 +81,67 @@ pub struct DecompressionFailed;
...
+        fn decompress(&self, input: &[u8], output: &mut [u8]) -> Result<(), DecompressionFailed> {
+            let (output_filled, rc) =
+                inflate::uncompress_slice(output, input, inflate::InflateConfig::default());
+            if rc != ReturnCode::Ok {
+                return Err(DecompressionFailed);
+            }
+
+            if output_filled.len() != output.len() {
+                return Err(DecompressionFailed);
+            }
```
cpu, line 113: nit: maybe fold these into one `if`?

```rust
if rc != ReturnCode::Ok || output_filled.len() != output.len() {
    return Err(DecompressionFailed);
}
```

rustls/src/compress.rs
```diff
@@ -75,3 +81,67 @@ pub struct DecompressionFailed;
...
+                CompressionLevel::Interactive => deflate::DeflateConfig::default(),
+                CompressionLevel::Amortized => deflate::DeflateConfig::new(9),
+            };
+            let (output_filled, rc) = deflate::compress_slice(&mut output, &input, config);
+            if rc != ReturnCode::Ok {
+                return Err(CompressionFailed);
+            }
+
+            let used = output_filled.len();
+            output.truncate(used);
```
cpu, line 144: nit: I think the `used` binding doesn't add much here. Inline?

rustls/src/client/client_conn.rs
```diff
@@ -215,6 +216,15 @@ pub struct ClientConfig {
 
     /// How to verify the server certificate chain.
     pub(super) verifier: Arc<dyn verify::ServerCertVerifier>,
+
+    /// How to decompress the server's certificate chain.
+    ///
+    /// If this is non-empty, the [RFC8779] certificate compression
+    /// extension is offered, and any compressed certificates are
+    /// transparently decompressed during the handshake.
```
cpu, line 224: I think there should be a mention here that the extension is only offered for TLS 1.3:

> This extension is only supported with TLS 1.3 [RFC8446] and newer; if TLS 1.2 [RFC5246] or earlier is negotiated, the peers MUST ignore this extension.

In the ECH PR we tried to make it hard to enable ECH without also enabling TLS1.3 as a protocol version. Should there be some consideration given to that here?

rustls/src/server/server_conn.rs
```diff
@@ -337,6 +338,15 @@ pub struct ServerConfig {
 
     /// Provides the current system time
     pub time_provider: Arc<dyn TimeProvider>,
+
+    /// How to compress the server's certificate chain.
+    ///
+    /// If a client supports this extension, and advertises support
+    /// for one of the compression algorithms included here, the
+    /// server certificate will be compressed according to [RFC8779].
```
cpu, line 346: ditto here RE: TLS 1.3 discussion

cpu (commented):
My promised second pass, this time done with a bit more care w.r.t matching against RFC 8879.

The zlib-rs parts were a bit tricky to review because the upstream seems to have no rustdocs. I filed an issue to ask about that, but was able to make do between reading the code and looking at the C zlib docs.

rustls/src/compress.rs
cpu: Realized after the fact this is load bearing, it prevents two mutable borrows of `output` :+1:

rustls/src/client/client_conn.rs
ctz: I don't think so. One of my goals in this PR is (aside from opting-in via the crate features) avoiding any explicit configuration to have things work in the common case (and, eg, requiring that callers clear out `ClientConfig::cert_decompressors` to use TLS1.2 wouldn't gel with that). I think that is appropriate for what should be treated as an opportunistic performance improvement.

(Whereas with ECH, it is actively harmful to privacy to get the mandatory configuration wrong.)

ctz: Will add a note to this comment.

rustls/src/enums.rs
```diff
@@ -69,6 +69,7 @@ enum_builder! {
         CertificateURL => 0x15,
```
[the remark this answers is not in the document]
ctz: Added a commit to drop this.

rustls/Cargo.toml
ctz: I couldn't work out how to use this crate without std, basically :)

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn compress(
+        &self,
+        input: Vec<u8>,
+        level: CompressionLevel,
+    ) -> Result<Vec<u8>, CompressionFailed>;
+}
+
+/// A hint for how many resources to dedicate to a compression.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum CompressionLevel {
```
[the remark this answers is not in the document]
ctz: I only looked at boringssl, and it doesn't (ref https://github.com/google/boringssl/blob/afd52e91dfed27ab7193be040f067900947b14ac/include/openssl/ssl.h#L3216-L3260 ).

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// `Err(DecompressionFailed)` should be returned if the `input` is in any
+    /// way malformed.
+    fn decompress(&self, input: &[u8], output: &mut [u8]) -> Result<(), DecompressionFailed>;
+}
+
+/// Returns the supported `CertDecompressor` implementations enabled
+/// by crate features.
+pub fn default_cert_decompressors() -> &'static [&'static dyn CertDecompressor] {
+    &[]
+}
```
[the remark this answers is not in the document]
ctz: moved up

rustls/src/compress.rs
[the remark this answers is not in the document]
cpu: :+1: Just an idle curiosity on my side, doesn't merit deep digging. Thanks!

rustls/src/client/client_conn.rs
cpu: This makes sense to me :+1: Thanks for adding a comment.

rustls/src/compress.rs
ctz: In zlib this is `#define Z_BEST_COMPRESSION       9` and actually I can get that out of zlib-rs::c_api -- done.

rustls/src/compress.rs
ctz: sorry -- old habits die hard -- fixed.

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use alloc::vec::Vec;
+use core::fmt::Debug;
+
+use crate::enums::CertificateCompressionAlgorithm;
+
+/// An available certificate decompression algorithm.
+pub trait CertDecompressor: Debug + Send + Sync {
+    /// Which algorithm this decompressor handles.
+    fn algorithm(&self) -> CertificateCompressionAlgorithm;
```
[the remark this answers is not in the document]
ctz: done this, for both traits and both impls

rustls/src/msgs/handshake_test.rs
cpu: > i think the current ordering is in keeping with the rest of the file (which is extremely bottom-up).

Reasonable :+1:

rustls/src/msgs/handshake.rs
```diff
@@ -2302,7 +2302,7 @@ impl<'a> CertificateStatus<'a> {
 
 // -- RFC8879 compressed certificates
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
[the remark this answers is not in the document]
ctz: Dropped this commit.

rustls/src/compress.rs
```diff
@@ -223,6 +229,144 @@ mod feat_brotli {
...
+            .map_err(|_| CompressionFailed)?;
+        if cache.len() == self.size {
+            cache.remove(0);
+        }
+        cache.push(Arc::clone(&new_entry));
+        Ok(new_entry)
+    }
+
+    #[cfg(not(feature = "std"))]
+    /// Create a dummy compression cache.
```
cpu, line 313: nit: I think we usually order this rustdoc -> cfg gates -> impl no?

rustls/src/compress.rs
```diff
@@ -223,6 +229,144 @@ mod feat_brotli {
...
+    ///
+    /// The cache is not supported for no_std builds.
```
cpu, line 315: This reads a little awkwardly, maybe: "No caching is performed by the dummy cache. It is primarily useful for no_std builds" ?

rustls/src/compress.rs
```diff
@@ -223,6 +229,144 @@ mod feat_brotli {
...
+    pub fn new(_unused: usize) -> Self {
+        Self {}
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub(crate) fn compression_for(
```
cpu, line 321: Instead of having one `CompressionCache` impl with two sets of functions, each gated by the `std` configuration, would it be more natural to have a trait and two impls? One that's a no-op cache for no-std and one that does caching? I think the only shared functionality is `uncached_compression` and I think that could be a free-standing fn both use. WDYT?

rustls/src/compress.rs
```diff
@@ -223,6 +229,144 @@ mod feat_brotli {
...
+            compressed: CompressedCertificatePayload {
+                alg: algorithm,
+                uncompressed_len,
+                compressed: PayloadU24(Payload::new(compressed)),
+            },
+        }))
+    }
+}
+
+#[cfg_attr(not(feature = "std"), allow(dead_code))]
```
cpu, line 353: I'm surprised this dead code allow is needed for no-std, isn't this fn ref'd by `pub(crate) compression_for(...)` in the `no-std` case?

cpu (approved):

rustls/src/compress.rs
```diff
@@ -223,6 +229,144 @@ mod feat_brotli {
...
+            .map_err(|_| CompressionFailed)?;
+        for (i, item) in cache.iter().enumerate() {
+            if item.algorithm == algorithm && item.original == encoding {
+                // this item is now MRU
+                let item = cache.remove(i);
+                cache.push(Arc::clone(&item));
+                return Ok(item);
+            }
+        }
+        drop(cache);
```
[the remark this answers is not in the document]
djc: Given that this is pretty linear and there's not much else happening between the binding and the drop, I think this is probably fine.

rustls/src/compress.rs
ctz: The complaint this addresses is about `CompressionCacheEntry::algorithm` and `CompressionCacheEntry::original` being written but never read.

cpu: Ahh! Ok that makes sense.

I found it's a little awkward to be using the `CompressionCacheEntry` for both the cached use case (where these fields are read, and the `original` is non-empty) and the non-cached case (where the fields are never read, and `original` is always empty) but I couldn't think of a better approach.

I think it's OK as-is. Thanks for explaining the `allow`.

rustls/src/compress.rs
ctz: I had a look at this and agree it's an improvement. To prevent this trait and the types it involves (`CertificatePayloadTls13`, `CompressionCacheEntry`), etc leaking into the public API I think I will have to add some localised disabling of the `private_interfaces` lint. I guess that lint is to prevent private traits from accidentally making unusable bits of public API, but hopefully it's alright here?

cpu: >  I guess that lint is to prevent private traits from accidentally making unusable bits of public API, but hopefully it's alright here?

Yeah, reading the [proposal that added the lints](https://rust-lang.github.io/rfcs/2145-type-privacy.html#private-in-public-lints) I think your summary is correct, and it doesn't seem applicable in this case.

I would be OK with a localized ignore. @djc WDYT?

cpu: If it turns into a can of worms the original approach didn't seem outrageous to me.

rustls/Cargo.toml
```diff
@@ -26,6 +26,7 @@ subtle = { version = "2.5.0", default-features = false }
 webpki = { package = "rustls-webpki", version = "0.102.4", features = ["alloc"], default-features = false }
 pki-types = { package = "rustls-pki-types", version = "1.7", features = ["alloc"] }
 zeroize = "1.7"
+zlib-rs = { version = "0.1", optional = true }
```
djc, line 31: I think it would be nice to front this with an explicitly named feature `zlib = ["zlib-rs"]` like we do for brotli. This is generally the better, more flexible approach (AIUI implicit features might be going away in 2024 edition).

README.md
```diff
@@ -57,7 +57,7 @@ to a wider set of architectures and environments, or compliance requirements.  S
 Specifying `default-features = false` when depending on rustls will remove the
 dependency on aws-lc-rs.
 
-Rustls requires Rust 1.63 or later.
+Rustls requires Rust 1.63 or later. It has an optional dependency on zlib-rs which requires 1.75 or later.
```
djc, line 60: Maybe these changes should be in the commit that introduces the use of zlib-rs? I think it might be nice to also make explicit that the zlib-rs dependency is "optional (and disabled by default)".

rustls/src/client/client_conn.rs
```diff
@@ -215,6 +216,18 @@ pub struct ClientConfig {
...
+    ///
+    /// This only applies to TLS1.3 connections.  It is ignored for
+    /// TLS1.2 connections.
+    ///
+    /// [RFC8779]: https://datatracker.ietf.org/doc/rfc8879/
+    pub cert_decompressors: Vec<&'static dyn compress::CertDecompressor>,
```
djc, line 232: Maybe this could/should be `Arc<[..]>` instead?

rustls/src/msgs/handshake.rs
```diff
@@ -1023,6 +1023,24 @@ impl ClientHelloPayload {
         self.find_extension(ExtensionType::EarlyData)
             .is_some()
     }
+
+    pub(crate) fn certificate_compression_extension(
+        &self,
+    ) -> Option<&Vec<CertificateCompressionAlgorithm>> {
```
djc, line 1029: Would be more idiomatic to yield a slice here.

rustls/src/server/tls13.rs
```diff
@@ -748,36 +749,34 @@ mod client_hello {
...
+                            entry.compressed_cert_payload(),
+                        ),
+                    }),
+                };
+
+                trace!("sending compressed certificate {:?}", c);
+                transcript.add_message(&c);
+                common.send_msg(c, true);
+            }
+            Err(_) => emit_certificate_tls13(transcript, common, cert_chain, ocsp_response),
```
djc, line 805: Linearize this by yielding out the error path first?

rustls/src/compress.rs
djc: This might be easier if you wrap the fields inside `CompressionCache::Enabled` into a private type?

ctz: Thanks. Settled on an enum-based formula for this which seems the most tidy.

rustls/src/compress.rs
```diff
@@ -72,3 +78,64 @@ pub struct DecompressionFailed;
...
+#[cfg(feature = "zlib-rs")]
+mod feat_zlib_rs {
+    use zlib_rs::c_api::Z_BEST_COMPRESSION;
+    use zlib_rs::{deflate, inflate, ReturnCode};
+
+    use super::*;
+
+    /// A certificate decompressor for the Zlib algorithm using the `zlib-rs` crate.
+    #[derive(Debug)]
+    pub struct ZlibRsDecompressor;
```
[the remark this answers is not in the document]
ctz: I have done that. It's also more in keeping with other instances of the same pattern elsewhere in the crate.

rustls/src/msgs/handshake.rs
```diff
@@ -956,18 +956,14 @@ impl ClientHelloPayload {
...
-            let mut seen = BTreeSet::new();
-
-            for kse in entries {
-                let grp = u16::from(kse.group);
-
-                if !seen.insert(grp) {
-                    return true;
-                }
-            }
+            has_duplicates::<_, _, u16>(
```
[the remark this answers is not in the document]
ctz: Instead of that, I've used `Option::map(...)::unwrap_or_default()` here

rustls/Cargo.toml
ctz: Should the feature gates after this change use `zlib-rs` or `zlib`? Both work, but `zlib-rs` seems more precise but less clear.

djc: I think that depends on whether we want the feature to be about the precise dependency (`zlib-rs`) vs about the capability (`zlib`). Do you think we'll ever want to have first-party support for multiple implementations of the same algorithm?

rustls/src/compress.rs
```diff
@@ -0,0 +1,77 @@
...
+    &[]
+}
+
+/// A content-less error for when `CertDecompressor::decompress` fails.
+#[derive(Debug)]
+pub struct DecompressionFailed;
+
+/// A content-less error for when `CertCompressor::compress` fails.
+#[derive(Debug)]
+pub struct CompressionFailed;
```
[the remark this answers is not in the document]
ctz: Not sure of the purpose of that? These are intended to be names for the unit type, really. If we (eg) impl `Display` for this, a) that would never be used, b) it implies that someone might one day expect to see that output and find it useful. That is not great for content-less error.

rustls/Cargo.toml
ctz: I think it's more about the capability -- if (say) there was a better option for zlib we'd probably move to that rather than supporting both.

README.md
ctz: Squashed this commit into that one.

rustls/src/compress.rs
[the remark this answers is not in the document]
cpu: > Not sure of the purpose of that?

What about a downstream project that wants to jam all of the errors it might get from Rustls into a `RustlsErr(dyn Error + Send + Sync)` variant? 

I don't feel strongly in any case. Wouldn't want to block this work on a nit :-)

rustls/src/compress.rs
```diff
@@ -226,6 +234,184 @@ mod feat_brotli {
 #[cfg(feature = "brotli")]
 pub use feat_brotli::{BROTLI_COMPRESSOR, BROTLI_DECOMPRESSOR};
 
+/// An LRU cache for compressions.
+///
+/// The prospect of being able to reuse a given compression for many connections
+/// means we can afford to spend more time on that compression (by passing
+/// `CompressionLevel::Amortized` to the compressor).
+#[derive(Debug)]
+pub enum CompressionCache {
```
cpu, line 275: Good idea vs using a trait like I originally suggested :+1: In retrospect that feels like the wrong suggestion since the only implementations are meant to be internal and there's a fixed number.

cpu (approved):

rustls/src/client/client_conn.rs
ctz: I feel this makes the common uses (like adding a value, or removing one) pretty wordy? (eg, converting to a vec, adding a value, then using Vec's `Into<Arc<[T]>>`) -- maybe I missed something though.

cpu:
Extra thanks to @sharksforarms for getting the ball rolling on this work initially. :tada: We appreciate it!

PR #1968, opened 2024-05-24
ChunkVecBuffer -> BufferQueue + improvements

pascaldekloe:
I got confused by the pop, thinking it reversed the ordering like stacks do.
The change names `ChunkVecBuffer` conform its functionality `BufferQueue` as
opposed to naming the inner mechanisms.

 * Vec drain only when needed
 * no reinsert, but peek
 * dequeue does not panic (as comment on pop stated)
 * non-zero check on data-size limit


djc (commented):
@pascaldekloe thanks for sharing your improvements to rustls.

In my opinion, this kind of PR is not great as a first contribution, because it contains substantial amounts of churn (like renaming types and methods) while adding relatively little value. Additionally many of the commits you've made here have a number of logical changes wrapped together, not all of which might be desirable. All in all, getting this through review takes will take substantial maintainer time with limited benefit.

As an alternative route, I would suggest closing this PR and submitting a smaller, more focused single change to get on the same page with the maintainers about potential value (maybe improving the documentation, since that's what had you confused)?

Also, given that this PR suggests you're working with rustls internals, it would be helpful to share what you're working on to make sure (if you intend to submit some of that work upstream) it is working in the right direction.

pascaldekloe:
The plan with this contribution was exactly to see if you'd welcome improvements, rather than a back and forth with justification on every move. 🤝

djc:
> The plan with this contribution was exactly to see if you'd welcome improvements, rather than a back and forth with justification on every move. 🤝

We definitely welcome improvements, and I agree the `ChunkVecBuf` name isn't great. On the other hand, I think it's pretty normal that changes have to be justified (unless they are obvious improvements).

pascaldekloe:
Precisely the "obvious improvements" it is. Better names, better separations, and small tweaks, all separated in small commits is what I believe this project needs and I am willing to give. Happy to hear if anything's wrong. I won't go into micro patches pushing against this works good enough. Your choice. All good either way.

djc:
No, I don't think all of these are obvious improvements, and the commits are not small enough. If your opinion of what is good enough differs from our opinion I'm afraid we won't be able to make much progress here.

PR #1971, opened 2024-05-27
ci: temp. pin nightly to 2024-05-22

cpu:
There is an upstream issue (rust-lang/rust#125474) that causes an ICE building Rustls tests with current nightly ([example](https://github.com/rustls/rustls/actions/runs/9245183158/job/25431321782)):

> thread 'rustc' panicked at compiler/rustc_codegen_ssa/src/back/link.rs:2700:27:
> index out of bounds: the len is 53 but the index is 53

`2024-05-22` looks to be the last nightly that worked without error. Let's pin CI to that for now while upstream sorts out the bug.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,8 @@ jobs:
       - name: Install ${{ matrix.rust }} toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: ${{ matrix.rust }}
+          # TODO(XXX): Revert to "matrix.rust" after rust-lang/rust#125474 is fixed.
+          toolchain: ${{ matrix.rust == 'nightly' && 'nightly-2024-05-22' || matrix.rust }}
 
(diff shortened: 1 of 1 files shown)


cpu:
> Build+test (nightly, ubuntu-latest) Expected — Waiting for status to be reported

Will have to do the annoying branch protection rule update dance for this one :roll_eyes:

djc (approved):

ctz:
> > Build+test (nightly, ubuntu-latest) Expected — Waiting for status to be reported
> 
> Will have to do the annoying branch protection rule update dance for this one 🙄

Alternatively could try something like just:

```
          toolchain: ${{ matrix.rust == "nightly" && "nightly-2024-05-22" || matrix.rust }}
```

cpu:
> Alternatively could try something like just:

Good idea :+1: Reworked.

ctz (approved):

PR #1973, opened 2024-05-28
Fix LTO setting and a clippy::use_self finding

cpu:
## cargo: fix bench profile LTO setting

Per [upstream docs](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) and a warning in my IDE, the `lto` setting's valid options are: false, true, "fat", "thin", or "off". This commit changes our "yes" value to true.

## msgs: fix clippy::use_self finding

I'm not sure yet why CI didn't flag this, but I see it locally:

```
warning: unnecessary structure name repetition
    --> rustls/src/msgs/handshake.rs:1385:52
     |
1385 |     pub(crate) fn new(cert: CertificateDer<'a>) -> CertificateEntry<'a> {
     |                                                    ^^^^^^^^^^^^^^^^^^^^ help: use the applicable keyword: `Self`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self
    --> rustls/src/lib.rs:310:5
     |
310  |     clippy::use_self,
     |     ^^^^^^^^^^^^^^^^
```

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1382,8 +1382,8 @@ impl<'a> Codec<'a> for CertificateEntry<'a> {
 }
 
 impl<'a> CertificateEntry<'a> {
-    pub(crate) fn new(cert: CertificateDer<'a>) -> CertificateEntry<'a> {
-        CertificateEntry {
+    pub(crate) fn new(cert: CertificateDer<'a>) -> Self {
+        Self {
             cert,
             exts: Vec::new(),
         }


cpu:
> It looks like we use --deny warnings on the stable clippy tasks, but not the nightly clippy tasks. Was that intentional?

@djc This feels like something you are likely to have thoughts on :-)

djc:
Yes, that was intentional. The stable ones only change every six weeks but nightly clippy can change everyday and can contain regressions (like false positives) that will get fixed before release. So getting warnings from nightly clippy feels like the right trade-off to me.

cpu:
That's fair, I just think they're hard to notice in practice unless you search them out. That's probably still the right trade-off.

cpu:
> a) we're throwing a lot of "rustc-check-cfg" warnings (also seeing these locally). I think my "fix" from https://github.com/rustls/rustls/pull/1942 was sufficient to fix stable, but isn't appeasing nightly.

This ended up being something simple. Fix PR: https://github.com/rustls/rustls/pull/1975

PR #1978, opened 2024-05-31
Tidy handshake_tests file

ctz:


rustls/src/msgs/handshake_test.rs
```diff
@@ -415,7 +415,7 @@ fn can_print_all_client_extensions() {
 
 #[test]
 fn can_clone_all_client_extensions() {
-    let _ = get_sample_server_hello_payload().extensions;
+    let _ = get_sample_client_hello_payload().extensions;
```
cpu: nice catch

cpu (approved):
Thanks! A nice clean up :broom:

djc (approved):
Nice! Is prefices an intentional latinism? 😄

PR #1982, opened 2024-06-03
Minor `#[allow()]` cleanups

ctz:
This is mostly just seeing what `#[allow()]`s we have that are no longer needed, by removing them all and seeing what warnings appear.


djc (approved):

cpu (approved):
Nice! Wish there was a tool that could flag these stale allows automatically.

PR #1986, opened 2024-06-03
Prepare 0.23.9

ctz:
Release notes headlines:

* **[RFC8879](https://datatracker.ietf.org/doc/html/rfc8879) certificate compression is now supported.** [Get started](https://rustls.github.io/rustls/prerelease/compress/index.html#getting-started) by enabling the `brotli` and/or `zlib` crate features, which depend on the `brotli` or `zlib-rs` crates. We recommend brotli as it has the widest deployment so far.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2217,7 +2217,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.8"
+version = "0.23.9"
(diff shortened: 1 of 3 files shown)


cpu (approved):

ctz:
> I don't feel strongly but it feels like useful context to have.

Added.

PR #1988, opened 2024-06-05
Limit fruitless processing of small messages

### LINKED ISSUE #717: DoS prevention for zero-length or very small fragments

### LINKED ISSUE #952: DoS potential due to renegotiation requests
* In TLS 1.2, we accept unlimited renegotiation requests from the peer. For each one, we send/queue an alert and keep going. This could cause us to allocate memory without bound. We should implement anti-DoS countermeasures that limit the number of renegotiation requests we process.
* ~~We do not verify that the `HelloRequest` or `ClientHello` we receive during a renegotiation request is well-formed. We should implement some form of well-formedness check. This should be trivial for HelloRequest. For ClientHello, we probably don't want to use our current ClientHello parser though. Perhaps we should just sanity check that it isn't huge?~~ [edit: we do parse it into a `HandshakePayload`, and that does validate it sufficiently.]

ctz:
This PR tightens up a number of cases where we can receive large numbers of small messages that do not make any progress and are not surfaced to the application.

These are:

- excessive warning-level alerts
- post-handshake renegotiation attempts
- key_update requests
- empty `TLSPlaintext` fragments

fixes #717
fixes #952

### DIFF
--- a/rustls/src/msgs/deframer.rs
+++ b/rustls/src/msgs/deframer.rs
@@ -26,6 +26,10 @@ pub struct MessageDeframer {
 
     /// If we're in the middle of joining a handshake payload, this is the metadata.
     joining_hs: Option<HandshakePayloadMeta>,
+
+    /// We limit consecutive empty fragments to avoid a route for the peer to send
+    /// us significant but fruitless traffic.
+    seen_consecutive_empty_fragments: u8,
 }
 
 impl MessageDeframer {
@@ -166,6 +170,20 @@ impl MessageDeframer {
                 // records, there MUST NOT be any other records between them."
                 // https://www.rfc-editor.org/rfc/rfc8446#section-5.1
                 return Err(self.set_err(PeerMisbehaved::MessageInterleavedWithHandshakeMessage));
+            }
+
+            match plain_payload_slice.len {
+                0 => {
+                    if self.seen_consecutive_empty_fragments
+                        == ALLOWED_CONSECUTIVE_EMPTY_FRAGMENTS_MAX
+                    {
+                        return Err(self.set_err(PeerMisbehaved::TooManyEmptyFragments));
+                    }
+                    self.seen_consecutive_empty_fragments += 1;
+                }
+                _ => {
+                    self.seen_consecutive_empty_fragments = 0;
+                }
             }
 
             // If it's not a handshake message, just return it -- no joining necessary.
@@ -704,6 +722,10 @@ const MAX_HANDSHAKE_SIZE: u32 = 0xffff;
 
 #[cfg(feature = "std")]
 const READ_SIZE: usize = 4096;
+
+/// cf. BoringSSL's `kMaxEmptyRecords`
+/// <https://github.com/google/boringssl/blob/dec5989b793c56ad4dd32173bd2d8595ca78b398/ssl/tls_record.cc#L124-L128>
+const ALLOWED_CONSECUTIVE_EMPTY_FRAGMENTS_MAX: u8 = 32;
 
 #[cfg(feature = "std")]
 #[cfg(test)]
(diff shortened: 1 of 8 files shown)


rustls/src/msgs/deframer.rs
```diff
@@ -705,6 +723,8 @@ const MAX_HANDSHAKE_SIZE: u32 = 0xffff;
 #[cfg(feature = "std")]
 const READ_SIZE: usize = 4096;
 
+const ALLOWED_CONSECUTIVE_EMPTY_FRAGMENTS_MAX: u8 = 32;
```
djc, line 726: Feels like this could be lower? Did you get this number from the Go impl?

djc (approved):

rustls/tests/api.rs
```diff
@@ -6882,6 +6882,187 @@ impl<'a> io::Write for FakeStream<'a> {
...
+            0x00,
+            0x00,
+        ]),
+    };
+    raw_server.encrypt_and_send(&msg, &mut client);
+    let err = client
+        .process_new_packets()
+        .unwrap_err();
+    assert_eq!(format!("{err:?}"),
+               "InappropriateHandshakeMessage { expect_types: [NewSessionTicket, KeyUpdate], got_type: HelloRequest }");
```
cpu, line 6912: Will this be brittle over time? Would it be better to look for the presence of a key word or two instead of doing an exact match over the `Debug` repr?

Same comment applies to the other instances in this commit but I won't flag them each individually.

rustls/tests/api.rs
```diff
@@ -6882,6 +6882,187 @@ impl<'a> io::Write for FakeStream<'a> {
...
+    let (mut client, mut server) = make_pair_for_configs(client_config, server_config);
+    do_handshake(&mut client, &mut server);
+
+    let mut raw_client = RawTls::new_client(client);
+
+    let msg = PlainMessage {
+        typ: ContentType::Handshake,
+        version: ProtocolVersion::TLSv1_3,
+        payload: Payload::new(vec![
+            u8::from(HandshakeType::ClientHello),
```
cpu, line 6973: Is there a reason to prefer bashing this together with hex constants instead of building up a `HandshakeMessagePayload` with a `HandshakePayload::ClientHello(ClientHelloPayload)`? It seems like it would be easier to reason about if workable.

cpu (commented):
Are the various limits being introduced sourced from another implementation, inspired by standards text, or best guesses at what seems reasonable?

ctz:
> Are the various limits being introduced sourced from another implementation, inspired by standards text, or best guesses at what seems reasonable?

They are effectively sourced from BoringSSL, via the values tested by the bogo test suite. With that said, I am happy if we want to diverge from those and have good rationale to do that.

cpu:
> They are effectively sourced from BoringSSL, via the values tested by the bogo test suite

SGTM. I don't have any reason to suggest we should deviate. Mostly just wanted to understand where the values came from (maybe worth mentioning in the PR desc).

rustls/tests/api.rs
ctz: replaced this and other instances  -- except the Message containing an alert, as `AlertMessagePayload` is not public and I'd like to avoid adding to `internal` exports.

ctz:
> Mostly just wanted to understand where the values came from (maybe worth mentioning in the PR desc).

I have added references/rationale to the values.

djc (approved):

PR #1995, opened 2024-06-07
fix lib.rs examples link

bmw:
reading https://docs.rs/rustls/latest/rustls/, i noticed the examples link there links to https://docs.rs/rustls/latest/rustls/examples/README.md which doesn't exist. this link doesn't work when building the documentation locally either

this proposes one simple way to fix that


cpu (approved):
Thanks!

PR #2001, opened 2024-06-12
ech: implement inner hello extension compression

cpu:
This commit extends the initial client-side ECH support from https://github.com/rustls/rustls/pull/1718 to enable the "SHOULD" recommendation of compressing extensions that are identical between the outer client hello and the inner client hello. This optimization reduces the overhead of ECH, shrinking the total client hello size.

Achieving this is a bit tricky (which is why it was done separately at the end!). To-be-compressed extensions must be in a contiguous block in the inner hello. When we encode that contiguous block for the encoded inner hello, we replace the block with a marker extension listing all of the extension types that were replaced. In the outer hello the compressed extensions must be present in the same relative order (but don't need to be contiguous). We also need the uncompressed inner hello with the uncompressed extensions in place to update the inner transcript.

With this feature implemented we can also activate a couple of bogo tests that were previously excluded.

### DIFF
--- a/rustls/src/msgs/enums.rs
+++ b/rustls/src/msgs/enums.rs
@@ -77,7 +77,7 @@ enum_builder! {
     /// from the various RFCs covering TLS, and are listed by IANA.
     /// The `Unknown` item is used when processing unrecognised ordinals.
     @U16
-    pub(crate) enum ExtensionType {
+    pub enum ExtensionType {
         ServerName => 0x0000,
         MaxFragmentLength => 0x0001,
         ClientCertificateUrl => 0x0002,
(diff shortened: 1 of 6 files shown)


rustls/src/msgs/enums.rs
```diff
@@ -77,7 +77,7 @@ enum_builder! {
     /// from the various RFCs covering TLS, and are listed by IANA.
     /// The `Unknown` item is used when processing unrecognised ordinals.
     @U16
-    pub(crate) enum ExtensionType {
+    pub enum ExtensionType {
```
djc, line 80: Why is this necessary/useful?

rustls/src/msgs/handshake.rs
```diff
@@ -868,7 +868,73 @@ impl TlsListElement for ClientExtension {
...
+            _ => self.session_id.encode(bytes),
+        }
+
+        self.cipher_suites.encode(bytes);
+        self.compression_methods.encode(bytes);
+
+        match purpose {
+            // Compressed extensions must be replaced in the encoded inner client hello.
+            Encoding::EchInnerHello { to_compress } if !to_compress.is_empty() => {
+                // Safety: not empty check in match guard.
```
djc, line 898: Maybe take this outside of the match arm?

djc (approved):
Nice!

ctz (approved):
A couple of things unrelated to this PR I noticed while reviewing:

The head comment in `examples/src/bin/ech-client.rs` has an example usage that wants to find ech-client in the rustls-provider-example crate; it has moved since.

It would be nice if `cargo run --bin ech-client -- --help` included an example usage.

Do we want to run ech-client in the connect-tests job?

rustls/src/msgs/enums.rs
cpu: Good question, I should have called this out somewhere.

With this changeset the existing `ExtensionType` enum ends up being embedded in the `ClientExtension::EncryptedClientHelloOuterExtensions` extension, and since `ClientExtension` type is `pub` we get a clippy warn without the visibility change:

```
warning: type `msgs::enums::ExtensionType` is more private than the item `msgs::handshake::ClientExtension::EncryptedClientHelloOuterExtensions::0`
   --> rustls/src/msgs/handshake.rs:563:41
    |
563 |     EncryptedClientHelloOuterExtensions(Vec<ExtensionType>),
    |                                         ^^^^^^^^^^^^^^^^^^ field `msgs::handshake::ClientExtension::EncryptedClientHelloOuterExtensions::0` is reachable at visibility `pub`
```

rustls/src/msgs/handshake.rs
cpu: Yeah :-/ I had tried a few arrangements but was trying hard to avoid any excessive copying after the clones we definitely require to pull content out of the outer hello for the inner hello.

I'll fiddle to see if there's a nicer way to arrange the match with the first element removal outside the arm.

cpu: > I'll fiddle to see if there's a nicer way to arrange the match with the first element removal outside the arm.

After some tries I can't come up with an arrangement that won't require more cloning/allocating. I'm going to leave this one as-is unless someone feels strongly.

cpu: Ohhh! Ok that makes sense. Thank you, I'll fold that in.

cpu: Done :ballot_box_with_check:

djc (approved):

PR #2003, opened 2024-06-12
`key_update` API and automatic key refreshing

### LINKED ISSUE #755: Generalize support for per-key AEAD limits from QUIC to TLS
In PR #750 we're planning to expose information about the AEAD algorithm's limits as far as the amount of data that is safe to encrypt and/or authenticate using a given AEAD key. The idea is that the application should do a key update or close the connection before hitting either limit.

It seems like the same thinking applies to TLS, both TLS 1.2 and TLS 1.3. We should consider having the TLS API, not just the QUIC API, exposes a good interface for dealing with this. In the case of TLS 1.2 then the only option would be to close the connection since we don't support renegotiation. For TLS 1.3 key update could be used.

BTW, I seem to remember that the limits mentioned in RFC 9001 were derived in a quite approximate way. We might want to redo the math before we decide what limits to expose.

Also, for a TLS connection, we should consider keeping track of the number of bytes encrypted and/or authenticated and refuse to exceed those limits. This is beyond what the QUIC stuff in PR #750 does.

### LINKED ISSUE #946: Provide an API for controlling traffic key updates
An application might know that it won't be sending application data for a while, and/or it won't be expecting data for a while, and it might want to roll the traffic key(s) forward. We should have a way for the application to make this happen.

Note that issue #755 tracks doing automatic key updates when "enough" data has been sent.

There are a lot of details to decide on w.r.t. the API design and functionality to implement. I don't think this is a high priority yet. I'm mostly filing this so that I can reference this issue from other issues.

ctz:
This PR:

- adds a new public API for sending TLS1.3 `key_update` requests. This is available in both the unbuffered and classic APIs.
- automatically sends TLS1.3 `key_update` requests when our encryption key approaches "exhaustion" due to the birthday bound[^1]

[^1]: Note: as before, we track this with the sequence number rather than counting blocks, so sending small messages underestimates the limit significantly. This errs on the side of safety: (for AES-GCM suites) we update keys each 16 million messages whether that represents 16MB or 256GB of data transfer.

fixes #946 
fixes #755

### DIFF
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -1554,6 +1554,14 @@ impl State<ClientConnectionData> for ExpectTraffic {
         Ok(self)
     }
 
+    fn send_key_update_request(&mut self, common: &mut CommonState) -> Result<(), Error> {
+        common.check_aligned_handshake()?;
+        common.send_msg(Message::build_key_update_request(), true);
+        self.key_schedule
+            .update_encrypter(common);
+        Ok(())
+    }
+
     fn export_keying_material(
         &self,
         output: &mut [u8],
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -364,7 +364,14 @@ impl CommonState {
 
     // Put m into sendable_tls for writing.
     fn queue_tls_message(&mut self, m: OutboundOpaqueMessage) {
+        self.perhaps_write_key_update();
         self.sendable_tls.append(m.encode());
+    }
+
+    pub(crate) fn perhaps_write_key_update(&mut self) {
+        if let Some(message) = self.queued_key_update_message.take() {
+            self.sendable_tls.append(message);
+        }
     }
 
     /// Send a raw TLS message, fragmenting it if needed.
@@ -703,12 +710,6 @@ impl CommonState {
         }
 
         self.send_plain_non_buffering(payload, limit)
-    }
-
-    pub(crate) fn perhaps_write_key_update(&mut self) {
-        if let Some(message) = self.queued_key_update_message.take() {
-            self.sendable_tls.append(message);
-        }
     }
 }
 
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -40,7 +40,6 @@ pub struct CommonState {
     pub(crate) has_received_close_notify: bool,
     #[cfg(feature = "std")]
     pub(crate) has_seen_eof: bool,
-    pub(crate) received_middlebox_ccs: u8,
     pub(crate) peer_certificates: Option<CertificateChain<'static>>,
     message_fragmenter: MessageFragmenter,
     pub(crate) received_plaintext: ChunkVecBuffer,
@@ -71,7 +70,6 @@ impl CommonState {
             has_received_close_notify: false,
             #[cfg(feature = "std")]
             has_seen_eof: false,
-            received_middlebox_ccs: 0,
             peer_certificates: None,
             message_fragmenter: MessageFragmenter::default(),
             received_plaintext: ChunkVecBuffer::new(Some(DEFAULT_RECEIVED_PLAINTEXT_LIMIT)),
@@ -648,6 +646,11 @@ impl CommonState {
                 .encode(),
         );
     }
+
+    pub(crate) fn received_tls13_change_cipher_spec(&mut self) -> Result<(), Error> {
+        self.temper_counters
+            .received_tls13_change_cipher_spec()
+    }
 }
 
 #[cfg(feature = "std")]
@@ -843,6 +846,7 @@ struct TemperCounters {
     allowed_warning_alerts: u8,
     allowed_renegotiation_requests: u8,
     allowed_key_update_requests: u8,
+    allowed_middlebox_ccs: u8,
 }
 
 impl TemperCounters {
@@ -875,6 +879,16 @@ impl TemperCounters {
             }
         }
     }
+
+    fn received_tls13_change_cipher_spec(&mut self) -> Result<(), Error> {
+        match self.allowed_middlebox_ccs {
+            0 => Err(PeerMisbehaved::IllegalMiddleboxChangeCipherSpec.into()),
+            _ => {
+                self.allowed_middlebox_ccs -= 1;
+                Ok(())
+            }
+        }
+    }
 }
 
 impl Default for TemperCounters {
@@ -891,6 +905,12 @@ impl Default for TemperCounters {
             // cf. BoringSSL `kMaxKeyUpdates`
             // <https://github.com/google/boringssl/blob/dec5989b793c56ad4dd32173bd2d8595ca78b398/ssl/tls13_both.cc#L35-L38>
             allowed_key_update_requests: 32,
+
+            // At most two CCS are allowed: one after each ClientHello (recall a second
+            // ClientHello happens after a HelloRetryRequest).
+            //
+            // note BoringSSL allows up to 32.
+            allowed_middlebox_ccs: 2,
         }
     }
 }
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -428,7 +428,13 @@ impl CommonState {
     pub(crate) fn start_encryption_tls12(&mut self, secrets: &ConnectionSecrets, side: Side) {
         let (dec, enc) = secrets.make_cipher_pair(side);
         self.record_layer
-            .prepare_message_encrypter(enc);
+            .prepare_message_encrypter(
+                enc,
+                secrets
+                    .suite()
+                    .common
+                    .confidentiality_limit,
+            );
         self.record_layer
             .prepare_message_decrypter(dec);
     }
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -213,13 +213,20 @@ impl CommonState {
                 payload.clone(),
             );
 
-        let remaining_encryptions = self
-            .record_layer
-            .remaining_write_seq()
-            .ok_or(EncryptError::EncryptExhausted)?;
-
-        if fragments.len() as u64 > remaining_encryptions.get() {
-            return Err(EncryptError::EncryptExhausted);
+        for f in 0..fragments.len() {
+            match self
+                .record_layer
+                .pre_encrypt_action(f as u64)
+            {
+                record_layer::PreEncryptAction::Nothing => {}
+                record_layer::PreEncryptAction::Close => {
+                    self.eager_send_close_notify(outgoing_tls)?;
+                    return Err(EncryptError::EncryptExhausted);
+                }
+                record_layer::PreEncryptAction::Refuse => {
+                    return Err(EncryptError::EncryptExhausted);
+                }
+            }
         }
 
         self.check_required_size(
@@ -298,20 +305,25 @@ impl CommonState {
     }
 
     fn send_single_fragment(&mut self, m: OutboundPlainMessage) {
-        // Close connection once we start to run out of
-        // sequence space.
-        if self
+        match self
             .record_layer
-            .wants_close_before_encrypt()
+            .next_pre_encrypt_action()
         {
-            self.send_close_notify();
-        }
+            record_layer::PreEncryptAction::Nothing => {}
 
-        // Refuse to wrap counter at all costs.  This
-        // is basically untestable unfortunately.
-        if self.record_layer.encrypt_exhausted() {
-            return;
-        }
+            // Close connection once we start to run out of
+            // sequence space.
+            record_layer::PreEncryptAction::Close => {
+                self.send_close_notify();
+                return;
+            }
+
+            // Refuse to wrap counter at all costs.  This
+            // is basically untestable unfortunately.
+            record_layer::PreEncryptAction::Refuse => {
+                return;
+            }
+        };
 
         let em = self.record_layer.encrypt_outgoing(m);
         self.queue_tls_message(em);
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -798,6 +798,10 @@ pub(crate) trait State<Data>: Send + Sync {
         Err(Error::HandshakeNotComplete)
     }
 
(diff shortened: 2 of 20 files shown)


rustls/src/record_layer.rs
```diff
@@ -181,16 +180,19 @@ impl RecordLayer {
...
-
-    /// Return true if we outright refuse to do anything with the
-    /// encryption key.
-    pub(crate) fn encrypt_exhausted(&self) -> bool {
-        self.write_seq >= SEQ_HARD_LIMIT
+    /// Return a remedial action when we are near to encrypting too many messages.
+    ///
+    /// `add` is added to the current sequence number.  `add` as `0` means
+    /// "the next message processed by `encrypt_outgoing`"
+    pub(crate) fn pre_encrypt_action(&self, add: u64) -> PreEncryptAction {
```
cpu, line 199: Would this be easier to understand at call-sites if there was a separate `pub(crate) fn next_pre_encrypt_action()` that calls `pre_encrypt_action(0)` (or similar)?

rustls/src/record_layer.rs
```diff
@@ -246,6 +241,18 @@ pub(crate) struct Decrypted<'a> {
...
 
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) enum PreEncryptAction {
+    /// No action is needed before calling `encrypt_outgoing`
+    Nothing,
+
+    /// A `close_notify` alert should be sent ASAP
+    Close,
+
+    /// Do not call `encrypt_outgoing` further, it will panic.
```
cpu, line 267: nit: maybe worth mentioning why it would panic? e.g. something about too many messages having been encrypted.

rustls/src/conn.rs
```diff
@@ -981,6 +982,25 @@ impl<Data> ConnectionCore<Data> {
             Err(e) => Err(e.clone()),
         }
     }
+
+    fn refresh_traffic_keys(&mut self) -> Result<(), Error> {
```
cpu, line 986: ordering nit: I think since `maybe_refresh_traffic_keys` calls `refresh_traffic_keys` their order is inverted.

rustls/tests/api.rs
```diff
@@ -7155,6 +7155,77 @@ fn test_refresh_traffic_keys() {
     check_both_directions(&mut client, &mut server);
 }
 
+#[test]
+fn test_automatic_refresh_traffic_keys() {
+    static LIMITED_SUITE: OnceLock<rustls::Tls13CipherSuite> = OnceLock::new();
+    let limited = LIMITED_SUITE.get_or_init(|| {
+        with_confidentiality_limit(provider::cipher_suite::TLS13_AES_128_GCM_SHA256, 1024)
```
cpu, line 7166: Is there a way to write this that would be a little bit less brittle w.r.t some of the constants?

An easy change would be to lift the 1024 limit from the `limited` ciphersuite into a binding that could be ref'd for the `for` range on L7186 and the assert on L7204.

It might be nice if the expected `transfer` values (30 and 57) could be calculated based on the actual encoding of some exemplar messages but that feels pretty heavy-weight. I would be OK with how it is now.

rustls/tests/common/mod.rs
```diff
@@ -1040,3 +1040,18 @@ impl RawTls {
         f(msg);
     }
 }
+
+pub fn with_confidentiality_limit(
+    original: SupportedCipherSuite,
+    limit: u64,
+) -> rustls::Tls13CipherSuite {
```
cpu, line 1047: I don't feel strongly, but since there's only one call-site for this fn it feels like it might be less cognitive overhead to inline this in `test_automatic_refresh_traffic_keys()`, or otherwise put it alongside that test instead of in `common/mod.rs` (at least until there's more callers?)

rustls/tests/api.rs
```diff
@@ -34,7 +34,7 @@ use rustls::{
     crypto::aws_lc_rs::hpke::ALL_SUPPORTED_SUITES,
     internal::msgs::base::PayloadU16,
     internal::msgs::handshake::{
-        EchConfigPayload, EchConfigContents, HpkeKeyConfig, HpkeSymmetricCipherSuite,
+        EchConfigContents, EchConfigPayload, HpkeKeyConfig, HpkeSymmetricCipherSuite,
```
cpu, line 37: Oops! thought I fixed the last of this stuff in https://github.com/rustls/rustls/pull/2002 but I forgot about the dance required to reformat `api.rs`. Thanks for fixing :+1:

cpu (commented):
Some comments from an early review pass. Looks solid!

rustls/tests/common/mod.rs
cpu: Makes sense :+1:

rustls/src/common_state.rs
```diff
@@ -366,9 +366,16 @@ impl CommonState {
 
     // Put m into sendable_tls for writing.
     fn queue_tls_message(&mut self, m: OutboundOpaqueMessage) {
+        self.perhaps_write_key_update();
```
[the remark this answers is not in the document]
ctz: Yes, this was found during writing `test_refresh_traffic_keys`, and it still reproduces it (spurious `DecryptError`s)

rustls/src/record_layer.rs
```diff
@@ -1,14 +1,13 @@
 use alloc::boxed::Box;
```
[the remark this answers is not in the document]
ctz: Added rationale.

rustls/src/record_layer.rs
```diff
@@ -1,14 +1,13 @@
...
 use crate::crypto::cipher::{InboundOpaqueMessage, MessageDecrypter, MessageEncrypter};
 use crate::error::Error;
 #[cfg(feature = "logging")]
 use crate::log::trace;
 use crate::msgs::message::{InboundPlainMessage, OutboundOpaqueMessage, OutboundPlainMessage};
 
-static SEQ_SOFT_LIMIT: u64 = 0xffff_ffff_ffff_0000u64;
-static SEQ_HARD_LIMIT: u64 = 0xffff_ffff_ffff_fffeu64;
+const SEQ_SOFT_LIMIT: u64 = 0xffff_ffff_ffff_0000u64;
+const SEQ_HARD_LIMIT: u64 = 0xffff_ffff_ffff_fffeu64;
```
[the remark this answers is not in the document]
ctz: Moved to bottom in this commit.

rustls/tests/api.rs
ctz: So I have hoisted a lot of this into `with_confidentiality_limit` (and renamed it). The remainder has `const CONFIDENTIALITY_LIMIT: usize = 1024;` plus a const fn that gives how big a TLS message is with a given payload.

rustls/src/client/tls13.rs
```diff
@@ -1569,6 +1569,14 @@ impl State<ClientConnectionData> for ExpectTraffic {
             .extract_secrets(Side::Client)
     }
 
+    fn send_key_update_request(&mut self, common: &mut CommonState) -> Result<(), Error> {
```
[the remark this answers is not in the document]
ctz: moved up

ctz:
This is ready, I think. I have hoisted the more incidental changes to the top of the PR.

rustls/src/common_state.rs
```diff
@@ -219,10 +221,15 @@ impl CommonState {
                 .pre_encrypt_action(f as u64)
             {
                 record_layer::PreEncryptAction::Nothing => {}
-                record_layer::PreEncryptAction::Close => {
-                    self.eager_send_close_notify(outgoing_tls)?;
-                    return Err(EncryptError::EncryptExhausted);
-                }
+                record_layer::PreEncryptAction::RefreshOrClose => match self.negotiated_version {
+                    Some(ProtocolVersion::TLSv1_3) => {
+                        self.refresh_traffic_keys_pending = Some(());
```
cpu, line 226: The mutation on L327 is decorated with a comment motivating why this is setting a flag and expecting the caller to take the action. Worth duplicating that here?

rustls/src/common_state.rs
```diff
@@ -533,25 +533,9 @@ impl CommonState {
...
-        debug!("Sending warning alert {:?}", AlertDescription::CloseNotify);
-
-        let iter = self
-            .message_fragmenter
-            .fragment_message(&m);
-
-        let written = self.write_fragments(outgoing_tls, iter);
-
-        Ok(written)
+        self.send_close_notify();
```
cpu, line 548: What was your impression of the `debug_assert!()` that is removed with the change to re-using `send_close_notify()`?

cpu (approved):

rustls/tests/api.rs
cpu: Looks much nicer, thank you!

rustls/src/common_state.rs
ctz: I think this whole codepath was only reachable if `CommonState::may_send_application_data` was true, which implies `self.record_layer.is_encrypting()` is true. In any case, the code underneath `send_close_notify` works for both cases (encrypted alerts, and not)

rustls/src/common_state.rs
ctz: Added.

rustls/src/record_layer.rs
```diff
@@ -181,16 +177,23 @@ impl RecordLayer {
...
+    /// "the next message processed by `encrypt_outgoing`"
+    pub(crate) fn pre_encrypt_action(&self, add: u64) -> PreEncryptAction {
+        let value = self.write_seq.saturating_add(add);
+        if value == SEQ_SOFT_LIMIT {
+            PreEncryptAction::Close
+        } else if value >= SEQ_HARD_LIMIT {
+            PreEncryptAction::Refuse
+        } else {
+            PreEncryptAction::Nothing
+        }
```
djc: Nit: I might've written this as a `match value {}`.

rustls/src/common_state.rs
```diff
@@ -213,13 +213,20 @@ impl CommonState {
...
-            .ok_or(EncryptError::EncryptExhausted)?;
-
-        if fragments.len() as u64 > remaining_encryptions.get() {
-            return Err(EncryptError::EncryptExhausted);
+        for f in 0..fragments.len() {
+            match self
+                .record_layer
+                .pre_encrypt_action(f as u64)
+            {
+                record_layer::PreEncryptAction::Nothing => {}
```
djc, line 223: Nit: maybe import `record_layer::PreEncryptAction`.

rustls/src/common_state.rs
```diff
@@ -51,6 +51,7 @@ pub struct CommonState {
     pub(crate) quic: quic::Quic,
     pub(crate) enable_secret_extraction: bool,
     temper_counters: TemperCounters,
+    pub(crate) refresh_traffic_keys_pending: Option<()>,
```
djc, line 54: Why `Option<()>` instead of just `bool`?

rustls/src/record_layer.rs
```diff
@@ -259,8 +259,11 @@ pub(crate) enum PreEncryptAction {
     /// No action is needed before calling `encrypt_outgoing`
     Nothing,
 
-    /// A `close_notify` alert should be sent ASAP
-    Close,
+    /// A `key_update` request should be sent ASAP.
+    ///
+    /// If that is not possible (for example, the connection is TLS1.2), a `close_notify`
+    /// alert should be sent instead.
+    RefreshOrClose,
```
djc, line 266: I'd have pulled this back into the originating commit.

rustls/src/common_state.rs
```diff
@@ -227,6 +227,7 @@ impl CommonState {
                         self.refresh_traffic_keys_pending = Some(());
                     }
                     _ => {
+                        error!("Traffic keys exhausted, closing connection to prevent security failure");
```
djc, line 230: I think we usually use all-lowercase log messages.

rustls/src/common_state.rs
```diff
@@ -307,21 +308,32 @@ impl CommonState {
...
+            &m,
+            self.record_layer
+                .next_pre_encrypt_action(),
+        ) {
+            // Alerts are always sendable -- never quashed by a PreEncryptAction.
+            (
+                OutboundPlainMessage {
+                    typ: ContentType::Alert,
+                    ..
+                },
```
djc, line 321: Since there is not actually any interaction between the matches for `m` and the next pre-encrypt action, I think this would be easier to follow if you handled the alert before even calling `next_pre_encrypt_action()`?

rustls/src/common_state.rs
```diff
@@ -307,21 +308,32 @@ impl CommonState {
...
             // sequence space.
-            record_layer::PreEncryptAction::RefreshOrClose => {
+            (_, record_layer::PreEncryptAction::RefreshOrClose) => {
                 match self.negotiated_version {
                     Some(ProtocolVersion::TLSv1_3) => {
                         // driven by caller, as we don't have the `State` here
                         self.refresh_traffic_keys_pending = Some(());
                     }
                     _ => {
+                        error!("Traffic keys exhausted, closing connection to prevent security failure");
```
djc, line 336: Lowercase this?

djc (commented):
Sorry for the slow review -- had just started before you added this to the merge queue.

rustls/src/common_state.rs
ctz: I was fixated on wanting to write `.take()` but I guess I can use mem::take with a bool.

rustls/src/client/tls13.rs
```diff
@@ -1554,6 +1554,14 @@ impl State<ClientConnectionData> for ExpectTraffic {
         Ok(self)
     }
 
+    fn send_key_update_request(&mut self, common: &mut CommonState) -> Result<(), Error> {
```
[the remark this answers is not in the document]
ctz: Have done this, though it's quite minor -- I would very much like to avoid any dependency from `common_state` to `tls13::key_schedule`.

ctz:
> Sorry for the slow review -- had just started before you added this to the merge queue.

Sorry about that -- see #2016 for follow-up

PR #2004, opened 2024-06-13
v0.23.10 release preparation

cpu:
### Proposed release notes

* **[draft-ietf-tls-esni-18](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18) encrypted client hello (ECH) is now supported** for client applications. See the `ech-client.rs` example for a complete end-to-end demonstration using DNS-over-HTTPS to retrieve a server's ECH configuration for building a Rustls `ClientConfig` using `with_ech()`.
* Additional safety limits guarding against fruitless processing of small messages have been added. Rustls will now limit excessive warning-level alerts, post-handshake renegotiation attempts, key update requests, and empty plaintext fragments.
* FIPS mode has been updated to exclude X25519 key exchange.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -135,10 +135,11 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
 /// These must be created via the [`ClientConfig::builder()`] or [`ClientConfig::builder_with_provider()`]
 /// function.
 ///
-/// Note that using [`ClientConfig::with_ech]` will produce a common configuration specific to
-/// the provided [`crate::client::EchConfig`] that may not be appropriate for all connections made
-/// by the program. In this case the configuration should only be shared by connections intended
-/// for domains that offer the provided [`crate::client::EchConfig`] in their DNS zone.
+/// Note that using [`ConfigBuilder<ClientConfig, WantsVersions>::with_ech()`] will produce a common
+/// configuration specific to the provided [`crate::client::EchConfig`] that may not be appropriate
+/// for all connections made by the program. In this case the configuration should only be shared
+/// by connections intended for domains that offer the provided [`crate::client::EchConfig`] in
+/// their DNS zone.
 ///
 /// # Defaults
 ///
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2177,7 +2177,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.9"
+version = "0.23.10"
 dependencies = [
  "aws-lc-rs",
  "base64 0.22.1",
@@ -2217,7 +2217,7 @@ dependencies = [
  "fxhash",
  "itertools 0.13.0",
  "rayon",
- "rustls 0.23.9",
+ "rustls 0.23.10",
  "rustls-pemfile 2.1.2",
  "rustls-pki-types",
  "tikv-jemallocator",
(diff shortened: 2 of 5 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -135,10 +135,11 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
 /// These must be created via the [`ClientConfig::builder()`] or [`ClientConfig::builder_with_provider()`]
 /// function.
 ///
-/// Note that using [`ClientConfig::with_ech]` will produce a common configuration specific to
-/// the provided [`crate::client::EchConfig`] that may not be appropriate for all connections made
-/// by the program. In this case the configuration should only be shared by connections intended
-/// for domains that offer the provided [`crate::client::EchConfig`] in their DNS zone.
+/// Note that using [`ConfigBuilder<ClientConfig, WantsVersions>::with_ech`] will produce a common
```
djc, line 138: Tiny nit: `with_ech()`?

djc (approved):

rustls/src/client/client_conn.rs
cpu: Fixed, thanks

djc:
One question I have is: is it explicitly documented what happens with TLS 1.2/1.3 interaction and ECH configuration? AIUI a client that is configured to do ECH will not be able to negotiate 1.2, even though it is configured to do so. Is that documented? Should we reject `ClientConfig`s that configure ECH with TLS 1.2 enabled? Should we have a mode for "opportunistic" ECH that only enables it when possible? (Given that it is a privacy feature, that might be pointless?)

cpu:
@djc Good questions, thanks for double checking this from more of an end user perspective.

> One question I have is: is it explicitly documented what happens with TLS 1.2/1.3 interaction and ECH configuration? 

Yes, the `with_ech()` builder fn is documented to say:

>This implicitly selects TLS 1.3 as the only supported protocol version to meet the requirement to support ECH.

I think that's the right place to document that.

> AIUI a client that is configured to do ECH will not be able to negotiate 1.2, even though it is configured to do so. 

Because `with_ech()` is only exposed on the `ConfigBuilder<ClientConfig, WantsVersions> ` state you _can't_ select both ECH and specific protocol versions. Providing an ECH config locks the config to TLS 1.3 only by design.

> Is that documented in all the right places? Should we reject ClientConfigs that configure ECH with TLS 1.2 enabled? 

I think this is covered by the above. You shouldn't be able to get into a state where a client config has both ECH configured and can do TLS 1.2.

> Should we have a mode for "opportunistic" ECH that only enables it when possible? (Given that it is a privacy feature, that might be pointless?)

The only thing we can do opportunistically without a `EchConfigList` read from DNS is to send GREASE ECH. I think in this case it would be best to let the calling application decide whether it wants that GREASE ECH offer to be sent. We can't do it "for free" (in the sense of requiring no new configuraton) because we need a selection of an HPKE suite to generate the dummy pubkey for the GREASE ECH offer.

ctz (approved):

djc:
> I think this is covered by the above. You shouldn't be able to get into a state where a client config has both ECH configured and can do TLS 1.2.

Ahh, that sounds nice!

cpu:
> Ahh, that sounds nice!

I got the idea [from one of the smarter programmers I know](https://github.com/rustls/rustls/pull/1718#discussion_r1441561147) :wink:

cpu:
* Pushed [v/0.23.10](https://github.com/rustls/rustls/releases/tag/v%2F0.23.10) tag
* Uploaded rustls v0.23.10 to registry `crates-io`
* Published [0.23.10 GitHub release](https://github.com/rustls/rustls/releases/tag/v%2F0.23.10)

PR #2013, opened 2024-06-19
(Examples) Feature: Use port-prefix naming to lookup correct ECHConfig

ckcr4lyf:
# Background

When looking up SVCB records, a port and scheme may be prefixed to the QNAME for non-standard ports. For HTTPS records, this practice is followed, except for the standard port (443). (Ref: https://datatracker.ietf.org/doc/html/rfc9460#section-9.1)

I.e. to connect to a HTTPS server (say, `example.com`) on port `4443` as an example, one should lookup the config of the QNAME `_4443._https.example.com` , rather than just `example.com` .

## Old Behavior

Even when the port is explicitly specified, the resolver picks the standard ECHConfig (note the config_id):

```
$ cargo run --package rustls-examples --bin ech-client -- rfc5746.mywaifu.best -p 4443 rfc5746.mywaifu.best

...[truncated]...

[2024-06-19T05:13:21Z DEBUG rustls::client::ech] selected ECH config ID 132 suite HpkeSuite { kem: DHKEM_X25519_HKDF_SHA256, sym: HpkeSymmetricCipherSuite { kdf_id: HKDF_SHA256, aead_id: AES_128_GCM } }
```

## After Fix

Correctly fetches the ECHConfig for port 4443 with correct config ID / public_name

```
$ cargo run --package rustls-examples --bin ech-client -- rfc5746.mywaifu.best -p 4443 rfc5746.mywaifu.best

...[truncated]...

[2024-06-19T05:05:23Z DEBUG rustls::client::ech] selected ECH config ID 240 suite HpkeSuite { kem: DHKEM_X25519_HKDF_SHA256, sym: HpkeSymmetricCipherSuite { kdf_id: HKDF_SHA256, aead_id: AES_128_GCM } } public_name DnsName("cia.gov")
```

## Miscellaneous

I also updated the debug log to print the `public_name` , I think it kinda helps for debugging purposes. I can remove that commit if its better to not have it.

### DIFF
--- a/examples/src/bin/ech-client.rs
+++ b/examples/src/bin/ech-client.rs
@@ -47,6 +47,8 @@ fn main() {
         .and_then(|d| d.deserialize())
         .unwrap_or_else(|e| e.exit());
 
+    let port = args.flag_port.unwrap_or(443);
+
     // Find raw ECH configs using DNS-over-HTTPS with Hickory DNS.
     let resolver_config = if args.flag_cloudflare_dns {
         ResolverConfig::cloudflare_https()
@@ -58,7 +60,7 @@ fn main() {
         true => None, // Force the use of the GREASE ext by skipping ECH config lookup
         false => match args.flag_ech_config {
             Some(path) => Some(read_ech(&path)),
-            None => lookup_ech_configs(&resolver, &args.arg_outer_hostname),
+            None => lookup_ech_configs(&resolver, &args.arg_outer_hostname, port),
         },
     };
 
@@ -119,10 +121,7 @@ fn main() {
         trace!("\nRequest {} of {}", i + 1, args.flag_num_reqs);
         let mut conn = rustls::ClientConnection::new(config.clone(), server_name.clone()).unwrap();
         // The "outer" server that we're connecting to.
-        let sock_addr = (
-            args.arg_outer_hostname.as_str(),
-            args.flag_port.unwrap_or(443),
-        )
+        let sock_addr = (args.arg_outer_hostname.as_str(), port)
             .to_socket_addrs()
             .unwrap()
             .next()
@@ -205,9 +204,17 @@ struct Args {
 fn lookup_ech_configs(
     resolver: &Resolver,
     domain: &str,
+    port: u16,
 ) -> Option<pki_types::EchConfigListBytes<'static>> {
+    // For non-standard ports, lookup the ECHConfig using port-prefix naming
+    // See: https://datatracker.ietf.org/doc/html/rfc9460#section-9.1
+    let qname_to_lookup = match port {
+        443 => domain.to_owned(),
+        port => format!("_{port}._https.{domain}"),
+    };
+
     resolver
-        .lookup(domain, RecordType::HTTPS)
+        .lookup(qname_to_lookup, RecordType::HTTPS)
         .ok()?
         .record_iter()
         .find_map(|r| match r.data() {
--- a/rustls/src/client/ech.rs
+++ b/rustls/src/client/ech.rs
@@ -144,8 +144,8 @@ impl EchConfig {
                     .find(|hpke| hpke.suite() == suite)
                 {
                     debug!(
-                        "selected ECH config ID {:?} suite {:?}",
-                        key_config.config_id, suite
+                        "selected ECH config ID {:?} suite {:?} public_name {:?}",
+                        key_config.config_id, suite, contents.public_name
                     );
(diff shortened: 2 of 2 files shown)


ckcr4lyf:
Note: I am somewhat familiar with rust basics, but not an expert in idiomatic rust. From what I can tell for stuff like `u16` its better to pass it "directly" rather than as a reference. Any feedback is appreciated.

examples/src/bin/ech-client.rs
```diff
@@ -58,7 +58,11 @@ fn main() {
         true => None, // Force the use of the GREASE ext by skipping ECH config lookup
         false => match args.flag_ech_config {
             Some(path) => Some(read_ech(&path)),
-            None => lookup_ech_configs(&resolver, &args.arg_outer_hostname),
+            None => lookup_ech_configs(
+                &resolver,
+                &args.arg_outer_hostname,
+                args.flag_port.unwrap_or(443),
```
cpu, line 64: Now that there's two usages of `args.flag_port` I think we should pull out a `let port = args.flag_port.unwrap_or(443)` binding on ~L49-ish so that we can use it both here and on L128 when constructing the `SocketAddr` without duplicating the `unwrap_or(...)` default.

examples/src/bin/ech-client.rs
```diff
@@ -205,9 +209,17 @@ struct Args {
 fn lookup_ech_configs(
     resolver: &Resolver,
     domain: &str,
+    port: u16,
 ) -> Option<pki_types::EchConfigListBytes<'static>> {
+    // For non-standard ports, lookup the ECHConfig using port-prefix naming
+    // See: https://datatracker.ietf.org/doc/html/rfc9460#section-9.1
+    let qname_to_lookup: String = match port {
+        443 => domain.to_string(),
+        other => format!("_{other}._https.{domain}").to_string(),
```
cpu, line 218: A few small nits here:

1. We prefer to omit the `String` typing information if it can be inferred.
2. I think `to_string()` on the `format!()` result is superfluous - it's already a `String`.
3. (Minor) I would be tempted to keep using `port` as the binding name for the "other" arm. I think it will read a bit more clearly in the format string

```rust
    let qname_to_lookup = match port {
        443 => domain.to_string(),
        port => format!("_{port}._https.{domain}"),
    };
```

cpu (approved):
Thanks! This looks great. I flagged a few small changes that I think would be nice to land.

I already had a TODO leftover to consider trying to upstream a higher level interface for the ECH config lookup into hickory-dns. This subtlety with respect to constructing the QNAME for ports other than 443 gives more motivation for that IMO :+1:

cpu:
>  From what I can tell for stuff like u16 its better to pass it "directly" rather than as a reference. Any feedback is appreciated.

Correct, because the primitive numeric types are marked [Copy](https://doc.rust-lang.org/std/marker/trait.Copy.html). The trait docs have more info if you're curious. :+1:

cpu:
> codecov/patch — 0.00% of diff hit (target 94.18%)

Just wanted to explicitly mention it's fine to ignore this check for an update to an example binary.

examples/src/bin/ech-client.rs
djc: Tiny nit: in cases like these where we're convertting from `&str` to `String` it can be helpful to use `to_owned()` instead of `to_string()`, which needs a little bit less work (because `ToString` depends on formatting machinery) and makes it clear that what's going on is more limited than conversion from anything to string.

ckcr4lyf: Thanks for the suggestions guys. It didn't even occur to me it would be "legal" to use the same name inside the arm, neat that it works.

cpu (approved):
Thanks! I think the last thing we need is to tidy up the commit history. Ideally this branch would be just two commits: the lookup change, and the additional logging. (Doing it all as one commit seems OK to me too if you prefer).

ckcr4lyf:
> Thanks! I think the last thing we need is to tidy up the commit history. Ideally this branch would be just two commits

Got it, what's the best way to do this? Close this PR and make a new one? Or rebase on my end and then push again?

cpu:
> Or rebase on my end and then push again?

^ This is what I'd usually do. Thanks!

ckcr4lyf:
Ok, I think I rebased and force pushed it to just two commits (not sure if I did it correctly)

cpu:
> (not sure if I did it correctly)

You got it :-)

PR #2014, opened 2024-06-19
Refactor integration tests to let rustfmt work

ctz:
This PR undoes the macrology that prevented rustfmt from formatting the integration tests.

That comes at some trade-offs:

- Having the test code continue to live in the least surprising place means opting-out of `autotests`
- There are some stubs in `tests/runners/*.rs` which also need to contain items that are global to the test program and must not be instantiated per-provider.
- `allow(clippy::duplicate_mod)` needs to be splattered in each test file.

However, at least `cargo fmt` works again.


rustls/tests/runners/macros.rs
```diff
@@ -1,37 +1,45 @@
-/// Instantiate the given test functions once for each built-in provider.
-///
-/// The selected provider module is bound as `provider`; you can rely on this
-/// having the union of the public items common to the `rustls::crypto::ring`
-/// and `rustls::crypto::aws_lc_rs` modules.
+//! Instantiate the given test functions once for each built-in provider.
```
cpu, line 1: I think this headline might need tweaking. WDYT about something like:  "Import and scaffold a built-in crypto provider" (?)

cpu (approved):
This looks good to me. Having to drop `autotests` is a bit unfortunate but I think it's the right call. We seldom add new tests but often add new test content. Having to do a special format dance and losing other IDE features is a greater burden day-to-day. Thanks for fixing!

One meta-request: WDYT about adding a `rustls/tests/README.md` or a new section in `CONTRIBUTING.md` that mentions this arrangement + its motivation at a high-level? I think it would be helpful to a new contributor to know why tests are arranged this way and how to add a new one so it will be run.

rustls/tests/runners/macros.rs
ctz: Reworded.

ctz:
> a new section in `CONTRIBUTING.md` that mentions this arrangement + its motivation at a high-level?

I've added a commit for this. [Rendered](https://github.com/rustls/rustls/blob/6a7917cce0aa52f87a76b91d0c404c53de387e1f/CONTRIBUTING.md)

cpu (approved):

PR #2015, opened 2024-06-20
docs: word iff => if

Borber:


ctz:
Thanks for the change! Iff is jargon, short for "if and only if" (see https://en.wikipedia.org/wiki/Iff) but I think it would be an improvement to expand all instances of iff to "if any only if" to improve clarity. I count 5 in total. Would you like to expand the scope of this PR a bit?

Borber:
> Thanks for the change! Iff is jargon, short for "if and only if" (see https://en.wikipedia.org/wiki/Iff) but I think it would be an improvement to expand all instances of iff to "if any only if" to improve clarity. I count 5 in total. Would you like to expand the scope of this PR a bit?

Of course, I'll take care of it.

ctz (approved):

djc (approved):

ctz:
Thanks!

PR #2016, opened 2024-06-21
`key_update` review follow-up

ctz:
Follow-on from #2003


cpu (approved):

rustls/src/client/tls13.rs
```diff
@@ -1555,8 +1555,7 @@ impl State<ClientConnectionData> for ExpectTraffic {
     }
 
     fn send_key_update_request(&mut self, common: &mut CommonState) -> Result<(), Error> {
-        common.check_aligned_handshake()?;
-        common.send_msg(Message::build_key_update_request(), true);
+        common.send_key_update_request()?;
```
djc, line 1558: Isn't it possible to deduplicate the `update_encrypter()` call too? Maybe the call to `common.send_key_update_request()` could be moved into `update_encrypter()`?

djc (approved):
Nice -- hope you felt that was still worth it.

rustls/src/client/tls13.rs
ctz: Ah, yes, I have moved it into `update_encrypter` and renamed that, so it's similar in scope to the existing `update_encrypter_and_notify`

PR #2022, opened 2024-06-24
Tighten up ticketer decryption

### LINKED ISSUE #2023: Partitioning Oracle Attacks against TLS Session Tickets
This is a copy of an advisory filed privately by @mradoy99, @xomex, and @jurajsomorovsky. Reproduced here publicly for transparency, and because github advisories cannot be published if they describe a negative result. However, we are taking aboard the defensive recommendations: see #2021 and #2022.

## Summary
We analyzed whether partitioning oracle attacks [1] are applicable to TLS session tickets. We found that a partitioning oracle attack might be applicable to Rustls. We will publish a paper about this at ESORICS 24 [3] (September 16-20, 2024). In our paper we conclude that the attack does **not provide a significant advantage over simple bruteforce** at the moment. Hence this is more a recommendation and **not a security issue**.

## Details
AES-GCM and ChaCha20-Poly1305 use Carter-Wegman MACs. These compute the tag by evaluating a polynomial. Unfortunately, this allows constructing inputs (ciphertexts) that are valid under multiple keys. These ciphertexts are called *splitting ciphertexts*. Len, Grubbs and Ristenpart [1] presented how to construct splitting ciphertexts without a valid ciphertext. Armour and Cid [2] showed how to use a valid ciphertext to compute a splitting ciphertext, which allows to circumvent certain plaintext format checks. As TLS servers issue tickets, the latter approach can also be used.
The attacker has to guess which key(s) could be used by the server and compute a ciphertext/ticket that validly decrypts under all guesses.

We found that Rustls resumes the session, depending on whether the servers STEK is included in the guessed set (resumption) or not (no resumption). While this may seem intuitive (as in the latter case, the ticket cannot be decrypted), the session should not be resumed in the first case, as the ticket was generated/modified by the attacker.

To fix the issue, we suggest switching from a AEAD scheme to a key-committing scheme like AES-CBC with an HMAC. We further recommend imposing length limits on the resumed tickets.

## PoC
The library was tested using the included demo server. Using our tool, we created valid ciphertext adopted to the format requirements of the session ticket parser of Rustls.
To evaluate the issue we needed to make minimal modifications to the included test server, such that we can supply a known STEK to the library/extract the used STEK from the library.
Our code then creates two splitting ciphertexts: One with the correct STEK in the key set, the other without the STEK in the key set. We then observe the libraries behavior when trying to resume the ticket (built from the splitting ciphertext).
We can provide you with the our code, but unfortunately it is not a simple plug+play code. To evaluate your library we recommend appending some bytes to the ticket plaintext before parsing it and check how your library reacts.This mimics our attack, as we created the splitting ciphertext such that the plaintext contains the plaintext from a valid ticket, and then some random suffix blocks.

## Impact
The vulnerability could allow attackers to recover the key that is used to encrypt session tickets (session ticket encryption key / STEK). With this key, an attacker can passively decrypt TLS traffic. In TLS 1.2 (and prior) the session issuing a ticket and sessions resuming a ticket are affected. In TLS 1.3 the 0-RTT data are affected, and (if no additional key exchange is performed) sessions where a ticket is resumed.
To mount this attack, the attacker needs prior knowledge of which keys are more likely, as the attack does not scale well to test many keys. In general it is slower than brute force against a single server.
Rustls only authenticates the state, but not nonce or any other server specific data. This allows an attacker to precompute a ticket and use it against all Rustls servers, which could make this attack more efficient than brute force. Additionally, the key name is not included in the ticket format, which could be added to the authenticated additional data.

If you have any further questions, please dont hesitate to ask.

Please add Sven Hebrok (@XoMEX) and Juraj Somorovsky (@jurajsomorovsky) as collaborators to this advisory. Thank you.

Kind regards,
Maximilian Radoy
Research Assistant
Paderborn University

[1] https://www.usenix.org/conference/usenixsecurity21/presentation/len
[2] https://link.springer.com/chapter/10.1007/978-3-030-92548-2_3 or https://eprint.iacr.org/2021/1296
[3] https://esorics2024.org/

ctz:
These changes aim to make it harder to mount any theoretical attack against AEAD decryption due to the AEADs we use being not key-committing. The best current attacks against this aren't better than brute force, so there is no security issue here.

The last commit is a little unrelated, but is along for the ride after I noticed the poor error in the test changed in the second commit.

fixes #2023

### DIFF
--- a/rustls/src/crypto/ring/ticketer.rs
+++ b/rustls/src/crypto/ring/ticketer.rs
@@ -7,6 +7,8 @@ use core::fmt;
 use core::fmt::{Debug, Formatter};
 use core::sync::atomic::{AtomicUsize, Ordering};
 
+use subtle::ConstantTimeEq;
+
 use super::ring_like::aead;
 use super::ring_like::rand::{SecureRandom, SystemRandom};
 use super::TICKETER_AEAD;
@@ -57,9 +59,15 @@ fn make_ticket_generator() -> Result<Box<dyn ProducesTickets>, GetRandomFailed> 
 
     let key = aead::UnboundKey::new(TICKETER_AEAD, &key).unwrap();
 
+    let mut key_name = [0u8; 16];
+    SystemRandom::new()
+        .fill(&mut key_name)
+        .map_err(|_| GetRandomFailed)?;
+
     Ok(Box::new(AeadTicketer {
         alg: TICKETER_AEAD,
         key: aead::LessSafeKey::new(key),
+        key_name,
         lifetime: 60 * 60 * 12,
         maximum_ciphertext_len: AtomicUsize::new(0),
     }))
@@ -72,6 +80,7 @@ fn make_ticket_generator() -> Result<Box<dyn ProducesTickets>, GetRandomFailed> 
 struct AeadTicketer {
     alg: &'static aead::Algorithm,
     key: aead::LessSafeKey,
+    key_name: [u8; 16],
     lifetime: u32,
 
     /// Tracks the largest ciphertext produced by `encrypt`, and
@@ -101,15 +110,27 @@ impl ProducesTickets for AeadTicketer {
             .fill(&mut nonce_buf)
             .ok()?;
         let nonce = aead::Nonce::assume_unique_for_key(nonce_buf);
-        let aad = aead::Aad::empty();
+        let aad = aead::Aad::from(self.key_name);
 
-        let mut ciphertext =
-            Vec::with_capacity(nonce_buf.len() + message.len() + self.key.algorithm().tag_len());
+        // ciphertext structure is:
+        // key_name: [u8; 16]
+        // nonce: [u8; 12]
+        // message: [u8, _]
+        // tag: [u8; 16]
+
+        let mut ciphertext = Vec::with_capacity(
+            self.key_name.len() + nonce_buf.len() + message.len() + self.key.algorithm().tag_len(),
+        );
+        ciphertext.extend(self.key_name);
         ciphertext.extend(nonce_buf);
         ciphertext.extend(message);
         let ciphertext = self
             .key
-            .seal_in_place_separate_tag(nonce, aad, &mut ciphertext[nonce_buf.len()..])
+            .seal_in_place_separate_tag(
+                nonce,
+                aad,
+                &mut ciphertext[self.key_name.len() + nonce_buf.len()..],
+            )
             .map(|tag| {
                 ciphertext.extend(tag.as_ref());
                 ciphertext
@@ -133,9 +154,26 @@ impl ProducesTickets for AeadTicketer {
             return None;
         }
 
-        // Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
-        let nonce = ciphertext.get(..self.alg.nonce_len())?;
-        let ciphertext = ciphertext.get(nonce.len()..)?;
+        let (alleged_key_name, ciphertext) = try_split_at(ciphertext, self.key_name.len())?;
+
+        let (nonce, ciphertext) = try_split_at(ciphertext, self.alg.nonce_len())?;
+
+        // checking the key_name is the expected one, *and* then putting it into the
+        // additionally authenticated data is duplicative.  this check quickly rejects
+        // tickets for a different ticketer (see `TicketSwitcher`), while including it
+        // in the AAD ensures it is authenticated independent of that check and that
+        // any attempted attack on the integrity such as [^1] must happen for each
+        // `key_label`, not over a population of potential keys.  this approach
+        // is overall similar to [^2].
+        //
+        // [^1]: https://eprint.iacr.org/2020/1491.pdf
+        // [^2]: "Authenticated Encryption with Key Identification", fig 6
+        //       <https://eprint.iacr.org/2022/1680.pdf>
+        if ConstantTimeEq::ct_ne(&self.key_name[..], alleged_key_name).into() {
+            #[cfg(debug_assertions)]
+            debug!("rejected ticket with wrong ticket_name");
+            return None;
+        }
 
         // This won't fail since `nonce` has the required length.
         let nonce = aead::Nonce::try_assume_unique_for_key(nonce).ok()?;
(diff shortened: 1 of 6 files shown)


rustls/src/crypto/ring/ticketer.rs
```diff
@@ -69,6 +71,10 @@ struct AeadTicketer {
     alg: &'static aead::Algorithm,
     key: aead::LessSafeKey,
     lifetime: u32,
+
+    /// Tracks the largest ciphertext produced by `encrypt`, and
+    /// uses it to early-reject `decrypt` queries that are too long.
+    maximum_ciphertext_len: AtomicUsize,
```
cpu, line 86: Can we fold in a short explanation about why a peer might be sending too-long-ciphertexts? E.g. something about avoiding allowing a partitioning oracle to be constructed? I think it could be helpful to motivate the control being added.

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -100,10 +106,23 @@ impl ProducesTickets for AeadTicketer {
...
     }
 
     /// Decrypt `ciphertext` and recover the original message.
     fn decrypt(&self, ciphertext: &[u8]) -> Option<Vec<u8>> {
+        if ciphertext.len()
+            > self
+                .maximum_ciphertext_len
+                .load(Ordering::SeqCst)
+        {
+            return None;
```
cpu, line 144: Does this case merit some kind of a warning log? It seems like it shouldn't occur under normal circumstances.

cpu (approved):
Looks good to me, just two non-blocking comments/questions.

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -100,10 +106,23 @@ impl ProducesTickets for AeadTicketer {
                 ciphertext
             })
             .ok()
+            .map(|r| {
```
djc, line 130: Don't love the `map()` for this side-effecty thing. Maybe just stick `?` on the `ok()` and deal with this separately?

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -93,14 +102,26 @@ impl ProducesTickets for AeadTicketer {
             .fill(&mut nonce_buf)
             .ok()?;
         let nonce = aead::Nonce::assume_unique_for_key(nonce_buf);
-        let aad = aead::Aad::empty();
-
-        let mut ciphertext =
-            Vec::with_capacity(nonce_buf.len() + message.len() + self.key.algorithm().tag_len());
+        let aad = aead::Aad::from(self.key_name);
```
djc, line 105: I'm a little confused by the notion of sticking the key name both in the AAD and in the ciphertext. If we stick it in the ciphertext, what's the value of also putting it in the AAD?

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -153,6 +178,15 @@ impl Debug for AeadTicketer {
     }
 }
 
+/// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
+fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
+    if mid > slice.len() {
```
djc, line 183: Nit, consider `match` for this, which is slightly more concise?

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -153,6 +178,15 @@ impl Debug for AeadTicketer {
     }
 }
 
+/// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
+fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
```
djc, line 182: `split_at_checked()` coming... soonish (but will take a while to trickle out to our MSRV).

rustls/src/msgs/codec.rs
```diff
@@ -290,7 +290,13 @@ impl<'a, T: Codec<'a> + TlsListElement + Debug> Codec<'a> for Vec<T> {
         let len = match T::SIZE_LEN {
             ListLength::U8 => usize::from(u8::read(r)?),
             ListLength::U16 => usize::from(u16::read(r)?),
-            ListLength::U24 { max } => Ord::min(usize::from(u24::read(r)?), max),
+            ListLength::U24 { max, error } => {
+                let actual_length = usize::from(u24::read(r)?);
+                if actual_length > max {
```
djc, line 295: `match` with symmetric arms?

djc (approved):

rustls/src/crypto/ring/ticketer.rs
ctz: I think we should avoid doing any extraneous work in this branch (or any other ticket rejection branches). However I'd be open to doing this gated by `#[cfg(debug_assertions)]`?

rustls/src/crypto/ring/ticketer.rs
ctz: ah, that's great. i went looking for `try_split_at` and didn't find that

rustls/src/crypto/ring/ticketer.rs
cpu: Sounds reasonable 👍

rustls/src/crypto/ring/ticketer.rs
ctz: I mentioned this a bit in the commit message: doing both is unnecessary, but also not harmful.

djc: I'm probably being stupid or lack your cryptography expertise, but to me this doesn't really follow from your commit message. Anyway -- if doing both is unnecessary, why do it? Also I think this merits a comment in the code where it provides context that isn't readily available from the commit message.

rustls/src/msgs/codec.rs
ctz: Something like 

```rust
            ListLength::U24 { max, error } => match usize::from(u24::read(r)?) {
                len if len > max => return Err(error),
                len => len,
            },
```
?

rustls/src/crypto/ring/ticketer.rs
cpu: New text is exactly what I was looking for :+1:  Thanks!

rustls/src/crypto/ring/ticketer.rs
ctz: Added a comment and some references, especially one to https://eprint.iacr.org/2022/1680.pdf where figure 6 matches what is going on here pretty well.

rustls/src/msgs/codec.rs
djc: Yup!

djc (approved):

PR #2024, opened 2024-06-25
Benchmark unbuffered API

### LINKED ISSUE #1816: Benchmark the `Unbuffered{Client,Server}Connection` API
We should extend our benchmarking programs to demonstrate/validate the performance benefit in these APIs.

ctz:
- [x] bulk performance
 - [x] handshake performance

fixes #1816


rustls/examples/internal/bench_impl.rs
```diff
@@ -588,6 +640,230 @@ impl KeyType {
...
+        }
+    }
+
+    fn read_and_discard(&mut self, mut expected: usize, input: &mut [u8]) -> usize {
+        let mut input_used = 0;
+
+        while expected > 0 {
+            match self {
+                Self::Client(client) => {
+                    match client.process_tls_records(&mut input[input_used..]) {
```
djc, line 846: Take this out of the match arm?

djc (commented):
Some early feedback!

rustls/examples/internal/bench_impl.rs
```diff
@@ -27,15 +28,123 @@ use rustls::{
...
+        }
+        Some(Command::ListSuites) => {
+            for bench in ALL_BENCHMARKS {
+                println!(
+                    "{:?} (key={:?} version={:?})",
+                    bench.ciphersuite, bench.key_type, bench.version
+                );
+            }
+        }
+        None => {
```
djc, line 78: This feels like a somewhat counterintuitive use of `Option`, maybe `Command::All` instead? The derived clap stuff looks great!

ctz: Did that, though the underlying value is still an Option, `fn command()` drops `None` to `Command::AllTests` (suggestion from https://github.com/clap-rs/clap/issues/975#issuecomment-1888988717 )

cpu (approved):
Really nice, great work :-)

cpu (approved):

rustls/examples/internal/bench_impl.rs
```diff
@@ -363,24 +502,18 @@ fn make_client_config(
     cfg
 }
 
-fn apply_work_multiplier(work: u64) -> u64 {
```
djc, line 366: I don't see a strong justification for this change in this commit... Maybe move it to a separate commit if the `Options` structure will be helpful later on?

rustls/examples/internal/bench_impl.rs
```diff
@@ -131,6 +250,15 @@ enum ResumptionParam {
 }
 
 impl ResumptionParam {
+    fn from_subcommand(cmd: &Command) -> Self {
```
djc: Nit: if we're going to use `Command` here, maybe it should go below `ResumptionParam`? (This is admittedly one of those cases where a strict ordering might not make sense.)

rustls/examples/internal/bench_impl.rs
```diff
@@ -252,24 +252,41 @@ fn bench_bulk(
         max_fragment_size,
     ));
 
+    let total_data = options.apply_work_multiplier(if plaintext_size < 8192 {
+        64 * 1024 * 1024
```
djc, line 400: Nit: potentially extract the `1024 * 1024` factor from the `if`? (A `match` would be slightly more concise.)

djc (approved):

rustls/examples/internal/bench_impl.rs
ctz: This was necessary because `apply_work_multiplier` before used the "ambient" environment, but now the value comes from clap and needs to be passed down from there. But there's a degree of looking ahead in putting it in the `Options` struct.

rustls/examples/internal/bench_impl.rs
ctz: I think `Command` warrants being pretty high up in this file given its value in explaining what comes after it.

djc: Fair!

PR #2025, opened 2024-06-25
add warn for elided_lifetimes_in_paths

cpu:
Our style guide [already recommended not eliding generic lifetimes](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#dont-elide-generic-lifetimes). This branch adds them wherever they were missing and adds the `elided_lifetimes_in_paths` clippy lint as a warn (h/t djc for showing that lint exists in https://github.com/rustls/webpki/pull/262).


rustls/src/lib.rs
```diff
@@ -319,7 +319,8 @@
     unreachable_pub,
     unused_import_braces,
     unused_extern_crates,
-    unused_qualifications
+    unused_qualifications,
+    elided_lifetimes_in_paths
```
djc, line 323: Maybe add this in alphabetic order?

djc (commented):
Thanks for taking this and running with it!

djc (approved):

rustls/src/lib.rs
cpu: Done.

rustls/src/server/hs.rs
```diff
@@ -564,9 +564,9 @@ impl State<ServerConnectionData> for ExpectClientHello {
 ///
 /// [`ResolvesServerCert`]: crate::server::ResolvesServerCert
 pub(super) fn process_client_hello<'a>(
-    m: &'a Message,
+    m: &'a Message<'_>,
```
ctz, line 567: I find this one confusing. In fact I don't think I understand what the lifetimes here actually are or which are different names for the same span. Like we have `Message<'m>` in the immediate caller, but then that is inferred in this function, where `'a` is the caller's scope. Then the return from the caller is `'static`, not `'m`? So can we delete/infer `'m` in the caller?

djc: I think it would make sense to use `'a` here (or rename `'a` to '`m`).

cpu: I think we can infer the `'m` on the `Message<'_>` lifetime in the callsite, but we still need the explicit `'m` there for the `Self` clause used for the `NextStateOrError` return  - it's used to signal the borrow comes from `m` and not `cx` here AFAICT

Something like:

```diff
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -540,7 +540,7 @@ impl State<ServerConnectionData> for ExpectClientHello {
     fn handle<'m>(
         self: Box<Self>,
         cx: &mut ServerContext<'_>,
-        m: Message<'m>,
+        m: Message<'_>,
     ) -> NextStateOrError<'m>
     where
         Self: 'm,
```

cpu: > (or rename 'a to 'm).

That sounds good to me. I'll push an adjustment.

cpu: > I'll push an adjustment.

Done

ctz (approved):

Issue #2019, opened 2024-06-23
Add support for determining the key exchange for connection


filiptronicek:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**

I would like to be able to determine the key exchange algorithm used in a TLS connection.

**Describe the solution you'd like**

I would like to be able to access the key exchange mechanism of a TLS connection, just like I can access the negotiated cipher suite name. It could potentially live somewhere alongside or inside the `.conn.negotiated_cipher_suite()` method.

**Describe alternatives you've considered**

I have tried to set up a `CryptoProvider` with only the desired kex group (`rustls_post_quantum::X25519Kyber768Draft00` in this case), which seems to work on sites including `pq.cloudflareresearch.com`. I have set it up like so:
```rs
let parent = aws_lc_rs::default_provider();
let kyber_provider = CryptoProvider {
    kx_groups: vec![&rustls_post_quantum::X25519Kyber768Draft00],
    ..parent
};

if CryptoProvider::install_default(kyber_provider).is_err() {
    return false;
}
```

The problem with this approach is that it does not work if I added more groups to the `CryptoProvider`, because I would not be able to tell which actually ended up being used. I could only tell that if the handshake succeeded, it had to be one of the ones I provided.

djc:
Why do you want to know this? What's the use case/motivation here? Just research?

filiptronicek:
@djc that's right, I am gathering data on the adoption of key exchange mechanisms. Apart from that, it seems providing some of the data tools like the Google Chrome DevTools show could be useful not just in research, but also things like e2e tests of TLS.
<img width="623" alt="image" src="https://github.com/rustls/rustls/assets/29888641/8f51ac69-e22b-4212-9f95-51dafa5526e3">

cpu:
>  Apart from that, it seems providing some of the data tools like the Google Chrome DevTools show could be useful not just in research, but also things like e2e tests of TLS.

I think providing details from the peer certificate chain is a harder sell but from the "Connection" info shown in your screenshot I think it's just the key exchange information that's missing. From my perspective it does seem reasonable to expose that information so I've drafted a PR in #2028 exposing it.

cpu:
Support for this was added to `main` as [`CommonState::negotiated_key_exchange_group`](https://rustls.github.io/rustls/prerelease/struct.CommonState.html#method.negotiated_key_exchange_group).

cpu:
[...]

PR #2028, opened 2024-06-27
common_state: expose key exchange group

cpu:
Similar to [`negotiated_cipher_suite()`](https://docs.rs/rustls/latest/rustls/struct.CommonState.html#method.negotiated_cipher_suite) it feels like there's utility in knowing the key exchange group that was used for the connection. In particular it also let us tighten up some FFHDE related unit tests where the expected key exchange was implied by success/failure but not verified precisely.  Both those unit tests, some similar ones interrogating the negotiated ciphersuite, and the post-quantum example binary were updated to also interrogate the KEX group.

Resolves https://github.com/rustls/rustls/issues/2019

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -17,7 +17,7 @@ use crate::client::client_conn::ClientConnectionData;
 use crate::client::common::ClientHelloDetails;
 use crate::client::ech::EchState;
 use crate::client::{tls13, ClientConfig, EchMode, EchStatus};
-use crate::common_state::{CommonState, HandshakeKind, State};
+use crate::common_state::{CommonState, HandshakeKind, KxState, State};
 use crate::conn::ConnectionRandoms;
 use crate::crypto::{ActiveKeyExchange, KeyExchangeAlgorithm};
 use crate::enums::{AlertDescription, CipherSuite, ContentType, HandshakeType, ProtocolVersion};
@@ -109,7 +109,11 @@ pub(super) fn start_handshake(
     let mut resuming = find_session(&server_name, &config, cx);
 
     let key_share = if config.supports_version(ProtocolVersion::TLSv1_3) {
-        Some(tls13::initial_key_share(&config, &server_name)?)
+        Some(tls13::initial_key_share(
+            &config,
+            &server_name,
+            &mut cx.common.kx_state,
+        )?)
     } else {
         None
     };
@@ -1035,6 +1039,7 @@ impl ExpectServerHelloOrHelloRetryRequest {
                     }
                 };
 
+                cx.common.kx_state = KxState::Start(skxg);
                 skxg.start()?
             }
             _ => offered_key_share,
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -12,7 +12,7 @@ use crate::check::inappropriate_handshake_message;
 use crate::client::common::{ClientAuthDetails, ClientHelloDetails, ServerCertDetails};
 use crate::client::ech::{self, EchState, EchStatus};
 use crate::client::{hs, ClientConfig, ClientSessionStore};
-use crate::common_state::{CommonState, HandshakeKind, Protocol, Side, State};
+use crate::common_state::{CommonState, HandshakeKind, KxState, Protocol, Side, State};
 use crate::conn::ConnectionRandoms;
 use crate::crypto::ActiveKeyExchange;
 use crate::enums::{
@@ -147,6 +147,7 @@ pub(super) fn handle_server_hello(
         KeySchedulePreHandshake::new(suite)
     };
 
+    cx.common.kx_state.complete();
     let shared_secret = our_key_share.complete(&their_key_share.payload.0)?;
 
     let mut key_schedule = key_schedule_pre_handshake.into_handshake(shared_secret);
@@ -228,6 +229,7 @@ fn validate_server_hello(
 pub(super) fn initial_key_share(
     config: &ClientConfig,
     server_name: &ServerName<'_>,
+    kx_state: &mut KxState,
 ) -> Result<Box<dyn ActiveKeyExchange>, Error> {
     let group = config
         .resumption
@@ -244,6 +246,7 @@ pub(super) fn initial_key_share(
                 .expect("No kx groups configured")
         });
 
+    *kx_state = KxState::Start(group);
     group.start()
 }
 
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -3,6 +3,7 @@ use alloc::vec::Vec;
 
 use pki_types::CertificateDer;
 
+use crate::crypto::SupportedKxGroup;
 use crate::enums::{AlertDescription, ContentType, HandshakeType, ProtocolVersion};
 use crate::error::{Error, InvalidMessage, PeerMisbehaved};
 #[cfg(feature = "logging")]
@@ -31,6 +32,7 @@ pub struct CommonState {
     pub(crate) side: Side,
     pub(crate) record_layer: record_layer::RecordLayer,
     pub(crate) suite: Option<SupportedCipherSuite>,
+    pub(crate) kx_state: KxState,
     pub(crate) alpn_protocol: Option<Vec<u8>>,
     pub(crate) aligned_handshake: bool,
     pub(crate) may_send_application_data: bool,
@@ -63,6 +65,7 @@ impl CommonState {
             side,
             record_layer: record_layer::RecordLayer::new(),
             suite: None,
+            kx_state: KxState::default(),
             alpn_protocol: None,
             aligned_handshake: true,
             may_send_application_data: false,
@@ -136,6 +139,22 @@ impl CommonState {
     /// This returns None until the ciphersuite is agreed.
     pub fn negotiated_cipher_suite(&self) -> Option<SupportedCipherSuite> {
         self.suite
+    }
+
+    /// Retrieves the key exchange group agreed with the peer.
+    ///
+    /// This function may return `None` depending on the state of the connection,
+    /// the type of handshake, and the protocol version.
+    ///
+    /// If [`CommonState::is_handshaking()`] is true this function will return `None`.
+    /// Similarly, if the [`CommonState::handshake_kind()`] is [`HandshakeKind::Resumed`]
+    /// and the [`CommonState::protocol_version()`] is TLS 1.2, then no key exchange will have
+    /// occurred and this function will return `None`.
+    pub fn negotiated_key_exchange_group(&self) -> Option<&'static dyn SupportedKxGroup> {
+        match self.kx_state {
+            KxState::Complete(group) => Some(group),
+            _ => None,
+        }
     }
 
     /// Retrieves the protocol version agreed with the peer.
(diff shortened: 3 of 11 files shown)


cpu:
Note due to the `NamedGroup::Unknown` variant (which, as discussed in https://github.com/rustls/rustls/pull/1845 should probably be named `NamedGroup::Other`) the consumer might have to do a bit of legwork with the result from our API. 

E.g. a connection made with the experimental `rustls-post-quantum` provider's KEX will give a connection that returns `NamedGroup::Unknown(25497)`, not something like `NamedGroup::X25519Kyber768Draft00`. You'd need to map the TLS identifier to a name yourself.

examples/src/bin/limitedclient.rs
```diff
@@ -53,6 +53,15 @@ fn main() {
         ciphersuite.suite()
     )
     .unwrap();
+    let kx_group = tls
```
djc, line 56: On balance, not convinced of the utility of adding this to every example? It's taken this long for someone to ask about the functionality which indicates to me that it's a much more rare requirement (compared to cipher suite).

djc (approved):

ctz (approved):
A few thoughts:

1) Shower thought: we could perhaps return `Option<&'static dyn SupportedKxGroup>`, but realistically the goals of a caller are better met by `NamedGroup`. Though `SupportedKxGroup` has a `Debug` bound, and also `fips()`, I guess the caller who wants that can just search through their provider for the named  one if they wanted that.

2) We can now implement bogo `-expect-curve-id`, `-on-initial-expect-curve-id`, and `-on-resume-expect-curve-id` options on top of this (either in this PR, or a later one).

3) There's a test called `resumption_combinations` that could usefully call this too, it should be set for all full handshakes and TLS1.3 resumptions, but not set for TLS1.2 resumptions.

djc:
> 1. Shower thought: we could perhaps return `Option<&'static dyn SupportedKxGroup>`, but realistically the goals of a caller are better met by `NamedGroup`. Though `SupportedKxGroup` has a `Debug` bound, and also `fips()`, I guess the caller who wants that can just search through their provider for the named  one if they wanted that.

Feels like that would align more with `negotiated_cipher_suite()` yielding `SupportedKxGroup`, I think I would prefer it over returning `NamedGroup` directly (going from `SupportedKxGroup` to `NamedGroup` is a trivial `name()` call away, while the other way around is less ergonomic).

examples/src/bin/limitedclient.rs
cpu: I think I'd like to keep it in the `rustls-post-quantum` client example where the KX choice is the fundamental point of the example. I'll remove the other instances :+1:

cpu: Removed from the other examples.

cpu:
> Shower thought: we could perhaps return Option<&'static dyn SupportedKxGroup>

I've adopted this idea, but it does come with a new complication. In the TLS 1.3 client case we don't hang on to a `&dyn SupportedKxGroup` very long. In `client/hs.rs`'s `start_handshake()` fn we almost immediately turn the `&dyn SupportedKxGroup` into a `Box<dyn ActiveKeyExchange>`. Once we've done that the `ActiveKeyExchange` trait doesn't facilitate going back to a `SupportedKxGroup` (and any changes to that trait would be breaking). 

We can't stash the `SupportedKxGroup` into the `CommonState` in `start_handshake()` because at this point in the handshake we don't know if the `SupportedKxGroup` will end up being used (e.g. based on HRR, resumption, etc) and we want `negotiated_key_exchange_group()` to return `None` at this stage. 

My solution here was to stash the `SupportedKxGroup` in the `ClientHelloDetails` and only extracting it to put into the `CommonState` later in the handshake once we've finished KX. It's a bit awkward but I think workable? If folks dislike this approach I would suggest we go back to exposing the `NamedGroup` since the `ActiveKeyExchange` trait does make that available.

> We can now implement bogo -expect-curve-id, -on-initial-expect-curve-id, and -on-resume-expect-curve-id options on top of this (either in this PR, or a later one).

Good idea, but I spent enough time fiddling with the above issue that I'd prefer to handle this in a follow-up PR on ~Tuesday if folks don't mind.

> There's a test called resumption_combinations that could usefully call this too, it should be set for all full handshakes and TLS1.3 resumptions, but not set for TLS1.2 resumptions.

I integrated this suggestion in this branch. Thanks! It was helpful and I ended up shifting where/when the KX group was put into the `CommonState` based on this test.

rustls/src/client/hs.rs
```diff
@@ -108,10 +108,11 @@ pub(super) fn start_handshake(
 
     let mut resuming = find_session(&server_name, &config, cx);
 
-    let key_share = if config.supports_version(ProtocolVersion::TLSv1_3) {
-        Some(tls13::initial_key_share(&config, &server_name)?)
+    let (group, key_share) = if config.supports_version(ProtocolVersion::TLSv1_3) {
+        let group_and_keyshare = tls13::initial_key_share(&config, &server_name)?;
+        (Some(group_and_keyshare.0), Some(group_and_keyshare.1))
```
cpu, line 113: This is a bit gross but I couldn't think of a nicer way to transform a `(&'static dyn SupportedKxGroup, Box<dyn ActiveKeyExchange>)` into a `(Option<&'static dyn SupportedKxGroup>, Option<Box<dyn ActiveKeyExchange>>)`.

Making this conditional emit an optional tuple instead of two optional values seemed like it made the rest of the code uglier.

Alternative suggestions welcome.

cpu:
> Some checks haven’t completed yet

:cloud_with_lightning_and_rain: Looks like [bad weather](https://www.githubstatus.com/incidents/9vwllhs2w1kj):
> Actions runs triggered by pull requests are experiencing start delays.

rustls/tests/api.rs
```diff
@@ -515,13 +519,33 @@ fn resumption_combinations() {
...
                 make_pair_for_configs(client_config.clone(), server_config.clone());
             do_handshake(&mut client, &mut server);
 
             assert_eq!(client.handshake_kind(), Some(HandshakeKind::Resumed));
             assert_eq!(server.handshake_kind(), Some(HandshakeKind::Resumed));
+            assert!(client
+                .negotiated_key_exchange_group()
+                .is_none());
+            assert!(server
+                .negotiated_key_exchange_group()
```
ctz, line 547: Interesting, I was expecting to see a `if version == &TLS12` somewhere around here.

cpu: Aha yes, I think I should be updating `cx.common.kx_group` in `handle_server_hello` lower than where I have it. I'll push a fix.

cpu: Fixed.

cpu:
I think the TLS 1.2 code paths are still setting the common state KX too soon. I will take another look at that & test coverage shortly.

rustls/src/client/hs.rs
```diff
@@ -160,6 +161,9 @@ pub(super) fn start_handshake(
         _ => None,
     };
 
+    let mut hello = ClientHelloDetails::new(extension_order_seed);
```
djc, line 164: It looks like this is the only callsite for `ClientHelloDetails::new()`? If so, I'd probably pass the `offered_group` in directly.

djc (approved):
LGTM!

ctz (approved):

rustls/src/common_state.rs
```diff
@@ -939,5 +952,13 @@ impl Default for TemperCounters {
     }
 }
 
+#[derive(Debug, Default)]
+pub(crate) enum KxState {
+    #[default]
```
ctz, line 957: This is new to me! 🧠

rustls/src/client/tls13.rs
```diff
@@ -147,6 +147,12 @@ pub(super) fn handle_server_hello(
         KeySchedulePreHandshake::new(suite)
     };
 
+    cx.common.kx_state = match cx.common.kx_state {
```
ctz, line 150: maybe this line and the one below could be just `cx.common.kx_state.complete()`?

(alongside a `fn complete(&mut self) { ... }` on `KxState`)

could take it or leave it

ctz (approved):

rustls/src/client/tls13.rs
cpu: I like :+1: I ended up replacing the `unreachable!` and went with:

```rust
    pub(crate) fn complete(&mut self) {
        debug_assert!(matches!(self, Self::Start(_)));
        if let Self::Start(group) = self {
            *self = Self::Complete(*group);
        }
    }
```

I think that's the right balance - if this assert invariant didn't hold in a prod build it feels better to have `negotiated_kx_group()` return `None` as opposed to panic'ing in `KxState::complete`.

cpu:
> We can now implement bogo -expect-curve-id, -on-initial-expect-curve-id, and -on-resume-expect-curve-id options on top of this (either in this PR, or a later one).

Unfortunately the API we're exposing doesn't make implementing these straightforward. 

For `-on-initial-expect-curve-id` and `-on-resume-expect-curve-id` we lack the granularity to return the initial curve vs the negotiated curve. I don't think we want to expose that level of detail to our consumers so these will stay unimplemented.

For `-expect-curve-id` bogo expects to get the matching curve back even for resumption connections (e.g. `CurveID-Resume-Client`). We return `None` from our API in this case because we didn't do a full KX. I don't think we should change that so we're left evaluating workarounds, but we're stuck here too since we can't easily conditionally assert equality on expected vs actual KX ignoring these specific tests because AFAIK the only signal to use to inform the skip is the test name containing "Resume" and that information isn't wired through to the shim.

djc:
> We return None from our API in this case because we didn't do a full KX.

Why not? Does the cipher suite getter return `None` for resumed connections?

cpu:
> Why not?

Because key exchange wasn't negotiated/performed. Do you have a non-bogo use-case in mind that would benefit from the information? 

> Does the cipher suite getter return None for resumed connections?

No, but the ciphersuite is actively used for a resumed connection.

ctz:
> Because key exchange wasn't negotiated/performed. Do you have a non-bogo use-case in mind that would benefit from the information?

I guess one point would be simplifying implementation of openssl `SSL_get_negotiated_group` with 100% fidelity in rustls-openssl-compat (to save you looking it up -- it returns the connection's group if there is one, or the original connection's group if not). But I'm wary of that as a justification for rustls API design decisions, and I think it makes more sense to me that this new API accurately reflects the state of the current connection.

djc:
> Because key exchange wasn't negotiated/performed. Do you have a non-bogo use-case in mind that would benefit from the information?

This caveat is not documented -- probably want to fix that?

I don't have a use case, just think it's not intuitive to many users.

cpu:
> This caveat is not documented -- probably want to fix that?

Agreed, I will fix and also add a pointer to the fn that returns whether it was a full handshake or not. 👍

rustls/src/common_state.rs
```diff
@@ -138,6 +141,18 @@ impl CommonState {
         self.suite
     }
 
+    /// Retrieves the key exchange group agreed with the peer.
+    ///
+    /// This returns None until the key exchange has been completed,
+    /// or if the [`CommonState::handshake_kind`] is
+    /// [`HandshakeKind::Resumed`].
```
ctz, line 148: There's a bit more nuance:

- for TLS1.2, there is never a key exchange in a resumption, so this condition is accurate
- for TLS1.3, there may or may not be a key exchange in a resumption, though as it happens rustls requires this as a matter of policy

djc (approved):

ctz (approved):

PR #2032, opened 2024-07-03
unbuffered: add missing deref for `CommonState`

ctz:
This has a number of important APIs, especially `alpn_protocol`, which are currently missing in the unbuffered API.


ctz:
(actually, I will add some calls that rely on this to `tests/unbuffered.rs`)

djc (approved):

cpu (approved):

PR #2033, opened 2024-07-05
Avoid markdown footnotes

ctz:
These are apparently non-standard(?) and don't show well in rust doc anyway.  They looked like:

Screenshot from 2024-07-05 13-34-31

Now this looks like:

image

As of https://github.com/rust-lang/rust/pull/127127 this is now a warning.


cpu (approved):
Thanks!

djc (approved):

PR #2034, opened 2024-07-06
Verify CertifiedKey consistency in ConfigBuilder with_single_cert methods

lvkv:
## Summary 

Building off of https://github.com/rustls/rustls/pull/1954, this PR updates `ConfigBuilder`'s `with_single_cert` functions to verify public and private key consistency using `CertifiedKey::keys_match`. This is part of a larger effort to verify `CertifiedKey` consistency over at https://github.com/rustls/rustls/issues/1918.

## Changes

* Refactor `AlwaysResolvesChain::new` to take an existing owned `CertifiedKey` instead of creating one
* Refactor `AlwaysResolvesChain::new_with_extras` to take an existing owned `CertifiedKey` instead of creating one
* Update `ConfigBuilder::with_single_cert` to verify key consistency with `CertifiedKey::keys_match`
* Update `ConfigBuilder::with_single_cert_with_ocsp` to verify key consistency with `CertifiedKey::keys_match`

### DIFF
--- a/rustls/src/server/builder.rs
+++ b/rustls/src/server/builder.rs
@@ -7,8 +7,8 @@ use pki_types::{CertificateDer, PrivateKeyDer};
 use crate::builder::{ConfigBuilder, WantsVerifier};
 use crate::crypto::CryptoProvider;
 use crate::error::Error;
-use crate::msgs::handshake::CertificateChain;
 use crate::server::{handy, ResolvesServerCert, ServerConfig};
+use crate::sign::CertifiedKey;
 use crate::time_provider::TimeProvider;
 use crate::verify::{ClientCertVerifier, NoClientAuth};
 use crate::{compress, versions, NoKeyLog};
@@ -74,7 +74,8 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
             .provider
             .key_provider
             .load_private_key(key_der)?;
-        let resolver = handy::AlwaysResolvesChain::new(private_key, CertificateChain(cert_chain));
+        let certified_key = CertifiedKey::new(cert_chain, private_key);
+        let resolver = handy::AlwaysResolvesChain::new(certified_key);
         Ok(self.with_cert_resolver(Arc::new(resolver)))
     }
 
@@ -100,11 +101,8 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
             .provider
             .key_provider
             .load_private_key(key_der)?;
-        let resolver = handy::AlwaysResolvesChain::new_with_extras(
-            private_key,
-            CertificateChain(cert_chain),
-            ocsp,
-        );
+        let certified_key = CertifiedKey::new(cert_chain, private_key);
+        let resolver = handy::AlwaysResolvesChain::new_with_extras(certified_key, ocsp);
         Ok(self.with_cert_resolver(Arc::new(resolver)))
     }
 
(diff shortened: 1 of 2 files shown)


lvkv:
Note: These changes don't (yet) pass tests. This is because any test that uses the `with_single_cert` builder functions will currently fail because no `SigningKey` yet implements `public_key`. This means that every `with_single_cert` and `with_single_cert_with_ocsp` invocation will result in an error `InconsistentKeys(Unknown)`, as it's impossible to determine key consistency without knowledge of the public key.

This means that these changes (as currently written) are blocked on `SigningKey`s implementing `public_key`.

lvkv:
Well, I suppose we could just black hole the "unknown" case too. I did that instead

ctz (approved):

rustls/src/server/builder.rs
```diff
@@ -101,7 +112,14 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
             .provider
             .key_provider
             .load_private_key(key_der)?;
+
         let certified_key = CertifiedKey::new(cert_chain, private_key);
+        match certified_key.keys_match() {
```
djc, line 117: I think this is a better way of spelling it:

```rust
match certified_key.keys_match() {
    // Don't treat unknown consistency as an error
    Ok(()) | Err(Error::InconsistentKeys(InconsistentKeys::Unknown)) => (),
    Err(err) => return Err(err),
}
```

rustls/src/server/builder.rs
```diff
@@ -64,6 +64,8 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
     /// but other `CryptoProviders` may not.
     ///
     /// This function fails if `key_der` is invalid.
+    ///
+    /// This function fails if `key_der` is found to be inconsistent with `cert_chain`.
```
djc, line 68: I don't love this documentation addition. For one thing, IMO it would be better not to repeat the first part of the sentence, and we could instead concatenate them together somehow. Second, without context, I think "found to be inconsistent" is pretty vague. Proposal:

"This function fails if `key_der` is invalid, or if the subject public key info from the private key does not match the public key for the end entity certificate from the `cert_chain`."

cpu (approved):
I agree with Djc's feedback but otherwise think this looks great. Thanks again @lvkv !

rustls/src/server/builder.rs
lvkv: Done! And agreed—this makes the intention and behavior more obvious

rustls/src/server/builder.rs
lvkv: Done!

cpu (approved):

cpu:
> CIFuzz / Fuzzing (pull_request) Failing after 1m

Just wanted to note this is failing across many branches and not something blocking/related to this work.

djc (approved):

cpu:
>> CIFuzz / Fuzzing (pull_request) Failing after 1m
>
> Just wanted to note this is failing across many branches and not something blocking/related to this work.

I think we should probably admin merge in this case? I took a quick peek at https://github.com/google/oss-fuzz/issues/12168 and wasn't able to spot an immediately obvious culprit, and I think pinning a working version might be non-trivial.

ctz:
> I think we should probably admin merge in this case?

AFAIK this job is not required for a successful merge.

cpu:
Oh, that makes things easier. My mistake

PR #2037, opened 2024-07-08
Prepare 0.23.11

ctz:
Draft release notes:

* **New feature**: API for determining whether a `CertifiedKey`'s certificate and private key matches: [`keys_match()`](https://rustls.github.io/rustls/prerelease/sign/struct.CertifiedKey.html#method.keys_match). This is called from existing fallible functions that accept a private key and certificate (for example, [`with_single_cert()`](https://rustls.github.io/rustls/prerelease/struct.ConfigBuilder.html#method.with_single_cert)) so these functions now detect this misconfiguration.
  
  This relies on a new -- optional -- member of the `SigningKey` trait: [`public_key()`](https://rustls.github.io/rustls/prerelease/sign/trait.SigningKey.html#method.public_key) -- downstream implementers of this trait can opt-in to this behavior by implementing this method.

* **New feature**: API for determining which key exchange group a connection used: [`negotiated_key_exchange_group()`](https://rustls.github.io/rustls/prerelease/server/struct.ServerConnection.html#method.negotiated_key_exchange_group)

* **New feature**: Automatic sending of TLS1.3 `key_update` messages  to avoid exceeding AEAD confidentiality limits. This is complemented with a new API for manual use, [`refresh_traffic_keys()`](https://rustls.github.io/rustls/prerelease/client/struct.ClientConnection.html#method.refresh_traffic_keys)
* Expose common connection items in unbuffered API ([docs](https://rustls.github.io/rustls/prerelease/client/struct.UnbufferedClientConnection.html#deref-methods-CommonState))

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2211,7 +2211,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.10"
+version = "0.23.11"
 dependencies = [
  "aws-lc-rs",
  "base64 0.22.1",
@@ -2252,7 +2252,7 @@ dependencies = [
  "fxhash",
  "itertools 0.13.0",
  "rayon",
- "rustls 0.23.10",
+ "rustls 0.23.11",
  "rustls-pemfile 2.1.2",
  "rustls-pki-types",
  "tikv-jemallocator",
@@ -2265,7 +2265,7 @@ dependencies = [
  "hickory-resolver",
  "regex",
  "ring",
- "rustls 0.23.10",
+ "rustls 0.23.11",
 ]
 
 [[package]]
@@ -2279,7 +2279,7 @@ dependencies = [
  "log",
  "mio",
  "rcgen",
- "rustls 0.23.10",
+ "rustls 0.23.11",
  "rustls-pemfile 2.1.2",
  "rustls-pki-types",
  "serde",
@@ -2297,7 +2297,7 @@ dependencies = [
  "num-bigint",
  "once_cell",
  "openssl",
- "rustls 0.23.10",
+ "rustls 0.23.11",
  "rustls-pemfile 2.1.2",
  "rustls-pki-types",
 ]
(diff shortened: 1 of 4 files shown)


djc (approved):
IMO we should get the change in to implement the public key stuff for the built-in key types so we can get rid of that caveat (and also merge #2034). Seems pretty simple/straightforward?

ctz:
I agree that it would be better to deliver this in one piece -- I will look into what that will take.

cpu:
> New feature: Automatic sending of TLS1.3 key_update messages, and API for the manual use: [refresh_traffic_keys()](https://rustls.github.io/rustls/prerelease/client/struct.ClientConnection.html#method.refresh_traffic_keys)

To align with the roadmap would it be sensible to mention the confidentiality/integrity limits that are part of the "automatic sending" of the key update messages? Maybe something like:

> New feature: Automatic sending of TLS1.3 key_update messages to avoid exceeding AEAD confidentiality/integrity limits. This is complemented with a new API for manual use, [refresh_traffic_keys()](https://rustls.github.io/rustls/prerelease/client/struct.ClientConnection.html#method.refresh_traffic_keys).

On a separate topic, should the changelog have an entry for the work in https://github.com/rustls/rustls/pull/2022 or do you think that's too internal to be of note?

ctz:
Reworded the key/cert checking item now this is fully landed -- please recheck.

> To align with the roadmap would it be sensible to mention the confidentiality/integrity limits that are part of the "automatic sending" of the key update messages? Maybe something like:

Thanks, taken that almost as is (dropped the integrity mention, as it doesn't really apply to TCP-TLS).

> On a separate topic, should the changelog have an entry for the work in #2022 or do you think that's too internal to be of note?

I think that is too unactionable/internal to warrant a headline.

djc:
"behaviour" -> "behavior" for the published release notes? 😛

djc (approved):

cpu (approved):
:rocket:

PR #2042, opened 2024-07-11
deps: aws-lc-rs 1.6 => 1.8

tryor:


djc:
1.6 and 1.8 are semver-compatible, so 1.6 is merely acting as a lower bound here (the ^ operator, unlike in other semver implementations, is implicit in Cargo manifests).

As such, I don't see a reason to merge this.

PR #2044, opened 2024-07-14
rfc7685 ClientHello Padding Extension

fox0:


rustls/src/error.rs
```diff
@@ -183,6 +183,8 @@ pub enum InvalidMessage {
     UnsupportedCurveType,
     /// A peer sent an unsupported key exchange algorithm.
     UnsupportedKeyExchangeAlgorithm(KeyExchangeAlgorithm),
+    /// The client MUST fill the padding extension completely with zero bytes.
+    InvalidPadding,
```
djc, line 187: Nit: please keep these in alphabetic order.

rustls/src/msgs/handshake.rs
```diff
@@ -542,16 +542,17 @@ impl TlsListElement for CertificateCompressionAlgorithm {
 
 #[derive(Clone, Debug)]
 pub enum ClientExtension {
+    ServerName(Vec<ServerName>),
```
djc, line 545: Nit: please avoid this unrelated change.

rustls/src/msgs/handshake.rs
```diff
@@ -567,16 +568,17 @@ pub enum ClientExtension {
 impl ClientExtension {
     pub(crate) fn ext_type(&self) -> ExtensionType {
         match *self {
+            Self::ServerName(_) => ExtensionType::ServerName,
```
djc, line 571: Please avoid.

rustls/src/msgs/handshake.rs
```diff
@@ -599,10 +601,10 @@ impl Codec<'_> for ClientExtension {
 
         let nested = LengthPrefixedBuffer::new(ListLength::U16, bytes);
         match *self {
+            Self::ServerName(ref r) => r.encode(nested.buf),
```
djc, line 604: Avoid.

rustls/src/msgs/handshake.rs
```diff
@@ -630,10 +633,10 @@ impl Codec<'_> for ClientExtension {
         let mut sub = r.sub(len)?;
 
         let ext = match typ {
+            ExtensionType::ServerName => Self::ServerName(Vec::read(&mut sub)?),
```
djc, line 636: Avoid.

ctz:
Could you explain what this is for? As it currently stands this PR doesn't achieve much (there is no code that includes the padding extension to a client hello). So what we are doing here is just parsing and then discarding an additional extension? It is not necessary to parse unknown extensions, they can just be 100% ignored.

AIUI the original reason the padding extension was introduced is no longer relevant: there are no longer a meaningful population of corporate middleboxes that confuse SSLv2 and later versions.

fox0:
> Could you explain what this is for? As it currently stands this PR doesn't achieve much (there is no code that includes the padding extension to a client hello).  So what we are doing here is just parsing and then discarding an additional extension? It is not necessary to parse unknown extensions, they can just be 100% ignored.

I'm trying to rewrite this application in rust (https://github.com/stopcenz/fakesni), but it requires TLS padding (RFC 7685).

If I understand your words correctly, then this code is not enough to establish a tls connection and specify the extensions in the required order.

Let's close this pull request, and then I'll add a working patch (perhaps with an example) and create a new pull request?

cpu:
> I'm trying to rewrite this application in rust (https://github.com/stopcenz/fakesni), but it requires TLS padding (RFC 7685).

Based on a machine translation it sounds like the padding mechanism is being used to try and split the client hello across multiple packets such that the cleartext SNI is in the second half (the linked page itself links to [this page](https://ntc.party/t/http-headerstls-padding-as-a-censorship-circumvention-method/168) to describe that technique).

This seems very niche, and unlikely to be reliable for long as more middleboxes and censorship technology updates to gracefully handle reassembly of the client hello. Post quantum key exchange will make this more common (e.g. see https://tldr.fail/) so I suspect this mechanism for bypass is of decreasing value over time. I don't know that a reworked PR for this would be accepted.

Have you evaluated using [encrypted client hello](https://www.memorysafety.org/blog/rustls-ech-support/) instead? This is a much more robust mechanism for protecting the server name in the client hello.

fox0:
ESNI (Encrypted SNI) worked for a while (SNI has been removed from Firefox 85).
ECH (Encrypted Client Hello) looks promising. Thank you.

Issue #2040, opened 2024-07-09
unbuffered: crash (attempt to subtract with overflow) with TLS 1.3


honzasp:
**Describe the bug**
When using the unbuffered client API to connect to certain servers (https://canihazip.com, a HTTPS server which simply returns the client's IP address), rustls crashes in `<rustls::msgs::deframer::DeframerSliceBuffer as rustls::msgs::deframer::FilledDeframerBuffer>::filled` due to a numeric overflow in subtraction.

```
[examples/src/bin/unbuffered-client.rs:64:15] state.unwrap() = EncodeTlsData
[examples/src/bin/unbuffered-client.rs:64:15] state.unwrap() = TransmitTlsData
sent 240B of data
[examples/src/bin/unbuffered-client.rs:64:15] state.unwrap() = BlockedHandshake
received 3118B of data
[examples/src/bin/unbuffered-client.rs:64:15] state.unwrap() = EncodeTlsData
discarded 127B from `incoming_tls`
thread 'main' panicked at rustls/src/msgs/deframer.rs:600:19:
attempt to subtract with overflow
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:144:5
   3: <rustls::msgs::deframer::DeframerSliceBuffer as rustls::msgs::deframer::FilledDeframerBuffer>::filled
             at /home/jan-spacek/prog/keyrock/rustls/rustls/src/msgs/deframer.rs:600:19
   4: rustls::msgs::deframer::FilledDeframerBuffer::len
             at /home/jan-spacek/prog/keyrock/rustls/rustls/src/msgs/deframer.rs:651:9
   5: rustls::msgs::deframer::DeframerSliceBuffer::is_empty
             at /home/jan-spacek/prog/keyrock/rustls/rustls/src/msgs/deframer.rs:561:9
   6: rustls::msgs::deframer::MessageDeframer::pop
             at /home/jan-spacek/prog/keyrock/rustls/rustls/src/msgs/deframer.rs:49:19
[...]

PR #2049, opened 2024-07-19
Rewrite message deframer

ctz:
This PR starts with a repro case for #2040 , with the end goal of a) that and all other tests passing, and b) the deframer code being something I feel I can live with.


rustls/src/msgs/deframer/buffers.rs
```diff
@@ -0,0 +1,115 @@
...
+    slice: &'b [u8],
+}
+
+impl<'b> Delocator<'b> {
+    pub(crate) fn new(slice: &'b [u8]) -> Self {
+        Self { slice }
+    }
+
+    pub(crate) fn slice_from_range(&'_ self, range: &Range<usize>) -> &'b [u8] {
+        self.slice.get(range.clone()).unwrap()
```
cpu, line 46: Is there a safety comment to add w.r.t to this `unwrap()`? It seems like it's safe with the assumption the `HandshakeDeframer` is the only source of `&Range<usize>` being provided to the `Delocator`, and only for spans fully contained within the larger slice?

rustls/src/msgs/deframer/buffers.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl<'b> Coalescer<'b> {
+    pub(crate) fn new(slice: &'b mut [u8]) -> Self {
+        Self { slice }
+    }
+
+    pub(crate) fn copy_within(&mut self, from: Range<usize>, to: Range<usize>) {
+        debug_assert!(from.len() == to.len());
+        self.slice.copy_within(from, to.start);
```
cpu, line 66: `copy_within` says:
> This function will panic if either range exceeds the end of the slice, or if the end of src is before the start.

Is that latter condition something that should have a `debug_assert!()`?

rustls/src/msgs/deframer/buffers.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    pub(crate) fn add_processed(&mut self, processed: usize) {
+        self.processed += processed;
+    }
+
+    pub(crate) fn take_discard(&mut self) -> usize {
+        self.processed = self
+            .processed
+            .saturating_sub(self.discard);
```
cpu, line 108: it seems slightly non-obvious why `processed` is updated when you take `self.discard`. Maybe worth a clarifying comment?

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// Spans covering individual handshake payloads, in order of receipt.
+    spans: Vec<FragmentSpan>,
+}
+
+impl HandshakeDeframer {
+    /// Accepts a message into the deframer.
+    ///
+    /// `containing_buffer` allows mapping the message payload to its position
+    /// in the input buffer, and thereby avoid retaining a borrow on the input
+    /// buffer.
```
cpu, line 21: Perhaps helpful to lift some of the later context w.r.t _why_ retaining a borrow is problematic into this part of the comment?

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,452 @@
...
+
+        match ready {
+            0 => HandshakeDeframerStatus::Empty,
+            _ => HandshakeDeframerStatus::Ready(ready),
+        }
+    }
+
+    pub(crate) fn is_active(&self) -> bool {
+        !matches!(self.status(), HandshakeDeframerStatus::Empty)
+    }
```
cpu, line 88: nit: since `HandshakeDeframerStatus` impl's `PartialEq` I think this is more natural as:

```rust
    pub(crate) fn is_active(&self) -> bool {
        self.status() != HandshakeDeframerStatus::Empty
    }
```

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// this will never be the last item.
+    fn requires_coalesce(&self) -> Option<usize> {
+        match self.spans.split_last() {
+            Some((_last, elements)) => elements
+                .iter()
+                .enumerate()
+                .find_map(|(i, span)| (!span.is_complete()).then_some(i)),
+            None => None,
+        }
+    }
```
cpu, line 165: nit: I think you can get rid of the outer `match` by using `and_then()`:

```rust
    fn requires_coalesce(&self) -> Option<usize> {
        self.spans.split_last().and_then(|(_last, elements)|{
            elements
                .iter()
                .enumerate()
                .find_map(|(i, span)| (!span.is_complete()).then_some(i))
        })
    }
```

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,452 @@
...
+        while let Some(i) = self.requires_coalesce() {
+            self.coalesce_one(i, Coalescer::new(containing_buffer));
+        }
+    }
+
+    /// Within `containing_buffer`, move `span[index+1]` to be contiguous
+    /// with `span[index]`.
+    fn coalesce_one(&mut self, index: usize, mut containing_buffer: Coalescer<'_>) {
+        let second = self.spans.remove(index + 1);
+        let mut first = self.spans.remove(index);
```
cpu, line 124: ISTM the safety of the `remove` index being in-bounds is assumed based on `coalesce_one` only being called when `requires_colesce()` is `Some(_)`, and that ensures there's always at least one more span than the one at the returned index, right? I think so, just double checking my understanding.

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,452 @@
...
+        // for a pair where the first is not complete.  move
+        // the second down towards the first, then reparse the contents.
+        while let Some(i) = self.requires_coalesce() {
+            self.coalesce_one(i, Coalescer::new(containing_buffer));
+        }
+    }
+
+    /// Within `containing_buffer`, move `span[index+1]` to be contiguous
+    /// with `span[index]`.
+    fn coalesce_one(&mut self, index: usize, mut containing_buffer: Coalescer<'_>) {
```
cpu, line 122: This might be one of those `fn`'s that deserves the ASCII art "here's some buffers", "here's what they look like before", "here's what they look like after" treatment. 

I think the logic is reasonable, but I had to draw some of it out on paper to really track what was going on :dizzy_face:

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct FragmentSpan {
+    /// version taken from containing message.
+    version: ProtocolVersion,
+
+    /// size of the handshake message header (excluding header)
+    size: Option<usize>,
```
cpu, line 312: I think this comment should mention the significance of `size: None`.

rustls/src/msgs/deframer.rs
```diff
@@ -503,6 +481,21 @@ impl DeframerVecBuffer {
     fn unfilled(&mut self) -> &mut [u8] {
         &mut self.buf[self.used..]
     }
+
+    /// Append `bytes` to the end of this buffer.
+    ///
+    /// Return the span saying where it went.
```
cpu, line 124: Should this say "range" instead of "span" to differentiate from a `FragmentSpan`?

rustls/src/msgs/deframer.rs
```diff
@@ -734,6 +727,7 @@ const ALLOWED_CONSECUTIVE_EMPTY_FRAGMENTS_MAX: u8 = 32;
 #[cfg(feature = "std")]
 #[cfg(test)]
 mod tests {
+    /*
```
cpu: Intentionally left commented out?

rustls/src/conn.rs
```diff
@@ -862,57 +878,197 @@ impl<Data> ConnectionCore<Data> {
...
+        }
+    }
+
+    fn take_handshake_message<'b>(
+        &mut self,
+        buffer: &'b mut [u8],
+        buffer_range: &mut BufferRange,
+    ) -> InboundPlainMessage<'b> {
+        // unwrap: guaranteed by caller checking `hs_deframer.status()`
+        let (message, discard) = self
```
cpu, line 898: This fn body feels small enough that it might be safer to in-line it at the call-site so that the pre-condition and the `unwrap` depending on it are colocated. That also removes the potential it gets called from somewhere the invariant doesn't hold. WDYT?

rustls/src/conn.rs
```diff
@@ -1006,3 +1162,32 @@ impl<Data> ConnectionCore<Data> {
 
 /// Data specific to the peer's side (client or server).
 pub trait SideData: Debug {}
+
+/// An InboundPlainMessage which does not borrow its payload.
```
cpu, line 1166: > /// An InboundPlainMessage which does not borrow its payload.

:arrow_down: 

> /// An InboundPlainMessage which does not borrow its payload, but references a range that can later be borrowed. 

?

cpu (commented):
This is a really nice direction. I know from trying to untangle parts of this code before that it's very non-trivial to navigate the complexity of the task and the requirements imposed by the borrow checker. Kudos for managing :bow: It was a really interesting exercise to read through this.

Here's an initial pass with some feedback. IMO this feels like a case where individual commits help for review, but it might be better to land as a one-and-done squashed commit. Either that or once it's further along it might be sensible to revisit the intermediate commits to make sure they build/test/lint cleanly.

lozanopo (approved):

rustls/src/msgs/deframer/handshake.rs
ctz: Correct, yes.

rustls/src/msgs/deframer/handshake.rs
ctz: Added such diagrams to the docstring of `coalesce`

rustls/src/msgs/deframer.rs
ctz: Added a commit before this one that deletes the tests for the code that is then removed a few commits later.

rustls/src/conn.rs
ctz: I kinda like this small function, since it's called in two places which should always do the precise same thing. I've addressed this by eliminating the unwrap by making it return an `Option`, and removing the hardcoded `Some`s from the two call sites

rustls/src/msgs/deframer/buffers.rs
ctz: Added some assertions using `self.slice.get` which have the same preconditions.

ctz:
I think this is ready to go from my side.

rustls/src/msgs/deframer/outer.rs
```diff
@@ -0,0 +1,236 @@
...
+    consumed: usize,
+}
+
+impl DeframerIter<'_, '_> {
+    pub(crate) fn bytes_consumed(&self) -> usize {
+        self.consumed
+    }
+}
+
+impl<'a, 'b> Iterator for DeframerIter<'a, 'b> {
```
djc, line 46: I guess this is similar as the point that @cpu made before, but (perhaps more concretely) this would be easier to review if it was introduced in the same commit that removed the original version of this.

djc (commented):
Here's some initial feedback. So far I have not done an in-depth review yet -- after starting to look at types called `Locator` and `Delocator` I was a bit scared so I first want to ask some questions:

* What is the conceptual problem that this refactoring is solving? Presumably you're now close enough that you can articulate it in writing. Would be good to have some high-level design comments on what the important constraints here are and why this is the/an optional solution.
* It looks like this is a rewrite from scratch, and it shows in the commit history, which makes it harder to review.

I guess as a main contributor of the previous deframer (though before it got further complexified in order to support the unbuffered case) I'm still trying to grasp the (design/concept) issues here and how this makes rustls (and our lives as maintainers) better.

lozanopo (approved):

ctz:
> * What is the conceptual problem that this refactoring is solving? Presumably you're now close enough that you can articulate it in writing. Would be good to have some high-level design comments on what the important constraints here are and why this is the/an optional solution.

I think for me the previous deframer was something I could hold my nose at while it worked, but then inscrutable once I had to debug an issue in it. I think the root of that was one piece of code with too many responsibilities. It was a reasonable process how it got there -- having it decrypt the messages and join together the handshake stream was done to achieve very real performance improvements, but it became no longer necessary once the inbound message types started borrowing their data.

But, to be more concrete about the actual issue that got me to giving up on the existing deframer: basically `DeframerSliceBuffer::take` was incorrect in a ~unfixable way. It _had_ to move the buffer (via the `mem::take`/`split_at_mut` trick) to obtain a return value of the correct lifetime, but then had no reasonable way to adjust for that in indices into the buffer that came from outside the `DeframerSliceBuffer`. This happened to work in the buffered API, because `DeframerSliceBuffer::take` was the very last call on any given `DeframerSliceBuffer` object. But I'm quite surprised the unbuffered API worked as well as it did, because even a cursory exercise of the `DeframerSliceBuffer` API after `take()` is called shows the fundamental issue:

```rust
    #[test]
    fn take_panics() {
        let mut buf = [1, 2, 3];
        let mut dsb = DeframerSliceBuffer::new(&mut buf);
        let rs = RawSlice::from(&dsb.filled()[..1]);

        assert_eq!(&[1], dsb.take(rs));

        // any of the following lines panic (not just because the assert fails)
        assert_eq!(&[2, 3], dsb.filled());
        assert_eq!(2, dsb.len());
        assert!(!dsb.is_empty());
    }
```

To reframe that in terms of a conceptual problem, I'd say it is symptomatic of not being precise about what code requires mutable access to the input buffer (and the complexity which came with achieving that). The main reduction in complexity in this PR is achieved in minimising and containing those areas.

The other structural issue I didn't like with the existing code was how much complexity budget was applied to having the deframer be involved with writing messages into the underlying buffer for received quic handshake messages. I think the situation at the end of this PR in that area is an improvement.

Having spent quite a bit of time on this area of the code so far, I also have some future improvements in mind:

- at the moment we have (at least) two places where errors can be fused -- probably we can remove the one in the deframer, at which point we'll be left with just `DeframerIter` that requires no context, and the connection-level fused error in the `Err` variant of `state`.
- I think it would be ideal if we can make code that handles a single message (eg, `process_main_protocol`) instead work on an iterator of such messages, which should mean we can realise the benefits of `trait State` impls that only borrow their input (at the moment, this doesn't happen because `into_owned` is called unconditionally on these for every state transition.)

ctz:
> * at the moment we have (at least) two places where errors can be fused -- probably we can remove the one in the deframer, at which point we'll be left with just `DeframerIter` that requires no context, and the connection-level fused error in the `Err` variant of `state`.

Have thought about it a bit more, and I'm going to do this in this PR.

rustls/src/msgs/deframer/outer.rs
```diff
@@ -0,0 +1,236 @@
+use core::mem;
+
+use crate::error::{Error, InvalidMessage};
+use crate::msgs::codec::Reader;
+use crate::msgs::message::{
+    read_opaque_message_header, InboundOpaqueMessage, MessageError, HEADER_SIZE,
+};
+
+/// A deframer of outer TLS messages.
```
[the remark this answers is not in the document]
ctz: In the process of removing error fusing from here, the `OuterDeframer` disappeared. So I moved the `DeframerIter` back down into `deframer/mod.rs`. The remaining docstrings talk about "TLS wire messages" rather than "outer".

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,543 @@
...
+            // a single flight
+            spans: Vec::with_capacity(16),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum HandshakeDeframerStatus {
+    /// We have the given number of messages ready.
+    Ready(usize),
```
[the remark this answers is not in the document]
ctz: Thanks for the comments!

Yes, this is unused. I think actually `status()` itself is not really paying its rent, especially since there are other status indicators not included in it. Have eliminated that.

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,543 @@
...
+use super::buffers::{Coalescer, Delocator, Locator};
+use crate::msgs::codec::{u24, Codec};
+use crate::msgs::message::InboundPlainMessage;
+use crate::{ContentType, ProtocolVersion};
+
+#[derive(Debug)]
+pub(crate) struct HandshakeDeframer {
+    /// Spans covering individual handshake payloads, in order of receipt.
+    spans: Vec<FragmentSpan>,
+}
```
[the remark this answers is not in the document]
ctz: Good call -- done.

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,543 @@
...
+        let next_span = self.deframer.spans.get(self.index)?;
+
+        if !next_span.is_complete() {
+            return None;
+        }
+
+        // if this is the last handshake message, then we'll end
+        // up with an empty `spans` and can discard the remainder
+        // of the input buffer.
+        let discard = if self.deframer.spans.len() - 1 == self.index {
```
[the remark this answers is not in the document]
ctz: It is actually pretty crucial that no "discard" operation happens while there is any partial or full message, as it moves the buffer and therefore invalidates our ideas about where the message fragment is.

rustls/src/msgs/deframer/outer.rs
ctz: ^ Have done this.

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -0,0 +1,543 @@
...
+        debug_assert!(containing_buffer.fully_contains(msg.payload));
+
+        // if our last span is incomplete, we can blindly add this as a new span --
+        // no need to attempt parsing it with `DissectHandshakeIter`.
+        //
+        // `coalesce()` will later move this new message to be contiguous with
+        // `_last_incomplete`, and reparse the result.
+        //
+        // we cannot merge these processes, because `coalesce` mutates the underlying
+        // buffer, and `msg` borrows it.
```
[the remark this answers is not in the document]
ctz: I view `InboundUnborrowedMessage` as a bit of a hack as part of convincing the borrow checker to determine the correct lifetime relationship between the buffer and object returned by `deframe()`. So while this would work equally well, perhaps better, I'm minded to keep it as is.

cpu (approved):

ctz:
Going to merge this with one review, but as always happy to address later comments in a follow-up PR.

stevefan1999-personal:
```
  left: InvalidCertificate(Expired)
 right: InvalidCertificate(UnknownIssuer)


failures:
    tests_with_aws_lc_rs::tests::tls13_packed_handshake
```
Seems like the captured data is expired

ctz:
Thanks for the ping -- will address this shortly.

PR #2059, opened 2024-07-29
Update CLI of example tlsserver-mio to support early data and disabling resumption (closes  #2058)

### LINKED ISSUE #2058: Arguments "--resumption" and "--tickets" don't change server's behavior for TLS1.3 (in example tlsserver-mio)
When I used the tlsserver-mio example for TLS1.3 session resumption, I noticed that the "--tickets" and "--resumption" arguments don't change the server's behavior as I understood it from the description in --help:
`--resumption        Support session resumption.`
`--tickets           Support tickets.`
The server's behavior for TLS1.3 is always the same regardless of the arguments. When using a TLS1.2, the tickets are e.g. only sent if "--tickets" is used.

This was my test setup:
- Start tlsserver-mio with set arguments
- Run openssl s_client (1. handshake): `openssl s_client -connect localhost:443 -verifyCAfile ./root_ca_cert.pem -tls1_3 -sess_out session.pem`
- Stop s_client
- Run s_client again (2. handshake): `openssl s_client -connect localhost:443 -verifyCAfile ./root_ca_cert.pem -tls1_3 -sess_in session.pem`

I tested four variations of the arguments with the tlsserver-mio. Below you can see the resulting behavior and what I would've expected:
1. **With "--tickets" / With "-resumption":** `cargo run --bin tlsserver-mio -- --certs ./server_cert.pem --key ./server_key.pem --tickets --resumption --verbose echo`
   - Expected: server sends tickets and also accepts their usage in the 2. handshake
   - What happend: server sends session tickets and also accepts their usage in the 2. handshake

2. **No "--tickets" / With "-resumption":** `cargo run --bin tlsserver-mio -- --certs ./server_cert.pem --key ./server_key.pem --resumption --verbose echo`
   - Expected: not sure what exactly to expect as resumption in TLS1.3 is mainly done with tickets
   - What happend: server sends session tickets and also accepts their usage in the 2. handshake

3. **With "--tickets" / No "-resumption":** `cargo run --bin tlsserver-mio -- --certs ./server_cert.pem --key ./server_key.pem --tickets --verbose echo`
   - Expected: not sure what exactly to expect as resumption in TLS1.3 is mainly done with tickets
   - What happend: server sends session tickets and also accepts their usage in the 2. handshake

4. **No "--tickets" / No "-resumption":** `cargo run --bin tlsserver-mio -- --certs ./server_cert.pem --key ./server_key.pem --verbose echo`
   - Expected: Server doesn't send session tickets after first handshake / Server doesn't accept session tickets for 2. handshake
   - What happend: server sends session tickets and also accepts their usage in the 2. handshake

---

I came across this when I was playing around with 0-RTT and first couldn't figure out how to make the tlsserver-mio accept early data.
I added the line `config.max_early_data_size = 500;` add the end of `fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {...}` and used the s_client with early data in the 2. handshake: `openssl s_client -connect localhost:443 -verifyCAfile ./root_ca_cert.pem -tls1_3 -sess_in session.pem -early_data early_data.txt` 

The server only accepts early data in the configuration 2 and 4 from above. For 1 and 3 it gives the warning "WARN  rustls::server::tls13] early_data with stateless resumption is not allowed".
It seems like that the server sees tickets always as stateless resumption and doesn't want to accet early data then. I didn't find this immediately obvious.

Would be great if anyone can give some background whether this behavior and the differences between TLS1.2 and TLS1.3 are on purpose or not. If this helps, I can also provide a small documentation patch what to enable 0RTT with the example server.

Leonie-Theobald:
Two changes:
1. Changed `--resumption` flag to `--no-resumption` to have an option to disable stateful resumption as resumption is the default anyways
3. Added new `max-early-data BYTES` flag that allows opting in early data without change example code
    - if `max-early-data`is set to 0 no early data is allowed / for non-zero values the config is adapted to allow a maximum of BYTES bytes of early data (maximum is u32::MAX)
    - early data is only supported for stateful resumption => therefore it cannot be used together with `--ticket` flag

### DIFF
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -444,8 +444,8 @@ Options:
                         CRLFILE. May be used multiple times.
     --require-auth      Send a fatal alert if the client does not complete client
                         authentication.
-    --resumption        Support session resumption.
-    --tickets           Support tickets.
+    --no-resumption     Disable stateful session resumption.
+    --tickets           Support tickets (stateless resumption).
     --protover VERSION  Disable default TLS version list, and use
                         VERSION instead.  May be used multiple times.
     --suite SUITE       Disable default cipher suite list, and use
@@ -472,7 +472,7 @@ struct Args {
     flag_ocsp: Option<String>,
     flag_auth: Option<String>,
     flag_require_auth: bool,
-    flag_resumption: bool,
+    flag_no_resumption: bool,
     flag_tickets: bool,
     arg_fport: Option<u16>,
 }
@@ -640,8 +640,8 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
 
     config.key_log = Arc::new(rustls::KeyLogFile::new());
 
-    if args.flag_resumption {
-        config.session_storage = rustls::server::ServerSessionMemoryCache::new(256);
+    if args.flag_no_resumption {
+        config.session_storage = Arc::new(rustls::server::NoServerSessionStorage {});
     }
 
     if args.flag_tickets {
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -446,6 +446,7 @@ Options:
                         authentication.
     --no-resumption     Disable stateful session resumption.
     --tickets           Support tickets (stateless resumption).
+    --max-early-data BYTES   Support receiving BYTES many bytes with 0RTT
     --protover VERSION  Disable default TLS version list, and use
                         VERSION instead.  May be used multiple times.
     --suite SUITE       Disable default cipher suite list, and use
@@ -474,6 +475,7 @@ struct Args {
     flag_require_auth: bool,
     flag_no_resumption: bool,
     flag_tickets: bool,
+    flag_max_early_data: u32,
     arg_fport: Option<u16>,
 }
 
@@ -646,6 +648,19 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
 
     if args.flag_tickets {
         config.ticketer = provider::Ticketer::new().unwrap();
+    }
+
+    if args.flag_max_early_data > 0 {
+        if !versions.contains(&&rustls::version::TLS13) {
+            panic!("Early data is only available for servers supporting TLS1.3");
+        }
+        if args.flag_no_resumption {
+            panic!("Early data requires resumption.");
+        }
+        if args.flag_tickets {
+            panic!("Early data is not supported for stateless resumption (--tickets).");
+        }
+        config.max_early_data_size = args.flag_max_early_data;
     }
 
     config.alpn_protocols = args
--- a/README.md
+++ b/README.md
@@ -234,8 +234,9 @@ Options:
                         CRLFILE. May be used multiple times.
     --require-auth      Send a fatal alert if the client does not complete client
                         authentication.
-    --resumption        Support session resumption.
-    --tickets           Support tickets.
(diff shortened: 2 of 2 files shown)


examples/src/bin/tlsserver-mio.rs
```diff
@@ -640,7 +640,9 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
 
     config.key_log = Arc::new(rustls::KeyLogFile::new());
 
-    if args.flag_resumption {
+    if args.flag_no_resumption {
+        config.session_storage = Arc::new(rustls::server::NoServerSessionStorage {});
+    } else {
         config.session_storage = rustls::server::ServerSessionMemoryCache::new(256);
```
ctz, line 648: Could remove this branch, as it is the same as the default.

examples/src/bin/tlsserver-mio.rs
```diff
@@ -640,14 +642,29 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
...
     }
 
     if args.flag_tickets {
         config.ticketer = provider::Ticketer::new().unwrap();
     }
 
+    if args.flag_max_early_data > 0 {
+        if !&versions.contains(&&rustls::version::TLS13) {
```
cpu, line 656: small nit: I think the `&` on `versions` is unnecessary here: 

```suggestion
        if !versions.contains(&&rustls::version::TLS13) {
```

examples/src/bin/tlsserver-mio.rs
```diff
@@ -444,8 +444,9 @@ Options:
                         CRLFILE. May be used multiple times.
     --require-auth      Send a fatal alert if the client does not complete client
                         authentication.
-    --resumption        Support session resumption.
-    --tickets           Support tickets.
+    --no-resumption     Disable stateful session resumption.
+    --tickets           Support tickets (stateless resumption).
+    --max-early-data BYTES   Support receiving BYTES many bytes with 0RTT
```
cpu, line 449: Because these flags are reproduced in the README you'll need to do an update dance to fix [this CI job that's failing](https://github.com/rustls/rustls/actions/runs/10147456601/job/28059069867?pr=2059).

Try running `./admin/pull-readme` in the commit that changes these flags, and then add the updated `README.md` content to the commit.

cpu (commented):
Thanks! Looking good :-)

examples/src/bin/tlsserver-mio.rs
ctz: It actually will be `./admin/pull-usage` for this one.

cpu: Oops, you're right. Thanks!

Leonie-Theobald: Thanks for the tip. The README is updated with [f1c774d](https://github.com/rustls/rustls/pull/2059/commits/f1c774d34bce1d7d5ca1a8af64569b42160394fc)

examples/src/bin/tlsserver-mio.rs
Leonie-Theobald: I have fixed it with https://github.com/rustls/rustls/pull/2059/commits/356822ba424cb91e70ba354230e3a1eaedea1cbc.

examples/src/bin/tlsserver-mio.rs
Leonie-Theobald: This is updated now.

cpu (approved):
Thank you

ctz (approved):
Thanks! Please squash the fixup commits.

PR #2060, opened 2024-07-30
Avoid hanging if the client disconnects while self.wants_write()

lasiotus:
Similarly to how Ok(0) (zero) is treated during reads below, rustls should exit the loop instead of spinning indefinitely upon receiving Ok(0).

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -589,7 +589,13 @@ impl<Data> ConnectionCommon<Data> {
             }
 
             while self.wants_write() {
-                wrlen += self.write_tls(io)?;
+                match self.write_tls(io)? {
+                    0 => {
+                        io.flush()?;
+                        return Ok((rdlen, wrlen)); // EOF.
+                    }
+                    n => wrlen += n,
+                }
             }
             io.flush()?;
 
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -2357,6 +2357,62 @@ fn server_complete_io_for_write() {
 }
 
 #[test]
+fn server_complete_io_for_write_eof() {
+    for kt in ALL_KEY_TYPES {
+        let (mut client, mut server) = make_pair(*kt);
+
+        do_handshake(&mut client, &mut server);
+
+        // Queue 20 bytes to write.
+        server
+            .writer()
+            .write_all(b"01234567890123456789")
+            .unwrap();
+        {
+            const BYTES_BEFORE_EOF: usize = 5;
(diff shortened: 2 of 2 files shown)


lasiotus:
> I think this deserves a test.

Will do, thanks! (Maybe tomorrow).

ctz (approved):
Thanks!

rustls/src/conn.rs
```diff
@@ -588,7 +588,14 @@ impl<Data> ConnectionCommon<Data> {
             }
 
             while self.wants_write() {
-                wrlen += self.write_tls(io)?;
+                match self.write_tls(io)? {
```
cpu, line 591: Commenting here because it isn't possible to leave feedback on the commit message directly:

There's a typo in your commit title: "Avoid hagning if the client disconnects" -> "Avoid *hanging* if the client disconnects"

cpu (approved):
Would you mind squashing your two commits & fixing the commit message typo? Thanks!

rustls/tests/api.rs
```diff
@@ -2349,6 +2349,62 @@ fn server_complete_io_for_write() {
...
+        do_handshake(&mut client, &mut server);
+
+        // Queue 20 bytes to write.
+        server
+            .writer()
+            .write_all(b"01234567890123456789")
+            .unwrap();
+        {
+            const BYTES_BEFORE_EOF: usize = 5;
+            let mut eof_writer: EofWriter<BYTES_BEFORE_EOF> = Default::default();
```
djc, line 2389: Suggest spelling this as `let mut eof_writer = EofWriter::<BYTES_BEFORE_EOF>::default();`.

rustls/tests/api.rs
```diff
@@ -2349,6 +2349,62 @@ fn server_complete_io_for_write() {
     }
 }
 
+#[derive(Default)]
+struct EofWriter<const N: usize> {
+    written: usize,
+}
+
+impl<const N: usize> std::io::Write for EofWriter<N> {
```
cpu, line 2357: nit picking I think this `EofWriter` should also be below the test that uses it.

lasiotus: Done.

rustls/tests/api.rs
lasiotus: All done.

cpu (approved):
Thanks :-)

cpu:
> rustls / Build+test (stable, windows-latest) (pull_request) Failing after 3m
> ...
> cpu force-pushed the main branch from 02fb64b to c5817bc 

I think this CI err will fall away with a rebase on main so I went ahead and did that. Hope you don't mind :-)

PR #2062, opened 2024-07-31
Support for Raw Public Keys (RFC 7250)

holodorum:
This PR implements [RFC 7250](https://datatracker.ietf.org/doc/html/rfc7250) (Raw Public Key support) for TLS 1.3. The relevant discussion can be found in [this issue](https://github.com/rustls/rustls/issues/423), with a [detailed comment](https://github.com/rustls/rustls/issues/423#issuecomment-1908082743) by @ctz explaining the necessary steps.

### Main API changes:

1. The `ServerCertVerifier` and `ClientCertVerifier` traits now include a `requires_raw_public_keys()` function, while the `ServerResolveCert` and `ClientResolveCert` traits now have a `only_raw_public_keys()` function. Both functions have default implementations that return `false`.
2. We provide implementations for `ServerResolveCert` and `ClientResolveCert` that always resolve to a specific raw public key (`AlwaysResolvesServerRawPublicKeys`, `AlwaysResolvesClientRawPublicKeys`).
3. We introduce `webpki::verify::verify_tls13_signature_with_spki()` for signature verification. These functions use `webpki::RawPublicKeyEntity`, which was added to `rustls-webpki` in [this PR](https://github.com/rustls/webpki/pull/275).
4. Two error values were added to `PeerIncompatible`: 
   - `IncorrectCertificateTypeExtension`, raised when the local side expects Raw Public Keys but the peer sends an incorrect certificate type extension.
   - `UnsolicitedCertificateTypeExtension`, raised when the peer sends a certificate type extension, but the local side does not expect Raw Public Keys.
5. New values were added to `handshake::ServerExtension` and `handshake::ClientExtension`:
   - `ClientExtension::ServerCertTypes` and `ClientExtension::ClientCertTypes` are used to communicate the `CertificateType`s supported by the client.
   - `ServerExtension::ServerCertType` and `ServerExtension::ClientCertType` are used to communicate the `CertificateType` chosen by the server.
6. The two certificate type extensions, `ServerExtension::ServerCertType` and `ServerExtension::ClientCertType`, are added to `server_conn::ClientHello` for use in the `Acceptor` API.
7. We needed to decrypt, modify, and re-encrypt the `ServerHello` message to test our implementation. For this, `tls13::key_schedule::derive_traffic_key()` and `tls13::key_schedule::derive_traffic_iv()` were made public.

### Interoperability testing and example implementation:

1. An example implementation of a server and client using Raw Public Keys can be found in `raw_key_openssl_interop.rs`. This implementation is tested against an OpenSSL client and server to ensure interoperability with our implementation.

### Considerations:

1. To minimize the impact on the API and ensure backward compatibility, we sometimes use a `CertificateDer` type for a Raw Public Key instead of the more concise `SubjectPublicKeyInfoDer`. This approach is used in:
   - `CommonState::peer_certificates()`
   - The `cert` field of the `CertifiedKey` struct.
2. Our implementation closely follows RFC 7250. The main deviation is our assumption of a-priori knowledge about the peer’s authentication method when using Raw Public Keys (as explained in [this comment](https://github.com/rustls/rustls/issues/423#issuecomment-1908082743)). This means we never offer a `RawPublicKey` alongside `X509`, though the RFC permits it.

### Contributors:

This PR was created in collaboration with @aochagavia.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -504,6 +504,15 @@ jobs:
 
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@stable
+  
+      - name: Install pixi
+        run: | 
+          curl -fsSL https://pixi.sh/install.sh | bash
+          echo "$HOME/.pixi/bin" >> $GITHUB_PATH
+
+      - name: Install OpenSSL 3.2 with pixi
+        run: |
+          pixi global install openssl==3.2
 
       - name: openssl version
         run: openssl version
@@ -513,3 +522,4 @@ jobs:
         run: cargo test --locked -- --include-ignored
         env:
           RUST_BACKTRACE: 1
+          
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2179,6 +2179,7 @@ dependencies = [
  "once_cell",
  "openssl",
  "rustls 0.23.15",
+ "rustls-pki-types",
 ]
 
 [[package]]
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,3 +21,7 @@ We recommend new users start by looking at `simpleclient.rs` and `simpleserver.r
 * `simple_0rtt_server.rs` - shows how to make a TLS1.3 that accepts multiple connections and prints early 0RTT data.
 * `server_acceptor.rs` - shows how to use the `Acceptor` API to create a server that generates a unique `ServerConfig` for each client. This example also shows how to use client authentication, CRL revocation checking, and uses `rcgen` to generate its own certificates.
 * `unbuffered-server.rs` - shows an advanced example of using Rustls lower-level APIs to implement a server that does not buffer any data inside Rustls.
+
+## Client-Server examples
+
+* A client-server example using Raw Public Keys (RFC 7250) can be found in [`raw_key_openssl_interop`](../openssl-tests/src/raw_key_openssl_interop.rs).
--- a/openssl-tests/src/raw_key_openssl_interop.rs
+++ b/openssl-tests/src/raw_key_openssl_interop.rs
@@ -1,0 +1,522 @@
+//! This module provides tests for the interoperability of raw public keys with OpenSSL, and also
+//! demonstrates how to set up a client-server architecture that utilizes raw public keys.
+//!
+//! The module also includes example implementations of the `ServerCertVerifier` and `ClientCertVerifier` traits, using  
+//! pre-configured raw public keys for the verification of the peer.  
+
+mod client {
+    use std::io::{self, Read, Write};
+    use std::net::TcpStream;
+    use std::sync::Arc;
+
+    use rustls::client::danger::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
+    use rustls::client::AlwaysResolvesClientRawPublicKeys;
+    use rustls::crypto::{
+        aws_lc_rs as provider, verify_tls13_signature_with_raw_key, WebPkiSupportedAlgorithms,
+    };
+    use rustls::sign::CertifiedKey;
+    use rustls::version::TLS13;
+    use rustls::{
+        CertificateError, ClientConfig, ClientConnection, DigitallySignedStruct, Error,
+        InconsistentKeys, PeerIncompatible, SignatureScheme, Stream,
+    };
+    use rustls_pki_types::pem::PemObject;
+    use rustls_pki_types::{
+        CertificateDer, PrivateKeyDer, ServerName, SubjectPublicKeyInfoDer, UnixTime,
+    };
+
+    /// Build a `ClientConfig` with the given client private key and a server public key to trust.
+    pub(super) fn make_config(client_private_key: &str, server_pub_key: &str) -> ClientConfig {
+        let client_private_key = Arc::new(provider::default_provider())
+            .key_provider
+            .load_private_key(
+                PrivateKeyDer::from_pem_file(client_private_key)
+                    .expect("cannot open private key file"),
+            )
+            .expect("cannot load signing key");
+        let client_public_key = client_private_key
+            .public_key()
+            .ok_or(Error::InconsistentKeys(InconsistentKeys::Unknown))
+            .expect("cannot load public key");
+        let client_public_key_as_cert = CertificateDer::from(client_public_key.to_vec());
+
+        let server_raw_key = SubjectPublicKeyInfoDer::from_pem_file(server_pub_key)
+            .expect("cannot open pub key file");
+
+        let certified_key = Arc::new(CertifiedKey::new(
+            vec![client_public_key_as_cert],
+            client_private_key,
+        ));
+
+        ClientConfig::builder_with_protocol_versions(&[&TLS13])
+            .dangerous()
+            .with_custom_certificate_verifier(Arc::new(SimpleRpkServerCertVerifier::new(vec![
+                server_raw_key,
+            ])))
+            .with_client_cert_resolver(Arc::new(AlwaysResolvesClientRawPublicKeys::new(
+                certified_key,
+            )))
+    }
+
+    /// Run the client and connect to the server at the specified port.
+    ///
+    /// This client reads a message and then writes 'Hello from the client' to the server.
+    pub(super) fn run_client(config: ClientConfig, port: u16) -> Result<String, io::Error> {
+        let server_name = "0.0.0.0".try_into().unwrap();
+        let mut conn = ClientConnection::new(Arc::new(config), server_name).unwrap();
+        let mut sock = TcpStream::connect(format!("[::]:{}", port)).unwrap();
+        let mut tls = Stream::new(&mut conn, &mut sock);
+
+        let mut buf = vec![0; 128];
+        let len = tls.read(&mut buf).unwrap();
+        let received_message = String::from_utf8_lossy(&buf[..len]).to_string();
+
+        let bytes_written = tls
+            .write("Hello from the client".as_bytes())
+            .unwrap_or("".len());
+        assert!(bytes_written > 0);
+        Ok(received_message)
+    }
+
+    /// Verifies the tls handshake signature of the server,
+    /// and that the server's raw public key is in the list of trusted keys.
+    ///
+    /// Note: when the verifier is used for Raw Public Keys the `CertificateDer` argument to the functions contains the SPKI instead of a X509 Certificate
+    #[derive(Debug)]
+    struct SimpleRpkServerCertVerifier {
+        trusted_spki: Vec<SubjectPublicKeyInfoDer<'static>>,
+        supported_algs: WebPkiSupportedAlgorithms,
+    }
+
+    impl SimpleRpkServerCertVerifier {
+        fn new(trusted_spki: Vec<SubjectPublicKeyInfoDer<'static>>) -> Self {
+            SimpleRpkServerCertVerifier {
+                trusted_spki,
+                supported_algs: Arc::new(provider::default_provider())
+                    .clone()
+                    .signature_verification_algorithms,
+            }
+        }
+    }
+
+    impl ServerCertVerifier for SimpleRpkServerCertVerifier {
+        fn verify_server_cert(
+            &self,
+            end_entity: &CertificateDer<'_>,
+            _intermediates: &[CertificateDer<'_>],
+            _server_name: &ServerName<'_>,
+            _ocsp_response: &[u8],
+            _now: UnixTime,
+        ) -> Result<ServerCertVerified, rustls::Error> {
+            let end_entity_as_spki = SubjectPublicKeyInfoDer::from(end_entity.as_ref());
+            match self
+                .trusted_spki
+                .contains(&end_entity_as_spki)
+            {
+                false => Err(rustls::Error::InvalidCertificate(
+                    CertificateError::UnknownIssuer,
+                )),
+                true => Ok(ServerCertVerified::assertion()),
+            }
+        }
+
+        fn verify_tls12_signature(
+            &self,
+            _message: &[u8],
+            _cert: &CertificateDer<'_>,
+            _dss: &DigitallySignedStruct,
+        ) -> Result<HandshakeSignatureValid, rustls::Error> {
+            Err(rustls::Error::PeerIncompatible(
+                PeerIncompatible::Tls12NotOffered,
+            ))
+        }
+
+        fn verify_tls13_signature(
+            &self,
+            message: &[u8],
+            cert: &CertificateDer<'_>,
+            dss: &DigitallySignedStruct,
+        ) -> Result<HandshakeSignatureValid, rustls::Error> {
+            verify_tls13_signature_with_raw_key(
+                message,
+                &SubjectPublicKeyInfoDer::from(cert.as_ref()),
+                dss,
+                &self.supported_algs,
+            )
+        }
+
+        fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
+            self.supported_algs.supported_schemes()
+        }
+
+        fn requires_raw_public_keys(&self) -> bool {
+            true
+        }
+    }
+}
+
+mod server {
+    use std::io::{self, ErrorKind, Read, Write};
+    use std::{net::TcpListener, sync::Arc};
+
+    use rustls::client::danger::HandshakeSignatureValid;
+    use rustls::crypto::aws_lc_rs as provider;
+    use rustls::crypto::verify_tls13_signature_with_raw_key;
+    use rustls::crypto::WebPkiSupportedAlgorithms;
+    use rustls::server::danger::{ClientCertVerified, ClientCertVerifier};
+    use rustls::server::AlwaysResolvesServerRawPublicKeys;
+    use rustls::sign::CertifiedKey;
+    use rustls::version::TLS13;
+    use rustls::Error;
+    use rustls::InconsistentKeys;
+    use rustls::{
+        CertificateError, DigitallySignedStruct, DistinguishedName, PeerIncompatible, ServerConfig,
+        ServerConnection, SignatureScheme,
+    };
+    use rustls_pki_types::pem::PemObject;
+    use rustls_pki_types::PrivateKeyDer;
+    use rustls_pki_types::{CertificateDer, SubjectPublicKeyInfoDer, UnixTime};
+
+    /// Build a `ServerConfig` with the given server private key and a client public key to trust.
+    pub(super) fn make_config(server_private_key: &str, client_pub_key: &str) -> ServerConfig {
+        let client_raw_key = SubjectPublicKeyInfoDer::from_pem_file(client_pub_key)
+            .expect("cannot open pub key file");
+
+        let server_private_key = provider::default_provider()
+            .key_provider
+            .load_private_key(
+                PrivateKeyDer::from_pem_file(server_private_key)
+                    .expect("cannot open private key file"),
+            )
+            .expect("cannot load signing key");
+        let server_public_key = server_private_key
+            .public_key()
+            .ok_or(Error::InconsistentKeys(InconsistentKeys::Unknown))
+            .expect("cannot load public key");
+        let server_public_key_as_cert = CertificateDer::from(server_public_key.to_vec());
+
+        let certified_key = Arc::new(CertifiedKey::new(
+            vec![server_public_key_as_cert],
+            server_private_key,
+        ));
+
+        let client_cert_verifier = Arc::new(SimpleRpkClientCertVerifier::new(vec![client_raw_key]));
+        let server_cert_resolver = Arc::new(AlwaysResolvesServerRawPublicKeys::new(certified_key));
+
+        ServerConfig::builder_with_protocol_versions(&[&TLS13])
+            .with_client_cert_verifier(client_cert_verifier)
+            .with_cert_resolver(server_cert_resolver)
+    }
+
+    /// Run the server at the specified port and accept a connection from the client.
+    ///
+    /// After the handshake is complete, the server writes 'Hello from the server' to the client.
+    /// The server then waits until reads it receives a message from the client and closes the connection.
+    pub(super) fn run_server(
+        config: ServerConfig,
+        listener: TcpListener,
+    ) -> Result<String, io::Error> {
+        let (mut stream, _) = listener.accept()?;
+
+        let mut conn = ServerConnection::new(Arc::new(config)).unwrap();
+        conn.complete_io(&mut stream)?;
+
+        conn.writer()
+            .write_all(b"Hello from the server")?;
+        conn.complete_io(&mut stream)?;
+
+        let mut buf = [0; 128];
+
+        loop {
+            match conn.reader().read(&mut buf) {
+                Ok(len) => {
+                    conn.send_close_notify();
+                    conn.complete_io(&mut stream)?;
+                    return Ok(String::from_utf8_lossy(&buf[..len]).to_string());
+                }
+                Err(err) if err.kind() == ErrorKind::WouldBlock => {
+                    conn.read_tls(&mut stream)?;
+                    conn.process_new_packets().unwrap();
+                }
+                Err(err) => {
+                    return Err(err);
+                }
+            };
+        }
+    }
+
+    /// Verifies the tls handshake signature of the client,
+    /// and that the client's raw public key is in the list of trusted keys.
+    ///
+    /// Note: when the verifier is used for Raw Public Keys the `CertificateDer` argument to the functions contains the SPKI instead of a X509 Certificate
+    #[derive(Debug)]
+    struct SimpleRpkClientCertVerifier {
+        trusted_spki: Vec<SubjectPublicKeyInfoDer<'static>>,
+        supported_algs: WebPkiSupportedAlgorithms,
+    }
+
+    impl SimpleRpkClientCertVerifier {
+        pub fn new(trusted_spki: Vec<SubjectPublicKeyInfoDer<'static>>) -> Self {
+            Self {
+                trusted_spki,
+                supported_algs: Arc::new(provider::default_provider())
+                    .clone()
+                    .signature_verification_algorithms,
+            }
+        }
+    }
+
+    impl ClientCertVerifier for SimpleRpkClientCertVerifier {
+        fn root_hint_subjects(&self) -> &[DistinguishedName] {
+            &[]
+        }
+
+        fn verify_client_cert(
+            &self,
+            end_entity: &CertificateDer<'_>,
+            _intermediates: &[CertificateDer<'_>],
+            _now: UnixTime,
+        ) -> Result<ClientCertVerified, rustls::Error> {
+            let end_entity_as_spki = SubjectPublicKeyInfoDer::from(end_entity.as_ref());
+            match self
+                .trusted_spki
+                .contains(&end_entity_as_spki)
+            {
+                false => Err(rustls::Error::InvalidCertificate(
+                    CertificateError::UnknownIssuer,
+                )),
+                true => Ok(ClientCertVerified::assertion()),
+            }
+        }
+
+        fn verify_tls12_signature(
+            &self,
+            _message: &[u8],
+            _cert: &CertificateDer<'_>,
+            _dss: &DigitallySignedStruct,
+        ) -> Result<HandshakeSignatureValid, rustls::Error> {
+            Err(rustls::Error::PeerIncompatible(
+                PeerIncompatible::Tls12NotOffered,
+            ))
+        }
+
+        fn verify_tls13_signature(
+            &self,
+            message: &[u8],
+            cert: &CertificateDer<'_>,
+            dss: &DigitallySignedStruct,
+        ) -> Result<HandshakeSignatureValid, rustls::Error> {
+            verify_tls13_signature_with_raw_key(
+                message,
+                &SubjectPublicKeyInfoDer::from(cert.as_ref()),
+                dss,
+                &self.supported_algs,
+            )
+        }
+
+        fn supported_verify_schemes(&self) -> Vec<SignatureScheme> {
+            self.supported_algs.supported_schemes()
+        }
+
+        fn requires_raw_public_keys(&self) -> bool {
+            true
+        }
+    }
+}
+
+mod tests {
+    use std::io::{BufRead, BufReader, Read, Write};
+    use std::net::TcpListener;
+    use std::process::{Command, Stdio};
+    use std::sync::mpsc::channel;
+    use std::thread;
+
+    use super::{client, server};
+    use crate::utils::verify_openssl3_available;
+
+    const SERVER_PRIV_KEY_FILE: &str = "../test-ca/ecdsa-p256/end.key";
+    const SERVER_PUB_KEY_FILE: &str = "../test-ca/ecdsa-p256/end.spki.pem";
+    const SERVER_CERT_KEY_FILE: &str = "../test-ca/ecdsa-p256/end.cert";
+    const CLIENT_PUB_KEY_FILE: &str = "../test-ca/ecdsa-p256/client.spki.pem";
+    const CLIENT_PRIV_KEY_FILE: &str = "../test-ca/ecdsa-p256/client.key";
+    const CLIENT_CERT_KEY_FILE: &str = "../test-ca/ecdsa-p256/client.cert";
+
+    fn tcp_listener() -> TcpListener {
+        TcpListener::bind("[::]:0").expect("Could not bind to random port")
+    }
+
+    #[test]
+    fn test_rust_server_and_rust_client() {
+        let listener = tcp_listener();
+        let port = listener.local_addr().unwrap().port();
+
+        let (sender, receiver) = channel();
+        let server_thread = thread::spawn(move || {
+            sender
+                .send(server::run_server(
+                    server::make_config(SERVER_PRIV_KEY_FILE, CLIENT_PUB_KEY_FILE),
+                    listener,
+                ))
+                .unwrap();
+        });
+
+        // Start the Rust client
+        let client_config = client::make_config(CLIENT_PRIV_KEY_FILE, SERVER_PUB_KEY_FILE);
+        match client::run_client(client_config, port) {
+            Ok(server_message) => {
+                assert_eq!(server_message, "Hello from the server");
+            }
+            Err(e) => {
+                panic!("Client failed to communicate with the server: {:?}", e);
+            }
+        }
+
+        // Wait for the server to finish and clean up the thread
+        let server_result = receiver.recv().unwrap();
+        server_thread
+            .join()
+            .expect("Failed to join server thread");
+
+        match server_result {
+            Ok(client_message) => {
+                assert_eq!(client_message, "Hello from the client");
+            }
+            Err(e) => {
+                panic!("Server failed to communicate with the client: {:?}", e);
+            }
+        }
+    }
+
+    #[test]
+    fn test_rust_server_with_openssl_client() {
+        verify_openssl3_available();
+
+        let listener = tcp_listener();
+        let port = listener.local_addr().unwrap().port();
+
+        let server_thread = thread::spawn(move || {
+            server::run_server(
+                server::make_config(SERVER_PRIV_KEY_FILE, CLIENT_PUB_KEY_FILE),
+                listener,
+            )
+            .expect("failed to run server to completion")
+        });
+
+        // Start the OpenSSL client
+        let mut openssl_client = Command::new("openssl")
+            .arg("s_client")
+            .arg("-connect")
+            .arg(format!("[::]:{:?}", port))
+            .arg("-enable_server_rpk")
+            .arg("-enable_client_rpk")
+            .arg("-key")
+            .arg(CLIENT_PRIV_KEY_FILE)
+            .arg("-cert")
+            .arg(CLIENT_CERT_KEY_FILE)
+            .arg("-tls1_3")
+            .arg("-debug")
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .spawn()
+            .expect("Failed to execute OpenSSL client");
+
+        let mut stdin = openssl_client.stdin.take().unwrap();
+        let mut stdout = openssl_client.stdout.take().unwrap();
+        let mut stdout_buf = [0; 1024];
+        let mut openssl_stdout = String::new();
+        let mut received_server_msg = false;
+        loop {
+            match stdout.read(&mut stdout_buf) {
+                Ok(0) => break,
+                Ok(len) => {
+                    let read = &stdout_buf[..len];
+
+                    std::io::stdout()
+                        .write_all(read)
+                        .unwrap();
+                    openssl_stdout.push_str(&String::from_utf8_lossy(read));
+                    if openssl_stdout.contains("Hello from the server") {
+                        received_server_msg = true;
+                        stdin
+                            .write_all(b"Hello, from openssl client!")
+                            .expect("Failed to write to stdin");
+                        break;
+                    }
+                }
+                Err(e) => panic!("Error reading from OpenSSL stdin: {e:?}"),
+            }
+        }
+
+        assert!(received_server_msg);
+        assert_eq!(server_thread.join().unwrap(), "Hello, from openssl client!");
+    }
+
+    #[test]
+    fn test_rust_client_with_openssl_server() {
+        verify_openssl3_available();
+
+        let listener = tcp_listener();
+        let port = listener.local_addr().unwrap().port();
+        drop(listener);
+
+        // Start OpenSSL server
+        let mut server_process = Command::new("openssl")
+            .arg("s_server")
+            .arg("-port")
+            .arg(port.to_string())
+            .arg("-cert")
+            .arg(SERVER_CERT_KEY_FILE)
+            .arg("-key")
+            .arg(SERVER_PRIV_KEY_FILE)
+            .arg("-verify")
+            .arg("1")
+            .arg("-enable_server_rpk")
+            .arg("-enable_client_rpk")
+            .arg("-tls1_3")
+            .arg("-debug")
+            .stdout(Stdio::piped())
+            .stdin(Stdio::piped())
+            .spawn()
+            .expect("Failed to start OpenSSL server");
+
+        // Read from the OpenSSL server's stdout and wait for "ACCEPT"
+        if let Some(stdout) = server_process.stdout.take() {
+            let stdout_reader = BufReader::new(stdout);
+            for line in stdout_reader.lines() {
+                match line {
+                    Ok(l) => {
+                        if l.contains("ACCEPT") {
+                            break;
+                        }
+                    }
+                    Err(e) => {
+                        panic!("Error reading from OpenSSL stdout: {:?}", e);
+                    }
+                }
+            }
+        }
+
+        // Write a message to the OpenSSL server's stdin
+        if let Some(mut stdin) = server_process.stdin.take() {
+            stdin
+                .write_all(b"Hello, from openssl server!")
+                .expect("Failed to write to stdin");
+        }
+
+        // Create the Rust client config and run the client
+        let client_config = client::make_config(CLIENT_PRIV_KEY_FILE, SERVER_PUB_KEY_FILE);
+        match client::run_client(client_config, port) {
+            Ok(server_message) => {
+                assert_eq!(server_message, "Hello, from openssl server!");
+            }
+            Err(_) => {
+                unreachable!("Client failed to communicate with the server");
+            }
+        }
+
+        // Ensure the OpenSSL server process is terminated
+        server_process
+            .kill()
+            .expect("Failed to kill OpenSSL server process");
+    }
+}
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -241,6 +241,41 @@ impl client::ResolvesClientCert for AlwaysResolvesClientCert {
     }
 }
 
+/// An exemplar `ResolvesClientCert` implementation that always resolves to a single
+/// [RFC 7250] raw public key.  
+///
+/// [RFC 7250]: https://tools.ietf.org/html/rfc7250  
+#[derive(Clone, Debug)]
+pub struct AlwaysResolvesClientRawPublicKeys(Arc<sign::CertifiedKey>);
+impl AlwaysResolvesClientRawPublicKeys {
+    /// Create a new `AlwaysResolvesClientRawPublicKeys` instance.
+    pub fn new(certified_key: Arc<sign::CertifiedKey>) -> Self {
+        Self(certified_key)
+    }
+}
+
+impl client::ResolvesClientCert for AlwaysResolvesClientRawPublicKeys {
+    fn resolve(
+        &self,
+        _root_hint_subjects: &[&[u8]],
+        _sigschemes: &[SignatureScheme],
+    ) -> Option<Arc<sign::CertifiedKey>> {
+        Some(Arc::clone(&self.0))
+    }
+
+    fn only_raw_public_keys(&self) -> bool {
+        true
+    }
+
+    /// Returns true if the resolver is ready to present an identity.
+    ///
+    /// Even though the function is called `has_certs`, it returns true  
+    /// although only an RPK (Raw Public Key) is available, not an actual certificate.  
+    fn has_certs(&self) -> bool {
+        true
+    }
+}
+
 test_for_each_provider! {
     use std::prelude::v1::*;
     use alloc::sync::Arc;
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -17,7 +17,9 @@ use crate::client::client_conn::ClientConnectionData;
 use crate::client::common::ClientHelloDetails;
 use crate::client::ech::EchState;
 use crate::client::{tls13, ClientConfig, EchMode, EchStatus};
-use crate::common_state::{CommonState, HandshakeKind, KxState, State};
+use crate::common_state::{
+    CommonState, HandshakeKind, KxState, RawKeyNegotationResult, RawKeyNegotiationParams, State,
+};
 use crate::conn::ConnectionRandoms;
 use crate::crypto::{ActiveKeyExchange, KeyExchangeAlgorithm};
 use crate::enums::{AlertDescription, CipherSuite, ContentType, HandshakeType, ProtocolVersion};
@@ -25,7 +27,9 @@ use crate::error::{Error, PeerIncompatible, PeerMisbehaved};
 use crate::hash_hs::HandshakeHashBuffer;
 use crate::log::{debug, trace};
 use crate::msgs::base::Payload;
-use crate::msgs::enums::{Compression, ECPointFormat, ExtensionType, PSKKeyExchangeMode};
+use crate::msgs::enums::{
+    CertificateType, Compression, ECPointFormat, ExtensionType, PSKKeyExchangeMode,
+};
 use crate::msgs::handshake::{
     CertificateStatusRequest, ClientExtension, ClientHelloPayload, ClientSessionTicket,
     ConvertProtocolNameList, HandshakeMessagePayload, HandshakePayload, HasServerExtensions,
@@ -339,6 +343,24 @@ fn emit_client_hello_for_retry(
         false
     };
 
+    if config
+        .client_auth_cert_resolver
+        .only_raw_public_keys()
+    {
+        exts.push(ClientExtension::ClientCertTypes(vec![
+            CertificateType::RawPublicKey,
+        ]));
+    }
+
+    if config
+        .verifier
+        .requires_raw_public_keys()
+    {
+        exts.push(ClientExtension::ServerCertTypes(vec![
+            CertificateType::RawPublicKey,
+        ]));
+    }
+
     // Extra extensions must be placed before the PSK extension
     exts.extend(extra_exts.iter().cloned());
 
@@ -648,6 +670,52 @@ pub(super) fn process_alpn_protocol(
             .map(|v| bs_debug::BsDebug(v))
     );
     Ok(())
+}
+
+pub(super) fn process_server_cert_type_extension(
+    common: &mut CommonState,
+    config: &ClientConfig,
+    server_cert_extension: Option<&CertificateType>,
+) -> Result<(), Error> {
+    let requires_server_rpk = config
+        .verifier
+        .requires_raw_public_keys();
+    let server_offers_rpk = matches!(server_cert_extension, Some(CertificateType::RawPublicKey));
+
+    let raw_key_negotation_params = RawKeyNegotiationParams {
+        peer_supports_raw_key: server_offers_rpk,
+        local_expects_raw_key: requires_server_rpk,
+        extension_type: ExtensionType::ServerCertificateType,
+    };
+    match raw_key_negotation_params.validate_raw_key_negotiation() {
+        RawKeyNegotationResult::Err(err) => {
+            Err(common.send_fatal_alert(AlertDescription::HandshakeFailure, err))
+        }
+        _ => Ok(()),
+    }
+}
+
+pub(super) fn process_client_cert_type_extension(
+    common: &mut CommonState,
+    config: &ClientConfig,
+    client_cert_extension: Option<&CertificateType>,
+) -> Result<(), Error> {
+    let requires_client_rpk = config
+        .client_auth_cert_resolver
+        .only_raw_public_keys();
+    let server_allows_rpk = matches!(client_cert_extension, Some(CertificateType::RawPublicKey));
+
+    let raw_key_negotation_params = RawKeyNegotiationParams {
+        peer_supports_raw_key: server_allows_rpk,
+        local_expects_raw_key: requires_client_rpk,
+        extension_type: ExtensionType::ClientCertificateType,
+    };
+    match raw_key_negotation_params.validate_raw_key_negotiation() {
+        RawKeyNegotationResult::Err(err) => {
+            Err(common.send_fatal_alert(AlertDescription::HandshakeFailure, err))
+        }
+        _ => Ok(()),
+    }
 }
 
 impl State<ClientConnectionData> for ExpectServerHello {
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -11,7 +11,7 @@ use crate::log::{debug, error, warn};
 use crate::msgs::alert::AlertMessagePayload;
 use crate::msgs::base::Payload;
 use crate::msgs::codec::Codec;
-use crate::msgs::enums::{AlertLevel, KeyUpdateRequest};
+use crate::msgs::enums::{AlertLevel, ExtensionType, KeyUpdateRequest};
 use crate::msgs::fragmenter::MessageFragmenter;
 use crate::msgs::handshake::{CertificateChain, HandshakeMessagePayload};
 use crate::msgs::message::{
@@ -24,7 +24,7 @@ use crate::suites::{PartiallyExtractedSecrets, SupportedCipherSuite};
 use crate::tls12::ConnectionSecrets;
 use crate::unbuffered::{EncryptError, InsufficientSizeError};
 use crate::vecbuf::ChunkVecBuffer;
-use crate::{quic, record_layer};
+use crate::{quic, record_layer, PeerIncompatible};
 
 /// Connection state common to both client and server connections.
 pub struct CommonState {
@@ -107,21 +107,27 @@ impl CommonState {
         !(self.may_send_application_data && self.may_receive_application_data)
     }
 
-    /// Retrieves the certificate chain used by the peer to authenticate.
+    /// Retrieves the certificate chain or the raw public key used by the peer to authenticate.
     ///
     /// The order of the certificate chain is as it appears in the TLS
     /// protocol: the first certificate relates to the peer, the
     /// second certifies the first, the third certifies the second, and
     /// so on.
     ///
+    /// When using raw public keys, the first and only element is the raw public key.
+    ///
     /// This is made available for both full and resumed handshakes.
     ///
-    /// For clients, this is the certificate chain of the server.
+    /// For clients, this is the certificate chain or the raw public key of the server.
     ///
-    /// For servers, this is the certificate chain of the client,
+    /// For servers, this is the certificate chain or the raw public key of the client,
     /// if client authentication was completed.
     ///
     /// The return value is None until this value is available.
+    ///
+    /// Note: the return type of the 'certificate', when using raw public keys is `CertificateDer<'static>`
+    /// even though this should technically be a `SubjectPublicKeyInfoDer<'static>`.
+    /// This choice simplifies the API and ensures backwards compatibility.
     pub fn peer_certificates(&self) -> Option<&[CertificateDer<'static>]> {
         self.peer_certificates.as_deref()
     }
@@ -885,6 +891,35 @@ enum Limit {
     #[cfg(feature = "std")]
     Yes,
     No,
+}
+
+#[derive(Debug)]
+pub(super) struct RawKeyNegotiationParams {
+    pub(super) peer_supports_raw_key: bool,
+    pub(super) local_expects_raw_key: bool,
+    pub(super) extension_type: ExtensionType,
+}
+
+impl RawKeyNegotiationParams {
+    pub(super) fn validate_raw_key_negotiation(&self) -> RawKeyNegotationResult {
+        match (self.local_expects_raw_key, self.peer_supports_raw_key) {
+            (true, true) => RawKeyNegotationResult::Negotiated(self.extension_type),
+            (false, false) => RawKeyNegotationResult::NotNegotiated,
+            (true, false) => RawKeyNegotationResult::Err(Error::PeerIncompatible(
+                PeerIncompatible::IncorrectCertificateTypeExtension,
+            )),
+            (false, true) => RawKeyNegotationResult::Err(Error::PeerIncompatible(
+                PeerIncompatible::UnsolicitedCertificateTypeExtension,
+            )),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) enum RawKeyNegotationResult {
+    Negotiated(ExtensionType),
+    NotNegotiated,
+    Err(Error),
 }
 
 /// Tracking technically-allowed protocol actions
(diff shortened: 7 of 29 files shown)


rustls/examples/internal/test_ca.rs
```diff
@@ -8,6 +8,7 @@ use std::str::FromStr;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
+use pki_types::SubjectPublicKeyInfoDer;
```
ctz: I expect this commit is pretty uncontroversial -- let's split it off as a separate PR and merge that first. That should reduce the noise in this one.

rustls/src/client/builder.rs
```diff
@@ -163,6 +162,35 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
...
+    ///
+    /// `cert_chain` is a vector of a DER-encoded RPK certificate.
+    /// `key_der` is a DER-encoded private key as PKCS#1, PKCS#8, or SEC1. The
+    /// `aws-lc-rs` and `ring` [`CryptoProvider`]s support all three encodings,
+    /// but other `CryptoProviders` may not.
+    ///
+    /// This function fails if `key_der` is invalid, or if the public_key cannot be parsed to a RPK cert
+    pub fn with_client_auth_rpk(
+        self,
+        public_key_der: SubjectPublicKeyInfoDer<'static>,
```
ctz, line 176: I would suggest removing this argument, and using `private_key.public_key()` below. That should remove a category of accidental errors in this API (mismatched private/public key).

ctz: This would mean this API is only usable for `SigningKey` impls that can produce the corresponding public key. This covers all of them in the base crate, but it's worth mentioning this detail in the doc for this function if we go this route.

rustls/examples/internal/test_ca.rs
holodorum: Great idea! Removed this commit from this PR and created a https://github.com/rustls/rustls/pull/2078. When that PR get's merged I'll do a rebase.

rustls/src/client/builder.rs
holodorum: I do like that approach, seems to be more robust. In that case we could remove `public_key_der` from `client::builder::with_client_auth_rpk()`, `server::builder::with_rpk()`, `AlwaysResolvesClientRawPublicKeys::new()` and `AlwaysResolvesServerRawPublicKeys::new()`.
Currently the `with_client_auth_rpk()/with_rpk()` functions verify whether or not the `public_key_der` can be read into a `RawPublicKeyEntity`. Now it's probably better to verify that `SigningKey::public_key()` does not return `None` right? 
Removing `public_key_der` from these functions might also make `get_client_spki()/client_spki()` in `tests::common::mod` redundant. If we would remove these I doubt there is much use for https://github.com/rustls/rustls/pull/2078.

holodorum: Thanks for the good suggestion, looks better now! As of now I feel that the changes made in  https://github.com/rustls/rustls/pull/2078 might not be necessary. But maybe you foresee testing scenarios where it is useful to have RPK pemfiles.

ctz:
> 1. The reason being that the openssl server does not return the `client_cert_type` extension, but instead sends a RawPublicKey directly. Is this common behaviour, and if so would it be better to do the validation of the `client_cert_type` and `server_cert_type` extensions somewhere else?

There is no reason for openssl to send `client_cert_type` as it is not going to ask for client authentication. If I add `-verify 1` then it does send it (though it fails later in the handshake with a decode error, not sure if that is something i've done wrong locally). That means, I think, that the call to `process_client_cert_type_extension` must happen later, during `tls12::ExpectCertificateRequest`.

Would be good to automate the running of rpk_example.rs against openssl in `openssl-tests/`.

examples/src/bin/rpk_example.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    Ok(key.unwrap())
+}
+
+fn create_rpk_client_config(
+    client_private_key: PrivateKeyDer<'static>,
+    server_spki_certs: Vec<SubjectPublicKeyInfoDer<'static>>,
+) -> ClientConfig {
+    let server_cert_verifier = Arc::new(SimpleRpkServerCertVerifier::new(server_spki_certs));
+    ClientConfig::builder_with_protocol_versions(&[&TLS12])
```
ctz, line 47: caution that hardcoding TLS1.2 here would prevent testing of TLS1.3 code

rustls/src/client/handy.rs
```diff
@@ -241,6 +242,44 @@ impl client::ResolvesClientCert for AlwaysResolvesClientCert {
...
+        Some(Arc::clone(&self.0))
+    }
+
+    fn only_raw_public_keys(&self) -> bool {
+        true
+    }
+
+    fn has_certs(&self) -> bool {
+        true
+    }
```
cpu, line 280: This feels a bit odd to me. There's not really a certificate available right?

Renaming the `fn` to something more about the resolver being ready to present an identity seems appealing but would be a breaking change. Maybe a compromise would be a comment here explaining some of the nuance?

rustls/src/client/hs.rs
```diff
@@ -645,6 +665,48 @@ pub(super) fn process_alpn_protocol(
...
+    common: &mut CommonState,
+    config: &ClientConfig,
+    server_cert_extension: Option<&CertificateType>,
+) -> Result<(), Error> {
+    let requires_server_rpk = config
+        .verifier
+        .requires_raw_public_keys();
+    let server_offers_rpk = server_cert_extension.map_or(false, |cert_type| {
+        *cert_type == CertificateType::RawPublicKey
+    });
```
cpu, line 678: I think this could be simplified to:

```rust
    let server_offers_rpk = matches!(server_cert_extension, Some(CertificateType::RawPublicKey));
```

rustls/src/client/hs.rs
```diff
@@ -645,6 +665,48 @@ pub(super) fn process_alpn_protocol(
...
+    common: &mut CommonState,
+    config: &ClientConfig,
+    client_cert_extension: Option<&CertificateType>,
+) -> Result<(), Error> {
+    let requires_client_rpk = config
+        .client_auth_cert_resolver
+        .only_raw_public_keys();
+    let server_allows_rpk = client_cert_extension.map_or(false, |cert_type| {
+        *cert_type == CertificateType::RawPublicKey
+    });
```
cpu, line 699: similar opportunity for simplification here:

```rust
    let server_allows_rpk = matches!(client_cert_extension, Some(CertificateType::RawPublicKey));
```

rustls/src/client/hs.rs
```diff
@@ -720,6 +782,21 @@ impl State<ClientConnectionData> for ExpectServerHello {
             ));
         }
 
+        cx.common.negotiated_version = Some(version);
+        // Validate server and client cert type extension
+        if !cx.common.is_tls13() {
```
cpu, line 787: It seems a little bit odd that there's a branching here based on protocol version. Can you expand on this design choice?

rustls/src/common_state.rs
```diff
@@ -273,6 +273,27 @@ impl CommonState {
         Ok(self.write_fragments(outgoing_tls, fragments))
     }
 
+    pub(crate) fn validate_rpk_negotiation(
+        &mut self,
+        local_expects_rpk: bool,
+        peer_supports_rpk: bool,
```
cpu, line 279: I'm a bit nervous about this API signature that takes two `bool` values. It seems like it would be easy to mistakenly switch the order around. Perhaps a dedicated `enum` that covers both sides of this unambiguously would be in order?

rustls/src/msgs/enums.rs
```diff
@@ -305,6 +307,19 @@ enum_builder! {
     }
 }
 
+enum_builder! {
+    /// The `CertificateType` enum send in the cert_type extensions.
```
cpu, line 311: typo: "send in the" -> "sent in the" ?

rustls/src/msgs/handshake.rs
```diff
@@ -1308,7 +1348,8 @@ pub struct ServerHelloPayload {
     pub(crate) session_id: SessionId,
     pub(crate) cipher_suite: CipherSuite,
     pub(crate) compression_method: Compression,
-    pub(crate) extensions: Vec<ServerExtension>,
+
+    pub extensions: Vec<ServerExtension>,
```
cpu, line 1352: I think our ordering style guide would want this field promoted to the first entry in the `struct` now that it's `pub`.

rustls/src/server/builder.rs
```diff
@@ -88,6 +88,33 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
         Ok(self.with_cert_resolver(Arc::new(resolver)))
     }
 
+    /// Set a single raw public key certificate and matching private key. This
```
cpu, line 91: I think I touched on this in previous comments, but is it appropriate to call this a "raw public key certificate"? I think "certificate" is fairly well established to mean the X.509v3 structure from RFC5280. I think a raw public key is a different beast and we should try not to conflate it with a certificate.

rustls/src/server/builder.rs
```diff
@@ -88,6 +88,33 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
+    /// certificate and key is used for all subsequent connections,
+    /// irrespective of things like SNI hostname.
+    ///
+    /// `key_der` is a DER-encoded private key as PKCS#1, PKCS#8, or SEC1. The
+    /// `aws-lc-rs` and `ring` [`CryptoProvider`]s support all three encodings,
+    /// but other `CryptoProviders` may not.
+    ///
+    /// This function fails if `key_der` is invalid, or if the public_key cannot be retrieved from the private_key.
```
cpu, line 99: ditto re: "public_key" -> "public key" and "private_key" -> "private key".

rustls/src/server/handy.rs
```diff
@@ -1,6 +1,8 @@
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
+use pki_types::CertificateDer;
+use std::vec;
```
cpu, line 5: nit: the `pki_types` import should be on its own, separated by whitespace from the `alloc`/`core` imports above it and the `crate` imports below it.

I also think `std::vec` should probably be `alloc::vec` for no-std?

rustls/src/server/handy.rs
```diff
@@ -199,6 +201,38 @@ impl server::ResolvesServerCert for AlwaysResolvesChain {
...
+            private_key,
+        )))
+    }
+}
+
+impl server::ResolvesServerCert for AlwaysResolvesServerRawPublicKeys {
+    fn resolve(&self, _client_hello: ClientHello<'_>) -> Option<Arc<sign::CertifiedKey>> {
+        Some(Arc::clone(&self.0))
+    }
+    fn only_raw_public_keys(&self) -> bool {
```
cpu, line 231: nit: whitespace between these fns.

rustls/src/server/hs.rs
```diff
@@ -156,9 +156,58 @@ impl ExtensionProcessing {
...
             ocsp_response.take();
         }
+        let requires_server_rpk = config
+            .cert_resolver
+            .only_raw_public_keys();
+        let client_allows_rpk = hello
+            .server_certificate_extension()
+            .map_or(false, |cert_type| {
+                cert_type[0] == CertificateType::RawPublicKey
+            });
```
cpu, line 166: similar feedback as the client side, I think `matches!` might be nicer here:

```rust
        let client_allows_rpk = matches!(hello.server_certificate_extension(), Some([CertificateType::RawPublicKey]));
```

rustls/src/server/hs.rs
```diff
@@ -156,9 +156,58 @@ impl ExtensionProcessing {
...
+            RpkNegotiationResult::NotNegotiated => {}
+        }
+        let requires_client_rpk = config
+            .verifier
+            .requires_raw_public_keys();
+        let client_offers_rpk = hello
+            .client_certificate_extension()
+            .map_or(false, |cert_type| {
+                cert_type[0] == CertificateType::RawPublicKey
+            });
```
cpu, line 191: same feedback here w.r.t `matches!`

rustls/src/server/server_conn.rs
```diff
@@ -147,6 +157,8 @@ impl<'a> ClientHello<'a> {
             server_name,
             signature_schemes,
             alpn,
+            server_cert_type,
+            client_cert_type,
```
cpu, line 161: Should these new fields also get a `trace!` line above, alongside the `trace!` for sni, sig schemes, etc?

rustls/src/server/server_conn.rs
```diff
@@ -136,6 +144,8 @@ impl<'a> ClientHello<'a> {
         server_name: &'a Option<DnsName<'_>>,
         signature_schemes: &'a [SignatureScheme],
         alpn: Option<&'a Vec<ProtocolName>>,
+        server_cert_type: Option<&'a [CertificateType]>,
+        client_cert_type: Option<&'a [CertificateType]>,
```
cpu, line 148: Is there perhaps a better way to represent this so that the server/client types can't be swapped accidentally?

rustls/tests/api.rs
```diff
@@ -47,6 +47,216 @@ use provider::cipher_suite;
 use provider::sign::RsaSigningKey;
 use rustls::ProtocolVersion::TLSv1_2;
 
+mod test_rpk {
+    use super::*;
+
+    #[test]
+    fn test_certificate_type_extension_errors() {
+        let kt = KeyType::Rsa2048;
+        //TODO this does not work for TLS13 yet because we don't encrypt the extensions and get the following error: Err(PeerMisbehaved(UnexpectedCleartextExtension))
```
cpu, line 56: style nit: I think the comments in this file should be written `//<space><content>` instead of `//<content>`

rustls/tests/common/mod.rs
```diff
@@ -523,6 +541,31 @@ pub fn make_server_config_with_client_verifier(
         .unwrap()
 }
 
+pub fn make_server_config_with_rpk_support_and_version(
+    kt: KeyType,
+    version: &[&'static rustls::SupportedProtocolVersion],
+) -> ServerConfig {
+    let mut client_verifier = MockClientVerifier::new(|| Ok(ClientCertVerified::assertion()), kt);
+    client_verifier.expect_raw_public_keys = true;
+    let client_verifier = Arc::new(client_verifier);
```
cpu, line 550: nit: this last binding doesn't add much value. Suggest using `with_client_cert_verifier(client_verifier.into())` or moving the `Arc::new()` inline with `with_client_cert_verifier()`.

rustls/tests/common/mod.rs
```diff
@@ -972,8 +1027,93 @@ impl Default for MockServerVerifier {
...
             ],
             expected_ocsp_response: None,
+            requires_raw_public_keys: false,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct MockClientVerifier {
+    parent: Arc<dyn ClientCertVerifier>,
```
cpu, line 1037: nit: private field after the pub fields.

examples/src/bin/rpk_example.rs
```diff
@@ -0,0 +1,390 @@
...
+use std::time::Duration;
+use std::{
+    env,
+    fs::File,
+    io::{BufReader, Read, Write},
+    net::{TcpListener, TcpStream},
+    result::Result,
+    sync::Arc,
+};
+use std::{io, thread};
```
cpu, line 23: I think these imports need to be updated to match our style guide, both in terms of ordering and also in terms of how items are grouped with parent modules.

examples/src/bin/rpk_example.rs
```diff
@@ -0,0 +1,390 @@
+use log::trace;
```
cpu, line 1: Can you add a `//!` comment at the top of this file that roughly matches the way the other examples introduce themselves?

I think `examples/README.md` could also use an update.

examples/src/bin/rpk_example.rs
```diff
@@ -0,0 +1,390 @@
...
+        );
+        self.supported_algs.supported_schemes()
+    }
+
+    fn requires_raw_public_keys(&self) -> bool {
+        true
+    }
+}
+
+fn main() -> Result<(), io::Error> {
```
cpu, line 304: I think this needs to be rearranged into a "top down" style where `main` is up top and all of the supporting functions/types are below it.

examples/src/bin/rpk_example.rs
cpu, line 1: It would also be nice if the `.github/workflows/daily-tests.yml` could be updated to include some coverage in the `example-tests` job for this new example (assuming its not too painful).

cpu (commented):
Here's some comments/suggestions from a rough first pass.

Beyond the mechanics of the implementation I'm also interested in how this interacts with revocation. This is one of the aspects of raw public keys that I think deserves explicit attention in our documentation. As I understand it there _isn't_ any mechanism for revocation checking when you've forgone certificates in favour of raw public keys. E.g. neither CRLs, nor OCSP will cover these entities. 

Does that match your understanding? Should a client or server configuration that's configured with both CRLs and raw public keys fail in an obvious manner?

cpu:
> Whenever possible I used the SubjectPublicKeyInfoDer type. However, sometimes the best option seems casting to a CertificateDer type. For example in AlwaysResolvesClientRawPublicKeys::new().

I'm curious what other reviewers think of this aspect. From my perspective it seems unforatunate/subpar to be doing this cast since we know the SPKI DER _isn't_ certificate DER. It feels like we want a more type-safe way of representing the union of the two but I'm not sure what kind of flexibility we have to rework this without breaking changes.

ctz:
> > Whenever possible I used the SubjectPublicKeyInfoDer type. However, sometimes the best option seems casting to a CertificateDer type. For example in AlwaysResolvesClientRawPublicKeys::new().
> 
> I'm curious what other reviewers think of this aspect. From my perspective it seems unfortunate/subpar to be doing this cast since we know the SPKI DER _isn't_ certificate DER. It feels like we want a more type-safe way of representing the union of the two but I'm not sure what kind of flexibility we have to rework this without breaking changes.

Speaking strictly at the level of public API changes, I would really like to avoid  the 99.9999% of users -- who have no interest in or use for RPK -- from having to think about this.

eg, `fn peer_certificates(&self)` -- I think it would be quite bad if we changed the return type of this to `enum CertOrRpk { ... }` and force those users to discard the statically-unreachable Rpk variant. Similarly, I would like to avoid adding a parallel `fn raw_public_key(&self)` because (frankly) I don't think this feature will ever gain common use to warrant separate public API surface and the mental load that adds to users.

rustls/src/client/builder.rs
```diff
@@ -163,6 +162,36 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
...
+    ///
+    /// `key_der` is a DER-encoded private key as PKCS#1, PKCS#8, or SEC1. The
+    /// `aws-lc-rs` and `ring` [`CryptoProvider`]s support all three encodings,
+    /// but other `CryptoProviders` may not.
+    ///
+    /// This function fails if `key_der` is invalid, or if the public_key cannot be retrieved from the private_key.
+    ///
+    /// Note: This API is only usable for `SigningKey` implementations that can produce the corresponding public key.
+    /// Please ensure that the `CryptoProvider` you are using supports this functionality.
+    pub fn with_client_auth_rpk(
```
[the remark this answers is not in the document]
djc: `with_client_auth_raw_public_key()` seems a little verbose to me, but `with_client_auth_raw()` or `with_client_auth_raw_key()` could be a good fit.

(`rpk` evokes an association with the RPKI for me, so I agree that avoiding it is probably better.)

djc:
> eg, `fn peer_certificates(&self)` -- I think it would be quite bad if we changed the return type of this to `enum CertOrRpk { ... }` and force those users to discard the statically-unreachable Rpk variant. Similarly, I would like to avoid adding a parallel `fn raw_public_key(&self)` because (frankly) I don't think this feature will ever gain common use to warrant separate public API surface and the mental load that adds to users.

Agreed. In Quinn we call this [`peer_identity()`](https://docs.rs/quinn/latest/quinn/struct.Connection.html#method.peer_identity) and model it as `Box<dyn Any>`. I don't think we should model it as `Any` here (since that would make life harder for the 99.9%) but might make sense to provide a method alias that is slightly less specific?

ctz:
> there is a decode error when running the example against openssl with `--verify 1`. However, this only happens when using  tls1.2 not tls1.3. Reading the logs it seems to be a problem related to the tag/length of the rpk, any idea why this is the case in tls1.2 (but not tls1.3)? 

RFC7250 has:

```
struct {
       select(certificate_type){

            // certificate type defined in this document.
            case RawPublicKey:
              opaque ASN.1_subjectPublicKeyInfo<1..2^24-1>;

           // X.509 certificate defined in [RFC 5246](https://datatracker.ietf.org/doc/html/rfc5246)
           case X.509:
             ASN.1Cert certificate_list<0..2^24-1>;

           // Additional certificate type based on
           // "TLS Certificate Types" subregistry
       };
   } Certificate;
```

This means the encoding of a single x509 certificate with value 0xaa is:

```
  00 00 04
     00 00 01
        aa
```

and the encoding of a single raw public key is:

```
  00 00 01
     aa
```

I think that means it is necessary to introduce an additional `HandshakePayload` variant, plus the context to know when to decode it (observe above, that `certificate_type` is not actually in the message anywhere.)

examples/src/bin/rpk_example.rs
djc: There is a tendency in some Rust projects to gather a boatload of relatively niche examples which will then incur maintenance overhead on things like dependency upgrades, so I've started pushing back on adding more examples for relatively niche things in those projects. Feels like this might be on of those things, and this could live in some other repo for people who care about raw keys?

rustls/src/client/handy.rs
```diff
@@ -241,6 +242,44 @@ impl client::ResolvesClientCert for AlwaysResolvesClientCert {
...
 
+#[derive(Debug)]
+pub(super) struct AlwaysResolvesClientRawPublicKeys(Arc<sign::CertifiedKey>);
+
+impl AlwaysResolvesClientRawPublicKeys {
+    pub(super) fn new(private_key: Arc<dyn sign::SigningKey>) -> Result<Self, Error> {
+        let public_key_as_cert = CertificateDer::from(
+            private_key
+                .public_key()
+                .unwrap()
```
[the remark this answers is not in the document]
holodorum: I'll add an error if `public_key()` returns `None`. Don't think there is a specific Error for this so far, any preferences for the Error that we return?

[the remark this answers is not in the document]
cpu: > Any preferences for the Error that we return?

No strong preference from me :-) `UnsupportedKey`?

examples/src/bin/rpk_example.rs
cpu: I haven't felt like our `examples/` dir has been a source of much maintenance burden (it feels relatively rare that I touch files in there except for breaking API changes). That said, I agree raw public keys is likely to be quite niche and so wouldn't be heartbroken if the example were kept external to the project.

@ctz Do you have a pref on this subject?

cpu:
> I might prefer a parallel function if that means we can avoid sticking not-certificate DER in a CertificateDer which IMO is also a pretty bad anti-pattern.

I was thinking similar here. I'm fully in agreement that we want to avoid making the subject of raw public keys front-and-center for the average user, but it doesn't seem too egregious to have a parallel function like this.

Cargo.toml
```diff
@@ -31,3 +31,7 @@ resolver = "2"
 [profile.bench]
 codegen-units = 1
 lto = true
+
+[patch.crates-io]
+rustls-webpki = { git = "https://github.com/holodorum/rustls-webpki.git", branch = "rpk_entity" }
```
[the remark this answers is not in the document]
cpu: a fresh and shiny [0.102.7](https://github.com/rustls/webpki/releases/tag/v%2F0.102.7) is now available :rocket:

ctz:
I guess that was just one example.  To completely avoid putting an RPK into a `CertificateDer` in the public API, we'd also need to provide RPK variants of:

1. `sign::CertifiedKey` (or modalise it, but that would be a breaking change)
2. `client::ResolvesClientCert` for (1), and its slot in `ClientConfig`
3. `server::ResolvesServerCert` for (1), and its slot in `ServerConfig`
4. `ServerCertVerifier`, and its slot in `ClientConfig`
5. `ClientCertVerifier`, and its slot in `ServerConfig`

(I'm not sure that's a complete list.)

holodorum:
> Beyond the mechanics of the implementation I'm also interested in how this interacts with revocation. This is one of the aspects of raw public keys that I think deserves explicit attention in our documentation. As I understand it there _isn't_ any mechanism for revocation checking when you've forgone certificates in favour of raw public keys. E.g. neither CRLs, nor OCSP will cover these entities.
> 
> Does that match your understanding? Should a client or server configuration that's configured with both CRLs and raw public keys fail in an obvious manner?

In the [RFC 7250 §6](https://datatracker.ietf.org/doc/html/rfc7250#section-6) a few comments are made regarding security considerations. There is no mention of either CRL or OCSP. Just different methods to authenticate raw public keys out-of-band.

The assumption we made following @ctz's [comment](https://github.com/rustls/rustls/issues/423#issuecomment-1908082743) is that when someone decides to use RPK's they have pretty specific a-priori knowledge about the server and how it will be authenticated (and vice versa). I'm not sure if it should happen that a server is using both CRLs and raw public keys (unless using the `Acceptor` api), in that case it might be a good idea to indeed fail. 

What do you think?

holodorum:
> > there is a decode error when running the example against openssl with `--verify 1`. However, this only happens when using  tls1.2 not tls1.3. Reading the logs it seems to be a problem related to the tag/length of the rpk, any idea why this is the case in tls1.2 (but not tls1.3)?
> 
> RFC7250 has:
> 
> ```
> struct {
>        select(certificate_type){
> 
>             // certificate type defined in this document.
>             case RawPublicKey:
>               opaque ASN.1_subjectPublicKeyInfo<1..2^24-1>;
> 
>            // X.509 certificate defined in [RFC 5246](https://datatracker.ietf.org/doc/html/rfc5246)
>            case X.509:
>              ASN.1Cert certificate_list<0..2^24-1>;
> 
>            // Additional certificate type based on
>            // "TLS Certificate Types" subregistry
>        };
>    } Certificate;
> ```
> 
> This means the encoding of a single x509 certificate with value 0xaa is:
> 
> ```
>   00 00 04
>      00 00 01
>         aa
> ```
> 
> and the encoding of a single raw public key is:
> 
> ```
>   00 00 01
>      aa
> ```
> 
> I think that means it is necessary to introduce an additional `HandshakePayload` variant, plus the context to know when to decode it (observe above, that `certificate_type` is not actually in the message anywhere.) Painful!
> 
> Do you need this feature for TLS1.2?

Thank you @ctz!! That clarifies things a lot. Very painful indeed.
One thing I don't understand yet is why this is different in TLS1.2 and TLS1.3, is there a default decoding done in TLS1.2 and a way to not use it?

rustls/src/client/hs.rs
holodorum: Reason for this is that we want to fail if in the `client_cert_type` or `server_cert_type` extension both a x509 and RPK are offered. However, in tls13 the extension is encrypted so we call `process_client_server/cert_type_extension()` in `tls13::ExpectEncryptedExtensions`. To prevent calling this function twice we added the branching.

ctz:
> One thing I don't understand yet is why this is different in TLS1.2 and TLS1.3

TLS1.3 has a `CertificateEntry` struct in the way, and ensuring there is only one is done with prose rather than the wire encoding.

rustls/src/client/handy.rs
holodorum: Added a comment that mentions the nuance

rustls/src/server/handy.rs
holodorum: Is it correct as it's now?

rustls/src/server/server_conn.rs
holodorum: I've added a `Server/ClientCertType` tuple struct to `server_conn.rs`, I think it's more robust like this. Or do you have a different method in mind? 
Also, what would be a good location to add these tuple structs?

aochagavia:
Thanks for the review work! I've been lurking here for a while, as I'm helping a bit behind the scenes. It looks like there are a few decisions still to be made, and I wanted to provide my 2 cents.

### Could we avoid supporting TLS 1.2?

It turns out that supporting TLS 1.2 adds some undesirable complexity. @cpu pointed out the branching it introduces, and @ctz that we'd probably need a new `HandshakePayload` variant. That's painful and it would be nice to forgo TLS 1.2 support if possible. Since the main stakeholder for this feature is Solana, and they dropped TLS 1.2 support in 2023 (see [this PR](https://github.com/solana-labs/solana/pull/28956)), it looks like we can indeed skip TLS 1.2 here :)

### Where should the example code live?

@djc [voiced some concerns](https://github.com/rustls/rustls/pull/2062#discussion_r1732333347) on letting the example code live in this repository, to which @cpu responded with his opinion. There's no consensus yet, but the issue is small enough that I guess it can be easily decided once @ctz gives tells us what he thinks.
### `CertificateDer` and SPKI DERs

@cpu [mentioned](https://github.com/rustls/rustls/pull/2062#issuecomment-2310726679) it's unfortunate to be representing an SPKI DER as a struct of the `CertificateDer` type, since they are _not_ the same. I think everyone agrees, but it looks like a significant amount of internal refactoring would be required to achieve a clean separation between the two (while preserving backwards-compatibility in the API). See, for instance, [this non-exhaustive list](https://github.com/rustls/rustls/pull/2062#issuecomment-2312853615) @ctz came up with.

I guess the issue boils down to choosing between (1) an inconsistency in the `CertificateDer` type, which sometimes contains an SPKI DER; and (2) consistent types at the cost of more complex internals. Personally, I feel inclined towards 1 (maybe you could consider Raw Public Keys as a kind of certificate, since the RFC says they are a new "certificate type", so the name `CertificateDer` is not _that_ inconsistent in the end). Regardless, I guess you'd better pick whatever is easiest to maintain long-term.

rustls/src/common_state.rs
holodorum: Yeah that is tricky indeed. I've added a struct `RawKeyNegotationParams`, with fields 'local_expects_rpk` and `peer_supports_rpk` as fields. It should be harder to mistakenly switch the order around now.

holodorum:
Thanks for the feedback and all the work so far. Except for the comments on `rpk_example.rs`, we've incorporated all the feedback. In case I wasn't sure whether our approach is correct I left the conversation unresolved. 
For now I think a few things need to happen: 

- [ ] Create an example that is in line with the rest of the examples, either in its own repository or in this repository. 
- [ ]  Basic positive interop tests against OpenSSL 3.2. 
- [ ]  Improve the tests for clients/servers rejecting the peer for missing RPK support and get full test coverage. 
- [ ] Update the documentation where necessary
- [ ] Depending on the outcome of the dialogue refactor the scenarios where a RPK is wrapped in a `CertificateDer`

If there's anything I'm forgetting please let me know!

rustls/src/client/hs.rs
cpu: Makes sense :+1: It sounds like consensus is moving towards not implementing this for TLS1.2, but if support stays it would be useful to capture this context at the call-site with a comment (IMO).

rustls/src/common_state.rs
```diff
@@ -883,6 +905,17 @@ enum Limit {
     No,
 }
 
+pub(crate) enum RpkNegotiationResult {
```
cpu, line 908: `#[derive(Debug)]`?

rustls/src/common_state.rs
```diff
@@ -883,6 +905,17 @@ enum Limit {
...
+    Negotiated,
+    NotNegotiated,
+    Err(Error),
+}
+
+pub(super) struct RawKeyNegotationParams {
```
cpu, line 914: ditto RE: `Debug`. I think it's almost universally helpful for new struct/enums.

cpu, line 914: I think you have a typo in this type name: `RawKeyNegotationParams` -> `RawKeyNegotiationParams`

rustls/src/error.rs
```diff
@@ -47,6 +47,9 @@ pub enum Error {
     /// The certificate verifier doesn't support the given type of name.
     UnsupportedNameType,
 
+    /// We couldn't retrieve a public key.
+    UnsupportedPublicKey,
```
cpu, line 51: bike-shedding: I wonder if there's a better name for this since it reflects that we couldn't retrieve a public key more than that the public key was in-hand, but unsupported. `UnknownPublicKey`?

rustls/src/server/server_conn.rs
```diff
@@ -196,6 +226,20 @@ impl<'a> ClientHello<'a> {
     pub fn cipher_suites(&self) -> &[CipherSuite] {
         self.cipher_suites
     }
+
+    /// Get the server certificate types offered in the ClientHello.
+    ///
+    /// Returns `None` if the client did not include a certificate type extension.
+    pub fn server_cert_types(&self) -> Option<&'a [CertificateType]> {
```
cpu, line 233: nit: I think it might be more natural to name this `server_cert_type()` since it's a singular `CertificateType` optionally returned? WDYT?

(If you agree, same feedback applies to `client_cert_types()`).

rustls/src/webpki/mod.rs
```diff
@@ -16,12 +16,16 @@ mod verify;
 pub use anchors::RootCertStore;
 pub use client_verifier::{ClientCertVerifierBuilder, WebPkiClientVerifier};
 pub use server_verifier::{ServerCertVerifierBuilder, WebPkiServerVerifier};
+
```
cpu, line 19: spurious diff?

rustls/src/server/handy.rs
```diff
@@ -275,14 +307,15 @@ mod sni_resolver {
 
     #[cfg(test)]
     mod tests {
+
```
cpu, line 310: spurious diff?

rustls/src/server/server_conn.rs
```diff
@@ -120,13 +121,26 @@ pub trait ResolvesServerCert: Debug + Send + Sync {
...
+    fn only_raw_public_keys(&self) -> bool {
+        false
+    }
 }
 
+/// A struct representing the server certificate types offered in the ClientHello.
+pub(super) struct ServerCertType<'a>(pub(super) &'a [CertificateType]);
+
+/// A struct representing the client certificate types offered in the ClientHello.
+pub(super) struct ClientCertType<'a>(pub(super) &'a [CertificateType]);
```
cpu, line 135: I think the visibility of these two items means they should be lower in the mod (below the `pub` items).

rustls/tests/api.rs
```diff
@@ -47,6 +47,210 @@ use provider::cipher_suite;
...
+            test_handshake_unexpected_server_hello(&client_config, &server_config);
+        }
+    }
+
+    fn test_handshake_incorrectly_altered_client_hello(
+        client_config: &ClientConfig,
+        server_config: &ServerConfig,
+    ) {
+        //If server requires rpk, client must send rpk
+        //If client sends nothing, x509 or x509 and rpk, server should return IncorrectCertificateTypeExtension
```
cpu, line 75: nit: leading space for comments, `// ...` vs `//...`. Ditto on L128/129.

rustls/src/server/server_conn.rs
cpu: Your approach looks good to me and I think where you defined the struct makes sense :+1:  Thanks!

cpu (commented):
Here's another round of feedback. Thanks!

> In the [RFC 7250 §6](https://datatracker.ietf.org/doc/html/rfc7250#section-6) a few comments are made regarding security considerations. There is no mention of either CRL or OCSP. Just different methods to authenticate raw public keys out-of-band.

Indeed :-( I think that's an oversight. I'm not sure it merits an errata but it feels like a missing piece to me.

> The assumption we made following @ctz's https://github.com/rustls/rustls/issues/423#issuecomment-1908082743 is that when someone decides to use RPK's they have pretty specific a-priori knowledge about the server and how it will be authenticated (and vice versa). 

I agree with the assumption Ctz & you folks are making. I'm mainly advocating that this patchset should try to explicitly document the need for that a-priori knowledge in very strong & obvious language.

> I'm not sure if it should happen that a server is using both CRLs and raw public keys (unless using the Acceptor api), in that case it might be a good idea to indeed fail.

You're right, I think the main concern I had (a client/server configuration with both CRLs and RPK configured) isn't possible based on this impl :+1: I think the `Acceptor` API shouldn't introduce any additional risk. 

> I guess the issue boils down to choosing between (1) an inconsistency in the CertificateDer type, which sometimes contains an SPKI DER; and (2) consistent types at the cost of more complex internals. Personally, I feel inclined towards 1 

I think I'm coming around to (1) as well short any other ideas that can be implemented within semver without complicating the API for the normal use-case.

rustls/src/server/server_conn.rs
holodorum: It's not necessarily the case that only a singular `CertificateType` is returned. When a client supports both x509 and rpk's a both `CertificateType`s are sent in the extension. The 'equivalent' function for handling server extensions, `HasServerExtensions::client_cert_type`, does always return a singular `CertificateType`.

rustls/src/error.rs
holodorum: In the spirit of bike-shedding: would it be appropriate to use `InconsistentKeys::Unknown`?

rustls/src/server/server_conn.rs
```diff
@@ -120,13 +121,26 @@ pub trait ResolvesServerCert: Debug + Send + Sync {
...
     fn resolve(&self, client_hello: ClientHello<'_>) -> Option<Arc<sign::CertifiedKey>>;
+
+    /// Return true when the server only supports raw public keys.
+    fn only_raw_public_keys(&self) -> bool {
+        false
+    }
 }
 
+/// A struct representing the server certificate types offered in the ClientHello.
+pub(super) struct ServerCertType<'a>(pub(super) &'a [CertificateType]);
```
djc, line 132: It's confusing that a type called `ServerCertType` (singular) contains a `[CertificateType]` (a slice of types -- plural). Which one makes sense? It doesn't seem like they can make sense at the same type.

And honestly, what's the value of introducing these newtypes in the first place?

holodorum: You're right, the client can offer multiple `CertificateType`s to the server so it should be plural. 
The added value is to make the `ClientHello::new()` signature more robust by making it harder to mix up the client and server certificate types. Do you think these ewtypes are redundant?

djc: I think so -- should be clear enough from the field/method context in which they appear.

rustls/src/error.rs
cpu: That feels reasonable to me, but if we go this direction we should update the part of the docs that say "Specific failure cases from [keys_match](https://docs.rs/rustls/latest/rustls/sign/struct.CertifiedKey.html#method.keys_match).".

rustls/src/server/server_conn.rs
cpu: Ahh! Ok, that makes sense. Thanks for explaining.

holodorum:
@cpu I think so too! Thanks for this round of feedback. Most of it is already processed, and beginning next week I'll finish it and push everything.

holodorum:
We've incorporated all the feedback and finished the todo's. Thanks for all the great help and feedback so far. Hopefully we can get this merged soon ;)

Two remarks about the OpenSSL tests and the example:
> - [x] Create an example that is in line with the rest of the examples, either in its own repository or in this repository.
> - [x] Basic positive interop tests against OpenSSL 3.2.

The 'example' is not really in line with the rest of the examples right now, since it is living in `openssl-tests`. There was some discussion over adding a niche example, maybe this is the best of both worlds. There is an example implementation for using raw keys, and we don't need to duplicate code to test against OpenSSL. 

The OpenSSL tests fail at the moment, most likely because the runner has OpenSSL 3.0 instead of 3.2. Is it better to add a new workflow for OpenSSL tests with OpenSSL 3.2 installed on the runner or is it okay to modify the existing workflow and run all tests against OpenSSL 3.2?

cpu:
> there was some discussion over adding a niche example, maybe this is the best of both worlds. There is an example implementation for using raw keys, and we don't need to duplicate code to test against OpenSSL.

Seems reasonable to me :+1: 

> The OpenSSL tests fail at the moment, most likely because the runner has OpenSSL 3.0 instead of 3.2. Is it better to add a new workflow for OpenSSL tests with OpenSSL 3.2 installed on the runner or is it okay to modify the existing workflow and run all tests against OpenSSL 3.2?

I think testing against one version is likely sufficient unless someone has strong opinions. 3.2 also ships certificate compression and it might be nice to have interop testing for that some day.

rustls/src/lib.rs
```diff
@@ -537,6 +538,7 @@ pub use crate::suites::{
 pub use crate::ticketer::TicketSwitcher;
 #[cfg(feature = "tls12")]
 pub use crate::tls12::Tls12CipherSuite;
+pub use crate::tls13::key_schedule::{derive_traffic_iv, derive_traffic_key};
```
ctz, line 541: Let's have this in `internal` please, not the crate top-level.

rustls/src/tls13/key_schedule.rs
```diff
@@ -770,6 +770,20 @@ impl KeySchedule {
     }
 }
 
+/// [HKDF-Expand-Label] where the output is an AEAD key.
+///
+/// [HKDF-Expand-Label]: <https://www.rfc-editor.org/rfc/rfc8446#section-7.1>
+pub fn derive_traffic_key(expander: &dyn HkdfExpander, aead_key_len: usize) -> AeadKey {
+    hkdf_expand_label_aead_key(expander, aead_key_len, b"key", &[])
```
ctz, line 777: This being a formerly private "detail" function, it now feels inappropriately low level to accept `aead_key_len` as a usize when it will panic for all values > 32. How about accepting a  `&Tls13AeadAlgorithm` instead -- all existing code paths actually get this value from one of these already.

rustls/tests/api.rs
```diff
@@ -47,6 +47,403 @@ use provider::cipher_suite;
...
+    use rustls::crypto::cipher::{
+        InboundOpaqueMessage, MessageDecrypter, MessageEncrypter, OutboundChunks,
+        OutboundPlainMessage,
+    };
+    use rustls::crypto::tls13::OkmBlock;
+    use rustls::{derive_traffic_iv, derive_traffic_key};
+    use rustls::{Connection, Tls13CipherSuite};
+
+    #[test]
+    fn succesful_raw_key_connection_and_correct_peer_certificates() {
```
ctz, line 61:
```suggestion
    fn successful_raw_key_connection_and_correct_peer_certificates() {
```

rustls/tests/client_cert_verifier.rs
```diff
@@ -137,73 +132,3 @@ fn client_verifier_fails_properly() {
         }
     }
 }
-
```
ctz, line 140: Perhaps pull out this code movement into an earlier separate commit?

rustls/src/tls13/key_schedule.rs
holodorum: Clear, that's indeed better. I changed the function signature in a seperate commit.

rustls/src/tls13/key_schedule.rs
```diff
@@ -770,6 +770,23 @@ impl KeySchedule {
     }
 }
 
+/// [HKDF-Expand-Label] where the output is an AEAD key.
+///
+/// [HKDF-Expand-Label]: <https://www.rfc-editor.org/rfc/rfc8446#section-7.1>
+pub fn derive_traffic_key(
+    expander: &dyn HkdfExpander,
+    aead_alg: &dyn Tls13AeadAlgorithm,
```
djc, line 778: Nit: please don't move the code and change it in the same commit, this makes it hard to review the changes.

rustls/src/client/builder.rs
```diff
@@ -163,6 +163,34 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
         Ok(self.with_client_cert_resolver(Arc::new(resolver)))
     }
 
+    /// Sets a raw public key and matching private key for use
+    /// in client authentication.
+    ///
+    /// `key_der` is a DER-encoded private key as PKCS#1, PKCS#8, or SEC1. The
```
djc, line 169: "DER-encoded private key as PKCS#1, PKCS#8, or SEC1" reads a little awkwardly to me. Maybe "private key encoded as PKCS#1, PKCS#8 or SEC1 DER"?

rustls/src/client/builder.rs
```diff
@@ -163,6 +163,34 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
...
+    /// `aws-lc-rs` and `ring` [`CryptoProvider`]s support all three encodings,
+    /// but other `CryptoProviders` may not.
+    ///
+    /// This function fails if `key_der` is invalid, or if the public key cannot be retrieved from the private key.
+    ///
+    /// Note: This API is only usable for `SigningKey` implementations that can produce the corresponding public key.
+    /// Please ensure that the `CryptoProvider` you are using supports this functionality.
+    pub fn with_client_auth_raw_key(
```
djc, line 177: Instead of adding this niche API surface, should we just expose `AlwaysResolvesClientRawPublicKeys` directly (maybe under a better name, even if it isn't consistent with the existing naming)?

rustls/src/client/handy.rs
```diff
@@ -241,6 +241,44 @@ impl client::ResolvesClientCert for AlwaysResolvesClientCert {
     }
 }
 
+#[derive(Debug)]
+pub(super) struct AlwaysResolvesClientRawPublicKeys(Arc<sign::CertifiedKey>);
+
+impl AlwaysResolvesClientRawPublicKeys {
+    pub(super) fn new(private_key: Arc<dyn sign::SigningKey>) -> Result<Self, Error> {
```
djc, line 248: What's the motivation to not just take an `Arc<sign::CertifiedKey>` directly?

rustls/src/common_state.rs
```diff
@@ -274,6 +278,28 @@ impl CommonState {
         Ok(self.write_fragments(outgoing_tls, fragments))
     }
 
+    pub(crate) fn validate_rpk_negotiation(
```
djc, line 281: Would prefer to avoid the `rpk` acronym in favor of something more explicit, like `raw_key` (both in the name and in bindings inside it). Also it looks like this doesn't actually use `self`, make it a method on `RawKeyNegotationParams` instead?

rustls/src/common_state.rs
```diff
@@ -274,6 +278,28 @@ impl CommonState {
...
+        }
+        if local_expects_rpk && !peer_supports_rpk {
+            return RawKeyNegotationResult::Err(Error::PeerIncompatible(
+                PeerIncompatible::IncorrectCertificateTypeExtension,
+            ));
+        }
+        if !local_expects_rpk && !peer_supports_rpk {
+            return RawKeyNegotationResult::NotNegotiated;
+        }
+        RawKeyNegotationResult::Negotiated(negotation_params.extension_type)
```
djc, line 300: I'd suggest spelling this as a tuple match:

```rust
match (local_expects_raw, peer_supports_raw) {
    ..
}
```

rustls/src/common_state.rs
```diff
@@ -887,6 +913,20 @@ enum Limit {
...
+    NotNegotiated,
+    Err(Error),
+}
+
+#[derive(Debug)]
+pub(super) struct RawKeyNegotiationParams {
+    pub(super) peer_supports_raw_key: bool,
+    pub(super) local_expects_raw_key: bool,
+    pub(super) extension_type: ExtensionType,
+}
```
djc, line 928: IMO `..Params` should come before `..Result`.

rustls/src/crypto/signer.rs
```diff
@@ -87,9 +91,11 @@ pub trait Signer: Debug + Send + Sync {
 
 /// A packaged-together certificate chain, matching `SigningKey` and
 /// optional stapled OCSP response.
+///
+/// Note: This struct is also used to represent a raw public key, when the client/server is configured to use raw public keys instead of certificates
```
djc, line 95: "This" -> "this", add a period at the end.

rustls/src/msgs/handshake.rs
```diff
@@ -1303,12 +1343,12 @@ impl HelloRetryRequest {
 
 #[derive(Clone, Debug)]
 pub struct ServerHelloPayload {
+    pub extensions: Vec<ServerExtension>,
```
djc, line 1346: Don't change the order and the code in the same commit, please?

rustls/src/msgs/handshake.rs
```diff
@@ -2037,6 +2077,22 @@ pub(crate) trait HasServerExtensions {
         }
     }
 
+    fn server_cert_type(&self) -> Option<&CertificateType> {
+        let ext = self.find_extension(ExtensionType::ServerCertificateType)?;
+        match *ext {
```
djc, line 2082: Prefer to avoid `ref` which are usually unnecessary.

rustls/src/server/builder.rs
```diff
@@ -88,6 +88,31 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
+    ///
+    /// `key_der` is a DER-encoded private key as PKCS#1, PKCS#8, or SEC1. The
+    /// `aws-lc-rs` and `ring` [`CryptoProvider`]s support all three encodings,
+    /// but other `CryptoProviders` may not.
+    ///
+    /// This function fails if `key_der` is invalid, or if the public key cannot be retrieved from the private key.
+    ///
+    /// Note: This API is only usable for `SigningKey` implementations that can produce the corresponding public key.
+    /// Please ensure that the `CryptoProvider` you are using supports this functionality.
+    pub fn with_raw_key(self, key_der: PrivateKeyDer<'static>) -> Result<ServerConfig, Error> {
```
djc, line 103: Same here as for client.

rustls/src/server/hs.rs
```diff
@@ -201,6 +206,83 @@ impl ExtensionProcessing {
...
+                    ));
+            }
+            RawKeyNegotationResult::Err(err) => {
+                return Err(cx
+                    .common
+                    .send_fatal_alert(AlertDescription::HandshakeFailure, err));
+            }
+            RawKeyNegotationResult::NotNegotiated => {}
+            RawKeyNegotationResult::Negotiated(_) => {
+                unreachable!();
```
djc, line 281: Nit: remove the `;` to make this more compact. Maybe add a comment on the arm why this is unreachable?

rustls/src/webpki/verify.rs
```diff
@@ -199,6 +199,30 @@ pub fn verify_tls13_signature(
         .map(|_| HandshakeSignatureValid::assertion())
 }
 
+/// Verify a message signature using a raw public key and the first TLS 1.3 compatible
+/// supported scheme.
+///
+/// This function verifies the `dss` signature over `message` using the subject public key from
+/// `cert`. See [WebPkiSupportedAlgorithms::mapping] for more information.
+pub fn verify_tls13_signature_with_spki(
```
djc, line 207: IMO this function's name should reference that it's about raw keys (only?).

openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -0,0 +1,499 @@
...
+
+/// Load the private key from the file at the given path
+fn load_private_key(path_to_private_key: &str) -> Result<PrivateKeyDer<'static>, io::Error> {
+    let file = File::open(path_to_private_key)?;
+    let mut reader = BufReader::new(file);
+    let key = rustls_pemfile::private_key(&mut reader)
+        .map_err(|_| std::io::Error::new(io::ErrorKind::InvalidData, "Invalid private key data"))?;
+
+    Ok(key.unwrap())
+}
```
djc, line 29: Nit: these should probably be somewhere near the bottom (before `mod tests`).

rustls/src/client/handy.rs
holodorum: Because it's in line with the `AlwaysResolvesClientCert::new()` function, where you also pass an `Arc<dyn sign::SigningKey>` and in the function an Arc<sign::CertifiedKey> is created. 

If it's better to just take `Arc<sign::CertifiedKey>` I can of course change it.

.github/workflows/build.yml
```diff
@@ -504,6 +504,15 @@ jobs:
...
         uses: dtolnay/rust-toolchain@stable
+  
+      - name: Install pixi
+        run: | 
+          curl -fsSL https://pixi.sh/install.sh | bash
+          echo "$HOME/.pixi/bin" >> $GITHUB_PATH
+
+      - name: Install OpenSSL 3.2 with pixi
+        run: |
+          pixi global install openssl==3.2
```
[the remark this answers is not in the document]
djc: Yeah, I don't really have a better recommendation. Just wanted to flag this, curious what the other maintainers think.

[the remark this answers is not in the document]
ctz: I looked for suitable PPAs yesterday but didn't find any. That would be the route I was going to suggest. I haven't heard of pixi, but I think if it does the job without getting in the way, that's fine for CI.

openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -0,0 +1,499 @@
...
+
+/// Load the raw public key from the file at the given path
+fn load_raw_key(path_to_rpk: &str) -> Vec<SubjectPublicKeyInfoDer<'static>> {
+    let file = File::open(path_to_rpk).expect("cannot open certificate file");
+    let mut reader = BufReader::new(file);
+    rustls_pemfile::public_keys(&mut reader)
+        .map(|result| result.unwrap())
+        .collect()
+}
```
ctz, line 292: This can become: `SubjectPublicKeyInfoDer::from_pem_file(path_to_rpk).expect("canont open public key file")`

openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -0,0 +1,499 @@
...
+
+/// Load the private key from the file at the given path
+fn load_private_key(path_to_private_key: &str) -> Result<PrivateKeyDer<'static>, io::Error> {
+    let file = File::open(path_to_private_key)?;
+    let mut reader = BufReader::new(file);
+    let key = rustls_pemfile::private_key(&mut reader)
+        .map_err(|_| std::io::Error::new(io::ErrorKind::InvalidData, "Invalid private key data"))?;
+
+    Ok(key.unwrap())
+}
```
ctz, line 302: Ditto, `PrivateKeyDer::from_pem_file(path_to_private_key).expect("cannot open private key file")`

rustls/Cargo.toml
```diff
@@ -55,6 +55,7 @@ hex = "0.4"
 log = "0.4.8"
 num-bigint = "0.4.4"
 rcgen = { version = "0.13", default-features = false, features = ["aws_lc_rs", "pem"] }
+rustls-pemfile = "2.1.3"
```
ctz, line 58: Reintroducing this dep shouldn't be needed.

rustls/tests/common/mod.rs
```diff
@@ -310,6 +325,13 @@ impl KeyType {
             .collect()
     }
 
+    pub fn get_spki(&self) -> SubjectPublicKeyInfoDer<'static> {
+        rustls_pemfile::public_keys(&mut io::BufReader::new(self.bytes_for("end.spki.pem")))
+            .map(|result| result.unwrap())
+            .next() // We only expect one RPK.
+            .unwrap()
```
ctz, line 332: Apologies, main has moved under you a bit. This can be replaced with:

```suggestion
        SubjectPublicKeyInfoDer::from_pem_slice(self.bytes_for("end.spki.pem"))
            .unwrap()
```
(ditto below)

Cargo.lock
```diff
@@ -494,9 +494,9 @@ checksum = "428d9aa8fbc0670b7b8d6030a7fadd0f86151cae55e4dbbece15f3780a3dfaf3"
 
 [[package]]
 name = "cc"
-version = "1.1.22"
+version = "1.1.23"
```
ctz, line 497: I think this commit can now be dropped?

rustls/src/tls13/key_schedule.rs
```diff
@@ -1045,7 +1046,7 @@ test_for_each_provider! {
         let expander = TLS13_AES_128_GCM_SHA256_INTERNAL
             .hkdf_provider
             .expander_for_okm(&traffic_secret);
-        let key = derive_traffic_key(expander.as_ref(), aead_alg.key_len());
+        let key = derive_traffic_key(expander.as_ref(), TLS13_AES_128_GCM_SHA256_INTERNAL.aead_alg);
```
ctz, line 1049: Just `aead_alg` here would be equivalent.

rustls/src/webpki/verify.rs
```diff
@@ -199,6 +199,30 @@ pub fn verify_tls13_signature(
         .map(|_| HandshakeSignatureValid::assertion())
 }
 
+/// Verify a message signature using a raw public key and the first TLS 1.3 compatible
+/// supported scheme.
+///
+/// This function verifies the `dss` signature over `message` using the subject public key from
+/// `cert`. See [WebPkiSupportedAlgorithms::mapping] for more information.
```
ctz, line 206: Perhaps drop this bit of comment?

rustls/src/tls13/key_schedule.rs
```diff
@@ -770,6 +770,21 @@ impl KeySchedule {
     }
 }
 
+/// [HKDF-Expand-Label] where the output is an AEAD key.
```
ctz, line 773: Reword commit message? (perhaps "move `derive_traffic_key` up in preparation for becoming public")

ctz (commented):
Thanks for persevering with this -- I feel it is pretty close.

rustls/tests/common/mod.rs
holodorum: Yup, I noticed after the CI crashing. Was just updating it. Pretty cool pattern to load the pemfiles I think.

rustls/src/tls13/key_schedule.rs
holodorum: It seems that `aead::AES_128_GCM` doesn't implement the `Tls13AeadAlgorithm` trait, or do I misunderstand what you mean?

rustls/src/client/builder.rs
djc: @ctz got opinions on this?

rustls/src/tls13/key_schedule.rs
ctz: Ah, true, it would have to mention `as &dyn Tls13AeadAlgorithm` which is pretty wordy, let's leave it how it is.

rustls/src/client/handy.rs
ctz: I think it is better in this case to take `sign::CertifiedKey`, as it does precisely what 99.9% people want (the public key from their private key); without the potential for getting them from divergent sources. That does mean someone who has a private key impl that can't get the public key will need to provide a whole trait impl, I think that is uncommon.

rustls/src/client/builder.rs
ctz: I agree about exposing the `handy::AlwaysResolvesClientRawPublicKey` exemplar implementation, mainly because I have duplicates of `AlwaysResolvesClientCert` and `AlwaysResolvesChain` in downstream projects for using alternate private key types (and have been too lazy thus far to fix that...).

I have a (very marginal) preference for losing the builder function after that, on the basis of a insubstantial worry that it could funnel new users towards this feature rather than away from it?

djc: > I have a (very marginal) preference for losing the builder function after that, on the basis of a insubstantial worry that it could funnel new users towards this feature rather than away from it?

That was one of my reasons as well.

cpu:
> cpu self-requested a review now

I will catch up with the progress in this branch ~tomorrow/friday. Thank you!

rustls/src/client/builder.rs
holodorum: Gotcha, `handy::AlwaysResolvesClientRawPublicKey` and `handy::AlwaysResolvesServerRawPublicKey` are now public and the builder functions are removed.

rustls/src/client/handy.rs
```diff
@@ -241,6 +241,32 @@ impl client::ResolvesClientCert for AlwaysResolvesClientCert {
     }
 }
 
+#[derive(Debug)]
+/// An exemplar `ResolvesClientCert` implementation that always resolves to a single raw public key.
+pub struct AlwaysResolvesClientRawPublicKeys(pub Arc<sign::CertifiedKey>);
```
djc, line 246: I think I'd prefer to keep the field private and a have simple `new()` constructor instead, to make it a little easier to potentially evolve the API if necessary.

(Nit: please put comments before attributes.)

rustls/src/server/handy.rs
```diff
@@ -199,6 +199,20 @@ impl server::ResolvesServerCert for AlwaysResolvesChain {
     }
 }
 
+#[derive(Debug)]
+/// An exemplar `ResolvesServerCert` implementation that always resolves to a single raw public key.
+pub struct AlwaysResolvesServerRawPublicKeys(pub Arc<sign::CertifiedKey>);
```
djc, line 204: (As for the client.)

djc (approved):
Thanks for all the work on this!

.github/workflows/build.yml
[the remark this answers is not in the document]
cpu: > I looked for suitable PPAs yesterday but didn't find any. 

That was going to be my first suggestion as well.

>  I tried installing it by downloading the source code and installing that, which took a lot longer than installing it via pixi

I think it'd be worth revisiting this later (let's land this as-is first). I bet with some cache configuration we get something competitive to Pixi but without the maintenance overhead.

rustls/tests/api.rs
```diff
@@ -47,6 +47,403 @@ use provider::cipher_suite;
...
+    use rustls::crypto::cipher::{
+        InboundOpaqueMessage, MessageDecrypter, MessageEncrypter, OutboundChunks,
+        OutboundPlainMessage,
+    };
+    use rustls::crypto::tls13::OkmBlock;
+    use rustls::internal::{derive_traffic_iv, derive_traffic_key};
+    use rustls::{Connection, Tls13CipherSuite};
+
+    #[test]
+    fn successful_raw_key_connection_and_correct_peer_certificates() {
```
cpu, line 61: The test coverage here is great, thank you! :bow:

rustls/src/client/builder.rs
```diff
@@ -12,7 +12,7 @@ use crate::error::Error;
 use crate::key_log::NoKeyLog;
```
cpu, line 12: Another comment without a good home: I think this commit could also update [`rustls/src/manual/features.rs`](https://github.com/rustls/rustls/blob/main/rustls/src/manual/features.rs) to mention that we support RFC 7250 RPKs thanks to this work.

examples/README.md
```diff
@@ -21,3 +21,7 @@ We recommend new users start by looking at `simpleclient.rs` and `simpleserver.r
 * `simple_0rtt_server.rs` - shows how to make a TLS1.3 that accepts multiple connections and prints early 0RTT data.
 * `server_acceptor.rs` - shows how to use the `Acceptor` API to create a server that generates a unique `ServerConfig` for each client. This example also shows how to use client authentication, CRL revocation checking, and uses `rcgen` to generate its own certificates.
 * `unbuffered-server.rs` - shows an advanced example of using Rustls lower-level APIs to implement a server that does not buffer any data inside Rustls.
+
+## Client-Server examples
+
+* A client-server example using Raw Public Keys (RFC 7250) can be found in [`raw_key_openssl_interop`](openssl-tests/src/raw_key_openssl_interop.rs).
```
cpu, line 27: I think this link target should be `../openssl-test/src/raw_key_openssl_interop.rs` no? I think it expands to `examples/openssl-tests/src
/raw_key_openssl_interop.rs`

openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -0,0 +1,522 @@
+//! This module provides tests for the interoperability of raw public keys with OpenSSL, and also
+//! demonstrates how to set up a client-server architecture that utilizes raw public keys.
+//!
+//! The module also includes an example implementations of the `ServerCertVerifier` and `ClientCertVerifier` trait, using
```
cpu, line 4: nit:
```suggestion
//! The module also includes example implementations of the `ServerCertVerifier` and `ClientCertVerifier` traits, using
```

cpu (commented):
Very nice work, thank you for sticking with this. 

I had a fair number of comments but they're all pretty insubstantial and should be easy fixes. I'm happy to +1 after that.

cpu:
Now might be a good point to update the PR description as well if you're up for it.

ctz (approved):
Modulo the outstanding comments, this LGTM

holodorum:
Thanks for all the thorough feedback and consistent input! I just processed the last outstanding comments (sorry that it took a bit longer), and I think it's ready to merge!

cpu (approved):
Great work, thanks again :-)

ripatel-fd:
@cpu @holodorum Hmm... This seems broken. After the latest rustls release, my peers stopped being able to handshake with rustls.

I believe this is because this implementation is not RFC 7250 compliant. rustls server configs that have an X.509 cert but no raw public key will now terminate handshakes with a TLS alert, if those servers see a ClientHello advertising both X.509 and RPK as a supported client certificate type. The correct behavior per [RFC 7250 Section 4.2](https://datatracker.ietf.org/doc/html/rfc7250#section-4.2) is to either omit the server_certificate_type extension, or to respond with client_certificate_type X.509.

I tried to hack around this problem on the rustls server side but stumbled across an API design issue.
The resolve function of the ResolvesClientCert::resolve and the CertVerifier traits don't provide the certificate type to callbacks. The selected client certificate type is part of handshake state and is required to correctly parse a certificate. I understand the design decision to not support multiple certificate types in the "safe" user API. For more advanced use cases, it would be nice to have the ability to support both via a custom ResolvesServerCert.

In pseudocode, I would expect the API to look more like so:

```
fn verify(cert, client_cert_type) {
    match(client_cert_type) {
        RPK => verify_rpk(cert),
        X.509 => verify_x509(cert),
    }
}
```

Do you mind if I open GitHub issues and provide test cases for the aforementioned two issues? Or is the above intended behavior? Many thanks in advance.

ripatel-fd:
@cpu Thank you for the quick response, I submitted these two issues: #2257 #2258.

PR #2065, opened 2024-08-01
Fix daily-tests failures

ctz:
Unfortunately feature unification turned on fips mode for every workspace item. Rather than chase around making that work, just avoid it by specifying the workspace item each time.

Fixing that uncovered a test that wanted to see the `HandshakePayloadTooLarge` error which didn't carry-over in the recent refactor. Restore that.

See passing run: https://github.com/rustls/rustls/actions/runs/10197371458


djc (approved):

cpu (approved):
Thanks :+1:

Issue #2021, opened 2024-06-24
aws-lc-rs ticketer can & should use RFC5077 "recommended ticket construction"


ctz:
The current default ticketers are compromise designs, because they are built on AEADs that have limited security margin when used with random nonces (and we avoid sequential nonces to avoid a privacy leak).

There are also theoretical concerns about these AEADs not providing key commitment, whereas the AES+HMAC generic composition does.

PR #2066, opened 2024-08-01
aws_lc_rs: implement RFC 5077 recommended ticketer

cpu:
Previously the `aws_lc_rs` crypto module shared its `Ticketer` implementation with `ring` using the "ring-like" mechanism we use for other shared implementations. This branch replaces it with an [RFC 5077](https://www.rfc-editor.org/rfc/rfc5077) based design that we believe addresses some theoretical concerns (https://github.com/rustls/rustls/issues/2023) with the design of the ring specific `Ticketer`, using APIs only available with `aws-lc-rs`.

For background, the ring ticketer is designed around an AEAD using random nonces. We must use an AEAD with `*ring*` because it doesn't expose an unauthenticated cipher/modes we could use for the RFC 5077 design. We must use random nonces with the AEAD design to avoid privacy leaks. The combination of these two constraints results in a design that has a more limited security margin (due to the size of the nonces and the implications of re-use).

This commit adds an `aws-lc-rs` Ticketer implementation that is a direct translation of the [RFC 5077 "Recommended Ticket Construction"](https://www.rfc-editor.org/rfc/rfc5077#section-4). It uses AES-256 in CBC mode w/ PKCS#7 padding for encryption and HMAC-SHA-256 for message authentication.

Resolves https://github.com/rustls/rustls/issues/2021

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/ticketer.rs
+++ b/rustls/src/crypto/aws_lc_rs/ticketer.rs
@@ -5,11 +5,14 @@ use core::fmt;
 use core::fmt::{Debug, Formatter};
 use core::sync::atomic::{AtomicUsize, Ordering};
 
-use subtle::ConstantTimeEq;
+use aws_lc_rs::cipher::{
+    DecryptionContext, PaddedBlockDecryptingKey, PaddedBlockEncryptingKey, UnboundCipherKey,
+    AES_256, AES_256_KEY_LEN, AES_CBC_IV_LEN,
+};
+use aws_lc_rs::{hmac, iv};
 
-use super::ring_like::aead;
 use super::ring_like::rand::{SecureRandom, SystemRandom};
-use super::TICKETER_AEAD;
+use super::unspecified_err;
 use crate::error::Error;
 #[cfg(feature = "logging")]
 use crate::log::debug;
@@ -24,8 +27,10 @@ impl Ticketer {
     /// Make the recommended `Ticketer`.  This produces tickets
     /// with a 12 hour life and randomly generated keys.
     ///
-    /// The encryption mechanism used is injected via `TICKETER_AEAD`;
-    /// it must take a 256-bit key and 96-bit nonce.
+    /// The `Ticketer` uses the [RFC 5077 §4] "Recommended Ticket Construction",
+    /// using AES 256 for encryption and HMAC-SHA256 for ciphertext authentication.
+    ///
+    /// [RFC 5077 §4]: https://www.rfc-editor.org/rfc/rfc5077#section-4
     #[cfg(feature = "std")]
     pub fn new() -> Result<Arc<dyn ProducesTickets>, Error> {
         Ok(Arc::new(crate::ticketer::TicketSwitcher::new(
@@ -37,8 +42,10 @@ impl Ticketer {
     /// Make the recommended `Ticketer`.  This produces tickets
     /// with a 12 hour life and randomly generated keys.
     ///
-    /// The encryption mechanism used is injected via `TICKETER_AEAD`;
-    /// it must take a 256-bit key and 96-bit nonce.
+    /// The `Ticketer` uses the [RFC 5077 §4] "Recommended Ticket Construction",
+    /// using AES 256 for encryption and HMAC-SHA256 for ciphertext authentication.
+    ///
+    /// [RFC 5077 §4]: https://www.rfc-editor.org/rfc/rfc5077#section-4
     #[cfg(not(feature = "std"))]
     pub fn new<M: crate::lock::MakeMutex>(
         time_provider: &'static dyn TimeProvider,
@@ -52,47 +59,62 @@ impl Ticketer {
 }
 
 fn make_ticket_generator() -> Result<Box<dyn ProducesTickets>, GetRandomFailed> {
-    Ok(Box::new(AeadTicketer::new()?))
+    // NOTE(XXX): Unconditionally mapping errors to `GetRandomFailed` here is slightly
+    //   misleading in some cases (e.g. failure to construct a padded block cipher encrypting key).
+    //   However, we can't change the return type expected from a `TicketSwitcher` `generator`
+    //   without breaking semver.
+    //   Tracking in https://github.com/rustls/rustls/issues/2074
+    Ok(Box::new(
+        Rfc5077Ticketer::new().map_err(|_| GetRandomFailed)?,
+    ))
 }
 
-/// This is a `ProducesTickets` implementation which uses
-/// any *ring* `aead::Algorithm` to encrypt and authentication
-/// the ticket payload.  It does not enforce any lifetime
-/// constraint.
-struct AeadTicketer {
-    alg: &'static aead::Algorithm,
-    key: aead::LessSafeKey,
+/// An RFC 5077 "Recommended Ticket Construction" implementation of a [`Ticketer`].
+struct Rfc5077Ticketer {
+    aes_encrypt_key: PaddedBlockEncryptingKey,
+    aes_decrypt_key: PaddedBlockDecryptingKey,
+    hmac_key: hmac::Key,
     key_name: [u8; 16],
     lifetime: u32,
-
-    /// Tracks the largest ciphertext produced by `encrypt`, and
-    /// uses it to early-reject `decrypt` queries that are too long.
-    ///
-    /// Accepting excessively long ciphertexts means a "Partitioning
-    /// Oracle Attack" (see <https://eprint.iacr.org/2020/1491.pdf>)
-    /// can be more efficient, though also note that these are thought
-    /// to be cryptographically hard if the key is full-entropy (as it
-    /// is here).
     maximum_ciphertext_len: AtomicUsize,
 }
 
-impl AeadTicketer {
-    fn new() -> Result<Self, GetRandomFailed> {
-        let mut key = [0u8; 32];
-        SystemRandom::new()
-            .fill(&mut key)
+impl Rfc5077Ticketer {
+    fn new() -> Result<Self, Error> {
+        let rand = SystemRandom::new();
+
+        // Generate a random AES 256 key to use for AES CBC encryption.
+        let mut aes_key = [0u8; AES_256_KEY_LEN];
+        rand.fill(&mut aes_key)
             .map_err(|_| GetRandomFailed)?;
 
-        let key = aead::UnboundKey::new(TICKETER_AEAD, &key).unwrap();
+        // Convert the raw AES 256 key bytes into encrypting and decrypting keys using CBC mode and
+        // PKCS#7 padding. We don't want to store just the raw key bytes as constructing the
+        // cipher keys has some setup overhead. We can't store just the `UnboundCipherKey` since
+        // constructing the padded encrypt/decrypt specific types consume the `UnboundCipherKey`.
+        let aes_encrypt_key =
+            UnboundCipherKey::new(&AES_256, &aes_key[..]).map_err(unspecified_err)?;
+        let aes_encrypt_key =
+            PaddedBlockEncryptingKey::cbc_pkcs7(aes_encrypt_key).map_err(unspecified_err)?;
 
+        // Convert the raw AES 256 key bytes into a decrypting key using CBC PKCS#7 padding.
+        let aes_decrypt_key =
+            UnboundCipherKey::new(&AES_256, &aes_key[..]).map_err(unspecified_err)?;
+        let aes_decrypt_key =
+            PaddedBlockDecryptingKey::cbc_pkcs7(aes_decrypt_key).map_err(unspecified_err)?;
+
+        // Generate a random HMAC SHA256 key to use for HMAC authentication.
+        let hmac_key = hmac::Key::generate(hmac::HMAC_SHA256, &rand).map_err(unspecified_err)?;
+
+        // Generate a random key name.
         let mut key_name = [0u8; 16];
-        SystemRandom::new()
-            .fill(&mut key_name)
+        rand.fill(&mut key_name)
             .map_err(|_| GetRandomFailed)?;
 
-        Ok(AeadTicketer {
-            alg: TICKETER_AEAD,
-            key: aead::LessSafeKey::new(key),
+        Ok(Self {
+            aes_encrypt_key,
+            aes_decrypt_key,
+            hmac_key,
             key_name,
             lifetime: 60 * 60 * 12,
             maximum_ciphertext_len: AtomicUsize::new(0),
@@ -100,7 +122,7 @@ impl AeadTicketer {
     }
 }
 
-impl ProducesTickets for AeadTicketer {
+impl ProducesTickets for Rfc5077Ticketer {
     fn enabled(&self) -> bool {
         true
     }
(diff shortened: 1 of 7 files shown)


rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -0,0 +1,354 @@
...
+        // is overall similar to [^2].
+        //
+        // [^1]: https://eprint.iacr.org/2020/1491.pdf
+        // [^2]: "Authenticated Encryption with Key Identification", fig 6
+        //       <https://eprint.iacr.org/2022/1680.pdf>
+        if ConstantTimeEq::ct_ne(&self.key_name[..], alleged_key_name).into() {
+            #[cfg(debug_assertions)]
+            debug!("rejected ticket with wrong ticket_name");
+            return None;
+        }
```
cpu, line 193: I wasn't 100% sure how applicable this ticket name "defense in depth" strategy is for the RFC 5077 construction, so I left it (and the `maximum_ciphertext_len` field) in-place.

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -0,0 +1,354 @@
...
+
+        // Split off the key name from the remaining ciphertext.
+        let (alleged_key_name, ciphertext) = try_split_at(ciphertext, self.key_name.len())?;
+
+        // checking the key_name is the expected one, *and* then putting it into the
+        // additionally authenticated data is duplicative.  this check quickly rejects
+        // tickets for a different ticketer (see `TicketSwitcher`), while including it
+        // in the AAD ensures it is authenticated independent of that check and that
```
ctz, line 181:
```suggestion
        // in the hmac input ensures it is authenticated independent of that check and that
```

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -0,0 +1,354 @@
...
+        // any attempted attack on the integrity such as [^1] must happen for each
+        // `key_label`, not over a population of potential keys.  this approach
+        // is overall similar to [^2].
+        //
+        // [^1]: https://eprint.iacr.org/2020/1491.pdf
+        // [^2]: "Authenticated Encryption with Key Identification", fig 6
+        //       <https://eprint.iacr.org/2022/1680.pdf>
```
ctz, line 188: I would remove this block as this construction is key-committing.

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// Encrypt `message` and return the ciphertext.
+    fn encrypt(&self, message: &[u8]) -> Option<Vec<u8>> {
+        // Convert the raw AES 128 key bytes into an encrypting key using CBC PKCS#7 padding.
+        let aes_key = cipher::UnboundCipherKey::new(&cipher::AES_128, &self.aes_key[..]).ok()?;
+        let aes_key = cipher::PaddedBlockEncryptingKey::cbc_pkcs7(aes_key).ok()?;
+
+        // Encrypt the ticket state - the cipher module handles generating a random IV of
+        // appropriate size, returning it in the `DecryptionContext`.
+        let mut encrypted_state = Vec::from(message);
+        let dec_ctx = aes_key
```
ctz, line 121: 👺 if you ignore the docs and trace down to what this actually does, yes, it makes and returns a random IV

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    fn lifetime(&self) -> u32 {
+        self.lifetime
+    }
+
+    /// Encrypt `message` and return the ciphertext.
+    fn encrypt(&self, message: &[u8]) -> Option<Vec<u8>> {
+        // Convert the raw AES 128 key bytes into an encrypting key using CBC PKCS#7 padding.
+        let aes_key = cipher::UnboundCipherKey::new(&cipher::AES_128, &self.aes_key[..]).ok()?;
+        let aes_key = cipher::PaddedBlockEncryptingKey::cbc_pkcs7(aes_key).ok()?;
```
ctz, line 116: Can we keep this instead of `self.aes_key` to avoid paying the key schedule costs for each encryption/decryption? (probably means keeping both a `PaddingBlockEncryptingKey` and a `PaddedBlockDecryptingKey`)

rustls/src/crypto/aws_lc_rs/ticketer.rs
ctz: See comments above. I think the `maximum_ciphertext_len` is a good DoS/shenanigans gate irrespective.

rustls/src/crypto/mod.rs
```diff
@@ -574,6 +574,15 @@ pub fn default_fips_provider() -> CryptoProvider {
     aws_lc_rs::default_provider()
 }
 
+/// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
```
ctz, line 577: I wonder if this is better kept in a new `src/polyfill.rs`, with a link to the upstream stabilisation issue? (For this one, https://github.com/rust-lang/rust/issues/119128)

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -0,0 +1,354 @@
+use alloc::boxed::Box;
```
djc, line 1: How do you feel about making an intermediate commit that copies the ring ticketer into this file, so that this commit becomes a diff between the old ticketer and the new ticketer? Not sure how much code they would share but that's kind of the point.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: This fell away.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Done :+1:

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Yeah :-( I ended up spelunking into the impl to determine this too. To be fair, using the `cipher` module is discouraged in general. I'll think about suggesting an upstream docs update.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Done.

rustls/src/crypto/mod.rs
cpu: Done.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Done.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: https://github.com/aws/aws-lc-rs/pull/488

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -21,11 +21,13 @@ use crate::server::ProducesTickets;
...
 
 impl Ticketer {
-    /// Make the recommended Ticketer.  This produces tickets
+    /// Make the recommended `Ticketer`.  This produces tickets
     /// with a 12 hour life and randomly generated keys.
     ///
-    /// The encryption mechanism used is injected via TICKETER_AEAD;
-    /// it must take a 256-bit key and 96-bit nonce.
+    /// The `Ticketer` uses the [RFC 5077 §4] "Recommended Ticket Construction",
+    /// using AES 128 for encryption and HMAC-SHA256 for ciphertext authentication.
```
ctz, line 28: Apologies, should've seen this in the previous review: let's use AES-256. The performance difference is not significant, but it feels unbalanced to use AES-128 to protect stronger keys even if the cryptographic strengths are so large that the difference is not meaningful in brute force settings. (Maybe a bit more meaningful in multi-key and post-quantum settings.)

Tried to comment all the places where there are mentions of AES-128 below, but there are also some in the commit message.

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -34,10 +36,13 @@ impl Ticketer {
         )?))
     }
 
-    /// Make the recommended Ticketer.  This produces tickets
+    /// Make the recommended `Ticketer`.  This produces tickets
     /// with a 12 hour life and randomly generated keys.
     ///
-    /// The encryption mechanism used is Chacha20Poly1305.
+    /// The `Ticketer` uses the [RFC 5077 §4] "Recommended Ticket Construction",
+    /// using AES 128 for encryption and HMAC-SHA256 for ciphertext authentication.
```
ctz, line 43: AES 128> ditto

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -51,97 +56,124 @@ impl Ticketer {
...
     maximum_ciphertext_len: AtomicUsize,
 }
 
-impl ProducesTickets for AeadTicketer {
+impl Rfc5077 {
+    fn new() -> Result<Self, Error> {
+        let rand = SystemRandom::new();
+
+        // Generate a random AES 128 key to use for AES CBC encryption.
+        let mut aes_key = [0u8; cipher::AES_128_KEY_LEN];
```
ctz, line 81: AES 128> ditto

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -51,97 +56,124 @@ impl Ticketer {
...
+        rand.fill(&mut aes_key)
+            .map_err(|_| GetRandomFailed)?;
+
+        // Convert the raw AES 128 key bytes into encrypting and decrypting keys using CBC mode and
```
ctz, line 85: AES 128> ditto

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -51,97 +56,124 @@ impl Ticketer {
...
+        // PKCS#7 padding. We don't want to store just the raw key bytes as constructing the
+        // cipher keys has some setup overhead. We can't store just the `UnboundCipherKey` since
+        // constructing the padded encrypt/decrypt specific types consume the `UnboundCipherKey`.
+        let aes_encrypt_key = cipher::UnboundCipherKey::new(&cipher::AES_128, &aes_key[..])
+            .map_err(unspecified_err)?;
+        let aes_encrypt_key = cipher::PaddedBlockEncryptingKey::cbc_pkcs7(aes_encrypt_key)
+            .map_err(unspecified_err)?;
+
+        // Convert the raw AES 128 key bytes into a decrypting key using CBC PKCS#7 padding.
+        let aes_decrypt_key = cipher::UnboundCipherKey::new(&cipher::AES_128, &aes_key[..])
```
ctz, line 95: AES 128> ditto

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Switched to AES 256 throughout (including commit message & PR desc)

rustls/src/lib.rs
```diff
@@ -650,6 +650,9 @@ pub mod time_provider;
 /// APIs abstracting over locking primitives.
 pub mod lock;
 
+/// Pollyfills for features that are not yet stabilized or available with current MSRV.
```
ctz, line 653:
```suggestion
/// Polyfills for features that are not yet stabilized or available with current MSRV.
```

ctz (approved):

cpu:

There's an instruction count increase for the `handshake_tickets_*` tests for aws-lc-rs with this ticketer, but the wall-time benchmarks show no difference. I usually think of the latter as more reliable as the former. Does that track with your thinking here @ctz?

ctz:
Yes, I think the walltime ones are generally more representative. But I'm not surprised that this is a little slower -- quite a lot of area is dedicated now to making AES-GCM go fast.

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -51,97 +56,124 @@ impl Ticketer {
...
-        .map_err(|_| GetRandomFailed)?;
-
-    Ok(Box::new(AeadTicketer {
-        alg: TICKETER_AEAD,
-        key: aead::LessSafeKey::new(key),
-        key_name,
-        lifetime: 60 * 60 * 12,
-        maximum_ciphertext_len: AtomicUsize::new(0),
-    }))
+    // NOTE(XXX): Unconditionally mapping errors to `GetRandomFailed` here is slightly
```
djc, line 59: File a follow-up issue or PR with the breaking change label?

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -51,97 +56,124 @@ impl Ticketer {
...
 
-/// This is a `ProducesTickets` implementation which uses
-/// any *ring* `aead::Algorithm` to encrypt and authentication
-/// the ticket payload.  It does not enforce any lifetime
-/// constraint.
-struct AeadTicketer {
-    alg: &'static aead::Algorithm,
-    key: aead::LessSafeKey,
+/// An RFC 5077 "Recommended Ticket Construction" implementation of a [`Ticketer`].
+struct Rfc5077 {
```
djc, line 67: Bikeshedding: I'd find `Rfc5077Ticketer` a clearer name for this.

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -5,11 +5,11 @@ use core::fmt;
...
 use core::sync::atomic::{AtomicUsize, Ordering};
 
-use subtle::ConstantTimeEq;
+use aws_lc_rs::{cipher, hmac, iv};
 
-use super::ring_like::aead;
 use super::ring_like::rand::{SecureRandom, SystemRandom};
-use super::TICKETER_AEAD;
+use super::unspecified_err;
+
```
djc, line 12: Nit: I think we usually keep `super` and `crate` imports in the same block?

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -51,97 +56,124 @@ impl Ticketer {
...
-/// This is a `ProducesTickets` implementation which uses
-/// any *ring* `aead::Algorithm` to encrypt and authentication
-/// the ticket payload.  It does not enforce any lifetime
-/// constraint.
-struct AeadTicketer {
-    alg: &'static aead::Algorithm,
-    key: aead::LessSafeKey,
+/// An RFC 5077 "Recommended Ticket Construction" implementation of a [`Ticketer`].
+struct Rfc5077 {
+    aes_encrypt_key: cipher::PaddedBlockEncryptingKey,
```
djc, line 68: What's the value of keeping this type, and a bunch of the other stuff used in `new()`, prefixed with `cipher`?

(For example, I can see the value for using `hmac::Key` but these other identifiers seem pretty specific/unambiguous.)

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -153,48 +185,52 @@ impl ProducesTickets for AeadTicketer {
...
 }
 
-impl Debug for AeadTicketer {
+impl Debug for Rfc5077 {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        // Note: we deliberately omit the key from the debug output.
-        f.debug_struct("AeadTicketer")
-            .field("alg", &self.alg)
+        // Note: we deliberately omit keys from the debug output.
+        f.debug_struct("Rfc5077Ticketer")
```
djc, line 233: The type name is currently inconsistent with the code.

djc (approved):
LGTM!

There are a number of little things that the new implementation cleans up that we could perhaps backport to the ring ticketer, like treating the `Ticketer` name in doc comment like code, adding an empty line between methods, and moving the initialization of the `Ticketer` type into a `new()` method. Might be nice to include those in the ring `Ticketer` before it gets copied? On the other hand, might not be very high value.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Done: https://github.com/rustls/rustls/issues/2074

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Yeah, in retrospect I agree. You can see based on the `Debug` impl falling out of sync I waffled on this a bit. My thinking with using `Rfc5077` was to avoid stutter from the module name but since this is an internal type it's never referenced by the name `ticketer::Rfc5077` and `Rfc5077Ticketer` seems better. I switched back to that name :+1:

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Fixed up.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: Dropped the `cipher::` prefixes.

rustls/src/crypto/aws_lc_rs/ticketer.rs
cpu: This fell away with the return to using `Rfc5077Ticketer` as the name.

cpu:
> There are a number of little things that the new implementation cleans up that we could perhaps backport to the ring ticketer, like treating the Ticketer name in doc comment like code, adding an empty line between methods, and moving the initialization of the Ticketer type into a new() method. Might be nice to include those in the ring Ticketer before it gets copied

Good call-out. It feels nice to port some of those improvements backwards to the original impl. I went ahead and did that before the copy over into `aws-lc-rs`.

PR #2073, opened 2024-08-05
Remove unused code in `ServerConnection::new`

DragonDev1906:
Looks like this was a result of a merge, as `common` isn't used anywhere and the same code exists in `ConnectionCore::for_server`.

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -564,9 +564,6 @@ mod connection {
         /// Make a new ServerConnection.  `config` controls how
         /// we behave in the TLS protocol.
         pub fn new(config: Arc<ServerConfig>) -> Result<Self, Error> {
-            let mut common = CommonState::new(Side::Server);
-            common.set_max_fragment_size(config.max_fragment_size)?;
-            common.enable_secret_extraction = config.enable_secret_extraction;
             Ok(Self {
                 inner: ConnectionCommon::from(ConnectionCore::for_server(config, Vec::new())?),
             })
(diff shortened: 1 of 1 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -564,9 +564,6 @@ mod connection {
         /// Make a new ServerConnection.  `config` controls how
         /// we behave in the TLS protocol.
         pub fn new(config: Arc<ServerConfig>) -> Result<Self, Error> {
-            let mut common = CommonState::new(Side::Server);
```
ctz, line 567: Thanks -- good spot! There is also another instance of this in `UnbufferedServerConnection::new` -- care to eliminate that one too?

ctz (approved):
Thanks!

rustls/src/server/server_conn.rs
```diff
@@ -564,9 +564,6 @@ mod connection {
...
-            common.set_max_fragment_size(config.max_fragment_size)?;
```
djc, line 568: Do we understand why lints didn't fire against this?

djc (approved):

rustls/src/server/server_conn.rs
ctz: It seems like `unused_assignments` only works for locals, not slots in arrays, tuples or structs. https://github.com/rust-lang/rust/issues/65467#issuecomment-994312394

I would have expected the assignment to `enable_secret_extraction` on a value that is immediately dropped to warn at least. The `set_max_fragment_size` line is OK as it potentially alters control flow.

cpu (approved):

PR #2075, opened 2024-08-07
Experiment: allow state machine to become blocked by extension point traits

ctz:
This is an experiment about how we could allow extension point traits called within the state machine to not produce their results immediately.

While that happens, the state machine is blocked, and attempts to move it forward return a non-fused error (in the buffered API) or a new `ConnectionState` variant (in the unbuffered API -- NYI).

When the application sees that, they have several options:

1. just blindly retry -- obvious, but also not helpful,
2. learn when the underlying operation completes, and then retry

There is a reasonable amount of plumbing required to make (2) work. Basically, the verifier (or whatever) and application need to know more about each other than they currently do.

Then the application could do:

```rust
match conn.process_new_packets() {
  Err(Error::PendingOperation(PendingOperation::ServerCertificateVerification)) => {
    rustls_platform_verifier::completion_future().await;
    break 'retry;
  },
...
}
```
 
A context-free `completion_future` like that would need to know which `ServerCertVerifier` impl it is concerned with. That could be done with thread local storage.

Alternatively the application could retain the concrete verifier used and ask it directly (not currently possible with `rustls-platform-verifier` because it hides its types behind the trait, but IMO a feasible API change.)

The extension point traits I'm thinking are in scope are: `Signer`, `ServerCertVerifier` and `ClientCertVerifier` for sure. Perhaps `ResolvesClientCert` and `ResolvesServerCert` (but I think `Acceptor` is a better solution) as a follow up. `ProducesTickets`, `StoresServerSessions`, `ClientSessionStore`, could theoretically come later but given the high performance requirements on those, I'd suggest a synchronous facade on top of a asynchronous backing store.

Things not called inside the state machine are not in scope: that includes all the low-level provider interfaces. I think the work there is a long tail, and needs compiler support for turning the whole library async -- this would require using things like RPITIT and AFIT.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -13,7 +13,7 @@ use super::hs::ClientContext;
 use crate::check::{inappropriate_handshake_message, inappropriate_message};
 use crate::client::common::{ClientAuthDetails, ServerCertDetails};
 use crate::client::{hs, ClientConfig};
-use crate::common_state::{CommonState, HandshakeKind, KxState, Side, State};
+use crate::common_state::{CommonState, HandshakeKind, KxState, Side, State, Transition};
 use crate::conn::ConnectionRandoms;
 use crate::crypto::KeyExchangeAlgorithm;
 use crate::enums::{AlertDescription, ContentType, HandshakeType, ProtocolVersion};
@@ -143,7 +143,7 @@ mod server_hello {
                     let sig_verified = verify::HandshakeSignatureValid::assertion();
 
                     return if must_issue_new_ticket {
-                        Ok(Box::new(ExpectNewTicket {
+                        Ok(Transition::Next(Box::new(ExpectNewTicket {
                             config: self.config,
                             secrets,
                             resuming_session: Some(resuming),
@@ -154,9 +154,9 @@ mod server_hello {
                             resuming: true,
                             cert_verified,
                             sig_verified,
-                        }))
+                        })))
                     } else {
-                        Ok(Box::new(ExpectCcs {
+                        Ok(Transition::Next(Box::new(ExpectCcs {
                             config: self.config,
                             secrets,
                             resuming_session: Some(resuming),
@@ -168,13 +168,13 @@ mod server_hello {
                             resuming: true,
                             cert_verified,
                             sig_verified,
-                        }))
+                        })))
                     };
                 }
             }
 
             cx.common.handshake_kind = Some(HandshakeKind::Full);
-            Ok(Box::new(ExpectCertificate {
+            Ok(Transition::Next(Box::new(ExpectCertificate {
                 config: self.config,
                 resuming_session: None,
                 session_id: server_hello.session_id,
@@ -185,7 +185,7 @@ mod server_hello {
                 suite,
                 may_send_cert_status,
                 must_issue_new_ticket,
-            }))
+            })))
         }
     }
 }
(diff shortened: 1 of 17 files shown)


cpu:
@ctz Is this something you're looking for early feedback on, or should reviewers wait until it's further along and CI is passing?

ctz:
> @ctz Is this something you're looking for early feedback on, or should reviewers wait until it's further along and CI is passing?

I think I'd like some feedback on the approach (so, i guess, just 07906b6cdf7045dbb0e0530dd45e81888b1288a6), but not any of the "wip" or "tmp" commits at this point.

djc:
After reading the approach commit and the OP I'm still a little confused about how the state machine can resume. I'm wondering how we might design a type-safe approach to doing these stuff. Personally, I still like the idea of expanding `Box<State>` into an enum with the specific states, after which we could (a) have a low-level API mode that externalizes this `State` enum type, and (b) add in new states that allow the caller to move to the new state. I suppose this would work in tandem with the unbuffered API, so that you couldn't do this kind of thing with the buffered API (but we could provide utilities that provide a default implementation of buffer handling).

rustls/src/client/tls12.rs
```diff
@@ -838,16 +838,15 @@ impl State<ClientConnectionData> for ExpectServerDone<'_> {
         let suite = st.suite;
 
         // 1. Verify the cert chain.
-        // 2. Verify any SCTs provided with the certificate.
-        // 3. Verify that the top certificate signed their kx.
-        // 4. If doing client auth, send our Certificate.
-        // 5. Complete the key exchange:
+        // 2. Verify that the top certificate signed their kx.
```
cpu: This commit seems like one that could be pulled out and landed separate from the async work

rustls/src/common_state.rs
```diff
@@ -845,6 +855,30 @@ pub(crate) trait State<Data>: Send + Sync {
     fn into_owned(self: Box<Self>) -> Box<dyn State<Data> + 'static>;
 }
 
+/// This is an output of [`State::handle()`]
+///
+/// It contains the next step in the state machine.
```
cpu, line 860: nit: seems like this comment should also indicate it may contain the reason the state is blocked on the current state as well as mentioning it contains the next state.

rustls/src/common_state.rs
```diff
@@ -845,6 +855,30 @@ pub(crate) trait State<Data>: Send + Sync {
...
+    Next(Box<dyn State<Data> + 'a>),
+
+    /// The current state is blocked.
+    ///
+    /// `current` is the current state; returned so it is not consumed.
+    /// Likewise, `message` is the current message.
+    /// `why` describes the operation that it is waiting on.
+    ///
+    /// `handle()` should be called again with the same message.  It will
+    /// make progress when possible.
```
cpu, line 874: Perhaps this is a place to discuss the more involved decision that could be done based on `why` to decide when to call `handle()` again?

rustls/src/error.rs
```diff
@@ -102,6 +102,12 @@ pub enum Error {
     /// [`keys_match`]: crate::crypto::signer::CertifiedKey::keys_match
     InconsistentKeys(InconsistentKeys),
 
+    /// A asynchronous operation is pending.
```
cpu, line 105: very small nit: A -> An

rustls/src/error.rs
```diff
@@ -102,6 +102,12 @@ pub enum Error {
...
+    ///
+    /// Once the relevant operation completes, re-try this call to
+    /// advance the handshake.
```
cpu, line 108: nit: I feel like "this call" might be ambiguous. Perhaps "re-try the operation that produced the `PendingOperation` error?" too verbose?

cpu (commented):
Some very minor comments I had while reviewing the diff to get a sense of the direction. I tried not to give feedback on the more WIP-commits.

cpu:
> After reading the approach commit and the OP I'm still a little confused about how the state machine can resume.

I feel a bit similar. I think the mechanisms getting started in the branch are fairly approachable, but I'm not sure I see the big picture yet. Perhaps it would be easier to understand if there was an external verifier impl (platform verifier or otherwise) updated to take advantage of this? Or some other more complete end-to-end demo? I know that's asking for a lot when the implementation in the library is in flux but I think it would help me better grasp the overall approach.

ctz:
Going to close this for now. I am a little concerned that doing something _weird_ like this is the wrong direction. It is kinda unclear what the status of the "keyword generics initiative" is, but if it ever comes to fruition (and assuming async works fully in traits by then...), that seems like vastly the preferable direction. Note that doesn't preclude making state machine representation changes in the mean time, which should be complementary.

PR #2077, opened 2024-08-08
Optimise crate size

### LINKED ISSUE #2076: Test data shouldn't be included in the crates.io releases
**Checklist**

**Describe the bug**
I recently skimmed over the changes  of a recent rustls update and noticed that there are several non-code test files included in the published packages:

That includes: 
* The testdata directory: https://diff.rs/rustls/0.22.4/0.23.12/src%2Ftestdata
* The binary representation of the ffdhe_groups: https://diff.rs/rustls/0.22.4/0.23.12/src%2Fmsgs%2Fffdhe_groups

**To Reproduce**
1. Download a release from crates.io and check the crates file for these folders

**Applicable Version(s)**

0.23.12

**Expected behavior**

For the testdata I would just expect that they are not included. They only seem to be needed for running the tests, which is not required for the downloaded crate from crates.io

For the ffdhe_groups I can see why they are needed. I personally would prefer defining them as constants in the rust code matching the relevant definitions in the linked RFC by using the hex representation. That would make it easier to verify that these constants are correct.

**Additional context**

With the recent xz accident such files became suspicious, although I explicitly do not want to imply that they are any problem.
Additionally these files  increase the size of the downloaded code.

ctz:
fixes #2076

This takes rustls.crate from 434KB to 334KB, which translates to saving crates.io (and, indeed, downloaders) about 50 GB per day.


rustls/src/msgs/ffdhe_groups.rs
```diff
@@ -60,39 +60,244 @@ impl<'a> FfdheGroup<'a> {
 ///
 /// [RFC 7919 Appendix A.1]: https://datatracker.ietf.org/doc/html/rfc7919#appendix-A.1
 pub const FFDHE2048: FfdheGroup<'static> = FfdheGroup {
-    p: include_bytes!("ffdhe_groups/ffdhe2048-modulus.bin"),
+    p: &[
```
ctz, line 63: On this commit, I checked we had working test coverage by introducing a one-bit error into all of these locally, and checking the validation caught it. It did.

djc (approved):

Issue #1920, opened 2024-04-25
Ensuring that a provider based on the one built-in is used


stormshield-gt:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**

I work on a very big project with multiple applications that uses rustls and which dependencies also use rustls. We have a custom provider based on aws-lc-rs, that we want to be used by all of our crates.
Besides, some of our dependency relies on a static provider has been installed in the current process.

My problem is sometimes, when new codes that need tls is added, the original aws-lc-rs is silently used instead of our custom one.

For instance, if in a crate someone does `rustls_platform_verifier::tls_config` without installing our custom provider before, no error will be thrown and this will use the classic aws-lc-rs provider

I thought how I could detect this with some tooling, like clippy or cargo deny and I didn't figure it out.

**Describe the solution you'd like**

I would like that ring and aws-lc-rs provider to be moved on a separate crate. They could still be added as dependency to ruslts under feature flag to keep the current behavior.

If aws-lc-rs provider were in a separated crate, I could use this crate directly for my custom provider. I could then deny the feature aws-lc-rs and ring with cargo deny

```toml
[[bans.features]]
name = "rustls"
deny = ["aws-lc-rs","aws_lc_rs", "ring"]
reason = "we use our own provider"
``` 
Then if somebody forgets to install the custom provider, it will be caught thanks to a panic in `get_default_or_install_from_crate_features()` . This can work right now for providers that aren't based on ring or aws-lc-rs, but in my case I must activate the aws-lc-rs feature so I can base my custom provider on it.

**Describe alternatives you've considered**

#### Option A: force users to always use `builder_with_provider`

I can tell users to not use the process default provider and add this to my clippy.toml:

```toml
disallowed-methods = [
    { path = "rustls::ClientConfig::builder", reason = "can silently use a wrong provider" },
    { path = "rustls::ServerConfig::builder", reason = "can silently use a wrong provider" },
    # But those methodes can also be called transitively..
    { path = "rustls_platform_verifier::tls_config", reason = "can silently use a wrong provider" },  
]
```
The problem is that I must also scan dependencies that my use the default provider and so each time a new version is issued.

Besides we have a patched dependency which relies on the process default because it makes the patch a lot simpler.

#### Option B: reexport rustls and dependencies that use rustls under an inner crate
 
Another solution is to have an inner crate that has rustls has a private dependency and force use to use my crate with cargo deny

```toml
deny = [
[...]

Issue #1961, opened 2024-05-20
Clean up crate feature naming


ctz:
- We should remove the `logging` feature, and instead use the `log` implicit feature.
- We should remove the `aws_lc_rs` feature, and instead use the `aws-lc-rs` implicit feature.

stormshield-gt:
I think it worth mentioning that [2024 edition will remove implicit features](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/cargo-remove-implicit-features.html)

djc:
Yes, although I think we should use explicit features bearing the exact name of the dependency.

PR #2082, opened 2024-08-14
Crypto: add feature to disable installation of built-in provider

stormshield-gt:
Fix https://github.com/rustls/rustls/issues/1920 , especially the problematic described in https://github.com/rustls/rustls/issues/1920#issuecomment-2255255668 .

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -40,6 +40,7 @@ aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs"]
 aws-lc-rs = ["aws_lc_rs"] # Alias because Cargo features commonly use `-`
 brotli = ["dep:brotli", "dep:brotli-decompressor", "std"]
 ring = ["dep:ring", "webpki/ring"]
+custom-provider = []
 tls12 = []
 read_buf = ["rustversion", "std"]
 fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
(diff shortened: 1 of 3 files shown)


ctz:
Addressing the coverage issue in #2083 (preexisting issue on main).

ctz:
This looks reasonable to me.

It would be good to add some documentation for this to https://rustls.github.io/rustls/prerelease/#crate-features, and perhaps a short mention in https://rustls.github.io/rustls/prerelease/crypto/struct.CryptoProvider.html#using-the-per-process-default-cryptoprovider .

stormshield-gt:
I added some comments where indicated, I'm not a native English speaker, so the wording might not be perfect.

rustls/Cargo.toml
```diff
@@ -40,6 +40,7 @@ aws_lc_rs = ["dep:aws-lc-rs", "webpki/aws_lc_rs"]
 aws-lc-rs = ["aws_lc_rs"] # Alias because Cargo features commonly use `-`
 brotli = ["dep:brotli", "dep:brotli-decompressor", "std"]
 ring = ["dep:ring", "webpki/ring"]
+custom_provider = []
```
djc, line 43: I think we wanted to standardize on dashes over underscores in feature names now?

djc (approved):

rustls/Cargo.toml
stormshield-gt: https://github.com/rustls/rustls/issues/1961 seems to go in that direction but don't mention the `read_buf` feature so I was confused

djc: Probably still use a dash here.

stormshield-gt: I've changed it to a dash in the latest push ;)

djc:
@ctz do you want to take another look at this?

ctz (approved):
Looks good, thanks!

PR #2086, opened 2024-08-21
Expose UnknownExtensions fields & to use in projects.

gituser-rs:
Hey. I am working on a Anti-Bot project and I need to access unknown extension fields to check if some extension is present. 

For example:
```
if let ClientExtension::Unknown(UnknownExtension {
    typ: ExtensionType::Unknown(28),
    ..
}) = ext
{
    Some(true)
} else {
    None
}
```

These fields & enums is private so I hope we can make them public 😃

### DIFF
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -441,7 +441,7 @@ pub mod internal {
         }
         pub mod enums {
             pub use crate::msgs::enums::{
-                AlertLevel, Compression, EchVersion, HpkeAead, HpkeKdf, HpkeKem, NamedGroup,
+                AlertLevel, Compression, EchVersion, HpkeAead, HpkeKdf, HpkeKem, NamedGroup, ExtensionType
             };
         }
         pub mod fragmenter {
@@ -452,7 +452,7 @@ pub mod internal {
                 CertificateChain, ClientExtension, ClientHelloPayload, DistinguishedName,
                 EchConfigContents, EchConfigPayload, HandshakeMessagePayload, HandshakePayload,
                 HpkeKeyConfig, HpkeSymmetricCipherSuite, KeyShareEntry, Random, ServerName,
-                SessionId,
+                SessionId, UnknownExtension
             };
         }
         pub mod message {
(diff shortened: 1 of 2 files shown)


djc (approved):
Seems harmless, I guess?

ctz:
The `internals` module and everything inside is not really intended for use in anything other than our tests, and is outside the API that we check for semver-compatible changes. So if you are using this in a downstream crate, that will be extremely fragile and I suggest you vendor this crate (where you can make this change).

gituser-rs:
> The `internals` module and everything inside is not really intended for use in anything other than our tests, and is outside the API that we check for semver-compatible changes. So if you are using this in a downstream crate, that will be extremely fragile and I suggest you vendor this crate (where you can make this change).

That's the only change i need. I don't think creating fork for such type of change is a good idea tbh.

djc:
> The `internals` module and everything inside is not really intended for use in anything other than our tests, and is outside the API that we check for semver-compatible changes. So if you are using this in a downstream crate, that will be extremely fragile and I suggest you vendor this crate (where you can make this change).

To be fair, it also changes only pretty rarely so it might just work out okay for a long time (but, no guarantees -- might want to pin those dependencies).

ctz:
I think I've said my piece on this. Note that I still plan to return to #1475 at some point, and that would remove `UnknownExtension` and also make it structurally hard to reinstate later.

ctz (changes requested):
Please run `cargo fmt`

cpu:
I'm worried merging this will send the wrong message to consumers, both about the scope of the `rustls` project and the `internals` module. I think users wanting to interrogate low-level details of the handshake are better served by using a crate tailored for that (e.g. [tls-parser](https://github.com/rusticata/tls-parser)). Adding this capability, and then turning around and removing it in subsequent refactoring seems like it will only cause strife and prolong future discussions about exposing other properties beyond unknown extensions that occur with some frequency.

djc:
I'm okay with closing this, FWIW -- had forgotten about the extension refactoring work that will get rid of this. We should get #1475 merged already...

cpu:
Thank you for the contribution. Apologies that this isn't the outcome you were hoping for, but I think there are options available to you. You might also be interested in pursing this change with a fork like https://github.com/3andne/craftls that has more appetite for API changes tailored to more niche use-cases.

Issue #2068, opened 2024-08-02
`no-std` support for targets w/o atomics?


bjoernQ:
Since `0.23.0` there is support for `no-std/alloc` which works fine on targets supporting atomics. Some targets however don't have support for atomics and usually the `portable-atomic` crate is used for that. It seems all the dependencies of Rustls support that but Rustls itself needs `alloc::sync::Arc` which is not available without atomics.

Is supporting those targets desired? Would a PR adding that something you might consider?

ctz:
Could you say which targets these are?

bjoernQ:
Microcontrollers like ESP32-C2, ESP32-C3 and ESP32-S2 (other members of the ESP32-family come with support for atomics)

ctz:
Thanks! I could see two routes for this:

1. Globally, replace all uses of Arc with a crate-local Arc type alias. For these targets, that alias would be Rc. Remove all the Send+Sync bounds on everything for those targets. The result would be rustls types cannot be moved between threads, so multiple threads using rustls would each need their own *Config values, etc.
2. For these targets, depend on `portable-atomic-util` and use its `Arc` as a polyfill.

Sounds like you were proposing (2)?

bjoernQ:
Yes, my initial idea was (2)

cpu:
If we take the 2nd approach, is there a way to prevent regression in our CI matrix? How do other no-std projects handle that? Presently our no-std testing is done w/ `x86_64-unknown-none` and that since that platform supports atomics it seems like it would be tricky to catch mistakes in the polyfill.

bjoernQ:
> If we take the 2nd approach, is there a way to prevent regression in our CI matrix? How do other no-std projects handle that? Presently our no-std testing is done w/ `x86_64-unknown-none` and since that platform supports atomics it seems like it would be tricky to catch mistakes in the polyfill.

I see how optionally depending on the polyfill will make things tricky. Thinking about it makes option (1) more appealing

djc:
I worry that making trait bounds optional depending on targets will be a pretty big pain in the ass?

bjoernQ:
> I worry that making trait bounds optional depending on targets will be a pretty big pain in the ass?

Yes - I didn't even dare to think about that initially

bjoernQ:
[...]

PR #2087, opened 2024-08-23
remove Send + Sync not needed from client session storage field

brodycj:
While working on a proposal to support using `alloc::rc::Rc`, as discussed as a possible solution for #2068, I discovered a case of extra `Send + Sync` that I think is not needed & one more thing that would need to be updated. Of course I could remove this as part of my proposal for #2068, but would rather keep this kind of cleanup in a separate proposal & commit if possible.

FYI I did already get a green build of this proposal in my own fork.

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -248,7 +248,7 @@ pub struct ServerConfig {
     pub max_fragment_size: Option<usize>,
 
     /// How to store client sessions.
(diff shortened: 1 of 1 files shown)


djc (approved):
Since we don't have these bounds on any of the other `Arc<dyn>`s in the `ServerConfig` (or `ClientConfig`), I suppose this is fine.

ctz (approved):
Thanks!

(I missed it initially, but this is good because `StoresServerSessions` already has `Send + Sync` as trait bounds. That should be a clippy lint!)

PR #2088, opened 2024-08-23
ALT DRAFT RFC: add option to use alloc::rc::Rc for no-std targets w/o built-in atomic ops - INITIAL HACK

brodycj:
**STATUS:** initial draft with some ugly hacks & ugly naming - needs some cleanup & discussion before merging - can we find a way to break this update into smaller & cleaner pieces?

---

## WHAT

implement idea 1 discussed in: #2068 /cc @bjoernQ

with the following new feature~~s~~ (need to discuss & improve naming):

* `withrcalias` - use an internal alias to use `alloc::rc::Rc` instead of `alloc::sync::Arc`; this removes `Send + Sync` from many pub traits, making it impossible to share config state between threads
* ~~`defaultproviderenabled` - use this feature to enable default provider API functions, which will not work together with the `withrcalias` feature - TODO item is to add this to the default feature list (I encountered some issues with this, think I encountered this: <https://github.com/rust-lang/cargo/issues/7160#issuecomment-789314185>)~~

~~I decided to keep these as separate options as well, may make sense to consider combining them in some form / fashion.~~

May include some other, very minor code & testing updates, for example:

- small update I proposed in #2087, as needed for this proposal to build with optional `withrcalias` feature enabled
- renamed an internal API unit test utility function to `use_default_with_exactly_one_provider` for improved clarity

## APPROACH

* add & use new, internal `alias` module to alias `alloc::rc::Rc` as `Arc` when the `withrcalias` feature is enabled
* add & use new, internal `pub_api_trait` & `internal_generic_state_trait` macros to define certain traits with or without `Send + Sync`, as needed for the build to succeed, depending on whether `withrcalias` feature is absent or present
* ~~use `defaultproviderenabled` feature configuration to enable or disable default provider API functions, which will not build or work when the `withrcalias` feature is enabled~~
* __UPDATED SOLUTION:__ using internal `crypto_default_provider` module to manage default crypto provider state in a cleaner manner, regardless of `withrcalias` feature, with thanks to `once_cell` with `critical-section` feature

~~NOTE that this includes an update I proposed in #2087, as needed to build when the `withrcalias` feature is enabled. _(PR #2087 is now merged & merged into this proposal)_~~

## RATIONALE

* adding an option to use `alloc::rc::Rc` instead of `alloc::sync::Arc` enables use with no-std targets w/o built-in atomic ops, as described in #2068, without the CI issues described in https://github.com/rustls/rustls/issues/2068#issuecomment-2269332639
* keeping default provider API functions behind `defaultproviderenabled` feature enables the build to succeed with `Send + Sync` removed from many pub API traits

## HOW TESTED

- `cargo build -p rustls -F withrcalias` (try building of package with `withrcalias` feature enabled)
- ~~`cargo build -F defaultproviderenabled` (try full build with `defaultproviderenabled` feature enabled)~~
- `cargo test -p rustls -F withrcalias` (unit test of package with `withrcalias` feature enabled)
- ~~`cargo test -F defaultproviderenabled` (full unit testing with `defaultproviderenabled` feature enabled)~~
- `cargo test -p rustls` ~~(less important - unit test of package with both new feature options not enabled)~~
- _now working with multiple cross-rs test targets_
- _build now tested with `thumbv6m-none-eabi`, with no atomic pointer functionality_

---

## TODO items & loose ends

- `git grep "pub.*trait.*Send"` shows some more public traits that should probably be updated to use `pub_api_trait` macro, to remove `Send + Sync` when the `withrcalias` feature is enabled - it may be ideal to add some more unit tests to check that `Send + Sync` is gone for these public traits with optional `withrcalias` feature
- it would be ideal if we can migrate from the new, internal macros to using conditional marker aliases, not sure if Rust will ever provide stable or even unstable support for this kind of thing
- fix & update the generated API documentation
- known CI testing issues that need to be resolved
- add new feature options to CI feature testing
- may have issue with `--all-features` cargo option (I am wondering if `custom-provider` could lead to a similar issue with built-in providers not enabled)
- resolve many warning messages
- clean up some ugly hacks
- resolve commented XXX & TODO items - `git grep XXX` should not show any new results with this proposal
- improve consistency of formatting of `#[cfg(...)]` formatting


rustls/tests/unbuffered.rs
```diff
@@ -1,6 +1,10 @@
 #![allow(clippy::duplicate_mod)]
 
 use std::num::NonZeroUsize;
+
+#[cfg(feature = "withrcalias")]
+use std::rc::Rc as Arc;
```
djc, line 6: Maybe we should expose `alias` via `internals` instead?

rustls/Cargo.toml
```diff
@@ -36,6 +36,11 @@ zlib-rs = { version = "0.2", optional = true }
 default = ["aws_lc_rs", "logging", "std", "tls12"]
 std = ["webpki/std", "pki-types/std", "once_cell/std"]
 logging = ["log"]
+# XXX TODO NAMING - ADD SEPARATORS
+# XXX TODO ADD THIS TO DEFAULT FEATURES
+defaultproviderenabled = []
```
djc, line 41: Would suggest splitting this out in a separate commit or PR, named `default-provider` with separate motivation on why this is hard to support on the targets you care about.

rustls/Cargo.toml
```diff
@@ -36,6 +36,11 @@ zlib-rs = { version = "0.2", optional = true }
...
+# XXX TODO NAMING - ADD SEPARATORS
+withrcalias = []
```
djc, line 43: For naming, I'd suggest `rc-as-arc`.

rustls/src/alias.rs
```diff
@@ -0,0 +1,5 @@
+#[cfg(not(feature = "withrcalias"))]
```
djc, line 1: This should just be an inline module in `lib.rs`.

rustls/src/builder.rs
```diff
@@ -1,5 +1,6 @@
+use crate::alias::Arc;
```
djc, line 1: `crate` imports should go in the block of `crate` and `super` imports (everywhere).

rustls/src/error.rs
```diff
@@ -634,7 +634,11 @@ mod other_error {
     ///
     /// Enums holding this type will never compare equal to each other.
     #[derive(Debug, Clone)]
-    pub struct OtherError(#[cfg(feature = "std")] pub Arc<dyn StdError + Send + Sync>);
+    pub struct OtherError(
+        #[cfg(not(feature = "withrcalias"))]
+        #[cfg(feature = "std")]
```
djc, line 639: Should merge double `cfg` attributes everywhere.

Why disable this for `withrcalias`?

djc (commented):
Here's some initial feedback.

rustls/src/trait_macros.rs
```diff
@@ -0,0 +1,33 @@
+/// pub trait that includes Send & Sync - supports use with alloc::sync::Arc
+#[cfg(not(feature = "withrcalias"))]
+macro_rules! pub_api_trait {
+    ($name:ident, $body:tt) => {
+        pub trait $name: core::fmt::Debug + Send + Sync $body
```
ctz, line 5: I haven't tried this, but I was assuming this would work:

```rust
#[cfg(target_has_atomic = "ptr")]
pub trait ThreadSafe : Send + Sync {}

#[cfg(not(target_has_atomic = "ptr"))]
pub trait ThreadSafe {}
```

(then each extension point trait would be `pub trait ResolvesServerCert: Debug + ThreadSafe { ... }`)

rustls/Cargo.toml
ctz, line 43: Can we actually do this with features? It causes a public API breaking change. Perhaps we can use `#[cfg(target_has_atomic = "ptr")]`, which simply echoes the non-portability of Arc?

rustls/src/trait_macros.rs
djc: Oh, good idea.

brodycj: I did already try this kind of idea, discovered that API user would need to add additional empty `impl` block for the added "ThreadSafe" trait. Looks like Rust unstable would support trait alias, not sure how well this would work for conditional usage of the `Send + Sync` marker traits.

rustls/src/crypto/mod.rs
```diff
@@ -1,12 +1,14 @@
...
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 
-#[cfg(not(feature = "std"))]
+#[cfg(all(feature = "defaultproviderenabled", not(feature = "std")))]
 use once_cell::race::OnceBox;
-#[cfg(feature = "std")]
+#[cfg(all(feature = "defaultproviderenabled", feature = "std"))]
 use once_cell::sync::OnceCell;
```
brodycj, line 10: I've been thinking it may be ideal to define & use another alias for `OnceCell` vs `OnceBox`. It may be a little cleaner to break out to another PR, not sure if a separate PR is worth the trouble for such a small update for this.

rustls/Cargo.toml
brodycj: I would totally like to get rid of the new features, do have a couple questions:

Any pointers or ideas how we might test this in CI (and on a local workstation)?

I would also love to add an option to keep the Arc support working with the portable-atomics crate. This could increase the complexity of the cfg conditions in multiple places. We could try to address this at some point in the future, with a little extra code churn.

rustls/src/error.rs
brodycj: I recall using Rc led to an issue with this due to lack of `Send + Sync` marker traits.

rustls/tests/unbuffered.rs
brodycj: > Maybe we should expose `alias` via `internals` instead?

Should be now done & resolved thanks!

rustls/src/alias.rs
brodycj: should be resolved: now moved into the inline module

rustls/src/crypto/mod.rs
```diff
@@ -588,6 +570,47 @@ pub fn default_fips_provider() -> CryptoProvider {
     aws_lc_rs::default_provider()
 }
 
+#[cfg(feature = "defaultproviderenabled")]
+mod crypto_default_provider {
+    use alloc::sync::Arc;
```
brodycj, line 575: I have already proposed this update in PR #2089. I am planning to update this cfg attribute to get rid of the new feature options.

.github/workflows/build.yml
```diff
@@ -372,17 +175,32 @@ jobs:
...
-      - name: Check semver
-        uses: obi1kenobi/cargo-semver-checks-action@v2
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Install cross (cross-rs) from GitHub
+        run: cargo install cross --git https://github.com/cross-rs/cross
+      # XXX TODO FIND WAY TO GET BUILD WORKING WITH BUILT-IN PROVIDERS
+      - run: cross build --no-default-features --features custom-provider,withrcalias --package rustls --target ${{ matrix.target }}
+      # XXX TRY with alloc::sync::Arc - THIS SHOULD FAIL DUE TO LACK OF SUPPORT for thumbv6m:
+      - run: cross build --no-default-features --features custom-provider --package rustls --target ${{ matrix.target }}
```
brodycj, line 201: FAILED AS EXPECTED due to missing support for `alloc::sync` on `thumbv6m` - https://github.com/rustls/rustls/actions/runs/10659143604/job/29541249343?pr=2088#step:6:15 - time to remove these lines:

```suggestion
```

rustls/Cargo.toml
```diff
@@ -24,7 +24,7 @@ brotli-decompressor = { version = "4.0.1", optional = true } # 4.0.1 required fo
 hashbrown = { version = "0.14", optional = true, default-features = false, features = ["ahash", "inline-more"] }
 log = { version = "0.4.8", optional = true }
 # remove once our MSRV is >= 1.70
-once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
+once_cell = { version = "1.16", default-features = false, features = ["alloc", "critical-section"] }
```
brodycj, line 27: Fails on `thumbv6m` - https://github.com/rustls/rustls/actions/runs/10661757574/job/29547782037?pr=2088#step:5:71

```
error[E0599]: no method named `compare_exchange` found for struct `portable_atomic::AtomicUsize` in the current scope
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/once_cell-1.19.0/src/race.rs:60:24
   |
60 |             self.inner.compare_exchange(0, value.get(), Ordering::AcqRel, Ordering::Acquire);
   |                        ^^^^^^^^^^^^^^^^ method not found in `AtomicUsize`
```

Fails in `race.rs` even though `race` features is not requested or needed from `once_cell`.

matklad: Could you try 

```
once_cell = { version = "1.16", default-features = false, features = ["critical-section"] }
```

That is, without alloc? It seems that the only thing enabled by `alloc` is `race`, so, if you need `race`, you should be able to remove `alloc`?

brodycj: Solved it - thanks & apologies!

rustls/Cargo.toml
brodycj: This option is now gone

rustls/Cargo.toml
brodycj: I will likely propose a different solution for this, as discussed elsewhere thanks.

rustls/src/builder.rs
brodycj: done ✅

rustls/src/crypto/mod.rs
```diff
@@ -226,45 +222,39 @@ impl CryptoProvider {
...
 
     /// Returns the default `CryptoProvider` for this process.
     ///
     /// This will be `None` if no default has been set yet.
-    pub fn get_default() -> Option<&'static Arc<Self>> {
-        PROCESS_DEFAULT_PROVIDER.get()
+    #[cfg(feature = "withrcalias")]
+    pub fn get_default() -> Option<Arc<Self>> {
+        static_default::get_default()
     }
```
brodycj, line 243: Unfortunately this API function has a somewhat different return type depending on the build configuration.

Existing return type with static-lifetime reference to `Arc` cannot be replicated with `Rc` alias due to differences in storing & cloning.

I would rather not break existing API usage for any existing users.

I think I need to add a note to the source to explain this.

cpu:
Just wanted to leave a note here that I'm going to wait to review this until it's out of draft status and the commit history is easier to approach. Please tag if there are specific questions I could try to answer in the meantime!

brodycj:
@cpu it looks to me like updating `aws-lc-rs` as proposed in PR #2118 is needed for cross testing to succeed in GitHub CI, hopeful that change will be merged in the near future.

I am now thinking of taking another approach which I may end up proposing in a brand-new PR. My apologies for all of the noise with this one!

djc:
> I am now thinking of taking another approach which I may end up proposing in a brand-new PR. My apologies for all of the noise with this one!

Going to flip this back to draft for now, then.

brodycj:
I would like to close this one for now. I think that the alternative I proposed in PR #2200 should be much cleaner & easier to understand. I do think this would be an interesting option to consider adding in the future, but it would likely be less drastic if we can finish & merge #2200 first.

PR #2089, opened 2024-08-26
default crypto provider improvements

brodycj:
WHAT _(updated)_

- add & use inline module to manage default crypto provider within `rustls/src/crypto/mod.rs`
- fix `rustls::crypto::CryptoProvider::install_default()` to return consistent result type with both `std` & `no-std` _build config_
- ~~use `Option:: unwrap_or_else()` to shorten an internal function~~
- ~~rename some default crypto provider test functions (I would like to pull these out of PR #2088)~~

---

WHY - UPDATED:

SEE PR: #2088 & ISSUE #2068

This refactoring & cleanup would help make it easier to update & maintain the code that manages the default crypto provider.

The code needed to store the default crypto provider already depends on the build environment, requiring `#[cfg` statements sprinkled through some of the existing import statements, `rustls::crypto::CryptoProvider` API functions, and configuration-specific static variables in `rustls/src/crypto/mod.rs`.

This issue with the various bits of conditional code is expected to compound as more build options are added over time, for example:

- replace `alloc::arc::Arc` with `alloc::rc::Rc`, with help from some type aliasing, to help support targets with non-atomic pointers
- option to use `once_cell` with `portable-atomic` & `critical-section`  crate options, as needed to support non-atomic pointers - and it should be possible to turn this option off in case there may be a more performant solution for certain build targets

~~In case of an embedded target with no atomic pointers or references, it would not be possible to build any part of this library using `alloc::arc::Arc`, and another solution such as using `alloc::rc::Rc` would be needed instead. For example, a type alias could be used to use `alloc::rc::Rc` in place of `Arc` (`alloc::arc::Arc`).~~

~~In this case, it would not be possible to support the existing default crypto provider API as it is depending on the `Send & Sync` support. Assuming that the existing default crypto provider API will be preserved, we would need to use a `cfg` directive to disable this default crypto provider functionality in this case. This would involve integrating `cfg` directive updates with some previously existing `cfg` directives in multiple places, becoming harder & harder to read, understand, and maintain.~~

Moving the default crypto provider functionality into an internal module would simplify the ability to keep or remove this functionality depending on the build target.

I think this refactoring should offer some further benefits, regardless of these anticipated updated:

- improve readability of the default crypto provider code in general
- improve `rustls::crypto::CryptoProvider::install_default` API consistency for `std` vs `non-std` build config in general

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -3,10 +3,6 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 
-#[cfg(not(feature = "std"))]
-use once_cell::race::OnceBox;
-#[cfg(feature = "std")]
-use once_cell::sync::OnceCell;
 use pki_types::PrivateKeyDer;
 use zeroize::Zeroize;
 
@@ -227,28 +223,15 @@ impl CryptoProvider {
     /// Call this early in your process to configure which provider is used for
     /// the provider.  The configuration should happen before any use of
     /// [`ClientConfig::builder()`] or [`ServerConfig::builder()`].
-    #[cfg(feature = "std")]
     pub fn install_default(self) -> Result<(), Arc<Self>> {
-        PROCESS_DEFAULT_PROVIDER.set(Arc::new(self))
-    }
-
-    /// Sets this `CryptoProvider` as the default for this process.
-    ///
-    /// This can be called successfully at most once in any process execution.
-    ///
-    /// Call this early in your process to configure which provider is used for
-    /// the provider.  The configuration should happen before any use of
-    /// [`ClientConfig::builder()`] or [`ServerConfig::builder()`].
-    #[cfg(not(feature = "std"))]
-    pub fn install_default(self) -> Result<(), Box<Arc<Self>>> {
-        PROCESS_DEFAULT_PROVIDER.set(Box::new(Arc::new(self)))
+        static_default::install_default(self)
     }
 
     /// Returns the default `CryptoProvider` for this process.
     ///
     /// This will be `None` if no default has been set yet.
     pub fn get_default() -> Option<&'static Arc<Self>> {
-        PROCESS_DEFAULT_PROVIDER.get()
+        static_default::get_default()
     }
 
     /// An internal function that:
@@ -318,11 +301,6 @@ impl CryptoProvider {
             && key_provider.fips()
     }
 }
-
-#[cfg(feature = "std")]
-static PROCESS_DEFAULT_PROVIDER: OnceCell<Arc<CryptoProvider>> = OnceCell::new();
-#[cfg(not(feature = "std"))]
-static PROCESS_DEFAULT_PROVIDER: OnceBox<Arc<CryptoProvider>> = OnceBox::new();
 
 /// A source of cryptographically secure randomness.
 pub trait SecureRandom: Send + Sync + Debug {
@@ -624,6 +602,44 @@ impl From<&[u8]> for SharedSecret {
 #[cfg_attr(docsrs, doc(cfg(feature = "fips")))]
 pub fn default_fips_provider() -> CryptoProvider {
     aws_lc_rs::default_provider()
+}
+
+mod static_default {
+    #[cfg(not(feature = "std"))]
+    use alloc::boxed::Box;
+    use alloc::sync::Arc;
+
+    #[cfg(not(feature = "std"))]
+    use once_cell::race::OnceBox;
+    #[cfg(feature = "std")]
+    use once_cell::sync::OnceCell;
+
+    use super::CryptoProvider;
+
+    #[cfg(feature = "std")]
+    pub(crate) fn install_default(
+        default_provider: CryptoProvider,
+    ) -> Result<(), Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER.set(Arc::new(default_provider))
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub(crate) fn install_default(
+        default_provider: CryptoProvider,
+    ) -> Result<(), Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER
+            .set(Box::new(Arc::new(default_provider)))
+            .map_err(|e| *e)
+    }
+
+    pub(crate) fn get_default() -> Option<&'static Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER.get()
+    }
+
+    #[cfg(feature = "std")]
+    static PROCESS_DEFAULT_PROVIDER: OnceCell<Arc<CryptoProvider>> = OnceCell::new();
+    #[cfg(not(feature = "std"))]
+    static PROCESS_DEFAULT_PROVIDER: OnceBox<Arc<CryptoProvider>> = OnceBox::new();
 }
 
 #[cfg(test)]


rustls/src/crypto/mod.rs
```diff
@@ -588,6 +566,46 @@ pub fn default_fips_provider() -> CryptoProvider {
...
+    use once_cell::race::OnceBox;
+    #[cfg(feature = "std")]
+    use once_cell::sync::OnceCell;
+
+    use crate::crypto::CryptoProvider;
+
+    #[cfg(feature = "std")]
+    static PROCESS_DEFAULT_PROVIDER: OnceCell<Arc<CryptoProvider>> = OnceCell::new();
+    #[cfg(not(feature = "std"))]
+    static PROCESS_DEFAULT_PROVIDER: OnceBox<Arc<CryptoProvider>> = OnceBox::new();
```
djc, line 585: These `static`s should go below the usage site in the functions.

rustls/src/crypto/mod.rs
```diff
@@ -588,6 +566,46 @@ pub fn default_fips_provider() -> CryptoProvider {
     aws_lc_rs::default_provider()
 }
 
+mod crypto_default_provider {
+    use alloc::sync::Arc;
+
+    #[cfg(not(feature = "std"))]
```
djc, line 572: Keep both `alloc` impls together (no empty line), please, even with the guard.

rustls/src/crypto/mod.rs
```diff
@@ -588,6 +566,46 @@ pub fn default_fips_provider() -> CryptoProvider {
...
+
+    use crate::crypto::CryptoProvider;
+
+    #[cfg(feature = "std")]
+    static PROCESS_DEFAULT_PROVIDER: OnceCell<Arc<CryptoProvider>> = OnceCell::new();
+    #[cfg(not(feature = "std"))]
+    static PROCESS_DEFAULT_PROVIDER: OnceBox<Arc<CryptoProvider>> = OnceBox::new();
+
+    #[cfg(feature = "std")]
+    pub(crate) fn install_default_provider(
```
djc, line 588: For these method names, I'd prefer just using the method names (`install_default()`, `get_default()`). Suffixing with `_provider` isn't all that helpful since they're in a `crypto_default_provider` module already.

rustls/src/crypto/mod.rs
```diff
@@ -588,6 +566,46 @@ pub fn default_fips_provider() -> CryptoProvider {
     aws_lc_rs::default_provider()
 }
 
+mod crypto_default_provider {
```
djc, line 569: Maybe call this `default_impl`?

djc (approved):
This makes sense to me.

cpu:
Having not reviewed #2088 yet I think I'd appreciate some elaboration in the accompanying text to help this change set stand alone.

At the moment the commit message only says "default crypto provider improvements" and doesn't get into any motivation or enumerate any improvements. Similarly the PR description says it makes it easier to configure the provider in 2088, but doesn't expand on _how_.

The actual mechanics of the diff seem reasonable (once djc's feedback is addressed).

rustls/src/crypto/mod.rs
brodycj: Sounds like I should move these static declarations below all functions, to the very end of this new "default implementation" or "default impl" block. I would totally favor this. In case this is not what you are suggesting can you give me very explicit instructions wheee to and under what I should move these static declarations. Just hoping to avoid potential confusion on my part thanks!

brodycj:
With the one clarification that I requested. all suggestions sounds good. I will try to get this proposal fixed within the next 1-2 days. Thanks!

rustls/src/crypto/mod.rs
cpu: > Sounds like I should move these static declarations below all functions, to the very end of this new "default implementation" or "default impl" block.

Yup! I think that's exactly right.

brodycj:
I have now updated the description, rebased, and included some more info in the commit message to include the details of the what & why (with `#` removed from the issue & PR number references to avoid extra GitHub spamming) ... keeping this PR in draft status for now thanks!

rustls/src/crypto/mod.rs
brodycj: done

rustls/src/crypto/mod.rs
brodycj: shortened ✅

rustls/src/crypto/mod.rs
brodycj: I came up with `mod static_default`, can change it again if needed.

rustls/src/crypto/mod.rs
```diff
@@ -256,15 +239,13 @@ impl CryptoProvider {
...
-        if let Some(provider) = Self::get_default() {
-            return provider;
-        }
-
-        let provider = Self::from_crate_features()
-            .expect("no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point");
-        // Ignore the error resulting from us losing a race, and accept the outcome.
-        let _ = provider.install_default();
-        Self::get_default().unwrap()
+        static_default::get_default().unwrap_or_else(|| {
```
djc, line 242: IMO the previous structure with the `let-else` was easier to follow than the chaining with `unwrap_or_else()` that you have here.

rustls/src/crypto/mod.rs
```diff
@@ -588,6 +564,47 @@ pub fn default_fips_provider() -> CryptoProvider {
...
+    #[cfg(not(feature = "std"))]
+    use alloc::boxed::Box;
+    use alloc::sync::Arc;
+
+    #[cfg(not(feature = "std"))]
+    use once_cell::race::OnceBox;
+    #[cfg(feature = "std")]
+    use once_cell::sync::OnceCell;
+
+    use crate::crypto::CryptoProvider;
```
djc, line 577: This could just be `super::CryptoProvider`.

rustls/src/crypto/mod.rs
```diff
@@ -226,28 +222,15 @@ impl CryptoProvider {
...
-    ///
-    /// This can be called successfully at most once in any process execution.
-    ///
-    /// Call this early in your process to configure which provider is used for
-    /// the provider.  The configuration should happen before any use of
-    /// [`ClientConfig::builder()`] or [`ServerConfig::builder()`].
-    #[cfg(not(feature = "std"))]
-    pub fn install_default(self) -> Result<(), Box<Arc<Self>>> {
-        PROCESS_DEFAULT_PROVIDER.set(Box::new(Arc::new(self)))
+        static_default::install_default(Arc::new(self))
```
djc, line 226: Suggest you change `static_default::install_default()` to just take `CryptoProvider`, and let it handle the `Arc`-wrapping.

djc (approved):
Here are a bunch of nits about the implementation. IMO @cpu's point still stands that, in isolation, this does not seem like an obvious improvement, mostly moving stuff around.

brodycj:
@djc thanks for taking the time to review this!

I may need a few days to go through your comments.... I would like to quote myself from what I wrote in both the updated PR description & in the updated commit message in terms of general motivation - please let me know if my motivation or any part of my motivation does not sound clear enough:

> This refactoring & cleanup would help make it easier to update & maintain the code that manages the default crypto provider.
> 
> The code needed to store the default crypto provider already depends on the build environment, requiring `#[cfg` statements sprinkled through some of the existing import statements, `rustls::crypto::CryptoProvider` API functions, and configuration-specific static variables in `rustls/src/crypto/mod.rs`.
> 
> This issue with the various bits of conditional code is expected to compound as more build options are added over time, for example:
> 
> * replace `alloc::arc::Arc` with `alloc::rc::Rc`, with help from some type aliasing, to help support targets with non-atomic pointers
> * option to use `once_cell` with `critical-section`  crate option, as needed to support non-atomic pointers - and it should be possible to turn this option off in case there may be a more performant solution for certain build targets
> 
> Moving the default crypto provider functionality into an internal module would simplify the ability to keep or remove this functionality depending on the build target.
> 
> I think this refactoring should offer some further benefits, regardless of these anticipated updated:
> 
> * improve readability of the default crypto provider code in general
> * improve `rustls::crypto::CryptoProvider::install_default` API consistency for `std` vs `non-std` build config in general

cpu:
>> @djc thanks for taking the time to review this!
>
> I may need a few days to go through your comments....

@brodycj Checking in on this PR: will you have time to address Djc's feedback?

rustls/src/crypto/mod.rs
brodycj: reverted (under protest, as I would prefer things to be more functional whenever possible)

rustls/src/crypto/mod.rs
brodycj: done

rustls/src/crypto/mod.rs
brodycj: Done ... my impression is that this could add another move operation on the `CryptoProvider` config data (move, not copy), but this is just my impression ... suspect that any possible performance impact should be negligible, if any. Definitely seems to be cleaner this way :)

rustls/src/crypto/mod.rs
```diff
@@ -588,6 +564,47 @@ pub fn default_fips_provider() -> CryptoProvider {
...
+        }
+    }
+
+    pub(crate) fn get_default() -> Option<&'static Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER.get()
+    }
+
+    #[cfg(feature = "std")]
+    type DefaultProviderStore = OnceCell<Arc<CryptoProvider>>;
+    #[cfg(not(feature = "std"))]
```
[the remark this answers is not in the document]
brodycj: Yes that would be more concise, with fewer lines of code. But ... I still like my proposed single definition of this static `PROCESS_DEFAULT_PROVIDER` as I can see it more easily in one (single) place.

I did update my proposal to move the internal `DefaultProviderStore` type alias below the static `PROCESS_DEFAULT_PROVIDER` definition, which I think would read the best.

If this still looks like too much extra indirection, I would be happy to remove this internal type alias in favor of the 2 alternative `static` definitions.

brodycj:
I have now addressed @djc most recent comments & rebased into a clean commit, with an updated commit message that contains the updated justification. Heads up that I will likely unplug for the rest of this week due to a religious holiday. Thanks!

rustls/src/crypto/mod.rs
```diff
@@ -626,6 +604,47 @@ pub fn default_fips_provider() -> CryptoProvider {
...
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub(crate) fn install_default(
+        default_provider: CryptoProvider,
+    ) -> Result<(), Arc<CryptoProvider>> {
+        match PROCESS_DEFAULT_PROVIDER.set(Box::new(Arc::new(default_provider))) {
+            Ok(()) => Ok(()),
+            Err(previous_default_provider) => Err(*previous_default_provider),
+        }
```
cpu, line 633: I think this `match` can be replaced with `map_err()`:
```rust
        PROCESS_DEFAULT_PROVIDER.set(Box::new(Arc::new(default_provider))).map_err(|e|*e)
```

rustls/src/crypto/mod.rs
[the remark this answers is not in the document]
cpu: I think I would prefer two static `PROCESS_DEFAULT_PROVIDER` definitions and no type aliases but wouldn't block the PR on that feedback.

rustls/src/crypto/mod.rs
```diff
@@ -626,6 +604,47 @@ pub fn default_fips_provider() -> CryptoProvider {
...
+    pub(crate) fn install_default(
+        default_provider: CryptoProvider,
+    ) -> Result<(), Arc<CryptoProvider>> {
+        PROCESS_DEFAULT_PROVIDER.set(Arc::new(default_provider))
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub(crate) fn install_default(
+        default_provider: CryptoProvider,
+    ) -> Result<(), Arc<CryptoProvider>> {
```
cpu, line 629: Is changing this return type from `Result<(), Box<Arc<Self>>>` a breaking change for no-std users? It feels like yes.

ctz: I think so, but it fixes the feature-controlled breaking change which I would argue is kind of worse.

cpu: True, and I think the no-std userbase is very small at this point.

cpu (approved):

rustls/src/crypto/mod.rs
brodycj: Done (with a little reformatting), need to rebase (yet) again (yet yet yet again)

brodycj:
rebased again with a few more updates & fixes:

- updates from main
- use `map_err()` to simplify internal fn for no-std (with minor reformatting)
- remove internal type alias for `PROCESS_DEFAULT_PROVIDER` (I still like it, but definitely less concise with the type alias)

ctz (approved):

djc (approved):

ctz:
Thanks!

PR #2090, opened 2024-08-26
chore: remove trailing spaces

brodycj:
Discovered the trailing spaces while working on doc fixes for PR #2088 - looks like this should be very, very limited in scope


djc (approved):

djc:
Thanks!

PR #2091, opened 2024-08-26
polyfill: allow dead code w/o built-in providers, no-std

cpu:
The dead code warning from `clippy` for `--no-default-features` builds has been breaking the daily powerset job ([exemplar](https://github.com/rustls/rustls/actions/runs/10565112794/job/29268946096)) since https://github.com/rustls/rustls/commit/1c33f4b78155b624a9fc6218ef48fae10edf10d9 landed.

Since it's a small polyfill function we allow it to be dead code and avoid more complicated `cfg_attr` expressions. Here's a [passing run](https://github.com/cpu/rustls/actions/runs/10567141405) with the change applied.

### DIFF
--- a/rustls/src/polyfill.rs
+++ b/rustls/src/polyfill.rs
@@ -1,5 +1,6 @@
 /// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
 // TODO(XXX): remove once stabilized in https://github.com/rust-lang/rust/issues/119128
+#[allow(dead_code)] // Complicated conditional compilation guards elided
 pub(crate) fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
     match mid > slice.len() {
         true => None,


rustls/src/polyfill.rs
```diff
@@ -1,5 +1,12 @@
 /// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
 // TODO(XXX): remove once stabilized in https://github.com/rust-lang/rust/issues/119128
+#[cfg_attr(
+    any(
+        not(any(feature = "aws-lc-rs", feature = "ring")),
+        not(feature = "std")
+    ),
+    allow(dead_code)
+)]
```
cpu, line 9: This boolean logic is a bit gnarly so I'd also be open to just slapping a `#[allow(dead_code)]` on this unconditionally. I thought I'd try to do it the right way first.

cpu:
Unrelated to this specific clippy finding I think our powerset job has been getting pretttttty chonky lately :-( Back when we added this step to the daily tests we only had ~5 features and the runtime was pretty manageable. We're up to ~9+ now and the power law is starting to punish us.

The last successful run was [3 weeks ago](https://github.com/rustls/rustls/actions/runs/10271930335/job/28423010308) and it took 15m to run through ~476 iterations. We added the `custom-provider` feature since then and now the powerset is ~940 iterations :skull: 

Perhaps worth thinking on how to trim that down?

rustls/src/polyfill.rs
djc: Something like `#[allow(dead_code)] // Complicated conditional compilation guards elided` sounds good to me.

cpu: WFM. Applied verbatim.

ctz (approved):

ctz:
> Unrelated to this specific clippy finding I think our powerset job has been getting pretttttty chonky lately :-( 
>
> Perhaps worth thinking on how to trim that down?

There are some options to tell cargo-hack to reduce the feature power set. eg, we could add:

grouping feature dependencies:

- `--group-features aws_lc_rs,aws-lc-rs` -- these are the same feature, but different spelling: aws-lc-rs activates aws_lc_rs
- `--group-features std,brotli` -- brotli activates std
- `--group-features std,read_buf` -- read_buf activates std
- `--group-features aws_lc_rs,fips` -- `fips` activates `aws_lc_rs`, so there is no reason to multiply those together

(I haven't tried any of these; just going from the docs.)

cpu:
> (I haven't tried any of these; just going from the docs.)

I can fiddle with some of this in a follow-up :+1:

PR #2092, opened 2024-08-27
Remove FFDHE constants from programs that don't use FFDHE

### LINKED ISSUE #2081: Rework layout of FFDHE support to be linker-friendly
A downstream project which does not use anything FFDHE-related, has, as its top line in `cargo bloat`:

```
0.1%   4.8%  29.0KiB               rustls rustls::msgs::ffdhe_groups::FfdheGroup::named_group
```

We should reorganise things until the linker can discard these.

ctz:
The constants in `rustls::ffdhe_groups` are kinda large, and aren't used unless someone specifically builds a FFDHE key exchange group and includes it in their crypto provider. But the linker doesn't know that, and so the whole lot are included in people's final binary:

```
$ cargo bloat --bin limitedclient
 File  .text     Size      Crate Name
 1.3%   9.4% 332.0KiB aws_lc_sys aws_lc_0_20_1_aes_gcm_encrypt_avx512
 1.3%   9.4% 332.0KiB aws_lc_sys aws_lc_0_20_1_aes_gcm_decrypt_avx512
>0.2%   1.7%  58.9KiB     rustls rustls::msgs::ffdhe_groups::FfdheGroup::named_group
 0.1%   0.5%  16.9KiB        std std::backtrace_rs::symbolize::gimli::Context::new
 0.1%   0.4%  13.8KiB        std std::backtrace_rs::symbolize::gimli::resolve
(...)
```

This PR ensures they are only referenced from each FFDHE key exchange group, so people pay for what they use. This is not a breaking API change, but to realise the code size benefits all key exchange implementors (FFDHE or not) should impl `ffdhe_group()`.

~~While this is not a breaking API change, anyone already using FFDHE will need to cooperate with the additional API surface: cc @s-arash @Taowyoo~~ 

fixes #2081

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -10,6 +10,7 @@ use once_cell::sync::OnceCell;
 use pki_types::PrivateKeyDer;
 use zeroize::Zeroize;
 
+use crate::msgs::ffdhe_groups::FfdheGroup;
 use crate::sign::SigningKey;
 pub use crate::webpki::{
     verify_tls12_signature, verify_tls13_signature, WebPkiSupportedAlgorithms,
@@ -411,6 +412,20 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
         })
     }
 
+    /// FFDHE group the `SupportedKxGroup` operates in.
+    ///
+    /// Return `None` if this group is not a FFDHE one.
+    ///
+    /// The default implementation calls `FfdheGroup::from_named_group`: this function
+    /// is extremely linker-unfriendly so it is recommended all key exchange implementers
+    /// provide this function.
+    ///
+    /// `rustls::ffdhe_groups` contains suitable values to return from this,
+    /// for example [`rustls::ffdhe_groups::FFDHE2048`][crate::ffdhe_groups::FFDHE2048].
+    fn ffdhe_group(&self) -> Option<FfdheGroup<'static>> {
+        FfdheGroup::from_named_group(self.name())
+    }
+
     /// Named group the SupportedKxGroup operates in.
     ///
     /// If the `NamedGroup` enum does not have a name for the algorithm you are implementing,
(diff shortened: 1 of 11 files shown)


cpu (approved):
Nice :+1:

cpu:
> @codecov codecov/patch Failing after 1s — 88.88% of diff hit (target 94.78%)

There's a small coverage gap for the `ffdhe_group()` fn of `ActiveKeyExchange` if you're keen to get this green.

ctz:
> There's a small coverage gap for the `ffdhe_group()` fn of `ActiveKeyExchange` if you're keen to get this green.

Added a trivial exercise of those.

cpu (approved):

ctz:
> Will this be accompanied with a `0.24` release (and not a `0.23.x` release)?

I would plan to put this in a 0.23.x release, and rustls-mbedtls-provider would need to implement `ffdhe_group()` to continue supporting FFDHE. But perhaps we could make the default impl of `ffdhe_group()` call the old function, and then remove that in a subsequent 0.23.x release? That would give you at least one release to track this release, rather than zero.

s-arash:
My concern is that since a new `0.23.13` release is sem-ver compatible with current `0.23.12`, there will be users who consider it safe to (or automatically) switch to `0.23.13` while still relying on `rustls-mbedcrypto-provider`  `v0.1.0`, and that breaks FFDHE support for them. We cannot fix that by releasing a new version of `rustls-mbedcrypto-provider`.

djc:
That's fair. @ctz could you live with holding this back until 0.24?

ctz:
I have reworked this to be the opposite: the default for the new `ffdhe_group()` function continues to link everything in, so there is no need for any lock-step change. Instead, the size improvements require everything to implement `ffdhe_group()` (but if you don't, the situation is identical to the current release).

cpu (approved):

rustls/src/crypto/mod.rs
```diff
@@ -417,6 +418,20 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
...
+    ///
+    /// Return `None` if this group is not a FFDHE one.
+    ///
+    /// The default implementation calls `FfdheGroup::from_named_group`: this function
+    /// is extremely linker-unfriendly so it is recommended all key exchange implementers
+    /// provide this function.
+    ///
+    /// `rustls::ffdhe_groups` contains suitable values to return from this,
+    /// for example [`rustls::ffdhe_groups::FFDHE2048`][crate::ffdhe_groups::FFDHE2048].
+    fn ffdhe_group(&self) -> Option<FfdheGroup<'static>> {
```
djc, line 431: Nit: I'd probably sort this before `name()`? (Also in the impls.)

rustls/src/crypto/mod.rs
```diff
@@ -493,6 +508,19 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    /// FFDHE group the `ActiveKeyExchange` is operating in.
+    ///
+    /// Return `None` if this group is not a FFDHE one.
+    ///
+    /// The default implementation returns `None`, so this function _must_ be provided
+    /// for FFDHE groups.
+    ///
+    /// `rustls::ffdhe_groups` contains suitable values to return from this,
+    /// for example [`rustls::ffdhe_groups::FFDHE2048`][crate::ffdhe_groups::FFDHE2048].
+    fn ffdhe_group(&self) -> Option<FfdheGroup<'static>> {
```
djc, line 522: Nit: maybe sort this before `pub_key()`?

djc (approved):
Nice!

rustls/src/crypto/mod.rs
ctz: I feel like this should live alongside `group` (different representation of the same data) so I'm minded to leave this one here.

rustls/src/crypto/mod.rs
ctz: reordered

PR #2093, opened 2024-08-27
ci: trim daily-tests powerset

cpu:
Reduces the runtime of the `cargo hack check` step of the `daily-tests.yml` workflow by grouping/excluding certain feature combos. Follow-up from https://github.com/rustls/rustls/pull/2091

Since the number of features in Rustls has grown the powerset check task has been getting very long, requiring ~940 builds without further configuration. In some cases we know certain features imply another, or are somewhat mutually exclusive in intent.

This commit adds `--group-features` and `--mutually-exclusive-features` configuration to the `cargo hack check` invocation, reducing the runtime greatly. Here's a [completed run](https://github.com/cpu/rustls/actions/runs/10598891676/job/29372553113) with the updated configuration (_4m duration_).

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -143,6 +143,15 @@ jobs:
         uses: taiki-e/install-action@cargo-hack
 
       - name: Check feature powerset
-        run: cargo hack check --feature-powerset --no-dev-deps
+        run: >
+            cargo hack check
+            --package rustls
+            --feature-powerset 
+            --no-dev-deps
+            --group-features aws_lc_rs,aws-lc-rs
+            --group-features fips,aws_lc_rs
+            --mutually-exclusive-features fips,ring
+            --mutually-exclusive-features custom_provider,aws_lc_rs
+            --mutually-exclusive-features custom_provider,ring
         env:
           RUSTFLAGS: --deny warnings


.github/workflows/daily-tests.yml
```diff
@@ -143,6 +143,17 @@ jobs:
         uses: taiki-e/install-action@cargo-hack
 
       - name: Check feature powerset
-        run: cargo hack check --feature-powerset --no-dev-deps
+        run: >
+            cargo hack check
+            --package rustls
```
cpu, line 148: Prev we were running the check for all workspace packages, but I think its primarily of value to `rustls` and we can skip some runtime avoiding doing it for `bogo`, `rustls-post-quantum`, etc.

.github/workflows/daily-tests.yml
```diff
@@ -143,6 +143,17 @@ jobs:
...
+            --feature-powerset 
+            --no-dev-deps
+            --group-features aws_lc_rs,aws-lc-rs
+            --group-features brotli,std
+            --group-features read_buf,std
+            --group-features fips,aws_lc_rs
+            --mutually-exclusive-features fips,ring
+            --mutually-exclusive-features custom_provider,aws_lc_rs
+            --mutually-exclusive-features custom_provider,ring
```
cpu, line 157: The `--group-features` config are from [the suggestions ctz made](https://github.com/rustls/rustls/pull/2091#issuecomment-2311723723). Unfortunately with those changes alone the powerset is still ~641 builds and seems too unwieldly.

`fips` and `ring` aren't truly mutually exclusive, but it seems unlikely folks that want FIPS are going to also want the ring provider so a `fips`, `aws-lc-rs`, `ring` combinations don't seem very useful. Adding that reduces the set to ~480 builds.

My read of the intent for the `custom_provider` feature is that it's unlikely you want that feature and one of the built-in provider features. Adding those last two `--mutually-exclusive-features` sets brings the set down to a very reasonable 160 builds.

WDYT? Do these changes seem like a good tradeoff between coverage & runtime?

ctz (approved):

.github/workflows/daily-tests.yml
ctz: I think this sounds like a good improvement. I feel pretty relaxed about this job taking tens of minutes each night, but naturally that leaves little budget if we want to add a feature.

djc: For things like `brotli` and `std`, I assume `--group-features` means that we don't exercise `brotli` without `std` (which is intended) but that we also don't exercise `std` without `brotli` (which is not)? Combined with the other `--group-features`, it means we never test `brotli` without `read_buf`.

I suppose this is fine for these particular cases, just something to be mindful of.

(It feels like `cargo-hack` could/should figure some of these out from the feature dependencies...)

djc (approved):

.github/workflows/daily-tests.yml
ctz: https://github.com/taiki-e/cargo-hack/issues/81 implies it does look at feature dependencies, and checking last night's run, there are no invocations with (eg) both `read_buf` and `std`, or `brotli` and `std`. So maybe we can get rid of those `--group-features`?

However there are for `fips` and `aws-lc-rs`, and the `--group-features aws-lc-rs,aws_lc_rs` is definitely warranted:

```
running `cargo check --no-default-features --features aws-lc-rs,brotli,custom-provider,read_buf,ring,tls12` on rustls (397/940)
running `cargo check --no-default-features --features aws_lc_rs,brotli,custom-provider,read_buf,ring,tls12` on rustls (398/940)
```

(there's no reasonable way for cargo hack to know that these are two spellings of the same feature.)

PR #2095, opened 2024-08-28
ci: cross-target build & test updates

brodycj:
__UPDATED:__ _update to do `cross test` over multiple targets_

Leaving out the following targets, need to use `portable-atomic` to get CA test working for these (already got these working in my personal branch):

- arm-linux-androideabi
- armv5te-unknown-linux-gnueabi

Another nice target for `no-std`, if we can work around issue with no atomic refs for issue #2068: `thumbv6m-none-eabi`

will likely defer these missing targets for issue #2100 (time permitting, of course)

---

NOTES - UPDATED:

I think this should only be seen as a very small, slightly-incremental step towards testing on more and more embedded, "no-std" target environments.

Building & testing with `cross-rs` for many embedded targets seems to be limited by the following issues:

- supported Rust toolchain has a very limited number of Tier 2 / Tier 3 targets (I am not sure if it has any tier 3 targets builit-in)
- limited number of targets supported by `cross` docker images (there may be some others out there in the wild)
- It looks like `aws-lc-rs` / `aws-lc-sys` requires some bindgen setup to build for most of the targets other than the normal Windows / /macOS / normal headless Linux targets. And it looks to me like even if we would not enable the `aws_lc_rs` feature, `cross test` would still include `aws-lc-rs` / `aws-lc-sys` due to this `dev-depenencies` entry needed for `rustls/examples/internal/test_ca.rs`:
  - `rcgen = { version = "0.13", default-features = false, features = ["aws_lc_rs", "pem"] }`

---

SOME HELPFUL RESOURCES:

- https://hongchai.medium.com/cross-compiling-armv7-linux-androideabi-using-rust-5b59c857353c
- https://doc.rust-lang.org/nightly/rustc/platform-support.html
- https://github.com/aws/aws-lc-rs/blob/main/.github/workflows/cross.yml
- https://doc.rust-lang.org/stable/rustc/platform-support.html

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -338,8 +338,25 @@ jobs:
         working-directory: rustls/
 
   cross:
-    name: Check cross compilation targets
+    name: cross-target testing
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+          # 32-bit Android (Linux) targets:
+          - armv7-linux-androideabi
+          - i686-linux-android
+          - thumbv7neon-linux-androideabi
+          # Other standard 32-bit (Linux) targets (SKIP bindgen for i686 only)
+          - i586-unknown-linux-gnu
+          - i686-unknown-linux-gnu
+          # exotic Linux targets:
+          - riscv64gc-unknown-linux-gnu
+          - s390x-unknown-linux-gnu
+          # additional target(s):
+          # NOTE: This could have some overlap with 64-bit ARM-style CPU on macOS CI host;
+          # may have similar ARM-style CPU overlap with standard Windows & possibly Linux in the future
+          - aarch64-linux-android
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
@@ -348,9 +365,12 @@ jobs:
 
       - name: Install rust toolchain
         uses: dtolnay/rust-toolchain@stable
-      - name: Install cross
-        uses: taiki-e/install-action@cross
-      - run: cross build -p rustls --locked --target i686-unknown-linux-gnu
+      - name: Install cross (cross-rs) from GitHub
+        run: cargo install cross --git https://github.com/cross-rs/cross
+      - name: Install bindgen feature & CLI for aws-lc-sys (as needed for many cross targets)
+        if: ${{ matrix.target != 'i686-unknown-linux-gnu' }}
+        run: cargo add --dev --features bindgen aws-lc-sys --package rustls --verbose && cargo install bindgen-cli --verbose
+      - run: cross test --package rustls --target ${{ matrix.target }}
 
   semver:
     name: Check semver compatibility


.github/workflows/build.yml
```diff
@@ -351,6 +351,28 @@ jobs:
...
+          - s390x-unknown-linux-gnu
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - run: cargo install cross --git https://github.com/cross-rs/cross
```
djc, line 372: Why can't we use `taiki-e/install-action@cross` here?

Couldn't i686 be part of the target matrix here?

cpu: > Couldn't i686 be part of the target matrix here?

Agreed - I think the bindgen installation bits could be conditional on the non-i686 target and then both jobs could be folded into one.

brodycj: Using that GitHub action did not work for me, kept getting CMake issues both with `taiki-e/install-action@cross` and now with this:

```yaml
      - name: Install cross
        uses: taiki-e/install-action@v2
        with:
          tool: cross
```

I will fold all of these jobs together, skipping the bindgen installation for the i686 (`i686-unknown-linux-gnu`) target.

.github/workflows/build.yml
```diff
@@ -349,8 +349,35 @@ jobs:
...
+      - run: cross build --package rustls --locked --target i686-unknown-linux-gnu
+      - run: cross test --package rustls --target i686-unknown-linux-gnu
+
+  cross-test-targets:
+    name: cross build-test - multiple targets
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+          - aarch64-linux-android
```
brodycj, line 364: I am not so convinced we should keep this `aarch64-linux-android` target, as this CPU type should already be covered by testing on normal macOS silicon. I would favor removing this in favor of adding some more exotic targets & other embedded CPU/OS combinations.

.github/workflows/build.yml
brodycj: I would actually like to remove the existing `i686-unknown-linux-gnu` target in favor of `i686-linux-android`, which seems to be working in the CI.

cpu: I'm not sure I understand `i686-unknown-linux-gnu` is cross-compiling fine in CI today, no?

brodycj: Alright I now have all the targets together, including `i686-unknown-linux-gnu`, multiple Android CPU build targets, and exotic `s390x-unknown-linux-gnu` target. I would love to add some more exotic & embedded targets, maybe in a separate PR (someday). There may be some overlap, should not be too bad at this point.

brodycj:
Shoud be ready now - review comments should be resolved

djc (approved):
LGTM, thanks!

Given that all the intermediate commits here just resolve TODOs introduced in the first commit, I don't think there's value in keeping them and we should just squash this. @brodycj can you do that?

cpu (approved):
Thanks! This is cool :-)

It seems like the increase in CI runtime is reasonable (all of the new jobs are still faster than the coverage task).

brodycj:
I just rebased with recent updates from `main`.

Yes this will definitely take some more CI resources, in the form of more parallel jobs. I did keep what I think are the most important cross targets toward the top of the matrix ... and did add some comments to help people understand what I think is the relative importance of these cross targets.

Assuming that we can start adding more and more embedded targets, as discussed in #2100, I would totally favor removing say a couple of the targets I added to reduce the chance of waiting for CI jobs stuck queued.

Thanks!

brodycj:
Seems to be held up by "Check cross compilation targets" job, which I had renamed in this PR..

brodycj:
I have pushed some updates to add a few more targets & keep the same `cross:` job key in `build.yaml`, hopefully won't be blocked for so long by GitHub again. I have added & updated some comments, think people should feel free to update or even remove some of the targets I added if this takes too heavy GitHub CI resources. Apologies for all of the churn in this PR.

djc:
I have removed the old job name from the merge queue guards. Right now the merge queue is not gating on any of the cross-compilation jobs. Not sure what we want to do with that going forward?

PR #2097, opened 2024-08-28
Ensure only max_early_data_size will be read after reading from ReadEarlyData

### LINKED ISSUE #2096: client can send more TLS1.3 early data than maximum if early data is read
Reported originally by @tahmid-23 :

---

### Summary
Reading early data on the server reduces the tracked received early data, which allows the client to send more early data than specified by `max_early_data_size`.

### Details
`ChunkVecBuffer` pops/shrinks chunks as it is read. In `EarlyDataState::take_received_plaintext`, it uses the total length of the `ChunkVecBuffer` to determine whether or not more early data can be accepted. However, if the server reads the early data, the `ChunkVecBuffer` will pop the data, thus reducing the perceived amount of early data that can be read.

### PoC
Adding this test to `rustls/tests/api.rs` demonstrates that this is possible.
In order for this to work, the client must behave maliciously. This can be achieved by making `rustls::client::client_conn::EarlyData::check_write_opt()` always return all requested bytes.
```rs
#[test]
fn malicious_early_data_client() {
    let (client_config, server_config) = early_data_configs();

    // first connection
    let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);

    // second connection
    let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
    assert!(client.early_data().is_some());
    assert_eq!(
        client
            .early_data()
            .unwrap()
            .bytes_left(),
        1234
    );

    // send early data on connection
    client
        .early_data()
        .unwrap()
        .write(&[b'Z'; 1234]) // maximum 1234 bytes
        .unwrap();

    transfer(&mut client, &mut server);
    server.process_new_packets().unwrap();

    // server begins to respond, but...
    transfer(&mut server, &mut client);
    // ...don't process anything on the client so that it remains in early state

    let mut received_early_data = [0u8; 1234];
    assert_eq!(
        server
            .early_data()
            .expect("early_data didn't happen")
            .read(&mut received_early_data)
            .expect("early_data failed unexpectedly"),
        1234
    );

    // Send as much early data as you'd like!
    // client sends early data, waits until server processes it, and then client sends more early data
    // an actual implementation over IO would require special timing
    for _ in 0..100 {
        // just keep sending under max early data size limit each time
        client
            .early_data()
            .unwrap()
            .write(&[b'Z'; 1234])
            .unwrap();
        transfer(&mut client, &mut server);
        // this should panic, too much early data was sent
        server.process_new_packets().unwrap();

        // server keeps responding...
        transfer(&mut server, &mut client);
        // client keeps itself in early state

        let mut received_early_data = [0u8; 1234];
        // fails here, all early data should be read at this point
        assert_eq!(
            server
                .early_data()
                .expect("early_data didn't happen")
                .read(&mut received_early_data)
                .expect("early_data failed unexpectedly"),
            0
        );
    }
}
```

### Impact
Any server that attempts to read early data before waiting for the handshake to complete is vulnerable.

---

In private discussion, we were sure that this was a bug, but unsure if this was a security bug, given RFC8446 doesn't actually require a server to limit anything to `max_early_data_size` (it's a "SHOULD"). However, we shall certainly fix this.

tahmid-23:
I'm not sure how to write a test for this
resolves #2096

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -988,7 +988,10 @@ impl State<ServerConnectionData> for Accepting {
 
 pub(super) enum EarlyDataState {
     New,
-    Accepted(ChunkVecBuffer),
+    Accepted {
+        received: ChunkVecBuffer,
+        left: usize,
+    },
     Rejected,
 }
 
@@ -1004,12 +1007,15 @@ impl EarlyDataState {
     }
 
     pub(super) fn accept(&mut self, max_size: usize) {
-        *self = Self::Accepted(ChunkVecBuffer::new(Some(max_size)));
+        *self = Self::Accepted {
+            received: ChunkVecBuffer::new(Some(max_size)),
+            left: max_size,
+        };
     }
 
     #[cfg(feature = "std")]
     fn was_accepted(&self) -> bool {
-        matches!(self, Self::Accepted(_))
+        matches!(self, Self::Accepted { .. })
     }
 
     pub(super) fn was_rejected(&self) -> bool {
@@ -1018,7 +1024,9 @@ impl EarlyDataState {
 
     fn pop(&mut self) -> Option<Vec<u8>> {
         match self {
-            Self::Accepted(ref mut received) => received.pop(),
+            Self::Accepted {
+                ref mut received, ..
+            } => received.pop(),
             _ => None,
         }
     }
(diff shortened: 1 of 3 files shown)


ctz (approved):
Thanks! This looks good to me. I will look into what the options are for testing this.

rustls/src/server/server_conn.rs
```diff
@@ -1055,7 +1071,12 @@ impl Debug for EarlyDataState {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
             Self::New => write!(f, "EarlyDataState::New"),
-            Self::Accepted(buf) => write!(f, "EarlyDataState::Accepted({})", buf.len()),
+            Self::Accepted { received, left } => write!(
+                f,
+                "EarlyDataState::Accepted{{ received: {}, left: {} }}",
```
djc, line 1076: Tiny nit: add a space before the `{{`, please?

djc (approved):
LGTM, thanks!

cpu (approved):

Issue #2085, opened 2024-08-17
early data questions


tahmid-23:
I have some questions about TLS1.3 early data.

1. Let's say I want to process all the early data on a connection, before the handshake is complete.
I would need to know when all the early data has finished reading. Can there ever be a situation where you process packets on the `ServerConnection`, `ReadEarlyData::read()` yields 0 bytes, yet there is still more early data that can be read? If so, then how can I tell when I am fully done reading the early data without waiting for handshake to complete?

2. I noticed that the tlsserver-mio example was recently updated to accommodate early data. However, when testing with OpenSSL s_client in echo mode, s_client says that the early data was rejected. Does tlsserver-mio really support early data? I believe that early data has to be read separately via `ReadEarlyData`, which doesn't seem to be used in the example code. If I'm mistaken, then is there an example of testing early data with tlsserver-mio?

3. I think that an analogous server example to simple_0rtt_client would be very useful. I would love to submit one, but my first point confuses me.

ctz:
Answering these in order:

1. At the moment the only signal is when the handshake ends, which (from the server's point of view) is the next message after the early data finishes. It would probably be sensible for an early data read to return EOF only when the `EndOfEarlyData` message has been processed.

2. Yes, AIUI the purpose of that change was to allow early data to be accepted, but not actually do anything with the data. It would make more sense to read the data too (in http mode: expect the request there, in echo mode: echo the early data, in forward mode: send the early data to the backend connection).

3. Yes, and it would be great if `simple_0rtt_client` could work with that.

PR #2102, opened 2024-08-29
Improve early data examples

tahmid-23:
Adds simple_0rtt_server, lets simple_0rtt_client connect to arbitrary servers (so it can work with simple_0rtt_server), makes tlsserver-mio read early data

questions
1. I need to add `--package rustls-examples` or else I get an error about no crypto provider being installed for both the existing 0rtt client and the new 0rtt server. might I be doing something wrong?
2. should I use an actual named args struct like in tlsclient-mio for the new options, or are the positional arguments fine?
3. if not, do I have the correct "syntax" for the arguments in the usage comment?
4. the handshaking code is a bit obtuse and mostly inspired by complete_io(). ideally I want to be able to progress the handshake without completion (since the point of early data is to not have to wait for the handshake completion). should some of this code be extracted to a new method? or maybe we can have a new function to handshake until early data can be read?
5. afaik early data is terminated by the EndOfEarlyData message, not by handshake completion, but usually these happen at the same time. should there be a new method to determine this, or is it not really worth it?
6. tlsserver-mio early data does not "work" with `openssl s_client -connect localhost:8443 -tls1_3 -sess_in sess.pem -early_data early.txt` if I don't specify --package rustls-examples (at least on my machine). something to do with fips feature being enabled causing a HRR to be sent due to mismatched kx groups, meaning that the early data gets dropped (I barely know what any of that means but this is what I think is happening). is this anything to worry about?

### DIFF
--- a/examples/src/bin/simple_0rtt_client.rs
+++ b/examples/src/bin/simple_0rtt_client.rs
@@ -1,5 +1,7 @@
 //! This is an example client that uses rustls for TLS, and sends early 0-RTT data.
 //!
+//! Usage: cargo r --bin simple_0rtt_client --package rustls-examples [domain name] [port] [path/to/ca.cert]
+//!
 //! You may set the `SSLKEYLOGFILE` env var when using this example to write a
 //! log file with key material (insecure) for debugging purposes. See [`rustls::KeyLog`]
 //! for more information.
@@ -9,17 +11,19 @@
 
 use std::io::{BufRead, BufReader, Write};
 use std::net::TcpStream;
+use std::str::FromStr;
 use std::sync::Arc;
+use std::{env, fs};
 
 use rustls::pki_types::ServerName;
 use rustls::RootCertStore;
 
-fn start_connection(config: &Arc<rustls::ClientConfig>, domain_name: &str) {
+fn start_connection(config: &Arc<rustls::ClientConfig>, domain_name: &str, port: u16) {
     let server_name = ServerName::try_from(domain_name)
         .expect("invalid DNS name")
         .to_owned();
     let mut conn = rustls::ClientConnection::new(Arc::clone(config), server_name).unwrap();
-    let mut sock = TcpStream::connect(format!("{}:443", domain_name)).unwrap();
+    let mut sock = TcpStream::connect(format!("{}:{}", domain_name, port)).unwrap();
     sock.set_nodelay(true).unwrap();
     let request = format!(
         "GET / HTTP/1.1\r\n\
@@ -66,11 +70,30 @@ fn start_connection(config: &Arc<rustls::ClientConfig>, domain_name: &str) {
 fn main() {
     env_logger::init();
 
-    let root_store = RootCertStore::from_iter(
-        webpki_roots::TLS_SERVER_ROOTS
-            .iter()
-            .cloned(),
-    );
+    let mut args = env::args();
+    args.next();
+    let domain_name = args
+        .next()
+        .unwrap_or("jbp.io".to_owned());
+    let port = args
+        .next()
+        .map(|port| u16::from_str(&port).expect("invalid port"))
+        .unwrap_or(443);
+
+    let mut root_store = RootCertStore::empty();
+    if let Some(cafile) = args.next() {
+        let certfile = fs::File::open(cafile).expect("Cannot open CA file");
+        let mut reader = BufReader::new(certfile);
+        root_store.add_parsable_certificates(
+            rustls_pemfile::certs(&mut reader).map(|result| result.unwrap()),
+        );
+    } else {
+        root_store.extend(
+            webpki_roots::TLS_SERVER_ROOTS
+                .iter()
+                .cloned(),
+        )
+    }
 
     let mut config = rustls::ClientConfig::builder()
         .with_root_certificates(root_store)
@@ -87,7 +110,7 @@ fn main() {
     // second will use early data if the server supports it.
 
     println!("* Sending first request:");
-    start_connection(&config, "jbp.io");
+    start_connection(&config, &domain_name, port);
     println!("* Sending second request:");
-    start_connection(&config, "jbp.io");
+    start_connection(&config, &domain_name, port);
 }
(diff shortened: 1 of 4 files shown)


ctz:
> questions
> 
> 1. I need to add `--package rustls-examples` or else I get an error about no crypto provider being installed for both the existing 0rtt client and the new 0rtt server. might I be doing something wrong?

This is unfortunately due to feature unification in the workspace, coming from `bogo/Cargo.toml` which turns on everything.

> 2. should I use an actual named args struct like in tlsclient-mio for the new options, or are the positional arguments fine?

I think this is fine, for the level of complexity here. But if we add any options or further arguments then it may be better to reach for a proper library.

> 4. the handshaking code is a bit obtuse and mostly inspired by complete_io(). ideally I want to be able to progress the handshake without completion (since the point of early data is to not have to wait for the handshake completion). should some of this code be extracted to a new method? or maybe we can have a new function to handshake until early data can be read?

I think this would be better motivated if we enabled `send_half_rtt_data` and wrote some data at the point of early data receipt. WDYT?

> 5. afaik early data is terminated by the EndOfEarlyData message, not by handshake completion, but usually these happen at the same time. should there be a new method to determine this, or is it not really worth it?

See point (1) on #2085

> 6. tlsserver-mio early data does not "work" with `openssl s_client -connect localhost:8443 -tls1_3 -sess_in sess.pem -early_data early.txt` if I don't specify --package rustls-examples (at least on my machine). something to do with fips feature being enabled causing a HRR to be sent due to mismatched kx groups, meaning that the early data gets dropped (I barely know what any of that means but this is what I think is happening). is this anything to worry about?

This is fine, it is because openssl prefers x25519 which is disabled in fips mode. A HRR is needed to complete such a handshake. The alternative is to add `-groups P-256` to the openssl command.

tahmid-23:
> I think this would be better motivated if we enabled `send_half_rtt_data` and wrote some data at the point of early data receipt. WDYT?

I think there is still some merit in being able to read the client's early data without sending half-RTT data if you choose not to have it enabled. It makes sense to show `send_half_rtt_data` in the example, but ideally you could handshake until early data is ready without `send_half_rtt_data`.

If we don't want to have the manual handshake code, would `complete_io()` finish once the client hello has been received, or is there a better solution? How might this work internally?

For clarification, the documentation for `send_half_rtt_data` says it "comes at the cost of a subtle weakening of the normal handshake integrity guarantees that TLS provides". What exactly does this mean? Is there any risk to sending half-RTT data if you are already accepting 0-RTT data?

> See point (1) on #2085

Wouldn't that be a breaking change?

ctz:
> If we don't want to have the manual handshake code, would `complete_io()` finish once the client hello has been received, or is there a better solution? How might this work internally?

I feel like `complete_io` is a function enabling simpler downstream programs, and making that take into account this subtle behaviour might make those programs more complex? At the more complex programs do not (or should not) use `complete_io`.

> For clarification, the documentation for `send_half_rtt_data` says it "comes at the cost of a subtle weakening of the normal handshake integrity guarantees that TLS provides". What exactly does this mean? Is there any risk to sending half-RTT data if you are already accepting 0-RTT data?

When a server receives the client's finished message and it is valid, it learns that the handshake was successful in both directions. If it sends data before then, it does so before the handshake integrity is knowable. (This whole explanation ignores client auth.)

examples/src/bin/simple_0rtt_server.rs
```diff
@@ -0,0 +1,107 @@
+//! This is an example server that streams 0-RTT early data from the client.
```
cpu, line 1: I think [`examples/README.md`](https://github.com/rustls/rustls/blob/main/examples/README.md) needs an update for this new example. 

Similarly, the [daily-tests.yml](https://github.com/rustls/rustls/blob/8c04dba680d19d203a7eda1951ad596f5fc2ae59/.github/workflows/daily-tests.yml#L59-L60) has a section that could use an update for this binary.

examples/src/bin/tlsserver-mio.rs
```diff
@@ -251,6 +251,19 @@ impl OpenConnection {
...
+                early_data
+                    .read_to_end(&mut buf)
+                    .unwrap();
+
+                if !buf.is_empty() {
+                    debug!("early data read {:?}", buf.len());
+                    self.incoming_plaintext(&buf);
+                    return;
+                }
+            }
```
cpu, line 265: Can you break this change out into its own commit alongside a commit message motivating the change?

examples/src/bin/simple_0rtt_client.rs
```diff
@@ -9,17 +11,19 @@
...
 use std::net::TcpStream;
+use std::str::FromStr;
 use std::sync::Arc;
+use std::{env, fs};
 
 use rustls::pki_types::ServerName;
 use rustls::RootCertStore;
 
-fn start_connection(config: &Arc<rustls::ClientConfig>, domain_name: &str) {
+fn start_connection(config: &Arc<rustls::ClientConfig>, domain_name: &str, port: u16) {
```
cpu, line 21: The changes to `simple_0rtt_client.rs` seem OK to me, but would be best in their own independent commit.

cpu (commented):
Some initial feedback, thanks!

examples/src/bin/simple_0rtt_server.rs
tahmid-23: this example doesn't terminate, it keeps accepting connections

cpu: Good point, and it doesn't have a `--help` or similar. Fair to skip the CI update then :+1:

cpu (approved):

cpu:
@djc @ctz Should I hold this for your reviews or is it acceptable to merge?

examples/src/bin/simple_0rtt_client.rs
```diff
@@ -1,5 +1,7 @@
 //! This is an example client that uses rustls for TLS, and sends early 0-RTT data.
 //!
+//! Usage: cargo r --bin simple_0rtt_client --package rustls-examples [domain name] [port] [path/to/ca.cert]
```
ctz, line 3: I have a feeling that a `--` might be needed here, after `--package rustls-examples`?

examples/src/bin/simple_0rtt_server.rs
```diff
@@ -0,0 +1,107 @@
...
+//!
+//! Usage: cargo r --bin simple_0rtt_server --package rustls-examples <path/to/cert.pem> <path/to/privatekey.pem>
```
ctz, line 3: Ditto (maybe? I haven't checked.)

tahmid-23: nope it's fine

ctz (approved):

PR #2103, opened 2024-08-30
[DRAFT] ci: add job to check that there are no warnings - DRAFT WIP

brodycj:
TODO:

- [ ] known to fail with a couple known warnings with release build & teset
- [ ] add steps to try build & test with some more combinations of features


rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -14,7 +14,7 @@ use aws_lc_rs::{hmac, iv};
 use super::ring_like::rand::{SecureRandom, SystemRandom};
 use super::unspecified_err;
 use crate::error::Error;
-#[cfg(feature = "logging")]
+#[cfg(debug_assertions)]
 use crate::log::debug;
```
brodycj, line 18:
```suggestion
```

I think it will still work in all cases if we remove these lines here & in `rustls/src/crypto/ring/ticketer.rs`, unforuntaely do not know how to explain why. I did try removing the `use crate::log::` lines from some other modules, did not seem to work with logging enabled. I would have to investigate further to explain this.

brodycj:
I have just rebased the updates

rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -14,7 +14,7 @@ use aws_lc_rs::{hmac, iv};
 use super::ring_like::rand::{SecureRandom, SystemRandom};
 use super::unspecified_err;
 use crate::error::Error;
-#[cfg(feature = "logging")]
+#[cfg(debug_assertions)]
```
ctz, line 17: this is a good change 👍🏽

.github/workflows/build.yml
```diff
@@ -484,3 +484,77 @@ jobs:
         run: cargo test --locked -- --include-ignored
         env:
           RUST_BACKTRACE: 1
+
```
ctz, line 487: However, I'm afraid I don't like the duplication here. Note we also already deny warnings for all features and no features, which should catch 99% of the issues -- see the clippy job.

djc: Maybe we can deny warnings on the powerset job to close some of the remaining gaps?

brodycj: > we also already deny warnings for all features and no features, which should catch 99% of the issues -- see the clippy job

Got it ... guessing the clippy job does not cover the release configuration?

The one issue I found with this was with release + logging combination.

> deny warnings on the powerset job

It looiks to me like the powerset job already does deny warnings, or am I mistaken here?

Incidentally I would like to switch to better spelled-out `--deny warnings` for improved clarity & consistency.

I did try running `cargo hack check` with the feature powerset (only on `--package rustls`) command from daily-tests.yml, seems to fail with this:

```sh
cargo check --no-default-features --features brotli,ring --package rustls
```

I will investigate and hopefully find a way to resolve this soon.

As an aside, I tried running `cargo hack test --package rustls` with the powerset, failed with `custom-package`, everything else seems to be OK. The test issue with the `custom-package` seems like a minor cfg issue in the tests, applied a quick workaround in my workarea for now.

cpu:
It seems to me the consensus here is:

* The change of the `#[cfg(feature = "logging")]` to `#[cfg(debug_assertions)]` gates is a good one, and we would accept a PR making that change in isolation
* The CI updates are not the direction we want to go. I think maintainers agree that it would be better to investigate why the powerset job is allowing warnings despite having its env set `RUSTFLAGS: --deny warnings`.

I'm going to close this draft since it hasn't had any activity recently and I think the best path forward would be one/two new PRs. Happy to reconsider if folks disagree with my summary.

Issue #1980, opened 2024-05-31
Server-side Encrypted Client Hello (ECH) support


cpu:
**Checklist**
* [x] I've searched the issue tracker for similar requests - https://github.com/rustls/rustls/issues/199 covers the general case. We want a more specific issue for server-side support.

**Is your feature request related to a problem? Please describe.**

Rustls should support [encrypted client hello (ECH)](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18) when operating as a server.

**Describe the solution you'd like**

I want to be able to implement a TLS server that supports ECH, operating in either of the supported [topologies](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18#section-3.1) - shared mode, or split mode.

I want to be able to ergonomically load ECH configurations and associated private key material in a server config. It should be possible to indicate which (if any) of the ECH configs should be offered as "retry configurations", to support graceful migration between ECH configs. Ideally it would be possible to change configurations without requiring a server reload.

**Describe alternatives you've considered**

N/A.

**Additional context**

Client side-support was implemented in https://github.com/rustls/rustls/pull/1718

icebluey:
The IETF has officially published RFC 9849: On March 3, 2026, the TLS Encrypted Client Hello (ECH) specification formally advanced to **Proposed Standard** (Standards Track), transitioning from its long-term Internet-Draft status (draft-ietf-tls-esni) to an official RFC.

The supporting standards were published at the same time:

- **RFC 9848**: Bootstrapping ECH configuration using DNS SVCB/HTTPS records.  
- **RFC 9934**: PEM file format for ECH configurations.

PR #2104, opened 2024-08-31
[DRAFT] Add EncryptedClientHello server support

tahmid-23:
adds ECH support for rustls servers

kind of rough around the edges but works
I left some // TODO with questions / things to be addressed. mostly style or spec clarifications that I would like help on
pending more tests if I figure out how they should be set up (and making the existing three work regularly without the cryptoprovider installation issue)

notes/questions not present in the code:
1. no distinction between split and shared mode. a shared mode server can just be another ServerConnection, which I think is ideal. otherwise you have to keep the ECH state in the ServerConnection and switch back and forth. this means it can use the same logic for split mode vs shared mode
2. is the EchAccepted concept/signature (maybe not the implementation just yet :stuck_out_tongue:) good? it's basically a temporary proxy that will run until the client and server can just talk to each other
3. should I remove the ech server out from the acceptor? I basically used it 'cuz it shares logic with regular accepting
4. from spec:
> When the client offers the "pre_shared_key" extension in ClientHelloInner, it SHOULD also include a GREASE "pre_shared_key" extension in ClientHelloOuter, generated in the manner described in [Section 6.1.2](https://www.ietf.org/archive/id/draft-ietf-tls-esni-20.html#grease-psk). The client MUST NOT use this extension to advertise a PSK to the client-facing server. (See [Section 10.12.3](https://www.ietf.org/archive/id/draft-ietf-tls-esni-20.html#flow-clienthello-malleability).) When the client includes a GREASE "pre_shared_key" extension, it MUST also copy the "psk_key_exchange_modes" from the ClientHelloInner into the ClientHelloOuter.
> When the client offers the "early_data" extension in ClientHelloInner, it MUST also include the "early_data" extension in ClientHelloOuter. This allows servers that reject ECH and use ClientHelloOuter to safely ignore any early data sent by the client per [[RFC8446](https://www.ietf.org/archive/id/draft-ietf-tls-esni-20.html#RFC8446)], [Section 4.2.10](https://rfc-editor.org/rfc/rfc8446#section-4.2.10).

does MUST mean the server has to verify?
5. I think only aws lc rs has hpkes. what to do about ring?

please criticize :slightly_smiling_face: 

Resolves https://github.com/rustls/rustls/issues/1980


cpu:
Thanks for getting the ball rolling here.

I suspect I won't have time to give this any substantial review for some time given the hours I have to allot to Rustls at the moment & the other work underway.

Shoring up test coverage (in particular, getting the bogo server-side ECH tests passing) and breaking this up into smaller commits would certainly help.

>  I think only aws lc rs has hpkes. what to do about ring?

It's fine to ignore ring. The client-side ECH support is already conditional on the basis of using `aws-lc-rs` and we should take the same approach here. There's nothing else to do until the upstream project adds HPKE or offers the static ECDH support required to implement HPKE in our crate like we did for `aws-lc-rs`.

cpu:
> does MUST mean the server has to verify?

Yes, that would be my reading.

tahmid-23:
> Shoring up test coverage (in particular, getting the bogo server-side ECH tests passing) and breaking this up into smaller commits would certainly help.

I got most of the bogo tests running except a few that I was unsure about (had to take some creative liberties to make it work :sweat_smile:), and I split off a few commits that were distinct from the main ECH implementation

cpu:
Out of curiosity, have you done any interop testing with different ECH compatible clients and a Rustls server? I suspect Chrome and Firefox are the most interesting interop targets but OpenSSL and Go have clients that could be tested as well.

> no distinction between split and shared mode. a shared mode server can just be another ServerConnection, which I think is ideal.

I'm unsure about this aspect of the design. I will have to spend time with the draft spec again but I think there are important reasons that split mode exists instead of simply suggesting a new connection from the handshaking server to another backend without consideration of ECH.

In general since it's probably wise to split as much complexity off from this branch as we can maybe it would be best to focus on only shared mode for now? (I suspect the it will be more common for the Rustls user-base anyway, split mode is probably of most use to larger CDN type applications).

> add hpke_suites to CryptoProvider

Is this necessary? We didn't do this for client-side support and I would have expected that we could use a similar strategy to keep the ECH suites separate from the provider struct.

tahmid-23:
> Out of curiosity, have you done any interop testing with different ECH compatible clients and a Rustls server? I suspect Chrome and Firefox are the most interesting interop targets but Curl/OpenSSL and Go have clients that could be tested as well.

Nope, I can look into it

> I'm unsure about this aspect of the design. I will have to spend time with the draft spec again but I think there are important reasons that split mode exists instead of simply suggesting a new connection from the handshaking server to another backend without consideration of ECH.

When reading the spec it didn't feel like they were very distinct to me. The backend is configured separately for accepting ECH.

I personally felt like a unified treatment can be applied to both. At the end of the day you need the split mode logic when you want to implement it, so why have two code paths? In a shared mode design, you need to switch between "frontend" mode and "backend" mode as well as determining if you've done an HRR In response to an ECH, OR if you're rejecting ECH as frontend due to bad configs. You also need to figure out which SNI to verify with. And you have to modify the way the ExpectClientHello works a bit to handle the inner ECH after the outer (or I'm just bad at Rust). Keeping track of state is annoying yet doable—but why bother with all of it?

> Is this necessary?

No

tahmid-23:
> Go have clients that could be tested as well.
also isn't this the whole point of bogo?

cpu:
> also isn't this the whole point of bogo?

The bogo ECH implementation != the Go stdlib implementation. The former was implemented before the later and is specifically geared for protocol testing. I suspect they're very similar, but it seems prudent to test with the stdlib impl as well.

tahmid-23:
> Out of curiosity, have you done any interop testing with different ECH compatible clients and a Rustls server? I suspect Chrome and Firefox are the most interesting interop targets but Curl/OpenSSL and Go have clients that could be tested as well.

Go works
Firefox works
Chrome I found a GREASE issue and now works
cURL I was getting some kind of weird issue in my makeshift dns `DoH: Too small type A for ech.localhost.direct` so I couldn't figure it out (also

cpu:
> Go works
> Firefox works
> Chrome I found a GREASE issue and now works

Nice, that's great :-)

> cURL I was getting some kind of weird issue in my makeshift dns DoH: Too small type A for ech.localhost.direct so I couldn't figure it out

There's a [Matrix chat room](https://matrix.to/#/#ech-dev:matrix.org) run by the Guardian Project folks for developers of ECH support to discuss implementation issues. The author of the curl/openssl ECH support (`@sftcd`) hangs around in there and has been helpful in the past. Perhaps he could suggest some avenues for debugging?

cpu (commented):
Here's a smattering of feedback from a first attempt to read through the code. While I appreciate your efforts in general I think my opinion is that this isn't a great task for an external contributor to pick up, and that I don't personally have the bandwidth available to coach you through the work that would be required to get this to a mergeable state. I'm sorry if that's disappointing to hear. 

I think this position is also supported by the history of how client-side ECH support went: there were a number of initial community contributed PRs that never quite made it across the line and instead it took dedicated effort and approaching the problem from the viewpoint of making as many discrete & reviewable changes as possible across a longer period to get the feature landed.

If you disagree with my perspective and want to continue to make progress then I think trying to pull out a design proposal and starting there (instead of with an impl) would be a better path forward. It's challenging dropping in to this diff and understanding why you made some of the choices you made.

Again, thank you for getting the ball rolling and taking a crack at this.

tahmid-23:
Thanks for the feedback.

I'm still interested in getting this to a more mergeable state, unless you think that server ECH is something that one of you will pick up in the near future and that it would be more efficient for one of you to start from scratch.

1. is the relative ordering and contents of commits supposed to be extremely motivating/highly logical? that's kind of what I'm getting from some of the comments you've made here and before. normally I just commit whenever... :stuck_out_tongue_closed_eyes:
2. I think that the design can be improved a lot, sure. I focused on getting to a working state when I had a lot of free time, and I can think about it more now that I know how it all works.
what does a design proposal look like to you?
3. can you tell me what "discrete, reviewable" changes look like? to me, it feels like ECH is a big feature that needs everything to work at once, bar some of the SHOULDs and GREASE stuff.
you outlined in one of your comments defining server config, handling outer, handling inner, HRR. is this the crux of what you mean?

when you told me to split the commits, I tried to just rebase the original commit into parts that I thought were logically independent units. I can try to make this flow better.

I think my main pain point is how I have the actual ECH frontend implemented. I can mainly try to address this aspect, rework everything around it, reorder it in the commit history, and then say when I think the PR is in a good place. Does this make sense?
And also, can I have some high-level suggestions on what you think an ECH frontend server design should look like at some point later? there is indeed a lot of IO logic that I made, and maybe I'm barking down the wrong tree.

djc:
Yes, we are pretty particular about using small(ish) commits that can each pass CI and sort of encompass one logical change per commit. You should carefully read [`CONTRIBUTING.md`](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md) if you haven't already. It would probably be helpful to also look at PRs that built up the client side of ECH:

* #1944
* #1963 
* #2001 
* #1718

cpu:
I'm going to close this for now based on the conclusions we reached in the previous comments. Thanks again,

PR #2106, opened 2024-09-02
Tidy logging macros

ctz:
Normalise the fact that logging macros live in `crate::log`, rather than a difference place depending on the `logging` feature.

ref https://github.com/rustls/hyper-rustls/pull/285#issuecomment-2324605158


djc (approved):

ctz:
I think this is pretty minor, merging with one approval.

Issue #2056, opened 2024-07-26
Productionise post-quantum support


ctz:
At the moment rustls-post-quantum has the following shortcomings:

- we offer it and may select it in TLS1.2 (draft doesn't say anything about this, but it actually cannot work)
- our selection of which `SupportedKxGroup` to send shares for in the client hello is limited to the first configured one, but for PQ we probably want to default to sending two: `X25519Kyber768Draft00` followed by `X25519` so we need improved configuration surface to express this.

(And, while I'm writing this down, we could also consider whether to do the optimisation where the X25519 setup computation is shared between the classical and hybrid algorithms. That's probably a different issue, and a pure optimisation.)

ctz:
> * our selection of which `SupportedKxGroup` to send shares for in the client hello is limited to the first configured one, but for PQ we probably want to default to sending two: `X25519Kyber768Draft00` followed by `X25519` so we need improved configuration surface to express this.
> 
> (And, while I'm writing this down, we could also consider whether to do the optimisation where the X25519 setup computation is shared between the classical and hybrid algorithms. That's probably a different issue, and a pure optimisation.)

I have been thinking about this a bit, and now have a design in mind that solves  these together in one reasonably composable API.

ctz:
We did this, and it was released in 0.23.22.

PR #2112, opened 2024-09-06
Allow `SupportedKxGroup` to be version-specific

ctz:
X25519Kyber768Draft00 and the forthcoming equivalent with ML-KEM are only defined for TLS1.3.

Therefore, a TLS1.2 client should not offer it, and server that decides on TLS1.2 should not select it.

This is one small step towards #2056.

### DIFF
--- a/rustls-post-quantum/src/lib.rs
+++ b/rustls-post-quantum/src/lib.rs
@@ -56,7 +56,7 @@ use rustls::crypto::{
     ActiveKeyExchange, CompletedKeyExchange, CryptoProvider, SharedSecret, SupportedKxGroup,
 };
 use rustls::ffdhe_groups::FfdheGroup;
-use rustls::{Error, NamedGroup, PeerMisbehaved};
+use rustls::{Error, NamedGroup, PeerMisbehaved, ProtocolVersion};
 
 /// A `CryptoProvider` which includes `X25519Kyber768Draft00` key exchange.
 pub fn provider() -> CryptoProvider {
@@ -126,6 +126,10 @@ impl SupportedKxGroup for X25519Kyber768Draft00 {
 
     fn name(&self) -> NamedGroup {
         NAMED_GROUP
+    }
+
+    fn usable_for_version(&self, version: ProtocolVersion) -> bool {
+        version == ProtocolVersion::TLSv1_3
     }
 }
 
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -403,12 +403,16 @@ impl ClientConfig {
             .find(|&scs| scs.suite() == suite)
     }
 
-    pub(super) fn find_kx_group(&self, group: NamedGroup) -> Option<&'static dyn SupportedKxGroup> {
+    pub(super) fn find_kx_group(
+        &self,
+        group: NamedGroup,
+        version: ProtocolVersion,
+    ) -> Option<&'static dyn SupportedKxGroup> {
         self.provider
             .kx_groups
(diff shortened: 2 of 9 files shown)


rustls-post-quantum/src/lib.rs
```diff
@@ -127,6 +127,10 @@ impl SupportedKxGroup for X25519Kyber768Draft00 {
     fn name(&self) -> NamedGroup {
         NAMED_GROUP
     }
+
+    fn usable_for_version(&self, version: ProtocolVersion) -> bool {
+        u16::from(version) >= u16::from(ProtocolVersion::TLSv1_3)
```
djc, line 132: Is this an instance where using `==` instead of `>=` would be akin to "failing closed" and might be better?

rustls/src/crypto/mod.rs
```diff
@@ -437,6 +437,13 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
     fn fips(&self) -> bool {
         false
     }
+
+    /// Return `true` if this should be offered/selected with the given version.
+    ///
+    /// The default implementation returns true for all versions.
+    fn usable_for_version(&self, _version: ProtocolVersion) -> bool {
```
djc, line 444: Maybe file an issue to remove the default impl for the next breaking change?

rustls/src/server/hs.rs
```diff
@@ -430,6 +430,7 @@ impl ExpectClientHello {
                 .provider
                 .kx_groups
                 .iter()
+                .filter(|skxg| skxg.usable_for_version(selected_version))
```
djc, line 433: I'd probably stick this in the `find()` argument.

rustls/src/client/hs.rs
```diff
@@ -242,16 +242,22 @@ fn emit_client_hello_for_retry(
...
+    let offered_groups = config
+        .provider
+        .kx_groups
+        .iter()
+        .filter(|skxg| {
+            supported_versions
+                .iter()
+                .any(|v| skxg.usable_for_version(*v))
+        })
+        .map(|skxg| skxg.name())
```
djc, line 255: `filter_map()`?

rustls/src/client/client_conn.rs
```diff
@@ -403,10 +403,15 @@ impl ClientConfig {
...
-    pub(super) fn find_kx_group(&self, group: NamedGroup) -> Option<&'static dyn SupportedKxGroup> {
+    pub(super) fn find_kx_group(
+        &self,
+        group: NamedGroup,
+        version: ProtocolVersion,
+    ) -> Option<&'static dyn SupportedKxGroup> {
         self.provider
             .kx_groups
             .iter()
+            .filter(|skxg| skxg.usable_for_version(version))
```
djc, line 414: Maybe combine `filter()` and `find()`?

djc (approved):

cpu:
@ctz Did you want to roll this into .13? It looks like it just needs some nits optionally addressed.

ctz:
No, I think this is quite low priority.

rustls-post-quantum/src/lib.rs
ctz: I guess maybe writing code now that works properly for TLS1.4 is probably premature by a decade or so. Will replace with a straight equality.

rustls/src/client/hs.rs
ctz: https://rust-lang.github.io/rust-clippy/master/index.html#/filter_map_bool_then clippy's on the ball!

PR #2118, opened 2024-09-09
Take aws-lc-rs 1.9

### LINKED ISSUE #1913: Rustls w/ aws-lc-rs on Windows requires NASM
The entire reason half the ecosystem is using rustls rather than openssl is because it "just works" without any dependencies.

Having upgraded to the latest version, I get the following error:
> error: failed to run custom build command for `aws-lc-sys v0.15.0`
> ...
> Missing dependency: nasm

It's very frustrating that `aws-lc` has been added to the default dependency list as many intermediate dependencies forget to add "default-features = false", making it in practice a hard dependency of rustls.

Please reconsider making `aws-lc` the default backend. The default backend should require *fewer* dependencies, not more. Any minor performance difference is a secondary concern.

ctz:
This should improve matters for people building aws-lc-rs on windows thanks to the [prebuilt-nasm feature](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/#prebuilt-nasm): fixes #1913.

It also removes `syn` from the dependencies of rustls; a diff of `cargo tree --edges normal --all-features` includes:

```
 │   └── zeroize v1.8.1
-│       └── zeroize_derive v1.4.2 (proc-macro)
-│           ├── proc-macro2 v1.0.86
-│           │   └── unicode-ident v1.0.12
-│           ├── quote v1.0.37
-│           │   └── proc-macro2 v1.0.86 (*)
-│           └── syn v2.0.77
-│               ├── proc-macro2 v1.0.86 (*)
-│               ├── quote v1.0.37 (*)
-│               └── unicode-ident v1.0.12
```


rustls/src/crypto/ring/kx.rs
```diff
@@ -102,11 +133,16 @@ struct KeyExchange {
...
+    pub_key_validator: fn(&[u8]) -> bool,
 }
 
 impl ActiveKeyExchange for KeyExchange {
     /// Completes the key exchange, given the peer's public key.
     fn complete(self: Box<Self>, peer: &[u8]) -> Result<SharedSecret, Error> {
+        match (self.pub_key_validator)(peer) {
+            true => {}
+            false => return Err(PeerMisbehaved::InvalidKeyShare.into()),
+        };
```
cpu, line 145: Did you think the `match` with the empty `true` arm was more readable than an `if !(self.pub_key_validator)(peer) { return Err(PeerMisbehaved::InvalidKeyShare.into()); }` ? The latter seems a bit more concise but I don't feel strongly.

rustls/src/crypto/ring/kx.rs
```diff
@@ -76,22 +91,38 @@ pub static X25519: &dyn SupportedKxGroup = &KxGroup {
...
 /// Ephemeral ECDH on secp256r1 (aka NIST-P256)
 pub static SECP256R1: &dyn SupportedKxGroup = &KxGroup {
     name: NamedGroup::secp256r1,
     agreement_algorithm: &agreement::ECDH_P256,
     fips_allowed: true,
+    pub_key_validator: secp256r1_validator,
 };
 
+fn secp256r1_validator(point: &[u8]) -> bool {
+    matches!((point.len(), point.first()), (65, Some(&0x04)))
```
cpu, line 111: Citation for the `0x04` constant? (I also think Djc's feedback to drop the `&` could be applied here).

If you agree, ditto for L123.

rustls/src/crypto/ring/kx.rs
```diff
@@ -76,22 +91,38 @@ pub static X25519: &dyn SupportedKxGroup = &KxGroup {
     //  are not compliant to SP 800-56Arev3."
     // -- <https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf>
     fips_allowed: false,
+
+    pub_key_validator: x25519_validator,
 };
 
+fn x25519_validator(point: &[u8]) -> bool {
+    point.len() == 32
+}
```
cpu, line 100: This `fn` feels small enough that maybe it could just be in-lined into the `pub_key_valdiator` field initialization as a closure? (If you agree, ditto for `SECP256R1` & `SECP384R1`).

cpu (approved):
> Prepare 0.23.13

Can you add some proposed release notes to the PR desc for this? Thank you! :bow:

rustls/src/crypto/ring/kx.rs
ctz: Added a citation of the 0x04 number, and also refactored this to remove the length checks (there's actually no need for that).

rustls/src/crypto/ring/kx.rs
ctz: I have inlined the x25519 one, and kept the other one separate so it has a name and the citation has one place to live.

ctz:
> > Prepare 0.23.13
> 
> Can you add some proposed release notes to the PR desc for this? Thank you! 🙇

Added

cpu:
Some ideas for the release notes to consider:

> Add new custom-provider crate feature to disable installation of built-in provider. Thanks to @stormshield-gt

Maybe this could have a brief mention of the motivation? Suggested wording: "This may be useful in situations where your application wishes to ensure consistent use of a custom crypto provider throughout the dependency tree." ?

I thought about the same feedback for the ticketer change but decided the motivation is too in the weeds to be worth calling out top-line in changelog. It's in the PR if anyone is especially curious about the change.

> Take aws-lc-rs 1.9 -- this [eases the build-time requirements on Windows](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/#prebuilt-nasm).

Can we add a bit more detail here? In particular I think it would be worth mentioning `nasm` explicitly since there are a number of folks that have reacted to that issue and might overlook this change otherwise.

brodycj:
> * **Add new `custom-provider` crate feature to disable installation of built-in provider**.

I thought this has already been present (and don't see this in the files changed), or am I mistaken somehow?

cpu:
> I thought this has already been present (and don't see this in the files changed), or am I mistaken somehow?

You're correct, the code for this already landed, but it hasn't been included in a published release yet.

This section of the PR description isn't just for new code in this branch, it's also proposing changelog content for the release that covers this branch's work and also the work that's queued but unreleased in `main`. 

Does that help clarify?

brodycj:
> it's also proposing changelog content for the release that covers this branch's work and also the work that's queued but unreleased in `main`.

Got it thanks @cpu ... I think it would be ideal if this could have been more clear in the description

PR #2120, opened 2024-09-11
Send flights of handshake messages in single message

### LINKED ISSUE #2041: Pack multiple handshake messages into single TLS message
At the moment each handshake message we send is contained in its own outer TLS message. This means:

- handshakes take a little more data that is actually necessary, due to repeated encryption overhead.
- there is likely some performance left on the table (since N encryptions of M bytes will be slower than one encryption of N * M bytes).

ctz:
This is generally only valuable for TLS1.3, where we're saving encryption overheads several times over. For TLS1.2 there are no encrypted handshake messages, so we're only saving 20 bytes on the wire in the server -> client direction.

fixes #2041


rustls/src/hash_hs.rs
```diff
@@ -98,10 +98,16 @@ impl HandshakeHash {
...
+        match &m.payload {
+            MessagePayload::Handshake { encoded, .. } => self.update_raw(encoded.bytes()),
+            MessagePayload::HandshakeFlight(payload) => self.update_raw(payload.bytes()),
+            _ => self,
         }
-        self
+    }
+
+    /// Hash/buffer an encoded handshake message.
+    pub(crate) fn add(&mut self, bytes: &[u8]) {
```
cpu, line 109: I'm chewing on the decision to add this wrapper instead of promoting `update_raw()`'s visibility. It feels like the main reason is to expose a name that's a better match to the new purpose of adding a handshake message that's already been encoded as an alternative to `add_message()`.

WDYT about renaming `update_raw()` to `add_raw()` instead?

rustls/src/hash_hs.rs
```diff
@@ -197,6 +207,42 @@ test_for_each_provider! {
...
+        hhb.add_message(&end_of_early_data_flight);
+        assert_eq!(hhb.start_hash(&SHA256).current_hash().as_ref(),
+                   SHA256.hash(b"\x0e\x00\x00\x00\x05\x00\x00\x00").as_ref());
+
+        let mut hh = HandshakeHashBuffer::new().start_hash(&SHA256);
+        hh.add_message(&server_hello_done_message);
+        hh.add_message(&app_data_ignored);
+        hh.add_message(&end_of_early_data_flight);
+        assert_eq!(hh.current_hash().as_ref(),
+                   SHA256.hash(b"\x0e\x00\x00\x00\x05\x00\x00\x00").as_ref());
```
cpu, line 242: I think the difference between these two test cases is in the coverage they produce, but it's not obvious from a first glance. Worth a comment that calls out the difference in approach for A vs B?

rustls/src/msgs/message/mod.rs
```diff
@@ -44,6 +48,10 @@ impl<'a> MessagePayload<'a> {
         }
     }
 
+    pub(crate) fn handshake_flight(encoded: Vec<u8>) -> Self {
```
cpu, line 51: nit: visibility is sub `pub` and so should be lower in order.

rustls/tests/unbuffered.rs
```diff
@@ -30,9 +30,6 @@ fn tls12_handshake() {
             "Ok(EncodeTlsData)",
             "Ok(TransmitTlsData)",
             "Ok(BlockedHandshake)",
-            "Ok(BlockedHandshake)",
-            "Ok(BlockedHandshake)",
-            "Ok(BlockedHandshake)",
```
cpu, line 35: it's interesting to observe the optimization so directly in this API's expected states :+1:

rustls/src/common_state.rs
```diff
@@ -974,5 +979,40 @@ impl KxState {
...
+            } else {
+                ProtocolVersion::TLSv1_2
+            },
+            payload: MessagePayload::handshake_flight(self.body),
+        };
+        common.send_msg(m, TLS13);
+    }
+}
+
+pub(crate) type HandshakeFlightTls12<'a> = HandshakeFlight<'a, false>;
```
cpu, line 1015: Is the TLS 1.2 use-case landed alongside the new types to avoid dead code warnings? I think this diff would be easier to digest if it were separated out like the TLS 1.3 case but I wager that might produce warnings and I don't feel very strongly.

rustls/src/common_state.rs
```diff
@@ -974,5 +979,40 @@ impl KxState {
     }
 }
 
+pub(crate) struct HandshakeFlight<'a, const TLS13: bool> {
```
cpu, line 982: I briefly considered the restrictiveness of the const generic bool here vs something more expressive but decided it was premature optimization and this seems quite reasonable after all.

rustls/tests/api.rs
```diff
@@ -5222,7 +5222,7 @@ fn test_client_sends_helloretryrequest() {
         let wrlen = server.write_tls(&mut pipe).unwrap();
         assert!(wrlen > 200);
         assert_eq!(pipe.writevs.len(), 1);
-        assert!(pipe.writevs[0].len() == 5); // server hello / encrypted exts / cert / cert-verify / finished
+        assert_eq!(pipe.writevs[0].len(), 2); // { server hello / encrypted exts / cert / cert-verify} / finished
```
cpu, line 5225: ultra nit: space in the comment before closing `}` for consistency with opening `// { server ..`

rustls/tests/api.rs
```diff
@@ -3652,9 +3652,9 @@ fn vectored_write_for_server_handshake_with_half_rtt_data() {
         let mut pipe = OtherSession::new(&mut client);
```
cpu, line 3652: Nice to see the improvements reflected here too :rocket:

cpu (approved):
This is really nice, outside of any perf benefits I think it has an improvement on the overall readability/structure of the code.

rustls/src/common_state.rs
```diff
@@ -974,5 +979,40 @@ impl KxState {
...
+    pub(crate) fn add(&mut self, hs: HandshakeMessagePayload<'_>) {
+        let start_len = self.body.len();
+        hs.encode(&mut self.body);
+        self.transcript
+            .add(&self.body[start_len..]);
+    }
+
+    pub(crate) fn finish(self, common: &mut CommonState) {
+        let m = Message {
+            version: if TLS13 {
```
djc, line 1004: Nit: `match` would save you one line. Maybe inline `m`?

rustls/src/common_state.rs
djc: In general, not sure I'm convinced of the benefit of this type alias.

rustls/src/msgs/message/mod.rs
djc: Given only one library caller and one test caller, maybe we don't need this at all?

djc (approved):
This looks great!

djc:
> This is really nice, outside of any perf benefits I think it has an improvement on the overall readability/structure of the code.

Agreed, -11 lines for an efficiency improvement seems like a really good deal.

rustls/src/common_state.rs
ctz: Yes I would like to avoid the true/false thing leaking out of this file, as far as possible.

My second choice would be to stop this type being generic like this, and just have two different constructors (eg, `new_tls12()`, `new_tls13()` or similar)  and rely on everything being inlined into their uses.

rustls/src/msgs/message/mod.rs
ctz: Dropped.

PR #2123, opened 2024-09-13
Lower noise in ci-bench

### LINKED ISSUE #1494: CI benchmarks: use cachegrind "client requests" for more precise instruction counts
From [this article](https://kobzol.github.io/rust/rustc/2023/09/23/rustc-runtime-benchmarks.html):

> Valgrind has support for [client requests](https://valgrind.org/docs/manual/manual-core-adv.html#manual-core-adv.clientreq), which allow the profiled program (amongst other things) to selectively enable and disable instrumentation for parts of the program.

They are not yet available for cachegrind, but it looks like support is coming in the next release. Using client requests would greatly improve our benchmarking harness, making it possible to ignore setup code when counting instructions. The functionality should be accessible through the [crabgrind](https://github.com/2dav/crabgrind) crate.

The current workaround to discard the instruction counts originating from setup code is to run the setup code in isolation and subtract the measured icount from the full benchmark's icount. This works well enough in practice, but gets confusing when you are investigating a regression, because the cachegrind diff does include the benchmark's setup code.

ctz:
For me this reduces maximum run-to-run noise to 0.29%, from about 0.61% on main.

There will be a discontinuity in benchmark numbers in this PR, as we're reducing the scope of what is measured.

fixes #1494

### DIFF
--- a/ci-bench/README.md
+++ b/ci-bench/README.md
@@ -2,7 +2,7 @@
 
 This crate is meant for CI benchmarking. It has two modes of operation:
 
-1. Measure CPU instructions using `cachegrind`.
+1. Measure CPU instructions using `callgrind`.
 2. Measure wall-time (runs each benchmark multiple times, leaving it to the caller to do statistical
    analysis).
 
@@ -35,10 +35,10 @@ handshake_no_resume_1.3_rsa_aes_client,4212770
 ...
 ```
 
-In the `cachegrind` subdirectory you will find output files emitted by the `cachegrind` tool, which
+In the `callgrind` subdirectory you will find output files emitted by the `callgrind` tool, which
 are useful to report detailed instruction count differences when comparing two benchmark runs. This
-subdirectory also contains log information from cachegrind itself (in `.log` files), which can be
-used to diagnose unexpected cachegrind crashes.
+subdirectory also contains log information from callgrind itself (in `.log` files), which can be
+used to diagnose unexpected callgrind crashes.
 
 ### Running all benchmarks in wall-time mode
 
@@ -86,14 +86,14 @@ here are some high-level considerations that can help you hack on the crate.
 ### Environment configuration
 
 An important goal of this benchmarking setup is that it should run with minimal noise. Measuring CPU
-instructions using `cachegrind` yields excellent results, regardless of your environment. The
+instructions using `callgrind` yields excellent results, regardless of your environment. The
 wall-time benchmarks, however, require a more elaborate benchmarking environment: running them on a
 laptop is too noisy, but running them on a carefully configured bare-metal server yields accurate
 measurements (up to 1% resolution, according to our tests).
 
 ### Instruction count mode
 
-Using `cachegrind` has some architectural consequences because it operates at the process level
+Using `callgrind` has some architectural consequences because it operates at the process level
 (i.e. it can count CPU instructions for a whole process, but not for a single function). The most
 important consequences when running in instruction count mode are:
 
--- a/ci-bench/README.md
+++ b/ci-bench/README.md
@@ -93,20 +93,16 @@ measurements (up to 1% resolution, according to our tests).
 
 ### Instruction count mode
 
-Using `callgrind` has some architectural consequences because it operates at the process level
-(i.e. it can count CPU instructions for a whole process, but not for a single function). The most
-important consequences when running in instruction count mode are:
+Instruction counting is done with `callgrind`, and is precise in the sense that only operations
+we want to measure are included.  We tell `callgrind` to start with collection disabled,
+with `--collect-atstart=no`.  Then we use
+[client requests](https://valgrind.org/docs/manual/cl-manual.html#cl-manual.clientrequests) via
+the [crabgrind crate](https://docs.rs/crabgrind/latest/crabgrind/) to enable and disable collection
+(see `callgrind::CountInstructions`).
 
-- Since we want to measure server and client instruction counts separately, the benchmark runner
-  spawns two child processes for each benchmark (one for the client, one for the server) and pipes
-  their stdio to each other for communication (i.e. stdio acts as the transport layer).
-- There is a no-op "benchmark" that measures the overhead of starting up the child process, so
-  we can subtract it from the instruction count of the real benchmarks and reduce noise.
-- Since we want to measure individual portions of code (e.g. data transfer after the handshake),
-  there is a mechanism to subtract the instructions that are part of a benchmark's setup.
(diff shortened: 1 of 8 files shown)


ci-bench/src/cachegrind.rs
```diff
@@ -291,3 +291,24 @@ pub fn diff(baseline: &Path, candidate: &Path, scenario: &str) -> anyhow::Result
 
     Ok(diff)
 }
+
+pub(crate) fn pause_instruction_count() -> PauseGuard {
+    PauseGuard::new()
+}
+
+pub(crate) struct PauseGuard;
```
cpu, line 304: Maybe worth an internal rustdoc comment to warn this won't do the right thing (AFAICT) if you end up creating multiple `PauseGuard`'s and drop one but not the other. I think the first that gets dropped would toggle collection off and the next drop would turn it back on?

Probably not a concern for real usage :shrug:

ci-bench/src/main.rs
```diff
@@ -384,6 +384,23 @@ fn aws_lc_rs_ticketer() -> Arc<dyn rustls::server::ProducesTickets> {
...
+        ..base
+    }
+}
+
+#[derive(Debug)]
+struct NotRandom;
+
+impl rustls::crypto::SecureRandom for NotRandom {
+    fn fill(&self, buf: &mut [u8]) -> Result<(), rustls::crypto::GetRandomFailed> {
+        buf.fill(0x5a);
```
cpu, line 399: oblig. https://xkcd.com/221/ :game_die: :laughing:

cpu (commented):
Nice!

djc (approved):

ci-bench/src/cachegrind.rs
```diff
@@ -291,3 +291,29 @@ pub fn diff(baseline: &Path, candidate: &Path, scenario: &str) -> anyhow::Result
 
     Ok(diff)
 }
+
+pub(crate) fn reset_instruction_count() {
+    #[cfg(target_os = "linux")]
+    crabgrind::callgrind::zero_stats();
```
ctz, line 297: sadly this actually does jack, because we're running this under _cachegrind_ and not _callgrind_  -- the version of valgrind we have in ci does not have a programmatic control interface for cachegrind (though does in later versions)

will have a look to see if we can use callgrind instead (we turn off the cache simulation, so they are should be very similar)

djc: Can we upgrade our version of valgrind?

ctz: Unfortunately the test runner is on debian stable, installing a newer valgrind from debian unstable would work, but would give us an odd mix of the two. I think if I get into changing the ansible scripts for this, I would rather spend the time installing docker to make it easier to (eg) rev the rust compiler too.

cpu: Will benchmarking from Docker introduce any significant overhead?

ctz: I think some networking/filesystem performance can be worse, but pure CPU should be identical.

ctz: callgrind seems to work well now, though renaming a bunch of types/files called CachegrindSomething remains to do.

ci-bench/src/main.rs
```diff
@@ -384,6 +384,23 @@ fn aws_lc_rs_ticketer() -> Arc<dyn rustls::server::ProducesTickets> {
     aws_lc_rs::Ticketer::new().unwrap()
 }
 
+fn derandomize(base: rustls::crypto::CryptoProvider) -> rustls::crypto::CryptoProvider {
```
cpu, line 387: nit: I think the `derandomize` signature is still using the fully qualified `CryptoProvider` type even though it's now imported in-scope.

ci-bench/src/cachegrind.rs
```diff
@@ -233,61 +233,55 @@ impl Sub for InstructionCounts {
...
-        .arg("--mod-funcname=s/17h[0-9a-f]+E\\.llvm\\.\\d+/E/")
+    // callgrind_annotate formats the callgrind output file, suitable for comparison with
+    // callgrind_differ
+    let callgrind_annotate_base = Command::new("callgrind_annotate")
         .arg(
             baseline
                 .join(CACHEGRIND_OUTPUT_SUBDIR)
                 .join(scenario),
         )
+        .arg("--auto=no")
```
cpu: Maybe worth a comment here indicating why we want to turn off automatic annotation of source files that contributed to the event count threshold?

cpu (approved):
:racehorse:

ci-bench/src/cachegrind.rs
```diff
@@ -233,61 +233,55 @@ impl Sub for InstructionCounts {
...
-    let diff =
-        String::from_utf8(cg_annotate.stdout).context("cg_annotate produced invalid UTF8")?;
+    let string_base = String::from_utf8(callgrind_annotate_base.stdout)
+        .context("callgrind_annotate produced invalid UTF8")?;
+    let string_candidate = String::from_utf8(callgrind_annotate_candidate.stdout)
+        .context("callgrind_annotate produced invalid UTF8")?;
 
-    fs::remove_file(tmp_path).ok();
+    // TODO: reinstate actual diffing, using `callgrind_differ` crate
+    let diff =
```
djc: Nit: inline the `diff` binding?

ci-bench/src/benchmark.rs
```diff
@@ -27,40 +26,13 @@ pub fn validate_benchmarks(benchmarks: &[Benchmark]) -> anyhow::Result<()> {
...
-    }
-
     Ok(())
 }
 
 /// Specifies how the results of a particular benchmark should be reported
 pub enum ReportingMode {
     /// All instructions are reported
     AllInstructions,
-    /// All instructions are reported, after subtracting the instructions of the setup code
```
djc, line 59: Is the enum still useful?

ci-bench/README.md
```diff
@@ -2,7 +2,7 @@
 
 This crate is meant for CI benchmarking. It has two modes of operation:
 
-1. Measure CPU instructions using `cachegrind`.
+1. Measure CPU instructions using `callgrind`.
```
djc, line 5: Should all of this be in the previous commit that actually switched?

djc (approved):
Nice!

ci-bench/src/benchmark.rs
ctz: removed and squashed removal into the original commit

PR #2126, opened 2024-09-19
examples: convert tlsserver-mio from docopt to clap

djc:
Not sure how best to test that this works...

I can take care of the other examples too, but figured it'd be good to get feedback on one first.


README.md
```diff
@@ -139,43 +139,8 @@ diving in to the more complex MIO examples.
...
-    --no-tickets        Disable session ticket support.
-    --no-sni            Disable server name indication support.
-    --insecure          Disable certificate verification.
-    --verbose           Emit log output.
-    --max-frag-size M   Limit outgoing messages to M bytes.
-    --version, -v       Show tool version.
-    --help, -h          Show this screen.
-```
+
+The MIO client example program is named `tlsclient-mio`.
```
djc, line 143: Given that you can trivially grab the interface from the command line (and you need a local clone to run the code anyway), it doesn't seem all that useful to replicate the full interface description here, making the README significantly longer.

ctz: Seems reasonable.

examples/src/bin/tlsserver-mio.rs
```diff
@@ -417,79 +416,65 @@ impl OpenConnection {
...
+    /// private key, the last should be a root CA).
+    #[clap(long)]
+    certs: Option<String>,
+    /// Perform client certificate revocation checking using the DER-encoded CRL from the given
+    /// files.
+    #[clap(long)]
+    crl: Vec<String>,
+    /// Read private key from the given file. This should be a private key in PEM format.
+    #[clap(long)]
+    key: Option<String>,
```
ctz, line 458:
```suggestion
    key: String,
```
(i think this should be mandatory, to get clap to generate the error message?)

examples/src/bin/tlsserver-mio.rs
```diff
@@ -417,79 +416,65 @@ impl OpenConnection {
...
+    #[clap(long)]
+    suite: Vec<String>,
+    /// Negotiate the given protocols using ALPN.
+    #[clap(long)]
+    proto: Vec<String>,
+    /// Read server certificates from the given file. This should contain PEM-format certificates
+    /// in the right order (the first certificate should certify the end entity, matching the
+    /// private key, the last should be a root CA).
+    #[clap(long)]
+    certs: Option<String>,
```
ctz, line 451:
```suggestion
    certs: String,
```
(ditto)

examples/src/bin/tlsserver-mio.rs
```diff
@@ -417,79 +416,65 @@ impl OpenConnection {
...
+    require_auth: bool,
+    /// Disable stateful session resumption.
+    #[clap(long)]
+    no_resumption: bool,
+    /// Support tickets (stateless resumption).
+    #[clap(long)]
+    tickets: bool,
+    /// Support receiving this many bytes with 0-RTT.
+    #[clap(long)]
+    max_early_data: u32,
```
ctz, line 477:
```suggestion
    #[clap(long, default_value = "0")]
    max_early_data: u32,
```
(otherwise this becomes mandatory)

examples/src/bin/tlsserver-mio.rs
ctz: if we do that, we can get rid of `args.key.as_ref().expect("--key option missing")` below

ctz:
Thanks for doing this, happy to do the legwork on the remaining samples to eliminate docopt altogether after this is finished.

README.md
```diff
@@ -197,59 +162,10 @@ TLS error: InvalidCertificate(Expired)
...
-                        VERSION instead.  May be used multiple times.
-    --suite SUITE       Disable default cipher suite list, and use
-                        SUITE instead.  May be used multiple times.
-    --proto PROTOCOL    Negotiate PROTOCOL using ALPN.
-                        May be used multiple times.
-    --verbose           Emit log output.
-    --version, -v       Show tool version.
-    --help, -h          Show this screen.
-```
+The MIO server example program is named `tlsserver-mio`.
```
cpu, line 168: Nit: whitespace after the `## ` heading to match the client example program header?

examples/src/bin/tlsserver-mio.rs
```diff
@@ -417,79 +417,65 @@ impl OpenConnection {
...
-#[derive(Debug, Deserialize)]
+/// Runs a TLS server on :PORT.  The default PORT is 443.
+///
+/// `echo' mode means the server echoes received data on each connection.
+///
+/// `http' mode means the server blindly sends a HTTP response on each connection.
+///
+/// `forward' means the server forwards plaintext to a connection made to `localhost:fport`.
+///
+/// `--certs' names the full certificate chain, `--key' provides the private key.
```
cpu, line 428: This looks pre-existing, but I think the opening backticks on a number of these lines are mismatched with closing single quotes:

```suggestion
/// Runs a TLS server on :PORT.  The default PORT is 443.
///
/// `echo` mode means the server echoes received data on each connection.
///
/// `http` mode means the server blindly sends a HTTP response on each connection.
///
/// `forward` means the server forwards plaintext to a connection made to `localhost:fport`.
///
/// `--certs` names the full certificate chain, `--key` provides the private key.
```

examples/src/bin/tlsserver-mio.rs
```diff
@@ -417,79 +416,65 @@ impl OpenConnection {
...
+    suite: Vec<String>,
+    /// Negotiate the given protocols using ALPN.
+    #[clap(long)]
+    proto: Vec<String>,
+    /// Read server certificates from the given file. This should contain PEM-format certificates
+    /// in the right order (the first certificate should certify the end entity, matching the
+    /// private key, the last should be a root CA).
+    #[clap(long)]
+    certs: String,
+    /// Perform client certificate revocation checking using the DER-encoded CRL from the given
```
cpu, line 453: nit: DER-encoded CRL**s** plural, to match "files"?

cpu (approved):
Very nice :+1:

cpu:
Separate from the docopt -> clap change, how do folks feel about the "mio" part of the example names? I've heard from at least one person looking for something sort of like a `openssl s_client` and `openssl s_server` that these examples might be better named `rustls_client` and `rustls_server`. 

It feels a little bit like the MIO part is an implementation detail, and perhaps not the "main draw" of the examples. WDYT?

examples/src/bin/tlsserver-mio.rs
ctz: Ah good, catch. These were originally old school GNU-style smart quotes; now this is a markdown doc `` is more suitable.

cpu: TIL "GNU-style smart quotes" :laughing: :water_buffalo:

djc:
> It feels a little bit like the MIO part is an implementation detail, and perhaps not the "main draw" of the examples. WDYT?

Seems fair. Since we have the `simpleclient`/`simpleserver` as well as minimal implementations, we could argue that these are less "example code" and more CLI interfaces, in which case renaming makes sense but I'd probably also migrate from using mio directly to just using tokio-rustls for these, since the rustls <-> mio interworkings seem extremely niche in practice.

cpu:
> I'd probably also migrate from using mio directly to just using tokio-rustls for these, since the rustls <-> mio interworkings seem extremely niche in practice.

I'm a little bit torn on that direction. If we made that change, would there be an argument the examples should live in the `tokio-rustls` repo instead? I also think there's some value in having an example that sits between the simple examples that use `complete_io` and something that defers to Tokio for the guts of the I/O. :thinking: On the other hand, I agree with you that few folks are likely to want to operate at that level and MIO feels niche.

djc:
> > I'd probably also migrate from using mio directly to just using tokio-rustls for these, since the rustls <-> mio interworkings seem extremely niche in practice.
> 
> I'm a little bit torn on that direction. If we made that change, would there be an argument the examples should live in the `tokio-rustls` repo instead? I also think there's some value in having an example that sits between the simple examples that use `complete_io` / the stream abstractions and something that defers to Tokio for the guts of the I/O. 🤔 On the other hand, I agree with you that few folks are likely to want to operate at that level and MIO feels niche.

You were the one arguing that these examples are effectively providing a CLI for rustls, which makes sense to me. I think any value from sitting between the simple examples and deferring to Tokio for I/O is extremely niche. But as discussed recently, I'm kind of down on having many examples...

> Edit; also don't want to derail this PR too much. Maybe best to file a separate issue?

Yeah, wasn't going to pick it up for this PR.

ctz (approved):

PR #2134, opened 2024-09-25
add DerefMut to common state for UnbufferedConnectionCommon

stevefan1999-personal:
This is needed because `CommonState::send_close_notify()` is needed to be called when you need to do `send_close_notify` for fatal case.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -792,6 +792,13 @@ impl<T> Deref for UnbufferedConnectionCommon<T> {
     }
 }
 
+impl<T> DerefMut for UnbufferedConnectionCommon<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.core.common_state
+    }
+}
+
+
 pub(crate) struct ConnectionCore<Data> {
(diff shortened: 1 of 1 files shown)


ctz:
Is this so you can call `CommonState::send_close_notify()`? Because there is `queue_close_notify()` on the unbuffered WriteTraffic state.

cpu:
>> Because there is queue_close_notify() on the unbuffered WriteTraffic state.
>>
> Exactly. I have to handle these cases:

I think the missing piece is why you can't use `queue_close_notify()` from `WriteTraffic` state for this. Can you be more explicit about that aspect?

cpu (approved):
I'm still not sure I understand the motivation fully, but I don't feel very strongly either.

> codecov/patch — 0.00% of diff hit (target 94.63%)

It would be nice to have a small smoketest to make sure we don't regress this by mistake.

Can you also squash your commits?

cpu:
@stevefan1999-personal ping?

djc:
Unless we have a clear(er) motivation I don't think we should merge this.

cpu:
> Unless we have a clear(er) motivation I don't think we should merge this.

Agreed, and the OP isn't responding so let's close this for now.

PR #2135, opened 2024-09-26
Eliminate an allocation in signature verification

ctz:
The signed message has a pretty tight upper bound, so we can avoid a Vec allocation here.

### DIFF
--- a/rustls/src/tls13/mod.rs
+++ b/rustls/src/tls13/mod.rs
@@ -1,4 +1,3 @@
-use alloc::vec::Vec;
 use core::fmt;
 
 use crate::crypto;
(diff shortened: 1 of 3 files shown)


rustls/src/tls13/mod.rs
```diff
@@ -87,22 +86,41 @@ impl fmt::Debug for Tls13CipherSuite {
...
     handshake_hash: &hash::Output,
-    context_string_with_0: &[u8],
-) -> Vec<u8> {
-    let mut msg = Vec::new();
-    msg.resize(64, 0x20u8);
-    msg.extend_from_slice(context_string_with_0);
-    msg.extend_from_slice(handshake_hash.as_ref());
-    msg
+    context_string_with_0: &[u8; 34],
+) -> VerifyMessage {
```
cpu, line 101: This feels like a constructor for `VerifyMesage` now. I think it'd be more natural as `VerifyMessage::new()`.

djc (approved):

cpu (approved):

PR #2136, opened 2024-09-26
Jbp hybrid kx optimisation

ctz:
This is draft while I write some extra tests, but the substance can certainly be reviewed.

This is towards #2056


rustls/src/crypto/mod.rs
```diff
@@ -505,6 +505,81 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    /// `hybrid_pair`. It only enables an optimization.
+    ///
+    /// # Background
+    /// Rustls always sends a presumptive key share in its ClientHello, using
+    /// (absent any other information) the first item in [`CryptoProvider::kx_groups`].
+    /// If the server accepts the client's selection, it can complete the handshake
+    /// using that key share.  If not, the server sends a `HelloRetryRequest` instructing
+    /// the client to send a different key share instead.
+    ///
+    /// This costs an extra round trip, and wastes the key exchange computation
```
cpu, line 525: Might be a smidge clearer to say:
```
// "This request costs an extra round trip ...
```
I think it better connects to the last sentence of the previous paragraph.

rustls/src/crypto/mod.rs
```diff
@@ -505,6 +505,81 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    /// 1. If the return value is None, nothing further happens.
+    /// 2. If the given [`ActiveKeyExchange::group()`] does not appear in
+    ///    [`CryptoProvider::kx_groups`], nothing further happens.
+    /// 3. The given [`ActiveKeyExchange::pub_key()`] is added to the `ClientHello`,
+    ///    after the hybrid entry.
+    ///
+    /// Then, one of three things may happen when the server replies to the `ClientHello`:
+    ///
+    /// 1. The server sends a `HelloRetryRequest`.  Everything is thrown away and
+    ///    we start again.  Note that `hybrid_pair` is never called in this case.
```
cpu, line 560: Maybe worth clarifying here. It's been called for the initial handshake and I think the intent is to communicate it won't be called again for the HRR right?

> Note that `hybrid_pair()` is not called again in this case.

I also think we like having parens on fn names.

rustls/src/crypto/mod.rs
```diff
@@ -505,6 +505,81 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    /// 2. The server agrees to our hybrid key exchange: rustls calls
+    ///    [`ActiveKeyExchange::complete()`] consuming `self`.  `into_hybrid_pair`
```
cpu, line 562: `into_hybrid_pair` -> `into_hybrid_pair()`?

rustls/src/crypto/mod.rs
```diff
@@ -505,6 +505,81 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    ///    consumed by [`ActiveKeyExchange::complete()`].
+    fn hybrid_pair(&self) -> Option<&dyn ActiveKeyExchange> {
+        None
+    }
+
+    /// Consumes this hybrid key exchange, yielding the underlying classical
+    /// key exchange.
+    ///
+    /// If [`ActiveKeyExchange::hybrid_pair()`] returns non-None, this
+    /// _must_ return non-None.
```
cpu, line 576: nit: I would find it easier to parse this without "non-None", maybe  "returns Some(x), this _must_ return Some(x)"`?

Should there also be some mention that it should be returning the _same_ `dyn ActiveKeyExchange` as `hybrid_pair()`?

rustls-post-quantum/src/lib.rs
```diff
@@ -158,6 +158,14 @@ impl ActiveKeyExchange for Active {
         Ok(SharedSecret::from(&combined.0[..]))
     }
 
+    fn hybrid_pair(&self) -> Option<&dyn ActiveKeyExchange> {
+        Some(self.x25519.as_ref())
+    }
+
+    fn into_hybrid_pair(self: Box<Self>) -> Option<Box<dyn ActiveKeyExchange>> {
+        Some(self.x25519)
+    }
```
cpu, line 167: It feels a little bit like these two things should be coupled together more tightly, like `hybrid_pair()` should be returning something that can act as a `&dyn ActiveKeyExchange`, but also be translated to a `Box<dyn ActiveKeyExchange>` once the decision point to use it has been reached vs asking the `ActiveKeyExchange` in two operations.

I'm not sure this is a fully baked thought yet but maybe it's something you already considered?

rustls/src/client/tls13.rs
```diff
@@ -88,14 +88,47 @@ pub(super) fn handle_server_hello(
...
+        // primary
+        our_key_share
+    } else if our_key_share
+        .hybrid_pair()
+        .map(|pair| pair.group())
+        == Some(their_key_share.group)
+    {
+        let secondary = our_key_share
+            .into_hybrid_pair()
+            .expect("into_hybrid_pair() must by supported if hybrid_pair() is");
```
cpu, line 104: Thinking back to my not fully baked review comment, I think this `expect()` might fall away if the `hybrid_pair()` return was also sufficiently capable of being turned into the `Box`'d form we need.

rustls/src/client/tls13.rs
```diff
@@ -88,14 +88,47 @@ pub(super) fn handle_server_hello(
...
+        debug_assert_eq!(secondary.group(), their_key_share.group);
```
cpu, line 105: good defensive check :+1:

cpu (commented):
Here's some initial feedback. I think the overall idea makes sense :+1:

rustls-post-quantum/src/lib.rs
ctz: The tricky part I think is avoiding consuming the original `ActiveKeyExchange` until the right time, and not breaking the API property that we want, which is that any given `ActiveKeyExchange` is only `complete()`ed at most once.

rustls/src/crypto/mod.rs
ctz: Actually I meant to talk about `into_hybrid_pair()` in this clause.

rustls/src/crypto/mod.rs
```diff
@@ -505,6 +505,83 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+
+    /// Consumes this hybrid key exchange, yielding the underlying classical
+    /// key exchange.
+    ///
+    /// If [`ActiveKeyExchange::hybrid_pair()`] returns Some(_), this
+    /// _must_ return Some(_), and return the same underlying `dyn ActiveKeyExchange`
+    /// object.
+    ///
+    /// See [`ActiveKeyExchange::hybrid_pair()`] for what this is for.
+    fn into_hybrid_pair(self: Box<Self>) -> Option<Box<dyn ActiveKeyExchange>> {
```
djc, line 581: Given the description of "consumes this hybrid exhcnage, yielding the underlying classical exchange", the name "into_hybrid_pair()` seems pretty confusing to me.

rustls-post-quantum/src/lib.rs
djc: But those sound invariants the Rust type system is pretty good at enforcing? I'm not sure I currently understand it enough to write it, but suggest you try and we can discuss the design?

rustls/src/client/hs.rs
```diff
@@ -300,8 +300,25 @@ fn emit_client_hello_for_retry(
...
-        exts.push(ClientExtension::KeyShare(vec![key_share]));
+        let mut shares = vec![KeyShareEntry::new(key_share.group(), key_share.pub_key())];
+
+        if retryreq.is_none() {
+            // Only for the initial client hello, see if we can send a second KeyShare
+            // for "free".  We only do this if the same algorithm is also supported
+            // separately by our provider for this version (`find_kx_group` looks that up).
+            if let Some(hybrid_pair) = key_share.hybrid_pair().filter(|pair| {
+                config
+                    .find_kx_group(pair.group(), ProtocolVersion::TLSv1_3)
```
djc, line 311: It seems kind of a confusing that `pair` has a single `group()`? Shouldn't it conceptually have two, or something?

rustls/src/crypto/mod.rs
ctz: mmm, it's not the clearest. i think the important points are: "if self is a hybrid key exchange, give me half that might be supported separately". so, maybe `into_hybrid_half`?

cpu: Would describing it as a "fallback_exchange()" be confusing? :house: :paintbrush:

ctz: for me "fallback" isn't conveying the right thing, but also now i think "hybrid" is a property of self, not the returned thing. will re-read the RFC draft and see if anything jumps out at me.

ctz: trying out `hybrid_component()` and `split_hybrid_component()` for size

ctz: Going to resolve this thread; I'm happy with this naming and I think it's matching the language in https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design-11 is a justifiable approach.

rustls-post-quantum/src/lib.rs
ctz: As an alternative approach, how do you both feel about this?

https://github.com/rustls/rustls/compare/main...jbp-hybrid-kx-optimisation-2

There is one decision point for whether we have a hybrid kx, but at the expense of some more surface (because we can't just call `pub_key()` and such on a different object.)

djc: This definitely looks like an improvement. I wonder if it would make sense to, instead of `into_hybrid()`, have `ActiveKeyExchange::complete()` yield an enum like `enum KeyExchangeResult { Complete(SharedSecret), Hybrid(Box<dyn ActiveHybridKeyExchange>) }` so that control over whether hybrid agreement is used rests with the trait implementer rather than protocol code?

Also `component` feels a little generic for the `ActiveHybridKeyExchange` methods, do we want "hybrid" to be fully generic or maybe clarify the purpose of "classical" vs "post_quantum"? And am I correct in thinking that the order doesn't matter?

@durumcrustulum not sure if you have time for this kind of thing, but perhaps you have Opinions?

ctz: > This definitely looks like an improvement. I wonder if it would make sense to, instead of `into_hybrid()`, have `ActiveKeyExchange::complete()` yield an enum like `enum KeyExchangeResult { Complete(SharedSecret), Hybrid(Box<dyn ActiveHybridKeyExchange>) }` so that control over whether hybrid agreement is used rests with the trait implementer rather than protocol code?

I don't think that works, because the decision is actually taken by the remote server; and that controls whether we call `complete()`, or `component_complete()` (and these are the operations which need a 100% watertight at-most-once guarantee). Maybe that will be clearer once I integrate this into the protocol parts of the crate.

(That's ignoring the desire I have to avoid breaking changes to `ActiveKeyExchange` for what remains a performance enhancement.)

djc: > I don't think that works, because the decision is actually taken by the remote server; and that controls whether we call `complete()`, or `component_complete()` (and these are the operations which need a 100% watertight at-most-once guarantee). Maybe that will be clearer once I integrate this into the protocol parts of the crate.

Ahh, yeah, that makes sense then.

rustls/src/crypto/mod.rs
```diff
@@ -483,6 +483,83 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    ///    consumed by [`ActiveKeyExchange::complete()`].
+    fn hybrid_component(&self) -> Option<&dyn ActiveKeyExchange> {
+        None
+    }
+
+    /// Consumes this hybrid key exchange, yielding the underlying classical
+    /// key exchange.
+    ///
+    /// If [`ActiveKeyExchange::hybrid_component()`] returns Some(_), this
+    /// _must_ return Some(_), and return the same underlying `dyn ActiveKeyExchange`
```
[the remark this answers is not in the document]
ctz: (fell away)

rustls-post-quantum/src/lib.rs
ctz: Ah, damn, this won't work after all.

The main issue is `into_hybrid()` consumes the receiver object, which makes it not useful in the not-hybrid case.

Changing the return type to return the original receiver (ie, `Result<Box<dyn ActiveHybridKeyExchange>, Box<dyn ActiveKeyExchange>>`) doesn't work because `self: Box<Self>` cannot become a `Box<dyn ActiveKeyExchange>` without `Self: Sized`, and `Self: Sized` means we can't call it on a trait object.

Back to the drawing board...

rustls/src/client/tls13.rs
```diff
@@ -88,14 +88,46 @@ pub(super) fn handle_server_hello(
...
 
-    if our_key_share.group() != their_key_share.group {
+    // Determine if the whole (`our_key_share`) or
+    // `our_key_share::hybrid_component()` was selected by the
+    // server.
+    let our_key_share = if our_key_share.group() == their_key_share.group {
+        our_key_share
+    } else if our_key_share
+        .hybrid_component()
+        .map(|component| component.group())
```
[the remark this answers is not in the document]
ctz: This fell away

rustls/src/client/tls13.rs
```diff
@@ -217,6 +217,46 @@ pub(super) fn handle_server_hello(
...
+
+impl KeyExchangeChoice {
+    /// Decide between `our_key_share` or `our_key_share.hybrid_component()`
+    /// based on the selection of the server expressed in `their_key_share`.
+    fn new(
+        config: &Arc<ClientConfig>,
+        cx: &mut ClientContext<'_>,
+        our_key_share: Box<dyn ActiveKeyExchange>,
+        their_key_share: &KeyShareEntry,
+    ) -> Option<Self> {
```
djc, line 233: Maybe `Result` would be better here, even if it's just `Result<Self, ()>`?

rustls/src/crypto/mod.rs
```diff
@@ -484,6 +484,87 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+    /// 1. The server sends a `HelloRetryRequest`.  Everything is thrown away and
+    ///    we start again.
+    /// 2. The server agrees to our hybrid key exchange: rustls calls
+    ///    [`ActiveKeyExchange::complete()`] consuming `self`.
+    /// 3. The server agrees to our classical key exchange: rustls calls
+    ///    [`ActiveKeyExchange::complete_hybrid_component()`] which
+    ///    discards the hybrid key data, and completes just the classical key exchange.
+    fn hybrid_component(&self) -> Option<(NamedGroup, &[u8])> {
+        None
+    }
```
djc, line 547: Still noodling on this, and the `unreachable!()` in `complete_hybrid_component()`... Why couldn't we yield an `enum` here that contains either `Self` or some new type (somewhat like `KeyExchangeChoice`) that represents the ability to do a hybrid key exchange?

ctz: I think the receiver would need to be `self: Box<Self>` or else the impls of it need to do run-time typing to get their state back, but that does mean you can't go back from that to `Box<dyn ActiveKeyExchange>` in the default impl.

Obviously passing the original `Box<dyn ActiveKeyExchange>` alongside a normal receiver (either the receiver borrows the object before it is moved into the argument, or it is moved too and the parameter can't be moved a second time.)

ctz: @djc @cpu any further thoughts on this?

cpu: I think I'm comfortable with the approach in-branch and haven't come up with any workable ideas that move the needle.

djc: Going to give it one more think tomorrow morning when I'm fresh.

djc: So started noodling around with this and still wondering whether it's possible to figure this out:

```rust
    fn hybrid_component(self: Box<Self>) -> MaybeHybrid {
        MaybeHybrid::Single(Box::new(self))
    }

enum MaybeHybrid {
    Single(Box<dyn ActiveKeyExchange>),
}
```

This doesn't work directly:

```rust
error[E0277]: the size for values of type `Self` cannot be known at compilation time
   --> rustls/src/crypto/mod.rs:546:29
    |
546 |         MaybeHybrid::Single(Box::new(self))
    |                             ^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
note: required for `alloc::boxed::Box<Self>` to implement `crypto::ActiveKeyExchange`
   --> rustls/src/crypto/mod.rs:596:28
    |
596 | impl<T: ActiveKeyExchange> ActiveKeyExchange for Box<T> {
    |      -                     ^^^^^^^^^^^^^^^^^     ^^^^^^
    |      |
    |      unsatisfied trait bound introduced here
    = note: required for the cast from `alloc::boxed::Box<alloc::boxed::Box<Self>>` to `alloc::boxed::Box<(dyn crypto::ActiveKeyExchange + 'static)>`
help: consider further restricting `Self`
    |
545 |     fn hybrid_component(self: Box<Self>) -> MaybeHybrid where Self: Sized {
    |                                                         +++++++++++++++++
```

But I'm wondering if there's a way out of this using either an extra trait or maybe an extra layer of boxing?

djc: Right, so what about this?

https://github.com/rustls/rustls/commit/71edf410ac5842971dbe5904c03f247ce83f17d1

ctz: doesn't seem to compile? 

```
error: reached the recursion limit while instantiating `<alloc::boxed::Box<alloc::boxed::Box<alloc::boxed::Box<alloc::boxed::Box<alloc::boxed::Box<...>>>>> as crypto::ActiveKeyExchange>::maybe_hybrid`
   --> rustls/src/crypto/mod.rs:550:29
```

i wonder if we can put the `maybe_hybrid` fn in its own trait, and then have a blanket impl of that trait for any type that impls `ActiveKeyExchange`?

ctz: > i wonder if we can put the `maybe_hybrid` fn in its own trait, and then have a blanket impl of that trait for any type that impls `ActiveKeyExchange`?

this breaks coherence

ctz:
I'm minded to merge this -- @djc do you have strong objections to that?

I think once we take breaking changes, we should do (sketch):

```rust
pub enum ActiveKeyExchange {
  Single(Box<dyn ActiveSingleKeyExchange>),
  Hybrid(Box<dyn ActiveHybridKeyExchange>),
}

pub trait ActiveSingleKeyExchange {
 // something like ActiveKeyExchange prior to this PR
}

pub trait ActiveHybridKeyExchange {
 // something like ActiveKeyExchange after this PR, but with no Option on hybrid_component()'s return type
}
```

djc (approved):
Makes sense to me. File an issue tagged with the appropriate label to follow up for the next breaking release?

PR #2137, opened 2024-09-28
no_std tech demo for removing std::io and some std aliases to core/alloc

stevefan1999-personal:
This PR serves as a `no_std` tech demo for removing the `std::io` requirement. As a result, many `cfg(feature = "std")` attributes, mostly related to `std::io` inclusion, can be removed. This allows the code that assumed buffered I/O to be compiled in a `no_std` environment, as an alternative to unbuffered I/O API if available.

This is just an initial evaluation, and incorporated some code from `no_std_io` before as a replacement for the standard I/O functionality. However, other alternatives such as `embedded-io` exist and could potentially be explored in the future. The choice of `no_std_io` before for this demo was made based on its simplicity and ease of integration, but we're open to considering other options as we further develop this feature. 

Right now, this has been further simplified and the `no_std_io` requirement has been eliminated, as I re-introduced a bare minimum of `no_std_io` code that is needed for rustls. However, it uses `Box<dyn core::error::Error>` to keep custom error, and may also need to change as discussion goes on.

It's important to note that this change doesn't affect the existing functionality for environments where `std` is available. While most of the unit tests should theoretically be compatible without std, but they still needed `std` at the moment for extra platform-specific stuff such as Mutex and Filesystem access.

My own private test under `std` shows that it works, but it needs a converter shim to convert types that implements `std::io::Read` and `std::io::Write` to `rustls::compat::io::Read` and `rustls::compat::io::Write`:
```rust
pub struct Converter<R>(R);

impl<R: std::io::Read> rustls::compat::io::Read for Converter<R> {
    fn read(&mut self, buf: &mut [u8]) -> rustls::compat::io::Result<usize> {
        self.0.read(buf)
    }
}

impl<R: std::io::Write> rustls::compat::io::Write for Converter<R> {
    fn write(&mut self, buf: &[u8]) -> rustls::compat::io::Result<usize> {
        self.0.write(buf)
    }

    fn flush(&mut self) -> rustls::compat::io::Result<()> {
        self.0.flush()
    }
}
```


rustls/src/vecbuf.rs
```diff
@@ -154,6 +151,23 @@ impl ChunkVecBuffer {
...
+    /// however, it consumes all the available chunks
+    #[cfg(not(feature = "std"))]
+    pub(crate) fn write_to(&mut self, wr: &mut dyn io::Write) -> io::Result<usize> {
+        if self.is_empty() {
+            return Ok(0);
+        }
+
+        let mut used = 0;
+        for chunk in self.chunks.iter() {
+            used += wr.write(&chunk)?;
```
paolobarbolini, line 165: This looks suspicious. What if the number of bytes written is less than the number returned?

stevefan1999-personal: Hmm, maybe we should use write_all for that.

stevefan1999-personal: Note that the original write_vectored definition can also have this hazard (https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored) except it only affects the first 64 chunks of taken buffer.

paolobarbolini: What I meant is what happens in this case:

1. Chunk 1: 16 bytes, Written: 16
2. Chunk 2: 16 bytes, Written 8
3. Chunk 3: 16 bytes, Written 8

You lost the last 8 bytes of chunk 2 and you'll write chunk 3 again on the next `write_to` call.

The third write in this example should never happen.

stevefan1999-personal: Yep, I was think of this kind of situation exactly. So I think adding the chunk length after `write_all`, then that should be safe.

djc:
So what is the use case for this exactly? Why not use the unbuffered API, which should not depend on `io`?

stevefan1999-personal:
> So what is the use case for this exactly? Why not use the unbuffered API, which should not depend on `io`?

Well, it is kinda niche I know, but I'm using it with smoltcp under a custom kernel written in Rust. So while I have to run under no_std, I do have ample amount of memory so that buffered APIs are affordable.

djc:
> Well, it is kinda niche I know, but I'm using it with smoltcp under a custom kernel written in Rust. So while I have to run under no_std, I do have ample amount of memory so that buffered APIs are affordable. What is still not affordable is a large kernel size, though, and surprisingly the unbuffered API actually enlarged the kernel binary size compared to buffered API.

I think it would be interesting to dig into that more.

stevefan1999-personal:
> > Well, it is kinda niche I know, but I'm using it with smoltcp under a custom kernel written in Rust. So while I have to run under no_std, I do have ample amount of memory so that buffered APIs are affordable. What is still not affordable is a large kernel size, though, and surprisingly the unbuffered API actually enlarged the kernel binary size compared to buffered API.
> 
> I think it would be interesting to dig into that more.

Alas, this is just a tech demo showing the possibility of how much work we needed to do to enable the no_std with buffered API, but until https://github.com/rust-lang/rfcs/issues/2262 is moving forward, I would still suggest the following few things: Make our own std::io::Error equivalent, make a minimal shim for Read/Write and isolate more real platform dependent stuff

djc:
> > > Well, it is kinda niche I know, but I'm using it with smoltcp under a custom kernel written in Rust. So while I have to run under no_std, I do have ample amount of memory so that buffered APIs are affordable. What is still not affordable is a large kernel size, though, and surprisingly the unbuffered API actually enlarged the kernel binary size compared to buffered API.
> > 
> > 
> > I think it would be interesting to dig into that more.
> 
> Alas, this is just a tech demo showing the possibility of how much work we needed to do to enable the no_std with buffered API, but until [rust-lang/rfcs#2262](https://github.com/rust-lang/rfcs/issues/2262) is moving forward, I would still suggest the following few things: Make our own std::io::Error equivalent, make a minimal shim for Read/Write and isolate more real platform dependent stuff

Why is a PR to the rustls project the right shape for "just a tech demo"?

stevefan1999-personal:
> > > I think it would be interesting to dig into that more.
> > 
> > 
> > Alas, this is just a tech demo showing the possibility of how much work we needed to do to enable the no_std with buffered API, but until [rust-lang/rfcs#2262](https://github.com/rust-lang/rfcs/issues/2262) is moving forward, I would still suggest the following few things: Make our own std::io::Error equivalent, make a minimal shim for Read/Write and isolate more real platform dependent stuff
> 
> Why is a PR to the rustls project the right shape for "just a tech demo"?

Of course, for discussion and figure out better ways to improve it. 

Instead of just talking it out as an issue or in the discussion tab (which wasn't enabled), I think setting up the example first would worth much more.

brody4hire:
I made `portable-io` crate with a similar idea to `compat::io` & tried it with these changes, I can raise a PR to propose this if there is any interest.

I may be able to make a fork with this kind of support if it is not wanted here.

I really like this idea, wonder how much interest there is from the user community.

cpu:
>> Why is a PR to the rustls project the right shape for "just a tech demo"?
>
> Of course, for discussion and figure out better ways to improve it.

I'm not sure this approach is leading to good results. It's been open for some months now and there hasn't been any significant discussion.

I would be more inclined to review the current implementation for providing feedback if the commit history were cleaned up and it was passing CI. However, in general for a task that touches a lot of Rustls I'm a little bit skeptical it's the right idea to dive into a specific solution without first articulating the problem clearly and weighing the available options. IME having maintainer buy-in on the problem & solution will make getting an implementation across the line easier.

stevefan1999-personal:
> > > Why is a PR to the rustls project the right shape for "just a tech demo"?
> > 
> > 
> > Of course, for discussion and figure out better ways to improve it.
> 
> I'm not sure this approach is leading to good results. It's been open for some months now and there hasn't been any significant discussion.
> 
> I would be more inclined to review the current implementation for providing feedback if the commit history were cleaned up and it was passing CI. However, in general for a task that touches a lot of Rustls I'm a little bit skeptical it's the right idea to dive into a specific solution without first articulating the problem clearly and weighing the available options. IME having maintainer buy-in on the problem & solution will make getting an implementation across the line easier.

Hmm, should I squash them together, or just like you also questioned, maybe get a new approach? I thought this would be the least invasive way to do it since it mostly mapped with the collection semantics in std...

For example, `std::vec::Vec` is actually an alias to `alloc::vec::Vec` which is obviously `no_std`, so we reduced an unnecessary `std` requirement into `no_std` equivalent without loss of generality, and I also did (mostly?) the same for `std::collections::HashMap`/`hashbrown`.

What would be your suggestion to have a more balanced way?

brody4hire:
@stevefan1999-personal I noticed that some recent PRs have been proposed & merged with multiple commits. I think it would really help if you could rebase & organize this into a cleaner & more limited number of commits.

I suspect this should be a new feature for no-std, in a similar fashion to how hashbrown is supported for no-std.

I think it would be ideal if we could find a way to reduce the amount of code vendored from std::io, especially from std::io::Error. I can think of a couple of ideas:

- https://github.com/wcampbell0x2a/no-std-io2 seems to be most recent & best supported, but is missing `IoSlice` as I reported in: https://github.com/wcampbell0x2a/no-std-io2/issues/17
- I have made `portable-io` crate, DOES include `IoSlice`, and I have tried this PR together with my `portable-io` crate in my own work area.

@stevefan1999-personal / @rustls team I really hope to see this finished and not abandoned or closed out. I think this could prove itself to be really nice for no-std.

stevefan1999-personal:
> @stevefan1999-personal I noticed that some recent PRs have been proposed & merged with multiple commits. I think it would really help if you could rebase & organize this into a cleaner & more limited number of commits.
> 
> I suspect this should be a new feature for no-std, in a similar fashion to how hashbrown is supported for no-std.
> 
> I think it would be ideal if we could find a way to reduce the amount of code vendored from std::io, especially from std::io::Error. I can think of a couple of ideas:
> 
> - https://github.com/wcampbell0x2a/no-std-io2 seems to be most recent & best supported, but is missing `IoSlice` as I reported in: https://github.com/wcampbell0x2a/no-std-io2/issues/17
> - I have made `portable-io` crate, DOES include `IoSlice`, and I have tried this PR together with my `portable-io` crate in my own work area.
> 
> @stevefan1999-personal / @rustls team I really hope to see this finished and not abandoned or closed out. I think this could prove itself to be really nice for no-std.

I would love to see the continuation of it, but right now my priorities are my company job, and I also got a paper to do. I will try to spend some time on reviving this, but I won't put in much attention and focus into it yet.

After all, this PR serves as a selling convincing rustls team to adopt no-std for buffered API, but it turns out they still aren't interested in it due to the underlying philosophical problem against the use of external crates for standard library, even if it proven to be very useful in practice. 

Anyway, right now it is an impasse with its implied silent treatment from rustls team. I guess we will have to wait for Rust team to settle on the core::io and the error issue around it in order to make a move.

ctz:
My preference on this is that the Rust Project allow downstream projects to build (-> build-std) and replace `std` as a library. That will allow users with esoteric platform requirements to supply polyfill std-equivalent libraries and insert them under rustls.

The idea that this PR suggests (AIUI, hundreds of thousands of crates that rely on `std::io` traits instead introduce their own compatibility layer) doesn't seem like a scalable one.

PR #2140, opened 2024-09-30
Migrate to pki-types PEM decoder

ctz:
This reworks all the test and example code we have to use the new pki-types PEM decoding.

Generally this went smoothly, and I'm very pleased that quite a lot of boilerplate fell away. We might want to `impl std::error::Error` on `pem::Error`?

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -5,11 +5,12 @@
 //
 
 use std::fmt::{Debug, Formatter};
-use std::io::{self, BufReader, Read, Write};
+use std::io::{self, Read, Write};
 use std::sync::Arc;
-use std::{env, fs, net, process, thread, time};
+use std::{env, net, process, thread, time};
 
 use base64::prelude::{Engine, BASE64_STANDARD};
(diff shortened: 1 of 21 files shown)


cpu (approved):

bogo/src/main.rs
```diff
@@ -611,8 +598,11 @@ fn make_server_cfg(opts: &Options) -> Arc<ServerConfig> {
             server::WebPkiClientVerifier::no_client_auth()
         };
 
-    let cert = load_cert(&opts.cert_file);
-    let key = load_key(&opts.key_file);
+    let cert: Vec<_> = CertificateDer::pem_file_iter(&opts.cert_file)
+        .unwrap()
+        .map(|cert| cert.unwrap())
+        .collect();
```
djc, line 604: Nit: might just be me, but I like `collect::Vec<_>()` better than `Vec<_> = fdsifhjsifjsdifjds.collect()`.

djc (approved):
Net -100 lines is a good deal!

PR #2147, opened 2024-10-04
Take aws-lc-rs 1.10

ctz:


ctz:
> ⚠️ There are significant wall-time differences

Sweet. This is the faster RSA impl using AVX512.

Cargo.lock
```diff
@@ -1435,7 +1435,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4979f22fdb869068da03c9f7528f8297c6fd2606bc3a4affe42e6a823fdb8da4"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
```
djc, line 1438: I wonder what caused this...

djc (approved):

Cargo.lock
ctz: Yes, very odd. Seems deterministic for `cargo update -p aws-lc-rs`. If I do an unconstrained `cargo update` it doesn't happen. It seems to be a result of accidentally keeping back aws-lc-fips-sys -- folded into this commit.

PR #2149, opened 2024-10-04
Prepare 0.23.14

ctz:
The main reason for this release is cutting a release that contains the minor perf improvements, for publishing benchmarks against aws-lc-rs 1.10. nb. this release does not require 1.10; there is not currently a need for that.

Draft release notes:

- **Breaking change for `no_std` users**: The return type of [`CryptoProvider::install_default()`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) would change depending on the `std` crate feature. This [was unintended](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility) and is corrected in this release. But that does mean this semver-compatible release contains a breaking API change, albeit only for users who omit the `std` crate feature. Our apologies in advance.
- **Performance improvements** especially for servers doing a full TLS1.3 handshake, and clients doing a resumed TLS1.2 handshake.
- **Improvements to example code**: demonstrate usage of TLS1.3 early data (thanks to @tahmid-23) & using clap for CLI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -448,7 +448,7 @@ version = "0.1.0"
 dependencies = [
  "base64",
  "env_logger",
- "rustls 0.23.13",
+ "rustls 0.23.14",
  "rustls-pki-types",
  "rustls-post-quantum",
 ]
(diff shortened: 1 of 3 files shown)


cpu (approved):
Exciting :-)

PR #2152, opened 2024-10-13
fix: renamed structs to correct naming inconsistency

hex1c:
Corrected the struct names in `crypto/aws_lc_rs/tls13.rs` that were incorrectly referred to as part of the Ring crypto library. This resolves potential confusion and ensures accuracy in the codebase.


djc (approved):
Thanks!

ctz (approved):
Thank you!

PR #2155, opened 2024-10-15
Improve receive performance

### LINKED ISSUE #2154: Poor asymptotic performance when reading plaintext incrementally
If we receive a large appdata message (16KB) and the application reads it one byte at a time, we spend a silly amount of time in memmove. That is because:

- `ChunkVecBuffer::read` for one byte has complexity of O(N)
- the application must call it once for each byte

So we end up with accidentally quadratic overall performance.

This also affects our benchmarking: currently each 16KB message is received in two 8KB reads, with 8KB of wasted memmove.

ctz:
When reading from a `ChunkVecBuffer` we `Vec::drain()` the read bytes. That involves a memmove of the unread bytes, and that is both expensive and avoidable.

A couple of unrelated nits coming along for the ride.

fixes #2154

### DIFF
--- a/rustls/src/vecbuf.rs
+++ b/rustls/src/vecbuf.rs
@@ -9,10 +9,10 @@ use std::io::Read;
 #[cfg(feature = "std")]
 use crate::msgs::message::OutboundChunks;
 
-/// This is a byte buffer that is built from a vector
-/// of byte vectors.  This avoids extra copies when
-/// appending a new byte vector, at the expense of
-/// more complexity when reading out.
+/// This is a byte buffer that is built from a deque of byte vectors.
+///
+/// This avoids extra copies when appending a new byte vector,
+/// at the expense of more complexity when reading out.
 pub(crate) struct ChunkVecBuffer {
     chunks: VecDeque<Vec<u8>>,
     limit: Option<usize>,
@@ -74,8 +74,9 @@ impl ChunkVecBuffer {
         len
     }
 
-    /// Take one of the chunks from this object.  This
-    /// function panics if the object `is_empty`.
+    /// Take one of the chunks from this object.
+    ///
+    /// This function returns `None` if the object `is_empty`.
     pub(crate) fn pop(&mut self) -> Option<Vec<u8>> {
         self.chunks.pop_front()
     }
--- a/rustls/src/vecbuf.rs
+++ b/rustls/src/vecbuf.rs
@@ -1,6 +1,6 @@
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
-use core::cmp;
+use core::{cmp, mem};
 #[cfg(feature = "std")]
 use std::io;
 #[cfg(feature = "std")]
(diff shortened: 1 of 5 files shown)


rustls/src/vecbuf.rs
```diff
@@ -9,10 +9,11 @@ use std::io::Read;
 #[cfg(feature = "std")]
 use crate::msgs::message::OutboundChunks;
 
-/// This is a byte buffer that is built from a vector
-/// of byte vectors.  This avoids extra copies when
-/// appending a new byte vector, at the expense of
-/// more complexity when reading out.
+/// This is a byte buffer that is built from a deque
+/// of byte vectors.
```
djc, line 13: Nit: unfold this? I think we're allowed to go until column 100 which I'm pretty sure this isn't hitting.

rustls/src/vecbuf.rs
```diff
@@ -45,11 +45,10 @@ impl ChunkVecBuffer {
...
     pub(crate) fn len(&self) -> usize {
-        let mut len = 0;
-        for ch in &self.chunks {
-            len += ch.len();
-        }
-        len
+        self.chunks
+            .iter()
+            .map(|ch| ch.len())
+            .sum()
```
djc: Does this have a performance benefit? Seems surprising.

rustls/src/vecbuf.rs
```diff
@@ -128,14 +145,19 @@ impl ChunkVecBuffer {
...
-            if used < buf.len() {
-                buf.drain(..used);
+    fn consume(&mut self, used: usize) {
+        // first, mark the rightmost extent of the used buffer
+        self.prefix_used += used;
+
+        // then reduce `prefix_used` by discarding wholly-covered
+        // buffers
+
+        while let Some(buf) = self.chunks.pop_front() {
```
djc, line 155: We currently do:

- `pop_front()`, check, then `push_front()` if necessary

Do you think it might be faster to do instead inspect front in place and only pop it if necessary?

rustls/src/vecbuf.rs
ctz: Reworded commit message.

djc: It doesn't look much more idiomatic to me, but it's one line less so I'll take it. 👍

rustls/src/vecbuf.rs
```diff
@@ -14,13 +14,23 @@ use crate::msgs::message::OutboundChunks;
 /// This avoids extra copies when appending a new byte vector,
 /// at the expense of more complexity when reading out.
 pub(crate) struct ChunkVecBuffer {
+    /// How many bytes have been consumed in the first chunk.
+    ///
+    /// Invariant: zero if `chunks.is_empty()`
+    /// Invariant: 0 <= `prefix_used` < `chunks[0].len()`
+    prefix_used: usize,
+
+    /// Subsequent chunks.
```
djc, line 23: Nit: the `Subsequent` here feels a little misleading to me, given that this actually contains all the chunks.

djc (approved):

rustls/src/vecbuf.rs
ctz: Ah, good spot! Leftover from when I had the first item separate from `chunks`

rustls/src/vecbuf.rs
```diff
@@ -44,11 +44,10 @@ impl ChunkVecBuffer {
...
     /// How many bytes we're storing
     pub(crate) fn len(&self) -> usize {
-        let mut len = 0;
-        for ch in &self.chunks {
-            len += ch.len();
-        }
-        len
+        self.chunks
+            .iter()
+            .map(|ch| ch.len())
```
cpu, line 59: I think this could be `map(Vec::len)`.

Alternatively, maybe the whole `map/sum` should be a fold?

```rust
 self.chunks
            .iter()
            .fold(0usize, |acc, chunk| acc + chunk.len())
            - self.prefix_used
```

I don't feel strongly in either case.

cpu (approved):

PR #2156, opened 2024-10-15
Prepare 0.23.15

ctz:
Draft release notes:

- **Improve performance** when reading plaintext data in small increments (#2154)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2192,9 +2192,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-pki-types"
-version = "1.9.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e696e35370c65c9c541198af4543ccd580cf17fc25d8e05c5a242b202488c55"
+checksum = "16f1201b3c9a7ee8039bcadc17b7e605e2945b27eee7631788c1bd2b0643674b"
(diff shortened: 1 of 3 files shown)


cpu (approved):

djc (approved):
Maybe it makes sense to also require pki-types 1.10? That would enable downstreams to rely on the `Error` impl for `pem::Error` without having to add a separate version constraint.

cpu (approved):

bogo/Cargo.toml
```diff
@@ -6,6 +6,6 @@ edition = "2021"
 [dependencies]
 base64 = "0.22"
 env_logger = "0.10" # 0.11 requires 1.71 MSRV even as a dev-dep (due to manifest features)
-pki-types = { package = "rustls-pki-types", version = "1.9", features = ["std"] }
+pki-types = { package = "rustls-pki-types", version = "1.10", features = ["std"] }
```
djc, line 9: Since bogo and examples depend on rustls already (and aren't published) I don't think making this change for them is useful (though it can't hurt, either).

djc: (Arguably they should use the reexport?)

djc: https://github.com/rustls/rustls/pull/2158.

djc (approved):

PR #2163, opened 2024-10-17
bench: avoid debug_assert

ctz:
This tool is invariably run in release mode; so using debug_assert to check we achieve the desired kind of handshake is less than ideal.


cpu (approved):
Good idea, thanks 👍

PR #2165, opened 2024-10-17
Remove crossbuilds and coverage from merge_checks

ctz:
These are the more costly jobs, and it seems unlikely they uniquely will capture rebase failures in a way that other jobs cannot (?).

This should improve our merge queue time, which is currently around 16-24 minutes.


cpu (approved):
Thanks :+1:

djc (approved):

PR #2166, opened 2024-10-17
client/tls12: remove stale ref to SCT verification

cpu:
The TLS 1.3 equivalent of this list of steps was already updated to reflect the removal of SCT verification. This commit does the same for the TLS 1.2 client code, renumbering steps as required.


ctz (approved):

cpu:
Docs-only. Going to merge w/ 1 review.

PR #2170, opened 2024-10-21
Change link to the NSS Key Log Format

yedayak:
The current link is 404.
I'm not sure that this is the official website, but this contains the correct information in any case.


ctz (approved):
Thanks! 

In the future we can link to https://datatracker.ietf.org/doc/draft-ietf-tls-keylogfile/ which will shortly be published as an RFC.

cpu (approved):
Thanks

Issue #2171, opened 2024-10-21
Reduce number of obsolete variants in CipherSuite enum


ctz:
There's not much reason to have an enum variant for `TLS_KRB5_EXPORT_WITH_DES_CBC_40_SHA` or `SSL_FORTEZZA_KEA_WITH_FORTEZZA_CBC_SHA`, so let's delete them.

PR #2172, opened 2024-10-21
Make ciphersuite enum smaller

### LINKED ISSUE #2138: Don't use Debug representation that much
Currently this innocent looking line may not be that much:

https://github.com/rustls/rustls/blob/211eabfa61d8af7643300e1e81dc33ce598e15f8/rustls/src/builder.rs#L245

But this needs the entire enum to implement Debug, and its representation can be huge, especially with CipherSuite:
https://github.com/rustls/rustls/blob/211eabfa61d8af7643300e1e81dc33ce598e15f8/rustls/src/enums.rs#L116-L491

I added a simple hack to suite_name:
```rust
            let suite_name = cs.common().suite;
            let suite_name: u16 = suite_name.into();
            return Err(Error::General(alloc::format!(
                "Ciphersuite {suite_name:?} requires {cs_kx:?} key exchange, but no {cs_kx:?}-compatible \
                key exchange groups were present in `CryptoProvider`'s `kx_groups` field",
            )));
```
This immediately shaved 7KB off the binary. 

Unfortunately, not all code can be eliminated like this, such as unwrap() calls which needs `Debug` as well, and this kind of hack might also lost its original (friendly-named) context. We should let the user to control this behavior, for example, we can just print the hex value of the suite name, so that the user can look it up themselves rather than adding this tax to the embedded device.

ctz:
We can't remove variants without a breaking API change (saved #2171 for later) but we can reduce the expense and size of `Debug` on it now.

This doesn't do anything for `fn to_str` but that function doesn't have the viral property that `Debug` has: if you don't want the strings in that function, just don't call it.

fixes #2138


djc (approved):

rustls/src/enums.rs
```diff
@@ -114,381 +114,381 @@ enum_builder! {
...
-        TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 => 0x009f,
-        TLS_DH_RSA_WITH_AES_128_GCM_SHA256 => 0x00a0,
-        TLS_DH_RSA_WITH_AES_256_GCM_SHA384 => 0x00a1,
-        TLS_DHE_DSS_WITH_AES_128_GCM_SHA256 => 0x00a2,
-        TLS_DHE_DSS_WITH_AES_256_GCM_SHA384 => 0x00a3,
-        TLS_DH_DSS_WITH_AES_128_GCM_SHA256 => 0x00a4,
-        TLS_DH_DSS_WITH_AES_256_GCM_SHA384 => 0x00a5,
-        TLS_DH_anon_WITH_AES_128_GCM_SHA256 => 0x00a6,
-        TLS_DH_anon_WITH_AES_256_GCM_SHA384 => 0x00a7,
+        TLS_RSA_WITH_NULL_MD5 => 0x0001 NoDebug,
```
djc, line 117: Don't love the random-looking identifier in here. Did you consider an attribute (with the downside of adding an extra line for each `NoDebug` variant) or maybe some comment-like syntax?

cpu (approved):
LGTM, but I concur with @djc that the `nodebug` syntax in the macro feels a little bit unusual. Not a deal breaker ofc.

rustls/src/enums.rs
ctz: I have tried an alternative syntax which calls back to negative trait impls. Still looks quite weird though, but at least it's now clear which items are included and excluded in `Debug::fmt`

cpu (approved):
New syntax looks like an improvement to me, thanks :+1:

PR #2173, opened 2024-10-21
automate dead link checking of documentation

cpu:
This branch adds [lychee](https://github.com/lycheeverse/lychee) via [lychee-action](https://github.com/lycheeverse/lychee-actio) to do fast link checking of the markdown files & generated rustdoc documentation. I've been using this successfully with some side projects and find it to be a high-quality tool. 

The link checking happens during the scheduled nightly docs build/deploy workflow in `docs.yml` so as to not block normal development if there are new broken links or flaky bad internet weather. There's a nice pattern to have Lychee automatically open issues for broken links, but it uses `peter-evans/create-issue-from-file` and seems to require write permission on the repo. I've skipped that for now and we'll just have to try and keep an eye on this workflow failing.

A few ignore regexps are added up front:

* One ignoring a file that's referenced by some generated rustdoc boilerplate but that doesn't seem to exist on-disk.

* Two ignoring specific URLs. These were failing for me locally with lychee but work in browser. I suspect due to WAF-like stuff in front of the servers swatting away automated traffic.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,6 +6,7 @@ permissions:
   id-token: write
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


ctz (approved):
Nice!

PR #2174, opened 2024-10-21
consistent APIs for connection FIPS indicator

cpu:
Previously `ClientConnection::fips()` could be used to determine if a `ClientConnection` was created from a FIPS-enabled `ClientConfig`, but there was no similar support for a `ServerConnection` - just `ServerConfig::fips()`.

This commit adds an equivalent `ServerConnection::fips()` fn and tidies up the internal code for where the fips status is shared to be in `CommonState` instead of `ClientConnectionData`.


ctz (approved):

djc (approved):
Nice!

PR #2175, opened 2024-10-22
post-quantum: move from kyber768 to ML-KEM768

ctz:
The aws-lc-rs API for this becomes not available if the fips feature is set (since the fipsmodule does not contain ML-KEM), so set the non-fips feature to fail the build if feature unification happens.

Originally I planned to move this into the main crate, unfortunately that is difficult for two reasons:

- the aforementioned difficulty with FIPS,
- `ML_KEM_768` remains gated behind unstable; I think setting that on the core crate would mean affixing the aws-lc-rs version, and that would mean being on a release treadmill for aws-lc-rs versions.


djc:
Is there a tracking issue on the aws-lc-rs side on why this is unstable?

.github/workflows/daily-tests.yml
```diff
@@ -129,7 +129,7 @@ jobs:
         run: cargo run --locked -p rustls-provider-example --example client
 
       - name: Check rustls-post-quantum client
-        run: cargo run --locked -p rustls-post-quantum --example client | grep 'kex=X25519Kyber768Draft00'
+        run: cargo run --locked --manifest-path=rustls-post-quantum/Cargo.toml --example client | grep 'kex=X25519MLKEM768'
```
djc, line 132: Neither this change nor the `Cargo.lock` addition seem like they belong in a separate commit?

ctz: split and squashed into the previous commits.

Cargo.toml
```diff
@@ -14,18 +14,17 @@ members = [
   "provider-example",
   # the main library and tests
   "rustls",
-  # experimental post-quantum algorithm support
-  "rustls-post-quantum",
   # rustls cryptography provider integration tests
   "rustls-provider-test",
 ]
+exclude = [
```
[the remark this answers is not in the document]
ctz: It would otherwise be included because of `bogo`'s path dependency on it. See https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

admin/all-workspace-members
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
ctz: Added some extra docs (linking the above cargo documentation) and used just `__doc__` -- not sure why I didn't do that before!

ctz:
Where this ended up is pretty messy and unsatisfying, but I expect it to be temporary and hopefully fixed in the next aws-lc-rs release. Then the second commit can be reverted.

.github/workflows/build.yml
```diff
@@ -452,7 +458,16 @@ jobs:
...
+      - run: |
+          set -x
+          for p in $(admin/all-workspace-members) ; do
+            if [ "$p" == "bogo" ] ; then
+              ALL_FEATURES="--features fips"
+            else
+              ALL_FEATURES="--all-features"
+            fi
+            cargo clippy --locked --package $p $ALL_FEATURES --all-targets -- $CLIPPY_PARAMS
+          done
```
cpu, line 470: At the point where we're writing Python anyway WDYT about expanding the scope of `all-workspace-members` to take these details as input, doing more of the looping/executing a cmd itself?

The shell scripting here isn't too egregious but it's also a pretty gross thing to have to use (does it need `set -e`?) and being embedded in YAML is like two terrible tastes that taste terrible together :laughing:

cpu (approved):

.github/workflows/build.yml
cpu: > I expect it to be temporary and hopefully fixed in the next aws-lc-rs release. Then the second commit can be reverted.

Sorry, wasn't thinking about this comment. Given this knowledge I think it's fine as-is.

ctz: It's a reasonable point -- I have hoisted all the clippy stuff into one script; this also reduces some of the repetition in the yaml.

cpu (approved):

PR #2181, opened 2024-10-25
Expose ConfigBuilder::crypto_provider()

djc:
This enables downstream projects (such as rustls-platform-verifier -- see https://github.com/rustls/rustls-platform-verifier/pull/150) to get the `CryptoProvider` from the `ConfigBuilder`, instead of having to pick one in parallel.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -120,9 +120,9 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
         sigschemes: &[SignatureScheme],
     ) -> Option<Arc<sign::CertifiedKey>>;
 
-    /// Return true if the client only supports raw public keys.  
-    ///  
-    /// See [RFC 7250](https://www.rfc-editor.org/rfc/rfc7250).  
+    /// Return true if the client only supports raw public keys.
+    ///
+    /// See [RFC 7250](https://www.rfc-editor.org/rfc/rfc7250).
     fn only_raw_public_keys(&self) -> bool {
         false
     }
@@ -316,10 +316,9 @@ impl ClientConfig {
         provider: Arc<CryptoProvider>,
     ) -> ConfigBuilder<Self, WantsVersions> {
         ConfigBuilder {
-            state: WantsVersions {
-                provider,
-                time_provider: Arc::new(DefaultTimeProvider),
-            },
+            state: WantsVersions {},
+            provider,
+            time_provider: Arc::new(DefaultTimeProvider),
             side: PhantomData,
         }
     }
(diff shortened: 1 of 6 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -120,9 +120,9 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
         sigschemes: &[SignatureScheme],
     ) -> Option<Arc<sign::CertifiedKey>>;
 
-    /// Return true if the client only supports raw public keys.  
-    ///  
-    /// See [RFC 7250](https://www.rfc-editor.org/rfc/rfc7250).  
+    /// Return true if the client only supports raw public keys.
```
ctz, line 123: Hm. Could had sworn I've seen rustfmt fix trailing whitespace in comments before.

ctz: nb. not objecting to fixing the whitespace here in passing, just confused how it got here in the first place.

ctz (approved):

rustls/src/client/client_conn.rs
djc: (I didn't even do anything to fix it, just set up VS Code to ditch trailing whitespace.)

djc:
Does this need some test coverage? Where/how?

djc:
I think this is semver-compatibility failure is probably not a big deal:

```
--- failure auto_trait_impl_removed: auto trait no longer implemented ---

Description:
A public type has stopped implementing one or more auto traits. This can break downstream code that depends on the traits being implemented.
        ref: https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.36.0/src/lints/auto_trait_impl_removed.ron

Failed in:
  type ConfigBuilder is no longer UnwindSafe, in /home/runner/work/rustls/rustls/rustls/src/builder.rs:164
  type ConfigBuilder is no longer RefUnwindSafe, in /home/runner/work/rustls/rustls/rustls/src/builder.rs:164
```

cpu (approved):
I trust your evaluation of the semver reporting :+1:

cpu:
> Does this need some test coverage? Where/how?

I think it wouldn't hurt & seems like an easy lift. Maybe update [these two tests](https://github.com/rustls/rustls/blob/626424311fbf8329eeccc4fa867974d7f6897c39/rustls/tests/process_provider.rs#L47-L71) to save the `ClientConfig` from `finish_client_config()` and assert the config provider matches default?

djc:
Added tests sort of like you suggested.

PR #2168, opened 2024-10-17
Faster TLS1.3 server resumption

ctz:
First commit fixes a small bit of wastefulness. Its effect is somewhat tempered by #2167 , but worth having.
Second commit expands the scope of #2120.

Indicative performance results:

- 0.23.15: `8019.6 handshakes/sec`
- after this: `8499.08 handshakes/sec`
- after this and #2167: `9666.8 handshakes/sec`

PR #2183, opened 2024-10-27
Bump version to 0.23.16

djc:
I don't think there's anything large enough that it's worth calling out here, but I'd like to get #2181 out for the benefit of https://github.com/rustls/rustls-platform-verifier/pull/150. So, proposed release notes highlights:

* Added support for RFC 7250 raw public keys.

(From #2168, #2173, #2172.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,9 +78,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.17"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23a1e53f0f5d86382dafe1cf314783b2044280f406e7e1506368220ad11b1338"
+checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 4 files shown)


cpu:
> I'd like to get https://github.com/rustls/rustls/pull/2181 out for the benefit of https://github.com/rustls/rustls-platform-verifier/pull/150.

SGTM. I wouldn't mind https://github.com/rustls/rustls/pull/2174 in a release for rustls-ffi.

> I don't think there's anything large enough that it's worth calling out here,

I think there's a few things worthy of mention in [the diff since last release](https://github.com/rustls/rustls/compare/v/0.23.15...main) :thinking: . Raw public key support (https://github.com/rustls/rustls/pull/2062) and the kyber -> ml-kem (https://github.com/rustls/rustls/pull/2175) change both seem worth some kind of mention.

djc:
> I think there's a few things worthy of mention in the diff since last release 🤔 . Raw public key support (#2062) and the kyber -> ml-kem (#2175) for example.

Ahh. ML-KEM is only in rustls-post-quantum IIUC but raw key support is definitely worth calling out. Stymied again by GitHub sorting PRs by creation date by default...

cpu:
> ML-KEM is only in rustls-post-quantum IIUC

Good point!

djc:
Added a line about the raw public keys.

cpu:
I think the fuzz crate lock file needs an update to fix ci

cpu (approved):

djc:
* Published rustls v0.23.16 at registry `crates-io`
* [new tag]           v/0.23.16 -> v/0.23.16
* [Release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.23.16)

stormshield-gt:
In the release note, it's mentioned that we moved from kyber768 to ML-KEM768 for `rustls-post-quantum` but this change is not published yet. Maybe a 0.2 version of `rustls-post-quantum` could be published? Or we are waiting on https://github.com/rustls/rustls/pull/2136 ?
I'm currently migrating to post-quantum, It would be better for me to directly migrate to ML-KEM rather than a 2-step process. Note that I'm fine with waiting, though.

ctz:
Yes I would like to land #2136 before releasing the next rustls-post-quantum.

Issue #1751, opened 2024-01-17
TLS1.3 performance problem


zh-jq-b:
See https://bytedance.larkoffice.com/docx/NEqXdDyq8op8mwx0XTWcwgionnf.

The tls1.3 performance isn't as good as OpenSSL 1.1.x when testing at high concurrency.

djc:
Why are you not testing aws-lc-rs support? Are your results in line with https://github.com/aochagavia/rustls-bench-results?

zh-jq-b:
I also add tls1.2 test result here https://bytedance.larkoffice.com/docx/NEqXdDyq8op8mwx0XTWcwgionnf.
The rustls 1.2 performance is better than tls1.3 & openssl (which is expected).
There are no much RPS difference between rustls 0.21 & rustls 0.22 ring & rustls 0.22 aws-lc in our test.

djc:
There have been some recent changes on the main branch. Would it be hard to try this with current rustls main?

zh-jq-b:
> There have been some recent changes on the main branch. Would it be hard to try this with current rustls main?

The max RPS (ring) is still around 54000.

I just bump the version here https://github.com/bytedance/g3/commit/0c2feafecb28e7d7c7c9b4186f52752dbb9c787c without any API changes.

cpu:
Are you building with gcc or clang?

zh-jq-b:
> Are you building with gcc or clang?

I didn't set a custom CC, so it should be gcc.

briansmith:
> I didn't set a custom CC, so it should be gcc.

recent clang should be faster.

zh-jq:
The boringssl performance is much better so I don't think it's the problem I reported here.
I will have an updated test with clang next week.

zh-jq-b:
Hi, I do a test with CC set to clang-18 today, the problem is still the same:
![image](https://github.com/rustls/rustls/assets/127708320/bc692e17-f84d-45c7-81e8-1bdf05877513)

zh-jq-b:
I have done a full test yesterday with an updated OS/clang, with the result updated in https://bytedance.larkoffice.com/docx/NEqXdDyq8op8mwx0XTWcwgionnf.
The TLS 1.2 performance of rustls is awesome, while the TLS 1.3 performance is still fall behind.

ctz:
[...]

PR #2187, opened 2024-10-30
server: default send_tls13_tickets 4 -> 2

cpu:
This change aligns the Rustls default for a `ServerConfig::builder()`'s [`send_tls13_tickets`](https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.send_tls13_tickets) value to match BoringSSL/OpenSSL's default of 2.

Previously we used 4 but this manifests as a performance gap in server-side resumption cost when comparing default configurations between TLS libraries due to the high cost of creating new tickets.

We believe a default of 2 strikes a better balance in general, and avoids folks repeatedly tripping into this during benchmarking (e.g. https://github.com/rustls/rustls/issues/1751#issuecomment-2445837285, https://github.com/rustls/rustls/pull/2167). Anyone that has a workload that prefers up-front server-side cost in favour of reduced client-side latency for batched connections can tweak the value to 4 or higher manually.

### DIFF
--- a/rustls/examples/internal/bench_impl.rs
+++ b/rustls/examples/internal/bench_impl.rs
@@ -611,7 +611,6 @@ fn make_server_config(
         cfg.session_storage = Arc::new(NoServerSessionStorage {});
     }
 
-    cfg.send_tls13_tickets = 2; // matches BoringSSL/OpenSSL default
     cfg.max_fragment_size = max_fragment_size;
     Arc::new(cfg)
 }
--- a/rustls/src/server/builder.rs
+++ b/rustls/src/server/builder.rs
@@ -137,7 +137,7 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
             enable_secret_extraction: false,
             max_early_data_size: 0,
             send_half_rtt_data: false,
-            send_tls13_tickets: 4,
+            send_tls13_tickets: 2,
             #[cfg(feature = "tls12")]
             require_ems: cfg!(feature = "fips"),
             time_provider: self.time_provider,
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -244,7 +244,7 @@ impl<'a> ClientHello<'a> {
 ///   implementation.
 /// * [`ServerConfig::alpn_protocols`]: the default is empty -- no ALPN protocol is negotiated.
 /// * [`ServerConfig::key_log`]: key material is not logged.
-/// * [`ServerConfig::send_tls13_tickets`]: 4 tickets are sent.
+/// * [`ServerConfig::send_tls13_tickets`]: 2 tickets are sent.
 /// * [`ServerConfig::cert_compressors`]: depends on the crate features, see [`compress::default_cert_compressors()`].
 /// * [`ServerConfig::cert_compression_cache`]: caches the most recently used 4 compressions
 /// * [`ServerConfig::cert_decompressors`]: depends on the crate features, see [`compress::default_cert_decompressors()`].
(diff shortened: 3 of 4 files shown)


djc:
> This change aligns the Rustls default for a `ServerConfig::builder()`'s [`send_tls13_tickets`](https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.send_tls13_tickets) value to match BoringSSL/OpenSSL's default of 2.
> 
> Previously we used 4 but this manifests as a performance gap in server-side resumption cost when comparing default configurations between TLS libraries due to the high cost of creating new tickets.
> 
> We believe a default of 2 strikes a better balance in general, and avoids folks repeatedly tripping into this during benchmarking (e.g. [#1751 (comment)](https://github.com/rustls/rustls/issues/1751#issuecomment-2445837285), #2167). Anyone that has a workload that prefers up-front server-side cost in favour of reduced client-side latency for batched connections can tweak the value to 4 or higher manually.

I don't disagree but I want to challenge this. I understand that, all other things being equal, it would be better to match others so that people benchmarking will be less likely to confuse themselves, but I also believe it is more important that we deliver good software to downstream (server) users than that we don't confuse benchmark readers.

At the same time, doesn't this make rustls worse at resuming sessions in practice? Do we send new tickets in resumed sessions? Why did we pick 4 in the first place? Preferring upfront server-side cost (how much cost are we talking about in practice?) in favor of reduced client-side latency seems like a pretty good trade-off for lots of people...

ctz:
> As such, doesn't this make rustls worse at resuming sessions in practice?

Yes, but I think in quite limited circumstances; see below.

> Do we send new tickets in resumed sessions?

Yes, `send_tls13_tickets` number of tickets are sent irrespective of the kind of handshake[^1].

> Why did we pick 4 in the first place?

For reference, that commit was d78079032 though I didn't record why I specifically chose 4 the default changed from 1 to 4, it was in the context of making TLS1.3 tickets non-reusable.

I _think_ I chose the number 4 on the basis of a typical limit of concurrent connections made by a HTTP1 client to a given host (the actual number varies wildly by browser, with RFC2616 specifying 2 but (eg) Chrome doing 6. However that measure is not relevant for HTTP2, nor HTTP3.

> Preferring upfront server-side cost (how much cost are we talking about in practice?) 

Seems the benchmarks here are saying the extra 2 tickets cost us 8%/6% (client/server) of a normal resumption, or ~0%/2% of a full handshake.

> in favor of reduced client-side latency seems like a pretty good trade-off for lots of people... I don't feel like the trade-off is all that crisply articulated in this PR's description.

The case where 2 is worse than 4 is pretty specific, I think. For the sake of argument, say the number is N:

1. The client starts with no tickets. It must do a full handshake. It now has `N` tickets.
2. the client may now start up to `N` resumed handshakes in parallel. Once those complete, it has up to `N*N` tickets (`N` from each of `N` handshakes).
3. the client may now start up to `N*N` resumed handshakes in parallel. Once those complete, it has `N*N*N` [^2] tickets.

To reiterate, this only limits the number of resumed handshakes a client can make _in parallel_ at any given point. In serial is fine: each time a handshake completes, it replenishes the client's cache by `N` (having consumed one ticket).

[^1]: Compare OpenSSL (but not BoringSSL) that sends only one ticket after a resumed handshake, but `SSL_CTX_set_num_tickets` (defaults to 2) for a full handshake.
[^2]: Somewhat unrelated, but rustls on the client limits this growth to 8 per host, to avoid this sort of growth pushing items for other hosts out of the session cache.

djc (approved):
Okay, I agree that this makes sense then!

PR #2192, opened 2024-11-01
Support multithreaded benchmarks

ctz:
This produces output like:

```
$ cargo run --release --example bench -- --threads 8 handshake TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
handshakes	TLSv1_2	Rsa2048	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256	client	server-auth	no-resume	threads	8	10155.44	10166.96	10092.23	10234.48	9095.63	10098.87	10075.14	9104.87	total	79023.63	per-thread	9877.95	handshakes/s
```

This means:

```
handshakes	TLSv1_2	Rsa2048	TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256	client	server-auth	no-resume
```

old output

```
threads	8
```

number of threads used

```
10155.44	10166.96	10092.23	10234.48	9095.63	10098.87	10075.14	9104.87
```

individual thread measurements

```
total	79023.63	per-thread	9877.95	handshakes/s
```

sum and average of thread measurements

### DIFF
--- a/admin/bench-measure.mk
+++ b/admin/bench-measure.mk
@@ -53,6 +53,8 @@ threads: $(BENCH)
 	  $^ --threads $$thr handshake-ticket TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 ; \
 	  $^ --threads $$thr handshake TLS13_AES_256_GCM_SHA384 ; \
 	  $^ --threads $$thr handshake-ticket TLS13_AES_256_GCM_SHA384 ; \
+	  $^ --threads $$thr bulk TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 ; \
+	  $^ --threads $$thr bulk TLS13_AES_256_GCM_SHA384 ; \
 	done
 
 clean:
--- a/admin/bench-measure.mk
+++ b/admin/bench-measure.mk
@@ -46,6 +46,15 @@ memory: $(BENCH)
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 1000
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 5000
 
+threads: $(BENCH)
+	for thr in $(shell admin/threads-seq.rs) ; do \
+	  $^ --threads $$thr handshake TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 ; \
+	  $^ --threads $$thr handshake-resume TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 ; \
+	  $^ --threads $$thr handshake-ticket TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 ; \
+	  $^ --threads $$thr handshake TLS13_AES_256_GCM_SHA384 ; \
+	  $^ --threads $$thr handshake-ticket TLS13_AES_256_GCM_SHA384 ; \
+	done
+
 clean:
 	rm -f perf-*.svg
 	cargo clean
--- a/admin/threads-seq.rs
+++ b/admin/threads-seq.rs
@@ -1,0 +1,51 @@
+#!/usr/bin/env -S cargo +nightly --quiet -Zscript
+
+//! # `admin/thread-seq N`
+//!
+//! This program prints a sequence of N integers for multithreaded
+//! performance testing.  The integers are numbers of threads to
+//! be sampled in a test.  The goal is to assist in graphing
+//! how per-thread throughput relates to concurrency.
+//!
+//! The sequence is (at most) length N, starts at 2, includes the
+//! number of CPU cores, and ends at 1.5 the number of CPU cores.
+//! It does not have repeated items.
+//!
+//! We exceed the number of cores specifically to see the
+//! "elbow" in the graph, when the number of threads tested
+//! exceeds the number of cores.  (This is good because otherwise
+//! -- assuming the software under test is perfectly scalable --
+//! the graph would be a straight line parallel with the x axis.)
+
+use std::{cmp, env, error, num::NonZeroUsize, str::FromStr, thread};
+
+fn main() -> Result<(), Box<dyn error::Error + Send + Sync + 'static>> {
+    let mut args = env::args();
+    args.next(); // skip argv[0]
+    let count = args
+        .next()
+        .map(|c| NonZeroUsize::from_str(&c))
+        .transpose()?
+        .unwrap_or(NonZeroUsize::new(16).unwrap())
+        .get();
+
+    let default_cpus = thread::available_parallelism()?;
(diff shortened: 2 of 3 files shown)


cpu (approved):
Nice :-)

admin/bench-measure.mk
```diff
@@ -46,6 +46,15 @@ memory: $(BENCH)
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 1000
 	$(MEMUSAGE) $^ memory TLS13_AES_256_GCM_SHA384 5000
 
+threads: $(BENCH)
+	for thr in 2 4 8 12 16 24 32 40 48 56 64 72 80 92 112 128; do \
```
djc, line 50: That seems like shitload of thread counts, is that worth it in terms of runtime? It doesn't seem like running at, say, non-powers of 2 will surface new information over above what we will get running only at 2 to the power 1..7.

rustls/examples/internal/bench_impl.rs
```diff
@@ -491,14 +498,24 @@ fn bench_bulk(
     let rounds = total_data / plaintext_size;
 
     if options.api.use_buffered() {
+        let threads = (0..options.threads.into())
+            .map(|_| {
+                let client_config = client_config.clone();
+                let server_config = server_config.clone();
+                thread::spawn(move || {
+                    bench_bulk_buffered(client_config, server_config, plaintext_size, rounds)
```
djc, line 506: Worth extracting this as a function parametrized with an `impl Fn(..)`?

(Might even consider making it a trait...)

rustls/examples/internal/bench_impl.rs
```diff
@@ -374,11 +403,39 @@ fn report_handshake_result(
...
         resume.label(),
-        (rounds as f64) / timings.server
     );
+
+    report_timings(&timings, rounds, |t| t.server);
+}
+
+fn report_timings(thread_timings: &[Timings], rounds: u64, which: impl Fn(&Timings) -> f64) {
+    let thread_count = thread_timings.len();
+    match thread_count {
```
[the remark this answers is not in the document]
ctz: pulled out the `1` arm into its own `if let` with an early return

admin/bench-measure.mk
ctz: I think maybe this list is overfitting on 80 cores (certainly it is not very useful for me when run locally on a 16 core laptop) -- replaced with a python script which explains the goal

admin/threads-seq
```diff
@@ -0,0 +1,38 @@
...
+# number of CPU cores, and ends at 1.5 the number of CPU cores.
+# It does not have repeated items.
+#
+# We exceed the number of cores specifically to see the
+# "elbow" in the graph, when the number of threads tested
+# exceeds the number of cores.  (This is good because otherwise
+# -- assuming the software under test is perfectly scalable --
+# the graph would be a straight line parallel with the x axis.)
+
+import os, sys, math, itertools
```
djc, line 20: I'd suggest that this could just as well be a little Rust script, which would be a little nicer IMO. Not a blocker by any means.

djc (approved):

cpu (approved):

admin/threads-seq
cpu: agreed, the more Python there is the more it's tempting to think about formatting, linting, Python versions, etc.

ctz: RiiR'd 🦀

admin/threads-seq
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env -S cargo +nightly --quiet -Zscript
```
djc, line 1: I didn't quite mean using the `-Zscript` support, but I guess it'll do? Would be nice to name it `.rs` for now so that we get syntax highlighting?

admin/threads-seq
```diff
@@ -0,0 +1,56 @@
...
+
+use std::{cmp, env, thread};
+
+fn main() {
+    let mut args = env::args();
+    args.next(); // skip argv[0]
+    let count = args
+        .next()
+        .map(|c| {
+            c.parse::<usize>()
```
djc, line 28: Nit: I prefer `usize::from_str(c)` over `c.parse::<usize>()`.

admin/threads-seq
```diff
@@ -0,0 +1,56 @@
...
+//!
+//! We exceed the number of cores specifically to see the
+//! "elbow" in the graph, when the number of threads tested
+//! exceeds the number of cores.  (This is good because otherwise
+//! -- assuming the software under test is perfectly scalable --
+//! the graph would be a straight line parallel with the x axis.)
+
+use std::{cmp, env, thread};
+
+fn main() {
```
djc, line 22: I think you can use `Result<(), Box<dyn StdError + Send + Sync + 'static>>` here and use `?` more? If you're going to make it script-like I'd probably ditch the "elaborate" `expect()` and `unwrap()` calls in favor of that.

djc (approved):

admin/threads-seq
ctz: What did you mean by rust script? Is there another way?

(I have other callers of this script -- ie, https://github.com/ctz/openssl-bench/pull/10/files )

djc: My suggestion was more like sticking a normal `threads-seq.rs` in an `examples` dir, without relying on nightly script support.

ctz: I think I prefer the -Zscript way, and AIUI it will stop needing nightly very very soon.

admin/threads-seq.rs
```diff
@@ -0,0 +1,51 @@
...
+    let count = args
+        .next()
+        .map(|c| usize::from_str(&c))
+        .unwrap_or(Ok(16))?;
+
+    let cpus = env::var("CPU_COUNT")
+        .map(|c| usize::from_str(&c))
+        .unwrap_or_else(|_| {
+            Ok(thread::available_parallelism()
+                .unwrap()
```
djc, line 34: Nit: also `?` this?

admin/threads-seq.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use std::{cmp, env, error, str::FromStr, thread};
+
+fn main() -> Result<(), Box<dyn error::Error + Send + Sync + 'static>> {
+    let mut args = env::args();
+    args.next(); // skip argv[0]
+    let count = args
+        .next()
+        .map(|c| usize::from_str(&c))
+        .unwrap_or(Ok(16))?;
```
djc, line 28: Nit: I'd instead do `.transpose()?.unwrap_or(16)`;

admin/threads-seq.rs
ctz: done (and also used nonzerousize everywhere)

PR #2193, opened 2024-11-05
Introduce `TicketRotator`, a version of `TicketSwitcher` with improved thread scalability

ctz:
The vastly overwhelming majority of ticket operations are "reads", as this covers encrypting a new ticket, decrypting an existing one, and checking if the ticketer needs rotating.

Since `TicketSwitcher` is used by default for every default ticketer, it is a bottleneck for multi-threaded servers as it serialised every single ticket creation and resumption attempt.

Here's what this PR looks like using the benchmark changes introduced in #2192 :

image

The Y axis here is handshakes per second _per thread_ --- the ideal shape of this graph is straight, flat lines -- that would mean perfect linear scalability with the number of threads. As you can see, where the ticketer is in play -- blue and yellow -- are improved in this PR.

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/ticketer.rs
+++ b/rustls/src/crypto/aws_lc_rs/ticketer.rs
@@ -33,7 +33,7 @@ impl Ticketer {
     /// [RFC 5077 §4]: https://www.rfc-editor.org/rfc/rfc5077#section-4
     #[cfg(feature = "std")]
     pub fn new() -> Result<Arc<dyn ProducesTickets>, Error> {
-        Ok(Arc::new(crate::ticketer::TicketSwitcher::new(
+        Ok(Arc::new(crate::ticketer::TicketRotator::new(
             6 * 60 * 60,
             make_ticket_generator,
         )?))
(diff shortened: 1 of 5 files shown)


rustls/src/crypto/aws_lc_rs/ticketer.rs
```diff
@@ -38,24 +38,6 @@ impl Ticketer {
...
-
-    /// Make the recommended `Ticketer`.  This produces tickets
-    /// with a 12 hour life and randomly generated keys.
-    ///
-    /// The `Ticketer` uses the [RFC 5077 §4] "Recommended Ticket Construction",
-    /// using AES 256 for encryption and HMAC-SHA256 for ciphertext authentication.
-    ///
-    /// [RFC 5077 §4]: https://www.rfc-editor.org/rfc/rfc5077#section-4
-    #[cfg(not(feature = "std"))]
-    pub fn new<M: crate::lock::MakeMutex>(
```
djc, line 50: This seems like a kind of big semver compatibility issue? Unless we have a way of finding everyone using rustls in no_std context (I guess releasing this is potentially a decent way of getting them into our issue tracker), I would be hesitant to release this as 0.23.x.

rustls/src/ticketer.rs
```diff
@@ -54,83 +52,49 @@ impl TicketSwitcher {
...
         let next = (self.generator)().ok()?;
-        if !are_recovering {
-            // Normal path, generate new next and place it in the state
-            let mut state = self.state.lock()?;
-            state.next = Some(next);
-            Some(state)
+
+        let mut write = self.state.write().ok()?;
+
+        if now > write.next_switch_time {
```
djc, line 85: I think this might be easier to follow if you do an early return for the "another thread beat us to it" case, at the cost of a little code duplication?

rustls/src/crypto/aws_lc_rs/ticketer.rs
ctz: I guess the other option is to reinstate this, and keep around the previous version of the ticketswitcher until the next breaking release? But I don't know where that leaves existing users (who already have a breaking API change if something in their dependency tree turns on `std`).

I think in the next breaking release we should do a pass to improve this pattern, so it looks like:

lock.rs:
```rust
struct StdMutex;
impl MakeMutex for StdMutex { ... obvious stuff ... }
```

elsewhere:
```rust
#[cfg(feature = "std")]
pub fn new(...) { new_detail::<StdMutex>() }

pub fn new_detail<M: crate::lock::MakeMutex>(...) { }
```

(this deals with that kind of feature-variant breaking change, but does not deal with the issue that `new_detail` leaks the otherwise-abstract details -- like "this uses a lock" -- into the public API.)

rustls/src/ticketer.rs
```diff
@@ -39,6 +41,7 @@ impl TicketSwitcher {
     /// longer than twice this duration.  `generator` produces a new
     /// `ProducesTickets` implementation.
     #[cfg(feature = "std")]
+    #[deprecated(note = "Use TicketRotator instead")]
```
djc, line 44: Nit: I think usual style is not to capitalize these.

djc (approved):
Nice!

cpu (approved):
Nice :+1:

The situation with the no-std ticketer features is unfortunate. Apologies we didn't catch that when it landed :weary:

mjguzik:
Mutex protection was indeed pretty bad.

So the fast path is still protected with a read-write lock, except taken in read mode. This still leaves performance on the table as the lock keeps being bounced during acquire (and presumably results in further cache misses when trying to use the target object).

What should work here and provide near linear scalability is changing the scheme to maintaining a reference to the ticketer which gets perodically updated instead.

Since I'm not fluent in Rust I'm not going to try to write my own patch, but I can outline this in pseudocode.

The fast path would do something like:
```
if worker->cached_ticketer != global_ticketer
       dropref(worker->cached_ticketer);
       worker->cached_ticketer = grabref(global_ticketer);
```

and that's it, any cacheline bouncing regarding concerning liveness of the thing will only be a perioc blip when it changes instead of happening on every use.

There should also be a single-threaded perf bump since 2 atomics are now elided.

However, it is plausible that even with this in place there is more bouncing. One would have to check with a profiler, but bottom line is that this code should scale almost perfectly and any slowdown should get sorted out.

ctz:
Hey, thanks for the note. I think what you're suggesting is what [arc-swap](https://docs.rs/arc-swap/latest/arc_swap/) does. However, AIUI that would lead to potentially many threads seeing and acting on an expired key. That could possibly lead to a large number of keys being cycled through quickly as those threads complete their rotation, and tickets produced during that time would be undecryptable. That is not a disaster, but seems a little undesirable. I guess we could have an additional mutex to serialize rotations.

mjguzik:
Taking a lock to do the swap was implied.

My general point is that if the current code can safely operate by taking a read lock in the fast path, then it can safely operate without it by instead holding a reference. This would both fix the scalability problem and speed up even single-threaded execution.

PR #2197, opened 2024-11-06
cancel-in-progress github action runs keyed on PR/ref

ctz:
This means multiple successive pushes to a ref or PR will only lead to one github action run being taken to completion -- the others will be started and then abandoned once they become obsolete.

See github docs: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-concurrency-groups

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,10 @@ permissions:
 
 on:
   push:
+    branches-ignore:
+    - 'gh-readonly-queue/**'
+    tags:
+    - '**'
   pull_request:
   merge_group:
   schedule:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,10 @@ on:
   schedule:
     - cron: '0 18 * * *'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build+test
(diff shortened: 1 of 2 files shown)


djc (approved):

cpu (approved):
This is great thank you! I've been doing this manually all the time and had no idea there was a Better Way :tm:

cpu:
> @github-merge-queue github-merge-queue bot removed this pull request from the merge queue due to failed status checks 1 minute ago

I started porting this over to other repos but maybe I should pause. Looks like there's a weird interaction with the merge queue?

djc:
Maybe needs an exception for `gh-readonly-queue` refs?

djc:
> > @github-merge-queue github-merge-queue bot removed this pull request from the merge queue due to failed status checks 1 minute ago
> 
> I started porting this over to other repos but maybe I should pause. Looks like there's a weird interaction with the merge queue?
> 
> > Canceling since a higher priority waiting request for 'rustls-refs/heads/gh-readonly-queue/main/pr-2197-828674444a3f64cd8ada3342f3c51fd686d52714' exists

I thought about this and figured I might wait for a week or so to see if we can shake out any issues.

ctz:
> Maybe needs an exception for `gh-readonly-queue` refs?

I've fixed it both ways: by including the action type in the key (so a "push" will never cancel a "merge_queue"), and also to stop building the merge queue pushes (see second commit).

ctz:
Hm, this is still not right.

ctz:
OK! I think this is working properly now:

> `Canceling since a higher priority waiting request for 'rustls-pull_request-2197' exists`

PR #2198, opened 2024-11-06
minor improvements to internal test imports

brodycj:


cpu (approved):
Thanks

ctz (approved):

djc (approved):
I'd prefer to move in the direction of `super::enums` instead of `crate::msgs::enums`.

(For the future, also not sure if this kind of thing is really worth a PR on its own.)

cpu:
> I would like to update this when I get a chance.

Let's close it for the meantime then. I don't think there's any value in having it left in draft status in the meantime, it just clutters up the PR list.

Thanks,

brodycj:
I just tried to push an update on this but it is now closed and I cannot reopen it. Can we reopen this?

cpu:
a

PR #2199, opened 2024-11-07
minor improvements to internal test imports

brodycj:
(similar to PR #2198, updated per minor review suggestion)

BRIEF CONTEXT: I am working on a proposal similar to PR #2088, with some updated imports; merging this separately would help me work on these updated imports in a slightly cleaner fashion.

### DIFF
--- a/rustls/src/msgs/handshake_test.rs
+++ b/rustls/src/msgs/handshake_test.rs
@@ -4,19 +4,13 @@ use std::{format, println, vec};
 
 use pki_types::{CertificateDer, DnsName};
 
-use super::enums::CertificateType;
-use super::handshake::{ServerDhParams, ServerKeyExchange, ServerKeyExchangeParams};
-use crate::enums::{
-    CertificateCompressionAlgorithm, CipherSuite, HandshakeType, ProtocolVersion, SignatureScheme,
-};
-use crate::error::InvalidMessage;
-use crate::msgs::base::{Payload, PayloadU16, PayloadU24, PayloadU8};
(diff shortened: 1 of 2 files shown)


cpu (approved):

rustls/src/msgs/handshake_test.rs
```diff
@@ -25,9 +19,14 @@ use crate::msgs::handshake::{
     HandshakePayload, HasServerExtensions, HelloRetryExtension, HelloRetryRequest, KeyShareEntry,
     NewSessionTicketExtension, NewSessionTicketPayload, NewSessionTicketPayloadTls13,
     PresharedKeyBinder, PresharedKeyIdentity, PresharedKeyOffer, ProtocolName, Random,
-    ServerEcdhParams, ServerExtension, ServerHelloPayload, ServerKeyExchangePayload, SessionId,
-    UnknownExtension,
+    ServerDhParams, ServerEcdhParams, ServerExtension, ServerHelloPayload, ServerKeyExchange,
+    ServerKeyExchangeParams, ServerKeyExchangePayload, SessionId, UnknownExtension,
 };
+
```
djc, line 25: I think `crate` and `super` imports usually go in the same block.

brodycj: Agreed & now fixed for the sake of consistency.

I generally prefer to keep some separation for the ease of reading & understanding, but this would be a separate discussion.

Please let me know if there is anything else you need for this one. Thanks!

djc (approved):

PR #2200, opened 2024-11-08
unstable feature to support targets with no atomic pointers - DRAFT DO NOT MERGE YET

brody4hire:
__STATUS:__ _see TODO items below_

Proposal to resolve issue: #2068

This is a proposal to support targets w/o atomic ptr by using portable-atomic `Arc` & `critical-section`, as optional features, _with help from a recently published contribution: <https://github.com/taiki-e/portable-atomic/pull/195>_

This solution requires use of Rust nightly in order for the required coercion to work with the _recent update from_ https://github.com/taiki-e/portable-atomic/pull/195 ... looks like the stabilization is blocked pending this: https://github.com/rust-lang/rust/issues/18598

Optional features added to `rustls/Cargo.toml`:

- `portable-atomic-arc` - enable use of `Arc` from portable-atomic-util
- `critical-section` - enable `critical-section` feature on once-cell & portable-atomic-util crates & include `portable-atomic-arc` feature - this is needed to support build on targets w/o atomic ptr

CI TESTING:

- this adds a new job to try `cross build` with `critical-section` on multiple targets including non-atomic ptr target
- Some additional CI testing with nightly Rust

COMPARISON TO ALTERNATIVE SOLUTIONS:

- use alias to `Rc` for target(s) with no atomic ptr - already proposed in PR #2088 - seems to imply some API mangling & use of an ugly macro to enable or disable use of `Send` & `Sync` traits; possibly harder to document & understand the updated API
- add & use casting macros to get this working with Rust stable - I have already done this in my work area, seems to imply some worse-looking API mangling, with more documentation burden & possibly harder to understand the updated API

---

__STATEMENT OF IMPACT__

This unstable `critical-section` feature would require using Rust nightly along with the unstable `--cfg portable_atomic_unstable_coerce_unsized` build option in order for the build to succeed with this feature.

The one form of impact I can think of is on users building rustls library with `--all-features` flag. I do NOT think this should affect anyone using rustls as a dependency for a higher-level library or application.

Alternative could be to use an unstable cfg option, like I had to do in <https://github.com/taiki-e/portable-atomic/pull/195>, but this seems to be inconsistent with the other rustls build options.

---

__XXX / TODO ITEMS:__

- [ ] resolve remaining XXX / TODO item(s) in `.github/workflows/build.yml`
- [ ] resolve remaining documentation TODO item(s) doc items in `rustls/src` (`rustls/src/lib.rs` & `rustls/src/crypto/mod.rs`)
- [ ] I noticed that https://docs.rs/rustls has references into documentation for Rust nightly, thinking these should point into documentation for Rust stable instead (moving this item out of PR #2285)
- [ ] check for any other remaining remaining XXX / TODO item(s) in this PR

/cc @bjoernQ @taiki-e

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,17 +76,32 @@ jobs:
       # nb. feature sets that include "fips" should be --release --
       # this is required for fips on windows.
       # nb. "--all-targets" does not include doctests
-      - name: cargo test (release; all features)
+      # unstable critical-section feature requires both `--cfg portable_atomic_unstable_coerce_unsized` & Rust nightly
+      - name: cargo test (release; all stable features)
+        if: ${{ !startsWith(matrix.rust, 'nightly') }}
+        run: cargo test --release --locked --lib $(admin/all-stable-features rustls) --all-targets
+        env:
+          RUST_BACKTRACE: 1
+      - name: cargo test (release; all features including unstable feature(s) - Rust nightly only - using unstable --cfg option)
+        if: startsWith(matrix.rust, 'nightly')
         run: cargo test --release --locked --all-features --all-targets
         env:
           RUST_BACKTRACE: 1
+          RUSTFLAGS: --cfg portable_atomic_unstable_coerce_unsized
 
       # nb. this is separate since `--doc` option cannot be combined with other target option(s) ref:
       # - https://doc.rust-lang.org/cargo/commands/cargo-test.html
-      - name: cargo test --doc (release; all-features)
+      - name: cargo test --doc (release; all stable features)
+        if: ${{ !startsWith(matrix.rust, 'nightly') }}
+        run: cargo test --release --locked $(admin/all-stable-features rustls) --doc
+        env:
+          RUST_BACKTRACE: 1
+      - name: cargo test --doc (release; all features including unstable feature(s) - Rust nightly only - using unstable --cfg option)
+        if: startsWith(matrix.rust, 'nightly')
         run: cargo test --release --locked --all-features --doc
         env:
           RUST_BACKTRACE: 1
+          RUSTFLAGS: --cfg portable_atomic_unstable_coerce_unsized
 
       - name: cargo test (debug; aws-lc-rs)
         run: cargo test --no-default-features --features aws_lc_rs,tls12,read_buf,logging,std --all-targets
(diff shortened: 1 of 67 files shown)


djc:
I think it's unlikely we will want to merge this in something close to its current form. rustls is a small project with a small group of maintainers and we don't necessarily have in-house expertise on embedded platforms and their needs. We also have API compatibility constraints that we want to hold up for the benefit of the (much larger) ecosystem of folks deploying rustls to more general purpose platforms (that is, where `std::sync::Arc` is available, at least..).

I haven't reviewed this in detail since I don't think it's close to being reviewable, in the sense of having (a) a high-level description of the constraints that made rustls a tricky fit for your platform and how you overcame these challenges and (b) a clean commit history with commits that incrementally change rustls to fit your needs.

Instead of this approach, I think a better solution might be to incrementally modularize rustls more such that you can build on top of the logic that rustls provides without the rustls project having to maintain a bunch of workarounds for the limits of your platform.

ctz:
Yes, I agree.

I'm also pretty concerned about incorporating types from another crate in our public API (even behind a feature gate); I feel that would make cutting a rustls 1.0 even harder.

I would perhaps be willing to entertain something like:

```rust
diff --git a/rustls/src/lib.rs b/rustls/src/lib.rs
index d8562932..dc428236 100644
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -674,6 +674,11 @@ pub mod lock;
 /// Polyfills for features that are not yet stabilized or available with current MSRV.
 pub(crate) mod polyfill;
 
+mod sync {
+    // This crate refers to Arc only via this point.
+    pub(crate) use alloc::sync::Arc;
+}
+
 #[cfg(any(feature = "std", feature = "hashbrown"))]
 mod hash_map {
     #[cfg(feature = "std")]
```

That would make such a fork much easier to maintain, as the difference would be limited to one place (assuming the replacement for Arc was truly interchangeable.)

brodycj:
Any pointers how we could start improving the structure of this project?

From a quick look through this project I did already start having some ideas on reducing the amount of dependency on `Arc` & providing an Arc-less API. Definitely not an overnight project but I will see how much further I can get with these ideas.

I personally may not mind that `mod sync` idea, though I suspect some others may be disappointerd ... I was really hoping we could find a way to do this without the forking.

bjoernQ:
Long-term seeing Rustls being more modular (and offering Arc-free APIs) would be a good thing.

I was also hoping for not going the fork-route but if that fork can be easily maintained that's better than nothing

djc:
> Any pointers how we could start improving the structure of this project?
> 
> From a quick look through this project I did already start having some ideas on reducing the amount of dependency on `Arc` & providing an Arc-less API. Definitely not an overnight project but I will see how much further I can get with these ideas.

In my mind we mostly use `Arc` to avoid cloning (allocations) and to wrap around trait objects. Both of these cases are kind of accidental complexity in that they are not really necessary for the code to function. It might be possible to have a `ClientConfig` trait and use generics to abstract over the use of `Arc<ClientConfig>` (for example, I think we could use abstract types in the handshake state types).

However, this is just spitballing and we'd have to carefully control how to allocate the extra complexity, so whoever takes this on needs to walk a fine line.

brodycj:
> ```rust
> +mod sync {
> +    // This crate refers to Arc only via this point.
> +    pub(crate) use alloc::sync::Arc;
> +}
> ```

If I would continue with this kind of idea, to help with the possible forking, should I continue in this PR or raise a brand new PR to do this?

> In my mind we mostly use `Arc` to avoid cloning (allocations) and to wrap around trait objects. Both of these cases are kind of accidental complexity in that they are not really necessary for the code to function.

After some trial-and-error in my personal workspace, I am now thinking of an idea to use config builders to build immutable config objects, that could be used from multiple threads, hopefully without all of the atomic locking requirements that come with `Arc`. I may need some time to consider this further ... I would really love to work on this, cannot guarantee near-term prioirity for this kind of idea.

cpu:
I think #2285 is our preferred way to support targets w/o atomic pointers. Let's close this PR for now since it has conflicts with main and doesn't seem likely to be merge-able in the near future. Thanks!

PR #2201, opened 2024-11-08
Static site for rustls.dev, including home for performance reports

ctz:
This is a more permanent and visible home for https://github.com/rustls/rustls/wiki/Performance-measurements-(2024%E2%80%9010%E2%80%9018) -- with more to come (for arm64, and multithreaded cases).

At the same time, I started a landing site for rustls.dev. Feedback on what we say there gratefully accepted!

To see what this looks like, schedule https://github.com/rustls/rustls/actions/workflows/docs.yml for the branch `jbp-landing-page` and then look at https://rustls.dev/ . Each time `main` changes (and nightly) this will be lost and you'll see a github 404.


djc:
Did you consider picking up something like [oranda](https://opensource.axo.dev/oranda/)? Seems like it might be a good fit.

ctz:
I haven't heard of that, will look into it.

ctz:
OK, I spent an hour or so with oranda. I like the way it can present github release notes, but unfortunately it only really supports a handful of built-in themes (and limited CSS customisation on top of that). I filed an issue about that, but in the meantime I think that possibility need not block this?

djc (approved):

djc:
The `Makefile` and custom `pulldown-cmark` stuff seems a little ugly, I'd prefer doing something with Zola or maybe another Rust script that uses Askama and the pulldown-cmark library to do the same stuff.

cpu (approved):
Zola seems pretty heavy weight for this case and I think I'd prefer the Makefile personally. A Rust script seems OK.

ctz:
Zola seems to work pretty well, stays out the way, and has good docs. I'm minded to write the absolute minimum of code for this.

cpu:
> Zola seems to work pretty well, stays out the way, and has good docs. I'm minded to write the absolute minimum of code for this.

My enthusiasm for static site generators was reduced by the experience of Hugo and Jekyll. Maybe Zola will be better, but if it's like those two then there's a non-zero amount of headache related to keeping up to date with the tooling and associated breaking changes. 

This branch is also hardcoding a Zola version in CI in a way that I think excludes Rennovate from helping us keep it in-sync with upstream releases. Is there a way to avoid that? 

I don't feel strongly enough to block this getting merged but given there's only one page with no significant site structure to contend with it feels like dragging in a whole site generation framework is overkill.

cpu:
> This branch is also hardcoding a Zola version in CI in a way that I think excludes Rennovate from helping us keep it in-sync with upstream releases. Is there a way to avoid that?

I noticed Curl has some special markup in some of its CI YAML that looks like it configures Renovate to update versions, e.g.: https://github.com/curl/curl/blob/774844ab5db84e2368389e6cd9f0b8f81852b738/.github/workflows/linux.yml#L38-L57

That might be an option if we stick with Zola.

ctz:
> My enthusiasm for static site generators was reduced by the experience of Hugo and Jekyll. Maybe Zola will be better, but if it's like those two then there's a non-zero amount of headache related to keeping up to date with the tooling and associated breaking changes.
> 
> This branch is also hardcoding a Zola version in CI in a way that I think excludes Rennovate from helping us keep it in-sync with upstream releases. Is there a way to avoid that?

I think these two points kinda cancel each other out a bit? For the purposes of static site generation, I'm happy to fix on a working version until we have a reason to change (for my own blog, I basically did that with jekyll, only upgrading it 3 times in 11 years).

cpu:
> For the purposes of static site generation, I'm happy to fix on a working version until we have a reason to change

I feel like that just shifts the pain, collecting it all up at the point where you have to jump many versions at once.

We're using so few of the features of Zora that it seems excessive weighed against a small makefile directly invoking the one thing we do want, markdown -> HTML for ~a single file but if I'm in the minority here I'm happy to go along with the crowd.

djc:
> I feel like that just shifts the pain, collecting it all up at the point where you have to jump many versions at once. Put another way: if we adopt Zora we're on the hook for future update pain.

I'm sympathetic and don't mind either route.

That said, I'm optimistic about the Zola route just based on the notion that the Rust culture around backwards compatibility is pretty good.

I think Joe will most likely end up being on the hook for maintenance so ultimately his call?

ctz:
I think one of two things will happen:

- either we use more of Zola's features, and therefore it was worth it, or
- we don't, and fall-back to something simpler is pretty trivial if that becomes needed

PR #2203, opened 2024-11-09
fix: do not send session_ticket(35) extension for TLS 1.3

### LINKED ISSUE #2202: `session_ticket` extension is sent by TLS 1.3 client
I am looking at
https://github.com/rustls/rustls/blob/d1c33f8641c1c69edc27d98047c38f7f852f55eb/rustls/src/client/hs.rs#L587-L592

This code pushes extension 35 (`session_ticket`) defined in https://datatracker.ietf.org/doc/html/rfc5077
This extension has nothing to do with TLS 1.3 session resumption mechanism.
https://www.rfc-editor.org/rfc/rfc8446 never mentions this extension and explicitly deprecated RFC 5077:

> This document supersedes and obsoletes previous versions of TLS,
> including version 1.2 [[RFC5246](https://www.rfc-editor.org/rfc/rfc5246)].  It also obsoletes the TLS ticket
> mechanism defined in [[RFC5077](https://www.rfc-editor.org/rfc/rfc5077)] and replaces it with the mechanism
> defined in [Section 2.2](https://www.rfc-editor.org/rfc/rfc8446#section-2.2).

So it seems this code should be changed like this:
```diff
diff --git a/rustls/src/client/hs.rs b/rustls/src/client/hs.rs
index 9ddad98a..24f9124a 100644
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -584,8 +584,8 @@ fn prepare_resumption<'a>(
     let resuming = match resuming {
         Some(resuming) if !resuming.ticket().is_empty() => resuming,
         _ => {
-            if config.supports_version(ProtocolVersion::TLSv1_3)
-                || config.resumption.tls12_resumption == Tls12Resumption::SessionIdOrTickets
+            if config.supports_version(ProtocolVersion::TLSv1_2)
+                && config.resumption.tls12_resumption == Tls12Resumption::SessionIdOrTickets
             {
                 // If we don't have a ticket, request one.
                 exts.push(ClientExtension::SessionTicket(ClientSessionTicket::Request));
```

Only if TLS 1.2 is enabled and its RFC 5077 is used we should request RFC 5077 ticket.
TLS 1.3 servers should send `new_session_ticket` extension (4) if they support TLS session resumption using new PSK mechanism regardless of whether `session_ticket` extension is sent.

link2xt:
Closes #2202

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -584,8 +584,8 @@ fn prepare_resumption<'a>(
     let resuming = match resuming {
         Some(resuming) if !resuming.ticket().is_empty() => resuming,
         _ => {
-            if config.supports_version(ProtocolVersion::TLSv1_3)
-                || config.resumption.tls12_resumption == Tls12Resumption::SessionIdOrTickets
+            if config.supports_version(ProtocolVersion::TLSv1_2)
+                && config.resumption.tls12_resumption == Tls12Resumption::SessionIdOrTickets
             {
                 // If we don't have a ticket, request one.
                 exts.push(ClientExtension::SessionTicket(ClientSessionTicket::Request));
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -6199,6 +6199,35 @@ fn test_server_rejects_clients_without_any_kx_groups() {
     );
 }
 
+/// Tests that session_ticket(35) extension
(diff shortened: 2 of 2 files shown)


djc (approved):
This change makes sense to me, thanks! Definitely needs some test coverage.

PR #2207, opened 2024-11-12
multithreading benchmarking write-up

ctz:
[Rendered](https://github.com/rustls/rustls/blob/jbp-thr-perf-results/website/content/perf/2024-11-28-threading/index.md)


cpu (approved):

djc:
Looks like the charts didn't make it into your PR? (At least they don't show up in the Rendered for me.)

ctz:
> Looks like the charts didn't make it into your PR? (At least they don't show up in the Rendered for me.)

Fixed

djc:
I think the log-scale x axis made things quite a bit better.

For the right-hand key numbers in those distribution charts you currently show N, min, mean, stddev, 99.9% and max. I think percentiles are generally a better (and the commonly accepted) way of thinking about latencies (which don't usually really follow a normal distribution), so I'd (a) skip N -- which is the same for all implementations anyway and (b) report P5, P50, P75 and P99 (or something like that). Might also be nice to see if you can better align the numbers vertically so that they're easier to vertically compare.

Also, still curious about OpenSSL numbers.

All in all, rustls perf is awesome, great work!

website/content/perf/2024-11-12-threading/index.md
```diff
@@ -0,0 +1,175 @@
...
+
+![htop output, showing 80 cores at high utilization](htop-80-99.png)
+
+Note that the below charts use a base 10 logarithm x axis, to adequately show the
+four distributions in one figure.
+
+![graph of TLS1.2 resumed handshake server latency](latency-resume-tls12-server.svg)
+
+Here we can clearly see an improvement between OpenSSL 3.0 and OpenSSL 3.3.
+rustls has the tightest distribution, followed by BoringSSL.
```
alex, line 164: Assuming this is describing the image above, that graph shows OpenSSL 3.3 having a tighter distribution than Boring

ctz: Because OpenSSL 3.3 is further along on the log scale, its distribution is visually narrower but covers a larger space of numbers. BoringSSL is (roughly -- reading off the graph) from 20-90us, whereas OpenSSL 3.3 is maybe 700-900us?

alex: Oh man, that was... not visually obvious. Obviously I should have read the axis closer though. I think a parenthetical pointing that out might help others avoid the same mistake. Suggested language:

> (OpenSSL 3.3's distribution may appear visually tighter than BoringSSL's, however the scale is logarithmic, so it is not.)

ctz:
I've addressed comments thus far; tomorrow I will regenerate the data at specific versions.

alex:
Oh, should this measure against 3.4.0, since its been out for about a month?

website/content/perf/2024-11-28-threading/index.md
```diff
@@ -0,0 +1,183 @@
...
+### How the benchmarks work
+Compared to [previous reports](/perf), the benchmark tool can now perform the same
+benchmarks in many threads simultaneously.  Each thread runs the same benchmarking
+operation as before, and threads do not contend with each other _except_ via
+the internals of the TLS library.
+
+As before, the benchmarking is performed by measuring a TLS client "connecting" to
+a TLS server over a memory buffer -- there is no network latency, system calls, or
+other overhead that would be present in a typical networked application.
+This arrangement actually should be the worse case for multithreaded testing: every
```
djc, line 27: "the worse case"?

djc (approved):
I'd still recommend getting rid of min, max (I think these are less relevant to the audience than P5/P99.9) and probably stddev. I'd probably also get rid of P75 and P99 (or P99.9) since these don't seem to add much information over the P50/P99.9 latencies that are already reported.

ctz:
> I'd still recommend getting rid of min, max (I think these are less relevant to the audience than P5/P99.9) and probably stddev. I'd probably also get rid of P75 and P99 (or P99.9) since these don't seem to add much information over the P50/P99.9 latencies that are already reported.

deleted min, max, stddev, P75 and P99.9. added P90. that leaves P5, P50, P90, P99. (this also avoids the formatting looking odd when where are fewer than 4 rows per table, heh)

website/content/perf/2024-11-28-threading/index.md
```diff
@@ -0,0 +1,181 @@
...
+This arrangement actually should be the worst case for multithreaded testing: every
+thread should be working all the time (rather than waiting for IO) and therefore
+contention on any locks in the library under test should be maximal.
+
+### Versions
+The benchmarking tool used for both OpenSSL and BoringSSL was [openssl-bench `02249496`](https://github.com/ctz/openssl-bench/tree/02249496d2963e2a0b694e7be3b37f0d85f8eccd).
+
+#### BoringSSL and OpenSSL
+The version of BoringSSL and its build instructions are the same
+as [the previous report](2024-10-18-report.md).
```
alex, line 36: Because this is an index.md, I don't think this type of relative link works

ctz: Doh, and our link checker just whinged about that. Will do a follow-up.

alex: That's why I noticed. I am not a markdown savant 😂 . (I'm excited to share these results)

PR #2208, opened 2024-11-13
Extract benchmark tool from core crate

ctz:
The end goal here is to allow this tool to depend on providers in other crates.

The second commit is a straight rename from `rustls-bench/src/bench_impl.rs` to `rustls-bench/src/main.rs` but git's rename detection is broken for it for some reason.


lattaai6:
Hello, I tried to solve the issue.

This is what I did:

Add optional dependencies for external crypto providers in the Cargo.toml file. This allows rustls to optionally depend on external crates for crypto implementations.

You can review changes in this commit: https://github.com/lattaai6/rustls-rustls-2208/commit/24526edd3973b334f35d30108657f56413d0de3c.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

rustls-bench/src/main.rs
```diff
@@ -832,16 +833,96 @@ impl Options {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, ValueEnum)]
+enum Provider {
+    None,
```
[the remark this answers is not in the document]
ctz: This was needed when built with no features ("references are always considered inhabited") but actually making the receiver `self` rather than a reference fixes that.

rustls-bench/src/bench_impl.rs
```diff
@@ -1,1332 +0,0 @@
-// This program does assorted benchmarking of rustls.
```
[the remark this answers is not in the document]
ctz: Actually squashing this into the previous commit works better.

rustls-bench/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "rustls-bench"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+clap = { version = "4", features = ["derive", "env"] }
```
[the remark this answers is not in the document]
cpu: I have a vague memory of bumping into an MSRV issue related to workspace inheritance that made it seem like we might need MSRV 1.64+ first. [I think this comment thread](https://github.com/rustls/rustls/pull/1718#discussion_r1626195490) had some details.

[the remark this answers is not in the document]
djc: Ahh yes, that's probably the reason we don't have it here yet.

rustls-bench/src/main.rs
```diff
@@ -685,11 +683,12 @@ fn bench_memory(params: &BenchmarkParam, conn_count: u64) {
 
 fn make_server_config(
     params: &BenchmarkParam,
+    options: &Options,
```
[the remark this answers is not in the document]
ctz: Added a commit that does this refactor

rustls-bench/src/main.rs
```diff
@@ -413,41 +419,32 @@ fn multithreaded(
...
-        params.key_type,
-        params.ciphersuite,
-        if clientauth == ClientAuth::Yes {
-            "mutual"
-        } else {
-            "server-auth"
+        params.proto.version,
+        params.proto.key_type,
+        params.proto.ciphersuite,
+        match params.client_auth {
```
djc, line 429: Nit: further steps, consider extracting this into a method on `Parameters`?

rustls-bench/src/main.rs
```diff
@@ -686,17 +652,11 @@ fn bench_memory(
...
-fn make_server_config(
-    params: &BenchmarkParam,
-    options: &Options,
-    client_auth: ClientAuth,
-    resume: ResumptionParam,
-    max_fragment_size: Option<usize>,
-) -> Arc<ServerConfig> {
-    let provider = Arc::new(options.provider.build());
-    let client_auth = match client_auth {
+fn make_server_config(params: &Parameters) -> Arc<ServerConfig> {
```
djc, line 655: Maybe this and `make_client_config()` can now be `Parameters::server_config()`?

rustls-bench/src/main.rs
```diff
@@ -785,6 +756,66 @@ fn lookup_matching_benches(name: &str) -> Vec<&BenchmarkParam> {
     r
 }
 
+/// General parameters common to several kinds of
+/// benchmark.
+#[derive(Clone)]
+struct Parameters {
+    /// Set by the user.
+    options: Options,
```
djc, line 764: Not sure if the `Options` type still makes sense? I thought that was directly used for argument parsing, but I noticed just now it is instantiated separately. I'd review whether it makes more sense to flatten it into `Parameters`.

Or, maybe these can actually be fields on `Parameters` instead of being recreated?

rustls-bench/src/main.rs
ctz: hoisted up the `client_auth` match into a fn on `ClientAuth`

rustls-bench/src/main.rs
ctz: Removed `Options` and inlined into `Parameters`. That means we need to resolve the provider in several places, but I think that's OK.

djc (approved):

rustls-bench/src/main.rs
```diff
@@ -686,103 +641,173 @@ fn bench_memory(
     }
 }
 
-fn make_server_config(
-    params: &BenchmarkParam,
-    options: &Options,
+fn lookup_matching_benches(name: &str) -> Vec<&BenchmarkParam> {
```
cpu, line 644: `name` might be better expressed as `ciphersuite_name`?

rustls-bench/src/main.rs
```diff
@@ -686,103 +641,173 @@ fn bench_memory(
...
+
+    if r.is_empty() {
+        panic!("unknown suite {:?}", name);
+    }
+
+    r
+}
+
+/// General parameters common to several kinds of
+/// benchmark.
```
cpu, line 658: nit: it doesn't seem like this headline is long enough to merit a linebreak.

cpu (approved):

PR #2209, opened 2024-11-13
StoresServerSessions: guarantee keys are high entropy

ctz:
In TLS1.2, these were wire encoding of a SessionID (which is a 0..32 byte string with a one-byte length prefix). Fix that, which we can do because `Vec<u8>` knows its length.

In TLS1.3 this was already 32 random bytes.

This opens the way to allow implementations to shard the storage by key.

In the future we could change the key type to be `[u8; 32]`.

---

I experimented a bit with sharding the server session cache like this, but it didn't seem to improve scalability in a measurable way: probably because the critical section is already very very small? But I think this commit is worth having anyway.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -145,7 +145,7 @@ impl Codec<'_> for SessionId {
     fn encode(&self, bytes: &mut Vec<u8>) {
         debug_assert!(self.len <= 32);
         bytes.push(self.len as u8);
-        bytes.extend_from_slice(&self.data[..self.len]);
+        bytes.extend_from_slice(self.as_ref());
     }
 
     fn read(r: &mut Reader<'_>) -> Result<Self, InvalidMessage> {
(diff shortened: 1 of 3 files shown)


cpu (approved):
SGTM :+1:

djc (approved):
I think you're arguing that `SessionId::as_ref()` is higher entropy than `SessionId::get_encoding()`. I don't find the Wikipedia entry on [Entropy](https://en.wikipedia.org/wiki/Entropy) very accessible for information-theoretical purposes, but don't both values have basically the same information-theoretical **amount** of randomness although the former has slightly higher **rate** of randomness?

ctz:
Yeah, they have the same total amount of entropy, and the byte length prefix doesn't add anything to that. But if one were to shard on, say, the lower 4 bits from the first byte of each key, you might have a surprising distribution (~all TLS1.2 keys would be 32 & 15 = 0, TLS1.3 would be uniformly distributed) -- https://github.com/bytedance/g3/blob/405bbe12ddb3ded6c1931dfd32f03efbc632656d/lib/g3-types/src/net/rustls/cache.rs#L76-L81

PR #2215, opened 2024-11-16
ci: enable bindgen feature on correct aws-lc-sys

ctz:
`cargo add` adds the latest version of a package, meaning we end up with two aws-lc-sys versions if there is a newer version than the one in our Cargo.lock.

`cargo add aws-lc-sys@*` ought to work, but chooses the wrong version and then complains about the `bindgen` feature missing. (ref: https://github.com/rust-lang/cargo/issues/14832)

Instead, just splat the line into Cargo.toml with sed.


djc (approved):
Nice, much better!

PR #2220, opened 2024-11-18
Move MSRV to 1.71

ctz:
As discussed on discord.

Most of this PR is actually adopting the new let-else syntax; this could be a later PR if desired.

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -198,52 +198,49 @@ mod client_hello {
                 (share.group == selected_kxg.name()).then_some((share, selected_kxg))
             });
 
-            let chosen_share_and_kxg = match chosen_share_and_kxg {
-                Some(s) => s,
-                None => {
-                    // We don't have a suitable key share.  Send a HelloRetryRequest
-                    // for the mutually_preferred_group.
-                    self.transcript.add_message(chm);
+            let Some(chosen_share_and_kxg) = chosen_share_and_kxg else {
+                // We don't have a suitable key share.  Send a HelloRetryRequest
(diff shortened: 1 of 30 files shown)


rustls/src/server/tls13.rs
```diff
@@ -285,16 +282,16 @@ mod client_hello {
                 for (i, psk_id) in psk_offer.identities.iter().enumerate() {
                     let now = self.config.current_time()?;
 
-                    let resume = match self
+                    let Some(resume) = self
```
djc, line 285: Pre-existing nit: IMO this wasn't very readable before and hasn't gotten better (maybe even slightly worse) with let-else. Maybe tweak that?

(I usually try to avoid multi-line scrutinees because they make it harder for me to make sense of the control flow. There are a few more examples in this commit...)

djc (approved):
Seems like a nice improvement!

Too bad GitHub syntax highlighting doesn't understand let-else...

rustls/src/server/tls13.rs
ctz: added a refactoring commit for this

djc: Nice!

cpu (approved):

algesten:
With ureq (also the new ureq3 which is RC 2), I try to maintain a conservative MSRV. This change forces me to fall in line with 1.71, which is unfortunate since I had hoped to stay on 1.65 for now. I know I'm in a minority that considers MSRV bumps borderline breaking change in semver terms.

djc:
@algesten someone just opened #2239, let's collect all the discussion there?

PR #2221, opened 2024-11-18
Avoid inefficient aws-lc-rs `open_within()`

ctz:
`open_within` is a specialised function useful for the case where we want the plaintext to end up at a different location to the ciphertext.  In TLS1.2 GCM decryption we have:

```
  0                       8
  nn nn nn nn nn nn nn nn cc cc cc ...
  `---------------------' `----------'
      explicit nonce       ciphertext
```

And we want to shift the ciphertext left during decryption. (This is, in fact, much less important now our message decryption functions return a borrow of the input buffer.)

In *ring*, this API is implemented by writing the plaintext to the desired location during decryption, making it entirely free.

In aws-lc-rs, this API is implemented by doing the decryption in-place, and then using `copy_within` to do the offsetting. This is not free.  `rustls-graviola` does the same thing.

Avoid it by adding a new `InboundOpaqueMessage::into_plain_message()` variant which accepts the subrange of the plaintext.

This only affects the receive direction of TLS1.2 GCM suites. But the improvement is very much worth having:

| test | vers | suite | frag | direction | before | after | unit | diff |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| bulk | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 | max_fragment_size:default | recv | 8828.1 | 9353.55 | MB/s | +6% |
| bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 | max_fragment_size:default | recv | 10439.4 | 11354.4 | MB/s | +8.8% |
| bulk | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 8003.23 | 8565.64 | MB/s | +7% |
| bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 9185.1 | 9917.08 | MB/s | +8% |

(measured on my laptop, not comparable to other numbers.)

This also should address the slightly mysterious closeness seen on this graph.

image

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/tls12.rs
+++ b/rustls/src/crypto/aws_lc_rs/tls12.rs
@@ -285,7 +285,7 @@ impl MessageDecrypter for GcmMessageDecrypter {
         let payload = &mut msg.payload;
         let plain_len = self
             .dec_key
-            .open_within(nonce, aad, payload, GCM_EXPLICIT_NONCE_LEN..)
+            .open_in_place(nonce, aad, &mut payload[GCM_EXPLICIT_NONCE_LEN..])
             .map_err(|_| Error::DecryptError)?
             .len();
 
@@ -293,8 +293,11 @@ impl MessageDecrypter for GcmMessageDecrypter {
             return Err(Error::PeerSentOversizedRecord);
         }
 
-        payload.truncate(plain_len);
-        Ok(msg.into_plain_message())
+        Ok(
+            msg.into_plain_message_range(
+                GCM_EXPLICIT_NONCE_LEN..GCM_EXPLICIT_NONCE_LEN + plain_len,
+            ),
+        )
     }
 }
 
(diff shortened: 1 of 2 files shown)


djc (approved):
Nice win!

cpu (approved):
Nice! Did you find this from manual examination or did it come up in profiling?

djc (approved):

ctz:
This was from manual examination, following thinking about how to eliminate https://github.com/ctz/graviola/blob/16878633f48a19a49f23295c76cadba981fdeef1/rustls-graviola/src/aead.rs#L284

PR #2222, opened 2024-11-19
Normalise syntax of `test_for_each_provider!`

ctz:
This uses `macro_rules_attribute` (a new dev-dep) to apply a declarative macro as a proc macro.  That has the advantage that the use-site syntax is standard rust, which means rustfmt can format it.

(If we agree this is a good direction, I will later do something similar to the integration tests.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1633,6 +1633,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "macro_rules_attribute"
+version = "0.2.0"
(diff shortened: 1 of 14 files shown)


ctz:
> Did you consider just porting `test_for_each_provider!()` to be a procedural macro? Would save one layer of abstraction at least... Although this seems like an okay solution too.

I think I kinda prefer declarative macros for this type of code generation -- probably just due to familiarity at this point.

djc (approved):

PR #2224, opened 2024-11-21
bench: optionally record individual handshake latencies

ctz:


rustls-bench/src/main.rs
```diff
@@ -268,13 +283,16 @@ fn bench_handshake(params: &Parameters) {
 }
 
 fn bench_handshake_buffered(
+    params: &Parameters,
```
djc, line 292: Argument order nit: take this last?

rustls-bench/src/main.rs
```diff
@@ -1352,6 +1406,14 @@ impl TempBuffers {
     }
 }
 
+fn wall_time() -> f64 {
+    duration_nanos(
+        SystemTime::now()
```
djc, line 1422: Shouldn't this be using `Instant` instead?

rustls-bench/src/main.rs
```diff
@@ -661,12 +667,17 @@ fn bench_memory(
     }
 }
 
-fn lookup_matching_benches(ciphersuite_name: &str) -> Vec<&BenchmarkParam> {
-    let r: Vec<&BenchmarkParam> = ALL_BENCHMARKS
+fn lookup_matching_benches(
+    ciphersuite_name: &str,
+    key_type: &Option<KeyType>,
```
djc, line 672: Since `KeyType` is `Copy` anyway, why take this by reference?

djc (approved):

rustls-bench/src/main.rs
```diff
@@ -99,6 +100,13 @@ struct Args {
     )]
     multiplier: f64,
 
+    #[arg(
+        long,
+        env = "BENCH_LATENCY",
+        help = "Writes individual handshake latency into files starting with this string.  The files are named by appending a role (client/server), a thread id, and 'latency.tsv' to the given string."
+    )]
+    latency: Option<String>,
```
cpu, line 108: maybe this would be more descriptive as `latency-prefix`? :thinking:

cpu (approved):

rustls-bench/src/main.rs
ctz: I don't think you can go from an Instant to something that can be written a file? Aside from that, this is intended to produce similar output files to the C++ version which uses `gettimeofday` here.

djc: Well, you can get a `Duration` by comparing two `Instant`s but I guess that doesn't help if you're trying to match C++ output.

PR #2228, opened 2024-11-21
ML-KEM-only key agreement

dconnolly:
This may be relevant to https://github.com/rustls/rustls/issues/2056

TLSWG has [approved ML-KEM-only codepoints](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8) for key agreement in TLS 1.3, this is a first stab at supporting that. Restructuring feedback welcome.

This is ML-KEM-768 to start, 512 and 1024 are planned but wanted to get eyes on the approach first.


ctz:
Thanks for this -- going to run with it to do the general hybrid structure refactoring I mentioned a few weeks back.

PR #2229, opened 2024-11-22
Manage dependencies via the workspace

djc:


Cargo.toml
```diff
@@ -35,6 +35,60 @@ default-members = [
...
+rsa = { version = "0.9", features = ["sha2"], default-features = false }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+sha2 = { version = "0.10", default-features = false }
+signature = "2"
+subtle = { version = "2.5.0", default-features = false }
+time = { version = "0.3.6", default-features = false }
+tikv-jemallocator = "0.6"
+tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"]}
+webpki = { package = "rustls-webpki", version = "0.102", features = ["alloc"], default-features = false }
```
ctz, line 86: I think this needs `0.102.8`

ctz (approved):
Personally I would've done this only for cases where the same version appears more than once (eg, "ecdsa" is specific to "provider-example", so it only need appear there). But this is fine too.

djc:
> Personally I would've done this only for cases where the same version appears more than once (eg, "ecdsa" is specific to "provider-example", so it only need appear there). But this is fine too.

I agree that the provider-examples in particular are a bit out of place, but in my experience when managing dependencies (updates/MSRV bumps etc) it is really helpful to have everything in one list.

cpu:
>  in my experience when managing dependencies (updates/MSRV bumps etc) it is really helpful to have everything in one list.

That's also my experience. It's really nice to have a one-stop-shop.

Thanks for doing this!

PR #2230, opened 2024-11-22
Periodic maintenance of 0.22 release branch

ctz:
Just getting this branch green again.


ctz:
Not sure what is up with cifuzz currently, but it is also broken on main.

djc:
> Not sure what is up with cifuzz currently, but it is also broken on main.

Pinged you on Discord about it -- looks like an upstream fuzzing corpus disappeared?

djc (approved):

PR #2238, opened 2024-11-25
once_cell dependency updates (for no-std usage)

brodycj:
- specify most recent version in top-level `Cargo.toml`
- remove redundant version specifier from `rustls/Cargo.toml`
- specify required features in `rustls/Cargo.toml` instead of top-level `Cargo.toml`
- move comment to the end of the line for `once_cell` dependency in `rustls/Cargo.toml`

__RATIONALE - updated:__

I think this would help ensure downstream user is using the more recent version of `once_cell`, as I think would be needed with the updates I proposed in PR #2200.

It looks to me like `once_cell` is only required for `no-std`, and usage may change somewhat further if PR #2200 is accepted. I was thinking that putting everything for `once_cell` (except for version spec) together could help improve cohesiveness and help further preparations for PR #2200.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -23,8 +23,7 @@ brotli = { workspace = true, optional = true }
 brotli-decompressor = { workspace = true, optional = true }
 hashbrown = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
-# only required for no-std
-once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
+once_cell = { workspace = true, default-features = false, features = ["alloc", "race"] } # only required for no-std
 ring = { workspace = true, optional = true }
 subtle = { workspace = true }
 webpki = { workspace = true }
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -23,7 +23,7 @@ brotli = { workspace = true, optional = true }
 brotli-decompressor = { workspace = true, optional = true }
 hashbrown = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
-once_cell = { workspace = true, default-features = false, features = ["alloc", "race"] } # only required for no-std
+once_cell = { workspace = true, default-features = false, features = ["alloc", "race"] } # (only required for no-std)
 ring = { workspace = true, optional = true }
 subtle = { workspace = true }
 webpki = { workspace = true }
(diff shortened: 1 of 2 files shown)


brodycj (commented):
Unfortunately I am thinking that this proposal does introduce another inconsistency. This proposal specifies features required for `once_cell` in `rustls/Cargo.toml`, while features for the other dependencies are now specified in top-level `Cargo.toml`.

I personally feel this is the best way to specify features that are required for `once_cell`, which is only needed to support no-std.

In terms of opinion & feeling, I feel that it would be best to specify features for the other dependencies in the package-specific `Cargo.toml`, but am not so sure how hard I should press this feeling. I would like to leave this whole proposal open for discussion, I wouldn't mind if we reject this proposal in favor of some other solution to improve the consistency with these dependencies.

djc:
I don't think we should take a separate PR for this. If #2200 has a specific lower bound it needs for once_cell, it should bump the dependency accordingly (and ideally not by requiring the latest version). `Cargo.toml` is used to describe the oldest version that is expected to work.

rustls/Cargo.toml
```diff
@@ -23,8 +23,7 @@ brotli = { workspace = true, optional = true }
 brotli-decompressor = { workspace = true, optional = true }
 hashbrown = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
-# only required for no-std
-once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
+once_cell = { workspace = true, default-features = false, features = ["alloc", "race"] } # only required for no-std
```
brodycj, line 26: minor nit:

```suggestion
once_cell = { workspace = true, default-features = false, features = ["alloc", "race"] } # (only required for no-std)
```

brodycj:
> I don't think we should take a separate PR for this.

👍

> If #2200 has a specific lower bound it needs for once_cell, it should bump the dependency accordingly (and ideally not by requiring the latest version). `Cargo.toml` is used to describe the oldest version that is expected to work.

Unfortunately I cannot tell for sure without some further analysis and testing.

I developed the proposed changes for #2200 assuming that we should be able to use the recent updates to `once_cell`, including some updates related to its use of `portable-atomic` and its `critical-section` option.

I personally think it would be ideal if we could find a way to test with dependencies all the way back to the minimum specified dependency versions, and I would be happy to raise a new issue if needed to discuss this. I do recall that `portable-atomic` is using Depedabot, which seems to update each minimum dependency version along with the updates to Cargo.lock. But yeah, this would probably be out of scope for the time being.

Closing this for now :)

PR #2240, opened 2024-11-25
replace unnecessary use of .map_or() with .is_some_and()

brodycj:
I noticed a warning about this in the clippy nightly output, along with a nice-looking suggestion how to rework this.

---

TODO:

- [x] check CI build status

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1107,7 +1107,7 @@ impl ClientHelloPayload {
     pub(crate) fn check_psk_ext_is_last(&self) -> bool {
         self.extensions
             .last()
-            .map_or(false, |ext| ext.ext_type() == ExtensionType::PreSharedKey)
+            .is_some_and(|ext| ext.ext_type() == ExtensionType::PreSharedKey)
     }
 
     pub(crate) fn psk_modes(&self) -> Option<&[PSKKeyExchangeMode]> {


rustls/src/lib.rs
```diff
@@ -320,6 +320,7 @@
     clippy::std_instead_of_core,
     clippy::use_self,
     clippy::upper_case_acronyms,
+    clippy::unnecessary_map_or,
```
brodycj, line 323: I am surprised this causes issue with Clippy stable as this looks like it was added in Rust 1.75:

- https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or

Am I missing anything here?

ctz: Raised a clippy bug for that -- it looks like the lint was only merged two weeks ago, so it cannot be in 1.75.

brodycj: I think rustversion was already added conditionally for `read_buf` option, was hoping I could use it to check that this warning does not come back. I can remove it, though.

ctz: This is not needed. We deny warnings in stable clippy runs (and don't for nightly clippy, as the noise is not worth it). Therefore, once this lint is in a stable version, regressions of this PR will fail CI.

The list in this file is (or should be) only for clippy lints that do not warn by default (eg, `clippy::use_self` is `allow` by default).

brodycj: I have now removed the clippy updates, rebased to just make the one-line change I would like to propose

cpu (approved):

djc (approved):

ctz (approved):

ctz:
Thanks!

PR #2241, opened 2024-11-26
update minimum serde & test all features w/minimum dependency versions

brodycj:
rationale:

- minimum `serde` version `1.0.103` is required for `cargo minimal-versions --direct --ignore-private test` to work (with or without `--all-features` option), seems to be due to issue between `serde` pre-1.0.103 & test(s) in `rustls/src/crypto/aws_lc_rs/hpke.rs`
- test with minimum dependency versions, all features could help ensure updates like proposed PR #2200 continue working with outdated dependency versions, especially for `once_cell` & `portable-atomic`

Note that while the code DOES seem to build with `serde` back to `1.0.0`, bumping the minimum to `1.0.3` which is still 7 years old seems very reasonable to me.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -346,8 +346,8 @@ jobs:
       - name: Install cargo-hack
         uses: taiki-e/install-action@cargo-hack
 
-      - name: Check direct-minimal-versions
-        run: cargo minimal-versions --direct --ignore-private check
+      - name: Test minimal-versions - direct (all features)
+        run: cargo minimal-versions --direct --ignore-private test --all-features
         working-directory: rustls/
 
   cross:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ rcgen = { version = "0.13", features = ["pem", "aws_lc_rs"], default-features = 
 regex = "1"
 ring = "0.17"
 rsa = { version = "0.9", features = ["sha2"], default-features = false }
-serde = { version = "1", features = ["derive"] }
+serde = { version = "1.0.103", features = ["derive"] }
 serde_json = "1"
 sha2 = { version = "0.10", default-features = false }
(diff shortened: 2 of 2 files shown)


djc (approved):

.github/workflows/build.yml
```diff
@@ -329,7 +329,7 @@ jobs:
           fail_ci_if_error: false
 
   minver:
-    name: Check minimum versions of direct dependencies
+    name: Test minimum versions of direct dependencies (with all features)
```
djc, line 332: I don't think we need to adapt the name here (which is a little tricky with merge queue setup).

brodycj: Disagreed on my part, too bad about the issue with the merge q, reverting this update anyways ...
```suggestion
    name: Check minimum versions of direct dependencies
```

brodycj:
I tried the rebase button, too bad it doesn't offer to squash the updates

brodycj:
Now rebased, with job name no longer updated, please let me know if I need to do anything else for this thanks

cpu (approved):
Thanks

cpu:
> Do we want to test at minimum versions? It seems excessive?

I don't feel strongly. It looks like it increases the job's runtime from ~22s to ~1.5m which isn't nothing, but also dwarfed by other tasks. :shrug:

brodycj:
IMHO ideal if we can test with minimum dependencies that are supported ... I don't know how many other projects are doing this. For example, my work in PR #2200 assumes recent update to `portable-atomic-util`, would be nice to have a way to check that we are not allowing too-old dependency version. I remain hopeful on this, leaving open for discussion 😄

djc:
I don't think the `check` -> `test` transition is really worth it. IMO it's unlikely that code that compiled at an older version does not work at that version when it is known to work in the current version.

Issue #2239, opened 2024-11-25
Please reconsider bumping MSRV to 1.71


tnull:
**Describe the bug**
The recently released version 0.23.18 bumped the MSRV from 1.63 to 1.71 in a patch release. This was done in #2220 with seemingly no strong necessity, IIUC mostly to adopt the `let ... else` syntax?

We are downstream users of `rustls` through a variety of intermediate dependencies and do rely on MSRV 1.63 (building on Debian stable). Shipping this MSRV bump hence breaks our builds. While this is a common issue, we're usually 'okay' with doing some pinning to ensure we can maintain the 1.63 MSRV. 

However, in this case we really, really don't want to pin `rustls` back due to security considerations. Given that the bump happened casually on a patch release means that we have no 'stable' channel that we can keep using (i.e., for which security updates would be provided).

As this doesn't only concern one crate/project, but quite a few projects that depend on or try to maintain an MSRV of 1.63, I'd like to ask you to reconsider the MSRV bump, i.e, the changes in #2220.

As we see it, possible solutions could be, for example:

- Revert #2220, simply stay on MSRV 1.63, at least until the next Debian stable release ships, or
- Intermittently revert #2220, possibly ship it as part of a minor release, and committing to providing backports of security fixes to the 0.23 branch.

algesten:
Reposting here ([original comment](https://github.com/rustls/rustls/pull/2220#issuecomment-2497850346)):

With ureq (also the new ureq3 which is RC 2), I try to maintain a conservative MSRV. This change forces me to fall in line with 1.71, which is unfortunate since I had hoped to stay on 1.65 for now. I know I'm in a minority that considers MSRV bumps borderline breaking change in semver terms.

For context, I'm [fighting an uphill battle](https://github.com/algesten/ureq/issues/877). Someone wanted me to ensure ureq 2.x was compatible with 1.63, and I gave up on that one :)

djc:
We discussed this in Discord last week:

@djc:
> re https://github.com/rustls/rustls/pull/2208#discussion_r1840572154 (and https://github.com/rustls/rustls-native-certs/pull/152), maybe it's time we just take the leap to 1.64 or 1.65?
> The end goal here is to allow this tool to depend on providers in other crates.
> The second commit is a straight rename from rustls-bench/src/bench_impl.rs to rustls-bench/src/main.rs but git&#3...
> I think we'd still be very safely on the conservative side of the ecosystem
> Might even consider going all the way to 1.70
[...]

PR #2243, opened 2024-11-26
Document a MSRV policy

ctz:
For discussion ref #2239 

[Rendered](https://github.com/rustls/rustls/blob/jbp-msrv-policy/SECURITY.md)


SECURITY.md
```diff
@@ -16,9 +16,40 @@ For example, as of 2024-04-18 the latest release is 0.23.4.
...
-policy. For example, while the Rustls 0.20.0 GitHub release note was created
-Jul, 2023 the actual release in `crates.io` was published in Sept. 2021._
+> [!NOTE]
+> We use the date of `crates.io` publication when evaluating the security
+> policy. For example, while the Rustls 0.20.0 GitHub release note was created
+> Jul, 2023 the actual release in `crates.io` was published in Sept. 2021.
+
+### Minimum Supported Rust Version
+
+Our MSRV will be no more recent than 9 versions old, or approximately 12 months.
```
djc, line 26: I'm a little wary about promising more than about 4-6 versions. Note that Tokio promises 6 months (4 versions) but are usually more conservative in practice. In that sense, I think it often pays off to underpromise and overdeliver.

Also note that (a) tokio-rustls represents about 80% of rustls recent downloads and (b) we bump the tokio-rustls MSRV whenever necessary for a Tokio bump so practically our usability (at least without some amount of pinning) is limited by that.

djc (approved):
Nice!

It feels like the MSRV section could use a bit more structure, the current versions feels like a somewhat lose collection of points. Maybe think about how that could be more cohesive? Maybe turn it into a bullet list?

ctz:
> Maybe think about how that could be more cohesive? Maybe turn it into a bullet list?

Restructured into bullets, and put the MSRV of new deps under its own subheading

SECURITY.md
```diff
@@ -16,9 +16,46 @@ For example, as of 2024-04-18 the latest release is 0.23.4.
...
-policy. For example, while the Rustls 0.20.0 GitHub release note was created
-Jul, 2023 the actual release in `crates.io` was published in Sept. 2021._
+> [!NOTE]
+> We use the date of `crates.io` publication when evaluating the security
+> policy. For example, while the Rustls 0.20.0 GitHub release note was created
+> Jul, 2023 the actual release in `crates.io` was published in Sept. 2021.
+
+### Minimum Supported Rust Version
+
+From time to time we will update our MSRV in the course of normal development,
```
cpu, line 26: nit: Likely well known to Rust developers, but can we expand MSRV in the first usage for everyone else?

cpu (approved):
Thanks for writing this down.

SECURITY.md
```diff
@@ -16,9 +16,40 @@ For example, as of 2024-04-18 the latest release is 0.23.4.
...
+
+Once we reach 1.0.0, we will not make MSRV changes in patch releases.
+
+We will not make MSRV changes to security maintenance branches.
+
+> [!NOTE]
+> For the avoidance of doubt: security maintenance branches exist for each
+> release line _that is not the latest_: so (at the time of writing) 0.22
+> and 0.21 have maintenance branches, but 0.23 does not and is released from
+> the `main` branch.
```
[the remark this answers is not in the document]
ctz: Yes, that is correct. To do otherwise would mean MSRV changes are only allowed on a breaking release. (ie, 0.24.0 or 1.0.0 would have to decide the MSRV for that entire release line).

djc:
Maybe further clarifications:

Once we do 1.x, would we do a security release for 1.0.x after we have released 1.1.0?

As for the point about adding optional dependencies, might want to make it explicit that you intend this to be about non-default optional dependencies only. ("default" and "optional" are technically distinct.)

djc (approved):

SECURITY.md
ctz: While toward the end we were feeling the pain, we only recently went from an N-19 MSRV (and N-16 before that, in April of this year), and I don't recall that being exceptionally painful. Though there is obviously an opportunity cost there that is difficult to measure.

So I think N-9 is conservative, but not exceptionally so when compared to our recent MSRV history.

My main concern with N-9 right now is that the ecosystem will move more quickly than usual to take advantage of the 2024 edition (which I think we saw with 1.56).

ctz:
> ships a 1.2.X branch that initially has an MSRV of 1.98 but then switches to an MSRV of rustc 1.99 in 1.2.1

I think we're saying we won't do that:

> "Once we reach 1.0.0, we will not make MSRV changes in patch releases."

This means the 1.2.0 branch would start with a MSRV of 1.99, and not change MSRV thereafter.

Ultimately the MSRV-aware resolver does mean there could be successful dependency solutions that do not include security fixes, if your MSRV does not allow for them. There's tooling already to help with that (cargo audit). The situation today is that you'd get a dependency solution that you cannot build, but does include security fixes.

I'd also say that this policy is a minimum, rather than a maximum; and:

> I think we'll need to revisit the maintenance policy at 1.0.0 to answer this, and also possibly adjust it to take into account (hopefully!) longer windows between breaking changes (eg, at the moment, if we have 1.x.x  which is current for 2 years, then cut 2.0.0, the 1.x.x branch isn't guaranteed any security maintenance at all.)

PR #2246, opened 2024-11-27
performance headlines / landing page

ctz:
This PR adds a little visual display of performance "headlines" to https://rustls.dev/perf/ which looks like:

image

These are just server TLS1.3 handshake performance, plus TLS1.3 AES-256-GCM transfer rates, which are think is a reasonable 

Most of this PR is made up of an extremely "clever" macro, which expands to an SVG I drew in inkscape and sets the dial transforms based on its inputs. I suggest this SVG is treated as write-only :)


djc (approved):
Looks fun!

I think the numerical differences (-47%) are a bit confusing? I think a factor ("1.9x slower") would be easier to grok.

website/templates/reports.html
```diff
@@ -1,6 +1,119 @@
...
+              <h2>{{ page.extra.transfer.recv.boringssl | round }}
+                {{ macros::compare(target=page.extra.transfer.recv.boringssl, base=page.extra.transfer.recv.rustls) }}
+                <footer>BoringSSL</footer>
+              </h2>
+            </div>
+          </div>
+        </div>
+
+      </div>
+      <!-- {{ page.extra.version.rustls }} vs {{ page.extra.version.openssl }} vs {{ page.extra.version.boringssl }} -->
```
cpu, line 107: Should this be left in?

cpu (approved):
Nice :+1: Glad to see some of the functionality of the site generator being helpful sooner than later.

ctz:
> Nice 👍 Glad to see some of the functionality of the site generator being helpful sooner than later.

I think just for autoreload it has been worth it so far

djc:
Could you update the screenshot?

I wonder if we could get this to do deploys for PRs...

ctz:
> Could you update the screenshot?

done

> I wonder if we could get this to do deploys for PRs...

apparently there is a thing called "GitHub Pages preview" but "This feature is only in alpha currently and is not available to the public!" so maybe this will become easy soon?

alternatively we could deploy prs to a throw-away netlify account or similar

djc (approved):

PR #2247, opened 2024-11-27
docs: update linking to aws-lc-rs crate etc.

brodycj:
- add add mono-spaced linking to the `aws-lc-rs` crate in several places in `README.md` & `rustls/src/lib.rs`
- add small clarification of when `aws-lc-rs` provider would be available

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ list of protocol features](https://docs.rs/rustls/latest/rustls/manual/_04_featu
 
 While Rustls itself is platform independent, by default it uses [`aws-lc-rs`] for implementing
 the cryptography in TLS.  See [the aws-lc-rs FAQ][aws-lc-rs-platforms-faq] for more details of the
-platform/architecture support constraints in aws-lc-rs.
+platform/architecture support constraints in [`aws-lc-rs`].
 
 [`ring`] is also available via the `ring` crate feature: see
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ list of protocol features](https://docs.rs/rustls/latest/rustls/manual/_04_featu
 
 While Rustls itself is platform independent, by default it uses [`aws-lc-rs`] for implementing
 the cryptography in TLS.  See [the aws-lc-rs FAQ][aws-lc-rs-platforms-faq] for more details of the
-platform/architecture support constraints in aws-lc-rs.
+platform/architecture support constraints in [`aws-lc-rs`].
```
djc, line 48: I don't think we commonly spell crate names as literals (at least I personally normally don't use code literals for crate names, except when used in the context of actual code).

djc (commented):
This doesn't seem like a material improvement to me.

brodycj:
Yes this is really minor, I thought this could be a small improvement in terms of consistency. I also thought that anything that could mean something to a computer (normal computer, not something that is using AI), like a function name or crate name, would often be shown with monospacing. I rest my case, don't mind if we just close this.

README.md
ctz: Yeah, agree. Probably some instants of not following this rule elsewhere in the crate.

Issue #2227, opened 2024-11-21
Panicking in coalesce_one


kvinwang:
**Checklist**
* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**
I have a Rocket-based web server with mTLS configuration. When I access the server using the Rust reqwest library, it crashes. However, it works fine when using curl as the client. After debugging, I discovered it crashed during the TLS handshake process. The relevant code is shown below.

**To Reproduce**
Compile and run the code below:
```rust
use rustls::server::Acceptor;
use std::collections::VecDeque;
use std::io;

struct ReplayInput {
    chunks: VecDeque<Vec<u8>>,
}
impl io::Read for ReplayInput {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        let Some(chunk) = self.chunks.pop_front() else {
            return Ok(0);
        };
        if buf.len() < chunk.len() {
            return Err(io::Error::new(
                io::ErrorKind::InvalidData,
                "buffer too small",
            ));
        }
        buf[..chunk.len()].copy_from_slice(&chunk);
        Ok(chunk.len())
    }
}

fn main() {
    let chunks = vec![
[...]

PR #2249, opened 2024-11-29
Improve fuzzing coverage and improve processed data tracking

### LINKED ISSUE #2159: Unbuffered API: `DecryptError` is reported by TLS client when server uses large-sized certificates
**Describe the bug**
This issue can be found under the situation:
 - Server uses large certificate that require numbers of TLS messages to be sent to client.
 - One or more TLS messages are divided into multiple data packets.

After some investigations, the issue can be found after the patch set https://github.com/rustls/rustls/pull/2049 was merged. My guess is that this might be a regression issue related to the deframer or TLS message processing logic.

When a previous TLS message has been decrypted but not discarded, and the current TLS message is incomplete, upon receiving the full message, the `process_more_input()` will start back from the *decrypted content* that is not expected.

**To Reproduce**
Unfortunately, I can't reproduce the issue with existing examples because this issue is found under specific network channel (the TLS messages are divided).

**Applicable Version(s)**
v/0.23.13
v/0.23.14

**Additional context**
`UnbufferedServerConnection` and `UnbufferedClientConnection` are used in my case.

ctz:
This PR started as improving fuzzing, but adding working fuzzing of the unbuffered API involved redoing the fix to #2227, and that fixed #2159 at the same time.

The first commit can reach #2227 quite quickly, if cherry picked before that fix.

I will also reinstate the larger corpus repo once this is done.

fixes #2159


djc:
(Already pushed up https://github.com/rustls/rustls-fuzzing-corpora earlier this week.)

ctz:
> (Already pushed up https://github.com/rustls/rustls-fuzzing-corpora earlier this week.)

Ah, sweet, thanks.

fuzz/fuzzers/server.rs
```diff
@@ -6,9 +6,47 @@ extern crate rustls;
 use std::io;
```
cpu, line 6: do the corpus files added in this commit have any interesting provenance?

fuzz/fuzzers/server.rs
```diff
@@ -6,9 +6,47 @@ extern crate rustls;
...
 use std::sync::Arc;
 
-use rustls::server::ResolvesServerCert;
+use rustls::server::{Acceptor, ResolvesServerCert};
 use rustls::{ServerConfig, ServerConnection};
 
+fuzz_target!(|data: &[u8]| {
+    match data.split_first() {
```
cpu, line 13: would this be more effective dispatching on the MSB instead?

fuzz/fuzzers/unbuffered.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::sync::Arc;
+
+use rustls::client::UnbufferedClientConnection;
+use rustls::server::{ResolvesServerCert, UnbufferedServerConnection};
+use rustls::unbuffered::{ConnectionState, UnbufferedStatus};
+use rustls::{ClientConfig, RootCertStore, ServerConfig, SideData};
+
+fuzz_target!(|data: &[u8]| {
+    let mut data = data.to_vec();
+    match data.split_first_mut() {
```
cpu, line 15: similar question here RE: MSB vs first byte.

fuzz/fuzzers/unbuffered.rs
```diff
@@ -0,0 +1,95 @@
...
+        Some((0x01, rest)) => server(rest),
+        Some((_, _)) | None => {}
+    }
+});
+
+fn client(data: &mut [u8]) {
+    let root_store = RootCertStore::empty();
+    let config = ClientConfig::builder()
+        .with_root_certificates(root_store)
+        .with_no_client_auth();
```
cpu, line 26: nit: inline `root_store`?

fuzz/fuzzers/unbuffered.rs
```diff
@@ -0,0 +1,95 @@
...
+    Client(UnbufferedClientConnection),
+    Server(UnbufferedServerConnection),
+}
+
+fn fuzz_unbuffered(mut data: &mut [u8], mut conn: ClientServer) {
+    while !data.is_empty() {
+        let consumed = match &mut conn {
+            ClientServer::Server(server) => {
+                let st = server.process_tls_records(data);
+                process(st)
```
cpu, line 50: nit: inline `st` here and on L53?

fuzz/fuzzers/unbuffered.rs
```diff
@@ -0,0 +1,95 @@
...
+        };
+    }
+}
+
+fn process<S: SideData>(status: UnbufferedStatus<'_, '_, S>) -> Option<usize> {
+    let UnbufferedStatus { discard, state } = status;
+
+    match state {
+        Ok(ConnectionState::EncodeTlsData(mut enc)) => {
+            let mut buffer = [0u8; 16384];
```
cpu, line 72: is this buffer size significant in some way? Maybe the literal could be written `16_384`?

fuzz/fuzzers/unbuffered.rs
```diff
@@ -0,0 +1,95 @@
...
+            enc.encode(&mut buffer).unwrap();
+        }
+        Ok(ConnectionState::TransmitTlsData(xmit)) => xmit.done(),
+        Ok(ConnectionState::WriteTraffic(_)) => return None,
+        Ok(ConnectionState::BlockedHandshake) => return None,
+        Ok(st) => panic!("unhandled state {st:?}"),
+        Err(_err) => return None,
```
cpu, line 79: nit: `Err(_)`?

fuzz/fuzzers/unbuffered.rs
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    Some(discard)
+}
+
+#[derive(Debug)]
+struct Fail;
```
cpu, line 86: nit: more descriptive name? `NoneCertResolver`?

cpu (approved):
Nice, :+1:

fuzz/fuzzers/server.rs
ctz: They are copies of the existing ones, with 0x00 or 0x01 prepended, so they reach `fuzz_buffered_api` and `fuzz_acceptor_api`.

fuzz/fuzzers/server.rs
cpu: This comment falls away with my understanding from [the other thread](https://github.com/rustls/rustls/pull/2249#discussion_r1869969109).

fuzz/fuzzers/unbuffered.rs
cpu: This comment falls away with my understanding from [the other thread](https://github.com/rustls/rustls/pull/2249#discussion_r1869969109).

fuzz/fuzzers/unbuffered.rs
ctz: fixed; should actually be `16_384 + 5`. Maybe in the future we could export this as a const.

PR #2256, opened 2024-12-04
Optimize HMAC key preparation for TLS1.2 PRF

ctz:
(benchmark check)


djc:
> So this was effective, but is not applicable to aws-lc-rs, and (I think) is best done during a breaking change to avoid the bifurcation. So will keep this in my back pocket for later.

Breaking change issue maybe?

PR #2259, opened 2024-12-05
Restructure hybrid key exchange implementation

ctz:
This PR incorporates and builds on @dconnolly's work in #2228 to factor out the "hybrid" part of rustls-post-quantum. That's in preparation to eventually support secp256r1MLKEM768 for FIPS builds -- when we want that, it will be just:

```rust
pub static secp256r1MLKEM768: &dyn SupportedKxGroup = &hybrid::Hybrid {
    classical: kx_group::SECP256R1,
    post_quantum: MLKEM768,
    name: NamedGroup::secp256r1MLKEM768,
    layout: hybrid::Layout {
        classical_share_len: 32,
        post_quantum_client_share_len: MLKEM768_ENCAP_LEN,
        post_quantum_server_share_len: MLKEM768_CIPHERTEXT_LEN,
        post_quantum_first: false,
    },
};
```

(Haven't done that yet, since I can't find any other implementations to test with.)

Along the way, I've added the new MLKEM `NamedGroup` codepoints to the core crate.

fixes #2228

### DIFF
--- a/rustls-post-quantum/Cargo.lock
+++ b/rustls-post-quantum/Cargo.lock
@@ -441,7 +441,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.19"
+version = "0.23.20"
 dependencies = [
  "aws-lc-rs",
  "log",
--- a/rustls-post-quantum/Cargo.lock
+++ b/rustls-post-quantum/Cargo.lock
@@ -460,7 +460,7 @@ checksum = "16f1201b3c9a7ee8039bcadc17b7e605e2945b27eee7631788c1bd2b0643674b"
 
 [[package]]
 name = "rustls-post-quantum"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "aws-lc-rs",
  "env_logger",
--- a/rustls-post-quantum/src/hybrid.rs
+++ b/rustls-post-quantum/src/hybrid.rs
@@ -1,0 +1,175 @@
+use rustls::crypto::{ActiveKeyExchange, CompletedKeyExchange, SharedSecret, SupportedKxGroup};
+use rustls::ffdhe_groups::FfdheGroup;
+use rustls::{Error, NamedGroup, ProtocolVersion};
+
+use crate::INVALID_KEY_SHARE;
+
+/// A generalization of hybrid key exchange.
+#[derive(Debug)]
+pub(crate) struct Hybrid {
+    pub(crate) classical: &'static dyn SupportedKxGroup,
+    pub(crate) post_quantum: &'static dyn SupportedKxGroup,
+    pub(crate) name: NamedGroup,
+    pub(crate) layout: Layout,
+}
+
+impl SupportedKxGroup for Hybrid {
+    fn start(&self) -> Result<Box<dyn ActiveKeyExchange>, Error> {
+        let classical = self.classical.start()?;
+        let post_quantum = self.post_quantum.start()?;
+
+        let combined_pub_key = self
+            .layout
+            .concat(post_quantum.pub_key(), classical.pub_key());
+
+        Ok(Box::new(ActiveHybrid {
+            classical,
+            post_quantum,
+            name: self.name,
+            layout: self.layout,
+            combined_pub_key,
+        }))
+    }
+
+    fn start_and_complete(&self, client_share: &[u8]) -> Result<CompletedKeyExchange, Error> {
+        let (post_quantum_share, classical_share) = self
+            .layout
+            .split_received_client_share(client_share)
+            .ok_or(INVALID_KEY_SHARE)?;
+
+        let cl = self
+            .classical
+            .start_and_complete(classical_share)?;
+        let pq = self
+            .post_quantum
+            .start_and_complete(post_quantum_share)?;
+
+        let combined_pub_key = self
+            .layout
+            .concat(&pq.pub_key, &cl.pub_key);
+        let secret = self
+            .layout
+            .concat(pq.secret.secret_bytes(), cl.secret.secret_bytes());
+
+        Ok(CompletedKeyExchange {
+            group: self.name,
+            pub_key: combined_pub_key,
+            secret: SharedSecret::from(secret),
+        })
+    }
+
+    fn ffdhe_group(&self) -> Option<FfdheGroup<'static>> {
+        None
+    }
+
+    fn name(&self) -> NamedGroup {
+        self.name
+    }
+
+    fn usable_for_version(&self, version: ProtocolVersion) -> bool {
+        version == ProtocolVersion::TLSv1_3
+    }
+}
+
+struct ActiveHybrid {
+    classical: Box<dyn ActiveKeyExchange>,
+    post_quantum: Box<dyn ActiveKeyExchange>,
+    name: NamedGroup,
+    layout: Layout,
+    combined_pub_key: Vec<u8>,
+}
+
+impl ActiveKeyExchange for ActiveHybrid {
+    fn complete(self: Box<Self>, peer_pub_key: &[u8]) -> Result<SharedSecret, Error> {
+        let (post_quantum_share, classical_share) = self
+            .layout
+            .split_received_server_share(peer_pub_key)
+            .ok_or(INVALID_KEY_SHARE)?;
+
+        let cl = self
+            .classical
+            .complete(classical_share)?;
+        let pq = self
+            .post_quantum
+            .complete(post_quantum_share)?;
+
+        let secret = self
+            .layout
+            .concat(pq.secret_bytes(), cl.secret_bytes());
+        Ok(SharedSecret::from(secret))
+    }
+
+    /// Allow the classical computation to be offered and selected separately.
+    fn hybrid_component(&self) -> Option<(NamedGroup, &[u8])> {
+        Some((self.classical.group(), self.classical.pub_key()))
+    }
+
+    fn complete_hybrid_component(
+        self: Box<Self>,
+        peer_pub_key: &[u8],
+    ) -> Result<SharedSecret, Error> {
+        self.classical.complete(peer_pub_key)
+    }
+
+    fn pub_key(&self) -> &[u8] {
+        &self.combined_pub_key
+    }
+
+    fn ffdhe_group(&self) -> Option<FfdheGroup<'static>> {
+        None
+    }
+
+    fn group(&self) -> NamedGroup {
+        self.name
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Layout {
+    /// Length of classical key share.
+    pub(crate) classical_share_len: usize,
+
+    /// Length of post-quantum key share sent by client
+    pub(crate) post_quantum_client_share_len: usize,
+
+    /// Length of post-quantum key share sent by server
+    pub(crate) post_quantum_server_share_len: usize,
+
+    /// Whether the post-quantum element comes first in shares and secrets.
+    ///
+    /// For dismal and unprincipled reasons, SECP256R1MLKEM768 has the
+    /// classical element first, while X25519MLKEM768 has it second.
+    pub(crate) post_quantum_first: bool,
+}
+
+impl Layout {
+    fn split_received_client_share<'a>(&self, share: &'a [u8]) -> Option<(&'a [u8], &'a [u8])> {
+        self.split(share, self.post_quantum_client_share_len)
+    }
+
+    fn split_received_server_share<'a>(&self, share: &'a [u8]) -> Option<(&'a [u8], &'a [u8])> {
+        self.split(share, self.post_quantum_server_share_len)
+    }
+
+    fn split<'a>(
+        &self,
+        share: &'a [u8],
+        post_quantum_share_len: usize,
+    ) -> Option<(&'a [u8], &'a [u8])> {
+        if share.len() != self.classical_share_len + post_quantum_share_len {
+            return None;
+        }
+
+        Some(match self.post_quantum_first {
+            true => share.split_at(post_quantum_share_len),
+            false => share.split_at(self.classical_share_len),
+        })
+    }
+
+    fn concat(&self, post_quantum: &[u8], classical: &[u8]) -> Vec<u8> {
+        match self.post_quantum_first {
+            true => [post_quantum, classical].concat(),
+            false => [classical, post_quantum].concat(),
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


rustls-post-quantum/src/lib.rs
```diff
@@ -59,12 +61,21 @@ use rustls::crypto::{
 use rustls::ffdhe_groups::FfdheGroup;
 use rustls::{Error, NamedGroup, PeerMisbehaved, ProtocolVersion};
 
-/// A `CryptoProvider` which includes `X25519MLKEM768` key exchange.
+mod mlkem;
```
cpu, line 60: I thought our pref'd ordering was `mod` -> `pub use` -> `use` but I don't think it's something made explicit in the [ordering](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#ordering) section of the contrib guide.

If my gut check on the expected order is correct I can add some text about it. WDYT?

rustls-post-quantum/src/lib.rs
```diff
@@ -59,12 +61,21 @@ use rustls::crypto::{
...
+/// A `CryptoProvider` which includes `X25519MLKEM768` and `MLKEM768`
+/// key exchanges.
 pub fn provider() -> CryptoProvider {
     let mut parent = default_provider();
+
     parent
         .kx_groups
         .insert(0, &X25519MLKEM768);
+
+    parent.kx_groups.insert(1, &MLKEM768);
```
cpu: feels a bit tidier as:
```rust
    parent
        .kx_groups
        .splice(0..0, [X25519MLKEM768, MLKEM768]);
```

rustls-post-quantum/src/mlkem.rs
```diff
@@ -0,0 +1,96 @@
...
+impl SupportedKxGroup for MLKEM768 {
+    fn start(&self) -> Result<Box<dyn ActiveKeyExchange>, Error> {
+        let decaps_key = kem::DecapsulationKey::generate(&ML_KEM_768)
+            .map_err(|_| Error::FailedToGetRandomBytes)?;
+
+        let encaps_key = decaps_key
+            .encapsulation_key()
+            .map_err(|_| Error::FailedToGetRandomBytes)?;
+
+        let pub_key_bytes = encaps_key.key_bytes().unwrap();
```
cpu, line 22: This is documented as returning an unspecified error for "Any failure to retrieve the EnscapsulationKey bytes.". It's not immediate obvious to me what sort of condition could cause that failure.

Can you add a justification comment for why the panic is unreachable, or should it be mapped to `Error::Other` or `Error::General`?

rustls-post-quantum/src/mlkem.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// [MLKEM]: https://datatracker.ietf.org/doc/draft-connolly-tls-mlkem-key-agreement
+#[allow(clippy::upper_case_acronyms)]
+#[derive(Debug)]
+pub struct MLKEM768;
+
+impl SupportedKxGroup for MLKEM768 {
+    fn start(&self) -> Result<Box<dyn ActiveKeyExchange>, Error> {
+        let decaps_key = kem::DecapsulationKey::generate(&ML_KEM_768)
+            .map_err(|_| Error::FailedToGetRandomBytes)?;
```
cpu, line 16: both `kem::DecapsulationKey::generate()` and `DecapsulationKey::encapsulation_key()` are doc'd as returning an error "when operation fails due to internal error."

Is `FailedToGetRandomBytes` the best choice for these error conditions?

rustls-post-quantum/Cargo.lock
```diff
@@ -441,7 +441,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.19"
+version = "0.23.20"
```
cpu, line 444: Based on the commit message I would have expected this commit to also update to `Cargo.toml` to change:
```toml
rustls = { version = "0.23.2", features = ["aws_lc_rs"], path = "../rustls" }
```
to
```toml
rustls = { version = "0.23.20", features = ["aws_lc_rs"], path = "../rustls" }
```

Am I missing something?

rustls-post-quantum/src/hybrid.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Length of post-quantum key share sent by client
+    pub(crate) post_quantum_client_share_len: usize,
+
+    /// Length of post-quantum key share sent by server
+    pub(crate) post_quantum_server_share_len: usize,
+
+    /// Whether the post-quantum element comes first in shares and secrets.
+    ///
+    /// For dismal and unprincipled reasons, SECP256R1MLKEM768 has the
+    /// classical element first, while X25519MLKEM768 has it second.
```
cpu, line 145: _sigh_ :sweat:

rustls-post-quantum/src/hybrid.rs
```diff
@@ -0,0 +1,178 @@
+use rustls::crypto::{ActiveKeyExchange, CompletedKeyExchange, SharedSecret, SupportedKxGroup};
+use rustls::ffdhe_groups::FfdheGroup;
+use rustls::{Error, NamedGroup, ProtocolVersion};
+
+use crate::INVALID_KEY_SHARE;
+
+/// A generalization of hybrid key exchange.
```
cpu, line 7: this is a nice refactor :+1:

cpu (approved):

rustls/src/msgs/enums.rs
```diff
@@ -223,6 +223,11 @@ enum_builder! {
         FFDHE4096 => 0x0102,
         FFDHE6144 => 0x0103,
         FFDHE8192 => 0x0104,
+        MLKEM512 => 0x0200,
+        MLKEM768 => 0x0201,
+        MLKEM1024 => 0x0202,
```
dconnolly, line 228: Besides 768, the other two are not offered yet, should we comment them out?

ctz: Having these here is ~free, and improves logging output if a peer ever mentions them.

rustls-post-quantum/Cargo.lock
ctz: Yikes. Good catch!

rustls-post-quantum/src/lib.rs
djc: Not sure what you're objecting too here? This order looks good to me, although I would not have the empty line between the `mod` declaration and any imports from that module.

I think my preferred order is (1) imports from other crates and other modules, (2) module declarations, (3) keep imports from locally-declared modules near their declarations.

rustls-post-quantum/src/lib.rs
djc: Agreed, though I think we can drop the left `0` in `0..0`.

rustls-post-quantum/src/mlkem.rs
```diff
@@ -0,0 +1,96 @@
...
+use rustls::crypto::{ActiveKeyExchange, CompletedKeyExchange, SharedSecret, SupportedKxGroup};
+use rustls::ffdhe_groups::FfdheGroup;
+use rustls::{Error, NamedGroup, PeerMisbehaved, ProtocolVersion};
+
+/// This is the [MLKEM] key exchange.
+///
+/// [MLKEM]: https://datatracker.ietf.org/doc/draft-connolly-tls-mlkem-key-agreement
+#[allow(clippy::upper_case_acronyms)]
+#[derive(Debug)]
+pub struct MLKEM768;
```
djc: Why not spell it as `MlKem768`?

rustls-post-quantum/src/mlkem.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        Ok(Box::new(Active {
+            decaps_key: Box::new(decaps_key),
+            encaps_key_bytes: Vec::from(pub_key_bytes.as_ref()),
+        }))
+    }
+
+    fn start_and_complete(&self, client_share: &[u8]) -> Result<CompletedKeyExchange, Error> {
+        let (ciphertext, shared_secret) = kem::EncapsulationKey::new(&ML_KEM_768, client_share)
+            .map_err(|_| INVALID_KEY_SHARE)
```
djc, line 32: Suggest moving the `?` here and writing the `and_then()` as linear code instead of inside a combinator.

rustls-post-quantum/src/lib.rs
```diff
@@ -245,8 +245,6 @@ impl CombinedShare {
     }
 }
 
-const NAMED_GROUP: NamedGroup = NamedGroup::Unknown(0x11ec);
```
djc, line 237: Maybe reorder the commits so that this commit falls away?

rustls-post-quantum/src/hybrid.rs
```diff
@@ -0,0 +1,178 @@
...
+        }))
+    }
+
+    fn start_and_complete(&self, client_share: &[u8]) -> Result<CompletedKeyExchange, Error> {
+        let Some((post_quantum_share, classical_share)) = self
+            .layout
+            .split_received_client_share(client_share)
+        else {
+            return Err(INVALID_KEY_SHARE);
+        };
```
djc, line 40: Because I don't like the multi-line scrutinee in the let-else, suggest using `ok_or(INVALID_KEY_SHARE)?` instead (also seems generally simpler).

rustls-post-quantum/src/hybrid.rs
```diff
@@ -0,0 +1,178 @@
...
+    combined_pub_key: Vec<u8>,
+}
+
+impl ActiveKeyExchange for ActiveHybrid {
+    fn complete(self: Box<Self>, peer_pub_key: &[u8]) -> Result<SharedSecret, Error> {
+        let Some((post_quantum_share, classical_share)) = self
+            .layout
+            .split_received_server_share(peer_pub_key)
+        else {
+            return Err(INVALID_KEY_SHARE);
```
djc, line 90: And here.

rustls-post-quantum/src/hybrid.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn split_received_server_share<'a>(&self, share: &'a [u8]) -> Option<(&'a [u8], &'a [u8])> {
+        if share.len() != self.classical_share_len + self.post_quantum_server_share_len {
+            return None;
+        }
+
+        Some(match self.post_quantum_first {
+            true => share.split_at(self.post_quantum_server_share_len),
+            false => share.split_at(self.classical_share_len),
+        })
+    }
```
djc, line 170: Looks like we could deduplicate this logic by extracting `self.post_quantum_{client,server}_share_len`?

rustls-post-quantum/src/lib.rs
```diff
@@ -43,214 +43,59 @@
...
-        out.0[..MLKEM768_CIPHERTEXT_LEN].copy_from_slice(ml_kem.as_ref());
-        out.0[MLKEM768_CIPHERTEXT_LEN..].copy_from_slice(x25519);
-        out
-    }
-}
+/// This is the [MLKEM] key exchange.
+///
+/// [MLKEM]: https://datatracker.ietf.org/doc/draft-connolly-tls-mlkem-key-agreement
+#[allow(clippy::upper_case_acronyms)]
+pub static MLKEM768: &dyn SupportedKxGroup = &mlkem::MLKEM768;
```
djc, line 95: Maybe clashes between spec-native naming and Rust naming, but I feel like this would more idiomatically named `ML_KEM_768`.

rustls-post-quantum/src/lib.rs
```diff
@@ -43,214 +43,59 @@
...
-    fn combine(x25519: SharedSecret, ml_kem: kem::SharedSecret) -> Self {
-        let mut out = CombinedSecret([0u8; COMBINED_SHARED_SECRET_LEN]);
-        out.0[..MLKEM768_SECRET_LEN].copy_from_slice(ml_kem.as_ref());
-        out.0[MLKEM768_SECRET_LEN..].copy_from_slice(x25519.secret_bytes());
-        out
-    }
-}
-
-struct CombinedShare(Vec<u8>);
+pub static X25519MLKEM768: &dyn SupportedKxGroup = &hybrid::Hybrid {
```
djc, line 79: Consider changing the order of commits to avoid the intra-PR changes? Can always add Deirdre as a co-author on your commits.

djc (approved):

rustls-post-quantum/src/lib.rs
ctz: Trying to keep these named the same as the IANA names, albeit uppercased to meet the rust static/const naming convention (that doesn't affect this one, but it's why `SECP256R1` is very shouty compared to the IANA `secp256r1`).

ctz: (realised this `clippy::upper_case_acryonyms` is not needed here; will remove)

djc: > Trying to keep these named the same as the IANA names, albeit uppercased to meet the rust static/const naming convention (that doesn't affect this one, but it's why `SECP256R1` is very shouty compared to the IANA `secp256r1`).

I guess that makes sense, though at the same time, I'll argue the other side: it feels to me like the Rust naming convention will be more strongly activated in the context of writing Rust code and there's little chance of confusion between `MLKEM768` and `ML_KEM_768` -- it's even likely there might be autocomplete to help out.

rustls-post-quantum/src/lib.rs
cpu: > Not sure what you're objecting too here?

This is ordering `use` imports, then declares a `mod`, then adds `pub use` for items from the mod. As mentioned I thought it should be declare `mod`, export items from mod `pub use`, import stuff with `use`.

> I think my preferred order is (1) imports from other crates and other modules, (2) module declarations, (3) keep imports from locally-declared modules near their declarations.

@ctz Is that your preference too? If so I think it should be put into CONTRIBUTING.md.

I don't feel strongly except that it shouldn't be ambiguous per-PR.

ctz: I don't have a strong opinion in this area. It kind of makes sense to me that the order would go 1) private `use`s, 2) `mod`s, 3) `pub use`s. In practice I think the number of files with all three should be pretty low -- I'd prefer if files that have a ton of `mod`s don't also have non-trivial code in.

rustls-post-quantum/src/lib.rs
ctz: Put these two commits into one.

rustls-post-quantum/src/lib.rs
cpu: I'll make a CONTRIB update for this. Thanks!

PR #2261, opened 2024-12-06
docs: reduce permission scope

cpu:
Only the `deploy` task needs the "pages/id-token:write" permissions. Let's try to limit the potential for badness by only giving that specific job step elevated permission instead of doing it at the workflow level.

This was [flagged](https://woodruffw.github.io/zizmor/audits/#remediation_2) by running Zizmor on the `.github/workflows` dir with Zizmor 0.7.0 in offline mode and seems like a sensible idea.

If you run this tool in **online** mode w/ a `GH_TOKEN` it also flags several [ref-confusion](https://woodruffw.github.io/zizmor/audits/#ref-confusion) findings for our usage of [install-action](https://github.com/taiki-e/install-action). I'm not 100% sure why it didn't flag the other instances where we specify dependent actions by tag (e.g. `checkout@v4`), but the fix in either case is to [pin our actions by SHA hash](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash). Since that's a bit more invasive I haven't made that change yet pending some discussion on whether it's something folks are on-board for.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,8 +2,6 @@ name: documentation
 
 permissions:
   contents: read
-  pages: write
-  id-token: write
 
 on:
   workflow_dispatch:
@@ -85,6 +83,9 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'rustls/rustls'
     needs: generate
+    permissions:
+      pages: write
+      id-token: write
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
@@ -92,4 +93,3 @@ jobs:
       - name: Deploy to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@v4
-


djc (approved):

cpu:
> My thoughts:: I'd prefer to concentrate on minimising any and all privileges for GHA (like this PR does), rather than try and affix on third party steps.

That aligns with my views. I appreciate that our CI is "boring". I won't bother with any follow-ups to switch to SHA pins.

> (This is also why I don't personally care for release-plz and similar, giving GHA rights to publish your crate feels like the first step on a very painful journey.)

I'm warmer to the idea depending on specific context but for a crate like Rustls it feels like the wrong tradeoff :+1:

djc:
> My thoughts:: I'd prefer to concentrate on minimising any and all privileges for GHA (like this PR does), rather than try and affix on third party steps. I don't think it's feasible to review those, so we should treat them as untrustworthy and pre-compromised.

Agreed that reviewing actions changes in any amount of depth is not feasible, so keeping scope minimal seems like the right direction to me.

> (This is also why I don't personally care for release-plz and similar, giving GHA rights to publish your crate feels like the first step on a very painful journey.)

There will be first-class support for [trusted publishing](https://github.com/rust-lang/rfcs/pull/3691) soonish so that might be a more interesting avenue to explore.

A5rocks:
> If you run this tool in **online** mode w/ a `GH_TOKEN` it also flags several [ref-confusion](https://woodruffw.github.io/zizmor/audits/#ref-confusion) findings for our usage of [install-action](https://github.com/taiki-e/install-action). I'm not 100% sure why it didn't flag the other instances where we specify dependent actions by tag (e.g. `checkout@v4`), but the fix in either case is to [pin our actions by SHA hash](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash). Since that's a bit more invasive I haven't made that change yet pending some discussion on whether it's something folks are on-board for.

I think this specific error code is because `taiki-e/install-action` has both a `v2` tag and a `v2` branch, so I suppose it's not immediately clear which one it uses in case they differ? Anyways that explains why `actions/checkout@v4` isn't flagged (no `v4` branch) and probably the solution would just be to tell `taiki-e/install-action` to only have one of either a tag or a branch...?

PR #2263, opened 2024-12-09
openssl-tests: install openssl from source

ctz:
At the moment CI is broken because pixi's latest release is broken (already reported to them by others).

I looked for a PPA, but didn't find anything suitable.

However a combination of `actions/cache` and 5 lines of shell to fetch and build seems like an acceptable solution to me, and is just as fast as pixi.

This PR also bumps the openssl version to 3.4.0, but that is incidental.


djc (approved):
Nice!

cpu (approved):
Thanks! IMO this is nicer than a 3rd party action (especially if it's going to break from under us!)

PR #2264, opened 2024-12-09
docs: update CONTRIBUTING.md to mention `mod` order

cpu:
Consensus from https://github.com/rustls/rustls/pull/2259#discussion_r1871801247 is that we prefer to order:

1. private `use`'s
2. `mod`'s
3. `pub use`'s

Also note that if there are a lot of `mod`'s (e.g. for a `lib.rs`) there should probably not be any non-trivial code alongside.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,6 +109,10 @@ single user, or just a few co-located users).
 
 The `#[cfg(test)] mod tests {}` module goes on the very bottom, if present.
 
+When defining new module definitions with `mod` they should be placed after
+private `use`'s, and before `pub use`'s. Files that have many `mod`'s should aim
+to avoid non-trivial code alongside the module definitions.
+
 #### Ordering for a given type
 
 For a given type, we prefer to order items as follows:


dconnolly (approved):

djc (approved):

CONTRIBUTING.md
```diff
@@ -109,6 +109,10 @@ single user, or just a few co-located users).
 
 The `#[cfg(test)] mod tests {}` module goes on the very bottom, if present.
 
+When defining new module definitions with `mod` they should be placed after
```
djc, line 112: I think the way I usually do it is a little different, which is to keep private `use`s together with public `use`s (since I think it's confusing to separate imports from the same source in different locations). For imports from a locally-defined module, those should be grouped with the module (whether declaration or definition).

ctz: > For imports from a locally-defined module, those should be grouped with the module (whether declaration or definition).

Yeah, agree. Here's an example of that happening: https://github.com/rustls/rustls/blob/main/rustls/src/conn.rs#L351-L352

cpu: Do either of you want to suggest an edit? I'm not sure the best way to translate this guidance into text.

djc: Figured this might be easiest as a separate PR: https://github.com/rustls/rustls/pull/2271.

Issue #2258, opened 2024-12-04
[Feature Request] Expose selected certificate types in Cert Verify callbacks


ripatel-fd:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**

rustls lacks support for multiple certificate types (you can only have one active certificate type per TLS config).
This is problematic for existing deployments of TLS that want to migrate certificate types, e.g. X.509 to Raw Public Keys (particularly P2P networks, where these migrations take a long time)

**Describe the solution you'd like**

The main problem is that certain callbacks (via public traits) don't provide the certificate types that were negotiated by the ClientHello/ServerHello exchange.

A practical way to allow multiple certificate types is by extending the following callbacks with additional parameters:
- ResolvesClientCert::resolve should provide the selected `client_certificate_type`. Currently, the callback has no way of knowing which certificate type the server would expect
- ServerCertVerifier::verify_tls13_signature should provide the selected `server_certificate_type`. 
- ClientCertVerifier::verify_tls13_signature should provide the selected `client_certificate_type`. 

If these 3 changes are made, it's possible to create a TLS config with a custom ResolvesClientCert and cert verifier that supports RawPublicKeys with X.509 fallback.

**Describe alternatives you've considered**

-

**Additional context**

-

ctz:
https://github.com/rustls/rustls/issues/423#issuecomment-1908082743 is background on the limited implementation of RFC7250 we have. 

>  for existing deployments of TLS that want to migrate certificate types

Raw public keys are not a type of certificate (except in the limited & misleading setting of TLS type names). It seems pretty sketchy to have one service that supports both, since they have such disjoint requirements and security properties.

> ServerCertVerifier::verify_tls13_signature should provide the selected server_certificate_type.
> ClientCertVerifier::verify_tls13_signature should provide the selected client_certificate_type.

In the implementation we have, there is a one-to-one mapping between RPK/certificate requirement and a given  `ServerCertVerifier`/`ClientCertVerifier` implementation. In other words, this is already known.

Undoing that, I think, would force every rustls user to think about this. I think that would apply an API complexity cost on the 99.99% of users who have no use for RPK.

ripatel-fd:
> Raw public keys are not a type of certificate (except in the limited & misleading setting of TLS type names). It seems pretty sketchy to have one service that supports both, since they have such disjoint requirements and security properties.

[...]

PR #2266, opened 2024-12-10
[DRAFT] Expose selected certificate types in Cert Verify callbacks 

dignifiedquire:
Working on a design to solve #2258 

Currently the design is based on the latest suggestion from @djc in #2258 

PKI Types branch: https://github.com/dignifiedquire/pki-types

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2326,8 +2326,7 @@ dependencies = [
 [[package]]
 name = "rustls-pki-types"
 version = "1.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "16f1201b3c9a7ee8039bcadc17b7e605e2945b27eee7631788c1bd2b0643674b"
+source = "git+https://github.com/dignifiedquire/pki-types?branch=feat-identity-der#7a278d28f272a76d10361c110b91d58a992fe610"
 
 [[package]]
 name = "rustls-post-quantum"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,3 +93,7 @@ zlib-rs = "0.4"
 [profile.bench]
 codegen-units = 1
 lto = true
+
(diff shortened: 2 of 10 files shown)


dignifiedquire:
@djc good idea, added the appropriate methods, I think this should work asfaict. They will need some naming work, so for now I just prefixed them with `yeet_` to make clear that these need better naems

ctz:
I would kind of expect `IdentityDer` to contain a certificate chain, not just an end-entity cert? That would make the invalid state of RPK + "intermediate" X509 certs unrepresentable at the level of the verifier traits.

dignifiedquire:
> I would kind of expect `IdentityDer` to contain a certificate chain, not just an end-entity cert?

That makes sense, but if we do this, how would we use this type in `(yeet_)verify_tls13_signature`, where only the `CertifcateDer` is passed in? would the chain just be left empty always?

dignifiedquire:
Any more thoughts on if I should fill this out with logic, or do you think this is not realistic to merge this?

cpu:
> Any more thoughts on if I should fill this out with logic, or do you think this is not realistic to merge this?

I'm fairly ambivalent personally but would be interested in what the other maintainers are thinking. 

From my side I think it's introducing complexity that seems to only support a use case had by users in one particular niche that evaluated TLS for their use case, found a traditional PKI a bad fit, and so deployed a hack that now complicates migration to a better solution. I'm sympathetic since the better solution wasn't available at the time, but I'm also not eager to complicate Rustls to support what seems like an improvement with a limited shelf-life (AIUI eventually all mixed deployments will be RPK only and new deployments should be RPK only from the jump).

ctz:
I think we have a preference on the shape that was proposed in #2560 and will bring forward some changes along those lines (see, eg groundwork in #2614)

PR #2267, opened 2024-12-10
Improve fuzzing coverage further

ctz:
Before this PR the client and server fuzzers were pretty limited. Both of them suffered from using real cryptography: fuzzers (obviously) cannot make progress in the face of this, and perform badly if the fuzzed program is non-deterministic, so they would never be able to complete a handshake. The server one was even worse than that: it didn't have any server auth key, so  the handshake would be instantly aborted after client hello handling.

This PR introduces a fuzzing-only CryptoProvider that is completely deterministic, and doesn't do any cryptography. The theory of this is fuzzing should be able to access more of the handshake. The corpus for these now contains a whole handshake, so the fuzzers start out in a more advanced position.

To quantify that, here are measurements of coverage in `src/server` or `src/client` using the base corpus (that is, files that are checked into the repo.)

Before this PR: 
- client: 22.02%
- server: 9.53%

After this PR:
- client: 40.98%
- server: 45.41%

After this PR, with 5 minutes fuzzing:
- client: 45.75%
- server: 52.48%

There's still future work to do here: no support for QUIC, client auth, etc.


rustls-fuzzing-provider/tests/smoke.rs
```diff
@@ -0,0 +1,138 @@
+use std::fs;
+use std::io::Write;
+
+use rustls::version::{TLS12, TLS13};
+use rustls::*;
```
djc, line 5: I'd prefer to avoid star imports even in the tests of the fuzzing utility.

djc (commented):
Very cool!

Should we also consider digging into Deterministic Simulation Testing, maybe with Antithesis? (See https://turso.tech/blog/introducing-limbo-a-complete-rewrite-of-sqlite-in-rust#can-we-match-sqlite-s-world-famous-reliability-.)

rustls-fuzzing-provider/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+        hash_provider: &Hash,
+        confidentiality_limit: u64::MAX,
+    },
+    kx: KeyExchangeAlgorithm::ECDHE,
+    sign: &[SIGNATURE_SCHEME],
+    prf_provider: &tls12::PrfUsingHmac(&Hmac),
+    aead_alg: &Aead,
+});
+
+use rustls::crypto::hash;
```
djc, line 113: Nit: move this to the top?

rustls-fuzzing-provider/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+}
+
+const HMAC_OUTPUT: &[u8] = b"HmacHmacHmacHmacHmacHmacHmacHmac";
+
+struct ActiveKeyExchange;
+
+impl crypto::ActiveKeyExchange for ActiveKeyExchange {
+    fn complete(self: Box<Self>, peer: &[u8]) -> Result<crypto::SharedSecret, Error> {
+        if peer == KX_PEER_SHARE {
```
djc, line 185: `match`?

rustls-fuzzing-provider/tests/smoke.rs
```diff
@@ -7,15 +8,69 @@ use rustls::{
...
+// Transcripts are written into the fuzzing corpus.
 
 #[test]
 fn pairwise_tls12() {
-    test_version(&TLS12);
+    let transcript = test_version(&TLS12);
+
+    fs::write(
+        "../fuzz/corpus/unbuffered/tls12-server.bin",
+        [&[0u8], &transcript.server_wrote[..]].concat(),
```
djc, line 20: Nit: are the `[..]` suffixes necessary for these?

djc (approved):

rustls-fuzzing-provider/tests/smoke.rs
djc: I prefer the punctuation for its conciseness over the words.

cpu (approved):
Very nice :+1:

cpu:
> Should we also consider digging into Deterministic Simulation Testing, maybe with Antithesis? (See https://turso.tech/blog/introducing-limbo-a-complete-rewrite-of-sqlite-in-rust#can-we-match-sqlite-s-world-famous-reliability-.)

I was looking at Antithesis last night. It seems really cool but AFAICT it's proprietary & only available to paying customers?

ctz:
I think this gets us closer to being able to do model checking and/or deterministic simulation testing, thanks to removing a large chunk of non-determinism. I think we'd also need to control for time, but we have the interface to do that now. However, I think the model/simulator would be pretty complex if it covered the whole library's APIs. But we could start with a simpler model, such as "data written into one side's `io::Write` appears from the other side's `io::Read`" which could be modeled/simulated with the cursor interface to `Vec<u8>`.

PR #2269, opened 2024-12-10
fix rustls-provider-example hpke no-std support

brodycj:
(with some very minor refactoring)

---

should be finally ready now ... apologies for so much noise ... looks like codecov is complaining about 2 lines of code for `no-std` since I don't think we can unit-test this; I don't think this should be an issue since CI does already verify that we can compile this for `no-std`

### DIFF
--- a/provider-example/src/hpke.rs
+++ b/provider-example/src/hpke.rs
@@ -1,8 +1,6 @@
 use alloc::boxed::Box;
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
-use std::error::Error as StdError;
 
 use hpke_rs_crypto::types::{AeadAlgorithm, KdfAlgorithm, KemAlgorithm};
 use hpke_rs_crypto::HpkeCrypto;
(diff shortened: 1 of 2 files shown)


provider-example/src/hpke.rs
```diff
@@ -209,18 +207,35 @@ impl HpkeOpener for HpkeRsReceiver {
...
+    }
 }
 
 #[cfg(not(feature = "std"))]
-fn other_err(err: impl Send + Sync + 'static) -> Error {
-    Error::General(alloc::format!("{}", err));
+mod error_util {
+    use core::fmt::Display;
+    use rustls::Error;
+    use rustls::OtherError;
```
brodycj, line 233:
```suggestion
```

provider-example/src/hpke.rs
```diff
@@ -209,18 +207,35 @@ impl HpkeOpener for HpkeRsReceiver {
...
+
+    #[cfg(not(feature = "std"))]
+    pub(crate) fn wrap_err(err: impl Display + Send + Sync + 'static) -> Error {
```
brodycj, line 236:
```suggestion
    pub(crate) fn wrap_err(err: impl Display) -> Error {
```

brodycj (commented):
should remove some things not needed for `no-std`

provider-example/src/hpke.rs
```diff
@@ -213,14 +211,28 @@ impl HpkeOpener for HpkeRsReceiver {
...
+        Error::Other(OtherError(Arc::new(err)))
+    }
 }
 
 #[cfg(not(feature = "std"))]
-fn other_err(err: impl Send + Sync + 'static) -> Error {
-    Error::General(alloc::format!("{}", err));
+mod error_util {
+    use core::any::Any;
+    use rustls::Error;
```
brodycj, line 231: oops forgot this, apologies

```suggestion
    use rustls::Error;
    use rustls::OtherError;
```

djc (approved):
Nice!

PR #2265, opened 2024-12-09
Support `certificate_authorities` extension in ClientHello

s-arash:
This PR adds support for the [`certificate_authorities`](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.4) extension in ClientHello by:

 - On the client side, adding a default method to `ServerCertVerifier` that asks for CA names to be sent with ClientHello,

 - On the server side, adding the method `certificate_authorities()` to the `ClientHello` type, which is provided to `ResolvesServerCert` when the server needs to pick a cert.

Closes #2235

PR #2270, opened 2024-12-11
server: remove ClientHello constructor

djc:
Slightly fewer lines of code via logging abstraction and better readability. Worth it?

cc @s-arash from reviewing #2265.


ctz (approved):
Nice!

Issue #2257, opened 2024-12-04
UnsolicitedCertificateTypeExtension is not RFC 7250 compliant


ripatel-fd:
**Checklist**
* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**

Consider the following scenario:
- a rustls server using X.509 certificates
- a third party client advertising both X.509 certificates and Raw Public Keys as supported client certificate types

If this client attempts to connect to the rustls server, the handshake will fail with TLS alert "handshake failure" and message "UnsolicitedCertificateTypeExtension".

This is not RFC 7250 compliant. 

**To Reproduce**

1. Run a rustls server with a X.509 cert and Raw Public Key support disabled
2. Craft a ClientHello with a client_certificate_types extension advertising both X.509 and Raw Public Keys (RFC 7250)
3. Attempt to connect

**Applicable Version(s)**

rustls 0.23.16

**Expected behavior**

The client and server share a supported certificate type (both understand X.509).

According to [RFC 7250 Section 4.2](https://datatracker.ietf.org/doc/html/rfc7250#section-4.2), the server should then behave as follows:

- Don't send an alert (RFC 7250 makes no specific allowances to fail if both peers support X.509)
- Respond with the client_certificate_type extension in ServerHello; set the cert type to X.509 
- Same for server_certificate_type

**Additional context**

Here's an example of this failure in the wild. Below is a ClientHello sent by third party software

```
DEBUG  12-04 20:24:30.682265 1431788 f0   0    src/waltz/tls/fd_tls.c(1013): HEXDUMP "client hello" (194 bytes at 0x7fffe35689b0)
        0000:  01 00 00 be 03 03 76 b0 fc 10 9e b7 09 a3 0f 1c  ......v.........
        0010:  7c 1d 98 8e f6 38 4a ad 12 60 0d e8 75 ce 80 a0  |....8J..`..u...
        0020:  54 2f cc 00 44 ea 00 00 02 13 01 01 00 00 93 00  T/..D...........
        0030:  2b 00 03 02 03 04 00 33 00 26 00 24 00 1d 00 20  +......3.&.$... 
        0040:  bd c5 6b dd 45 85 c6 e2 a4 f5 aa 02 f9 7f 05 81  ..k.E...........
        0050:  a5 f4 42 83 3e 7a 73 e1 15 8d 9b 7c 42 ae 6b 31  ..B.>zs....|B.k1
[...]

PR #2274, opened 2024-12-11
Fix RFC 7250 Compliance (#2257)

holodorum:
This pull requests fixes issue #2257 by changing the way the server validates the `client_certificate_type_extension` and the `server_certificate_type_extension` in the `ServerHello`.  

Considerations
- Initially the validation logic for the certificate type extensions resided in `common_state` and validated the negotation of raw keys only, hence we called the struct `RawKeyNegotationParams`. The validation logic is different for the server and client now and we are not only negotiating raw keys, I moved this to `client::hs::CertificateTypeExtensionParams` and `server::hs::CertificateTypeExtensionParams`.

### DIFF
--- a/openssl-tests/src/raw_key_openssl_interop.rs
+++ b/openssl-tests/src/raw_key_openssl_interop.rs
@@ -418,36 +418,46 @@ mod tests {
             .spawn()
             .expect("Failed to execute OpenSSL client");
 
-        let mut stdin = openssl_client.stdin.take().unwrap();
-        let mut stdout = openssl_client.stdout.take().unwrap();
-        let mut stdout_buf = [0; 1024];
-        let mut openssl_stdout = String::new();
-        let mut received_server_msg = false;
-        loop {
-            match stdout.read(&mut stdout_buf) {
-                Ok(0) => break,
-                Ok(len) => {
-                    let read = &stdout_buf[..len];
-
-                    std::io::stdout()
-                        .write_all(read)
-                        .unwrap();
-                    openssl_stdout.push_str(&String::from_utf8_lossy(read));
-                    if openssl_stdout.contains("Hello from the server") {
-                        received_server_msg = true;
-                        stdin
-                            .write_all(b"Hello, from openssl client!")
-                            .expect("Failed to write to stdin");
-                        break;
-                    }
-                }
-                Err(e) => panic!("Error reading from OpenSSL stdin: {e:?}"),
-            }
-        }
+        let stdin = openssl_client.stdin.take().unwrap();
+        let stdout = openssl_client.stdout.take().unwrap();
+        let received_server_msg =
+            process_openssl_client_interaction(stdin, stdout, "Hello, from openssl client!");
 
         assert!(received_server_msg);
         assert_eq!(server_thread.join().unwrap(), "Hello, from openssl client!");
         openssl_client.wait().unwrap();
+    }
+
+    fn process_openssl_client_interaction(
+        mut stdin: std::process::ChildStdin,
+        mut stdout: std::process::ChildStdout,
+        message: &str,
+    ) -> bool {
+        let mut stdout_buf = [0; 1024];
+        let mut openssl_stdout = String::new();
+
+        loop {
+            let len = match stdout.read(&mut stdout_buf) {
+                Ok(0) => break,
+                Ok(len) => len,
+                Err(e) => panic!("Error reading from OpenSSL stdin: {e:?}"),
+            };
+
+            let read = &stdout_buf[..len];
+            std::io::stdout()
+                .write_all(read)
+                .unwrap();
+            openssl_stdout.push_str(&String::from_utf8_lossy(read));
+
+            if openssl_stdout.contains("Hello from the server") {
+                stdin
+                    .write_all(message.as_bytes())
+                    .expect("Failed to write to stdin");
+                return true;
+            }
+        }
+
+        false
     }
 
     #[test]
--- a/openssl-tests/src/raw_key_openssl_interop.rs
+++ b/openssl-tests/src/raw_key_openssl_interop.rs
@@ -329,6 +329,9 @@ mod tests {
     use std::sync::mpsc::channel;
     use std::thread;
 
+    use rustls::pki_types::pem::PemObject;
+    use rustls::pki_types::{CertificateDer, PrivateKeyDer};
+
     use super::{client, server};
     use crate::utils::verify_openssl3_available;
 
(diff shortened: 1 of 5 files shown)


rustls/src/client/hs.rs
```diff
@@ -1181,6 +1189,33 @@ impl State<ClientConnectionData> for ExpectServerHelloOrHelloRetryRequest {
...
+        match (client_expects_raw_key, server_supports_raw_key) {
+            (true, true) => CertificateTypeExtensionNegotationResult::Negotiated(
+                self.extension_type,
+                CertificateType::RawPublicKey,
+            ),
+            (false, false) => CertificateTypeExtensionNegotationResult::NotNegotiated,
+            (true, false) => CertificateTypeExtensionNegotationResult::Err(
+                Error::PeerIncompatible(PeerIncompatible::IncorrectCertificateTypeExtension),
+            ),
+            (false, true) => CertificateTypeExtensionNegotationResult::Err(
```
cpu, line 1212: The coverage gap flagged for this branch seems worth considering if it isn't too tricky. WDYT?

cpu (approved):
This seems reasonable to me. Thanks for the quick fix!

rustls/src/client/hs.rs
holodorum: To be honest I'm not sure if it's possible to get to the `(false, true)` arm. 
There are already tests sending an unsolicited certificate type extension in the `ServerHello`, but then an `UnsolicitedEncryptedExtension` error is raised. 
If there is truly no way to get to that situation we could also remove the `UnsolicitedCertificateTypeExtension` of course.

rustls/src/client/hs.rs
```diff
@@ -1181,6 +1189,33 @@ impl State<ClientConnectionData> for ExpectServerHelloOrHelloRetryRequest {
     }
 }
 
+#[derive(Debug)]
+struct CertificateTypeExtensionParams {
```
djc, line 1193: I don't think defining this type adds much value. It would be better to just define a function, maybe called `check_certificate_type()`.

rustls/src/client/hs.rs
```diff
@@ -1181,6 +1189,33 @@ impl State<ClientConnectionData> for ExpectServerHelloOrHelloRetryRequest {
...
+    server_negotiated: Option<CertificateType>,
+    client_expects: CertificateType,
+    extension_type: ExtensionType,
+}
+
+impl CertificateTypeExtensionParams {
+    fn validate_certificate_type_extension(&self) -> CertificateTypeExtensionNegotationResult {
+        let server_supports_raw_key = self.server_negotiated == Some(CertificateType::RawPublicKey);
+        let client_expects_raw_key = self.client_expects == CertificateType::RawPublicKey;
+        match (client_expects_raw_key, server_supports_raw_key) {
```
djc, line 1203: It seems like you could `match` `server_negotiated` and `client_expects` directly instead of first defining a `bool`, which would make this logic easier to follow.

rustls/src/client/hs.rs
```diff
@@ -1228,3 +1245,85 @@ fn low_quality_integer_hash(mut x: u32) -> u32 {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::msgs::enums::{CertificateType, ExtensionType};
+    use crate::{Error, PeerIncompatible};
+
+    use super::check_certificate_type;
+
+    #[test]
+    fn test_check_certificate_type() {
```
djc, line 1257: Not sure this test actually carries its weight? The compiler already checks that the `match` expression in `check_certificate_type()` is exhaustive, and I doubt whether adding an exact transliteration of the `match` in test form is actually useful.

Is this important to coverage? Can we generate the coverage in a more "real"/integrated way?

openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -388,6 +388,38 @@ mod tests {
         }
     }
 
+    fn process_openssl_client_interaction(
```
djc, line 391: Nit: please move the utility function below the tests that use it.

rustls/src/client/hs.rs
djc: Yup, much better.

rustls/src/client/hs.rs
holodorum: I agree with you. The tests in `api.rs` should already test most arms in an integrated way. 
Oddly enough before adding these tests the code coverage was complaining, after removing these tests and a new push it passes...

rustls/src/client/hs.rs
```diff
@@ -1181,6 +1179,25 @@ impl State<ClientConnectionData> for ExpectServerHelloOrHelloRetryRequest {
     }
 }
 
+fn check_certificate_type(
```
djc, line 1182: Suggest this takes a `requires_raw_keys: bool` instead of `client_extpects: CertificateType` and a reference to `common` so that it can take care of the `send_fatal_alert()` which is currently duplicated in both callers.

rustls/src/server/hs.rs
```diff
@@ -213,22 +212,25 @@ impl ExtensionProcessing {
...
+            CertificateType::X509
         };
+        let client_supports = hello
+            .server_certificate_extension()
+            .map(|certificate_types| certificate_types.to_vec())
+            .unwrap_or_default();
 
         self.process_cert_type_extension(
-            raw_key_negotation_params.validate_raw_key_negotiation(),
+            check_certificate_type(
```
djc, line 229: What's the reason to keep `process_cert_type_extension()` and `check_certificate_type()` split? It looks like we could save some duplication/overhead by merging them.

rustls/src/server/hs.rs
```diff
@@ -735,3 +739,55 @@ pub(crate) enum HandshakeHashOrBuffer {
     Buffer(HandshakeHashBuffer),
     Hash(HandshakeHash),
 }
+
+fn check_certificate_type(
+    client_supports: Vec<CertificateType>,
+    server_supports: CertificateType,
```
djc, line 745: Suggest using `requires_raw_keys: bool` instead of `server_supports: CertificateType`.

rustls/src/server/hs.rs
```diff
@@ -735,3 +739,55 @@ pub(crate) enum HandshakeHashOrBuffer {
...
+        return Ok(None);
+    }
+
+    if client_supports.contains(&server_supports) {
+        Ok(Some((extension_type, server_supports)))
+    } else {
+        Err(Error::PeerIncompatible(
+            PeerIncompatible::IncorrectCertificateTypeExtension,
+        ))
+    }
```
djc, line 758: Nit: this will be a little more compact as a `match`.

openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -428,32 +428,10 @@ mod tests {
...
-                        stdin
-                            .write_all(b"Hello, from openssl client!")
-                            .expect("Failed to write to stdin");
-                        break;
-                    }
-                }
-                Err(e) => panic!("Error reading from OpenSSL stdin: {e:?}"),
-            }
-        }
+        let stdin = openssl_client.stdin.take().unwrap();
```
djc, line 431: It looks like this commit only changes the test code introduced in the previous commit? If so, just squash it.

holodorum: There are also some changes to the already existing tests `test_rust_server_with_openssl_client`. Is it still best to squash in that case?

rustls/src/server/hs.rs
holodorum: I'm not sure if that's clearer. Now the idea is that if the server supports either `X509` or a `RawPublicKey` and `client_supports` contains this certificate type we should continue with the handshake. 
Initially we were using a `bool`, but then it becomes more complicated to verify the cases where the client supports both `CertificateType`'s and the server supports only one I think.

djc: I'm not sure why/how? `server_supports: CertificateType` provides 1 bit of information as well, so what is the difference?

openssl-tests/src/raw_key_openssl_interop.rs
djc: I think you should have a commit that does the test refactoring for the existing test prior to the main commit (which adds a new test which also needs the newly extracted helper).

cpu:
Just to check my understanding here: is this blocked on another round of iteration based on djc's most recent feedback?

rustls/src/server/hs.rs
holodorum: You're right, I didn't look at it that way. Thank you

holodorum:
The semver test is failing, with the message `pub fn removed or renamed`, before rebasing on main the test didn't fail. Any idea, why it fails and how to solve that?

brody4hire:
I tried merging this with recent updates from main in https://github.com/brody4hire/rustls/pull/8 - semver seems to be OK now. But codecov does seem to indicate some missing test coverage.

I tried a couple mutations in https://github.com/brody4hire/rustls/pull/8 based on the missing test coverage, they do not seem to trigger any CI testing failures. I think we need to add the missing test coverage.

rustls/src/server/hs.rs
```diff
@@ -239,52 +233,63 @@ impl ExtensionProcessing {
...
+            Err(err) => {
                 return Err(cx
                     .common
                     .send_fatal_alert(AlertDescription::HandshakeFailure, err));
             }
-            RawKeyNegotationResult::NotNegotiated => {}
-            RawKeyNegotationResult::Negotiated(_) => unreachable!(
-                "The extension type should only ever be ClientCertificateType or ServerCertificateType"
-            ),
+            Ok((_,_)) => return Err(Error::General("The extension type should only ever be ClientCertificateType or ServerCertificateType".to_string())),
```
ctz, line 292: I would like to avoid new uses of "Error::General" -- in this case I think this statement is statically unreachable. I would suggest this instead:

- `Ok((_,_)) => unreachable!(),` here.
- `debug_assert!(extension_type == ExtensionType::ClientCertificateType || extension_type == ExtensionType::ServerCertificateType);` at the top of this function to record its valid inputs
- then drop the artificial test introduced at the bottom

ctz (commented):
I'm pretty happy with this. One comment, and then I feel this is mergable. A rebase will fix the CI issue now.

ctz (approved):
Thanks! Appreciate your patience.

holodorum:
> Thanks! Appreciate your patience.

Thank you all too for the feedback and patience😁

PR #2277, opened 2024-12-12
improve provider-example CI testing & coverage testing

brodycj:
Recently merged PR #2269 had a CI failure since some affected code was not included in the coverage testing. This update is also needed to remove coverage failure from PR #2200 which is still in progress & should be almost ready.

I think the provider-example testing should also be included in the Build+test CI job, as proposed here.


djc (approved):

cpu:
Seems uncontroversial enough to merge without waiting on further reviews.

PR #2278, opened 2024-12-12
examples: ech-client should process all HTTPS records

### LINKED ISSUE #2276: example ech-client.rs panics on some URLs
**Checklist**
* [x] I've searched the issue tracker for similar bugs. (Well, not really, I'm confident this sample hasn't been tried with these URLs:-)

**Describe the bug**
With some of my ECH test URLs, the sample ech-client.rs panics. I think that's for some cases with multi-valued HTTPS RRs, some of which are ok and some of which contain bad or unsupported values.

The output I see includes: 
```
thread 'main' panicked at src/main.rs:61:14:
called `Result::unwrap()` on an `Err` value: InvalidEncryptedClientHello(NoCompatibleConfig)
```
(Note, I have copied the source for this sample and done some minor edits, so line 61 above would be line 70 in the repo here.)

**To Reproduce**

Using the sample with these URLs sometimes has this behaviour, other times works. (Presumably dependent on the ordering of DNS answers, but not verified.)

https://curves1-ng.test.defo.ie/echstat.php?format=json
https://curves1-ng.test.defo.ie:15443/echstat.php?format=json
https://curves2-ng.test.defo.ie/echstat.php?format=json
https://curves2-ng.test.defo.ie:15443/echstat.php?format=json
https://curves3-ng.test.defo.ie/echstat.php?format=json
https://curves3-ng.test.defo.ie:15443/echstat.php?format=json
https://v4-ng.test.defo.ie/echstat.php?format=json
https://v4-ng.test.defo.ie:15443/echstat.php?format=json

The names for those URLs have various multi-valued HTTPS RRs. If you load https://test.defo.ie/iframe_tests.html you can see a rough description on those (search for the names in the page).

The public name in the (well-encoded) HTTPS RRs for all these should be 'public.test.defo.ie'.

**Applicable Version(s)**
0.23.19 (I think)

**Expected behavior**
Ingest the correct DNS answers and skip over the bogus ones.

I'm not sure if this is purely an issue with the sample code or if it perhaps runs deeper. (Afraid my rustiness doesn't extend that far:-)

cpu:
Previously in the `ech-client.rs` example we only pulled `EchConfigListBytes` from the _first_ HTTPS record `ech-config` SVC param we found when performing a DNS-over-HTTPS lookup using HickoryDNS.

Instead we should process all of the HTTPS records, collect up all of the `EchConfigListBytes` from each we can, and then use the first compatible ECH config from all of the available lists.

As reported in #2276 you could observe this causing issues when running:
```
cargo run --package rustls-examples --bin ech-client -- \
   --host curves1-ng.test.defo.ie \
   --path "echstat.php?format=json" \
   public.test.defo.ie curves1-ng.test.defo.ie
```

Before this change, the client would panic approximately 50% of the time after finding no supported ECH configs from the list in the first HTTPS record. The ECH configs in the other records were not examined.

With this change the client acts as expected and successfully negotiates ECH with the server after finding a good config from the whole set.

Along the way I tried to reduce some of the combinator-soup in favour of simpler logic.

Resolves #2276

### DIFF
--- a/examples/src/bin/ech-client.rs
+++ b/examples/src/bin/ech-client.rs
@@ -30,7 +30,7 @@ use clap::Parser;
 use hickory_resolver::config::{ResolverConfig, ResolverOpts};
 use hickory_resolver::proto::rr::rdata::svcb::{SvcParamKey, SvcParamValue};
 use hickory_resolver::proto::rr::{RData, RecordType};
-use hickory_resolver::{Resolver, TokioResolver};
+use hickory_resolver::{ResolveError, Resolver, TokioResolver};
 use log::trace;
 use rustls::client::{EchConfig, EchGreaseConfig, EchStatus};
 use rustls::crypto::aws_lc_rs;
@@ -55,7 +55,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
         true => None, // Force the use of the GREASE ext by skipping ECH config lookup
         false => match args.ech_config {
             Some(path) => Some(read_ech(&path)?),
-            None => lookup_ech_configs(&resolver, &args.inner_hostname, args.port).await,
+            None => lookup_ech_configs(&resolver, &args.inner_hostname, args.port).await?,
         },
     };
 
@@ -197,7 +197,7 @@ async fn lookup_ech_configs(
     resolver: &TokioResolver,
     domain: &str,
     port: u16,
-) -> Option<EchConfigListBytes<'static>> {
+) -> Result<Option<EchConfigListBytes<'static>>, ResolveError> {
     // For non-standard ports, lookup the ECHConfig using port-prefix naming
     // See: https://datatracker.ietf.org/doc/html/rfc9460#section-9.1
     let qname_to_lookup = match port {
@@ -205,10 +205,9 @@ async fn lookup_ech_configs(
         port => format!("_{port}._https.{domain}"),
     };
 
-    resolver
+    Ok(resolver
         .lookup(qname_to_lookup, RecordType::HTTPS)
-        .await
-        .ok()?
+        .await?
         .record_iter()
         .find_map(|r| match r.data() {
             RData::HTTPS(svcb) => svcb
@@ -222,7 +221,7 @@ async fn lookup_ech_configs(
                 }),
             _ => None,
         })
-        .map(Into::into)
+        .map(Into::into))
 }
 
 fn read_ech(path: &str) -> Result<EchConfigListBytes<'static>, Box<dyn Error>> {
--- a/examples/src/bin/ech-client.rs
+++ b/examples/src/bin/ech-client.rs
@@ -44,18 +44,24 @@ use rustls::RootCertStore;
 async fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::parse();
 
-    // Find raw ECH configs using DNS-over-HTTPS with Hickory DNS.
-    let resolver_config = if args.use_cloudflare_dns {
-        ResolverConfig::cloudflare_https()
-    } else {
-        ResolverConfig::google_https()
-    };
-    let resolver = Resolver::tokio(resolver_config, ResolverOpts::default());
     let server_ech_config = match args.grease {
         true => None, // Force the use of the GREASE ext by skipping ECH config lookup
         false => match args.ech_config {
             Some(path) => Some(read_ech(&path)?),
-            None => lookup_ech_configs(&resolver, &args.inner_hostname, args.port).await?,
+            None => {
+                // Find raw ECH configs using DNS-over-HTTPS with Hickory DNS.
+                let resolver_config = if args.use_cloudflare_dns {
+                    ResolverConfig::cloudflare_https()
+                } else {
+                    ResolverConfig::google_https()
+                };
+                lookup_ech_configs(
+                    &Resolver::tokio(resolver_config, ResolverOpts::default()),
+                    &args.inner_hostname,
+                    args.port,
+                )
+                .await?
+            }
         },
     };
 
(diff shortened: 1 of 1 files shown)


cpu:
> As reported in https://github.com/rustls/rustls/issues/2276 you could observe this causing issues when running:
>
>cargo run --package rustls-examples --bin ech-client -- \
>   --host curves1-ng.test.defo.ie \
>   --path "echstat.php?format=json" \
>   public.test.defo.ie curves1-ng.test.defo.ie

I haven't added this host to the [daily-test CI job](https://github.com/rustls/rustls/blob/deb04c286ce074db1a9ebed50e90462d26554065/.github/workflows/daily-tests.yml#L118-L126). Perhaps I should?

examples/src/bin/ech-client.rs
```diff
@@ -43,18 +43,28 @@ use rustls::RootCertStore;
...
-    // Find raw ECH configs using DNS-over-HTTPS with Hickory DNS.
-    let resolver_config = if args.use_cloudflare_dns {
-        ResolverConfig::cloudflare_https()
-    } else {
-        ResolverConfig::google_https()
-    };
-    let resolver = Resolver::tokio(resolver_config, ResolverOpts::default());
-    let server_ech_config = match args.grease {
-        true => None, // Force the use of the GREASE ext by skipping ECH config lookup
+    // Determine the ECH config to use.
```
djc, line 46: This commit feels like it has a bunch of stuff going on at the same time... Maybe split it a little?

examples/src/bin/ech-client.rs
```diff
@@ -65,11 +75,15 @@ async fn main() {
...
-            .unwrap()
-            .into(),
-        None => {
+    // If the ECH config list is empty, use a GREASE ECH config.
+    // Otherwise, use the first compatible ECH config from one of the ECH config lists.
+    let ech_mode = match server_ech_configs.is_empty() {
+        false => server_ech_configs
+            .into_iter()
+            .find_map(|list| EchConfig::new(list, ALL_SUPPORTED_SUITES).ok())
+            .map(Into::into)
```
djc, line 84: I feel like the use of `Into` is a little opaque here (also this particular case could/should probably go inside the `find_map()`?) -- the type we're casting to isn't really clear here, so maybe go with some kind of explicit conversion (even just wrapping in `T::from()`)?

examples/src/bin/ech-client.rs
```diff
@@ -201,37 +215,50 @@ struct Args {
...
-    // For non-standard ports, lookup the ECHConfig using port-prefix naming
-    // See: https://datatracker.ietf.org/doc/html/rfc9460#section-9.1
+) -> Vec<EchConfigListBytes<'static>> {
     let qname_to_lookup = match port {
         443 => domain.to_owned(),
         port => format!("_{port}._https.{domain}"),
     };
 
-    resolver
+    let Ok(lookup) = resolver
```
djc, line 236: Hmm, should we be silently discarding the error here?

examples/src/bin/ech-client.rs
```diff
@@ -201,37 +215,50 @@ struct Args {
...
+        };
+
+        ech_config_lists.extend(
+            svcb.svc_params()
                 .iter()
-                .find_map(|sp| match sp {
+                .filter_map(|params| match params {
                     (SvcParamKey::EchConfigList, SvcParamValue::EchConfigList(e)) => {
-                        Some(e.clone().0)
+                        Some(e.clone().0.into())
```
djc, line 254: Here, too, I feel the code would be easier to follow if you skipped the type annotation on `ech_config_lists` definition in favor of calling `EchConfigListBytes::from()` explicitly.

examples/src/bin/ech-client.rs
cpu: Broken up into smaller pieces now. I should have done that from the start :-)

examples/src/bin/ech-client.rs
cpu: Fixed up.

examples/src/bin/ech-client.rs
cpu: Yeah I think it'd be better to return the error. Updated.

examples/src/bin/ech-client.rs
cpu: Fixed.

examples/src/bin/ech-client.rs
```diff
@@ -44,18 +44,24 @@ use rustls::RootCertStore;
...
-    } else {
-        ResolverConfig::google_https()
-    };
-    let resolver = Resolver::tokio(resolver_config, ResolverOpts::default());
     let server_ech_config = match args.grease {
         true => None, // Force the use of the GREASE ext by skipping ECH config lookup
         false => match args.ech_config {
             Some(path) => Some(read_ech(&path)?),
-            None => lookup_ech_configs(&resolver, &args.inner_hostname, args.port).await?,
+            None => {
```
djc, line 51: Maybe this would be a bit nicer as a `match (args.grease, args.ech_config) { .. }`? At least it would save some rightward drift.

cpu: Good call, it does make a nice improvement. I applied as a separate commit (and also tacked on one more updating the rustdoc at the top of file).

djc (approved):

ctz (approved):
Nice!

(Do we need to sync `tests/ech_fetch.rs` in rustls-ffi? I guess probably not with any urgency given it is just test infra...)

cpu:
> (Do we need to sync tests/ech_fetch.rs in rustls-ffi? I guess probably not with any urgency given it is just test infra...)

Yeah, the `client.c` example will also need an update. I filed https://github.com/rustls/rustls-ffi/issues/503 for this before I forget. I think I can pick it up next week.

PR #2281, opened 2024-12-13
post-quantum key exchange writeup

ctz:
(still working on this)

[Rendered](https://github.com/rustls/rustls/blob/jbp-pqkx-writeup/website/content/perf/2024-12-12-pq-kx/index.md)


ctz:
This is ready for review now. The code for the non-optimised case is on [jbp-pqkx-comparative-benchmarks](https://github.com/rustls/rustls/compare/main...jbp-pqkx-comparative-benchmarks)

Manual documentation run is https://github.com/rustls/rustls/actions/runs/12372917287/job/34532172546 (deploy step is expected to fail, but prior step contains the website generation, and internal/external link checking.)

djc (approved):
Nice work!

I feel like the article kind of buries the lede ("rustls can do PQC handshakes faster than OpenSSL handles regular ones"), and I wonder if it might be worth restructuring to put that conclusion up front (maybe also in the title), and then back it up with the full story?

cpu (approved):
I'm not sure if I'm missing something obvious, or looking at an older rendering, but in the part of the writeup that says "Instead, we can do:" and has the second diagram I feel like it's not super clear what the difference is between that diagram, and the one before it. Could we maybe use some colour here? Something like:

* Diagram 1: `[ML-KEM share (green)][X25519 share (yellow)]    [X25519 share (orange)]` 
* Diagram 2: `[ML-KEM share (green)][X25519 share (yellow)]    [X25519 share (yellow)]`

djc:
> Yeah I think I'm trying to tell two stories. Maybe this could be clearer as two completely separate pages?
> 
> one page about the optimization
> another about the overall cost of pq key exchange, and the comparison to openssl

I think both of those might be too thin on their own so combining them could still make sense, just consider more clearly separating these two topics and presenting them in optimal order (which I suggest would be to have the OpenSSL comparison come first).

cpu:
> I think both of those might be too thin on their own so combining them could still make sense

FWIW I was thinking the same :+1:

ctz:
> I think both of those might be too thin on their own so combining them could still make sense

OK, I think that reads a lot better. WDYT?

cpu (approved):
I think this version reads better. Thanks!

PR #2285, opened 2024-12-18
top-level `sync::Arc` alias to support fork for targets with no atomic ptr

### LINKED ISSUE #2068: `no-std` support for targets w/o atomics?

brody4hire:
Adds an internal `sync` module aliasing the `Arc` implementation to allow downstream forks of rustls targetting architectures without atomic pointers to replace the implementation with another implementation such as `portable_atomic_util::Arc` in one central location.

This update should make it more straightforward for a fork of rustls to overwrite the top-level import of `alloc::sync::Arc` in `rustls/src/lib.rs` with another implementation such as `portable_atomic_util::Arc`.

(A fork would also need to single test import of `Arc` in `rustls/tests/common/mod.rs`.)

To prevent accidental use of the non-aliased `Arc` the project `.clippy.toml` is upated to add `std::sync::Arc` to the `disallowed-types`. Since we only want this to apply to the main crate the CI clippy invocations targetting auxiliary crates are updated to ignore `disallowed-types` findings. We can improve this in the future with more targetted lint configuration from `Cargo.toml` settings.

---

Resolves #2068 (ref: https://github.com/rustls/rustls/issues/2068)

---

Replaces rejected PR: https://github.com/rustls/rustls/pull/2200

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,1 +1,6 @@
 upper-case-acronyms-aggressive = true
+
+# only intended to affect main rustls crate - need to allow disallowed-types in all other crates in Clippy CI job
+disallowed-types = [
+  { path = "std::sync::Arc", reason = "must use Arc from sync module to support downstream forks targetting architectures without atomic ptrs" },
+]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -464,7 +464,11 @@ jobs:
         with:
           components: clippy
       # We want to be free of any warnings, so deny them.
-      - run: ./admin/clippy -- --deny warnings
+      # - Allow `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for the main `rustls` crate.
+      - run: ./admin/clippy -- --deny warnings --allow clippy::disallowed_types
+      # - Keep the main crate free of all warnings.
+      - run: cargo clippy -p rustls -- --deny warnings
 
   clippy-nightly:
     name: Clippy (Nightly)
@@ -483,8 +487,12 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           components: clippy
-      # do not deny warnings, as nightly clippy sometimes has false negatives
-      - run: ./admin/clippy
+      # Do not deny warnings, as nightly clippy sometimes has false negatives.
+      # - Ignore `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
+      - run: ./admin/clippy -- --allow clippy::disallowed_types
+      # Check the main crate for any Clippy nightly warnings, but do not deny them.
+      - run: cargo clippy -p rustls
 
   check-external-types:
     name: Validate external types appearing in public API
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -1,5 +1,4 @@
 use alloc::boxed::Box;
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 
@@ -8,6 +7,7 @@ use zeroize::Zeroize;
 
 use crate::msgs::ffdhe_groups::FfdheGroup;
 use crate::sign::SigningKey;
+use crate::sync::Arc;
 pub use crate::webpki::{
     verify_tls12_signature, verify_tls13_signature, verify_tls13_signature_with_raw_key,
     WebPkiSupportedAlgorithms,
@@ -695,7 +695,6 @@ pub fn default_fips_provider() -> CryptoProvider {
 mod static_default {
     #[cfg(not(feature = "std"))]
     use alloc::boxed::Box;
-    use alloc::sync::Arc;
     #[cfg(feature = "std")]
     use std::sync::OnceLock;
 
@@ -703,6 +702,7 @@ mod static_default {
     use once_cell::race::OnceBox;
 
     use super::CryptoProvider;
+    use crate::sync::Arc;
 
     #[cfg(feature = "std")]
     pub(crate) fn install_default(
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -401,6 +401,14 @@ mod log {
 #[macro_use]
 mod test_macros;
 
+/// This internal `sync` module aliases the `Arc` implementation to allow downstream forks
+/// of rustls targetting architectures without atomic pointers to replace the implementation
+/// with another implementation such as `portable_atomic_util::Arc` in one central location.
+mod sync {
+    #[allow(clippy::disallowed_types)]
+    pub(crate) type Arc<T> = alloc::sync::Arc<T>;
+}
+
 #[macro_use]
 mod msgs;
 mod common_state;
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -6,10 +6,11 @@ use std::fmt::Debug;
 use std::io::{self, IoSlice, Read, Write};
 use std::ops::{Deref, DerefMut};
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex};
+use std::sync::Mutex;
 use std::{fmt, mem};
 
 use pki_types::{CertificateDer, IpAddr, ServerName, UnixTime};
+
 use rustls::client::{verify_server_cert_signed_by_trust_anchor, ResolvesClientCert, Resumption};
 use rustls::crypto::{ActiveKeyExchange, CryptoProvider, SharedSecret, SupportedKxGroup};
 use rustls::internal::msgs::base::Payload;
--- a/rustls/tests/common/mod.rs
+++ b/rustls/tests/common/mod.rs
@@ -3,13 +3,14 @@
 
 use std::io;
 use std::ops::DerefMut;
-use std::sync::{Arc, OnceLock};
+use std::sync::OnceLock;
 
 use pki_types::pem::PemObject;
 use pki_types::{
     CertificateDer, CertificateRevocationListDer, PrivateKeyDer, PrivatePkcs8KeyDer, ServerName,
     SubjectPublicKeyInfoDer, UnixTime,
 };
+
 use rustls::client::danger::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
 use rustls::client::{
     AlwaysResolvesClientRawPublicKeys, ServerCertVerifierBuilder, WebPkiServerVerifier,
@@ -28,9 +29,13 @@ use rustls::{
     DigitallySignedStruct, DistinguishedName, Error, InconsistentKeys, NamedGroup, ProtocolVersion,
     RootCertStore, ServerConfig, ServerConnection, SideData, SignatureScheme, SupportedCipherSuite,
 };
+
 use webpki::anchor_from_trusted_cert;
 
 use super::provider;
+
+// Import `Arc` here for tests - can be overwritten to test with another `Arc` such as `portable_atomic_util::Arc`
+pub use std::sync::Arc;
 
 macro_rules! embed_files {
     (
--- a/rustls/tests/unbuffered.rs
+++ b/rustls/tests/unbuffered.rs
@@ -1,7 +1,6 @@
 #![allow(clippy::duplicate_mod)]
 
 use std::num::NonZeroUsize;
-use std::sync::Arc;
 
 use rustls::client::{ClientConnectionData, EarlyDataError, UnbufferedClientConnection};
 use rustls::server::{ServerConnectionData, UnbufferedServerConnection};
--- a/rustls/src/builder.rs
+++ b/rustls/src/builder.rs
@@ -1,5 +1,4 @@
 use alloc::format;
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt;
 use core::marker::PhantomData;
@@ -8,6 +7,7 @@ use crate::client::EchMode;
 use crate::crypto::CryptoProvider;
 use crate::error::Error;
 use crate::msgs::handshake::ALL_KEY_EXCHANGE_ALGORITHMS;
+use crate::sync::Arc;
 use crate::time_provider::TimeProvider;
 use crate::versions;
 #[cfg(doc)]
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -1,4 +1,3 @@
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::marker::PhantomData;
 
@@ -10,6 +9,7 @@ use crate::client::{handy, ClientConfig, EchMode, ResolvesClientCert};
 use crate::error::Error;
 use crate::key_log::NoKeyLog;
 use crate::msgs::handshake::CertificateChain;
+use crate::sync::Arc;
 use crate::versions::TLS13;
 use crate::webpki::{self, WebPkiServerVerifier};
 use crate::{compress, verify, versions, WantsVersions};
@@ -89,10 +89,10 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
 
 /// Container for unsafe APIs
 pub(super) mod danger {
-    use alloc::sync::Arc;
     use core::marker::PhantomData;
 
     use crate::client::WantsClientCert;
+    use crate::sync::Arc;
     use crate::{verify, ClientConfig, ConfigBuilder, WantsVerifier};
 
     /// Accessor for dangerous configuration options.
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -1,4 +1,3 @@
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::marker::PhantomData;
 use core::ops::{Deref, DerefMut};
@@ -20,6 +19,7 @@ use crate::msgs::enums::NamedGroup;
 use crate::msgs::handshake::ClientExtension;
 use crate::msgs::persist;
 use crate::suites::SupportedCipherSuite;
+use crate::sync::Arc;
 #[cfg(feature = "std")]
 use crate::time_provider::DefaultTimeProvider;
 use crate::time_provider::TimeProvider;
@@ -511,10 +511,9 @@ pub enum Tls12Resumption {
 
 /// Container for unsafe APIs
 pub(super) mod danger {
-    use alloc::sync::Arc;
-
     use super::verify::ServerCertVerifier;
     use super::ClientConfig;
+    use crate::sync::Arc;
 
     /// Accessor for dangerous configuration options.
     #[derive(Debug)]
@@ -611,7 +610,6 @@ impl EarlyData {
 
 #[cfg(feature = "std")]
 mod connection {
-    use alloc::sync::Arc;
     use alloc::vec::Vec;
     use core::fmt;
     use core::ops::{Deref, DerefMut};
@@ -625,6 +623,7 @@ mod connection {
     use crate::conn::{ConnectionCommon, ConnectionCore};
     use crate::error::Error;
     use crate::suites::ExtractedSecrets;
+    use crate::sync::Arc;
     use crate::ClientConfig;
 
     /// Stub that implements io::Write and dispatches to `write_early_data`.
--- a/rustls/src/client/common.rs
+++ b/rustls/src/client/common.rs
@@ -1,11 +1,11 @@
 use alloc::boxed::Box;
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 
 use super::ResolvesClientCert;
 use crate::log::{debug, trace};
 use crate::msgs::enums::ExtensionType;
 use crate::msgs::handshake::{CertificateChain, DistinguishedName, ServerExtension};
+use crate::sync::Arc;
 use crate::{compress, sign, SignatureScheme};
 
 #[derive(Debug)]
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -1,11 +1,10 @@
-use alloc::sync::Arc;
-
 use pki_types::ServerName;
 
 use crate::enums::SignatureScheme;
 use crate::error::Error;
 use crate::msgs::handshake::CertificateChain;
 use crate::msgs::persist;
+use crate::sync::Arc;
 use crate::{client, sign, NamedGroup};
 
 /// An implementer of `ClientSessionStore` which does nothing.
@@ -279,7 +278,6 @@ impl client::ResolvesClientCert for AlwaysResolvesClientRawPublicKeys {
 #[cfg(test)]
 #[macro_rules_attribute::apply(test_for_each_provider)]
 mod tests {
-    use alloc::sync::Arc;
     use std::prelude::v1::*;
 
     use pki_types::{ServerName, UnixTime};
@@ -294,6 +292,7 @@ mod tests {
     use crate::msgs::handshake::SessionId;
     use crate::msgs::persist::Tls13ClientSessionValue;
     use crate::suites::SupportedCipherSuite;
+    use crate::sync::Arc;
 
     #[test]
     fn test_noclientsessionstorage_does_nothing() {
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -1,6 +1,5 @@
 use alloc::borrow::ToOwned;
 use alloc::boxed::Box;
-use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::ops::Deref;
@@ -37,6 +36,7 @@ use crate::msgs::handshake::{
 };
 use crate::msgs::message::{Message, MessagePayload};
 use crate::msgs::persist;
+use crate::sync::Arc;
 use crate::tls13::key_schedule::KeyScheduleEarly;
 use crate::SupportedCipherSuite;
 
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1,6 +1,5 @@
 use alloc::borrow::ToOwned;
 use alloc::boxed::Box;
-use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 
@@ -31,6 +30,7 @@ use crate::msgs::message::{Message, MessagePayload};
 use crate::msgs::persist;
 use crate::sign::Signer;
 use crate::suites::{PartiallyExtractedSecrets, SupportedCipherSuite};
+use crate::sync::Arc;
 use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
 use crate::verify::{self, DigitallySignedStruct};
 
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -1,5 +1,4 @@
 use alloc::boxed::Box;
-use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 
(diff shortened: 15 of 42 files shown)


djc:
@brodycj every time you push to a PR in this repo, you generate notifications for everyone watching the repo. If you make a fork in your personal account, you should be able to run CI workflows there too. For your DO NOT MERGE branches, I think that would be a better option.

rustls/src/crypto/mod.rs
```diff
@@ -124,7 +124,7 @@ pub use crate::suites::CipherSuiteCommon;
 ///
 /// ```
 /// # #[cfg(feature = "aws_lc_rs")] {
-/// # use std::sync::Arc;
```
ctz, line 127: I think we should stick to using `std::sync::Arc` for non-unit tests, and lose `rustls::internal::alias::atomic_sync::Arc`. Generally `internal` is used for tests to reuse code they otherwise absolutely cannot.

brodycj: @ctz I think that sounds reasonable, will try to fix this sometime later today.

brodycj:
Hey checking in on this, please let me know if I should rebase it again.

ctz (approved):

.github/workflows/build.yml
```diff
@@ -192,6 +192,76 @@ jobs:
         run: cargo test --locked --release --no-run
         working-directory: rustls
 
+  overwrite-atomic-sync-build-targets:
```
djc, line 195: I don't think we should merge this -- I don't think this repository's maintainers should be on the hook for enforcing that this works, given the level of complexity here.

rustls/src/lib.rs
```diff
@@ -401,6 +401,10 @@ mod log {
 #[macro_use]
 mod test_macros;
 
+// For top-level `Arc` type import - may be overwritten to use a custom `Arc` implementation
```
djc, line 404: I think this should be an inline module, and I think we should call it `sync` to match similar modules in `std` and `tokio`, instead of inventing a different name here.

.github/workflows/build.yml
ctz: Ah! To add to the malaise I'm not sure I support this PR without some sort of defense of the changes.

I wonder if clippy `disallowed-types` (and co) could be used instead? (With a specific `allow` just for `crate::atomic_sync`?

.github/workflows/build.yml
```diff
@@ -383,15 +453,25 @@ jobs:
...
+        uses: dtolnay/rust-toolchain@nightly
       - name: Install cross (cross-rs) from GitHub
         run: cargo install cross --git https://github.com/cross-rs/cross
       - name: Install bindgen feature & CLI for aws-lc-sys (as needed for many cross targets)
         if: ${{ matrix.target != 'i686-unknown-linux-gnu' }}
         run: cargo add --dev --features bindgen 'aws-lc-sys@>0.20' --package rustls --verbose && cargo install bindgen-cli --verbose
       - run: cross test --package rustls --target ${{ matrix.target }}
 
+      # ############################################################################
+      # TEST with portable_atomic_util::Arc - MUST COME AT THE END OF THIS CI JOB
```
cpu, line 467: I think this comment would be better if it had some mention of why the ordering matters. Presumably because it's changing the source code of the checked out repo?

If we stick to this approach I think I'd mildly prefer to see a stand-alone script that applies the edits, runs the needed tests, and then undoes the edits in an exit trap handler. 

I think there's general aversion to complex bash scripts so this isn't ideal either but it seems better to have it all in one place and somewhat atomic.

admin/use-portable-atomic-arc
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Install dependencies to use `once_cell` & `portable_atomic_util::Arc` with critical-section
+# (cannot use `once_cell` with `--no-default-features` here due to workspace config in top-level `Cargo.toml`)
+cargo add once_cell -F alloc,race,critical-section -p rustls
```
cpu, line 7: optional nit: I'd prefer to see long-form opts used where possible. The purpose of `-F` is less obvious than `--features` and we're not optimizing for shortest commands.

admin/use-portable-atomic-arc
```diff
@@ -0,0 +1,15 @@
...
+cargo add portable-atomic --no-default-features -F critical-section -p rustls
+cargo add portable-atomic-util --no-default-features -F alloc -p rustls
+
+# Update `rustls/src`: wverwrite `rustls/src/atomic_sync.rs` to use `portable_atomic_util::Arc`
```
cpu, line 11: typo:

```suggestion
# Update `rustls/src`: overwrite `rustls/src/atomic_sync.rs` to use `portable_atomic_util::Arc`
```

rustls/src/atomic_sync.rs
```diff
@@ -0,0 +1,7 @@
+// NOTE: This `atomic_sync` module is intended to make it easier for a fork to
+// use another implementation of `Arc` such as `portable_atomic_util::Arc`,
+// which can help support build targets with no atomic pointer.
+// This module also makes it really easy for CI to overwrite this import
+// for build & unit testing with `portable_atomic_util::Arc`.
```
cpu, line 5: I think we could improve this comment by describing more of how it makes it easier for these consumers. E.g. it's a level of indirection that gives a central place to change the `Arc` implementation. I think using Rustdoc tripple-slashes for the comment would be also be preferred.

```suggestion
/// The `atomic_sync` module indirects the `Arc` implementation to allow downstream forks
/// of rustls targetting architectures without atomics to replace the implementation in one
/// central location.
///
/// See `admin/use-portable-atomic-arc` and its use in CI for a demonstration.
```

rustls/tests/api.rs
```diff
@@ -42,6 +43,7 @@ use rustls::{
 use super::*;
 
 mod common;
+// no need to import `common::Arc` - included by wildcard import below
```
cpu, line 46: This feels like something that was useful to call out to reviewers but isn't likely to be useful to a reader of the code in the future. WDYT about deleting this comment addition?

rustls/tests/common/mod.rs
```diff
@@ -28,10 +29,14 @@ use rustls::{
     DigitallySignedStruct, DistinguishedName, Error, InconsistentKeys, NamedGroup, ProtocolVersion,
     RootCertStore, ServerConfig, ServerConnection, SideData, SignatureScheme, SupportedCipherSuite,
 };
+
 use webpki::anchor_from_trusted_cert;
 
 use super::provider;
 
+// Import `Arc` here for tests - can be overwritten to test with another Arc  such as `portable_atomic_util::Arc`
```
cpu, line 37: nits: consistent backtick usage, extra space between "Arc" and "such":

```suggestion
// Import `Arc` here for tests - can be overwritten to test with another `Arc` such as `portable_atomic_util::Arc`
```

rustls/tests/common/mod.rs
```diff
@@ -28,10 +29,14 @@ use rustls::{
...
+pub use std::sync::Arc;
```
cpu, line 38: It feels awkward that the concrete `Arc` type is selected separately for the library and the tests. It doesn't feel like there's a case where you want to change out one and not the other.

I imagine potential solutions to that have their own downsides. WDYT?

rustls/tests/unbuffered.rs
```diff
@@ -17,6 +16,7 @@ use rustls::{
 use super::*;
 
 mod common;
+// no need to import `common::Arc` - included by wildcard import below
```
cpu, line 19: Ditto here RE: my earlier feedback about the utility of this sort of comment.

cpu (commented):
This feels closer to the right set of tradeoffs than #2200 but I'm hoping we can dial in some of the aspects called out by ctz & djc in their reviews before merging.

Thanks!

rustls/tests/common/mod.rs
brody4hire: Hmm ... my thought is that tests would have to follow the library API ... fork changes the library API then may need to update the tests.

I recall earlier discussion of adding `internal::alias` to export an undocumented alias & help alleviate this issue with the tests. This was my initial approach, changed as discussed earlier here: https://github.com/rustls/rustls/pull/2285#discussion_r1891378527

rustls/src/lib.rs
brody4hire: Done but I don't like this 100%. I was able to hack something with help from `sed` into `admin/use-portable-atomic-arc` but don't feel this would be the best long-term solution.

I also do not like having what is normally a standard Rust import for `Arc` coming right in the middle of the internal crate imports. Here is an example from `rustls/src/builder.rs`:

```rs
use crate::client::EchMode;
use crate::crypto::CryptoProvider;
use crate::error::Error;
use crate::msgs::handshake::ALL_KEY_EXCHANGE_ALGORITHMS;
use crate::sync::Arc;
use crate::time_provider::TimeProvider;
```

I would favor something more like this:

```rs
use crate::sync::Arc;

use crate::client::EchMode;
use crate::crypto::CryptoProvider;
use crate::error::Error;
use crate::msgs::handshake::ALL_KEY_EXCHANGE_ALGORITHMS;
use crate::time_provider::TimeProvider;
```

Thoughts?

djc: I don't think we'd want to make an exception for how we handle this import in formatting.

.github/workflows/build.yml
brody4hire: I have added some info here.

I totally share the adversion to complex shell scripts.

.github/workflows/build.yml
brody4hire: This is now gone, I have updated the clippy configuration as suggested. I think using this clippy configuration together with the `overwrite-atomic-sync-test` job should be good enough to validate this. A downstream fork can always do quick hacks & adaptations if needed while waiting for potential future regressions.

.github/workflows/build.yml
brody4hire: I have now removed this whole block. As I said in <https://github.com/rustls/rustls/pull/2285#discussion_r1924001033> I think it should be good enough  to test this scenario in `overwrite-atomic-sync-test`.

We could always add this kind of further testing in the future if needed.

admin/use-portable-atomic-arc
brody4hire: should be done

.github/workflows/build.yml
djc: I'd still prefer to skip all the added CI stuff except for the clippy disallowed types bits. I think the latter is sufficient to help us keep this working, and I'd rather avoid us maintaining all the cargo/bash stuff that patch rustls in CI.

.github/workflows/build.yml
```diff
@@ -192,6 +192,39 @@ jobs:
         run: cargo test --locked --release --no-run
         working-directory: rustls
 
+  overwrite-atomic-sync-test:
+    name: overwrite atomic_sync - build & test with portable_atomic_util::Arc
```
brody4hire, line 196: I think we should update this job name if we end up keeping it

.github/workflows/build.yml
brody4hire: I can remove this but thought there was some interest in improving some of this CI stuff from these comments like this: https://github.com/rustls/rustls/pull/2285#discussion_r1923792155

I did some work on improving this in response to other comments.

This also would have been a nice demonstration, as pointed out by verbiage suggested by @cpu: https://github.com/rustls/rustls/pull/2285#discussion_r1923812261

cpu: > I can remove this but thought there was some interest in improving some of this CI stuff from these comments like this: https://github.com/rustls/rustls/pull/2285#discussion_r1923792155

I made those comments in the spirit of improving the implementation if consensus was that it was the right overall direction. I've personally been swayed by djc's argument that it makes the most sense to not have it in this repo. I think @ctz is also supportive of that direction (?)

Sorry if that made some extra work.

brody4hire: I will remove the CI additions then. We can always add this in a separate PR, if (ever) desired. Yes "extra work" is sometimes normal part of the game, sometimes needed till we get a chance to make the final decision.

brody4hire: I have removed the CI additions and rebased. Please let me know if anything else is needed from me.

The CI additions would have been nice but unfortunately with some cost of additional complexity and potentially a little confusing for newcomers. Maybe a separate CI yml script with some better structure & more comments. We can always add this later if someone has extra time to look into this.

This would totally help another fork support these targets with minimal adjustments needed. I may do this with a few more things as well, supporting the idea in #2137 with `portable-io` for example.

rustls/src/lib.rs
```diff
@@ -401,6 +401,15 @@ mod log {
 #[macro_use]
 mod test_macros;
 
+/// This internal `atomic_sync` module aliases the `Arc` implementation to allow downstream forks
+/// of rustls targetting architectures without atomic pointers to replace the implementation
+/// with another implementation such as `portable_atomic_util::Arc` in one central location.
+mod sync {
+    // NOTE: update here may require update in `admin/use-portable-atomic-arc`.
```
cpu, line 408: I think this comment is out of date now.

cpu (approved):

rustls/src/lib.rs
brody4hire: Dang I just removed & force-pushed again thanks

.clippy.toml
```diff
@@ -1 +1,6 @@
 upper-case-acronyms-aggressive = true
+
+# only intended to affect main rustls crate
+disallowed-types = [
+  { path = "std::sync::Arc", reason = "must use Arc from atomic_sync module to support downstream forks targetting architectures without atomic ptrs" },
```
djc, line 5: "atomic_sync" -> "sync"

.github/workflows/build.yml
```diff
@@ -464,9 +464,13 @@ jobs:
         with:
           components: clippy
       # - we want to be free of any warnings, so deny them
-      # - disable incompatible_msrv as it does not understand that we apply our
+      # - allow clippy::disallowed_types as workaround since disallowed_types configured in .clippy.toml
+      #   is only intended for main rustls crate
+      # - allow clippy::incompatible_msrv as it does not understand that we apply our
       #   MSRV to the just the core crate.
-      - run: ./admin/clippy -- --deny warnings --allow clippy::incompatible_msrv
+      - run: ./admin/clippy -- --deny warnings --allow clippy::disallowed_types --allow clippy::incompatible_msrv
```
djc, line 471: Why do we need these changes?

rustls/src/lib.rs
```diff
@@ -401,6 +401,14 @@ mod log {
 #[macro_use]
 mod test_macros;
 
+/// This internal `atomic_sync` module aliases the `Arc` implementation to allow downstream forks
```
djc, line 404: `atomic_sync` -> `sync`

djc (commented):
This is looking much better, some small nits remaining.

rustls/src/atomic_sync.rs
brody4hire, line 1: need to remove this now empty file, apologies

.clippy.toml
```diff
@@ -1 +1,6 @@
 upper-case-acronyms-aggressive = true
+
+# only intended to affect main rustls crate - need to allow disallowed-types in all other crates in Clippy CI job
+disallowed-types = [
+  { path = "std::sync::Arc", reason = "must use Arc from sync module to support downstream forks targetting architectures without atomic ptrs" },
+]
```
brody4hire, line 6: I was hoping to configure this `disallowed-types` *only* for the main rustls crate & remove the workaround I made to ignore this for all other crates in CI. I tried configuring this in `[lints.clippy]` in `rustls/Cargo.toml` but to no avail. Any ideas or pointers?

.github/workflows/build.yml
brody4hire: As I tried to explain in the comments, this is a workaround since I couldn't figure out how to configure the `disallowed-types` *only* for the main rustls crate. This is definitely ugly but I couldn't figure out another solution. Any ideas or pointers would be much appreciated on this.

We _could_ also remove this in favor of configuring `disallowed-types = "allow"` in Config.toml for all other crates. This would mean quite a few more changes, which I would not likely favor from my side.

I would love to see this supported on the Rust Cargo side.

.github/workflows/build.yml
```diff
@@ -464,9 +464,13 @@ jobs:
...
           components: clippy
       # - we want to be free of any warnings, so deny them
-      # - disable incompatible_msrv as it does not understand that we apply our
+      # - allow clippy::disallowed_types as a workaround since disallowed_types configured in .clippy.toml
+      #   is only intended for main rustls crate
+      # - allow clippy::incompatible_msrv as it does not understand that we apply our
       #   MSRV to the just the core crate.
-      - run: ./admin/clippy -- --deny warnings --allow clippy::incompatible_msrv
+      - run: ./admin/clippy -- --deny warnings --allow clippy::disallowed_types --allow clippy::incompatible_msrv
+      # - now deny all warnings in main crate
```
brody4hire, line 472:
```suggestion
      # - keep the main crate free of all Clippy warnings
```

.github/workflows/build.yml
brody4hire: So sorry I am not sure I understand your comment 100%. I just pushed an update to add a TODO comment. I am thinking it may be easiest if you can contribute the required verbiage, like cpu did for some other comments, and put your name in the commit co-authored-by lines.

.github/workflows/build.yml
```diff
@@ -463,10 +463,12 @@ jobs:
...
-      # - we want to be free of any warnings, so deny them
-      # - disable incompatible_msrv as it does not understand that we apply our
-      #   MSRV to the just the core crate.
-      - run: ./admin/clippy -- --deny warnings --allow clippy::incompatible_msrv
+      # We want to be free of any warnings, so deny them.
+      # - Allow `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
+      - run: ./admin/clippy -- --deny warnings --allow clippy::disallowed_types
+      # - keep the main crate free of all Clippy warnings
+      - run: cargo clippy -p rustls -- --deny warnings
```
brody4hire, line 471: I think we need to finish review & hopeful merging of other PR #2310 first, to keep that update separate.

.github/workflows/build.yml
```diff
@@ -485,8 +487,12 @@ jobs:
...
         with:
           components: clippy
-      # do not deny warnings, as nightly clippy sometimes has false negatives
-      - run: ./admin/clippy
+      # Do not deny warnings, as nightly clippy sometimes has false negatives.
+      # - Ignore `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
+      - run: ./admin/clippy -- --allow clippy::disallowed_types # --allow clippy::incompatible_msrv
+      # Check the main crate for all Clippy nightly warnings; do not deny.
+      - run: cargo clippy -p rustls
```
brody4hire, line 495: I am proposing this update assuming that my proposal in PR #2311 will be rejected for now. As much as I would like to have CI deny warnings from Clippy nightly, I think this needs consensus from the @rustls maintainers and would rather not wait for this.

brody4hire (commented):
I think these updates should be ready, would like to wait for PR #2310 to keep that update separate as I stated below.

.github/workflows/build.yml
```diff
@@ -463,10 +463,12 @@ jobs:
...
           components: clippy
-      # - we want to be free of any warnings, so deny them
-      # - disable incompatible_msrv as it does not understand that we apply our
-      #   MSRV to the just the core crate.
-      - run: ./admin/clippy -- --deny warnings --allow clippy::incompatible_msrv
+      # We want to be free of any warnings, so deny them.
+      # - Allow `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
+      - run: ./admin/clippy -- --deny warnings --allow clippy::disallowed_types
+      # - keep the main crate free of all Clippy warnings
```
brody4hire, line 470: nit for consistency with other comments here:
```suggestion
      # - Keep the main crate free of all Clippy warnings.
```

.github/workflows/build.yml
```diff
@@ -485,8 +487,12 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           components: clippy
-      # do not deny warnings, as nightly clippy sometimes has false negatives
-      - run: ./admin/clippy
+      # Do not deny warnings, as nightly clippy sometimes has false negatives.
+      # - Ignore `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
+      - run: ./admin/clippy -- --allow clippy::disallowed_types # --allow clippy::incompatible_msrv
```
brody4hire, line 493:
```suggestion
      - run: ./admin/clippy -- --allow clippy::disallowed_types
```

(oops I almost missed this one)

.github/workflows/build.yml
```diff
@@ -485,8 +487,12 @@ jobs:
...
+      # Check the main crate for all Clippy nightly warnings; do not deny.
```
brody4hire, line 494:
```suggestion
      # Check the main crate for all Clippy nightly warnings, but do not deny them.
```

(another nit, I think this would give us slightly better consistency.)

.github/workflows/build.yml
brody4hire, line 495: Side question: these updates make me start to kinda wonder what is the point of keeping Clippy nightly as a CI job.

I think it may be ideal to keep this Clippy nightly CI job and deny only a subset of new Clippy nightly warnings, maybe just deny certain new Clippy nightly warnings as we see them suggest updating our code?

brody4hire (commented):
I just found a few more nits & added a side question. I should be able to resolve the nits in the next few hours or so.

.github/workflows/build.yml
djc: I think they serve us well -- we will often notice them even as warnings only, which will avoid them breaking CI when new clippy lints are stabilized.

brody4hire: Thanks `djc` that does make total sense to me. My apologies.

The one thing I want to say in my own defense is that I generally like to understand anything that I propose changing as much as I can, and like to clearly question in order to understand anything that doesn't make 100% sense to me. But yes I think I was (a little) out of line with the questioning and don't think my tone was right, my apologies.

cpu:
I'm confused by the state of this PR. It seemed like we whittled it down to 1 commit and a general approach that was OK to myself and Ctz, but now the commit history is a bit of a mess and it's in draft status.

What happened? Can you get this back to just the scope of the PR title and the minimum number of commits to implement exclusively what's needed for supporting a fork replacing the `Arc` impl?

brody4hire:
I will rebase this in a few minutes, my apologies. This kind of draft mess is what happens when I do too much mindless multi-tasking.

.github/workflows/build.yml
```diff
@@ -464,7 +464,11 @@ jobs:
         with:
           components: clippy
       # We want to be free of any warnings, so deny them.
-      - run: ./admin/clippy -- --deny warnings
+      # - Allow `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
```
brody4hire, line 468: minor nit, I will fix this while rebasing:

```suggestion
      #   is only intended for the main `rustls` crate.
```

brody4hire:
I have slightly updated the title, and I think this should be ready now. @rustls maintainers please DO feel free to update the title of this PR & the title of the commit message or any other possible nits you may see when merging, and give yourself co-authored-by credit in the commit message.

I have rebased this now that CI update needed from #2310 is merged, and I made a few updates only to cleanup some comments in the CI changes (and remove a commented-out option that was removed in other PR #2310). Unfortunately some workarounds were needed in the CI due to the configuration I made in `.clippy.toml` since I was not able to figure out a better solution. I would be happy if anyone can find a better solution either today or at some point in the future for this.

My intention has been to keep the proposed updates exclusively within the state of the PR title, and I have made a bit of effort to split some other updates out to separate PRs to keep these updates as focused & atomic as possible. Yes quite a bit of work over the past few months but I am really happy if we can improve support for the extra-tiny embedded devices as proposed here. I would also like to thank the @rustls maintainers for your efforts on reviewing my work and for supporting all of the @rustls requirements in general over the years.

cpu (approved):
Thanks. I updated the PR description.

brody4hire:
@cpu I just updated the description again, fixed the module name, added a note in parens: "(A fork would also need to update import of `Arc` in `rustls/tests/common/mod.rs`.)" & put "replaces #2200" in parens. Please feel free to edit again if needed.

I DO think that the updated description is what we should put in the commit message and would be happy to rebase again with this updated description if needed (your decision).

I do also think that the info concerning where a fork would need to update the import in `rustls/tests` should also be in the commit message. I think it would be ideal if we could document this info somewhere such as adding a new `DEVELOPING.md` file and would be happy to add this if you think it is needed. But I would rather not hold up the final merge too much longer.

cpu:
> just updated the description again: I fixed the module name, added a note in parens: "(A fork would also need to update import of Arc in rustls/tests/common/mod.rs.)" & put "replaces https://github.com/rustls/rustls/pull/2200" in parens. Please feel free to edit again if needed.

Those edits seem fine. Thanks.

> I DO think that the updated description is what we should put in the commit message and would be happy to rebase again with this updated description if needed (your decision).

Agreed, update the commit message to match please.

> I do also think that the info concerning where a fork would need to update the import in rustls/tests should also be in the commit message. 

Sure, sounds reasonable.

> I think it would be ideal if we could document this info somewhere such as adding a new DEVELOPING.md file and would be happy to add this if you think it is needed. But I would rather not hold up the final merge too much longer.

We have a `CONTRIBUTING.md` but I don't think this is a broadly applicable enough concern to be worth adding. I expect one or two downstream projects would need to know about this, not the average Rustls contributor.

.github/workflows/build.yml
```diff
@@ -464,7 +464,11 @@ jobs:
         with:
           components: clippy
       # We want to be free of any warnings, so deny them.
-      - run: ./admin/clippy -- --deny warnings
+      # - Allow `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for the main `rustls` crate.
+      - run: ./admin/clippy -- --deny warnings --allow clippy::disallowed_types
+      # - Keep the main crate free of all Clippy warnings.
```
brody4hire, line 470:
```suggestion
      # - Keep the main crate free of all warnings.
```

.github/workflows/build.yml
```diff
@@ -483,8 +487,12 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           components: clippy
-      # do not deny warnings, as nightly clippy sometimes has false negatives
-      - run: ./admin/clippy
+      # Do not deny warnings, as nightly clippy sometimes has false negatives.
+      # - Ignore `clippy::disallowed_types` as a workaround since `disallowed_types` configured in `.clippy.toml`
+      #   is only intended for main `rustls` crate.
+      - run: ./admin/clippy -- --allow clippy::disallowed_types
+      # Check the main crate for all Clippy nightly warnings, but do not deny them.
```
brody4hire, line 494:
```suggestion
      # Check the main crate for all nightly warnings, but do not deny them.
```

brody4hire (commented):
I would like to fix these minor nits in the CI comments when I push an update to update the commit message. Apologies for so much noise in this PR 😅😅

brody4hire:
I have updated the title slightly yet again, made a couple more minor updates to the description & pushed an updated commit with the updated description (reformatted to fit decent git line width 75-77 characters). I did touch a couple CI comments again in the following force-push diff:

- https://github.com/rustls/rustls/compare/a067a3f1183c75dd808ef92d625b31d78372f765..f8de6aaa234bd8ae4cbe4a2c1967141946885e2a

I have also added `---` separators to both the PR description & commit to separate the Resolves message, etc. I did keep the resolves message but NOT the reference to the "replaces" rejected PR in the commit message, also keeping of course the co-authored-by line in the commit message.

I think this should be good now. @cpu & other @rustls maintainers please feel free to fix-up anything else you may see if ready to merge & of course keep yourself in the co-authored-by credit. Thanks again for all of your review efforts.

ctz (approved):
FWIW I checked CI broke when I added a use of `alloc::sync::Arc`

PR #2286, opened 2024-12-18
Take aws-lc-rs 1.12

ctz:
This is mostly about undoing hacks we had because rustls-post-quantum was incompatible with the `rustls/fips` feature.


cpu (approved):
Nice

djc (approved):

PR #2288, opened 2024-12-19
Move rustls-post-quantum into the core crate

ctz:
The goal here is:

- move the important parts of rustls-post-quantum into the core crate
- add X25519MLKEM768 to the default key exchange algorithms for the aws-lc-rs provider (note, MLKEM768 is not included alone, for reasons given in the commit message, but remains available separately)
- rustls-post-quantum merely reexports things it previously contained. This will be the final release of rustls-post-quantum.


ctz:
> --- failure function_missing: pub fn removed or renamed ---
> --- failure pub_static_missing: pub static is missing ---

I think this is https://github.com/obi1kenobi/cargo-semver-checks/issues/573 -- I'm pretty sure this is semver compatible, otherwise the "semver trick" could never work.

ctz:

😅

rustls/src/crypto/aws_lc_rs/pq/mod.rs
```diff
@@ -1,12 +1,12 @@
-//! This crate provides a [`rustls::crypto::CryptoProvider`] that includes
-//! a hybrid[^1], post-quantum-secure[^2] key exchange algorithm --
-//! specifically [X25519MLKEM768], as well as a non-hybrid
+//! This module provides a hybrid[^1], post-quantum-secure[^2] key exchange
```
ctz, line 1: Have realised these docs are not visible. I'll think about where to move them.

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -224,11 +226,32 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
...
 pub mod kx_group {
     pub use super::kx::{SECP256R1, SECP384R1, X25519};
+    pub use super::pq::{MLKEM768, X25519MLKEM768};
 }
 
-pub use kx::ALL_KX_GROUPS;
+/// A list of the default key exchange groups supported by this provider.
+///
+/// This does not contain MLKEM768; by default MLKEM768 is only offered
+/// in hybrid with X25519.
```
dconnolly, line 238: re the comment message: 

> This maintains the existing `ALL_KX_GROUPS` and
introduces `DEFAULT_KX_GROUPS`.  These are different
because (for now) `DEFAULT_KX_GROUPS` does not contain
ML-KEM-768.

> That is out of abundance of caution -- in case ML-KEM
does not fare well in its first years of being (partially)
load-bearing for internet traffic.  This seems the most
conservative choice.

Fair enough. There is (at time of writing) activity in TLSWG to adopt and possibly make the MLKEM-only key agreements more 'real' and on track to become an RFC than they currently are, including possibly 'Recommended=Y': if that happens I may raise shuffling the `DEFAULT_KX_GROUPS` around again 😅

rustls/src/crypto/aws_lc_rs/pq/mod.rs
```diff
@@ -62,3 +62,44 @@ const INVALID_KEY_SHARE: Error = Error::PeerMisbehaved(PeerMisbehaved::InvalidKe
 const X25519_LEN: usize = 32;
 const MLKEM768_CIPHERTEXT_LEN: usize = 1088;
 const MLKEM768_ENCAP_LEN: usize = 1184;
+
+#[cfg(bench)]
```
djc, line 66: Why inline these here instead of keeping them in a `benches/` module?

djc (commented):
Very exciting stuff!

So if I understand correctly, the aws-lc-rs default provider will now prioritize X25519MLKEM768 as the key exchange, and that will have a substantial effect on how handshakes will be conducted in the wild. I wonder whether, despite the careful API compatibility, this is the kind of thing that might merit a semver-incompatible release with clear notice in the release notes?

Once it's been deployed for a while, we could bring 0.23 up too, but this feels like a big change with substantial risk at breakage.

rustls/src/crypto/aws_lc_rs/pq/mod.rs
cpu: My first thoughts are in `rustls/src/crypto/mod.rs`, or maybe a manual page ala FIPS.md?

rustls/src/crypto/aws_lc_rs/pq/mlkem.rs
```diff
@@ -50,6 +50,14 @@ impl SupportedKxGroup for MlKem768 {
         NamedGroup::MLKEM768
     }
 
+    fn fips(&self) -> bool {
+        // AUDITORS:
+        // At the time of writing, the ML-KEM implementation in AWS-LC-FIPS module 3.0
+        // is FIPS-pending.  Some regulatory regimes (eg, FedRAMP rev 5 SC-13) allow
+        // use of implementations in this state, as if they are already approved.
+        super::super::fips()
```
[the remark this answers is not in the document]
cpu: > In the next breaking release, should we make fips() return a tri-state enum? (And if so, file an issue about that?)

That feels reasonable to me. It's not a 1:1 comparison but the Go FIPS [module selection logic](https://github.com/golang/go/issues/70200#issuecomment-2490017956) exposes a "inprocess" choice for folks that have flexibility and don't want to wait for the tremendously slow certification process to complete to use the good stuff.

cpu (commented):
> I wonder whether, despite the careful API compatibility, this is the kind of thing that might merit a semver-incompatible release with clear notice in the release notes? Or just start by adding a simple way to enable this without affecting the default providers? (Like crypto::aws_lc_rs::post_quantum_provider().)

I'm optimistic, but share some of Djc's concerns about whether promoting the hybrid kx to default in a point release is going to cause trouble.

It feels like the separate `rustls-post-quantum` crate probably didn't get a lot of exposure/testing and so I see some appeal in a more gradual process where we make post-qc options more accessible (maybe a crate feature to opt-in to default post-qc without code change?) while concurrently collecting up more work to justify a semver-incompatible release. Hopefully by the time we have a good set of changes for a version bump we'll also have more downstream experience with post-qc and can retire the feature flag and make it a default.

djc:
Oh, I also like a Cargo feature as a mechanism to enable this, and something that we can easily shift from opt-in to opt-out over time.

rustls/src/crypto/ring/kx.rs
```diff
@@ -118,7 +118,8 @@ fn uncompressed_point(point: &[u8]) -> bool {
     matches!(point.first(), Some(0x04))
 }
 
-/// A list of all the key exchange groups supported by rustls.
+/// A list of all the key exchange groups supported by this provider.
+#[allow(unreachable_pub, dead_code)]
```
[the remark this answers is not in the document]
ctz: I'll try something different here. cfg won't work, because this file is instantiated twice.

rustls/src/crypto/aws_lc_rs/pq/mod.rs
ctz: This is aligned with the other benchmarks that live in `src/crypto/ring/kx.rs` in an inline `mod benchmarks`. There's no reason they actually have to be written in "unit benchmark" style, since everything they test are public APIs. But putting them there means they get duplicated easily for both providers.

djc: Fair enough.

ctz:
> Oh, I also like a Cargo feature as a mechanism to enable this, and something that we can easily shift from opt-in to opt-out over time.

I've added a commit to do that. Here's what I'm thinking about a release note for this:

* **New crate feature: `post-quantum`**: setting this feature adjusts the `aws-lc-rs` provider's `DEFAULT_KX_GROUPS` order, so `X25519MLKEM768` is the most preferred key exchange algorithm. This has some performance impact: handshakes which offer TLS1.3 are larger and slower.
 
  Without this feature, `X25519MLKEM768` is the least preferred (so can still be used, if the server prefers).

  We expect to add this feature to the crate's default features in a future minor release.

  The `rustls-post-quantum` crate now depends on the core `rustls` crate and activates this feature.

ctz:
(I'm also aware that the commit ordering of this PR no longer makes much sense; it's on my TODO list.)

rustls/src/crypto/aws_lc_rs/pq/mlkem.rs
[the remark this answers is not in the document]
ctz: I've adjusted the comment here and filed #2309 for future.

rustls/src/crypto/aws_lc_rs/pq/mod.rs
ctz: I have moved this into the "defaults" page of the manual.

ctz:
I think this is ready for a second look. Rough changes since last time:

- controlling the priority of X25519MLKEM768 under a crate feature, called `prefer-post-quantum`. the `rustls-post-quantum` crate sets this feature, but otherwise it is not set by default.
- included a reference to the SP800-227 draft that was recently released
- fix a couple of HRR bugs found by bogo when X25519MLKEM768 is highest priority

rustls/src/crypto/aws_lc_rs/pq/mlkem.rs
```diff
@@ -50,6 +50,21 @@ impl SupportedKxGroup for MlKem768 {
...
+    fn fips(&self) -> bool {
+        // AUDITORS:
+        // At the time of writing, the ML-KEM implementation in AWS-LC-FIPS module 3.0
+        // is FIPS-pending.  Some regulatory regimes (eg, FedRAMP rev 5 SC-13) allow
+        // use of implementations in this state, as if they are already approved.
+        //
+        // We follow this liberal interpretation, and say MlKem768 is FIPS-compliant
+        // if the underlying library is in FIPS mode.
+        //
+        // TODO: adjust the `fips()` function return type to allow more policies to
```
djc, line 62: Nit: link the new issue you created yesterday?

rustls-post-quantum/README.md
```diff
@@ -8,64 +8,8 @@ Rustls is a modern TLS library written in Rust.
...
-    kx_groups: vec![
-        rustls_post_quantum::X25519MLKEM768,
-        aws_lc_rs::kx_group::X25519,
-        rustls_post_quantum::MLKEM768,
-    ],
-    ..parent
-};
-```
-
+This crate is vestigal, as post-quantum key exchange support
```
djc, line 11: "vestigal" -> "vestigial" but also maybe just avoid that word which is likely to fall outside the vocabulary of at least some of the audience here, and instead give a plain language recommendation as to what to do? Like: "Functionality of this crate became part of the rustls library since the 0.23.22 release. When using that version of the crate, use the `prefer-post-quantum` Cargo feature to control whether to prefer using post-quantum algorithms instead of using this crate."

rustls-post-quantum/src/lib.rs
```diff
@@ -1,2 +1,5 @@
+//! This crate is vestigal, as post-quantum key exchange support
```
djc, line 1: (As in the README.)

rustls/src/manual/defaults.rs
```diff
@@ -26,4 +26,42 @@ the implementation security will be improved.  We think this is an uncommon case
 Both provide roughly the same classical security level, but x25519 has better performance and
 it's _much_ more likely that both peers will have good quality implementations.
 
+### How can use the post-quantum-secure key exchange `X25519MLKEM768`?
```
djc, line 29: "How can use the" reads awkwardly.

rustls/src/manual/defaults.rs
```diff
@@ -26,4 +26,42 @@ the implementation security will be improved.  We think this is an uncommon case
...
+
+[`X25519MLKEM768`] -- a hybrid[^1], post-quantum-secure[^2] key exchange
+algorithm -- is available when using the aws-lc-rs provider.
+
+The `post-quantum` crate feature makes `X25519MLKEM768` the
```
djc, line 34: Feature flag name changed?

rustls/src/manual/features.rs
```diff
@@ -36,6 +37,7 @@ APIs ([`CryptoProvider`] for example).
       in browsers.  It is also not supported by the WebPKI, because the
       CA/Browser Forum Baseline Requirements do not support it for publicly
       trusted certificates.
+[^2]: See [the documentation][crate::manual::_05_defaults#how-can-use-the-post-quantum-secure-key-exchange-x25519mlkem768]
```
djc, line 40: (This link will change if you update the header.)

djc (approved):
Nice!

rustls-post-quantum/README.md
cpu: I was going to leave the same feedback about mentioning the `prefer-post-quantum` flag for folks transitioning from the legacy crate. :+1:

rustls/src/manual/defaults.rs
```diff
@@ -26,4 +26,42 @@ the implementation security will be improved.  We think this is an uncommon case
...
+for example, by [Chrome] and [Cloudflare].
+
+You may see unexpected connection failures (such as [tldr.fail])
+-- [please report these to us][interop-bug].
+
+The two components of this key exchange are well regarded:
+X25519 alone is already used by default by rustls, and tends to have
+higher quality implementations than other elliptic curves.
+ML-KEM-768 was standardized by NIST in [FIPS203].
+
```
cpu, line 48: WDYT about adding some mention that pure ML-KEM isn't in the default set? Could probably just reuse some text from 3f674313c421dcecfb1e70125b121305878eb6ba's commit message.

rustls/src/manual/features.rs
```diff
@@ -12,6 +12,7 @@ APIs ([`CryptoProvider`] for example).
 * ECDSA, Ed25519 or RSA server authentication by clients `*`
 * ECDSA, Ed25519[^1] or RSA server authentication by servers `*`
 * Forward secrecy using ECDHE; with curve25519, nistp256 or nistp384 curves `*`
+* Post-quantum hybrid key exchange with [X25519MLKEM768](https://datatracker.ietf.org/doc/draft-kwiatkowski-tls-ecdhe-mlkem/) [^2] `*`
```
cpu, line 15: :tada:

cpu (approved):
Great work :stars:

rustls-post-quantum/README.md
ctz: Adopted this sort of language.

ctz:
I think I've addressed those comments -- thanks. I also removed the link to `rustls-post-quantum` from the list of alternate providers.

djc (approved):

djc:
I'm confused why the semver compatibility lints are failing here. Maybe send some feedback upstream?

ctz:
> I'm confused why the semver compatibility lints are failing here. Maybe send some feedback upstream?

This is tracked upstream as https://github.com/obi1kenobi/cargo-semver-checks/issues/638

PR #2293, opened 2024-12-29
rustls-post-quantum: 0.2.1 -> 0.2.2

cpu:
Releases a crate update that takes aws-lc-rs 1.12 without the 'non-fips' feature flag, and the stable KEM API.

There's a separate plan in-flight in https://github.com/rustls/rustls/pull/2288 to move `rustls-post-quantum` into the core crate, and to publish a stub re-exporting from `rustls` as the final `rustls-post-quantum@0.2.2` release.

However, for https://github.com/rustls/rustls-ffi/issues/507 it would be helpful if there was a `rustls-post-quantum` I could use today that doesn't activate `non-fips` in `aws-lc-rs`. When we run `cbindgen` in that crate we enable all features so we can translate feature-conditional APIs into equivalent C `#if defined ...` guards in our `.h`. That makes a feature flag that activates post-quantum support tricky, because fips & rustls-post-quantum 0.2.1 are mutually exclusive. The required work already landed in https://github.com/rustls/rustls/pull/2286, but wasn't published.

Ultimately I can also just wait for https://github.com/rustls/rustls/pull/2288 to be finalized. WDYT @ctz?

### Proposed release notes

Updates the `aws-lc-rs` dependency to [1.12](https://github.com/aws/aws-lc-rs/releases/tag/v1.12.0). Notably this release switches to a stable ML-KEM API, and includes a [FIPS module with ML-KEM support](https://aws.amazon.com/blogs/security/aws-lc-fips-3-0-first-cryptographic-library-to-include-ml-kem-in-fips-140-3-validation/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2558,7 +2558,7 @@ checksum = "d2bf47e6ff922db3825eb750c4e2ff784c6ff8fb9e13046ef6a1d1c5401b0b37"
 
 [[package]]
 name = "rustls-post-quantum"
-version = "0.2.1"
+version = "0.2.2"
 dependencies = [
  "aws-lc-rs",
  "criterion",
--- a/rustls-post-quantum/Cargo.toml
+++ b/rustls-post-quantum/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-post-quantum"
-version = "0.2.1"
+version = "0.2.2"
 edition = "2021"
 rust-version = "1.71"
 license = "Apache-2.0 OR ISC OR MIT"


djc (approved):

djc:
Feels like some version of #2288 is pretty close, so depends on how urgent/important your work in rustls-ffi is?

djc (approved):

djc:
How do we feel conceptually about including stuff in rustls-ffi that's explicitly not part of rustls? I guess we may already have that for verifier stuff?

ctz (approved):

cpu:
> How do we feel conceptually about including stuff in rustls-ffi that's explicitly not part of rustls? I guess we may already have that for verifier stuff?

Good question. I don't think there's been any discussion on that. 

As you mentioned `rustls-platform-verifier` is being brought in by default since 0.14. The to-be-released 0.15 also adds an optional feature for cert compression that brings along zlib-rs and brotli2. Both feel OK to me since in general the state of C dependency management makes separating those things out feel very unappealing. In contrast I think Rust benefits from a richer ecosystem where it makes more sense for `rustls` and `rustls-platform-verifier` to be separate.

For `rustls-post-quantum` in particular I don't have any hesitation since it seems clear that it's going to be part of `rustls` itself soon :tm: 

> Feels like some version of https://github.com/rustls/rustls/pull/2288 is pretty close, so depends on how urgent/important your work in rustls-ffi is?

Thinking about what's left in https://github.com/rustls/rustls-ffi/pull/518 I think it's likely that #2288 will be finished first. 

Maybe using a cargo patch would be sufficient for now and I could revive this branch if 2288 ends up taking more than a couple weeks to land.

Let's go with that plan unless there's any extra input from Ctz RE: his availability for 2288.

PR #2294, opened 2024-12-30
Server fuzzing harness improvements

ctz:
First commit aims to fix https://oss-fuzz.com/testcase-detail/4979312264019968 introduced in 9fd25dcf49ad2ec346c40b14137142f998a32c23, aka

```
thread '<unnamed>' panicked at fuzzers/server.rs:53:47:
called `Result::unwrap()` on an `Err` value: Custom { kind: InvalidData, error: "message buffer full" }
==407== ERROR: libFuzzer: deadly signal
```

Second is to cover `Accepted::into_connection`

### DIFF
--- a/fuzz/fuzzers/server.rs
+++ b/fuzz/fuzzers/server.rs
@@ -50,7 +50,10 @@ fn fuzz_acceptor_api(data: &[u8]) {
     let mut stream = io::Cursor::new(data);
 
     loop {
-        let rd = server.read_tls(&mut stream).unwrap();
+        let rd = server
+            .read_tls(&mut stream)
+            .unwrap_or(0);
+
         match server.accept() {
             Ok(Some(_)) | Err(_) => {
                 break;
--- a/fuzz/fuzzers/server.rs
+++ b/fuzz/fuzzers/server.rs
@@ -6,7 +6,7 @@ extern crate rustls;
 use std::io;
 use std::sync::Arc;
 
-use rustls::server::Acceptor;
+use rustls::server::{Accepted, Acceptor};
 use rustls::{ServerConfig, ServerConnection};
 
 fuzz_target!(|data: &[u8]| {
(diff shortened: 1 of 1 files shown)


cpu (approved):
> first commit aims to fix https://oss-fuzz.com/testcase-detail/4979312264019968 

hmm. Did this get flagged to you by a notification of some kind? I thought I added my email to the CC list but haven't seen any messages come through.

ctz:
> hmm. Did this get flagged to you by a notification of some kind? I thought I added my email to the CC list but haven't seen any messages come through.

I didn't get a notification, and don't exactly know how this works. I _think_ there is a process going from finding a crash, to minimising it, and then finally filing an issue -- over successive days. I think we only get notifications when an issue is filed? [There is not yet one for this issue](https://issues.oss-fuzz.com/issues?q=rustls).

(I not looking at oss-fuzz regularly, I went looking today to see what coverage we were getting, as part of recreating and reinstating https://github.com/rustls/rustls-fuzzing-corpora)

djc:
I think there is a list of Google accounts in Git somewhere that get notified?

cpu:
> I think there is a list of Google accounts in Git somewhere that get notified?

Yeah I think that was https://github.com/google/oss-fuzz/pull/11631

>  I think there is a process going from finding a crash, to minimising it, and then finally filing an issue -- over successive days. I think we only get notifications when an issue is filed?

Ahh ok that would explain it!

fuzz/fuzzers/server.rs
```diff
@@ -65,3 +56,37 @@ fn fuzz_acceptor_api(data: &[u8]) {
...
+            .with_no_client_auth()
+            .with_cert_resolver(rustls_fuzzing_provider::server_cert_resolver()),
+    ));
+
+    let mut server = match maybe_server {
+        Ok(conn) => conn,
+        Err(_) => return,
+    };
+
+    service_connection(stream, &mut server);
```
djc, line 74: Nit: in this case I think the `service_connection()` call fits on a single line within the `match` arm? Don't see much reason to break it out.

djc (approved):

PR #2299, opened 2025-01-08
docs: update @cpu maintainer status

cpu:
All good things come to an end :) Thanks to ISRG for their support!

For whatever it's worth I expect to have some part-time availability for Rustls-related work in the future if the right project/funder came along. My contact info is not hard to find.


ctz (approved):
🫡

PR #2303, opened 2025-01-16
Implement BufRead for Reader, Stream, and StreamOwned

goffrie:
This is basically a reroll of #870 against the latest version of the code.

New in this PR is an alternate method `Reader::into_first_chunk` (other naming suggestions welcome) that acts like `fill_buf` but returns a slice with a longer lifetime, at the cost of consuming `self` (needed in order to steal the `&mut ConnectionCommon` borrow). This makes it possible to implement AsyncBufRead in tokio_rustls without necessarily implementing BufRead directly on ConnectionCommon. It's also needed to implement BufRead for StreamOwned.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -25,7 +25,7 @@ mod connection {
     use alloc::vec::Vec;
     use core::fmt::Debug;
     use core::ops::{Deref, DerefMut};
-    use std::io;
+    use std::io::{self, BufRead, Read};
 
     use crate::common_state::{CommonState, IoState};
     use crate::error::Error;
@@ -47,7 +47,7 @@ mod connection {
         /// Read TLS content from `rd`.
         ///
         /// See [`ConnectionCommon::read_tls()`] for more information.
-        pub fn read_tls(&mut self, rd: &mut dyn io::Read) -> Result<usize, io::Error> {
+        pub fn read_tls(&mut self, rd: &mut dyn Read) -> Result<usize, io::Error> {
             match self {
                 Self::Client(conn) => conn.read_tls(rd),
                 Self::Server(conn) => conn.read_tls(rd),
@@ -108,7 +108,7 @@ mod connection {
         pub fn complete_io<T>(&mut self, io: &mut T) -> Result<(usize, usize), io::Error>
         where
             Self: Sized,
-            T: io::Read + io::Write,
+            T: Read + io::Write,
         {
             match self {
                 Self::Client(conn) => conn.complete_io(io),
@@ -205,7 +205,7 @@ mod connection {
         }
     }
 
-    impl io::Read for Reader<'_> {
+    impl Read for Reader<'_> {
         /// Obtain plaintext data received from the peer over this TLS connection.
         ///
         /// If the peer closes the TLS session cleanly, this returns `Ok(0)`  once all
@@ -268,6 +268,29 @@ mod connection {
             }
 
             self.check_no_bytes_state()
+        }
+    }
+
+    impl BufRead for Reader<'_> {
+        /// Obtain a chunk of plaintext data received from the peer over this TLS connection.
+        /// This reads the same data as [`Reader::read()`], but returns a reference instead of
+        /// copying the data.
+        ///
+        /// The caller should call [`Reader::consume()`] afterward to advance the buffer.
+        ///
+        /// See [`Reader::into_first_chunk()`] for a version of this function that returns a
+        /// buffer with a longer lifetime.
+        fn fill_buf(&mut self) -> io::Result<&[u8]> {
+            Reader {
+                // reborrow
+                received_plaintext: self.received_plaintext,
+                ..*self
+            }
+            .into_first_chunk()
+        }
+
+        fn consume(&mut self, amt: usize) {
+            self.received_plaintext.consume(amt)
         }
     }
 
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -173,7 +173,7 @@ mod connection {
         pub(super) has_seen_eof: bool,
     }
 
-    impl Reader<'_> {
+    impl<'a> Reader<'a> {
         /// Check the connection's state if no bytes are available for reading.
         fn check_no_bytes_state(&self) -> io::Result<()> {
             match (self.has_received_close_notify, self.has_seen_eof) {
@@ -187,6 +187,20 @@ mod connection {
                 // connection still going, but needs more data: signal `WouldBlock` so that
                 // the caller knows this
                 (false, false) => Err(io::ErrorKind::WouldBlock.into()),
+            }
+        }
+
+        /// Obtain a chunk of plaintext data received from the peer over this TLS connection.
+        ///
+        /// This method consumes `self` so that it can return a slice whose lifetime is bounded by
+        /// the [`ConnectionCommon`] that created this `Reader`.
+        pub fn into_first_chunk(self) -> io::Result<&'a [u8]> {
+            match self.received_plaintext.chunk() {
+                Some(chunk) => Ok(chunk),
+                None => {
+                    self.check_no_bytes_state()?;
+                    Ok(&[])
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


rustls/src/conn.rs
```diff
@@ -257,6 +271,43 @@ mod connection {
         }
     }
 
+    impl BufRead for Reader<'_> {
+        /// Obtain a chunk of plaintext data received from the peer over this TLS connection.
```
djc, line 275: Where are all these docs coming from? Ideally we'd reference the other place rather than making copies at this level of detail (which easily falls out of sync).

rustls/src/conn.rs
```diff
@@ -189,6 +189,20 @@ mod connection {
                 (false, false) => Err(io::ErrorKind::WouldBlock.into()),
             }
         }
+
+        /// Obtain a chunk of plaintext data received from the peer over this TLS connection.
+        /// This does the same thing as [`Reader::fill_buf`], but unlike that method, this
+        /// method returns a slice whose lifetime is bounded by the [`ConnectionCommon`]
+        /// that created this `Reader`.
+        pub fn into_first_chunk(self) -> io::Result<&'a [u8]> {
```
djc, line 197: Why is this necessary/useful? It seems unrelated to the `ReadBuf` impl.

djc (commented):
Can you restate in the PR description what you are trying to achieve with this PR? Have you benchmarked your proposed tokio-rustls changes?

rustls/src/conn.rs
goffrie: Fair enough, it was copied from `Read::read` / `Read::read_buf` methods (which were already copy pasted)

rustls/src/conn.rs
goffrie: It's necessary to be able to implement `tokio::io::AsyncBufRead` in `tokio_rustls`, since otherwise `rustls::Reader::fill_buf` returns a buffer with too short of a lifetime (the lifetime of the `Reader`, which is constructed temporarily in tokio_rustls - not of the `ConnectionCommon`). Neither method is strictly more useful than the other though.

djc:
I'll let @ctz have a look at this first.

ctz (commented):
I think this looks fairly reasonable? Naturally it needs tests.

Mostly unrelated: I've also considered adding a `pub fn take_chunk(&mut self) -> io::Result<Option<Vec<u8>>>` to `Reader`, to allow move-style buffered IO (rather than the copy-style IO implied by all the `std::io` traits.) I don't think that helps or hinders with this addition though.

cpu:
>  Naturally it needs tests.

If this is motivated by a downstream need in tokio-rustls it would also be nice to see a draft implementing that side before this is merged.

goffrie:
Here is the downstream PR: https://github.com/rustls/tokio-rustls/pull/100

I will add tests.

goffrie:
I added a number of tests for BufRead behaviour and cleaned up the code a bit too. I've extended the PR to implement BufRead for Stream/StreamOwned as it seemed logical and some tests that I wanted to extend were relying on it.

goffrie:
This also provides a foundation to easily resolve https://github.com/rustls/tokio-rustls/issues/55 (without rewriting the whole thing to use the unbuffered API): https://github.com/rustls/tokio-rustls/commit/655c39656154ad38a4b634d34b596a92a084ee7c?w=1

rustls/src/vecbuf.rs
```diff
@@ -142,7 +142,14 @@ impl ChunkVecBuffer {
         Ok(offs)
     }
 
-    fn consume(&mut self, used: usize) {
+    /// Returns the first contiguous chunk of data, or None if empty.
```
djc, line 145: Nit: since this is a simpler method than both `consume()` and `write_to()`, please move it to the bottom of the impl block.

rustls/src/conn.rs
djc, line 197: Since this is not covered by the commit description ("Implement BufRead for Reader"), please split it into a separate commit.

rustls/src/conn.rs
djc: Should remove this from the original commit.

rustls/src/stream.rs
```diff
@@ -204,6 +226,21 @@ where
     }
 }
 
+impl<C, T, S> BufRead for StreamOwned<C, T>
```
djc, line 229: Nit: let's split this into a separate commit.

rustls/src/stream.rs
```diff
@@ -41,15 +41,8 @@ where
...
-}
 
-impl<'a, C, T, S> Read for Stream<'a, C, T>
-where
-    C: 'a + DerefMut + Deref<Target = ConnectionCommon<S>>,
-    T: 'a + Read + Write,
-    S: SideData,
-{
-    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+    fn prepare_read(&mut self) -> Result<()> {
```
djc, line 45: I'd like to see a separate commit for the extraction of `prepare_read()`, too. This commit ("Implement BufRead for Stream/StreamOwned") has too much going on right now.

rustls/src/stream.rs
```diff
@@ -62,24 +55,53 @@ where
...
-                break;
-            }
+impl<'a, C, T, S> BufRead for Stream<'a, C, T>
+where
+    C: 'a + DerefMut + Deref<Target = ConnectionCommon<S>>,
+    T: 'a + Read + Write,
+    S: 'a + SideData,
+{
+    fn fill_buf(&mut self) -> Result<&[u8]> {
+        Stream {
```
djc, line 96: Why do we need to reconstitute a `Stream` here?

rustls/src/conn.rs
```diff
@@ -294,13 +294,12 @@ mod connection {
...
-                None => {
-                    self.check_no_bytes_state()?;
-                    Ok(&[])
-                }
+            Reader {
+                // reborrow
+                received_plaintext: self.received_plaintext,
+                ..*self
             }
+            .into_first_chunk()
```
djc, line 289: I'd suggest using this commit to introduce `into_first_chunk()` itself.

rustls/src/stream.rs
goffrie: It reborrows, which is necessary since the helper takes `self` by value. I'll add a comment.

rustls/src/conn.rs
djc, line 197: Is there a reason this has to take `self` owned, instead of taking `&self`?

I don't think it makes sense to document this by pointing to `fill_buf()`, which doesn't exist yet and also feels like it creates a sort of circular reference. Just call out instead that the return value's lifetime extends to that of the underlying `ConnectionCommon`.

rustls/src/conn.rs
```diff
@@ -271,6 +271,29 @@ mod connection {
         }
     }
 
+    impl BufRead for Reader<'_> {
+        /// Obtain a chunk of plaintext data received from the peer over this TLS connection.
+        /// This reads the same data as [`Reader::read`], but returns a reference instead of
```
djc, line 276: Nit: please refer to functions and methods with trailing `()` in order to distinguish from fields.

djc (approved):
Nice, this now has a pretty clean commit history. Some minor questions left.

rustls/src/conn.rs
goffrie: Yes, because the inner reference is mutable. `&'a &'b mut T` can't convert to a `&'b T`, only `&'a T` (the mutable reference is only "temporarily" frozen for `'a`). This is unlike `&'a &'b T` where the inner `&'b T` can be copied out. Taking `self` by value instead consumes the `&'b mut T` borrow and lets us convert it into an immutable one.

The method could be written as taking `&'a self` but it would not be callable since the Reader is generally short lived and wouldn't be borrowable for `'a`.

goffrie:
I'm not sure if I should be worried about the benchmark results - it looks like it could just be random variation? I don't really see how this change could lead to a regression but only in two random handshake tests

cpu (approved):
Thanks! This is nice and tidy :sparkles: 

> I'm not sure if I should be worried about the benchmark results - it looks like it could just be random variation?

I think it's safe to ignore here.

ctz (approved):
Thanks!

PR #2305, opened 2025-01-20
Fix SSL 2 version constant to 0x0002

droe:
SSL 2 uses a version field of 0x0002, not 0x0200.  This is confirmed not only in the original Netscape spec [1] and RFC draft of the time [2], but also in major implementations such as OpenSSL [3] and Wireshark [4] as well as packet captures of SSL 2 traffic [5].

[1] https://www-archive.mozilla.org/projects/security/pki/nss/ssl/draft02.html
[2] https://datatracker.ietf.org/doc/html/draft-hickman-netscape-ssl-00
[3] https://github.com/openssl/openssl/blob/OpenSSL_0_9_6m/ssl/ssl2.h#L66-L71
[4] https://github.com/wireshark/wireshark/blob/release-4.4/epan/dissectors/packet-tls-utils.h#L266-L277
[5] https://github.com/Lekensteyn/wireshark-notes/blob/master/ssl3/sslv2.pcapng

### DIFF
--- a/rustls/src/enums.rs
+++ b/rustls/src/enums.rs
@@ -95,7 +95,7 @@ enum_builder! {
     /// The `Unknown` item is used when processing unrecognised ordinals.
     #[repr(u16)]
     pub enum ProtocolVersion {
-        SSLv2 => 0x0200,
+        SSLv2 => 0x0002,
         SSLv3 => 0x0300,
         TLSv1_0 => 0x0301,
         TLSv1_1 => 0x0302,


bdaehlie:
This seems correct. I checked BoringSSL myself:

#define SSL2_VERSION 0x0002

And confirmed the spec says:

#define SSL_CLIENT_VERSION			0x0002
#define SSL_SERVER_VERSION			0x0002

cpu (approved):
Thanks :+1: 

I think the incorrect value has been there since [the very first commit](https://github.com/rustls/rustls/commit/b24ddffcb9e9e162dda0f8dcfba5f83517963cd3): https://github.com/rustls/rustls/blob/b24ddffcb9e9e162dda0f8dcfba5f83517963cd3/src/proto_enums.rs#L35

Likely unnoticed since SSLv2 was never implemented here.

ctz:
Thanks! The SSLv2 item here probably doesn't belong there in retrospect, as it implies that SSLv2 frames share some sort of structure with SSL3 and later headers, which they don't.

droe:
@ctz In the context of v2 compatible handshakes (RFC 6101 Appendix E), when parsing a v2 ClientHello message, implementations did at some point have to read a version field that could have had the value 0x0002 or 0x030x.  In that limited sense, version 0x0002 was in the same code point namespace as the other version constants.

PR #2306, opened 2025-01-21
Incidental/preparatory parts of #2288

ctz:
This PR sloughs off the incidental parts of #2288, to reduce the size of that PR.

I think this is pretty uncontroversial, though note that (eg) `rustls::crypto::ring::DEFAULT_KX_GROUPS` is a new public item.

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/mod.rs
+++ b/rustls/src/crypto/aws_lc_rs/mod.rs
@@ -50,7 +50,7 @@ pub fn default_provider() -> CryptoProvider {
 fn default_kx_groups() -> Vec<&'static dyn SupportedKxGroup> {
     #[cfg(feature = "fips")]
     {
-        ALL_KX_GROUPS
+        DEFAULT_KX_GROUPS
             .iter()
             .filter(|cs| cs.fips())
             .copied()
@@ -58,7 +58,7 @@ fn default_kx_groups() -> Vec<&'static dyn SupportedKxGroup> {
     }
     #[cfg(not(feature = "fips"))]
     {
-        ALL_KX_GROUPS.to_vec()
+        DEFAULT_KX_GROUPS.to_vec()
     }
 }
 
@@ -224,9 +224,13 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
 /// All defined key exchange groups supported by aws-lc-rs appear in this module.
 ///
 /// [`ALL_KX_GROUPS`] is provided as an array of all of these values.
+/// [`DEFAULT_KX_GROUPS`] is provided as an array of this provider's defaults.
 pub mod kx_group {
     pub use super::kx::{SECP256R1, SECP384R1, X25519};
 }
+
+/// A list of the default key exchange groups supported by this provider.
+pub static DEFAULT_KX_GROUPS: &[&dyn SupportedKxGroup] = ALL_KX_GROUPS;
 
 /// A list of all the key exchange groups supported by this provider.
 pub static ALL_KX_GROUPS: &[&dyn SupportedKxGroup] =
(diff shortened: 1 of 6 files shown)


rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -224,16 +224,20 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
...
 pub mod kx_group {
     pub use super::kx::{SECP256R1, SECP384R1, X25519};
 }
 
-pub use kx::ALL_KX_GROUPS;
+/// A list of the default key exchange groups supported by this provider.
+pub static DEFAULT_KX_GROUPS: &[&dyn SupportedKxGroup] = kx::ALL_KX_GROUPS;
+
+/// A list of all the key exchange groups supported by this provider.
+pub static ALL_KX_GROUPS: &[&dyn SupportedKxGroup] = kx::ALL_KX_GROUPS;
```
djc, line 236: What is the reason for redefining `ALL_KX_GROUPS` here rather than redefining as before? Seems like that generates some potential for confusion.

djc (approved):

rustls/src/crypto/aws_lc_rs/mod.rs
ctz: This is attempting to avoid a `unreachable_pub` warning down the road. They way that happens is:

- for ring, `kx.rs` is included and has a `pub static ALL_KX_GROUPS ...`, and `pub use`s it directly because it contains all the kx groups. That works.
- for aws-lc-rs, `kx.rs` is included and has a `pub static ALL_KX_GROUPS ...`, but cannot use it directly because there are additional kx groups (the post-quantum ones) that need adding (prepending and appending) to it.

I guess an alternative is to remove `ALL_KX_GROUPS` from `kx.rs`? That doesn't seem so bad.

ctz: (previously: https://github.com/rustls/rustls/pull/2288#discussion_r1893654949)

ctz: > I guess an alternative is to remove ALL_KX_GROUPS from kx.rs? 

Did this, and added a note why to the commit message.

rustls/tests/unbuffered.rs
```diff
@@ -1422,3 +1314,123 @@ fn server_receives_incorrect_first_handshake_message() {
         _ => panic!("unexpected alert sending state"),
     };
 }
+
+const TLS12_CLIENT_TRANSCRIPT: &[&str] = &[
```
cpu, line 1318: Need a gut-check here: would it be overkill to try and lose the `Ok(..)` part of each of these while tidying up? It's a small bit of noise IMO since there's no `Err(..)` instances. It feels like something we could apply mechanically at the call-sites with a helper fn, but also like inconsequential nit pickery. This is a nice tidy on its own :+1:

cpu (approved):
Nice! Thanks for breaking this out.

rustls/tests/unbuffered.rs
djc: So I think these come from here:

```rust
fn advance_server(
    conn: &mut UnbufferedConnectionCommon<ServerConnectionData>,
    buffers: &mut Buffers,
    actions: Actions,
    transcript: &mut Vec<String>,
) -> State {
    let UnbufferedStatus { discard, state } = conn.process_tls_records(buffers.incoming.filled());

    transcript.push(format!("{:?}", state));

    let state = match state.unwrap() {
        ConnectionState::ReadEarlyData(mut state) => {
            let mut records = vec![];
            let mut peeked_len = state.peek_len();
```

I think it would be an improvement to `unwrap()` before we transcribe. 👍

cpu: Ahh yeah that seems like a better idea indeed. Thanks!

brody4hire (commented):
nit: with this many changes I wish we could have moved [features] to come before [dependencies] & [build-dependencies]. I think an investigative user would rather see [features] come first, and this is more consistent with some other crates such as portable-atomic & tokio. I wish I would have seen this earlier. Oh well.

PR #2307, opened 2025-01-21
chore: enforce taplo toml formatting

brody4hire:
This would have helped me with some of the toml updates needed for PR #2285.

I am proposing here that we keep the line width at 100, vertically expanding some of the dependency feature lists. This might have made it a little easier for me when adjusting some of the dependency features when working on the rejected draft PR #2200.

I would be happy to set the line width to 120 & revert many of the formatting updates I proposed, if needed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -506,6 +506,20 @@ jobs:
         working-directory: rustls/
         run: cargo check-external-types
 
+  taplo:
+      name: Taplo
+      runs-on: ubuntu-latest
+      steps:
+        - name: Checkout sources
+          uses: actions/checkout@v4
+          with:
+            persist-credentials: false
+
+        - name: Install rust toolchain
+          uses: dtolnay/rust-toolchain@stable
+        - run: cargo install taplo-cli --locked
+        - run: taplo format --check
+
   openssl-tests:
     name: Run openssl-tests
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ exclude = [
 ]
 
 default-members = [
+  # ---
   "examples",
   "rustls",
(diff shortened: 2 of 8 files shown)


djc:
While I appreciate automating consistency, IMO spreading some dependency's features over multiple lines makes `Cargo.toml` harder to read rather than easier, so I don't really like this change as currently submitted.

djc (approved):

Cargo.toml
```diff
@@ -31,6 +31,7 @@ exclude = [
 ]
 
 default-members = [
+  # ---
```
ctz, line 34: what do these comments mean?

ctz: Ah, makes sense. I think this is OK.

rustls/Cargo.toml
```diff
@@ -116,6 +116,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.cargo_check_external_types]
 allowed_external_types = [
-    "rustls_pki_types",
-    "rustls_pki_types::*",
+  # ---
```
brody4hire, line 119: Needed a comment to keep taplo from compressing these, I can try to come up with a nicer comment here.

cpu (approved):

.github/workflows/build.yml
```diff
@@ -506,6 +506,20 @@ jobs:
...
+      steps:
+        - name: Checkout sources
+          uses: actions/checkout@v4
+          with:
+            persist-credentials: false
+
+        - name: Install rust toolchain
+          uses: dtolnay/rust-toolchain@stable
+        - run: cargo install taplo-cli --locked
+        - run: taplo format --check
```
brody4hire, line 521: aargh extra indenting my bad, PR coming to fix this

PR #2308, opened 2025-01-21
REJECTED: reorganize features in Cargo.toml; use document_features; document `std` feature

brody4hire:
STATUS: Closing as REJECTED; I would like to find a way to improve support from the Rust tooling (someday).

---

- organize features as discussed in https://github.com/rustls/rustls/pull/2306#issuecomment-2605412127
- add missing documentation for `std` feature
- use `document_features` to document features in `rustls/Cargo.toml` & maintain consistency with doc comments

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -14,6 +14,65 @@ autotests = false
 exclude = ["src/testdata", "tests/**"]
 build = "build.rs"
 
+[features]
+default = ["aws_lc_rs", "logging", "std", "tls12"]
+
+## Enable some more advanced connection support APIs.
+std = ["webpki/std", "pki-types/std", "once_cell/std"]
+
+## Makes the rustls crate depend on the [`aws-lc-rs`] crate.
+## Use `rustls::crypto::aws_lc_rs::default_provider().install_default()` to
+## use it as the default `CryptoProvider`, or provide it explicitly
+## when making a `ClientConfig` or `ServerConfig`.
+##
+## Note that aws-lc-rs has additional build-time dependencies like cmake.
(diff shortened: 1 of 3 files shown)


rustls/Cargo.toml
```diff
@@ -14,6 +14,65 @@ autotests = false
 exclude = ["src/testdata", "tests/**"]
 build = "build.rs"
 
+[features]
+default = ["aws_lc_rs", "logging", "std", "tls12"]
+
+## Enable some more advanced connection support APIs.
+std = ["webpki/std", "pki-types/std", "once_cell/std"]
```
brody4hire, line 21: I think this documentation needs to be more specific, would appreciate any pointers to help improve this.

rustls/Cargo.toml
```diff
@@ -31,21 +90,8 @@ webpki = { workspace = true }
...
-custom-provider = []
-fips = ["aws_lc_rs", "aws-lc-rs?/fips"]
-logging = ["log"]
-read_buf = ["rustversion", "std"]
-ring = ["dep:ring", "webpki/ring"]
-std = ["webpki/std", "pki-types/std", "once_cell/std"]
-tls12 = []
-zlib = ["dep:zlib-rs"]
+# doc utility
+document-features = "0.2.10"
```
brody4hire, line 94: Unfortunately it looks like `document-features` needs to be in `[dependencies]` for this to work. I would like to find another solution but don't know how long it would take. I will totally understand if we need to leave this part out.

ctz: I'm afraid I don't think moving text from one file to another is a good justification for new dependencies. In general I would like to avoid proc-macro dependencies in rustls at all costs.

PR #2310, opened 2025-01-22
CI: deny all Clippy warnings including clippy::incompatible_msrv

brody4hire:
It looks to me like there is no need to ignore `clippy::incompatible_msrv` in Clippy CI task to succeed with `--deny warnings`.

Context comes from this comment: https://github.com/rustls/rustls/pull/2285#discussion_r1925385243

For PR #2285 I would like to update CI to allow a single warning in most crates & deny all other warnings. I have already tried this with XXX TODO comments; PR #2285 still has green status with these updates.

Considering that ignoring `clippy::incompatible_msrv` is not what we want for PR #2285, and that CI can succeed without ignoring this warning, I think we should remove this allow option as proposed here. I would also like to keep this update separate from PR #2285.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -463,10 +463,8 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      # - we want to be free of any warnings, so deny them
-      # - disable incompatible_msrv as it does not understand that we apply our
-      #   MSRV to the just the core crate.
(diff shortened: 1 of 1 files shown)


cpu (approved):
If clippy/CI are happy with this change, so am I. Thanks!

djc (approved):

ctz (approved):
Neat, I guess the reason for this has gone away -- though I don't think clippy has addressed the underlying false positive. But we can put this back if it is ever needed.

PR #2311, opened 2025-01-22
CI: deny all Clippy nightly warnings

brody4hire:
Based on other PR that I think should be reviewed & hopefully merged first:

- [ ] #2310

---

My impression is that new Clippy nightly warnings have not been much of an issue over the past year and it has been pretty easy to resolve the few issues I have seen come up.

I think the ideal situation for PR #2285 is to assume that we will only ignore a single warning that it introduces for most crates, and deny all other warnings for Clippy nightly as well.

I do totally think this would require a discussion & decision among the @rustls maintainers for consideration, will totally understand if this proposal need to be rejected & closed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -463,10 +463,8 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      # - we want to be free of any warnings, so deny them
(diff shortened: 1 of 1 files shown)


djc:
I don't think we should do this. clippy on nightly can be a little bit flakey and we only warn on those for a reason. I don't see a good reason to change that now.

PR #2312, opened 2025-01-23
implement quic multipath encryption

divagant-martian:
At https://github.com/n0-computer we are working on implementing [multipath](https://datatracker.ietf.org/doc/draft-ietf-quic-multipath/) in [`quinn`](https://github.com/quinn-rs/quinn). Multipath requires a few changes in the nonce calculation for packet protection (described in section 4.1).

This PR implements those changes to the best of my knowledge (which is fairly limited for this repo). The tests are based on the multipath implementation for [`picoquic`](https://github.com/private-octopus/picoquic/tree/master)

On an implementation note, I opted for separate functions to avoid disrupting implementations that do not support multipath. Let me know if you would rather have them merged

### DIFF
--- a/rustls/src/crypto/cipher.rs
+++ b/rustls/src/crypto/cipher.rs
@@ -214,18 +214,32 @@ impl Nonce {
     /// This is `iv ^ seq` where `seq` is encoded as a 96-bit big-endian integer.
     #[inline]
     pub fn new(iv: &Iv, seq: u64) -> Self {
-        let mut nonce = Self([0u8; NONCE_LEN]);
-        codec::put_u64(seq, &mut nonce.0[4..]);
+        let mut seq_bytes = [0u8; NONCE_LEN];
+        codec::put_u64(seq, &mut seq_bytes[4..]);
+        Self::new_from_seq(iv, seq_bytes)
+    }
 
-        nonce
-            .0
-            .iter_mut()
+    /// Creates a unique nonce based on the multipath `path_id`, the `iv` and packet number `pn`.
+    ///
+    /// The nonce is computed as the XOR between the `iv` and the 96-bit big-endian integer formed
+    /// by concatenating `path_id` and `pn`.
+    pub fn for_path(path_id: u32, iv: &Iv, pn: u64) -> Self {
+        let mut seq_bytes = [0u8; NONCE_LEN];
+        seq_bytes[0..4].copy_from_slice(&path_id.to_be_bytes());
+        codec::put_u64(pn, &mut seq_bytes[4..]);
+        Self::new_from_seq(iv, seq_bytes)
+    }
+
+    /// Creates a unique nonce based on the `iv` and sequence number `seq`.
+    #[inline]
+    fn new_from_seq(iv: &Iv, mut seq: [u8; NONCE_LEN]) -> Self {
+        seq.iter_mut()
             .zip(iv.0.iter())
-            .for_each(|(nonce, iv)| {
-                *nonce ^= *iv;
+            .for_each(|(s, iv)| {
+                *s ^= *iv;
             });
 
-        nonce
+        Self(seq)
     }
 }
 
@@ -346,5 +360,21 @@ impl MessageDecrypter for InvalidMessageDecrypter {
         _seq: u64,
     ) -> Result<InboundPlainMessage<'a>, Error> {
         Err(Error::DecryptError)
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Using test values provided in the spec in
+    /// <https://www.ietf.org/archive/id/draft-ietf-quic-multipath-15.html#section-2.4>
+    #[test]
+    fn multipath_nonce() {
+        const PATH_ID: u32 = 3;
+        const PN: u64 = 54321;
+        const IV: [u8; 16] = 0x6b26114b9cba2b63a9e8dd4fu128.to_be_bytes();
+        const EXPECTED_NONCE: [u8; 16] = 0x6b2611489cba2b63a9e8097eu128.to_be_bytes();
+        let nonce = Nonce::for_path(PATH_ID, &Iv::copy(&IV[4..]), PN).0;
+        assert_eq!(EXPECTED_NONCE[4..], nonce);
     }
 }
--- a/rustls/src/crypto/ring/quic.rs
+++ b/rustls/src/crypto/ring/quic.rs
@@ -130,9 +130,14 @@ impl quic::PacketKey for PacketKey {
         packet_number: u64,
         header: &[u8],
         payload: &mut [u8],
+        path_id: Option<u32>,
     ) -> Result<quic::Tag, Error> {
         let aad = aead::Aad::from(header);
-        let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, packet_number).0);
+        let nonce_value = match path_id {
+            Some(path_id) => Nonce::for_path(path_id, &self.iv, packet_number),
+            None => Nonce::new(&self.iv, packet_number),
+        };
+        let nonce = aead::Nonce::assume_unique_for_key(nonce_value.0);
         let tag = self
             .key
             .seal_in_place_separate_tag(nonce, aad, payload)
@@ -142,20 +147,28 @@ impl quic::PacketKey for PacketKey {
 
     /// Decrypt a QUIC packet
     ///
-    /// Takes the packet `header`, which is used as the additional authenticated data, and the
-    /// `payload`, which includes the authentication tag.
+    /// Takes a `packet_number` and optional `path_id`, used to derive the nonce; the packet
+    /// `header`, which is used as the additional authenticated data, and the `payload`, which
+    /// includes the authentication tag.
     ///
-    /// If the return value is `Ok`, the decrypted payload can be found in `payload`, up to the
-    /// length found in the return value.
+    /// On success, returns the slice of `payload` containing the decrypted data.
+    ///
+    /// When provided, the `path_id` is used for multipath ecryption as described in
+    /// <https://www.ietf.org/archive/id/draft-ietf-quic-multipath-15.html#section-2.4>.
     fn decrypt_in_place<'a>(
         &self,
         packet_number: u64,
         header: &[u8],
         payload: &'a mut [u8],
+        path_id: Option<u32>,
     ) -> Result<&'a [u8], Error> {
         let payload_len = payload.len();
         let aad = aead::Aad::from(header);
-        let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, packet_number).0);
+        let nonce_value = match path_id {
+            Some(path_id) => Nonce::for_path(path_id, &self.iv, packet_number),
+            None => Nonce::new(&self.iv, packet_number),
+        };
+        let nonce = aead::Nonce::assume_unique_for_key(nonce_value.0);
         self.key
             .open_in_place(nonce, aad, payload)
             .map_err(|_| Error::DecryptError)?;
@@ -249,7 +262,7 @@ mod tests {
         let mut buf = PLAIN.to_vec();
         let (header, payload) = buf.split_at_mut(4);
         let tag = packet
-            .encrypt_in_place(PN, header, payload)
+            .encrypt_in_place(PN, header, payload, None)
             .unwrap();
         buf.extend(tag.as_ref());
 
@@ -270,7 +283,7 @@ mod tests {
 
         let (header, payload_tag) = buf.split_at_mut(4);
         let plain = packet
-            .decrypt_in_place(PN, header, payload_tag)
+            .decrypt_in_place(PN, header, payload_tag, None)
             .unwrap();
 
         assert_eq!(plain, &PLAIN[4..]);
@@ -383,7 +396,7 @@ mod tests {
         let tag = server
             .local
             .packet
-            .encrypt_in_place(1, &server_header, &mut server_payload)
+            .encrypt_in_place(1, &server_header, &mut server_payload, None)
             .unwrap();
         let (first, rest) = server_header.split_at_mut(1);
         let rest_len = rest.len();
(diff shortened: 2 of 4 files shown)


rustls/src/quic.rs
```diff
@@ -701,6 +701,23 @@ pub trait PacketKey: Send + Sync {
...
+    /// Encrypts a multipath QUIC packet
+    ///
+    /// Takes a `path_id` and `packet_number`, used to derive the nonce; the packet `header`, which is used as
+    /// the additional authenticated data; and the `payload`. The authentication tag is returned if
+    /// encryption succeeds.
+    ///
+    /// Fails if and only if the payload is longer than allowed by the cipher suite's AEAD algorithm.
+    ///
+    /// See <https://www.ietf.org/archive/id/draft-ietf-quic-multipath-11.html#name-nonce-calculation>.
+    fn encrypt_in_place_multipath(
```
djc, line 713: I don't think we want to make a semver-incompatible release for this, so if this trait needs new methods those need to have a default implementation.

divagant-martian: I'm not sure what would be a good default implementation. Silently ignoring the path_id and using the non-multipath method will not produce the right data. Is there a better option to returning an error? The appropriate ones would then be `Error::EncryptError` and `Error::DecryptError`, is that correct?

divagant-martian: ok this is what I did, let me know if there's a better option

djc: I think that's probably a reasonable solution in this case.

rustls/src/crypto/cipher.rs
```diff
@@ -213,18 +213,32 @@ impl Nonce {
...
+    pub fn new_multipath(iv: &Iv, pn: u64, path_id: u32) -> Self {
+        let mut seq_bytes = [0u8; NONCE_LEN];
+        seq_bytes[0..4].copy_from_slice(&path_id.to_be_bytes());
+        codec::put_u64(pn, &mut seq_bytes[4..]);
+        Self::new_from_seq(iv, seq_bytes)
+    }
+
+    /// Creates a unique nonce based on the `iv` and sequence number `seq`.
+    #[inline]
+    pub fn new_from_seq(iv: &Iv, mut seq: [u8; NONCE_LEN]) -> Self {
```
djc, line 234: Should this be private?

rustls/src/crypto/cipher.rs
```diff
@@ -213,18 +213,32 @@ impl Nonce {
...
+    }
 
-        nonce
-            .0
-            .iter_mut()
+    /// Creates a unique nonce based on the `iv`, the packet number `pn` and multipath `path_id`.
+    ///
+    /// The nonce is computed as the XOR between the `iv` and the 96-bit big-ending integer formed
+    /// by concatenating `path_id` and `pn`.
+    pub fn new_multipath(iv: &Iv, pn: u64, path_id: u32) -> Self {
```
djc, line 225: Maybe call this `for_path`? I'd also keep `path_id` as the first argument.

rustls/src/crypto/ring/quic.rs
```diff
@@ -140,6 +140,24 @@ impl quic::PacketKey for PacketKey {
         Ok(quic::Tag::from(tag.as_ref()))
     }
 
+    fn encrypt_in_place_multipath(
```
djc, line 143: Maybe replace `_multipath` with `_for_path`?

rustls/src/crypto/ring/quic.rs
```diff
@@ -164,6 +182,26 @@ impl quic::PacketKey for PacketKey {
         Ok(&payload[..plain_len])
     }
 
+    fn decrypt_in_place_multipath<'a>(
```
djc, line 185: Replace `_multipath` with `_for_path()`?

rustls/src/crypto/ring/quic.rs
```diff
@@ -413,4 +451,81 @@ mod tests {
...
+        let mut buf = PAYLOAD.to_vec();
+        let tag = packet
+            .encrypt_in_place_multipath(PATH_ID, PN, HEADER, &mut buf)
+            .unwrap();
+        buf.extend_from_slice(tag.as_ref());
+
+        assert_eq!(buf.as_slice(), EXPECTED);
+    }
+
+    // This test is based on `multipath_aead_test` in `picoquictest/multipath_test.c`
```
djc, line 495: Maybe turn this into a GitHub link (to a specific revision)? (For both tests.)

djc (commented):
I think this is looking pretty good modulo nits. Suggest you keep this as draft until you have something working end-to-end/need to depend on it in a published crate.

divagant-martian:
@djc thanks for the review, just updated based on suggestions. 

I can make this a draft but we I think it would be benefit for everyone if this is included in the next release. I expect the API and implementation to not change since the spec changes are fairly simple unless there's something we are missing? I hope not but there's clearly that risk. That being said, keeping it open requires us to depend on a fork for development, find a timely release for `rustls` for when our fork of `quinn`, and eventually og `quinn` itself, are ready in terms of multipath. I would rather not try to rush the `rustls` team pestering you all with a release when we needed it but instead allow you to include it in the next one whenever that happens, since that will most likely happen before us having a full multipath implementation we are happy with.

Let me know what you think!

djc:
> I can make this a draft but we I think it would be benefit for everyone if this is included in the next release. I expect the API and implementation to not change since the spec changes are fairly simple unless there's something we are missing? I hope not but there's clearly that risk. That being said, keeping it open requires us to depend on a fork for development, find a timely release for `rustls` for when our fork of `quinn`, and eventually og `quinn` itself, are ready in terms of multipath. I would rather not try to rush the `rustls` team pestering you all with a release when we needed it but instead allow you to include it in the next one whenever that happens, since that will most likely happen before us having a full multipath implementation we are happy with.

We make semver-compatible releases quite often and are pretty fast at doing it. I can virtually guarantee that we can get it done within like one workday of whenever your downstream work is ready, but I would really prefer to hold off on merging this until that is done.

> ETA: codecov is unhappy, but it wants me to test the default impl of the multipath methods. Should I add these tests?

Seems fine to me, but maybe @ctz has a different opinion?

Also, can you file a followup issue with the `next-major-release` label to drop these default impls?

divagant-martian:
> But I would really prefer to hold off on merging this until that is done

Understood, we'll be knocking in your door hopefully not too far in the future

> can you file a followup issue with the next-major-release label to drop these default impls?

Issue created https://github.com/rustls/rustls/issues/2315 but it's pending labels (I can't add them myself)

djc:
> > can you file a followup issue with the next-major-release label to drop these default impls?
> 
> Issue created #2315 but it's pending labels (I can't add them myself)

Thanks, added the label.

ctz:
> Seems fine to me, but maybe @ctz has a different opinion?

I generally do not write tests for trivial functions just for the coverage. This PR currently looks great to me coverage-wise.

cpu:
the diff looks good to me too, thanks! If you had time I feel like this could be squashed into a singular commit instead of three.

cpu:
Is this ready to be flipped out of draft status w/ a rebase & squash or would folks still prefer to see this held until the end-to-end feature works with Quinn? 

If the latter, could someone link up the Quinn PR it's blocked on? I took a quick scan and didn't spot an obvious candidate.

djc:
FYI, main is open for breaking changes for 0.24 now.

divagant-martian:
@djc would you mind approving the workflows to know what's the state of tests at this point? 🙏

cpu:
> would you mind approving the workflows to know what's the state of tests at this point?

Apologies if this is already something you're aware of, but you can also fork our repo, enable GitHub actions (it's a button in the actions tab after you fork), and then push your branch to your fork named something like `ci/my-change`. The `ci/` prefix will make the actions run. Doing a process like that will let you check the state of CI at any point without needing to wait for approvals or to push updates to the PR before they're ready.

(_We should probably add that to our contributing docs somewhere_)

divagant-martian:
Now that breaking changes are allowed I removed the default implementations (would close #2315, makes codecov happy)

The spec now also gives a small test case that I included, so this is ready from my perspective

Also note the broken link failure is in `tlsvulns.rs`, so unrelated

rustls/src/quic.rs
```diff
@@ -731,6 +731,23 @@ pub trait PacketKey: Send + Sync {
...
+    /// Encrypts a multipath QUIC packet
+    ///
+    /// Takes a `path_id` and `packet_number`, used to derive the nonce; the packet `header`, which is used as
+    /// the additional authenticated data; and the `payload`. The authentication tag is returned if
+    /// encryption succeeds.
+    ///
+    /// Fails if and only if the payload is longer than allowed by the cipher suite's AEAD algorithm.
+    ///
+    /// See <https://www.ietf.org/archive/id/draft-ietf-quic-multipath-11.html#name-nonce-calculation>.
+    fn encrypt_in_place_for_path(
```
djc: Instead of adding these methods, how about we make `path_id` an `Option` for `encrypt_in_place()`/`decrypt_in_place()`? (It should become last in the argument list.)

divagant-martian: since multipath encryption is compatible with non multipath encryption, using None and 0 are equivalent, there's that option as well and in general I think the suggestion is a reasonable change. 

A drawback is that from an API perspective it makes mandatory a parameter very few will use - at least for now -

There's also the question of how deep to make the merge, since ultimately `Nonce::for_path(0,..)` and `Nonce::new(..)` are equivalent

divagant-martian: the change is in [43312ae](https://github.com/rustls/rustls/pull/2312/commits/43312aee6f4f8d62270356d9279e1f6c6ea117c9) not yet squashed so that it can be reviewed, let me know what you think.

I left the option for clarity, and did not merge the nonce calculation to avoid potentially wasting resources copying zeros around

djc:
@ctz / @cpu opinions on single method vs different methods?

rustls/src/crypto/cipher.rs
```diff
@@ -214,18 +214,32 @@ impl Nonce {
...
     #[inline]
     pub fn new(iv: &Iv, seq: u64) -> Self {
-        let mut nonce = Self([0u8; NONCE_LEN]);
-        codec::put_u64(seq, &mut nonce.0[4..]);
+        let mut seq_bytes = [0u8; NONCE_LEN];
+        codec::put_u64(seq, &mut seq_bytes[4..]);
+        Self::new_from_seq(iv, seq_bytes)
+    }
+
+    /// Creates a unique nonce based on the `iv`, the packet number `pn` and multipath `path_id`.
```
cpu: This comment describes the arguments in the order `iv`, `pn` and then `path_id`, but the function itself takes `path_id` as the first argument. Should one of them change to be in-sync?

rustls/src/crypto/cipher.rs
```diff
@@ -214,18 +214,32 @@ impl Nonce {
...
+    ///
+    /// The nonce is computed as the XOR between the `iv` and the 96-bit big-ending integer formed
```
cpu: small typo here I think. "big-ending" -> "big-endian"?

rustls/src/crypto/cipher.rs
```diff
@@ -214,18 +214,32 @@ impl Nonce {
     /// This is `iv ^ seq` where `seq` is encoded as a 96-bit big-endian integer.
```
cpu, line 214: Anchoring a comment here arbitrarily: You've got a small typo in the commit message header "implment" -> "implement"

rustls/src/crypto/cipher.rs
```diff
@@ -348,3 +362,18 @@ impl MessageDecrypter for InvalidMessageDecrypter {
         Err(Error::DecryptError)
     }
 }
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Using test values provided in the spec
```
cpu: nit: can you link the specific spec + section here?

rustls/src/quic.rs
```diff
@@ -745,6 +762,24 @@ pub trait PacketKey: Send + Sync {
         payload: &'a mut [u8],
     ) -> Result<&'a [u8], Error>;
 
+    /// Decrypt a multipath QUIC packet
+    ///
+    /// Takes a `path_id` and `packet_number`, used to derive the nonce; the packet `header`, which is used as
+    /// the additional authenticated data; and the `payload`. The authentication tag is returned if
+    /// encryption succeeds.
```
cpu: I think this last bit should say "if decryption succeeds"

rustls/src/quic.rs
```diff
@@ -745,6 +762,24 @@ pub trait PacketKey: Send + Sync {
...
+    ///
+    /// If the return value is `Ok`, the decrypted payload can be found in `payload`, up to the
+    /// length found in the return value.
```
cpu: Does this need to be reworded? I'm getting thrown by the "the length found in the return value part" since it looks like the signature returns `&[u8]` in the success case and the impl is giving `Ok(&payload[..plain_len])` where from the docs alone I'd assume it'd return `plain_len` and the caller has to dig it out of `payload`.

cpu (approved):
Thanks! Here's a first pass of feedback.

> @ctz / @cpu opinions on single method vs different methods?

I feel like I don't have a great intuition on the downstream consumers of this part of rustls + how prevalent multi-path quic is to have a strong pref. If this is fairly niche w.r.t other QUIC stacks maybe a separate fn to keep it off the happy-path for everyone else would be better.

cpu:
> cpu approved these changes now
> cpu self-requested a review now

Sorry, hit the review submit button with approved selected when I wanted to just comment.

djc:
> @ctz / @cpu opinions on single method vs different methods?

Also, @Ralith who might have stronger opinions in this case?

rustls/src/crypto/cipher.rs
djc: I think `path_id` should go last in both.

rustls/src/quic.rs
divagant-martian: this comes from the existing method https://github.com/rustls/rustls/blob/e2f6115548b54e15ec52517208e4b4c6f27aa7db/rustls/src/crypto/ring/quic.rs#L143-L150

rustls/src/crypto/cipher.rs
divagant-martian: is there consensus about this? this is the first argument because it was explicitly requested as such https://github.com/rustls/rustls/pull/2312#discussion_r1928625093

ctz:
> @ctz / @cpu opinions on single method vs different methods?

Not sure, but I can explain my preference:

- if both are always going to be required, then one method
- if there is the prospect of one or the other being deprecated, or failing to gain traction, and eventually become unneeded, then two methods

rustls/src/crypto/cipher.rs
divagant-martian: :+1: I'll keep it in mind for the final squash. Leaving it untouched while the pr settles - to facilitate reviewing

rustls/src/quic.rs
cpu: If you agree it's not accurate add a commit up-front fixing them both?

rustls/src/crypto/cipher.rs
cpu: I don't feel strongly except that the docs + args be consistent. We can go with whatever past Djc or present Djc prefers :laughing:

rustls/src/quic.rs
divagant-martian: ok after checking it I agree it's misleading. What about
```rust
/// If the return value is `Ok`, it references the slice of `payload` containing the decrypted payload
```

suggestions welcome

cpu: How about something like:

> /// On success, returns the slice of `payload` containing the decrypted data

divagant-martian:
Thanks for the feedback. Based on your comments, the methods are separate, and not required. This means restoring the default implementation (not covered by tests). This reduces the burden on users who have no interest in multipath and makes the set of users affected by a potential deprecation much smaller.

The multipath methods and their docs have all path as starting argument, making them consistent among them, and their docs describe them in the listed order.

Docs about return value meaning for decryption methods should now be clear as well

PTAL

Ralith:
> Also, @Ralith who might have stronger opinions in this case?

I don't love that this makes the use of an implementation which hasn't been updated a runtime error, but it's not exactly an easy one to miss, and I'm also sympathetic to concerns about it being unsupported. Long-term maybe we can deprecate the old version?

I don't feel very strongly, since in practice there'll only be a few implementers and most users won't be customizing this.

djc:
> I don't love that this makes the use of an implementation which hasn't been updated a runtime error, but it's not exactly an easy one to miss, and I'm also sympathetic to concerns about it being unsupported. Long-term maybe we can deprecate the old version?

Not entirely sure what you mean by "this" and "it's" -- sorry... Can you clear this up? Note that this PR targets `main`, which has already bumped to 0.24.0-dev.0 and thus has no backwards compatibility constraints.

divagant-martian:
I suppose it's about the default implementation using an error. Would it be an improvement to add to the docs that by default the function errors? Alternatively, and harder to miss is to make the default implementation be `unimplemented!()`

Ralith:
> Can you clear this up?

I mean the choice to introduce new methods with a provided implementation rather than break the existing trait.

A third option might be to introduce a new trait, but that pushes churn downstream, which isn't great either.

> Would it be an improvement to add to the docs that by default the function errors?

No, I think the main risk is users inadvertently supplying implementations which haven't been updated, and those users won't notice the doc change. I think if we want to avoid churn then the current draft is the right way to go.

divagant-martian:
Is there anything missing to move this pr forward?

djc:
No, please squash your changes so we stick with a single method with `path_id: Option<_>` argument.

divagant-martian:
reverted to single method and squashed, please take another look

djc (approved):

PR #2313, opened 2025-01-23
add missing doc for default `std` feature

brody4hire:


rustls/src/lib.rs
```diff
@@ -273,6 +273,8 @@
 //! Here's a list of what features are exposed by the rustls crate and what
 //! they mean.
 //!
+//! - `std` (enabled by default): enable some more advanced connection support APIs that require Rust `std` library.
```
brody4hire, line 276: Any suggestions on how to make this statement more specific?

@rustls maintains please DO feel free to update this as may be fit & take co-authored-by credit.

I suspect the last few words "require Rust `std` library" could use some better wording, @rustls maintainers please feel free to reword (and take co-authored-by credit).

djc: "Enable the high-level (buffered) `Connection` API, which relies on std API."?

ctz: "Enable the high-level (buffered) Connection API and other functionality which relies on `std`"?

(there are a lot of small details that depend on `std`, like error types, caches, even the basic constructors of `ServerConfig`, and I don't think this is a good place to document all that. Instead I think rustdoc already does a good-enough job of that.)

rustls/src/lib.rs
```diff
@@ -272,6 +272,10 @@
 //! # Crate features
 //! Here's a list of what features are exposed by the rustls crate and what
 //! they mean.
+//!
+//! - `std` (enabled by default): enable the high-level (buffered) Connection API and other functionality
+//!   which relies on the `std` library.
+
```
brody4hire, line 278: oops I will remove this

```suggestion
```

ctz (approved):

cpu (approved):
Thanks

Issue #2316, opened 2025-01-28
Panic in `write_tls`


ConradIrwin:
**Checklist**
* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**

At [Zed](https://zed.dev) we run tokio_rustls in production with a few tens of thousands of active users. About once a day we see a panic with the following trace:

Notably the "slice of length 24" seems to always be that size; but the "range start index" varies considerably (from 40 - 216 in the last week).

I tried following the code, but I think there's some inlining going on that makes it tricky to decipher. I also tried to grep for magic number "24", but there was no obvious use that made sense. 

```
Panic: range start index 40 out of range for slice of length 24
core::slice::index::slice_start_index_len_fail::heeb7c73b0ee7f95b
rustls::vecbuf::ChunkVecBuffer::write_to::he63d8bd98360b066
_$LT$tokio_rustls..client..TlsStream$LT$IO$GT$$u20$as$u20$tokio..io..async_write..AsyncWrite$GT$::poll_shutdown::he45ac55ffd33d1df
h2::proto::connection::Connection$LT$T$C$P$C$B$GT$::poll::h8d7d535d8032d8e7
_$LT$hyper..proto..h2..client..H2ClientFuture$LT$B$C$T$GT$$u20$as$u20$core..future..future..Future$GT$::poll::h945af6542e475182
tokio::runtime::task::raw::poll::h603eb9699b70c2c3
tokio::runtime::scheduler::multi_thread::worker::Context::run_task::h9f5a9bf687bfba0a
tokio::runtime::scheduler::multi_thread::worker::run::he5a038213beb3f9a
tokio::runtime::task::raw::poll::h96fd9b5e4d5c3cd2
core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h4c6d91adc37e555d
std::sys::pal::unix::thread::Thread::new::thread_start::h1bd1b9c95010bf71
_pthread_start
thread_start
```

Running through [symbolicate](https://github.com/ConradIrwin/symbolicate) gives me:

```
/rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/slice/index.rs:57:	core::slice::index::slice_end_index_len_fail::h19b63a1d9e216190
[...]

PR #2319, opened 2025-01-29
Defensiveness improvements to `ChunkVecBuffer`

ctz:
In pursuit of #2316 I took #2318, added some extra checks (which are release-mode asserts where something external has gone wrong).

Note that on main right now, `ChunkVecBuffer::consume()` is externally callable with any value (via `BufRead::consume`) but this is not the case in any released version and therefore not under suspicion for #2316. However, this PR adds a release-mode assert for misuse of that.

During investigating this bug, I did this to model-check `ChunkVecBuffer`:

1. Wrote a very simplified model of `ChunkVecBuffer`, which is slow.

2. Had a fuzz target that takes as input a sequence of

```rust
#[derive(Arbitrary, Debug)]
enum Operation {
    Append(Data),
    Pop,
    ConsumeFirstChunk { max: usize },
    WriteTo { max: usize },
}
```

3. Performs those actions against a `ChunkVecBuffer` and a model, in lockstep. 
5. After each operation, ensures the real and model versions agree what the contents of the buffer is.

That fuzzer is not included in this PR (since it cloned the whole of `ChuckVecBuffer`, and therefore is not useful.)

In any case, I ran it for many millions of iterations with no panics, assertions or divergence :(

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -290,7 +290,8 @@ mod connection {
         }
 
         fn consume(&mut self, amt: usize) {
-            self.received_plaintext.consume(amt)
+            self.received_plaintext
+                .consume_first_chunk(amt)
         }
     }
 
--- a/rustls/src/vecbuf.rs
+++ b/rustls/src/vecbuf.rs
@@ -142,7 +142,20 @@ impl ChunkVecBuffer {
         Ok(offs)
     }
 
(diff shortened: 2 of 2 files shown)


ctz:
(first commit will fall away when #2318 is merged)

cpu (approved):

ctz:
Going to merge this with one review, as it's pretty low risk.

PR #2332, opened 2025-02-06
unbuffered: introduce `PeerClosed` state

### LINKED ISSUE #1895: unbuffered: After `Closed` no `WriteTraffic` state arrives
**Checklist**

**Describe the bug**
I've updated my `pipebuf` crate Rustls wrapper in crate `pipebuf_rustls` to add unbuffered Rustls support. However my tests which succeed in the buffered mode fail for the unbuffered mode. The failing test does the following:

- Client sends one byte
- Client closes connection
- Server sends one byte
- Server closes connection

The byte sent by the server cannot be processed through the unbuffered interface because the `process_tls_records` call does not give a `WriteTraffic` state again after `Closed`. Without access to `WriteTraffic`, I can't send the data. The buffered interface does not have this limitation.

**To Reproduce**
Check out `pipebuf_rustls` from github (https://github.com/uazu/pipebuf_rustls).  Run a single test using `cargo test --no-default-features --features unbuffered byte_each_way`. There is an assertion failure due to 1 byte sent but 0 bytes received.

**Applicable Version(s)**
0.23.4 on Linux

**Expected behavior**
I'd expect this case to be supported. I don't know how you'd want to support it with your unbuffered API though.

**Additional context**
The reason I added unbuffered support to `pipebuf_rustls` was that this is a much better fit to the `pipebuf` model. I added `PBufRd::data_mut` to `pipebuf` 0.3.1 to support Rustls unbuffered mode (which requires `&mut [u8]` for incoming data). So I support the unbuffered effort. Just it doesn't cover all necessary cases yet.

ctz:
`Closed` was, and remains, the bi-directional, terminal closure state.

Because `rustls::unbuffered::ConnectionState` is a `non_exhaustive` enum, this is not a breaking API change. However, it demands a clear release note because downstream users will now encounter the new state.

fixes #1895

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -811,6 +811,7 @@ impl<Data> From<ConnectionCore<Data>> for ConnectionCommon<Data> {
 pub struct UnbufferedConnectionCommon<Data> {
     pub(crate) core: ConnectionCore<Data>,
     wants_write: bool,
+    emitted_peer_closed: bool,
 }
 
 impl<Data> From<ConnectionCore<Data>> for UnbufferedConnectionCommon<Data> {
@@ -818,6 +819,7 @@ impl<Data> From<ConnectionCore<Data>> for UnbufferedConnectionCommon<Data> {
         Self {
             core,
             wants_write: false,
+            emitted_peer_closed: false,
         }
     }
 }
--- a/rustls/src/conn/unbuffered.rs
+++ b/rustls/src/conn/unbuffered.rs
@@ -131,6 +131,18 @@ impl<Data> UnbufferedConnectionCommon<Data> {
                 .core
                 .common_state
                 .has_received_close_notify
+                && !self.emitted_peer_closed
+            {
+                self.emitted_peer_closed = true;
+                break (buffer.pending_discard(), ConnectionState::PeerClosed);
+            } else if self
+                .core
+                .common_state
+                .has_received_close_notify
+                && self
+                    .core
+                    .common_state
+                    .has_sent_close_notify
             {
                 break (buffer.pending_discard(), ConnectionState::Closed);
             } else if self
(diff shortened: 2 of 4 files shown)


rustls/tests/unbuffered.rs
```diff
@@ -870,6 +922,39 @@ fn write_traffic<T: SideData, R, F: FnMut(WriteTraffic<T>) -> R>(
...
+    status: UnbufferedStatus<'_, '_, T>,
+    mut f: F,
+) -> (R, usize) {
+    let UnbufferedStatus { discard, state } = status;
+    let state = state.unwrap();
+    if let ConnectionState::ReadTraffic(state) = state {
+        (f(state), discard)
+    } else {
+        panic!("unexpected client state {state:?} (wanted ReadTraffic)");
+    }
```
djc, line 935: I would find these clearer when written with a `match` expression.

djc (approved):
Nice!

rustls/src/conn.rs
```diff
@@ -811,13 +811,15 @@ impl<Data> From<ConnectionCore<Data>> for ConnectionCommon<Data> {
 pub struct UnbufferedConnectionCommon<Data> {
     pub(crate) core: ConnectionCore<Data>,
     wants_write: bool,
+    emitted_peer_closed: bool,
```
cpu, line 814: Would this be a touch clearer named `emitted_peer_closed_state` to disambiguate with the actual alert on the wire or am I over-thinking?

rustls/src/conn/unbuffered.rs
```diff
@@ -131,6 +131,18 @@ impl<Data> UnbufferedConnectionCommon<Data> {
                 .core
                 .common_state
                 .has_received_close_notify
+                && !self.emitted_peer_closed
```
cpu, line 134: I'm starting to find the big chain of `if/else if`'s and their relatively complicated boolean expressions tricky to follow. It's quite dense.

I don't think it's necessary to clean up in this branch but I wonder if we could lift out some intermediate vars at the minimum so that the density is reduced a little bit.

rustls/tests/unbuffered.rs
```diff
@@ -870,6 +922,39 @@ fn write_traffic<T: SideData, R, F: FnMut(WriteTraffic<T>) -> R>(
...
+    let UnbufferedStatus { discard, state } = status;
+    let state = state.unwrap();
+    if let ConnectionState::PeerClosed = state {
+        discard
+    } else {
+        panic!("unexpected client state {state:?} (wanted PeerClosed)");
+    }
+}
+
+fn closed<T: SideData>(status: UnbufferedStatus<'_, '_, T>) -> usize {
```
cpu, line 948: it seemed like `closed`, `peer_closed` and `read_traffic` could be made into one helper, but in practice because `ConnectionState` doesn't impl `PartialEq` the contortions I had to make to get it working didn't seem like a net improvement.

cpu (approved):
Nice, :+1:

rustls/src/conn/unbuffered.rs
ctz: Yeah, agree. Even though I find the way the early data state is injected into `process_tls_records_common` a bit distasteful as a special case, I think actually it's a good direction to split the logic up a bit.

djc: Would suggest taking that out of this PR, though.

PR #2334, opened 2025-02-10
cleanup: use internal alias for std::error::Error (etc.) - REJECTED

brody4hire:
I think this would be a little cleaner and would make it a little easier to switch to using `core::error::Error` once this is possible.

---

__TODO__

- [ ] cleanup remaining uses of `std::error::Error`


djc:
This increases the level of indirection, and I don't think it provides enough value to compensate for this. I'm generally not a fan of type aliases because they obfuscate the underlying meaning without adding meaningful abstraction.

PR #2337, opened 2025-02-11
Expose SingleCertAndKey server cert resolver

djc:
Motivated by:

- https://github.com/hickory-dns/hickory-dns/pull/2764

(Should we start deprecating OCSP API?)

## Proposed release notes

- Export `SingleCertAndKey` implementation of `ResolvesServerCert` (was already used internally).
- Expose `CertifiedKey::from_der()` to help create `CertifiedKey`s with necessary checks.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -8,7 +8,7 @@ use crate::builder::{ConfigBuilder, WantsVerifier};
 use crate::client::{handy, ClientConfig, EchMode, ResolvesClientCert};
 use crate::error::Error;
 use crate::key_log::NoKeyLog;
-use crate::msgs::handshake::CertificateChain;
+use crate::sign::{CertifiedKey, SingleCertAndKey};
 use crate::sync::Arc;
 use crate::versions::TLS13;
 use crate::webpki::{self, WebPkiServerVerifier};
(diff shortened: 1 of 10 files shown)


rustls/src/server/handy.rs
```diff
@@ -168,9 +168,9 @@ impl server::ProducesTickets for NeverProducesTickets {
 
 /// Something which always resolves to the same cert chain.
 #[derive(Debug)]
-pub(super) struct AlwaysResolvesChain(Arc<sign::CertifiedKey>);
+pub(super) struct SingleCertAndKey(Arc<sign::CertifiedKey>);
```
djc: Not married to this name in particular, but `AlwaysResolvesChain` seems pretty idiosyncratic to me.

ctz (approved):
Thoughts about doing the matching change for `client::handy::AlwaysResolvesClientCert` for symmetry?

(there's a copy of that in `rustls/tests/api.rs` which could be eliminated in case we do)

rustls/src/crypto/signer.rs
```diff
@@ -123,6 +125,29 @@ pub struct CertifiedKey {
 }
 
 impl CertifiedKey {
+    /// Create a new `CertifiedKey` from a certificate chain and DER-encoded private key.
+    ///
+    /// Attempt to parse the private key with the given [`CryptoProvider`]'s [`KeyProvider`] and
+    /// verify that it matches the public key in the first certificate of the `cert_chain`.
```
cpu, line 150: nit: maybe add "if possible" here so that the way this code eats `Error::InconsistentKeys(InconsistentKeys::Unknown)` when `keys_match` isn't wired up doesn't come as a surprise?

rustls/src/client/builder.rs
```diff
@@ -8,7 +8,7 @@ use crate::builder::{ConfigBuilder, WantsVerifier};
 use crate::client::{handy, ClientConfig, EchMode, ResolvesClientCert};
```
cpu, line 8: I think this commit title has a typo: "SingleCertAndReady" -> "SingleCertAndKey"

cpu (approved):
Nice improvement :clap:

> (Should we start deprecating OCSP API?)

It feels a little bit premature to me :man_shrugging:

rustls/src/crypto/signer.rs
djc: Makes sense, done.

rustls/src/client/builder.rs
djc: Nice catch! Brains...

djc:
* Published rustls v0.23.23 at registry `crates-io`
* [new tag]           v/0.23.23 -> v/0.23.23
* [Release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.23.23)

PR #2338, opened 2025-02-11
unbuffered: do not prematurely consume data

### LINKED ISSUE #2031: unbuffered: No way to refuse a ReadTraffic if there is no space
**Checklist**

**Describe the bug**

The scenario is that Rustls unbuffered is being used between four fixed-sized buffers (in/out on each side), each of which might have a varying level of fill depending on downstream/upstream processing. So depending on the size of the record being decoded, there might not be space to output the results. In simple terms both data to decode and space to put the result must be present before a Rustls operation can occur successfully. 

ReadTraffic appears to support this in terms of having a `peek_len` method to get the length that is required. If there is not space, presumably the intention is that I should not call `next_record` and I shouldn't apply the `discard`.  However even if `next_record` is not called, the operation already seems to have been committed to, and the data is lost.

**To Reproduce**

The code I have to reproduce this is in a couple of crate versions which aren't published. I'd have to give you a TGZ.

**Applicable Version(s)**

0.23.10 on Linux

**Expected behavior**

If the intention of unbuffered is to allow working with fixed-size buffers, then there always needs to be a way to refuse an operation if the caller doesn't have enough space to store the result. So I'd expect that if I didn't commit (via `next_record` or whatever) then the operation wouldn't be commited internally in Rustls either. But it's not the details of the API that are important, but just having the means to handle the situation. If there was some other way to check how much space is required at some other point, e.g. before calling `process_tls_records`, that would also be okay. Even something like "don't call it unless you have 32KB free in both outgoing buffers". That's more wasteful but could also work. 

**Additional context**

I think it may be worth putting unbuffered under a cargo feature (e.g. "beta") and specifying that it isn't under semver, to give you some flexibility to iterate until things are more solid.

I also have some difficulties with WriteTraffic, but more solvable. Since in this scenario my output buffers may be fixed-size but not fully allocated, I don't want to reallocate the buffers to the maximum unless really necessary. So I need to estimate how much space you require for the `WriteTraffic::encrypt` call. So I'm estimating 100 bytes more than the unencrypted data length. This mostly works, but sometimes I was seeing much larger requirements in terms of InsufficientSize being returned. It seems that other stuff may get included in the write unrelated to the data I'm sending, which I can't account for in my estimate. 

Also, when my output space is limited, I was attempting to send data 100 bytes shorter than the space I have free. But this may be complicated by the other stuff that Rustls wants to send. I tried to back off if I get a failure, and send zero bytes to a zero-sized buffer get a measurement of the hidden data that needs sending. However I have not yet fully debugged this code because of the issues on the ReadTraffic side.

So it may be helpful to have better documentation on buffer sizing requirements and the expectations that Rustls has of the caller.

Again, as with the other issue (#1895) there is no rush. I'm trying to implement this to test the feasibility of running with fixed-sized buffers. But I don't have a deadline.

ctz:
This alters `ReadTraffic` and `ReadEarlyData` so that their `next_record()` functions are where the data is consumed, allowing `peek_len()` functions work as intended.

fixes #2031

### DIFF
--- a/rustls/src/conn/unbuffered.rs
+++ b/rustls/src/conn/unbuffered.rs
@@ -53,15 +53,15 @@ impl<Data> UnbufferedConnectionCommon<Data> {
                 break (buffer.pending_discard(), execute(self, incoming_tls, value));
             }
 
-            if let Some(chunk) = self
+            if !self
                 .core
                 .common_state
                 .received_plaintext
-                .pop()
+                .is_empty()
             {
                 break (
                     buffer.pending_discard(),
-                    ReadTraffic::new(self, incoming_tls, chunk).into(),
+                    ReadTraffic::new(self, incoming_tls).into(),
                 );
             }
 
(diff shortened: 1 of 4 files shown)


rustls/src/conn/unbuffered.rs
```diff
@@ -316,50 +316,51 @@ impl<Data> fmt::Debug for ConnectionState<'_, '_, Data> {
...
-        } else {
-            self.taken = true;
-            Some(Ok(AppDataRecord {
+        self.chunk = self
+            .conn
+            .core
+            .common_state
+            .received_plaintext
+            .pop();
+        self.chunk.as_ref().map(|ch| {
```
djc, line 349: Nit: rename `ch` to `payload` (or maybe `chunk`)?

rustls/src/vecbuf.rs
```diff
@@ -86,6 +86,13 @@ impl ChunkVecBuffer {
         len
     }
 
+    /// Inspect the first chunk from this object.
```
djc, line 89: Nit: I'd sort this below the `&mut` `pop()` and `read_buf()`, probably?

rustls/src/conn/unbuffered.rs
```diff
@@ -316,50 +316,51 @@ impl<Data> fmt::Debug for ConnectionState<'_, '_, Data> {
...
 
     /// Decrypts and returns the next available app-data record
     // TODO deprecate in favor of `Iterator` implementation, which requires in-place decryption
     pub fn next_record(&mut self) -> Option<Result<AppDataRecord<'_>, Error>> {
-        if self.taken {
-            None
-        } else {
-            self.taken = true;
-            Some(Ok(AppDataRecord {
+        self.chunk = self
```
djc, line 343: On `ReadTraffic` `Drop`, should we put back the `chunk` if `Some`?

(I guess this is covered by your commit message saying that `next_record()` remains one-shot.)

rustls/src/conn/unbuffered.rs
```diff
@@ -369,52 +369,47 @@ impl<'c, 'i, Data> ReadTraffic<'c, 'i, Data> {
...
     /// decrypts and returns the next available app-data record
     // TODO deprecate in favor of `Iterator` implementation, which requires in-place decryption
     pub fn next_record(&mut self) -> Option<Result<AppDataRecord<'_>, Error>> {
-        if self.taken {
-            None
-        } else {
-            self.taken = true;
-            Some(Ok(AppDataRecord {
+        self.chunk = self.conn.pop_early_data();
+        self.chunk.as_ref().map(|ch| {
```
djc, line 398: As above.

rustls/src/server/server_conn.rs
```diff
@@ -908,6 +908,10 @@ impl DerefMut for UnbufferedServerConnection {
 }
 
 impl UnbufferedConnectionCommon<ServerConnectionData> {
+    pub(crate) fn peek_early_data(&self) -> Option<&[u8]> {
```
djc, line 911: Nit: move this down a bit, too?

rustls/src/server/server_conn.rs
```diff
@@ -1051,6 +1055,13 @@ impl EarlyDataState {
         matches!(self, Self::Rejected)
     }
 
+    fn peek(&self) -> Option<&[u8]> {
+        match self {
```
djc, line 1059: Nit: I think it would be more idiomatic to use match ergonomics (`match &self { .. }` and drop the `ref` keyword).

djc (approved):
Not a very in-depth review, I don't have all the unbuffered API context swapped in anymore.

rustls/src/conn/unbuffered.rs
ctz: No. This is the point at which the chunk is consumed. We need to save it in `self.chunk` to get a place for `AppDataRecord` to borrow from.

PR #2340, opened 2025-02-13
Fix daily tests

ctz:
There were two issues here:

1. Busted hickory
Taking hickory-proto 0.25.0-alpha.5 without hickory-resolver produced something that built but couldn't do any lookups: https://github.com/rustls/rustls/actions/runs/13292247381/job/37115669161

  We previously pinned this because prereleases have no particular stability guarantees, but since we should be only using `--locked` dependencies, this is actively unhelpful and this dependency is only in example and test code.

2. Embarrassing overlooking of unbuffered examples
Fallout from #2332 

Working run with these changes: https://github.com/rustls/rustls/actions/runs/13307896223/workflow

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -766,6 +766,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


ctz:
(merging with one review because I think it's mostly uncontroversial)

cpu (approved):

PR #2342, opened 2025-02-13
Take rustls-webpki 0.103.0 and improve certificate error reporting

ctz:
This is a preview of the error reporting improvements we can do once we release webpki 0.103.0. I think it looks good, and if other folks agree I think we can go ahead and release that.

### DIFF
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -540,6 +540,48 @@ impl From<CertificateError> for AlertDescription {
     }
 }
 
+impl fmt::Display for CertificateError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            #[cfg(feature = "std")]
+            Self::NotValidForNameContext {
+                expected,
+                presented,
+            } => {
+                write!(
+                    f,
+                    "certificate not valid for name {:?}; certificate ",
+                    expected.to_str()
+                )?;
+
+                match presented.as_slice() {
+                    &[] => write!(
+                        f,
+                        "is not valid for any names (according to its subjectAltName extension)"
+                    ),
+                    [one] => write!(f, "is only valid for {}", one),
+                    many => {
+                        write!(f, "is only valid for ")?;
+
+                        let n = many.len();
+                        let all_but_last = &many[..n - 1];
+                        let last = &many[n - 1];
+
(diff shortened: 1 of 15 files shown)


cpu (commented):
This looks good to me. I think it'll be a nice UX improvement downstream and seems to have the least downsides compared to the other approaches we kicked around in discussion. Thanks!

rustls/src/error.rs
```diff
@@ -405,6 +422,16 @@ impl PartialEq<Self> for CertificateError {
...
+            (
+                NotValidForNameDetail {
+                    expected: le,
+                    presented: lp,
+                },
+                NotValidForNameDetail {
+                    expected: re,
+                    presented: rp,
+                },
+            ) => le.eq(re) && lp.eq(rp),
```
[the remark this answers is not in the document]
ctz: did this, and also used tuples which i think is a bit more readable

rustls/src/error.rs
```diff
@@ -540,6 +540,48 @@ impl From<CertificateError> for AlertDescription {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            #[cfg(feature = "std")]
+            Self::NotValidForNameDetail {
+                expected,
+                presented,
+            } => {
+                write!(
+                    f,
+                    "certificate not valid for name {:?}. certificate ",
```
djc, line 553: Nit: don't love the period in here, maybe semi-colon or comma?

rustls/src/error.rs
```diff
@@ -577,6 +577,42 @@ impl fmt::Display for CertificateError {
                     }
                 }
             }
+
+            Self::ExpiredDetail { time, not_after } => write!(
+                f,
+                "certificate expired: current UNIX time is {}, \
```
djc, line 583: Maybe "current" -> "verification", or even "current UNIX time" -> "verified at {} (UNIX)"?

djc (approved):
Nice.

I don't like the `..Detail` suffix much on the error variant names, because it doesn't read very naturally? I guess I'd stick with `..Context` which is admittedly also not great.

ctz:
> (Have already approved this -- would you like me to make another pass?)

I think it's OK; but as always happy to address any follow-ups in a later PR.

PR #2344, opened 2025-02-17
Take semver-compatible dependencies

ctz:
Just a plain `cargo update` to track *ring* performance updates.

I tried to prod renovate to do this, but it's ignoring me. Possibly because it is on strike due to #2317?


ctz:
> I tried to prod renovate to do this, but it's ignoring me. Possibly because it is on strike due to #2317?

That doesn't make sense.

cpu (approved):

PR #2345, opened 2025-02-17
unbuffered: introduce dangerous_extract_secrets, analogous to buffered API

edef1c:
This moves the actual *implementation* of dangerous_extract_secrets into ConnectionCore, since that's a) where the goods actually are, and b) shared between the unbuffered and buffered APIs. This is only made `pub(crate)` and outfitted with the appropriate facades for public API, for both the new and old `dangerous_extract_secrets`.

The test is largely copied from the buffered API tests, and doesn't provide full coverage. They are testing the same underlying code, so this seems minor compared to creating further unnecessary duplication.

# Proposed release notes
* Expose `dangerous_extract_secrets` on UnbufferedClientConnection, UnbufferedServerConnection, and UnbufferedConnectionCommon

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -18,7 +18,7 @@ use crate::log::trace;
 use crate::msgs::enums::NamedGroup;
 use crate::msgs::handshake::ClientExtension;
 use crate::msgs::persist;
-use crate::suites::SupportedCipherSuite;
+use crate::suites::{ExtractedSecrets, SupportedCipherSuite};
 use crate::sync::Arc;
 #[cfg(feature = "std")]
 use crate::time_provider::DefaultTimeProvider;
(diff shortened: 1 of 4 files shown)


djc (approved):
LGTM, thanks!

djc:
(If you'd like a release containing these changes, suggest adding a commit with a version bump -- update both of the `Cargo.lock` files -- and propose some release notes in the PR description.)

edef1c:
> (If you'd like a release containing these changes, suggest adding a commit with a version bump -- update both of the `Cargo.lock` files -- and propose some release notes in the PR description.)

I've added release notes, but I probably won't have time to migrate my own codebase to the unbuffered API for a while yet. So assuming this is the conditional ask I parsed it as, I'm personally fine if this gets merged without a bump and rolled into a release later.

ctz (approved):

cpu (approved):
Nice and tidy, thanks!

PR #2347, opened 2025-02-19
bogo: implement -wait-for-debugger in shim

cpu:
Also updates the `runme` script to conditionally exclude the `-test.timeout` flag given to the BoGo runner when `-wait-for-debugger` is in use.

The implementation in the shim depends on UNIX SIGSTOP signalling, and so is conditional to UNIX-like systems. Providing the flag on another platform will panic with an error message.

Having this support makes it possible to debug the Rustls-side of the BoGo testing process by having the BoringSSL-side Go test runner start the shim process in a suspended state. In this mode the test runner turns off some timeouts, reduces parallelism to 1 task at a time, and prints the PID of the shim process to attach to with a Rust-friendly debugger. By using `delve` or another Go debugger it's also possible to debug the runner-side in conjugation with the shim. This should be a nice alternative to the tedious print-style-debugging I've used in the past :-)

### DIFF
--- a/bogo/Cargo.toml
+++ b/bogo/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 base64 = { workspace = true }
 env_logger = { workspace = true }
+nix = { version = "0.29", default-features = false, features = ["signal"] }
 rustls = { path = "../rustls", features = ["aws_lc_rs", "ring", "tls12"] }
 rustls-post-quantum = { path = "../rustls-post-quantum", optional = true }
 
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -10,6 +10,10 @@ use std::sync::Arc;
 use std::{env, net, process, thread, time};
 
 use base64::prelude::{BASE64_STANDARD, Engine};
+#[cfg(unix)]
+use nix::sys::signal::{self, Signal};
+#[cfg(unix)]
+use nix::unistd::Pid;
 use rustls::client::danger::{HandshakeSignatureValid, ServerCertVerified, ServerCertVerifier};
 use rustls::client::{
     ClientConfig, ClientConnection, EchConfig, EchGreaseConfig, EchMode, EchStatus, Resumption,
@@ -107,6 +111,7 @@ struct Options {
     expect_curve_id: Option<NamedGroup>,
     on_initial_expect_curve_id: Option<NamedGroup>,
     on_resume_expect_curve_id: Option<NamedGroup>,
+    wait_for_debugger: bool,
 }
 
 impl Options {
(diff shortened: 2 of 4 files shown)


bogo/Cargo.toml
```diff
@@ -8,6 +8,7 @@ base64 = { workspace = true }
 env_logger = { workspace = true }
 rustls = { path = "../rustls", features = ["aws_lc_rs", "ring", "tls12"] }
 rustls-post-quantum = { path = "../rustls-post-quantum", optional = true }
+nix = { version = "0.29", default-features = false, features = ["signal"] }
```
djc, line 11: Nit: alphabetic order?

djc (approved):

bogo/src/main.rs
```diff
@@ -1600,6 +1606,14 @@ pub fn main() {
             "-server-preference" => {
                 opts.server_preference = true;
             }
+            "-wait-for-debugger" => {
+                #[cfg(windows)] {
```
djc, line 1610: This formatting looks surprising to me, is rustfmt skipping this? I think the usual formatting would be like

```rust
#[cfg(windows)]
{
    panic(..);
}
```

cpu: > is rustfmt skipping this?

oh bother, I think it must be. I'll take a look.

cpu: Hmmmm I can't get a local `cargo fmt` to complain, but it does seem to be looking at `bogo/src/main.rs`:

```
[daniel@noire:~/Code/Rust/rustls/bogo]$ git rev-parse HEAD
edd6a68c50d6e65bf37ee71884bcf17619d4e6ed

[daniel@noire:~/Code/Rust/rustls/bogo]$ cargo fmt --all --verbose --check | grep bogo
[bin (2021)] "/home/daniel/Code/Rust/rustls/bogo/src/main.rs"
[test (2021)] "/home/daniel/Code/Rust/rustls/rustls/tests/bogo.rs"
rustfmt --edition 2021 --check /home/daniel/Code/Rust/rustls/bogo/src/main.rs <snipped>
```

bogo/Cargo.toml
cpu: Fixed. Thanks!

bogo/src/main.rs
cpu: Fixed manually for now.

djc: Weird -- it doesn't seem to be in a macro context, either.

ctz: I think maybe rustfmt gives up if it can't find a way to format some code to fit in the line width? And this code is pretty tall and wide.

cpu:
> cpu force-pushed the cpu-bogo-wait-for-debugger branch from caaa692 to ff191bb 

Addressed conflicts and also tacked on a commit like https://github.com/rustls/rustls/commit/ede4b26fa5e24de5ced0ea874fd3a97e18e18a9e for the BoGo shim. A `cargo clippy` from the `bogo` dir was flagging `Arc` usage.

djc (approved):

ctz (approved):

PR #2350, opened 2025-02-23
specify recent once_cell version in a single place

brody4hire:


Cargo.toml
```diff
@@ -67,7 +67,7 @@ log = { version = "0.4.8" }
 macro_rules_attribute = "0.2"
 mio = { version = "1", features = ["net", "os-poll"] }
 num-bigint = "0.4.4"
-once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
+once_cell = { version = "1.20.3", default-features = false, features = ["alloc", "race"] }
```
cpu, line 70: Can we use "1" here?

cpu (commented):
Is there a reason you opened this as a draft? The general change seems good to me.

Cargo.toml
brody4hire: Yes but I would *personally* downvote that idea.

I would personally vote against supporting too many outdated dependencies unless there is some kind of strong and documented motivation for this, supporting older MSRV could be an interesting example.

Outdated dependencies may have some bugs, potentially even security issues, that have already been resolved. Also I do not recall there is CI testing with the older dependency versions.

Leaving this decision up to you. I started this as a draft until I had the second commit ready. I would be happy to squash if needed.

rustls/Cargo.toml
```diff
@@ -40,7 +40,7 @@ brotli-decompressor = { workspace = true, optional = true }
 hashbrown = { workspace = true, optional = true }
 log = { workspace = true, optional = true }
 # only required for no-std
-once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
+once_cell = { workspace = true, default-features = false, features = ["alloc", "race"] }
```
ctz, line 43: I think the `, default-features = false, features = ["alloc", "race"] ` can be dropped here?

Cargo.toml
ctz: I suggest keeping this as "1.16". The lowest we can write here is "1.7" (introduction of the "race" feature).

Generally the minor number should be moved forward only when a specific API is required; this gives cargo the widest possible set of versions to select from. And the patch number should only be specified if we have a known, specific requirement on a fix it introduced.

djc: If we really want to sweat the details on this, we'd use cargo-minimal-versions to find (and then potentially test in CI) that we depend on the minimal version.

I don't think we should unnecessarily bump the requirement.

ctz: 👍🏽 we do that here https://github.com/rustls/rustls/blob/main/.github/workflows/build.yml#L340-L343

djc: But I guess that only checks in one direction: whether the required version is still good enough. The more ambitious version would be to also do the bisection to figure out the oldest version that still allows us to pass the test suite.

rustls/Cargo.toml
brody4hire: done

PR #2353, opened 2025-02-24
docs: minor improvements to CryptoProvider doc

brody4hire:
__STATUS:__ I think this should be ready for review. I made comments on a couple things I noticed, don't think they should be blocking. Thanks.

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -110,7 +110,7 @@ pub use crate::suites::CipherSuiteCommon;
 ///
 /// # Making a custom `CryptoProvider`
 ///
-/// Your goal will be to populate a [`crypto::CryptoProvider`] struct instance.
+/// Your goal will be to populate an instance of this [`crypto::CryptoProvider`] struct.
 ///
 /// ## Which elements are required?
 ///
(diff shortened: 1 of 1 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -120,7 +120,7 @@ pub use crate::suites::CipherSuiteCommon;
 /// provider (dynamically).
 ///
 /// For example, if we want to make a provider that just overrides key loading in the config builder
-/// API ([`ConfigBuilder::with_single_cert`] etc.), it might look like this:
+/// API (with [`ConfigBuilder::with_single_cert`], etc.), it might look like this:
```
brody4hire, line 123: Without these updates, the line seems to breaks right after `API (` in my browser.

With these updates, the line seems to break after `API (with` in my browser.

I think this is a slight improvement, wonder if anyone can think of something better for this.

djc (approved):

brody4hire:
I just pushed another update, guessing it will miss the merge queue. I guess I will open another PR for this.

rustls/src/crypto/mod.rs
```diff
@@ -110,7 +110,7 @@ pub use crate::suites::CipherSuiteCommon;
 ///
 /// # Making a custom `CryptoProvider`
 ///
-/// Your goal will be to populate a [`crypto::CryptoProvider`] struct instance.
+/// Your goal will be to populate an instance of this [`crypto::CryptoProvider`] struct.
```
brody4hire, line 113: I would like to do this update a little differently, thankfully it missed the merge queue. I would like to remove the brackets, as it seems pretty useless for this page to reference itself here. Using `CryptoProvider` with no brackets also seems more consistent with how this is done elsewhere for `CryptoProvider` in the documentation. PR should be coming soon.

PR #2355, opened 2025-02-24
docs: improve some more links, etc.

brody4hire:
__STATUS:__ KEEPING AS DRAFT FOR NOW - KNOWN TODO ITEM(S) BELOW

- update link to custom provider example in top-level documentation
- update wording in __Making a custom `CryptoProvider`__ section
- replace extra self-linking with unlinked reference to `CryptoProvider` struct, which I think is is more consitent with other parts of documentation for `CryptoProvider`
- update reference to RustCrypto which now uses "upper camel case" instead of kebab-case
- use backticks for multiple references in __Example code__ section
- use backticks for a couple more references in `crypto/mod.rs` that I missed in recent other PR #2353
- adding brackets to add linking to an enum & 2 traits referenced in some text in `crypto/mod.rs`
- add the word "custom" before reference to `provider-example/`

__TODO__

- [x] resolve a couple dead links I found in: https://docs.rs/portable-rustls/latest/portable_rustls/manual/_03_howto/index.html#customising-private-key-usage
- ~~add more explicit reference to the manual that is referenced here & there~~ - see PR #2356
- [x] link to https://github.com/rustls/rustls-cng/blob/dev/src/signer.rs could use some improvement in howto.rs

__FURTHER INVESTIGATION DEFERRED FOR NOW__

- I noticed that some (I think most) but not all links to methods show `()`, found some methods without `()` for example:
  - `ServerConfig::require_ems` & `ClientConfig::require_ems` in lib.rs
  - `crypto::signer::SigningKey::choose_scheme` in `manual/howto.rs`
  - `ConfigBuilder::with_single_cert` in `crypto/mod.rs`

---

__RELATED UPDATES__

- PR #2353 - minor updates to `CryptoProvider` already merged
- PR #2356 - proposed updates for "detailed manual"

---

__MOVING FORWARD__

My apologies for so much noise. I think this should be ready for review. A bunch of things are combined, and I would be happy to separate some of these things if needed.

I have raised some more PRs that may be related to this. I would be happy to combine or separate changes out in a different way if needed.

### DIFF
--- a/rustls/src/manual/howto.rs
+++ b/rustls/src/manual/howto.rs
@@ -6,7 +6,7 @@ However, if your private key resides in a HSM, or in another process, or perhaps
 another machine, rustls has some extension points to support this:
 
 The main trait you must implement is [`sign::SigningKey`][signing_key]. The primary method here
-is [`choose_scheme`][choose_scheme] where you are given a set of [`SignatureScheme`s][sig_scheme] the client says
+is [`choose_scheme()`][choose_scheme] where you are given a set of [`SignatureScheme`s][sig_scheme] the client says
 it supports: you must choose one (or return `None` -- this aborts the handshake). Having
 done that, you return an implementation of the [`sign::Signer`][signer] trait.
 The [`sign()`][sign_method] performs the signature and returns it.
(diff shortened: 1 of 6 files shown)


rustls/src/manual/howto.rs
```diff
@@ -6,7 +6,7 @@ However, if your private key resides in a HSM, or in another process, or perhaps
 another machine, rustls has some extension points to support this:
 
 The main trait you must implement is [`sign::SigningKey`][signing_key]. The primary method here
-is [`choose_scheme`][choose_scheme] where you are given a set of [`SignatureScheme`s][sig_scheme] the client says
+is [`choose_scheme()`][choose_scheme] where you are given a set of [`SignatureScheme`s][sig_scheme] the client says
```
brody4hire, line 9: reverting this for now:

```suggestion
is [`choose_scheme`][choose_scheme] where you are given a set of [`SignatureScheme`s][sig_scheme] the client says
```

It looks to me like links like this do not naturally show `()` for member functions:

    [`ConfigBuilder::with_single_cert`]

I would like to investigate this sometime later as I DO see some other links like this:

    It is used for functions like [`ClientConfig::builder()`] and [`ServerConfig::builder()`].

brody4hire: never mind ... I think we DO want to keep the `()` here, which is both more consistent with reference to `sign()` method later in this paragraph and is more clear since we do not show the path to this method here

ctz:
Just a reminder that every PR or change to a PR notifies every maintainer. This PR especially has been very noisy.

ctz (approved):

cpu (approved):

PR #2359, opened 2025-02-25
Avoid handshake message round-tripping for binders

ctz:
Verifying a _received_ `ClientHello` binder should be done against the original received bytes, not our re-encoding of them.

This previously worked, because we required and tested that we could accurately round-trip handshake messages. But that is a less-robust approach.

(change split out from #1475)


djc (approved):

djc:
Nice. Was just thinking about #1475 this morning.

cpu (approved):
:+1:

PR #2361, opened 2025-02-28
Prevent resumption between "incompatible" clients

### LINKED ISSUE #2284: Default to more precise segregation of resumed sessions
It is currently possible to construct two `ClientConfig`s that share a resumption store; this in itself is desirable (eg, see #2260 and openssl-compat). However, those `ClientConfig`s may diverge in ways that make sharing their resumed sessions incorrect. Such as:

1. variance of server certificate validation (in a minor sense, such as different root certs, or a major sense, like whether certificate validation happens at all).
2. variance of client authentication identities,

Getting into that situation doesn't happen by magic, and an application would need to write code to arrange it, but I think it would be surprising.

The best behaviour would be to have resumption keyed on a wider set of data  than just the server name (AFAICT this would mean multiple breaking API changes, because the verifier traits have no equality relation).

An alternative and partial measure would be to reverify the server's certificate on resumption -- this would cost significant performance and only address (1) but not (2). Allowing that separately might be a good idea for other reasons, as we don't really have any reason to assume the output of cert validation is stable over a session.

ctz:
"Compatible" here means they have ~interchangeable security, which means they have the same server certificate verifier and same potentially-offered client credentials.
    
"Same" is defined by `Arc` equality, which means a rustls user wishing to arrange for multiple `ClientConfig`s to share a `resumption` _also_ now need to share the `client_auth_cert_resolver` and `verifier`.  A non-`dangerous` function is provided which does this, and also provides a convenient place to document the problem we are solving here.

If we think this is a reasonably-shaped solution to this problem, I will do similar for servers.

fixes #2284

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -166,6 +166,22 @@ pub struct ClientConfig {
     pub alpn_protocols: Vec<Vec<u8>>,
 
     /// How and when the client can resume a previous session.
+    ///
+    /// # Sharing `resumption` between `ClientConfig`s
+    /// In a program using many `ClientConfig`s it may improve resumption rates
+    /// (which has a significant impact on connection performance) if those
+    /// configs share a single `Resumption`.
+    ///
+    /// However, resumption is only allowed between two `ClientConfig`s if their
+    /// `client_auth_cert_resolver` (ie, potential client authentication credentials)
+    /// and `verifier` (ie, server certificate verification settings) are
+    /// the same (according to `Arc::ptr_eq`).
+    ///
+    /// To illustrate, imagine two `ClientConfig`s `A` and `B`.  `A` fully validates
+    /// the server certificate, `B` does not.  If `A` and `B` shared a resumption store,
+    /// it would be possible for a session originated by `B` to be inserted into the
+    /// store, and then resumed by `A`.  This would give a false impression to the user
+    /// of `A` that the server certificate is fully validated.
     pub resumption: Resumption,
 
     /// The maximum size of plaintext input to be emitted in a single TLS record.
--- a/rustls/src/msgs/persist.rs
+++ b/rustls/src/msgs/persist.rs
@@ -4,6 +4,7 @@ use core::cmp;
 use pki_types::{DnsName, UnixTime};
 use zeroize::Zeroizing;
 
+use crate::client::ResolvesClientCert;
 use crate::enums::{CipherSuite, ProtocolVersion};
 use crate::error::InvalidMessage;
 use crate::msgs::base::{PayloadU8, PayloadU16};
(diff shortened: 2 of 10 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -369,6 +369,35 @@ impl ClientConfig {
...
+    /// `client_auth_cert_resolver` (ie, potential client authentication credentials)
+    /// and `verifier` (ie, server certificate verification settings) are
+    /// the same (according to `Arc::ptr_eq`).
+    ///
+    /// To illustrate, imagine two `ClientConfig`s `A` and `B`.  `A` fully validates
+    /// the server certificate, `B` does not.  If `A` and `B` shared a resumption store,
+    /// it would be possible for a session originated by `B` to be inserted into the
+    /// store, and then resumed by `A`.  This would give a false impression to the user
+    /// of `A` that the server certificate is fully validated.
+    pub fn share_resumption_with(&mut self, other: &Self) {
```
djc, line 395: Do we know that anyone needs this? It feels a little YAGNI.

rustls/src/msgs/persist.rs
```diff
@@ -213,6 +223,8 @@ pub struct ClientSessionCommon {
     epoch: u64,
     lifetime_secs: u32,
     server_cert_chain: Arc<CertificateChain<'static>>,
+    server_cert_verifier: Weak<dyn ServerCertVerifier>,
```
djc, line 226: So does this mean that persistent client caches now become impossible?

djc:
What about the server? Is there something that prevents similar incompatible resumption issues across server configs?

rustls/src/msgs/persist.rs
ctz: The client session types have been in-memory-only for a while (in contrast to the server types, which have a serialized representation). If/when we reinstate that, we'll also need to address `Tls12ClientSessionValue::suite`.

ctz:
> What about the server? Is there something that prevents similar incompatible resumption issues across server configs?

Yeah I'd probably do the same for `ClientCertVerifier` and `ResolvesServerCert` on servers. Deserialized sessions would be free from the equality restrictions.

cpu (approved):

rustls/src/client/client_conn.rs
ctz: Do you mean the whole feature, or this function? The problem this function is solving comes from `ClientConfig::verifier` being private, so someone can't write the body of this function themselves.

djc: I mean this function -- not so much whether people can write it another way, but whether anyone will need it.

ctz: Removed for now.

ctz:
(Plan to address the coverage gaps here shortly.)

PR #2366, opened 2025-03-07

djc:
Avoid warnings about [RUSTSEC-2025-0009](https://rustsec.org/advisories/RUSTSEC-2025-0009.html).


ctz (approved):

djc:
@ctz are we ignoring this semver compatibility stuff? Have you already looked into ways of `allow`ing certain lints locally?

PR #2367, opened 2025-03-07
Avoid semver errors on discriminant changes

djc:


ctz (approved):
lol, here's me thinking "what if we add a dummy structure variant to every public enum"

cpu:
Thanks for figuring this out 👍

PR #2369, opened 2025-03-07
Allow `AeadKey` to be constructed outside crate

ctz:
Previously this was constructable, but only for 256-bit keys via `From<[u8; 32]>`.

Instead of just publicising `AeadKey::new`, recast the type as an enum containing variants for the two usual symmetric key lengths (this prevents misuses like empty keys, one byte keys, etc).  Unfortunately for semver reasons this needs to be wrapped in the original struct type.

The underlying rationale for this change is to support some changes I want to make in quinn; this is draft while I solidify those.

### DIFF
--- a/rustls/src/crypto/cipher.rs
+++ b/rustls/src/crypto/cipher.rs
@@ -270,52 +270,55 @@ const TLS12_AAD_SIZE: usize = 8 + 1 + 2 + 2;
 
 /// A key for an AEAD algorithm.
 ///
-/// This is a value type for a byte string up to `AeadKey::MAX_LEN` bytes in length.
-pub struct AeadKey {
-    buf: [u8; Self::MAX_LEN],
-    used: usize,
+/// This is a value type for a byte string of 128- or 256-bits.
+pub struct AeadKey(AeadKeyInner);
+
+enum AeadKeyInner {
+    /// 128-bit key
+    B128([u8; 16]),
+    /// 256-bit key
+    B256([u8; 32]),
 }
 
 impl AeadKey {
+    /// Implemented only for 16-byte or 32-byte `buf`.
     #[cfg(feature = "tls12")]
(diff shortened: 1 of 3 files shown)


rustls/src/crypto/cipher.rs
```diff
@@ -270,52 +270,56 @@ const TLS12_AAD_SIZE: usize = 8 + 1 + 2 + 2;
...
 ///
-/// This is a value type for a byte string up to `AeadKey::MAX_LEN` bytes in length.
-pub struct AeadKey {
-    buf: [u8; Self::MAX_LEN],
-    used: usize,
+/// This is a value type for a byte string of 128- or 256-bits.
+#[non_exhaustive]
+pub struct AeadKey(AeadKeyInner);
+
+#[non_exhaustive]
```
djc, line 277: Don't think these need `#[non_exhaustive]` -- the former because its fields are private anyway, the latter because the type itself is private.

rustls/src/crypto/cipher.rs
```diff
@@ -270,52 +270,56 @@ const TLS12_AAD_SIZE: usize = 8 + 1 + 2 + 2;
...
-        key.used = buf.len();
-        key
-    }
-
-    pub(crate) fn with_length(self, len: usize) -> Self {
-        assert!(len <= self.used);
-        Self {
-            buf: self.buf,
-            used: len,
+    pub(crate) fn from_slice(buf: &[u8]) -> Self {
```
djc, line 287: It feels like this should be fallible, or just skip it entirely?

djc:
In an experimental project I'm working on I included an `Unreachable` variant in my `Error` type so that I can yield those instead of calling `unreachable!()`. Do you think that would be a good strategy for this 1.2 key block splitting stuff? The `unwrap()` stuff is not ideal... (To be fair, fake `Result` types are also not great...)

ctz:
Will bring this back sometime later.

PR #2370, opened 2025-03-08
Add kernel connection API

### LINKED ISSUE #2362: Better support for kTLS
**Checklist**
* [x] I've searched the issue tracker for similar requests: The last kTLS issue was #198 which was resolved by adding the equivalent of `dangerous_extract_secrets` in #1112.

### Is your feature request related to a problem? Please describe.
I have been attempting to make use of rustls to set up and drive a kTLS connection. As things stand today, you can use rustls to drive the TLS handshake, then extract the traffic secrets via `dangerous_extract_secrets`, and use those to configure the socket. This part works great!

The problems start once kTLS is set up. Sending/receiving application works great, but you will also receive control messages (any TLS message other than `application_data`) that need to be dealt with. In practice, all that it is really practical to do is:
- handle `close_notify`
- abort on receiving any other alert
- abort on receiving `key_update`
- ignore any other non-`application_data` message

You can see this in the [ktls crate](https://github.com/rustls/ktls/blob/main/ktls/src/ktls_stream.rs#L117), my code ends up looking fairly similar.

### Describe the solution you'd like
kTLS already handles encryption/decryption and message (de)fragmentation (i.e. the TLS record layer).  I would like to have a client variant where I can pass in an (effectively) `InboundPlainMessage` and get back a sequence of actions/events (e.g. send this control message, update tx/rx secrets, peer closed, etc.).

### Describe alternatives you've considered
1. Don't do anything. Continue to ignore most TLS control messages and abort the connection on `key_update`.
2. Expose the actual application secret and other rustls bits (such as storage for session tickets) and reimplement most of the TLS protocol

### Additional context
I have been poking around the rustls source to see what would need to be done to expose an API for this. The biggest challenge has proven to be extracting a TLS record layer abstraction (note: not just `RecordLayer`) out of `CommonState` without breaking things.

If you decide you're willing to accept code for this then I would be willing to do the work to get it in.

swlynch99:
This is an attempt at addressing #2362 by following the API suggested there. 

It introduces a new `ExternalConnection` type which allows users to pass received session tickets back to rustls and perform key updates. An `ExternalConnection` can be constructed from an `Unbuffered(Client|Server)Connection` by calling `dangerous_into_external_connection` which will return both an `ExtractedSecrets` and an `ExternalConnection`. It does not include support any other connection types at this time. However, it should be easy enough to extend in the future as more as new features become needed.

Internally this is implemented by adding a new `ExternalState` trait and a `State::into_external_state` conversion method. `ExternalState` is implemented for the `ExpectTraffic` state for tls 1.2 and 1.3, and for server and client modes, respectively. 

I have also included a complete example that shows how to use `ExternalConnection` with a kTLS connection. It ends up serving more as an example in getting kTLS to work using rustls than actually using `ExternalConnection`, since `ExternalConnection` has a pretty minimal API surface, but I think it works better that way.

This PR ended up being quite a bit larger than I thought it would be initially. If there's anything I can do that would help make this easier to review (e.g. moving the ktls example to a separate PR), please let me know.

Closes #2362

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -16,6 +16,7 @@ use crate::conn::{ConnectionCore, UnbufferedConnectionCommon};
 use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::error::Error;
+use crate::kernel::KernelConnection;
 use crate::log::trace;
 use crate::msgs::enums::NamedGroup;
 use crate::msgs::handshake::ClientExtension;
@@ -898,6 +899,23 @@ impl UnbufferedClientConnection {
     /// Should be used with care as it exposes secret key material.
     pub fn dangerous_extract_secrets(self) -> Result<ExtractedSecrets, Error> {
         self.inner.dangerous_extract_secrets()
+    }
+
+    /// Extract secrets and a [`KernelConnection`] object.
+    ///
+    /// This allows you use rustls to manage keys and then manage encryption and
+    /// decryption yourself (e.g. for kTLS).
+    ///
+    /// Should be used with care as it exposes secret key material.
+    ///
+    /// See the [`crate::kernel`] documentations for details on prerequisites
+    /// for calling this method.
+    pub fn dangerous_into_kernel_connection(
+        self,
+    ) -> Result<(ExtractedSecrets, KernelConnection<ClientConnectionData>), Error> {
+        self.inner
+            .core
+            .dangerous_into_kernel_connection()
     }
 }
 
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -897,6 +897,8 @@ impl UnbufferedClientConnection {
 
     /// Extract secrets, so they can be used when configuring kTLS, for example.
     /// Should be used with care as it exposes secret key material.
+    #[deprecated = "dangerous_extract_secrets() does not support session tickets or \
+                    key updates, use dangerous_into_kernel_connection() instead"]
     pub fn dangerous_extract_secrets(self) -> Result<ExtractedSecrets, Error> {
         self.inner.dangerous_extract_secrets()
     }
--- a/rustls/src/conn/kernel.rs
+++ b/rustls/src/conn/kernel.rs
@@ -1,0 +1,269 @@
+//! Kernel connection API.
+//!
+//! This module gives you the bare minimum you need to implement a TLS connection
+//! that does its own encryption and decryption while still using rustls to manage
+//! connection secrets and session tickets. It is intended for use cases like kTLS
+//! where you want to use rustls to establish the connection but want to use
+//! something else to do the encryption/decryption after that.
+//!
+//! There are only two things that [`KernelConnection`] is able to do:
+//! 1. Compute new traffic secrets when a key update occurs.
+//! 2. Save received session tickets sent by a server peer.
+//!
+//! That's it. Everything else you will need to implement yourself.
+//!
+//! # Entry Point
+//! The entry points into this API are
+//! [`UnbufferedClientConnection::dangerous_into_kernel_connection`][client-into]
+//! and
+//! [`UnbufferedServerConnection::dangerous_into_kernel_connection`][server-into].
+//!
+//! In order to actually create an [`KernelConnection`] all of the following
+//! must be true:
+//! - the connection must have completed its handshake,
+//! - the connection must have no buffered TLS data waiting to be sent, and,
+//! - the config used to create the connection must have `enable_extract_secrets`
+//!   set to true.
+//!
+//! This sounds fairly complicated to achieve at first glance. However, if you
+//! drive an unbuffered connection through the handshake until it returns
+//! [`WriteTraffic`] then it will end up in an appropriate state to convert
+//! into an external connection.
+//!
+//! [client-into]: crate::client::UnbufferedClientConnection::dangerous_into_kernel_connection
+//! [server-into]: crate::server::UnbufferedServerConnection::dangerous_into_kernel_connection
+//! [`WriteTraffic`]: crate::unbuffered::ConnectionState::WriteTraffic
+//!
+//! # Cipher Suite Confidentiality Limits
+//! Some cipher suites (notably AES-GCM) have vulnerabilities where they are no
+//! longer secure once a certain number of messages have been sent. Normally,
+//! rustls tracks how many messages have been written or read and will
+//! automatically either refresh keys or emit an error when approaching the
+//! confidentiality limit of the cipher suite.
+//!
+//! [`KernelConnection`] has no way to track this. It is the responsibility
+//! of the user of the API to track approximately how many messages have been
+//! sent and either refresh the traffic keys or abort the connection before the
+//! confidentiality limit is reached.
+//!
+//! You can find the current confidentiality limit by looking at
+//! [`CipherSuiteCommon::confidentiality_limit`] for the cipher suite selected
+//! by the connection.
+//!
+//! [`CipherSuiteCommon::confidentiality_limit`]: crate::CipherSuiteCommon::confidentiality_limit
+//! [`KernelConnection`]: crate::kernel::KernelConnection
+
+use core::marker::PhantomData;
+
+use alloc::boxed::Box;
+
+use crate::client::ClientConnectionData;
+use crate::common_state::Protocol;
+use crate::msgs::codec::Codec;
+use crate::msgs::handshake::{CertificateChain, NewSessionTicketPayloadTls13};
+use crate::quic::Quic;
+use crate::{CommonState, ConnectionTrafficSecrets, Error, ProtocolVersion, SupportedCipherSuite};
+
+/// A kernel connection.
+///
+/// This does not directly wrap a kernel connection, rather it gives you the
+/// minimal interfaces you need to implement a well-behaved TLS connection on
+/// top of kTLS.
+///
+/// See the [`crate::kernel`] module docs for more details.
+pub struct KernelConnection<Data> {
+    state: Box<dyn KernelState>,
+
+    peer_certificates: Option<CertificateChain<'static>>,
+    quic: Quic,
+
+    negotiated_version: ProtocolVersion,
+    protocol: Protocol,
+    suite: SupportedCipherSuite,
+
+    _data: PhantomData<Data>,
+}
+
+impl<Data> KernelConnection<Data> {
+    pub(crate) fn new(state: Box<dyn KernelState>, common: CommonState) -> Result<Self, Error> {
+        Ok(Self {
+            state,
+
+            peer_certificates: common.peer_certificates,
+            quic: common.quic,
+            negotiated_version: common
+                .negotiated_version
+                .ok_or(Error::HandshakeNotComplete)?,
+            protocol: common.protocol,
+            suite: common
+                .suite
+                .ok_or(Error::HandshakeNotComplete)?,
+
+            _data: PhantomData,
+        })
+    }
+
+    /// Retrieves the ciphersuite agreed with the peer.
+    pub fn negotiated_cipher_suite(&self) -> SupportedCipherSuite {
+        self.suite
+    }
+
+    /// Retrieves the protocol version agreed with the peer.
+    pub fn protocol_version(&self) -> ProtocolVersion {
+        self.negotiated_version
+    }
+
+    /// Update the traffic secret used for encrypting messages sent to the peer.
+    ///
+    /// Returns the new traffic secret and initial sequence number to use.
+    ///
+    /// In order to use the new secret you should send a TLS 1.3 key update to
+    /// the peer and then use the new traffic secrets to encrypt any future
+    /// messages.
+    ///
+    /// Note that it is only possible to update the traffic secrets on a TLS 1.3
+    /// connection. Attempting to do so on a non-TLS 1.3 connection will result
+    /// in an error.
+    pub fn update_tx_secret(&mut self) -> Result<(u64, ConnectionTrafficSecrets), Error> {
+        // The sequence number always starts at 0 after a key update.
+        self.state
+            .update_secrets(Direction::Transmit)
+            .map(|secret| (0, secret))
+    }
+
+    /// Update the traffic secret used for decrypting messages received from the
+    /// peer.
+    ///
+    /// Returns the new traffic secret and initial sequence number to use.
+    ///
+    /// You should call this method once you receive a TLS 1.3 key update message
+    /// from the peer.
+    ///
+    /// Note that it is only possible to update the traffic secrets on a TLS 1.3
+    /// connection. Attempting to do so on a non-TLS 1.3 connection will result
+    /// in an error.
+    pub fn update_rx_secret(&mut self) -> Result<(u64, ConnectionTrafficSecrets), Error> {
+        // The sequence number always starts at 0 after a key update.
+        self.state
+            .update_secrets(Direction::Receive)
+            .map(|secret| (0, secret))
+    }
+}
+
+impl KernelConnection<ClientConnectionData> {
+    /// Handle a `new_session_ticket` message from the peer.
+    ///
+    /// This will register the session ticket within with rustls so that it can
+    /// be used to establish future TLS connections.
+    ///
+    /// # Getting the right payload
+    ///
+    /// This method expects to be passed the inner payload of the handshake
+    /// message. This means that you will need to parse the header of the
+    /// handshake message in order to determine the correct payload to pass in.
+    /// The message format is described in [RFC 8446 section 4][0]. `payload`
+    /// should not include the `msg_type` or `length` fields.
+    ///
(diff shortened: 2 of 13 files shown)


ctz:
First, I think the `dangerous_into_external_connection()` API is an excellent way to do this and maybe I'd go further to suggest that this PR could deprecate the `dangerous_extract_secrets()` API, and eviscerate its internal implementation (eg, by making `dangerous_extract_secrets()` just do `let (secrets, _) = self.dangerous_into_external_connection()?; Ok(secrets)`).

I think there is space here to reduce the public API scope of `ExternalConnection` quite a bit. Observing the idea that KTLS users are exchanging complexity for big performance and density improvements, I think that committing to keeping things like the peer's certificate around while servicing the connection is not in keeping with that goal. So I would suggest removing much of that API surface for now -- since the caller needs to ensure the handshake is complete _before_ calling `dangerous_into_external_connection()`, if they need the peer certificate, handshake shape, ALPN, etc, they can obtain this before the conversion. (That doesn't stop us from having to store it internally, but gives us some future leeway in how we do that to improve density.)

djc:
Agreed. Also, a bit of bikeshedding: I don't love `External` as it feels a little generic. Was thinking maybe `Traffic` would be a good prefix for this that is a little more meaningful in communicating how it's different?

swlynch99:
I'm going to make changes as new commits for now since it is easier to see what is changing. I'll rework things into a more coherent commit history before this gets merged.

Quick summary of the new changes before I respond to individual comments:
- I have made `ExternalConnection` generic over `Data` so that users with a server-side connection cannot call `handle_new_session_ticket`. This is currently just a facade using `PhantomData` over non-generic internals but it gives freedom in the future if something needs access to `Data`. My main motivation for this, though, is disallowing calls to `handle_new_session_ticket` on server side connections.
- I have removed a bunch of the public API methods from `ExternalConnection`.
- I have reimplemented `dangerous_into_secrets` on top of `dangerous_into_external_connection`.  The one change in behaviour is that it now emits an error if there are messages stored in `sendable_tls` instead of giving unusable secrets.

> I'd go further to suggest that this PR could deprecate the dangerous_extract_secrets() API, and eviscerate its internal implementation.

I have done all of this except the deprecation. I'll add a commit for the deprecation bit once we've decided on the final name for the type.

> I think there is space here to reduce the public API scope of ExternalConnection quite a bit.

I have reduced the API surface down to just the negotiated protocol version and the chosen cipher suite. The protocol version needs to be kept anyways and the cipher suite is the only way to get the confidentiality limit (and is needed for kTLS besides). This has not appreciably reduced the size of `ExternalConnection`, unfortunately, since the quic data is ~600B and everything else is ~80B.

> I don't love External as it feels a little generic. Was thinking maybe Traffic would be a good prefix for this that is a little more meaningful in communicating how it's different?

Up front, I don't have a strong preference here. This is your bikeshed so ultimately you get the final say :)

My thought process behind `ExternalConnection` was something like "it's a connection that's managed externally to rustls" -> `ExternalConnection`. Not sure if that makes it better or worse.

`TrafficConnection` sounds a little off to me. It might be because all the other prefixes are adjectives?. As an alternative, what if we just called it `Connection` like in the quic module? Then you have `external::Connection`, `external::ClientConnection`, and `external::ServerConnection`. I'm not sure if that's better, on second thought, but I'll leave it here as an alternative.

djc:
> `TrafficConnection` sounds a little off to me. It might be because all the other prefixes are adjectives?. As an alternative, what if we just called it `Connection` like in the quic module? Then you have `external::Connection`, `external::ClientConnection`, and `external::ServerConnection`. I'm not sure if that's better, on second thought, but I'll leave it here as an alternative.

That's fair. I think my beef with "external" is that it leaves open the question of "external to what"?

swlynch99:
@djc I've been thinking on and off about the naming for the last few days and have not been able to come up with a better name. So I'm putting this back on your side of the court: if you want me to rename type+module to `TrafficConnection` and `traffic` then I'll do so, if you think it is ok staying as `ExternalConnection` + `external`, then I'll leave it as is. I don't have a strong enough opinion on this to stall the PR over it :)

I will say that when I named it my thought process was that it is a connection that is "external to rustls". i.e. that the user has to implement the connection themselves, whereas for the other connection types the user just provides/receives bytes and rustls handles the rest.

I think I have addressed all the points brought up so far. The only thing left is deprecating `dangerous_external_secrets` which I will do after naming is settled. Let me know if there is anything I can do to make the review easier, as I realize this is a big PR. Otherwise, I'll leave this until you guys have bandwidth to review.

ctz:
I don't mind the `ExternalConnection` naming, but WDYT about `KernelConnection`?

swlynch99:
I have gone ahead and done the rename. I have also deprecated `dangerous_extract_secrets` in favour of the new `dangerous_into_kernel_connection` method. Finally, I have cleaned up the commit history into something a bit more coherent.

rustls/src/conn/kernel.rs
```diff
@@ -0,0 +1,223 @@
...
+///
+/// The only difference for kernel client vs server connections is that client
+/// connections have support for recording new session tickets.
+pub type KernelClientConnection = KernelConnection<ClientConnectionData>;
+
+/// A kernel server connection.
+///
+/// The only difference for kernel client vs server connections is that client
+/// connections have support for recording new session tickets.
+pub type KernelServerConnection = KernelConnection<ServerConnectionData>;
```
djc, line 23: I'd prefer to get rid of these type aliases -- I think these generally don't make code clearer but instead obfuscate it in more indirection.

rustls/src/conn/kernel.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.state
+            .handle_new_session_ticket(&mut cx, &nst)
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub(crate) enum Direction {
+    Transmit,
+    Receive,
+}
```
djc, line 200: Nit: move this to the bottom of the module.

rustls/src/lib.rs
```diff
@@ -533,6 +533,72 @@ pub mod unbuffered {
     };
 }
 
+/// Kernel connection API.
```
djc, line 536: Nit: I think this docstring should be inside the `conn/kernel.rs` file. Could we just do `pub use conn::kernel;`?

rustls/src/server/tls12.rs
```diff
@@ -998,7 +999,30 @@ impl State<ServerConnectionData> for ExpectTraffic {
...
+            "TLS 1.2 connections do not support traffic secret updates".into(),
+        ))
+    }
+
+    fn handle_new_session_ticket(
+        &mut self,
+        _cx: &mut ExternalContext<'_>,
+        _message: &NewSessionTicketPayloadTls13,
+    ) -> Result<(), Error> {
+        Err(Error::InappropriateMessage {
```
djc, line 1023: Don't we have a better error variant for wrong protocol version? If not, maybe we should create one?

rustls/src/server/tls13.rs
```diff
@@ -1499,11 +1500,38 @@ impl State<ServerConnectionData> for ExpectTraffic {
...
         self
     }
 }
 
+impl ExternalState for ExpectTraffic {
+    fn update_secrets(&mut self, dir: Direction) -> Result<ConnectionTrafficSecrets, Error> {
+        let side = match dir {
+            Direction::Transmit => Side::Server,
+            Direction::Receive => Side::Client,
+        };
```
djc, line 1517: Nit: suggest inlining these matches everywhere.

rustls/src/tls13/key_schedule.rs
```diff
@@ -534,26 +534,31 @@ impl KeyScheduleTraffic {
...
+        &mut self,
+        side: Side,
+    ) -> Result<ConnectionTrafficSecrets, Error> {
+        let secret = self.next_application_traffic_secret(side);
+        let (key, iv) = expand_secret(
+            &secret,
+            self.ks.suite.hkdf_provider,
+            self.ks.suite.aead_alg.key_len(),
+        );
+        let secrets = self
```
djc, line 547: Nit: no need to bind `secrets`, just inline it.

rustls/src/tls13/key_schedule.rs
```diff
@@ -577,6 +582,15 @@ impl KeyScheduleTraffic {
     }
 }
 
+fn expand_secret(secret: &OkmBlock, hkdf: &'static dyn Hkdf, aead_key_len: usize) -> (AeadKey, Iv) {
```
djc, line 585: Nit: please move the extraction of this function into a separate commit.

rustls/tests/runners/kernel.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+#[cfg(feature = "aws_lc_rs")]
+#[path = "."]
+mod tests_with_aws_lc_rs {
+    provider_aws_lc_rs!();
+
+    #[path = "../kernel.rs"]
+    mod tests;
+}
```
djc, line 22: Can we do this without adding yet another small file?

examples/src/bin/ktls-client.rs
```diff
@@ -0,0 +1,666 @@
+//! This is an example showing how to use rustls' external connection API in
```
djc, line 1: I'm not sure we want to maintain all this stuff in this crate...

rustls/src/client/client_conn.rs
```diff
@@ -851,6 +851,8 @@ impl UnbufferedClientConnection {
 
     /// Extract secrets, so they can be used when configuring kTLS, for example.
     /// Should be used with care as it exposes secret key material.
+    #[deprecated = "dangerous_extract_secrets does not support session tickets or \
```
djc, line 854: Nit: suffix function/method names with `()` to make their usage more obvious.

djc:
I think this is generally the right direction. I'm on the fence about the example code -- while it's nice to have a worked example, it seems to spend a lot of code on the details of the kernel API, while the API surface we have to maintain in order to make it work appears to be substantially smaller.

rustls/src/server/tls12.rs
swlynch99: This gets handled higher up in https://github.com/swlynch99/rustls/blob/external-connection/rustls/src/conn/kernel.rs#L226. I'm happy to change the error here, but this is dead code so I just mirrored what rustls returns in this case for a normal connection.

rustls/src/server/tls13.rs
swlynch99: I'm not sure where you want to inline these? This match statement is different based on whether we are on the server or client side of the connection and we don't have that info in `KernelConnection`

rustls/tests/runners/kernel.rs
swlynch99: Sure. Where should I put the tests? I can stick them under `unbuffered`

examples/src/bin/ktls-client.rs
swlynch99: That's reasonable. This is really the example I wished I had when I was trying to figure out kTLS. I'm fine if it lives elsewhere as long as I can link it in the docs.

If you would rather not maintain it then let me know and I'll move it to its own repo.

rustls/src/conn/kernel.rs
```diff
@@ -0,0 +1,223 @@
...
+/// See the [`crate::kernel`] module docs for more details.
+pub struct KernelConnection<Data> {
+    state: Box<dyn ExternalState>,
+
+    peer_certificates: Option<CertificateChain<'static>>,
+    quic: Quic,
+
+    negotiated_version: ProtocolVersion,
+    protocol: Protocol,
+    suite: SupportedCipherSuite,
```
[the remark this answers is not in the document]
swlynch99: - `peer_certificates`, `quic`, and `protocol` are needed to handle session ticket updates. They are passed to the state via `KernelContext` (which I've just renamed from `ExternalContext`)
- `protocol_version` is used for some error checks
- `suite` is there purely to be exposed to the user. It is needed for kTLS and is also the only way to get at the `confidentiality_limit`, which I'd like to make sure is always available.

examples/src/bin/ktls-client.rs
ctz: I think we should use this example as inspiration when integrating this new API into the ktls crate, but probably not keeping it in this repo.

rustls/src/server/tls13.rs
djc: I meant like:

```rust
self.key_schedule.refresh_traffic_secret(match dir {
    ..
})
```

rustls/tests/runners/kernel.rs
djc: Yup, that sounds okay.

rustls/src/server/tls12.rs
ctz: I don't mind that; maybe we could put a `&'static str` in there so we can work out which one was incorrect, if someone who reports seeing such an error.

swlynch99: Here's what I have done instead:
- For `server::tls12::ExpectTraffic` I have added a useful error message that matches what `KernelConnection::handle_new_session_ticket` returns
- For `client::tls1[23]::ExpectTraffic` I have put `unreachable!` with a useful error message. I have also added a note that the method is only really supposed to be called for client connections.

All three are dead code, but the client ones should definitely never be called (and I've added a note to the `KernelState` documentation about this) whereas the server-side of the connection should just be straight-up unreachable.

examples/src/bin/ktls-client.rs
swlynch99: I have removed the example. Depending on how things go I may follow up later with a link to either a separate example repo or the ktls crate in a follow-up PR.

swlynch99:
Pinging on this. I think I have addressed and/or responded to all the comments, so I'm leaving this comment mostly to mark that I consider this PR to be waiting on a reviewer.

rustls/src/conn/kernel.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// # Errors
+    /// This method will return an error if:
+    /// - This connection is not a TLS 1.3 connection (in TLS 1.2 session tickets
+    ///   are sent as part of the handshake).
+    /// - This connection is a server connection.
```
ctz, line 212: nit: since this is defined on `KernelConnection<ClientConnectionData>` I don't think `KernelConnection<ServerConnectionData>` gets to return an error.

rustls/src/conn/kernel.rs
```diff
@@ -0,0 +1,273 @@
...
+        // here. 
+        if self.protocol_version() != ProtocolVersion::TLSv1_3 {
+            return Err(Error::General(
+                "TLS 1.2 session tickets may not be sent once the handshake has completed".into(),
+            ));
+        }
+
+        let mut reader = Reader::init(payload);
+        let nst = NewSessionTicketPayloadTls13::read(&mut reader)?;
+        reader.expect_empty("NewSessionTicket")?;
```
ctz, line 230:
```suggestion
        let nst = NewSessionTicketPayloadTls13::read_bytes(payload)?;
```

rustls/src/conn/kernel.rs
```diff
@@ -218,7 +218,7 @@ impl KernelConnection<ClientConnectionData> {
     pub fn handle_new_session_ticket(&mut self, payload: &[u8]) -> Result<(), Error> {
         // We want to return a more specific error here first if this is called
         // on a non-TLS 1.3 connection since a parsing error isn't the real issue
-        // here.
```
ctz: Please squash these whitespace fixes into previous commit. (ditto below)

rustls/src/conn/kernel.rs
swlynch99: Good catch! That was left over from when I made `KernelConnection` generic. I must have missed it when making changes.

swlynch99:
I think that's everything resolved. I have also rebased on `main`, so everything should be up to date.

ctz (approved):

rustls/src/tls13/key_schedule.rs
```diff
@@ -578,26 +578,30 @@ impl KeyScheduleTraffic {
...
-            aead_key_len: usize,
-        ) -> (AeadKey, Iv) {
-            let expander = hkdf.expander_for_okm(secret);
-
-            (
-                hkdf_expand_label_aead_key(expander.as_ref(), aead_key_len, b"key", &[]),
-                hkdf_expand_label(expander.as_ref(), b"iv", &[]),
-            )
-        }
+    pub(crate) fn refresh_traffic_secret(
```
djc, line 581: I think this rename means this commit won't compile, can you fix that?

djc (approved):
Thanks for all your work on this!

rustls/src/tls13/key_schedule.rs
swlynch99: This was not a rename (if you look at the diff you'll notice `extract_secrets` still exists below). However, `refresh_traffic_secrets` is part of the wrong commit here, so I've moved the diff adding it to [99b1da4](https://github.com/rustls/rustls/pull/2370/commits/99b1da43204855335856f358668fd2276e526aa4).

PR #2379, opened 2025-03-13
Refactor: prefer to take reference on match scrutinee

ctz:
This makes everything take advantage of match ergonomics.

(After this, there are no remaining `ref` uses in this repo.)


djc (approved):
Yay match ergonomics!

cpu (approved):
🥳

PR #2381, opened 2025-03-14
Add warning about server-side cross-config resumption

ctz:


rustls/src/server/server_conn.rs
```diff
@@ -242,6 +242,30 @@ impl<'a> ClientHello<'a> {
...
+/// # Sharing resumption storage between `ServerConfig`s
+/// In a program using many `ServerConfig`s it may improve resumption rates
+/// (which has a significant impact on connection performance) if those
+/// configs share [`ServerConfig::session_storage`] or [`ServerConfig::ticketer`].
+///
+/// However, caution is needed: other fields influence the security of a session
+/// and resumption between them can be surprising.  If sharing
+/// [`ServerConfig::session_storage`] or [`ServerConfig::ticketer`] between two
+/// `ServerConfig`s, you should also evaluate the following fields and also
+/// ensure they are equivalent:
```
cpu, line 254: editorial: the extra "also" reads awkwardly to me:

```suggestion
/// `ServerConfig`s, you should also evaluate the following fields and
/// ensure they are equivalent:
```

rustls/src/server/server_conn.rs
```diff
@@ -242,6 +242,30 @@ impl<'a> ClientHello<'a> {
...
+///
+/// * `ServerConfig::verifier` -- client authentication requirements,
```
cpu, line 256: should this field be in `[]`'s to match the `cert_resolver` instance below?

cpu (approved):

rustls/src/server/server_conn.rs
ctz: Unfortunately this is a private field, so asking to link it causes a warning.

rustls/src/server/server_conn.rs
```diff
@@ -242,6 +242,30 @@ impl<'a> ClientHello<'a> {
 /// * [`ServerConfig::cert_compression_cache`]: caches the most recently used 4 compressions
 /// * [`ServerConfig::cert_decompressors`]: depends on the crate features, see [`compress::default_cert_decompressors()`].
 ///
+/// # Sharing resumption storage between `ServerConfig`s
```
djc, line 245: Tiny nit: I think an empty line after the heading would make this easier to scan.

djc (approved):

PR #2382, opened 2025-03-15
Allow changing the limit on plaintext buffer size

belowm:
This trivial PR adds `Connection::set_plaintext_buffer_limit`, analog to the existing `Connection::set_buffer_limit` (which changes send buffer sizes).

**Usecase**: In my scenario, I cannot handle the decoded plaintext intermediately while feeding encrypted data into rustls. If a larger packet arrives, rustls may reject decoding it because there is a [hardcoded 16kb limit](https://github.com/rustls/rustls/blob/main/rustls/src/common_state.rs#L1041) on how many bytes the `ChunkVecBuffer` is allowed to hold. 

While I could implement additional buffering in my application, this introduces unnecessary overhead. Providing a method to adjust the plaintext buffer limit would eliminate the need for an extra layer of buffering.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -135,6 +135,16 @@ mod connection {
             }
         }
 
+        /// Sets a limit on the internal plaintext buffer.
+        ///
+        /// See [`ConnectionCommon::set_plaintext_buffer_limit()`] for more information.
+        pub fn set_plaintext_buffer_limit(&mut self, limit: Option<usize>) {
+            match self {
+                Self::Client(client) => client.set_plaintext_buffer_limit(limit),
+                Self::Server(server) => server.set_plaintext_buffer_limit(limit),
+            }
+        }        
+
         /// Sends a TLS1.3 `key_update` message to refresh a connection's keys
         ///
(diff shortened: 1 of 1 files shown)


rustls/src/conn.rs
```diff
@@ -517,6 +527,16 @@ impl<Data> ConnectionCommon<Data> {
         self.sendable_tls.set_limit(limit);
     }
 
+    /// Sets a limit on the internal buffers used to buffer decoded plaintext.
+    ///
+    /// See [`Self::set_buffer_limit`] for more information on how limits are applied.
+    pub fn set_plaintext_buffer_limit(&mut self, limit: Option<usize>) {
+        self.core
```
ctz, line 534: Could we have a test that observes that this behaves as you want? I would suggest one that firstly observes the "`received plaintext buffer full`" case in default configuration, and then calls this method and observes the error no longer happens. Let me know if you want help with writing that test.

ctz (commented):
I don't think I necessarily object to this as a new API.

cpu:
@belowm Are you interested in rebasing this branch and addressing [ctz's feedback](https://github.com/rustls/rustls/pull/2382#discussion_r2027424079) ?

PR #2383, opened 2025-03-15
Prepare 0.23.24

### LINKED ISSUE #2031: unbuffered: No way to refuse a ReadTraffic if there is no space

ctz:
ref rustls/rustls-platform-verifier#163

Proposed release notes:

* **New feature**: More detailed and helpful error reporting for common certificate errors, such as name mismatches and certificate expiry. Users who `std::fmt::Display` the rustls `Error` type will take advantage of this automatically. Users handling `CertificateError` variants individually should note the new variants, such as `CertificateError::NotValidForNameContext` (compare `CertificateError::NotValidForName`).

   ```shell
   $ cargo -q run --bin tlsclient-mio -- --http wrong.host.badssl.com
  TLS error: invalid peer certificate: certificate not valid for name "wrong.host.badssl.com";
   certificate is only valid for DnsName("*.badssl.com") or DnsName("badssl.com")
  Connection closed
  ```

  The old `CertificateError` variants (such as `NotValidForName`, `Expired`, etc.) remain usable, and may be produced by both the default and third-party certificate verification traits.
   
* **New feature**: Allow KTLS handoff for unbuffered API users, by introducing `dangerous_extract_secrets()`. Thanks to @edef1c.
* **Bug fix**: Unbuffered connections now consume data during the `next_record()` function, rather than production of the state. This fixes #2031.
* **Bug fix**: Build speed improvement for aws-lc-rs `fips` users.
* **Behavior change**: Clients no longer offer resumption between different `ClientConfig`s that share a resumption store but do not share server certificate verification and client authentication credentials. If you share a resumption store between multiple `ClientConfig`s, please ensure their server certificate verification and client authentication credentials are also shared. Please read the new documentation on the `ClientConfig::resumption` item for details.
  
  Additionally, if you share a resumption store or ticketer between multiple `ServerConfig`s, please see the new documentation on `ServerConfig` about this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -494,7 +494,7 @@ dependencies = [
  "base64",
  "env_logger",
  "nix",
- "rustls 0.23.23",
+ "rustls 0.23.24",
  "rustls-post-quantum",
 ]
 
@@ -1362,7 +1362,7 @@ dependencies = [
  "ipnet",
  "once_cell",
  "rand 0.9.0",
- "rustls 0.23.23 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustls 0.23.23",
  "thiserror 2.0.12",
  "tinyvec",
  "tokio",
@@ -1387,7 +1387,7 @@ dependencies = [
  "parking_lot",
  "rand 0.9.0",
  "resolv-conf",
- "rustls 0.23.23 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustls 0.23.23",
  "smallvec",
  "thiserror 2.0.12",
  "tokio",
@@ -2581,6 +2581,21 @@ dependencies = [
 [[package]]
 name = "rustls"
 version = "0.23.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47796c98c480fce5406ef69d1c76378375492c3b0a0de587be0c1d9feb12f395"
+dependencies = [
+ "log",
+ "once_cell",
+ "ring",
+ "rustls-pki-types",
+ "rustls-webpki 0.102.8",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
+name = "rustls"
+version = "0.23.24"
 dependencies = [
  "aws-lc-rs",
  "base64",
@@ -2610,26 +2625,11 @@ dependencies = [
 ]
 
 [[package]]
-name = "rustls"
-version = "0.23.23"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "47796c98c480fce5406ef69d1c76378375492c3b0a0de587be0c1d9feb12f395"
-dependencies = [
- "log",
- "once_cell",
- "ring",
- "rustls-pki-types",
- "rustls-webpki 0.102.8",
- "subtle",
- "zeroize",
-]
-
-[[package]]
 name = "rustls-bench"
 version = "0.1.0"
 dependencies = [
  "clap",
- "rustls 0.23.23",
+ "rustls 0.23.24",
  "rustls-post-quantum",
  "tikv-jemallocator",
 ]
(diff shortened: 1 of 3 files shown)


djc (approved):

cpu (approved):
LGTM. I went through the commits since .23 and didn't spot anything that felt relevant that wasn't covered.

> Users handling CertificateError variants individually should note the new variants, such as CertificateError::NotValidForNameContext (compare CertificateError::NotValidForName).

Is there more to say about the purpose of `CertificateError::NotValidForName` now that the `Context` variant exists? Is it retained for semver purposes only or can it still be emitted? It looks to me like the former but I only did a quick scan.


> Behavior change: Clients no longer offer resumption between different ClientConfigs that share a resumption store ....

WDYT about using this as an opportunity to make a small note about how this should also be considered for reusing server configs with some of the text from https://github.com/rustls/rustls/pull/2381 ?

ctz:
> Is there more to say about the purpose of `CertificateError::NotValidForName` now that the `Context` variant exists? Is it retained for semver purposes only or can it still be emitted? It looks to me like the former but I only did a quick scan.

Added a note -- I think the answer is that all the previous variants remain both for semver reasons but also for non-webpki verifiers (eg, platform-verifier) which I expect will continue to use them.

> WDYT about using this as an opportunity to make a small note about how this should also be considered for reusing server configs with some of the text from #2381 ?

Added a reference for this.

PR #2387, opened 2025-03-20
Upgrade to hickory-resolver 0.25

djc:


ctz (approved):

cpu (approved):
Nice! 

I will kick the daily tasks job when I'm at a real keyboard to double check the ech example still does it's thing

cpu:
> I will kick the daily tasks job when I'm at a real keyboard

All good (minus an unrelated powerset bug, will open an issue for that): https://github.com/rustls/rustls/actions/runs/13969447498/job/39107384691

PR #2408, opened 2025-04-01
ci: pin cmake to 3.28 on linux

cpu:
cmake 4.0 is not compatible w/ aws-lc until the forthcoming release. The runner image switched the version out from under us. Pin it back until one side or the other sorts itself out.

The runner-image-provided `cmake@4.0` is in `/usr/local/cmake` and has higher `$PATH` priority. Rather than munge with the `PATH` and change other things we symbolic link the installed `cmake@3.28` over-top.

Fixes the [build failures on `main`](https://github.com/rustls/rustls/actions/runs/14202742201).

See https://github.com/aws/aws-lc-rs/issues/755, https://github.com/actions/runner-images/issues/11926

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,6 +56,12 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
 
+      - name: Install Cmake 3.x on Linux
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get install cmake=3.28.3-1build7
+          sudo ln -sf /usr/bin/cmake /usr/local/bin/cmake
+
       - name: Install NASM for aws-lc-rs on Windows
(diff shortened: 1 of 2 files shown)


cpu:
> rustls / Build+test (stable, windows-latest) (pull_request)Failing after 5m

argh, I was hoping it was just the Linux runners that had the issue. I'm not sure I know how to easily remedy this on Windows :thinking:

benj5378:
@cpu see fix for Windows and Ubuntu here 
- https://github.com/FreeCAD/FreeCAD/pull/20581

cpu:
Closing this - I think an `aws-lc-fips-sys` 0.13.5 w/ CMake 4.0 support was released. Let's take that update instead of reverting CMake.

PR #2412, opened 2025-04-03
Add cargo deny check in CI

djc:


deny.toml
```diff
@@ -0,0 +1,10 @@
+[licenses]
+version = 2
+allow = ["Apache-2.0", "BSD-3-Clause", "ISC", "MIT", "MPL-2.0", "OpenSSL", "Unicode-3.0"]
```
cpu, line 3: I think we [need one more here](https://github.com/rustls/rustls/actions/runs/14237900888/job/39900863213?pr=2412) for Zlib:

```
error[rejected]: failed to satisfy license requirements
  ┌─ registry+https://github.com/rust-lang/crates.io-index#zlib-rs@0.5.0:4:12
  │
4 │ license = "Zlib"
  │            ━━━━
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: license is not explicitly allowed
  ```

File: deny.toml [hunk 2]
```diff
@@ -0,0 +1,10 @@
+[licenses]
```
cpu, line 1: The ["taplo" CI check for YAML formatting](https://github.com/rustls/rustls/actions/runs/14237900888/job/39900863210?pr=2412) is upset (but doesn't say much about why... :thinking: ):

> ERROR taplo:format_files: the file is not properly formatted path="/home/runner/work/rustls/rustls/deny.toml"

cpu (approved):
LGTM mod the two CI fails. Thanks!

deny.toml
djc: It wanted me to indent by two (instead of four spaces) -- fixed.

ctz (approved):

PR #2420, opened 2025-04-08
Invert working of `SignatureScheme::supported_in_tls13`

### LINKED ISSUE #2419: TLS 1.3 SignatureSchemes are not extendable on the server
# Context
I'm writing a custom `CryptoProvider` that makes new PQC algorithms from `liboqs` available to Rustls. This is for research, I'm aware that `liboqs` isn't production-ready crypto.
So far I have implemented the key exchange to use `ML-KEM` and signatures to work with `ML-DSA`. Right now, I'm trying to get a simple client-server integration test to run. The self-signed signatures have been generated by OpenSSL with `oqsprovider` and the keys get loaded correctly.

# The Error
You can think of my test to look something like this:
* Initialize a client and server config with the custom provider
* Start the server on a separate thread and bind to local IP
* After a short delay initiate a TCP connection to the server
* Both the server and the client try to write some predetermined data

The server thread panics with a `PeerIncompatible(NoSignatureSchemesInCommon)` error.

# Things I've debugged
* My `SigningKey::choose_scheme()` implementation gets an empty slice in the `offered` parameter. So the problem happens earlier in the process
* Logging with `trace` gave me the following insights:
  *  The last trace I get on the server is "sending certificate HandshakeMessagePayload {..."
  * The client advertises the expected SignatureScheme enums of ML-DSA I've defined and the server accepts them

# The problem (as far as I understand it)
Looking at the source of Rustls I've tracked down the origin of the `NoSignatureSchemesInCommon` error and traced back how the SignatureSchemes get chosen.
The line where I suspect the crux lies: https://github.com/rustls/rustls/tree/main/rustls/src/server/tls13.rs#L150C13-L150C27
This line filters all `SignatureScheme` variants to only the ones defined in `SignatureScheme::supported_in_tls13()` and since the client is only advertising self-defined `SignatureScheme::Unknown` variants this always results in the vector becoming empty.

If there are any extension points in Rustls I have missed that might fix this problem or if my assessment seems incorrect, please let me know.

ctz:
This was previously an allow-list, which stands in the way of people shipping providers for other algorithms that are not specifically outlawed in RFC8446.

fixes #2419

### DIFF
--- a/rustls/src/enums.rs
+++ b/rustls/src/enums.rs
@@ -1,6 +1,7 @@
 #![allow(non_camel_case_types)]
 #![allow(missing_docs)]
 use crate::msgs::codec::{Codec, Reader};
+use crate::msgs::enums::HashAlgorithm;
 
 enum_builder! {
     /// The `AlertDescription` TLS protocol enum.  Values in this enum are taken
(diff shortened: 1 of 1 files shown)


rustls/src/enums.rs
```diff
@@ -544,15 +545,40 @@ impl SignatureScheme {
...
-                | Self::RSA_PSS_SHA384
-                | Self::RSA_PSS_SHA256
-                | Self::ED25519
+        let [hash, sign] = self.to_array();
+
+        // This covers both disallowing SHA1:
+        //
+        //  "These values refer solely to signatures which appear in certificates
+        //   (see Section 4.4.2.2) and are not defined for use in signed TLS
+        //   handshake messages"
```
djc, line 554: This quote doesn't seem to have anything to do with SHA1?

rustls/src/enums.rs
```diff
@@ -544,15 +545,40 @@ impl SignatureScheme {
...
+            | HashAlgorithm::SHA1
+            | HashAlgorithm::SHA224 => return false,
+            _ => (),
+        };
+
+        // RSA-PKCS1 is also disallowed for TLS1.3:
+        //
+        //  "These values refer solely to signatures which appear in certificates
+        //   (see Section 4.4.2.2) and are not defined for use in signed TLS
+        //   handshake messages"
```
djc, line 573: Does this repeat the same quote? Not sure if that's useful.

djc (approved):
Nice!

rustls/src/enums.rs
ctz: Ah, I've cut too much. The full section is:

```
Legacy algorithms:  Indicates algorithms which are being deprecated
      because they use algorithms with known weaknesses, specifically
      SHA-1 which is used in this context with either (1) RSA using
      RSASSA-PKCS1-v1_5 or (2) ECDSA.  These values refer solely to
      signatures which appear in certificates (see Section 4.4.2.2) and
      are not defined for use in signed TLS handshake messages, although
      they MAY appear in "signature_algorithms" and
      "signature_algorithms_cert" for backward compatibility with
      TLS 1.2.
```

rustls/src/enums.rs
ctz: And this one uncut is:

```
RSASSA-PKCS1-v1_5 algorithms:  Indicates a signature algorithm using
      RSASSA-PKCS1-v1_5 [RFC8017] with the corresponding hash algorithm
      as defined in [SHS].  These values refer solely to signatures
      which appear in certificates (see Section 4.4.2.2) and are not
      defined for use in signed TLS handshake messages, although they
      MAY appear in "signature_algorithms" and
      "signature_algorithms_cert" for backward compatibility with
      TLS 1.2.
```

ctz: Going to replace these quotes with references to the paragraphs.

djc:
Maybe there should also be a line or 2 about why not "failing closed" is the better solution here? (Even if that is just "because it aligns more closely with the spec".)

ctz:
> Maybe there should also be a line or 2 about why not "failing closed" is the better solution here? (Even if that is just "because it aligns more closely with the spec".)

Added a bit to the docstring for this function.

PR #2421, opened 2025-04-08
Reduce use of library internals in tests

ctz:
This PR reduces the reliance of our tests on internal types for the future benefit of #1475.

Instead of that, `tests::common::encoding` is some very stupid code that can produce a variety of client hellos. Making that separate also means we're not testing rustls against itself, and the test code is not fighting against the type-safe internals code when constructing invalid input (see, eg, `server_rejects_sni_with_illegal_dns_name`).

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -15,11 +15,8 @@ use rustls::client::{ResolvesClientCert, Resumption, verify_server_cert_signed_b
 use rustls::crypto::{ActiveKeyExchange, CryptoProvider, SharedSecret, SupportedKxGroup};
 use rustls::internal::msgs::base::Payload;
 use rustls::internal::msgs::codec::Codec;
-use rustls::internal::msgs::enums::{AlertLevel, CertificateType, Compression, ExtensionType};
-use rustls::internal::msgs::handshake::{
-    ClientExtension, ClientHelloPayload, HandshakeMessagePayload, HandshakePayload, Random,
-    ServerExtension, SessionId,
-};
+use rustls::internal::msgs::enums::{AlertLevel, CertificateType, ExtensionType};
+use rustls::internal::msgs::handshake::{ClientExtension, HandshakePayload, ServerExtension};
 use rustls::internal::msgs::message::{Message, MessagePayload, PlainMessage};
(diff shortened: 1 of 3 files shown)


rustls/tests/common/mod.rs
```diff
@@ -1477,3 +1477,182 @@ mod plaintext {
         }
     }
 }
+
+/// Deeply inefficient, test-only TLS encoding helpers
+pub mod encoding {
```
djc, line 1482: Nit: top-down ordering?

rustls/tests/api.rs
```diff
@@ -6706,7 +6562,14 @@ fn test_acceptor() {
...
     // Minimal valid 1-byte application data message is not a handshake message
     acceptor
-        .read_tls(&mut [0x17, 0x03, 0x03, 0x00, 0x01, 0x00].as_ref())
+        .read_tls(
+            &mut encoding::message_framing(
+                ContentType::ApplicationData,
+                ProtocolVersion::TLSv1_2,
+                vec![0x00],
+            )
+            .as_slice(),
```
djc, line 6571: Maybe these smaller changes could make for a separate commit?

djc (commented):
Seems nice! This commit has a lot going on, do you think it can profitably split up in some smaller stages?

rustls/tests/common/mod.rs
```diff
@@ -1477,3 +1477,181 @@ mod plaintext {
...
+            let mut share = len_u16(SOME_POINT_ON_P256.to_vec());
+            share.splice(0..0, NamedGroup::secp256r1.to_array());
+
+            Extension {
+                typ: ExtensionType::KeyShare,
+                body: len_u16(share),
+            }
+        }
+    }
+    /// Prefix with u8 length
```
cpu, line 1627: nit: whitespace separator

cpu (approved):
Very nice :clap:

djc (approved):
Nice! I think the commit history would be nicer if it flowed the other way:

- `message_framing()` (last commit with bits it needs from the new module)
- `handshake_framing()` (half of the second commit)

This would split the introduction of the `encoding` module starting with the lower-level bits and working up to the higher-level functions.

PR #2424, opened 2025-04-10
add TLS 1.3 PSK support

elagergren-spideroak:
This PR adds TLS 1.3 PSK support.


djc:
Can you fix up CI? Let's drop the separate examples for now.

elagergren-spideroak:
> Can you fix up CI?

Sure thing. I fixed bogo locally, just haven't pushed it yet. For semver should I bump the version in the cargo manifest?

> Let's drop the separate examples for now.

👍

elagergren-spideroak:
@djc okay, I've verified that bogo passes locally.

djc:
I don't think we're in a state where we want to merge semver-incompatible changes. As such:

- We can't add a `HandshakeKind` variant right now
- We can't remove a `PeerMisbehaved` variant right now (but we can `#[deprecate]` it and stop emitting it)

rustls/src/client/hs.rs
```diff
@@ -459,6 +524,7 @@ fn emit_client_hello_for_retry(
                 config.provider.secure_random,
                 input.server_name.clone(),
                 &chp_payload,
+                Arc::clone(&config.provider),
```
djc, line 527: Nit: pass a reference here.

rustls/src/client/hs.rs
```diff
@@ -503,24 +569,28 @@ fn emit_client_hello_for_retry(
...
-        (_, Some(tls13_session)) => Some((
-            tls13_session.suite(),
-            tls13::fill_in_psk_binder(&tls13_session, &transcript_buffer, &mut chp),
-        )),
-
-        // No early key schedule in other cases.
-        _ => None,
-    };
+    // Derive the TLS 1.3 key schedule for sending early data.
+    let tls13_early_data_key_schedule = tls13_psk
```
djc, line 573: I don't think the refactoring here (from a two-element `match` to an `and_then()` with inner `match` is an improvement. Suggest to separate refactoring from other changes into a separate commit.

rustls/src/client/hs.rs
```diff
@@ -602,62 +679,121 @@ fn emit_client_hello_for_retry(
...
+    }
 
     // If the selected cipher suite can't select from the session's, we can't resume.
-    if let Some(suite) = suite {
-        suite.can_resume_from(tls13.suite())?;
+    if suite.is_some_and(|suite| {
+        suite
+            .can_resume_from(resuming.suite())
+            .is_none()
+    }) {
```
djc, line 763: Nit: this style with a multi-line condition header is pretty hard to read.

rustls/src/msgs/persist.rs
```diff
@@ -188,10 +186,14 @@ impl Tls12ClientSessionValue {
...
     }
 
-    pub(crate) fn ticket(&mut self) -> Arc<PayloadU16> {
+    pub(crate) fn ticket(&self) -> Arc<PayloadU16> {
         Arc::clone(&self.common.ticket)
     }
 
+    pub(crate) fn ticket_bytes(&self) -> &[u8] {
+        self.common.ticket.0.as_slice()
+    }
```
djc, line 195: If the return value should be treated specially if it `is_empty()`, suggest yielding an `Option` here.

rustls/src/client/tls13.rs
```diff
@@ -63,102 +66,261 @@ static DISALLOWED_TLS13_EXTS: &[ExtensionType] = &[
...
-    our_key_share: Box<dyn ActiveKeyExchange>,
+    our_key_share: Option<Box<dyn ActiveKeyExchange>>,
     mut sent_tls13_fake_ccs: bool,
     server_hello_msg: &Message<'_>,
     ech_state: Option<EchState>,
 ) -> hs::NextStateOrError<'static> {
+    // Early data requires a PSK.
+    //
+    // TODO(eric): Express this with types, like
+    // `../server/tls13::InputSecrets`.
```
djc, line 91: When do you plan to address these TODOs?

rustls/src/client/tls13.rs
```diff
@@ -63,102 +66,261 @@ static DISALLOWED_TLS13_EXTS: &[ExtensionType] = &[
...
+    // Any future values that are allocated must ensure that the
+    // transmitted protocol messages unambiguously identify which
+    // mode was selected by the server; at present, this is
+    // indicated by the presence of the "key_share" in the
+    // ServerHello."
+    let mode = if their_key_share.is_some() {
+        PSKKeyExchangeMode::PSK_DHE_KE
+    } else {
+        PSKKeyExchangeMode::PSK_KE
+    };
```
djc, line 137: Nit, write this more directly as a `match`.

rustls/src/client/tls13.rs
```diff
@@ -302,77 +505,312 @@ pub(super) fn initial_key_share(
     group.start()
 }
 
-/// This implements the horrifying TLS1.3 hack where PSK binders have a
-/// data dependency on the message they are contained within.
-pub(super) fn fill_in_psk_binder(
```
djc, line 307: Split the refactoring from the logic additions (a one-variant enum would be fine for this).

rustls/src/crypto/mod.rs
```diff
@@ -653,6 +655,200 @@ impl From<Vec<u8>> for SharedSecret {
...
+/// A TLS 1.3 preshared key.
+#[derive(Clone)]
+pub struct PresharedKey {
+    identity: Zeroizing<Box<[u8]>>,
+    secret: Zeroizing<Box<[u8]>>,
+    hash_alg: Tls13HashAlg,
+    /// Whether early data is allowed.
+    ///
+    /// This is (max_early_data, _, ALPN)
+    early_data: Option<(u32, CipherSuite, Option<Vec<u8>>)>,
```
djc, line 667: Nit: let's use a struct for the tuple here.

rustls/src/enums.rs
```diff
@@ -495,6 +496,24 @@ enum_builder! {
...
+
+        let alg = match self {
+            TLS13_AES_128_GCM_SHA256
+            | TLS13_CHACHA20_POLY1305_SHA256
+            | TLS13_AES_128_CCM_SHA256
+            | TLS13_AES_128_CCM_8_SHA256 => HashAlgorithm::SHA256,
+            TLS13_AES_256_GCM_SHA384 => HashAlgorithm::SHA384,
+            _ => return None,
+        };
+        Some(alg)
```
djc, line 513: Drop the `alg` binding and yield the value of the `match` directly.

rustls/src/server/hs.rs
```diff
@@ -302,7 +310,13 @@ pub(super) struct ExpectClientHello {
     pub(super) session_id: SessionId,
     #[cfg(feature = "tls12")]
     pub(super) using_ems: bool,
+    /// Have we sent a HelloRetryRequest?
     pub(super) done_retry: bool,
+    /// Did we send a "key_share" extension in our
+    /// HelloRetryRequest?
+    ///
+    /// Only has meaning if `self.done_retry` is true.
```
djc, line 318: This suggests these should be modeled with an enum (in a separate commit).

rustls/src/server/server_conn.rs
```diff
@@ -304,6 +314,15 @@ pub struct ServerConfig {
...
     pub ticketer: Arc<dyn ProducesTickets>,
 
+    /// Retrieves external preshared keys.
+    pub preshared_keys: Arc<dyn SelectsPresharedKeys>,
+
+    /// Only allow preshared key handshakes.
+    ///
+    /// The server will abort the connection if the client does
+    /// not provide a valid preshared key.
+    pub only_allow_preshared_keys: bool,
```
djc, line 324:
```rust
enum PresharedKeySelection {
     Enabled(Arc<dyn SelectsPresharedKeys>),
     Required(Arc<dyn SelectsPresharedKeys>),
     Disabled,
}
```

rustls/src/server/server_conn.rs
```diff
@@ -110,6 +110,16 @@ pub trait ProducesTickets: Debug + Send + Sync {
...
 }
 
+/// Selects TLS 1.3 external preshared keys.
+pub trait SelectsPresharedKeys: Debug + Send + Sync {
+    /// Retrieves a preshared key.
+    ///
+    /// To help prevent adversaries from discovering identities
+    /// supported by the server, identity comparisons should be
+    /// performed in constant time.
+    fn load_psk(&self, identity: &[u8]) -> Option<Arc<PresharedKey>>;
```
djc, line 120: If we need constant-time here, suggest we change the API so that rustls itself handles the identity comparison.

djc (commented):
Okay, that is a lot of stuff...

I think this needs a bunch of work to clean it up. There should be a clean commit history (potentially as separate PRs so we can merge some of it sooner). I also think there should be a bunch of refactoring here so that the PSK/resumption code is more isolated from the rest of the handshake.

rustls/src/client/tls13.rs
elagergren-spideroak: I don't think they're critical for this feature,  so either I'll delete the comments or I can make those changes on a subsequent PR.

rustls/src/client/client_conn.rs
```diff
@@ -525,6 +561,33 @@ pub enum Tls12Resumption {
...
+    /// # Warning
+    ///
+    /// This mode lacks forward security. Unless you know that
+    /// you need PSK-only key establishment, you should use
+    /// [`PskWithDhe`][Self::PskWithDhe].
+    ///
+    /// See [RFC 9257] for more guidance.
+    ///
+    /// [RFC 9257]: https://www.rfc-editor.org/rfc/rfc9257.html
+    PskOnly,
```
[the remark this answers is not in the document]
elagergren-spideroak: > These variant names aren't great. Let's avoid repeating the `Psk` prefix.

`PskKexMode::Only` is a little weird.

> Are "exchange modes" how these are described in the specifications?

RFC 8446 refers to them as "Pre-Shared Key Exchange Modes" and the extension is called "psk_key_exchange_modes".

elagergren-spideroak:
> I don't think we're in a state where we want to merge semver-incompatible changes. As such:
> 
> * We can't add a `HandshakeKind` variant right now

This is probably fine since rustls doesn't make any important decisions with `HandshakeKind`, but if we don't add a new variant then it's unclear which variant we should return when external PSKs are used. `HandshakeKind::Resumed` is the closest, but external PSKs are different from session resumption. Specifically, you can use an external PSK without ever having previously performed a handshake.

elagergren-spideroak:
> Okay, that is a lot of stuff...
> 
> I think this needs a bunch of work to clean it up. There should be a clean commit history (potentially as separate PRs so we can merge some of it sooner). I also think there should be a bunch of refactoring here so that the PSK/resumption code is more isolated from the rest of the handshake.

Let me know if you'd like me to roll up the commit history. I typically wait to squash commits before committing so it doesn't disrupt code review.

Let me know which parts you'd like moved to separate PRs, if any. I tried to make this PR as minimal as possible, but some refactoring (particularly of resumption handling) was necessary, otherwise the code would've been terribly difficult to follow, which is unacceptable for security critical code.

djc:
> > I don't think we're in a state where we want to merge semver-incompatible changes. As such:
> > 
> > * We can't add a `HandshakeKind` variant right now
> 
> This is probably fine since rustls doesn't make any important decisions with `HandshakeKind`, but if we don't add a new variant then it's unclear which variant we should return when external PSKs are used. `HandshakeKind::Resumed` is the closest, but external PSKs are different from session resumption. Specifically, you can use an external PSK without ever having previously performed a handshake.

I think using `Resumed` is probably fine, if you add documentation on the variant that it can be used this way -- and file an issue to make `HandshakeKind` `#[non_exhaustive]` in the future.

> Let me know if you'd like me to roll up the commit history. I typically wait to squash commits before committing so it doesn't disrupt code review.
>
> Let me know which parts you'd like moved to separate PRs, if any. I tried to make this PR as minimal as possible, but some refactoring (particularly of resumption handling) was necessary, otherwise the code would've been terribly difficult to follow, which is unacceptable for security critical code.

We typically work with smallish commits that each compile and can pass CI (though we don't usually test whether exactly each commit does). I suggest you break this up into smaller pieces, especially by (1) splitting refactoring into separate commits, (2) splitting client from server. Also try to minimize unrelated changes (like renaming `secret` to `psk`). We don't usually incrementally review, so keeping all WIP commits around just makes things harder.

This is a pretty large amount of code to review. Maybe your organization could consider sponsoring the integration effort?

mooreoak:
> This is a pretty large amount of code to review. Maybe your organization could consider sponsoring the integration effort?

What would this cost/involve?

djc:
> > This is a pretty large amount of code to review. Maybe your organization could consider sponsoring the integration effort?
> 
> What would this cost/involve?

Send me an email? You can trivially find my address from my homepage. Feel free to include a proposal for what you would consider reasonable.

ctz:
Going to close this for now, pending future work on PSK support.

PR #2425, opened 2025-04-11
Add `prefer-post-quantum` to default features

ctz:
This enables, for dependents who use default-features, X25519MLKEM768 as the most-preferred key exchange.

This is another step on the way to enabling this by default. In a future release, I plan to deprecate this feature, make it do nothing, and therefore unconditionally have `X25519MLKEM768` as the most preferred key exchange in the default aws-lc-rs provider.

Users who wish to opt-out from that change would then have to construct a customized `CryptoProvider` in their code.

### DIFF
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -15,7 +15,7 @@ exclude = ["src/testdata", "tests/**"]
 build = "build.rs"
 
 [features]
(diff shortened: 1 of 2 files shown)


djc (approved):

cpu (approved):
🚀

PR #2426, opened 2025-04-13
Improve invalid key purpose errors

djc:
Depends on:

- https://github.com/rustls/webpki/pull/337

### DIFF
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -2,9 +2,11 @@ use alloc::format;
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::fmt;
-use pki_types::{ServerName, UnixTime};
 #[cfg(feature = "std")]
 use std::time::SystemTimeError;
+
+use pki_types::{ServerName, UnixTime};
+use webpki::KeyUsage;
 
 use crate::enums::{AlertDescription, ContentType, HandshakeType};
 use crate::msgs::handshake::{EchConfigPayload, KeyExchangeAlgorithm};
@@ -434,6 +436,17 @@ pub enum CertificateError {
     /// The certificate is being used for a different purpose than allowed.
     InvalidPurpose,
 
+    /// The certificate is being used for a different purpose than allowed.
+    ///
+    /// This variant is semantically the same as `InvalidPurpose`, but includes
+    /// extra data to improve error reports.
+    InvalidPurposeContext {
+        /// Extended key purpose that was required by the application.
+        required: ExtendedKeyPurpose,
+        /// Extended key purposes that were presented in the peer's certificate.
+        presented: Vec<ExtendedKeyPurpose>,
+    },
+
     /// The certificate is valid, but the handshake is rejected for other
     /// reasons.
     ApplicationVerificationFailure,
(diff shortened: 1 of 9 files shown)


rustls/src/error.rs
```diff
@@ -5,6 +5,7 @@ use core::fmt;
 use pki_types::{ServerName, UnixTime};
 #[cfg(feature = "std")]
 use std::time::SystemTimeError;
+use webpki::KeyUsage;
```
cpu, line 8: I think this file has a small pre-existing import order style issue. Want to fix it up now?

Both the `pki_types` and `webpki` imports should be separate from the `alloc/core/std` imports above.

cpu (commented):
Nice :-)

Will hold approving until the cargo patch is dropped, but the implementation LGTM.

rustls/src/error.rs
djc: Yup, fixed!

djc:
> Nice :-)
> 
> Will hold approving until the cargo patch is dropped, but the implementation LGTM.

I think having rustls-webpki 0.103.2 be out in the wild before rustls 0.23.27 is released would lead to suboptimal error handling, so I kinda want to get my approvals for this PR before I publish rustls-webpki 0.103.2.

rustls/src/error.rs
```diff
@@ -426,6 +428,17 @@ pub enum CertificateError {
     /// The certificate is being used for a different purpose than allowed.
     InvalidPurpose,
 
+    /// The certificate is being used for a different purpose than allowed.
+    ///
+    /// This variant is semantically the same as `InvalidPurpose`, but includes
+    /// extra data to improve error reports.
+    InvalidPurposeContext {
```
ctz, line 435: Do we want a long-form `Display` for this variant?

rustls/src/webpki/mod.rs
```diff
@@ -91,7 +93,18 @@ fn pki_error(error: webpki::Error) -> Error {
             CertRevocationListError::BadSignature.into()
         }
 
+        #[allow(deprecated)]
         RequiredEkuNotFound => CertificateError::InvalidPurpose.into(),
+        RequiredEkuNotFoundContext(webpki::RequiredEkuNotFoundContext { required, present }) => {
```
ctz, line 98: Not an issue being introduced in this PR, but I note that we don't have coverage for either `RequiredEkuNotFound` or `RequiredEkuNotFoundContext`

ctz (approved):

rustls/src/webpki/mod.rs
djc: Any suggestion where/how is the best place to add coverage?

ctz: I think I would start a server using client credentials (ie, `KeyType::get_client_chain()` rather than `KeyType::get_chain()`, ditto for the key), then make a client and call `do_handshake_until_error`.

cpu (approved):

rustls/src/webpki/mod.rs
djc: Added a test, which seems to fail despite looking equal...

```
  left: InvalidCertificate(InvalidPurposeContext { required: ServerAuth, presented: [ClientAuth] })
 right: InvalidCertificate(InvalidPurposeContext { required: ServerAuth, presented: [ClientAuth] })
```

djc: Fixed.

djc:
* Published rustls v0.23.27 at registry `crates-io`
* [new tag]           v/0.23.27 -> v/0.23.27
* [Release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.23.27)

PR #2428, opened 2025-04-14
Take latest bogo test suite

ctz:


bogo/src/main.rs
```diff
@@ -1757,7 +1769,12 @@ pub fn main() {
         exec(&opts, sess, i);
         if opts.resume_with_tickets_disabled {
             opts.tickets = false;
-            server_cfg = Some(make_server_cfg(&opts));
+
+            if opts.side == Side::Server {
```
djc, line 1773: Nit: I think `match` would be nicer here than `if`.

djc (approved):
Nice!

cpu (approved):
Thanks :-)

PR #2430, opened 2025-04-15
Tighten up decoding of empty messages

ctz:
This PR makes us more correct when parsing messages with defined lower length bounds, like:

```
      opaque DistinguishedName<1..2^16-1>;
```

This notation says `DistinguishedName` must be a non-empty string. Our decoding now enforces this.

There's some risk that this introduces new interop failures, but I think it is quite low.

There are quite a lot of commits, but they are all quite small.


rustls/src/msgs/handshake.rs
```diff
@@ -220,8 +220,11 @@ impl TlsListElement for NamedGroup {
     };
 }
 
+/// RFC8446: `SignatureScheme supported_signature_algorithms<2..2^16-2>;`
 impl TlsListElement for SignatureScheme {
-    const SIZE_LEN: ListLength = ListLength::U16;
+    const SIZE_LEN: ListLength = ListLength::NonZeroU16 {
```
djc, line 227: Should we generalize to `ListLength::U16Min` so we can enforce the `2..`?

rustls/src/msgs/handshake.rs
```diff
@@ -550,8 +550,11 @@ impl CertificateStatusRequest {
 
 // ---
 
+/// RFC8446: `PskKeyExchangeMode ke_modes<1..255>;`
 impl TlsListElement for PSKKeyExchangeMode {
-    const SIZE_LEN: ListLength = ListLength::U8;
+    const SIZE_LEN: ListLength = ListLength::NonZeroU8 {
+        empty_error: InvalidMessage::IllegalEmptyList("PskKeyExchangeModes"),
```
djc, line 566: I prefer the spelling you picked for the error string, but should we keep it in sync with the code?

rustls/src/msgs/base.rs
```diff
@@ -192,6 +192,43 @@ impl fmt::Debug for PayloadU8 {
     }
 }
 
+/// An arbitrary, unknown-content, non-empty, u8-length-prefixed payload
+#[derive(Clone, Eq, PartialEq)]
+pub(crate) struct PayloadNonZeroU8(pub(crate) Vec<u8>);
```
djc, line 231: Instead of duplicating, have you considered having a `const` generic argument for the minimum?

djc (commented):
Nice!

rustls/src/msgs/handshake.rs
ctz: The `2..` is enforced, but via a different route: if the length is non-zero, then we read the items (each 2 bytes in length) until the end is reached. If the length is 1 (or any odd value) then the last item read will fail, with `InvalidMessage::MissingData`.

rustls/src/msgs/handshake.rs
ctz: Added a renaming commit for this type.

rustls/src/msgs/base.rs
ctz: Have tried this for the PayloadU16 type, will extend it to the U8 type if you think it's an improvement? I'm on the fence.

djc: Why are you on the fence?

ctz: Hiding the mysterious type parameter via a type alias, which breaks the existing tuple struct uses.

I also tried this kind of thing:

```rust
pub trait Cardinality: Clone + Eq + PartialEq {
    const MIN: usize;
}

#[derive(Clone, Eq, PartialEq)]
pub struct MaybeEmpty;
impl Cardinality for MaybeEmpty {
    const MIN: usize = 0;
}

#[derive(Clone, Eq, PartialEq)]
pub struct NonEmpty;
impl Cardinality for NonEmpty {
    const MIN: usize = 1;
}
```

and then have `PayloadU16<C: Cardinality = MaybeEmpty>` for existing code, and `PayloadU16<NonEmpty>` for the changes in this PR. That also breaks the existing tuple struct uses (because the need for a `PhantomData<C>`) but I suppose that is not a huge cost.

djc: I'd just get rid of the type aliases, which I don't like anyway (except perhaps as a transition mechanism). In general avoiding the duplication in favor of abstraction seems like a good thing?

rustls/src/msgs/base.rs
```diff
@@ -148,50 +149,75 @@ impl fmt::Debug for PayloadU16 {
...
-impl PayloadU8 {
+impl<C: Cardinality> PayloadU8<C> {
     pub(crate) fn encode_slice(slice: &[u8], bytes: &mut Vec<u8>) {
         (slice.len() as u8).encode(bytes);
         bytes.extend_from_slice(slice);
     }
 
     pub(crate) fn new(bytes: Vec<u8>) -> Self {
-        Self(bytes)
+        debug_assert!(C::MIN <= bytes.len());
```
djc, line 170: Tiny nit: feels like more conventional to write `bytes.len() >= C::MIN`?

djc (approved):

rustls/src/msgs/base.rs
```diff
@@ -148,50 +149,75 @@ impl fmt::Debug for PayloadU16 {
...
     }
 }
 
+pub trait Cardinality: Clone + Eq + PartialEq {
+    const MIN: usize;
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct MaybeEmpty;
+impl Cardinality for MaybeEmpty {
```
cpu, line 217: nit: I think we'd usually want whitespace between the `struct` and `impl` here and on L216 no?

cpu (approved):
rad. This is the kind of diff that makes me <3 Rust

PR #2437, opened 2025-04-22
[SECURITY(deps)] update dependency ring to >=0.17.12 to remove panic vulnerability GHSA-4p46-pwfr-66x6

Shifta-Robel:
GHSA-ID GHSA-4p46-pwfr-66x6
Vulnerability description from Github

> 
"ring::aead::quic::HeaderProtectionKey::new_mask() may panic when overflow checking is enabled. In the QUIC protocol, an attacker can induce this panic by sending a specially-crafted packet. Even unintentionally it is likely to occur in 1 out of every 2**32 packets sent and/or received.

On 64-bit targets operations using ring::aead::{AES_128_GCM, AES_256_GCM} may panic when overflow checking is enabled, when encrypting/decrypting approximately 68,719,476,700 bytes (about 64 gigabytes) of data in a single chunk. Protocols like TLS and SSH are not affected by this because those protocols break large amounts of data into small chunks. Similarly, most applications will not attempt to encrypt/decrypt 64GB of data in one chunk.

Overflow checking is not enabled in release mode by default, but RUSTFLAGS="-C overflow-checks" or overflow-checks = true in the Cargo.toml profile can override this. Overflow checking is usually enabled by default in debug mode."

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ rand_core = { version = "0.6", features = ["getrandom"] }
 rayon = "1.7"
 rcgen = { version = "0.13", features = ["pem", "aws_lc_rs"], default-features = false }
 regex = "1"
-ring = "0.17"
+ring = ">=0.17.12"
 rsa = { version = "0.9", features = ["sha2"], default-features = false }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"


djc:
In Cargo's implementation of semver, the `^` operator is implied, so our `0.17` dependency can trivially be used with anything between `0.17.0` and `0.17.99999`. As such, this change is unnecessary -- downstream users can already use rustls with newer ring.

Shifta-Robel:
yea ring = "0.17" allows newer versions, but users can still end up with vulnerable versions like 0.17.8 by default, like I did. Upgrading to >=0.17.12 guarantees everyone gets the safe version.

djc:
The Cargo tooling encourages upgrading proactively to the latest versions which are generally believed to be the best (with the fewest bugs and security issues), so we believe it is reasonable to leave this responsibility with downstream users rather than proactively constraining their choice of dependency versions.

paolobarbolini:
I will jump in and also add, as I have seen this done many times, that `>=0.17.12` would NOT be what you want anyway, because it would allow ANY version higher than `0.17.12`, including breaking changes like `0.18`, `0.19`, `1.0`, and so on. The correct thing, if you really wanted to do this, would have been to write `0.17.12`, which, as already explained, means `>=0.17.12, <0.18` (`,` stands for `AND`).

Shifta-Robel:
Thank you for the clarification, @djc . @paolobarbolini  your'e absolutely right the correct constraint should be ">=0.17.12, <0.18.0".
I still believe preventing vulnerable versions by default to be better, but I respect your perspective.

your work here is much appreciated, thank you for all you do!

PR #2429, opened 2025-04-14
Allow client ALPN protocols to be specified at connection level

CodeMan62:
This PR implements issue #2260  by allowing ALPN protocols to be specified at the connection level instead of only in ClientConfig. This enables clients to reuse a single config for connections that differ only in their ALPN protocols.
The implementation adds new_with_alpn() methods to both client connection types while maintaining backward compatibility. The handshake code prioritizes connection-level protocols over config-level protocols when provided.
I have run tests locally all are going well no fails in issues

PR #2438, opened 2025-04-23
Add support for connection-level ALPN protocol configuration

djc:
Cleanup of:

- #2429

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -510,6 +510,36 @@ fn alpn() {
     );
 }
 
+#[test]
+fn connection_level_alpn_protocols() {
+    let mut server_config = make_server_config(KeyType::Rsa2048);
+    server_config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+    let server_config = Arc::new(server_config);
+
+    // Config specifies `h2`
+    let mut client_config = make_client_config(KeyType::Rsa2048);
+    client_config.alpn_protocols = vec![b"h2".to_vec()];
+    let client_config = Arc::new(client_config);
+
+    // Client relies on config-specified `h2`, server agrees
+    let mut client =
+        ClientConnection::new(client_config.clone(), server_name("localhost")).unwrap();
+    let mut server = ServerConnection::new(server_config.clone()).unwrap();
+    do_handshake_until_error(&mut client, &mut server).unwrap();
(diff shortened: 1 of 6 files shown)


rustls/tests/api.rs
```diff
@@ -8389,3 +8389,32 @@ fn assert_client_sends_hello_with_one_hybrid_key_share(msg: &mut Message) -> Alt
 }
 
 const CONFIDENTIALITY_LIMIT: u64 = 1024;
+#[test]
+fn connection_level_alpn_protocols() {
```
ctz, line 8393: This test actually does nothing but call some constructors; it does not actually observe the effect of that. (in other words, if `new_with_alpn()` did say `let alpn_protocols = config.alpn_protocols.clone();`, no tests fail.)

 I suggest we lose it, and in `alpn_test_error` further up in this file, add usage of `new_with_alpn()`. Something like this patch:
 

```patch
index e214b05e..a7a9eaae 100644
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -454,6 +454,25 @@ fn alpn_test_error(
     let server_config = Arc::new(server_config);
 
     for version in rustls::ALL_VERSIONS {
+        // per-connection configuration
+        let mut server = ServerConnection::new(Arc::clone(&server_config)).unwrap();
+        let mut client = ClientConnection::new_with_alpn(
+            Arc::new(make_client_config_with_versions(
+                KeyType::Rsa2048,
+                &[version],
+            )),
+            server_name("localhost"),
+            client_protos.clone(),
+        )
+        .unwrap();
+        assert_eq!(client.alpn_protocol(), None);
+        assert_eq!(server.alpn_protocol(), None);
+        let error = do_handshake_until_error(&mut client, &mut server);
+        assert_eq!(client.alpn_protocol(), agreed);
+        assert_eq!(server.alpn_protocol(), agreed);
+        assert_eq!(error.err(), expected_error);
+
+        // config-level configuration
         let mut client_config = make_client_config_with_versions(KeyType::Rsa2048, &[version]);
         client_config
             .alpn_protocols


```

djc: Was already working on a test rewrite, have a look?

ctz (approved):

rustls/tests/api.rs
ctz: LGTM!

PR #2440, opened 2025-04-24
Derive iv with custom length

### LINKED ISSUE #2439: No way to derive iv value with custom length in TLS 1.3 (GOST)
**Checklist**

I'm writing a custom CryptoProvider that makes support for nation-state cryptography (GOST). The standard is established by Russian government and described in [rfc9367](https://datatracker.ietf.org/doc/html/rfc9367). It covers four new CipherSuites with two AEAD ciphers and different iv length for each cipher: 8-byte for "magma" and 16-byte for "kuznyechik". In general, GOST iv derivation process is same as [rfc8446#section-7.1](https://datatracker.ietf.org/doc/html/rfc8446#section-7.1) describes, but GOST uses different values of hash lengths in expand function `HKDF-Expand(PRK, info, L) -> OKM` (i.e. `L = 8` and `L = 16`). Rustls's deriving iv logic is placed in `derive_traffic_iv` function, which has no variants for custom iv length differs from `NONCE_LEN` (i.e. `L = 12`).

Make `derive_traffic_iv` function derive secret for custom iv length, like `derive_traffic_key` do for custom key length.

#2440

igrkir:
Fixes #2439

### DIFF
--- a/provider-example/src/aead.rs
+++ b/provider-example/src/aead.rs
@@ -44,14 +44,14 @@ impl Tls12AeadAlgorithm for Chacha20Poly1305 {
     fn encrypter(&self, key: AeadKey, iv: &[u8], _: &[u8]) -> Box<dyn MessageEncrypter> {
         Box::new(Tls12Cipher(
             chacha20poly1305::ChaCha20Poly1305::new_from_slice(key.as_ref()).unwrap(),
-            Iv::copy(iv),
+            Iv::new(iv),
         ))
     }
 
     fn decrypter(&self, key: AeadKey, iv: &[u8]) -> Box<dyn MessageDecrypter> {
         Box::new(Tls12Cipher(
             chacha20poly1305::ChaCha20Poly1305::new_from_slice(key.as_ref()).unwrap(),
-            Iv::copy(iv),
+            Iv::new(iv),
         ))
     }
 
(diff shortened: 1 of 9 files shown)


djc:
Please fix the clippy lints, and make sure there are no backwards incompatible changes in the public API.

ctz:
Unfortunately I don't think we can change `crypto::cipher::Nonce` in a non-breaking release. However, it does not appear elsewhere in any API (it is merely a helper for gluing together a sequence number and `Iv`), so we can introduce a different function or type for this calculation.

igrkir:
Many thanks for replies!

There is one more place where I have to change public API. It is new method `fn expander_iv_len(&self) -> Option<usize>` of `pub trait Tls13AeadAlgorithm`. It serves like a `fn key_len(&self) -> usize` to get iv length specific to implementation (`CryptoProvider`). Return value is optional, just for backward capability. I have no idea how to implement it while keeping public API untouched. Could you advice me how to make it properly.

Darkspirit:
https://www.consilium.europa.eu/en/policies/sanctions-against-russia-explained/#services
> The ban concerns the following services to Russia or Russian persons:
> [...]
>    IT consultancy and legal advice
> [...]
>    intellectual property rights and trade secrets (related to goods and technology covered by other sanctions)
>    software for the management of enterprises
>    software for industrial design and manufacture

https://x.com/EuromaidanPress/status/1915766524410827101

ctz:
We have discussed this internally and decided not to accept this contribution.

PR #2441, opened 2025-04-28
Improve representation of SNI, ALPN and protocol version extensions

ctz:
These are diversions that happened in #1475

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -1794,11 +1794,10 @@ impl<'a> CertificatePayloadTls13<'a> {
         false
     }
 
-    pub(crate) fn end_entity_ocsp(&self) -> Vec<u8> {
+    pub(crate) fn end_entity_ocsp(&self) -> &[u8] {
         self.entries
             .first()
             .and_then(CertificateEntry::ocsp_response)
-            .map(|resp| resp.to_vec())
             .unwrap_or_default()
     }
 
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -340,7 +340,6 @@ impl TlsListElement for ProtocolName {
 pub(crate) trait ConvertProtocolNameList {
     fn from_slices(names: &[&[u8]]) -> Self;
     fn to_slices(&self) -> Vec<&[u8]>;
-    fn as_single_slice(&self) -> Option<&[u8]>;
 }
 
 impl ConvertProtocolNameList for Vec<ProtocolName> {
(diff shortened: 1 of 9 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -359,16 +358,48 @@ impl ConvertProtocolNameList for Vec<ProtocolName> {
...
             .collect::<Vec<&[u8]>>()
     }
+}
+
+/// RFC7301 encodes a single protocol name as `Vec<ProtocolName>`
+#[derive(Clone, Debug)]
+pub struct SingleProtocolName(ProtocolName);
+
+impl SingleProtocolName {
+    pub(crate) fn new(name: impl Into<Vec<u8>>) -> Self {
```
djc, line 375: Given that this is internal, I'd externalize the conversion to `Vec<u8>` (by doing away with the `impl Into<_>` here).

rustls/src/msgs/handshake.rs
```diff
@@ -228,100 +228,130 @@ impl TlsListElement for SignatureScheme {
...
-        match self {
-            Self::HostName(name) => {
-                (name.as_ref().len() as u16).encode(bytes);
-                bytes.extend_from_slice(name.as_ref().as_bytes());
-            }
-            Self::IpAddress(r) => r.encode(bytes),
-            Self::Unknown(r) => r.encode(bytes),
+        let server_name_list = LengthPrefixedBuffer::new(Self::SIZE_LEN, bytes);
+
+        if let ServerNamePayload::SingleDnsName(dns_name) = self {
```
djc, line 267: Nit: let-else to reduce rightward drift?

rustls/src/msgs/handshake.rs
```diff
@@ -228,100 +228,130 @@ impl TlsListElement for SignatureScheme {
...
+
+                // Skip over hostname items that are actually IP addresses.
+                HostNamePayload::IpAddress(_invalid) => {
+                    warn!(
+                        "Illegal SNI extension: ignoring IP address presented as hostname ({:?})",
+                        _invalid
+                    );
+                    found = Some(Self::IpAddress);
+                }
+            }
```
djc, line 320: I feel like merging these two `match`es might be more readable? Also potentially hoist the `found.is_some()` into the match arm pattern rather than nesting it?

djc (approved):
Very nice!

rustls/src/msgs/handshake.rs
```diff
@@ -570,7 +570,65 @@ impl TlsListElement for KeyShareEntry {
...
+///
 /// RFC8446: `ProtocolVersion versions<2..254>;`
+#[derive(Clone, Copy, Debug, Default)]
+pub struct SupportedProtocolVersions {
+    pub(crate) tls13: bool,
+    pub(crate) tls12: bool,
+}
+
+impl SupportedProtocolVersions {
+    pub(crate) fn any(&self, func: impl Fn(ProtocolVersion) -> bool) -> bool {
```
cpu, line 625: WDYT about calling the `func` callback something like `filter` or `allowed` instead? I think it's a bit clearer.

rustls/src/msgs/handshake.rs
```diff
@@ -228,100 +228,130 @@ impl TlsListElement for SignatureScheme {
...
-    pub(crate) typ: ServerNameType,
-    pub(crate) payload: ServerNamePayload,
+    /// RFC6066: `ServerName server_name_list<1..2^16-1>`
+    const SIZE_LEN: ListLength = ListLength::NonZeroU16 {
+        empty_error: InvalidMessage::IllegalEmptyList("ServerNames"),
+    };
 }
-
-impl Codec<'_> for ServerName {
+/// Simplified encoding/decoding for a `ServerName` extension payload to/from `DnsName`
```
cpu, line 256: nit: missing whitespace separator here

cpu (approved):

rustls/src/msgs/handshake.rs
cpu: Though since the stdlib [uses `f`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) for this, I can be convinced it doesn't matter at all :-)

PR #2443, opened 2025-04-30
Reorganize extensions testing

ctz:
The goal of this PR is to provide a single internal API surface for all our extension testing needs. A similar idea was previously included in #1475; this is a variation that replaces that.

That allows us to remove `ClientExtension` and `ServerExtension` from `rustls::internals`, and then further privatize types that were reachable from there.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -228,7 +228,7 @@ impl TlsListElement for SignatureScheme {
 }
 
 #[derive(Clone, Debug)]
-pub enum ServerNamePayload<'a> {
+pub(crate) enum ServerNamePayload<'a> {
     /// A successfully decoded value:
     SingleDnsName(DnsName<'a>),
 
@@ -358,7 +358,7 @@ impl HostNamePayload {
 
 wrapped_payload!(
     /// RFC7301: `opaque ProtocolName<1..2^8-1>;`
-    pub struct ProtocolName, PayloadU8<NonEmpty>,
+    pub(crate) struct ProtocolName, PayloadU8<NonEmpty>,
 );
 
 /// RFC7301: `ProtocolName protocol_name_list<2..2^16-1>`
@@ -494,7 +494,7 @@ impl TlsListElement for PresharedKeyBinder {
 }
 
 #[derive(Clone, Debug)]
-pub struct PresharedKeyOffer {
+pub(crate) struct PresharedKeyOffer {
     pub(crate) identities: Vec<PresharedKeyIdentity>,
     pub(crate) binders: Vec<PresharedKeyBinder>,
 }
@@ -532,7 +532,7 @@ impl TlsListElement for ResponderId {
 }
 
 #[derive(Clone, Debug)]
-pub struct OcspCertificateStatusRequest {
+pub(crate) struct OcspCertificateStatusRequest {
     pub(crate) responder_ids: Vec<ResponderId>,
     pub(crate) extensions: PayloadU16,
 }
@@ -553,7 +553,7 @@ impl Codec<'_> for OcspCertificateStatusRequest {
 }
 
 #[derive(Clone, Debug)]
-pub enum CertificateStatusRequest {
+pub(crate) enum CertificateStatusRequest {
     Ocsp(OcspCertificateStatusRequest),
     Unknown((CertificateStatusType, Payload<'static>)),
 }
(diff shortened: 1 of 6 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -1397,6 +1423,29 @@ impl Codec<'_> for HelloRetryRequest {
...
+        )?;
+
+        let mut bytes = Vec::new();
+        self.legacy_version.encode(&mut bytes);
+        HELLO_RETRY_REQUEST_RANDOM.encode(&mut bytes);
+        self.session_id.encode(&mut bytes);
+        self.cipher_suite.encode(&mut bytes);
+        Compression::Null.encode(&mut bytes);
+        bytes.extend_from_slice(&extensions);
+        Ok(bytes)
```
djc, line 1446: Can we lean on `payload_encode()` here to avoid duplicating (some of) this?

rustls/src/msgs/handshake.rs
```diff
@@ -997,7 +997,33 @@ pub struct ClientHelloPayload {
...
+            self.extensions
+                .iter()
+                .map(|ext| (ext.ext_type(), ext.get_encoding()))
+                .collect(),
+            extras,
+            visitor,
+        )?;
+
+        let mut bytes = Vec::new();
+        self.client_version.encode(&mut bytes);
```
djc, line 1018: Can we lean on `payload_encode()` to avoid some duplication here?

ctz:
@djc ready for another look -- this no longer adds any code to the core crate.

djc (commented):
I'm not sure I'm a big fan.

I think you have the original goal that most of the tests should be on the outside of the crate, but I am not entirely sure that I understand the value of this ideal. It seems reasonable to me to have tests written on the inside of the crate boundary, especially since we can also use modules as an abstraction boundary. On the other hand, it comes with this crutch of the `internals` module which is like semi-public but not stable, which works around rather than with a bunch of the tooling.

But the idea of growing a second implementation of encoding and decoding messages seems suboptimal to me -- this is adding a whole bunch of code for the sole purpose of supporting some tests that want to inspect/change messages that contain extensions, which means we now have to review/maintain a lot of test-only code in addition to the production library code that is our actual product.

I agree with is that it is better to have higher-level tests that are not tightly integrated to low-level (internal) APIs, but I think I prefer some kind of mid-level API that avoids too much integration without requiring what is substantially a rewrite of the logic used in the library -- and I think it would be okay to have that inside the library.

(I also think both our 8.5k-line `tests/api.rs` and 3.4k-line `msgs/handshake.rs` could both more organization/splitting up.)

ctz:
> I'm not sure I'm a big fan.

I think the main issue here is we are testing several different things, but testing them in the same way. The original purpose of `tests/api.rs` is to (as the name hints) exercise the public API in realistic, representative and complete ways, so we can be sure that the public surface is sufficient to do all the things we want to. Naturally, if we move those tests into the crate, we lose that.

The other property of those tests is they don't become invalid from testing rustls against itself. In other words, if we screw up the encoding and decoding of a `ClientHello` in a consistent way, these tests will pass even though we no longer interop with other implementations. That is fine, but is undesirable for protocol-level tests and for those we want an implementation that is independent (bogo, openssl, or a some straightforward encoding/decoding written in a different style that this PR introduced).

As time has gone on we've added to `tests/api.rs` things like protocol error testing, to give coverage of the cases that bogo doesn't cover. Those things don't need to be tested outside the crate. Perhaps putting them inside the crate might allow them to become smaller tests, but if not we'd still need to (say) create a `ClientConnection` with the boilerplate that entails. Maybe we could find a way of reusing `tests/common/mod.rs` both inside and outside the crate, maybe with a `#[path]` or something.

Lastly, the other constraint is the desire to have tightly defined and misuse-resistant internal APIs, that those by design make it more difficult to construct erroneous messages for testing. As an example, ffac73a071248807bc96cc84230518aaa38b5b3f makes it impossible to represent the clienthello issued by  https://github.com/rustls/rustls/blob/7f8df8979e299ba15230ba1ecce319f56dc80571/rustls/tests/api.rs#L1549 , that would prevent moving this test into the core crate and reusing the same extension representation.

tldr: onboard with moving the protocol tests into the crate, but i still think some of the encoding cases need to diverge from the "real" code

ctz:
> Maybe we could find a way of reusing `tests/common/mod.rs` both inside and outside the crate, maybe with a `#[path]` or something.

This seems to be extremely difficult:

- a crate can't refer to itself via its name, so code that works outside the crate does not work inside it, and vv.
- a testing-only crate that contains common/mod.rs is basically unusable inside the crate, because all the types are from a separate compilation of the same code (and so none of the types match).

PR #2451, opened 2025-05-06
manual: add a short howto debugging section

cpu:
Cover how to collect verbose logs with `RUST_LOG` and offer some starting point instructions on collecting a pcap and using `SSLKEYLOGFILE`.

The new issue template for bug reports is updated to include a pointer to the new manual section.

### DIFF
--- a/rustls/src/manual/howto.rs
+++ b/rustls/src/manual/howto.rs
@@ -54,4 +54,71 @@ messages are never delimited by the close of the TCP connection), it can uncondi
 ignore `UnexpectedEof` errors from rustls.
 
 [^2]: <https://datatracker.ietf.org/doc/html/rfc8446#section-6.1>
+
+# Debugging
+
+If you encounter a bug with Rustls it can be helpful to collect up as much diagnostic
+information as possible.
+
+## Collecting logs
+
+If your bug reproduces with one of the [Rustls examples] you can use the
+[`RUST_LOG`] environment variable to increase the log verbosity. If you're using
+your own application, you may need to configure it with a logging backend
+like `env_logger`.
+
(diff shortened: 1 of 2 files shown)


rustls/src/manual/howto.rs
```diff
@@ -54,4 +54,71 @@ messages are never delimited by the close of the TCP connection), it can uncondi
...
+capture to `rustls.pcap`.
+
+Another approach is to use `tcp and port XXXX` instead of `tcp and dst host XX.XX.XX.XX`
+to capture all traffic to a specific port instead of a specific host server.
+
+[`tcpdump`]: https://www.redhat.com/en/blog/introduction-using-tcpdump-linux-command-line
+
+### SSLKEYLOGFILE
+
+If the bug you are reporting happens after the initial handshake you may also wish to
```
djc, line 107: "after the initial handshake" reads a little confusing to me, and might be interpreted by some as the whole handshake? Maybe something like "If the bug you are reporting happens later in the handshake phase [..]"?

djc (approved):

djc:
This is great, thanks for writing it up!

djc:
The `cargo-check-external-types` sync got cherrypicked into #2453 to unblock CI. 👍

rustls/src/manual/howto.rs
cpu: I replaced it with "after data is encrypted". I think that's the most important part and removes any need to understand when it happens in the overall handshake.

cpu:
> The cargo-check-external-types sync got cherrypicked into #2453 to unblock CI. 👍

Thanks! Rebased.

ctz (approved):
Nice!

PR #2452, opened 2025-05-07
Update dependencies

djc:


ctz:
> ```
> thread 'main' panicked at rustls/src/verifybench.rs:222:14:
> called `Result::unwrap()` on an `Err` value: InvalidCertificate(UnknownIssuer)
> ```

This one relied on one of the recently-removed roots:

>         Issuer: C = IE, O = Baltimore, OU = CyberTrust, CN = Baltimore CyberTrust Root

ctz (approved):

PR #2453, opened 2025-05-07
Update verifybench test data

ctz:
This was four years old, and the servo tests required a now-removed root.

The process for this was:

- delete all the old data
- run admin/capture-certdata
- adjust the time in verifybench.rs, and adjust chains that changed in length


djc:
Cherry-pick @cpu's check external types upgrade?

Maybe add a short README to the testdata dir and/or some comments to the verifybench module about this update process?

ctz:
> Cherry-pick @cpu's check external types upgrade?
> 
> Maybe add a short README to the testdata dir and/or some comments to the verifybench module about this update process?

Done both of these.

djc (approved):

PR #2458, opened 2025-05-15
Small clippy fixes

cpu:
The first two commits fix the [clippy task failing for `main`](https://github.com/rustls/rustls/actions/runs/15052009760/job/42308873568) from new findings spit out by the brand new [1.87 release](https://blog.rust-lang.org/2025/05/15/Rust-1.87.0/). The third commit fixes nightly clippy for me locally. If it's too noisy of a diff we can drop the 3rd commit and take only the 1st/2nd for now.

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/sign.rs
+++ b/rustls/src/crypto/aws_lc_rs/sign.rs
@@ -117,7 +117,7 @@ impl RsaSigningKey {
             }
         }
         .map_err(|key_rejected| {
-            Error::General(format!("failed to parse RSA private key: {}", key_rejected))
+            Error::General(format!("failed to parse RSA private key: {key_rejected}"))
         })?;
 
         Ok(Self {
@@ -428,7 +428,7 @@ mod tests {
         ));
 
         let k = any_supported_type(&key).unwrap();
-        assert_eq!(format!("{:?}", k), "EcdsaSigningKey { algorithm: ECDSA }");
+        assert_eq!(format!("{k:?}"), "EcdsaSigningKey { algorithm: ECDSA }");
         assert_eq!(k.algorithm(), SignatureAlgorithm::ECDSA);
 
         assert!(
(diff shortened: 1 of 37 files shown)


djc (approved):

rustls/src/crypto/aws_lc_rs/sign.rs
```diff
@@ -117,7 +117,7 @@ impl RsaSigningKey {
             }
         }
         .map_err(|key_rejected| {
-            Error::General(format!("failed to parse RSA private key: {}", key_rejected))
+            Error::General(format!("failed to parse RSA private key: {key_rejected}"))
```
ctz, line 120: I think this commit comes from  `cargo clippy --fix`, but I think `cargo clippy --fix --all-features --all-targets` produces a more uniform result. eg, this link also fires on `src/crypto/ring/sign.rs`

cpu: > I think this commit comes from cargo clippy --fix, but I think cargo clippy --fix --all-features --all-targets produces a more uniform result. 

Ahh yes, you're right. Let me rework.

ctz: I prefer your commit message, but here's the same commit I came up with https://github.com/rustls/rustls/commit/9b0c22d94b82524d1d4723f94afc5123fbafa0ca

cpu: I regenerated that commit (and while reviewing the diff, also noticed a small manual cleanup opportunity I tacked on separate from the machine generated diff: 4d4a01436e40ad62ab54d37646ae1bbb9c803508)

ctz (approved):

rustls/src/crypto/aws_lc_rs/sign.rs
cpu: Ohhh right, that was an intentional thing because of `no-std`: 

```
error: unused variable: `invalid`
   --> rustls/src/msgs/handshake.rs:311:44
    |
311 |                 HostNamePayload::IpAddress(invalid) => {
    |                                            ^^^^^^^ help: if this is intentional, prefix it with an underscore: `_invalid`
    |
    = note: `-D unused-variables` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_variables)]`
````

Will drop :sweat:

ctz: shower thought: i wonder if there is a way to avoid this by the logging macros becoming `drop(format!(<args>))` or something? But obviously we'd need to be confident that the compiler doesn't actually do the formatting...

PR #2460, opened 2025-05-15
Rework clippy setup

ctz:
Prior to this PR, running `admin/clippy` (which is a common part of my workflow) produces a wall of 150 warnings. This PR fixes that by moving the clippy configuration into `rustls/`.


cpu (approved):
Good idea

ctz:
(overlooked nightly clippy in ci)

djc (approved):

PR #2465, opened 2025-05-21
Enable more lints on internal crates

ctz:
Internal crates now enable the default-allow lints that the core crate does, with the exception of ones about documentation for public items.


djc (approved):
Nice!

Personally I feel like the `Arc::clone()` idiom isn't worth it...

cpu (approved):

PR #2466, opened 2025-05-21
Fix "Format (unstable)" job in CI

ctz:
This has been quietly complaining to itself[^1] for a little bit since 177cb0cef4d97b09335948b9265112deb03f2b23 as it wanted to undo the `style_edition = "2024"` changes. Fix that, and then do a pass tidying imports.

[^1]:  recall it is a `continue-on-error` job to avoid actually relying on the behaviour of nightly rustfmt


djc (approved):
Nice!

cpu (approved):

PR #2467, opened 2025-05-21
Fix nightly  docs

ctz:
The intention is that these appear at https://rustls.dev/docs/ but this was broken in af0229b8e9f4ba759490f69d626d81cbc5a03434 because we stopped copying across all cargo doc's output files. That meant https://rustls.dev/docs/ appeared without its CSS.


cpu (approved):
Thanks!

PR #2470, opened 2025-05-23
Fix bug in crypto::aws_lc_rs::pq::hybrid::Layout

cjpatton:
The share splitter function swaps the order of the shares whenever `post_quantum_first` is `true`.


ctz (approved):
In retrospect it probably wasn't a great idea to make this more flexible than we needed and could test.

djc:
Why can't we test it?

PR #2471, opened 2025-05-26
Support `secp256r1mlkem768`

ctz:
This adds `rustls::crypto::aws_lc_rs::kx_group::SECP256R1MLKEM768`. It is added to `ALL_KX_GROUPS`, but not `DEFAULT_KX_GROUPS`; this means that -- for now -- it requires explicit action to use (like plain MLKEM768).

Generally I don't expect this to be widely deployed, and (eg) isn't supported right now by pq.cloudflareresearch.com. But I found that openquantumsafe.org does support it, so add a quick interop test with that server.

This also acts as a regression test for #2470

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -131,6 +131,8 @@ jobs:
       - name: Check rustls-post-quantum client
         run: cargo run --locked -p rustls-post-quantum --example client | grep 'kex=X25519MLKEM768'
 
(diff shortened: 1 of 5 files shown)


djc:
Nice!

cpu (approved):
LGTM. A manual kick of the [daily-tests](https://github.com/rustls/rustls/actions/runs/15258855075) run seems happy too.

PR #2473, opened 2025-05-29
Convert more low-level integration tests into unit tests

ctz:
Previously: #2472, #2462

This concludes the set of tests that #1475 touched.

Then, the size of `rustls::internals` is minimised, which demonstrates that these changes have allowed (eg) `ServerExtension` and `ClientExtension` to become private.

Finally, the `CertificateType` enum type is added to the public API. This is the sort of issue that the api.rs-level tests are meant to find, but was entangled in the noise of `rustls::internal::*` use.


rustls/src/lib.rs
```diff
@@ -647,6 +647,7 @@ pub mod server {
     #[cfg(feature = "std")]
     pub use server_conn::{AcceptedAlert, Acceptor, ReadEarlyData, ServerConnection};
 
+    pub use crate::msgs::enums::CertificateType;
```
djc, line 650: Should it also be moved from `crate::msgs::enums` to `crate::enums`?

djc (approved):
Seems nice!

PR #2476, opened 2025-05-30
feat: expose the number of received TLS1.3 resumption tickets

Frando:
This adds a public function to get the count of TLS1.3 resumption tickets that were received in a client-side rustls session.

Currently, this information is not available outside of the session store. Downstream users may want to access this, though, to keep a connection open until tickets are received. When doing a 0RTT connection, it may happen that the application-side logic is done before the tickets are received. In that case, if expecting to do further 0rtt connections in the future, it may be beneficial to keep the connection open until the tickets are received. With this patch, this is possible without having to wrap or hack around the session store.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -774,6 +774,11 @@ mod connection {
             self.inner.core.data.ech_status
         }
 
+        /// Returns the number of TLS1.3 tickets that have been received.
+        pub fn tls13_tickets_received(&self) -> u32 {
+            self.inner.tls13_tickets_received
+        }
+
         /// Return true if the connection was made with a `ClientConfig` that is FIPS compatible.
         ///
         /// This is different from [`crate::crypto::CryptoProvider::fips()`]:
@@ -918,6 +923,11 @@ impl UnbufferedClientConnection {
         self.inner
             .core
             .dangerous_into_kernel_connection()
+    }
+
+    /// Returns the number of TLS1.3 tickets that have been received.
+    pub fn tls13_tickets_received(&self) -> u32 {
+        self.inner.tls13_tickets_received
     }
 }
 
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -1527,6 +1527,10 @@ impl ExpectTraffic {
             protocol: cx.common.protocol,
             quic: &cx.common.quic,
         };
+        cx.common.tls13_tickets_received = cx
+            .common
+            .tls13_tickets_received
+            .saturating_add(1);
         self.handle_new_ticket_impl(&mut kcx, nst)
     }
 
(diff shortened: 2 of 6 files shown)


rustls/src/quic.rs
```diff
@@ -219,6 +219,11 @@ mod connection {
         pub fn is_early_data_accepted(&self) -> bool {
             self.inner.core.is_early_data_accepted()
         }
+
+        /// Returns the number of TLS1.3 tickets that have been received.
+        pub fn tls13_session_tickets_received(&self) -> u32 {
```
djc, line 224: Nit: let's drop the `session_` segment of this, similar to the other method and the inner API?

djc (approved):
This seems reasonable to me, thanks!

djc:
(Please squash your commits.)

rustls/src/quic.rs
Frando: :+1:

rustls/src/client/client_conn.rs
```diff
@@ -774,6 +774,11 @@ mod connection {
             self.inner.core.data.ech_status
         }
 
+        /// Returns the number of TLS1.3 tickets that have been received.
+        pub fn tls13_tickets_received(&self) -> u32 {
```
cpu, line 778: I think the `UnbufferedClientConnection` needs a `pub` accessor like this as well.

rustls/src/client/tls13.rs
```diff
@@ -1527,6 +1527,7 @@ impl ExpectTraffic {
             protocol: cx.common.protocol,
             quic: &cx.common.quic,
         };
+        cx.common.tls13_tickets_received += 1;
```
cpu, line 1530: It might be unwarranted paranoia, but I'm idly wondering whether this should be a saturating or wrapping add to avoid a panic on overflow. I don't think there is a session ticket limit enforced earlier in this codepath, though I'm also not confident it's feasible to transmit ~four billion tickets on one conn :laughing: 

What do other folks think?

rustls/tests/api.rs
```diff
@@ -4753,6 +4753,7 @@ fn tls13_stateful_resumption() {
     // full handshake
     let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
     let (full_c2s, full_s2c) = do_handshake(&mut client, &mut server);
+    assert_eq!(client.tls13_tickets_received(), 2);
```
cpu, line 4756: If the unbuffered conn API is updated it'd be nice to sneak an update in to one of its unit tests in `rustls/tests/unbuffered.rs`. Maybe the `eary_data()` test?

cpu (approved):
Thanks!

rustls/src/client/tls13.rs
djc: Hah... I guess it can't hurt to do a `saturating_add()` here?

rustls/tests/api.rs
Frando: I added the assert.

rustls/src/client/client_conn.rs
cpu: Thanks!

cpu (approved):

PR #2478, opened 2025-06-02
Eliminate redundant `HandshakeMessagePayload::typ` field 

ctz:
This is quite a big but somewhat uninteresting change.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -67,7 +67,7 @@ macro_rules! wrapped_payload(
 );
 
 #[derive(Clone, Copy, Eq, PartialEq)]
-pub struct Random(pub(crate) [u8; 32]);
+pub(crate) struct Random(pub(crate) [u8; 32]);
 
 impl fmt::Debug for Random {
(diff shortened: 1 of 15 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -2765,7 +2765,40 @@ impl HandshakePayload<'_> {
...
+            EncryptedExtensions(_) => HandshakeType::EncryptedExtensions,
+            KeyUpdate(_) => HandshakeType::KeyUpdate,
+            Finished(_) => HandshakeType::Finished,
+            CertificateStatus(_) => HandshakeType::CertificateStatus,
+            MessageHash(_) => HandshakeType::MessageHash,
+            Unknown((t, _)) => *t,
+        }
+    }
+
+    fn wire_handshake_type(&self) -> HandshakeType {
```
djc, line 2797: Maybe this should be a method on `HandshakeType` instead?

djc (approved):

rustls/src/msgs/handshake.rs
ctz: since this only has one caller, and exists primarily to be a place for the comment to live, i think i prefer to keep it private

PR #2479, opened 2025-06-02
Introduce specific error for unsupported signatures

ctz:
`BadSignature` convolves two cases with different causes: a genuinely bad signature, where we positively identified that a signature was incorrect; and where we don't support a signature algorithm and can't say whether it was valid or not.


rustls/src/error.rs
```diff
@@ -413,6 +413,9 @@ pub enum CertificateError {
     /// issuer.
     BadSignature,
 
+    /// A signature inside a certificate or on a handshake was made with an unsupported algorithm.
+    UnsupportedSignatureAlgorithm,
```
djc, line 417: Could we have the value for the signature algorithm in here? (Plus potentially the TLS version if needed to disambiguate?)

ctz: I think the most useful additional data would be the public key and signature `AlgorithmIdentifier `s (though not that exact type, as it is `'static`), and certainly webpki's `verify_signature` function has that data at the point it produces the source errors here.

ctz: -> https://github.com/rustls/webpki/issues/356

Issue #2485, opened 2025-06-11
Self supplied CipherSuites without TLS_EMPTY_RENEGOTIATION_INFO_SCSV


SpaceMoehre:
**Checklist**
* [ X] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
When supplying a self defined list of cipher suites to a crypto provider and using this to open a TLS connection the ciphersuite TLS_EMPTY_RENEGOTIATION_INFO_SCSV - id:255 is always appended to the list. Is there maybe already a way to prevent that?

**Describe the solution you'd like**
An optional flag in the `CryptoProvider` . if set, TLS_EMPTY_RENEGOTIATION_INFO_SCSV will not be pushed to ciphersuites.

**Describe alternatives you've considered**
A way to supply the vec for client hello directly.

**Additional context**
I found that for the client hello message, there will always be this ciphersuite pushed to final vec.

`src/client/hs.rs`:line 442:
```rust 
    let mut cipher_suites: Vec<_> = config
        .provider
        .cipher_suites
        .iter()
        .filter_map(|cs| match cs.usable_for_protocol(cx.common.protocol) {
            true => Some(cs.suite()),
            false => None,
        })
        .collect();
    // We don't do renegotiation at all, in fact.
    cipher_suites.push(CipherSuite::TLS_EMPTY_RENEGOTIATION_INFO_SCSV);
```

no matter the content of my vec, there will always be this Ciphersuite pushed.

This is the code I'm working with:

```rust
use rustls::{crypto::ring::cipher_suite::*, CipherSuite, CipherSuiteCommon, SupportedCipherSuite, Tls12CipherSuite};

#[tokio::main(flavor = "current_thread")]
async fn main() -> Result<(), Box<dyn std::error::Error>> {


    let cipher_suites: Vec<rustls::SupportedCipherSuite> = vec![
        TLS13_AES_128_GCM_SHA256,
        TLS13_CHACHA20_POLY1305_SHA256,
        TLS13_AES_256_GCM_SHA384,
        TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
        TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
        TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
        TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
[...]

PR #2486, opened 2025-06-11
Only include renegotiation SCSV for TLS1.2 attempts

ctz:
On reviewing #2485, I noticed we only want to include this for hellos where TLS1.2 is offered.


cpu (approved):
Might be nice to have a test to catch unintended regression if it's not a big pain.

cpu (approved):
🚀

djc (approved):

PR #2489, opened 2025-06-12
Commence 0.24 development branch

ctz:
So far this is a version bump, an MSRV bump, and followups.

The target branch, dev-0.24, is equal with main.

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -734,10 +734,7 @@ impl<Data> ConnectionCommon<Data> {
     /// [`reader()`]: ConnectionCommon::reader
     pub fn read_tls(&mut self, rd: &mut dyn io::Read) -> Result<usize, io::Error> {
         if self.received_plaintext.is_full() {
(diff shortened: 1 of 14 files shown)


ctz:
> Error running benchmarks

Ran rustup on benchmarking machine

rustls/src/conn.rs
```diff
@@ -734,10 +734,7 @@ impl<Data> ConnectionCommon<Data> {
     /// [`reader()`]: ConnectionCommon::reader
     pub fn read_tls(&mut self, rd: &mut dyn io::Read) -> Result<usize, io::Error> {
         if self.received_plaintext.is_full() {
-            return Err(io::Error::new(
-                io::ErrorKind::Other,
-                "received plaintext buffer full",
-            ));
+            return Err(io::Error::other("received plaintext buffer full"));
```
djc, line 737: Would suggest that these belong in the commit that bumps the MSRV.

djc (approved):

PR #2493, opened 2025-06-13
Reduce size of ciphersuite enum

ctz:
fixes #2172


cpu:
> These now come from IANA, with significant filtering.

Out of curiosity, was this manual or programmatic?

ctz:
> > These now come from IANA, with significant filtering.
> 
> Out of curiosity, was this manual or programmatic?

I pushed the script here: https://github.com/ctz/junk-drawer/blob/main/iana-csv-scripts/iana.py (but I view it as a one-off)

Issue #2074, opened 2024-08-07
Generalize error return from TicketSwitcher generator fn


cpu:
The [`TicketSwitcher`](https://docs.rs/rustls/0.23.12/rustls/ticketer/struct.TicketSwitcher.html) type's generator callback is expressed as:

```rust
 generator: fn() -> Result<Box<dyn ProducesTickets>, GetRandomFailed>,
```

The choice of `GetRandomFailed` as the only possible error feels over-fit to the `*ring*` based CHACHA20_POLY1305 construction. Other errors might occur when creating a ticketer in the abstract. For example in the process of writing an RFC 5077 ticketer in https://github.com/rustls/rustls/pull/2066 we had to map other possible errors to `GetRandomFailed` because changing this constructor to return `Error` would break semver.

We should revisit this for the next breaking release.

ctz:
Fixed via #2494 / #2534

PR #2494, opened 2025-06-14
Remove deprecated TicketSwitcher, use TicketRotator instead.

bdaehlie:


djc:
I think this should target the 0.24 branch. Is there an issue about this? If so, suggest mentioning it in the PR description.

bdaehlie:
I didn't see an issue, I can file one if you want. I will switch branches.

bdaehlie:
I was doing this in preparation for working on issue #2074

djc:
(Changed base branch to dev-0.24.)

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -40,7 +40,7 @@ impl Ticketer {
     pub fn new<M: crate::lock::MakeMutex>(
         time_provider: &'static dyn TimeProvider,
     ) -> Result<Arc<dyn ProducesTickets>, Error> {
-        Ok(Arc::new(crate::ticketer::TicketSwitcher::new::<M>(
+        Ok(Arc::new(crate::ticketer::TicketRotator::new::<M>(
```
ctz: This is very strange; I don't think this actually will compile. And it doesn't on `main` either, not since this code was first introduced back for 0.23.5!

Which is mainly a concern because this PR's CI is green it that shows we don't have any coverage of `cargo build --lib --no-default-features --features ring,hashbrown`.

I suggest we delete this entire function, and the matching one in `aws_lc_rs/ticketer.rs`. And then alter the feature gates on these files from `#[cfg(any(feature = "std", feature = "hashbrown"))]` to just `#[cfg(feature = "std")]`

ctz: I suggest: `git cherry-pick 6584005e8c22979548daac188d78abf57aa2a71b` and then using `git rebase -i` to put that commit above this one.

djc:
I proactively merged main into dev-0.24, so this needs to rebase -- then Joe's commit should fall away.

bdaehlie:
Closing this in favor of #2513

PR #2501, opened 2025-06-17
Pending merge of dev-0.24

ctz:


djc:
I'm not sure what the purpose of this PR is at this point, but if it's going to generate notifications on every merge to dev-0.24 I'd prefer to keep it closed until we're getting ready to actually merge it.

ctz:
> I'm not sure what the purpose of this PR is at this point, but if it's going to generate notifications on every merge to dev-0.24 I'd prefer to keep it closed until we're getting ready to actually merge it.

I opened this to see if we can use this bit of github UI to help keep the branch up to date:

image

(but if it's noisy, it seems that assistance is not a good trade-off)

Issue #2119, opened 2024-09-10
Reevaluate additions to traits with default implementations


ctz:
In quite a few places we've avoided a breaking change by adding functions to traits with default implementations, but that may not be the best design. Examples:

- [`SupportedKxGroup::ffdhe_group()`](https://docs.rs/rustls/latest/rustls/crypto/trait.SupportedKxGroup.html#method.ffdhe_group) (ditto `ActiveKeyExchange`)
- `SupportedKxGroup::usable_for_version()`
- `SigningKey::public_key()`

When we next open for breaking changes, let's review all these and prefer to make them explicit.

ctz:
Furthermore:

- `ActiveKeyExchange::hybrid_component()` introduced in #2136  -- see https://github.com/rustls/rustls/pull/2136#issuecomment-2511339286 for one idea

ctz:
Additionally:

- https://github.com/rustls/rustls/pull/2256#issuecomment-2517882500 gives a ~6% TLS1.2 resumption speed bump, though is not applicable to aws-lc-rs's TLS1.2 PRF implementation. I decided it wasn't worth it with the bifurcation of the `Prf` trait. However, a cut-over version would probably be cleaner.

ctz:
Also:

* #2315

ctz:
I believe all these are now covered.

ctz:
> * `ActiveKeyExchange::hybrid_component()` introduced in [Implement post-quantum hybrid KX optimisation #2136](https://github.com/rustls/rustls/pull/2136)  -- see [Implement post-quantum hybrid KX optimisation #2136 (comment)](https://github.com/rustls/rustls/pull/2136#issuecomment-2511339286) for one idea

Actually, no, this one remains.

PR #2504, opened 2025-06-18
Move `usable_for_version()` from `SupportedKxGroup` to `NamedGroup`

ctz:
The fact returned by this function is not specific to _an implementation_ described by `SupportedKxGroup`, but a standardised algorithm named by `NamedGroup`.  It would be an error for two different implementations of `X25519MLKEM768` (say) to differ in their implementation of this function; so this rearrangement makes that impossible.

The trade-off of this arrangement is that it `NamedGroup::Unknown(_)` is unhandled, which means a correct implementation of new TLS1.3-only groups requires us to add a variant.  The second commit resolves this
for existing IANA reservations.

This is further work toward #2119


rustls/src/client/client_conn.rs
```diff
@@ -430,11 +430,15 @@ impl ClientConfig {
         group: NamedGroup,
         version: ProtocolVersion,
     ) -> Option<&'static dyn SupportedKxGroup> {
-        self.provider
-            .kx_groups
-            .iter()
-            .copied()
-            .find(|skxg| skxg.usable_for_version(version) && skxg.name() == group)
+        match group.usable_for_version(version) {
```
djc: Nit: maybe do an early return for `false` to reduce rightward drift?

rustls/src/client/hs.rs
```diff
@@ -262,8 +262,8 @@ fn emit_client_hello_for_retry(
         .provider
         .kx_groups
         .iter()
-        .filter(|skxg| supported_versions.any(|v| skxg.usable_for_version(v)))
         .map(|skxg| skxg.name())
+        .filter(|group| supported_versions.any(|v| group.usable_for_version(v)))
```
djc: Nit: consider using `filter_map()`?

rustls/src/msgs/enums.rs
```diff
@@ -197,6 +198,22 @@ impl NamedGroup {
             _ => KeyExchangeAlgorithm::ECDHE,
         }
     }
+
+    pub fn usable_for_version(&self, version: ProtocolVersion) -> bool {
+        match version {
+            ProtocolVersion::TLSv1_3 => true,
+            _ => !matches!(
```
djc, line 220: Makes me wonder if this could be keyed off of `SupportedProtocolVersion` instead of `ProtocolVersion`...

djc (approved):

rustls/src/msgs/enums.rs
```diff
@@ -197,6 +213,29 @@ impl NamedGroup {
...
+                self,
+                // <https://datatracker.ietf.org/doc/draft-ietf-tls-mlkem/>
```
cpu: nit: since these ref's are on the `NamedGroup` enum variants, do they add much here? I'm lightly leaning towards thinking they make the `matches!` harder to read.

If the intent is to demonstrate why they're not appropriate for < TLS 1.3, maybe they should point to a specific section instead of the overall document?

rustls/src/msgs/enums.rs
ctz: At the moment `SupportedProtocolVersion` values don't get close to the callsites of this, so this would be a pretty big refactor to arrange. Will look at that in the future.

rustls/src/msgs/enums.rs
ctz: removed these. i was hoping to quote a sentence from each making their restriction to TLS1.3 clear, but actually they tend not to mention this except in the title.

Issue #2448, opened 2025-05-06
TLS error: invalid peer certificate: BadSignature


stephane-dubois:
**Checklist**
* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**
When hitting a specific website, I get this error:
TLS error: invalid peer certificate: BadSignature
Hitting the same website with a browser doesn't give any error.

**To Reproduce**
cargo run --bin tlsclient-mio -- --http --cafile ~/custom-ca.pem mywebsite.mydomain.com 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `/Users/duboisst/Workspace/RustExample/rustls/target/debug/tlsclient-mio --http --cafile /Users/duboisst/custom-ca.pem mywebsite.mydomain.com`
TLS error: invalid peer certificate: BadSignature
Connection closed

**Applicable Version(s)**
Rustls 0.23.27
macOS 15.3

**Expected behavior**
cargo run --bin tlsclient-mio -- --http --cafile ~/custom-ca.pem mywebsite.mydomain.com 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `/Users/duboisst/Workspace/RustExample/rustls/target/debug/tlsclient-mio --http --cafile /Users/duboisst/custom-ca.pem mywebsite.mydomain.com`
HTTP/1.1 200 OK

**Additional context**
This is private website not exposed on internet so you wont be able to reproduce, but I can provide the certificate and the ca file if needed

djc:
Does it work with 0.23.26?

stephane-dubois:
No it doesn't work with 0.23.26

stephane-dubois:
Which TLS stack?

$ nmap --script ssl-enum-ciphers -p 443 mywebsite.mydomain.com
 
PORT    STATE SERVICE
443/tcp open  https
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384 (ecdh_x25519) - A
|     compressors: 
|       NULL
[...]

PR #2505, opened 2025-06-19
Support `_ABSENT_PARAMS` PKCS#1 signature algorithms

ctz:
This is the final part of #2448

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ subtle = { version = "2.5.0", default-features = false }
 time = { version = "0.3.6", default-features = false }
 tikv-jemallocator = "0.6"
 tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"] }
-webpki = { package = "rustls-webpki", version = "0.103.2", features = ["alloc"], default-features = false }
+webpki = { package = "rustls-webpki", version = "0.103.3", features = ["alloc"], default-features = false }
 webpki-roots = "1"
 x25519-dalek = "2"
 x509-parser = "0.17"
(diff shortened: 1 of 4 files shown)


cpu (approved):
This seems reasonable but I'm also not sure I follow what you mean by the commit message for the first commit w.r.t `RSA_PKCS1_3072_8192_SHA384` being a "more-restricted" version of the 2048 equivalent.

Can you expand on that a little bit for my own education?

ctz:
> This seems reasonable but I'm also not sure I follow what you mean by the commit message for the first commit w.r.t `RSA_PKCS1_3072_8192_SHA384` being a "more-restricted" version of the 2048 equivalent.
> 
> Can you expand on that a little bit for my own education?

*ring*'s `RSA_PKCS1_2048_8192_SHA384` is basically what TLS1.3 calls `SignatureScheme` `rsa_pkcs1_sha384`, with the added restriction that public keys must be 2048-8192 bits in size. `RSA_PKCS1_3072_8192_SHA384` is the same thing, but public key size restriction is 3072-8192 bits. This is, I assume, intended to meet [CNSA](https://datatracker.ietf.org/doc/html/rfc9151) standards which require minimum 3072-bit RSA.

So, my argument here is that there are no signatures that are accepted by `RSA_PKCS1_3072_8192_SHA384` that are not also accepted by `RSA_PKCS1_2048_8192_SHA384`. Since this list is additive (every algorithm is allowed a chance to match and verify the signature, and first one accepting the signature wins) having both in the list doesn't make sense.

cpu:
> So, my argument here is that there are no signatures that are accepted by RSA_PKCS1_3072_8192_SHA384 that are not also accepted by RSA_PKCS1_2048_8192_SHA384.

That makes sense. Thanks for the explanation!

PR #2506, opened 2025-06-19
Make `SigningKey::public_key()` mandatory

ctz:
Retain the ability to opt-out, but this must now be done explicitly.

re #2119

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ clap = { version = "4.3.21", features = ["derive", "env"] }
 crabgrind = "=0.1.9" # compatible with valgrind package on GHA ubuntu-latest
 criterion = "0.6"
 der = "0.7"
-ecdsa = "0.16.8"
+ecdsa = { version = "0.16.8", features = ["pem"] }
 env_logger = "0.11"
 fxhash = "0.2.1"
 hashbrown = { version = "0.15", default-features = false, features = ["default-hasher", "inline-more"] }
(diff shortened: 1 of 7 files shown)


ctz:
(deeply underbaked!)

djc (approved):

Cargo.toml
```diff
@@ -55,7 +55,7 @@ clap = { version = "4.3.21", features = ["derive", "env"] }
 crabgrind = "=0.1.9" # compatible with valgrind package on GHA ubuntu-latest
 criterion = "0.6"
 der = "0.7"
-ecdsa = "0.16.8"
+ecdsa = { version = "0.16.8", features = ["pem"] }
```
cpu, line 58: I was curious what part of `ecdsa` required the `pem` feature since the new `provider-example` code is using DER. To save anyone else looking, AFAICT the answer is the blanket `impl<C> EncodePublicKey for VerifyingKey<C>`. Odd but OK.

provider-example/src/sign.rs
```diff
@@ -39,6 +39,16 @@ impl SigningKey for EcdsaSigningKeyP256 {
         }
     }
 
+    fn public_key(&self) -> Option<SubjectPublicKeyInfoDer<'_>> {
```
cpu, line 42: nit: Maybe this could be introduced up-front separate from making it mandatory?

Cargo.toml
ctz: The other odd thing about this is how the `EncodePublicKey` trait lives in the `pkcs8` crate (PKCS#8 doesn't say anything about public keys.)

PR #2507, opened 2025-06-19
Make *CertVerifier::root_hint_subjects return Arc<Vec<>> instead of &[]

### LINKED ISSUE #2497: rustls::server::danger::ClientCertVerifier cannot return changeable root_hint_subjects
Because of the method signature of [ClientCertVerifier::root_hint_subjects](https://docs.rs/rustls/latest/rustls/server/danger/trait.ClientCertVerifier.html#tymethod.root_hint_subjects)

```
fn root_hint_subjects(&self) -> &[DistinguishedName];
```

one cannot easily implement the trait if the available roots can change over the lifetime of the verifier, because the returned slice must live as long as `self`. This makes it hard to implement a verifier that supports reloading its contents dynamically (for example to support periodically renewing certificates).

It would be better if it had this signature:

```
fn root_hint_subjects(&self) -> impl Iterator<Item = DistinguishedName>;
```

so a naive verifier could still return a cheap slice but a more sophisticated verifier could return an iterator that implements `Drop` which allows older roots to be eventually freed after new ones are loaded.

But that is an incompatible change, so I'm not sure what would be best.

## Workarounds (not great)

Use a verifier that does not support reloading its contents. Rebuild `ServerConfig` periodically with a new verifier supplied with the freshest roots each time. But this relies on applications being able to accept a new `ServerConfig` mid-flight, which is uncommon and probably should not be expected.

Make a reloadable verifier that interns all of the root distinguished name lists it has ever seen. When the set changes, this effectively leaks the old set. The hope would be that even though root certificates are periodically renewed their distinguished names stay the same every time, so the leak happens rarely in practice.

vandry:
Returning &[DistinguishedName] makes it difficult to implement the traits for verifiers that support dynamic reloading because the returned slice must live as long as &self, even after the underlying roots have changed and the verifier wants to expire the old root_hint_subjects and replace them with new ones.

I considered several different alternate return types including a proxy object (like Box<dyn AsRef<[DistinguishedName]>>) that could keep track of when the caller is done with the return value but settled on Arc as anything more complicated did not seem worthwhile and Arc is already extensively used throughout the crate. Plain Vec<> would also be reasonable: although it would commit the API to forcing a Vec clone on every call, that's what all of the existing callers do anyway.

Fixes #2497

### DIFF
--- a/rustls-test/src/lib.rs
+++ b/rustls-test/src/lib.rs
@@ -1250,7 +1250,7 @@ impl Default for MockServerVerifier {
 #[derive(Debug)]
 pub struct MockClientVerifier {
     pub verified: fn() -> Result<ClientCertVerified, Error>,
-    pub subjects: Vec<DistinguishedName>,
+    pub subjects: Arc<Vec<DistinguishedName>>,
     pub mandatory: bool,
     pub offered_schemes: Option<Vec<SignatureScheme>>,
     expect_raw_public_keys: bool,
@@ -1269,7 +1269,7 @@ impl MockClientVerifier {
                 .build()
                 .unwrap(),
             verified,
-            subjects: get_client_root_store(kt).subjects(),
+            subjects: Arc::new(get_client_root_store(kt).subjects()),
             mandatory: true,
             offered_schemes: None,
             expect_raw_public_keys: false,
@@ -1283,8 +1283,8 @@ impl ClientCertVerifier for MockClientVerifier {
         self.mandatory
     }
 
-    fn root_hint_subjects(&self) -> &[DistinguishedName] {
-        &self.subjects
+    fn root_hint_subjects(&self) -> Arc<Vec<DistinguishedName>> {
+        self.subjects.clone()
     }
 
     fn verify_client_cert(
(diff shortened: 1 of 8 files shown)


ctz:
Thanks for taking care of this! CI is showing a few other instances that need attention in bogo/ and openssl-tests/, but other than that this looks substantively good to me.

rustls-test/src/lib.rs
```diff
@@ -1283,8 +1283,8 @@ impl ClientCertVerifier for MockClientVerifier {
         self.mandatory
     }
 
-    fn root_hint_subjects(&self) -> &[DistinguishedName] {
-        &self.subjects
+    fn root_hint_subjects(&self) -> Arc<Vec<DistinguishedName>> {
+        Arc::clone(&self.subjects)
```
djc: We prefer to avoid the `Arc::clone()` idiom in favor of just doing `self.subjects.clone()` (also in a few other spots).

djc (approved):

rustls-test/src/lib.rs
vandry: Done

cpu (approved):

PR #2509, opened 2025-06-20
Reduce small `Vec<Enum>` uses in extensions

ctz:
Instead, where possible, extract just the information we need during encoding. 

"Where possible" -- specifically where the vector only encodes a peers capabilities rather than its preferences. This is true for both of these extensions, but for things like `signature_schemes` a different approach (that also retains the preference) would be needed.

This reduces some complexity, but the main goal is reducing the size of the extensions struct: `sizeof(Option<Vec<PskKeyExchangeMode>>)` is 24 bytes, but `sizeof(Option<PskKeyExchangeModes>)` is 2 bytes.

This follows on from #2508 and #2502.

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -643,6 +643,39 @@ impl CertificateStatusRequest {
 // ---
 
 /// RFC8446: `PskKeyExchangeMode ke_modes<1..255>;`
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct PskKeyExchangeModes {
+    pub(crate) psk_dhe: bool,
+    pub(crate) psk: bool,
+}
+
+impl Codec<'_> for PskKeyExchangeModes {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        let inner = LengthPrefixedBuffer::new(PskKeyExchangeMode::SIZE_LEN, bytes);
+        if self.psk_dhe {
+            PskKeyExchangeMode::PSK_DHE_KE.encode(inner.buf);
+        }
+        if self.psk {
+            PskKeyExchangeMode::PSK_KE.encode(inner.buf);
+        }
+    }
+
+    fn read(reader: &mut Reader<'_>) -> Result<Self, InvalidMessage> {
+        let mut psk_dhe = false;
+        let mut psk = false;
+
+        for ke in TlsListIter::<PskKeyExchangeMode>::new(reader)? {
+            match ke? {
+                PskKeyExchangeMode::PSK_DHE_KE => psk_dhe = true,
+                PskKeyExchangeMode::PSK_KE => psk = true,
+                _ => continue,
+            };
+        }
+
+        Ok(Self { psk_dhe, psk })
+    }
+}
+
 impl TlsListElement for PskKeyExchangeMode {
     const SIZE_LEN: ListLength = ListLength::NonZeroU8 {
         empty_error: InvalidMessage::IllegalEmptyList("PskKeyExchangeModes"),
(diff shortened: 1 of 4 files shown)


rustls/src/msgs/handshake.rs
```diff
@@ -209,6 +209,32 @@ impl UnknownExtension {
     }
 }
 
+#[derive(Clone, Debug, Default)]
```
djc: Nit: let's implement `Copy` for this?

rustls/src/server/tls12.rs
```diff
@@ -90,14 +90,14 @@ mod client_hello {
             // it means that only the uncompressed point format is
             // supported"
             // - <https://datatracker.ietf.org/doc/html/rfc8422#section-5.1.2>
-            let ecpoints_ext = client_hello
+            let supported_ec_point_formats = client_hello
                 .ec_point_formats
-                .as_deref()
-                .unwrap_or(&[ECPointFormat::Uncompressed]);
+                .clone()
```
djc: (Should avoid the `clone()` here.)

rustls/src/msgs/handshake_test.rs
```diff
@@ -779,7 +779,7 @@ fn sample_client_hello_payload() -> ClientHelloPayload {
             cookie: Some(PayloadU16::new(vec![1, 2, 3])),
             signature_schemes: Some(vec![SignatureScheme::ECDSA_NISTP256_SHA256]),
             session_ticket: Some(ClientSessionTicket::Request),
-            ec_point_formats: Some(ECPointFormat::SUPPORTED.to_vec()),
+            ec_point_formats: Some(SupportedEcPointFormats { uncompressed: true }),
```
djc: Maybe implement `Default` for `SupportedEcPointFormats`?

rustls/src/server/tls12.rs
```diff
@@ -173,19 +173,6 @@ mod client_hello {
                 ));
             }
 
-            let ecpoint = ECPointFormat::SUPPORTED
-                .iter()
-                .find(|format| ecpoints_ext.contains(format))
-                .cloned()
-                .ok_or_else(|| {
-                    cx.common.send_fatal_alert(
```
djc, line 181: `NoEcPointFormatsInCommon` doesn't seem to reappear in this diff, is that a problem?

rustls/src/msgs/handshake.rs
```diff
@@ -636,6 +636,39 @@ impl CertificateStatusRequest {
 // ---
 
 /// RFC8446: `PskKeyExchangeMode ke_modes<1..255>;`
+#[derive(Clone, Debug)]
```
djc: `Copy`, `Default`?

djc (approved):

rustls/src/msgs/handshake_test.rs
ctz: removed the derive and impl'd with `uncompressed: true`

rustls/src/server/tls12.rs
ctz: extracting to a separate commit to explain this

rustls/src/msgs/handshake.rs
```diff
@@ -209,6 +209,41 @@ impl UnknownExtension {
...
+        if self.uncompressed {
+            ECPointFormat::Uncompressed.encode(inner.buf);
+        }
+    }
+
+    fn read(r: &mut Reader<'_>) -> Result<Self, InvalidMessage> {
+        let mut spf = Self {
+            uncompressed: false,
+        };
+
```
djc: Tiny nit: consider just initializing a local `uncompressed` here and only initializing `Self { uncompressed }` in the return value expression?

djc (approved):

cpu (approved):

Issue #2496, opened 2025-06-15
Small proposal for 0.24


Tudyx:
While working with `rustls`, I've noticed 2 small inconsistencies in the types used. They don't seem to be mentioned in https://github.com/rustls/rustls/issues/2400, I want to add them to the discussion for 0.24

- `ServerName` is used most of the time for SNI, but not for [`ResolversServerCertUsingSni::add`](https://docs.rs/rustls/0.23.27/rustls/server/struct.ResolvesServerCertUsingSni.html#method.add) where it's just a `&str`.
- [`ResolversServerCertUsingSni::add`](https://docs.rs/rustls/0.23.27/rustls/server/struct.ResolvesServerCertUsingSni.html#method.add)) uses a `CertifiedKey` and [`ConfigBuilder::with_client_auth_cert`](https://docs.rs/rustls/0.23.27/rustls/struct.ConfigBuilder.html#method.with_client_auth_cert)) uses a certificate chain and a private key and then build a `CertifiedKey` under the hood.

If this was intentional in the API design, please feel free to ignore this issue

djc:
Good suggestions! I think these are mostly artifacts of how the API has evolved over time, and I think it would make sense to re-assess whether we can improve on this.

djc:
I implemented #2510 based on your suggestion.

With regards to your second suggestion, this doesn't seem entirely trivial because the `ResolvesServerCertUsingSni` doesn't retain access to a `CryptoProvider` which it would need if it wanted to be able to construct `CertifiedKey`s with the `from_der()` constructor that can check whether the certificate chain and key match.

djc:
I don't think there's anything left to do here... feel free to reopen if you disagree.

PR #2510, opened 2025-06-20
server: use DnsName to represent SNI

djc:
See #2496.

### DIFF
--- a/rustls/src/server/handy.rs
+++ b/rustls/src/server/handy.rs
@@ -192,7 +192,6 @@ impl server::ResolvesServerCert for AlwaysResolvesServerRawPublicKeys {
 
 #[cfg(any(feature = "std", feature = "hashbrown"))]
 mod sni_resolver {
-    use alloc::string::{String, ToString};
     use core::fmt::Debug;
 
     use pki_types::{DnsName, ServerName};
@@ -208,7 +207,7 @@ mod sni_resolver {
     /// on client-supplied server name (via SNI).
     #[derive(Debug)]
     pub struct ResolvesServerCertUsingSni {
-        by_name: HashMap<String, Arc<sign::CertifiedKey>>,
+        by_name: HashMap<DnsName<'static>, Arc<sign::CertifiedKey>>,
     }
 
     impl ResolvesServerCertUsingSni {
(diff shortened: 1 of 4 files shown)


rustls/src/server/handy.rs
```diff
@@ -224,14 +223,7 @@ mod sni_resolver {
...
         /// chain is syntactically faulty.
-        pub fn add(&mut self, name: &str, ck: sign::CertifiedKey) -> Result<(), Error> {
-            let server_name = {
-                let checked_name = DnsName::try_from(name)
-                    .map_err(|_| Error::General("Bad DNS name".into()))
-                    .map(|name| name.to_lowercase_owned())?;
-                ServerName::DnsName(checked_name)
-            };
-
+        pub fn add(&mut self, name: DnsName<'static>, ck: sign::CertifiedKey) -> Result<(), Error> {
```
cpu, line 225: Does this need a call to `name.to_lowercase_owned()` to match the way the conversion from `&str` to `DnsName` was previously being done, or was that conversion to lowercase unnecessary in the first place?

The pre-existing `sni_resolver_lower_cases_configured_names()` unit tests reads like it's checking this still works as expected, but I'm not sure I see what layer is handling the conversion.

rustls/src/server/handy.rs
```diff
@@ -224,14 +223,7 @@ mod sni_resolver {
         /// This function fails if `name` is not a valid DNS name, or if
```
cpu: I think the rustdoc comment needs the "This function fails if `name` is not a valid DNS name" part removed.

rustls/src/server/handy.rs
ctz: I think the answer here is that `Hash` and `Eq` are both case-insensitive for this type. So while "LOCALHOST" is inserted into the hashmap, queries for "localhost", "Localhost", etc all resolve to it.

djc: That's right, we get it for free thanks to using a proper type.

- `PartialEq`: https://github.com/rustls/pki-types/blob/main/src/server_name.rs#L260
- `Hash`: https://github.com/rustls/pki-types/blob/main/src/server_name.rs#L274

rustls/src/server/handy.rs
djc: Right, done.

ctz (approved):

rustls/src/server/handy.rs
cpu: Ah! That makes perfect sense. Thanks

PR #2518, opened 2025-06-24
client: pass all of ClientHelloInput into tls12 handle_server_hello()

djc:
As was already done for 1.3 in

- #2516

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -187,8 +187,6 @@ pub(super) fn start_handshake(
             config,
             resuming,
             random,
-            #[cfg(feature = "tls12")]
-            using_ems: false,
             sent_tls13_fake_ccs: false,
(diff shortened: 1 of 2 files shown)


rustls/src/client/hs.rs
```diff
@@ -225,8 +223,6 @@ pub(super) struct ClientHelloInput {
     pub(super) config: Arc<ClientConfig>,
     pub(super) resuming: Option<persist::Retrieved<ClientSessionValue>>,
     pub(super) random: Random,
-    #[cfg(feature = "tls12")]
-    pub(super) using_ems: bool,
```
djc, line 229: This turned out to be dead in a way that wasn't obvious. The `handle_server_hello()` in `tls12` overwrites this value with `server_hello.extended_master_secret_ack.is_some()`.

PR #2522, opened 2025-06-25
Simplify the simpleserver example by using rustls::Stream

### LINKED ISSUE #2520: Why is simpleserver example not using rustls::Stream?
I just spent about an hour trying to understand the simpleserver example to make a simple TLS stream over TCP:
https://github.com/rustls/rustls/blob/6cc7a45fae74717a1c9bb4733ecdf0f6a8b75cc2/examples/src/bin/simpleserver.rs#L41-L49
because I was thrown off by the fact that we don't pass the `TcpStream` when creating the `ServerConnection`, and that seemed to imply that this example (using `conn.reader()` as my final reader) would not work if the message was too large, because it was unclear to me how it would be able to still refer to the `TcpStream` when doing many calls to `read` on that...

Only to finally figure out that one can actually use `rustls::Stream`, like on the client side, to get a working stream with none of the crazy `read_tls()`/`write_tls()`/`complete_io()` concerns.

My question is, why is that example not using that already?
The current writing seems error-prone: people might not immediately notice that they need to interlace calls to `read` on the `connection.reader()` and calls to `complete_io`, or actually just use `Stream` instead.
I could just as well have not noticed that we don't pass the `TcpStream` when creating the `ServerConnection`...

New proposed writing:
```rust
	let listener = TcpListener::bind(format!("[::]:{}", 4443)).unwrap();
	let (mut tcp_stream, _) = listener.accept()?;
	let mut conn = rustls::ServerConnection::new(Arc::new(config))?;
	let mut tls_stream = rustls::Stream::new(&mut conn, &mut tcp_stream);

	tls_stream.write_all(b"Hello from the server")?;
	let mut buf = [0; 64];
	let len = tls_stream.read(&mut buf)?;
	println!("Received message from client: {:?}", &buf[..len]);
```

@cpu since you authored the relevant file

(I can open a PR if you like)

Ten0:
Resolves #2520

This should hopefully make it easier to understand what the idiomatic way to use rustls in the most basic way (as a wrapper over an existing `impl Read+Write`) is.

cc @cpu

### DIFF
--- a/examples/src/bin/simpleserver.rs
+++ b/examples/src/bin/simpleserver.rs
@@ -36,16 +36,14 @@ fn main() -> Result<(), Box<dyn StdError>> {
         .with_single_cert(certs, private_key)?;
 
     let listener = TcpListener::bind(format!("[::]:{}", 4443)).unwrap();
-    let (mut stream, _) = listener.accept()?;
-
+    let (mut tcp_stream, _) = listener.accept()?;
     let mut conn = rustls::ServerConnection::new(Arc::new(config))?;
-    conn.complete_io(&mut stream)?;
+    let mut tls_stream = rustls::Stream::new(&mut conn, &mut tcp_stream);
 
-    conn.writer()
-        .write_all(b"Hello from the server")?;
-    conn.complete_io(&mut stream)?;
+    tls_stream.write_all(b"Hello from the server")?;
+    tls_stream.flush()?;
     let mut buf = [0; 64];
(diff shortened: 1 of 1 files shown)


ctz (approved):
I think this is OK. I think we don't generally want to encourage `rustls::Stream` use, as it's pretty narrowly applicable and very hard to make a robust program with (especially for a server!) but actually this is mostly a comment about blocking IO and thread-per-client models.

PR #2523, opened 2025-06-25
Enable bogo OCSP stapling tests

ctz:
This enables some pre-existing bogo tests for OCSP stapling, that illustrates we can't raise an error that produces the right alert in this case.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -127,6 +127,7 @@ struct Options {
     on_initial_expect_curve_id: Option<NamedGroup>,
     on_resume_expect_curve_id: Option<NamedGroup>,
     wait_for_debugger: bool,
+    ocsp: OcspValidation,
 }
 
 impl Options {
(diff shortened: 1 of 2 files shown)


bogo/src/main.rs
```diff
@@ -437,6 +441,10 @@ impl ServerCertVerifier for DummyServerAuth {
         _ocsp: &[u8],
         _now: UnixTime,
     ) -> Result<ServerCertVerified, Error> {
+        match self.ocsp {
+            OcspValidation::Reject => return Err(CertificateError::InvalidOcspResponse.into()),
+            OcspValidation::None => {}
+        }
```
cpu: Maybe clearer as:

```rust
if let OcspValidation::Reject = self.ocsp {
    return Err(CertificateError::InvalidOcspResponse.into())
}
```

bogo/src/main.rs
```diff
@@ -465,6 +473,15 @@ impl ServerCertVerifier for DummyServerAuth {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
+enum OcspValidation {
+    /// Totally ignore `ocsp_response` value
+    None,
```
cpu, line 479: WDYT about deriving `Default` and marking `None` as the default variant?

djc (approved):

PR #2524, opened 2025-06-26
client: initial support for preshared keys

djc:
Early feedback welcome!

Still missing: early data support.

Based on the implementation in

- #2424 

cc @elagergren-spideroak


djc:
Have not heard back from the client, closing this for now.

TaeHagen:
Apparently, to talk to quickrelay (for FaceTime), Apple requires TLS-PSK support. I got it working with this branch, I had to make a few tweaks to get it to work, main one is using "ext binder" instead of "res binder" as the HKDF tag.

I'd be willing to fund getting this ported to the latest version of rustls and upstreamed. Alternatively, I could work on porting it myself, however, given it is a substantial changeset and you have opinionated standards (which is fair, it is your project after all), it might just be easier to pay for a few days of your time. Please let me know how you want to proceed.

PR #2525, opened 2025-06-26
bogo: enable client certificate selection tests

ctz:
This enables most `CertificateSelection-Client-*` tests, which demonstrates the extent to which the `ResolvesClientCert` trait is good or not.

This PR also does some minor fiddling in config.json.in, with the eventual goal of making the order of `DisabledTests` irrelevent, and therefore being able to sort the sections of this file.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -230,6 +230,8 @@ impl Options {
 struct Credentials {
     default: Credential,
     additional: Vec<Credential>,
+    /// Some(-1) means `default`, otherwise index into `additional`
+    expect_selected: Option<isize>,
 }
 
 impl Credentials {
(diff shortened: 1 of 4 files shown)


bogo/src/main.rs
```diff
@@ -230,6 +226,50 @@ impl Options {
...
 }
 
+#[derive(Debug, Default)]
+struct Credentials {
+    default: Credential,
+    additional: Vec<Credential>,
+}
+
+impl Credentials {
+    fn most_recent_mut(&mut self) -> &mut Credential {
```
djc: Nit: maybe just call this `last_mut()`? There's no real "recency" involved anyway...

PR #2527, opened 2025-06-27
Further extend number of bogo tests

ctz:
Final part of reviewing bogo `DisabledTests` and enabling easy ones, but documenting the remainder.


bogo/src/main.rs
```diff
@@ -1974,26 +2005,69 @@ pub fn main() {
...
     }
 }
 
+#[derive(Debug)]
+struct KeyLogMemo(Mutex<KeyLogMemoInner>);
+
+impl KeyLogMemo {
+    fn new() -> Self {
+        Self(Mutex::new(KeyLogMemoInner::default()))
+    }
```
djc: Guessing you'll get this for free if you derive `Default`? (Even if you don't, suggest turning this into a manual `Default` impl.)

djc (approved):

PR #2529, opened 2025-06-27
Update 0.24.x from main

djc:


djc:
I guess there's no way to get a man-made merge commit into the main branch via the GitHub PR UI...

PR #2530, opened 2025-06-28
bogo: replace sleep with orderly TCP closure

ctz:
Aside from anything else this takes bogo tests from 4.7s to 1.7s.


bogo/src/main.rs
```diff
@@ -1195,6 +1194,25 @@ fn flush(sess: &mut Connection, conn: &mut net::TcpStream) {
     conn.flush().unwrap();
 }
 
+fn orderly_close(conn: &mut net::TcpStream) {
```
cpu: nit: this file has the same issues elsewhere but I think `orderly_close()` should be positioned after `after_read()`.

PR #2531, opened 2025-06-30
Adopt webpki 0.103.4

djc:
## Proposed release notes

- Improved errors for invalid signature algorithms; added variants to both `CertificateError` and `CertRevocationError` to replace the `UnsupportedSignatureAlgorithm` variant (now deprecated) in order to provide more context.
- Added support for post-quantum ML-DSA signature schemes when compiled with the `aws-lc-rs-unstable` Cargo feature.

Depends on:

- https://github.com/rustls/webpki/pull/361

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1991,7 +1991,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07033963ba89ebaf1584d767badaa2e8fcec21aedea6b8c0346d487d49c28667"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -2865,7 +2865,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.28"
+version = "0.23.29"
 dependencies = [
  "aws-lc-rs",
  "base64",
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -275,7 +275,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.28"
+version = "0.23.29"
 dependencies = [
  "log",
  "once_cell",
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls"
-version = "0.23.28"
+version = "0.23.29"
 edition = "2021"
 rust-version = "1.71"
 license = "Apache-2.0 OR ISC OR MIT"


rustls/src/webpki/mod.rs
```diff
@@ -1,14 +1,13 @@
...
+use std::sync::Arc;
 
 use pki_types::CertificateRevocationListDer;
 use webpki::{CertRevocationList, InvalidNameContext, OwnedCertRevocationList};
 
 use crate::error::{
     CertRevocationListError, CertificateError, Error, ExtendedKeyPurpose, OtherError,
 };
-#[cfg(feature = "std")]
-use crate::sync::Arc;
```
ctz: unnecessary change perhaps?

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -174,6 +174,12 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
         webpki_algs::RSA_PKCS1_2048_8192_SHA256_ABSENT_PARAMS,
         webpki_algs::RSA_PKCS1_2048_8192_SHA384_ABSENT_PARAMS,
         webpki_algs::RSA_PKCS1_2048_8192_SHA512_ABSENT_PARAMS,
+        #[cfg(all(feature = "aws-lc-rs-unstable", not(feature = "fips")))]
```
djc: It occurs to me that the `fips` feature isn't great for additivity, because we're treating it as a feature but in this context it actually shrinks the API surface. Bit of a risk, but might not be much we can do?

djc:
@ctz added a commit for adding ML-DSA and release notes. Suppose that makes this complete and ready for release?

djc:
> @ctz added a commit for adding ML-DSA and release notes. Suppose that makes this complete and ready for release?

Any suggestions on how we should defend the new feature in CI?

rustls/src/crypto/aws_lc_rs/mod.rs
```diff
@@ -222,6 +228,12 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
             SignatureScheme::RSA_PKCS1_SHA256,
             &[webpki_algs::RSA_PKCS1_2048_8192_SHA256],
         ),
+        #[cfg(all(feature = "aws-lc-rs-unstable", not(feature = "fips")))]
```
ctz: I feel like this is -- even behind a feature named "unstable" -- pretty aggressive for the core crate.  WDYT about doing this in the rustls-post-quantum crate (in a new major version?) I realise it is quite a bit more work and wrangling.

djc: Makes sense, will redo!

djc:
* Published rustls v0.23.29 at registry `crates-io`
* [new tag]           v/0.23.29 -> v/0.23.29
* [Release notes](https://github.com/rustls/rustls/releases/tag/v%2F0.23.29)

PR #2528, opened 2025-06-27
Generalize error return from TicketRotator generator fn.

bdaehlie:

PR #2534, opened 2025-06-30
Ticketer lifetime and doc cleanups

### LINKED ISSUE #2533: specification of ticketer lifetimes is confusing
In the aws-lc-rs ticketer, there is code that calculates lifetimes in two places. In the first instance, there is a comment stating that there is a 12 hour lifetime, then the calculation used is "6 * 60 * 60", presumably but not obviously because in the general ticketer.rs lifetime is calculated like this:

fn lifetime(&self) -> u32 {
        self.lifetime * 2
}

That in and of itself is a bit confusing unless you happen to know about that * 2. Makes it seem like the comment and the code are out of sync.

Then, later in the aws-lc-rs ticketer, there is this line:

lifetime: 60 * 60 * 12,

I don't know enough yet to know if there is a problem here (probably not?), but having two different ways of specifying these lifetimes, one presumed to be doubled and one not, seems error-prone.

ctz:
(for after #2528)

fixes #2533

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/ticketer.rs
+++ b/rustls/src/crypto/aws_lc_rs/ticketer.rs
@@ -39,7 +39,7 @@ impl Ticketer {
     #[cfg(feature = "std")]
     pub fn new() -> Result<Arc<dyn ProducesTickets>, Error> {
         Ok(Arc::new(crate::ticketer::TicketRotator::new(
-            6 * 60 * 60,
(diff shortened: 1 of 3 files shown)


djc (approved):

ctz:
> (for after #2528)

now incorporates this commit

ctz:
first commit was reviewed  by me in #2528, the remainder reviewed here by djc

Issue #945, opened 2022-01-19
`export_keying_material()` API doesn't conform with TLS 1.3 recommendations


briansmith:
The TLS 1.3 spec says in https://datatracker.ietf.org/doc/html/rfc8446#appendix-E.1.4:
> [B]ecause these secrets can be used to compute any
   exporter value, they SHOULD be erased as soon as possible.  If the
   total set of exporter labels is known, then implementations SHOULD
   pre-compute the inner Derive-Secret stage of the exporter computation
   for all those labels, then erase the [early_]exporter_master_secret,
   followed by each inner value as soon as it is known that it will not
   be needed again.

However, the design of the `export_keying_material()` API makes this erasure recommendation impossible to implement, because we currently support calling `export_keying_material()` at any time after the handshake completes, and so we have to retain the exporter key.

Note that TLS doesn't provide any way to rotate the exporter key; the `KeyUpdate` mechanism only applies to traffic keys. This design pretty much forces us to implement the erasure recommendation above, if we want to have long-lived connections that minimize their ability to expose secrets.

I propose:
* We remove `export_keying_material()` from the public API.
* Add a "handshake" callback. (Note that this suggested interface is intended to be easily extensible in a way that maintains SemVer compatibility):

```rust
pub trait HandshakeComplete {
     /// Called just after the the peer's handshake messages and identity (if any) have been successfully validated.
     ///
     ///
      pub fn handshake_complete(&self, handshake_data: HandshakeData) -> Result<(), Box<dyn Debug>>
}

pub struct HandshakeData {
     /// <The current export_keying_material API>
     fn export_keying_material(&mut self, ....) -> ....;
```

* When the handshake completes, call this callback.
* Move `current_exporter_secret` out of `KeyScheduleTraffic` to the implementation(s) of `HandshakeData`. (It should probably be lazily constructed the first time `export_keying_material` is called, since most applications don't need it.)
* The new `export_keying_material()` API should implement additional validation of its parameters as suggested in the TLS 1.3 specification and [the and common sense (e.g. that the label and other parameters that should be non-empty aren't empty).](https://datatracker.ietf.org/doc/html/rfc5705#section-4).
* Since `HandshakeData` is moved into the callback, its contents (in particular, the exporter key) will be destroyed automatically, implementing the semantics that the TLS 1.3 spec recommends.

Besides being safer, this would also:
* Slightly reduce the memory requirements for each connection
[...]

PR #2540, opened 2025-07-02
Do not retain master secret during terminal key schedule state

ctz:
This PR is a step toward #945, but for now arranges that the TLS1.3 "master secret" is erased in the terminal state. Doing that means that secrets derived from that are impossible to rederive later.

Unconditionally deriving the resumption secret has some performance consequences:

- for clients connected to servers that won't offer resumption, it is wasted work.
- for clients that receive several tickets, it is a saving for subsequent tickets.
- for servers this doesn't alter anything (see the optimisation landed in 17518c386 previously)

### DIFF
--- a/rustls/src/tls13/key_schedule.rs
+++ b/rustls/src/tls13/key_schedule.rs
@@ -12,58 +12,6 @@ use crate::msgs::message::Message;
 use crate::suites::PartiallyExtractedSecrets;
 use crate::{ConnectionTrafficSecrets, KeyLog, Tls13CipherSuite, quic};
 
-/// The kinds of secret we can extract from `KeySchedule`.
-#[derive(Debug, Clone, Copy, PartialEq)]
-enum SecretKind {
-    ResumptionPskBinderKey,
-    ClientEarlyTrafficSecret,
-    ClientHandshakeTrafficSecret,
-    ServerHandshakeTrafficSecret,
-    ClientApplicationTrafficSecret,
-    ServerApplicationTrafficSecret,
-    ExporterMasterSecret,
-    ResumptionMasterSecret,
-    DerivedSecret,
-    ServerEchConfirmationSecret,
-    ServerEchHrrConfirmationSecret,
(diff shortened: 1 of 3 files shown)


rustls/src/tls13/key_schedule.rs
```diff
@@ -343,49 +344,118 @@ impl KeyScheduleHandshake {
...
+            current_server_traffic_secret,
+            current_exporter_secret,
+            resumption_master_secret,
+        }
+    }
+}
+
+/// Client-side key schedule before the finished message is sent.
+///
+/// This differs from `KeyScheduleBeforeFinished`
```
djc: Uh, did you forget to finish this docstring?

rustls/src/tls13/key_schedule.rs
```diff
@@ -745,6 +714,55 @@ impl KeySchedule {
...
     }
+}
+
+impl Deref for KeySchedule {
+    type Target = KeyScheduleSuite;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
```
djc, line 741: How much does it cost to not have this?

djc (approved):
Makes me wonder (again): should we have a separate types for an in-handshake connection, vs the long-term traffic connection, or would that affect the ergonomics too much? Maybe there should be an inner `TrafficConnection` type that you can turn a `ClientConnection` or `ServerConnection` into that has minimal state?

rustls/src/tls13/key_schedule.rs
ctz: good catch!

rustls/src/tls13/key_schedule.rs
ctz: added a temporary commit showing this. my feeling is it looks better with the deref. wdyt?

djc: Agreed!

PR #2541, opened 2025-07-03
Excise `read_buf` support and its crate feature

ctz:
[It seems this doesn't have a clear future](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Future.20of.20read_buf/with/526874872) so dropping it in 0.24 seems reasonable. I'm generally open to solving the same problem in a different way in the future.


cpu (approved):
The [clippy finding in CI](https://github.com/rustls/rustls/actions/runs/16043956473/job/45270848881?pr=2541) is odd and doesn't seem obviously related to this diff, did the dev-0.24 branch protection rules let it slip in?

cpu:
This'll result in some similar tidying in [rustls-ffi](https://github.com/rustls/rustls-ffi/blob/f0ef08dfde7bd9b1a20b8b8325b4b18c7a1fd7bc/librustls/src/rustls.h#L1890-L1910) when 0.24 is ready :+1:

djc:
> The [clippy finding in CI](https://github.com/rustls/rustls/actions/runs/16043956473/job/45270848881?pr=2541) is odd and doesn't seem obviously related to this diff, did the dev-0.24 branch protection rules let it slip in?

I think I broke the CI enforcement of the branch protection a while back when I migrated us to the rulesets stuff. I'll try to fix.

djc:
> > The [clippy finding in CI](https://github.com/rustls/rustls/actions/runs/16043956473/job/45270848881?pr=2541) is odd and doesn't seem obviously related to this diff, did the dev-0.24 branch protection rules let it slip in?
> 
> I think I broke the CI enforcement of the branch protection a while back when I migrated us to the rulesets stuff. I'll try to fix.

Should be better now.

PR #2545, opened 2025-07-04
Correct calculation of ServerHello ECH confirmation

ctz:
This was another case where we were relying on the ability to round-trip received messages.  That assumption as abandoned in 692f981446. This meant that we calculated the ECH confirmation based on a different extension ordering than the one the server sent, which meant it didn't match.

bogo and the cloudflare test server continued to work, as they just happened to choose the same extension ordering as us.  But OpenSSL didn't.

### DIFF
--- a/rustls/src/client/ech.rs
+++ b/rustls/src/client/ech.rs
@@ -449,6 +449,7 @@ impl EchState {
         self,
         ks: &mut KeyScheduleHandshakeStart,
         server_hello: &ServerHelloPayload,
+        server_hello_encoded: &Payload<'_>,
         hash: &'static dyn Hash,
     ) -> Result<Option<EchAccepted>, Error> {
         // Start the inner transcript hash now that we know the hash algorithm to use.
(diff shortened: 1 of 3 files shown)


ctz:
https://github.com/rustls/rustls/actions/runs/16075349921/job/45369135745

cpu (approved):
Nice catch

rustls/src/client/ech.rs
```diff
@@ -785,6 +807,16 @@ impl EchState {
...
 
+/// The last eight bytes of the ServerHello's random, taken from a Handshake message containing it.
+///
+/// This has:
+/// - a HandshakeType (1 byte),
+/// - an exterior length (3 bytes),
+/// - the legacy_version (2 bytes), and
+/// - the balance of the random field (24 bytes).
+const SERVER_HELLO_ECH_CONFIRMATION_SPAN: core::ops::Range<usize> =
+    (1 + 3 + 2 + 24)..(1 + 3 + 2 + 32);
```
cpu, line 818: this feels a little bit brittle, but I'm not sure I have concrete suggestions on how to make it better :-/

ctz: i've added a small test for this

PR #2547, opened 2025-07-10
Refactor `ServerSessionValue`

ctz:
The goal here is to use a tighter type for TLS1.2 master secrets, which pays off in the next PR.

This is technically a breaking change (and the following PR definitely is), so it is against dev-0.24.


rustls/src/msgs/persist.rs
```diff
@@ -343,46 +343,42 @@ impl Codec<'_> for ServerSessionValue {
...
-                Err(_) => return Err(InvalidMessage::InvalidServerName),
-            };
-
-            Some(dns_name)
-        } else {
-            None
+        let sni = match u8::read(r)? {
+            1 => {
+                let dns_name = PayloadU8::<MaybeEmpty>::read(r)?;
+                let dns_name = match DnsName::try_from(dns_name.0.as_slice()) {
```
djc, line 515: Nit: inline the last `dns_name` binding? (`Some(match .. { .. })`)

rustls/src/msgs/persist.rs
```diff
@@ -297,21 +297,205 @@ static MAX_FRESHNESS_SKEW_MS: u32 = 60 * 1000;
...
+#[derive(Debug)]
+pub struct Tls12ServerSessionValue {
+    #[doc(hidden)]
+    pub common: CommonServerSessionValue,
     pub(crate) master_secret: Zeroizing<PayloadU8>,
     pub(crate) extended_ms: bool,
+}
+
+impl Tls12ServerSessionValue {
+    pub(crate) fn new(
```
djc, line 340: Seems like `Tls12ServerSessionValue::new()` and `Tls13ServerSessionValue::new()` are pretty crappy as far as abstractions go, consider dropping them (assuming they don't have many callsites)?

rustls/src/msgs/persist.rs
```diff
@@ -362,14 +365,18 @@ impl Tls12ServerSessionValue {
...
     }
 
     fn read(r: &mut Reader<'_>) -> Result<Self, InvalidMessage> {
         Ok(Self {
             common: CommonServerSessionValue::read(r)?,
-            master_secret: Zeroizing::new(PayloadU8::read(r)?),
+            master_secret: Zeroizing::new(
+                r.take(48)
+                    .and_then(|slice| slice.try_into().ok())
+                    .ok_or(InvalidMessage::MessageTooShort)?,
```
djc: Don't love `and_then().ok_or()`, I think `match` might be clearer even if slightly more verbose?

djc (approved):
Nice!

cpu (approved):

rustls/src/msgs/persist.rs
ctz: Yeah they're kinda crappy, but inlining them would leak quite of lot of extra imports and implementation details into already quite busy files (`server/tls13.rs` and co.)

ctz: Maybe a better idea is to use the `Common*` type constructor as well. I'll try that and see how it looks.

rustls/src/msgs/persist.rs
```diff
@@ -343,46 +343,42 @@ impl Codec<'_> for ServerSessionValue {
...
         let age_obfuscation_offset = u32::read(r)?;
 
         Ok(Self {
             sni,
-            version: v,
-            cipher_suite: cs,
-            master_secret: ms,
-            extended_ms: ems == 1u8,
-            client_cert_chain: ccert,
+            version,
```
djc: Could also consider inlining all/most of these directly into the `Self` initializer (maybe in a separate commit).

djc (commented):
(Sorry for the late feedback, something to think about.)

rustls/src/msgs/persist.rs
djc: Yeah, that's what I was thinking.

PR #2548, opened 2025-07-10
Optimize HMAC key preparation for TLS1.2 PRF

ctz:
Coming after #2547 this is a performance optimisation for TLS1.2 (formerly #2256). It doesn't apply to aws-lc-rs but will apply to most other providers.

re #2119

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/tls12.rs
+++ b/rustls/src/crypto/aws_lc_rs/tls12.rs
@@ -1,13 +1,14 @@
 use alloc::boxed::Box;
 
 use aws_lc_rs::{aead, tls_prf};
+use zeroize::Zeroizing;
 
 use crate::crypto::cipher::{
     AeadKey, InboundOpaqueMessage, Iv, KeyBlockShape, MessageDecrypter, MessageEncrypter,
     NONCE_LEN, Nonce, Tls12AeadAlgorithm, UnsupportedOperationError, make_tls12_aad,
 };
-use crate::crypto::tls12::Prf;
-use crate::crypto::{ActiveKeyExchange, KeyExchangeAlgorithm};
+use crate::crypto::tls12::{Prf, PrfSecret};
+use crate::crypto::{ActiveKeyExchange, KeyExchangeAlgorithm, SharedSecret};
 use crate::enums::{CipherSuite, SignatureScheme};
 use crate::error::Error;
 use crate::msgs::fragmenter::MAX_FRAGMENT_LEN;
@@ -428,19 +429,6 @@ fn gcm_iv(write_iv: &[u8], explicit: &[u8]) -> Iv {
 struct Tls12Prf(&'static tls_prf::Algorithm);
 
 impl Prf for Tls12Prf {
-    fn for_secret(&self, output: &mut [u8], secret: &[u8], label: &[u8], seed: &[u8]) {
-        // safety:
-        // - [1] is safe because our caller guarantees `secret` is non-empty; this is
-        //   the only documented error case.
-        // - [2] is safe in practice because the only failure from `derive()` is due
-        //   to zero `output.len()`; this is outlawed at higher levels
-        let derived = tls_prf::Secret::new(self.0, secret)
-            .unwrap() // [1]
-            .derive(label, seed, output.len())
-            .unwrap(); // [2]
-        output.copy_from_slice(derived.as_ref());
-    }
-
     fn for_key_exchange(
         &self,
         output: &mut [u8; 48],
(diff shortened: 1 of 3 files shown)


rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -449,17 +437,47 @@ impl Prf for Tls12Prf {
...
+    }
+
     fn fips(&self) -> bool {
         super::fips()
     }
 }
+
+// nb: we can't put a `tls_prf::Secret` in here because it is
+// consumed by `tls_prf::Secret::derive()`
+struct Tls12PrfSecret(&'static tls_prf::Algorithm, Zeroizing<[u8; 48]>);
```
djc: Suggest naming the fields.

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -449,17 +437,47 @@ impl Prf for Tls12Prf {
...
                 .secret_bytes(),
-            label,
-            seed,
-        );
+        )
+        .prf(output, label, seed);
         Ok(())
     }
 
+    fn new_secret(&self, secret: &[u8; 48]) -> Box<dyn PrfSecret> {
```
djc, line 448: Suggest taking an owned array if we're going to store it -- especially for secrets? (Up the chain as relevant.)

ctz: Ah, I was taking a reference here specifically because it is a secret -- to avoid stack-to-stack moves leaving copies everywhere? But I'm not sure 100% a) whether that actually happens in release mode, b) whether taking a reference does actually avoid that.

djc: That's fair -- kinda interesting that the "correct" Rust semantics are kind of opposed to best practices from a machine/assembly implementation point of view. Feel free to ignore my comment.

cpu:
> It doesn't apply to aws-lc-rs but will apply to most other providers.

Can you save me looking deeper and say why? :-)

ctz:
> > It doesn't apply to aws-lc-rs but will apply to most other providers.
> 
> Can you save me looking deeper and say why? :-)

Since for FIPS the TLS-PRF is a separate approved algorithm, so we use that rather than building it from whole cloth using HMAC.

Unfortunately the aws-lc-rs APIs for that (both the rust-level and lower-level) aren't shaped to allow for this optimisation (which fundamentally to do the HMAC key processing once-per-connection rather than once-per-secret -- there are 2-3 such secrets per connection). The references for that are [here](https://github.com/aws/aws-lc-rs/blob/a2ed3357324b31c542a24462c2f98a282ec3d970/aws-lc-rs/src/tls_prf.rs#L66) where the underlying object is constructed, [here](https://github.com/aws/aws-lc-rs/blob/a2ed3357324b31c542a24462c2f98a282ec3d970/aws-lc-rs/src/tls_prf.rs#L81) where the PRF is run (which consumes the object), and [here](https://github.com/aws/aws-lc/blob/04a6cf021aa5c2e089d593967e5999d0b69f93cf/crypto/fipsmodule/tls/kdf.c#L137) which is the lower-level function -- observe it takes the whole secret as a each time.

I haven't raised an issue with aws-lc/aws-lc-rs for this because it's a kinda minor benefit, and the change would be in the fips module and would probably be quite grim.

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -449,17 +437,61 @@ impl Prf for Tls12Prf {
...
+            alg: self.alg,
+            secret: self.secret.as_ref(),
+        }
+        .prf(output, label, seed)
+    }
+}
+
+struct Tls12PrfSecretBorrow<'a> {
+    alg: &'static tls_prf::Algorithm,
+    secret: &'a [u8],
```
djc: Could this be `&'a [u8; 48]`?

rustls/src/crypto/aws_lc_rs/tls12.rs
```diff
@@ -449,17 +437,61 @@ impl Prf for Tls12Prf {
...
+        //   the only documented error case.
+        // - [2] is safe in practice because the only failure from `derive()` is due
+        //   to zero `output.len()`; this is outlawed at higher levels
+        let derived = tls_prf::Secret::new(self.alg, self.secret)
+            .unwrap() // [1]
+            .derive(label, seed, output.len())
+            .unwrap(); // [2]
+        output.copy_from_slice(derived.as_ref());
+    }
+}
```
djc: Maybe this would be easier to understand/less duplicative if you introduced an `enum` to wrap over `&'a [u8]` and `Zeroizing<[u8; 48]>`?

djc (approved):

rustls/src/crypto/aws_lc_rs/tls12.rs
ctz: Added an enum over the master secret and key exchange case. I think this is a lot cleaner!

rustls/src/crypto/aws_lc_rs/tls12.rs
ctz: No, as the key exchange case is variable length. But it could move the `SharedSecret` into here.

PR #2551, opened 2025-07-10
client: touchup `WriteEarlyData` rustdoc

cpu:
Quick fix for a docs nit I noticed while looking up some API details.

Model the rustdoc for this after what we say for the server equivalent [`ReadEarlyData` rustdoc](https://docs.rs/rustls/latest/rustls/server/struct.ReadEarlyData.html).

The previous text refers to `write_early_data()`, but that's an internal fn and not helpful to API doc readers.


rustls/src/client/client_conn.rs
```diff
@@ -651,7 +651,11 @@ mod connection {
     use crate::suites::ExtractedSecrets;
     use crate::sync::Arc;
 
-    /// Stub that implements io::Write and dispatches to `write_early_data`.
+    /// Allows writing of early data in resumed TLS 1.3 connections
```
djc: Tiny nit: I think convention in this crate is to end initial sentence with a period?

rustls/src/client/client_conn.rs
```diff
@@ -651,7 +651,11 @@ mod connection {
...
+    ///
+    /// "Early data" is also known as "0-RTT data".
+    ///
+    /// This structure implements [`std::io::Write`].
```
djc: Another tiny nit: I think "structure" is a weird fit here, maybe "type"?

rustls/src/client/client_conn.rs
cpu: Fixed.

rustls/src/client/client_conn.rs
cpu: Added a commit fixing this in the pre-existing `ReadEarlyData` struct docs and then matched it in the new `WriteEarlyData` docs.

PR #2553, opened 2025-07-11
crypto: require matching keys for CertifiedKey construction

djc:
One more part of #2119.

### DIFF
--- a/rustls/src/crypto/signer.rs
+++ b/rustls/src/crypto/signer.rs
@@ -135,6 +135,7 @@ impl ResolvesServerCert for SingleCertAndKey {
 /// certificates.
 ///
 /// [RFC 7250]: https://tools.ietf.org/html/rfc7250
+#[non_exhaustive]
 #[derive(Clone, Debug)]
 pub struct CertifiedKey {
     /// The certificate chain or raw public key.
--- a/rustls/src/crypto/signer.rs
+++ b/rustls/src/crypto/signer.rs
@@ -139,7 +139,7 @@ impl ResolvesServerCert for SingleCertAndKey {
 #[derive(Clone, Debug)]
 pub struct CertifiedKey {
     /// The certificate chain or raw public key.
-    pub cert: Vec<CertificateDer<'static>>,
+    pub cert_chain: Vec<CertificateDer<'static>>,
 
     /// The certified key.
     pub key: Arc<dyn SigningKey>,
@@ -178,9 +178,9 @@ impl CertifiedKey {
     ///
     /// The cert chain must not be empty. The first certificate in the chain
     /// must be the end-entity certificate.
-    pub fn new(cert: Vec<CertificateDer<'static>>, key: Arc<dyn SigningKey>) -> Self {
+    pub fn new(cert_chain: Vec<CertificateDer<'static>>, key: Arc<dyn SigningKey>) -> Self {
         Self {
-            cert,
+            cert_chain,
             key,
             ocsp: None,
         }
@@ -202,7 +202,7 @@ impl CertifiedKey {
 
     /// The end-entity certificate.
     pub fn end_entity_cert(&self) -> Result<&CertificateDer<'_>, Error> {
-        self.cert
+        self.cert_chain
             .first()
             .ok_or(Error::NoCertificatesPresented)
     }
--- a/rustls/src/crypto/signer.rs
+++ b/rustls/src/crypto/signer.rs
@@ -186,6 +186,24 @@ impl CertifiedKey {
         }
     }
 
+    /// Make a new `CertifiedKey` from a raw private key.
+    ///
+    /// Unlike [`CertifiedKey::new()`], this does not check that the end-entity certificate's
+    /// subject key matches `key`'s public key.
+    ///
+    /// This avoids parsing the end-entity certificate, which is useful when using client
+    /// certificates that are not fully standards compliant, but known to usable by the peer.
+    pub fn new_unchecked(
+        cert_chain: Vec<CertificateDer<'static>>,
+        key: Arc<dyn SigningKey>,
+    ) -> Self {
+        Self {
+            cert_chain,
+            key,
+            ocsp: None,
+        }
+    }
+
     /// Verify the consistency of this [`CertifiedKey`]'s public and private keys.
     /// This is done by performing a comparison of SubjectPublicKeyInfo bytes.
     pub fn keys_match(&self) -> Result<(), Error> {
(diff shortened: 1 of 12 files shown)


ctz:
Haven't reviewed this in detail, but I think we need to continue avoiding:

- requiring integrators of secure elements, key store APIs, etc. to be able to output public keys for existing private ones, and format those public keys in a specific format. that is especially true when it comes to PKCS#11 which represents the relationship between private keys and their certificates in the API, so us doing that separately is a duplication of effort.
- requiring that every certificate is parsable by the webpki crate -- so if we have a checking `new()` we'd need a `new_unchecked()`

rustls/src/error.rs
```diff
@@ -1001,8 +975,8 @@ impl fmt::Display for Error {
             Self::BadMaxFragmentSize => {
                 write!(f, "the supplied max_fragment_size was too small or large")
             }
-            Self::InconsistentKeys(why) => {
-                write!(f, "keys may not be consistent: {why:?}")
+            Self::InconsistentKeys => {
+                write!(f, "public key does not match private key")
```
ctz:
```suggestion
                write!(f, "public key in certificate does not match private key")
```

rustls/src/crypto/signer.rs
```diff
@@ -67,7 +68,7 @@ pub trait SigningKey: Debug + Send + Sync {
     ///
     /// If an implementation does not have the ability to derive this,
     /// it can return `None`.
-    fn public_key(&self) -> Option<SubjectPublicKeyInfoDer<'_>>;
+    fn public_key(&self) -> SubjectPublicKeyInfoDer<'_>;
```
ctz: Obsolete rustdoc comment above this.

djc:
Would be good to get these things mentioned/described in comments at least:

> * requiring integrators of secure elements, key store APIs, etc. to be able to output public keys for existing private ones, and format those public keys in a specific format. that is especially true when it comes to PKCS#11 which represents the relationship between private keys and their certificates in the API, so us doing that separately is a duplication of effort.

What specifically are the constraints here? Does this mean we can't even make the change to have `SigningKey::public_key()` return a non-optional value? What constraints does the TLS protocol conceptually impose on `CertifiedKey` values? Does the protocol not depend on getting the SPKI out of the certificates anyway?

> * requiring that every certificate is parsable by the webpki crate -- so if we have a checking `new()` we'd need a `new_unchecked()`

Why not?

djc:
Let's do

- #2554 

first.

ctz:
> Would be good to get these things mentioned/described in comments at least:
> 
> > * requiring integrators of secure elements, key store APIs, etc. to be able to output public keys for existing private ones, and format those public keys in a specific format. that is especially true when it comes to PKCS#11 which represents the relationship between private keys and their certificates in the API, so us doing that separately is a duplication of effort.
> 
> What specifically are the constraints here? Does this mean we can't even make the change to have `SigningKey::public_key()` return a non-optional value? 

To be concrete, let's talk about two APIs: CNG and PKCS#11.

CNG: for that, one would start by selecting a certificate and then call [`CryptAcquireCertificatePrivateKey`](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptacquirecertificateprivatekey) with `CRYPT_ACQUIRE_COMPARE_KEY_FLAG`. If that call succeeds, it is already guaranteed that the private key and certificate match. So I think this is a case where returning `None` from `public_key()` would be justified -- doing the work again is duplicative.

PKCS#11: if you wanted to find a certificate and its matching key, you would search by `CKA_ID` which has the same value for certificates, private keys and public keys corresponding to the same key pair. If you were to do that, there's not a good reason to then do the same work again to match up private keys and certificates.

But let's ignore that for a minute. `public_key()` would be implementable by reading the `CKA_PUBLIC_KEY_INFO` attribute from the private key. But that attribute is optional:

> DER-encoding of the SubjectPublicKeyInfo for the associated public key (MAY be empty; DEFAULT derived from the underlying private key data; MAY be manually set for specific key types; if set; MUST be consistent with the underlying private key data)

That means a token whose private keys has `CKA_PUBLIC_KEY_INFO` empty would be a bit stuck and unable to use those tokens with rustls.

> What constraints does the TLS protocol conceptually impose on `CertifiedKey` values?
> Does the protocol not depend on getting the SPKI out of the certificates anyway?

It requires verifying signatures with the public key contained in a certificate, but that itself doesn't involve extracting it as a separate quantity, and of course that never needs a private key to produce its SPKI either.

> > * requiring that every certificate is parsable by the webpki crate -- so if we have a checking `new()` we'd need a `new_unchecked()`
> 
> Why not?

We have had users in the past (https://github.com/libp2p/rust-libp2p/issues/5487) who benefited from an escape hatch for this. But more generally I think the tight policy we have on the webpki crate is more defensible if it can be entirely removed from rustls when users require that.

djc:
Okay, so relative to `main`, instead of the current approach:

- Make `CertifiedKey` non-exhaustive
- Rename `CertifiedKey::cert` to `cert_chain`
- Introduce `new_unchecked()` and use it for raw keys etc
- Drop `keys_match()` and make `new()` fallible if keys inconsistent -- should this error or succeed on unknown?

What do you think?

ctz:
I think that sounds good.

> should this error or succeed on unknown?

I think erroring is fine; can point people toward `new_unchecked()` in the rustdoc?

rustls/src/crypto/signer.rs
```diff
@@ -186,6 +186,21 @@ impl CertifiedKey {
         }
     }
 
+    /// Make a new `CertifiedKey` from a raw private key.
+    ///
+    /// Unlike [`CertifiedKey::new()`], this does not check that the end-entity certificate's
+    /// subject key matches `key`'s public key.
+    pub fn new_unchecked(
```
cpu: I think this could maybe use a mention that it avoids parsing the certificate :thinking: 

IIRC that's the main reason people reach for this escape hatch: not so much because they don't care about the correct cert/private key matching, but because their certificate is busted up and has something we'll reject in the process of trying to make that determination.

WDYT about:

```
    /// Make a new `CertifiedKey` from a raw private key.
    ///
    /// Unlike [`CertifiedKey::new()`], this does not check that the end-entity certificate's
    /// subject key matches `key`'s public key.
    ///
    /// This avoids Rustls parsing the end-entity certificate, which can avoid parsing errors for
    /// client certificates that are not standards compliant, but are otherwise usable for the TLS
    /// handshake.
```

cpu (approved):
LGTM. I had one suggested doc update to consider.

rustls/src/crypto/signer.rs
djc: Good suggestion -- tightened up the language a bit more.

PR #2558, opened 2025-07-13
Clarify the ambiguous process-level CryptoProvider error

JonathanBrouwer:
I have personally hit this error after adding a new dependency that enabled the `ring` feature without me realizing.
I think the instruction of the error, to call `CryptoProvider::install_default()`, does not help the user understand why this error suddenly appeared.

Yesterday I saw this [reddit thread](https://www.reddit.com/r/rust/comments/1lyekqv/rustls_ring_and_awslcrs/) hit the same problem, so lets fix this. I'm open to different wording than I came up with, I'm not an expert on this crate.

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -246,7 +246,11 @@ impl CryptoProvider {
         }
 
         let provider = Self::from_crate_features()
-            .expect("no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point");
+            .expect(r###"
+Could not automatically determine the process-level CryptoProvider from Rustls crate features.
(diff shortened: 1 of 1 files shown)


JonathanBrouwer:
Sounds good, let me know if there's anything I need to do to make that happen. Should I rebase on that branch?

ctz (approved):
Thanks!

cpu (approved):
Thanks :+1:

cpu:
> Sounds good, let me know if there's anything I need to do to make that happen. Should I rebase on that branch?

Thanks for the offer. I went ahead and opened the backport PR to make life easy: https://github.com/rustls/rustls/pull/2561

PR #2559, opened 2025-07-14
Don't propagate WouldBlock errors from complete_io()

### LINKED ISSUE #2556: Potential deadlock when both client and server use `rustls::Stream` on nonblocking TCP
**Checklist**

**Describe the bug**
The application I have here boils down to a simple client-server pair, where each party is sending a large amount of data to the other, while simultaneously reading the data from the other party.

The underlying sockets have been set to non-blocking. There is an I/O thread that spins and keeps trying to read from/write to a `rustls::Stream`, with proper buffers on the application side (so it is always accepting newly read data). The issue is that after a bit, the app deadlocks for both parties.

After a bit of investigation I think the issue is that, when the `rustls::Stream` tries to `read`, it first tries to 'complete prior IO'. This involves writing any completed TLS records to the underlying socket. Unfortunately, if the underlying socket is full and won't accept new data (i.e. WouldBlock), 'write TLS' just fails, so 'complete prior IO' just fails.
Then the stream *never attempts to read from the underlying socket at all* but instead directly returns `WouldBlock`. As a result, if the TCP stream is full on both directions, neither party is reading from the TCP stream (because neither party can successfully 'write TLS'), and the TCP stream never gets emptied, so no more progress is made.

**Applicable Version(s)**
Tested on 0.22.4 and 0.23.29

djc:
Fixes #2556.

cc @zhaowenlan1779

### DIFF
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -552,9 +552,8 @@ impl<Data> ConnectionCommon<Data> {
     ///   once.
     ///
     /// The return value is the number of bytes read from and written
-    /// to `io`, respectively.
-    ///
-    /// This function will block if `io` blocks.
+    /// to `io`, respectively. Once both `read()` and `write()` yield `WouldBlock`,
+    /// this function will propagate the error.
     ///
     /// Errors from TLS record handling (i.e., from [`process_new_packets`])
     /// are wrapped in an `io::ErrorKind::InvalidData`-kind error.
@@ -572,8 +571,8 @@ impl<Data> ConnectionCommon<Data> {
         let mut eof = false;
         let mut wrlen = 0;
         let mut rdlen = 0;
-
         loop {
+            let (mut blocked_write, mut blocked_read) = (None, None);
             let until_handshaked = self.is_handshaking();
 
             if !self.wants_write() && !self.wants_read() {
@@ -582,15 +581,22 @@ impl<Data> ConnectionCommon<Data> {
             }
 
             while self.wants_write() {
-                match self.write_tls(io)? {
-                    0 => {
+                match self.write_tls(io) {
+                    Ok(0) => {
                         io.flush()?;
                         return Ok((rdlen, wrlen)); // EOF.
                     }
-                    n => wrlen += n,
+                    Ok(n) => wrlen += n,
+                    Err(err) if err.kind() == io::ErrorKind::WouldBlock => {
+                        blocked_write = Some(err);
+                        break;
+                    }
+                    Err(err) => return Err(err),
                 }
             }
-            io.flush()?;
+            if wrlen > 0 {
+                io.flush()?;
+            }
 
             if !until_handshaked && wrlen > 0 {
                 return Ok((rdlen, wrlen));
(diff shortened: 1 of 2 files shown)


ctz (approved):
I think the doc comment on `Stream` and `StreamOwned` should explain the behaviour if the underlying stream is non-blocking.

At the moment `StreamOwned` says the stream should be blocking, but if we think it can work and are willing to test it, that is something we could remove?

djc:
> I think the doc comment on `Stream` and `StreamOwned` should explain the behaviour if the underlying stream is non-blocking.

Instead of copy-pasting a bunch of stuff, added a pointer to `ConnectionCommon::complete_io()`, which has quite a bit of detail on the expected semantics.

> At the moment `StreamOwned` says the stream should be blocking, but if we think it can work and are willing to test it, that is something we could remove?

I think the proposed behavior is reasonable. Do you have any thoughts on what aspects of this you'd like to see tested?

(Changed base branch to `rel-0.23`.)

djc:
I guess this might be unobservable in practice, but if we want to be careful we can just leave it out of 0.23.x.

```
--- failure method_requires_different_generic_type_params: method now requires a different number of generic type parameters ---

Description:
A method now requires a different number of generic type parameters than it used to. Uses of this method that supplied the previous number of generic types will be broken.
        ref: https://doc.rust-lang.org/reference/items/generics.html
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.42.0/src/lints/method_requires_different_generic_type_params.ron

Failed in:
  rustls::ConnectionCommon::complete_io takes 0 generic types instead of 1, in /home/runner/work/rustls/rustls/rustls/src/conn.rs:601
  rustls::Connection::complete_io takes 0 generic types instead of 1, in /home/runner/work/rustls/rustls/rustls/src/conn.rs:112
```

djc:
> Unless we plan to 100%-without-fail obviate this on main in another way (eg, by deleting `complete_io`) I suggest this should be targeted at main first, then cherry-picked afterwards. Otherwise `main` will be behind a backport branch, which is a recipe for future regressions.

Ahh, see, I was brought up in the old school of VCS management where we try to commit to the oldest relevant branch first, which means you can merge forward the complete old branch into the new one. I guess that doesn't work so well with the GitHub PR workflow...

rustls/src/conn.rs
```diff
@@ -636,11 +639,13 @@ impl<Data> ConnectionCommon<Data> {
...
 
-            match (eof, until_handshaked, self.is_handshaking()) {
-                (_, true, false) => return Ok((rdlen, wrlen)),
-                (_, false, _) => return Ok((rdlen, wrlen)),
-                (true, true, true) => return Err(io::Error::from(io::ErrorKind::UnexpectedEof)),
-                (..) => {}
+            let blocked = blocked_write.zip(blocked_read);
+            match (eof, until_handshaked, self.is_handshaking(), blocked) {
+                (_, true, false, _) => return Ok((rdlen, wrlen)),
+                (_, false, _, _) => return Ok((rdlen, wrlen)),
```
thalesfragoso, line 645: I think we should check the values of `rdlen`,  `wrlen` and `blocked` here. As in, if `rdlen == wrlen == 0` and `blocked.is_some()` then we should return `Err(WouldBlock)` instead.

PR #2560, opened 2025-07-14
Draft: Generalize API for extensible non-X.509 support

rrauch:
### What

This draft PR introduces an approach for cleaner, more extensible support of non-X.509 identity schemes, such as Raw Public Keys. The focus is on a long-term design that better represents the realities of modern TLS. It also solves issue #2258.

### Why

Historically, TLS and X.509 certificates were tightly coupled, both by convention and implementation. However, the TLS protocol has been evolving towards a more flexible design. RFC 7250 added support for `Raw Public Keys` (RPK) to TLS 1.2, and RFC 8446 (TLS 1.3) went further by making `certificate_type` negotiation a standard part of the handshake.

The specification is extensible, currently allowing:
*   X.509 (certificate type id `0`)
*   RPK (certificate type id `2`)
*   Custom/private schemes (certificate type id range `224-255`)

*(Note: OpenPGP, which used certificate type id `19`, was specified in RFC 6091, but its support was removed in RFC 8446 due to a lack of adoption.)*

Rustls' initial RPK support (PR #2062) minimized changes to the public API by reusing existing types (`CertificateDer` to hold RPKs) and traits (`CertVerifier`). This required adding specialized methods like `requires_raw_public_keys` and hardcoded logic to the handshake to differentiate between schemes.

While this approach reduced the scope of the PR and preserved backward compatibility, it limits how cleanly Rustls can support schemes beyond X.509, as highlighted in discussions #2258 and #2266. The `CertVerifier` traits, designed exclusively for X.509, now have potentially invalid states and are awkward to use for other identity types.

Furthermore, the current design does not scale to support future schemes and does *not* support custom/private types at all. For me, enabling custom schemes is the primary motivation for this PR.

### How

1.  **Introduced `Identity` Abstraction:** This PR introduces a new `TlsIdentity` trait to decouple the API from X.509 certificates. Implementations for X.509 and RPK are included. A peer's identity is accessible via `peer_identity()`, while `peer_certificates()` is kept for backward compatibility. Convenience methods like `Identity::as_certificates` and `Identity::as_public_key` are provided, with `Identity::as_any` available for custom types.

2.  **New Verifier/Resolver Traits:** New, scheme-neutral traits have been introduced: `ServerIdVerifier`, `ClientIdVerifier`, `ResolvesServerIdentity`, and `ResolvesClientIdentity`. Implementations are only required for new or custom schemes. This PR includes implementations for X.509 (wrapping the existing `CertVerifier` traits for compatibility) and new, purpose-built `RpkVerifier`/`RpkResolver` traits for RPK users.

3.  **Dynamic Handshake Logic:** Hardcoded handshake logic has been replaced with a dynamic approach that delegates negotiation to the new verifier/resolver traits. X.509-specific code has been moved into its respective trait implementation, making the main code paths identity-scheme neutral.

4.  **Support for Custom Types:** The `CertificateType` enum was extended to support the custom/private use range (224-255). Basic range support was added to the `enum_builder`.

5.  **Terminology Update:** Throughout the codebase, `certificate`/`cert` has been replaced with `identity`/`id` where appropriate to reflect the new abstractions.

### Impact

The changes are necessarily invasive but establish a foundation for long-term API evolution. The number of changed LOC is substantial, though many are simple renames.

#### Code Changes

The most meaningful changes are in:
*   `identity.rs`: Contains the main `Identity` type,  the `TlsIdentity` & `IdentitySigner` traits and their respective implementations.
*   `verify.rs`: Adds the new `ServerIdVerifier` & `ClientIdVerifier` traits, compatibility wrappers for X.509, and the new `ServerRpkVerifier` & `ClientRpkVerifier` traits.
*   `server_conn.rs` & `client_conn.rs`:  Adds new identity resolvers and public setters for using non-X.509 schemes.
*   `common_state.rs`: Adds the `peer_identity()` method and fields to store the negotiated certificate types.
*   Handshake logic: Updated to use the new dynamic negotiation. The changes are mostly related to TLS 1.3, TLS 1.2 remains limited to X.509 certificate support.
*   `builder.rs`: Adds `with_*_rpk_resolver` and `with_*_identity_resolver` methods to supplement existing ones.

#### Performance

Performance should be unaffected. Any significant regression should be considered a bug.

#### Backward Compatibility

Although the changes are substantial, care has been taken to minimize breakage.

1.  **RPK Users:** This is a breaking change. Users must migrate to the new, purpose-built RPK traits (e.g., `ClientRpkVerifier` instead of `ClientCertVerifier`) and use the new `with_rpk_*` configuration methods. The peer's key should be accessed via `peer_identity().as_public_key()`. These changes result in a cleaner, more correct API for RPK usage.

2.  **Direct Config Field Access:** For users who directly access the public `ClientConfig::client_auth_cert_resolver` or `ServerConfig::cert_resolver` fields, 100% backward compatibility cannot be maintained, as these fields have been renamed and re-typed. Affected users must use the similarly named methods instead (e.g. change from `config.cert_resolver = X;` to `config.cert_resolver(X);`.
 *Note: Users who rely on the builder methods are **not** affected.*

Overall, this has been an interesting endeavor. I believe these changes are needed in one form or another. Looking forward to everyone's feedback and opinions.

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -641,7 +641,7 @@ pub(super) mod danger {
     use crate::sync::Arc;
     use crate::verify::{
         ServerCertVerifier, ServerCertVerifierCompat, ServerIdVerifier, ServerRpkVerifier,
-        ServerRpkVerifierCompat,
+        ServerRpkVerifierWrapper,
     };
 
     /// Accessor for dangerous configuration options.
@@ -653,17 +653,17 @@ pub(super) mod danger {
 
     impl DangerousClientConfig<'_> {
         /// Overrides the default `ServerIdVerifier` with something else.
-        pub fn certificate_verifier(&mut self, verifier: Arc<dyn ServerCertVerifier>) {
-            self.identity_verifier(Arc::new(ServerCertVerifierCompat::from(verifier)))
+        pub fn set_certificate_verifier(&mut self, verifier: Arc<dyn ServerCertVerifier>) {
+            self.set_identity_verifier(Arc::new(ServerCertVerifierCompat::from(verifier)))
         }
 
         /// Overrides the default `ServerIdVerifier` with something else.
-        pub fn rpk_verifier(&mut self, verifier: Arc<dyn ServerRpkVerifier>) {
-            self.identity_verifier(Arc::new(ServerRpkVerifierCompat::from(verifier)))
+        pub fn set_rpk_verifier(&mut self, verifier: Arc<dyn ServerRpkVerifier>) {
+            self.set_identity_verifier(Arc::new(ServerRpkVerifierWrapper::from(verifier)))
         }
 
         /// Overrides the default `ServerIdVerifier` with something else.
-        pub fn identity_verifier(&mut self, verifier: Arc<dyn ServerIdVerifier>) {
+        pub fn set_identity_verifier(&mut self, verifier: Arc<dyn ServerIdVerifier>) {
             self.cfg.verifier = verifier;
         }
     }
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -1,3 +1,10 @@
+use alloc::vec::Vec;
+use core::marker::PhantomData;
+use core::ops::{Deref, DerefMut};
+use core::{fmt, mem};
+
+use pki_types::{ServerName, UnixTime};
+
 use super::handy::NoClientSessionStorage;
 use super::hs::{self, ClientHelloInput};
 #[cfg(feature = "std")]
@@ -11,6 +18,7 @@ use crate::crypto;
 use crate::crypto::{CryptoProvider, SupportedKxGroup};
 use crate::enums::{CertificateType, CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::error::Error;
+use crate::identity::SigningIdentity;
 use crate::kernel::KernelConnection;
 use crate::log::trace;
 use crate::msgs::enums::NamedGroup;
@@ -23,11 +31,6 @@ use crate::time_provider::DefaultTimeProvider;
 use crate::time_provider::TimeProvider;
 use crate::unbuffered::{EncryptError, TransmitTlsData};
 use crate::{DistinguishedName, KeyLog, WantsVersions, compress, sign, verify, versions};
-use alloc::vec::Vec;
-use core::marker::PhantomData;
-use core::ops::{Deref, DerefMut};
-use core::{fmt, mem};
-use pki_types::{ServerName, UnixTime};
 
 /// A trait for the ability to store client session data, so that sessions
 /// can be resumed in future connections.
@@ -146,6 +149,18 @@ impl From<Arc<dyn ResolvesClientCert>> for ResolvesClientCertCompat {
     }
 }
 
+impl<T: ResolvesClientCert + 'static> From<Arc<T>> for ResolvesClientCertCompat {
+    fn from(value: Arc<T>) -> Self {
+        Self(value)
+    }
+}
+
+impl<T: ResolvesClientCert + 'static> From<T> for ResolvesClientCertCompat {
+    fn from(value: T) -> Self {
+        Self(Arc::new(value))
+    }
+}
+
 impl ResolvesClientIdentity for ResolvesClientCertCompat {
     fn resolve(
         &self,
@@ -192,6 +207,18 @@ pub(crate) struct ResolvesClientRpkWrapper(Arc<dyn ResolvesClientRpk>);
 impl From<Arc<dyn ResolvesClientRpk>> for ResolvesClientRpkWrapper {
     fn from(value: Arc<dyn ResolvesClientRpk>) -> Self {
         Self(value)
+    }
+}
+
+impl<T: ResolvesClientRpk + 'static> From<Arc<T>> for ResolvesClientRpkWrapper {
+    fn from(value: Arc<T>) -> Self {
+        Self(value)
+    }
+}
+
+impl<T: ResolvesClientRpk + 'static> From<T> for ResolvesClientRpkWrapper {
+    fn from(value: T) -> Self {
+        Self(Arc::new(value))
     }
 }
 
@@ -342,10 +369,10 @@ pub struct ClientConfig {
     /// How to verify the server certificate chain.
     pub(super) verifier: Arc<dyn verify::ServerIdVerifier>,
 
-    /// How to decompress the server's certificate chain.
+    /// How to decompress the server's identity.
     ///
     /// If this is non-empty, the [RFC8779] certificate compression
-    /// extension is offered, and any compressed certificates are
+    /// extension is offered, and any compressed identities are
     /// transparently decompressed during the handshake.
     ///
     /// This only applies to TLS1.3 connections.  It is ignored for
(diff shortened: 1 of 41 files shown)


rustls/src/crypto/signer.rs
```diff
@@ -207,6 +208,69 @@ impl CertifiedKey {
...
+        self.key.deref()
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct KeyPair {
```
ctz, line 228: I think this naming is much too generic. I am concerned that a typical user of rustls _does not_ have a use for RPK, but this naming makes it sound enticingly useful.

Maybe we could have an `rustls::rpk` module (or something) collecting these things, whose doc comment says "if you stumbled upon this this isn't what you want"

rustls/src/identity.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::msgs::handshake;
+use crate::msgs::handshake::{
+    CertificateChain, CertificatePayloadTls13, CertificateStatus, IdentityEntry,
+};
+use crate::sign::{CertifiedKey, SigningKey};
+use crate::sync::Arc;
+use core::fmt::Debug;
+use pki_types::{CertificateDer, SubjectPublicKeyInfoDer};
+use std::any::Any;
+use std::prelude::rust_2015::{Box, Vec};
```
ctz: This is odd. Use `alloc::vec::Vec` and `alloc::boxed::Box`?

rustls/src/lib.rs
```diff
@@ -391,10 +391,10 @@ use log;
...
 #[cfg(not(feature = "log"))]
 mod log {
-    macro_rules! trace    ( ($($tt:tt)*) => {{}} );
-    macro_rules! debug    ( ($($tt:tt)*) => {{}} );
-    macro_rules! error    ( ($($tt:tt)*) => {{}} );
-    macro_rules! _warn    ( ($($tt:tt)*) => {{}} );
+    macro_rules! trace ( ($($tt:tt)*) => {{}} );
+    macro_rules! debug ( ($($tt:tt)*) => {{}} );
+    macro_rules! error ( ($($tt:tt)*) => {{}} );
+    macro_rules! _warn ( ($($tt:tt)*) => {{}} );
```
ctz, line 397: unnecessary whitespace change

rustls/src/client/client_conn.rs
```diff
@@ -1,10 +1,3 @@
-use alloc::vec::Vec;
-use core::marker::PhantomData;
-use core::ops::{Deref, DerefMut};
```
ctz: Please run `cargo +nightly fmt-unstable` to sort out the import ordering/groupings.

rustls/src/client/client_conn.rs
```diff
@@ -121,17 +133,98 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
...
-    fn only_raw_public_keys(&self) -> bool {
-        false
-    }
-
     /// Return true if any certificates at all are available.
     fn has_certs(&self) -> bool;
 }
 
+#[derive(Debug)]
+pub(crate) struct ResolvesClientCertCompat(Arc<dyn ResolvesClientCert>);
```
ctz, line 144: I appreciate the idea of this compatibility layer, but I think `ResolvesClientIdentity` can replace `ResolvesClientCert` in its entirety. (main is open for breaking changes right now, but I suspect it wasn't when you started these changes.)

rustls/src/client/client_conn.rs
```diff
@@ -121,17 +133,98 @@ pub trait ResolvesClientCert: fmt::Debug + Send + Sync {
...
+
+    fn supported_certificate_types(&self) -> &[CertificateType] {
+        &[CertificateType::X509]
+    }
+}
+
+pub trait ResolvesClientRpk: fmt::Debug + Send + Sync {
+    fn resolve(
+        &self,
+        root_hint_subjects: &[&[u8]],
```
ctz, line 197: If we keep this, I think this parameter doesn't make sense for something that only dispenses RPKs (which can't be filtered/named by X509 distinguished names?

rustls/src/client/hs.rs
```diff
@@ -634,30 +649,50 @@ pub(super) fn process_server_cert_type_extension(
...
-        config
-            .verifier
-            .requires_raw_public_keys(),
-        server_cert_extension.copied(),
-        ExtensionType::ServerCertificateType,
-    )
+) -> Result<(), Error> {
+    // X.509 is implied of no extension present
+    let accepted = server_cert_extension
+        .map(|c| c)
```
ctz, line 655: Suspect this `map()` has no effect.

rustls/src/client/tls13.rs
```diff
@@ -912,10 +913,19 @@ impl State<ClientConnectionData> for ExpectCertificateRequest {
             })
             .cloned();
 
+        let cert_type = cx
+            .common
+            .negotiated_client_cert_type()
```
ctz, line 918: Generally let's not insert things into `cx.common` early in the handshake and then extract them later. Instead, thread these into the handshake states and expose them in the public API at exactly the same point as `cx.common.peer_identity` is changed.

rustls/src/msgs/base.rs
```diff
@@ -60,17 +74,62 @@ impl Payload<'static> {
...
+    fn from(value: &'a [u8]) -> Self {
+        Self::Borrowed(value)
+    }
+}
+
+impl<'a> From<&'a Vec<u8>> for Payload<'a> {
+    fn from(value: &'a Vec<u8>) -> Self {
+        Self::Borrowed(value.as_slice())
+    }
+}
```
ctz, line 73: This is weird.

rustls/src/msgs/handshake.rs
```diff
@@ -1663,76 +1676,93 @@ extension_struct! {
...
+        Self::read_payload(&mut sub)
     }
 }
 
-#[derive(Debug)]
-pub(crate) struct CertificateEntry<'a> {
-    pub(crate) cert: CertificateDer<'a>,
-    pub(crate) extensions: CertificateExtensions<'a>,
+#[derive(Debug, Clone)]
+pub struct IdentityEntry<'a> {
```
ctz, line 1704: These types should be named -- as far as possible -- after types in RFC8446. I suggest we keep this as `CertificateEntry` but make it a rust enum with `RawPublicKey` and `X509` variants. ref: https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2

rustls/src/msgs/handshake.rs
```diff
@@ -2528,7 +2506,7 @@ impl HandshakePayload<'_> {
             ClientHello(x) => x.encode(bytes),
             ServerHello(x) => x.encode(bytes),
             HelloRetryRequest(x) => x.encode(bytes),
-            Certificate(x) => x.encode(bytes),
+            Certificate(x) => Codec::encode(x, bytes),
```
ctz: This is weird.

rustls/src/enums.rs
```diff
@@ -584,6 +584,9 @@ enum_builder! {
     pub enum CertificateType {
         X509 => 0x00,
         RawPublicKey => 0x02,
+        
+        !Ranges:
+          Custom => 224..=255,
```
ctz, line 589: This seems to be unused/unmotivated? FWIW this type can already represent the full range of standard values (unassigned ones via the `Unknown(_)` variant)

djc:
Conceptually this is definitely interesting to me -- decoupling TLS and X.509 seems like a worthwhile way to further modularize rustls and make it usable across a broader range of use cases. Ideally we'd broaden the interface while trying not to meaningfully degrade the ergonomics for the vast majority that wants to just use certificates.

On the implementation side, I think this needs a lot of work to be viable. I'd suggest picking much smaller chunks to happen at a time, starting with picking client or server and working out what it needs.

Some aspects of your PR seem to be affected by LLM-assisted coding. While I'm a happy Copilot user, I'd strongly suggest writing any descriptive prose yourself and making sure to closely self-review any LLM-generated code before submitting it to us for review. We usually work with small commits exactly because they make review easier. Most of this is explained in our [`CONTRIBUTING.md`](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md).

cpu:
Since CI is red, there are conflicts, and there's been no response to the initial round of feedback I'm going to close this to keep the PR queue tidy.

PR #2550, opened 2025-07-10
post-quantum: add unstable ML-DSA support

djc:
Fixes #2417.

PR #2564, opened 2025-07-15
Include MLKEM in bogo tests of aws-lc-rs

ctz:
This is available by default now.

(found in review of #2550)


ctz:
> I think this needs a tweak for `Compliance-fips-202205-TLS-[Client|Server]-MLKEM` ? 🤔
> 
> Both are [failing in bogo](https://github.com/rustls/rustls/actions/runs/16298447314/job/46026390856?pr=2564) as unexpected success.

Doh. I didn't try that combination locally.

PR #2565, opened 2025-07-16
Audit public enums for exhaustiveness

### LINKED ISSUE #2398: Audit public enums for exhaustivity
Should we make more enums `#[non_exhaustive]`?

ctz:
(would like to have written `#[expect()]` here, but our MSRV precludes that.)

fixes #2398

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -377,6 +377,15 @@ don't provide additional type safety. Using the
 [newtype idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html)
 is one alternative when an abstraction boundary is worth the added complexity.
 
+#### Type exhaustiveness
+
+Public enums should be marked as _either_ `#[non_exhaustive]` or `#[allow(clippy::exhaustive_enums)]`.
+The latter is suitable for enums that are already complete by definition.  For example:
+`enum CoinFlip { Heads, Tails }` is complete.  Err on the side of marking something `#[non_exhaustive]`.
+
+The same applies to structs, with the detail that no manual marking is needed for
+structures with at least one private field.
+
 ## Design and Architecture
 
 Some general concepts about how the library should fit together:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,6 +144,22 @@ Note that we usually also practice top-down ordering here; where these are in
 conflict, make a choice that you think makes sense. For getters and setters, the
 order should typically mirror the order of the fields in the type definition.
 
+#### Attribute ordering
+
+Order attributes so that documentation appears first, and the attributes with the
+most effect on the meaning and function of the type appear last.  For example:
+
+```rust
+/// Doc comment always first
+#[cfg(feature-gates)]
+#[allow(lint-configuration)]
+#[non_exhaustive]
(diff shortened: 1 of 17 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -520,6 +520,7 @@ impl Default for Resumption {
 
 /// What mechanisms to support for resuming a TLS 1.2 session.
 #[derive(Clone, Copy, Debug, PartialEq)]
+#[non_exhaustive]
```
djc, line 522: Style nit: do we put `non_exhaustive` "outside" the derives or "inside" them? Current usage is not completely consistent though it seems like the majority is on the outside. To me, since non-exhaustiveness feels like a "bigger" change to a type than its derives, I have been inclined to keep it before derives.

rustls/src/common_state.rs
```diff
@@ -888,6 +888,7 @@ pub(crate) struct Context<'a, Data> {
 
 /// Side of the connection.
 #[derive(Clone, Copy, Debug, PartialEq)]
+#[allow(clippy::exhaustive_enums)]
```
djc, line 890: I think I would also be inclined to put lint suppressions before derives.

djc (approved):
Looks great!

Did you use your tooling for this? You should publish/publicize that a bit!

djc:
Should we have a blurb about this in `CONTRIBUTING.md`?

ctz:
> Did you use your tooling for this? You should publish/publicize that a bit!

I've been porting that to a few clippy lints, and in doing that found clippy already supported the exhaustiveness parts, since 1.51!

djc:
> > Did you use your tooling for this? You should publish/publicize that a bit!
> 
> I've been porting that to a few clippy lints, and in doing that found clippy already supported the exhaustiveness parts, since 1.51!

Hah, awesome!

ctz:
> Should we have a blurb about this in `CONTRIBUTING.md`?

Added one about exhaustiveness and another above attribute ordering.

CONTRIBUTING.md
```diff
@@ -144,6 +144,22 @@ Note that we usually also practice top-down ordering here; where these are in
...
+attributes appear last.  For example:
+
+```rust
+/// Doc comment always first
+#[cfg(feature-gates)]
+#[allow(lint-configuration)]
+#[non_exhaustive]
+#[derive(Clone, Debug)]
+pub struct Foo;
+```
```
djc: How would you order these as import? From most to least important, I'd suggest:

- `cfg` (can turn the entire thing off)
- `non-exhaustive` (controls important downstream behavior)
- derive stuff (trait implementations)
- lint stuff (irrelevant to downstreams)

But, I'd probably order like this:

- Docstring
- cfg
- non-exhaustive
- lints
- derive

I guess maybe because I feel like the derives should not concern themselves with the attributes/the procedural macros should be close to the syntax they use as input?

ctz: Probably important isn't the right word. The way I think about this is how much effect each attribute has on the type. `derive` has the most effect, then `non_exhaustive` . Lints, cfg gates and comments have ~no effect on the type.

ctz: Reworded.

PR #2566, opened 2025-07-16
Audit public structs for exhaustiveness

### LINKED ISSUE #2399: Audit public struct types for exhaustivity
Should we make more structs `#[non_exhaustive]`?

ctz:
After #2565 

fixes #2399

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -587,6 +587,7 @@ impl Secrets {
 }
 
 /// Keys used to communicate in a single direction
+#[allow(clippy::exhaustive_structs)]
 pub struct DirectionalKeys {
     /// Encrypts or decrypts a packet's headers
     pub header: Box<dyn HeaderProtectionKey>,
@@ -771,6 +772,7 @@ pub trait PacketKey: Send + Sync {
 }
 
 /// Packet protection keys for bidirectional 1-RTT communication
+#[allow(clippy::exhaustive_structs)]
 pub struct PacketKeySet {
     /// Encrypts outgoing packets
     pub local: Box<dyn PacketKey>,
@@ -862,6 +864,7 @@ impl Suite {
 }
 
 /// Complete set of keys used to communicate with the peer
+#[allow(clippy::exhaustive_structs)]
 pub struct Keys {
     /// Encrypts outgoing packets
     pub local: DirectionalKeys,
--- a/rustls/src/webpki/verify.rs
+++ b/rustls/src/webpki/verify.rs
@@ -55,6 +55,7 @@ pub fn verify_server_name(
 
 /// Describes which `webpki` signature verification algorithms are supported and
 /// how they map to TLS [`SignatureScheme`]s.
+#[allow(clippy::exhaustive_structs)]
 #[derive(Clone, Copy)]
 pub struct WebPkiSupportedAlgorithms {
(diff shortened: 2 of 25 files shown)


rustls/src/webpki/verify.rs
```diff
@@ -55,8 +55,8 @@ pub fn verify_server_name(
 
 /// Describes which `webpki` signature verification algorithms are supported and
 /// how they map to TLS [`SignatureScheme`]s.
+#[allow(clippy::exhaustive_structs)]
 #[derive(Clone, Copy)]
-#[allow(unreachable_pub)]
```
cpu, line 59: nit: the few small re-orderings of traits seem fine to do alongside adding an `#[allow(...)]`, but I think it'd be a touch tidier if this removal was its own commit. It feels just at the cusp of being notable enough to stand alone (also super nit picky :laughing:)

rustls/src/client/builder.rs
```diff
@@ -96,6 +96,7 @@ pub(super) mod danger {
     use crate::{ClientConfig, ConfigBuilder, WantsVerifier, verify};
 
     /// Accessor for dangerous configuration options.
+    #[non_exhaustive]
```
djc: Why is `cfg` public anyway?

rustls/src/crypto/signer.rs
```diff
@@ -135,6 +135,7 @@ impl ResolvesServerCert for SingleCertAndKey {
 /// certificates.
 ///
 /// [RFC 7250]: https://tools.ietf.org/html/rfc7250
+#[non_exhaustive]
```
djc: I think this one will merged via #2553.

rustls/src/quic.rs
```diff
@@ -587,6 +587,7 @@ impl Secrets {
 }
 
 /// Keys used to communicate in a single direction
+#[non_exhaustive]
```
djc: Hmm, not sure about this one...

rustls/src/quic.rs
```diff
@@ -771,6 +772,7 @@ pub trait PacketKey: Send + Sync {
 }
 
 /// Packet protection keys for bidirectional 1-RTT communication
+#[non_exhaustive]
```
djc: And both this and `Keys` feel like they could probably stick to being exhaustive, though I didn't find any Quinn code relying on this.

rustls/src/conn/unbuffered.rs
```diff
@@ -170,6 +170,7 @@ impl<Data> UnbufferedConnectionCommon<Data> {
 }
 
 /// The current status of the `UnbufferedConnection*`
+#[allow(clippy::exhaustive_structs)]
```
djc: Is this externally constructed? Not convinced it's definitionally complete. Same for the other `unbuffered` types.

djc (commented):
Seeing the ratio between `non_exhaustive` and `#[allow(clippy::exhaustive_structs)]`, not totally convinced the lint is worth it here...

rustls/src/client/builder.rs
ctz: added a commit resolving this

rustls/src/webpki/verify.rs
ctz: split these changes into two precursor commits

Issue #2401, opened 2025-03-25
Evaluate whether CryptoProvider should become a trait


djc:
This would enable additional flexibility in the future (allow us to add trait methods with default impls).

As a type (in its current definition), it benefits from update syntax, which allows downstream implementations to reuse most of an existing implementation and replace a part of it.

However, even if we have a trait, we could also provide a "default" type that can still be used like that.

`CryptoProvider` was originally conceived as a trait but reworked to be a type in #1628. Reviewing the PR description, I think most of the advantages would still hold if we switched back to a trait (+ type).

IIRC around the QUIC changes for 0.23 (around #1798) there was some motivation to make changed to `CryptoProvider` (add more fields) in a way that the current design does not allow.

ctz:
> As a type (in its current definition), it benefits from update syntax, which allows downstream implementations to reuse most of an existing implementation and replace a part of it.

I noticed today that the update syntax doesn't work well if someone wants to avoid and eliminate a default from their program. It's not especially surprising, because for this to work pretty extreme compiler magic would need to be happening. That means `examples/src/bin/limitedclient` actually does have AES compiled-in, despite what its head comment says.

The other goal I think we could adopt for this ticket is allowing `const` construction of `CryptoProvider`s.

djc:
https://github.com/rust-lang/rust/issues/132162 is coming, but will presumably take a while to be within our MSRV.

stevefan1999-personal:
Speaking as a crpyto provider implementor, for the short run this would be a little painful, but for the long run I totally support this decision :)

ctz:
> The other goal I think we could adopt for this ticket is allowing `const` construction of `CryptoProvider`s.

BTW #2675 addressed this.

I think the sticking point here is whether we either:

a) commit to having boilerplate in each provider crate, mapping structures to the trait
b) have those types in the core crate, in which case we have more code but not the stability benefit

djc:
I think I'm okay with closing this for 0.24 (and, presumably, 1.0).

Issue #1659, opened 2023-12-01
Specify ciphersuites for TLS 1.2 and TLS 1.3 separately


jsha:
First discussed here: https://github.com/rustls/rustls/pull/564#issuecomment-826049679, and brought up again here: https://github.com/rustls/rustls/pull/1628#pullrequestreview-1756467489.

Right now, we specify cipher suites and protocol versions independently. That introduces a possible error case in config building: since our cipher suites are TLS 1.2-only or TLS 1.3-only, it's possible to specify "Only allow TLS 1.3; only allow `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` (a TLS 1.2 cipher suite)". We check that error at config building time, and it's one of two places during config building that the user has to handle an error.

We also have three paths through config building when specifying protocols:

 - `ClientConfig::builder_with_protocols()`
 - `ClientConfig::builder().with_protocols()`
 - `ClientConfig::builder_with_provider().with_protocols()`

It would be nice to reduce this to one path:

 - `ClientConfig::builder_with_provider(CryptoProvider { tls12_cipher_suites: vec![], ..RING })`
 
Also we want to support eliminating the `tls12` Cargo feature (#224). In practice that means each `SupportedCipherSuite` needs to contain a `&'static SupportedProtocolVersion`, so that the linker only has a path to the TLS 1.2 implementation code if there was any TLS 1.2 cipher suite referenced in the application code.

sayrer:
Yeah, there is something kind of off here. In my older ECH efforts, I made a builder that could only use TLS 1.3 cypher suites, by adding such a thing to those enums. The idea being that code attempting to offer ECH where TLS 1.2 was a possibility wouldn't even compile. I think that is the desired result, but I don't have a strong opinion on how to do it.

PR #2568, opened 2025-07-17
Turn `SupportedProtocolVersion`s into capabilities and eliminate `tls12` crate feature

### LINKED ISSUE #224: Allow unused protocol versions to be discarded at link-time

ctz:
This fixes #224 . I've manually verified this works using `limitedclient`, but as noted on [#2401](https://github.com/rustls/rustls/issues/2401#issuecomment-3083716092) this doesn't work as-is because the call to `aws_lc_rs::default_provider()` links in all the cipher suites anyway.

```
$ nm -C target/debug/limitedclient | grep rustls::client::tls13 | wc -l
87
$ nm -C target/debug/limitedclient | grep rustls::client::tls12 | wc -l
0
```
(in other words, none of the TLS1.2 handshake machinery exists in a program that only uses TLS1.3)

Note the naming of `rustls::version::TLS12_VERSION` is janky but temporary: after this I want to eliminate `SupportedProtocolVersion` and all the config builder steps to do with version configuration will fall away. During that `rustls::version::TLS12_VERSION` will become `rustls::version::TLS12` (etc.)

After _that_ `SupportedCipherSuite` will disappear in the course of addressing #1659.  Instead, someone configures a set of `Tls12CipherSuite`s and `Tls13CipherSuite`s, and their program references `rustls::version::TLS12`/`rustls::version::TLS13` only via these.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -232,7 +232,7 @@ struct ExpectCertificate {
     randoms: ConnectionRandoms,
     using_ems: bool,
     transcript: HandshakeHash,
-    pub(super) suite: &'static Tls12CipherSuite,
+    suite: &'static Tls12CipherSuite,
     may_send_cert_status: bool,
     must_issue_new_ticket: bool,
 }
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -4,7 +4,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 
 use pki_types::ServerName;
-pub(super) use server_hello::handle_server_hello;
+pub(crate) use server_hello::{TLS12_HANDLER, Tls12Handler};
 use subtle::ConstantTimeEq;
 
 use super::client_conn::ClientConnectionData;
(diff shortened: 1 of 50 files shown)


djc:
Previous feedback from Discord:

> I think maybe the `Tls13CipherSuite` / `Tls12CipherSuite` -> version capability could be an associated const on the impl instead of having be a field on the type?

More generally -- do we need the separate "capability" type, or could we just use the cipher suite types as the capability?

> Note the naming of rustls::version::TLS12_VERSION is janky but temporary: after this I want to eliminate SupportedProtocolVersion and all the config builder steps to do with version configuration will fall away. During that rustls::version::TLS12_VERSION will become rustls::version::TLS12 (etc.)

This sounds great! Only took us [4 years](https://github.com/rustls/rustls/pull/564#issuecomment-826049679)... 😄

ctz:
> Previous feedback from Discord:
> 
> > I think maybe the `Tls13CipherSuite` / `Tls12CipherSuite` -> version capability could be an associated const on the impl instead of having be a field on the type?

I tried this:

```patch
diff --git a/rustls/src/server/hs.rs b/rustls/src/server/hs.rs
index 5b14ba217..78fceeb02 100644
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -27,7 +27,7 @@ use crate::msgs::persist;
 use crate::server::common::ActiveCertifiedKey;
 use crate::server::{ClientHello, ServerConfig, tls13};
 use crate::sync::Arc;
-use crate::{SupportedCipherSuite, suites};
+use crate::{SupportedCipherSuite, Tls12CipherSuite, suites};
 
 pub(super) type NextState<'a> = Box<dyn State<ServerConnectionData> + 'a>;
 pub(super) type NextStateOrError<'a> = Result<NextState<'a>, Error>;
@@ -494,8 +494,7 @@ impl ExpectClientHello {
                         extra_exts: self.extra_exts,
                     },
                 ),
-            SupportedCipherSuite::Tls12(suite) => suite
-                .protocol_version
+            SupportedCipherSuite::Tls12(suite) => Tls12CipherSuite::VERSION
                 .server
                 .handle_client_hello(
                     cx,
diff --git a/rustls/src/tls12/mod.rs b/rustls/src/tls12/mod.rs
index a0ec09fc5..3fc3b4623 100644
--- a/rustls/src/tls12/mod.rs
+++ b/rustls/src/tls12/mod.rs
@@ -83,6 +83,8 @@ impl Tls12CipherSuite {
     pub fn fips(&self) -> bool {
         self.common.fips() && self.prf_provider.fips() && self.aead_alg.fips()
     }
+
+    pub(crate) const VERSION: &'static Tls12Version = crate::version::TLS12_VERSION;
 }
 
 impl From<&'static Tls12CipherSuite> for SupportedCipherSuite {
```

Assuming that's what you mean, unfortunately it doesn't work because the reference to `TLS12_VERSION` is made statically by `ExpectClientHello` (via the associated const) and that unconditionally brings in both protocol versions.
 
> More generally -- do we need the separate "capability" type, or could we just use the cipher suite types as the capability?

I think there are quite a few ways of arranging this. The main goal is keeping the details private and sealed -- so we can have private types in the "handoff" interface, and change those interfaces at will.

> > Note the naming of rustls::version::TLS12_VERSION is janky but temporary: after this I want to eliminate SupportedProtocolVersion and all the config builder steps to do with version configuration will fall away. During that rustls::version::TLS12_VERSION will become rustls::version::TLS12 (etc.)
> 
> This sounds great! Only took us [4 years](https://github.com/rustls/rustls/pull/564#issuecomment-826049679)... 😄

😅

djc:
> I think there are quite a few ways of arranging this. The main goal is keeping the details private and sealed -- so we can have private types in the "handoff" interface, and change those interfaces at will.

A sealed trait for the versioned cipher suite types would do that, I think? Seems nicer than introducing the separate zero-sized types.

ctz:
> > I think there are quite a few ways of arranging this. The main goal is keeping the details private and sealed -- so we can have private types in the "handoff" interface, and change those interfaces at will.
> 
> A sealed trait for the versioned cipher suite types would do that, I think? Seems nicer than introducing the separate zero-sized types.

Tried this too. Unfortunately I think it needs to be a dyn trait (a static trait introduces a static reference to that code), so `impl Tls12Handler for Tls12CipherSuite` is great but there's no way to get a `dyn Tls12Handler` from a `Tls12CipherSuite` that is not unconditional. Hm.

rustls/src/versions.rs
```diff
@@ -65,8 +67,18 @@ pub static TLS13_VERSION: &Tls13Version = &Tls13Version {};
...
+    pub(crate) client: &'static dyn crate::client::Tls12Handler,
+}
+
+// TODO
+impl PartialEq for Tls12Version {
+    fn eq(&self, _: &Self) -> bool {
+        true
+    }
+}
+impl Eq for Tls12Version {}
```
djc: Nit: empty line?

rustls/src/versions.rs
```diff
@@ -65,8 +67,18 @@ pub static TLS13_VERSION: &Tls13Version = &Tls13Version {};
...
 /// There is one value of this type.  It is `TLS12_VERSION`.
 #[non_exhaustive]
-#[derive(Debug, Eq, PartialEq)]
-pub struct Tls12Version {}
+#[derive(Debug)]
+pub struct Tls12Version {
+    pub(crate) client: &'static dyn crate::client::Tls12Handler,
+}
+
+// TODO
```
djc: When/why?

rustls/src/client/tls12.rs
```diff
@@ -40,29 +40,45 @@ mod server_hello {
...
-        pub(in crate::client) randoms: ConnectionRandoms,
-        pub(in crate::client) transcript: HandshakeHash,
-        pub(in crate::client) input: ClientHelloInput,
-    }
+    pub(crate) static TLS12_HANDLER: &dyn Tls12Handler = &Handler;
 
-    impl CompleteServerHelloHandling {
-        pub(in crate::client) fn handle_server_hello(
-            mut self,
+    pub(crate) trait Tls12Handler: Sealed + Send + Sync + core::fmt::Debug {
```
djc: Trait definition below type implementing it? Maybe move `fmt::Debug` before `Send + Sync` (because more notable) and import `fmt` at the top of the module?

In terms of the commit history, would suggest splitting this first into a free function, to deal with all the `self.` removals, and then introducing the `Handler`.

rustls/src/client/tls12.rs
```diff
@@ -40,29 +40,45 @@ mod server_hello {
...
+            randoms: ConnectionRandoms,
             suite: &'static Tls12CipherSuite,
+            transcript: HandshakeHash,
+            tls13_supported: bool,
+            input: ClientHelloInput,
+        ) -> hs::NextStateOrError<'static>;
+    }
+
+    #[derive(Debug)]
+    struct Handler;
```
djc, line 50: Could this be replaced by an existing type? Maybe `Tls12Version` or `Tls12CipherSuite`?

rustls/src/server/tls12.rs
```diff
@@ -46,40 +47,50 @@ mod client_hello {
...
+        fn handle_client_hello(
+            &self,
             cx: &mut ServerContext<'_>,
             server_key: ActiveCertifiedKey<'_>,
             chm: &Message<'_>,
             client_hello: &ClientHelloPayload,
             selected_kxg: &'static dyn SupportedKxGroup,
             sigschemes_ext: Vec<SignatureScheme>,
             tls13_enabled: bool,
+            mut cch: CompleteClientHelloHandling,
```
djc: Nits: same here about isolating the `self` -> `cch` renames. Also suggest moving `cch` up front in the argument list.

ctz:
> This feels pretty type tetrissy and unintuitively complex. Do we have any tests that defend against this yet? I think we should have some because it seems easy to regress.

Not so far. The tests I have in mind (and have been doing manually) basically build `limitedclient` and then observe whether or not TLS1.2-specific symbols are present or not. Ideally that sort of test would use the same mechanism to check that those symbols _are_ present in `simpleclient`.

Maybe I'll add this test now?  The caveat is that it has to come with some unpalatable temporary publicising of aws-lc-rs provider internals. That could go away with #2401 though.

djc:
> Maybe I'll add this test now? The caveat is that it has to come with some unpalatable temporary publicising of aws-lc-rs provider internals. That could go away with #2401 though.

How internal do you mean? Just a particular symbol name? That seems straightforward to me.

ctz:
> > Maybe I'll add this test now? The caveat is that it has to come with some unpalatable temporary publicising of aws-lc-rs provider internals. That could go away with #2401 though.
> 
> How internal do you mean? Just a particular symbol name? That seems straightforward to me.

Actually it doesn't need any internals, just some statics that have the value of (eg)  `aws_lc_rs::default_provider().secure_random`. Addressing this first in #2573

rustls/src/versions.rs
ctz: impl'd this for `SupportedProtocolVersion` instead

rustls/src/versions.rs
ctz: impl'd this for `SupportedProtocolVersion` instead

ctz:
> Do we have any tests that defend against this yet? I think we should have some because it seems easy to regress.

Added a commit for this.

rustls/src/client/tls12.rs
ctz: So we need a value of the type to take a dyn reference. For `Tls12Version` that value would contain itself. For `Tls12CipherSuite` that would need to refer to itself too, via the `protocol_version` field.

The other way of structuring this is to use function pointers. That is a bit less idiomatic rust and does less well at hiding the types of arguments (but can be made to work).

PR #2571, opened 2025-07-18
docs CI: run for pull request, but skip deploys

ctz:
We have an deployment protection that limits this
to only stuff reaching main.


.github/workflows/docs.yml
```diff
@@ -82,7 +83,7 @@ jobs:
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
-    if: github.repository == 'rustls/rustls'
+    if: github.repository == 'rustls/rustls' && (github.event_name == 'push' || github.event_name == 'schedule')
```
djc: Do we want to do scheduled deploys? Seems like we should deploy only push to `main`?

ctz: altered this to just be anything that builds main

PR #2573, opened 2025-07-23
Fix and defend `limitedclient`

ctz:
`limitedclient` is meant to demonstrate how rustls's design scales down, and allows people to make quite small client/servers executables in exchange for limited TLS compatibility. Unfortunately that wasn't defended so broke a while back.

### DIFF
--- a/examples/tests/limitedclient.rs
+++ b/examples/tests/limitedclient.rs
@@ -1,0 +1,61 @@
+//! Check that `limitedclient` is successful in its goal of not linking in any
+//! AES code.
+//!
+//! We also check `simpleclient` (which includes everything) as a baseline to
+//! detect errors in the test code.
+
+// Don't assume binutils are available everywhere, or that `nm` has a
+// portable interface.
+#![cfg(target_os = "linux")]
+
+use std::process::Command;
+
+#[test]
+fn simpleclient_contains_aes_symbols() {
+    assert!(count_aes_symbols_in_executable(env!("CARGO_BIN_EXE_simpleclient")) > 0);
+}
+
+#[test]
+fn limitedclient_does_not_contain_aes_symbols() {
+    let limitedclient = env!("CARGO_BIN_EXE_limitedclient");
+    if fips_mode(limitedclient) {
+        println!("FIPS mode includes the entirety of the fipsmodule due to dynamic linking");
+        return;
+    }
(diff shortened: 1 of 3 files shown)


examples/tests/limitedclient.rs
```diff
@@ -0,0 +1,53 @@
...
+    symbols_in_executable(exe)
+        .lines()
+        .any(|sym| sym.starts_with("aws_lc_fips_"))
+}
+
+fn count_aes_symbols_in_executable(exe: &str) -> usize {
+    symbols_in_executable(exe)
+        .lines()
+        .inspect(|sym| println!("candidate symbol {sym:?}"))
+        .filter(is_aws_lc_rs_aes)
```
cpu: Was there a reason you wanted to avoid writing `is_aws_lc_rs_aes` as a closure? It feels pretty small and with only one call-site maybe not worth being a top-level fn.

examples/tests/limitedclient.rs
```diff
@@ -0,0 +1,53 @@
+// Don't assume binutils are available everywhere, or that `nm`
+// has a portable interface.
+#![cfg(target_os = "linux")]
```
cpu: Maybe also worth a quick description of the high-level goal of the test w.r.t `limitedclient` as a rustdoc module comment?

examples/tests/limitedclient.rs
ctz: originally i was thinking that this test had to also work with the ring provider, but it doesn't

examples/tests/limitedclient.rs
```diff
@@ -0,0 +1,55 @@
...
+        .any(|sym| sym.starts_with("aws_lc_fips_"))
+}
+
+fn count_aes_symbols_in_executable(exe: &str) -> usize {
+    symbols_in_executable(exe)
+        .lines()
+        .inspect(|sym| println!("candidate symbol {sym:?}"))
+        .filter(|sym| sym.starts_with("aws_lc_") && sym.ends_with("_EVP_aead_aes_128_gcm_tls13"))
+        .inspect(|sym| println!("found aes symbol {sym:?}"))
+        .count()
```
djc: IMO this would be clearer as a `for ln in symbols_in_executable(exe).lines()` loop.

djc (approved):

PR #2574, opened 2025-07-24
Prepare to delete version-selection API

ctz:
This PR withdraws use of version selection APIs from tests and examples.

The first commit introduces a new API that alters a `CryptoProvider` to restrict it to specific protocol versions. Not super happy with the naming of this.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -53,7 +53,7 @@ use rustls::{
     AlertDescription, CertificateCompressionAlgorithm, CertificateError, Connection,
     DigitallySignedStruct, DistinguishedName, Error, HandshakeKind, InvalidMessage, NamedGroup,
     PeerIncompatible, PeerMisbehaved, ProtocolVersion, RootCertStore, Side, SignatureAlgorithm,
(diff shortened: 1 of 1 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -304,6 +304,32 @@ See the documentation of the CryptoProvider type for more information.
             && secure_random.fips()
             && key_provider.fips()
     }
+
+    /// Return a new `CryptoProvider` that only supports TLS1.3.
+    pub fn with_only_tls13(self) -> Self {
```
djc, line 309: This seems suboptimal because I suppose we still ship the other versions cipher suites etc in the binary, right? Should we have separate `crypto::aws_lc_rs::tls12_provider()` and `crypto::aws_lc_rs::tls13_provider()` functions instead?

ctz: Very good point. I think once we do #2401 we should end up with consts for `DEFAULT_PROVIDER`, `JUST_TLS12_PROVIDER`, `JUST_TLS13_PROVIDER` (or whatever) so I think this function is a stop-gap to support the work on the version API.

Issue #2400, opened 2025-03-25
Evaluate whether CryptoProvider should become a trait


djc:
This would enable additional flexibility in the future (allow us to add trait methods with default impls).

As a type (in its current definition), it benefits from update syntax, which allows downstream implementations to reuse most of an existing implementation and replace a part of it.

However, even if we have a trait, we could also provide a "default" type that can still be used like that.

cpu:
Maybe worth going back and linking to some of the earlier discussion on this topic. IIRC we landed the `CryptoProvider` as a trait initially, and then changed it to a struct afterwards.

djc:
That's right:

- https://github.com/rustls/rustls/pull/1628

djc:
I seem to remember there was some motivation around the QUIC work from around #1798.

zh-jq-b:
How about https://github.com/rustls/rustls/issues/1876 and https://github.com/rustls/rustls/issues/1877 ?

fenollp:
It seems this is missing from the list:
* Renamed feature `logging` to `log`: https://github.com/rustls/rustls/commit/192ab85a5c2c865b9cca9b2a8c987de33a102667
* Dropped feature `aws_lc_rs`: https://github.com/rustls/rustls/commit/afa22bb8e105c67b5bf4927e8a04de50cf583b8d
* Dropped provider from default features: https://github.com/rustls/rustls/commit/30f3753a70c3b138841675b6381c0b63a3005414

Could `0.24.0-dev.0` be published? It'd help downstream dependents merge a few changes to `main`.
I'm not sure how the usual development cycle works.

EDIT: others are also in my place: https://github.com/rustls/tokio-rustls/pull/118 ;)

djc:
Why do you want to depend on rustls 0.24 now?

This is a tracking issue for work that we don't want to forget, not for work that is already done.

ctz:
I think a prerelease may be a bit premature. For example, one outcome for #2395 could be withdrawing all crate features that control the built-in providers, and shipping that code as separate crates. If we go in that direction, it would waste work in the meantime to align upstream crate features with ours.

fenollp:
I forked a few of my dependencies to be able to use https://github.com/rustls/rustls/blob/8b96a51666c863605d1be1d9f71183cfd483552c/rustls/Cargo.toml#L17 
* https://github.com/seanmonstar/reqwest/pull/2785
* https://github.com/rustls/tokio-rustls/pull/119
[...]

Issue #1723, opened 2024-01-04
reduce bifurcation of library internals by using low-level API in high-level API


cpu:
**Is your feature request related to a problem? Please describe.**

The gradual introduction of a lower-level API that grants callers more control over allocation has bifurcated the internal structure of Rustls somewhat. This can be problematic with respect to test coverage and maintenance: e.g. most of our existing test coverage is using the higher-level API.

**Describe the solution you'd like**

We should make a concerted effort to implement the higher-level API on top of the lower-level API, reducing the bifurcation and ensuring that our existing test coverage also applies to the lower level API without needing to duplicate test cases.

cpu:
I've spent some time thinking about this. Here's my thought process so far. @ctz, @djc does this match your expectations?

It seems like [`ClientConnection`](https://github.com/rustls/rustls/blob/582516e241c8330275a2149304c322e4b80baece/rustls/src/client/client_conn.rs#L546-L548) and [`ServerConnection`](https://github.com/rustls/rustls/blob/582516e241c8330275a2149304c322e4b80baece/rustls/src/server/server_conn.rs#L456-L458) should have their [`ConnectionCommon<Data>`](https://github.com/rustls/rustls/blob/582516e241c8330275a2149304c322e4b80baece/rustls/src/conn.rs#L347-L352) member replaced with an adapter layer that holds an [`UnbufferedConnectionCommon<Data>`](https://github.com/rustls/rustls/blob/582516e241c8330275a2149304c322e4b80baece/rustls/src/conn.rs#L694-L698) and additional buffers (`ChunkVecBuffer` instances?) that can augment the `UnbufferedConnectionCommon` state for use when handling each `UnbufferedStatus` we encounter while processing.

I'm imagining that the loop over `UnbufferedConnectionCommon<Data>`'s `process_tls_records_common()` results (e.g. [this code](https://github.com/rustls/rustls/blob/582516e241c8330275a2149304c322e4b80baece/examples/src/bin/unbuffered-client.rs#L60-L191) in the client example) needs to be done in the adapter layer's `process_new_packets()` fn, but using a single call to `process_tls_records_common()` instead of in a loop `while open_connection { ... }`.  Each call would handle one  `UnbufferedStatus` state, discarding from the adapter's internal buffer at the end.

[...]

PR #2575, opened 2025-07-25
Fix: Do not try to call deframer on junk data

schreter:
After receiving close notify, the TLS stream ends. The receive buffer can contain additional junk data received past close notify record, which obviously cannot be interpreted anymore.

However, `UnbufferedConnectionCommon::process_tls_records_common()` tried to interpret this junk data.

Add a check for `has_received_close_notify` to prevent calling the deframer on junk data. Also update the test to test also with a longer junk data (original junk data didn't trigger the bug).

### DIFF
--- a/rustls/src/conn/unbuffered.rs
+++ b/rustls/src/conn/unbuffered.rs
@@ -80,7 +80,13 @@ impl<Data> UnbufferedConnectionCommon<Data> {
                 );
             }
 
-            let deframer_output =
+            let deframer_output = if self
+                .core
+                .common_state
+                .has_received_close_notify
+            {
+                None
+            } else {
                 match self
                     .core
                     .deframe(None, buffer.filled_mut(), &mut buffer_progress)
@@ -93,7 +99,8 @@ impl<Data> UnbufferedConnectionCommon<Data> {
                         };
                     }
                     Ok(r) => r,
-                };
+                }
+            };
 
             if let Some(msg) = deframer_output {
                 let mut state =
--- a/rustls/tests/unbuffered.rs
+++ b/rustls/tests/unbuffered.rs
@@ -547,40 +547,49 @@ fn full_closure_server_to_client() {
 #[test]
 fn junk_after_close_notify_received() {
     // cf. test_junk_after_close_notify_received in api.rs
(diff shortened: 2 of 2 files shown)


schreter:
@djc This addresses the deframer processing junk data as mentioned in #959.

ctz (approved):
Thanks!

For other reviewers: the equivalent buffered behaviour is here https://github.com/rustls/rustls/blob/76cc2cabbe8b47dc3399115e1a8d4b5cc74681c8/rustls/src/conn.rs#L712-L714 and tested by `test_read_tls_artificial_eof_after_close_notify`. So this difference is another thing potentially addressed by #1723.

The idea of handling alerts in the state machine I think is a good one, but beware that the state machine operates on a sequence of whole TLS messages, so it has a limited effect on layers above that (and whether to read data at all is _several_ layers above that). The other thing to note here is that every state in every state machine  would need to handle alerts in exactly the same way; that commonality is why it doesn't currently happen in this way. But that is nothing some additional layering (eg, make `State::handle` concrete and handle alerts, before passing off to the state-specific code)

cpu (approved):

schreter:
@ctz & @cpu Thanks for the review and merging the fix.

I just have one question: what is the strategy to get fixes to `crates.io`? There is a document about how to make a release, but I didn't find details who/when is going to do it. I guess I'll simply wait until it gets picked up by the next release :-), since we are not under time pressure to consume the fix in our project.

cpu:
> I just have one question: what is the strategy to get fixes to crates.io? 

Right now `main` has breaking changes as we work towards [a 0.24 release](https://github.com/rustls/rustls/issues/2400). I think the best way to get this fix onto `crates.io` in a timely manner is for us to backport it to the [`rel-0.23`](https://github.com/rustls/rustls/tree/rel-0.23) branch so we could include it in a 0.23.30 release.

I kicked the process off over here: https://github.com/rustls/rustls/pull/2576

cpu:
> I just have one question: what is the strategy to get fixes to crates.io?

The fix is now available on crates.io as part of [Rustls v0.23.30](https://github.com/rustls/rustls/releases/tag/v%2F0.23.30). 

Thanks!

PR #2579, opened 2025-07-26
ML-DSA signing keys

### LINKED ISSUE #2577: Can't seem to load ML-DSA private key with rustls-post-quantum 0.2.3
**Checklist**
* [X] I've searched the issue tracker for similar bugs.

**Describe the bug**
Hi! Many thanks for adding PQC DSA beyond the PQC KEM support that rustls-post-quantum already accomplished, both much appreciated and I understand that this new addition came clearly labeled "add _unstable_ ML-DSA support" but I can't seem to get it to load any ML-DSA private key despite trying:

- mine generated with OpenSSL 3.5.1 (which has native ML-DSA support since 3.5.0)
- mine generated on another computer with OpenSSL 3.4.1 and oqsprovider 0.9.0
- the one corresponding to the test certificate linked [in the PR thread](https://github.com/rustls/webpki/pull/348#issuecomment-2984029964)

The error I am stuck at is "failed to parse private key as RSA, ECDSA, or EdDSA" which exists in 4 different places in rusts 0.23.29 source but I managed to track it down specifically to [aws_lc_rs/sign.rs#L36](https://github.com/rustls/rustls/blob/v/0.23.29/rustls/src/crypto/aws_lc_rs/sign.rs#L36) which is all about RSA, ECDSA and EdDSA indeed... If I understand its Cargo.toml correctly, rustls-post-quantum (with "aws-lc-rs-unstable" feature enabled) is not at all a fork of rustls (and "aws-lc-rs" feature) but the former actually depend on the latest stable release of the latter, but is it normal that loading my PQC creds happens in the non-PQC part? 🙂 

I mean, it is entirely possible that I misunderstood the "Experimental support for ML-DSA signature schemes is now available via rustls-post-quantum 0.2.3" announcement: perhaps for now the support is only for validating a certificate presented by a remote peer as opposed to loading your own ML-DSA files like I am trying to do?

**To Reproduce**
I guess the best way would be to try to add Mutual TLS support in the existing rustls-post-quantum/examples/client.rs or to implement a corresponding "server.rs" but what I have been trying so far is simply with rocket.rs (from main branch, not the latest 0.5.1 release which still tries to parse keys by itself even if the crypto provider is overridden) with the very first instruction being `rustls_post_quantum::provider().install_default().unwrap();` as per rustls documentation or `let _ = rustls_post_quantum::provider().install_default();` as inferred from rocket documentation.

PS: if like me you/someone tries to patch rustls to help figure out what is going on in rustls-post-quantum, beware that unpatched rustls can still be brought in by rocket and tokio-rustls for example.

**Applicable Version(s)**
rustls-post-quantum 0.2.3
rustls 0.23.29
https://github.com/rwf2/Rocket/tree/master since May the 4th

**Expected behavior**
Same as always with TLS the only difference being credentials of the ML-DSA type instead of RSA, ECDSA or EdDSA. Further explained in the "Load/validate and parse Quantum certificates generated by oqsprovider" issue #2417 with now the addition of files generated by OpenSSL>=3.5.0 itself.

**Additional context**
Having edited rustls-post-quantum/src/lib.rs I am quite confident that I successfully forced rocket to use rustls-post-quantum instead of any other crypto provider, but private key parsing still seems to go from rustls-post-quantum into non-PQC rustls code; maybe more ML-DSA patches are needed in rustls and/or rustls-pemfile?

djc:
Fixes #2577.

## Proposed release notes

* Support using ML-DSA keys for signing when the `aws-lc-rs-unstable` feature is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2735,7 +2735,7 @@ dependencies = [
 
 [[package]]
 name = "rustls-post-quantum"
-version = "0.2.3"
+version = "0.2.4"
 dependencies = [
  "aws-lc-rs",
  "criterion",
--- a/rustls-post-quantum/Cargo.toml
+++ b/rustls-post-quantum/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-post-quantum"
-version = "0.2.3"
+version = "0.2.4"
 edition = "2021"
 rust-version = "1.79"
 license = "Apache-2.0 OR ISC OR MIT"


rustls/src/crypto/signer.rs
```diff
@@ -243,8 +243,8 @@ impl CertifiedKey {
     }
 }
 
-#[cfg_attr(not(any(feature = "aws-lc-rs", feature = "ring")), allow(dead_code))]
-pub(crate) fn public_key_to_spki(
+/// Convert a public key and algorithm identifier into a SubjectPublicKeyInfo.
```
ctz:
```suggestion
/// Convert a public key and algorithm identifier into a `SubjectPublicKeyInfo`.
///
/// In the returned encoding, `alg_id` is used as the `algorithm` field, and `public_key`
/// is wrapped inside an ASN.1 `BIT STRING` and then used as the `subjectPublicKey`
/// field.
```

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
     {
         provider.signature_verification_algorithms = SUPPORTED_SIG_ALGS;
+        provider.key_provider = &key_provider::PqAwsLcRs;
     }
     provider
 }
 
+#[cfg(feature = "aws-lc-rs-unstable")]
+mod key_provider {
+    #[cfg(feature = "aws-lc-rs-unstable")]
```
cpu: If the entire `mod` is behind a `cfg(feature = "aws-lc-rs-unstable")` do these handful of imports need their own duplicated `cfg` guard?

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+
+            for &alg in ML_DSA_SIGNING_ALGS {
+                let key_pair = match &key_der {
+                    PrivateKeyDer::Pkcs8(pkcs8) => {
+                        match PqdsaKeyPair::from_pkcs8(alg, pkcs8.secret_pkcs8_der()) {
+                            Ok(key_pair) => key_pair,
+                            Err(_) => continue,
+                        }
+                    }
+                    // To do: support `PqdsaKeyPair::from_raw_private_key()`?
```
cpu: Can you write this "TODO:"? I think that form is most common vs "to do" and some IDEs will pick up as a special annotation.

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+
+            Some(Box::new(PqdsaSigner {
+                key: self.inner.clone(),
+                kind: self.kind,
+            }))
+        }
+
+        fn public_key(&self) -> Option<SubjectPublicKeyInfoDer<'_>> {
+            Some(public_key_to_spki(
+                self.kind.alg_id(),
```
cpu: This will want to be `&self.kind.alg_id()` once you take rustls 0.23.31 and add the import for `public_key_to_spki()`.

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+                return Ok(Arc::new(PqdsaSigningKey {
+                    kind: if alg == &ML_DSA_44_SIGNING {
+                        PqdsaKeyKind::MlDsa44
+                    } else if alg == &ML_DSA_65_SIGNING {
+                        PqdsaKeyKind::MlDsa65
+                    } else if alg == &ML_DSA_87_SIGNING {
+                        PqdsaKeyKind::MlDsa87
+                    } else {
+                        unreachable!("unexpected ML-DSA signing algorithm")
+                    },
```
cpu: I think this can be made nicer if you flip things around so that `ML_DSA_SIGNING_ALGS` is a slice of `PqdsaKeyKind`, and then have a `impl From<PqdsaKeyKind> for &PqdsaSigningAlgorithm` to use on L66. It avoids this `unreachable!()` and moves this conversion logic closer to the enum where I think it fits better.

Something like this: https://github.com/rustls/rustls/commit/b231c9e69ead1ad478919b96b25b0d9dc7993891

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+                }));
+            }
+
+            if let Ok(key) = sign::any_supported_type(&key_der) {
+                return Ok(key);
+            }
+
+            Err(Error::General(
+                "failed to parse private key as ML-DSA, RSA, ECDSA, or EdDSA".into(),
+            ))
```
cpu: nit: would a `match` be more idiomatic here?

```rust
            match sign::any_supported_type(&key_der) {
                Ok(key) => Ok(key),
                Err(_) => Err(Error::General(
                    "failed to parse private key as ML-DSA, RSA, ECDSA, or EdDSA".into(),
                )),
            }
```

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+
+    struct PqdsaSigningKey {
+        kind: PqdsaKeyKind,
+        inner: Arc<PqdsaKeyPair>,
+    }
+
+    impl SigningKey for PqdsaSigningKey {
+        fn choose_scheme(&self, offered: &[SignatureScheme]) -> Option<Box<dyn Signer>> {
+            let scheme = self.kind.scheme();
+            if !offered.contains(&scheme) {
```
cpu: nit: inline `scheme`?

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+
+        fn public_key(&self) -> Option<SubjectPublicKeyInfoDer<'_>> {
+            Some(public_key_to_spki(
+                self.kind.alg_id(),
+                self.inner.public_key(),
+            ))
+        }
+
+        fn algorithm(&self) -> SignatureAlgorithm {
+            SignatureAlgorithm::Unknown(255)
```
cpu, line 112: Is there something to cite for the use of `255` here?

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,184 @@ pub fn provider() -> CryptoProvider {
...
+
+    impl Debug for PqdsaSigner {
+        fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+            f.debug_struct("PqdsaSigner")
+                .field("scheme", &self.kind.scheme())
+                .finish_non_exhaustive()
+        }
+    }
+
+    #[cfg(feature = "aws-lc-rs-unstable")]
```
cpu: this `cfg` can fall away because it's inside the `mod` too, like the import ones.

cpu (commented):
FWIW I got this branch building locally to fiddle with review feedback using this small patch: https://github.com/rustls/rustls/commit/7ae2f2b63434bc73f1a70431b236d79f323cea78

What kind of test coverage did you have in-mind for this to be ready to go? There's some low-level ML-DSA test vectors recently added to Wycheproof (https://github.com/C2SP/wycheproof/pull/146) but I'm not sure how well they'll map into this context.

djc:
I was thinking we could do a handshake with a freshly generated key? I guess that would require rcgen support maybe.

rustls-post-quantum/src/lib.rs
djc: They don't, cleaned it up.

rustls-post-quantum/src/lib.rs
djc: Done.

rustls-post-quantum/src/lib.rs
djc: Fixed.

rustls-post-quantum/src/lib.rs
djc: Ahh, good idea!

rustls-post-quantum/src/lib.rs
djc: Yup.

rustls-post-quantum/src/lib.rs
djc: Done.

rustls-post-quantum/src/lib.rs
djc: Added some comments.

rustls-post-quantum/src/lib.rs
djc: Done.

rustls/src/crypto/signer.rs
djc: This is now #2662.

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,181 @@ pub fn provider() -> CryptoProvider {
...
+                            Err(_) => continue,
+                        }
+                    }
+                    // TODO: support `PqdsaKeyPair::from_raw_private_key()`?
+                    _ => {
+                        return Err(Error::General(
+                            "ML-DSA keys must be provided in PKCS#8 format".into(),
+                        ));
+                    }
+                };
```
ctz: AIUI this will early-reject (eg) a PKCS#1 RSA key. Maybe we should do the `sign::any_supported_type` thing first?

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,181 @@ pub fn provider() -> CryptoProvider {
...
+
+                return Ok(Arc::new(PqdsaSigningKey {
+                    kind,
+                    inner: Arc::new(key_pair),
+                }));
+            }
+
+            match sign::any_supported_type(&key_der) {
```
ctz, line 72: It might be clearer to dispatch to `rustls::crypto::aws_lc_rs::default_provider().key_provider` here?

rustls-post-quantum/src/lib.rs
```diff
@@ -25,10 +25,181 @@ pub fn provider() -> CryptoProvider {
...
+            match self {
+                Self::MlDsa44 => alg_id::ML_DSA_44,
+                Self::MlDsa65 => alg_id::ML_DSA_65,
+                Self::MlDsa87 => alg_id::ML_DSA_87,
+            }
+        }
+    }
+}
+
 /// Keep in sync with the `SUPPORTED_SIG_ALGS` in `rustls::crypto::aws_lc_rs`.
```
ctz, line 196: Missing the recent additions to the `SignatureScheme::ECDSA_NISTP521_SHA512` mapping.

rustls-post-quantum/src/lib.rs
ctz: (Probably not, actually.)

rustls-post-quantum/src/lib.rs
djc: Ah, good catch -- and I think it would also have failed with non-ML-DSA PKCS#8 keys. Restructured.

rustls-post-quantum/src/lib.rs
djc: Ah, added those.

rustls-post-quantum/src/lib.rs
```diff
@@ -280,3 +280,109 @@ static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms = WebPkiSupportedAlgorithms
...
+        CertificateParams, CertifiedIssuer, ExtendedKeyUsagePurpose, IsCa, KeyPair, KeyUsagePurpose,
+    };
+    use rustls::pki_types::PrivateKeyDer;
+    use rustls::{
+        ClientConfig, ClientConnection, ConnectionCommon, RootCertStore, ServerConfig,
+        ServerConnection, SideData,
+    };
+
+    #[test]
+    fn ml_kem() {
```
ctz:
```suggestion
    fn ml_dsa() {
```

djc:
* Published rustls-post-quantum v0.2.4 at registry `crates-io`
* [new tag]           rustls-post-quantum-v/0.2.4 -> rustls-post-quantum-v/0.2.4
* [Release notes](https://github.com/rustls/rustls/releases/tag/rustls-post-quantum-v%2F0.2.4)

PR #2576, opened 2025-07-26
0.23.30 release prep & 2575 backport

cpu:
This is a backport of https://github.com/rustls/rustls/pull/2575 to `rel-0.23` to prepare a 0.23.30 release with the fix.

### Proposed release notes

* Fixes a bug with the unbuffered connection API that could result in deframing junk data after a close notify alert was received. 
* Updates `Connection::complete_io()` to yield a `WouldBlock` error when both read/write operations are blocked.

PR #2580, opened 2025-07-26
sign: make public_key_to_spki() public

djc:
In order to support

- #2579

Can we add this to

- #2576 

?

Do we want to tweak the name before we make it public?


rustls/src/crypto/signer.rs
```diff
@@ -207,8 +207,8 @@ impl CertifiedKey {
     }
 }
 
-#[cfg_attr(not(any(feature = "aws_lc_rs", feature = "ring")), allow(dead_code))]
-pub(crate) fn public_key_to_spki(
+/// Convert a public key and algorithm identifier into a SubjectPublicKeyInfo.
```
cpu:
```suggestion
/// Convert a public key and algorithm identifier into `SubjectPubicKeyInfoDer`.
```

ctz: careful :)
```suggestion
/// Convert a public key and algorithm identifier into `SubjectPublicKeyInfoDer`.
```

Issue #2572, opened 2025-07-23
There is no way to rety the ECH rejected hanshake


matszczygiel:
**Checklist**
* [ ] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
Currently, there seems to be no way to retry the rejected ECH handshake in case the server responds with proper ECH config. In such a case, the `Error::PeerIncompatible(PeerIncompatible::ServerRejectedEncryptedClientHello(configs))` error is returned. However, there's no way to construct a new TLS connection since `configs` use the private type `EchConfigPayload`. Ideally, there should be a way to construct `EchConfig` from it, but the functionality seems to be private.

**Describe the solution you'd like**
The solution would be to either implement the retry functionality within the handshake logic, or to expose `EchConfigPayload` type as public with its conversion to `EchConfig`.

**Describe alternatives you've considered**
Currently there seems to be none.

**Additional context**
None.

cpu:
Thanks for pointing out this oversight 👍 

I think it makes sense to offer a way to construct an `EchMode` from a `EchConfigPayload` sourced from the `Error::PeerIncompatible` error and `&[&'static dyn Hpke],`. I don't think we should do the retry internal to Rustls.

Exporting `EchConfigPayload` might be the best way. If we use the `pki_types` repr we'll have to serialize into the error, and then deserialize on `EchMode` construction and we might as well save the work since we had to parse the config from server.

cpu:
> Describe alternatives you've considered
> Currently there seems to be none.

It's not ideal (or something I recommend depending on) but if you're desperate for a short term workaround you can use some bits from `rustls::internal::msgs::handshake::*` to get the `EchConfigPayload` to use with `rustls::internal::msgs::codec::Codec` to serialize to TLS encoded data you can wrap with `pki_types::EchConfigListBytes` for the pub `EchConfig::new()` constructor.

On that subject, I think the test usages we have for these APIs won't motivate keeping them in `rustls::internal` for long. There's ~3 test usages:

* There's `rustls/tests/api.rs` where we are constructing some bogus data that could probably be replaced by static bytes for each supported suite.
* There's `rustls/tests/ech.rs`, but these feel more like unit tests than integration tests and could be moved internal to the messages module.
[...]

PR #2585, opened 2025-07-28
Support building ECH config from rejection err w/ retry configs, remove more internal API exports

cpu:
As discussed in #2572 we had a number of internal message types being exported through the unstable `rustls::internal::*` namespace, but they were only used in places where better alternatives exist. This branch cleans those up, removing a number of internal exports.

At the same time, we want a way to allow users to build an `EchConfig` from retry configurations delivered by a server when it rejected an initial ECH negotiation. To do this we promote the `PeerIncompatible::ServerRejectedEncryptedClientHello` variant to a top-level error variant, `Error::ServerRejectedEncryptedClientHello(EncryptedClientHelloRejected)`. The inner `EncryptedClientHelloRejected` type can be used with `EchConfig::for_retry()` to construct a new `EchConfig` by picking a compatible retry config (or returning an error). The type is otherwise opaque, except for its ability to also provide the TLS encoded `EchConfigListBytes` of the retry configs if desired.

Resolves https://github.com/rustls/rustls/issues/2572


djc:
Those names are pretty wordy. Maybe skip the Server prefix (clients don't get to reject ECH presumably?) and shorten EncryptedClientHello to Ech as is already the case in other public API? (Especially for the inner type?)

cpu:
> Those names are pretty wordy. Maybe skip the Server prefix (clients don't get to reject ECH presumably?) and shorten EncryptedClientHello to Ech as is already the case in other public API? (Especially for the inner type?)

Good idea. :+1: I also added a few more helpful cross-links to the rust docs.

rustls/src/msgs/handshake.rs
```diff
@@ -189,7 +189,7 @@ impl AsRef<[u8]> for SessionId {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct UnknownExtension {
+pub(crate) struct UnknownExtension {
```
ctz, line 192: :chefskiss:

ctz (approved):
Very nice. Appreciate the continued reductions in `rustls::internals`.

cpu:
Since we're not cutting a release from `main` imminently I'm going to merge this. I'd be happy to iterate further if there's any new feedback.

PR #2591, opened 2025-08-04
Be precise about the requirements of `hmac::Tag` type

ctz:
This adds a counterpart to the `hmac::Tag` type for values that are published on the wire. This saves a minor bit of performance because zeroization contains optimization barriers that make it more expensive than just a memset.

### DIFF
--- a/rustls/src/tls12/mod.rs
+++ b/rustls/src/tls12/mod.rs
@@ -3,7 +3,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt;
 
-use zeroize::Zeroize;
+use zeroize::{Zeroize, Zeroizing};
 
 use crate::common_state::{CommonState, Side};
 use crate::conn::ConnectionRandoms;
@@ -207,7 +207,7 @@ impl ConnectionSecrets {
         )
     }
 
-    fn make_key_block(&self) -> Vec<u8> {
+    fn make_key_block(&self) -> Zeroizing<Vec<u8>> {
         let shape = self.suite.aead_alg.key_block_shape();
 
         let len = (shape.enc_key_len + shape.fixed_iv_len) * 2 + shape.explicit_nonce_len;
@@ -220,7 +220,7 @@ impl ConnectionSecrets {
         self.master_secret_prf
             .prf(&mut out, b"key expansion", &randoms);
 
-        out
+        Zeroizing::new(out)
     }
 
     pub(crate) fn suite(&self) -> &'static Tls12CipherSuite {


rustls/src/crypto/hmac.rs
```diff
@@ -45,11 +46,38 @@ impl Tag {
...
     }
 }
 
 impl AsRef<[u8]> for Tag {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+/// A HMAC tag, stored as a value, that is non-secret.
```
djc: Maybe add some context on what exactly it means for a tag to be public, and also why it matters/what's different about it?

rustls/src/tls12/mod.rs
```diff
@@ -300,6 +301,21 @@ impl Drop for ConnectionSecrets {
     }
 }
 
+struct KeyBlock(Vec<u8>);
```
djc: Could this just be `zeroize::Zeroizing`?

djc (approved):
Nice!

cpu:
@ctz Do you think this is ready to revive or is there a blocker?

ctz:
> @ctz Do you think this is ready to revive or is there a blocker?

I will get back to this soon.

rustls/src/crypto/hmac.rs
```diff
@@ -20,23 +20,30 @@ pub trait Hmac: Send + Sync {
...
-        tag
+        Self(PublicTag::new(bytes))
+    }
+
+    /// Declare this tag is public.
+    ///
+    /// Uses of this function should explain why this tag is public.
+    pub(crate) fn into_public(self) -> PublicTag {
+        let pt = self.0.clone();
+        core::mem::forget(self);
```
djc: Nit: maybe import `mem`? Maybe `pt` -> `public`?

rustls/src/crypto/hmac.rs
```diff
@@ -45,11 +52,45 @@ impl Tag {
...
     }
 }
 
 impl AsRef<[u8]> for Tag {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+/// A HMAC tag, stored as a value, that is non-secret.
```
cpu: bike-shedding: What about putting the most relevant part in the headline up-front like:

> A non-secret HMAC tag, stored as a value.

rustls/src/crypto/hmac.rs
```diff
@@ -20,23 +20,30 @@ pub trait Hmac: Send + Sync {
 }
 
 /// A HMAC tag, stored as a value.
+///
```
cpu, line 24: Thinking of my latter feedback for the non-secret headline docstring, perhaps this one should be:

> A secret HMAC tag, stored as a value.

rustls/src/crypto/hmac.rs
```diff
@@ -45,11 +52,45 @@ impl Tag {
...
 
 impl AsRef<[u8]> for Tag {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+/// A HMAC tag, stored as a value, that is non-secret.
+///
+/// A value of this type is not zeroized on drop.
```
cpu: nit: maybe `**not**`? Might as well lean on Rustdoc having nice markdown.

rustls/src/crypto/hmac.rs
```diff
@@ -45,11 +52,45 @@ impl Tag {
...
+///
+/// A tag is "public" if it is published on the wire. For example, the
```
cpu: bike-shedding a bit to try and connect this up to the private tag:

> A tag is "public" if it is published on the wire, as opposed to being used as key material. For example ....

rustls/src/crypto/hmac.rs
```diff
@@ -45,11 +52,45 @@ impl Tag {
...
+    used: usize,
+}
+
+impl PublicTag {
+    /// Build a tag by copying a byte slice.
+    ///
+    /// The slice can be up to [`Tag::MAX_LEN`] bytes in length.
+    pub(crate) fn new(bytes: &[u8]) -> Self {
+        let mut tag = Self {
+            buf: [0u8; Tag::MAX_LEN],
```
cpu, line 87: nit: `0` vs `0u8`? 

_(Thinking along the lines of how we avoid explicit type annotations elsewhere, but know this is super nit-picky :laughing:)_

rustls/src/crypto/tls12.rs
```diff
@@ -92,17 +92,28 @@ pub trait PrfSecret: Send + Sync {
 }
 
 pub(crate) fn prf(out: &mut [u8], hmac_key: &dyn hmac::Key, label: &[u8], seed: &[u8]) {
+    // Whether `out` is public or secret is up to the caller: the caller must zeroize
+    // `out` if it is secret, irrespective of whether the intermediate values are
+    // zeroized in this function.  Therefore there is no reason to zeroize them at this
+    // level.
```
cpu: Sorry, I'm not sure I follow here. Can you help me out?

If the caller is extracting into `out` for the purpose of a secret usage, they can/should zeroize `out` once done, but this function isn't zeroizing the intermediate values and the caller can't either. I think what I'm missing is an assertion that the intermediates aren't helpful for cryptanalysis.

rustls/src/tls12/mod.rs
```diff
@@ -231,18 +231,18 @@ impl ConnectionSecrets {
         &self.master_secret
     }
 
-    fn make_verify_data(&self, handshake_hash: &hash::Output, label: &[u8]) -> Vec<u8> {
-        let mut out = vec![0u8; 12];
+    fn make_verify_data(&self, handshake_hash: &hash::Output, label: &[u8]) -> [u8; 12] {
+        let mut out = [0u8; 12];
```
cpu, line 235: nit: another `0u8` to `0` if you don't hate that feedback.

cpu (approved):

rustls/src/crypto/hmac.rs
ctz: Oooh this is a bit of a style question isn't it. I prefer to write the suffix to help readers, even though in this case the actual type is a few lines above. And that's not a consistent approach either, because obviously we're relying on inference in zillions of places outside this.

rustls/src/crypto/tls12.rs
ctz: Thanks -- yeah, you're right. I have dropped this commit. I think the way to achieve this is to have a separate `fn prf` whose output is public and whose intermediates can be too. But doing that would touch the trait interface layer and I think that probably isn't worth it.

rustls/src/crypto/hmac.rs
djc: Yeah, I tend to prefer relying on inference where possible because I think fewer tokens will generally be easier to consume.

ctz:
> [documentation / Generate pre-release documentation (pull_request)](https://github.com/rustls/rustls/actions/runs/17234862416/job/48897271065?pr=2591)Failing after 2m

Azure's networking being deserving of its reputation

rustls/src/crypto/hmac.rs
ctz: I think if we go in that direction we'd need some tooling to help with consistency. Though we don't have consistency right now (there are about 110 places with the suffix, and about 50 without currently.)

cpu: >  (there are about 110 places with the suffix, and about 50 without currently.)

Seems very reasonable to leave this as-is and consider a tidying pass if/when someone has time to make it enforceable in CI.

PR #2592, opened 2025-08-06
server: implement `Debug` for `Acceptor`

lucab:
This adds a partial `Debug` implementation for `Acceptor`, in order to make it easier to embed and debug-print it in external consumers (e.g. `tokio-rustls::LazyConfigAcceptor`).

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -876,6 +876,17 @@ mod connection {
         }
     }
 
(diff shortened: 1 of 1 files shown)


lucab:
This was driven by https://github.com/rustls/tokio-rustls/pull/120, but then it was decided to do the `tokio-rustls` change in a different way, so this is not required anymore. Closing.

Issue #1877, opened 2024-03-27
Feature Request: Avoid panicking when ring and aws_lc_rs are both specified


Alvenix:
**Checklist**
* [x] I've searched the issue tracker for similar requests

I know the current design for crypto providers is probably well discussed but I wanted to add one data point against the current behavior for reference. This issue is mostly rephrasing my discussion at [1876](https://github.com/rustls/rustls/issues/1876#issuecomment-2023403756):

**Is your feature request related to a problem? Please describe.**
When I accidentally enabled these two features (directly or indirectly through a crate) I had few panics that I had to debug. I consider myself somewhat familiar with Rust and (and cargo features) so debugging this was easy although annoying. However, this could be a little road blocker for people less familiar with Rust.

**Describe the solution you'd like**
Changing the internal function from_crate_features() to choose either `ring` or `aws_lc_rs` CryptoProvider when both features are specified, instead of returning None. In addition, removing the related panic. In addition, specify the behavior so it is no longer ambiguous.

**Describe alternatives you've considered**
Setup the default crypto provider on main. However, sometimes when developing a library I probably shouldn't do that. I would have to document to the users, that they need to setup the crypto providers if they don't want to have a panic accidentally.

djc:
I agree that the current setup is not ideal for libraries. I think we should have gone with something like this:

```diff
diff --git a/rustls/src/crypto/mod.rs b/rustls/src/crypto/mod.rs
index 74b19ea6..9f029aed 100644
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -215,6 +215,29 @@ pub struct CryptoProvider {
 }
 
 impl CryptoProvider {
+    /// Returns the default `CryptoProvider` for this process.
+    ///
+    /// If a crypto provider has been installed using [`install_default()`], return that.
+    /// Otherwise, return the default provider for the crate features in use. If both
+    /// first-party providers are enabled, this will return the aws-lc-rs provider.
+    #[cfg(any(feature = "aws_lc_rs", feature = "ring"))]
+    pub fn installed_or_default() -> Arc<Self> {
+        match PROCESS_DEFAULT_PROVIDER.get() {
+            Some(provider) => provider.clone(),
+            None => {
+                #[cfg(feature = "aws_lc_rs")]
+                {
+                    return Arc::new(aws_lc_rs::default_provider());
+                }
+
+                #[cfg(all(feature = "ring", not(feature = "aws_lc_rs")))]
[...]

PR #2596, opened 2025-08-08
feat: CryptoProvider::default_available

### LINKED ISSUE #2595: Add fn to check whether a default `CryptoProvider` is available
**Checklist**
related to #1877

When crate features do not provide an unambiguous default provider, and no default provider has been installed by the user, several methods (e.g. `ServerConfig::builder`) panic in `get_default_or_install_from_crate_features`. There appears to be no way to check in advance whether this will panic or not

An associated function `CryptoProvider::default_available()` that returns `true` if `get_default_or_install_from_crate_features` will not panic, and false if it will panic

I read the discussion in #1877. I don't have strong opinions about whether the panicking behavior should be changed, but i would like to be able to detect the configuration issue before the panic occurs

prestwich:
Closes #2595 

Partially addresses #1877

Adds an associated fn on `CryptoProvider` that can be used as a preflight check for uses of the default `CryptoProvider` via (e.g. `ClientConfig::builder`

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -234,6 +234,53 @@ impl CryptoProvider {
         static_default::get_default()
     }
 
+    /// Checks whether a default `CryptoProvider` has been configured, or can
+    /// be configured implicitly by inferring from crate features. Returns true
+    /// if so, false otherwise.
+    ///
+    /// This can be used to check whether [`ServerConfig::builder`] or
+    /// [`ClientConfig::builder`] will panic if called.
+    ///
+    /// ```
+    /// # use rustls::crypto::CryptoProvider;
+    /// # fn _example() {
+    /// if !CryptoProvider::default_available() {
+    ///     // No provider is available. You must call
+    ///     // CryptoProvider::install_default() before using
+    ///     // the ServerConfig::builder() or ClientConfig::builder() APIs.
+    /// }
+    /// # }
+    /// ```
+    ///
+    /// [`ServerConfig::builder`]: crate::server::ServerConfig::builder
+    /// [`ClientConfig::builder`]: crate::client::ClientConfig::builder
+    pub fn default_available() -> bool {
+        if Self::get_default().is_some() {
+            return true;
+        }
+
+        #[cfg(all(
+            feature = "ring",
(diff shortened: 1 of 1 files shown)


prestwich:
oh yeah, i would definitely prefer that solution as it prevents duplicating the complex `cfg`s the way this PR currently does

cpu:
> One thing we could do instead of this is make from_crate_features() public. 

I'm supportive. We carry a similar [re-implementation downstream](
https://github.com/rustls/rustls-ffi/blob/1bf0b6cdc4b6e9f3ffba4fc5728bf87f8d989eb8/librustls/src/crypto_provider.rs#L577-L594) in rustls-ffi.

prestwich:
The combinator version of this is a little ugly due to the mismatched types `&'static Arc<CryptoProvider>` vs `CryptoProvider`

Naive approach:

```
CryptoProvider::get_default()
    .map(drop)
    .or_else(|| CryptoProvider::from_crate_features().map(drop))
    .is_some()
```

Seems more legible to write:

```
CryptoProvider::get_default().is_some() || CryptoProvider::from_crate_features().is_some() 
```

ctz (approved):

rustls/src/crypto/mod.rs
```diff
@@ -261,7 +261,23 @@ See the documentation of the CryptoProvider type for more information.
     /// providers), or specify no providers, or the feature `custom-provider` is activated.
     /// In all cases the application should explicitly specify the provider to use
     /// with [`CryptoProvider::install_default`].
-    fn from_crate_features() -> Option<Self> {
+    ///
+    ///
```
cpu: nit: I think one empty line is enough separation here

```suggestion
```

cpu (approved):

rustls/src/crypto/mod.rs
prestwich: removed in ea8b1a4

prestwich:
CI failure seems unrelated?

cpu:
> CI failure seems unrelated?

Agreed. Want to squash your branch down to 1 tidy commit and then I can bypass the failed task and merge?

prestwich:
squashed 👍

prestwich:
thanks y'all :)

PR #2599, opened 2025-08-13
Remove version selection API

ctz:
This removes the version selection parts of the config builder API: `.with_protocol_versions()`, `.with_safe_default_protocol_versions()`, etc.

A further step towards #1659.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -6,31 +6,28 @@ use pki_types::{CertificateDer, PrivateKeyDer};
 use super::client_conn::Resumption;
 use crate::builder::{ConfigBuilder, WantsVerifier};
 use crate::client::{ClientConfig, EchMode, ResolvesClientCert, handy};
+use crate::enums::ProtocolVersion;
 use crate::error::Error;
 use crate::key_log::NoKeyLog;
 use crate::sign::{CertifiedKey, SingleCertAndKey};
 use crate::sync::Arc;
-use crate::versions::TLS13;
 use crate::webpki::{self, WebPkiServerVerifier};
-use crate::{WantsVersions, compress, verify, versions};
+use crate::{compress, verify};
 
-impl ConfigBuilder<ClientConfig, WantsVersions> {
+impl ConfigBuilder<ClientConfig, WantsVerifier> {
     /// Enable Encrypted Client Hello (ECH) in the given mode.
     ///
-    /// This implicitly selects TLS 1.3 as the only supported protocol version to meet the
-    /// requirement to support ECH.
+    /// This requires TLS 1.3 as the only supported protocol version to meet the requirement
+    /// to support ECH.  At the end, the config building process will return an error if either
+    /// TLS1.3 _is not_ supported by the provider, or TLS1.2 _is_ supported.
     ///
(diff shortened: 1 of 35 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -345,6 +346,47 @@ See the documentation of the CryptoProvider type for more information.
             ..self
         }
     }
+
+    pub(crate) fn consistency_check(&self) -> Result<(), Error> {
```
djc, line 350: Should we privatize the fields and only expose getters?

djc:
> - All builder construction is fallible, allowing the basic checks on the provider to report their error as early as possible.

This seems unattractive. Why not report all errors at once, at the end?

ctz:
> > * All builder construction is fallible, allowing the basic checks on the provider to report their error as early as possible.
> 
> This seems unattractive. Why not report all errors at once, at the end?

I did this. In a couple of cases it added fallibility at the end of the builder process to accommodate, but most of these were already fallible.

rustls/src/client/builder.rs
```diff
@@ -6,31 +6,46 @@ use pki_types::{CertificateDer, PrivateKeyDer};
...
     /// In this case the configuration should only be shared by connections intended for domains
     /// that offer the provided [`crate::client::EchConfig`] in their DNS zone.
-    pub fn with_ech(
-        self,
-        mode: EchMode,
-    ) -> Result<ConfigBuilder<ClientConfig, WantsVerifier>, Error> {
-        let mut res = self.with_protocol_versions(&[&TLS13][..])?;
-        res.state.client_ech_mode = Some(mode);
-        Ok(res)
+    pub fn with_ech(mut self, mode: EchMode) -> Result<Self, Error> {
```
djc: This is still intermediately fallible, can we postpone this failure, too?

djc:
> I did this. In a couple of cases it added fallibility at the end of the builder process to accommodate, but most of these were already fallible.

Hmm, the transition to rustls-platform-verifier showed me that adding fallibility where none was previously needed can have substantial transitive effects. I guess my previous call not to have multiple builder methods fail should probably be nuanced to say that it would be better to limit the number of failure sites but if we can reduce the prevalence of fallible paths through the builder states (potentially by adding more states) I think that might be worth it? The goal would be to have strictly fewer fallibility points than before by coalescing fallibility points where possible.

But, all of that could be done in followup -- these changes are hairy enough already.

rustls/src/crypto/mod.rs
ctz: sorry, which fields?

djc: The `CryptoProvider` fields... `consistency_check()` provides a single-time check, but if we made the fields private we could do this at construction time and ensure these invariants hold throughout the lifetime of the `CryptoProvider`.

(I realize this has always been the case, just something that came to mind as I reviewed pulling out this `consistency_check()`.)

ctz: ah! yes, i think that will naturally happen in #2401 -- i was thinking that would come with some different concrete provider types (including ones that are const-constructable, and others that are suitable for runtime configuration)

djc: Yup, that makes sense.

PR #2602, opened 2025-08-22
Split cipher suite configuration in `CryptoProvider` by version

### LINKED ISSUE #1659: Specify ciphersuites for TLS 1.2 and TLS 1.3 separately

ctz:
I experimented with a bunch of ways of doing this which didn't lead to One Big Commit (like, introducing new fields, and moving things over to using the new fields in successive commits, and then removing the old field) but it ended up with too much back-and-forth especially in tests.

fixes #1659

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -358,24 +358,20 @@ impl ClientConfig {
     /// also configured.
     pub(crate) fn supports_version(&self, v: ProtocolVersion) -> bool {
         self.provider
-            .cipher_suites
-            .iter()
+            .iter_cipher_suites()
             .any(|cs| cs.version().version() == v)
     }
 
     #[cfg(feature = "std")]
     pub(crate) fn supports_protocol(&self, proto: Protocol) -> bool {
         self.provider
-            .cipher_suites
-            .iter()
+            .iter_cipher_suites()
             .any(|cs| cs.usable_for_protocol(proto))
     }
 
     pub(super) fn find_cipher_suite(&self, suite: CipherSuite) -> Option<SupportedCipherSuite> {
         self.provider
-            .cipher_suites
-            .iter()
-            .copied()
+            .iter_cipher_suites()
             .find(|&scs| scs.suite() == suite)
     }
 
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -353,13 +353,8 @@ impl ClientConfig {
         self.supports_version(ProtocolVersion::TLSv1_3)
     }
 
-    /// We support a given TLS version if it's quoted in the configured
-    /// versions *and* at least one ciphersuite for this version is
-    /// also configured.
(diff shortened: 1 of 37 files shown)


rustls/src/client/client_conn.rs
```diff
@@ -353,13 +353,20 @@ impl ClientConfig {
...
-    /// We support a given TLS version if it's quoted in the configured
-    /// versions *and* at least one ciphersuite for this version is
-    /// also configured.
+    /// We support a given TLS version if at least one ciphersuite for the version
+    /// is available.
     pub(crate) fn supports_version(&self, v: ProtocolVersion) -> bool {
-        self.provider
-            .iter_cipher_suites()
-            .any(|cs| cs.version().version() == v)
+        match v {
```
djc: Maybe move this code into the `CryptoProvider` so we can avoid more duplication?

rustls/src/server/hs.rs
```diff
@@ -581,15 +581,23 @@ impl ExpectClientHello {
...
-                // Reduce our supported ciphersuites by the certified key's algorithm.
-                suite.usable_for_signature_algorithm(sig_key_algorithm)
-                // And version
-                && suite.version().version() == selected_version
-                // And protocol
-                && suite.usable_for_protocol(protocol)
-                // And support one of key exchange groups
-                && (ecdhe_possible && suite.usable_for_kx_algorithm(KeyExchangeAlgorithm::ECDHE)
-                || ffdhe_possible && suite.usable_for_kx_algorithm(KeyExchangeAlgorithm::DHE))
+                match suite {
```
djc: How about linearizing this to handle the `Tls12` case last/outside of the `match`?

djc (approved):
Looks nice! Other than the comments, I think there are some other potential further ways to clean this up, but that can wait.

PR #2604, opened 2025-08-26
Avoid computing an extra `A(n)` term in TLS1.2 PRF

ctz:
This rejigs this loop to calculate the current rather than next term. That removes calculation of the next term in the final iteration.


rustls/src/crypto/tls12.rs
```diff
@@ -92,17 +92,26 @@ pub trait PrfSecret: Send + Sync {
...
 pub(crate) fn prf(out: &mut [u8], hmac_key: &dyn hmac::Key, label: &[u8], seed: &[u8]) {
-    // A(1)
-    let mut current_a = hmac_key.sign(&[label, seed]);
+    let mut previous_a: Option<hmac::Tag> = None;
 
     let chunk_size = hmac_key.tag_len();
     for chunk in out.chunks_mut(chunk_size) {
+        let a_i = match previous_a {
+            None => {
+                // A(0) = HMAC_hash(secret, label + seed)
```
djc: Comment outside the arm instead of inside for tighter formatting?

rustls/src/crypto/tls12.rs
```diff
@@ -92,17 +92,22 @@ pub trait PrfSecret: Send + Sync {
...
+        let a_i = match previous_a {
+            // A(0) = HMAC_hash(secret, label + seed)
+            None => hmac_key.sign(&[label, seed]),
+            // A(i) = HMAC_hash(secret, A(i - 1))
+            Some(previous_a) => hmac_key.sign(&[previous_a.as_ref()]),
+        };
+
         // P_hash[i] = HMAC_hash(secret, A(i) + seed)
-        let p_term = hmac_key.sign(&[current_a.as_ref(), label, seed]);
+        let p_term = hmac_key.sign(&[a_i.as_ref(), label, seed]);
```
cpu: pre-existing, but should the comment above this line read `P_hash[i] = HMAC_hash(secret, A(i) + label + seed)`? I was looking at 5246 while reviewing the diff and noticed the omission of `label` on L106.

The text as-is matches what [RFC 5246 says](https://datatracker.ietf.org/doc/html/rfc5246#section-5), but that text also doesn't reference `label` outside of the `PRF(secret, label, seed)` construction where our earlier comments (e.g. for `A(0) = HMAC_hash(secret, label + seed)`) do. With that in mind it feels like L106 should have it for consistency, or we should match RFC text throughout. WDYT?

cpu (approved):

rustls/src/crypto/tls12.rs
ctz: Ah yes, that part of the RFC is pretty confusing with its different definition of `seed` inside the definition of `P_hash` and outside. I'll fix the comment.

PR #2605, opened 2025-08-27
impl `Display` for `AlertDescription`

### LINKED ISSUE #1307: Could AltertDescription::CertificateRequired be renamed to `ClientCertificateRequired`
I was getting this error and was thinking "duh of course I need a certificate, but I've definitely setup the server with a certificate"
Turns out the server was configured with mutual TLS and the error was referring to the missing client certificate.

If we renamed it from `CertificateRequired` to `ClientCertificateRequired` I think that would help avoid confusions like mine?

I saw this comment:
```rust
    /// The `AlertDescription` TLS protocol enum.  Values in this enum are taken
```
But its not clear from the comment if the variant names are taken from the RFC's as well or just the error codes.
I dont know which particular RFCs its talking about so I couldnt check if that was the case.

ctz:
Use this to provide some human-readable descriptions of what these mean, and hopefully are moderately understandable for people who see them.

Failing that, they are unique strings and some additional comments are nearby if someone looks at the source where they appear(?)

This looks like:

```
$ cargo -q run --bin tlsclient-mio -- --http rc4.badssl.com
TLS error: received fatal alert: peer was unable to negotiate an acceptable set of security parameters from the options we offered
Connection closed
```

Also open to a less wordy-option, eg a direct expansion of the variant name (so in the above, it would be "TLS error: received fatal alert: handshake failed"). WDYT?

fixes #1307


rustls/src/enums.rs
```diff
@@ -47,6 +47,124 @@ enum_builder! {
     }
 }
 
+impl core::fmt::Display for AlertDescription {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            // this is normal.
+            Self::CloseNotify => write!(f, "peer cleanly closed the connection"),
```
djc: Hmm, we use `AlertDescription` both as send and as receive types, right? Should we try to come up with wording that more naturally also reflects the send side?

ctz: Yeah we use it in both cases. Maybe the `Display` code for `Error::AlertReceived` should be contributing the `the peer ` part? Then keep the verb and descriptors here?

(though that would mean a bit of rewording)

djc: Yeah, that sounds nice!

ctz: did this, and reviewed the result using `cargo -q test alert_display -- --nocapture | llm` to check syntax and grammar

djc (approved):

Issue #2406, opened 2025-04-01
TLS 1.3 early_exporter_master_secret


Itsusinn:
**Checklist**
* [x] I've searched the issue tracker for similar requests

I think an interface similiar to [export_keying_material](https://docs.rs/rustls/latest/rustls/quic/enum.Connection.html#method.export_keying_material) (might be `export_early_keying_material`?) could be support

In [RFC8446 # Section 7.5](https://datatracker.ietf.org/doc/html/rfc8446#section-7.5)

[RFC5705] defines keying material exporters for TLS in terms of the TLS pseudorandom function (PRF).  This document replaces the PRF with HKDF, thus requiring a new construction.  The exporter interface remains the same.

The exporter value is computed as:

TLS-Exporter(label, context_value, key_length) = HKDF-Expand-Label(Derive-Secret(Secret, label, ""), "exporter", Hash(context_value), key_length)

Where Secret is either the early_exporter_master_secret or the exporter_master_secret.  Implementations MUST use the exporter_master_secret unless explicitly specified by the application.  The early_exporter_master_secret is defined for use in settings where an exporter is needed for 0-RTT data.  A separate interface for the early exporter is RECOMMENDED; this avoids the exporter user accidentally using an early exporter when a regular one is desired or vice versa.

djc:
What do you need this for? What is your use case?

Itsusinn:
@djc 

To facilitate the discussion, let's consider a scenario involving a QUIC-based application-layer protocol. In this protocol, the client must send an authentication packet to the server; otherwise, all other operations will be suspended until the authentication packet is received.

When not using 0-RTT, the authentication packet consists of two components: a username and a password. For security reasons, the password is encrypted using export_keying_material. This design allows the application-layer protocol to achieve its own form of 0-RTT functionality.

However, when 0-RTT is enabled, if the password continues to be encrypted with export_keying_material, the authentication packet must wait for the TLS 1.3 1-RTT handshake to complete before being sent. This dependency would prevent the application-layer protocol from maintaining true 0-RTT behavior.

ctz:
I think it's fair to say we understand the general utility of early data exporters. Generally we don't add features unless motivated by a specific downstream usage (in other words, we want to see at least one downstream project ready to call the new API). With that said I've read tuic-protocol/tuic#264 through a translator and it kind of makes sense.

PR #2608, opened 2025-08-28
Replace exporter API

### LINKED ISSUE #945: `export_keying_material()` API doesn't conform with TLS 1.3 recommendations

ctz:
This represents the exporter key material inside its own object, where the application controls the lifetime.  The application chooses when to take the object, and can do that at most once for a given connection.

The application can choose to keep the object alive indefinitely alongside the connection to obtain an equivalent to the previous behaviour.

An application that does nothing with exporters means the lifetime is as long as the connection (as before) which is OK because a secret which is used for nothing has no value.

This also makes it easier to do #2406, I think.

fixes #945 and follows the suggestion made in https://github.com/rustls/rustls/issues/945#issuecomment-1385503870

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -14,8 +14,8 @@ use crate::check::{inappropriate_handshake_message, inappropriate_message};
 use crate::client::common::{ClientAuthDetails, ServerCertDetails};
 use crate::client::{ClientConfig, hs};
 use crate::common_state::{CommonState, HandshakeKind, KxState, Side, State};
+use crate::conn::ConnectionRandoms;
 use crate::conn::kernel::{Direction, KernelContext, KernelState};
-use crate::conn::{BytesExporter, ConnectionRandoms};
 use crate::crypto::KeyExchangeAlgorithm;
 use crate::enums::{AlertDescription, ContentType, HandshakeType, ProtocolVersion};
 use crate::error::{Error, InvalidMessage, PeerIncompatible, PeerMisbehaved};
@@ -1299,9 +1299,10 @@ impl State<ClientConnectionData> for ExpectFinished {
             false => None,
         };
 
+        cx.common.exporter = Some(st.secrets.into_exporter());
+
         Ok(Box::new(ExpectTraffic {
             extracted_secrets,
-            exporter: st.secrets.into_exporter(),
             _cert_verified: st.cert_verified,
             _sig_verified: st.sig_verified,
             _fin_verified,
@@ -1329,7 +1330,6 @@ impl State<ClientConnectionData> for ExpectFinished {
 struct ExpectTraffic {
     // only available if `config.enable_secret_extraction` is true
     extracted_secrets: Option<Result<PartiallyExtractedSecrets, Error>>,
-    exporter: Box<dyn BytesExporter>,
     _cert_verified: verify::ServerCertVerified,
     _sig_verified: verify::HandshakeSignatureValid,
     _fin_verified: verify::FinishedMessageVerified,
@@ -1356,16 +1356,6 @@ impl State<ClientConnectionData> for ExpectTraffic {
             }
         }
         Ok(self)
-    }
-
-    fn export_keying_material(
-        &self,
-        output: &mut [u8],
-        label: &[u8],
-        context: Option<&[u8]>,
-    ) -> Result<(), Error> {
-        self.exporter
-            .export_keying_material(output, label, context)
     }
 
     fn into_external_state(
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1360,13 +1360,14 @@ impl State<ClientConnectionData> for ExpectTraffic {
         Ok(())
     }
 
-    fn extract_secrets(&self) -> Result<PartiallyExtractedSecrets, Error> {
-        self.secrets
-            .extract_secrets(Side::Client)
-    }
-
-    fn into_external_state(self: Box<Self>) -> Result<Box<dyn KernelState + 'static>, Error> {
-        Ok(self)
+    fn into_external_state(
+        self: Box<Self>,
+    ) -> Result<(PartiallyExtractedSecrets, Box<dyn KernelState + 'static>), Error> {
+        Ok((
+            self.secrets
+                .extract_secrets(Side::Client)?,
+            self,
+        ))
     }
 
     fn into_owned(self: Box<Self>) -> hs::NextState<'static> {
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -14,8 +14,8 @@ use crate::check::{inappropriate_handshake_message, inappropriate_message};
 use crate::client::common::{ClientAuthDetails, ServerCertDetails};
 use crate::client::{ClientConfig, hs};
 use crate::common_state::{CommonState, HandshakeKind, KxState, Side, State};
-use crate::conn::ConnectionRandoms;
 use crate::conn::kernel::{Direction, KernelContext, KernelState};
+use crate::conn::{BytesExporter, ConnectionRandoms};
 use crate::crypto::KeyExchangeAlgorithm;
 use crate::enums::{AlertDescription, ContentType, HandshakeType, ProtocolVersion};
 use crate::error::{Error, InvalidMessage, PeerIncompatible, PeerMisbehaved};
@@ -1301,7 +1301,7 @@ impl State<ClientConnectionData> for ExpectFinished {
 
         Ok(Box::new(ExpectTraffic {
             extracted_secrets,
-            secrets: st.secrets,
+            exporter: st.secrets.into_exporter(),
             _cert_verified: st.cert_verified,
             _sig_verified: st.sig_verified,
             _fin_verified,
@@ -1329,7 +1329,7 @@ impl State<ClientConnectionData> for ExpectFinished {
 struct ExpectTraffic {
     // only available if `config.enable_secret_extraction` is true
     extracted_secrets: Option<Result<PartiallyExtractedSecrets, Error>>,
-    secrets: ConnectionSecrets,
+    exporter: Box<dyn BytesExporter>,
     _cert_verified: verify::ServerCertVerified,
     _sig_verified: verify::HandshakeSignatureValid,
     _fin_verified: verify::FinishedMessageVerified,
@@ -1364,9 +1364,8 @@ impl State<ClientConnectionData> for ExpectTraffic {
         label: &[u8],
         context: Option<&[u8]>,
     ) -> Result<(), Error> {
-        self.secrets
-            .export_keying_material(output, label, context);
-        Ok(())
+        self.exporter
+            .export_keying_material(output, label, context)
     }
 
     fn into_external_state(
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -1293,7 +1293,14 @@ impl State<ClientConnectionData> for ExpectFinished {
 
         cx.common
             .start_traffic(&mut cx.sendable_plaintext);
+
+        let extracted_secrets = match st.config.enable_secret_extraction {
+            true => Some(st.secrets.extract_secrets(Side::Client)),
+            false => None,
+        };
+
         Ok(Box::new(ExpectTraffic {
+            extracted_secrets,
             secrets: st.secrets,
             _cert_verified: st.cert_verified,
             _sig_verified: st.sig_verified,
@@ -1320,6 +1327,8 @@ impl State<ClientConnectionData> for ExpectFinished {
 
 // -- Traffic transit state --
 struct ExpectTraffic {
+    // only available if `config.enable_secret_extraction` is true
+    extracted_secrets: Option<Result<PartiallyExtractedSecrets, Error>>,
     secrets: ConnectionSecrets,
     _cert_verified: verify::ServerCertVerified,
     _sig_verified: verify::HandshakeSignatureValid,
@@ -1361,13 +1370,13 @@ impl State<ClientConnectionData> for ExpectTraffic {
     }
 
     fn into_external_state(
-        self: Box<Self>,
+        mut self: Box<Self>,
     ) -> Result<(PartiallyExtractedSecrets, Box<dyn KernelState + 'static>), Error> {
-        Ok((
-            self.secrets
-                .extract_secrets(Side::Client)?,
-            self,
-        ))
+        let extracted_secrets = self
+            .extracted_secrets
+            .take()
+            .expect("into_external_state() => enable_secret_extraction")?;
+        Ok((extracted_secrets, self))
     }
 
     fn into_owned(self: Box<Self>) -> hs::NextState<'static> {
--- a/rustls/src/conn.rs
+++ b/rustls/src/conn.rs
@@ -1,5 +1,5 @@
 use alloc::boxed::Box;
-use core::fmt::Debug;
+use core::fmt::{self, Debug};
 use core::mem;
 use core::ops::{Deref, DerefMut, Range};
 #[cfg(feature = "std")]
@@ -33,6 +33,7 @@ mod connection {
 
     use crate::ConnectionCommon;
     use crate::common_state::{CommonState, IoState};
+    use crate::conn::KeyingMaterialExporter;
     use crate::error::Error;
     use crate::msgs::message::OutboundChunks;
     use crate::suites::ExtractedSecrets;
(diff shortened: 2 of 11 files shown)


rustls/src/client/tls12.rs
```diff
@@ -1293,7 +1293,14 @@ impl State<ClientConnectionData> for ExpectFinished {
 
         cx.common
             .start_traffic(&mut cx.sendable_plaintext);
+
+        let extracted_secrets = match st.config.enable_secret_extraction {
```
djc: I think we can use `.then()` here? And maybe inline it into the `ExpectTraffic` initializer?

rustls/src/server/tls12.rs
```diff
@@ -948,7 +948,17 @@ impl State<ServerConnectionData> for ExpectFinished {
 
         cx.common
             .start_traffic(&mut cx.sendable_plaintext);
+
+        let extracted_secrets = match self.config.enable_secret_extraction {
```
djc: As above.

rustls/src/client/tls12.rs
```diff
@@ -1361,13 +1370,13 @@ impl State<ClientConnectionData> for ExpectTraffic {
...
     ) -> Result<(PartiallyExtractedSecrets, Box<dyn KernelState + 'static>), Error> {
-        Ok((
-            self.secrets
-                .extract_secrets(Side::Client)?,
-            self,
-        ))
+        let extracted_secrets = self
+            .extracted_secrets
+            .take()
+            .expect("into_external_state() => enable_secret_extraction")?;
```
djc: I don't find it that obvious what this means or what the invariant here is. Can you add some comments that add a little more context? (Same for the `server`.)

rustls/src/tls12/mod.rs
```diff
@@ -109,7 +109,7 @@ impl fmt::Debug for Tls12CipherSuite {
 pub(crate) struct ConnectionSecrets {
     pub(crate) randoms: ConnectionRandoms,
     suite: &'static Tls12CipherSuite,
-    master_secret: [u8; 48],
+    master_secret: Zeroizing<[u8; 48]>,
```
djc, line 132: This is nice -- shouldn't we also zeroize `master_secret_prf`? (If not, add a comment?)

rustls/src/conn.rs
```diff
@@ -358,6 +358,22 @@ https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof"
 #[cfg(feature = "std")]
 pub use connection::{Connection, Reader, Writer};
 
+/// This trait is for any object that can export keying material.
+///
+/// There are several such internal implementations, depending on the context
+/// and protocol version.
+pub(crate) trait BytesExporter: Send + Sync {
```
djc: Introduce this in the same commit as its users? The trait name doesn't really evoke the `export_keying_material()` functionality for me.

rustls/src/conn.rs
```diff
@@ -1221,12 +1221,20 @@ impl<Data> ConnectionCore<Data> {
...
-        match self.state.as_ref() {
-            Ok(st) => st
-                .export_keying_material(output.as_mut(), label, context)
-                .map(|_| output),
-            Err(e) => Err(e.clone()),
-        }
+        // promote fused error
+        self.state
+            .as_ref()
+            .map_err(|e| e.clone())?;
```
djc: This seems less clear than

```rust
if let Err(err) = &self.state {
     return Err(err.clone());
}
```

? Same for these other blocks below.

rustls/src/conn.rs
```diff
@@ -358,6 +354,49 @@ https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof"
...
+    /// material derived from the master session secret using `label`
+    /// and `context` for diversification. Ownership of the buffer is taken
+    /// by the function and returned via the Ok result to ensure no key
+    /// material leaks if the function fails.
+    ///
+    /// See RFC5705 for more details on what this does and is for.
+    ///
+    /// This function is not meaningful if `output.len()` is zero and will
+    /// return an error in that case.
+    pub fn export_keying_material<T: AsMut<[u8]>>(
```
djc: This method name now seems overly elaborate in the context of the type name.

rustls/src/conn.rs
```diff
@@ -421,32 +460,15 @@ impl<Data> ConnectionCommon<Data> {
...
-    /// See RFC5705 for more details on what this does and is for.
+    /// Returns an object that can derive key material from the agreed connection secrets.
     ///
-    /// For TLS1.3 connections, this function does not use the
-    /// "early" exporter at any point.
+    /// This function can be called at most once per connection, it returns
+    /// an error on subsequent calls.
     ///
-    /// This function fails if called prior to the handshake completing;
+    /// This function returns an error if called prior to the handshake completing;
```
djc: I think this might be clearer if you turn it into a bullet list:

```
This function will error when:

- called prior to the handshake completing;
- or if it is called more than once per connection.
```

rustls/src/conn.rs
djc: (I guess this code ends up being deleted in a later commit, so might not be as important.)

rustls/src/client/tls12.rs
ctz: read `=>` as "implies". to put it another way, there should be no calls to `into_external_state()` for connections whose config does not have `enable_secret_extraction` set

rustls/src/tls12/mod.rs
ctz: will add a comment. whatever is behind the impl for `master_secret_prf` needs to be zeroizing its secret parts in `drop()`. eg, for `PrfSecretUsingHmac` this happens via the `drop()` for `impl hmac::Key`. eventually once you get to the bottom of the large chain of drops, aws-lc-rs calls `HMAC_CTX_cleanup` which zeroes the secret parts.

ctz:
> Having looked previously at the `secrets_extraction` stuff which we guard with a proactive `Config::enable_secret_extraction` field, should we do the same thing for the exporter? I'd guess exporting secrets is pretty rare so making it opt-in seems nice.

While the naming is not great (but standard) exporters are 100% safe and don't affect the security of the underlying connection or session. So there's not much reason for restricting it like that.

rustls/src/client/tls12.rs
ctz: reworded this

cpu:
@ctz did you have revisions to this branch you haven't pushed yet? I was starting to look at it and noticed some djc feedback is resolved but there wasn't a push w/ updated code.

rustls/src/conn.rs
ctz: -> `derive()` with a docs note mentioning the names used in other libraries

rustls/src/conn.rs
ctz: -> `Exporter` (the public object name remains named `KeyingMaterialExporter`)

ctz:
> @ctz did you have revisions to this branch you haven't pushed yet? I was starting to look at it and noticed some djc feedback is resolved but there wasn't a push w/ updated code.

Sorry, I got sidetracked halfway through addressing them. All done now.

rustls/src/client/tls12.rs
```diff
@@ -1361,13 +1368,13 @@ impl State<ClientConnectionData> for ExpectTraffic {
...
     ) -> Result<(PartiallyExtractedSecrets, Box<dyn KernelState + 'static>), Error> {
-        Ok((
-            self.secrets
-                .extract_secrets(Side::Client)?,
-            self,
-        ))
+        let extracted_secrets = self
+            .extracted_secrets
+            .take()
+            .expect("calling into_external_state() requires enable_secret_extraction")?;
```
cpu: I believe the `expect()` here is safe because the only call-site of this mod-internal fn at the time of this commit is in `ConnectionCore::dangerous_into_kernel_connection()`, another crate-internal fn that checks `enable_secret_extraction`.

Since this fn already returns a `Result` would it be clearer to have the call-site do the `expect()` unwrapping so that it's in the same place as the invariant check?

rustls/src/server/tls12.rs
```diff
@@ -1009,13 +1019,13 @@ impl State<ServerConnectionData> for ExpectTraffic {
...
     ) -> Result<(PartiallyExtractedSecrets, Box<dyn KernelState + 'static>), Error> {
-        Ok((
-            self.secrets
-                .extract_secrets(Side::Server)?,
-            self,
-        ))
+        let extracted_secrets = self
+            .extracted_secrets
+            .take()
+            .expect("calling into_external_state() requires enable_secret_extraction")?;
```
cpu: same feedback here w.r.t where `expect()` is called.

rustls/src/client/tls12.rs
djc: Might be nicer to do this as an `Error` variant (maybe like `Error::Unreachable`)?

(Same for 1.3.)

rustls/src/client/tls12.rs
```diff
@@ -1320,6 +1325,8 @@ impl State<ClientConnectionData> for ExpectFinished {
 
 // -- Traffic transit state --
 struct ExpectTraffic {
+    // only available if `config.enable_secret_extraction` is true
```
cpu: nit: would it be clearer to say "only Some if..." ? Available or not available makes me think more of the presence of the field itself, and not necessarily what's in the `Option`.

rustls/src/server/tls12.rs
```diff
@@ -968,6 +976,8 @@ impl State<ServerConnectionData> for ExpectFinished {
 
 // --- Process traffic ---
 struct ExpectTraffic {
+    // only available if `config.enable_secret_extraction` is true
```
cpu: if you agree with my feedback about saying "Some" instead of "available" it applies here as well.

rustls/src/tls12/mod.rs
cpu: WDYT about adding a mention of this on the `PrfSecret` trait definition as well as here on the `master_secret_prf` field?

rustls/src/conn.rs
```diff
@@ -358,6 +358,26 @@ https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof"
...
+/// This trait is for any object that can export keying material.
+///
+/// The terminology comes from [RFC5705](https://datatracker.ietf.org/doc/html/rfc5705)
+/// but doesn't really involve "exporting" key material (in the usual meaning of "export"
+/// -- of moving an artifact from one domain to another) but is best thought of as key
+/// diversification using an existing secret.  That secret is implicit in this interface,
+/// so is assumed to be held by `self`.
+///
+/// There are several such internal implementations, depending on the context
+/// and protocol version.
```
cpu, line 411: maybe another place to hint to an implementer to `Zeroize` on drop in their impl?

djc (approved):
Good stuff!

rustls/src/conn.rs
```diff
@@ -358,6 +358,26 @@ https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof"
...
+pub(crate) trait Exporter: Send + Sync {
+    /// Fills in `output` with derived keying material.
+    ///
+    /// This is deterministic depending on a base secret (implicit in `self`),
+    /// plus the `label` and `context` values.
+    ///
+    /// Must fill in `output` entirely, or return an error.
+    fn derive(&self, output: &mut [u8], label: &[u8], context: Option<&[u8]>) -> Result<(), Error>;
```
cpu: I think this follows from `export_keying_material()` but it seems odd to have `output` as the first arg instead of the last. We're inconsistent on this in other places (e.g. `crypto/tls13.rs` puts it last in `expand_*()` but places like `crypto/tls12.rs`'s `prf()`, and the pre-existing `export_keying_materia()` put it first.

I think I prefer seeing it last if all things were equal. WDYT?

rustls/src/tls12/mod.rs
```diff
@@ -319,6 +313,28 @@ impl ConnectionSecrets {
...
+    master_secret_prf: Box<dyn crypto::tls12::PrfSecret>,
+}
+
+impl Exporter for Tls12Exporter {
+    fn derive(&self, output: &mut [u8], label: &[u8], context: Option<&[u8]>) -> Result<(), Error> {
+        let mut randoms = Vec::new();
+        randoms.extend_from_slice(&self.randoms.client);
+        randoms.extend_from_slice(&self.randoms.server);
+        if let Some(context) = context {
+            assert!(context.len() <= 0xffff);
```
cpu: `assert!(context.len() <= u16::MAX as usize);` to avoid a magic number?

rustls/src/conn.rs
```diff
@@ -1213,32 +1243,12 @@ impl<Data> ConnectionCore<Data> {
...
-        }
-
-        // promote fused error
-        if let Err(err) = &self.state {
-            return Err(err.clone());
+    pub(crate) fn exporter(&mut self) -> Result<KeyingMaterialExporter, Error> {
+        match self.common_state.exporter.take() {
+            Some(inner) => Ok(KeyingMaterialExporter { inner }),
+            None if self.common_state.is_handshaking() => Err(Error::HandshakeNotComplete),
+            None => Err(Error::General("exporter already used".into())),
```
cpu, line 1250: This feels like it _might_ be something worth its own error type. Did you consider that and decide it was overkill?

rustls/src/tls12/mod.rs
cpu: This falls away in cbf6a80b4f3428e580fc0260e4ff371acba1fe93

rustls/src/client/tls12.rs
ctz: This is now its own `Error::Unreachable` variant (and the parent is passing that on, rather than pulling it out and panicking on it)

ctz: I've tried this, PTAL

rustls/src/conn.rs
ctz: Did this, from the new public API down.

rustls/src/tls12/mod.rs
```diff
@@ -319,6 +313,28 @@ impl ConnectionSecrets {
...
 }
 
+pub(crate) struct Tls12Exporter {
+    randoms: ConnectionRandoms,
+    master_secret_prf: Box<dyn crypto::tls12::PrfSecret>,
+}
+
+impl Exporter for Tls12Exporter {
+    fn derive(&self, label: &[u8], context: Option<&[u8]>, output: &mut [u8]) -> Result<(), Error> {
+        let mut randoms = Vec::new();
```
djc: Nit: `Vec::with_capacity()`? Or could this be an array?

rustls/src/conn.rs
ctz: WDYT about some sort of `Error::ApiMisuse(ApiMisuse::ExporterAlreadyUsed)`? I think there are a number of other candidates which are similar.

cpu: SGTM!

djc: Sounds nice.

djc (approved):

rustls/src/tls12/mod.rs
cpu: I thought about the `with_capacity()` suggestion when I was reviewing this but I decided the optional `context` without a fixed length made it seem less appealing. Maybe I was premature in dismissing the idea?

djc: I think even doing a `Vec::with_capacity(64)` would avoid one allocation at trivial cost.

ctz: Did this (albeit with the full calculation)

rustls/src/conn.rs
ctz: Will do this in a follow-up

PR #2615, opened 2025-09-01
small tweaks from 2602

cpu:
Two very tiny nits noticed reviewing https://github.com/rustls/rustls/pull/2602 post-merge.

The first is just a minor tweak to avoid a clunkier debug repr in an example panic message. The second fixes up a change made in https://github.com/rustls/rustls/pull/2602/commits/720113c20a9bc189549855b416fb00cde352bb73 that switched some callers from copying the `SupportedCipherSuite` iter items to cloning them. `SupportedCipherSuite` is `Copy` so I think its clearer to use `copied()` on the `Iterator`.

### DIFF
--- a/examples/src/bin/tlsclient-mio.rs
+++ b/examples/src/bin/tlsclient-mio.rs
@@ -333,7 +333,10 @@ fn filter_suites(mut provider: CryptoProvider, suites: &[String]) -> CryptoProvi
 
     for s in suites {
         if !known_suites.contains(&s.to_lowercase()) {
-            panic!("cannot look up ciphersuite '{s}'. should be one of {known_suites:?}");
+            panic!(
+                "cannot look up ciphersuite '{s}'. should be one of {known_suites}",
+                known_suites = known_suites.join(", ")
+            );
         }
     }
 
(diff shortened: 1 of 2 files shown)


examples/src/bin/tlsclient-mio.rs
```diff
@@ -333,7 +333,10 @@ fn filter_suites(mut provider: CryptoProvider, suites: &[String]) -> CryptoProvi
 
     for s in suites {
         if !known_suites.contains(&s.to_lowercase()) {
-            panic!("cannot look up ciphersuite '{s}'. should be one of {known_suites:?}");
+            panic!(
+                "cannot look up ciphersuite '{s}'. should be one of {known_suites}",
```
djc: If we're going to tweak this -- don't love `.` in the error (and then not capitalizing the next sentence), maybe `;` or `--` instead?

djc (approved):
Nice!

examples/src/bin/tlsclient-mio.rs
cpu: Good points. I replaced the "cannot look up" part with "unsupported". The former reads like it was some kind of lookup error and not that it's just plain an unknown/unsupported ciphersuite. 

> maybe ; or -- instead?

I used `;`, I fear the emdash has been lost to the LLMs.

examples/src/bin/tlsclient-mio.rs
```diff
@@ -333,7 +333,10 @@ fn filter_suites(mut provider: CryptoProvider, suites: &[String]) -> CryptoProvi
...
+                known_suites = known_suites.join(", ")
```
ctz, line 338: this equals sign usage is new to me!

PR #2616, opened 2025-09-01
Enable & address `clippy::cloned_instead_of_copied`

ctz:
Seems like a pretty ironclad lint? Not clear to me why it's `allow` by default.

re #2615


cpu (approved):
Good idea!

djc (approved):

PR #2617, opened 2025-09-02
Add option to relax SNI validation

### LINKED ISSUE #2403: Option to relax SNI validation
* [✅] I've searched the issue tracker for similar requests

We occasionally encounter TLS clients that send invalid SNIs. As some other TLS libraries accept such invalid SNIs, our customers expect us to work with their non-compliant TLS clients as well.

I think a simple solution would be to add the option to configure the rustls server to accept (and ignore) invalid SNIs in ClientHello (similar to what was done for IP addresses as SNI in #1881).

s-arash:
Closes #2403

### DIFF
--- a/rustls/src/msgs/handshake.rs
+++ b/rustls/src/msgs/handshake.rs
@@ -284,6 +284,16 @@ impl ServerNamePayload<'_> {
     const SIZE_LEN: ListLength = ListLength::NonZeroU16 {
         empty_error: InvalidMessage::IllegalEmptyList("ServerNames"),
     };
+
+    /// Get the `DnsName` out of this `ServerNamePayload` if it contains one.
+    /// The returned `DnsName` will be normalized (converted to lowercase).
+    pub(crate) fn to_dns_name_normalized(&self) -> Option<DnsName<'static>> {
+        match self {
+            Self::SingleDnsName(dns_name) => Some(dns_name.to_lowercase_owned()),
+            Self::IpAddress => None,
+            Self::Invalid => None,
+        }
+    }
 }
 
 /// Simplified encoding/decoding for a `ServerName` extension payload to/from `DnsName`
@@ -341,9 +351,7 @@ impl<'a> Codec<'a> for ServerNamePayload<'a> {
                 }
 
                 HostNamePayload::IpAddress(_invalid) => {
-                    warn!(
-                        "Illegal SNI extension: ignoring IP address presented as hostname ({_invalid:?})"
-                    );
+                    warn!("Illegal SNI extension: IP address presented as hostname ({_invalid:?})");
                     Some(Self::IpAddress)
                 }
 
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -1,4 +1,4 @@
-use alloc::borrow::ToOwned;
+use alloc::borrow::{Cow, ToOwned};
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 
@@ -287,6 +287,15 @@ impl ExpectClientHello {
             .config
             .supports_version(ProtocolVersion::TLSv1_2);
 
+        cx.data.sni = self
+            .config
+            .invalid_sni_policy
+            .accept(client_hello.server_name.as_ref())
+            .map_err(|e| {
+                cx.common
+                    .send_fatal_alert(AlertDescription::IllegalParameter, e)
+            })?;
+
         // Are we doing TLS1.3?
         let version = if let Some(versions) = &client_hello.supported_versions {
             if versions.tls13 && tls13_enabled {
@@ -354,7 +363,7 @@ impl ExpectClientHello {
         // Choose a certificate.
         let certkey = {
             let client_hello = ClientHello {
-                server_name: &cx.data.sni,
+                server_name: cx.data.sni.as_ref().map(Cow::Borrowed),
                 signature_schemes: &sig_schemes,
                 alpn: client_hello.protocols.as_ref(),
                 client_cert_types: client_hello
@@ -639,8 +648,6 @@ impl State<ServerConnectionData> for ExpectClientHello {
 /// doesn't depend on a `ServerConfig` being available and extract everything needed to build a
 /// [`ClientHello`] value for a [`ResolvesServerCert`].
 ///
-/// Note that this will modify `data.sni` even if config or certificate resolution fail.
-///
 /// [`ResolvesServerCert`]: crate::server::ResolvesServerCert
 pub(super) fn process_client_hello<'m>(
     m: &'m Message<'m>,
@@ -664,40 +671,16 @@ pub(super) fn process_client_hello<'m>(
     // No handshake messages should follow this one in this flight.
     cx.common.check_aligned_handshake()?;
 
-    // Extract and validate the SNI DNS name, if any, before giving it to
-    // the cert resolver. In particular, if it is invalid then we should
-    // send an Illegal Parameter alert instead of the Internal Error alert
-    // (or whatever) that we'd send if this were checked later or in a
-    // different way.
-    //
-    // [RFC6066][] specifies that literal IP addresses are illegal in
-    // `ServerName`s with a `name_type` of `host_name`.
-    //
-    // Some clients incorrectly send such extensions: we choose to
-    // successfully parse these (into `ServerNamePayload::IpAddress`)
-    // but then act like the client sent no `server_name` extension.
-    //
-    // [RFC6066]: https://datatracker.ietf.org/doc/html/rfc6066#section-3
-    let sni = match &client_hello.server_name {
-        Some(ServerNamePayload::SingleDnsName(dns_name)) => Some(dns_name.to_lowercase_owned()),
-        Some(ServerNamePayload::IpAddress) => None,
-        Some(ServerNamePayload::Invalid) => {
-            return Err(cx.common.send_fatal_alert(
-                AlertDescription::IllegalParameter,
-                PeerMisbehaved::ServerNameMustContainOneHostName,
-            ));
+    if done_retry {
+        let ch_sni = client_hello
+            .server_name
+            .as_ref()
+            .and_then(ServerNamePayload::to_dns_name_normalized);
+        if cx.data.sni != ch_sni {
+            return Err(PeerMisbehaved::ServerNameDifferedOnRetry.into());
         }
-        None => None,
-    };
-
-    // save only the first SNI
-    if let (Some(sni), false) = (&sni, done_retry) {
-        // Save the SNI into the session.
-        // The SNI hostname is immutable once set.
-        assert!(cx.data.sni.is_none());
-        cx.data.sni = Some(sni.clone());
-    } else if cx.data.sni != sni {
-        return Err(PeerMisbehaved::ServerNameDifferedOnRetry.into());
+    } else {
+        assert!(cx.data.sni.is_none())
     }
 
     let sig_schemes = client_hello
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -1,3 +1,4 @@
+use alloc::borrow::Cow;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::fmt;
@@ -23,7 +24,9 @@ use crate::error::Error;
 use crate::kernel::KernelConnection;
 use crate::log::trace;
 use crate::msgs::base::Payload;
-use crate::msgs::handshake::{ClientHelloPayload, ProtocolName, ServerExtensionsInput};
+use crate::msgs::handshake::{
+    ClientHelloPayload, ProtocolName, ServerExtensionsInput, ServerNamePayload,
+};
 use crate::msgs::message::Message;
 use crate::suites::ExtractedSecrets;
 use crate::sync::Arc;
@@ -31,7 +34,9 @@ use crate::sync::Arc;
 use crate::time_provider::DefaultTimeProvider;
 use crate::time_provider::TimeProvider;
 use crate::vecbuf::ChunkVecBuffer;
-use crate::{DistinguishedName, KeyLog, NamedGroup, WantsVerifier, compress, sign, verify};
+use crate::{
+    DistinguishedName, KeyLog, NamedGroup, PeerMisbehaved, WantsVerifier, compress, sign, verify,
+};
 
 /// A trait for the ability to store server session data.
 ///
@@ -133,7 +138,7 @@ pub trait ResolvesServerCert: Debug + Send + Sync {
 /// A struct representing the received Client Hello
 #[derive(Debug)]
 pub struct ClientHello<'a> {
-    pub(super) server_name: &'a Option<DnsName<'a>>,
+    pub(super) server_name: Option<Cow<'a, DnsName<'a>>>,
     pub(super) signature_schemes: &'a [SignatureScheme],
     pub(super) alpn: Option<&'a Vec<ProtocolName>>,
     pub(super) server_cert_types: Option<&'a [CertificateType]>,
@@ -151,7 +156,7 @@ impl<'a> ClientHello<'a> {
     ///
     /// Returns `None` if the client did not supply a SNI.
     pub fn server_name(&self) -> Option<&DnsName<'_>> {
-        self.server_name.as_ref()
+        self.server_name.as_deref()
     }
 
     /// Get the compatible signature schemes.
@@ -435,6 +440,9 @@ pub struct ServerConfig {
     ///
     /// [RFC8779]: https://datatracker.ietf.org/doc/rfc8879/
     pub cert_decompressors: Vec<&'static dyn compress::CertDecompressor>,
+
+    /// Policy for how an invalid Server Name Indication (SNI) value from a client is handled.
+    pub invalid_sni_policy: InvalidSniPolicy,
 }
 
 impl ServerConfig {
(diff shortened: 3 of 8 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -435,6 +437,9 @@ pub struct ServerConfig {
     ///
     /// [RFC8779]: https://datatracker.ietf.org/doc/rfc8879/
     pub cert_decompressors: Vec<&'static dyn compress::CertDecompressor>,
+
+    /// Determines how invalid SNI in ClientHello is handled.
```
cpu: Editorial suggestion:

```suggestion
    /// Policy for how an invalid Server Name Indication (SNI) value from a client is handled.
```

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,50 @@ impl ServerConfig {
     }
 }
 
+/// This Policy determines how invalid SNI should be handled by the rustls server.
```
cpu:
```suggestion
/// A policy describing how invalid an invalid Server Name Indication (SNI) value from a client is handled by the server.
```

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,50 @@ impl ServerConfig {
...
+///
+/// The only valid form of SNI according to relevant RFCs ([RFC6066], [RFC1035]) is
+/// non-IP-address HostNames.
```
cpu:
```suggestion
/// non-IP-address host name, however some misconfigured clients may send a bare IP address, or
/// another invalid value. Some servers may wish to ignore these invalid values instead of producing
/// an error.
///
/// By default, Rustls will ignore invalid values that are an IP address (the most common misconfiguration)
/// and error for all other invalid values.
///
/// When an SNI value is ignored, Rustls treats the client as if it sent no SNI at all.
```

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,50 @@ impl ServerConfig {
...
+///
+/// [RFC1035]: https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.1
+/// [RFC6066]: https://datatracker.ietf.org/doc/html/rfc6066#section-3
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
+#[non_exhaustive]
+pub enum InvalidSniPolicy {
+    /// Reject all ClientHello messages that contain invalid SNI
```
cpu:
```suggestion
    /// Reject all ClientHello messages that contain an invalid SNI value.
```

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,50 @@ impl ServerConfig {
...
+    RejectAll,
+    /// Ignore SNI in ClientHello messages if they are IP addresses.
```
cpu:
```suggestion
    /// Ignore an invalid SNI value in ClientHello messages if the value is an IP address.
```

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,50 @@ impl ServerConfig {
...
+            .server_name
+            .as_ref()
+            .is_none_or(|sni| match self {
+                Self::RejectAll => matches!(sni, ServerNamePayload::SingleDnsName(_)),
+                Self::IgnoreIpAddresses => matches!(
+                    sni,
+                    ServerNamePayload::SingleDnsName(_) | ServerNamePayload::IpAddress
+                ),
+                Self::IgnoreAll => true,
+            })
```
cpu: I think this could be simplified with something like:

```rust
        let Some(server_name) = client_hello.server_name.as_ref() else {
            return true;
        };

        match (self, server_name) {
            (Self::RejectAll, ServerNamePayload::SingleDnsName(_)) => true,
            (Self::IgnoreIpAddresses, ServerNamePayload::SingleDnsName(_) | ServerNamePayload::IpAddress) => true,
            (Self::IgnoreAll, _) => true,
            _ => false
        }
```

WDYT?

cpu (commented):
Nice! Here's a first review pass. Mostly docs suggestions :-)

rustls/src/server/server_conn.rs
s-arash: Looks good!

rustls/src/server/hs.rs
```diff
@@ -333,6 +333,17 @@ impl ExpectClientHello {
...
             .supports_version(ProtocolVersion::TLSv1_2);
 
+        if !self
+            .config
+            .invalid_sni_policy
+            .is_acceptable(client_hello)
+        {
+            return Err(cx.common.send_fatal_alert(
+                AlertDescription::IllegalParameter,
+                InvalidMessage::InvalidServerName,
```
s-arash: Should I send `PeerMisbehaved::ServerNameMustContainOneHostName` here instead?

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,58 @@ impl ServerConfig {
...
+    /// Check if any potential invalid SNI in ClientHello is acceptable (i.e., ignorable)
+    /// by this policy.
+    ///
+    /// A `false` return value means there is an invalid SNI in ClientHello that is
+    /// rejected by this policy.
+    pub(super) fn is_acceptable(&self, client_hello: &ClientHelloPayload) -> bool {
+        let Some(server_name) = client_hello.server_name.as_ref() else {
+            return true;
+        };
+        match (self, server_name) {
```
s-arash: This is the clippy recommendation: 
```
matches!((self, server_name), (Self::RejectAll, ServerNamePayload::SingleDnsName(_)) | (
                 Self::IgnoreIpAddresses,
                 ServerNamePayload::SingleDnsName(_) | ServerNamePayload::IpAddress,
             ) | (Self::IgnoreAll, _))
```
Which I don't like. Should I apply this clippy recommendation, or work around it?

cpu: Yuck, I don't like that either :laughing: 

I would be tempted to allow the clippy lint here but we can also wait to see what djc and ctz think. Knowing djc he might have another round of code-golf to suggest that will side-step the concern with an even more improved formulation.

s-arash: @ctz @djc any suggestions?

rustls/src/server/hs.rs
djc: Why did you move this up? Ideally it should remain in the existing `match` that checks the server name validity, otherwise it is not locally obvious whether the `Invalid` case has been properly handled.

rustls/src/server/server_conn.rs
```diff
@@ -540,6 +545,58 @@ impl ServerConfig {
...
+    IgnoreAll,
+}
+
+impl InvalidSniPolicy {
+    /// Check if any potential invalid SNI in ClientHello is acceptable (i.e., ignorable)
+    /// by this policy.
+    ///
+    /// A `false` return value means there is an invalid SNI in ClientHello that is
+    /// rejected by this policy.
+    pub(super) fn is_acceptable(&self, client_hello: &ClientHelloPayload) -> bool {
```
djc: Nit: suggest simplifying this name to just `accept()`.

How about defining it more like this:

```rust
fn accept(&self, server_name: Option<&ServerNamePayload<'_>>) -> Result<Option<DnsName<'static>>, Error> { .. }
```

and moving the entire existing `match` into this function?

rustls/src/server/server_conn.rs
djc: Suggest doing it like this (as per suggestion above):

```rust
match (server_name, self) {
     (Some(ServerNamePayload::SingleDnsName(dns_name)), _) => Some(dns_name.to_lowercase_owned()),
     ...
}
```

rustls/tests/api.rs
```diff
@@ -7951,3 +7951,96 @@ impl ActiveKeyExchange for FakeHybridActive {
 }
 
 const CONFIDENTIALITY_LIMIT: u64 = 1024;
+
+#[test]
+fn server_invalid_sni_policy() {
+    const SERVER_NAME_GOOD: &str = "LXXXxxxXXXR";
+    const SERVER_NAME_BAAD: &str = "[XXXxxxXXX]";
```
djc: Nit: "BAAD" -> "BAD"?

djc:
(Please squash your commits.)

rustls/src/server/server_conn.rs
ctz: I also think this is OK, albeit I would have `match (self, server_name)` (but that is a pretty weak opinion.)

djc: Don't care all that much about the order here; I felt like the first (overwhelmingly) likely case ignoring `self` was a decent argument to have `server_name` self.

rustls/src/server/hs.rs
s-arash: The existing match is in `fn process_client_hello`, which happens before there is a server config. I moved it to the first function called after there is a server config.

rustls/src/server/server_conn.rs
s-arash: I can do that. Doing that would require changing the type of `ServerConnectionData::sni` from `DnsName` to `ServerNamePayload`. Would you like me to do that?

rustls/tests/api.rs
s-arash: I was trying to match the length of `SERVER_NAME_GOOD` 😅.

rustls/src/server/server_conn.rs
djc: That seems like it might cause a bunch of pain in APIs that rely on this field...

rustls/src/server/server_conn.rs
```diff
@@ -533,6 +538,54 @@ impl ServerConfig {
...
+        let Some(server_name) = client_hello.server_name.as_ref() else {
+            return true;
+        };
+        match server_name {
```
s-arash: I changed this match expression to silence Clippy. Let me know if this looks OK!

rustls/src/server/server_conn.rs
ctz: > I can do that. Doing that would require changing the type of `ServerConnectionData::sni` from `DnsName` to `ServerNamePayload`. Would you like me to do that?

Sorry, I'm not sure I understand the link here?

s-arash: Sorry I guess I misunderstood the ask. 
We just want to change the fn signature, but not actually use the returned `DnsName`? I say that because `fn process_client_hello` sets the SNI on the `ServerContext`, which happens before there is a server config, and therefore before we can call `InvalidSniPolicy::accept`.
And I think it needs to happen there, because `process_client_hello` is also called to construct an `Accepted` value in the `Acceptor` API; and `Accepted` needs the SNI.

djc: Okay, so we can't have `InvalidSniPolicy::accept()` take `ServerNamePayload` because we need `DnsName` in `ClientHello`, before we have the config? I think we could instead have `Accepted::client_hello()` take the `DnsName` directly from `payload`? Then we can postpone storing the `DnsName` into `ServerConnectionData` until later, right?

rustls/src/msgs/handshake.rs
```diff
@@ -367,6 +367,35 @@ impl<'a> From<&DnsName<'a>> for ServerNamePayload<'static> {
...
+    }
+}
+
+impl From<ServerNameNormalized> for DnsName<'static> {
+    fn from(server_name: ServerNameNormalized) -> Self {
+        server_name.0
+    }
+}
+
+impl<'a> TryFrom<&'a ServerNamePayload<'a>> for ServerNameNormalized {
```
djc: Adding a type for this seems overly complex to me. Suggest making it `ServerNamePayload::into_dns_name() -> Option<DnsName<'_>>`.

rustls/src/server/hs.rs
```diff
@@ -710,40 +718,16 @@ pub(super) fn process_client_hello<'m>(
...
-    // but then act like the client sent no `server_name` extension.
-    //
-    // [RFC6066]: https://datatracker.ietf.org/doc/html/rfc6066#section-3
-    let sni = match &client_hello.server_name {
-        Some(ServerNamePayload::SingleDnsName(dns_name)) => Some(dns_name.to_lowercase_owned()),
-        Some(ServerNamePayload::IpAddress) => None,
-        Some(ServerNamePayload::Invalid) => {
-            return Err(cx.common.send_fatal_alert(
-                AlertDescription::IllegalParameter,
-                PeerMisbehaved::ServerNameMustContainOneHostName,
```
djc, line 687: It seems surprising that `ServerNameMustContainOneHostName` doesn't appear in the new version anymore.

s-arash: I'll have `InvalidSniPolicy::accept` return this error instead, which would keep the existing behavior.

rustls/src/msgs/handshake.rs
s-arash: Good suggestion!

rustls/src/server/server_conn.rs
```diff
@@ -964,8 +1025,13 @@ impl Accepted {
     /// Get the [`ClientHello`] for this connection.
     pub fn client_hello(&self) -> ClientHello<'_> {
         let payload = Self::client_hello_payload(&self.message);
+        let server_name = payload
+            .server_name
+            .as_ref()
+            .and_then(ServerNamePayload::to_dns_name_normalized)
```
s-arash, line 1061: I'm not super happy that we allocate for `server_name` every time this public function is called now. But not sure how to avoid it if we don't want `fn process_client_hello` to set the `ServerConnectionData::sni` field.

ctz (approved):
I think this is alright, and I'm specifically happy with the public API changes. Maybe the internal changes could be different but determining that is not easy in a review setting, so maybe I'll experiment with that later. (The ideas there are whether we can make use of `impl Eq for DnsName` being case insensitive, or perhaps remove the dependency on `cx.data.sni` in the "varying SNI check" code.

rustls/src/msgs/handshake.rs
```diff
@@ -284,6 +284,16 @@ impl ServerNamePayload<'_> {
     const SIZE_LEN: ListLength = ListLength::NonZeroU16 {
         empty_error: InvalidMessage::IllegalEmptyList("ServerNames"),
     };
+
+    /// Get the `DnsName` out of this `ServerNamePayload` if it contains one.
+    /// The returned `DnsName` will be normalized (converted to lowercase).
+    pub(crate) fn to_dns_name_normalized(&self) -> Option<DnsName<'static>> {
+        match self {
+            ServerNamePayload::SingleDnsName(dns_name) => Some(dns_name.to_lowercase_owned()),
```
djc: Not sure why this isn't triggering `clippy::use_self`...

rustls/src/server/server_conn.rs
```diff
@@ -533,6 +541,59 @@ impl ServerConfig {
...
+    /// "Ignoring SNI" means accepting the ClientHello message, but acting as if the client sent no SNI.
+    IgnoreAll,
+}
+
+impl InvalidSniPolicy {
+    /// Returns the valid SNI value, or ignores the invalid SNI value if allowed by this policy; otherwise returns
+    /// an error.
+    pub(super) fn accept(
+        &self,
+        server_name_payload: Option<&ServerNamePayload<'_>>,
```
djc: Nit: rename this to `payload`, no need to precisely reflect the type name.

rustls/src/server/server_conn.rs
```diff
@@ -533,6 +541,59 @@ impl ServerConfig {
...
+    ) -> Result<Option<DnsName<'static>>, Error> {
+        let Some(server_name_payload) = server_name_payload else {
```
djc: Nit: and the variable, too.

djc (approved):
Alright, some small nits -- and can you squash this? Then we can merge it.

djc:
(FYI, if you want this on 0.23, it will need a backport.)

s-arash:
> (FYI, if you want this on 0.23, it will need a backport.)

We probably won't need it in 0.23. I can open a backport PR if you'd like me to.

rustls/src/server/server_conn.rs
```diff
@@ -533,6 +541,59 @@ impl ServerConfig {
...
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
+#[non_exhaustive]
+pub enum InvalidSniPolicy {
+    /// Reject all ClientHello messages that contain an invalid SNI value.
+    RejectAll,
+    /// Ignore an invalid SNI value in ClientHello messages if the value is an IP address.
+    ///
+    /// "Ignoring SNI" means accepting the ClientHello message, but acting as if the client sent no SNI.
+    #[default]
+    IgnoreIpAddresses,
```
s-arash, line 567: Speaking of a new version, with this change we have the opportunity to make `InvalidSniPolicy::RejectAll` the default, since there is no backcompat concern. What do the maintainers think?

djc: That's actually relatively recent:

- #1881

I think it's probably better to keep it for compatibility with the wider ecosystem.

rustls/src/server/hs.rs
```diff
@@ -664,40 +671,16 @@ pub(super) fn process_client_hello<'m>(
...
-            ));
+    if done_retry {
+        let ch_sni = client_hello
+            .server_name
+            .as_ref()
+            .and_then(ServerNamePayload::to_dns_name_normalized);
+        if cx.data.sni != ch_sni {
+            return Err(PeerMisbehaved::ServerNameDifferedOnRetry.into());
```
s-arash, line 680: Should we send an alert here?

rustls/src/msgs/handshake.rs
ctz: `Self` here refers to a different lifetime

djc:
> > (FYI, if you want this on 0.23, it will need a backport.)
> 
> We probably won't need it in 0.23. I can open a backport PR if you'd like me to.

If you don't need it, let's not bother.

rustls/src/server/hs.rs
ctz: There doesn't seem to be one specified in the standard, so if we did it would be one of the very generic alerts (ie, "invalid parameter") so I think this is OK.

s-arash: Got it!

shraddha-1508:
Which version is this feature targeted for? Looking forward to having this feature available.

djc:
> Which version is this feature targeted for? Looking forward to having this feature available.

0.24.0.

PR #2618, opened 2025-09-03
Add API for early exporter

### LINKED ISSUE #2406: TLS 1.3 early_exporter_master_secret

ctz:
This provides:

- `fn exporter()` on `WriteEarlyData` for clients
- `fn exporter()` on `ReadEarlyData` for servers

Since BoringSSL/bogo does not support this (though Go does!) I added an adhoc test in openssl-tests.

fixes #2406

### DIFF
--- a/openssl-tests/src/early_exporter.rs
+++ b/openssl-tests/src/early_exporter.rs
@@ -1,0 +1,148 @@
+use std::io::{Read, Write};
+use std::net::{TcpListener, TcpStream};
+use std::sync::Arc;
+use std::{str, thread};
+
+use openssl::ssl::{SslConnector, SslMethod, SslSession, SslStream};
+use rustls::ServerConfig;
+use rustls::crypto::aws_lc_rs as provider;
+use rustls::pki_types::pem::PemObject;
+use rustls::pki_types::{CertificateDer, PrivateKeyDer};
+
+use crate::utils::verify_openssl3_available;
+
+#[test]
+fn test_early_exporter() {
+    verify_openssl3_available();
+
+    // full handshake + one resumption
+    const ITERS: usize = 2;
+
+    let listener = TcpListener::bind(("localhost", 0)).unwrap();
+    let port = listener.local_addr().unwrap().port();
+
+    let server_thread = thread::spawn(move || {
+        let mut config = ServerConfig::builder_with_provider(provider::default_provider().into())
+            .with_no_client_auth()
+            .with_single_cert(load_certs(), load_private_key())
+            .unwrap();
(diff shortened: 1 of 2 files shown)


openssl-tests/src/early_exporter.rs
```diff
@@ -0,0 +1,150 @@
...
+
+            server
+                .complete_io(&mut tcp_stream)
+                .unwrap();
+
+            tcp_stream.flush().unwrap();
+        }
+    });
+
+    let mut connector = openssl::ssl::SslConnector::builder(SslMethod::tls()).unwrap();
```
cpu: nit: import `SslConnector`?

openssl-tests/src/early_exporter.rs
```diff
@@ -0,0 +1,150 @@
...
+        let stream = TcpStream::connect(("localhost", port)).unwrap();
+
+        let mut config = connector.configure().unwrap();
+        config.set_connect_state();
+
+        if let Some(sess) = &session {
+            unsafe { config.set_session(sess).unwrap() };
+        }
+
+        #[allow(deprecated)] // doesn't seem to be a way to call `write_early_data` early enough?
```
cpu: not sure I understand this comment fully. There's not a way to call `write_early_data()` early enough with the non-deprecated alternative (whatever it is)?

ctz: I was struggling to understand this API. It turns out `SslStream` is the thing I wanted, and you can make one outside calling `connect()`.

PR #2620, opened 2025-09-04
Introduce `ApiMisuse` error

ctz:
Follow-up from #2608

In the direction of #889

### DIFF
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -1154,6 +1154,11 @@ pub enum ApiMisuse {
     ///
     /// [`Acceptor::accept()`]: crate::server::Acceptor::accept()
     AcceptorPolledAfterCompletion,
+
+    /// Incorrect sample length provided to [`quic::HeaderProtectionKey::encrypt_in_place()`][]
+    ///
+    /// [`quic::HeaderProtectionKey::encrypt_in_place()`]: crate::quic::HeaderProtectionKey::encrypt_in_place()
+    InvalidQuicHeaderProtectionSampleLength,
 }
 
 impl From<ApiMisuse> for Error {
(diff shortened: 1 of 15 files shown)


ctz:
> [rustls / Run benchmarks (pull_request)](https://github.com/rustls/rustls/actions/runs/17487119362/job/49668367329?pr=2620)Failing after 1m

Regression in nightly-2025-09-05 compared to nightly-2025-09-04. Assuming this will be fixed by https://github.com/rust-lang/rust/pull/146133 but if not I'll file a bug

rustls/tests/api.rs
```diff
@@ -4632,6 +4632,14 @@ fn early_data_is_available_on_resumption() {
         .unwrap()
         .exporter()
         .unwrap();
+    assert_eq!(
+        Some(Error::ApiMisuse(ApiMisuse::ExporterAlreadyUsed)),
```
djc: Nit: I guess there's quite a bit of this in the existing code, but personally I think observed to be on left and expected on the right is more idiomatic. (I forget if we've had this discussion before?)

rustls/src/error.rs
```diff
@@ -1166,6 +1166,9 @@ pub enum ApiMisuse {
 
     /// QUIC attempted with a configuration that does not support TLS1.3.
     QuicApiRequiresTls13Support,
+
+    /// QUIC attempted with a configuration that does not support a ciphersuite that supports QUIC.
+    QuicApiRequiresAtLeastOneCipherSuiteSupportingQuic,
```
djc: Can we get this a bit more concise? `NoQuicCompatibleCipherSuites`?

Maybe the `Api` part of the `QuicApi` prefix also doesn't add much value?

djc (commented):
How much `General` usage does this leave? Is getting rid of all of it too ambitious?

cpu (approved):
Nice! Thank you

ctz:
> How much `General` usage does this leave?

14 in the core crate.

PR #2623, opened 2025-09-06
QUIC: add resumption data set and read via quic connection

wanlebing:
The application layer can provision TLS session resumption data via QUIC connections. A common use case is QUIC 0-RTT, where resumption data serves to store and retrieve 0-RTT parameters  which is bound to session ticket.

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -117,6 +117,42 @@ mod connection {
                 Self::Server(conn) => conn.core.exporter(),
             }
         }
+
+        /// Set the resumption data to embed in future resumption tickets supplied to the client.
+        ///
+        /// This method is only available for server connections. For client connections,
+        /// resumption data is received from the server and cannot be set.
+        ///
+        /// Defaults to the empty byte string. Must be less than 2^15 bytes to allow room for other
+        /// data. Should be called while `is_handshaking` returns true to ensure all transmitted
+        /// resumption tickets are affected.
+        ///
+        /// Integrity will be assured by rustls, but the data will be visible to the client. If secrecy
+        /// from the client is desired, encrypt the data separately.
+        pub fn set_resumption_data(&mut self, data: &[u8]) -> Result<(), Error> {
+            match self {
+                Self::Server(conn) => {
+                    conn.set_resumption_data(data);
+                    Ok(())
+                }
+                Self::Client(_) => Err(Error::General(
+                    "set_resumption_data is only available for server connections".into(),
+                )),
+            }
+        }
+
+        /// Retrieves the resumption data supplied by the peer, if any.
+        ///
+        /// For server connections, returns `Some` if and only if a valid resumption ticket
+        /// has been received from the cli
+        /// For client connections, this method is not typically used as clients receive
+        /// resumption tickets from servers.
+        pub fn received_resumption_data(&self) -> Option<&[u8]> {
+            match self {
+                Self::Server(conn) => conn.received_resumption_data(),
+                Self::Client(_) => None,
+            }
+        }
     }
 
     impl Deref for Connection {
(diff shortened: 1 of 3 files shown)


rustls/src/quic.rs
```diff
@@ -117,6 +117,42 @@ mod connection {
...
+        /// from the client is desired, encrypt the data separately.
+        pub fn set_resumption_data(&mut self, data: &[u8]) -> Result<(), Error> {
+            match self {
+                Self::Server(conn) => {
+                    conn.set_resumption_data(data);
+                    Ok(())
+                }
+                Self::Client(_) => Err(Error::General(
+                    "set_resumption_data is only available for server connections".into(),
+                )),
```
cpu: Once https://github.com/rustls/rustls/pull/2620 lands I think this will be a good candidate for an `ApiMisuse` error. Ditto for the other `Error::General` usages.

rustls/src/quic.rs
```diff
@@ -117,6 +117,42 @@ mod connection {
...
+            }
+        }
+
+        /// Retrieves the resumption data supplied by the peer, if any.
+        ///
+        /// For server connections, returns `Some` if and only if a valid resumption ticket
+        /// has been received from the cli
```
cpu: Small typo here: "cli" -> "client."

rustls/src/quic.rs
```diff
@@ -313,6 +349,31 @@ mod connection {
...
+        /// Retrieves the resumption data supplied by the client, if any.
+        ///
+        /// Returns `Some` if and only if a valid resumption ticket has been received from the client.
+        pub fn received_resumption_data(&self) -> Option<&[u8]> {
+            self.inner
+                .core
+                .data
+                .received_resumption_data
+                .as_ref()
+                .map(|x| &x[..])
```
cpu: I think this can be simplified using `as_deref()`.

rustls/tests/api.rs
```diff
@@ -5266,6 +5266,62 @@ mod test_quic {
...
+            client_config,
+            quic::Version::V1,
+            server_name("localhost"),
+            client_params.to_vec(),
+        )
+        .unwrap();
+
+        let mut client_conn = quic::Connection::Client(client);
+        assert!(client_conn.set_resumption_data(test_data).is_err());
+        assert_eq!(client_conn.received_resumption_data(), None);
```
cpu: It'd be nice to see test coverage that has `received_resumption_data()` returning `Some(_)`

cpu (commented):
Thanks for the PR. Without thinking too much about the API or end goal here's some initial feedback.

FWIW there's ~three CI failures, but only ([the formatting check](https://github.com/rustls/rustls/actions/runs/17514981178/job/49753217764?pr=2623)) is relevant. The audit one will be addressed by https://github.com/rustls/rustls/pull/2624 and the benchmarking one is an unrelated nightly regression.

rustls/src/quic.rs
```diff
@@ -117,6 +117,42 @@ mod connection {
...
+        /// Integrity will be assured by rustls, but the data will be visible to the client. If secrecy
+        /// from the client is desired, encrypt the data separately.
+        pub fn set_resumption_data(&mut self, data: &[u8]) -> Result<(), Error> {
+            match self {
+                Self::Server(conn) => {
+                    conn.set_resumption_data(data);
+                    Ok(())
+                }
+                Self::Client(_) => Err(Error::General(
+                    "set_resumption_data is only available for server connections".into(),
```
ctz: I think this means this function should be defined only on `ServerConnection`; a caller who has a `Connection` can do this match themselves.

rustls/src/quic.rs
```diff
@@ -117,6 +117,42 @@ mod connection {
...
+        /// Retrieves the resumption data supplied by the peer, if any.
+        ///
+        /// For server connections, returns `Some` if and only if a valid resumption ticket
+        /// has been received from the cli
+        /// For client connections, this method is not typically used as clients receive
+        /// resumption tickets from servers.
+        pub fn received_resumption_data(&self) -> Option<&[u8]> {
+            match self {
+                Self::Server(conn) => conn.received_resumption_data(),
+                Self::Client(_) => None,
```
ctz: Ditto, this means this function should be defined only on `ServerConnection`.

rustls/src/quic.rs
wanlebing: Ok

rustls/src/quic.rs
wanlebing: Ok

rustls/src/quic.rs
wanlebing: Ok

rustls/tests/api.rs
wanlebing: Okay

rustls/src/quic.rs
wanlebing: ok

ctz:
(just needs a `cargo fmt`)

PR #2625, opened 2025-09-06
error: display other error for Display

BiagioFesta:
`OtherError` implements `Display`, why not using for `Display` of `CertificateError`?

### DIFF
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -765,6 +765,8 @@ impl fmt::Display for CertificateError {
                 Ok(())
             }
 
+            Self::Other(other) => write!(f, "{other}"),
+
             other => write!(f, "{other:?}"),
         }
     }


BiagioFesta:
CI failing but don't think because of this change

rustls/src/error.rs
```diff
@@ -765,7 +765,7 @@ impl fmt::Display for CertificateError {
                 Ok(())
             }
 
-            other => write!(f, "{other:?}"),
+            other => write!(f, "{other}"),
```
ctz: This needs to be different in some way. Either we need to write a full set of handling for each variant, and then call `Display` on the value contained in the `Other(_)` variant, or we need to pick out the `Other(_)` variant and keep calling `Debug` on the catch-all case.

At the moment it unconditionally recurses forever.

ctz (changes requested):

rustls/src/error.rs
BiagioFesta: sorry about that, I didn’t look through it carefully (took it a bit lightly since it seemed like a trivial change)

ctz (approved):

ctz:
Thanks!

PR #2626, opened 2025-09-07
Use nightly compiler less

ctz:
Nightly is pretty regularly broken (not surprising), so prefer to use stable+RUSTC_BOOTSTRAP=1 in cases where we want to use an unstable feature, but not a bleeding-edge compiler.


djc (approved):
Good idea!

PR #2628, opened 2025-09-08
Clarify and extend tests for plaintext buffer limits

ctz:
This contains and therefore fixes #2382

### DIFF
--- a/rustls/tests/api.rs
+++ b/rustls/tests/api.rs
@@ -6607,9 +6607,13 @@ fn test_secret_extraction_disabled_or_too_early() {
 
 #[test]
 fn test_received_plaintext_backpressure() {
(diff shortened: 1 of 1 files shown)


rustls/tests/api.rs
```diff
@@ -6607,95 +6607,104 @@ fn test_secret_extraction_disabled_or_too_early() {
 
 #[test]
 fn test_received_plaintext_backpressure() {
-    let kt = KeyType::Rsa2048;
-    let provider = provider::default_provider();
-    let plaintext_limit = 16_384;
+    fn test(limit: Option<usize>, plaintext_limit: usize) {
```
djc: Maybe this becomes a top-level helper function? The extra indentation level is a bit of a bummer IMO.

Issue #916, opened 2022-01-10
Highlight always-taken branches (especially missing `else`) in code coverage metrics 


briansmith:
Our code coverage measurements are really rudimentary The way coverage information is displayed is optimized for finding cases where branches are never taken. However, it is important that we find cases where a branch is always taken, as in the following pattern where there is no else:

```
if condition {
     do something
}
```
We need a way to visualize that the empty implicit `else {}` is never taken.

#904 is a concrete example of a branch where we never test the implicit `else {}`:

```rust
  let mut exts = Vec::new();
    if !supported_versions.is_empty() {
        exts.push(ClientExtension::SupportedVersions(supported_versions));
    }
```

It would actually be very wrong for us to omit the "supported versions" extension, so doing it conditionally is wrong. It seems like `!supported.versions.is_empty()` likely always evaluates to `true`, so we are likely to be OK, but this is far from obvious. Regardless, the `ext.push()` should be done unconditionally.

If we can get code coverage presentation to highlight more of these cases, we might find more serious bugs in "100% covered" code.

PR #2629, opened 2025-09-08
Enable branch coverage measurement

ctz:
This enables branch coverage collection. The llvm html output is very detailed but inscrutable. The codecov output is very simple but clear at least.

This decreases our coverage numbers because "partial" lines are no longer counted toward the headline line coverage (which is right, since only part of the line was executed). But I think this is accurate and good, actually.

re #916

### DIFF
--- a/admin/coverage
+++ b/admin/coverage
@@ -2,7 +2,11 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


ctz:
I think this looks good but makes me want to address some coverage shortfalls.

PR #2636, opened 2025-09-10
verify: introduce PeerIdentity type

djc:
Next step towards the vision from

- #2560

Previously:

- #2634 (-25)
- #2633 (+26)
- #2632 (-48)
- #2614 (-78)

This does expose the existence of non-certificate identities in the `CommonState::peer_identity()` API that replaces `peer_certificates()`, but otherwise confines the complexity mostly to verifier implementations and resumption. I don't think this is the end state, there is probably more that can be done here.

It might also be possible to shave off smaller commit.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -17,7 +17,9 @@ use crate::common_state::{CommonState, HandshakeKind, KxState, Side, State};
 use crate::conn::ConnectionRandoms;
 use crate::conn::kernel::{Direction, KernelContext, KernelState};
 use crate::crypto::KeyExchangeAlgorithm;
-use crate::enums::{AlertDescription, ContentType, HandshakeType, ProtocolVersion};
+use crate::enums::{
+    AlertDescription, CertificateType, ContentType, HandshakeType, ProtocolVersion,
+};
 use crate::error::{Error, InvalidMessage, PeerIncompatible, PeerMisbehaved};
 use crate::hash_hs::HandshakeHash;
 use crate::log::{debug, trace, warn};
@@ -35,7 +37,7 @@ use crate::suites::PartiallyExtractedSecrets;
 use crate::sync::Arc;
 use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
 use crate::verify::{
-    self, CertificateIdentity, DigitallySignedStruct, ServerIdentity, SignatureVerificationInput,
+    self, DigitallySignedStruct, PeerIdentity, ServerIdentity, SignatureVerificationInput,
 };
 
 mod server_hello {
@@ -169,12 +171,7 @@ mod server_hello {
 
                     // Since we're resuming, we verified the certificate and
                     // proof of possession in the prior session.
-                    cx.common.peer_certificates = Some(
-                        resuming
-                            .server_cert_chain()
-                            .clone()
-                            .into_owned(),
-                    );
+                    cx.common.peer_identity = Some(resuming.peer_identity().clone());
                     cx.common.handshake_kind = Some(HandshakeKind::Resumed);
                     let cert_verified = verify::ServerCertVerified::assertion();
                     let sig_verified = verify::HandshakeSignatureValid::assertion();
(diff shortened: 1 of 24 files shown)


rustls/src/client/tls12.rs
```diff
@@ -883,20 +878,19 @@ impl State<ClientConnectionData> for ExpectServerDone<'_> {
         // 5. emit a Finished, our first encrypted message under the new keys.
 
         // 1.
-        let (end_entity, intermediates) = st
-            .server_cert
-            .cert_chain
-            .split_first()
-            .ok_or(Error::NoCertificatesPresented)?;
```
djc, line 890: No certificates, client/1.2: yield `Error::NoCertificatesPresented`.
No certificates, client/1.3: yield `Error::NoCertificatesPresented`.
No certificates, server/1.2 (if `mandatory`): send fatal alert `CertificateRequired`, yield `Error::NoCertificatesPresented`.
No certificates, server/1.3 (if `mandatory`): send fatal alert `CertificateRequired`, yield `Error::NoCertificatesPresented`.

Should the client send an alert for this, too?

ctz: I checked and there is no RFC-defined alert for this. `DecodeError` is not right because an empty `Certificate` message is allowed by the grammar. `CertificateRequired` is only specified to be sent by servers. So probably `BadCertificate`?

rustls/src/msgs/base.rs
```diff
@@ -81,6 +81,20 @@ impl fmt::Debug for Payload<'_> {
...
 }
 
+impl<'a> Codec<'a> for SubjectPublicKeyInfoDer<'a> {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        codec::u24(self.as_ref().len() as u32).encode(bytes);
+        bytes.extend(self.as_ref());
+    }
+
+    fn read(r: &mut Reader<'a>) -> Result<Self, InvalidMessage> {
+        let len = codec::u24::read(r)?.0 as usize;
```
ctz:
```suggestion
        let len = Self::SIZE_LEN.read(r)?;
```

rustls/src/msgs/codec.rs
```diff
@@ -274,6 +277,17 @@ impl Codec<'_> for () {
     }
 }
 
+impl TlsListElement for CertificateDer<'_> {
+    const SIZE_LEN: ListLength = ListLength::U24 {
```
ctz: Preexisting: this ought to be `ListLength::NonZeroU24`

rustls/src/common_state.rs
```diff
@@ -140,8 +139,8 @@ impl CommonState {
     /// Note: the return type of the 'certificate', when using raw public keys is `CertificateDer<'static>`
     /// even though this should technically be a `SubjectPublicKeyInfoDer<'static>`.
     /// This choice simplifies the API and ensures backwards compatibility.
-    pub fn peer_certificates(&self) -> Option<&[CertificateDer<'static>]> {
-        self.peer_certificates.as_deref()
+    pub fn peer_identity(&self) -> Option<&PeerIdentity> {
+        self.peer_identity.as_ref()
```
ctz, line 130: This change should also change the doc comment.

rustls/src/server/tls13.rs
```diff
@@ -908,6 +913,7 @@ struct ExpectCertificateOrCompressedCertificate {
     suite: &'static Tls13CipherSuite,
     key_schedule: KeyScheduleTrafficWithClientFinishedPending,
     send_tickets: usize,
+    expected_type: Option<CertificateType>,
```
ctz: Matter of opinion, but I would include `identity`, `cert` or `certificate` in this field name in exchange for `expected`

rustls/src/client/tls12.rs
djc: Going to handle this in a followup PR.

rustls/src/server/tls13.rs
djc: Renamed to `expected_certificate_type` to stick with the RFC vocabulary.

rustls/src/common_state.rs
djc: Updated the comment. Removed some stuff which is now better explained in `CertificateIdentity`.

rustls/src/msgs/codec.rs
djc: Added a prefix commit for this. It seems to fail some client certificate resolver tests, presumably because we're using empty certificates in some of these tests, but I couldn't figure out where/why/how.

rustls/src/msgs/base.rs
djc: Fixed this up in a prefix commit.

rustls/src/msgs/codec.rs
djc: Moved this to

- #2637

PR #2637, opened 2025-09-10
Enforce non-zero certificate lengths

djc:
Extracted from

- #2636

### DIFF
--- a/rustls/src/msgs/codec.rs
+++ b/rustls/src/msgs/codec.rs
@@ -310,6 +310,13 @@ pub(crate) enum ListLength {
 
     /// U24 with imposed upper bound
     U24 { max: usize, error: InvalidMessage },
+
+    /// U24 but non-empty, with imposed upper bound
+    NonZeroU24 {
+        max: usize,
+        empty_error: InvalidMessage,
+        too_many_error: InvalidMessage,
+    },
 }
 
 impl ListLength {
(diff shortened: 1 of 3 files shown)


rustls/src/msgs/codec.rs
```diff
@@ -274,6 +275,14 @@ impl Codec<'_> for () {
     }
 }
 
+impl TlsListElement for CertificateDer<'_> {
+    const SIZE_LEN: ListLength = ListLength::NonZeroU24 {
```
ctz: Sorry, this one should be `ListLength::U24`. This line is the second half of this definition in TLS1.2:

```
      opaque ASN.1Cert<1..2^24-1>;

      struct {
          ASN.1Cert certificate_list<0..2^24-1>;
      } Certificate;
```

In other words, the list of certificates can be empty, but individual certificates cannot.

The `impl Codec for CertificateDer...` part above, is the first. So actually that needs _not_ to use `Self::SIZE_LEN` (which is the size prefix for a vector of `CertificateDer`s, and that can be empty)  but instead a `ListLength::NonZeroU24 { ... }`

djc: Ah yeah, I figured it was something like that but wanted to focus first on fixing the other PR.

ctz (approved):

PR #2640, opened 2025-09-11
Improve errors for empty certificate chains

djc:
Followup from

- #2636


djc:
Reduced this to a single commit, since the other one was deemed to be controversial by the test suite.

ctz:
I think the previous commit was OK, but it needs to be inside a closure with `.ok_or_else` because otherwise the alert is sent in the success case.

(Clippy should be able to lint on this -- what's up with that?)

djc:
> I think the previous commit was OK, but it needs to be inside a closure with `.ok_or_else` because otherwise the alert is sent in the success case.

Ah, nice catch!

> (Clippy should be able to lint on this -- what's up with that?)

- https://github.com/rust-lang/rust-clippy/issues/15677

ctz (approved):

PR #2644, opened 2025-09-12
Derive protocol version from cipher suite

djc:
Seems nice to avoid storing duplicate state?


ctz:
I guess CI is showing a case in early data where the protocol version is committed, but the ciphersuite isn't yet. So if we're going to hang the version off the ciphersuite then we might need to commit that earlier, and possibly be prepared for that to change once we hear from the server.

djc:
Right, I was looking at this yesterday but might not be worth doing right now.

PR #2645, opened 2025-09-13
Reduce use of SupportedCipherSuite

djc:
This gets rid of some of the `versions` indirection stuff, not sure if that preserves the linker DCE.

Would like some feedback before I dig into the remaining test failures! (Could also split this into multiple PRs if that helps -- it got away from me a little.)


rustls/src/crypto/mod.rs
```diff
@@ -582,9 +582,9 @@ pub trait ActiveKeyExchange: Send + Sync {
     fn complete_for_tls_version(
         self: Box<Self>,
         peer_pub_key: &[u8],
-        tls_version: &SupportedProtocolVersion,
+        tls_version: ProtocolVersion,
     ) -> Result<SharedSecret, Error> {
-        if tls_version.version() != ProtocolVersion::TLSv1_2 {
+        if tls_version != ProtocolVersion::TLSv1_2 {
```
ctz: Note this is a wider type, so `complete_for_tls_version(..., ProtocolVersion::TLSv1_0)` is admitted now. So this branch is probably more accurately `if tls_version >= ProtocolVersion::TLSv1_3` perhaps.

rustls/src/tls12.rs
```diff
@@ -86,8 +86,13 @@ impl Tls12CipherSuite {
 }
 
 impl Suite for Tls12CipherSuite {
-    fn handler(&self) -> &'static dyn crate::server::ServerHandler<Self> {
-        crate::server::TLS12_HANDLER
+    fn handlers(
```
ctz: The failure of `limitedclient_does_not_contain_tls12_code` bisects to this commit.

djc:
The benchmarks don't look great. Are these diffs big enough to matter? Should I experiment with `&dyn Suite` instead of `T: Suite` to avoid duplicating a bunch of code?

ctz:
> The benchmarks don't look great. Are these diffs big enough to matter? Should I experiment with `&dyn Suite` instead of `T: Suite` to avoid duplicating a bunch of code?

I don't think these are large differences, and we are still up over this set of changes as a whole (see https://github.com/rustls/rustls/pull/2656)

djc:
CI is passing now -- thanks to @taiki-e for fixing the cargo-minimal-versions issue!

Issue #2569, opened 2025-07-17
Delegated Credentials Support


sjcobb2022:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
Delegated Credentials are currently unupported.

**Describe the solution you'd like**
Following https://datatracker.ietf.org/doc/rfc9345/

 **Work**
I have created a fork to start working on implementing this extension. This issue is purely for visibility.

If this is not in the spirit of rustls or there are any comments, please reach out.

djc:
Thanks -- personally I think this might make sense for rustls, but other maintainers may feel differently.

I suggest you share something of a design/implementation plan for early feedback before finishing your implementation.

Also, please make sure you're aware of our [contributing guidelines](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md). In particular, for a larger feature like this, it would probably help to do preliminary refactoring in smaller commits/PRs that you submit for feedback eagerly.

sjcobb2022:
Sounds good. Thanks @djc

sjcobb2022:
Did not mean to close this. The comment button and close issue are too close together haha

djc:
Do you have a particular use case for this? Why do you need this?

sjcobb2022:
Currently I am writing a http client that can avoid being fingerprinted, similar to [this](https://github.com/lwthiker/curl-impersonate).

delegated credentials are used by firefox and adjacent browsers.

djc:
Are you aware that rustls doesn't intend to support TLS fingerprinting workaround mechanisms?

sjcobb2022:
I was not aware of this, however I would still be interested in contributing.

sjcobb2022:
I mean the things that I need to configure are:

- cipher suites (including order)
- compression
- TLS versions
- extensions

At least for firefox, all the necessary extensions are available and configurable.  

The only extension missing that I require to have complete control of the JA3 fingerprint is delegated credentials

djc:
See the pinned issue #2498.

sjcobb2022:
I see. regardless it would be nice to get my hands dirty with implementating it in rustls. I will most likely move to craftls in the long run though.

PR #2646, opened 2025-09-13
RFC: Initial client-side delegated credentials

sjcobb2022:
Hi there,

As mentioned in #2569 I have started on the delegated credentials implementation. 

This is a relatively rough PR, without comprehensive testing or anything. Mainly just opening it for some direction and to determine whether the team thinks that this is necessary / worth investing time in this route.

If necessary, I can split this PR further. Perhaps into just the type definitions (stuff in src/msgs), and then separate out client and server impls after that. Both the client and server use very similar validation, for which we could create a shared function.

There might be some other things that I have missed, for example not sure if we should do DTLS1.3 + TLS1.3 for DC or just TLS1.3

#### TODOS:
- [ ] e2e tests for client side handshake
- [ ] Determine if correct error types are used
- [ ] Cleanup comments (most copied from RFC for reference whilst coding)
- [ ] General cleanup (potential to move some logic to shared place for client / server validation
- [ ] DelegationUsage and KeyUsage as X.509 extension

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -228,7 +228,11 @@ fn emit_client_hello_for_retry(
         ),
         // RFC 9345: a client willing to use delegated credentials SHALL send this extension
         delegated_credential_algorithms: if supported_versions.tls13 {
-            Some(config.verifier.supported_verify_schemes())
+            Some(
+                config
+                    .verifier
+                    .supported_verify_schemes(),
+            )
         } else {
             None
         },
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -226,6 +226,12 @@ fn emit_client_hello_for_retry(
                 .verifier
                 .supported_verify_schemes(),
         ),
+        // RFC 9345: a client willing to use delegated credentials SHALL send this extension
+        delegated_credential_algorithms: if supported_versions.tls13 {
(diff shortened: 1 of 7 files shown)


djc:
I don't think we would merge this if you only need this to pass fingerprinting tests.

sjcobb2022:
> I don't think we would merge this if you only need this to pass fingerprinting tests.

This is not for any fingerprinting, more as a personal challenge to see if I can implement it here.

ctz:
I'm afraid I don't think delegated credentials has wide-enough (or growing) deployment to make sense for this library. AIUI only Digicert, Cloudflare, and Firefox support it; so that alone is not enough for any web deployment. I guess a non-webpki non-web browser deployment would be possible, but it seems this is not what we're talking about.

sjcobb2022:
Alright I'll close this PR. Thanks for the feedback.

cpu:
If you decide to contribute work in the future I think it would also be helpful to look at how the commit history for other merged feature branches look compared to your own. If this branch was going to continue I think it would have been hard to review in the current state.

sjcobb2022:
Yeh I'm aware that it was a total mess haha. I was not expecting this PR to get merged, as it was mainly a POC to see if there was interest in the feature.

Thanks for the feedback, appreciate it.

PR #2647, opened 2025-09-15
rustls-post-quantum: track local crate

ctz:


djc:
What's the motivation here? I was hoping to finish up

- https://github.com/rustls/rustls/pull/2579

and get it released before doing this.

ctz:
Fair enough! I was doing some type name changes in rust analyzer and it broke this crate (because those names don't exist in 0.23.) Let's hold off on this.

PR #2652, opened 2025-09-15
`CryptoProvider` as a trait

ctz:
This is me playing around in the direction of #2401. Not ready for review yet, but grateful for thoughts on the end-goal?

Things I like so far:

- `ConstCryptoProvider` means these can be const constructable, include with struct-update syntax. That makes it easy to provide values for "the default, but TLS1.3-only".
- `OwnedCryptoProvider` means you can still do runtime alterations.
- `DefaultCryptoProvider` separates out the "process default" thing, and I plan to go further in this direction to make use of this explicit, and keep the errors external to the library.

Things I think suck:

- Naming
- Back-and-forth to avoid this being one big commit
- Docs are all over the place
- Arc wrapping of the provider is either 100% necessary or 100% waste depending on the underlying use. If we could have a `Box<dyn CryptoProvider + Clone>` then we could avoid that, but I don't think that is actually possible.

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -241,8 +241,8 @@ impl dyn CryptoProvider {
 /// # mod fictious_hsm_api { pub fn load_private_key(key_der: pki_types::PrivateKeyDer<'static>) -> ! { unreachable!(); } }
 /// use rustls::crypto::aws_lc_rs;
 ///
-/// pub fn provider() -> rustls::crypto::OwnedCryptoProvider {
-///   rustls::crypto::OwnedCryptoProvider {
+/// pub fn provider() -> rustls::crypto::ConstCryptoProvider {
+///   rustls::crypto::ConstCryptoProvider {
 ///     key_provider: &HsmKeyLoader,
 ///     ..aws_lc_rs::default_provider()
 ///   }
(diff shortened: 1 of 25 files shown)


djc:
> * Arc wrapping of the provider is either 100% necessary or 100% waste depending on the underlying use. If we could have a `Box<dyn CryptoProvider + Clone>` then we could avoid that, but I don't think that is actually possible.

Don't have time to look at the PR in more detail right now, but I think we can achieve `Box<dyn CryptoProvider>` fairly straightforwardly by providing our own `clone()`-like method as part of the trait.

rustls/src/crypto/mod.rs
```diff
@@ -53,6 +53,143 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
...
+    fn signature_verification_algorithms(&self) -> WebPkiSupportedAlgorithms;
+
+    /// Source of cryptographically secure random numbers.
+    fn secure_random(&self) -> &'static dyn SecureRandom;
+
+    /// Provider for loading private [`SigningKey`]s from [`PrivateKeyDer`].
+    fn key_provider(&self) -> &'static dyn KeyProvider;
+}
+
+impl dyn InternalCryptoProvider {
```
djc: Why not make these part of the trait as provided methods, instead of a separate `impl dyn`?

rustls/src/crypto/mod.rs
```diff
@@ -53,6 +53,143 @@ pub use crate::msgs::handshake::KeyExchangeAlgorithm;
 pub use crate::rand::GetRandomFailed;
 pub use crate::suites::CipherSuiteCommon;
 
+/// Controls core cryptography used by rustls.
+///
+/// Generalises over multiple concrete representations of `CryptoProvider`
+pub trait InternalCryptoProvider: Send + Sync + Debug {
```
djc: Arguably these names want to trade places I think? As in, `CryptoProvider` should be the trait whereas the type (which should probably become non-public) becomes `InternalCryptoProvider`?

rustls/src/crypto/mod.rs
```diff
@@ -295,7 +295,7 @@ impl dyn InternalCryptoProvider {
 /// [`ServerConfig::fips()`]/[`ClientConfig::fips()`] return `true`.
 #[allow(clippy::exhaustive_structs)]
 #[derive(Debug, Clone)]
-pub struct CryptoProvider {
+pub struct OwnedCryptoProvider {
```
djc, line 298: I think in 1.0 we should not make this part of the rustls' crate public API.

ctz:
I think I'm going to pursue #2675 instead of this.

djc:
> I think I'm going to pursue #2675 instead of this.

It's not obvious to me that that addresses the original motivation (as I see it, at least) for #2401, in the sense that we could not change field types or add fields with a const-constructable design.

PR #2653, opened 2025-09-15
fix dead code warns w/o std feature

cpu:
The [daily-tests.yml](https://github.com/rustls/rustls/actions/workflows/daily-tests.yml) `cargo hack` job has been sad since https://github.com/rustls/rustls/pull/2608  landed.

The only call-site for the `pub(crate)` `ConnectionCore::early_exporter()` fn is gated by the `std` feature, so it should be too.

Similarly, the mod private `Accepted` struct is only useful with `std` enabled and should be gated.

Here's a [manual dispatch](https://github.com/cpu/rustls/actions/runs/17746300794) of `daily-tests.yml` for this branch showing the fixes pass.

### DIFF
--- a/rustls/src/server/server_conn.rs
+++ b/rustls/src/server/server_conn.rs
@@ -15,18 +15,22 @@ use crate::builder::ConfigBuilder;
 use crate::common_state::{CommonState, Side};
 #[cfg(feature = "std")]
 use crate::common_state::{Protocol, State};
-use crate::conn::{ConnectionCommon, ConnectionCore, UnbufferedConnectionCommon};
+#[cfg(feature = "std")]
+use crate::conn::ConnectionCommon;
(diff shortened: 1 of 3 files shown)


rustls/src/server/server_conn.rs
```diff
@@ -1051,6 +1053,7 @@ impl UnbufferedConnectionCommon<ServerConnectionData> {
 ///
 /// Contains the state required to resume the connection through [`Accepted::into_connection()`].
 pub struct Accepted {
+    #[cfg(feature = "std")]
```
djc: Isn't `Accepted` without `connection` kinda pointless?

cpu: Fair! I'll lift it to the type itself and update.

PR #2654, opened 2025-09-16
`--deny warnings` for feature combination builds

ctz:
This would have caught #2653 synchronously in the standard PR builds.

As it is these warnings were hit but don't cause an error: https://github.com/rustls/rustls/actions/runs/17742255530/job/50418930016#step:5:16


djc (approved):
Nice!

PR #2656, opened 2025-09-16
Unify hello traits

djc:
Another chunk from

- #2645


djc:
Cleaned up the commit history quite a bit to introduce smaller changes.

PR #2658, opened 2025-09-18
Test QUIC exporter and ensure it is usable

ctz:
The existing `test_quic_exporter` didn't actually touch any QUIC types.

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -98,26 +98,6 @@ mod connection {
                 Self::Server(conn) => conn.alert(),
             }
         }
-
-        /// Returns an object that can derive key material from the agreed connection secrets.
-        ///
-        /// See [RFC5705][] for more details on what this is for.
-        ///
-        /// This function can be called at most once per connection.
-        ///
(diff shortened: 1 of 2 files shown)


rustls/src/quic.rs
```diff
@@ -98,26 +98,6 @@ mod connection {
...
-        ///   [`CommonState::is_handshaking`] first).
-        /// - if called more than once per connection.
-        ///
-        /// [RFC5705]: https://datatracker.ietf.org/doc/html/rfc5705
-        pub fn exporter(&mut self) -> Result<KeyingMaterialExporter, Error> {
-            match self {
-                Self::Client(conn) => conn.core.exporter(),
-                Self::Server(conn) => conn.core.exporter(),
-            }
-        }
```
ctz, line 120: I am generally preferring not to have functions like this, when they are trivial to provide externally and not commonly used.

djc: Maybe we should get rid of more of the wrappers over client + server or at least their methods? We also now have the potential option of abstracting over them...

PR #2664, opened 2025-09-19
Reorganise integration-level tests

ctz:
A typical build+retest cycle is a little quicker from avoiding re-linking as many test programs. Most importantly the tests are no longer in a 8300-line file.


rustls-test/src/lib.rs
```diff
@@ -1715,6 +1715,40 @@ where
     }
 }
 
+/// Check `reader` has available exactly `bytes`
```
djc, line 1748: Nit: this stuff doesn't seem to belong in this commit ("Address clippy::use_self in OtherSession")?

rustls-test/src/lib.rs
```diff
@@ -707,6 +707,34 @@ pub fn make_pair_for_arc_configs(
     )
 }
 
+/// Return a client and server config that don't share a common cipher suite
+pub fn make_disjoint_suite_configs(provider: &CryptoProvider) -> (ClientConfig, ServerConfig) {
```
djc: Nit: I think this could just take ownership of the `provider`?

(Would probably prefer that the refactor happen in a separate commit from the move.)

rustls/tests/raw_keys.rs
```diff
@@ -0,0 +1,122 @@
+#![allow(clippy::disallowed_types, clippy::duplicate_mod)]
```
djc, line 1: I think the `rpk` name is fairly idiosyncratic, maybe expand to `raw_keys`?

djc (commented):
Looks great to me!

PR #2669, opened 2025-09-23
Use generics instead of enum wrappers for connection types

djc:
Replaces `ClientConnection` with `Connection<Client>`, which seems like a pretty nice way to move forward. Makes generics quite a bit simpler: instead of `impl DerefMut<Target = ConnectionCommon<impl SideData>>` we can now say `ConnectionCommon<impl SideData>`).

Remaining steps:

- QUIC
- Unbuffered

Maybe having `Client` and `Server` types is a little weird and they should be `ClientData` and `ServerData`?

Would like early feedback before resolving all CI issues.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -21,6 +21,7 @@
 
 use core::fmt::{Debug, Formatter};
 use std::io::{self, Read, Write};
+use std::ops::DerefMut;
 use std::sync::{Arc, Mutex};
 use std::{env, net, process, thread, time};
 
@@ -34,7 +35,7 @@ use rustls::client::danger::{
 };
 use rustls::client::{
     ClientConfig, ClientConnection, EchConfig, EchGreaseConfig, EchMode, EchStatus, Resumption,
-    Tls12Resumption, WebPkiServerVerifier,
+    Tls12Resumption, WebPkiServerVerifier, WriteEarlyData,
 };
 use rustls::crypto::aws_lc_rs::hpke;
 use rustls::crypto::hpke::{Hpke, HpkePublicKey};
@@ -49,13 +50,14 @@ use rustls::server::danger::{
     ClientCertVerified, ClientCertVerifier, ClientIdentity, SignatureVerificationInput,
 };
 use rustls::server::{
-    ClientHello, ProducesTickets, ServerConfig, ServerConnection, WebPkiClientVerifier,
+    ClientHello, ProducesTickets, ReadEarlyData, ServerConfig, ServerConnection,
+    WebPkiClientVerifier,
 };
 use rustls::{
-    AlertDescription, CertificateCompressionAlgorithm, CertificateError, Connection,
+    AlertDescription, CertificateCompressionAlgorithm, CertificateError, ConnectionCommon,
     DistinguishedName, Error, HandshakeKind, InvalidMessage, NamedGroup, PeerIncompatible,
-    PeerMisbehaved, ProtocolVersion, RootCertStore, Side, SignatureAlgorithm, SignatureScheme,
-    client, compress, server, sign,
+    PeerMisbehaved, ProtocolVersion, RootCertStore, Side, SideData, SignatureAlgorithm,
+    SignatureScheme, client, compress, server, sign,
 };
 
 static BOGO_NACK: i32 = 89;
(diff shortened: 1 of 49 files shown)


ctz:
I'm not super sure about this. I had a look at this branch in terms of docs, autocomplete and llvm-lines output:

For docs, the first docs for the most important type in the whole library looks like:

image

Aside from knowing to scroll halfway down the page, and finding `impl Connection<Server>` etc, there is really no way to go from here to understanding what types for `Data` are acceptable, or where to find them, or what precisely to write to get something working (either as a value, or a type, or what constructors are available for this type.) Apparently rustdoc outputs these in the order they are encountered in, so maybe it's under our control to have `impl Connection<Client>` appear "above the fold"?

Autocomplete - rust-analyzer gleefully guides you to code that doesn't actually compile, like:

```rust
let mut conn = rustls::Connection::new(Arc::new(config), server_name).unwrap();
```

(and then leaves you to read the compiler error and follow one of the two suggestions, only one of which can ever produce code that compiles.)

For llvm-lines for the bogo runner, this grows lines from 38k to 41k, and most of that is coming from the entirely of `bogo::exec` being monomorphised an extra time.

Generally I think I have a visceral reaction to generics in rust, and I will avoid them at all costs, so take this with a pinch of salt.

djc:
> I'm not super sure about this.

Totally fair, and I'm okay with abandoning this PR if we decide it's not an improvement. I think I still like it despite your concerns, so will argue with your points a little:

> I had a look at this branch in terms of docs, autocomplete and llvm-lines output:
> 
> For docs, the first docs for the most important type in the whole library looks like:
> 
> [cut the screenshot]
> 
> Aside from knowing to scroll halfway down the page, and finding `impl Connection<Server>` etc, there is really no way to go from here to understanding what types for `Data` are acceptable, or where to find them, or what precisely to write to get something working (either as a value, or a type, or what constructors are available for this type.) Apparently rustdoc outputs these in the order they are encountered in, so maybe it's under our control to have `impl Connection<Client>` appear "above the fold"?

Well, the first problem here is that there is no meaningful doc string for `Connection` because it used to be a mostly internal type. Here's a screenshot of what it looks like when that is cleaned up a bit. Note that this renames the overly generic `Data` type argument to `Side` to make it more obvious, and adds eager bounds to make it clearer as well. (Implemented in later commits.)

Screenshot 2025-09-23 at 15 00 50

I agree that construction is less ergonomic, and that this is an important entry point, and rustdoc mostly does a decent job with representing the cascade of `Deref` implementations from `ClientConfig` -> `ConnectionCommon<ClientConnectionData>` -> `CommonState`. Apart from the constructors, arguably the generic methods for `Connection` are more relevant to most users than the side-specific ones, so I'd consider that a small win.

> Autocomplete - rust-analyzer gleefully guides you to code that doesn't actually compile, like:
> 
> ```rust
> let mut conn = rustls::Connection::new(Arc::new(config), server_name).unwrap();
> ```
> 
> (and then leaves you to read the compiler error and follow one of the two suggestions, only one of which can ever produce code that compiles.)

Right. As a potential workaround, we could move these methods to the `Config` types instead. That would have the slight advantage of enabling chaining in the case where you only need a single connection.

> For llvm-lines for the bogo runner, this grows lines from 38k to 41k, and most of that is coming from the entirely of `bogo::exec` being monomorphised an extra time.

It feels to me like abstracting over connections to the degree that bogo does it is pretty rare and probably shouldn't be a heavily weighted factor in deciding how to proceed with this. I think making it more obvious how to abstract over both types of connections has some value that might help in downstream crates (like tokio-rustls), and generally feels nicer to me in the sense of exposing our internal abstractions. I also feel like it opens up potential avenues for finding better ways to abstract over buffered/unbuffered/QUIC connections.

> Generally I think I have a visceral reaction to generics in rust, and I will avoid them at all costs, so take this with a pinch of salt.

I'd be curious if you can unpack this more. In general we're already using them a fair amount (for `Deref` tower), and I feel like that is largely better (mostly, easier to maintain) than duplicating wrappers all over the place -- or using macros, which I guess I have more of a visceral reaction to!

djc:
> I'm not super sure about this.

Do you think we should invite the Maston/Bluesky mob to give feedback?

cpu:
@djc Is this still something you want to see move forward? Should it be set as a draft?

djc:
> @djc Is this still something you want to see move forward? Should it be set as a draft?

Yes, would like to revisit this before 0.24.0 after discussion in https://github.com/rustls/rustls/pull/3149#discussion_r3613581239.

PR #2672, opened 2025-09-24
crypto: clean up SigningKey API

djc:


rustls/src/crypto/aws_lc_rs/sign.rs
```diff
@@ -15,40 +15,6 @@ use crate::enums::{SignatureAlgorithm, SignatureScheme};
 use crate::error::Error;
 use crate::sync::Arc;
 
-/// Parse `der` as any ECDSA key type, returning the first which works.
-///
-/// Both SEC1 (PEM section starting with 'BEGIN EC PRIVATE KEY') and PKCS8
-/// (PEM section starting with 'BEGIN PRIVATE KEY') encodings are supported.
-pub fn any_ecdsa_type(der: &PrivateKeyDer<'_>) -> Result<Arc<dyn SigningKey>, Error> {
```
djc, line 44: This was `pub` while the new API is private, by virtue of `EcdsaSigningKey` being private itself. I don't remember the reason we offered this as separate API, does it have to do something with linker DCE? If we want to keep this, seems reasonable to expose `EcdsaSigningKey` type directly.

rustls/src/crypto/aws_lc_rs/sign.rs
```diff
@@ -15,20 +15,6 @@ use crate::enums::{SignatureAlgorithm, SignatureScheme};
 use crate::error::Error;
 use crate::sync::Arc;
 
-/// Parse `der` as any EdDSA key type, returning the first which works.
-///
-/// Note that, at the time of writing, Ed25519 does not have wide support
-/// in browsers.  It is also not supported by the WebPKI, because the
-/// CA/Browser Forum Baseline Requirements do not support it for publicly
-/// trusted certificates.
-pub fn any_eddsa_type(der: &PrivatePkcs8KeyDer<'_>) -> Result<Arc<dyn SigningKey>, Error> {
```
djc, line 80: Same here for `Ed25519SigningKey`.

rustls/src/crypto/aws_lc_rs/sign.rs
ctz: Yeah I think it was a theoretical desire to keep a given `EcdsaSigningKey` single-curve. But then using that is a pain in the ass, so this function was plastering over that.

This is way less of a problem now due to `KeyProvider` being the main way to start using a private key.

rustls/src/crypto/aws_lc_rs/sign.rs
ctz: I think this was here only for hiding the type, and possibly (foolish) consistency.

rustls/src/crypto/aws_lc_rs/sign.rs
```diff
@@ -347,6 +333,21 @@ impl SigningKey for Ed25519SigningKey {
...
+    type Error = Error;
+
+    /// Parse `der` as any EdDSA key type, returning the first which works.
+    ///
+    /// Note that, at the time of writing, Ed25519 does not have wide support
+    /// in browsers.  It is also not supported by the WebPKI, because the
+    /// CA/Browser Forum Baseline Requirements do not support it for publicly
+    /// trusted certificates.
+    fn try_from(der: &PrivatePkcs8KeyDer<'_>) -> Result<Self, Error> {
+        // TODO: Add support for Ed448
```
ctz: I think this type name is badly named if the intention is it can support ed448. I suggest removing this comment and keeping the type name as-is (and also maybe fixing the second argument in its constructor?  But much less important given it is private.)

rustls/src/crypto/ring/sign.rs
```diff
@@ -380,6 +366,21 @@ impl SigningKey for Ed25519SigningKey {
     }
 }
 
+impl TryFrom<&PrivatePkcs8KeyDer<'_>> for Ed25519SigningKey {
+    type Error = Error;
+
+    /// Parse `der` as any EdDSA key type, returning the first which works.
```
ctz:
```suggestion
    /// Parse `der` an Ed25519 private key.
```

rustls/src/crypto/aws_lc_rs/sign.rs
djc: Cleaned this up.

ctz (approved):

PR #2673, opened 2025-09-24
Take MSRV of 1.83

ctz:
According to our MSRV policy, that means the earliest release of 0.24 could be November. That is OK I think.

Aside from the improvements in this PR, the impetus is static-in-consts support which is needed for a much better formulation of #2652.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -29,8 +29,8 @@ in the course of normal development, subject to these constraints:
 - Our MSRV will be no more recent than 9 versions old, or approximately 12 months.
 
 > [!TIP]
-> At the time of writing, the most recent Rust release is 1.88.  That means
-> our MSRV could be as recent as 1.79. As it happens, it is 1.79.
+> At the time of writing, the most recent Rust release is 1.90.  That means
+> our MSRV could be as recent as 1.81. As it happens, it is 1.83.
 
 - Our MSRV policy only covers the core library crate: it does not cover tests
   or example code, and is not binding on our dependencies.
--- a/rustls/src/polyfill.rs
+++ b/rustls/src/polyfill.rs
@@ -1,9 +1,0 @@
-/// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
-// TODO(XXX): remove once MSRV reaches 1.80
-#[allow(dead_code)] // Complicated conditional compilation guards elided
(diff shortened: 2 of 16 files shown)


djc (approved):
Nice!

djc:
I guess webpki needs an MSRV bump too?

```rust
error[E0277]: the trait bound `webpki::Error: core::error::Error` is not satisfied
   --> rustls/src/webpki/mod.rs:134:49
    |
134 |         _ => CertificateError::Other(OtherError(Arc::new(error))).into(),
    |                                                 ^^^^^^^^^^^^^^^ the trait `core::error::Error` is not implemented for `webpki::Error`
    |
    = note: required for the cast from `alloc::sync::Arc<webpki::Error>` to `alloc::sync::Arc<(dyn core::error::Error + Send + Sync + 'static)>`
```

SECURITY.md
```diff
@@ -29,8 +29,8 @@ in the course of normal development, subject to these constraints:
 - Our MSRV will be no more recent than 9 versions old, or approximately 12 months.
 
 > [!TIP]
-> At the time of writing, the most recent Rust release is 1.88.  That means
-> our MSRV could be as recent as 1.79. As it happens, it is 1.79.
+> At the time of writing, the most recent Rust release is 1.90.  That means
+> our MSRV could be as recent as 1.81. As it happens, it is 1.83 (this branch is pre-release).
```
cpu: This note about pre-release status feels like something that will be easy to forget to update later. I'm lightly feeling like it's not worth it?

rustls/src/polyfill.rs
```diff
@@ -1,9 +0,0 @@
-/// Non-panicking `let (nonce, ciphertext) = ciphertext.split_at(...)`.
-// TODO(XXX): remove once MSRV reaches 1.80
-#[allow(dead_code)] // Complicated conditional compilation guards elided
-pub(crate) fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
```
cpu, line 4: nice :tada:

ctz:
(Will bring back a1fbc8fcaec11d674ad5e4d26b61a8d6be311e60 later.)

SECURITY.md
ctz: Have removed this, but note that this paragraph will need updating before release so it makes sense :)

PR #2675, opened 2025-09-24
Allow const-constructable `CryptoProvider`

ctz:
This is an easier alternative #2652, using `Cow` 🐮 to generalise over borrowed/owned vectors of options. This was both easier to do, and the result is more flexible (because, eg, customising key exchange groups does not mandate allocating for the ciphersuites).

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -19,6 +19,7 @@
 )]
 
 use core::fmt::{Debug, Formatter};
+use std::borrow::Cow;
 use std::io::{self, Read, Write};
 use std::sync::{Arc, Mutex};
 use std::{env, net, process, thread, time};
@@ -223,6 +224,7 @@ impl Options {
         if let Some(groups) = &self.groups {
             provider
                 .kx_groups
+                .to_mut()
                 .retain(|kxg| groups.contains(&kxg.name()));
         }
 
@@ -312,9 +314,9 @@ impl SelectedProvider {
                 // this includes rustls-post-quantum, which just returns an altered
                 // version of `aws_lc_rs::default_provider()`
                 CryptoProvider {
-                    kx_groups: aws_lc_rs::ALL_KX_GROUPS.to_vec(),
-                    tls12_cipher_suites: aws_lc_rs::ALL_TLS12_CIPHER_SUITES.to_vec(),
-                    tls13_cipher_suites: aws_lc_rs::ALL_TLS13_CIPHER_SUITES.to_vec(),
+                    kx_groups: Cow::Borrowed(aws_lc_rs::ALL_KX_GROUPS),
+                    tls12_cipher_suites: Cow::Borrowed(aws_lc_rs::ALL_TLS12_CIPHER_SUITES),
+                    tls13_cipher_suites: Cow::Borrowed(aws_lc_rs::ALL_TLS13_CIPHER_SUITES),
                     ..aws_lc_rs::default_provider()
                 }
             }
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -230,8 +230,14 @@ impl Options {
 
         match (self.tls12_supported(), self.tls13_supported()) {
             (true, true) => provider,
-            (true, false) => provider.with_only_tls12(),
-            (false, true) => provider.with_only_tls13(),
+            (true, false) => CryptoProvider {
+                tls13_cipher_suites: Default::default(),
+                ..provider
+            },
+            (false, true) => CryptoProvider {
+                tls12_cipher_suites: Default::default(),
+                ..provider
+            },
             _ => panic!("nonsense version constraint"),
         }
     }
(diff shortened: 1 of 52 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -189,15 +190,15 @@ pub struct CryptoProvider {
     ///
     /// A valid `CryptoProvider` must ensure that all cipher suites are accompanied by at least
     /// one matching key exchange group in [`CryptoProvider::kx_groups`].
-    pub tls12_cipher_suites: Vec<&'static Tls12CipherSuite>,
+    pub tls12_cipher_suites: Cow<'a, [&'static Tls12CipherSuite]>,
```
djc: One downside of `Cow` is that it requires `alloc` (because it depends on `ToOwned`). Might be worth doing our own special-cased `Cow`-like enum?

ctz: I mean, we could, but `Vec` is also from `alloc` too right?

djc: Yeah, my point is that `Cow` has no option of "scaling down" to no-`alloc`, while our own enum could `cfg` out its `Owned` variant. Maybe we're too far from non-`alloc` being a reasonable use case for that to be relevant, though?

bogo/src/main.rs
```diff
@@ -314,7 +314,7 @@ impl SelectedProvider {
                 // this includes rustls-post-quantum, which just returns an altered
                 // version of `aws_lc_rs::default_provider()`
                 CryptoProvider {
-                    kx_groups: aws_lc_rs::DEFAULT_KX_GROUPS.to_vec(),
+                    kx_groups: aws_lc_rs::ALL_KX_GROUPS.to_vec(),
```
djc: Should we just land this separately?

rustls/src/crypto/mod.rs
```diff
@@ -179,7 +180,7 @@ pub use crate::suites::CipherSuiteCommon;
 /// [`ServerConfig::fips()`]/[`ClientConfig::fips()`] return `true`.
 #[allow(clippy::exhaustive_structs)]
 #[derive(Debug, Clone)]
-pub struct CryptoProvider {
+pub struct CryptoProvider<'a> {
```
djc: Should carefully weigh the value of allowing this to be non-`'static` -- seems useless in most of the common cases?

examples/src/bin/ech-client.rs
```diff
@@ -113,15 +113,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-    let mut config = rustls::ClientConfig::builder_with_provider(
-        aws_lc_rs::default_provider()
-            .with_only_tls13()
-            .into(),
-    )
-    .with_ech(ech_mode)
-    .with_root_certificates(root_store)
-    .with_no_client_auth()?;
+    // Construct a rustls client config with a TLS1.3-only provider, and ECH enabled.
+    let mut config =
```
djc, line 117: Would squash this ("Prefer to use const version providers") with the previous commit.

rustls/tests/api.rs
```diff
@@ -52,7 +52,10 @@ fn alpn_test_error(
 
     let server_config = Arc::new(server_config);
 
-    for version_provider in all_versions(&provider) {
+    for version_provider in [
```
djc: Can we define an `VERSION_PROVIDERS` `const` at some level that would avoid repeating this all over the place?

djc:
Could we go all in on this and store `&'static CryptoProvider` in our config types, instead of `Arc`-wrapping?

ctz:
> Could we go all in on this and store `&'static CryptoProvider` in our config types, instead of `Arc`-wrapping?

I think that would preclude run-time configuration (of the type done by tlsclient-mio, or rustls-openssl-compat). But we could possibly treat `CryptoProvider` as a value directly? (After all, there is already Arc-sharing via the config types itself.)

djc:
> > Could we go all in on this and store `&'static CryptoProvider` in our config types, instead of `Arc`-wrapping?
> 
> I think that would preclude run-time configuration (of the type done by tlsclient-mio, or rustls-openssl-compat).

I was thinking those cases could probably be solved by using [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) (or even `LazyLock` in simpler cases).

ctz:
> I was thinking those cases could probably be solved by using [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) (or even `LazyLock` in simpler cases).

Not sure I really understand this. Where is the 'static-lifetime storage for these? Beware that there is one `CryptoProvider` config per connection, based off parsing (eg) a configuration file, CLI arguments, or textual API parameters.

djc:
> > I was thinking those cases could probably be solved by using [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) (or even `LazyLock` in simpler cases).
> 
> Not sure I really understand this. Where is the 'static-lifetime storage for these? Beware that there is one `CryptoProvider` config per connection, based off parsing (eg) a configuration file, CLI arguments, or textual API parameters.

Yeah, you wouldn't really be able to do a different one per connection. But I think the common case would be one per process, which would be feasible.

rustls/src/crypto/mod.rs
ctz: Thoughts on something like https://github.com/rustls/pki-types/tree/ci/jbp-cow-slice ?

djc: Looks good to me.

ctz: (Personally I think this is a bit of a stretch, and we are significantly far away from no-alloc being feasible.)

djc: > (Personally I think this is a bit of a stretch, and we are significantly far away from no-alloc being feasible.)

That's also fair. I can live with just skipping this step. Just wanted to call it out.

ctz: OK, let's keep this idea in our back pocket.

PR #2682, opened 2025-09-29
Privitise `docsrs` cfg flag; fix nightly

ctz:
cfg is a global namespace, and we don't want to trigger this in dependencies.

Related to https://github.com/rust-lang/rust/pull/138907


djc (approved):
Makes sense. Should we backport all/part of this to rel-0.23?

PR #2684, opened 2025-09-30
Tidy bogo configuration

ctz:
Janitorial changes to config.json.in


djc (approved):
Nice! How did you find out which keys were unused?

ctz:
> Nice! How did you find out which keys were unused?

`./runme | python check.py` outputs this -- it's most useful when upgrading bogo (when they alter the name of a test) but I neglected to do that when last upgrading.

PR #2676, opened 2025-09-24
Limit capabilities of signing keys

djc:
Based on

- #2672 

I noticed that we were inconsistent about treating secrets (or private capabilities) properly in that we wrapped `SigningKey` values and the `CertifiedKey` type that contained them in `Arc` all over the place. Instead, wrap `SigningKey` values in `Box` and resolve eagerly to `CertifiedSigner` values which are sort of single use.

Still a bit messy but happy for early feedback!

PR #2685, opened 2025-09-30
Owned states

djc:
Following up on https://github.com/rustls/rustls/pull/2676#pullrequestreview-3268000058:

> One thing we can look at in the future is when/how `CertificatePayloadTls13::into_owned` is called, and if it is perhaps keep an `Arc<[CertificateDer]>` in there instead.

Not quite there yet but these seem like improvements.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -186,7 +186,7 @@ impl CommonState {
     pub(crate) fn process_main_protocol<Data>(
         &mut self,
         msg: Message<'_>,
-        mut state: Box<dyn State<Data>>,
+        state: Box<dyn State<Data>>,
         data: &mut Data,
         sendable_plaintext: Option<&mut ChunkVecBuffer>,
     ) -> Result<Box<dyn State<Data>>, Error> {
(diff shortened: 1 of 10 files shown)


rustls/src/common_state.rs
```diff
@@ -837,9 +834,7 @@ pub(crate) trait State<Data>: Send + Sync {
         self: Box<Self>,
         cx: &mut Context<'_, Data>,
         message: Message<'m>,
-    ) -> Result<Box<dyn State<Data> + 'm>, Error>
```
djc, line 840: This stuff is from

- https://github.com/rustls/rustls/pull/1597

I think the last bullet point from that PR ends up not making sense, because we effectly require owned states -- see the only external caller above, `CommonState::process_main_protocol()`. It requires quite a bit of boilerplate, though, so IMO it's better to rip this part out.

djc:
I'm surprised that this has substantial performance impact.

ctz:
The goal of this arrangement was that a single flight of multiple messages (as is quite common) could avoid owning the state data related to accepting the messages before the final one (because these were borrowed from the intermediate messages, and didn't need to be stored).

djc:
> The goal of this arrangement was that a single flight of multiple messages (as is quite common) could avoid owning the state data related to accepting the messages before the final one (because these were borrowed from the intermediate messages, and didn't need to be stored).

I looked into it but actually doing that from where we are is not trivial. The deframer needs mutable access to the deframer buffer, but the state will want to retain immutable references into the same buffer. This is probably solvable (splitting the deframer buffer for each deframed message), but that is a bunch of work.

djc:
Also, the certificate chains are the only borrowed data so far in the states. If we want to stick them in an `Arc` (the origin question of this PR), I don't think that's feasible.

PR #2686, opened 2025-09-30
Clarify internals

djc:
The less opinionated parts of

- #2669

I think these might still be useful even if we don't want to merge that PR?

(In particular, playing with borrowed states required making it obvious that the `Data` (`Side`) argument is `'static`, which is easier if there is a trait bound already.)

### DIFF
--- a/rustls/src/conn/kernel.rs
+++ b/rustls/src/conn/kernel.rs
@@ -81,7 +81,7 @@ pub struct KernelConnection<Side> {
     protocol: Protocol,
     suite: SupportedCipherSuite,
 
-    _data: PhantomData<Side>,
+    _side: PhantomData<Side>,
 }
 
 impl<Side> KernelConnection<Side> {
@@ -99,7 +99,7 @@ impl<Side> KernelConnection<Side> {
                 .suite
                 .ok_or(Error::HandshakeNotComplete)?,
 
-            _data: PhantomData,
+            _side: PhantomData,
         })
     }
 
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -836,7 +836,7 @@ impl<'a, Side: SideData> From<&'a mut ConnectionCommon<Side>> for Context<'a, Si
     fn from(conn: &'a mut ConnectionCommon<Side>) -> Self {
         Self {
             common: &mut conn.core.common_state,
-            data: &mut conn.core.data,
+            data: &mut conn.core.side,
             sendable_plaintext: Some(&mut conn.sendable_plaintext),
         }
     }
@@ -901,7 +901,7 @@ impl<Side: SideData> Deref for UnbufferedConnectionCommon<Side> {
 
 pub(crate) struct ConnectionCore<Side: SideData> {
     pub(crate) state: Result<Box<dyn State<Side>>, Error>,
-    pub(crate) data: Side,
+    pub(crate) side: Side,
     pub(crate) common_state: CommonState,
     pub(crate) hs_deframer: HandshakeDeframer,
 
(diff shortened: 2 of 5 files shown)


rustls/src/conn/mod.rs
```diff
@@ -889,9 +889,9 @@ impl<T> Deref for UnbufferedConnectionCommon<T> {
     }
 }
 
-pub(crate) struct ConnectionCore<Data> {
-    pub(crate) state: Result<Box<dyn State<Data>>, Error>,
-    pub(crate) data: Data,
+pub(crate) struct ConnectionCore<Side> {
+    pub(crate) state: Result<Box<dyn State<Side>>, Error>,
+    pub(crate) data: Side,
```
cpu: should this field be renamed `side`?

rustls/src/conn/mod.rs
```diff
@@ -900,8 +900,8 @@ pub(crate) struct ConnectionCore<Data> {
     seen_consecutive_empty_fragments: u8,
 }
 
-impl<Data> ConnectionCore<Data> {
-    pub(crate) fn new(state: Box<dyn State<Data>>, data: Data, common_state: CommonState) -> Self {
+impl<Side> ConnectionCore<Side> {
+    pub(crate) fn new(state: Box<dyn State<Side>>, data: Side, common_state: CommonState) -> Self {
```
cpu: same Q: re `data` param -> `side`.

rustls/src/conn/kernel.rs
```diff
@@ -81,10 +81,10 @@ pub struct KernelConnection<Data> {
     protocol: Protocol,
     suite: SupportedCipherSuite,
 
-    _data: PhantomData<Data>,
+    _data: PhantomData<Side>,
```
cpu: `_side`?

rustls/src/conn/mod.rs
```diff
@@ -443,7 +443,20 @@ impl ConnectionRandoms {
...
 
-/// Interface shared by client and server connections.
+/// TLS connection state with side-specific data (`Side`).
+///
+/// This is one of the core abstractions of the rustls API. It represents a single connection
+/// to a peer, and holds all the state associated with that connection. Note that it does
+/// not hold any IO objects: the application is responsible for reading and writing TLS records.
+/// If you want an object that does hold IO objects, see [`Stream`] and [`StreamOwned`].
+///
+/// This object is generic over the `Side` type parameter, which must be either. While much of
```
cpu: "which must be either" Is this sentence cut off? I'm not parsing it.

rustls/src/conn/mod.rs
djc: I guess it could be? I like `data` better for the field, but don't care a lot.

cpu: I feel like if `data` is better for the field, maybe it's better for the generic param too?

djc: To me the difference is about external vs internal. From the external perspective, I think `Side` is clearer (for example, interpreting `ConnectionCommon<Data>` vs `ConnectionCommon<Side>` in the docs). For a field (or method argument) the perspective is internal, and we're considering the role of the value in the context of the type.

ctz (approved):

djc:
@ctz did you have opinions on `side` vs `data` and `Side` vs `Data` (see @cpu's comments)? Need a tie-breaker!

rustls/src/conn/mod.rs
ctz: I would avoid general terms ("data") at all costs, so prefer `side` and `Side` for field and type names.

djc: Replaced these `data` fields and arguments with `side`.

rustls/src/conn/mod.rs
djc: Simplified to:

```rust
/// This object is generic over the `Side` type parameter, which must implement the marker trait
/// [`SideData`]. This is used to store side-specific data.
```

PR #2687, opened 2025-09-30
Update verifier trait and method names

djc:
These are no longer just for certificates, and the abbrevation in the type name (and method name) was a little unidiomatic.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -45,7 +45,7 @@ use rustls::pki_types::{
     CertificateDer, EchConfigListBytes, PrivateKeyDer, ServerName, SubjectPublicKeyInfoDer,
 };
 use rustls::server::danger::{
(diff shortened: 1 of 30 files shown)


cpu (approved):
What about the `ServerCertVerified` marker type returned from `verify_identity()`?

(Also maybe worth renaming the `NoCertificateVerification` impl in the `tlsclient-mio.rs` example to something like `NoServerVerification`?)

ctz:
> What about the `ServerCertVerified` marker type returned from `verify_identity()`?

This one I agree on.

> (Also maybe worth renaming the `NoCertificateVerification` impl in the `tlsclient-mio.rs` example to something like `NoServerVerification`?)

This one less so, because that verifier is requiring only certificates (via accepting the default `supported_certificate_types()`)

ctz (approved):

PR #2689, opened 2025-09-30
bogo: tighten up error classification

ctz:
We had quite a few tests which just wanted to see any kind of `Error::PeerMisbehaved(_)`, which is quite non-specific. This PR "tightens" those errors to be more specific, and prefers to use bogo's naming if that is unique.


bogo/src/main.rs
```diff
@@ -1159,15 +1159,15 @@ fn handle_err(opts: &Options, err: Error) -> ! {
...
             quit(":UNEXPECTED_EXTENSION:")
         }
-        // The TLS-ECH-Client-UnsolicitedInnerServerNameAck test is expected to fail with
-        // :UNEXPECTED_EXTENSION: when we receive an unsolicited inner hello SNI extension.
-        // We treat this the same as any unexpected enc'd ext and return :PEER_MISBEHAVIOUR:.
-        // Convert to the expected if this error occurs when we're configured w/ ECH.
         Error::PeerMisbehaved(PeerMisbehaved::UnsolicitedEncryptedExtension)
-            if opts.ech_config_list.is_some() =>
-        {
+        | Error::PeerMisbehaved(PeerMisbehaved::UnsolicitedServerHelloExtension)
```
djc: Maybe do `|` on the inside of `Error::PeerMisbehaved(..)`?

djc (approved):
Nice!

PR #2690, opened 2025-10-01
Make `aws_lc_rs::DEFAULT_KX_GROUPS` correct in fips mode

ctz:


djc (approved):
It is not obvious to me why this is more correct than what we had previously. Can you expand the PR description and/or commit message?

ctz:
> It is not obvious to me why this is more correct than what we had previously. Can you expand the PR description and/or commit message?

Added an explanation to the commit message.

PR #2692, opened 2025-10-01
Allow use of StdError in no-std contexts

### LINKED ISSUE #2691: OtherError should allow no_std by using core::error::Error
This is needed for RustCrypto provider as we assume no_std by default, we see that OtherError uses an Arc<StdError> but std::error::Error is already lifted to core::error::Error

djc:
Fixes #2691.

Arguably an oversight from

- #2673

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,7 +91,9 @@ subtle = { version = "2.5.0", default-features = false }
 time = { version = "0.3.6", default-features = false }
 tikv-jemallocator = "0.6"
 tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"] }
-webpki = { package = "rustls-webpki", version = "0.103.5", features = ["alloc"], default-features = false }
+webpki = { package = "rustls-webpki", version = "=0.104.0-alpha.1", features = [
+  "alloc",
+], default-features = false }
 webpki-roots = "1"
 x25519-dalek = "2"
 x509-parser = "0.18"
--- a/provider-example/src/hpke.rs
+++ b/provider-example/src/hpke.rs
@@ -206,14 +206,19 @@ impl HpkeOpener for HpkeRsReceiver {
     }
 }
 
+/// While rustls supports `core::error::Error`, hpke-rs's support is conditional on `std`.
 #[cfg(feature = "std")]
 fn other_err(err: impl core::error::Error + Send + Sync + 'static) -> Error {
     Error::Other(OtherError(alloc::sync::Arc::new(err)))
 }
 
+/// Since hpke-rs does not implement `core::error::Error` for `no_std`, we fall back to
+/// using a string representation of the error.
 #[cfg(not(feature = "std"))]
-fn other_err(_err: impl core::any::Any) -> Error {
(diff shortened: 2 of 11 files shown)


djc:
I guess this depends on

- https://github.com/rustls/webpki/pull/388

stevefan1999-personal:
Checkout #2694

ctz:
> Fixes #2691.
> 
> Arguably an oversight from
> 
> * [Take MSRV of 1.83 #2673](https://github.com/rustls/rustls/pull/2673)

Eh, I did this precise thing in that PR and then dropped it for the same reason that this PR doesn't build :)

Cargo.toml
```diff
@@ -91,7 +91,9 @@ subtle = { version = "2.5.0", default-features = false }
 time = { version = "0.3.6", default-features = false }
 tikv-jemallocator = "0.6"
 tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"] }
-webpki = { package = "rustls-webpki", version = "0.103.5", features = ["alloc"], default-features = false }
+webpki = { package = "rustls-webpki", version = "=0.104.0-alpha.1", features = [
```
djc, line 94: I find breaking this into multiple lines pretty ugly/hard to read, but that's what our Taplo config says.

Also, I think https://github.com/tombi-toml/tombi might be the new hotness?

provider-example/src/hpke.rs
```diff
@@ -206,14 +206,19 @@ impl HpkeOpener for HpkeRsReceiver {
     }
 }
 
+/// While rustls supports `core::error::Error`, hpke-rs's support is conditional on `std`.
 #[cfg(feature = "std")]
 fn other_err(err: impl core::error::Error + Send + Sync + 'static) -> Error {
     Error::Other(OtherError(alloc::sync::Arc::new(err)))
 }
 
+/// Since hpke-rs does not implement `core::error::Error` for `no_std`, we fall back to
```
djc, line 215: This is pretty ugly but seems like the best workaround for now?

cpu: Seems reasonable given this is just an example and not something I expect to be used in production contexts.

rustls/src/error.rs
```diff
@@ -466,22 +466,15 @@ pub enum CertificateError {
...
     /// A signature inside a certificate or on a handshake was made with an unsupported algorithm.
-    #[deprecated(
-        since = "0.23.29",
-        note = "use `UnsupportedSignatureAlgorithmContext` instead"
-    )]
-    UnsupportedSignatureAlgorithm,
-
-    /// A signature inside a certificate or on a handshake was made with an unsupported algorithm.
-    UnsupportedSignatureAlgorithmContext {
+    UnsupportedSignatureAlgorithm {
```
ctz, line 469: I guess third-party validators can just leave these empty? Seems a bit ugly though.

djc: Do you want to retain no-context variants? Or wrap this in an `Option<Context>` thing?

Happy to follow up and make this better.

PR #2694, opened 2025-10-01
refactor: replace custom error handling with OtherError for consistency

stevefan1999-personal:
Superseds #2692


stevefan1999-personal:
Seems like also blocked by https://github.com/rustls/rustls/issues/2691#issuecomment-3355638411

djc:
I think having some constructors and/or `From` impls for `OtherError` would make sense. I'd prefer to add them separately from the changes I did in #2692 and adding them one by one to evaluate whether each is worth adding.

PR #2696, opened 2025-10-02
client: pass negotiated certificate type into resolve()

djc:


bogo/src/main.rs
```diff
@@ -579,6 +579,7 @@ impl MultipleClientCredentialResolver {
 impl client::ResolvesClientCert for MultipleClientCredentialResolver {
     fn resolve(
         &self,
+        _negotiated_type: CertificateType,
         root_hint_subjects: &[&[u8]],
         sig_schemes: &[SignatureScheme],
```
ctz, line 584: Thoughts on wrapping these in a struct, ala `ClientHello` for the server-side?

ctz (approved):

bogo/src/main.rs
djc: Yeah, I think that makes sense. Thoughts on the name? `ServerHello`?

ctz: `IdentityRequest` maybe? (these fields come from a TLS message called `CertificateRequest` that does not necessarily request a certificate)

PR #2697, opened 2025-10-03
client: bundle resolve() arguments up in a type

djc:
Following up on

- https://github.com/rustls/rustls/pull/2696#discussion_r2401152220

### DIFF
--- a/rustls-test/src/lib.rs
+++ b/rustls-test/src/lib.rs
@@ -472,11 +472,9 @@ impl KeyType {
         .unwrap()
     }
 
-    pub fn ca_distinguished_name(&self) -> Vec<u8> {
+    pub fn ca_distinguished_name(&self) -> DistinguishedName {
         match self {
-            Self::Rsa2048 => {
-                b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 2048 CA" as &'static [u8]
-            }
+            Self::Rsa2048 => &b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 2048 CA"[..],
             Self::Rsa3072 => b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 3072 CA",
             Self::Rsa4096 => b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 4096 CA",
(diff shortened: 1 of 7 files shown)


rustls-test/src/lib.rs
```diff
@@ -471,16 +471,19 @@ impl KeyType {
         .unwrap()
     }
 
-    pub fn ca_distinguished_name(&self) -> &'static [u8] {
+    pub fn ca_distinguished_name(&self) -> Vec<u8> {
         match self {
-            Self::Rsa2048 => b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 2048 CA",
+            Self::Rsa2048 => {
+                b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 2048 CA" as &'static [u8]
```
ctz: is `&b"abc"[..]` more idiomatic?

rustls-test/src/lib.rs
```diff
@@ -471,19 +471,28 @@ impl KeyType {
         .unwrap()
     }
 
-    pub fn ca_distinguished_name(&self) -> Vec<u8> {
-        match self {
-            Self::Rsa2048 => {
-                b"0\x1f1\x1d0\x1b\x06\x03U\x04\x03\x0c\x14ponytown RSA 2048 CA" as &'static [u8]
+    pub fn ca_distinguished_name(&self) -> DistinguishedName {
+        DistinguishedName::from(
```
ctz: does suffix `.into()` work here, given the constraint of the return type?

rustls/src/client/client_conn.rs
```diff
@@ -139,7 +139,7 @@ impl IdentityRequest<'_> {
     /// `x509-parser`.
     ///
     /// [`DistinguishedName`]: crate::DistinguishedName
-    pub fn root_hint_subjects(&self) -> &[&[u8]] {
+    pub fn root_hint_subjects(&self) -> &[DistinguishedName] {
```
ctz, line 144: I would reiterate in the doc comment here that this type does not validate or guarantee the encoding.

ctz (approved):

rustls-test/src/lib.rs
djc: I guess it might be! At least it's more concise/gets better formatting.

rustls-test/src/lib.rs
djc: Yup, nice.

rustls/src/client/client_conn.rs
djc: Done. Feel free to nitpick my docs.

PR #2698, opened 2025-10-03
unbuffered: eliminate allocation in unbuffered bulk encrypt path

hyrumcoop:
## Summary
This PR eliminates allocations in the unbuffered API bulk encrypt path without modifying the public API. This is achieved by adding an `encrypt_into` method on the `MessageEncrypter` trait that writes the encrypted record directly into a caller-provided buffer.

To be less intrusive, I introduced the structs `PrefixedPayloadBorrowed` and `OutboundOpaqueMessageBorrowed` alongside their owned counterparts. This temporarily duplicates some logic but leaves the original encrypt path untouched.

## Benchmarks

Hardware: AMD Ryzen 7 5800XT  
Command: `BENCH_MULTIPLIER=32 taskset -c 4 ./target/release/rustls-bench --api=unbuffered bulk TLS13_AES_128_GCM_SHA256`

| Allocator | Version     | Throughput (MB/s) | Δ vs prev |
|-----------|-------------|-------------------|-----------|
| system    | prev        | 5022.06           | —         |
| system    | zero-alloc  | 5253.56           | +4.6%     |
| jemalloc  | prev        | 5024.51           | —         |
| jemalloc  | zero-alloc  | 5243.84           | +4.4%     |

Across both allocators, throughput improved by ~4-5%.

## Motivation
Beyond the modest throughput increase, removing allocations in this path:
- Reduces fragmentation caused by frequent allocation of variable-sized records, improving long-run consistency.
- Reduces allocator contention under load.
- Provides a foundation for future zero-copy optimizations.

## Follow-up
- **Zero-copy encrypt:** Once crypto providers expose an out-of-place API, we can eliminate the intermediate copy. In local experiments with aws-lc-rs, this yielded 15–35% higher throughput depending on machine, with even better multi-core scaling due to reduced DRAM traffic.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -625,14 +625,16 @@ impl CommonState {
             written += len;
         }
 
+        let mut outgoing_remaining = &mut outgoing_tls[written..];
         for m in fragments {
-            let em = self
+            let (em, new_outgoing_remaining) = self
                 .record_layer
-                .encrypt_outgoing(m)
+                .encrypt_outgoing_into(m, outgoing_remaining)
                 .encode();
 
+            outgoing_remaining = new_outgoing_remaining;
+
             let len = em.len();
-            outgoing_tls[written..written + len].copy_from_slice(&em);
             written += len;
         }
 
--- a/rustls/src/crypto/aws_lc_rs/tls12.rs
+++ b/rustls/src/crypto/aws_lc_rs/tls12.rs
@@ -13,7 +13,8 @@ use crate::enums::{CipherSuite, ProtocolVersion, SignatureScheme};
 use crate::error::Error;
 use crate::msgs::fragmenter::MAX_FRAGMENT_LEN;
 use crate::msgs::message::{
-    InboundPlainMessage, OutboundOpaqueMessage, OutboundPlainMessage, PrefixedPayload,
+    InboundPlainMessage, OutboundOpaqueMessage, OutboundOpaqueMessageBorrowed,
+    OutboundPlainMessage, PrefixedPayload, PrefixedPayloadBorrowed,
 };
 use crate::suites::{CipherSuiteCommon, ConnectionTrafficSecrets};
 use crate::tls12::Tls12CipherSuite;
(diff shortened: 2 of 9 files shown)


djc:
Looks very interesting! The `main` branch is currently targeting 0.24 so I think we'd prefer to avoid duplication in favor of tweaking existing API.

hyrumcoop:
Thanks for the feedback! That makes perfect sense.

I’ve been exploring a few ways to adjust the `MessageEncrypter` API so implementors can optionally take more control over their encryption data path without introducing unnecessary complexity or bloat to Rustls.

Once I’ve worked through the details, I’ll open a separate design discussion issue for the MessageEncrypter trait so we can agree on the right direction before committing to any specific changes.

djc:
Sounds great! FWIW, we already greatly appreciate the clear PR description and clean commit history.

hyrumcoop:
Hi @djc, thanks again for the feedback! I wanted to clarify a couple of things before continuing further:
- I'm assuming there's a requirement to avoid breaking changes to the `MessageEncrypter` trait to maintain compatibility with third-party crypto providers. Is that correct?
- You mentioned preferring API changes over duplication. Could you clarify which duplication you were referring to?
  - There is duplication in the aws-lc-rs/ring provider logic that I can easily resolve (we can concisely write `encrypt` in terms of `encrypt_into`)
  - There's also duplication in the sense that the `encrypt_into` method introduces a parallel data path (`PrefixedPayloadBorrowed`, `OutboundOpaqueMessageBorrowed`, `RecordLayer::encrypt_outgoing_into`) that mirrors the existing `encrypt` flow

djc:
> * I'm assuming there's a requirement to avoid breaking changes to the `MessageEncrypter` trait to maintain compatibility with third-party crypto providers. Is that correct?

No -- the `main` branch currently represents rustls 0.24, which will be allowed to break any and all public API as we're moving towards a rustls 1.0 release with a long-term stable API. So, unless you have requirements where you'd like to use these changes with rustls 0.23.x, we would prefer to optimize the API and avoid duplication for 0.24.x releases. (We can also publish 0.24.0-alpha releases if that helps you -- and these should be of high quality except for increased API churn -- although I'm not sure we want to do so for a whole bunch of downstream dependencies, like tokio-rustls and hyper-rustls.)

> * You mentioned preferring API changes over duplication. Could you clarify which duplication you were referring to?
>   * There is duplication in the aws-lc-rs/ring provider logic that I can easily resolve (we can concisely write `encrypt` in terms of `encrypt_into`)
>   * There's also duplication in the sense that the `encrypt_into` method introduces a parallel data path (`PrefixedPayloadBorrowed`, `OutboundOpaqueMessageBorrowed`, `RecordLayer::encrypt_outgoing_into`) that mirrors the existing `encrypt` flow

I mainly meant the latter, though in general I think we'd prefer to avoid duplication where that semantically makes sense.

PR #2701, opened 2025-10-06
Expose constructors for OtherError

djc:
These cannot be a blanket `From` impls because that conflicts with an implementation in `core`.

Replaces

- #2694 

Follow up from

- #2692

### DIFF
--- a/rustls/src/error.rs
+++ b/rustls/src/error.rs
@@ -1236,7 +1236,14 @@ mod other_error {
     /// Enums holding this type will never compare equal to each other.
     #[allow(clippy::exhaustive_structs)]
     #[derive(Debug, Clone)]
-    pub struct OtherError(pub Arc<dyn StdError + Send + Sync>);
+    pub struct OtherError(Arc<dyn StdError + Send + Sync>);
+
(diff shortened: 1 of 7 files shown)


rustls/src/error.rs
```diff
@@ -1236,7 +1236,35 @@ mod other_error {
...
-    pub struct OtherError(pub Arc<dyn StdError + Send + Sync>);
+    pub struct OtherError(Arc<dyn StdError + Send + Sync>);
+
+    impl OtherError {
+        /// Create a new `OtherError` from any error type.
+        pub fn new(err: impl StdError + Send + Sync + 'static) -> Self {
+            Self(Arc::new(err))
+        }
+
+        /// Create a new `OtherError` from any type that implements `fmt::Display`.
```
ctz: Not sure about this; I think people who actually need it can put this entirely in their program and that will be a hint that their error types should impl Error?

ctz (approved):

rustls/src/error.rs
djc: That's fair, I can move this into provider-example.

PR #2705, opened 2025-10-13
Rework hybrid key exchange arrangements

### LINKED ISSUE #2119: Reevaluate additions to traits with default implementations

ctz:
Have an enum that allows `SupportedKxGroup::start()` to return more specific trait types, and then trace that through everywhere.

`HybridKeyExchange` is an extension to `ActiveKeyExchange` that adds the hybrid key exchange parts.

This is the final TODO for, and therefore fixes #2119

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -17,7 +17,7 @@ use crate::client::common::ClientHelloDetails;
 use crate::client::ech::EchState;
 use crate::client::{ClientConfig, EchMode, EchStatus, tls13};
 use crate::common_state::{CommonState, HandshakeKind, KxState, State};
-use crate::crypto::{ActiveKeyExchange, CryptoProvider, KeyExchangeAlgorithm};
+use crate::crypto::{CryptoProvider, KeyExchangeAlgorithm, StartedKeyExchange};
 use crate::enums::{
     AlertDescription, CertificateType, CipherSuite, ContentType, HandshakeType, ProtocolVersion,
 };
@@ -58,7 +58,7 @@ pub(crate) struct ExpectServerHello {
     //
     // If this is `None` then we do not support early data.
     pub(super) early_data_key_schedule: Option<KeyScheduleEarly>,
-    pub(super) offered_key_share: Option<Box<dyn ActiveKeyExchange>>,
+    pub(super) offered_key_share: Option<StartedKeyExchange>,
     pub(super) suite: Option<SupportedCipherSuite>,
     pub(super) ech_state: Option<EchState>,
 }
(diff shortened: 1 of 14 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -476,10 +477,15 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
     /// public key. The key exchange can be completed by calling [`ActiveKeyExchange::complete()`]
     /// or discarded.
     ///
+    /// Most implementations will want to return the `StartedKeyExchange::Single(_)` variant.
+    /// Hybrid key exchange algorithms, which are constructed from two underlying algorithms,
+    /// may wish to return `StartedKeyExchange::Hybrid(_)` variant which allows the exchange
```
djc: Nit: "allows the exchange to be completed together or separately", maybe "exchanges" (plural)?

rustls/src/crypto/mod.rs
```diff
@@ -524,6 +530,39 @@ pub trait SupportedKxGroup: Send + Sync + Debug {
     }
 }
 
+/// Return value from [`SupportedKxGroup::start()`]
```
djc: Nit: IIRC prevailing style in this project is to end these with a `.`.

(Also for the variants and the inherent method.)

rustls/src/client/hs.rs
```diff
@@ -250,12 +250,16 @@ impl ExpectServerHelloOrHelloRetryRequest {
...
-            let offered_hybrid = offered_key_share
-                .hybrid_component()
-                .and_then(|(group_name, _)| {
-                    config.find_kx_group(group_name, ProtocolVersion::TLSv1_3)
-                })
-                .map(|skxg| skxg.name());
+            let offered_hybrid = match &offered_key_share {
+                StartedKeyExchange::Hybrid(h)
+                    if config
+                        .find_kx_group(h.component().0, ProtocolVersion::TLSv1_3)
```
djc: I think we could do `h.group()` here instead of `h.component().0`, which seems nicer?

Separately, there seems to be some repetition around for `Hybrid(h) if config.find_kx_group(h.group(), ProtocolVersion::TLSv1_3) -> Some(h.group())`, maybe a helper for this would be nice?

ctz: Replaced this with `HybridKeyExchange::component_separately_supported()`

rustls/src/crypto/mod.rs
```diff
@@ -689,6 +648,104 @@ pub trait ActiveKeyExchange: Send + Sync {
     fn group(&self) -> NamedGroup;
 }
 
+/// For hybrid key exchanges, allows the "classical" half to be completed separately.
```
djc: Make this first line more of a high-level description of the trait?

rustls/src/crypto/mod.rs
```diff
@@ -689,6 +648,104 @@ pub trait ActiveKeyExchange: Send + Sync {
...
+
+    /// Obtain the value as a `dyn ActiveKeyExchange`
+    fn as_key_exchange(&self) -> &(dyn ActiveKeyExchange + 'static);
+
+    /// Remove the ability to do hybrid key exchange on this object.
+    fn into_key_exchange(self: Box<Self>) -> Box<dyn ActiveKeyExchange>;
+
+    /// Returns true if the component group appears in `supported` and is usable
+    /// for `version`.
+    fn component_separately_supported(
```
djc: This still feels pretty awkward and the name doesn't seem very meaningful to me.

ctz: added a commit that hoists the whole match pattern upwards, wdyt?

PR #2712, opened 2025-10-17
crypto: store PeerIdentity in CertifiedKey/CertifiedSigner

djc:
Proliferating the use of `PeerIdentity`.

Not yet in here: renaming `CertifiedKey`/`CertifiedSigner`. Any name suggestions? Maybe `Credentials` and `SigningIdentity`? Also probably want to move `ocsp` to scope it to X509 only?

### DIFF
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -32,7 +32,7 @@ use rustls::crypto::{CryptoProvider, aws_lc_rs as provider};
 use rustls::pki_types::pem::PemObject;
 use rustls::pki_types::{CertificateDer, CertificateRevocationListDer, PrivateKeyDer};
 use rustls::server::WebPkiClientVerifier;
-use rustls::{ProtocolVersion, RootCertStore};
+use rustls::{PeerIdentity, ProtocolVersion, RootCertStore};
 
 // Token for our listening socket.
 const LISTENER: mio::Token = mio::Token(0);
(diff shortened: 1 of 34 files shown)


examples/src/bin/tlsserver-mio.rs
```diff
@@ -626,7 +626,11 @@ fn make_config(args: &Args) -> Arc<rustls::ServerConfig> {
     let (versions, provider) = args.provider();
     let mut config = rustls::ServerConfig::builder_with_provider(provider.into())
         .with_client_cert_verifier(client_auth)
-        .with_single_cert_with_ocsp(Arc::from(certs), privkey, Arc::from(ocsp))
+        .with_single_cert_with_ocsp(
+            Arc::new(PeerIdentity::from_cert_chain(certs).unwrap()),
+            privkey,
+            Arc::from(ocsp),
+        )
```
ctz, line 633: Not so sure about this change, mainly the `PeerIdentity` part when describing our own certificates. That, I would argue, is not the peer's. Maybe we want to rename `PeerIdentity` to just `Identity`?

ctz (commented):
LGTM!

examples/src/bin/tlsserver-mio.rs
djc: I was planning to review/rework names in a later PR. Want to just merge this and tweak names as follow-up? This is kinda big already.

Issue #2397, opened 2025-03-25
Evolution of the TLS connection state machine


djc:
There are two mutually exclusive ways forward being proposed:

* Small, modular core: "externalize" state machine transitions by using publicly visible enums (with states represented using mostly-opaque type with a minimal public API). Facilitates #850 et al by providing a low-level API that allows the caller to deal with all state machines that require application input, similar to the `Acceptor` approach. Combined with an unbuffered API, there is also an expectation that this might scale down better to potential low-code size environments and FFI constraints.

* Async everything: embrace the language-level async paradigm and rewrite the state machine as an `async fn`. In addition to I/O, model other application inputs (key/certificate availability, certificate verification feedback, etc) as events that wake the executor. Seems like it would require a custom executor/runtime. Would avoid boilerplate from serializing/making explicit particular state machine states.

conradludgate:
I think there's an additional benefit to exposing the low level state machine - it allows easier access to spawn_blocking for any key exchange or signing operations. 

At work we have a solution that just about works for TLS 1.3 where we use rayon for [`StartHandshake::into_stream`](https://docs.rs/tokio-rustls/latest/tokio_rustls/server/struct.StartHandshake.html#method.into_stream) which manages to encapsulate the certificate signing and key exchange and this reduced P99s considerably under high connection churn. This solution doesn't support HelloRetryRequest or TLS 1.2 though as the KX and signatures are deferred and occur during the regular await. Having access to the state machine would make it easier for us to wrap the appropriate transitions into a CPU-bound thread pool.

Issue #1976, opened 2024-05-31
poll based encrypt/decrypt/sign api to support both async and sync call


taikulawo:
**Checklist**
* [x] I've searched the issue tracker for similar requests
https://github.com/rustls/rustls/issues/850#issuecomment-1544457213
https://github.com/rustls/rustls/issues/341
https://github.com/rustls/rustls/pull/1648
https://github.com/rustls/rustls/issues/1808
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
https://docs.rs/rustls/latest/rustls/crypto/cipher/trait.MessageDecrypter.html

convert crypto api to poll based

from
```rs
pub trait MessageDecrypter: Send + Sync {
    /// Decrypt the given TLS message `msg`, using the sequence number
    /// `seq` which can be used to derive a unique [`Nonce`].
    fn decrypt<'a>(
        &mut self,
        msg: InboundOpaqueMessage<'a>,
        seq: u64,
    ) -> Result<InboundPlainMessage<'a>, Error>;
}
```
to
```rs
pub trait MessageDecrypter: Send + Sync {
    /// Decrypt the given TLS message `msg`, using the sequence number
    /// `seq` which can be used to derive a unique [`Nonce`].
    fn decrypt<'a>(
        &mut self,
        cx: &mut Context<'a>,
        msg: InboundOpaqueMessage<'a>,
        seq: u64,
    ) -> Result<InboundPlainMessage<'a>, Error>;
}
```
so that we can call `tokio#poll_read` to make encrypt/decrypt async

every encrypt/decrypt should call through trait, so we can hack it

**Additional context**
rustls only support poll method. user decide whether sync or async. If sync, just do and return `Poll::Ready`

ctz:
I think It's exceedingly unlikely we'll do this. `decrypt` (and `encrypt`) are computation-bound and relatively small computations at that, and not IO-bound. What is your use case?

djc:
Have you looked at tokio-rustls and the [unbuffered](https://docs.rs/rustls/latest/rustls/unbuffered/index.html) API?

taikulawo:
> I think It's exceedingly unlikely we'll do this. `decrypt` (and `encrypt`) are computation-bound and relatively small computations at that, and not IO-bound. What is your use case?

I want to hook asymmetric cryptography sign and encrypt/decrypt phase, send material to hardware cluster over internet then computed and send back. 

[...]

PR #2713, opened 2025-10-17
Externalise async ops

ctz:
This is (I think) a promising route toward #850.

Basically this extracts all calls to `ServerVerifier::verify_identity` from the state machine, and requires they happen elsewhere. For buffered connections, this just happens synchronously in `process_new_packets`. For unbuffered connections, there is a new state where the caller can take the input to this operation, and then feed in the result later (there is also a route to just doing the normal verification, which follows the configured behaviour.)

I think this general shape seems OK, and can be applied to other operations, and doesn't do much violence to the state machine. So far it is only done for server certificate verification by clients.

### DIFF
--- a/rustls-test/src/lib.rs
+++ b/rustls-test/src/lib.rs
@@ -798,6 +798,7 @@ pub fn do_unbuffered_handshake(
                         .unwrap();
                 }
                 ConnectionState::TransmitTlsData(data) => data.done(),
+                ConnectionState::VerifyServerIdentity(vsi) => vsi.complete(),
                 st => unreachable!("unexpected connection state: {st:?}"),
             }
 
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -21,6 +21,7 @@ use crate::msgs::message::{InboundPlainMessage, Message, MessagePayload};
 use crate::record_layer::Decrypted;
 use crate::suites::ExtractedSecrets;
 use crate::vecbuf::ChunkVecBuffer;
+use crate::verify::PeerVerified;
 
 // pub so that it can be re-exported from the crate root
 pub mod kernel;
@@ -873,6 +874,7 @@ pub struct UnbufferedConnectionCommon<Side: SideData> {
     pub(crate) core: ConnectionCore<Side>,
     wants_write: bool,
     emitted_peer_closed_state: bool,
+    verification_result: Option<Result<PeerVerified, Error>>,
 }
 
 impl<Side: SideData> From<ConnectionCore<Side>> for UnbufferedConnectionCommon<Side> {
@@ -881,6 +883,7 @@ impl<Side: SideData> From<ConnectionCore<Side>> for UnbufferedConnectionCommon<S
             core,
             wants_write: false,
             emitted_peer_closed_state: false,
+            verification_result: None,
         }
     }
 }
--- a/rustls/src/conn/unbuffered.rs
+++ b/rustls/src/conn/unbuffered.rs
@@ -13,6 +13,7 @@ use crate::conn::SideData;
 use crate::error::Error;
 use crate::msgs::deframer::buffers::DeframerSliceBuffer;
 use crate::server::ServerConnectionData;
+use crate::verify;
 
 impl UnbufferedConnectionCommon<ClientConnectionData> {
     /// Processes the TLS records in `incoming_tls` buffer until a new [`UnbufferedStatus`] is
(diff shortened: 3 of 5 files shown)


djc:
I don't love this. It feels like a kludge to retrofit low-level needs into a high-level API. I think we should make a proper low-level API and rebuild the high-level API on top of this.

- We recently discussed on Discord that we might drop the `Read`/`Write`-based `read_tls()` and `write_tls()` API. For `read_tls()`, I'm thinking that we could expose something like the current deframer buffer trait externally and implement it for just `Vec<u8>` so that `read_tls()` just takes an `impl TlsInput` which is a `Vec<u8>` in most cases. This moves the buffer out of the connection.
- We've discussed a few times if we could turn the `dyn State` thing into an explicit (mostly private?), non-exhaustive enum of states. If we do that, we could make these explicit states and allow callers to potentially switch between a lower-level and higher-level register much more easily.

ctz:
> I don't love this. It feels like a kludge to retrofit low-level needs into a high-level API. I think we should make a proper low-level API and rebuild the high-level API on top of this.

I guess, reading between the lines, that you are saying the unbuffered API is high-level? Or that the `dyn State` pattern is high level?

> * We recently discussed on Discord that we might drop the `Read`/`Write`-based `read_tls()` and `write_tls()` API. For `read_tls()`, I'm thinking that we could expose something like the current deframer buffer trait externally and implement it for just `Vec<u8>` so that `read_tls()` just takes an `impl TlsInput` which is a `Vec<u8>` in most cases. This moves the buffer out of the connection.
> * We've discussed a few times if we could turn the `dyn State` thing into an explicit (mostly private?), non-exhaustive enum of states. If we do that, we could make these explicit states and allow callers to potentially switch between a lower-level and higher-level register much more easily.

I'm not sure either of these points are altered by this PR; the second one particularly seems to be made easier by representing more steps in the state machine?

(I do still have a concern that the second pattern makes it hard for people to write forwards-compatible uses of rustls, which is demonstrated by this PR -- introducing new states which are impossible for a existing code to deal with.)

djc:
> > I don't love this. It feels like a kludge to retrofit low-level needs into a high-level API. I think we should make a proper low-level API and rebuild the high-level API on top of this.
> 
> I guess, reading between the lines, that you are saying the unbuffered API is high-level? Or that the `dyn State` pattern is high level?

I think the unbuffered API is clearly low-level. I think the `dyn State` pattern is higher-level than it needs to be.

> > * We recently discussed on Discord that we might drop the `Read`/`Write`-based `read_tls()` and `write_tls()` API. For `read_tls()`, I'm thinking that we could expose something like the current deframer buffer trait externally and implement it for just `Vec<u8>` so that `read_tls()` just takes an `impl TlsInput` which is a `Vec<u8>` in most cases. This moves the buffer out of the connection.
> > * We've discussed a few times if we could turn the `dyn State` thing into an explicit (mostly private?), non-exhaustive enum of states. If we do that, we could make these explicit states and allow callers to potentially switch between a lower-level and higher-level register much more easily.
> 
> I'm not sure either of these points are altered by this PR; the second one particularly seems to be made easier by representing more steps in the state machine?

Sure -- I like the part where we have more steps in the state machine. I don't like the inversion of control of injecting requirement thingies.

> (I do still have a concern that the second pattern makes it hard for people to write forwards-compatible uses of rustls, which is demonstrated by this PR -- introducing new states which are impossible for a existing code to deal with.)

That's a fair concern.

taikulawo:
Hi, I'm watching `async supports  on rustls` a very long time since I open the issue. https://github.com/rustls/rustls/issues/1976.

I don't think `add a new state` or `put some state out of state machine` is a silver bullet to solve all async based operation.

unbuffered API is hard to write correctly If user don't understand the underlying rustls internal API completely.

I agree https://github.com/rustls/rustls/issues/2397#:~:text=Async%20everything%3A%20embrace%20the%20language%2Dlevel%20async%20paradigm%20and%20rewrite%20the%20state%20machine%20as%20an%20async%20fn

We should async everything, state machine, public api, and trait(for example [Signer](https://docs.rs/rustls/latest/rustls/sign/trait.Signer.html#tymethod.sign)) , and then release a major version

ctz:
> We should async everything, state machine, public api, and trait(for example [Signer](https://docs.rs/rustls/latest/rustls/sign/trait.Signer.html#tymethod.sign)) , and then release a major version

Thanks for the comment! Unfortunately async dyn traits aren't supported in standard rust, so to do this we'd have to take a dependency, such as https://github.com/spastorino/dynosaur. But dynosaur is not stable and would become part of the public API of this crate, which we hope to stabilise in a 1.0 release.

ctz:
> * We've discussed a few times if we could turn the `dyn State` thing into an explicit (mostly private?), non-exhaustive enum of states. If we do that, we could make these explicit states and allow callers to potentially switch between a lower-level and higher-level register much more easily.

Thinking about this point more, I think a per-side, per-version enum of states can work. I don't think anything larger than that can, without client-only applications including server code, or TLS1.3-only clients including TLS1.2 code. So maybe the end result would look like:

- a top-level, generalised client or server handshake enum. Like this PR, this would expose the non-message dependencies rather than calling a synchronous trait. Under which would sit:
- a trait-based, late-bound version- and side-specific handshake, which is constructed from:
- a concrete enum tracing all states including inputs and outputs. I think this could be made "functionally pure" with the benefit of making it testable (currently the individual handshake states are annoying to test in isolation, because they have complex dependency on the contents of `ClientContext`

PR #2715, opened 2025-10-20
Fix unconditionally used underscore bindings

ctz:
The need for these fell away after the demise of the `quic` and `tls12` crate features.

### DIFF
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -115,33 +115,33 @@ mod cache {
 
         fn set_tls12_session(
             &self,
-            _server_name: ServerName<'static>,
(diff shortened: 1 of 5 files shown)


ctz:
> There should be a clippy lint for these!

There is -- `clippy::used_underscore_binding` -- but it also (as you would expect) complains about the condition ones (for logging, mostly). At some point I will look at our lints and move stuff to `expect`, which should tidy up quite a lot of this stuff.

djc:
> At some point I will look at our lints and move stuff to `expect`, which should tidy up quite a lot of this stuff.

Maybe it deserves an entry in the tracking issue?

PR #2716, opened 2025-10-20
Improve public API around identities

djc:
Following up from

- #2712

Also tries to limit the public API exposed in the crate root, to make the documentation easier to consume.

Not sure I've cleaned up all the documentation yet, but I think this clarifies things? Happy to discuss names etc.

### DIFF
--- a/rustls/src/crypto/signer.rs
+++ b/rustls/src/crypto/signer.rs
@@ -17,9 +17,11 @@ use crate::{SignerPublicKey, x509};
 
 /// Server certificate resolver which always resolves to the same identity and key.
 ///
-/// For use with [`ConfigBuilder::with_server_credential_resolver()`].
+/// For use with [`ConfigBuilder::with_server_credential_resolver()`] or
+/// [`ConfigBuilder::with_client_credential_resolver()`].
 ///
 /// [`ConfigBuilder::with_server_credential_resolver()`]: crate::ConfigBuilder::with_server_credential_resolver
+/// [`ConfigBuilder::with_client_credential_resolver()`]: crate::ConfigBuilder::with_client_credential_resolver
 #[derive(Debug)]
 pub struct SingleCredential(Credentials);
 


rustls/src/lib.rs
```diff
@@ -521,11 +521,7 @@ pub use crate::common_state::{CommonState, HandshakeKind, IoState, Side};
 #[cfg(feature = "std")]
 pub use crate::conn::{Connection, Reader, Writer};
 pub use crate::conn::{ConnectionCommon, KeyingMaterialExporter, SideData, kernel};
-pub use crate::error::{
-    ApiMisuse, CertRevocationListError, CertificateError, EncryptedClientHelloError, Error,
-    ExtendedKeyPurpose, InconsistentKeys, InvalidMessage, OtherError, PeerIncompatible,
-    PeerMisbehaved, RejectedEch,
-};
+pub use crate::error::Error;
```
djc, line 524: This leaves `Error` accessible as both `error::Error` and `Error`. It doesn't seem like a big deal to me, but alternatively we could not re-export `Error` at the top level?

bogo/src/main.rs
```diff
@@ -550,7 +550,7 @@ impl SigningKey for FixedSignatureSchemeSigningKey {
 
 #[derive(Debug)]
 struct FixedSignatureSchemeServerCertResolver {
-    cert_key: CertifiedKey,
+    cert_key: Credentials,
```
ctz: Maybe fix the field/parameter names at the same time? (ditto elsewhere in this commit)

rustls/src/crypto/signer.rs
```diff
@@ -114,7 +114,7 @@ impl CertifiedKey {
         )
     }
 
-    /// Make a new CertifiedKey, with the given identity and key.
+    /// Make a new Credentials, with the given identity and key.
```
ctz:
```suggestion
    /// Make a new `Credentials`, with the given identity and key.
```

rustls/src/crypto/signer.rs
```diff
@@ -16,21 +16,21 @@ use crate::server::{ClientHello, ParsedCertificate, ServerCredentialResolver};
 use crate::sync::Arc;
 use crate::{AlertDescription, InvalidMessage, SignerPublicKey, x509};
 
-/// Server certificate resolver which always resolves to the same certificate and key.
+/// Server certificate resolver which always resolves to the same identity and key.
```
ctz, line 18: Preexisting: docs need updating as this now impls `ClientCredentialResolver`

rustls/src/crypto/signer.rs
```diff
@@ -61,7 +61,7 @@ impl ServerCredentialResolver for SingleCredential {
 /// as the return value of their `resolve()` methods.
 #[non_exhaustive]
 #[derive(Debug)]
-pub struct CertifiedSigner {
+pub struct SelectedCredential {
```
ctz: Preexisting: consistency of vertical whitespace and field ordering compared to `Credentials`

rustls/src/lib.rs
ctz: I think I'd prefer the latter.

djc: Are you sure? We could also find a way to hide it from the `error` module. It seems like a significant ergonomic papercut if everyone who wants to reference `Error` in downstream projects has to import from `::error` explicitly.

(Even if we decide this is the way forward, I'd like to tackle it in a subsequent PR.)

bogo/src/main.rs
djc: Done.

rustls/src/crypto/signer.rs
djc: Fixed.

rustls/src/crypto/signer.rs
djc: Added a commit for this.

rustls/src/crypto/signer.rs
djc: Added a commit for this.

rustls/src/lib.rs
ctz: Not sure, and it's a pretty weak preference.

My general feeling is that most people don't actually need to name the Error type (they either print/log an error, or chuck it in `dyn std::error::Error`, or wrap it their own power-set error type, or maybe just panic). I think it is less common to need to name the error type (maybe they are handling error variants individually, or producing errors of type `rustls::Error`).

djc: https://github.com/search?q=%22use+rustls%3A%3AError%3B%22+language%3ARust&type=code&ref=advsearch

And that's just the literal one-item import -- I don't think GitHub Code Search can do wildcards so we could search for `use rustls::{*Error*};`...

PR #2721, opened 2025-10-23
Add `HandshakeKind::ResumedWithHelloRetryRequest`

ctz:
The impetus for this is a later PR, where this clarifies a test.

But, more generally, I kind of expect it would be useful for downstream users to hook up `handshake_kind()` to monitoring/metrics, and this change improves the resolution of that.

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -312,7 +312,10 @@ mod client_hello {
                     .handshake_kind
                     .get_or_insert(HandshakeKind::Full);
             } else {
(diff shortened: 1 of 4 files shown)


rustls/src/server/tls13.rs
```diff
@@ -312,7 +312,10 @@ mod client_hello {
                     .handshake_kind
                     .get_or_insert(HandshakeKind::Full);
             } else {
-                cx.common.handshake_kind = Some(HandshakeKind::Resumed);
+                cx.common.handshake_kind = match st.done_retry {
```
djc, line 315: Nit: hoist `Some`?

PR #2722, opened 2025-10-23
Separate resumption across security domains (redux)

ctz:
This is a different approach to solving #2284 and effectively reverts #2361. It is actually described in that bug (the paragraph starting "The best behaviour ...") but that wasn't possible at the time.

This is a draft -- looking for any thoughts on the overall approach rather than any details at this point. Relies on some (non-breaking) changes in pki-types and webpki too.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -8,12 +8,12 @@ use core::ops::Deref;
 
 use pki_types::ServerName;
 
-use super::config::{ClientConfig, ClientCredentialResolver, ClientSessionKey, Tls12Resumption};
-use super::connection::ClientConnectionData;
-use super::ech::{EchMode, EchState, EchStatus};
-use super::{ClientHelloDetails, tls13};
-use crate::bs_debug;
+use super::Tls12Resumption;
 use crate::check::inappropriate_handshake_message;
+use crate::client::ech::EchState;
+use crate::client::{
+    ClientConnectionData, ClientHelloDetails, ClientSessionKey, EchMode, EchStatus, tls13,
+};
 use crate::common_state::{CommonState, HandshakeKind, KxState, State};
 use crate::crypto::cipher::Payload;
 use crate::crypto::kx::{KeyExchangeAlgorithm, StartedKeyExchange};
@@ -38,7 +38,7 @@ use crate::sync::Arc;
 use crate::tls12::Tls12CipherSuite;
 use crate::tls13::Tls13CipherSuite;
 use crate::tls13::key_schedule::KeyScheduleEarly;
-use crate::verify::ServerVerifier;
+use crate::{ClientConfig, bs_debug};
 
 pub(super) type NextState = Box<dyn State<ClientConnectionData>>;
 pub(super) type NextStateOrError = Result<NextState, Error>;
@@ -1016,7 +1016,6 @@ impl ClientSessionValue {
                     .tls12_session(key)
                     .map(ClientSessionValue::Tls12)
             })
-            .and_then(|resuming| resuming.compatible_config(config.verifier(), config.resolver()))
             .and_then(|resuming| {
                 let now = config
                     .current_time()
@@ -1056,21 +1055,6 @@ impl ClientSessionValue {
         match self {
             Self::Tls13(v) => Some(v),
             Self::Tls12(_) => None,
-        }
-    }
-
-    fn compatible_config(
-        self,
-        server_cert_verifier: &Arc<dyn ServerVerifier>,
-        client_creds: &Arc<dyn ClientCredentialResolver>,
-    ) -> Option<Self> {
-        match &self {
-            Self::Tls13(v) => v
-                .compatible_config(server_cert_verifier, client_creds)
-                .then_some(self),
-            Self::Tls12(v) => v
-                .compatible_config(server_cert_verifier, client_creds)
-                .then_some(self),
         }
     }
 }
--- a/rustls/src/lib.rs
+++ b/rustls/src/lib.rs
@@ -371,8 +371,6 @@ mod test_macros;
 mod sync {
     #[expect(clippy::disallowed_types)]
     pub(crate) type Arc<T> = alloc::sync::Arc<T>;
-    #[expect(clippy::disallowed_types)]
-    pub(crate) type Weak<T> = alloc::sync::Weak<T>;
 }
 
 #[expect(unnameable_types)]
--- a/rustls-test/tests/api/kx.rs
+++ b/rustls-test/tests/api/kx.rs
@@ -312,6 +312,7 @@ fn test_client_sends_share_for_less_preferred_group() {
     ));
 
     // second handshake; HRR'd from secp384r1 to X25519
+    // (but resuming is possible, since the session storage is shared)
     let (mut client_2, mut server) = make_pair_for_configs(client_config_2, server_config);
     do_handshake(&mut client_2, &mut server);
     assert_eq!(
@@ -322,7 +323,7 @@ fn test_client_sends_share_for_less_preferred_group() {
     );
     assert_eq!(
         client_2.handshake_kind(),
-        Some(HandshakeKind::FullWithHelloRetryRequest)
+        Some(HandshakeKind::ResumedWithHelloRetryRequest)
     );
 }
 
--- a/rustls-test/tests/api/resume.rs
+++ b/rustls-test/tests/api/resume.rs
@@ -17,6 +17,7 @@ use rustls_test::{
     ClientConfigExt, ClientStorage, ClientStorageOp, ErrorFromPeer, KeyType, ServerConfigExt,
     do_handshake, do_handshake_until_error, make_client_config, make_client_config_with_auth,
     make_pair, make_pair_for_arc_configs, make_pair_for_configs, make_server_config, transfer,
+    webpki_server_verifier_builder,
 };
 
 use super::{ALL_VERSIONS, provider};
@@ -65,7 +66,12 @@ fn client_only_attempts_resumption_with_compatible_security() {
         // disallowed case: unmatching `verifier`
         let mut client_config = ClientConfig::builder(Arc::new(version_provider.clone()))
             .dangerous()
-            .with_custom_certificate_verifier(client_config.verifier().clone())
+            .with_custom_certificate_verifier(
+                webpki_server_verifier_builder(kt.client_root_store(), &version_provider)
+                    .allow_unknown_revocation_status()
+                    .build()
+                    .unwrap(),
+            )
             .with_client_credential_resolver(client_config.resolver().clone())
             .unwrap();
         client_config.resumption = base_client_config.resumption.clone();
--- a/rustls/.clippy.toml
+++ b/rustls/.clippy.toml
@@ -2,5 +2,4 @@ upper-case-acronyms-aggressive = true
 
 disallowed-types = [
   { path = "std::sync::Arc", reason = "must use Arc from sync module to support downstream forks targeting architectures without atomic ptrs" },
-  { path = "std::sync::Weak", reason = "must use Weak from sync module to support downstream forks targeting architectures without atomic ptrs" },
 ]
--- a/rustls/src/client/handy.rs
+++ b/rustls/src/client/handy.rs
@@ -210,7 +210,6 @@ impl client::ClientCredentialResolver for FailResolveClientCert {
 
 #[cfg(test)]
 mod tests {
-    use core::hash::Hasher;
     use core::time::Duration;
     use std::prelude::v1::*;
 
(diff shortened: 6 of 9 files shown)


ctz:

Yikes! Working on this.

rustls/src/client/hs.rs
```diff
@@ -441,7 +441,11 @@ impl ClientHelloInput {
         cx: &mut ClientContext<'_>,
         config: Arc<ClientConfig>,
     ) -> Result<Self, Error> {
-        let mut resuming = ClientSessionValue::retrieve(&server_name, &config, cx);
+        let session_key = ClientSessionKey {
+            partition: [0u8; 32], // TODO
```
djc: Nit: no need for the `u8` suffix?

rustls/src/crypto/signer.rs
```diff
@@ -186,7 +194,7 @@ pub struct SelectedCredential {
 
 /// A peer's identity, depending on the negotiated certificate type.
 #[non_exhaustive]
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
```
djc, line 202: Have been thinking: should we push `Identity` (et al) down into pki-types? I know there are some downstream crates that expose an identity type for client authentication.

rustls/src/client/client_conn.rs
```diff
@@ -116,7 +118,7 @@ impl ClientSessionKey<'_> {
 
 /// A trait for the ability to choose a certificate chain and
 /// private key for the purposes of client authentication.
-pub trait ClientCredentialResolver: fmt::Debug + Send + Sync {
+pub trait ClientCredentialResolver: fmt::Debug + Any + Send + Sync {
```
djc: Why is this necessary here?

bogo/src/main.rs
```diff
@@ -514,6 +514,10 @@ impl ServerVerifier for DummyServerAuth {
     fn request_ocsp_response(&self) -> bool {
         true
     }
+
+    fn hash_config(&self, h: &mut dyn core::hash::Hasher) {
```
djc: Shall we just import `Hasher` everywhere? This commit is a little inconsistent about. Not sure if you were just following the lowest resistance route from the IDE or if you're worried about a potential conceptual collision with our own `Hash` traits, but given that we don't have a `Hasher` trait, I feel like we shouldn't worry about it?

rustls/src/lib.rs
```diff
@@ -672,3 +672,18 @@ mod sealed {
     #[expect(unnameable_types)]
     pub trait Sealed {}
 }
+
+mod core_hash_polyfill {
+    /// Working around the garbage design of `core::hash::Hash`
```
djc: I like how it's colorful but would also appreciate a little more context on why its garbagity requires the gymnastics here.

rustls/src/crypto/signer.rs
```diff
@@ -58,6 +58,10 @@ impl ClientCredentialResolver for SingleCredential {
     fn supported_certificate_types(&self) -> &'static [CertificateType] {
         self.types
     }
+
+    fn hash_config(&self, h: &mut dyn Hasher) {
+        self.hash(&mut crate::core_hash_polyfill::DynHasher(h));
```
djc: Let's import this?

rustls/src/client/client_conn.rs
```diff
@@ -413,6 +416,49 @@ impl ClientConfig {
             .current_time()
             .ok_or(Error::FailedToGetCurrentTime)
     }
+
+    /// A token which partitions this config's use of the [`Self::resumption`] store.
+    pub(super) fn partition(&self) -> [u8; 32] {
```
djc: I don't think `partition` is a great name for this -- I guess I don't love it for the `ClientSessionKey` field, either... It seems pretty generic and more usually used as a verb?

rustls/src/client/client_conn.rs
```diff
@@ -413,6 +416,49 @@ impl ClientConfig {
...
+            .try_into()
+            .unwrap()
+    }
+}
+
+struct HashAdapter<'a>(&'a mut dyn hash::Context);
+
+impl Hasher for HashAdapter<'_> {
+    fn finish(&self) -> u64 {
+        unimplemented!()
```
djc: Should we warn trait implementers not to call this?

rustls/src/client/hs.rs
```diff
@@ -442,7 +442,7 @@ impl ClientHelloInput {
         config: Arc<ClientConfig>,
     ) -> Result<Self, Error> {
         let session_key = ClientSessionKey {
-            partition: [0u8; 32], // TODO
+            partition: config.partition(),
```
djc: Hmm, I'd probably squash the earlier commit into this one.

rustls/src/client/config.rs
```diff
@@ -391,6 +391,22 @@ impl ClientConfig {
...
+    /// This is the object that determines which credentials to use for client
+    /// authentication.
+    pub fn resolver(&self) -> &Arc<dyn ClientCredentialResolver> {
+        &self.client_auth_cert_resolver
+    }
+
+    /// Alter the resolver for this client configuration.
+    ///
+    /// This is the object that determines which credentials to use for client
+    /// authentication.
```
djc, line 247: Hmm, do you have a known use case for this? We could just skip it and add it later if someone complains? Otherwise, should this document a bunch of caveats about the restrictions/partitioning?

rustls/src/client/client_conn.rs
```diff
@@ -433,6 +433,10 @@ impl ClientConfig {
             .ok_or(Error::FailedToGetCurrentTime)
     }
 
+    pub(super) fn verifier(&self) -> &Arc<dyn verify::ServerVerifier> {
```
djc: Nit: suggest moving boring getters to the bottom of the `impl`.

rustls/src/client/client_conn.rs
```diff
@@ -291,10 +291,10 @@ pub struct ClientConfig {
     pub require_ems: bool,
 
     /// Provides the current system time
-    pub time_provider: Arc<dyn TimeProvider>,
+    pub(super) time_provider: Arc<dyn TimeProvider>,
```
djc: Is this because these are included in the key? Maybe make that explicit in the commit message?

rustls/src/client/client_conn.rs
```diff
@@ -291,10 +291,10 @@ pub struct ClientConfig {
...
 
     /// Source of randomness and other crypto.
-    pub(crate) provider: Arc<CryptoProvider>,
+    pub(super) provider: Arc<CryptoProvider>,
```
djc: Should we rename the getter to `provider()` for conciseness?

djc (commented):
This is looking pretty nice!

rustls/src/lib.rs
```diff
@@ -673,3 +673,20 @@ const TEST_PROVIDERS: &[&crypto::CryptoProvider] = &[
     #[cfg(feature = "ring")]
     &crypto::ring::DEFAULT_PROVIDER,
 ];
+
+mod core_hash_polyfill {
```
djc, line 561: Nit: maybe avoid the need to leak this module name into all the imports by re-"exporting" (`pub(crate)`) `DynHasher` from the crate root?

rustls/src/client/client_conn.rs
```diff
@@ -412,6 +416,49 @@ impl ClientConfig {
...
+            .iter_cipher_suites()
+            .map(|cs| cs.hash_provider())
+            .find(|h| h.output_len() >= 32)
+            .expect("no suitable cipher suite available (with |H| >= 32)"); // this is -- in practice -- all cipher suites
+
+        let mut h = hash.start();
+        let mut adapter = HashAdapter(h.as_mut());
+
+        // Include TypeId of impl, so two different types with different non-configured
+        // behaviour do not collide even if their `hash_config()`s are the same.
```
djc: Nit: "behavior".

rustls/src/client/client_conn.rs
djc: Maybe `config_hash`?

rustls/src/client/config.rs
```diff
@@ -481,6 +480,38 @@ impl Hasher for HashAdapter<'_> {
...
+
+    /// How to verify the server certificate chain.
+    verifier: Arc<dyn verify::ServerVerifier>,
+
+    /// How to decide what client auth certificate/keys to use.
+    client_auth_cert_resolver: Arc<dyn ClientCredentialResolver>,
+}
+
+impl SecurityDomain {
+    pub(crate) fn new(
```
djc, line 477: What is the value in this? Do you see risks with having all the fields `pub(crate)`?

rustls/src/client/config.rs
```diff
@@ -503,13 +473,75 @@ impl SecurityDomain {
         verifier: Arc<dyn verify::ServerVerifier + 'static>,
         time_provider: Arc<dyn TimeProvider + 'static>,
     ) -> Self {
+        // Use a hash function that outputs at least 32 bytes.
```
djc, line 483: Maybe change the commit order to introduce the `SecurityDomain` before the `partition` stuff so stuff doesn't have to move around? Also means you could fill in the partition values as its introduced.

djc (commented):
Some more suggestions.

rustls/src/client/client_conn.rs
ctz: I actually didn't realise this wasn't needed. Turns out every type impls Any

rustls/src/client/config.rs
ctz: The goal is to ensure that the `partition` field (pending name change) changes when any of the fields it depends on changes. This type just brings together all the dependencies on that calculation.

djc: Maybe add a comment to that effect?

rustls/src/client/config.rs
ctz: Removed this, and refactored the tests that were using it.

rustls/src/client/hs.rs
ctz: I think I would like to keep the `ClientSessionKey` change separate from the "hashing config" one. But I will reorder them (hashing config comes first, and then `ClientSessionKey` immediately takes advantage of it)

rustls/src/client/client_conn.rs
ctz: Ugh. Good point. This is 100% reachable from `hash_config` impls. I will implement it.

rustls-test/src/lib.rs
```diff
@@ -624,29 +624,35 @@ pub fn make_client_config_with_raw_key_support(
 pub trait ClientConfigExt {
     fn finish(self, kt: KeyType) -> ClientConfig;
     fn finish_with_creds(self, kt: KeyType) -> ClientConfig;
+    fn add_root_certs(
+        self,
+        kt: KeyType,
+    ) -> rustls::ConfigBuilder<ClientConfig, rustls::client::WantsClientCert>;
```
djc: Nit: import types?

rustls-test/tests/api/compress.rs
```diff
@@ -240,9 +240,10 @@ fn test_cert_decompression_by_server_fails() {
 #[cfg(feature = "zlib")]
 #[test]
 fn test_cert_decompression_by_server_would_result_in_excessively_large_cert() {
+    use rustls_test::ClientConfigExt;
```
djc: Nit: import at the top level?

rustls-test/tests/api/compress.rs
```diff
@@ -253,7 +254,10 @@ fn test_cert_decompression_by_server_would_result_in_excessively_large_cert() {
         Arc::new(Identity::from_cert_chain(vec![big_cert]).unwrap()),
         key,
     );
-    client_config.client_auth_cert_resolver = Arc::new(SingleCredential::from(big_cert_and_key));
+    let client_config = rustls::ClientConfig::builder(Arc::new(provider))
```
djc: Nit: import type?

rustls/src/client/client_conn.rs
```diff
@@ -360,6 +360,22 @@ impl ClientConfig {
         &self.provider
     }
 
+    /// Return the resolver for this client configuration.
```
djc: Nit: move these below `dangerous()` (which takes `&mut self`)?

rustls/src/client/client_conn.rs
```diff
@@ -356,7 +356,7 @@ impl ClientConfig {
     }
 
     /// Return the crypto provider used to construct this client configuration.
-    pub fn crypto_provider(&self) -> &Arc<CryptoProvider> {
+    pub fn provider(&self) -> &Arc<CryptoProvider> {
```
djc: Nit: IMO the rename should be in a different commit.

rustls/src/client/config.rs
```diff
@@ -206,8 +207,11 @@ pub struct ClientConfig {
     ///
     /// However, resumption is only allowed between two `ClientConfig`s if their
     /// `client_auth_cert_resolver` (ie, potential client authentication credentials)
-    /// and `verifier` (ie, server certificate verification settings) are
-    /// the same (according to `Arc::ptr_eq`).
+    /// and `verifier` (ie, server certificate verification settings):
+    ///
+    /// - are the same type (determined by hashing their `TypeId`), and
+    /// - input the same data into [`verify::ServerVerifier::hash_config()`] and
```
djc: Nit: drop the `verify::` prefix from the visible docs?

rustls/src/client/client_conn.rs
ctz: Done in a separate commit so dangerous appears before pub functions with an immutable receiver.

rustls/src/crypto/signer.rs
ctz: Yeah those seem a pretty minor construction on top of the existing certificate/spki types?

ctz:
This should also address/ameliorate https://github.com/quinn-rs/quinn/issues/2299

PR #2728, opened 2025-10-30
tests: move API tests into rustls-test

djc:
Results in a bunch of errors like this:

```rust
error[E0433]: failed to resolve: there are too many leading `super` keywords
  --> rustls-test/tests/raw_keys.rs:14:5
   |
14 | use super::provider;
   |     ^^^^^ there are too many leading `super` keywords
```

@ctz as the mastermind of the macro gunk here, care to have a look?

### DIFF
--- a/rustls-test/Cargo.toml
+++ b/rustls-test/Cargo.toml
@@ -3,7 +3,6 @@ name = "rustls-test"
 version = "0.1.0"
 edition = "2021"
 publish = false
-autotests = false
 
 [features]
 aws-lc-rs = ["rustls/aws-lc-rs"]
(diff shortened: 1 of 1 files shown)


rustls-test/Cargo.toml
```diff
@@ -4,5 +4,37 @@ version = "0.1.0"
 edition = "2021"
 publish = false
```
ctz, line 6: Try `autotests = false` here

djc: Ah!

djc:
Added a bunch of commits to make our integration test setup more boring/predictable.

PR #2731, opened 2025-11-03
quic: move key update test out of provider-specific code

djc:
I found that the aws-lc-rs provider depends on a number of private APIs. Additionally, I was angling for a way to make tests less dependent on macros (since they get in the way of Rust-Analyzer doing well).


rustls/src/quic.rs
```diff
@@ -1039,7 +1039,70 @@ mod tests {
...
+    use crate::quic::{HeaderProtectionKey, Secrets, Version};
+    use crate::{Side, TEST_PROVIDERS};
+
+    #[test]
+    fn key_update_test_vector() {
+        fn equal_okm(x: &OkmBlock, y: &OkmBlock) -> bool {
+            x.as_ref() == y.as_ref()
+        }
+
+        for provider in TEST_PROVIDERS {
```
djc, line 1053: So adding a loop in a bunch of tests might be slightly ugly, but it doesn't require any macros and seems substantially easier to understand than the macro indirections.

rustls/src/quic.rs
```diff
@@ -1039,7 +1039,70 @@ mod tests {
...
+            let aes_128_gcm = provider
+                .tls13_cipher_suites
+                .iter()
+                .find(|cs| cs.common.suite == CipherSuite::TLS13_AES_128_GCM_SHA256)
+                .unwrap();
+            let quic_suite = aes_128_gcm.quic.unwrap();
+
+            let mut secrets = Secrets::new(
```
djc, line 1061: `Secrets::new()` is private, so was trying to keep this outside of provider-specific tests.

ctz (approved):

rustls/src/quic.rs
ctz: True. The macros way was allowing `TLS13_AES_128_GCM_SHA256` to be referenced directly, rather than looked up, but that alone doesn't seem worth the squeeze.

Issue #2732, opened 2025-11-03
ConfigBuilder::with_single_cert_with_ocsp does not work as intended with empty `ocsp`


vuongDang:
**Checklist**
* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**

From the documentation and from previous version of rustls the function
[`ConfigBuilder::with_single_cert_with_ocsp`](https://github.com/rustls/rustls/blob/main/rustls/src/server/builder.rs#L86) should ignore the `ocsp` parameter if it's zero-length.

Starting from version 0.23.23 it seems like even when given an empty ocsp the server will require for one which results
in `Error code: SEC_ERROR_OCSP_MALFORMED_RESPONSE` when querying the server with Firefox.

**To Reproduce**

1.  Create a config with [`ConfigBuilder::with_single_cert_with_ocsp`](https://github.com/rustls/rustls/blob/main/rustls/src/server/builder.rs#L86)  and give it an empty vec for `ocsp`
2. Start a server with this config
3. Query the server with https
4. Get `Error code: SEC_ERROR_OCSP_MALFORMED_RESPONSE` on Firefox

**Applicable Version(s)**
- Windows 
- Rust: 1.90
- rustls: 0.23.34

**Expected behavior**
Following documentation and previous behavior I expect [`ConfigBuilder::with_single_cert_with_ocsp`](https://github.com/rustls/rustls/blob/main/rustls/src/server/builder.rs#L86)  to ignore the `ocsp` parameter when empty. 
Hence the server won't use ocsp when started. 

**Additional context**
Browsing quickly into the latest code I saw that: 
```
pub fn with_single_cert_with_ocsp(
        self,
        cert_chain: Vec<CertificateDer<'static>>,
        key_der: PrivateKeyDer<'static>,
        ocsp: Vec<u8>,
    ) -> Result<ServerConfig, Error> {
        let mut certified_key =
            CertifiedKey::from_der(cert_chain, key_der, self.crypto_provider())?;
        certified_key.ocsp = Some(ocsp);
        Ok(self.with_cert_resolver(Arc::new(SingleCertAndKey::from(certified_key))))
    }
```
I suspect `certified_key.ocsp` is supposed to be `None` when `ocsp` is an empty vec. 
I quickchecked the code so I don't have a good vision of the whole codebase to state that the bug is actually this.

I can do a PR if you agree on the issue and it helps.

ctz:
[...]

PR #2733, opened 2025-11-04
fix: only add ocsp in `ServerConfig` if `ocsp` parameter is not empty

vuongDang:
Fixes rustls/rustls#2732

Quick fix to the function `ConfigBuilder::with_single_cert_with_ocsp` to set the ocsp configuration
only if the given parameter is not 0-length.


djc:
Preparing 0.23.35 release in

- #2734

PR #2737, opened 2025-11-04
Support variable-length IVs in cipher operations

jedisct1:
Allow the `Iv` and `Nonce` types to support a variable length instead of a hardcoded size of 12 bytes.

The current built-in AEADs use a 96 bit nonce, but other AEADs may have different requirements.

### DIFF
--- a/rustls/src/crypto/cipher.rs
+++ b/rustls/src/crypto/cipher.rs
@@ -5,7 +5,7 @@ use core::fmt;
 use zeroize::Zeroize;
 
 use crate::enums::{ContentType, ProtocolVersion};
-use crate::error::Error;
+use crate::error::{ApiMisuse, Error};
 use crate::msgs::codec;
 pub use crate::msgs::message::{
     BorrowedPayload, InboundOpaqueMessage, InboundPlainMessage, OutboundChunks,
@@ -23,6 +23,11 @@ pub trait Tls13AeadAlgorithm: Send + Sync {
 
     /// The length of key in bytes required by `encrypter()` and `decrypter()`.
     fn key_len(&self) -> usize;
+
+    /// The length of IV in bytes required by `encrypter()` and `decrypter()`.
+    fn iv_len(&self) -> usize {
+        NONCE_LEN
+    }
 
     /// Convert the key material from `key`/`iv`, into a `ConnectionTrafficSecrets` item.
     ///
@@ -174,44 +179,64 @@ impl dyn MessageDecrypter {
 }
 
 /// A write or read IV.
-#[derive(Default)]
-pub struct Iv([u8; NONCE_LEN]);
+#[derive(Default, Clone)]
+pub struct Iv {
+    buf: [u8; Self::MAX_LEN],
+    used: usize,
+}
 
 impl Iv {
-    /// Create a new `Iv` from a byte array, of precisely `NONCE_LEN` bytes.
-    pub fn new(value: [u8; NONCE_LEN]) -> Self {
-        Self(value)
+    /// Create a new `Iv` from a byte slice.
+    ///
+    /// Returns an error if the length of `value` exceeds [`Self::MAX_LEN`].
+    pub fn new(value: &[u8]) -> Result<Self, Error> {
+        if value.len() > Self::MAX_LEN {
+            return Err(ApiMisuse::IvLengthExceedsMaximum {
+                actual: value.len(),
+                maximum: Self::MAX_LEN,
+            }
+            .into());
+        }
+        let mut buf = [0u8; Self::MAX_LEN];
+        buf[..value.len()].copy_from_slice(value);
+        Ok(Self {
+            buf,
+            used: value.len(),
+        })
     }
 
-    /// Create a new `Iv` from a byte slice, of precisely `NONCE_LEN` bytes.
-    pub fn copy(value: &[u8]) -> Self {
-        debug_assert_eq!(value.len(), NONCE_LEN);
-        let mut iv = Self::new(Default::default());
-        iv.0.copy_from_slice(value);
-        iv
+    /// Return the IV length.
+    #[expect(clippy::len_without_is_empty)]
+    pub fn len(&self) -> usize {
+        self.used
     }
+
+    /// Maximum supported IV length.
+    pub const MAX_LEN: usize = 16;
 }
 
 impl From<[u8; NONCE_LEN]> for Iv {
     fn from(bytes: [u8; NONCE_LEN]) -> Self {
-        Self(bytes)
+        Self::new(&bytes).expect("NONCE_LEN is within MAX_LEN")
     }
 }
 
 impl AsRef<[u8]> for Iv {
     fn as_ref(&self) -> &[u8] {
-        self.0.as_ref()
+        &self.buf[..self.used]
     }
 }
 
 /// A nonce.  This is unique for all messages on a connection.
-#[expect(clippy::exhaustive_structs)]
-pub struct Nonce(pub [u8; NONCE_LEN]);
+pub struct Nonce {
+    buf: [u8; Iv::MAX_LEN],
+    len: usize,
+}
 
 impl Nonce {
     /// Combine an `Iv` and sequence number to produce a unique nonce.
     ///
-    /// This is `iv ^ seq` where `seq` is encoded as a 96-bit big-endian integer.
+    /// This is `iv ^ seq` where `seq` is encoded as a big-endian integer.
     #[inline]
     pub fn new(iv: &Iv, seq: u64) -> Self {
         Self::new_inner(None, iv, seq)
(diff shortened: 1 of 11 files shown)


ctz:
Thanks for the contribution -- is this to allow for AEGIS support, or something else? Just interested in the motivation.

jedisct1:
Hi @ctz !

Indeed, this is to support AEGIS. However, other recent AEAD schemes like ROCCA-S and HiAE also require a 128-bit nonce.

More generally, this is to make the trait more future-proof.

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,148 @@ impl dyn MessageDecrypter {
...
+        self.len == 0
+    }
+
+    /// Convert to a fixed-size array of length `N`.
+    ///
+    /// Panics if the nonce length is not `N`.
+    ///
+    /// For standard nonces, use `nonce.to_array::<NONCE_LEN>()` or just `nonce.to_array()`
+    /// which defaults to `NONCE_LEN`.
+    pub fn to_array<const N: usize>(&self) -> [u8; N] {
```
djc: I don't think we want to panic here. The easiest solution might be to yield a `Result`? A better solution might be to parametrize the entire `Iv` with `const N: usize`, but not sure how complex that ends up getting.

jedisct1: Hi @djc and thanks for your feedback.

Parametrizing the entire `Iv` would introduce much more extensive changes, so indeed the easiest solution is to make the function return a `Result`.

Done!

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
-#[derive(Default)]
-pub struct Iv([u8; NONCE_LEN]);
+#[derive(Default, Clone)]
+pub struct Iv {
+    buf: [u8; Self::MAX_LEN],
+    used: usize,
+}
 
 impl Iv {
+    /// Maximum supported IV length.
```
djc: Nit: move this to the bottom of the impl block.

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
+    pub const MAX_LEN: usize = 16;
+
     /// Create a new `Iv` from a byte array, of precisely `NONCE_LEN` bytes.
     pub fn new(value: [u8; NONCE_LEN]) -> Self {
-        Self(value)
+        Self::copy(&value)
```
djc: Better to drop this, and rename `copy()` to `new()`, IMO.

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
+        }
+    }
+
+    /// Return the IV length.
+    pub fn len(&self) -> usize {
+        self.used
+    }
+
+    /// Return whether the IV is empty.
+    pub fn is_empty(&self) -> bool {
```
djc: I guess this is just here to appease clippy? Better to suppress the lint, IMO.

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
+    }
+
+    /// Return the nonce length.
+    pub fn len(&self) -> usize {
+        self.len
+    }
 
-        Self(seq_bytes)
+    /// Return whether the nonce is empty.
+    pub fn is_empty(&self) -> bool {
```
djc: As above.

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
+        self.len == 0
+    }
+
+    /// Convert to a fixed-size array of length `N`.
+    ///
+    /// Returns an error if the nonce length is not `N`.
+    ///
+    /// For standard nonces, use `nonce.to_array::<NONCE_LEN>()?` or just `nonce.to_array()?`
+    /// which defaults to `NONCE_LEN`.
+    pub fn to_array<const N: usize>(&self) -> Result<[u8; N], Error> {
```
djc, line 285: Nit: move this before `as_bytes()` and `len()`?

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
+        if self.len != N {
+            return Err(Error::General(alloc::format!(
```
djc: I think we should not do `General` for this. Maybe this is an `ApiMisuse` case?

rustls/src/crypto/cipher.rs
```diff
@@ -174,74 +179,150 @@ impl dyn MessageDecrypter {
...
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        &self.buf[..self.len]
+    }
+}
+
+impl AsRef<[u8]> for Nonce {
+    fn as_ref(&self) -> &[u8] {
+        &self.buf[..self.len]
```
djc: I don't think it makes sense to do both of these?

rustls/src/tls13/key_schedule.rs
```diff
@@ -907,11 +915,24 @@ pub fn derive_traffic_key(
...
+    derive_traffic_iv_with_len(expander, NONCE_LEN)
+}
+
+/// [HKDF-Expand-Label] where the output is an IV with a specified length.
+///
+/// For standard TLS 1.3 cipher suites, use [`derive_traffic_iv`] which uses
+/// the default IV length.
+///
+/// [HKDF-Expand-Label]: <https://www.rfc-editor.org/rfc/rfc8446#section-7.1>
+pub fn derive_traffic_iv_with_len(expander: &dyn HkdfExpander, iv_len: usize) -> Iv {
```
djc: Since we don't have to maintain API compatibility right now, suggest we just change the `derive_traffic_iv()` arguments instead of adding a parallel API.

rustls/src/crypto/cipher.rs
jedisct1: Done.

rustls/src/crypto/cipher.rs
jedisct1: Yeah, this is just for clippy. Like the vast majority of `is_empty()` methods that people add :) 

Removed.

rustls/src/crypto/cipher.rs
jedisct1: Done.

rustls/src/crypto/cipher.rs
jedisct1: `Deref` removed to only keep `AsRef`.

rustls/src/crypto/cipher.rs
jedisct1: Changed for an `ApiMisuse` case.

rustls/src/tls13/key_schedule.rs
jedisct1: I tried to avoid API changes, but if a signature change is acceptable, this is indeed better.

Changed, thanks!

rustls/src/crypto/cipher.rs
djc: Since it's public API, definitely.

rustls/src/crypto/cipher.rs
jedisct1: Done

rustls/src/crypto/cipher.rs
djc: I seem to remember there was a question here about making `new()` fallible, but that is gone now? Seems kinda important?

jedisct1: `new()` is now fallible.

Thanks!

rustls/src/crypto/cipher.rs
```diff
@@ -187,14 +187,22 @@ pub struct Iv {
 
 impl Iv {
     /// Create a new `Iv` from a byte slice.
-    pub fn new(value: &[u8]) -> Self {
-        debug_assert!(value.len() <= Self::MAX_LEN);
+    ///
+    /// Returns an error if the length of `value` exceeds [`Self::MAX_LEN`].
+    pub fn new(value: &[u8]) -> Result<Self, Error> {
+        if value.len() > Self::MAX_LEN {
+            return Err(crate::error::ApiMisuse::IvLengthExceedsMaximum {
```
djc: Nit: please import `ApiMisuse` here...

rustls/src/crypto/cipher.rs
```diff
@@ -550,12 +558,29 @@ mod tests {
...
+    #[test]
+    fn iv_length_exceeds_maximum() {
+        let too_long_iv = [0xAAu8; Iv::MAX_LEN + 1];
+        let result = Iv::new(&too_long_iv);
+
+        assert!(result.is_err());
+        assert!(matches!(
+            result,
+            Err(Error::ApiMisuse(
+                crate::error::ApiMisuse::IvLengthExceedsMaximum {
```
djc: ... and here.

rustls/src/crypto/cipher.rs
```diff
@@ -550,12 +558,29 @@ mod tests {
...
         assert_eq!(nonce_bytes[NONCE_LEN - 1], 0xFE);
     }
+
+    #[test]
+    fn iv_length_exceeds_maximum() {
+        let too_long_iv = [0xAAu8; Iv::MAX_LEN + 1];
+        let result = Iv::new(&too_long_iv);
+
+        assert!(result.is_err());
+        assert!(matches!(
```
djc: Don't think having both of these is useful.

djc (approved):
Nice!

rustls/src/crypto/cipher.rs
jedisct1: Done. `ApiMisuse` has been imported and is used directly.

djc:
Thanks for the iterations!

PR #2742, opened 2025-11-05
Tighten treatment of empty OCSP responses

ctz:
This ensures we both a) never send empty OCSP responses, and b) also never accept them.

a) has a related recent OpenSSL bug, and #2733 is related. But even after #2733 was possible by writing `credentials.ocsp = Some([])` or equiv. As it happens, the bogo shim does that so this provides test coverage for this change.
b) was the other end of the OpenSSL bug, and now we match BoringSSL and gnutls. It is a logical extension of #2637, #2430.


djc (approved):
Nice!

cpu (approved):

PR #2746, opened 2025-11-06
Remove intermediate `CommonState.received_plaintext` buffer in unbuffered API

Suficio:
Recently I tried to make use of the unbuffered API in a project and in an effort to understand the internals I was surprised to see that it uses an intermediate buffer `CommonState.received_plaintext` instead of decrypting in-place.

Similarly for issue #1723 it is difficult to implement the buffered API on top of the unbuffered API if this will lead to an extra copy.

The approach is simple:

- Introduce `UnborrowedPayload` which is an index into the user-passed buffer, the indexes of the range are managed by the existing `Locator` and `Delocator` utils.
- Pass in a `&mut Option<UnborrowedPayload>` into `CommonState.process_main_protocol` which gets populated if application data becomes available.

There are no breaking API changes that I am aware of (don't know the codebase very well).

There are two further improvements to make from here in a different PR:

- `ConnectionState::ReadTraffic` can be simplified to avoid necessity to loop in user code
  - This seems desirable since unbuffered API benchmarks already call `ReadTraffic::next_record().unwrap().unwrap()`
- `CommonState.received_plaintext` can be moved into `ConnectionCommon`

I don't have an appropriate benchmarking server set up for this use-case but basic server benchmark runs show over +10% throughput improvements in unbuffered benchmarks without meaningful regressions that I identified.

### DIFF
--- a/rustls/src/conn/unbuffered.rs
+++ b/rustls/src/conn/unbuffered.rs
@@ -336,42 +336,19 @@ pub struct ReadTraffic<'c, 'i, Side: SideData> {
     _conn: &'c mut UnbufferedConnectionCommon<Side>,
 
     payload: Payload<'i>,
-
-    // `payload` needs to remain allocated in order to meet lifetime
-    // requirements of `next_record`
-    is_terminated: bool,
 }
 
 impl<'c, 'i, Side: SideData> ReadTraffic<'c, 'i, Side> {
     fn new(_conn: &'c mut UnbufferedConnectionCommon<Side>, payload: Payload<'i>) -> Self {
-        Self {
-            _conn,
-            payload,
-            is_terminated: false,
-        }
+        Self { _conn, payload }
     }
 
     /// Decrypts and returns the next available app-data record
-    // TODO deprecate in favor of `Iterator` implementation, which requires in-place decryption
-    pub fn next_record(&mut self) -> Option<Result<AppDataRecord<'_>, Error>> {
-        if self.is_terminated {
-            return None;
-        }
-        self.is_terminated = true;
-        Some(Ok(AppDataRecord {
+    pub fn record(&self) -> AppDataRecord<'_> {
+        AppDataRecord {
             discard: 0,
             payload: self.payload.bytes(),
-        }))
-    }
-
-    /// Returns the payload size of the next app-data record *without* decrypting it
-    ///
-    /// Returns `None` if there are no more app-data records
-    pub fn peek_len(&self) -> Option<NonZeroUsize> {
-        if self.is_terminated {
-            return None;
         }
-        NonZeroUsize::new(self.payload.bytes().len())
     }
 }
 
--- a/examples/src/bin/unbuffered-async-client.rs
+++ b/examples/src/bin/unbuffered-async-client.rs
@@ -8,8 +8,8 @@ use std::sync::Arc;
 use rustls::client::{ClientConnectionData, UnbufferedClientConnection};
 use rustls::crypto::aws_lc_rs::DEFAULT_PROVIDER;
 use rustls::unbuffered::{
-    AppDataRecord, ConnectionState, EncodeError, EncryptError, InsufficientSizeError,
-    UnbufferedStatus, WriteTraffic,
+    ConnectionState, EncodeError, EncryptError, InsufficientSizeError, UnbufferedStatus,
+    WriteTraffic,
 };
 use rustls::{ClientConfig, RootCertStore};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
@@ -58,29 +58,23 @@ async fn converse(
         } = conn.process_tls_records(&mut incoming_tls[..incoming_used]);
 
         match dbg!(state.unwrap()) {
-            ConnectionState::ReadTraffic(mut state) => {
-                while let Some(res) = state.next_record() {
-                    let AppDataRecord {
-                        discard: new_discard,
-                        payload,
-                        ..
-                    } = res?;
-                    discard += new_discard;
+            ConnectionState::ReadTraffic(state) => {
+                let record = state.record();
+                discard += record.discard;
 
-                    if payload.starts_with(b"HTTP") {
-                        let response = core::str::from_utf8(payload)?;
-                        let header = response
-                            .lines()
-                            .next()
-                            .unwrap_or(response);
+                if record.payload.starts_with(b"HTTP") {
+                    let response = core::str::from_utf8(record.payload)?;
+                    let header = response
+                        .lines()
+                        .next()
+                        .unwrap_or(response);
 
-                        println!("{header}");
-                    } else {
-                        println!("(.. continued HTTP response ..)");
-                    }
-
-                    received_response = true;
+                    println!("{header}");
+                } else {
+                    println!("(.. continued HTTP response ..)");
                 }
+
+                received_response = true;
             }
 
             ConnectionState::EncodeTlsData(mut state) => {
--- a/examples/src/bin/unbuffered-client.rs
+++ b/examples/src/bin/unbuffered-client.rs
@@ -9,8 +9,8 @@ use std::sync::Arc;
 use rustls::client::{ClientConnectionData, EarlyDataError, UnbufferedClientConnection};
 use rustls::crypto::aws_lc_rs::DEFAULT_PROVIDER;
 use rustls::unbuffered::{
-    AppDataRecord, ConnectionState, EncodeError, EncryptError, InsufficientSizeError,
-    UnbufferedStatus, WriteTraffic,
+    ConnectionState, EncodeError, EncryptError, InsufficientSizeError, UnbufferedStatus,
+    WriteTraffic,
 };
 use rustls::{ClientConfig, RootCertStore};
 
@@ -63,29 +63,23 @@ fn converse(
         } = conn.process_tls_records(&mut incoming_tls[..incoming_used]);
 
         match dbg!(state.unwrap()) {
-            ConnectionState::ReadTraffic(mut state) => {
-                while let Some(res) = state.next_record() {
-                    let AppDataRecord {
-                        discard: new_discard,
-                        payload,
-                        ..
-                    } = res?;
-                    discard += new_discard;
+            ConnectionState::ReadTraffic(state) => {
+                let record = state.record();
+                discard += record.discard;
 
-                    if payload.starts_with(b"HTTP") {
-                        let response = core::str::from_utf8(payload)?;
-                        let header = response
-                            .lines()
-                            .next()
-                            .unwrap_or(response);
+                if record.payload.starts_with(b"HTTP") {
+                    let response = core::str::from_utf8(record.payload)?;
+                    let header = response
+                        .lines()
+                        .next()
+                        .unwrap_or(response);
 
-                        println!("{header}");
-                    } else {
-                        println!("(.. continued HTTP response ..)");
-                    }
-
-                    received_response = true;
+                    println!("{header}");
+                } else {
+                    println!("(.. continued HTTP response ..)");
                 }
+
+                received_response = true;
             }
 
             ConnectionState::EncodeTlsData(mut state) => {
--- a/examples/src/bin/unbuffered-server.rs
+++ b/examples/src/bin/unbuffered-server.rs
@@ -80,29 +80,23 @@ fn handle(
         } = conn.process_tls_records(&mut incoming_tls[..incoming_used]);
 
         match dbg!(state.unwrap()) {
-            ConnectionState::ReadTraffic(mut state) => {
-                while let Some(res) = state.next_record() {
-                    let AppDataRecord {
-                        discard: new_discard,
-                        payload,
-                        ..
-                    } = res?;
-                    discard += new_discard;
+            ConnectionState::ReadTraffic(state) => {
+                let record = state.record();
+                discard += record.discard;
 
-                    if payload.starts_with(b"GET") {
-                        let response = core::str::from_utf8(payload)?;
-                        let header = response
-                            .lines()
-                            .next()
-                            .unwrap_or(response);
+                if record.payload.starts_with(b"GET") {
+                    let response = core::str::from_utf8(record.payload)?;
+                    let header = response
+                        .lines()
+                        .next()
+                        .unwrap_or(response);
 
-                        println!("{header}");
-                    } else {
-                        println!("(.. continued HTTP request ..)");
-                    }
-
-                    received_request = true;
+                    println!("{header}");
+                } else {
+                    println!("(.. continued HTTP request ..)");
                 }
+
+                received_request = true;
             }
 
             ConnectionState::ReadEarlyData(mut state) => {
--- a/fuzz/fuzzers/unbuffered.rs
+++ b/fuzz/fuzzers/unbuffered.rs
@@ -70,11 +70,7 @@ fn process<S: SideData>(status: UnbufferedStatus<'_, '_, S>) -> Option<usize> {
         Ok(ConnectionState::TransmitTlsData(xmit)) => xmit.done(),
         Ok(ConnectionState::WriteTraffic(_)) => return None,
         Ok(ConnectionState::BlockedHandshake) => return None,
-        Ok(ConnectionState::ReadTraffic(mut read)) => loop {
-            let Some(_app_data) = read.next_record() else {
-                break;
-            };
-        },
+        Ok(ConnectionState::ReadTraffic(_)) => {}
         Ok(st) => panic!("unhandled state {st:?}"),
         Err(_) => return None,
     };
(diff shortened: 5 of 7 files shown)


Suficio:
I am not quite sure how to go about addressing the failing test case `read_traffic_not_consumed_too_early`.

The test case expects the unbuffered API to buffer the data such that it remains available in `ReadTraffic::peek_len()` until `ReadTraffic::next_record()` is called. 

This is a snippet of the failing test case with my own comments added.
```rust
// REVIEW: Peek length of incoming data which has already been allocated in the user buffer
let (_, discard) = read_traffic(
    server.process_tls_records(client_to_server_buf.filled()),
    |rt| assert_eq!(rt.peek_len(), NonZeroUsize::new(5)),
);
assert!(discard > 0);
// REVIEW: Discard data from the buffer
client_to_server_buf.discard(discard);

// REVIEW: Test case expects that data remained buffered in `CommonState` but already deallocated from the user buffer
let (_, discard) = read_traffic(
    server.process_tls_records(client_to_server_buf.filled()),
    |rt| assert_eq!(rt.peek_len(), NonZeroUsize::new(5)),
);
assert_eq!(discard, 0);
```

This is the related issue and PR: #2031 #2338

@ctz would appreciate some guidance here

ctz:
> @ctz would appreciate some guidance here

I think it would be OK to vary the details here, so long as the overall behaviour (which you've correctly described) is still attainable.

I think that would look like:

- making use of `AppDataRecord::discard` field to delay the discard to the right point
- keeping internal state that tracks whether the undiscarded prefix of the user buffer contains a decrypted record (the `read_traffic_not_consumed_too_early` test would exercise this)

(there may be other ways of achieving it too, or the above may equally be insufficient.)

Suficio:
>  keeping internal state that tracks whether the undiscarded prefix of the user buffer contains a decrypted record

@ctz I think this is impossible to do with in-place decryption on the unbuffered API.

To pass this test I would have to re-introduce a buffer. My impression is that the original issue requests behavior that is incompatible with an unbuffered API since `CommonState` state will advance whether the user calls discard or not.
 
Breaking down the current unbuffered workflow:
 
 1. Data is written into buffer
 2. Data is deframed and passed to the protocol handler
 3. If frame is application data then a new vector is allocated and added to the `CommonState` buffer
 4. Application checks the length of the allocated vector and decide to process
 
 With this PR:
 
 1. Data is written into buffer
 2. Data is deframed and passed to the protocol handler
 3. If frame is application data then we return a reference to the data
 4. Application checks the length of the slice and decide to process
 
Note that in both cases the frame was written into the buffer in it's entirety. The current unbuffered API will actually consume and allocate the data before the user has a chance to check it's payload.

Rewriting the test case slightly:
```
// REVIEW: Process and re-allocate data in internal buffer
let res = server.process_tls_records(client_to_server_buf.filled());
// REVIEW: Peek length of allocated data
let (_, discard) = read_traffic(res, |rt| assert_eq!(rt.peek_len(), NonZeroUsize::new(5)));
```

In short, I would advocate to remove this test case as it mandates using a buffer in the unbuffered API.

Suficio:
I would also like to highlight some API simplifications that can be made as a consequence of this PR:
https://github.com/Suficio/rustls/commit/6fea1f4c43d7bf2ae979a40ca9de2a7501b296fe

I have not included these in this PR since they contain breaking changes to the API.

Suficio:
Also highlight some results from my benchmark run, these were performed on an `r5a.large` AWS instance:

  |   |   |   |   | Current MB/s | Proposed MB/s | % Change
-- | -- | -- | -- | -- | -- | -- | --
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1851.59 | 2205.29 | 19.1025010936547
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 | max_fragment_size:default | recv | 2185.55 | 2540.8 | 16.2544897165473
bulk-unbuffered | TLSv1_3 | TLS13_AES_128_GCM_SHA256 | max_fragment_size:default | recv | 2187.87 | 2522.27 | 15.2842719174357
bulk-unbuffered | TLSv1_3 | TLS13_AES_128_GCM_SHA256 | max_fragment_size:default | recv | 2189.52 | 2523.81 | 15.2677299134057
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 | max_fragment_size:10000 | recv | 2122.16 | 2429.16 | 14.4663927319335
bulk-unbuffered | TLSv1_3 | TLS13_AES_128_GCM_SHA256 | max_fragment_size:10000 | recv | 2132.09 | 2423.77 | 13.6804731507582
bulk-unbuffered | TLSv1_3 | TLS13_AES_128_GCM_SHA256 | max_fragment_size:10000 | recv | 2136.14 | 2425.39 | 13.5407791624145
bulk-unbuffered | TLSv1_3 | TLS13_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1927.55 | 2185.39 | 13.3765661072346
bulk-unbuffered | TLSv1_3 | TLS13_AES_128_GCM_SHA256 | max_fragment_size:default | recv | 2187.73 | 2476.48 | 13.1986122601966
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1950.12 | 2207.5 | 13.1981621643796
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1874.97 | 2119.61 | 13.0476754294735
bulk-unbuffered | TLSv1_3 | TLS13_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1932.69 | 2183.22 | 12.9627617465812
bulk-unbuffered | TLSv1_3 | TLS13_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1935.49 | 2184.19 | 12.8494593100455
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1877.49 | 2118.34 | 12.8282973544466
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1877.01 | 2116.94 | 12.7825637583178
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1955.09 | 2204.4 | 12.7518426261707
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | recv | 1956.76 | 2203.68 | 12.6188188638361
bulk-unbuffered | TLSv1_3 | TLS13_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1888.21 | 2123.5 | 12.461008044656
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1879.93 | 2113.64 | 12.4318458666014
bulk-unbuffered | TLSv1_3 | TLS13_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1889.63 | 2124.52 | 12.4304758074332
bulk-unbuffered | TLSv1_3 | TLS13_AES_256_GCM_SHA384 | max_fragment_size:10000 | recv | 1888.75 | 2123.13 | 12.4092653871608
bulk-unbuffered | TLSv1_3 | TLS13_CHACHA20_POLY1305_SHA256 | max_fragment_size:default | recv | 854.35 | 901.4 | 5.50711066892959
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 | max_fragment_size:default | recv | 857.3 | 903.52 | 5.39134492009799
bulk-unbuffered | TLSv1_3 | TLS13_CHACHA20_POLY1305_SHA256 | max_fragment_size:10000 | recv | 835.37 | 878.78 | 5.19649975459975
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 | max_fragment_size:10000 | recv | 830.06 | 871.76 | 5.02373322410429
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 | max_fragment_size:10000 | recv | 830.22 | 871.86 | 5.01553805015538
bulk-unbuffered | TLSv1_2 | TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 | max_fragment_size:default | send | 1669.3 | 1752.98 | 5.01287965015276

djc:
> I have not included these in this PR since they contain breaking changes to the API.

FYI, `main` is currently open for breaking changes.

Suficio:
I made two additional commits:
- https://github.com/rustls/rustls/pull/2746/commits/57a463226c5fbbc7f3144851a5c3247d4ed132eb Removes the `read_traffic_not_consumed_too_early` test case since this PR does not re-introduce the original issue and is no longer testable since the test case relied on internal buffering which this PR removes.
- https://github.com/rustls/rustls/pull/2746/commits/34bb8528b0c4d56985f36613fbb6be7746d72371 PR simplifying the `ReadTraffic` API

> FYI, main is currently open for breaking changes.

@djc I originally did not include the changes to avoid focusing on the breaking changes and instead the performance improvement. However, since the behavior of `ReadTraffic::next_record` changed I figure it was appropriate to include the PR that also removes it.

ctz:
> In short, I would advocate to remove this test case as it mandates using a buffer in the unbuffered API.

At the minute the unbuffered API does use the user's buffer to store data that is unprocessed between calls into the library (and strongly requires the user to not mutate/truncate it except via the `discard` output). I think above I was trying to say that we could repeat this use for unaccepted `ReadTraffic` data. I think it is quite important that a user of the unbuffered API can avoid immediately processing a `ReadTraffic` state. If we can't do that, then such a caller must itself buffer that data; potentially an unlimited amount; so what we've done is move the buffering requirement outside the library while paying the cost of a complex API.

SuficioAC:
@ctz

> At the minute the unbuffered API does use the user's buffer to store data that is unprocessed between calls

Are you sure?

All `State` implementations for `ExpectTraffic` currently call `CommonState::take_received_plaintext` which has implementation:

```rust
pub(crate) fn take_received_plaintext(&mut self, bytes: Payload<'_>) {
        self.temper_counters.received_app_data();
        self.received_plaintext
            .append(bytes.into_vec());
}
```

> I think it is quite important that a user of the unbuffered API can avoid immediately processing a ReadTraffic state.

There are two options:
1. They can avoid calling `process_tls_records` which advances the state machine
2. They can buffer the data for later

I care about this too which is why I removed the unconditional allocation in `CommonState::take_received_plaintext` which would always buffer the data for the user even if they didn't want to process.

> If we can't do that, then such a caller must itself buffer that data; potentially an unlimited amount; so what we've done is move the buffering requirement outside the library 

This logic remains unchanged in this PR. This PR only removes an unconditional allocation in the existing process.

> so what we've done is move the buffering requirement outside the library

This is not a fair analysis.

If a user requires buffering they are free to do so, however, at this moment `rustls::unbuffered` will always unconditionally buffer all the data received. See `CommonState::take_received_plaintext`.

> paying the cost of a complex API.

I understand this is subjective but I will highlight the before and after implementation of the unbuffered benchmark to highlight the effects of the API changes.

```diff
match client.process_tls_records(&mut input[input_used..]) {
-  UnbufferedStatus { state: Ok(ConnectionState::ReadTraffic(mut rt)), discard, .. } => {
+  UnbufferedStatus { state: Ok(ConnectionState::ReadTraffic(rt)), discard, .. } => {
    input_used += discard;
-   let record = rt.next_record().unwrap().unwrap();
+   let record = rt.record();
    input_used += record.discard;
    expected -= record.payload.len();
  }
  st => panic!("unexpected read state: {st:?}"),
}
```

You will find similar simplifications in the unit tests, where the application state additionally went from `Vec<Vec<u8>>`  to `Vec<u8>`.

```diff
ConnectionState::ReadTraffic(mut state) => {
- let mut records = vec![];
- let mut peeked_len = state.peek_len();
-
- while let Some(res) = state.next_record() {
-   let payload = res.unwrap().payload.to_vec();
-   assert_eq!(NonZeroUsize::new(payload.len()), peeked_len);
-   records.push(payload);
-   peeked_len = state.peek_len();
- }
-
- assert_eq!(None, peeked_len);
-
- State::ReceivedAppData { records }
+ let record = state.record().payload.to_vec();
+ State::ReceivedAppData { record }
}
```

Where is the added complexity?

ctz:
> Are you sure?

Yes, during the handshake. Handshake messages can be received both encrypted and fragmented. To be processed they are encrypted and reassembled within the caller's buffer, over successive calls. (Not talking about application data, where you're right in your assessment of the current situation.)

> There are two options:
> 
> 1. They can avoid calling `process_tls_records` which advances the state machine
> 2. They can buffer the data for later

So I guess the alternate answer for #2031 is: never call `process_tls_records` with incoming data of length N unless you can also accept plaintext data of size N? I can live with that.

> I understand this is subjective but I will highlight the before

Sorry, I am comparing here the complexity of the buffered and unbuffered APIs.

Suficio:
@ctz 

> Yes, during the handshake.

I tried not to touch any of the handshaking logic as part of this PR by limiting myself only to the application data pipeline, do you feel something has changed here?

> So I guess the alternate answer for https://github.com/rustls/rustls/issues/2031 is: never call process_tls_records with incoming data of length N unless you can also accept plaintext data of size N? I can live with that.

For an application that can reject application data using this criteria the logic would be:

1. Call `process_tls_records` and receive `ReadTraffic`
2. Decide whether application is willing to process application data based on its contents or length
3. If data can be processed immediately, process
4. If data may not be processed immediately, allocate the data into a separate buffer _(current behavior)_ or keep the `ReadTraffic` struct around until you are ready to process.
5. If data won't be processed we continue calling `process_tls_records` and receive the next `ReadTraffic`

Eitherway, you can call `record().len()` to check the length of the application data and size your application data buffer immediately or process in-place.

> never call `process_tls_records` with incoming data of length N

You keep calling `process_tls_records` but you don't actually have to do anything with the data. By the time `process_tls_records` is called the data has already been allocated in the user's buffer.

_As a side note, `ReadTraffic` may expose some TLS header data to give applications more information about how to process the upcoming packet (allowing skipping decryption too)_

> Sorry, I am comparing here the complexity of the buffered and unbuffered APIs.

My higher-level goal is to implement the buffered API in terms of the unbuffered API, the problem right now is that both the receive and send side of the unbuffered API are buffered causing a performance penalty. The unbuffered pipeline needs to be optimized before that can happen.

rustls/src/conn/unbuffered.rs
```diff
@@ -331,51 +333,45 @@ impl<Side: SideData> fmt::Debug for ConnectionState<'_, '_, Side> {
...
+    _conn: &'c mut UnbufferedConnectionCommon<Side>,
 
-    // owner of the latest chunk obtained in `next_record`, as borrowed by
-    // `AppDataRecord`
-    chunk: Option<Vec<u8>>,
+    payload: Payload<'i>,
+
+    // `payload` needs to remain allocated in order to meet lifetime
+    // requirements of `next_record`
+    is_terminated: bool,
```
ctz: Perhaps this could be achieved using `payload: Option<Payload<'i>>`, and then doing `Option::take()` in `next_record()`?

ctz (approved):

rustls/src/conn/unbuffered.rs
ctz: This fell away in a later commit.

rustls/src/common_state.rs
```diff
@@ -890,11 +890,46 @@ pub(crate) trait State<Side>: Send + Sync {
...
+        }
+    }
+
+    pub(crate) fn reborrow<'b>(self, delocator: &Delocator<'b>) -> Payload<'b> {
+        match self {
+            Self::Unborrowed(range) => Payload::Borrowed(delocator.slice_from_range(&range)),
+            Self::Owned(payload) => Payload::Owned(payload),
+        }
+    }
+}
```
djc, line 965: I think this whole thing could make do with some comments both (a) motivating the complexity and (b) explaining the mechanics here and how they're intended to be used.

Suficio: Added the comments to the original commit: 0bacb3c1961f57277abadcaff6739314b4f0282d

Suficio:
> I guess it's an improvement? 

I would say that a 5-19% increase in receive throughput on the unbuffered API is an improvement 😅

Since plaintext data is now decrypted in-place it opens up an opportunity to remove even more allocations in application-level protocols.

PR #2747, opened 2025-11-07
tests: port server tests to TEST_PROVIDERS

djc:


rustls/src/server/test.rs
```diff
@@ -1,19 +1,38 @@
...
 use crate::msgs::message::{Message, MessagePayload};
-use crate::{CommonState, Side};
+use crate::pki_types::pem::PemObject;
+use crate::pki_types::{CertificateDer, PrivateKeyDer};
+use crate::server::{ServerConfig, ServerConnection};
+use crate::suites::CipherSuiteCommon;
+use crate::sync::Arc;
+use crate::tls12::Tls12CipherSuite;
+use crate::version::TLS12_VERSION;
+#[cfg(any(feature = "aws-lc-rs", feature = "ring"))]
```
ctz: This whole file is under that gate, so this gate could be removed.

PR #2751, opened 2025-11-09
Reduce `OutboundOpaqueMessage` API

Suficio:
I hope to make progress towards issue #1723 by removing the buffering requirement for the outbound pipeline. This PR arose as a small opportunity to simplify logic in that direction.

This PR contains two reductions to the `OutboundOpaqueMessage` API:

1. All uses of `OutboundOpaqueMessage::read` were immediately followed by `OutboundOpaqueMessage::into_plain_message`. Moving `read` directly into `PlainMessage` removes a bit of roundabout type wrangling and removes the ownership requirement of the data within `OutboundOpaqueMessage`.

2. Out of all message types, `PlainMessage`, `OutboundPlainMessage`, etc. `OutboundOpaqueMessage` was the only one that defined a `new` constructor. This PR removes that constructor, leaving only one method to initialize the type.

### DIFF
--- a/fuzz/fuzzers/fragment.rs
+++ b/fuzz/fuzzers/fragment.rs
@@ -6,15 +6,15 @@ extern crate rustls;
 use rustls::internal::msgs::base::Payload;
 use rustls::internal::msgs::codec::Reader;
 use rustls::internal::msgs::fragmenter::MessageFragmenter;
-use rustls::internal::msgs::message::{Message, OutboundOpaqueMessage, PlainMessage};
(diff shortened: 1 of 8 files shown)


djc:
Thanks for looking at this. I think it would be good to split this into two separate commits for the two independent changes you made. See also a bunch of related changes I queued up after reviewing this PR:

- #2752

SuficioAC:
@djc 

>  think it would be good to split this into two separate commits for the two independent changes you made.

Done

Your changes also look good!

PR #2754, opened 2025-11-10
crypto: access ticketer via crypto provider

### LINKED ISSUE #1876: Feature request: a way to set/get default ticketer dynamically
**Checklist**

Currently I am trying to make Rocket use the default CryptoProvider if it is set. [This](https://github.com/rwf2/Rocket/pull/2758#issuecomment-2021688150) is the only blocking issue. 

Either ticketer is added as a field on CryptoProvider or add a way to set/get default ticketer separately.

I considered simply using ring Ticketer but this may be unexpected to the users who change CryptoProvider.

djc:
Fixes #1876 by providing an optional `TicketerFactory` implementation in `CryptoProvider`s. 

@cpu in that issue:

> I'm not sure that we want to link the `ProducesTicketer` implementation into the `CryptoProvider` struct as a field but I'm curious what others think. 🤔
> 
> There's not really a "default" ticketer from my perspective, just two convenient implementations of `dyn ProducesTickets` made available for the built in cryptography provider options. From that viewpoint I think what's missing is a way to get the right one depending on whether you have unambiguously set crate features such that it's possible to pick the right one without more thought.

I think it's probably helpful to have crypto providers provide a default option. In this PR, I've made the `CryptoProvider` field an `Option` so that providers can still opt out of providing it (although that makes usage little uglier). I think the API I came up with (independently) mirrors @ctz's suggestion in the issue. With this setup, callers can still provider their own `TicketProducer` implementation but crypto providers have a way of exposing a default one for their crypto stack.

Quoting myself:

> So I guess it would be cool if we could reformulate the `ring_like` `AeadTicketer` (which somewhat confusingly lives in `crypto::ring::ticketer`) to rely on, maybe, a `Tls13AeadAlgorithm` implementation instead of using the `ring_like::aead` primitives directly?

I don't think this is workable, since the existing AEAD traits are very focused on messages.

### DIFF
--- a/rustls/src/crypto/aws_lc_rs/mod.rs
+++ b/rustls/src/crypto/aws_lc_rs/mod.rs
@@ -13,8 +13,11 @@ use crate::error::{Error, OtherError};
 use crate::rand::GetRandomFailed;
 #[cfg(feature = "std")]
 use crate::sync::Arc;
+#[cfg(feature = "std")]
+use crate::ticketer::TicketRotator;
+use crate::tls12::Tls12CipherSuite;
+use crate::tls13::Tls13CipherSuite;
 use crate::webpki::WebPkiSupportedAlgorithms;
-use crate::{TicketRotator, Tls12CipherSuite, Tls13CipherSuite};
 
 /// Hybrid public key encryption (HPKE).
 pub mod hpke;
@@ -30,6 +33,8 @@ pub(crate) mod kx;
 pub(crate) mod quic;
 #[cfg(feature = "std")]
 pub(crate) mod ticketer;
+#[cfg(feature = "std")]
+use ticketer::Rfc5077Ticketer;
 pub(crate) mod tls12;
 pub(crate) mod tls13;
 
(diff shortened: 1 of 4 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -227,6 +227,9 @@ pub struct CryptoProvider {
 
     /// Provider for loading private [`SigningKey`]s from [`PrivateKeyDer`].
     pub key_provider: &'static dyn KeyProvider,
+
+    /// Provider for creating [`TicketProducer`]s for stateless session resumption.
+    pub ticketer_factory: Option<&'static dyn TicketerFactory>,
```
ctz, line 232: Thoughts on making this mandatory? A genuine lack of ability to make a ticketer could be expressed by returning an error from `ticketer()`?

ctz (approved):

rustls/src/crypto/mod.rs
djc: It seems reasonable to me. Will follow up.

PR #2758, opened 2025-11-11
Enqueue key update message on demand

Suficio:
I am hoping to make progress towards in-place encryption in the unbuffered pipeline and this change resulted as a minor improvement along the way.

Currently the key update notify message is separately allocated in `CommonState` where it is encrypted and later written to the `CommonState.sendable_tls` buffer.

The critical issue with this approach is `CommonState::enqueue_key_update_notification` seems to only be called within a receive context where, if in-place encryption is implemented, it will not have access to the outgoing buffer.

This PR moves the initialization of the key update message into `CommonState::perhaps_write_key_update` and simultaneously removes the separate allocation within `CommonState`.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -13,7 +13,7 @@ use crate::msgs::base::Payload;
 use crate::msgs::codec::Codec;
 use crate::msgs::enums::{AlertLevel, KeyUpdateRequest};
 use crate::msgs::fragmenter::MessageFragmenter;
-use crate::msgs::handshake::{HandshakeMessagePayload, ProtocolName};
+use crate::msgs::handshake::{HandshakeMessagePayload, HandshakePayload, ProtocolName};
 use crate::msgs::message::{
     InboundPlainMessage, Message, MessagePayload, OutboundChunks, OutboundOpaqueMessage,
     OutboundPlainMessage, PlainMessage,
@@ -51,7 +51,7 @@ pub struct CommonState {
     message_fragmenter: MessageFragmenter,
     pub(crate) received_plaintext: ChunkVecBuffer,
     pub(crate) sendable_tls: ChunkVecBuffer,
-    queued_key_update_message: Option<Vec<u8>>,
+    queued_key_update_message: bool,
 
     /// Protocol whose key schedule should be used. Unused for TLS < 1.3.
     pub(crate) protocol: Protocol,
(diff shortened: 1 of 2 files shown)


rustls/src/common_state.rs
```diff
@@ -88,7 +88,7 @@ impl CommonState {
             message_fragmenter: MessageFragmenter::default(),
             received_plaintext: ChunkVecBuffer::new(Some(DEFAULT_RECEIVED_PLAINTEXT_LIMIT)),
             sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
-            queued_key_update_message: None,
+            queued_key_update_message: false,
```
djc, line 91: Suggest renaming this field, maybe to `send_key_update` or `key_update_requested`.

rustls/src/common_state.rs
```diff
@@ -460,7 +460,19 @@ impl CommonState {
     }
 
     pub(crate) fn perhaps_write_key_update(&mut self) {
-        if let Some(message) = self.queued_key_update_message.take() {
+        if self.queued_key_update_message {
+            self.queued_key_update_message = false;
+
+            let message = PlainMessage::from(Message {
```
djc, line 466: On balance, I think I prefer keeping this outlined in the `Message` constructor, as before.

djc:
It looks like bogo doesn't like this -- perhaps the new delay between requesting and encoding the message is problematic...

ctz:
> It looks like bogo doesn't like this -- perhaps the new delay between requesting and encoding the message is problematic...

Yeah, the key update message needs to be encrypted with the current encryption keys, not the "next" ones.

It's a larger refactor, but we could extract a type that represents the ability to encrypt one final message from `RecordLayer` (which would bind together the previous MessageEncrypter and the sequence number). Then insert that type in the `queued_key_update_message` slot and consume it when writing the message.

Suficio:
> Yeah, the key update message needs to be encrypted with the current encryption keys, not the "next" ones.

Missed this during test, will be more thorough next time.

Shame, thought this would be an easy win but it is definitely not worth to pursue if it requires any further refactoring since it's a bit of a shot in the dark before I have a solid proposal for in-place encryption.

Happy to leave this PR open or close if we want to pick up this idea in the near future.

ctz:
> Happy to leave this PR open or close if we want to pick up this idea in the near future.

Going to close this for now. If you find in your other refactoring that it is required, please shout and I can address it quite quickly.

PR #2759, opened 2025-11-13
Extract ring crypto provider into rustls-ring crate

djc:


rustls/src/quic.rs
```diff
@@ -826,14 +826,16 @@ impl PacketKeySet {
     }
 }
 
-pub(crate) struct KeyBuilder<'a> {
+/// Helper for building QUIC packet and header protection keys
+pub struct KeyBuilder<'a> {
```
djc, line 830: This seems boring enough to expose it publicly? Used in the rustls-ring QUIC tests. We could alternatively make it `#[doc(hidden)]`....

rustls-ring/src/sign.rs
```diff
@@ -367,19 +365,66 @@ fn load_key(
         .load_private_key(der)
 }
 
+/// Prepend stuff to `bytes_a` + `bytes_b` to put it in a DER SEQUENCE.
+pub(crate) fn wrap_concat_in_sequence(bytes_a: &[u8], bytes_b: &[u8]) -> Vec<u8> {
```
djc, line 369: This is copied from `rustls::x509`, rather than exposing it publicly from that crate (which doesn't need the two top-level functions here). Alternatively, could expose it under `internals` or make it otherwise `#[doc(hidden)]`?

rustls/src/crypto/ring/ticketer.rs
```diff
@@ -105,8 +102,6 @@ impl TicketProducer for AeadTicketer {
                 .maximum_ciphertext_len
                 .load(Ordering::SeqCst)
         {
-            #[cfg(debug_assertions)]
-            debug!("rejected over-length ticket");
```
djc, line 109: Not sure if it's worth bringing in (or exposing) logging stuff for these...

rustls-ring/src/tls12.rs
```diff
@@ -415,13 +415,13 @@ fn gcm_iv(write_iv: &[u8], explicit: &[u8]) -> Iv {
     Iv::new(&iv).expect("IV length is NONCE_LEN, which is within MAX_LEN")
 }
 
+const MAX_FRAGMENT_LEN: usize = 16384;
```
djc, line 415: Copied from the main crate. Could instead expose it somehow? Seems boring enough and unlikely to change that copying is okay.

rustls/src/crypto/tls12.rs
```diff
@@ -91,7 +91,8 @@ pub trait PrfSecret: Send + Sync {
     fn prf(&self, output: &mut [u8], label: &[u8], seed: &[u8]);
 }
 
-pub(crate) fn prf(out: &mut [u8], hmac_key: &dyn hmac::Key, label: &[u8], seed: &[u8]) {
+#[doc(hidden)]
+pub fn prf(out: &mut [u8], hmac_key: &dyn hmac::Key, label: &[u8], seed: &[u8]) {
```
djc, line 95: I don't think there's an easy way to replace this with previously public API? Since this seems fairly unlikely to change, feels okay to expose this as hidden.

rustls/src/lib.rs
```diff
@@ -670,6 +670,4 @@ mod sealed {
 const TEST_PROVIDERS: &[&crypto::CryptoProvider] = &[
     #[cfg(feature = "aws-lc-rs")]
     &crypto::aws_lc_rs::DEFAULT_PROVIDER,
-    #[cfg(feature = "ring")]
```
djc, line 673: So my plan here didn't quite work out -- the compiler complains that the types are different, ostensibly because there are two different compilations of rustls being used. Propose to just leave it out for this PR and figure out a solution after this lands, before externalizing the aws-lc-rs provider.

rustls-ring/src/sign.rs
ctz: I believe this and `wrap_in_octet_string` can move from the core crate as this file is the only caller of them.

rustls/src/crypto/ring/ticketer.rs
ctz: yeah, agree with dropping the logging

rustls-ring/src/tls12.rs
ctz: Could do an associated const on `OutboundPlainMessage`? But can be left for later.

rustls-ring/src/sign.rs
djc: Did that, but the underlying `asn1_wrap()` is still used in `DistinguishedName::in_sequence()` and `public_key_to_spki()`.

rustls/src/lib.rs
ctz: Yeah I think I found this before when introducing `rustls-test`; the `cfg(test)` compilation of rustls is not the same one as used by dev-dependencies that also depend on rustls. Pain in the ass.

rustls/src/crypto/tls12.rs
ctz: Seems OK for now. I would generally prefer for tests to be next to the impl, but that would need some in-crate `cfg(test)` hash functions and HMAC. Not a complete imposition.

rustls/src/quic.rs
ctz: Yeah that seems alright.

rustls-ring/src/tls12.rs
djc: We actually have more users on the inbound side, so not sure attaching it to `OutboundPlainMessage` makes sense.

rustls-ring/src/sign.rs
ctz: I think this is OK for now. Eventually we could deduplicate them into pki-types?

PR #2767, opened 2025-11-19
Simplify plaintext send

Suficio:
I am hoping to make some progress towards a proposal for in-place encryption, this PR arose as an opportunity along the way to simplify some related logic.

- `CommonState::send_plain` is only called with parameter `Limit::Yes` in `CommonState::buffer_plaintext` leading to dead code
- `CommonState::send_plain_non_buffering` is only called with `Limit::No` in `CommonState::flush_plaintext` which can call `CommonState::send_appdata_encrypt` directly

There is a minor issue that may occur in the current `CommonState::buffer_plaintext` and `CommonState::send_early_plaintext` where if the send buffer is full it will queue empty `OutboundChunks` in `CommonState.sendable_tls`. This is fixed by this PR.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -341,7 +341,7 @@ impl CommonState {
             return 0;
         }
 
-        self.send_appdata_encrypt(data[..len].into(), Limit::No)
+        self.send_appdata_encrypt(data[..len].into())
     }
 
     /// Fragment `m`, encrypt the fragments, and then queue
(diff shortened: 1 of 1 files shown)


rustls/src/common_state.rs
```diff
@@ -345,31 +345,17 @@ impl CommonState {
     }
 
     /// Like send_msg_encrypt, but operate on an appdata directly.
-    fn send_appdata_encrypt(&mut self, payload: OutboundChunks<'_>, limit: Limit) -> usize {
```
djc: I like the direction here, but it would be nice to make these changes in smaller commits so that it's more obvious when reviewing that each step is correct.

Suficio: I split the second commit into two, let me know if I should try simplify it futher.

djc: Yes, please split it more. Suggest doing one commit per function where you are removing the `limit` argument.

(Also please keep the first line of commit messages below 80 or so chars.)

Suficio: Third time's the charm, split the removal of the `limit` argument into two commits.

rustls/src/common_state.rs
```diff
@@ -456,7 +456,10 @@ impl CommonState {
         }
 
         while let Some(buf) = sendable_plaintext.pop() {
-            self.send_plain_non_buffering(buf.as_slice().into(), Limit::No);
+            if buf.is_empty() {
+                continue;
+            }
```
djc: This seems unrelated to the other changes in this commit?

Suficio:
I fixed the remaining code coverage issues.

- Modified two test cases to ensure that empty messages are not sent.
- Removed empty message check in `CommonState::start_outgoing_traffic` since all empty plaintext messages that enter the outgoing buffer are caught by the existing two checks.

rustls/src/common_state.rs
djc: This seems unrelated to other changes in this commit?

Suficio: Yes, it has been removed from the commit and the following commits have been edited.

Suficio:
Rebased after https://github.com/rustls/rustls/pull/2783

PR #2768, opened 2025-11-20
Clean up public API (enums)

djc:
This started because the aws-lc-rs provider has dependencies on a bunch of HPKE types that were actually living in the `msgs` hierarchy. In turn, that got me looking at moving more public types out of `msgs` (to make the internal division more clear), and then had me feeling dissatisfied with making the top-level `enums` module kind of a dumping ground. Happy to discuss if some of these are moving in the wrong direction, and whether a different hierarchy/organization would make more sense.

Also nice: moving more niche types out of the crate root (which is kind of a prime real estate from a documentation perspective).

### DIFF
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -27,7 +27,8 @@ use crate::sync::Arc;
 use crate::time_provider::DefaultTimeProvider;
 use crate::time_provider::TimeProvider;
 use crate::unbuffered::{EncryptError, TransmitTlsData};
-use crate::{DistinguishedName, KeyLog, WantsVerifier, compress, verify};
+use crate::verify::DistinguishedName;
+use crate::{KeyLog, WantsVerifier, compress, verify};
 
 /// A trait for the ability to store client session data, so that sessions
 /// can be resumed in future connections.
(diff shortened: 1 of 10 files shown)


ctz (approved):
LGTM!

I'm not 100% about putting protocol-level details like `AlertDescription` into error.rs, but equally I agree that enums.rs is quite unprincipled[^1] and I don't have a better suggestion. The same goes for `NamedGroup` et al in `crypto::`, but it's a big improvement on them being in the root.

[^1]: originally enums.rs was intended to be auto-generated from IANA, but I don't really like that idea now, and I prefer to keep hand-crafted information close to the definitions.

djc:
> Do you think it might make sense to start on that alongside changes like this as a draft .md doc in-repo?

It does make some amount of sense, but I find myself unmotivated to do the legwork until the release is closer.

Also, if we end up missing something we can always update the release notes later as people point things out, and for many of the changes Rust-Analyzer will be pretty good at pointing out the fix for downstream projects.

Also also, I think we should not release 0.24.0 until after we've ported our whole stack and a few well-known downstreams which will give us a decent way to note which incompatibilities people are likely to run into.

PR #2769, opened 2025-11-21
Move record_layer module into crypto::cipher

djc:


djc:
In a previous run this failed `limitedclient_does_not_contain_tls12_code`, but I have a hard time coming up with reasons why...

PR #2772, opened 2025-11-21
crypto: introduce kx module

djc:
We could consider moving `NamedGroup` into `kx`, too...


rustls/src/crypto/kx.rs
```diff
@@ -383,6 +380,297 @@ pub enum KeyExchangeAlgorithm {
     ECDHE,
 }
 
+/// This module contains parameters for FFDHE named groups as defined
+/// in [RFC 7919 Appendix A](https://datatracker.ietf.org/doc/html/rfc7919#appendix-A).
+pub mod ffdhe {
```
ctz: I would have this in a separate file due to the large constants.

PR #2775, opened 2025-11-21
Rework handshake "alignment" checks

ctz:
This is a precursor to moving away from the state machine having unlimited mutable access to `CommonState`.

But so far is just replacing a bool with a zero-size type that demonstrates the dataflow from the handshake deframer to the parts of code that want to know that there are no pending handshake fragments. That is needed either because the message is supposed to be at the end of the flight or because the decryption keying is being changed.

### DIFF
--- a/rustls/src/msgs/deframer/handshake.rs
+++ b/rustls/src/msgs/deframer/handshake.rs
@@ -372,6 +372,7 @@ impl FragmentSpan {
 /// Proof type that the handshake deframer is aligned.
 ///
 /// See [`HandshakeDeframer::aligned()`] for more details.
+#[must_use]
 #[derive(Clone, Copy)]
 pub(crate) struct HandshakeAlignedProof(());
 


rustls-test/tests/api/api.rs
```diff
@@ -1555,6 +1555,27 @@ fn large_client_hello_acceptor() {
...
+#[test]
+fn excess_client_hello_acceptor() {
+    // this is a trivial ClientHello, followed by a fragment of a ClientHello
+    let mut hello = encoding::basic_client_hello(vec![]);
+    let hello2 = hello[4..10].to_vec();
+    let hello2 = encoding::handshake_framing(HandshakeType::ClientHello, hello2);
+    hello.extend(hello2);
+    let hello = encoding::message_framing(ContentType::Handshake, ProtocolVersion::TLSv1_2, hello);
+
+    let mut acceptor = rustls::server::Acceptor::default();
```
djc: Import `Acceptor`? Not sure if you're doing this on purpose but it generally makes refactoring harder, and IMO also makes the code have lower signal vs noise ratio (assuming the name is reasonably non-conflicting).

rustls/src/msgs/deframer/handshake.rs
```diff
@@ -368,6 +369,12 @@ impl FragmentSpan {
     }
 }
 
+/// Proof type that the handshake deframer is aligned.
+///
+/// See [`HandshakeDeframer::aligned()`] for more details.
+#[derive(Clone)]
```
djc: If we're going to make this `Clone`, might as well make it `Copy`?

rustls/src/common_state.rs
```diff
@@ -17,6 +17,7 @@ use crate::log::{debug, error, trace, warn};
 use crate::msgs::alert::AlertMessagePayload;
 use crate::msgs::codec::Codec;
 use crate::msgs::deframer::buffers::{Delocator, Locator};
+use crate::msgs::deframer::handshake::HandshakeAlignedProof;
```
djc: Suggest re-exporting from the `deframer`, I don't think exposing internal `deframer` structure is helpful.

rustls-test/tests/api/api.rs
ctz: Not really doing it on purpose, just following prevailing uses in this file.

rustls/src/common_state.rs
ctz: Do you mean doing `pub(crate) use handshake::HandshakeAlignedProof;` at one module layer above? Note that this isn't done for any other type in `deframer` or below.

djc: Yes, that is what I mean. Maybe it would make sense to do that for everything inside `deframer` in the same commit?

ctz: Doing this supported by #2779

ctz:
(note new commit about these checks being missing for QUIC prior to this PR.)

PR #2779, opened 2025-11-24
Flatten `msgs::deframer` module

ctz:


rustls/src/msgs/deframer/mod.rs
```diff
@@ -6,7 +6,9 @@ use crate::msgs::codec::Reader;
 use crate::msgs::message::{HEADER_SIZE, MessageError, read_opaque_message_header};
 
 pub(crate) mod buffers;
-pub(crate) mod handshake;
+pub(crate) use handshake::HandshakeDeframer;
```
djc: I like to put the imports just below the relevant `mod` statement from which they're importing.

rustls/src/msgs/deframer/mod.rs
```diff
@@ -5,9 +5,12 @@ use crate::error::{Error, InvalidMessage};
 use crate::msgs::codec::Reader;
 use crate::msgs::message::{HEADER_SIZE, MessageError, read_opaque_message_header};
 
+mod buffers;
 mod handshake;
 
-pub(crate) mod buffers;
+pub(crate) use buffers::{
+    BufferProgress, DeframerSliceBuffer, DeframerVecBuffer, Delocator, Locator,
+};
```
djc, line 11: I meant more like this:

```rust
mod buffers;
pub(crate) use buffers::{..};

mod handshake;
pub(crate) use handshake::{..};
```

PR #2780, opened 2025-11-24
crypto: further tweaks to the public API

djc:


rustls/src/crypto/mod.rs
```diff
@@ -367,7 +367,7 @@ impl Borrow<[&'static Tls13CipherSuite]> for CryptoProvider {
 /// how they map to TLS [`SignatureScheme`]s.
 #[expect(clippy::exhaustive_structs)]
 #[derive(Clone, Copy)]
```
ctz: I think this rename implies that this type is used for more than it actually is. If someone provides their own verifiers, the value is not inspected by rustls.

The documentation on the `CryptoProvider::signature_verification_algorithms`  field does an OK job of saying that, but the docs on this type doesn't.

djc: Ohh, interesting. I've ditched the relevant commits from this PR for now.

PR #2781, opened 2025-11-24
crypto: merge message types

djc:
During work on an earlier PR I noticed that we have a bunch of duplication around the message types:

- `PlainMessage`
- `InboundPlainMessage`
- `InboundOpaqueMessage`
- `OutboundOpaqueMessage`
- `OutboundPlainMessage`

This PR refactors these types into a single `EncodedMessage` that abstracts over 4 different payload types:

- `Plain`
- `InboundOpaque` (previously `BorrowedPayload`)
- `OutboundOpaque` (previously `PrefixedPayload`)
- `OutboundPlain` (previously `OutboundChunks`)

(`InboundPlainMessage` is now just the same as `PlainMessage`: `EncodedMessage<Plain<'_>>`.)

Picked the name `EncodedMessage` to avoid conflicts with `Message`, which we use for a fully parsed message.

### DIFF
--- a/rustls/src/crypto/cipher/mod.rs
+++ b/rustls/src/crypto/cipher/mod.rs
@@ -1,22 +1,18 @@
 use alloc::boxed::Box;
 use alloc::string::ToString;
-use alloc::vec::Vec;
 use core::fmt;
 
 use zeroize::Zeroize;
 
 use crate::enums::{ContentType, ProtocolVersion};
-use crate::error::{ApiMisuse, Error, InvalidMessage};
-use crate::msgs::base::hex;
-use crate::msgs::codec::{self, Codec, Reader};
-use crate::msgs::message::read_opaque_message_header;
+use crate::error::{ApiMisuse, Error};
+use crate::msgs::codec;
 use crate::suites::ConnectionTrafficSecrets;
 
-mod inbound;
-pub use inbound::InboundOpaque;
-
-mod outbound;
-pub use outbound::{OutboundOpaque, OutboundPlain};
+mod messages;
+pub use messages::{
+    EncodedMessage, InboundOpaque, MessageError, OutboundOpaque, OutboundPlain, Payload,
+};
 
 mod record_layer;
 pub(crate) use record_layer::{Decrypted, PreEncryptAction, RecordLayer};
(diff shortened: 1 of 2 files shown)


rustls/src/crypto/cipher/mod.rs
```diff
@@ -464,26 +461,37 @@ impl PlainMessage {
...
 }
 
-/// An externally length'd payload
+/// A decrypted TLS frame payload with external length.
+///
+/// It is used for joining and fragmenting messages.
 #[non_exhaustive]
 #[derive(Clone, Eq, PartialEq)]
-pub enum Payload<'a> {
+pub enum Plain<'a> {
```
ctz: I would probably do this rename in a separate commit? But also the doc comment is quite specific about what this is for, whereas this type is used very generally (eg, for the bodies of extensions)

rustls/src/msgs/message.rs
```diff
@@ -169,10 +169,10 @@ impl Message<'_> {
...
 
-impl TryFrom<PlainMessage> for Message<'static> {
+impl TryFrom<EncodedMessage<Plain<'_>>> for Message<'_> {
     type Error = InvalidMessage;
 
-    fn try_from(plain: PlainMessage) -> Result<Self, Self::Error> {
+    fn try_from(plain: EncodedMessage<Plain<'_>>) -> Result<Self, Self::Error> {
         Ok(Self {
             version: plain.version,
             payload: MessagePayload::new(plain.typ, plain.version, plain.payload.bytes())?
```
ctz: The performance issue is here, because the critical codepath switched from `TryFrom<InboundPlainMessage<'a>> for Message<'a> ` (deleted below) to this impl, which copies everything as it was formerly impl'd for `Message<'static>`

rustls/src/crypto/cipher/mod.rs
ctz: (i'm also not sure if this is a good name, since in some contexts it can be used to contain a ciphertext of some description, eg in the context of ECH extensions, or a finished message, or handshake hash)

rustls/src/msgs/message.rs
djc: Ahh, thanks for looking into it.

rustls/src/crypto/cipher/mod.rs
djc: Dropped this rename. Clarified the docstring.

rustls/src/msgs/message.rs
djc: Fixed that. There's still seemingly some performance impact, though it's much smaller. Do you think it's worth looking into? Happy to do that myself.

ctz: i think the resumption-only server-only finding is maybe worth looking into, at least for a limited time

rustls/src/crypto/cipher/inbound.rs
```diff
@@ -106,33 +106,7 @@ impl<'a> InboundOpaque<'a> {
...
-/// It results from decryption.
-#[expect(clippy::exhaustive_structs)]
-#[derive(Debug)]
-pub struct InboundPlainMessage<'a> {
-    /// The content type of the message.
-    pub typ: ContentType,
-    /// The protocol version of the message.
-    pub version: ProtocolVersion,
-    /// The decrypted payload of the message.
-    pub payload: &'a [u8],
```
ctz, line 142: The other performance reduction is due to this commit, where the replacement for `InboundPlainMessage` of `EncodedMessage<Payload>` is a little larger, but most importantly Drop. That is not necessary, really, because it is always a `Payload::Borrowed` value. That bloats out basically every function that accepts one by value.

I pushed 35a4b9d54acb6c98a30f35d67131c335e57f55a3 but it probably warrants a bit more care/squashing into this commit or another.

djc: Squashed and rebased, but we should both check if anything weird crept in during the rebase.

rustls/src/crypto/cipher/inbound.rs
```diff
@@ -106,33 +106,7 @@ impl<'a> InboundOpaque<'a> {
...
-    /// We passthrough ChangeCipherSpec messages in the deframer without decrypting them.
-    /// Note: this is prior to the record layer, so is unencrypted. See
-    /// third paragraph of section 5 in RFC8446.
-    pub(crate) fn is_valid_ccs(&self) -> bool {
-        self.typ == ContentType::ChangeCipherSpec && self.payload == [0x01]
-    }
-}
-
-/// Decode a TLS1.3 `TLSInnerPlaintext` encoding.
+/// Decode a TLS1.3 `TLSInnerPayloadtext` encoding.
```
ctz: find-replace casualty here

rustls/src/conn/mod.rs
```diff
@@ -925,7 +925,7 @@ impl<Side: SideData> ConnectionCore<Side> {
             }
         };
 
-        // Should `InboundPlainMessage` resolve to plaintext application
+        // Should `EncodedMessage<Payload>` resolve to plaintext application
```
ctz, line 928: outdated comment
```suggestion
        // Should `EncodedMessage` resolve to plaintext application
```

rustls/src/conn/mod.rs
```diff
@@ -1114,6 +1114,8 @@ impl<Side: SideData> ConnectionCore<Side> {
                 return Err(PeerMisbehaved::MessageInterleavedWithHandshakeMessage.into());
             }
 
+            let message: EncodedMessage<&[u8]> = message;
```
ctz: apologies this can be removed; it is me forcing the compiler to tell me where the types are wrong

rustls/src/conn/mod.rs
```diff
@@ -1269,7 +1271,7 @@ impl<Side: SideData> ConnectionCore<Side> {
 /// Data specific to the peer's side (client or server).
 pub trait SideData: Debug {}
 
-/// An InboundPlainMessage which does not borrow its payload, but
+/// An [`EncodedMessage<Payload<'_>>`] which does not borrow its payload, but
```
ctz, line 1272: outdated comment

rustls/src/conn/mod.rs
djc: Killed it.

rustls/src/conn/mod.rs
djc: This seems to be removed in my local branch in the right commit...

rustls/src/crypto/cipher/inbound.rs
djc: Fixed.

rustls/src/conn/mod.rs
djc: This also seems to be fixed.

PR #2782, opened 2025-11-24
Promote lint configuration to workspace

ctz:


rustls-test/tests/api/api.rs
```diff
@@ -1016,21 +1016,21 @@ fn assert_lt(left: usize, right: usize) {
 #[test]
 fn connection_types_are_not_huge() {
     // Arbitrary sizes
-    assert_lt(mem::size_of::<ServerConnection>(), 1600);
-    assert_lt(mem::size_of::<ClientConnection>(), 1600);
+    assert_lt(size_of::<ServerConnection>(), 1600);
```
djc, line 1019: This change looks weird combined with the removal of the `std::mem` import. Where is this coming from now? Is `size_of()` just in the prelude now?

ctz: Yep, since 1.80

PR #2784, opened 2025-11-26
Rework sending alerts

ctz:
This is another simplification of the interface between the handshake state machine and rest of the library. Now every TLS alert is sent only based on a `rustls::Error` being produced. That means there is simply no possibility of (eg) sending two alerts in a certain codepath, and simplifies code that depended on `Context`/`CommonState` solely to send an alert.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -1818,8 +1818,7 @@ fn after_read(opts: &Options, sess: &mut Connection, conn: &mut net::TcpStream) 
 
 fn orderly_close(conn: &mut net::TcpStream) {
     // assuming we just flush()'d, we will write no more.
-    conn.shutdown(net::Shutdown::Write)
-        .unwrap();
+    let _ = conn.shutdown(net::Shutdown::Write);
 
     // wait for EOF
     let mut buf = [0u8; 32];


bogo/src/main.rs
```diff
@@ -1754,6 +1754,10 @@ fn handle_err(opts: &Options, err: Error) -> ! {
         Error::PeerMisbehaved(PeerMisbehaved::RejectedEarlyDataInterleavedWithHandshakeMessage) => {
             quit(":DECRYPTION_FAILED_OR_BAD_RECORD_MAC:")
         }
+        Error::PeerMisbehaved(PeerMisbehaved::ReceivedIllegalAlertLevel(_)) => quit(":BAD_ALERT:"),
+        Error::PeerMisbehaved(PeerMisbehaved::ReceivedIllegalWarningAlert(_)) => {
+            quit(":BAD_ALERT:")
+        }
```
djc: Nit: merge these arms?

rustls/src/error/mod.rs
```diff
@@ -979,6 +979,8 @@ pub enum PeerMisbehaved {
     OfferedIncorrectCompressions,
     PskExtensionMustBeLast,
     PskExtensionWithMismatchedIdsAndBinders,
+    ReceivedIllegalAlertLevel(AlertDescription),
+    ReceivedIllegalWarningAlert(AlertDescription),
```
djc: Nit: in the interest of conciseness, maybe skip the `Received` which kinda intuitively makes sense?

Also on the `Illegal` token, could we make do with something that is more common in our error stack already, like `Invalid`, `Unsupported` or `Unknown`? Or do you view all of these as semantically distinct? (If so, perhaps we should write that up somewhere.)

rustls/src/error/mod.rs
```diff
@@ -979,6 +979,8 @@ pub enum PeerMisbehaved {
     OfferedIncorrectCompressions,
     PskExtensionMustBeLast,
     PskExtensionWithMismatchedIdsAndBinders,
+    ReceivedIllegalAlertLevel(AlertDescription),
```
djc: Seems weird to put the `AlertDescription` in here when the `AlertLevel` is the part that's causing trouble?

rustls/src/error/mod.rs
```diff
@@ -1016,6 +1042,40 @@ pub enum PeerMisbehaved {
     UnsolicitedEchExtension,
 }
 
+impl From<PeerMisbehaved> for AlertDescription {
```
djc, line 1047: I think this could be `From<&PeerMisbehaved>` (and same for `PeerIncompatible`), avoiding the `clone()` calls in the `TryFrom<&Error>` impl?

rustls/src/error/mod.rs
```diff
@@ -154,6 +154,31 @@ pub enum Error {
...
 }
 
+/// Determine which alert should be sent for a given error.
+///
+/// If this mapping fails, no alert is sent.
+impl TryFrom<&Error> for AlertDescription {
+    type Error = ();
+
+    fn try_from(error: &Error) -> Result<Self, Self::Error> {
+        match error {
```
djc: Nit: alternatively, maybe hoist the `Ok(..)` and yeet the `Err(())` directly by explicitly `return`ing?

rustls/src/error/mod.rs
```diff
@@ -154,6 +154,31 @@ pub enum Error {
     Other(OtherError),
 }
 
+/// Determine which alert should be sent for a given error.
+///
+/// If this mapping fails, no alert is sent.
+impl TryFrom<&Error> for AlertDescription {
```
djc, line 160: Would prefer to see this in a commit that also uses it.

rustls/src/common_state.rs
```diff
@@ -257,14 +257,19 @@ impl CommonState {
...
-            Err(e @ Error::InappropriateMessage { .. })
-            | Err(e @ Error::InappropriateHandshakeMessage { .. }) => {
-                Err(self.send_fatal_alert(AlertDescription::UnexpectedMessage, e))
-            }
             Err(e) => Err(e),
         }
     }
 
+    pub(crate) fn send_fatal_alert_for_error(&mut self, error: &Error) {
+        if let Ok(alert) = AlertDescription::try_from(error) {
```
djc: let-else this to reduce rightward drift?

rustls/src/common_state.rs
```diff
@@ -257,14 +257,19 @@ impl CommonState {
...
             Ok(next) => Ok(next),
-            Err(e @ Error::InappropriateMessage { .. })
-            | Err(e @ Error::InappropriateHandshakeMessage { .. }) => {
-                Err(self.send_fatal_alert(AlertDescription::UnexpectedMessage, e))
-            }
             Err(e) => Err(e),
         }
     }
 
+    pub(crate) fn send_fatal_alert_for_error(&mut self, error: &Error) {
```
djc: So this silently does nothing when the `TryFrom` impl yields `Err(())`. Maybe the name should be adapted to reflect that? Like `maybe_send_fatal_alert()`? (The `for_error` stanza doesn't seem super-useful given that it's part of the signature, but I assume it was needed to disambiguate from existing methods.)

rustls/src/common_state.rs
```diff
@@ -257,14 +257,19 @@ impl CommonState {
         };
         match state.handle(&mut cx, msg) {
```
djc: Nit: you cleaned this up later, but I think that should just be in this commit.

rustls/src/conn/mod.rs
```diff
@@ -942,6 +942,11 @@ impl<Side: SideData> ConnectionCore<Side> {
             let opt_msg = match res {
                 Ok(opt_msg) => opt_msg,
                 Err(e) => {
+                    self.common_state
```
djc, line 941: Did you test this commit on its own? Seems to me that we end up sending fatal alerts twice if we do it here but also still have all the old stuff that's only removed in the next commit -- which is to say, it feels like these should be squashed?

rustls/src/conn/mod.rs
```diff
@@ -1170,32 +1170,6 @@ impl<Side: SideData> ConnectionCore<Side> {
         }
     }
 
-    fn handle_deframe_error(&mut self, error: Error, state: Option<&dyn State<Side>>) -> Error {
-        match error {
-            error @ Error::InvalidMessage(_) => {
-                if self.common_state.is_quic() {
-                    self.common_state.quic.alert = Some(AlertDescription::DecodeError);
```
djc, line 1170: Where/how is this covered? Is there test coverage for it?

rustls/src/client/ech.rs
```diff
@@ -847,13 +846,6 @@ pub(crate) struct EchAccepted {
     pub(crate) sent_extensions: Vec<ExtensionType>,
 }
 
-pub(crate) fn fatal_alert_required(
```
djc, line 857: I don't see the caller disappear in this commit?

rustls/src/client/ech.rs
```diff
@@ -517,7 +517,6 @@ impl EchState {
         &self,
         hrr: &HelloRetryRequest,
         cs: &Tls13CipherSuite,
-        common: &mut CommonState,
```
djc, line 522: So didn't this trigger a warning for an unused value in the previous commit?

rustls/src/server/connection.rs
```diff
@@ -351,6 +349,16 @@ mod buffered {
     pub struct AcceptedAlert(ChunkVecBuffer);
 
     impl AcceptedAlert {
+        pub(super) fn promote(
```
djc: Naming nit: maybe `from_error()` or `for_error()`?

djc (commented):
Very nice!

rustls/src/error/mod.rs
ctz: Added, as a u8 (I originally had this, but then didn't want to publicise the AlertLevel enum solely for a case where the value is not part of the enum!)

rustls/src/error/mod.rs
ctz: removed `Received`

there are quite a lot of `Illegal`s in `PeerMisbehaved`. generally my thinking around this wording is:

- Illegal: never allowed by the standard
- Unsupported: allowed by the standard, but not handled by rustls
- Unknown: possibly allowed by a future version of the standard? but not handled by rustls

not going to claim that these are all used consistently though!

rustls/src/error/mod.rs
ctz: Have done this. Alternatively we could make both of these Copy?

rustls/src/conn/mod.rs
ctz: Good spot -- the code builds but doesn't pass tests at this point. I was trying to keep the substantive new code away from the horde of the next commit. Will squash them.

rustls/src/conn/mod.rs
ctz: It turns out that even on main, this is dead code. There is no place where a quic connection can reach `handle_deframe_error()`. Have added a commit that improves coverage of `alert()` in the quic API.

rustls/src/error/mod.rs
djc: Making them `Copy` seems reasonable.

rustls/src/client/ech.rs
ctz: Squashed this hunk into the previous commit.

rustls/src/client/ech.rs
ctz: Yes it does; resolved in this commit. The problem is both commits are already pretty noisy. WDYT?

djc: I'd prefer squashing all the stuff so we end up with clean commits. I actually want to see how/where cleanup happens when it happens...

ctz: I have squashed this too.

rustls/src/common_state.rs
```diff
@@ -255,14 +247,18 @@ impl CommonState {
...
-            | Err(e @ Error::InappropriateHandshakeMessage { .. }) => {
-                Err(self.send_fatal_alert(AlertDescription::UnexpectedMessage, e))
-            }
-            Err(e) => Err(e),
-        }
+
+        state.handle(&mut cx, msg)
+    }
+
+    pub(crate) fn try_send_fatal_alert(&mut self, error: &Error) {
```
djc: Sorry, nit: don't love `try_` for this since it usually denotes fallibility.

djc (approved):

rustls/src/error/mod.rs
djc: (Unresolving this, feel free to resolve and address in a different PR.)

PR #2786, opened 2025-11-28
Ticket generations

### LINKED ISSUE #2778: The test of TicketRotator maybe not correct
When test "ticketrotator_remains_usable_over_temporary_ticketer_creation_failure", the origin code use a `fail_generator` and set the time to `now + 10s` to  simulate the situation where the rotation fail. But, when `encrpt` or `decrypt` later, it will still use `UnixTime::now()` and obviously it's not yet time for rotation, so the test will always pass.

So, was the original design that **even if the rotation failed, the previous generator would still be used?**

https://github.com/rustls/rustls/blob/ad3c1c6ed86638f8376728c78ac3f61e4340675f/rustls/src/ticketer.rs#L181-L212

`encrypt` will `UnixTime::now()`:
https://github.com/rustls/rustls/blob/ad3c1c6ed86638f8376728c78ac3f61e4340675f/rustls/src/ticketer.rs#L112-L113

`decrypt` will `UnixTime::now()`:
https://github.com/rustls/rustls/blob/ad3c1c6ed86638f8376728c78ac3f61e4340675f/rustls/src/ticketer.rs#L118-L119

djc:
Fixes #2778.

cc @BamLubi I hope this addresses your concerns!

### DIFF
--- a/rustls/src/ticketer.rs
+++ b/rustls/src/ticketer.rs
@@ -39,38 +39,62 @@ impl TicketRotator {
             generator,
             lifetime,
             state: RwLock::new(TicketRotatorState {
-                current: Generation {
+                current: Some(Generation {
                     producer: generator()?,
                     expires_at: UnixTime::now()
                         .as_secs()
                         .saturating_add(lifetime.as_secs()),
-                },
+                }),
                 previous: None,
             }),
         })
     }
 
     fn encrypt_at(&self, message: &[u8], now: UnixTime) -> Option<Vec<u8>> {
-        self.maybe_roll(now)?
-            .current
-            .producer
-            .encrypt(message)
+        let state = self.maybe_roll(now)?;
+
+        // If we have a current ticketer, use it. We don't need to check its
+        // expiration time; if it would have expired, we would have rolled above.
+        if let Some(current) = &state.current {
+            return current.producer.encrypt(message);
+        }
+
+        // If we don't have a previous ticketer, we can't encrypt.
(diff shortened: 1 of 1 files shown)


rustls/src/ticketer.rs
```diff
@@ -39,38 +39,62 @@ impl TicketRotator {
...
+                return Some(plain);
+            }
+        }
+
+        // If we don't have a previous ticketer, we can't decrypt.
+        let Some(prev) = &state.previous else {
+            return None;
+        };
+
+        // If the previous ticketer is more than one `lifetime` old, decline to encrypt.
```
ctz: Isn't this guaranteed to be true? How do we end up with a `state.previous` otherwise?

s/encrypt/decrypt/ in comment

djc: Here's a timeline:

- t = 0, we generate a ticketer
- t = 1 lifetime, we fail to generate a ticketer, the old ticketer is now `previous` and `current` becomes `None`
- t = 1.5 lifetimes <-- we are here, `previous` is more than one but less than 2 lifetimes old so we still allow encryption with it
- t = 2.0 lifetimes, we stop encrypting tickets

This is more or less copying the new encryption behavior, which may not make sense?

ctz: Sorry, I misread. That timeline does reflect what I think is a desired behaviour.

djc:
(Fixed the typo.)

PR #2789, opened 2025-12-01
Clarify `ExpectCertificate` reuse by `ExpectCompressedCertificate`

ctz:


rustls/src/client/tls13.rs
```diff
@@ -1024,15 +1011,18 @@ impl State<ClientConnectionData> for ExpectCertificate {
...
+        self.handle_cert_payload(require_handshake_msg_move!(
             m,
             HandshakeType::Certificate,
             HandshakePayload::CertificateTls13
-        )?;
+        )?)
+    }
+}
 
+impl ExpectCertificate {
```
djc, line 1008: Inherent impl before trait impl?

PR #2791, opened 2025-12-02
Minor handshake code tidying

ctz:


rustls/src/server/hs.rs
```diff
@@ -35,15 +35,20 @@ pub(super) type NextState = Box<dyn State<ServerConnectionData>>;
 pub(super) type NextStateOrError = Result<NextState, Error>;
 pub(super) type ServerContext<'a> = crate::common_state::Context<'a, ServerConnectionData>;
 
-#[derive(Default)]
```
ctz, line 37: Possibly this commit is not worth it...

rustls/src/server/hs.rs
```diff
@@ -34,15 +35,20 @@ pub(super) type NextState = Box<dyn State<ServerConnectionData>>;
...
+    pub(super) config: &'a ServerConfig,
+    pub(super) client_hello: &'a ClientHelloPayload,
 }
 
-impl ExtensionProcessing {
-    pub(super) fn new(extra_exts: ServerExtensionsInput<'static>) -> Self {
+impl<'a> ExtensionProcessing<'a> {
+    pub(super) fn new(
+        extra_exts: ServerExtensionsInput<'static>,
+        config: &'a ServerConfig,
```
djc: Nit: `config` is more boring than `hello` so switch the order?

rustls/src/server/hs.rs
djc: I kinda like it.

PR #2792, opened 2025-12-03
Deduplicate outgoing

Suficio:
As before I am hoping to make progress towards a proposal for in-place encryption and this PR arose as an opportunity to simplify things along the way.

`CommonState::write_plaintext` and `CommonState::send_appdata_encrypt` largely share the same internal logic but duplicate most of their implementations, this PR re-organizes the internals such that they may be shared generically.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -382,12 +382,11 @@ impl CommonState {
         {
             PreEncryptAction::Nothing => {}
 
-            // Close connection once we start to run out of
-            // sequence space.
+            // Close connection once we start to run out of sequence space.
             PreEncryptAction::RefreshOrClose => {
                 match self.negotiated_version {
                     Some(ProtocolVersion::TLSv1_3) => {
-                        // driven by caller, as we don't have the `State` here
+                        // Driven by caller, as we don't have the `State` here.
                         self.refresh_traffic_keys_pending = true;
                     }
                     _ => {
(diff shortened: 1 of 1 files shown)


Suficio:
@djc @ctz

Codecov is unable to pass on this PR since the buffered API does not have a unit test for the `EncryptExhausted` case. There is a test case on the unbuffered API, `tls12_connection_fails_after_key_reaches_confidentiality_limit`.

I couldn't implement a similar test since the error is ignored on the buffered API.

I can propagate this error by defining a `EncryptExchausted` type which can be turned into `EncryptError` or `io::Error` depending on the API used. At that point, I should be able to implement a similar unit test.

Let me know how to proceed.

Suficio:
I was wrong, there is an equivalent test for the buffered API, and the case that CodeCov is failing on is documented as untestable.

How can this PR proceed?

Suficio:
@ctz @djc May I get a review on this PR?

Codecov no longer fails after a rebase. The `s390x` failure appears unrelated to the changes in the PR.

This has been rebased to main.

djc:
There are [a lot of changes](https://github.com/rustls/rustls/pull/2837) in progress around `CommonState`, and it looks like the current iteration of the unbuffered API might go away. As such, please bear with us for a bit -- we'll make sure to reconsider your changes before 0.24 is released.

Suficio:
@djc

> the current iteration of the unbuffered API might go away

This PR is aligned with that goal.

It distills logic from the unbuffered and buffered API into a common unit, thus making the unbuffered API easier to remove.

Another issue with the current setup of `CommonState` is that `send_single_fragment` handles alert logic in the plaintext path, preventing features such as in-place encryption.

I am mostly interested in making progress on in-place encryption the same way that https://github.com/rustls/rustls/pull/2746 enabled in-place decryption and removed the necessity of `CommonState.received_plaintext`.

PR #2797, opened 2025-12-08
Delay exposure of `peer_identity()` until handshake `Finished` message is confirmed

ctz:
This PR also does quite a bit of tidying of handshake state code. More to come!

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -163,9 +163,6 @@ mod server_hello {
                     cx.common
                         .start_encryption_tls12(&secrets, Side::Client);
 
-                    // Since we're resuming, we verified the certificate and
-                    // proof of possession in the prior session.
-                    cx.common.peer_identity = Some(resuming.peer_identity().clone());
                     cx.common.handshake_kind = Some(HandshakeKind::Resumed);
                     let cert_verified = verify::PeerVerified::assertion();
                     let sig_verified = verify::HandshakeSignatureValid::assertion();
@@ -174,6 +171,8 @@ mod server_hello {
                         Ok(Box::new(ExpectNewTicket {
                             config,
                             secrets,
+                            // Since we're resuming, we verified the certificate and
+                            // proof of possession in the prior session.
                             peer_identity: resuming.peer_identity().clone(),
                             resuming_session: Some(resuming),
                             session_id: server_hello.session_id,
@@ -783,7 +782,6 @@ impl State<ClientConnectionData> for ExpectServerDone {
                     signature,
                 })?
         };
-        cx.common.peer_identity = Some(identity.clone());
 
         // 3.
         if let Some(client_auth) = &st.client_auth {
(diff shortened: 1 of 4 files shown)


rustls/src/client/hs.rs
```diff
@@ -55,9 +53,11 @@ pub(crate) struct ExpectServerHello {
     //
     // If this is `None` then we do not support early data.
     pub(super) early_data_key_schedule: Option<KeyScheduleEarly>,
+    pub(super) ech_state: Option<EchState>,
```
djc: I think I somewhat tried to maintain chronological ordering; as in, when was this value created during the handshake. That seems to me more meaningful (if harder to maintain) than alphabetical order (though consistency is great).

rustls/src/client/tls12.rs
```diff
@@ -174,6 +174,7 @@ mod server_hello {
                         Ok(Box::new(ExpectNewTicket {
                             cert_verified,
                             config,
+                            peer_identity: resuming.peer_identity().clone(),
```
djc, line 176: There seems to be quite a bit of extra cloning in this commit, which feels ugly (even though it does not seem to have an effect on performance). Where does that come from? It doesn't seem obvious to me.

rustls/src/server/tls13.rs
```diff
@@ -1276,6 +1272,7 @@ impl State<ServerConnectionData> for ExpectFinished {
         // Application data may now flow, even if we have client auth enabled.
         cx.common
             .start_traffic(&mut cx.sendable_plaintext);
+        cx.common.peer_identity = self.peer_identity;
```
djc, line 1275: While I think minimizing writes into `CommonState` (particularly as a replacement for maintaining handshake state) is a good idea, IMO delaying this should not be a goal in itself. Rather, I imagine at a lower-level of abstraction we would yeet the peer's identity to the caller as soon as we don't need it for the handshake itself anymore.

ctz: It's not really delaying things as a goal in itself, but instead recognising that the handshake is provisional until the finished message (which demonstrates the integrity of the whole handshake) is confirmed.

djc: Yeah, but the identity can't change after we have established/received it, right?

ctz: True, it cannot change. I think my point is that surfacing it as soon as it and a valid signature by it is received is not great (in TLS1.2 without EMS) because the amount of handshake covered by the signature isn't enough. That means an application that observes the certificate before the finished message is validated has the potential to be misled. It's a pretty marginal point.

rustls/src/client/hs.rs
ctz: Going to put this in a separate PR because it's a pain in the ass to iterate on.

rustls/src/server/tls13.rs
djc: It does feel like a good way to build in misuse-resistance, I suppose. Let's keep it for this PR, anyway.

rustls/src/client/hs.rs
ctz: #2798

rustls/src/client/tls12.rs
ctz: The fundamental here is avoiding using `CommonState` as a storage medium for states, so there's an extra clone (now the state contains its own copy of the identity for issuing or receiving tickets.) It is unfortunate. I think the route I'd like to take there is making `clone()` cheap on `Identity`.

PR #2799, opened 2025-12-10
More cleanups of `CommonState` interactions

ctz:


rustls/src/client/tls13.rs
```diff
@@ -212,6 +212,13 @@ impl ClientHandler<Tls13CipherSuite> for Handler {
 
         emit_fake_ccs(&mut sent_tls13_fake_ccs, cx.common);
 
+        cx.common.handshake_kind = Some(match (&resuming_session, st.done_retry) {
```
ctz, line 215: @djc curious if you prefer this.

rustls/src/server/tls13.rs
```diff
@@ -273,16 +273,17 @@ mod client_hello {
                 emit_fake_ccs(cx.common);
             }
 
-            if full_handshake {
-                cx.common
-                    .handshake_kind
-                    .get_or_insert(HandshakeKind::Full);
+            cx.common.handshake_kind = Some(if full_handshake {
```
ctz: Or this?

rustls/src/client/tls13.rs
djc: Definitely this!

PR #2802, opened 2025-12-12
Rework how buffered plaintext is released

ctz:
Instead of ... *all this* ... just have the buffered connection's `process_new_packets()` try to send anything buffered after each state transition.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -187,7 +187,6 @@ impl CommonState {
         data: &mut Data,
         plaintext_locator: &Locator,
         received_plaintext: &mut Option<UnborrowedPayload>,
-        sendable_plaintext: Option<&mut ChunkVecBuffer>,
     ) -> Result<Box<dyn State<Data>>, Error> {
         // Drop CCS messages during handshake in TLS1.3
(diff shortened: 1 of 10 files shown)


rustls/src/common_state.rs
```diff
@@ -442,30 +440,22 @@ impl CommonState {
...
+        while let Some(buf) = plaintext.pop() {
             self.send_appdata_encrypt(buf.as_slice().into());
         }
     }
 
     /// Mark the connection as ready to send and receive application data.
-    ///
-    /// Also flush `sendable_plaintext` if it is `Some`.
-    pub(crate) fn start_traffic(&mut self, sendable_plaintext: &mut Option<&mut ChunkVecBuffer>) {
+    pub(crate) fn start_traffic(&mut self) {
```
djc, line 450: So this is setting `self.may_receive_application_data` twice, right?

djc (approved):

rustls/src/common_state.rs
ctz: this should set both `may_receive_application_data` and `may_send_application_data` to true.

djc: Ahh, my mistake.

PR #2807, opened 2025-12-15
Tidy record layer

ctz:
This PR simplifies the `RecordLayer` structures. This simplification is achieved by moving the TLS1.2-specific complexity out and keeping it in `client/tls12.rs` and `server/tls12.rs`.

At the end, there is one structure per direction, with the necessary bookkeeping data alongside the current `MessageEncrypter`/`MessageDecrypter`. This arrangement should better support a future duplex-type API like #2795.

### DIFF
--- a/rustls/src/crypto/cipher/record_layer.rs
+++ b/rustls/src/crypto/cipher/record_layer.rs
@@ -238,8 +238,14 @@ pub(crate) enum PreEncryptAction {
     Refuse,
 }
 
-const SEQ_SOFT_LIMIT: u64 = 0xffff_ffff_ffff_0000u64;
-const SEQ_HARD_LIMIT: u64 = 0xffff_ffff_ffff_fffeu64;
+/// When to take action to avoid sequence space exhaustion.
+///
+/// This gives a margin in which any action can have an effect, prior to `SEQ_HARD_LIMIT`
+/// being reached.
+const SEQ_SOFT_LIMIT: u64 = u64::MAX - 0xffff;
+
+/// When to refuse further encryptions.
+const SEQ_HARD_LIMIT: u64 = u64::MAX - 1;
 
 #[cfg(test)]
 mod tests {


ctz:
Hm, I wasn't expecting that sort of performance difference. Will look again at this.

rustls/src/client/tls12.rs
```diff
@@ -187,9 +174,11 @@ mod server_hello {
                             // proof of possession in the prior session.
                             peer_identity: resuming.peer_identity().clone(),
                             resuming_session: Some(resuming),
+                            resuming_encrypter: Some(enc),
```
djc: I guess it's no coincidence that `resuming_session` and `resuming_encrypter` are always `Some`/`None` together. Suggest combining them in a single `Option`?

rustls/src/common_state.rs
```diff
@@ -30,7 +31,8 @@ pub struct CommonState {
     pub(crate) negotiated_version: Option<ProtocolVersion>,
     pub(crate) handshake_kind: Option<HandshakeKind>,
     pub(crate) side: Side,
-    pub(crate) record_layer: RecordLayer,
+    pub(crate) record_layer_in: RecordLayerIn,
```
djc: I like the direction but not sure about these names. Maybe `InputRecords`/`OutputRecords`, or `EncryptionState`/`DecryptionState`? I like the latter more, not sure how important "record layer" as a term of art is here.

ctz: On one hand "record layer" is a term of art in the standards, but on the other it doesn't match 100% with the responsibilities of these objects.

ctz: (did this rename, but haven't renamed the containing file and plan to extend the responsibilities of these objects a bit)

djc:
Did you find anything to explain the perf difference?

ctz:
> Did you find anything to explain the perf difference?

https://github.com/rustls/rustls/compare/ef58b0cef992747e17ca59eb193d446c33a6d0c0..2dc379f6e0e1fb4852d42e00888e8725f18bd96b addressed the earlier large change in server resumption performance. Specifically the second hunk means this branch https://github.com/rustls/rustls/blob/76228b82657f1d5f3f7a7b8d01a7099a516919a1/rustls/src/server/tls12.rs#L825-L846 is not always taken.

PR #2808, opened 2025-12-17
Handshake state tidying

ctz:
Sloughing off further smaller commits from my current work on making the protocol state machine independent of  `CommonState`, and thereby splitting that up into smaller chunks.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -176,7 +176,7 @@ impl CommonState {
         self.handshake_kind
     }
 
-    pub(crate) fn is_tls13(&self) -> bool {
+    fn is_tls13(&self) -> bool {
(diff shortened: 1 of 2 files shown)


rustls/src/common_state.rs
```diff
@@ -353,14 +353,14 @@ impl CommonState {
     /// Like send_msg_encrypt, but operate on an appdata directly.
     fn send_appdata_encrypt(&mut self, payload: OutboundPlain<'_>) -> usize {
         let len = payload.len();
-        let iter = self
+        for m in self
```
djc: I kind of prefer the previous style of binding the iter, because I find multi-line loop headers harder to read.

djc:
(Needs a rebase.)

rustls/src/common_state.rs
ctz: Reinstated that part of this commit.

PR #2809, opened 2025-12-18
feat: add `with_capacity()` fn to `RootCertStore`

abd0-omar:


ctz:
Hi, because `RootCertStore::anchors` is pub, there not a reason to reflect the whole of `Vec`'s public API in this type?

One thing we could do instead of this is withdraw the `empty()` constructor, encouraging callers to consider a better way of initialising this.

abd0-omar:
> Hi, because `RootCertStore::anchors` is pub, there not a reason to reflect the whole of `Vec`'s public API in this type?
> 
> One thing we could do instead of this is withdraw the `empty()` constructor, encouraging callers to consider a better way of initialising this.

Oh, I missed that `roots` is public, it is indeed the changes are redundant.

regarding `empty()` it's convenient to use but if you'd prefer to use manual initialization, I'm happy to update the PR to remove `empty()` and refactor where it's used

ctz:
I think it would be interesting to see how that refactoring ends up looking. I don't think the `empty()` constructor is very useful as a public API, given it creates an object that is not useful without further work. Maybe we could have a more useful constructor that takes an `impl IntoIterator<Item = CertificateDer<'a>>` and calls `add()` for each one?

abd0-omar:
Yes, I like this direction. I just initially hesitated because I didn't want to implement `FromIterator` for fallible collection

I'll update the PR to remove the `empty()` constructor and refactor internal call sites with the new `try_from_iter` constructor

Does `try_from_iter` sounds like the right name or would you prefer something else like new?

PR #2815, opened 2025-12-19
Take next versions of hpke-rs cinematic universe

ctz:
renovate wasn't managing this, because they need to be done in lock-step.


djc:
Renovate does have some features that could support this kind of thing:

https://github.com/rust-lang/rustup/blob/main/.github/renovate.json#L44

PR #2816, opened 2025-12-19
Rework dependencies from `tls13::key_schedule`

ctz:
This is further work toward restricting the dependency of things on `CommonState`. This makes everything in `tls13::key_schedule` have a write-only dependency on `CommonState` (ie, it never reads fields). This allows a future change to replace this `CommonState` with a more general event handler trait.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -468,7 +468,7 @@ impl CommonState {
     /// Send a raw TLS message, fragmenting it if needed.
     pub(crate) fn send_msg(&mut self, m: Message<'_>, must_encrypt: bool) {
         {
-            if let Protocol::Quic = self.protocol {
+            if self.protocol.is_quic() {
                 if let MessagePayload::Alert(_) = m.payload {
                     // alerts are sent out-of-band in QUIC mode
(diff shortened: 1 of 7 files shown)


rustls/src/common_state.rs
```diff
@@ -881,8 +877,17 @@ impl Side {
...
+    /// TCP-TLS, standardized in RFC5246 and RFC8446
     Tcp,
-    Quic,
+    /// QUIC, standardized in RFC9001
+    #[cfg_attr(not(feature = "std"), expect(dead_code))]
+    Quic(quic::Version),
+}
+
+impl Protocol {
+    pub(crate) fn quic(&self) -> bool {
```
djc: Nit: I prefer `is_quic()` as the method name.

rustls/src/common_state.rs
```diff
@@ -651,10 +651,6 @@ impl CommonState {
         }
     }
 
-    pub(crate) fn is_quic(&self) -> bool {
-        self.protocol == Protocol::Quic
-    }
```
djc, line 656: Might be nice to make this a separate commit.

PR #2818, opened 2025-12-22
Toward making low-level state machine functionally pure (omnibus PR)

ctz:
Previously: #2832, #2833, #2834, #2854, #2855, #2897, #2917, #2924

These are the final reorganisations to finish this work. Later on I will likely replace the `Output` trait and `Event` enum with a single dyn-dispatch trait. It will have more boilerplate, but should be quicker.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -144,7 +144,6 @@ pub struct ConnectionOutputs {
     pub(crate) exporter: Option<Box<dyn Exporter>>,
     pub(crate) early_exporter: Option<Box<dyn Exporter>>,
     pub(crate) fips: FipsStatus,
-    pub(crate) tls13_tickets_received: u32,
 }
 
 impl ConnectionOutputs {
@@ -206,13 +205,6 @@ impl ConnectionOutputs {
         self.handshake_kind
     }
 
-    /// Returns the number of TLS1.3 tickets that have been received.
-    ///
-    /// Only clients receive tickets, so this is zero for servers.
-    pub fn tls13_tickets_received(&self) -> u32 {
-        self.tls13_tickets_received
-    }
-
     /// Return the FIPS validation status of the connection.
     ///
     /// This is different from [`crate::crypto::CryptoProvider::fips()`]:
@@ -257,11 +249,6 @@ impl Output for ConnectionOutputs {
             Event::ProtocolVersion(ver) => {
                 self.negotiated_version = Some(ver);
             }
-            Event::ReceivedTicket => {
-                self.tls13_tickets_received = self
-                    .tls13_tickets_received
-                    .saturating_add(1)
-            }
             _ => unreachable!(),
         }
     }
@@ -279,7 +266,6 @@ impl Default for ConnectionOutputs {
             exporter: None,
             early_exporter: None,
             fips: FipsStatus::Unvalidated,
-            tls13_tickets_received: 0,
         }
     }
 }
@@ -696,6 +682,8 @@ pub(crate) struct ReceivePath {
     /// We limit consecutive empty fragments to avoid a route for the peer to send
     /// us significant but fruitless traffic.
     seen_consecutive_empty_fragments: u8,
+
+    pub(crate) tls13_tickets_received: u32,
 }
 
 impl ReceivePath {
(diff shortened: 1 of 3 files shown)


rustls/src/common_state.rs
```diff
@@ -300,14 +300,11 @@ pub(crate) struct SendPath {
 }
 
 impl SendPath {
-    pub(crate) fn maybe_send_fatal_alert(&mut self, error: &Error) {
+    pub(crate) fn maybe_send_fatal_alert(output: &mut dyn Output, error: &Error) {
```
djc: If it doesn't take `self`, detach it from `SendPath`?

rustls/src/conn/mod.rs
```diff
@@ -752,19 +732,13 @@ impl IoState {
     }
 }
 
-pub(crate) struct ConnectionCore<Side: SideData> {
+pub(crate) struct ConnectionProcessor<'a> {
```
djc: This seems fairly unprincipled. What happens to this next?

rustls/src/common_state.rs
```diff
@@ -1133,6 +1133,63 @@ impl Output for CaptureAppData<'_> {
...
+            EventDisposition::StartTraffic => {
+                self.recv.emit(Event::StartTraffic);
+                self.other.emit(Event::StartTraffic);
+            }
+            _ => self.other.emit(ev),
+        }
+    }
+}
+
+pub(crate) struct OutputSplitNonReceivePath<'a> {
```
djc: This also seems pretty crappy...

rustls/src/common_state.rs
ctz: Will make this a free function.

rustls/src/conn/mod.rs
ctz: This is so this can be used both during the handshake, and then during the traffic stage (where the receive and send are split.) These eventually look something like:

(handshake)
```rust
        let mut p = ConnectionProcessor::<Side> {
            state: mem::replace(&mut self.state, Err(Error::HandshakeNotComplete)),
            recv: &mut self.common.recv,
            other: &mut OutputSplitNonReceivePath {
                outputs: &mut self.common.outputs,
                protocol: self.common.protocol,
                quic: &mut self.common.quic,
                send: &mut self.common.send,
                side: &mut self.side,
            },
        };
        let rc = p.process_new_packets(input, finish);
```

and 

(traffic read-side)
```rust
    pub fn read<'a>(
        mut self,
        received_tls: &'a mut impl TlsInputBuffer,
    ) -> Result<ReceiveTrafficState<'a, Side>, ErrorWithAlert> {
        let mut send = SendAdapter::Unlocked(&self.send);
        let mut proc = ConnectionProcessor::<Side> {
            state: Ok(self.state),
            recv: &mut self.recv,
            other: &mut send,
        };
        let received_plain =
            match proc.process_new_packets(received_tls, ProcessFinishCondition::AppData) {
(...)
```

djc: Well, these three things don't seem like they are good types:

- `ConnectionProcessor` -- seems like it should be free function with arguments, it doesn't feel like a meaningful bundle of state.
- `OutputSplitNonReceivePath` -- that's not a name, but just a bunch of syllables strung together... What does it mean, how is it used?
- Same for `OutputSplitReceivePath`... is it time to split/replace the `Output` god trait yet?

ctz: `ConnectionProcessor`: now a function

simplified the names of the two other types.

djc:
Also 2+% regressions on no crypto benchmarks might be significant?

ctz:
> Also 2+% regressions on no crypto benchmarks might be significant?

These went away with LTO.

rustls/src/conn/mod.rs
```diff
@@ -752,113 +731,143 @@ impl IoState {
...
-}
-
-impl<Side: SideData> ConnectionCore<Side> {
-    pub(crate) fn new(state: Box<dyn State>, side: Side) -> Self {
-        Self {
-            state: Ok(state),
-            side,
+pub(crate) fn process_new_packets(
+    input: &mut dyn TlsInputBuffer,
+    state_err: &mut Result<Box<dyn State>, Error>,
```
djc: Nit: I don't think the `_err` suffix is very clear here -- it's not always an `Err`, right (in fact it's usually not?)? I'd probably just call it `state` -- the type makes it clear that it's wrapped.

PR #2832, opened 2026-01-09
Toward making low-level state machine functionally pure (pt 1)

ctz:
The eventual goal here is to bring into control the amount of dependency the low-level handshake state machines have on the rest of the crate. At the moment the `Context` parameter gives direct mutable access to a shed load of state.

At the end of this set of PRs, each state no longer has access to `Context`, instead it can merely surface events (along the lines of "here are some new encryption keys", or "here is the peer's certificate chain") to the next layer up. This ends up with:

```patch
 pub(crate) trait State<Side>: Send + Sync {
     fn handle<'m>(
         self: Box<Self>,
-        cx: &mut Context<'_, Side>,
         input: Input<'m>,
+        output: &mut dyn Output,
     ) -> Result<Box<dyn State<Side>>, Error>;
```

(Much later the plan is still to eliminate this trait, but that is several steps away from here.)

Omnibus PR: #2818

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -446,8 +446,7 @@ impl CommonState {
         self.start_outgoing_traffic();
     }
 
-    /// Mark the connection as ready to send application data.
-    pub(crate) fn start_outgoing_traffic(&mut self) {
+    fn start_outgoing_traffic(&mut self) {
         self.may_send_application_data = true;
         debug_assert!(self.encrypt_state.is_encrypting());
     }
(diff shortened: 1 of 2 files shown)


ctz:
> How did you arrive at the design of a single method with an enum argument, rather than a trait with one method per event kind?

I have two things in mind with that:

- always ensuring the data flow is one-way (ie, it's much less reasonable for a return value or getter to creep into the `Output` trait later)
- reducing boilerplate when directing events to a class with a specific interest. Like, much later on quic.rs ends up with:

```patch
+impl Output for Quic {
+    fn emit(&mut self, ev: Event<'_>) {
+        match ev {
+            Event::EncryptMessage(m) => self.send_msg(m, true),
+            Event::QuicEarlySecret(sec) => self.early_secret = sec,
+            Event::QuicHandshakeSecrets(sec) => self.hs_secrets = Some(sec),
+            Event::QuicTrafficSecrets(sec) => self.traffic_secrets = Some(sec),
+            Event::QuicTransportParameters(params) => self.params = Some(params),
+            Event::PlainMessage(m) => self.send_msg(m, false),
+            _ => {}
+        }
+    }
+}
+
```

djc:
Fair enough!

PR #2833, opened 2026-01-10
Toward making low-level state machine functionally pure (pt 2)

ctz:
Previously #2832 

Omnibus PR: #2818

### DIFF
--- a/rustls/src/tls13/key_schedule.rs
+++ b/rustls/src/tls13/key_schedule.rs
@@ -3,7 +3,7 @@
 use alloc::boxed::Box;
 use core::ops::Deref;
 
-use crate::common_state::{CommonState, Event, Output, Protocol, Side};
+use crate::common_state::{Event, Output, Protocol, Side};
 use crate::conn::Exporter;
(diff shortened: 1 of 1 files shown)


rustls/src/common_state.rs
```diff
@@ -844,13 +867,25 @@ pub(crate) trait Output {
...
+    ApplicationProtocol(ProtocolName),
+    CipherSuite(SupportedCipherSuite),
     EarlyExporter(Box<dyn Exporter>),
     EncryptMessage(Message<'a>),
+    EnqueueKeyUpdateNotification,
     Exporter(Box<dyn Exporter>),
     HandshakeKind(HandshakeKind),
     KeyExchangeGroup(&'static dyn SupportedKxGroup),
+    MessageDecrypter(Box<dyn MessageDecrypter>, HandshakeAlignedProof),
+    MessageDecrypterWithTrialDecryption(Box<dyn MessageDecrypter>, usize, HandshakeAlignedProof),
```
djc: Maybe named fields here, with the usize field? (Could do it for all 2+ or 3+ field variants maybe.)

PR #2834, opened 2026-01-10
Toward making low-level state machine functionally pure (pt 3) 

ctz:
Previously #2832, #2833 

Omnibus PR: #2818

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -1081,7 +1081,7 @@ impl<'a, const TLS13: bool> HandshakeFlight<'a, TLS13> {
             .add(&self.body[start_len..]);
     }
 
-    pub(crate) fn finish(self, common: &mut CommonState) {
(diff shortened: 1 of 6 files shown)


rustls/src/common_state.rs
```diff
@@ -179,6 +179,16 @@ impl CommonState {
         matches!(self.negotiated_version, Some(ProtocolVersion::TLSv1_3))
     }
 
+    pub(super) fn into_kernel_parts(
+        self,
+    ) -> Option<(ProtocolVersion, Protocol, SupportedCipherSuite, quic::Quic)> {
+        if let (Some(negotiated_version), Some(suite)) = (self.negotiated_version, self.suite) {
```
djc: Tiny nit: I think this would read better with a `match`? Maybe also destructuring.

djc (approved):

PR #2838, opened 2026-01-13
crypto: clean up (hybrid) key exchange API

djc:


rustls/src/crypto/kx/mod.rs
```diff
@@ -6,11 +6,203 @@ use core::ops::Deref;
...
 use crate::enums::ProtocolVersion;
-use crate::error::Error;
+use crate::error::{Error, PeerMisbehaved};
 
 pub mod ffdhe;
 use ffdhe::FfdheGroup;
 
+/// A generalization of hybrid key exchange.
+#[derive(Debug)]
+pub(crate) struct Hybrid {
```
ctz: would eventually like to make this `pub` (not a comment on this PR)

djc: Made it public to prevent warnings when aws-lc-rs is not enabled.

djc:
I've left `Hybrid` and `HybridLayout` exhaustive -- I think that's what makes sense for these? Wanted to call it out though.

PR #2841, opened 2026-01-14
Externalize aws-lc crypto provider

djc:


rustls/Cargo.toml
```diff
@@ -15,17 +15,14 @@ exclude = ["src/testdata", "tests/**"]
 
 [features]
 default = ["log", "std"]
-
-aws-lc-rs = ["dep:aws-lc-rs", "webpki/aws-lc-rs", "aws-lc-rs/aws-lc-sys", "aws-lc-rs/prebuilt-nasm"]
```
ctz, line 18: note `[package.metadata.docs.rs]` section below

djc: Fixed!

djc:
Gave up on the 2 examples in the builder docs for now, and switched them to `ignore`. According to the rustdoc team on Zulip, there is no way to make this work without actually making the `TEST_PROVIDER` part of the public API, which IMO isn't worth it.

rustls/Cargo.toml
```diff
@@ -15,15 +15,13 @@ exclude = ["src/testdata", "tests/**"]
 
 [features]
 default = ["log", "std"]
-aws-lc-rs = ["dep:aws-lc-rs", "webpki/aws-lc-rs", "aws-lc-rs/aws-lc-sys", "aws-lc-rs/prebuilt-nasm"]
 brotli = ["dep:brotli", "dep:brotli-decompressor", "std"]
-fips = ["aws-lc-rs", "aws-lc-rs?/fips", "webpki/aws-lc-rs-fips"]
+fips = []
```
ctz, line 19: I think maybe we should redesign things so this goes away?

ctz (approved):

rustls/Cargo.toml
djc: Definitely! But not on this PR?

ctz: Yeah for sure not in this PR.

PR #2844, opened 2026-01-15
Enable all features by default in rustls-test crate

djc:
Because I can't enable all features across the workspace (due to conflicts between aws-lc-rs fips and unstable), I set `rust-analyzer.cargo.features` to `[]` to use default features. However, this disabled all diagnostics for integration tests in rustls-test, which are gated on provider features.


ctz:
> ####

Hm, I think there's some unintended feature unification happening now, probably ci-bench needs to change or opt-out of default features?

PR #2845, opened 2026-01-15
ci: bring back integration tests for cross-compiled targets

djc:
In

- https://github.com/rustls/rustls/pull/2728

I moved the integration tests out of the rustls crate and into rustls-test. However, the cross-compiled target workflow in CI explicitly uses `-p rustls`, so it was not testing as much as before. Restore the previous behavior.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -420,6 +420,7 @@ jobs:
         if: ${{ matrix.target != 'i686-unknown-linux-gnu' }}
         run: cargo add --dev --features bindgen 'aws-lc-sys@>0.20' --package rustls --verbose
       - run: cross test --package rustls --target ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -420,6 +420,7 @@ jobs:
         if: ${{ matrix.target != 'i686-unknown-linux-gnu' }}
         run: cargo add --dev --features bindgen 'aws-lc-sys@>0.20' --package rustls --verbose
       - run: cross test --package rustls --target ${{ matrix.target }}
+      - run: cross test --package rustls-test --features aws-lc-rs --target ${{ matrix.target }}
```
ctz, line 423: perhaps a `--features ring` too? though maybe not, since the cross runner under qemu is dog-slow.

djc: Plus we just saw that ring no longer works on i586 at least. I would say rustls is the software under test here, and we don't need to run multiple providers.

PR #2846, opened 2026-01-15
Drop the fips feature

djc:


ctz (approved):

rustls/src/server/config.rs
```diff
@@ -683,7 +684,7 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
             max_early_data_size: 0,
             send_half_rtt_data: false,
             send_tls13_tickets: 2,
-            require_ems: cfg!(feature = "fips"),
+            require_ems,
```
ctz, line 687: See some docs:

```
    /// The default is `true` if the `fips` crate feature is enabled,
    /// `false` otherwise.
```

Some other instances of docs on this feature.

Issue #2252, opened 2024-12-01
Support prioritizing ChaCha20 for clients without AES hardware acceleration


james58899:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
If `ignore_client_order` is enabled on the server, there will be a performance impact on clients without AES hardware acceleration.

**Describe the solution you'd like**
I hope rustls can support similar functionality to OpenSSL's `PrioritizeChaCha`.
When ChaCha20-Poly1305 is first in the cipher list provided by the client, ChaCha20-Poly1305 is preferred instead, because these clients may not support AES hardware acceleration.

**Describe alternatives you've considered**
Provides a hook for selecting cipher during the handshake phase for manual selection.

**Additional context**
While I could `disable ignore_client_order` to respect client priority, AES128 may be faster than AES256 for file transfers, and there are many clients that prefer AES256.
I hope that AES128 will be used first for clients with AES acceleration, and ChaCha20 will be used first for clients without AES acceleration.

ctz:
I think this is something I'd consider. Though I probably wouldn't do it either by a special case for chacha, or a  callback. I think probably a two-tier client priority/server priority list, or maybe [copying golang's homework](https://go.dev/blog/tls-cipher-suites).

djc:
> I think this is something I'd consider. Though I probably wouldn't do it either by a special case for chacha, or a  callback. I think probably a two-tier client priority/server priority list, or maybe [copying golang's homework](https://go.dev/blog/tls-cipher-suites).

The Go approach sounds really good to me.

djc:
> > AES128 may be faster than AES256 for file transfers, and there are many clients that prefer AES256
> 
> 
> 
> Users that need compliance with CNSA-2.0, have to use AES256, regardless of the (fairly small) performance benefits that AES128 could offer.

As I understand it, this issue is only about ordering, not about enabling/disabling cipher suites.

crazyscot:
I'm very keen to see this. I have an old operational machine (~2008) which lacks on-chip support for AES; happy to help test, or perhaps even contribute though I don't have a great deal of bandwidth.

[...]

PR #2849, opened 2026-01-16
Prefer AES-128-GCM, matching Go

djc:
Reviewing https://go.dev/blog/tls-cipher-suites because of

- #2252 

this seems like an easy win. Might be nice to backport this to 0.23 as well?

### DIFF
--- a/rustls-aws-lc-rs/src/lib.rs
+++ b/rustls-aws-lc-rs/src/lib.rs
@@ -194,12 +194,12 @@ const SIX_HOURS: Duration = Duration::from_secs(6 * 60 * 60);
 /// This will be [`ALL_TLS12_CIPHER_SUITES`] sans any supported cipher suites that
 /// shouldn't be enabled by most applications.
 pub static DEFAULT_TLS12_CIPHER_SUITES: &[&Tls12CipherSuite] = &[
-    tls12::TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
     tls12::TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+    tls12::TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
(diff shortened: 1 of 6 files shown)


ctz:
Note https://docs.rs/rustls/latest/rustls/manual/_05_defaults/index.html#why-is-aes-256-preferred-over-aes-128 (happy for you to just delete this part of the manual in this PR)

djc:
> Note https://docs.rs/rustls/latest/rustls/manual/_05_defaults/index.html#why-is-aes-256-preferred-over-aes-128 (happy for you to just delete this part of the manual in this PR)

Done.

bogo's `EarlyData-WriteAfterEncryptedExtensions` seems to pass locally for me both as a single test and in the entire test suite, and I have no clue why this makes the i586 compilation fail again for ring.

stormshield-gt:
In the go blog post, they said TLS has no post-quantum key exchange (in 2021), but that's no longer true nowadays. Given that fact, do you think that the recommendation for AES-128 still holds ? 

> The larger key is only useful in multi-user and post-quantum settings, which are not relevant to TLS, which generates sufficiently random IVs and has no post-quantum key exchange support. Since the larger key doesn’t have any benefit, we prefer AES-128 for its speed.

ctz:
I don't believe grover's algorithm is expected to provide any practical benefit on attacking either AES-128 or AES-256. ref: https://csrc.nist.gov/Projects/Post-Quantum-Cryptography/faqs#question_LVQL

djc:
Also Go still ships with AES-128-GCM as the default:

https://cs.opensource.google/go/go/+/master:src/crypto/tls/cipher_suites.go;l=272

stormshield-gt:
Thanks for the clarification !

stormshield-gt:
For the curious, here is a more detailed explanation of why 128 bits is enough by the same author of the blog post: https://words.filippo.io/post-quantum-age/#128-bits-are-enough

PR #2852, opened 2026-01-17
bogo: adjust server message size estimates

ctz:
Attempting to address https://github.com/rustls/rustls/actions/runs/21082234271/job/60638317584


djc (approved):

djc:
Can we make this more principled, like using the hash output len of our first cipher suite or something?

ctz:
> Can we make this more principled, like using the hash output len of our first cipher suite or something?

Going to merge this as-is to fix CI, but my preference would be to delete the length predictions and have these tests feed in TLS messages one-at-a-time.

PR #2854, opened 2026-01-18
Toward making low-level state machine functionally pure (pt 4)

ctz:
Previously #2832, #2833, #2834

Omnibus PR: #2818

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -166,7 +166,7 @@ mod connection {
             let inner =
                 ConnectionCore::for_client(config, name, exts, Protocol::Quic(quic_version))?;
             Ok(Self {
-                inner: inner.into(),
+                inner: ConnectionCommon::new(inner, quic_version),
             })
         }
 
@@ -267,7 +267,8 @@ mod connection {
             };
 
             let core = ConnectionCore::for_server(config, exts, Protocol::Quic(quic_version))?;
-            Ok(Self { inner: core.into() })
+            let inner = ConnectionCommon::new(core, quic_version);
+            Ok(Self { inner })
         }
 
         /// Explicitly discard early data, notifying the client
@@ -372,9 +373,19 @@ mod connection {
         core: ConnectionCore<Side>,
         deframer_buffer: DeframerVecBuffer,
         sendable_plaintext: ChunkVecBuffer,
+        version: Version,
     }
 
     impl<Side: SideData> ConnectionCommon<Side> {
+        fn new(core: ConnectionCore<Side>, version: Version) -> Self {
+            Self {
+                core,
+                deframer_buffer: DeframerVecBuffer::default(),
+                sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
+                version,
+            }
+        }
+
         /// Return the TLS-encoded transport parameters for the session's peer.
         ///
         /// While the transport parameters are technically available prior to the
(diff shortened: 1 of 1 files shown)


rustls/src/msgs/persist.rs
```diff
@@ -231,6 +235,13 @@ impl ClientSessionCommon {
     }
 }
 
+/// A "template" for future TLS1.3 client session values.
+pub(crate) struct Tls13ClientSessionPrecursor {
```
djc: Don't love the name (maybe `Tls13ClientSessionInput`?), and I think it would be nice to have a comment that better explains why this exists/why it's useful to group/thread these through together.

rustls/src/msgs/persist.rs
```diff
@@ -78,22 +78,30 @@ pub struct Tls13ClientSessionValue {
 
 impl Tls13ClientSessionValue {
     pub(crate) fn new(
-        suite: &'static Tls13CipherSuite,
+        precursor: &Tls13ClientSessionPrecursor,
```
djc: Feels like it should get ownership of the precursor?

rustls/src/conn/kernel.rs
```diff
@@ -74,26 +72,21 @@ use crate::{CommonState, ConnectionTrafficSecrets, Error, SupportedCipherSuite};
...
-    protocol: Protocol,
     suite: SupportedCipherSuite,
 
     _side: PhantomData<Side>,
 }
 
 impl<Side> KernelConnection<Side> {
     pub(crate) fn new(state: Box<dyn KernelState>, common: CommonState) -> Result<Self, Error> {
-        let (negotiated_version, protocol, suite, quic) = common
+        let (negotiated_version, suite) = common
```
djc, line 83: The interaction with the kTLS API is a little unclear to me. Maybe the commit message could explain why/how this simplifies it?

rustls/src/server/hs.rs
```diff
@@ -94,15 +94,15 @@ impl<'a> ExtensionProcessing<'a> {
 
                 self.extensions.selected_protocol =
                     Some(SingleProtocolName::new(selected_protocol.clone()));
-                cx.emit(Event::ApplicationProtocol(selected_protocol));
-                true
+                cx.emit(Event::ApplicationProtocol(selected_protocol.clone()));
+                Some(selected_protocol)
```
djc: So here we're passing it on in 3 different ways:

- Store in `self.extensions.selected_protocol`
- Emit an event into the `Context`
- Return to the caller

Feels like overkill?

rustls/src/common_state.rs
```diff
@@ -639,24 +640,17 @@ impl CommonState {
...
 
-    pub(crate) fn should_update_key(
-        &mut self,
-        key_update_request: &KeyUpdateRequest,
-    ) -> Result<bool, Error> {
-        match key_update_request {
-            KeyUpdateRequest::UpdateNotRequested => Ok(false),
-            KeyUpdateRequest::UpdateRequested => Ok(self.queued_key_update_message.is_none()),
-            _ => Err(InvalidMessage::InvalidKeyUpdate.into()),
+    pub(crate) fn maybe_enqueue_key_update_notification(&mut self) -> bool {
```
djc: A little verbose for my taste. Maybe just `enqueue_key_update()` or `ensure_key_update_queued()`?

rustls/src/server/hs.rs
ctz: Separated the "choice" and "action" part. But I think all these things need to be done.

PR #2855, opened 2026-01-18
Toward making low-level state machine functionally pure (pt 5)

ctz:
Previously #2832, #2833, #2834, #2854

Omnibus PR: #2818

### DIFF
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -5,9 +5,9 @@ use core::{fmt, mem};
 use pki_types::ServerName;
 
 use super::config::ClientConfig;
-use super::hs::{self, ClientHelloInput};
+use super::hs::ClientHelloInput;
 use crate::client::EchStatus;
-use crate::common_state::{CommonState, EarlyDataEvent, Event, Output, Protocol, Side};
+use crate::common_state::{CommonState, Context, EarlyDataEvent, Event, Output, Protocol, Side};
 use crate::conn::{ConnectionCore, UnbufferedConnectionCommon};
 #[cfg(doc)]
 use crate::crypto;
@@ -274,7 +274,7 @@ impl ConnectionCore<ClientConnectionData> {
         common_state.fips = config.fips();
         let mut data = ClientConnectionData::new(common_state);
 
-        let mut cx = hs::ClientContext {
+        let mut cx = Context {
             data: &mut data,
             // `start_handshake` won't read plaintext
             plaintext_locator: &Locator::new(&[]),
--- a/rustls/src/client/tls13.rs
+++ b/rustls/src/client/tls13.rs
@@ -8,8 +8,7 @@ use super::config::{ClientConfig, ClientSessionKey, ClientSessionStore};
 use super::connection::ClientConnectionData;
 use super::ech::EchStatus;
 use super::hs::{
-    self, ClientContext, ClientHandler, ClientHelloInput, ClientSessionValue, ExpectServerHello,
-    GroupAndKeyShare,
+    self, ClientHandler, ClientHelloInput, ClientSessionValue, ExpectServerHello, GroupAndKeyShare,
 };
 use super::{ClientAuthDetails, ClientHelloDetails, ServerCertDetails};
 use crate::check::inappropriate_handshake_message;
@@ -64,7 +63,7 @@ impl ClientHandler<Tls13CipherSuite> for Handler {
         server_hello: &ServerHelloPayload,
         input: &Input<'_>,
         mut st: ExpectServerHello,
-        cx: &mut ClientContext<'_>,
+        output: &mut dyn Output,
     ) -> hs::NextStateOrError {
         // Start our handshake hash, and input the server-hello.
         let mut transcript = st
@@ -103,7 +102,7 @@ impl ClientHandler<Tls13CipherSuite> for Handler {
 
         // We always send a key share when TLS 1.3 is enabled.
         let our_key_share = st.offered_key_share.unwrap();
-        let our_key_share = KeyExchangeChoice::new(&config, cx, our_key_share, their_key_share)
+        let our_key_share = KeyExchangeChoice::new(&config, output, our_key_share, their_key_share)
             .map_err(|_| PeerMisbehaved::WrongGroupForKeyShare)?;
 
         let (key_schedule_pre_handshake, in_early_traffic) =
@@ -146,7 +145,7 @@ impl ClientHandler<Tls13CipherSuite> for Handler {
                 _ => {
                     debug!("Not resuming");
                     // Discard the early data key schedule.
-                    cx.emit(Event::EarlyData(EarlyDataEvent::Rejected));
+                    output.emit(Event::EarlyData(EarlyDataEvent::Rejected));
                     resuming_session.take();
                     (
                         KeySchedulePreHandshake::new(Side::Client, protocol, suite),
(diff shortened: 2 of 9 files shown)


rustls/src/client/connection.rs
```diff
@@ -576,4 +576,8 @@ impl ClientConnectionData {
     }
 }
 
+impl Output for ClientConnectionData {
```
djc, line 582: Hmm, really not a fan of these `Output` impls for the `SideData` types... Isn't there some other way?

Maybe we could wrap `&mut ClientConnectionData` and `&mut CommonStatus` in a type that can act as `dyn Output`?

Feels to me like there should be a single `Output` impl (per side).

rustls/src/quic.rs
```diff
@@ -272,9 +272,7 @@ mod connection {
         /// Explicitly discard early data, notifying the client
         ///
         /// Useful if invariants encoded in `received_resumption_data()` cannot be respected.
-        ///
-        /// Must be called while `is_handshaking` is true.
-        pub fn reject_early_data(&mut self) {
+        pub fn reject_early_data(&mut self) -> Result<(), Error> {
```
djc: Maybe better to just remove this API for now?

rustls/src/vecbuf.rs
```diff
@@ -125,6 +113,18 @@ impl ChunkVecBuffer {
         take
     }
 
+    /// For a proposed append of `len` bytes, how many
+    /// bytes should we actually append to adhere to the
+    /// currently set `limit`?
+    pub(crate) fn apply_limit(&self, len: usize) -> usize {
+        if let Some(limit) = self.limit {
+            let space = limit.saturating_sub(self.len());
+            core::cmp::min(len, space)
```
djc: Suggest `Ord::min()` instead, which doesn't need an import.

While you're move this, `let` `else`? Might be nice to make it its own commit.

rustls/src/server/tls13.rs
```diff
@@ -386,7 +384,12 @@ mod client_hello {
                         expected_certificate_type: cert_types.client,
                     }))
                 }
-            } else if doing_early_data == EarlyDataDecision::Accepted && !st.protocol.is_quic() {
+            } else if matches!(doing_early_data, EarlyDataDecision::Accepted { .. })
+                && !st.protocol.is_quic()
+            {
+                let EarlyDataDecision::Accepted { max_length } = doing_early_data else {
+                    unreachable!(); // FIXME: `else if let`
```
djc: `FIXME`?

rustls/src/common_state.rs
```diff
@@ -745,6 +745,11 @@ pub(crate) trait State<Side>: Send + Sync {
         Err(ApiMisuse::EarlyDataRejectedAtWrongTime.into())
     }
 
+    #[allow(dead_code)]
```
djc: Why does this remain dead?

rustls/src/client/tls13.rs
```diff
@@ -510,6 +513,11 @@ impl State<ClientConnectionData> for ExpectEncryptedExtensions {
             _ => None,
         };
 
+        let ech = Ech {
```
djc, line 533: I'd feel better if this were two commits:

- Wrap `ech_retry_configs` in `Ech`
- Move `ech_status` into it

rustls/src/client/connection.rs
djc: Thought about this some more, maybe `CommonState` wants to live inside `ClientConnectionData`/`ServerConnectionData`?

rustls/src/common_state.rs
ctz: At this point, the only no-std API is the unbuffered one and this does not expose either API. Clarified with `cfg_attr` and `expect`

rustls/src/quic.rs
ctz: I can't actually see any callers of this on github either...

djc: Maybe everyone who tried it found that it didn't work anyway? 😛

ctz: -> #2861

rustls/src/server/tls13.rs
ctz: removed the fixme, though obvously still `else if let` would be preferable

rustls/src/client/connection.rs
ctz: I have tried that in a trailing commit. Seems tidier, though at the expense of another layer of deref.

rustls/src/common_state.rs
```diff
@@ -657,6 +588,149 @@ impl CommonState {
...
+}
+
+impl fmt::Debug for CommonState {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("CommonState")
+            .finish_non_exhaustive()
+    }
+}
+
+pub(crate) fn process_main_protocol<Data: SideData>(
```
djc: This feels like it deserves a separate commit...

rustls/src/client/connection.rs
```diff
@@ -578,9 +579,28 @@ impl Output for ClientConnectionData {
...
+    }
+}
+
+impl Deref for ClientConnectionData {
+    type Target = CommonState;
+
+    fn deref(&self) -> &Self::Target {
+        &self.common
+    }
+}
```
djc, line 602: Hmm, I guess this is publicly visible? Maybe avoid `Deref` for this and use something like `SideData::common()`?

ctz: Note that, at this point, `CommonState` is a public type as is `ClientConnectionData`, but neither can be constructed outside the crate, and no values of `ClientConnectionData` are exposed.

(In a few commits time, this deref is moved into a private trait.)

rustls/src/common_state.rs
ctz: Extracted to a prior commit.

rustls/src/common_state.rs
```diff
@@ -656,6 +586,78 @@ impl CommonState {
     }
 }
 
+pub(crate) fn process_main_protocol<Data>(
```
djc: Given the dependency on `process_main_protocol`, this should go above it. Also suggest changing the argument order to have `common` last.

rustls/src/conn/mod.rs
```diff
@@ -1233,7 +1233,8 @@ impl<Side: SideData> ConnectionCore<Side> {
 }
 
 /// Data specific to the peer's side (client or server).
-pub trait SideData: Debug + Deref<Target = CommonState> + DerefMut {
+#[expect(private_bounds)]
+pub trait SideData: Output + Debug + Deref<Target = CommonState> + DerefMut {
```
djc: Should we add `Sealed` here?

rustls/src/client/connection.rs
```diff
@@ -585,6 +585,10 @@ impl crate::conn::SideData for ClientConnectionData {
     }
 }
 
+impl Output for ClientConnectionData {
```
djc, line 582: So I thought the point of the previous commit was that the `Output` impl would move to `SideData` (from `CommonState`) instead of adding more impls?

ctz: The end goal is that there are a number of smaller objects, all of whom `impl Output` and are fed events independently . For example, in the next PR, `ConnectionCommon` is split into 3 parts - one of those is for public API presentation purposes (one of the many current `ConnectionCommon` responsibilities). That usage would eventually be shared for every type of connection (eg, a `ServerConnection`, a `quic::ServerConnection`, etc.)

For that strategy to work, the levels above need to do a certain amount of "routing". Once this is all solidified, I think codifying where the events should end up can be done more explicitly, eg by each` Event` variant knowing a category.

rustls/src/conn/mod.rs
ctz: I'll bring that forward from a later commit.

PR #2857, opened 2026-01-19
refactor: change mutable references to immutable where applicable and reduce redundant clone

Tunglies:


Tunglies:
I usually use Clippy to look for general improvements.

```bash
> cargo clippy --all-targets --all-features -- -D warnings -D clippy::needless_collect -D clippy::redundant_clone -D clippy::needless_pass_by_ref_mut -D clippy::needless_collect
```

ctz:
> @ctz I think these changes might cause some conflicts with your in-flight work, should we hold off on merging?

I think this should be OK.

PR #2860, opened 2026-01-19
perf: reduce indirection and stack pressure by using Option<&T> instead of &Option<T>

Tunglies:
All tests passed on my machine.

### DIFF
--- a/rustls-bench/src/main.rs
+++ b/rustls-bench/src/main.rs
@@ -866,7 +866,7 @@ impl Parameters {
     }
 
     fn open_latency_file(&self, role: &str) -> LatencyOutput {
-        LatencyOutput::new(&self.latency_prefix, role)
+        LatencyOutput::new(self.latency_prefix.as_deref(), role)
     }
 }
 
@@ -875,9 +875,9 @@ struct LatencyOutput {
 }
 
 impl LatencyOutput {
-    fn new(prefix: &Option<String>, role: &str) -> Self {
+    fn new(prefix: Option<&str>, role: &str) -> Self {
         let thread_id = thread::current().id();
-        let output = prefix.as_ref().map(|prefix| {
+        let output = prefix.map(|prefix| {
             let file_name = format!("{prefix}-{role}-{thread_id:?}-latency.tsv");
             File::create(&file_name).expect("cannot open latency output file")
         });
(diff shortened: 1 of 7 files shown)


djc (commented):
Is this another clippy lint? Please link to the clippy website for more context. Also see https://github.com/rustls/rustls/pull/2858 for how we prefer to handle this going forward.

Will be curious to see if there's a substantial performance impact.

Tunglies:
Nope, there is no rules for this. Actually I tried to migrate some part of rules from [here](https://github.com/clash-verge-rev/clash-verge-rev/blob/dev/Cargo.toml) which I maintained, did not found any suitable for rustls yet.

rustls-bench/src/main.rs
```diff
@@ -875,9 +875,9 @@ struct LatencyOutput {
 }
 
 impl LatencyOutput {
-    fn new(prefix: &Option<String>, role: &str) -> Self {
+    fn new(prefix: Option<&String>, role: &str) -> Self {
```
djc: Maybe turn this into `Option<&str>` instead? `&String` always looks a little unidiomatic to me.

rustls/src/server/tls12.rs
```diff
@@ -417,14 +417,11 @@ mod client_hello {
         })
     }
 
-    fn emit_certificate_req(
-        flight: &mut HandshakeFlightTls12<'_>,
-        config: &ServerConfig,
-    ) -> Result<bool, Error> {
+    fn emit_certificate_req(flight: &mut HandshakeFlightTls12<'_>, config: &ServerConfig) -> bool {
```
djc: This seems unrelated, please pull it out into a separate commit?

Tunglies: Add by unexpected, since tried some additional clippy rules.

djc:
I want to merge clean commits:

- One commit removing `Result` types where unnecessary
- Another commit moving away from `&Option<T>`

Tunglies:
Let me do the removing `&Option<T>` first in this PR.

rustls/src/server/tls12.rs
```diff
@@ -754,7 +754,7 @@ fn emit_ticket(
     cx: &mut ServerContext<'_>,
     ticketer: &dyn TicketProducer,
     now: UnixTime,
-) -> Result<(), Error> {
+) {
```
djc: Sorry: you left in another case of `Result` removal...

Tunglies:
Should be fine now. Thanks for reviewing.

PR #2874, opened 2026-01-28
Move signature verification algorithms into provider crates

djc:


rustls-test/tests/api/server_cert_verifier.rs
```diff
@@ -568,7 +568,7 @@ fn client_check_server_certificate_helper_api() {
                 &correct_roots,
                 &identity.intermediates,
                 UnixTime::now(),
-                webpki::ALL_VERIFICATION_ALGS,
+                provider::ALL_VERIFICATION_ALGS,
```
djc, line 571: Because of this change, aws-lc-rs and ring need to be in the same commit.

rustls/src/crypto/mod.rs
```diff
@@ -369,6 +368,111 @@ impl Borrow<[&'static Tls13CipherSuite]> for CryptoProvider {
     }
 }
 
+/// Describes which `webpki` signature verification algorithms are supported and
+/// how they map to TLS [`SignatureScheme`]s.
+#[expect(clippy::exhaustive_structs)]
+#[derive(Clone, Copy)]
+pub struct WebPkiSupportedAlgorithms {
```
djc, line 378: This does not have any webpki dependencies, but its only use is in the `WebPki{Client,Server}Verifier`s. Expose it directly via `crypto` so that the provider crates can depend on it without necessarily pulling in a webpki dependency.

rustls/Cargo.toml
```diff
@@ -28,7 +29,7 @@ log = { workspace = true, optional = true }
 # only required for no-std
 once_cell = { workspace = true }
 subtle = { workspace = true }
-webpki = { workspace = true }
+webpki = { workspace = true, optional = true }
```
djc, line 32: Avoid dragging in webpki when it should be fully optional (really only required for our default verifier implementations).

rustls-aws-lc-rs/Cargo.toml
```diff
@@ -11,16 +11,17 @@ categories = ["network-programming", "cryptography"]
 autobenches = false
 
 [features]
-default = ["aws-lc-rs/aws-lc-sys", "std"]
-fips = ["aws-lc-rs/fips", "webpki/aws-lc-rs-fips"]
+default = ["aws-lc-sys", "std"]
```
ctz, line 14: Hm, why `aws-lc-sys`? I consider that a private feature of `aws-lc-rs` (because it is not documented)

rustls-aws-lc-rs/src/verify.rs
```diff
@@ -0,0 +1,331 @@
...
+                _ => return Err(InvalidSignature),
+            };
+        }
+        signature::UnparsedPublicKey::new(self.verification_alg, public_key)
+            .verify(message, signature)
+            .map_err(|_| InvalidSignature)
+    }
+
+    fn fips(&self) -> bool {
+        self.in_fips_submission && try_fips_mode().is_ok()
```
ctz: Maybe a followup PR to add `fips_status()` support? This is a prime candidate for the additional information.

rustls-ring/src/verify.rs
```diff
@@ -0,0 +1,222 @@
+// Copyright 2015 Brian Smith.
```
ctz: Generally this crate does not have copyright headers, and the source file has been entirely replaced over the last 11 years. I think we can lose this?

rustls/src/server/config.rs
```diff
@@ -627,6 +630,7 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
     /// This function fails if `key_der` is invalid, or if the
     /// `SubjectPublicKeyInfo` from the private key does not match the public
     /// key for the end-entity certificate from the `cert_chain`.
+    #[cfg(feature = "webpki")]
```
ctz, line 633: Hm, not sure about this one or the one below?

rustls/src/crypto/signer.rs
```diff
@@ -127,6 +133,7 @@ impl Credentials {
     ///
     /// This constructor should be used with all [`SigningKey`] implementations
     /// that can provide a public key, including those provided by rustls itself.
+    #[cfg(feature = "webpki")]
```
ctz, line 136: Also not sure about this? On one hand it does require webpki crate, but on the the other it can deliver most of its contract without it...

rustls-aws-lc-rs/src/verify.rs
ctz: Also, on this line: `super::fips()` should be usable here instead of `try_fips_mode().is_ok()`

rustls/src/crypto/signer.rs
ctz: Not a good comment, please overlook.

rustls-aws-lc-rs/Cargo.toml
djc: Basically, as I understand it, aws-lc-rs does not compile with `default-features = false`. It reports:

```
one of the following features must be specified: `aws-lc-sys`, `non-fips`, or `fips`.
```

So `aws-lc-sys` seems reasonable. Could be `non-fips`, too, but that's equally undocumented. Shall I file an upstream bug?

ctz: AFAICT `non-fips` is incompatible with `fips`. Seems like `aws-lc-sys` is the only option!

rustls-aws-lc-rs/src/verify.rs
djc: Fixed it up in this PR/commit.

rustls-ring/src/verify.rs
djc: Removed.

rustls/src/crypto/signer.rs
djc: We could only validate if `webpki` is available, and use `new_unchecked()` if not?

rustls/src/server/config.rs
djc: We could only validate if `webpki` is available, and use `new_unchecked()` if not?

rustls/src/crypto/signer.rs
ctz: Yeah I don't even agree with a comment a made less than an hour ago. We should try to avoid varying runtime behaviour, especially API-usability error behaviour, on the basis of cargo features. The usability of cargo features is so awful I would like not to inflict them on anyone.

PR #2876, opened 2026-01-28
Deny warnings across CI

djc:
Stacked on top of

- #2874


openssl-tests/src/validate_ffdhe_params.rs
```diff
@@ -57,6 +57,7 @@ fn get_ffdhe_params_from_openssl(ffdhe_group: NamedGroup) -> (Vec<u8>, Vec<u8>)
 }
 
 /// Parse PEM-encoded DH parameters, returning `(p, g)`
+#[allow(clippy::result_large_err)]
```
ctz: `expect`? Also surprised because this function doesn't return a `Result`

djc: Something to do with the closures inside -- I'll add a comment.

PR #2877, opened 2026-01-28
Move stream and key_log_file modules into new rustls-util crate

djc:
New crate, so cargo semver-checks failure is expected.


djc:
@ctz please take a quick look at the clippy fixes here.

PR #2880, opened 2026-01-28
bogo: manipulate encoded server session value directly

djc:
We also fuzz our `ServerSessionValue` parsing. Not sure how to work around that?


ctz:
> We also fuzz our `ServerSessionValue` parsing. Not sure how to work around that?

For that sort of thing I like this kind of approach: https://github.com/rustls/rustls/blob/main/rustls/src/msgs/deframer/mod.rs#L89

ctz (approved):

PR #2881, opened 2026-01-28
webpki as a feature option - simpler alternative

brody2consult:
DRAFT with TODO items:

It appears to me that a major objective of other PR #2874 is to keep use of webpki as an optional feature. This is a quick proposal to achieve this in a much simpler fashion, with what I think should be minimal API impact & avoid possible impact on third-party providers. A third-party provider could elect to use webpki from rustls by using rustls with the webpki option.

I did quickly check that this should not affect any more pub API functions than would be affected by other PR #2874.

STATUS: This is in ROUGH DRAFT status, open for discussion. Still need to resolve many import warnings & a few "XXX" items. I would not be surprised by build failures at this point.

Thanks for your consideration.

MAJOR TODO ITEMS:

- [ ] enable webpki by default to avoid impact on many users
- [ ] resolve imports no longer needed
- [ ] check & resolve build issues
- [ ] resolve "XXX" items
- [ ] cleanup commits - I do think the "refactor: move" commit should come first, by itself

/cc @djc - major credit to other PR #2874 for concept & idea

### DIFF
--- a/rustls/src/error/mod.rs
+++ b/rustls/src/error/mod.rs
@@ -1146,13 +1146,12 @@ pub enum ExtendedKeyPurpose {
     Other(Vec<usize>),
 }
 
+#[cfg(feature = "webpki")]
 impl ExtendedKeyPurpose {
     pub(crate) fn for_values(values: impl Iterator<Item = usize>) -> Self {
         let values = values.collect::<Vec<_>>();
         match &*values {
-            #[cfg(feature = "webpki")]
             ExtendedKeyUsage::CLIENT_AUTH_REPR => Self::ClientAuth,
-            #[cfg(feature = "webpki")]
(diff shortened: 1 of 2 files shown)


djc:
Can you describe conceptually how this is different from my PR? You wrote a lot of words in the PR description but left out this pretty important aspect.

brody2consult:
I tried to keep this PR more focused on simply keeping webpki as an optional feature in the main rustls crate, while the other PR goes further and includes what looks like some webpki implementation in aws-lc-rs & ring providers to keep webpki optional for these providers as well. This proposal was simpler in that the aws-lc-rs & ring providers would need to use rustls with the webpki option enabled.

I noticed what looks like quite a bit of duplicated code in the other PR, that is why I had tried raising this proposal for discussion.

Now that the other PR is merged, I will not work further on this one. Thanks.

PR #2887, opened 2026-01-30
Reduce internal API to only enable fuzzing

djc:


djc:
Pretty surprised there are a bunch of `no_crypto` perf changes.

ctz:
> Pretty surprised there are a bunch of `no_crypto` perf changes.

The performance thing only measures the difference between the branch base and tip; it doesn't for example measure the result of merging a branch. Run locally for main vs this branch rebased, this shows:

```
| handshake_session_id_1.3_no_crypto_client | 2207903 | 2207903 | 0 (0.00%) |
```

PR #2888, opened 2026-01-30
client: use builder pattern to make connections

### LINKED ISSUE #2850: Make a builder for `ClientConnection`s?
Today we have `ClientConnection::new()` (which takes `config` and `name`) and `ClientConnection::new_with_alpn()` (which additionally takes `alpn_protocols`). There is a request to add support for RFC 9149 which allows the client to suggest how many tickets it would like to receive. This seems like it should be sent on a per-connection (not per-config) basis. We should probably consider how to make the API for this modular/extensible.

(It might make sense for providing a buffer to write the `ClientHello` into directly on creating the connection, which may also warrant a slightly different API.)

djc:
Fixes #2850.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -98,7 +98,10 @@ pub fn main() {
                 let server_name = ServerName::try_from(opts.host_name.as_str())
                     .unwrap()
                     .to_owned();
-                let sess = ClientConnection::new(config.clone(), server_name).unwrap();
+                let sess = config
+                    .connect(server_name)
+                    .build()
+                    .unwrap();
                 exec(&opts, Connection::Client(sess), &key_log, i);
             }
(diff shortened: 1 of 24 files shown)


ctz (approved):
I'm working towards deleting the unbuffered API so not suggesting this needs to happen in this PR, but it would be good to think about how this would fit together with that? Maybe a `fn unbuffered() -> UnbufferedClientConnection` or whatever on the builder type?

djc:
> I'm working towards deleting the unbuffered API so not suggesting this needs to happen in this PR, but it would be good to think about how this would fit together with that? Maybe a `fn unbuffered() -> UnbufferedClientConnection` or whatever on the builder type?

Yes, I think something like that could make sense.

Should we just kill the unbuffered API now if we already know we're going to delete it?

PR #2890, opened 2026-01-31
Correct reachable `Error::Unreachable` for `KernelConnection::update_*_secret()`

ctz:


rustls-test/tests/api/quic.rs
```diff
@@ -282,23 +282,25 @@ fn test_quic_no_tls13_error() {
     client_config.alpn_protocols = vec![b"foo".into()];
     let client_config = Arc::new(client_config);
 
-    assert!(
+    assert_eq!(
```
djc, line 285: IMO "actual, expected" is a little more idiomatic than "expected, actual", so I'd prefer not to proliferate the latter.

ctz:
> Could/should we ban `Result::is_err()` via clippy?

Added a commit for this.

PR #2891, opened 2026-01-31
Update provider documentation for 0.24

djc:


rustls/src/crypto/mod.rs
```diff
@@ -92,14 +93,14 @@ pub use crate::suites::CipherSuiteCommon;
 ///
 /// Supply the provider when constructing your [`ClientConfig`] or [`ServerConfig`]:
 ///
-/// - [`ClientConfig::builder()`]
-/// - [`ServerConfig::builder()`]
+/// - [`ClientConfig::builder()`][crate::ClientConfig::builder()`]
+/// - [`ServerConfig::builder()`][crate::ServerConfig::builder()`]
```
ctz:
```suggestion
/// - [`ClientConfig::builder()`][crate::ClientConfig::builder()]
/// - [`ServerConfig::builder()`][crate::ServerConfig::builder()]
```

rustls/src/crypto/mod.rs
```diff
@@ -92,14 +93,14 @@ pub use crate::suites::CipherSuiteCommon;
...
 ///
 /// When creating and configuring a webpki-backed client or server certificate verifier, a choice of
 /// provider is also needed to start the configuration process:
 ///
-/// - [`client::WebPkiServerVerifier::builder()`]
-/// - [`server::WebPkiClientVerifier::builder()`]
+/// - [`WebPkiServerVerifier::builder()`][crate::client::WebPkiServerVerifier::builder()`]
+/// - [`WebPkiClientVerifier::builder()`][crate::server::WebPkiClientVerifier::builder()`]
```
ctz:
```suggestion
/// - [`WebPkiServerVerifier::builder()`][crate::client::WebPkiServerVerifier::builder()]
/// - [`WebPkiClientVerifier::builder()`][crate::server::WebPkiClientVerifier::builder()]
```

rustls/src/lib.rs
```diff
@@ -42,16 +42,29 @@
...
+//! * [`rustls-ring`] - a provider that uses the [`ring`] crate for cryptography. This
+//!   provider is easier to build on a variety of platforms, but has a more limited feature set
+//!   (for example, it does not support post-quantum algorithms).
 //!
 //! See the documentation for [`crypto::CryptoProvider`] for details on how providers are
 //! selected.
 //!
+//! (For rustls versions prior to 0.24, both of these providers are shipped as part of the rustls
+//! crate, and Cargo features are used to select the preferred provider. The `aws-lc-rs` feature
+//! is enabled by default.)
```
ctz:
```suggestion
//! (For rustls versions prior to 0.24, both of these providers were shipped as part of the rustls
//! crate, and Cargo features were used to select the preferred provider. The `aws-lc-rs` feature
//! was enabled by default.)
```

ctz (approved):

rustls/src/lib.rs
djc: We're still shipping 0.23.x releases and may continue doing so for a while, right?

ctz: Yeah, I guess this is a case where sharing text between a repo README and crate documentation is not ideal. I think a README covering multiple versions is OK, but a crate release I think should be specific to that release (and therefore use past tense when mentioning previous releases.)

djc: Fair enough...

PR #2892, opened 2026-01-31
Use monotonic time for the TicketRotator

### LINKED ISSUE #1466: Use monotonic time for ticket refresh/lifetimes
I think the ticketer code should be relying only on monotonic time so tickets have the desired lifetime irrespective of system clock jumps.

djc:
Fixes #1466.

### DIFF
--- a/rustls/src/ticketer.rs
+++ b/rustls/src/ticketer.rs
@@ -3,8 +3,7 @@ use alloc::vec::Vec;
 use core::mem;
 use core::time::Duration;
 use std::sync::{RwLock, RwLockReadGuard};
-
-use pki_types::UnixTime;
+use std::time::Instant;
 
 use crate::crypto::TicketProducer;
 use crate::error::Error;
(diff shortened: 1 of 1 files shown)


ctz (approved):
An additional thing to consider here is reconsidering the use of `UnixTime` in the `msgs::persist` types. (Though #1466 is definitely fully addressed by this PR and is not asking for more.)

djc:
> An additional thing to consider here is reconsidering the use of `UnixTime` in the `msgs::persist` types. (Though #1466 is definitely fully addressed by this PR and is not asking for more.)

So I commented in that issue 2.5 years ago:

> One issue I thought about: this pretty much only works for a single process run, right? If the machine persists tickets across reboots or if tickets are shared via a network cache, there would still be no monotonicity guarantee (and it might even be worse than using a `SystemTime`)?

Are you sure using `Instant` for things that get persisted across peers/processes is the right design?

Right now, the value of `Instant::now()` on my machine is 

```
[rustls/src/ticketer.rs:235:18] expiry = Instant {
    tv_sec: 358328,
    tv_nsec: 855860750,
}
```

PR #2893, opened 2026-01-31
Split supported verification algorithms off from provider

djc:
What do we think of doing something like this?

It feels like projects depending on, say, the platform verifier shouldn't have to drag around the provider's signature verification algorithms, even if that means a small ergonomic hit.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -27,7 +27,7 @@ use rustls::crypto::hpke::{Hpke, HpkePublicKey};
 use rustls::crypto::kx::NamedGroup;
 use rustls::crypto::{
     Credentials, CryptoProvider, Identity, SelectedCredential, SignatureScheme, Signer, SigningKey,
-    SingleCredential,
+    SingleCredential, WebPkiSupportedAlgorithms,
 };
 use rustls::enums::{
     ApplicationProtocol, CertificateCompressionAlgorithm, CertificateType, ProtocolVersion,
@@ -845,6 +845,13 @@ impl SelectedProvider {
         }
     }
 
+    fn verify_algs(&self) -> WebPkiSupportedAlgorithms {
+        match *self {
+            Self::AwsLcRs | Self::AwsLcRsFips => rustls_aws_lc_rs::SUPPORTED_SIG_ALGS,
+            Self::Ring => rustls_ring::SUPPORTED_SIG_ALGS,
+        }
+    }
+
     fn supports_ech(&self) -> bool {
         match *self {
             Self::AwsLcRs | Self::AwsLcRsFips => true,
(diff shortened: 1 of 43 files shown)


ctz:
I think this sort of effort should start with rustls-platform-verifier having no dependency on `CryptoProvider::signature_verification_algorithms`. Currently it uses it on every platform!

djc:
> I think this sort of effort should start with rustls-platform-verifier having no dependency on `CryptoProvider::signature_verification_algorithms`. Currently it uses it on every platform!

Ahh, I forgot that `verify_server_cert()` was just one part of the verifier, and we also have the `verify_tls1*_signature()` methods which do still end up using the `WebPkiSupportedAlgorithms`.

So do you think this conceptually makes no sense? It still feels a little weird that rustls itself does not use `CryptoProvider::signature_verification_algorithms` directly.

ctz:
> > I think this sort of effort should start with rustls-platform-verifier having no dependency on `CryptoProvider::signature_verification_algorithms`. Currently it uses it on every platform!
> 
> Ahh, I forgot that `verify_server_cert()` was just one part of the verifier, and we also have the `verify_tls1*_signature()` methods which do still end up using the `WebPkiSupportedAlgorithms`.
> 
> So do you think this conceptually makes no sense? It still feels a little weird that rustls itself does not use `CryptoProvider::signature_verification_algorithms` directly.

In an ideal world there would be a base `CryptoProvider` and refinements to it (eg, one with HPKE suites for use with ECH, another with signature verification for built-in certificate verification, another with key loading for servers or clients proving their identity, and then all combinations of these). The outcome would be that all `CryptoProvider` configuration surface would deal with all of these, and the API for (eg) ECH would become unavailable if your configured provider can't do that, and nobody would have to deal with a fragmented set of items like in this PR.

However, I don't know how to actually do that without creating a system of multi-dimension generic types that would make a libboost maintainer go "ehh, steady on!"

djc:
> However, I don't know how to actually do that without creating a system of multi-dimension generic types that would make a libboost maintainer go "ehh, steady on!"

I'll take that as a challenge! What if `CryptoProvider` becomes a trait and we have various subtraits? `ServerConfig` and `ClientConfig` both take a `<P>` for their provider and a `P: CryptoProvider` can additionally be `QuicProvider` or `HpkeProvider`. Doesn't seem all that multidimensional?

ctz:
I guess the multidimensional part would be someone wanting to do quic+client auth+default server verifier? So they would have to find/construct a type that `impl CryptoProvider + QuicProvider + AuthProvider + VerifierProvider` or whatever

djc:
Sure, but I would argue that's additive and so only constitutes a single dimension? It doesn't seem very onerous, particularly if it's mostly just AwsLcRsProvider providing a bunch of stuff, the Ring provider providing less stuff, etc.

PR #2897, opened 2026-02-03
Toward making low-level state machine functionally pure (pt 6)

djc:
Previously #2832, #2833, #2834, #2854, #2855

Omnibus PR: #2818

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -173,77 +173,30 @@ impl CommonState {
 
 impl Output for CommonState {
     fn emit(&mut self, ev: Event<'_>) {
-        match ev {
-            Event::ApplicationProtocol(protocol) => {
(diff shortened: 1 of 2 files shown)


rustls/src/common_state.rs
```diff
@@ -752,13 +663,131 @@ impl Output for CommonState {
...
+            negotiated_kx_group: None,
+            alpn_protocol: None,
+            peer_identity: None,
+            exporter: None,
+            early_exporter: None,
+            fips: FipsStatus::Unvalidated,
+            tls13_tickets_received: 0,
+        }
+    }
+}
```
djc: This looks like it can be derived.

djc: I guess `FipsStatus` does not implement `Default` yet -- but I think implementing `Default` to mean `Unvalidated` would be reasonable. Can do it after this is merged.

djc:
@ctz I suppose this needs a rubber stamp from you -- I did not make any changes.

PR #2900, opened 2026-02-03
Move remaining items out of msgs::persist

djc:
They weren't really msgs anyway.


rustls/src/server/mod.rs
```diff
@@ -40,3 +50,264 @@ pub mod danger {
 
 #[cfg(test)]
 mod test;
+
+// --- Server types ---
+#[non_exhaustive]
```
ctz: preexisting - `non_exhaustive` may not be needed now this is not a public type?

djc: Dropped it.

PR #2905, opened 2026-02-06
Remove the unbuffered API

djc:
This will shortly be replaced with an API that can share more code with the high-level API.

@ctz let me know if I removed some stuff that you'll want, or if there's stuff I should remove that I didn't.


ctz (approved):

maxim-pulsar:
Hi, are there still plans to reintroduce support for the functionality provided by the old unbuffered API in the new high-level API?

djc:
@maxim-pulsar have you already tried using the current API on `main`? Feel free to report any issues you've found with it.

PR #2906, opened 2026-02-06
Flatten out `Deref`s in public API 

ctz:
This changes `Connection` into a trait rather than enum, and ensures that `ClientConnection` and `ServerConnection` impls it.  This removes `ConnectionCommon` and `CommonState` derefs, and allows those types to become private.

This becomes important later, because the use of `Deref` commits us to using those parts for the internal representation of (eg) a `ClientConnection`.

This comes with the pain that `use rustls::Connection` is needed for most users. I think that is OK these days due to IDE improvements (eg, my IDE just inserts the right `use`).

This is draft as it should come after #2905 which will allow further withdrawing of `ConnectionCommon`, `SideData`, etc. from the public API.

### DIFF
--- a/rustls-test/tests/api/quic.rs
+++ b/rustls-test/tests/api/quic.rs
@@ -4,12 +4,12 @@
 
 use std::sync::Arc;
 
+use rustls::HandshakeKind;
 use rustls::client::Resumption;
 use rustls::error::{
     AlertDescription, ApiMisuse, Error, InvalidMessage, PeerIncompatible, PeerMisbehaved,
 };
-use rustls::quic::{self, ConnectionCommon, Side};
-use rustls::{HandshakeKind, SideData};
+use rustls::quic::{self, Connection, Side};
 use rustls_test::{
     ClientStorage, KeyType, encoding, make_client_config, make_server_config, server_name,
 };
@@ -17,9 +17,9 @@ use rustls_test::{
 use super::provider;
 
 // Returns the sender's next secrets to use, or the receiver's error.
-fn step<L: SideData, R: SideData>(
-    send: &mut ConnectionCommon<L>,
-    recv: &mut ConnectionCommon<R>,
+fn step(
+    send: &mut impl Connection,
+    recv: &mut impl Connection,
 ) -> Result<Option<quic::KeyChange>, Error> {
     let mut buf = Vec::new();
     let change = loop {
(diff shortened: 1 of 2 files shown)


djc:
How about we just drop `Connection` in 0.24.0 and see how bad people miss it?

It feels kind of low-value and it's easy-ish to make a downstream replacement for it where needed.

And we could add it back in 0.24.1 if people really want it.

Or, could we solve this using something more like #2669?

ctz:
> How about we just drop `Connection` in 0.24.0 and see how bad people miss it?

Willing to try that --  293b39e  is what that looks like for the QUIC API. WDYT?

djc:
> > How about we just drop `Connection` in 0.24.0 and see how bad people miss it?
> 
> Willing to try that -- [293b39e](https://github.com/rustls/rustls/commit/293b39eea4a3af7c61885e2358e94563bb32c586) is what that looks like for the QUIC API. WDYT?

Looks okay to me.

Still feels like this is worse than #2669? I guess this is also going to monomorphize `bogo::exec()`?

> This changes `Connection` into a trait rather than enum, and ensures that `ClientConnection` and `ServerConnection` impls it. This removes `ConnectionCommon` and `CommonState` derefs, and allows those types to become private.

But these are kind of two different steps though, right? The `CommonState` derefs makes it so that we don't have duplicate the `CommonState` API all over the place, which this brings back.

> This becomes important later, because the use of `Deref` commits us to using those parts for the internal representation of (eg) a `ClientConnection`.

Why is this a problem exactly?

ctz:
> Still feels like this is worse than #2669? I guess this is also going to monomorphize `bogo::exec()`?



> > This becomes important later, because the use of `Deref` commits us to using those parts for the internal representation of (eg) a `ClientConnection`.
> 
> Why is this a problem exactly?

In the other work to implement the buffered API in terms of the new state-based API, `ClientConnection` ends up looking like:

```rust
    pub struct ClientConnection {
        state: ClientState,
        buffers: ConnectionBuffers,
    }
```

`ClientState` is an enum, where the final state just doesn't have a value of `CommonState` any more -- it has been split into its constituent parts (in a lossy way, mind). So it is functionally impossible to have a deref for `CommonState` for that.

It is also not possible for a `Connection<Client>` to work, because while you could have an associated type on `Client` naming `ClientState`, most of the code needs to reveal the enum variants to work.  So you'd end up with something similar to this code, with most of the public API in `impl Connection<Client>` for the instead of `impl ClientConnection`

ctz:
I think basically here I am rediscovering the parts that @cpu found in in https://github.com/rustls/rustls/issues/1723#issuecomment-1946621961 (second half)

djc:
Okay, I think that makes sense. My sense is that we could loose the inherent methods on the connection types in favor of only implementing the `Connection` trait, but maybe that's worse because everyone needs the `Connection` import then?

ctz:
I have reworked this to be a half-way version, where there still is a deref to `ConnectionOutputs`. That remains possible with the new API, and avoids some of the excessive duplication.

djc:
> I have reworked this to be a half-way version, where there still is a deref to `ConnectionOutputs`. That remains possible with the new API, and avoids some of the excessive duplication.

That seems like an improvement.

The duplication here still seems ugly but probably okay for a while we flesh out the larger API transition.

djc (approved):

PR #2907, opened 2026-02-06
client: make client cache values serializable

### LINKED ISSUE #2287: Barriers to implementing persistent `ClientSessionStore`
[Semi-related discussion](https://github.com/quinn-rs/quinn/pull/1912#discussion_r1884667793). I'm looking at `rustls::client::ClientSessionStore`, and trying to think about what it would take for a user to create an implementation that saves sessions to disk so that they can be accessed from subsequent processes. AFAICT, there may be obstacles to that presently? Even aside from concerns about blocking versus async. I wanted to create this issue both to serve as a list of things that could be helpful on this matter, and so that someone can point out if I'm mistaken about the way things are.

- There is no way for the user to construct a `Tls12ClientSessionValue`
  - Even if there was, there is no way to access any of its fields
    - `ClientSessionCommon` is part of its public API via `Deref`, but is not publicly importable
- There is no way for the user to construct a `Tls13ClientSessionValue`
  - Even if there was, it only exposes the fields needed for QUIC, but not those of `ClientSessionCommon`
- It's not clear what the security implications would be if `remove_tls12_session` wasn't abided by
  - There may be other similar aspects of it worth documenting more
- It doesn't mesh well with asynchronous implementations. However, this is not a hard-blocker, as there are possible workaround a user could do. For example, a user could implement storage operations by spawning a task or sending a message in a channel to do the storage asynchronously, and pre-fetch the data for a server name from disk before initializing a TLS client.

djc:
Including some modernization of the `Reader` abstraction.

Fixes #2287.

### DIFF
--- a/rustls-test/src/lib.rs
+++ b/rustls-test/src/lib.rs
@@ -9,8 +9,7 @@ use rustls::client::danger::{
     HandshakeSignatureValid, PeerVerified, ServerIdentity, ServerVerifier,
 };
 use rustls::client::{
-    ClientSessionKey, ServerVerifierBuilder, Tls13ClientSessionValue, WantsClientCert,
-    WebPkiServerVerifier,
+    ClientSessionKey, ServerVerifierBuilder, Tls13Session, WantsClientCert, WebPkiServerVerifier,
 };
 use rustls::crypto::cipher::{
     EncodedMessage, InboundOpaque, MessageDecrypter, MessageEncrypter, Payload,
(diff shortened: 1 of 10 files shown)


rustls-test/src/lib.rs
```diff
@@ -1873,7 +1872,7 @@ impl rustls::client::ClientSessionStore for ClientStorage {
     fn set_tls12_session(
         &self,
         key: ClientSessionKey<'static>,
-        value: rustls::client::Tls12ClientSessionValue,
+        value: rustls::client::Tls12Ticket,
```
ctz: I like the idea of renaming this, it's very wordy, but it's not a ticket really. Just `client::Tls12Session` and `client::Tls13Session`?

ctz (approved):

rustls-test/src/lib.rs
djc: Done.

PR #2908, opened 2026-02-07
server: improve resumption APIs

djc:
I did not end up going with having `ServerSession` appear in as the value in the `ServerSessionStore` trait methods. That ends up being a bit of a pain for testing in particular, and adds quite a bit of complexity.

Here are some changes that I think might make this code easier to reason about. I dropped a commit that avoided allocating when storing session values, because it didn't seem much of an improvement.

### DIFF
--- a/rustls/src/server/tls13.rs
+++ b/rustls/src/server/tls13.rs
@@ -29,7 +29,7 @@ use crate::log::{debug, trace, warn};
 use crate::msgs::{
     CERTIFICATE_MAX_SIZE_LIMIT, CertificatePayloadTls13, Codec, HandshakeMessagePayload,
     HandshakePayload, KeyUpdateRequest, Message, MessagePayload, NewSessionTicketPayloadTls13,
-    Reader, SizedPayload,
+    PresharedKeyIdentity, Reader, SizedPayload,
 };
 use crate::suites::PartiallyExtractedSecrets;
 use crate::sync::Arc;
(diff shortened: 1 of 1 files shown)


rustls/src/server/tls13.rs
```diff
@@ -441,8 +441,19 @@ mod client_hello {
 
         let now = config.current_time()?;
         for (i, psk_id) in psk_offer.identities.iter().enumerate() {
-            let Some(mut session) =
-                attempt_tls13_ticket_decryption(psk_id.identity.bytes(), &config)
+            let encoded = match config.ticketer.as_deref() {
```
ctz: I think this is a case where the function is giving a useful name to the process for documentation purposes. If you feel strongly about inlining these, could we put the documentation in a comment instead?

rustls/src/server/tls13.rs
```diff
@@ -481,27 +489,6 @@ mod client_hello {
         Ok(None)
     }
 
-    fn check_binder(
```
ctz, line 459: Ditto.

rustls/src/server/tls13.rs
djc: Dropped this.

rustls/src/server/tls13.rs
djc: Dropped.

PR #2915, opened 2026-02-11
Remove nested connection modules

djc:


ctz:
Onboard with doing this, but it's going to cause some obnoxious conflicts elsewhere. Later?

djc:
Yup, makes sense. Will wait until after your stuff has landed.

PR #2917, opened 2026-02-12
Toward making low-level state machine functionally pure (pt 7)

ctz:
This one is mostly moving things around: moving things specific to the buffered API up to that level, and moving common receive-side actions into `ReceivePath`.

From: #2818
Previously:  #2832, #2833, #2834, #2854, #2855, #2897


djc:
Also, I was using the name `TlsInputBuffer` for `ReceivedData` which I like better because it's a little more specific.

ctz:
> No need to block on this here, but in my recent exploration I think we want to move/mirror `has_seen_eof` into the `ReceivedData` implementer.

Eventually I have this live in a `ConnectionBuffers` internal struct, which is all the buffers required for the std & buffered connection API.

PR #2918, opened 2026-02-12
Fixing daily-tests and CONTRIBUTING.md

ctz:
Fixing https://github.com/rustls/rustls/actions/runs/21918054244


cpu:
Does the `Api` type in `rustls-bench/src/main.rs` still make sense or should it also be removed?

ctz:
> Does the `Api` type in `rustls-bench/src/main.rs` still make sense or should it also be removed?

I'm going to use this enum again soon to test the new unbuffered API.

PR #2936, opened 2026-02-25
msgs: deframer buffer tweaks

djc:
Clean this up a little bit.

### DIFF
--- a/rustls/src/msgs/deframer/buffers.rs
+++ b/rustls/src/msgs/deframer/buffers.rs
@@ -5,146 +5,6 @@ use std::io;
 
 use crate::msgs::MAX_WIRE_SIZE;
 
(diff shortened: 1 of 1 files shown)


rustls/src/msgs/deframer/buffers.rs
```diff
@@ -273,45 +273,6 @@ impl DeframerVecBuffer {
     }
 }
 
-/// A borrowed version of [`DeframerVecBuffer`] that tracks discard operations
```
ctz, line 276: Later on I rename this to `SliceBuffer`, make it public, and remove unused parts in the impl block (so it retains `new()` and adds a new `fn into_used(self) -> usize`. Don't mind if it's deleted here because it's simple and the result is different in every respect.

PR #2940, opened 2026-02-25
Externalize input buffers

djc:
This is an incremental step towards externalizing buffer management. After this, I think we can drop the received plaintext buffer in favor of yielding a slice iterator from `process_new_packets()`, and other simplifications.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -46,7 +46,7 @@ use rustls::server::{
     self, ClientHello, PreferClientOrder, PreferServerOrder, ServerConfig, ServerConnection,
     ServerSessionKey, WebPkiClientVerifier,
 };
-use rustls::{Connection, DistinguishedName, HandshakeKind, RootCertStore, compress};
+use rustls::{Connection, DistinguishedName, HandshakeKind, RootCertStore, VecInput, compress};
 use rustls_aws_lc_rs::hpke;
 
 pub fn main() {
@@ -154,6 +154,7 @@ fn exec(opts: &Options, mut sess: impl Connection + 'static, key_log: &KeyLogMem
     conn.write_all(&opts.shim_id.to_le_bytes())
         .unwrap();
 
+    let mut input = VecInput::default();
     loop {
         if !sent_message && (opts.queue_data || (opts.queue_data_on_resume && count > 0)) {
             if !opts
@@ -162,7 +163,13 @@ fn exec(opts: &Options, mut sess: impl Connection + 'static, key_log: &KeyLogMem
             {
                 flush(&mut sess, &mut conn);
                 for message_size_estimate in &opts.queue_early_data_after_received_messages {
-                    read_n_bytes(opts, &mut sess, &mut conn, *message_size_estimate);
+                    read_n_bytes(
+                        opts,
+                        &mut input,
+                        &mut sess,
+                        &mut conn,
+                        *message_size_estimate,
+                    );
                 }
                 println!("now ready for early data");
             }
@@ -188,7 +195,7 @@ fn exec(opts: &Options, mut sess: impl Connection + 'static, key_log: &KeyLogMem
         }
 
         if sess.wants_read() {
-            read_all_bytes(opts, &mut sess, &mut conn);
+            read_all_bytes(opts, &mut input, &mut sess, &mut conn);
         }
 
         if opts.side == Side::Server && opts.enable_early_data {
@@ -414,37 +421,69 @@ fn server(conn: &mut dyn Any) -> &mut ServerConnection {
         .unwrap()
 }
 
-fn read_n_bytes(opts: &Options, sess: &mut impl Connection, conn: &mut net::TcpStream, n: usize) {
+fn read_n_bytes(
+    opts: &Options,
+    input: &mut VecInput,
+    sess: &mut impl Connection,
+    conn: &mut net::TcpStream,
+    n: usize,
+) {
     let mut bytes = [0u8; MAX_MESSAGE_SIZE];
     match conn.read(&mut bytes[..n]) {
         Ok(count) => {
             println!("read {count:?} bytes");
-            sess.read_tls(&mut io::Cursor::new(&mut bytes[..count]))
+            input
+                .read(&mut io::Cursor::new(&mut bytes[..count]))
                 .expect("read_tls not expected to fail reading from buffer");
         }
         Err(err) if err.kind() == io::ErrorKind::ConnectionReset => {}
         Err(err) => panic!("invalid read: {err}"),
     };
 
-    after_read(opts, sess, conn);
+    after_read(opts, input, sess, conn);
 }
 
-fn read_all_bytes(opts: &Options, sess: &mut impl Connection, conn: &mut net::TcpStream) {
-    match sess.read_tls(conn) {
+fn read_all_bytes(
+    opts: &Options,
+    input: &mut VecInput,
+    sess: &mut impl Connection,
+    conn: &mut net::TcpStream,
+) {
+    match input.read(conn) {
         Ok(_) => {}
         Err(err) if err.kind() == io::ErrorKind::ConnectionReset => {}
         Err(err) => panic!("invalid read: {err}"),
     };
 
-    after_read(opts, sess, conn);
+    after_read(opts, input, sess, conn);
 }
 
-fn after_read(opts: &Options, sess: &mut impl Connection, conn: &mut net::TcpStream) {
-    if let Err(err) = sess.process_new_packets() {
+fn after_read(
+    opts: &Options,
+    input: &mut VecInput,
+    sess: &mut impl Connection,
+    conn: &mut net::TcpStream,
+) {
+    //dbg!("process new packets", input.slice_mut().len());
+    if let Err(err) = sess.process_new_packets(input) {
+        //dbg!(&err);
         flush(sess, conn); /* send any alerts before exiting */
         orderly_close(conn);
         handle_err(opts, err);
     }
+    /*
+    let mut plaintext = [0u8; MAX_MESSAGE_SIZE];
+    let mut reader = sess.reader();
+    loop {
+        match reader.read(&mut plaintext) {
+            Ok(0) => break,
+            Ok(read) => {
+                dbg!(read);
+            }
+            Err(err) if err.kind() == io::ErrorKind::WouldBlock => break,
+            Err(err) => panic!("invalid plaintext read: {err:?}"),
+        }
+    }*/
 }
 
 fn flush(sess: &mut impl Connection, conn: &mut net::TcpStream) {
--- a/ci-bench/src/main.rs
+++ b/ci-bench/src/main.rs
@@ -21,7 +21,7 @@ use rustls::enums::ProtocolVersion;
 use rustls::server::{NoServerSessionStorage, ServerSessionMemoryCache, WebPkiClientVerifier};
 use rustls::{
     ClientConfig, ClientConnection, Connection, HandshakeKind, RootCertStore, ServerConfig,
-    ServerConnection,
+    ServerConnection, VecInput,
 };
 use rustls_test::KeyType;
 
@@ -237,6 +237,7 @@ fn main() -> anyhow::Result<()> {
                                     &bench.params,
                                     resumption_kind,
                                 ),
+                                input: VecInput::default(),
                             },
                             bench.kind,
                             resumed_reps,
@@ -252,6 +253,7 @@ fn main() -> anyhow::Result<()> {
                                     &bench.params,
                                     resumption_kind,
                                 ),
+                                input: VecInput::default(),
                             },
                             bench.kind,
                             resumed_reps,
@@ -296,6 +298,7 @@ fn main() -> anyhow::Result<()> {
                                     handshake_buf,
                                 },
                                 config: ServerSideStepper::make_config(params, resumption_kind),
+                                input: VecInput::default(),
                             },
                             bench.kind,
                             RESUMED_HANDSHAKE_RUNS,
@@ -314,6 +317,7 @@ fn main() -> anyhow::Result<()> {
                                 },
                                 resumption_kind,
                                 config: ClientSideStepper::make_config(params, resumption_kind),
+                                input: VecInput::default(),
                             },
                             bench.kind,
                             RESUMED_HANDSHAKE_RUNS,
@@ -686,6 +690,7 @@ struct ClientSideStepper<'a> {
     io: StepperIo<'a>,
     resumption_kind: ResumptionKind,
     config: Arc<ClientConfig>,
+    input: VecInput,
 }
 
 impl ClientSideStepper<'_> {
@@ -739,7 +744,13 @@ impl BenchStepper for ClientSideStepper<'_> {
             if !client.is_handshaking() && !client.wants_write() {
                 break;
             }
-            read_handshake_message(&mut client, self.io.reader, self.io.handshake_buf).await?;
+            read_handshake_message(
+                &mut self.input,
+                &mut client,
+                self.io.reader,
+                self.io.handshake_buf,
+            )
+            .await?;
         }
 
         // Session ids and tickets are no longer part of the handshake in TLS 1.3, so we need to
@@ -747,7 +758,13 @@ impl BenchStepper for ClientSideStepper<'_> {
         if self.resumption_kind != ResumptionKind::No
             && client.protocol_version().unwrap() == ProtocolVersion::TLSv1_3
         {
-            read_handshake_message(&mut client, self.io.reader, self.io.handshake_buf).await?;
+            read_handshake_message(
+                &mut self.input,
+                &mut client,
+                self.io.reader,
+                self.io.handshake_buf,
+            )
+            .await?;
         }
 
         Ok(client)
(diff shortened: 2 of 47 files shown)


bogo/src/main.rs
```diff
@@ -413,33 +414,52 @@ fn server(conn: &mut dyn Any) -> &mut ServerConnection {
...
+) {
     let mut bytes = [0u8; MAX_MESSAGE_SIZE];
     match conn.read(&mut bytes[..n]) {
         Ok(count) => {
             println!("read {count:?} bytes");
-            sess.read_tls(&mut io::Cursor::new(&mut bytes[..count]))
-                .expect("read_tls not expected to fail reading from buffer");
+            buf.read(
+                &mut io::Cursor::new(&mut bytes[..count]),
+                sess.is_handshaking(),
```
ctz: Just picking one instance of this, but `is_handshaking()` is different to "currently reassembling a handshake message". The first is a phase of the connection, the second is possible at all times of a connection because handshake messages can be received after the handshake, and can be large.

I would suggest if we do this change, we can just stop with that policy in `VecBuffer` and make it willing to grow the buffer as required?

djc: Inserted a commit that always uses the "handshake" 64k limit, instead of using a smaller one.

rustls/src/conn/mod.rs
```diff
@@ -1071,6 +1060,19 @@ pub(crate) trait TlsInputBuffer {
...
+    /// This is used to detect unclean connection closures, which are signalled by a TCP EOF
+    /// without a preceding TLS `close_notify` alert. In that case, the connection is closed
+    /// without the usual TLS closure semantics, and any data received so far should be
+    /// treated as suspect due to potential truncation attacks.
+    fn has_seen_eof(&self) -> bool;
+
+    /// Mark the connection as closed, so that no further data will be accepted.
+    ///
+    /// This will be called when the connection receives a TLS `close_notify` alert.
+    fn set_closed(&mut self);
```
ctz: These two only benefit connections that use std::io concepts for their "top" interface. So I think these should be in an additional trait which incorporates `TlsInputBuffer`, and is specific to those connection types?

djc: I was aiming to make incremental steps. Since we currently only accept input via `std::io` (even after this PR) this takes the minimal approach of externalizing the buffer. Later on, we could change stuff to bypass I/O (but it feels like even input paths that don't use `std` I/O need to be able to signal EOF?).

ctz: I think at the API level, if we don't have any std-style IO we only need to expose whether the read-side has terminated with a close_notify. Everything else builds on top of that, including the policy to have plaintext IO fail with an error if the TLS IO sees EOF without a close_notify.

djc: Currently `UnexpectedEof` errors are delayed -- they're not returned from `read_tls()` but from the `Reader`. I think we could fix that but it seems to be pretty involved. Here's what I have so far:

```diff
diff --git a/bogo/src/main.rs b/bogo/src/main.rs
index 2152e4b46..8c8bd8213 100644
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -145,7 +145,6 @@ fn exec(opts: &Options, mut sess: impl Connection + 'static, key_log: &KeyLogMem
         net::SocketAddr::from((net::Ipv4Addr::LOCALHOST, opts.port)),
     ];
     let mut conn = net::TcpStream::connect(&addrs[..]).expect("cannot connect");
-    let mut sent_shutdown = false;
     let mut sent_exporter = false;
     let mut sent_key_update = false;
     let mut quench_writes = false;
@@ -186,7 +185,13 @@ fn exec(opts: &Options, mut sess: impl Connection + 'static, key_log: &KeyLogMem
             flush(&mut sess, &mut conn);
         }
 
+        if opts.shut_down_after_handshake && !sess.is_handshaking() {
+            sess.send_close_notify();
+            break;
+        }
+
         if sess.wants_read() {
+            //dbg!(sess.is_handshaking());
             read_all_bytes(opts, &mut sess, &mut conn);
         }
 
@@ -378,11 +383,6 @@ fn exec(opts: &Options, mut sess: impl Connection + 'static, key_log: &KeyLogMem
             Err(err) => panic!("unhandled read error {err:?}"),
         };
 
-        if opts.shut_down_after_handshake && !sent_shutdown && !sess.is_handshaking() {
-            sess.send_close_notify();
-            sent_shutdown = true;
-        }
-
         if quench_writes && len > 0 {
             println!("unquenching writes after {len:?}");
             quench_writes = false;
@@ -432,6 +432,9 @@ fn read_all_bytes(opts: &Options, sess: &mut impl Connection, conn: &mut net::Tc
     match sess.read_tls(conn) {
         Ok(_) => {}
         Err(err) if err.kind() == io::ErrorKind::ConnectionReset => {}
+        Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => {
+            quit_err(":CLOSE_WITHOUT_CLOSE_NOTIFY:")
+        }
         Err(err) => panic!("invalid read: {err}"),
     };
 
diff --git a/rustls-test/src/lib.rs b/rustls-test/src/lib.rs
index 1d192cd8f..11ec9f185 100644
--- a/rustls-test/src/lib.rs
+++ b/rustls-test/src/lib.rs
@@ -227,11 +227,12 @@ pub fn transfer(left: &mut impl Connection, right: &mut impl Connection) -> usiz
     total
 }
 
-pub fn transfer_eof(conn: &mut impl Connection) {
+pub fn transfer_eof(conn: &mut impl Connection) -> Result<(), io::Error> {
     let empty_buf = [0u8; 0];
     let empty_cursor: &mut dyn io::Read = &mut &empty_buf[..];
-    let sz = conn.read_tls(empty_cursor).unwrap();
+    let sz = conn.read_tls(empty_cursor)?;
     assert_eq!(sz, 0);
+    Ok(())
 }
 
 pub enum Altered {
@@ -1672,7 +1673,10 @@ impl<'a, C: Connection> OtherSession<'a, C> {
 impl<C: Connection> io::Read for OtherSession<'_, C> {
     fn read(&mut self, mut b: &mut [u8]) -> io::Result<usize> {
         self.reads += 1;
-        self.sess.write_tls(&mut b)
+        match &self.last_error {
+            None => self.sess.write_tls(&mut b),
+            Some(err) => Err(io::Error::new(io::ErrorKind::InvalidData, err.clone())),
+        }
     }
 }
 
diff --git a/rustls-test/tests/api/api.rs b/rustls-test/tests/api/api.rs
index fa336d5d3..a4d22ff5e 100644
--- a/rustls-test/tests/api/api.rs
+++ b/rustls-test/tests/api/api.rs
@@ -1552,10 +1552,15 @@ fn large_client_hello() {
     let hello = include_bytes!("../data/bug2227-clienthello.bin");
     let mut cursor = io::Cursor::new(hello);
     loop {
-        if server.read_tls(&mut cursor).unwrap() == 0 {
-            break;
+        match server.read_tls(&mut cursor) {
+            Ok(0) => break, // Clean EOF (after close_notify)
+            Ok(_) => {
+                server.process_new_packets().unwrap();
+            }
+            // Reached end of test data (no close_notify expected in test data)
+            Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => break,
+            Err(e) => panic!("Unexpected error: {e}"),
         }
-        server.process_new_packets().unwrap();
     }
 }
 
diff --git a/rustls-test/tests/api/io.rs b/rustls-test/tests/api/io.rs
index 6dfd2fd6d..028a00390 100644
--- a/rustls-test/tests/api/io.rs
+++ b/rustls-test/tests/api/io.rs
@@ -844,7 +844,6 @@ fn test_client_stream_read(stream_kind: StreamKind, read_kind: ReadKind) {
 
         {
             let mut pipe = OtherSession::new(&mut server);
-            transfer_eof(&mut client);
 
             let stream: Box<dyn BufRead> = match stream_kind {
                 StreamKind::Ref => Box::new(Stream::new(&mut client, &mut pipe)),
@@ -865,7 +864,6 @@ fn test_server_stream_read(stream_kind: StreamKind, read_kind: ReadKind) {
 
         {
             let mut pipe = OtherSession::new(&mut client);
-            transfer_eof(&mut server);
 
             let stream: Box<dyn BufRead> = match stream_kind {
                 StreamKind::Ref => Box::new(Stream::new(&mut server, &mut pipe)),
@@ -980,13 +978,13 @@ fn client_stream_handshake_error() {
         assert!(rc.is_err());
         assert_eq!(
             format!("{rc:?}"),
-            "Err(Custom { kind: InvalidData, error: AlertReceived(HandshakeFailure) })"
+            "Err(Custom { kind: InvalidData, error: PeerIncompatible(NoCipherSuitesInCommon) })"
         );
         let rc = client_stream.write(b"hello");
         assert!(rc.is_err());
         assert_eq!(
             format!("{rc:?}"),
-            "Err(Custom { kind: InvalidData, error: AlertReceived(HandshakeFailure) })"
+            "Err(Custom { kind: InvalidData, error: PeerIncompatible(NoCipherSuitesInCommon) })"
         );
     }
 }
@@ -1002,13 +1000,13 @@ fn client_streamowned_handshake_error() {
     assert!(rc.is_err());
     assert_eq!(
         format!("{rc:?}"),
-        "Err(Custom { kind: InvalidData, error: AlertReceived(HandshakeFailure) })"
+        "Err(Custom { kind: InvalidData, error: PeerIncompatible(NoCipherSuitesInCommon) })"
     );
     let rc = client_stream.write(b"hello");
     assert!(rc.is_err());
     assert_eq!(
         format!("{rc:?}"),
-        "Err(Custom { kind: InvalidData, error: AlertReceived(HandshakeFailure) })"
+        "Err(Custom { kind: InvalidData, error: PeerIncompatible(NoCipherSuitesInCommon) })"
     );
 
     let (_, _) = client_stream.into_parts();
@@ -1799,16 +1797,12 @@ fn server_closes_uncleanly() {
         );
 
         transfer(&mut server, &mut client);
-        transfer_eof(&mut client);
+        let err = transfer_eof(&mut client).unwrap_err();
+        assert!(matches!(err, err if err.kind() == io::ErrorKind::UnexpectedEof));
         let io_state = client.process_new_packets().unwrap();
         assert!(!io_state.peer_has_closed());
         check_read(&mut client.reader(), b"from-server!");
 
-        check_read_err(
-            &mut client.reader() as &mut dyn Read,
-            io::ErrorKind::UnexpectedEof,
-        );
-
         // may still transmit pending frames
         transfer(&mut client, &mut server);
         server.process_new_packets().unwrap();
@@ -1845,16 +1839,12 @@ fn client_closes_uncleanly() {
         );
 
         transfer(&mut client, &mut server);
-        transfer_eof(&mut server);
+        let err = transfer_eof(&mut server).unwrap_err();
+        assert!(matches!(err, err if err.kind() == io::ErrorKind::UnexpectedEof));
         let io_state = server.process_new_packets().unwrap();
         assert!(!io_state.peer_has_closed());
         check_read(&mut server.reader(), b"from-client!");
 
-        check_read_err(
-            &mut server.reader() as &mut dyn Read,
-            io::ErrorKind::UnexpectedEof,
-        );
-
         // may still transmit pending frames
         transfer(&mut server, &mut client);
         client.process_new_packets().unwrap();
diff --git a/rustls/src/conn/mod.rs b/rustls/src/conn/mod.rs
index 1600d5168..243a76232 100644
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -248,23 +248,17 @@ mod connection {
     pub struct Reader<'a> {
         pub(super) received_plaintext: &'a mut ChunkVecBuffer,
         pub(super) has_received_close_notify: bool,
-        pub(super) has_seen_eof: bool,
     }
 
     impl<'a> Reader<'a> {
         /// Check the connection's state if no bytes are available for reading.
         fn check_no_bytes_state(&self) -> io::Result<()> {
-            match (self.has_received_close_notify, self.has_seen_eof) {
+            match self.has_received_close_notify {
                 // cleanly closed; don't care about TCP EOF: express this as Ok(0)
-                (true, _) => Ok(()),
-                // unclean closure
-                (false, true) => Err(io::Error::new(
-                    io::ErrorKind::UnexpectedEof,
-                    UNEXPECTED_EOF_MESSAGE,
-                )),
+                true => Ok(()),
                 // connection still going, but needs more data: signal `WouldBlock` so that
                 // the caller knows this
-                (false, false) => Err(io::ErrorKind::WouldBlock.into()),
+                false => Err(io::ErrorKind::WouldBlock.into()),
             }
         }
 
@@ -339,9 +333,6 @@ mod connection {
         }
     }
 
-    const UNEXPECTED_EOF_MESSAGE: &str = "peer closed connection without sending TLS close_notify: \
-https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof";
-
     /// A structure that implements [`std::io::Write`] for writing plaintext.
     pub struct Writer<'a> {
         sink: &'a mut dyn PlaintextSink,
@@ -528,7 +519,6 @@ pub(crate) struct ConnectionCommon<Side: SideData> {
     deframer_buffer: VecBuffer,
     pub(crate) received_plaintext: ChunkVecBuffer,
     pub(crate) sendable_plaintext: ChunkVecBuffer,
-    pub(crate) has_seen_eof: bool,
     pub(crate) fips: FipsStatus,
 }
 
@@ -539,7 +529,6 @@ impl<Side: SideData> ConnectionCommon<Side> {
             deframer_buffer: VecBuffer::default(),
             received_plaintext: ChunkVecBuffer::new(Some(DEFAULT_RECEIVED_PLAINTEXT_LIMIT)),
             sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
-            has_seen_eof: false,
             fips,
         }
     }
@@ -643,7 +632,6 @@ impl<Side: SideData> ConnectionCommon<Side> {
             // Are we done? i.e., have we processed all received messages, and received a
             // close_notify to indicate that no new messages will arrive?
             has_received_close_notify,
-            has_seen_eof: self.has_seen_eof,
         }
     }
 
@@ -663,7 +651,10 @@ impl<Side: SideData> ConnectionCommon<Side> {
 
         let res = self.deframer_buffer.read(rd);
         if let Ok(0) = res {
-            self.has_seen_eof = true;
+            return Err(io::Error::new(
+                io::ErrorKind::UnexpectedEof,
+                UNEXPECTED_EOF_MESSAGE,
+            ));
         }
         res
     }
@@ -1094,4 +1085,7 @@ pub(crate) trait StateMachine: Sized {
     ) -> Result<(PartiallyExtractedSecrets, Box<dyn KernelState + 'static>), Error>;
 }
 
+const UNEXPECTED_EOF_MESSAGE: &str = "peer closed connection without sending TLS close_notify: \
+https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof";
+
 const DEFAULT_RECEIVED_PLAINTEXT_LIMIT: usize = 16 * 1024;
```

This still incurs a bunch of bogo failures that I cannot (yet) quite wrap my head around. 

Is this the direction you had in mind? Since it seems fairly subtle, I still feel like shrinking the `TlsInputBuffer` trait after this PR has landed could make more sense -- since we can dismantle some of the relevant `Connection` APIs first (like dropping the `Reader` completely in favor of yielding an iterator of plaintext chunks from `process_new_packets()`).

djc: (Also, since you have #959 fixes presumably by moving close notify into the state machine, it might not make sense to tweak it too much here?)

ctz: Yeah in the new API you entirely lose the ability to feed in new received bytes after a close_notify. As a result of that, the buffered API layered on top of that simply stops reading (which is a change in behaviour, as before it would read but discard everything.)

> Currently `UnexpectedEof` errors are delayed -- they're not returned from `read_tls()` but from the `Reader`. I think we could fix that but it seems to be pretty involved.

The purpose of that delay/ordering is to avoid truncating the plaintext read side (which would be a risk if the error is raised before all pending plaintext has been read), and put the errors where the application is best placed to handle them:

Whether the truncation is a problem is really a property of the application protocol, so it should be delivered in-band with that stream. As a concrete example, in HTTP0.9 the truncation is always bad. In HTTP1.1 it's sometimes OK, depending on protocol state. In HTTP2 it's always fine, because there is an application-level closure message that works around TLS libraries having rubbish support for `close_notify`. So, putting the error in the `read_tls` layer means communicating that up to to the next layer in a different way, outside the library.

Sorry, that's a bit of a wall of text. I think I'd like to delay any larger public API changes here until #2953 is landed, as I've put a significant amount of work into ensuring the new API can be used to implement the old API as-is. But maybe it makes sense to land [efa92aa](https://github.com/rustls/rustls/pull/2940/commits/efa92aa6052f825b8ee3d092ddc333388c7c0701) even before that?

djc: > I think I'd like to delay any larger public API changes here until #2953 is landed, as I've put a significant amount of work into ensuring the new API can be used to implement the old API as-is.

Okay, I'll hold off for a bit. I do think the digesting the change to externalize the buffer soon will be helpful.

ci-bench/src/main.rs
```diff
@@ -237,6 +237,7 @@ fn main() -> anyhow::Result<()> {
                                     &bench.params,
                                     resumption_kind,
                                 ),
+                                buf: VecInput::default(),
```
ctz: i suggest we adopt a house style that values of this type are called `input` rather than `buf`, to discern later externalised output buffers?

djc: Done.

rustls/src/conn/mod.rs
djc: So getting back to this thread, I'm not exactly clear on how to improve on the current state -- and if those are invasive changes, I think it might be better to get this large (and largely mechanical) change merged and iterate more on the error/EOF handling separately.

Thoughts?

ctz: Yes, that sounds like a fair way forward.

djc:
Okay, just waiting for review then. Not sure why the benchmark data isn't getting updated...

ctz (approved):

PR #2942, opened 2026-02-25
Adjust --release profile for benchmarking

ctz:
Turns on LTO for `--release`, which should affect the CI benchmarks somewhat, at the expense of slower `--release` builds. That effect isn't transitive onto our dependents.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -249,13 +249,13 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
 
       - name: Smoke-test benchmark program (ring)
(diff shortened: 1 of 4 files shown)


djc:
Why not the other way around?

ctz:
> Why not the other way around?

Do you mean why not change the CI benchmark runner to use `--profile=bench`?

djc:
> > Why not the other way around?
> 
> Do you mean why not change the CI benchmark runner to use `--profile=bench`?

Yes.

ctz:
Will do that instead. Will mean some unexplained edges on the historical graphs (ie, not tied to a commit in this repo).

PR #2945, opened 2026-02-27
Model state machine as enum; expose its type internally

ctz:
Importantly, this withdraws the test `limited_no_tls12_symbols`.

I think it can come back later -- when we've done everything needed to address items in #850 -- by extracting a trait over all operations that expose the innards of (eg) `StateMachine::Tls12()` -- ie it would become `StateMachine::Tls12(Box<dyn TypeErased>)`

### DIFF
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -518,7 +518,7 @@ impl ExpectClientHello {
 
     fn with_version<T: Suite + 'static>(
         mut self,
-        mut input: ClientHelloInput<'_>,
+        input: ClientHelloInput<'_>,
         output: &mut dyn Output,
     ) -> Result<ServerState, Error>
     where
@@ -560,23 +560,27 @@ impl ExpectClientHello {
             })
             .collect::<Vec<_>>();
 
+        let mut sig_schemes = input.sig_schemes.clone();
         if T::VERSION == ProtocolVersion::TLSv1_2 {
-            input.sig_schemes.retain(|scheme| {
+            sig_schemes.retain(|scheme| {
                 client_suites
                     .iter()
                     .any(|&suite| suite.usable_for_signature_scheme(*scheme))
             });
         } else if T::VERSION == ProtocolVersion::TLSv1_3 {
-            input
-                .sig_schemes
-                .retain(SignatureScheme::supported_in_tls13);
+            sig_schemes.retain(SignatureScheme::supported_in_tls13);
         }
 
         // Choose a certificate.
         let credentials = self
             .config
             .cert_resolver
-            .resolve(&ClientHello::new(&input, sni.as_ref(), T::VERSION))?;
+            .resolve(&ClientHello::new(
+                &input,
+                &sig_schemes,
+                sni.as_ref(),
+                T::VERSION,
+            ))?;
         self.sni = sni;
 
         let (suite, skxg) = self.choose_suite_and_kx_group(
@@ -757,7 +761,7 @@ pub(crate) trait ServerHandler<T>: fmt::Debug + Sealed + Send + Sync {
 pub(crate) struct ClientHelloInput<'a> {
     pub(super) message: &'a Message<'a>,
     pub(super) client_hello: &'a ClientHelloPayload,
-    pub(super) sig_schemes: Vec<SignatureScheme>,
+    pub(super) sig_schemes: &'a Vec<SignatureScheme>,
     pub(super) proof: HandshakeAlignedProof,
 }
 
@@ -795,7 +799,7 @@ impl<'a> ClientHelloInput<'a> {
         Ok(ClientHelloInput {
             message: &input.message,
             client_hello,
-            sig_schemes: sig_schemes.to_owned(),
+            sig_schemes,
             proof,
         })
     }
--- a/rustls/src/server/config.rs
+++ b/rustls/src/server/config.rs
@@ -405,12 +405,13 @@ pub struct ClientHello<'a> {
 impl<'a> ClientHello<'a> {
     pub(super) fn new(
         input: &'a ClientHelloInput<'a>,
+        signature_schemes: &'a [SignatureScheme],
         sni: Option<&'a DnsName<'static>>,
         version: ProtocolVersion,
     ) -> Self {
         Self {
             server_name: sni.map(Cow::Borrowed),
-            signature_schemes: &input.sig_schemes,
+            signature_schemes,
             alpn: input.client_hello.protocols.as_ref(),
             server_cert_types: input
                 .client_hello


rustls/src/conn/mod.rs
```diff
@@ -954,7 +954,10 @@ pub trait SideData: private::SideData {}
 pub(crate) mod private {
     use super::*;
 
-    pub(crate) trait SideData: Output + Debug {}
+    pub(crate) trait SideData: Debug {
```
djc: Not a fan of keeping the same name for this as for the public trait.

rustls/src/client/connection.rs
```diff
@@ -293,7 +293,7 @@ impl ConnectionCore<ClientSide> {
             .send
             .set_max_fragment_size(config.max_fragment_size)?;
         common_state.fips = config.fips();
-        let mut data = ClientSide::new();
+        let mut data = ClientConnectionData::new();
```
djc, line 303: The first two commits seem like a neat implementation strategy, but I'm not sure keeping them separate makes sense for the future/review? Seems like a lot of flipflopping.

rustls/src/client/hs.rs
```diff
@@ -31,13 +33,54 @@ use crate::msgs::{
...
 
+#[expect(private_interfaces)]
+pub(crate) enum StateMachine {
+    ExpectServerHello(Box<ExpectServerHello>),
+    ExpectServerHelloOrHelloRetryRequest(Box<ExpectServerHelloOrHelloRetryRequest>),
+    Tls12(tls12::StateMachine),
+    Tls13(tls13::StateMachine),
+}
+
+impl StateMachine {}
```
djc: Drop this?

rustls/src/client/hs.rs
```diff
@@ -31,13 +33,54 @@ use crate::msgs::{
...
+use crate::suites::{PartiallyExtractedSecrets, Suite, SupportedCipherSuite};
 use crate::sync::Arc;
 use crate::tls12::Tls12CipherSuite;
 use crate::tls13::Tls13CipherSuite;
-use crate::tls13::key_schedule::KeyScheduleEarlyClient;
+use crate::tls13::key_schedule::{KeyScheduleEarlyClient, KeyScheduleTrafficSend};
 use crate::{ClientConfig, bs_debug};
 
+#[expect(private_interfaces)]
+pub(crate) enum StateMachine {
```
djc: Don't love `StateMachine` name. How about `ClientState`?

rustls/src/client/tls12.rs
```diff
@@ -33,6 +33,39 @@ use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
 use crate::tls13::key_schedule::KeyScheduleTrafficSend;
 use crate::verify::{self, DigitallySignedStruct, ServerIdentity, SignatureVerificationInput};
 
+#[expect(clippy::enum_variant_names, private_interfaces)]
+pub(crate) enum StateMachine {
+    ExpectCertificate(Box<ExpectCertificate>),
```
djc: I guess `enum_variant_names` was complaining about the shared `Expect` prefixes? I think ditching those is a good idea...

rustls/src/client/tls12.rs
```diff
@@ -54,7 +87,7 @@ mod server_hello {
             Input { message, .. }: &Input<'_>,
             st: ExpectServerHello,
             output: &mut dyn Output,
-        ) -> Result<Box<dyn State>, Error> {
+        ) -> Result<hs::StateMachine, Error> {
```
djc: Nit: suggest dropping the `hs::` prefix everywhere.

rustls/src/client/tls12.rs
```diff
@@ -33,6 +33,39 @@ use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
...
+pub(crate) enum StateMachine {
+    ExpectCertificate(Box<ExpectCertificate>),
+    ExpectCertificateStatusOrServerKx(Box<ExpectCertificateStatusOrServerKx>),
+    ExpectServerKx(Box<ExpectServerKx>),
+    ExpectServerDoneOrCertReq(Box<ExpectServerDoneOrCertReq>),
+    ExpectServerDone(Box<ExpectServerDone>),
+    ExpectNewTicket(Box<ExpectNewTicket>),
+    ExpectCcs(Box<ExpectCcs>),
+    ExpectFinished(Box<ExpectFinished>),
+    ExpectTraffic(Box<ExpectTraffic>),
```
djc: Maybe add a commit that drops the `Box` for `ExpectTraffic`? Could be nice.

rustls/src/conn/mod.rs
```diff
@@ -957,7 +959,18 @@ pub(crate) mod private {
     pub(crate) trait SideData: Debug {
         /// Data storage type.
         type Data: Debug + Output;
+        /// State machine type.
+        type StateMachine: StateMachine;
```
djc: IMO this type should just be called `State`.

rustls/src/vecbuf.rs
```diff
@@ -92,6 +92,20 @@ impl ChunkVecBuffer {
...
+            1 => self.chunks.pop_front().unwrap(),
+            _ => {
+                let mut v = Vec::with_capacity(self.len());
+                while let Some(c) = self.chunks.pop_front() {
+                    v.extend_from_slice(&c);
+                }
+                v
+            }
+        }
+    }
```
djc, line 106: This seems unrelated to `ErrorWithAlert`?

rustls/src/error/mod.rs
```diff
@@ -1539,3 +1541,58 @@ mod other_error {
 }
 
 pub use other_error::OtherError;
+
+/// An [`Error`] along with the (possibly encrypted) alert to send to
```
djc, line 1545: Suggest squashing this with the next commit to make clear how it's used.

rustls/src/client/connection.rs
```diff
@@ -211,14 +215,15 @@ impl ClientConnectionBuilder {
...
-        Ok(ClientConnection {
-            inner: ConnectionCommon::from(ConnectionCore::for_client(
-                config,
-                name,
-                ClientExtensionsInput::from_alpn(alpn_protocols),
-                Protocol::Tcp,
-            )?),
-        })
+        let fips = config.fips();
+        let mut inner = ConnectionCommon::from(ConnectionCore::for_client(
```
djc: Maybe replace the `ConnectionCommon` `From` impl with a `new()` constructor that takes `FipsStatus` directly? Otherwise you could forget to overwrite the default `fips` value.

rustls/src/server/connection.rs
```diff
@@ -434,35 +411,41 @@ impl io::Read for ReadEarlyData<'_> {
 ///
 /// Contains the state required to resume the connection through [`Accepted::into_connection()`].
 pub struct Accepted {
+    // invariant: `connection.core.state` is `Ok(hs::StateMachine::ChooseConfig)`
```
djc: Could we just store `ChooseConfig` directly? If not, why?

rustls/src/server/connection.rs
```diff
@@ -620,11 +585,25 @@ impl ConnectionCore<ServerSide> {
...
+                Vec::new(),
+                protocol,
+            ))
+            .into(),
             ServerConnectionData::default(),
             common,
         ))
     }
+
+    pub(crate) fn for_server_acceptor(protocol: Protocol) -> Self {
```
djc: Nit: maybe drop `_server` from this name? It doesn't seem useful.

rustls/src/server/hs.rs
```diff
@@ -341,6 +351,87 @@ pub(super) struct CertificateTypes {
...
+            resumption_data: Vec::new(),
+        }
+    }
+
+    fn set_resumption_data(&mut self, resumption_data: &[u8]) -> Result<(), Error> {
+        self.resumption_data = resumption_data.to_vec();
+        Ok(())
+    }
+
+    pub(crate) fn handle<'m>(
```
djc, line 375: Nit: `handle()` before `set_resumption_data()`.

rustls/src/client/tls12.rs
ctz: Did that, but left the struct names as-is (because `Certificate` is a type name is less descriptive)

rustls/src/vecbuf.rs
ctz: Called from the `ErrorWithAlert::new` ctor

djc:
The benchmarks are pretty unhappy...

rustls/src/conn/mod.rs
```diff
@@ -946,12 +946,15 @@ impl Output for SideCommonOutput<'_> {
...
-pub trait SideData: private::SideData {}
+pub trait SideData: private::Side {}
 
 pub(crate) mod private {
     use super::*;
 
-    pub(crate) trait SideData: Output + Debug {}
+    pub(crate) trait Side: Debug {
+        /// Data storage type.
+        type Data: Debug + Output;
```
djc: Tiny nit: I like going from rarer to more common, so the previous `Output + Debug` seems preferable to me.

rustls/src/client/tls12.rs
```diff
@@ -33,10 +34,43 @@ use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
 use crate::tls13::key_schedule::KeyScheduleTrafficSend;
 use crate::verify::{self, DigitallySignedStruct, ServerIdentity, SignatureVerificationInput};
 
+#[expect(private_interfaces)]
+pub(crate) enum StateMachine {
```
djc: Maybe just call this `Tls12State`, then?

rustls/src/vecbuf.rs
```diff
@@ -92,6 +92,20 @@ impl ChunkVecBuffer {
...
+            0 => Vec::new(),
+            1 => self.chunks.pop_front().unwrap(),
+            _ => {
+                let mut v = Vec::with_capacity(self.len());
+                while let Some(c) = self.chunks.pop_front() {
+                    v.extend_from_slice(&c);
+                }
+                v
+            }
+        }
```
djc: This seems a little ugly? I think there's a way to do this neatly (that is, without `.len()` -> `unwrap()`.

rustls/src/conn/mod.rs
```diff
@@ -232,6 +235,13 @@ mod connection {
         ///
         /// [`Connection::process_new_packets()`]: crate::Connection::process_new_packets
         fn is_handshaking(&self) -> bool;
+
+        /// Return the FIPS validation status of the connection.
+        ///
+        /// This is different from [`crate::crypto::CryptoProvider::fips()`]:
```
djc: Nit: hide the `crate::crypto` prefix from the displayed text?

rustls/src/server/connection.rs
```diff
@@ -319,26 +311,25 @@ impl Acceptor {
...
-            Ok(None) => {
-                self.inner = Some(connection);
-                return Ok(None);
-            }
-            Err(err) => return Err(AcceptedAlert::from_error(err, connection.core.common.send)),
+        if let Err(e) = connection.process_new_packets() {
+            return Err(AcceptedAlert::from_error(e, connection.core.common.send));
+        }
+
+        let Ok(hs::ServerState::ChooseConfig(_)) = connection.core.state else {
```
djc: Nit: drop the `hs::` prefix?

rustls/src/server/connection.rs
```diff
@@ -319,26 +311,25 @@ impl Acceptor {
...
-        let sig_schemes = match ClientHelloInput::from_input(&input) {
-            Ok(ClientHelloInput { sig_schemes, .. }) => sig_schemes,
-            Err(err) => {
-                return Err(AcceptedAlert::from_error(err, connection.core.common.send));
-            }
+        let Ok(hs::ServerState::ChooseConfig(choose_config)) = core::mem::replace(
+            &mut connection.core.state,
+            Err(Error::Unreachable("Accepted misused state")),
+        ) else {
+            unreachable!();
```
djc: Why is this unreachable? Maybe add a comment?

rustls/src/server/hs.rs
```diff
@@ -31,7 +31,8 @@ use crate::tls13::Tls13CipherSuite;
 use crate::tls13::key_schedule::KeyScheduleTrafficSend;
 
 pub(crate) enum ServerState {
-    Accepting(super::connection::Accepting),
+    ReadClientHello(ReadClientHello),
```
djc, line 38: Since these states are somewhat unlike the other ones, maybe add some short comments describing their intended use?

rustls/src/client/tls12.rs
ctz: A lot of the performance issues here are due to bloated type sizes, so this can't/shouldn't be done generically.

PR #2949, opened 2026-02-27
Improve Output type safety

djc:
Breaks down some of the more dynamic aspects of the `Output` trait setup to bring back more type safety, hopefully without compromising too much on modularity.

Also less code, and up to 6.5% wins on no crypto benchmarks.

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -75,38 +75,6 @@ impl CommonState {
     }
 }
 
-impl Output for CommonState {
-    fn emit(&mut self, _: Event<'_>) {
-        unreachable!();
-    }
-
-    fn output(&mut self, ev: OutputEvent<'_>) {
-        if let OutputEvent::ProtocolVersion(ver) = ev {
-            self.recv.negotiated_version = Some(ver);
-            self.send.negotiated_version = Some(ver);
-        }
-
-        self.outputs.handle(ev);
-    }
-
-    fn send_msg(&mut self, msg: Message<'_>, must_encrypt: bool) {
-        self.send.send_msg(msg, must_encrypt);
-    }
-
-    fn start_traffic(&mut self) {
-        self.recv.may_receive_application_data = true;
-        self.send.start_outgoing_traffic();
-    }
-
-    fn receive(&mut self) -> &mut ReceivePath {
-        &mut self.recv
-    }
-
-    fn send(&mut self) -> &mut dyn SendOutput {
-        &mut self.send
-    }
-}
-
 impl Deref for CommonState {
     type Target = ConnectionOutputs;
 
@@ -142,7 +110,7 @@ pub struct ConnectionOutputs {
 }
 
 impl ConnectionOutputs {
-    fn handle(&mut self, ev: OutputEvent<'_>) {
+    pub(crate) fn handle(&mut self, ev: OutputEvent<'_>) {
         match ev {
             OutputEvent::ApplicationProtocol(protocol) => {
                 self.alpn_protocol = Some(ApplicationProtocol::from(protocol.as_ref()).to_owned())
@@ -650,7 +618,7 @@ impl Default for SendPath {
 pub(crate) struct ReceivePath {
     side: Side,
     pub(crate) decrypt_state: DecryptionState,
-    may_receive_application_data: bool,
+    pub(crate) may_receive_application_data: bool,
     /// If the peer has signaled end of stream.
     pub(crate) has_received_close_notify: bool,
     temper_counters: TemperCounters,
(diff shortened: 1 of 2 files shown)


rustls-test/tests/api/quic.rs
```diff
@@ -766,6 +766,7 @@ fn packet_key_api() {
 fn test_quic_exporter() {
     let provider = provider::DEFAULT_TLS13_PROVIDER;
     for &kt in KeyType::all_for_provider(&provider) {
+        dbg!(kt);
```
ctz: intended/useful?

rustls/src/tls13/key_schedule.rs
```diff
@@ -134,12 +134,10 @@ impl KeyScheduleEarly {
...
         );
 
-        if self.ks.protocol.is_quic() {
+        if let Some(quic) = output.quic() {
             // If 0-RTT should be rejected, this will be clobbered by ExtensionProcessing
             // before the application can see.
-            output.emit(Event::QuicEarlySecret(Some(
-                client_early_traffic_secret.clone(),
-            )));
+            quic.early_secret = Some(client_early_traffic_secret.clone());
```
ctz: Would be preferable for testability, I think, to have this access be via a trait, and be write-only. This also avoids the previous mire we were in where random other types were used for storage of values.

rustls/src/tls13/key_schedule.rs
```diff
@@ -344,15 +342,15 @@ impl KeyScheduleHandshakeStart {
...
+            quic.hs_secrets = Some(quic::Secrets::new(
                 client_secret.clone(),
                 server_secret.clone(),
                 self.ks.suite,
                 self.ks.suite.quic.unwrap(),
                 self.ks.side,
-                quic_version,
-            )));
+                quic.version,
+            ));
```
ctz: Example of a two-way dependency, where the value output to `quic.hs_secrets` depends on the value read via `output.quic()` rather than the immutable version known by this object.

rustls/src/common_state.rs
```diff
@@ -1229,6 +1252,8 @@ pub(crate) trait Output {
     fn received_plaintext(&mut self, _payload: Payload<'_>) {}
 
     fn start_traffic(&mut self);
+
+    fn send(&mut self) -> &mut SendPath;
```
ctz: Suggest extracting a trait here rather than using the concrete `SendPath` for testability? All methods on the trait should return unit.

rustls/src/common_state.rs
```diff
@@ -908,7 +908,7 @@ impl ReceivePath {
     fn reject_renegotiation_request(
         &mut self,
         msg: &Message<'_>,
-        output: &mut dyn Output,
+        send: &mut SendPath,
```
ctz: This is not possible later, as the caller no longer has a `SendPath` to give. It has an `impl Output {}` which locks a Mutex and dispatches events into it.

djc:
Okay, so I should rework this to have specific traits for both `SendPath` and `Quic`?

rustls/src/tls13/key_schedule.rs
djc: Remind me why this useful/necessary? Should we document it somewhere?

(I agree that it was very messy before, but at a much smaller scale I don't see many issues?)

rustls/src/common_state.rs
djc: Why should they all yield unit?

rustls/src/tls13/key_schedule.rs
ctz: It basically means to test a given piece of code, several others are needed at the same time. Whereas if a dependency is output-only then it cannot have the same impact.

rustls/src/common_state.rs
ctz: Same reason as above -- to ensure the dependency is output-only.

ctz:
> Okay, so I should rework this to have specific traits for both `SendPath` and `Quic`?

I think that would be good!

rustls-test/tests/api/quic.rs
djc: Not intended, removed.

rustls/src/tls13/key_schedule.rs
djc: Introduced a `QuicOutput` trait like this.

rustls/src/tls13/key_schedule.rs
djc: Handled by the `QuicOutput` trait.

rustls/src/common_state.rs
djc: Extracted a `SendOutput` trait.

rustls/src/common_state.rs
djc: This should be solved with the `SendOutput` trait.

djc:
Second commit broke a few QUIC tests. Going to look into that some other time.

PR #2953, opened 2026-03-02
Introduce new state-based API, and move buffered APIs to use it

### LINKED ISSUE #959: Do we do the right thing when there's data after a `close_notify` alert?

### LINKED ISSUE #1723: reduce bifurcation of library internals by using low-level API in high-level API

### LINKED ISSUE #2397: Evolution of the TLS connection state machine

### LINKED ISSUE #2761: Consider a reader-writer split for the unbuffered API

### LINKED ISSUE #3047: Confusing behavior for `send_close_notify()`
I have some rustls code like this:
```rs
client_session.send_close_notify();
assert!(client_session.wants_write());
while client_session.wants_write() {
    client_session.write_tls(&mut buf).unwrap();
}


assert!(client_session.writer().write(b"rawr!!!!").unwrap() > 0);
assert!(client_session.wants_write());
panic!("I shouldn't hit this.");
```

yields:

```
running 1 test
test tls_hello_world ... 
thread 'tls_hello_world' (8925) panicked at tests/tls.rs:173:5:
I shouldn't hit this.
FAILED

failures:

failures:
    tls_hello_world

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.03s
```

I'm not sure why but I expected writing plaintext to fail after writing the close_notify alert on the underlying TLS stream. Every LLM is also convinced that rustls should reject writing more application data as well, which has been a frustrating debugging experience.

I peeked some of the source code and it seems like nothing guards against writing extra application data once the close_notify has been written.

I guess my question is, is this expected behavior? If I'm using rustls to implement my own I/O-based TlsStreams, what kind of behavior should I expose to users?

ctz:
This is still quite messy, and hence draft. Things to do:

- [x] extract more run-up commits from the "introduce new state-based API" commit
- [x] ensure the overally shape of the new API is uniform between client and server (currently not, around the `ServerOutputs` type

fixes #2397 fixes #1723 fixes #959 fixes #2795 fixes #2698 fixes #2761 fixes #3047

### DIFF
--- a/examples/src/bin/async-client.rs
+++ b/examples/src/bin/async-client.rs
@@ -1,0 +1,164 @@
+//! This is a simple client using rustls' state-based API, in conjunction with asynchronous I/O via
+//! tokio.  The application manages its own receive buffer.
+
+use core::error::Error;
+use std::sync::Arc;
+
+use rustls::client::{ClientHandshake, ClientTraffic};
+use rustls::crypto::cipher::OutboundPlain;
+use rustls::state::{ReceiveTrafficState, SliceInput, TlsInputBuffer};
+use rustls::{ClientConfig, RootCertStore};
+use rustls_aws_lc_rs::DEFAULT_PROVIDER;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpStream;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let root_store = RootCertStore {
+        roots: webpki_roots::TLS_SERVER_ROOTS.into(),
+    };
+
+    let config = ClientConfig::builder(Arc::new(DEFAULT_PROVIDER))
+        .with_root_certificates(root_store)
+        .with_no_client_auth()?;
+
+    let config = Arc::new(config);
+
+    let mut incoming_tls = vec![0; INCOMING_TLS_BUFSIZE];
+
+    let mut state = config
+        .connect(SERVER_NAME.try_into()?)
+        .start_handshake()?;
+    let mut sock = TcpStream::connect(format!("{SERVER_NAME}:{PORT}")).await?;
+
+    let mut incoming_used = 0;
+
+    // complete handshake
+    let traffic = loop {
+        state = match state {
+            ClientHandshake::SendClientFlight(mut send) => {
+                while let Some(data) = send.take_data() {
+                    sock.write_all(&data).await?;
+                }
+                send.into_next()
+            }
+
+            ClientHandshake::AwaitServerFlight(recv) => {
+                recv_tls(&mut sock, &mut incoming_tls, &mut incoming_used).await?;
+                let mut buffer = SliceInput::new(&mut incoming_tls[..incoming_used]);
+                let next = match recv.input_data(&mut buffer) {
+                    Ok(next) => next,
+                    Err((mut err_alert, _outputs)) => {
+                        while let Some(data) = err_alert.take_tls_data() {
+                            sock.write_all(&data).await?;
+                        }
+                        return Err(err_alert.error.into());
+                    }
+                };
+
+                let used = buffer.into_used();
+                // note, this is a very inefficient buffer representation.
+                incoming_tls.copy_within(used..incoming_used, 0);
+                incoming_used -= used;
+                next
+            }
+
+            ClientHandshake::Complete(traffic) => break traffic,
+
+            _ => todo!(),
+        };
+    };
+
+    let ClientTraffic {
+        mut send,
+        mut receive,
+        outputs,
+    } = *traffic;
+
+    println!(
+        "connected to server with version={:?} cipher suite={:?}",
+        outputs.protocol_version(),
+        outputs.negotiated_cipher_suite()
+    );
+
+    let request = [
+        b"GET / HTTP/1.1\r\n\
+        Host: ",
+        SERVER_NAME.as_bytes(),
+        b"\r\n\
+        Connection: close\r\n\
+        Accept-Encoding: identity\r\n\
+        \r\n",
+    ];
+    let mut tls_data = send.write(OutboundPlain::new(&request))?;
+    tls_data.push(send.close());
+    for chunk in tls_data {
+        sock.write_all(&chunk).await?;
+    }
+
+    loop {
+        let mut buffer = SliceInput::new(&mut incoming_tls[..incoming_used]);
+        let received = match receive.read(&mut buffer) {
+            Ok(received) => received,
+            Err(mut err_alert) => {
+                while let Some(data) = err_alert.take_tls_data() {
+                    sock.write_all(&data).await?;
+                }
+                return Err(err_alert.error.into());
+            }
+        };
+
+        receive = match received {
+            ReceiveTrafficState::Await(receive) => {
+                if recv_tls(&mut sock, &mut incoming_tls, &mut incoming_used).await? == 0 {
+                    println!("server uncleanly closed connection");
+                    break;
+                }
+                receive
+            }
+
+            ReceiveTrafficState::WakeSender(wake) => {
+                // sender already closed; nothing to do
+                wake.into_next()
+            }
+
+            ReceiveTrafficState::Available(data) => {
+                println!(
+                    "received response: {:?}",
+                    String::from_utf8_lossy(data.data)
+                );
+                let (used, next_state) = data.into_next();
+                buffer.discard(used);
+                let used = buffer.into_used();
+                incoming_tls.copy_within(used..incoming_used, 0);
+                incoming_used -= used;
+                next_state
+            }
+
+            ReceiveTrafficState::CloseNotify => {
+                println!("server cleanly closed connection");
+                break;
+            }
+        };
+    }
+
+    Ok(())
+}
+
+async fn recv_tls(
+    sock: &mut TcpStream,
+    incoming_tls: &mut [u8],
+    incoming_used: &mut usize,
+) -> Result<usize, Box<dyn Error>> {
+    let read = sock
+        .read(&mut incoming_tls[*incoming_used..])
+        .await?;
+    println!("received {read}B of data");
+    *incoming_used += read;
+    Ok(read)
+}
+
+const SERVER_NAME: &str = "jbp.io";
+const PORT: u16 = 443;
+
+const INCOMING_TLS_BUFSIZE: usize = 16 * 1024;
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -75,12 +75,6 @@ impl ClientConnection {
         self.core.side.early_data.is_accepted()
     }
 
-    /// Extract secrets, so they can be used when configuring kTLS, for example.
-    /// Should be used with care as it exposes secret key material.
-    pub fn dangerous_extract_secrets(self) -> Result<ExtractedSecrets, Error> {
-        self.core.dangerous_extract_secrets()
-    }
-
     /// Return the connection's Encrypted Client Hello (ECH) status.
     pub fn ech_status(&self) -> EchStatus {
         self.core.side.ech_status
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -11,22 +11,25 @@ use crate::client::EchStatus;
 use crate::common_state::{CommonState, ConnectionOutputs, EarlyDataEvent, Event, Protocol, Side};
 use crate::conn::private::SideOutput;
 use crate::conn::{
-    Connection, ConnectionCommon, ConnectionCore, IoState, KeyingMaterialExporter, Reader,
+    Buffers, Connection, ConnectionCore, IoState, KeyingMaterialExporter, PlaintextSink, Reader,
     SideCommonOutput, Writer,
 };
 #[cfg(doc)]
 use crate::crypto;
+use crate::crypto::cipher::OutboundPlain;
 use crate::enums::ApplicationProtocol;
 use crate::error::Error;
 use crate::log::trace;
-use crate::msgs::ClientExtensionsInput;
+use crate::msgs::{ClientExtensionsInput, Delocator, TlsInputBuffer};
 use crate::quic::QuicOutput;
 use crate::suites::ExtractedSecrets;
 use crate::sync::Arc;
 
 /// This represents a single TLS client connection.
 pub struct ClientConnection {
-    inner: ConnectionCommon<ClientSide>,
+    core: ConnectionCore<ClientSide>,
+    fips: FipsStatus,
+    buffers: Buffers,
 }
 
 impl fmt::Debug for ClientConnection {
@@ -56,13 +59,7 @@ impl ClientConnection {
     /// in this case the data is lost but the connection continues.  You
     /// can tell this happened using `is_early_data_accepted`.
     pub fn early_data(&mut self) -> Option<WriteEarlyData<'_>> {
-        if self
-            .inner
-            .core
-            .side
-            .early_data
-            .is_enabled()
-        {
+        if self.core.side.early_data.is_enabled() {
             Some(WriteEarlyData::new(self))
         } else {
             None
@@ -75,28 +72,28 @@ impl ClientConnection {
     /// handshake then the server will not process the data.  This
     /// is not an error, but you may wish to resend the data.
     pub fn is_early_data_accepted(&self) -> bool {
-        self.inner.core.is_early_data_accepted()
+        self.core.side.early_data.is_accepted()
     }
 
     /// Extract secrets, so they can be used when configuring kTLS, for example.
     /// Should be used with care as it exposes secret key material.
     pub fn dangerous_extract_secrets(self) -> Result<ExtractedSecrets, Error> {
-        self.inner.dangerous_extract_secrets()
+        self.core.dangerous_extract_secrets()
     }
 
     /// Return the connection's Encrypted Client Hello (ECH) status.
     pub fn ech_status(&self) -> EchStatus {
-        self.inner.core.side.ech_status
+        self.core.side.ech_status
     }
 
     fn write_early_data(&mut self, data: &[u8]) -> io::Result<usize> {
-        self.inner
-            .core
+        self.core
             .side
             .early_data
             .check_write(data.len())
             .map(|sz| {
-                self.inner
+                self.core
+                    .common
                     .send
                     .send_early_plaintext(&data[..sz])
             })
@@ -104,74 +101,256 @@ impl ClientConnection {
 
     /// Returns the number of TLS1.3 tickets that have been received.
     pub fn tls13_tickets_received(&self) -> u32 {
-        self.inner
-            .core
+        self.core
             .common
             .recv
             .tls13_tickets_received
+    }
+
+    fn current_io_state(&self) -> IoState {
+        let common_state = &self.core.common;
+        IoState {
+            tls_bytes_to_write: common_state.send.sendable_tls.len(),
+            plaintext_bytes_to_read: self.buffers.received_plaintext.len(),
+            peer_has_closed: common_state
+                .recv
+                .has_received_close_notify,
+        }
     }
 }
 
 impl Connection for ClientConnection {
     fn read_tls(&mut self, rd: &mut dyn io::Read) -> Result<usize, io::Error> {
-        self.inner.read_tls(rd)
+        if self
+            .buffers
+            .received_plaintext
+            .is_full()
+        {
+            return Err(io::Error::other("received plaintext buffer full"));
+        }
+
+        if self
+            .core
+            .common
+            .recv
+            .has_received_close_notify
+        {
+            return Ok(0);
+        }
+
+        let res = self.buffers.deframer_buffer.read(rd);
+        if let Ok(0) = res {
+            self.buffers.has_seen_eof = true;
+        }
+        res
     }
 
     fn write_tls(&mut self, wr: &mut dyn io::Write) -> Result<usize, io::Error> {
-        self.inner.write_tls(wr)
+        self.core
+            .common
+            .send
+            .sendable_tls
+            .write_to(wr)
     }
 
     fn wants_read(&self) -> bool {
-        self.inner.wants_read()
+        // We want to read more data all the time, except when we have unprocessed plaintext.
+        // This provides back-pressure to the TCP buffers. We also don't want to read more after
+        // the peer has sent us a close notification.
+        //
+        // In the handshake case we don't have readable plaintext before the handshake has
+        // completed, but also don't want to read if we still have sendable tls.
+        self.buffers
+            .received_plaintext
+            .is_empty()
+            && !self
+                .core
+                .common
+                .recv
+                .has_received_close_notify
+            && (self
+                .core
+                .common
+                .send
+                .may_send_application_data
+                || self
+                    .core
+                    .common
+                    .send
+                    .sendable_tls
+                    .is_empty())
     }
 
     fn wants_write(&self) -> bool {
-        self.inner.wants_write()
+        !self
+            .core
+            .common
+            .send
+            .sendable_tls
+            .is_empty()
     }
 
     fn reader(&mut self) -> Reader<'_> {
-        self.inner.reader()
+        let has_received_close_notify = self
+            .core
+            .common
+            .recv
+            .has_received_close_notify;
+        Reader {
+            received_plaintext: &mut self.buffers.received_plaintext,
+            // Are we done? i.e., have we processed all received messages, and received a
+            // close_notify to indicate that no new messages will arrive?
+            has_received_close_notify,
+            has_seen_eof: self.buffers.has_seen_eof,
+        }
     }
 
     fn writer(&mut self) -> Writer<'_> {
-        self.inner.writer()
+        Writer::new(self)
     }
 
     fn process_new_packets(&mut self) -> Result<IoState, Error> {
-        self.inner.process_new_packets()
+        loop {
+            let Some(payload) = self
+                .core
+                .process_new_packets(&mut self.buffers.deframer_buffer, None)?
+            else {
+                break;
+            };
+
+            let payload =
+                payload.reborrow(&Delocator::new(self.buffers.deframer_buffer.slice_mut()));
+            self.buffers
+                .received_plaintext
+                .append(payload.into_vec());
+            self.buffers.deframer_buffer.discard(
+                self.core
+                    .common
+                    .recv
+                    .deframer
+                    .take_discard(),
+            );
+        }
+
+        // Release unsent buffered plaintext.
+        if self
+            .core
+            .common
+            .send
+            .may_send_application_data
+            && !self
+                .buffers
+                .sendable_plaintext
+                .is_empty()
+        {
+            self.core
+                .common
+                .send
+                .send_buffered_plaintext(&mut self.buffers.sendable_plaintext);
+        }
+
+        Ok(self.current_io_state())
     }
 
     fn exporter(&mut self) -> Result<KeyingMaterialExporter, Error> {
-        self.inner.exporter()
+        self.core.exporter()
     }
 
     fn dangerous_extract_secrets(self) -> Result<ExtractedSecrets, Error> {
-        self.inner.dangerous_extract_secrets()
+        self.core.dangerous_extract_secrets()
     }
 
     fn set_buffer_limit(&mut self, limit: Option<usize>) {
-        self.inner.set_buffer_limit(limit)
+        self.buffers
+            .sendable_plaintext
+            .set_limit(limit);
+        self.core
+            .common
+            .send
+            .sendable_tls
+            .set_limit(limit);
     }
 
     fn set_plaintext_buffer_limit(&mut self, limit: Option<usize>) {
-        self.inner
-            .set_plaintext_buffer_limit(limit)
+        self.buffers
+            .received_plaintext
+            .set_limit(limit);
     }
 
     fn refresh_traffic_keys(&mut self) -> Result<(), Error> {
-        self.inner.refresh_traffic_keys()
+        self.core
+            .common
+            .send
+            .refresh_traffic_keys()
     }
 
     fn send_close_notify(&mut self) {
-        self.inner.send_close_notify();
+        self.core
+            .common
+            .send
+            .send_close_notify()
     }
 
     fn is_handshaking(&self) -> bool {
-        self.inner.is_handshaking()
+        !(self
+            .core
+            .common
+            .send
+            .may_send_application_data
+            && self
+                .core
+                .common
+                .recv
+                .may_receive_application_data)
     }
 
     fn fips(&self) -> FipsStatus {
-        self.inner.fips
+        self.fips
+    }
+}
+
+impl PlaintextSink for ClientConnection {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let len = self
+            .core
+            .common
+            .send
+            .buffer_plaintext(buf.into(), &mut self.buffers.sendable_plaintext);
+        self.core
+            .common
+            .send
+            .maybe_refresh_traffic_keys();
+        Ok(len)
+    }
+
+    fn write_vectored(&mut self, bufs: &[io::IoSlice<'_>]) -> io::Result<usize> {
+        let payload_owner: Vec<&[u8]>;
+        let payload = match bufs.len() {
+            0 => return Ok(0),
+            1 => OutboundPlain::Single(bufs[0].deref()),
+            _ => {
+                payload_owner = bufs
+                    .iter()
+                    .map(|io_slice| io_slice.deref())
+                    .collect();
+
+                OutboundPlain::new(&payload_owner)
+            }
+        };
+        let len = self
+            .core
+            .common
+            .send
+            .buffer_plaintext(payload, &mut self.buffers.sendable_plaintext);
+        self.core
+            .common
+            .send
+            .maybe_refresh_traffic_keys();
+        Ok(len)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
     }
 }
 
@@ -179,7 +358,7 @@ impl Deref for ClientConnection {
     type Target = ConnectionOutputs;
 
     fn deref(&self) -> &Self::Target {
-        &self.inner
+        &self.core.common.outputs
     }
 }
 
@@ -210,16 +389,15 @@ impl ClientConnectionBuilder {
         let alpn_protocols = alpn_protocols.unwrap_or_else(|| config.alpn_protocols.clone());
         let fips = config.fips();
         Ok(ClientConnection {
-            inner: ConnectionCommon::new(
-                ConnectionCore::for_client(
-                    config,
-                    name,
-                    ClientExtensionsInput::from_alpn(alpn_protocols),
-                    None,
-                    Protocol::Tcp,
-                )?,
-                fips,
-            ),
+            core: ConnectionCore::for_client(
+                config,
+                name,
+                ClientExtensionsInput::from_alpn(alpn_protocols),
+                None,
+                Protocol::Tcp,
+            )?,
+            buffers: Buffers::new(),
+            fips,
         })
     }
 }
@@ -242,7 +420,6 @@ impl<'a> WriteEarlyData<'a> {
     /// once this reaches zero.
     pub fn bytes_left(&self) -> usize {
         self.sess
-            .inner
             .core
             .side
             .early_data
@@ -269,7 +446,7 @@ impl<'a> WriteEarlyData<'a> {
     /// [RFC8446 appendix E.5.1]: https://datatracker.ietf.org/doc/html/rfc8446#appendix-E.5.1
     /// [`Connection::exporter()`]: crate::conn::Connection::exporter()
     pub fn exporter(&mut self) -> Result<KeyingMaterialExporter, Error> {
-        self.sess.inner.core.early_exporter()
+        self.sess.core.early_exporter()
     }
 }
 
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -394,6 +394,16 @@ impl ClientConnectionBuilder {
             fips,
         })
     }
+
+    /// Finalize the builder and create a [`ClientHandshake`].
+    pub fn start_handshake(self) -> Result<ClientHandshake, Error> {
+        let Self {
+            config,
+            name,
+            alpn_protocols,
+        } = self;
+        ClientHandshake::new(config, name, alpn_protocols)
+    }
 }
 
 /// Allows writing of early data in resumed TLS 1.3 connections.
@@ -505,7 +515,7 @@ impl EarlyData {
         )
     }
 
-    fn is_accepted(&self) -> bool {
+    pub(super) fn is_accepted(&self) -> bool {
         matches!(
             self.state,
             EarlyDataState::Accepted | EarlyDataState::AcceptedFinished
@@ -542,7 +552,7 @@ impl EarlyData {
         }
     }
 
-    fn check_write(&mut self, sz: usize) -> io::Result<usize> {
+    pub(super) fn check_write(&mut self, sz: usize) -> io::Result<usize> {
         self.check_write_opt(sz)
             .ok_or_else(|| io::Error::from(io::ErrorKind::InvalidInput))
     }
@@ -564,7 +574,7 @@ impl EarlyData {
         }
     }
 
-    fn bytes_left(&self) -> usize {
+    pub(super) fn bytes_left(&self) -> usize {
         self.left
     }
 }
@@ -580,7 +590,7 @@ enum EarlyDataState {
 }
 
 pub(crate) struct ClientConnectionData {
-    early_data: EarlyData,
+    pub(super) early_data: EarlyData,
     ech_status: EchStatus,
 }
 
@@ -590,6 +600,14 @@ impl ClientConnectionData {
             early_data: EarlyData::new(),
             ech_status: EchStatus::default(),
         }
+    }
+
+    pub(crate) fn early_data_is_enabled(&self) -> bool {
+        self.early_data.is_enabled()
+    }
+
+    pub(crate) fn ech_status(&self) -> EchStatus {
+        self.ech_status
     }
 }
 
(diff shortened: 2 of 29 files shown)


ctz:
@djc this is ready for a first look now.

Things I don't particularly like:

- the underlying code for buffered is more duplicated than i would like, this is because there is no abstraction over the two state-based APIs.
- the "state-based API" naming is not very descriptive; suggestions welcome!
- objects with move-based APIs are kept in Options/Results, which is generally painful

Things specifically to come later:

1. add support for the new API to rustls-bench
2. on the client, expose the certificate verification process in the state-based API, allowing async verifiers (the buffered API would continue to call the trait)
3. similarly, on the server, expose the certificate verification process in the state-based API
4. potentially rework how the send path encryption works; this would not work with the current aws-lc-rs/ring API but would give a route to a zero-alloc, zero-copy, zero-bzero application data send path

djc:
Suggest to start by extracting all the boring stuff into a separate PR to get it merged:

- Error documentation tweaks
- Extraction of `ConnectionBuffers`
- Record the ability for half-RTT data
- Renaming `DeframerVecBuffer`
- Insert trait for `ConnectionOutputs`
- Rework `SendPath::write_plaintext()`
- Box `ChooseConfig` state

And also:

- Introducing `PreviousConnectionError`

Suggest renaming the "Transfer buffered API" commits to have "client:" and "server:" first in the commit message to make the difference stand out more.

rustls/src/conn/mod.rs
```diff
@@ -851,6 +856,25 @@ impl<Side: SideData> ConnectionCore<Side> {
         }
     }
 
+    pub(crate) fn process_new_packets(
```
ctz: I reintroduced this helper, as it ends up being called from several places.

djc: Suggest also extracting this into the prefix PR. See also https://github.com/rustls/rustls/pull/2970/changes/b9edf7830e971ddd100bbadedff63b8c4629d408 (from #2970).

rustls/src/conn/receive.rs
```diff
@@ -424,6 +426,35 @@ impl ReceivePath {
...
+
+impl FinishCondition for FinishHandshake {
+    type State = bool;
+    fn start(recv: &ReceivePath) -> Self::State {
+        recv.may_receive_application_data
+    }
+    fn is_done(state: &Self::State, recv: &ReceivePath) -> bool {
+        *state != recv.may_receive_application_data
+    }
+}
```
djc: Suggest deferring the introduction of this into a commit that actually uses it.

rustls/src/client/state.rs
```diff
@@ -0,0 +1,393 @@
...
+use crate::msgs::{ClientExtensionsInput, TlsInputBuffer};
+use crate::state::{ReceiveTraffic, SendTraffic};
+use crate::sync::Arc;
+use crate::{
+    ClientConfig, CommonState, ConnectionOutputs, Error, ExtractedSecrets, KeyingMaterialExporter,
+};
+
+/// State-based Client TLS API.
+#[non_exhaustive]
+pub enum ClientState {
```
djc: So this still adds substantial parallel API. In my mind, I was hoping there was a way to maybe introspect/transition from a normal `ClientConnection` to the contained state. Do you think that's feasible? If not, why not?

rustls/src/client/state.rs
```diff
@@ -0,0 +1,393 @@
...
+            Self::Traffic(_) => f
+                .debug_tuple("Traffic")
+                .finish_non_exhaustive(),
+        }
+    }
+}
+
+/// A handshake state where we are required to send data to the peer.
+pub struct SendClientFlight {
+    inner: Box<ConnectionCore<ClientSide>>,
```
djc, line 110: Seems to me that this is an additional `Box` wrapper that the "normal" API doesn't have, and as we're moving between client and server flights this would cause a bunch of additional allocations and deallocations? Feels to me like a lower-level API like this should do fewer of those things on behalf of the caller rather than more.

rustls/src/client/state.rs
```diff
@@ -0,0 +1,393 @@
...
+
+    /// The receive side of the connection.
+    ///
+    /// This starts out as `Some()`, and callers may `take()` or insert `None`
+    /// to denote closure of the receive side (whether cleanly or otherwise).
+    pub receive: Option<ReceiveTraffic<ClientSide>>,
+
+    /// Facts about the connection established during the handshake.
+    pub outputs: ClientOutputs,
+}
```
djc, line 271: In my mind `ClientOutputs` should not be part of `ClientTraffic`, because it is "ejected" to the caller as part of the transition to `ClientTraffic`.

rustls/src/client/state.rs
```diff
@@ -0,0 +1,393 @@
...
+    ///
+    /// This starts out as `Some()`, and callers may `take()` or insert `None`
+    /// to denote closure of the send side.
+    pub send: Option<SendTraffic>,
+
+    /// The receive side of the connection.
+    ///
+    /// This starts out as `Some()`, and callers may `take()` or insert `None`
+    /// to denote closure of the receive side (whether cleanly or otherwise).
+    pub receive: Option<ReceiveTraffic<ClientSide>>,
```
djc: Hmm, I don't love this `Option` setup. IMO we should just allow the caller to `split()`, which yields both parts separately? (Not sure if the converse `merge()` operation would be useful as well.)

rustls/src/client/state.rs
```diff
@@ -0,0 +1,393 @@
...
+
+impl ClientOutputs {
+    /// Return the connection's Encrypted Client Hello (ECH) status.
+    pub fn ech_status(&self) -> EchStatus {
+        self.ech_status
+    }
+
+    pub(crate) fn is_early_data_accepted(&self) -> bool {
+        self.early_data_was_accepted
+    }
```
djc: Adding trivial getters for these doesn't seem all that useful?

rustls/src/conn/mod.rs
```diff
@@ -792,6 +791,30 @@ impl<Side: SideData> ConnectionCore<Side> {
             )
     }
 
+    #[inline]
```
djc, line 555: This seems to appear as new here (in "Introduce new state-based API"), did it get moved? Could that be split into a separate commit? Or does this still need to rebase on `main`?

rustls/src/conn/receive.rs
```diff
@@ -530,11 +552,70 @@ impl<'m> Output<'m> for CaptureAppData<'_, '_, 'm> {
     }
 }
 
-pub(super) struct JoinOutput<'a> {
-    pub(super) outputs: &'a mut ConnectionOutputs,
-    pub(super) quic: Option<&'a mut dyn QuicOutput>,
-    pub(super) send: &'a mut dyn SendOutput,
-    pub(super) side: &'a mut dyn SideOutput,
+pub(crate) struct DropAppData<'a, 'j> {
```
djc: This seems to mirror/be ignorant of changes I've made elsewhere. Does it still make sense?

rustls/src/msgs/deframer/buffers.rs
```diff
@@ -125,15 +124,13 @@ impl TlsInputBuffer for VecInput {
 
 /// A borrowed version of [`VecInput`] that tracks discard operations
 #[derive(Debug)]
-#[expect(unreachable_pub)]
```
djc, line 127: Seems to me that the input method (`SliceInput` vs `VecInput`) should be fully independent from whether we use the buffered or the state-based API.

rustls/src/state.rs
```diff
@@ -0,0 +1,320 @@
...
+use crate::msgs::Delocator;
+pub use crate::msgs::{SliceInput, TlsInputBuffer, VecInput};
+use crate::sync::Arc;
+use crate::tls13::key_schedule::KeyScheduleTrafficSend;
+use crate::{Error, SideData};
+
+/// The send-side of a connection.
+///
+/// You can use this object to send data to the peer.
+pub struct SendTraffic(pub(crate) Arc<Mutex<SendPath>>);
```
djc, line 20: Similarly, seems to me that the splitting up ought to be fully independent of both "owned vs borrowed buffers" and "state-based API". (I can see how it might be useful to combine them to prove out that you're on the right path, but as a way of getting them merged I think we'll have an easier time when focusing on one thing at a time.)

djc (commented):
Here's a more in-depth first round of review of "Introduce new state-based API". Since I've given some potentially disruptive feedback, going to stop here to see how you want to continue from here.

rustls/src/client/state.rs
ctz: The purpose of this is so that a transition from (eg) `SendClientFlight` to `AwaitServerFlight` is fast. Losing the box means the entirety of `ConnectionCore` is moved from place to place, which is slow (it's a memmove of like 500 bytes). That is especially important, because even considering a single newly received byte involves these operations for the caller:

- prior to calling `AwaitServerFlight::input_data`, prepare a place for the return value (stack allocation of hundreds of bytes)
- call `AwaitServerFlight::input_data`, which quickly finds there is nothing to do
- (most likely `next_state` is inlined) move `self.inner` into a new `AwaitServerFlight`
- return from the function
- caller now moves the new state to where-ever they wanted it

Because the value is moved, there is no allocation/deallocation except at the start and end of the handshake.

rustls/src/client/state.rs
ctz: I think that would mean most callers end up writing their own copy of this structure. Maybe that is OK though.

rustls/src/msgs/deframer/buffers.rs
ctz: The buffering in the buffered API is much more than whether the TLS-side input is buffered.

rustls/src/client/state.rs
djc: Why's that?

ctz: So say we get rid of the `Option`s in this structure. The API becomes kinda unusable, because the first call to `ReceiveTraffic::read` moves the `receive` field, and it's an absolute requirement that it is replaced with something valid if the caller wants to continue with that structure (there is no sentinel value for a `ReceiveTraffic`, and the caller can't create one).

The other option is to require callers to destructure this to continue (or, equivalently, call some `split()` method as you mentioned). I think that is OK. But the caller would need to then put the `receive` and `send` fields somewhere, and they would need to be in `Option`s for the same reason.

rustls/src/conn/receive.rs
ctz: Will look again at this.

rustls/src/client/state.rs
djc: Yes, I think callers will likely want to destructure this -- for example, in an async world you might want to run send and receive on different tasks like Tokio already allows for TCP connections. So I think either they're kept together (no `Option`) or they're split and it's up to the caller.

rustls/src/msgs/deframer/buffers.rs
djc: That's fair, but turning it around, doing all the other things unbuffered/with externalized state/whatever doesn't necessitate doing non-I/O TLS input in the same PR.

rustls/src/client/state.rs
ctz: Just addressed this one.

djc:
Reflecting a bit more on this at a high-level:

- IMO we should get rid of the `ConnectionBuffers` entirely even in the high-level API. For users that aren't using rustls via a higher-level wrapper like tokio-rustls, I think it will still be ergonomic enough, and avoids the "hidden" costs of buffering internally.

- The work on externalizing the `VecInput` buffer in #2940 suggested to me that the current trait interface might not be complete (in particular, for expressing EOF/`UnexpectedEof` handling). So adding the `SliceInput` alternative, which AIUI currently has no affordances for EOF handling, doesn't seem like a great idea at this stage. I also think buffering inside a `Vec` is not a great limitation for now, as we evolve the trait to make that pluggable.

- That leaves the current traits (resolver and verifier etc), and if we want to abstract over them for a higher-level API I think there might be more ergonomic/lower boilerplate ways to do that.

rustls/src/client/state.rs
ctz: I think this addressed now by `ClientTraffic` being a destructure-only type.

rustls/src/state.rs
ctz: I am keen to avoid the situation with the unbuffered API, where the new API was actually unusable in practice (initially, because it didn't do closure properly). Using the new API to supplant the internals of the existing API is a good way of demonstrating that (because, I think the existing API has been shown to be complete and usable in most real cases.)

rustls/src/client/state.rs
ctz: These are now pub fields.

rustls/src/conn/mod.rs
ctz: Hoisted out to a separate commit.

ctz:
Notwithstanding the other comments here, and other ideas for future changes to the high-level API, I think this is ready for more review.

While that is happening I will look into the performance findings, which seems to be rooted in more allocator usage.

djc:
I think there are two questions here:

1. What is the high-level end state of the 0.24 API?
2. What is the best route to get there from where we are today?

This PR ends up at this state:

```rust
#[non_exhaustive]
pub enum ClientState {
    /// Handshake data should be transmitted to the peer.
    SendClientFlight(SendClientFlight),

    /// We are awaiting handshake data from the peer.
    AwaitServerFlight(AwaitServerFlight),

    /// The handshake has completed and application data may now flow.
    Traffic(Box<ClientTraffic>),
}
```

Coming from the monolithic `ClientConnection` we had in 0.23, in my mind the ideal low-level API for 0.24 looks like more or less directly exposing the existing `hs::ClientState` with assorted public methods on the state types which do all the things you need to do, and some extra states exposed for the local transitions currrently driven via trait objects. The `ClientConnection` would still exist, as a wrapper of `ClientState` which additionally carries a collection of trait objects (or maybe even a monolithic trait object) which takes care of verification and resolution (and cipher suite selection, I suppose).

However, the state of this PR doesn't feel like it's on a path between 0.23 and the API I've had in mind for 0.24. That might be because my perception of the 0.24 API is different from your perception, or because this seemed like the easiest path to get there (or a bit of both).

Here's roughly the path that I would take:

- Unconditionally externalize the input buffer: this is nice because IMO the high-level API can bear this being done without loss of ergonomics but while losing a bunch of complexity from encompassing I/O (thus #2940).
- Unconditionally externalize `received_plaintext` and `sendable_plaintext`: I think we can come up with reasonably ergonomic ways of doing this that are ergonomic enough for the high-level API, still without splitting into high-level and low-level API.
- Split existing states to accomodate locally-driven state transitions for verification and resolution, hoisting out the use of the respective trait objects.
- Move these trait objects out of `ClientConfig`, into the `ClientConnection` or `ClientConnectionData` types.
- Make the existing `ClientState` public with along side the existing `ClientConnection` type, with the latter driving the former.

Without wanting detract from your work on this PR, it's not obvious to me how it helps on this path in my mind.

ctz:
> 1. What is the high-level end state of the 0.24 API?

1. a high-level, functionally-equivalent-to-0.23 (ie, pipe-like, with std::io abstractions) connection type. It's OK to make breaking changes, but not shovel lots of complexity onto every caller, or move to a different abstraction (specifically, the abstraction of a std::io pipe, which is the only standard interface that is common in rust for this.)
2. a lower-level, more flexible, more async-amenable, no-read-buffered, no_std and high-performance interface, with a minimum of exposed API surface. it must be possible to evolve details under this, so it shouldn't be so low level that adding or varying protocol features (for example, in 0.23, certificate compression) involve a breaking API or functional change. I think I've expressed that before, but I don't want to expose the low-level handshake types for this reason.

I think if we start down the road away from something fundamentally the shape of (1), it should start with a PR that replaces [the front page of the documentation](https://rustls.dev/docs/#design-overview).

To be clear, this PR is not moving anywhere near (1), it is introducing (2) and showing that #1723 is possible using it, and addressing a bunch of other shortcomings with the public API.

> Split existing states to accomodate locally-driven state transitions for verification and resolution

This is [mentioned above](https://github.com/rustls/rustls/pull/2953#issuecomment-4031998474) so I think this is a place where we agreeing.

> Move these trait objects out of ClientConfig, into the ClientConnection or ClientConnectionData types.

Right, I would like the traits only to be known at the `ClientConnection` level, with the lower level users being required to feed in dependencies directly (where they could simply call the traits themselves). That's in a similar way to how the new `ServerState` API funnels every caller through the `Acceptor`-like flow, except currently that doesn't go far enough as the `Accepter` still uses the traits for choosing a credential.

djc:
> 1. a high-level, functionally-equivalent-to-0.23 (ie, pipe-like, with std::io abstractions) connection type. It's OK to make breaking changes, but not shovel lots of complexity onto every caller, or move to a different abstraction (specifically, the abstraction of a std::io pipe, which is the only standard interface that is common in rust for this.)

..

> I think if we start down the road away from something fundamentally the shape of (1), it should start with a PR that replaces [the front page of the documentation](https://rustls.dev/docs/#design-overview).

I thought we were more aligned on this.

Yes, I think we should move away from some aspects of (1). First, the diagram on the front page is simplified -- there's actually a `process_new_packets()` step in the middle that is left out. I also think its simplicity is misleading -- acting like it is a pipe would require something like `complete_io()`, 100 lines of fairly dense code which has had its fair share of bugs in the past and which has recently been moved out of the core crate.

What I'm proposing is some modifications that I think provide a minor hit to the ergonomics/developer experience while reducing the complexity for many/most users (and for the maintainers), in a way that reminds me of Gabriel's Worse is Better essay.

- `handle_tls()` takes a `&mut dyn TlsInputBuffer` and yields an `impl Iterator<Item = &[u8]>` with chunks of received plaintext. It also takes a `&mut Vec<u8>` (or another trait object) for buffering TLS output. We could provide a `TlsInputBuffer` implementation that has a `read()` method for reading from a `&mut dyn io::Read`.
- `write_plaintext()` takes a `&[u8]` with plaintext data input and a `&mut Vec<u8>` or trait object similar to `read_tls()`.

Why it's better:

- Avoid data copies for this high-level API (as well as any lower-level API underpinning it).
- Externalize/clarify the requirement that input must be in a contiguous buffer.
- No more large `std` feature guards on the core API.
- Low-level and high-level API are more similar, since they both work with external buffers.

ctz:
OK, I think we are actually mostly aligned on these things, but are discussing the order. Maybe a way forward is to drop the final three commits here and defer #1723 until later, while keeping an eye that it is and remains possible in other changes to the high-level API?

djc:
> OK, I think we are actually mostly aligned on these things, but are discussing the order. Maybe a way forward is to drop the final three commits here and defer #1723 until later, while keeping an eye that it is and remains possible in other changes to the high-level API?

My problem so far is mostly with the 3rd commit, which adds a lot of new parallel public API surface. Do you think most of that will survive if we implement the other things discussed here?

ctz:
Yes, I think so. I don't see a route where `fn handle_tls(&mut self, input: &mut dyn TlsInputBuffer, output: &mut Vec<u8>) -> Result<impl Iterator<Item = &[u8]>, Error>` can address things like #2761 and #850?

rustls/src/client/state.rs
```diff
@@ -0,0 +1,383 @@
...
+/// and then progress them separately.
+///
+/// For example:
+///
+/// - if you do not require any information from `outputs`, you can drop it
+///   now to save memory.
+/// - if you are only receiving data, you can drop `send` now.  That is best done
+///   by calling [`SendTraffic::close()`].
+///
+/// Note that for good behaviour, it is usually necessary to service `receive`.  This
```
djc: Nit: "behaviour" -> "behavior".

rustls/src/client/state.rs
```diff
@@ -0,0 +1,383 @@
...
+
+        Self {
+            outputs: inner.common.outputs,
+            early_data_was_accepted,
+            ech_status,
+        }
+    }
+}
+
+impl fmt::Debug for ClientOutputs {
```
djc: Feels like this should be derived, and there should be a `Debug` impl for `ConnectionOutputs`.

rustls/src/client/state.rs
```diff
@@ -0,0 +1,383 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.outputs
+    }
+}
+
+impl DerefMut for ClientOutputs {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.outputs
+    }
+}
```
djc, line 355: Why is this useful? Seems like by the time we build a `ClientOutputs`, it is immutable from the perspective of rustls?

rustls/src/conn/receive.rs
```diff
@@ -510,11 +527,21 @@ impl<'m> Output<'m> for CaptureAppData<'_, '_, 'm> {
...
-    pub(super) send: &'a mut dyn SendOutput,
-    pub(super) side: &'a mut dyn SideOutput,
+pub(crate) struct JoinOutput<'a> {
+    pub(crate) outputs: &'a mut dyn ConnectionOutput,
+    pub(crate) quic: Option<&'a mut dyn QuicOutput>,
+    pub(crate) send: &'a mut dyn SendOutput,
+    pub(crate) side: &'a mut dyn SideOutput,
+}
+
+struct Discard;
```
djc, line 531: When could/should this be used without dropping stuff? Should we instead have a different output trait for traffic states?

rustls/src/client/state.rs
```diff
@@ -0,0 +1,383 @@
...
+        let inner = Box::new(ConnectionCore::for_client(
+            config,
+            name,
+            input,
+            None,
+            Protocol::Tcp,
+        )?);
+
+        Ok(Self::SendClientFlight(SendClientFlight { inner }))
+    }
```
djc: Should this instead be a method on `ClientConnectionBuilder`?

rustls/src/server/state.rs
```diff
@@ -0,0 +1,612 @@
...
+        }
+    }
+}
+
+/// We have received a `ClientHello` and it is time to choose a `ServerConfig` for this connection.
+pub struct ChooseConfig {
+    // invariant: `inner.state` is `Err(_)` with a meaningless error,
+    // it should be reinserted before further use.
+    inner: Box<ConnectionCore<ServerSide>>,
+    choose_config: Box<hs::ChooseConfig>,
```
djc, line 222: Nit: just import `ChooseConfig`?

rustls/src/server/state.rs
```diff
@@ -0,0 +1,612 @@
...
+                .as_deref(),
+            client_cert_types: client_hello
+                .client_certificate_types
+                .as_deref(),
+            cipher_suites: &client_hello.cipher_suites,
+            certificate_authorities: client_hello
+                .certificate_authority_names
+                .as_deref(),
+            named_groups: client_hello.named_groups.as_deref(),
+        };
```
djc: This should probably be abstracted out as a constructor on `ClientHello` or something?

rustls/src/server/state.rs
```diff
@@ -0,0 +1,612 @@
...
+
+    /// Expose the raw received `ClientHello` bytes.
+    ///
+    /// This is the reassembled and complete `ClientHello` handshake message as specified
+    /// in (eg) RFC8446.  In other words, the bytes start with `legacy_version`.  There is
+    /// no header or length prefix.
+    ///
+    /// Rustls has parsed and checked the structure of this encoding, but full processing
+    /// of the message only completes later so callers of this function should be defensive.
+    pub fn client_hello_bytes(&self) -> &[u8] {
```
djc: Who/what is this for? I don't think we expose this today?

rustls/src/server/state.rs
```diff
@@ -0,0 +1,612 @@
...
+
+        Self {
+            outputs: inner.common.outputs,
+            server_name,
+            received_resumption_data,
+        }
+    }
+}
+
+impl fmt::Debug for ServerOutputs {
```
djc: As for client.

rustls/src/server/state.rs
```diff
@@ -0,0 +1,612 @@
...
+
+impl Deref for ServerOutputs {
+    type Target = ConnectionOutputs;
+
+    fn deref(&self) -> &Self::Target {
+        &self.outputs
+    }
+}
+
+impl DerefMut for ServerOutputs {
```
djc, line 524: As for client.

rustls/src/server/state.rs
```diff
@@ -0,0 +1,612 @@
...
+            };
+            ServerState::ChooseConfig(ChooseConfig {
+                inner,
+                choose_config,
+            })
+        }
+
+        Ok(_) => ServerState::AwaitClientFlight(AwaitClientFlight { inner }),
+
+        Err(_) => panic!("TODO: withdraw error fusing in core.state"),
```
djc: What is this waiting for?

examples/src/bin/async-client.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    let config = Arc::new(config);
+
+    let mut incoming_tls = vec![0; INCOMING_TLS_BUFSIZE];
+    converse(&config, &mut incoming_tls).await?;
+
+    Ok(())
+}
+
+async fn converse(
```
djc: This function name seems a little idiosyncratic, and different from what we use elsewhere?

rustls/src/client/connection.rs
```diff
@@ -56,16 +93,14 @@ impl ClientConnection {
...
-            .inner
-            .core
-            .side
-            .early_data
-            .is_enabled()
-        {
-            Some(WriteEarlyData::new(self))
-        } else {
-            None
+        match &mut self.state {
```
djc: Maybe something like `let Ok(InnerState::Handshake(ClientState::AwaitServerFlight(st))) = &mut self.state else { return None; }` to reduce rightward drift?

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
     /// is not an error, but you may wish to resend the data.
     pub fn is_early_data_accepted(&self) -> bool {
-        self.inner.core.is_early_data_accepted()
-    }
-
-    /// Extract secrets, so they can be used when configuring kTLS, for example.
-    /// Should be used with care as it exposes secret key material.
-    pub fn dangerous_extract_secrets(self) -> Result<ExtractedSecrets, Error> {
-        self.inner.dangerous_extract_secrets()
+        match &self.state {
```
djc, line 112: Same here.

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
-                    .send_early_plaintext(&data[..sz])
-            })
+        match &self.state {
+            Ok(InnerState::Handshake(st)) => st.ech_status(),
+            Ok(InnerState::Traffic { outputs, .. }) => outputs.ech_status,
+            Err(_) => self
+                .err_outputs
+                .as_ref()
+                .map(|o| o.ech_status)
+                .unwrap_or(EchStatus::Rejected),
```
djc: Nit: would replace `map().unwrap_or()` with a `match` expression.

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
+            peer_has_closed: self.has_received_close_notify(),
+        }
+    }
+
+    fn has_received_close_notify(&self) -> bool {
+        matches!(&self.state, Ok(InnerState::Traffic { receive: None, .. }))
+    }
+
+    fn write_or_buffer_appdata(&mut self, data: OutboundPlain<'_>) -> io::Result<usize> {
+        Ok(match &mut self.state {
```
djc: let-else?

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
+            _ => self
+                .buffers
+                .sendable_plaintext
+                .append_limited_copy(data),
+        })
+    }
+
+    /// Act on a potential state transition from a handshake-related state to `new`.
+    fn post_handshake_state(&mut self, new: ClientState) -> InnerState {
+        if let ClientState::Traffic(traffic) = new {
```
djc: let-else?

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
     }
 
     fn write_tls(&mut self, wr: &mut dyn io::Write) -> Result<usize, io::Error> {
-        self.inner.write_tls(wr)
+        self.buffers.sendable_tls.write_to(wr)
     }
 
     fn wants_read(&self) -> bool {
-        self.inner.wants_read()
+        // We want to read more data all the time, except when we have unprocessed plaintext.
```
djc, line 243: Should there be a precursor commit to inline this logic from wherever it currently lives?

(This is where it feels like the "Transfer" commits should really be squashed into the new state commit...)

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
     }
 
     fn wants_write(&self) -> bool {
-        self.inner.wants_write()
+        !self.buffers.sendable_tls.is_empty()
     }
 
     fn reader(&mut self) -> Reader<'_> {
-        self.inner.reader()
+        let has_received_close_notify = self.has_received_close_notify();
```
djc, line 263: As above.

rustls/src/client/connection.rs
```diff
@@ -75,117 +110,401 @@ impl ClientConnection {
...
     }
 
     fn writer(&mut self) -> Writer<'_> {
-        self.inner.writer()
+        Writer::new(self)
     }
 
     fn process_new_packets(&mut self) -> Result<IoState, Error> {
-        self.inner.process_new_packets()
+        loop {
```
djc, line 278: Here too, the diff misses key steps.

rustls/src/server/connection.rs
```diff
@@ -651,7 +964,7 @@ impl crate::conn::SideData for ServerSide {}
 
 impl crate::conn::private::Side for ServerSide {
     type Data = ServerConnectionData;
-    type State = ServerState;
+    type State = super::hs::ServerState;
```
djc: Can we just import this.

rustls/src/server/connection.rs
```diff
@@ -40,14 +48,14 @@ impl ServerConnection {
...
-                fips,
-            ),
+            state: Ok(InnerState::Handshake(ServerState::new())),
+            buffers: Buffers::new(),
+            config: Some(config),
+            err_outputs: None,
+            early_data_received: false,
+            received_early_data: ChunkVecBuffer::new(None),
+            early_exporter: None,
+            fips,
```
djc: This change should not be in this commit.

Does it mean `ConnectionCommon` goes unused after this?

rustls/src/conn/mod.rs
```diff
@@ -517,186 +473,7 @@ impl ConnectionRandoms {
...
-///
-/// This is one of the core abstractions of the rustls API. It represents a single connection
-/// to a peer, and holds all the state associated with that connection. Note that it does
-/// not hold any IO objects: the application is responsible for reading and writing TLS records.
-/// If you want an object that does hold IO objects, see `rustls_util::Stream` and
-/// `rustls_util::StreamOwned`.
-///
-/// This object is generic over the `Side` type parameter, which must implement the marker trait
-/// [`SideData`]. This is used to store side-specific data.
-pub(crate) struct ConnectionCommon<Side: SideData> {
```
djc, line 514: Code deletions should be squashed into the commits that make them unnecessary.

djc:
> Yes, I think so. I don't see a route where `fn handle_tls(&mut self, input: &mut dyn TlsInputBuffer, output: &mut Vec<u8>) -> Result<impl Iterator<Item = &[u8]>, Error>` can address things like #2761 and #850?

They don't directly. But I think we could get there by making smaller steps on the road to modularity, which IMO would be easier to review for correctness and completeness. Whereas it feels like the success of this PR depends on your keeping track of loose ends that need to be cleaned up/addressed.

rustls/src/server/state.rs
ctz: Removed, was a bit of a thought that it could be useful for people wanting to calculate fingerprints; naturally the clienthello is available from the input buffer anyway as it is now external.

rustls/src/client/state.rs
ctz: Resolved the reason for this in #3002

rustls/src/server/state.rs
djc: Sure -- but let's not add it in this commit?

examples/src/bin/async-client.rs
ctz: inlined (was a hang-over from the reintroduced example)

rustls/src/server/state.rs
ctz: This is in a type called `ChooseConfig`

rustls/src/server/connection.rs
ctz: Have done this by renaming the public type away from `ServerState`

rustls/src/client/state.rs
ctz: done, and this is no longer `pub`

rustls/src/client/connection.rs
ctz: Added this, which inlines and removes `ConnectionCommon`. Then added a couple of commits which address coverage gaps caused.

rustls/src/client/connection.rs
ctz: I think this is addressed by the change above, though it'd be fair to characterise this hunk as "take away a bunch of dense code and replace it with some different dense code"

rustls/src/server/connection.rs
ctz: Resolved as above by the separate commit eliminating `ConnectionCommon`

rustls/src/server/state.rs
ctz: Added a commit which does this. It's a bit painful and not exactly where I'd like it to be (which is -- every function that deals with a `Side::State` accepts it by value and returns the new value on success. However, that means every such function has to consume whatever contains the `Side::State`, which might be annoying/impossible.)

rustls/src/common_state.rs
```diff
@@ -197,6 +197,19 @@ impl ConnectionOutput for ConnectionOutputs {
     }
 }
 
+impl fmt::Debug for ConnectionOutputs {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ConnectionOutputs")
```
djc, line 220: Should this destructure the type to make sure we don't miss anything in the future?

rustls/src/conn/mod.rs
```diff
@@ -501,195 +459,16 @@ impl ConnectionRandoms {
     }
 }
 
-/// TLS connection state with side-specific data (`Side`).
```
djc, line 504: Why is it now impossible to share all this logic between client and server?

djc:
Even apart from the performance situation, I'm still pretty uncomfortable with the state of this PR. The commits are large and not that clean, with quite a lot of code added in the core of the library. Still suspect this is not the right way to get to where we want to go.

rustls/src/conn/mod.rs
ctz: As mentioned above:

> this is because there is no abstraction over the two state-based APIs.

We could introduce one? I think it would be quite large and not dyn compatible, but should be doable?

djc: Large in what way? Could it be not an owned abstraction, but a borrowing (whether `&mut` or `&`) that is only used to deduplicate behavior but not state at rest?

djc:
Because you pushed this 3 days ago, I was assuming this already represented the approach we discussed over dinner last week of only splitting, but I guess that's not the case?

ctz:
That was a 100% rebase; I am working on a separate PR currently which aims for a "vertical slice" of this one.

ctz:
Will be bringing back an alternative formulation of this idea.

PR #2954, opened 2026-03-03
Provider feature and doctests updates

ctz:


rustls/src/lib.rs
```diff
@@ -289,24 +289,15 @@
 //! - `std` (enabled by default): enable the high-level (buffered) Connection API and other functionality
```
ctz, line 283: I left the mentions of `std` deliberately.

PR #2957, opened 2026-03-04
Move references to ECH RFC9849

ctz:


djc (approved):
I'd probably put a space between "RFC" and the number.

ctz:
> I'd probably put a space between "RFC" and the number.

I think the prevailing style is no space:

```
$ grep -rE '//.+RFC[0-9]{4}' * | wc -l
145

$ grep -rE '//.+RFC [0-9]{4}' * | wc -l
73
```

Though IETF themselves use a space uniformly, so will follow that.

djc:
Would be nice to address all the other prevailing uses too, IMO.

PR #2962, opened 2026-03-07
Add updated benchmarking report

ctz:


djc:
I thought @bdaehlie was asking for `main` rather than latest release?

bdaehlie:
Ideally we would get numbers for stable release and main, since main is so far ahead of stable right now. If there is a recent alpha or whatever cut off main that would be fine too.

ctz:
I have added SVGs (names ending `*-next.svg`) which allow comparison with rustls current main.

Haven't referenced them from the report, or included the numbers anywhere. I think that is best done when we have made a numbered release.

PR #2964, opened 2026-03-07
`#[inline]` functions called by provider ciphers

ctz:


ctz:
(couldn't see any benefit from this comparing `rustls-bench bulk TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384`)

PR #2965, opened 2026-03-08
Alter protocol enum representation

### LINKED ISSUE #2829: Enums with an Unknown catch-all variant can potentially break at runtime
I noticed that most public enums include an `Unknown(someInt)` variant, to allow people to handle variants not explicitly defined. It's a good goal, but I wanted to point out a potential compatibility hazard with the pattern. (I ran into this a while ago with hyper, wrote about it [here](https://seanmonstar.com/blog/pattern-matching-and-backwards-compatibility/#open-ended-variants), but inlined and adapted to rustls for this issue.)

If a user was matching on an a new unknown variant, and rustls decided to turn that into an known variant, even though the enum is marked `non_exhaustive`, the user would see breakage at runtime.

```rust
match val {
    // .. other variants
    TlsVersion::Unknown(0x0305) => println!("some future 1.4"),
    _ => ()
}
```

What I've found in cases like these where the underlying bytes are needed by users it to put the enum inside a struct, expose constants, and a method to get the underlying byte. This way, the user could write this and it will still work in the future:

```rust
match val {
    // .. other variants
    unknown if unknown.code() == 0x0305 => println!("some future 1.4"),
    _ => ()
}
```

ctz:
This should show some modest CPU improvement, and a bit of heap usage reduction too.

fixes #2829

### DIFF
--- a/rustls/src/enums.rs
+++ b/rustls/src/enums.rs
@@ -132,9 +132,9 @@ impl AsRef<[u8]> for ApplicationProtocol<'_> {
 enum_builder! {
     /// The `HandshakeType` TLS protocol enum.  Values in this enum are taken
     /// from the various RFCs covering TLS, and are listed by IANA.
-    /// The `Unknown` item is used when processing unrecognized ordinals.
-    #[repr(u8)]
-    pub enum HandshakeType {
+    pub struct HandshakeType(pub u8);
+
+    enum HandshakeTypeName {
         HelloRequest => 0x00,
         ClientHello => 0x01,
         ServerHello => 0x02,
@@ -161,9 +161,9 @@ enum_builder! {
 enum_builder! {
     /// The `ContentType` TLS protocol enum.  Values in this enum are taken
     /// from the various RFCs covering TLS, and are listed by IANA.
-    /// The `Unknown` item is used when processing unrecognized ordinals.
-    #[repr(u8)]
-    pub enum ContentType {
+    pub struct ContentType(pub u8);
+
+    pub(crate) enum ContentTypeName {
         ChangeCipherSpec => 0x14,
         Alert => 0x15,
         Handshake => 0x16,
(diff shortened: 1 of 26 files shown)


rustls/src/enums.rs
```diff
@@ -132,9 +132,8 @@ impl AsRef<[u8]> for ApplicationProtocol<'_> {
 enum_builder! {
     /// The `HandshakeType` TLS protocol enum.  Values in this enum are taken
     /// from the various RFCs covering TLS, and are listed by IANA.
-    /// The `Unknown` item is used when processing unrecognized ordinals.
     #[repr(u8)]
-    pub enum HandshakeType {
+    pub enum HandshakeType names HandshakeTypeName {
```
djc: I don't love hiding the true nature of these types (as in, structs with consts) behind an enum-like macro facade. Can we at least keep the macro input more similar to the output of the macro?

(Bonus points if we can ditch the macro in favor of a sealed trait...)

ctz: Just pushed an alternative to this which more closely matches the output code. WDYT?

> (Bonus points if we can ditch the macro in favor of a sealed trait...)

I think this would require repetition of the enum variants? (Absent proc macros, which I wouldn't consider for this use -- I'd prefer to do code generation.)

djc: > Just pushed an alternative to this which more closely matches the output code. WDYT?

Definitely an improvement. I'd probably sort the `Name` enum before the newtype wrapper in the input, but not a big deal.

> > (Bonus points if we can ditch the macro in favor of a sealed trait...)
> 
> I think this would require repetition of the enum variants? (Absent proc macros, which I wouldn't consider for this use -- I'd prefer to do code generation.)

I might prefer that if we could have generic test code that can exercise the variants being equivalent. Not sure how hard that gets...

ctz: > I might prefer that if we could have generic test code that can exercise the variants being equivalent. Not sure how hard that gets...

That might be a good future route, yes, if we can apply some reflection proc macros to the various types under `cfg(test)`?

PR #2971, opened 2026-03-10
Preparations for new state-based API

ctz:
This is groundwork for #2953

### DIFF
--- a/rustls/src/server/connection.rs
+++ b/rustls/src/server/connection.rs
@@ -429,7 +429,7 @@ impl io::Read for ReadEarlyData<'_> {
 pub struct Accepted {
     // invariant: `connection.core.state` is `Err(_)` and requires restoring
     connection: ConnectionCommon<ServerSide>,
-    choose_config: ChooseConfig,
+    choose_config: Box<ChooseConfig>,
 }
 
 impl Accepted {
--- a/rustls/src/server/hs.rs
+++ b/rustls/src/server/hs.rs
@@ -38,7 +38,7 @@ pub(crate) enum ServerState {
     ReadClientHello(ReadClientHello),
 
     /// Choose a [`ServerConfig`] based on the received ClientHello.
-    ChooseConfig(ChooseConfig),
+    ChooseConfig(Box<ChooseConfig>),
 
     /// Processing the received ClientHello.
     ClientHello(Box<ExpectClientHello>),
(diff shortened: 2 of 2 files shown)


rustls/src/conn/mod.rs
```diff
@@ -739,6 +729,7 @@ impl IoState {
 pub(crate) fn process_new_packets<'a, Side: SideData>(
     input: &mut dyn TlsInputBuffer,
     state: &mut Result<Side::State, Error>,
+    finish: ProcessFinishCondition,
```
ctz: This is always fixed for a given call site, so it could be a generic.

rustls/src/conn/mod.rs
```diff
@@ -851,6 +856,25 @@ impl<Side: SideData> ConnectionCore<Side> {
         }
     }
 
+    pub(crate) fn process_new_packets(
```
djc: I think the abstraction in https://github.com/rustls/rustls/pull/2970/changes/01b03315b523b99b5a79061056b942e4242456b4 is nicer, maybe cherry-pick it here? IMO it would also be good to separate the extraction of another method layer from the introduction of `ProgressFinishCondition`.

rustls/src/conn/mod.rs
```diff
@@ -687,6 +700,24 @@ impl<Side: SideData> DerefMut for ConnectionCommon<Side> {
     }
 }
 
+pub(crate) struct ConnectionBuffers {
```
djc: Is the `Connection` prefix doing any meaningful work here?

rustls/src/common_state.rs
```diff
@@ -101,29 +101,6 @@ pub struct ConnectionOutputs {
 }
 
 impl ConnectionOutputs {
-    pub(crate) fn handle(&mut self, ev: OutputEvent<'_>) {
```
djc, line 103: Should we additionally (but in a separate commit) rename `ConnectionOutputs` (the struct) to `StoredOutputs` (or something) to make the distinction a bit more obvious/usage more explicit?

rustls/src/conn/mod.rs
djc: If we had enums as `const` generics, that might be actually attractive...

ctz: could also do (massively overwrought):

```
trait FinishCondition {
  type State;
  fn start(recv: &ReceivePath) -> Self::State;
  fn continue(state: Self::State, recv: &ReceivePath) -> bool;
}

struct AppData;

impl FinishCondition for AppData {
  type State = ();
  fn start(_recv: &ReceivePath) -> Self::State { () }
  fn continue(_state: Self::State, _recv: &ReceivePath) -> bool { true }
}

etc
```

djc: Hard to say if that makes the code easier to follow...

ctz: I've tried this out, I think it makes the loop condition specifically a bit more readable. WDYT?

djc: Honestly, I've found both versions pretty inscrutable, not least because one of the values ends up being unused here. Maybe it would make more sense to squash this into the commit that uses it in the next PR, and we can iterate on it from there?

rustls/src/common_state.rs
ctz: Going to defer this til later, I think maybe `HandshakeOutputs` for the public type would be more appropriate.

Issue #2741, opened 2025-11-05
ECH Client: Allow passing arbitrary OuterSNI


ckcr4lyf:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
I Was trying to pass a custom OuterSNI (i.e. NOT the `public_name` from the ECHConfig), does not seem possible.

**Describe the solution you'd like**
A way to pass a customer OuterSNI

**Describe alternatives you've considered**
N/A?

**Additional context**
As per the I-D, Section 7.1 (https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-25#section-7.1-10)

```
Once the server has chosen the correct ECHConfig, it MAY verify that the value in the ClientHelloOuter "server_name" extension matches the value of ECHConfig.contents.public_name, and abort with an "illegal_parameter" alert if these do not match. This optional check allows the server to limit ECH connections to only use the public SNI values advertised in its ECHConfigs. The server MUST be careful not to unnecessarily reject connections if the same ECHConfig id or keypair is used in multiple ECHConfigs with distinct public names.
```

However it is not required to be the same. In fact, I've tried to hardcode the value hackily, see: https://github.com/ckcr4lyf/rustls/commit/ac50e7450f0b7d574618c74055b1d86a87af57cb#commitcomment-169688828

and it seems to work with the ECH-supporting NGINX I am running. Would it be possible to let the caller pass some value they want in here? Is the team open to this?

I could try and cook up something, though I am not 100% sure which struct this best lives in (EchMode maybe?)

ckcr4lyf:
This is where the outer_name is cloned from the public_name:

https://github.com/rustls/rustls/blob/6e1cdfcdebd02529b927087816d4ce2559eac91c/rustls/src/client/ech.rs#L378

djc:
What is the use case? What are you trying to achieve here?

ckcr4lyf:
I think clients may find it valuable to use arbitrary outer SNIs to hide in a larger anonymity set, since outer SNIs are what can be observed by eavesdroppers; and if we just use the `public_name` it can more easily be tied to a potential "inner host".

Furthermore since it is allowed by the spec (i.e not abusing it), I thought it would be good to expose control over it to the client.

djc:
Seems reasonable -- I think we'd be open to supporting that. It seems like something that could be a setter on `EchConfig` maybe?

ckcr4lyf:
[...]

PR #2973, opened 2026-03-11
ECH: Allow client to specify customer outer SNI

ckcr4lyf:
Regarding client behavor, [Section 6.1 of RFC9849](https://www.rfc-editor.org/rfc/rfc9849.html#section-6.1-7.5.1) states:

> It SHOULD place the value of ECHConfig.contents.public_name in the "server_name" extension. Clients that do not follow this step, or place a different value in the "server_name" extension, risk breaking the retry mechanism described in [Section 6.1.6](https://www.rfc-editor.org/rfc/rfc9849.html#rejected-ech) or failing to interoperate with servers that require this step to be done; see [Section 7.1](https://www.rfc-editor.org/rfc/rfc9849.html#client-facing-server).

Note the use of `SHOULD` and not `MUST`.

Furthermore, regarding server behavior, [Section 7.1 of RFC9849](https://www.rfc-editor.org/rfc/rfc9849.html#section-7.1-10) states:

> Once the server has chosen the correct ECHConfig, it MAY verify that the value in the ClientHelloOuter "server_name" extension matches the value of ECHConfig.contents.public_name and abort with an "illegal_parameter" alert if these do not match. This optional check allows the server to limit ECH connections to only use the public SNI values advertised in its ECHConfigs.

As such, the client should be able to set a customer outer SNI if it wants, for example to further mask the domain from middleboxes (e.g. set the OuterSNI to `cloudflare-ech.com` even for a website not related to Cloudflare at all).

This is my attempt at adding such a configurable option based on some discussion in #2741 , and since the RFC has now been published.

**Disclaimer: I know some basic rust but not super good with idiomatic rust; so my use of `.clones()` etc may be overkill. sorry about that; open to changes of course.**

### DIFF
--- a/examples/src/bin/ech-client.rs
+++ b/examples/src/bin/ech-client.rs
@@ -92,7 +92,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
                 .find_map(|list| {
                     let ech_config = EchConfig::new(list, ALL_SUPPORTED_SUITES).ok()?;
                     let ech_config = match &args.outer_sni {
-                        Some(sni) => ech_config.with_outer_sni(DnsName::try_from(sni.as_str()).ok()?.to_owned()),
+                        Some(sni) => ech_config.with_outer_sni(
+                            DnsName::try_from(sni.as_str())
+                                .ok()?
+                                .to_owned(),
+                        ),
                         None => ech_config,
                     };
                     Some(ech_config)


ckcr4lyf:
To run the example and se eit in action (have wireshark open for extra verification):

```
cargo run --package rustls-examples --bin ech-client -- --port 3443 --path "plain" rfc9849.mywaifu.best rfc9849.mywaifu.best --outer-sni alphabet-boys.com
```

image

djc:
Thanks for working on this!

I wonder whether setting up ECH via the `ClientConfig` is still the right way? Now that we have a builder for client connections, maybe it would make more sense to set it up via that path, making it easier to (for example) have different outer SNIs for different connections while reusing a common config for other stuff. @ctz and @cpu thoughts?

ctz:
Could you explain in detail what your direct use is for this?

I am pretty unsure about it generally, mainly because the working group is continuing to explore other ways of achieving the same goal:

- https://datatracker.ietf.org/doc/draft-sullivan-tls-signed-ech-updates
- https://datatracker.ietf.org/doc/draft-thomson-tls-ech-pnmasq

I would also note that this could be achieved entirely outside of rustls, by simply processing `ECHConfig` (an encoding established by standard) prior to supplying it to the library (whether on an initial connection or retry).

ckcr4lyf:
Hiya ctz,
> Could you explain in detail what your direct use is for this?

To be fair, my use is entirely academic; try and test different ECH implementations to see how they respond to different public names, see how different censors behave in regards to it, and possibly use it for censorship circumvention. 
My idea was to use rustls for this since the ech client example was already a good starting point; and I thought others may benefit from the feature.

> the working group is continuing to explore other ways of achieving the same goal

I'd say it's similar but not the same; in my case the client decides the OuterSNI to set (e.g. it may want to set `cloudflare-ech.com` for connections it makes), rather than the server advertising it via ECHConfig.

> I would also note that this could be achieved entirely outside of rustls, by simply processing ECHConfig (an encoding established by standard) prior to supplying it to the library (whether on an initial connection or retry).

That is true, after the ECHConfig has been resolved, the client could "fake" / override the public_name in there, but I was kinda thinking overriding the outer SNI could be configurable from the TLS Client without manipulating the advertised ECHConfig.

cpu:
> To be fair, my use is entirely academic; try and test different ECH implementations to see how they respond to different public names, see how different censors behave in regards to it, and possibly use it for censorship circumvention.

IMO this isn't sufficient justification for upstreaming a new configuration knob. I think this sort of research project is better served by a local fork with the changes you require.

ckcr4lyf:
Gotcha, my bad. Thanks for the feedback. If anyone needs it in the future at least hopefully this serves as a reference.

PR #2974, opened 2026-03-11
conn: hoist discarding buffer data

djc:


djc:
Feels surprising that this (the first commit) is worse for performance.

rustls/src/msgs/deframer/buffers.rs
```diff
@@ -147,6 +147,7 @@ pub(crate) trait TlsInputBuffer {
     ///
     /// Rustls guarantees it will not `discard()` more bytes than are returned
     /// from `slice_mut()`.
+    #[expect(dead_code)]
     fn discard(&mut self, num_bytes: usize);
```
ctz, line 185: Not sure I understand this now. Doesn't this mean that all buffers collect data forever and don't throw it away?

djc: I think it's just calling `discard()` on the concrete type, before the cast to `TlsInputBuffer`.

djc:
I had Claude look at the performance effect here, and it suggested an improvement:

> The benchmark regression is not caused by the buffer discard logic change itself. It's caused by LLVM inlining noise: the code size changes in `ReceivePath::process_new_packets()` cause LLVM to change its inlining decisions, resulting in `Tag::drop()` (which zeroizes a 64-byte `[u8; 64]` array) being inlined into every drop site within `tls12::prf()`. Since `prf()` creates/drops multiple `Tag` values per loop iteration and is called several times during a TLS 1.2 handshake, duplicating the zeroize code at each site adds ~24,000 instructions — virtually the entire regression.
>
> The fix is in hmac.rs: adding `#[inline(never)]` to `Tag::drop()`. This is appropriate because:
>
> 1. The zeroize call on a 64-byte buffer is substantial work relative to function call overhead — inlining it provides no benefit
> 2. `Tag::drop()` is called from many sites (every `Tag` value that goes out of scope), so deduplication saves code
> 3. It stabilizes instruction counts against unrelated code layout changes elsewhere

This sounds plausible enough to me, so let's see what the benchmarks say.

rustls/src/msgs/deframer/buffers.rs
ctz: I don't think that's a reasonable idea unfortunately. It means correct operation is happening outside the trait, so another type won't work?

djc: I'm not sure I understand why that is the case. `TlsInputBuffer` has exactly the same API, so as soon as we hoist the concrete buffer out of this function call, we can just call that?

PR #2976, opened 2026-03-12
Avoid inlining tag zeroization

djc:
Suggested by Claude for

- #2974 

> The benchmark regression is not caused by the buffer discard logic change itself. It's caused by LLVM inlining noise: the code size changes in `ReceivePath::process_new_packets()` cause LLVM to change its inlining decisions, resulting in `Tag::drop()` (which zeroizes a 64-byte `[u8; 64]` array) being inlined into every drop site within `tls12::prf()`. Since `prf()` creates/drops multiple `Tag` values per loop iteration and is called several times during a TLS 1.2 handshake, duplicating the zeroize code at each site adds ~24,000 instructions — virtually the entire regression.
>
> The fix is in hmac.rs: adding `#[inline(never)]` to `Tag::drop()`. This is appropriate because:
>
> 1. The zeroize call on a 64-byte buffer is substantial work relative to function call overhead — inlining it provides no benefit
> 2. `Tag::drop()` is called from many sites (every `Tag` value that goes out of scope), so deduplication saves code
> 3. It stabilizes instruction counts against unrelated code layout changes elsewhere

This sounds plausible enough to me, so let's see what the benchmarks say.

### DIFF
--- a/rustls-aws-lc-rs/src/hpke.rs
+++ b/rustls-aws-lc-rs/src/hpke.rs
@@ -916,6 +916,7 @@ impl LabeledSuiteId {
 struct AeadKey<const KEY_LEN: usize>([u8; KEY_LEN]);
 
 impl<const KEY_LEN: usize> Drop for AeadKey<KEY_LEN> {
+    #[inline(never)]
     fn drop(&mut self) {
         self.0.zeroize()
     }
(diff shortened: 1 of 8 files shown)


djc:
So these result in no significant performance effects, although there are some non-significant slowdowns. Do you think it's worth it to take these changes, which might help improve benchmark stability by avoiding inlining flipflopping?

ctz:
does this produce any effect layered onto #2974?

djc:
> does this produce any effect layered onto #2974?

Doing this for just the HMAC tag did seem to make any perf effects from that PR insignificant.

PR #2980, opened 2026-03-17
Tweak SECURITY.md to provide guidance to sloperators

djc:
Anything else we could mention here? I think we're talking to the LLMs themselves as much as we're talking to their operators...


ctz:
Thanks, good stuff.

PR #2981, opened 2026-03-17
webpki: avoid panic when misconfigured without verification algorithms

djc:


ctz (approved):
See also #2984 -- I don't mind which direction we go in.

djc:
> See also #2984 -- I don't mind which direction we go in.

Suggest we do both? This still seems nice as in-depth no-panicking defense, since the invariant isn't locally obvious.

PR #2982, opened 2026-03-17
Apply KU flood limit to all traffic-time handshake messages

ctz:

### DIFF
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -560,33 +560,33 @@ impl Default for TemperCounters {
 }
 
 pub(crate) struct TrafficTemperCounters {
-    allowed_key_update_requests: u8,
+    allowed_consecutive_handshake_messages: u8,
 }
 
 impl TrafficTemperCounters {
-    pub(crate) fn received_key_update_request(&mut self) -> Result<(), Error> {
-        match self.allowed_key_update_requests {
-            0 => Err(PeerMisbehaved::TooManyKeyUpdateRequests.into()),
+    pub(crate) fn received_handshake_message(&mut self) -> Result<(), Error> {
+        match self.allowed_consecutive_handshake_messages {
+            0 => Err(PeerMisbehaved::TooManyConsecutiveHandshakeMessagesAfterHandshake.into()),
             _ => {
-                self.allowed_key_update_requests -= 1;
+                self.allowed_consecutive_handshake_messages -= 1;
                 Ok(())
             }
         }
     }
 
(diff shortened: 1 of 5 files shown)


rustls/src/conn/receive.rs
```diff
@@ -560,22 +560,22 @@ impl Default for TemperCounters {
...
-                self.allowed_key_update_requests -= 1;
+                self.allowed_consecutive_handshake_messages -= 1;
                 Ok(())
             }
         }
     }
 
     pub(crate) fn received_app_data(&mut self) {
-        self.allowed_key_update_requests = Self::INITIAL_KEY_UPDATE_REQUESTS;
+        self.allowed_consecutive_handshake_messages = Self::INITIAL_KEY_UPDATE_REQUESTS;
```
djc: Should we also rename `INITIAL_KEY_UPDATE_REQUESTS`?

Was wondering if we could pull this out of the traffic state and into `ReceivePath` directly. Seems like this could just be in `process_new_packets()`, since even during the handshake we don't expect more than 32 consecutive handshake messages?

ctz: > Was wondering if we could pull this out of the traffic state and into `ReceivePath` directly. Seems like this could just be in `process_new_packets()`, since even during the handshake we don't expect more than 32 consecutive handshake messages?

We just moved it here (5433d04a96d0b33f3566a32c63ba05b26d1c4393), I'd prefer it if behaviours which are specific to a state machine (version and side-specific) are in the state machine, rather than layered on top.

djc: > We just moved it here ([5433d04](https://github.com/rustls/rustls/commit/5433d04a96d0b33f3566a32c63ba05b26d1c4393)), I'd prefer it if behaviours which are specific to a state machine (version and side-specific) are in the state machine, rather than layered on top.

This is not side-specific, right? Does it need to be version-specific?

(if it is both, I agree.)

ctz: It's version-specific in the sense that TLS1.2 doesn't have any post-handshake handshake messages, so there is nothing to count. (As a result, `TrafficTemperCounters` is not constructed in any TLS1.2 files.) Of course, if a TLS1.2 implementation does receive a handshake message, that is permanently fatal to the connection so restricting how many times that can occur isn't that important.

I guess it's side-specific in the sense that the messages involved and their cost are side-specific. (As of this PR everything is treated equally, but in the future extended key update or post-handshake auth could change that.)

djc: Fair enough!

PR #2983, opened 2026-03-17
Check provider DHE `ffdhe_group()` correctness

ctz:


ctz:
> For example, do we obviously not use any crypto provider in the pre-accepted phases of the server acceptor?

Yes, there is no `CryptoProvider` or `ServerConfig` at all at this point.

> (Obviously a 7-argument constructor is also not the greatest.)

Not sure what this is referring to?

djc:
> > (Obviously a 7-argument constructor is also not the greatest.)
> 
> Not sure what this is referring to?

A potential future `CryptoProvider::new()`...

PR #2984, opened 2026-03-17
Enforce that `WebPkiSupportedAlgorithms` are sensible at construction

ctz:

### DIFF
--- a/rustls/src/crypto/mod.rs
+++ b/rustls/src/crypto/mod.rs
@@ -371,17 +371,24 @@ impl Borrow<[&'static Tls13CipherSuite]> for CryptoProvider {
 
 /// Describes which `webpki` signature verification algorithms are supported and
 /// how they map to TLS [`SignatureScheme`]s.
-#[expect(clippy::exhaustive_structs)]
+///
+/// Create one with [`WebPkiSupportedAlgorithms::new`], which can be done in const-context.
 #[derive(Clone, Copy)]
 pub struct WebPkiSupportedAlgorithms {
     /// A list of all supported signature verification algorithms.
     ///
     /// Used for verifying certificate chains.
     ///
-    /// The order of this list is not significant.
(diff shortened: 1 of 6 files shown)


rustls/src/crypto/mod.rs
```diff
@@ -392,13 +399,34 @@ pub struct WebPkiSupportedAlgorithms {
...
+    pub const fn new(
+        all: &'static [&'static dyn SignatureVerificationAlgorithm],
+        mapping: &'static [(
+            SignatureScheme,
+            &'static [&'static dyn SignatureVerificationAlgorithm],
+        )],
+    ) -> Self {
+        let s = Self { all, mapping };
+        assert!(!all.is_empty());
+        assert!(!mapping.is_empty());
```
djc: Suggest yielding `ApiMisuse` for these instead? If not, we should document the function to panic. (Assuming we can call `unwrap()` in `const` I guess.)

rustls/src/crypto/mod.rs
```diff
@@ -392,13 +399,34 @@ pub struct WebPkiSupportedAlgorithms {
...
+
+        let mut i = 0;
+        while i < s.mapping.len() {
+            assert!(!s.mapping[i].1.is_empty());
+            i += 1;
+        }
```
djc, line 433: This looks pretty old-school C-like... Is that to make it `const` compatible? Would be nice to use something a little more idiomatic if possible.

ctz: Yeah, no iterators or for loops in const

djc: > Yeah, no iterators or for loops in const

Maybe add a comment about that to make it a little more obvious.

rustls/src/crypto/mod.rs
ctz: Unfortunately no `unwrap()`, but can be manually inlined as match is const.

PR #2985, opened 2026-03-17
docs: fix typo in SECURITY.md security bug advice

cpu:
Follow-up to https://github.com/rustls/rustls/pull/2980


djc (approved):
Oops, thanks for catching & fixing!

PR #2986, opened 2026-03-18
SECURITY.md: add threat model

ctz:
Happy to hear any feedback on this.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -61,7 +61,7 @@ MSRV than this policy.
 
 Before reporting a security bug, make sure to:
 
-- Consider the threat model. Misconfiguration that is unlikely to happen accidentally is
+- Consider the threat model below. Misconfiguration that is unlikely to happen accidentally is
   unlikely to be a security bug.
 - If applicable, compare the behavior to other TLS implementations. If the behavior is consistent
   with other implementations, it is less likely to be a security bug.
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -78,3 +78,112 @@ We'll then:
...
+Mitigations:
+
+- The entire crate which processes items on this trust boundary is `forbid(unsafe_code)`.  This means all
+  code within is the memory safe-subset of Rust.  This ameliorates impact of items like integer overflows (generally
+  reducing their impact to denial-of-service), but has little impact on other threats.
+- We fuzz this interface, looking for reachable panics.  The project is registered with OSS-Fuzz which provides
+  compute for this effort.  Fuzzing is performed with a mock provider of cryptography, which is intended to make
+  both pre-auth and post-auth code paths reachable to the fuzzer (at the cost of fuzzing not covering the actual
+  cryptography implementations).
+- We have [studied and explain](https://rustls.dev/docs/manual/_01_impl_vulnerabilities/index.html#a-review-of-tls-implementation-vulnerabilities)
```
cpu: typo: "explain" -> "explained" (?)

SECURITY.md
```diff
@@ -78,3 +78,112 @@ We'll then:
...
+
+- random material generation
+- cryptography operations (encryption, decryption, hashing, key agreement, key derivation, signing, verification)
+- zeroization of long-term, ephemeral and intermediate secret key material
+- side-channel safety of cryptographic operations
+- correct error reporting (especially for cases such as signature verification and input validation)
+
+The core `rustls` crate cannot perform securely if these items are faulty, and it is not
+a security finding that (for example) rustls does not detect random material generation which produces
+the number 4 repeatedly.
```
cpu, line 152: even if chosen by a fair dice roll !?! 🎲

SECURITY.md
```diff
@@ -78,3 +78,112 @@ We'll then:
...
+### Boundary: network-originated input
+
+Everything arriving from the wire is treated as adversarially crafted. This is the primary attack surface.
+
+Specific threats (non-exhaustive):
+
+- Integer overflow or underflow in length fields,
+- Buffer over-read during fragment reassembly,
+- Infinite loops, 
+- Reachable loops with inappropriate and attacker-controlled complexity,
```
cpu: Is there perhaps more to say w.r.t this topic (or the memory consumption part)?

I'm thinking something along the lines of needing there to be an amplification, or some increase that isn't itself dwarfed by other normal operations (network overhead, cryptographic overhead, TLS handshake resources, etc).

SECURITY.md
```diff
@@ -61,7 +61,7 @@ MSRV than this policy.
 
 Before reporting a security bug, make sure to:
 
-- Consider the threat model. Misconfiguration that is unlikely to happen accidentally is
+- Consider the threat model. Misconfiguration that is unlikely to be happen accidentally is
   unlikely to be a security bug.
```
cpu: Could this section of the pre-existing text have some sort of forward pointer to the Threat model section below?  Ideally a security report would explicitly reference the boundary, or control, being bypassed and the implication under the threat model.

SECURITY.md
```diff
@@ -78,3 +78,112 @@ We'll then:
...
+
+In addition, the core `rustls` crate itself:
+
+- performs zeroization of key material values it holds,
+- compares secret and public values in constant time,
+- correctly propagates and handles errors,
+- avoids `Debug` impls on any type that contains secret key material,
+- eschews support for problematic protocol features such as RSA encryption, and CBC-mode ciphersuites.
+
+### Boundary: public API
```
cpu, line 162: I'm not sure whether it's A) relevant enough to mention B) a good fit for this particular section, but I wonder whether we should have some mention of trusting the host on which our code is executed. E.g. that something like attaching a BPF filter, or mucking around with the Rust<->OS interfaces is out of scope.

SECURITY.md
```diff
@@ -78,3 +78,112 @@ We'll then:
...
+The core `rustls` library is a TLS protocol implementation, and requires additional items to
+become useful.  These items are therefore in scope:
+
+- the default certificate verifiers based on `rustls-webpki`.
+- the cryptography providers published from the rustls repository (currently `rustls-ring` and `rustls-aws-lc-rs`;
+  but not the underlying libraries).
+
+These items are out of scope (security reports for them will be treated as normal bug reports):
+
+- example code and test code,
```
cpu: perhaps worth mentioning benchmarking code explicitly as well (though arguably I suppose you could call that "test code")

SECURITY.md
```diff
@@ -78,3 +78,112 @@ We'll then:
...
+misuse-resistant.
+
+By "fully attacker-controlled", we specifically include:
+
+- on-path attackers (manipulating traffic to a real, honest peer)
+- malicious peers (whether pre- or post-authentication)
+- honest but broken peers (again, pre- or post-authentication)
+
+The core `rustls` library is a TLS protocol implementation, and requires additional items to
+become useful.  These items are therefore in scope:
```
cpu, line 97: What about other Rustls ecosystem crates? Thinking of things like rustls-platform-verifier, tokio-rustls, hyper-rustls, etc. 

I suppose it's a bit muddy and maybe we should have forward pointers from those repos (and confirm they are set up to take GHSA reports), or instead document in those crates that we prefer reports to go to rustls/rustls?

ctz: Enabled private vuln reporting for rustls-platform-verifier, tokio-rustls, hyper-rustls & openssl-probe.

SECURITY.md
ctz: I think it might be a bit much to discuss that here. In a similar vein would be unsound or memory-unsafe code existing in the same memory space as rustls. Generally I would describe both as: beyond our control and responsibility.

SECURITY.md
ctz: I think this is a bit tricky. It would be nice to have uniform across the rustls org. But there are projects here that are more maintainable and maintained than others, and with different assurance needs.

cpu: Agreed, punting on that is OK with me.

PR #2987, opened 2026-03-19
server: prefer ChaCha20 when it is the client's first preference.

brian-pane:


brian-pane:
This addresses issue #[2252](https://github.com/rustls/rustls/issues/2252): "Support prioritizing ChaCha20 for clients without AES hardware acceleration."

The selection logic is based on [this patch](https://github.com/james58899/rustls/commit/ab9ea616d13e745f18139ff44e5fde61c525b7e2) by @[james58899](https://github.com/james58899)

rustls/src/server/config.rs
```diff
@@ -87,6 +87,9 @@ pub struct ServerConfig {
     /// which is supported by the client.
     pub ignore_client_order: bool,
```
brian-pane: Ideally I'd like to remove this, because the same thing can now be done more cleanly by making a implementation of the new `CipherSuiteSelector` trait that prefers the client ordering. Would a breaking change be okay here?

djc: Yes, the `main` branch is currently open for breaking changes.

djc (commented):
(I definitely think `ignore_client_order` argument should be removed.)

rustls/src/server/config.rs
ctz: Yeah `ignore_client_order` should be removed in this PR. I think for the time being, and given the improved defaults given by this PR, it is OK to not support that option via a pre-baked `CipherSuiteSelector`

brian-pane: Thanks, I've removed `ignore_client_order`.

brian-pane:
The CI failure on `rustls / audit` looks like a pre-existing condition: there's a dependency on a version of the `digest` crate that's recently been yanked from crates.io.

djc:
Yup, feel free to ignore that.

djc:
So my question is if it would be feasible/desirable to combine this with the `ServerCredentialResolver` trait? If I'm not mistaken, these are called at roughly the same time.

```rust
        // Choose a certificate.
        let credentials = self
            .config
            .cert_resolver
            .resolve(&ClientHello::new(
                &input,
                &sig_schemes,
                sni.as_ref(),
                T::VERSION,
            ))?;
        self.sni = sni;

        let (suite, skxg) = self.choose_suite_and_kx_group(
            suites,
            credentials.signer.scheme(),
            input
                .client_hello
                .named_groups
                .as_deref()
                .unwrap_or_default(),
            &input.client_hello.cipher_suites,
        )?;
```

brian-pane:
> So my question is if it would be feasible/desirable to combine this with the `ServerCredentialResolver` trait? If I'm not mistaken, these are called at roughly the same time.

The two traits are used very close together, but I anticipate their implementations may evolve independently.

`ServerCredentialResolver` currently has _N_ = 8 implementations:
* `SingleCredential` for simple configurations with one server credential chain
* `ServerNameResolver` for SNI support
* and half a dozen different special cases for testing

With this PR, there's just one implementation of `CipherSuiteSelector` so far, but I can easily see that growing to _M_ >= 4 in the future, with new variants like these (which could live either inside or outside rustls) :
* `ClientPreferenceCipherSuiteSelector` that prefers the client's choice of cipher suite
* `FilteredCipherSuiteSelector` that lets the server operator disable cipher suites supported by the `CryptoProvider` (useful when there's a future vulnerability or implementation bug found in some cipher suite, and operators need to roll out a config fix to turn it off quickly, for example)
* and special cases for testing.

Having the `ServerCredentialResolver` and `CipherSuiteSelector` as separate traits would limit the number of potential implementations to _N_ + _M_ instead of _N_ x _M_.

ctz:
> So my question is if it would be feasible/desirable to combine this with the `ServerCredentialResolver` trait? If I'm not mistaken, these are called at roughly the same time.

two things:

a) they have a data dependency (chosen ciphersuite is used to alter the input to the `ServerCredentialResolver`); theoretically that could be addressed with careful API design
b) they are pretty different configuration surfaces, really, and squashing them together seems like it would cause that configuration surface to be present repeatedly on implementations (this is restating and agreeing with the point made above)

djc:
> two things:
> 
> a) they have a data dependency (chosen ciphersuite is used to alter the input to the `ServerCredentialResolver`); theoretically that could be addressed with careful API design b) they are pretty different configuration surfaces, really, and squashing them together seems like it would cause that configuration surface to be present repeatedly on implementations (this is restating and agreeing with the point made above)

Fair enough.

rustls/src/server/config.rs
```diff
@@ -695,3 +693,71 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
         })
     }
 }
+
+/// A filter that chooses the cipher suite to use for a TLS session.
+pub trait CipherSuiteSelector: Debug + Send + Sync {
```
djc: Nit: the trait definition should go below its implementor.

rustls/src/server/config.rs
```diff
@@ -695,3 +693,71 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
+            CipherSuite::TLS_PSK_WITH_CHACHA20_POLY1305_SHA256,
+        ];
+
+        if client_suites
+            .first()
+            .is_some_and(|suite| CHACHA20_SUITES.contains(suite))
+        {
+            // Make a copy of the server's supported cipher suite list with the ChaCha20 ciphers
+            // moved to the front, and then find the first match with the client's list.
+            let mut prioritized_suites: Vec<SupportedCipherSuite> =
```
djc: Feels like it should be possible to do this without allocating?

rustls/src/server/hs.rs
```diff
@@ -595,7 +600,10 @@ impl ExpectClientHello {
         sig_scheme: SignatureScheme,
         client_groups: &[NamedGroup],
         client_suites: &[CipherSuite],
-    ) -> Result<(&'static T, &'static dyn SupportedKxGroup), PeerIncompatible> {
+    ) -> Result<(SupportedCipherSuite, &'static dyn SupportedKxGroup), PeerIncompatible>
```
djc: This doesn't seem very nice. Perhaps the trait should have different methods for 1.2 and 1.3?

rustls/src/server/test.rs
```diff
@@ -123,6 +124,99 @@ fn test_process_client_hello(hello: ClientHelloPayload) -> Result<(), Error> {
     .map(|_| ())
 }
 
+// Process the `ClientHelloPayload` and return the `CipherSuite` from the resulting ServerHello.
+fn select_cipher_suite(
```
djc: Nit: the helper function should be defined below its callers.

rustls/src/server/config.rs
brian-pane: Yes, I'll push an update to eliminate the allocation.

rustls/src/server/hs.rs
brian-pane: I don't follow the part about having different methods for 1.2 and 1.3. Is there a use case where the cipher suite selection logic will need to differ based on the protocol version?

djc: I mean that I don't like that this unspecializes this method from working on one cipher suite type two working on both types of cipher suites at the same time. I guess we don't even need two different methods, as long as you the trait method abstracts over `T: Suite`?

brian-pane: I experimented with abstracting the trait method over `T: Suite`, like this:
```
pub trait CipherSuiteSelector: Debug + Send + Sync {
    /// Choose a cipher suite, given the server's and client's order lists of supported options.
    fn select_cipher_suite<T: Suite>(
        &self,
        client_suites: &[CipherSuite],
        server_suites: &[&'static T],
    ) -> Option<&'static T>;
}
```
but doing that prevents the trait from being `dyn` compatible, which in turn prevents putting a reference to the trait in the `ServerConfig`.

Creating two separate methods in the trait for 1.2 and 1.3 would fix that but would create a new problem: `choose_suite_and_kx_group` would need to decide which method to call based on the type `T: Suite` with which it's instantiated, which would violate the open-closed principle. (I.e., every time a new version of TLS is added, that function will have to change to call a corresponding new trait method.)

djc: > Creating two separate methods in the trait for 1.2 and 1.3 would fix that but would create a new problem: `choose_suite_and_kx_group` would need to decide which method to call based on the type `T: Suite` with which it's instantiated, which would violate the open-closed principle. (I.e., every time a new version of TLS is added, that function will have to change to call a corresponding new trait method.)

Not sure that's a big deal in practice. @ctz thoughts?

brian-pane: Okay, I just pushed an update that uses separate methods for 1.2 and 1.3 in the trait.

brian-pane:
I have an alternative structure in mind that might be a little cleaner.

In the current revision, the selector API has a separate method per `Suite` subclass:
```
pub trait CipherSuiteSelector: Debug + Send + Sync {
    /// Choose a cipher suite, given the server's and client's order lists of supported options.
    fn select_tls12_cipher_suite(
        &self,
        client_suites: &[CipherSuite],
        server_suites: &[&'static Tls12CipherSuite],
    ) -> Option<&'static Tls12CipherSuite>;

    /// Choose a cipher suite, given the server's and client's order lists of supported options.
    fn select_tls13_cipher_suite(
        &self,
        client_suites: &[CipherSuite],
        server_suites: &[&'static Tls13CipherSuite],
    ) -> Option<&'static Tls13CipherSuite>;
}
```
and the `ServerConfig` contains a single `CipherSuiteSelector`:
```
struct ServerConfig {
    // ...
    cipher_suite_selector: &dyn CipherSuiteSelector,
}
```

The alternative would have just one method, but the `CipherSuiteSelector` would be specialized on `T: Suite`:
```
pub trait CipherSuiteSelector<T: Suite>: Debug + Send + Sync {
    /// Choose a cipher suite, given the server's and client's order lists of supported options.
    fn select_cipher_suite(
        &self,
        client_suites: &[CipherSuite],
        server_suites: &[&'static T],
    ) -> Option<&'static T>;
}
```
and then `ServerConfig` would hold a separate `CipherSuiteSelector` per TLS version:
```
struct ServerConfig {
    // ...
    cipher_suite_selector_tls12: &dyn CipherSuiteSelector<Tls12CipherSuite>,
    cipher_suite_selector_tls13: &dyn CipherSuiteSelector<Tls13CipherSuite>,
}
```

With that alternative design, the details of choosing whether to use `config.cipher_suite_selector_tls12` or `config.cipher_suite_selector_tls13` would be encapsulated in the `CipherSuiteSelectorDispatcher`.

Thoughts on which is better?

djc:
I think I slightly prefer the former design, since IMO it better limits the blast radius of the additional complexity from selecting for different versions separately. @ctz thoughts?

ctz:
I prefer the first option - two methods on one trait, in one slot in `ServerConfig`

brian-pane:
Thanks. In that case, this is ready for review.

rustls/src/server/config.rs
```diff
@@ -695,3 +694,86 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
     }
 }
+
+/// A simple implementation of CipherSuiteSelector that balances client and server preferences.
+///
+/// If the server is configured to ignore the client's cipher suite ordering, but the first
+/// cipher suite in the client's preferred list uses ChaCha20, the server prioritizes ChaCha20
+/// cipher suites.
+#[derive(Debug)]
+struct DefaultCipherSuiteSelector;
```
djc, line 704: I think this PR should contain two commits:

- One to introduce the new trait-based API
- One that changes the cipher suite selection algorithm to prefer ChaCha20 where appropriate

(Can also do two different PRs.)

rustls/src/server/config.rs
```diff
@@ -695,3 +694,86 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
+        {
+            // First, look for a match among the ChaCha20 ciphers supported by the server.
+            if let Some(suite) = server_suites
+                .iter()
+                .filter(|server_suite| CHACHA20_SUITES.contains(&server_suite.suite()))
+                .find(|server_suite| client_suites.contains(&server_suite.suite()))
+                .copied()
+            {
+                return Some(suite);
+            }
```
djc: This seems like a fairly roundabout way of expressing what it is trying to do. Maybe try doing it without all the combinators and seeing if that could be more concise/easier to understand at a glance?

rustls/src/server/hs.rs
```diff
@@ -806,3 +806,32 @@ impl HandshakeHashOrBuffer {
...
+
+impl CipherSuiteSelectorDispatcher<Self> for Tls13CipherSuite {
+    fn select_cipher_suite(
+        selector: &dyn CipherSuiteSelector,
+        client_suites: &[CipherSuite],
+        server_suites: &[&'static Self],
+    ) -> Option<&'static Self> {
+        selector.select_tls13_cipher_suite(client_suites, server_suites)
+    }
+}
```
djc, line 837: Okay, this seems fairly ugly -- I better understand your alternative proposal now. I wonder if we can make a middle ground, like this?

```rust
pub trait CipherSuiteSelector: Tls12CipherSuiteSelector + Tls13CipherSuiteSelect {}
```

I seem to remember that, while `dyn` objects can only have a single trait bound, you can still have multiple super traits.

rustls/src/server/config.rs
```diff
@@ -695,3 +694,86 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
+        const CHACHA20_SUITES: [CipherSuite; 8] = [
+            CipherSuite::TLS13_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_ECDHE_PSK_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_DHE_PSK_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_RSA_PSK_WITH_CHACHA20_POLY1305_SHA256,
+            CipherSuite::TLS_PSK_WITH_CHACHA20_POLY1305_SHA256,
+        ];
```
ctz: I would recast this as `pub(crate) fn is_chacha20(&self) -> bool { matches!(cs, Self::TLS13_CHACHA20_POLY1305_SHA256 | etc) }` on the `CipherSuite` type

rustls/src/server/config.rs
brian-pane: > I think this PR should contain two commits:
> 
>     * One to introduce the new trait-based API
> 
>     * One that changes the cipher suite selection algorithm to prefer ChaCha20 where appropriate

Does that first commit still need to support `ignore_client_order`?

rustls/src/server/config.rs
brian-pane: Do you anticipate any future use cases that would need to check whether a cipher suite uses, say, AES? If so, I could generalize that to `pub(crate) fn get_stream_cipher_family(&self) -> StreamCipherFamily` where `StreamCipherFamily` is an enum with values `AES`, `CHACHA20`, etc. If not, I'll just do the `is_chacha20`.

rustls/src/server/config.rs
djc: I think it makes sense to remove support for `ignore_client_order` in the same commit as introducing the trait-based API -- I was trying to be agnostic about the ordering of the two commits.

rustls/src/server/config.rs
ctz: As it's an internal API, I would do the simplest option to start with.

djc: I think we should keep it simple for now.

brian-pane:
I split this into two commits, one that introduces the `CipherSuiteSelector` trait and another that prioritizes ChaCha20.

There is one new CI [failure](https://github.com/rustls/rustls/actions/runs/23551935041/job/68568029058?pr=2987), but it seems to be an unrelated environmental issue.

rustls/src/server/config.rs
brian-pane: In the latest revision, this part became simpler. How does the readability look in the new version?

rustls/src/server/hs.rs
brian-pane: I tried prototyping the `pub trait CipherSuiteSelector: Tls12CipherSuiteSelector + Tls13CipherSuiteSelect` approach. I ran into the same problem, though: it still needed something like the `CipherSuiteSelectorDispatcher` to figure how to call the right `CipherSuiteSelector` method based on the `Suite` subclass.

brian-pane:
Closing this now that the `CipherSuiteSelector` interface has landed separately in #3012

PR #2989, opened 2026-03-20
Add RA-TLS 0xFFBB custom TLS extension for SGX attestation

bfoing:
## Summary

This PR adds support for a custom TLS extension (type 0xFFBB) used by Intel SGX Remote Attestation TLS (RA-TLS). This enables the TLS handshake to carry an attestation challenge from the client and an SGX quote from the server, allowing mutual attestation during connection establishment.

## Motivation

In confidential computing environments using Intel SGX, the TLS handshake is the natural point to perform remote attestation. The client sends a challenge nonce via the 0xFFBB extension in ClientHello, and the server responds with its SGX quote in the EncryptedExtensions. This allows the client to verify that it is communicating with a genuine SGX enclave before exchanging any application data.

This is part of the RA-TLS protocol described in the Intel SGX documentation and used by frameworks such as Gramine and Teaclave.

## Changes

- Added 0xFFBB extension type constant
- ClientHello: allow sending an arbitrary challenge payload in the 0xFFBB extension
- Server: extract the challenge from ClientHello and expose it to the application
- Server: allow the application to attach a quote (attestation response) in EncryptedExtensions via the 0xFFBB extension
- Client: extract the server quote from EncryptedExtensions and expose it for verification

## Compatibility

- The extension is only sent when explicitly enabled by the application
- No changes to default behaviour — existing applications are unaffected
- Uses the experimental/private-use extension range (0xFFBB)

## Related

- Intel SGX RA-TLS: https://arxiv.org/abs/1801.05863
- Gramine RA-TLS: https://gramine.readthedocs.io/en/stable/attestation.html

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -169,6 +169,7 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
             max_fragment_size: None,
             client_auth_cert_resolver,
             versions: self.state.versions,
+            ratls_challenge: None,
             enable_sni: true,
             verifier: self.state.verifier,
             key_log: Arc::new(NoKeyLog {}),
--- a/rustls/src/client/client_conn.rs
+++ b/rustls/src/client/client_conn.rs
@@ -242,6 +242,15 @@ pub struct ClientConfig {
     #[cfg(feature = "tls12")]
     pub require_ems: bool,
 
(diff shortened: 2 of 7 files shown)


djc:
Is this intended for mainline rustls inclusion? I think it's way too niche for that (and this seems to be targeted at the 0.23 release branch while making semver-incompatible changes, which is obviously not going to fly).

It's probably better to file an issue to discuss the potential for changes to make rustls extensible enough to support this kind of thing out of tree.

PR #2990, opened 2026-03-20
Enable the key log file tests to run on macOS

brian-pane:


rustls-util/src/key_log_file.rs
```diff
@@ -155,7 +155,14 @@ mod tests {
     #[test]
     fn test_env_var_cannot_be_written() {
         init();
-        let mut inner = KeyLogFileInner::new(Some("/dev/full".into()));
+
+        #[cfg(target_os = "linux")]
+        const UNWRITABLE_FILE: &str = "/dev/full";
+
+        #[cfg(target_os = "macos")]
+        const UNWRITABLE_FILE: &str = "/dev/urandom";
```
brian-pane, line 163: In my testing, this pseudo-file is unwritable even by root on macOS.

djc:
Did you see this test failing on macOS? Because I have not seen it fail despite doing all of my development on macOS.

(The audit failure should go away when you rebase on top of #2994.)

ctz:
> Did you see this test failing on macOS? Because I have not seen it fail despite doing all of my development on macOS.

Prior to this PR the test is gated on `target_os = "linux"`

PR #2993, opened 2026-03-21
server encrypted client hello support

jellevandenhooff:
## Current state

This PR implements support for TLS encrypted hello, or ECH, as described in [RFC 9849](https://datatracker.ietf.org/doc/rfc9849/) for rustls servers. It implements support for ECH in shared mode where the rustls server decrypts the ECH (frontend) and also handles the TLS connection (backend). The code is setup for later implementation of split mode where the frontend runs on a SNI-inspecting reverse proxy.

The code tries to implement the spec faithfully and to adhere to existing rustls style. The logic implemented in this PR, roughly split out by commit, is all that is necessary to run ECH on the server:

- Wire format wrangling of ECH. The server-side implementation of ECH must do byte-level surgery to reconstruct the inner client hello. This PR introduces new machinery for working with raw client hello messages.
- ECH key management. Trial decryption of ECH messages requires finding the right key given a configuration ID, and the PR has an API for both providing keys and internal O(1) lookup of the right key.
- Trial decryption of ECH messages. To decrypt ECH you need to construct AAD, try the appropriate key(s) and gracefully fail.
- A state machine to handle ECH decryption with hello retry support, where the key gets re-used and state carries over.
- Integration of ECH with the existing TLS 1.3 handshake building on the state machine.
- A new status and configuration API (see below).
- API tests and support in the bogo shim for running ECH tests. All tests pass, except those expect split-mode support.

For servers that do not have ECH configured this PR changes behavior from before slightly: if ECH is offered in a GREASE extension, the support will respond with GREASE ECH in a hello retry request. This is so an observer on the wire cannot tell if the server does or does not have ECH configured. I think making this a non-configurable default is good for the TLS ecosystem. 

### New API surface

```rs
// --- Configuration ---

  // New field on ServerConfig:
  pub struct ServerConfig {
      /// ECH server keys for decrypting Encrypted Client Hello offers.
      pub ech_keys: Arc<dyn EchKeyResolver>,
      // ...
  }

  // --- Key management ---

  /// Trait for dynamic ECH key resolution (e.g. hot-swap on rotation).
  pub trait EchKeyResolver: Debug + Send + Sync {
      fn resolve(&self) -> EchKeys;
  }

  /// Immutable set of ECH server keys, cheap to clone (Arc'd).
  #[derive(Clone, Debug)]
  pub struct EchKeys;

  impl EchKeys {
      pub fn new(keys: Vec<EchServerKey>) -> Self;
      pub fn empty() -> Self;
  }

  /// Simplest EchKeyResolver: wraps a fixed EchKeys.
  pub struct FixedEchKeys;

  impl FixedEchKeys {
      pub fn new(keys: Vec<EchServerKey>) -> Self;
      pub fn empty() -> Self;
  }

  impl EchKeyResolver for FixedEchKeys { ... }
  impl From<Vec<EchServerKey>> for FixedEchKeys { ... }

  /// A server-side ECH key: ECHConfig + HPKE private key.
  pub struct EchServerKey;

  impl EchServerKey {
      /// Parse from raw ECHConfig bytes + private key.
      pub fn from_raw(
          config_bytes: &[u8],
          private_key_bytes: Vec<u8>,
          hpke_suites: &[&'static dyn Hpke],
      ) -> Result<Self, Error>;

      /// Exclude from retry_configs on rejection (for key rotation).
      pub fn with_retry(self, is_retry: bool) -> Self;
  }

  // --- Handshake introspection ---

  /// New method on ClientHello (in cert resolver callback):
  impl ClientHello<'_> {
      /// True when this is the decrypted ECH inner hello.
      /// Returning Err triggers ECH rewind to the outer hello.
      pub fn is_ech_inner(&self) -> bool;
  }

  /// New method on ServerConnection:
  impl ServerConnection {
      /// Returns ECH frontend info if ECH was accepted (shared mode).
      pub fn ech_accepted(&self) -> Option<&EchFrontendInfo>;
  }

  /// Same on quic::ServerConnection:
  impl quic::ServerConnection {
      pub fn ech_accepted(&self) -> Option<&EchFrontendInfo>;
  }

  /// Information from successful ECH termination.
  #[non_exhaustive]
  #[derive(Clone, Debug)]
  pub struct EchFrontendInfo;

  impl EchFrontendInfo {
      pub fn config_id(&self) -> u8;
      pub fn outer_server_name(&self) -> Option<&DnsName<'_>>;
  }
  ```

## Old description 

This PR introduces encrypted client hello support addressing #1980. Caveat: I wrote a lot of this PR with Claude... but I think it's pretty good. 

This implementation supports ECH in both shared mode and split mode and works with tls-over-tcp and quic. A demo showing all this runs at https://ech.jelle.dev and works with Chrome and Firefox on my laptop (when I point them at a DoH resolver). The code passes all ECH bogo tests, and each commit should pass format, lint, and tests by itself. Support for ECH is built up over small commits that slowly add more features. Let me know if you'd like more work on this, I'd be interested to get it to a place that it might get merged. At the same time, I also understand it's a lot of code you might want to instead work on yourselves.

### Design decisions and questions

There's a couple of things that I kind of decided along the way that might or might not be good ideas. Happy to change those.

- The backend always supports running as a split-mode backend, accepting decrypted inner client hellos. The RFC says not to do this but I think other libraries also do this. Making this a choice would require a configuration knob specifying to run as a backend or not.

- Rewind support, which is what happens when the application decides it does not like the SNI in the inner hello and rustls reverts back to the outer hello, is implemented by treating any errors by the certificate resolver as a request to rewind. I think that's a fine choice and I'd rather not introduce new APIs to otherwise approve or reject an inner hello.

- There is no check on the outer hello's SNI, so if a client sends eg. google.com, even though the backend has no certificate for that domain, it'd still get accepted.

- There is a somewhat fairly detailed status API to get information on the ECH handshake (eg. it shows the used config ID.) That seems quite nice for monitoring how, for example, HTTPS DNS records get propagated, but I don't know exactly what the API should look like.

- The API for getting ECH keys mirrors the certificate getting API.

- There is proxy support in this PR. Maybe that should not yet go in there. Although the API seems slim and good enough for the demo perhaps it's a bit early.

- The (re-)constructing of the ECH is a bunch of gnarly wrangling of bytes. I don't quite know how to make that cleaner or less scary.

### Demo details

The demo's code is a bit of a mess but still perhaps helpful to understand how the proxy API in this PR is used. The code lives at https://github.com/jellevandenhooff/rustls-server-ech-demo and depends on some small changes to [tokio-rustls](https://github.com/jellevandenhooff/tokio-rustls/tree/rustls-ech), [quinn](https://github.com/jellevandenhooff/quinn/tree/rustls-ech) and [h3](https://github.com/jellevandenhooff/h3/tree/rustls-ech). The split mode proxy for TCP is quite simple (https://github.com/jellevandenhooff/rustls-server-ech-demo/blob/main/src/split_proxy.rs#L111): to proxy a connection you decode one or two hello messages, pass them to rustls, and get back the message you should send to the backend. Proxying quic is quite a bit more hairy and I don't fully trust the code yet but it seems to work.

If I go to https://proxied-a.ech.jelle.dev/ in Chrome I get the following status:
```
Proxy metadata:
Outer SNI: public.ech.jelle.dev
Inner SNI (decrypted): proxied-a.ech.jelle.dev
Proxy ECH result: decrypted (ECH accepted)
Route mode: split (inner CH forwarded)

Backend metadata:
Protocol: QUIC (HTTP/3)
server_name (SNI): proxied-a.ech.jelle.dev
ECH status: accepted (split mode)
```
This means that the connection is over quic and is using ECH ("ECP accepted", different outer SNI from inner SNI). This is running in split routing mode, which means the proxy has decrypted the ECH and forwarded the inner ECH to the appropriate backend. That backend does know it's using ECH because the hello has a different shape from a normal hello.
(And the reason I am excited about split mode is because it splits responsibilities and secrecy nicely: the backends do not get the ECH keys, so they cannot decode other requests going into the proxy; meanwhile, the proxy does not decrypt the actual traffic.)


djc:
Do you have a (serious) use case for this? What are you trying to achieve?

jellevandenhooff:
Serious is a tricky question. I would like to run a reverse proxy, and I would like for it to use ECH. Is my anonymity set going to be anywhere near as large as Cloudflare? If this code merged, I would use it. If it did not, I might use a fork, or I might decide to not care.

My specific use case is that I would like to offer people personal (sub-)domains that I reverse proxy to their personal servers (think self-hosting), without me decrypting their TLS, and without anyone inspecting SNI being able to see who is trying to access anyone else's page.

jellevandenhooff:
I cleaned up the commit stack a little more. Now the "server: integrate ECH into TLS 1.3 handshake" commit is a bit smaller because I moved out the application-facing status API and some HelloRetryRequest plumbing.

Like I tried to say before, I am open to all kinds of feedback and/or approaches with this PR. If you want to skim the commits and their structure and give high-level feedback, that works, or if you'd want me to read everything line by line and make sure there's no more LLM-isms before you take a look at it, that works also. I can split things into several PRs so that each commit goes by itself and we can talk about them on their own; I haven't done that yet because I figured high-level feedback would be helpful earlier. I would be excited to contribute something meaningful to the TLS ecosystem and am flexible in approach.

Thanks.

djc:
In skimming some of the early commits, here is some high-level feedback:

- should fix code style to match the style guide in CONTRIBUTING.md, notably on item ordering (top-down)
- looks like there are lots of free functions, where it might make more sense to attach things to types
- lots of u8 slices as opposed to more strongly typed things
- for adding code, splitting into different commits tends to have limited value
- on other hand, changes to existing code should ideally be factored out into separate commits where possible

jellevandenhooff:
Thanks for taking a look. I believe I addressed your comments:
- The ordering in `ech.rs` is now top-down (or at least my understanding of top-down! sorry, I am used to go ordering).
- The free functions are gone and logic now lives in appropriate impl blocksl
- The u8 slices are gone (though some in tests remain). If this is too much new types happy to change.
- All code in `ech.rs` is now in one big commit (besides the proxy stuff). The wiring up of the ECH logic in the actual handshake is still spread over multiple commits (first basic, then API, retry, and rewind). If you'd prefer those also combined I can do that as well, though at that point maybe the whole PR view would show the same.

I appreciate the feedback; I think `ech.rs` is a lot more legible now.

iadev09:
We built a small local testbed around this PR and got the server-side ECH path working successfully, including a reqwest-based H3 client path and a simpler H2 comparison path.

Looking forward to this being merged. We used this local harness while testing, in case it is useful:

https://github.com/iadev09/ech-server

jellevandenhooff:
@iadev09 cool!

Pushed some more improvements. Most of the manual byte wrangling now goes through a ClientHelloFields struct so that there is only one new ClientHello parser. It would be great to integrate with the existing message parsing code but I don't quite know how to structure that or if it would make sense. Also tried to make the handshake -> ech state machine API a bit more obvious.

iadev09:
I think fully folding this into the existing typed ClientHello parsing may be a bit awkward, because the ECH path seems to need a wire-level view too: exact bytes for AAD, extension ordering, and raw extension bodies.

A smaller step that might still help would be to move the zero-copy ClientHelloFields / raw extension iterator pieces into msgs/client_hello.rs as a crate-private wire-level helper, and let server/ech.rs use that instead of keeping its own local parser.

jellevandenhooff:
Moved the message parsing code into msgs/client_hello.rs. It is a nice improvement, thanks for the suggestion. Also got rid of the new-types for byte slices that now seemed unnecessary.

jellevandenhooff:
Okay, final change for now: moved the confirmation signal computation to ech.rs. Now the integration in the handshake is really quite slim, so squashed the three commits (initial integration; retry support; rewind support) into one.

icebluey:
# RFC 9849 Compliance Gaps in this PR

1. **Certificate validation on ECH rejection still uses the original inner name, not `public_name`.**

   RFC 9849 requires the client, on the rejection path, to verify that the server certificate is valid for `ECHConfig.contents.public_name` (`rfc9849.txt:991-992`).

   In rustls, certificate verification still uses `self.hs.session_key.server_name` in `src/client/tls13.rs:1137`, and that name comes from the original connection target in `src/client/hs.rs:420`. It is not switched to the outer `public_name`.

   The outer SNI is correctly set to `public_name` in `src/client/hs.rs:614`, but the verification identity is not. This is the most important hard RFC mismatch.

2. **The recommended padding scheme is implemented incorrectly, and `maximum_name_length = 0` is treated as `255`.**

   RFC 9849 requires clients to first add padding derived from `maximum_name_length`, then add the final 32-byte alignment padding (`rfc9849.txt:818-834`).

   In `src/client/ech.rs:702`:

   - `maximum_name_length == 0` is rewritten to `255`, which does not match the RFC semantics for `M = 0`.
   - The code computes the first-stage padding (`max(0, M - D)` or `M + 9`) but does not actually append that padding to the encoded inner ClientHello. It only uses that value in the later alignment calculation.

   As a result, the RFC-recommended padding scheme is not implemented as specified.

3. **A shared-mode server accepts a network `type=inner` ClientHello, but RFC 9849 requires `illegal_parameter`.**

   RFC 9849 is explicit:

   - invalid `ECHClientHello.type` must abort,
   - a split-mode frontend receiving `inner` must abort,
   - a shared-mode server directly receiving `inner` must also abort

   (`rfc9849.txt:1156-1167`).

   However, the server currently treats `EncryptedClientHello::Inner` as `AcceptedInnerDirect` in `src/server/ech.rs:386`. The normal server handshake path can reach that branch from network input via `src/server/hs.rs:736`.

   That behavior does not match the RFC role constraints for shared mode.

4. **The alert level for an invalid `ECHClientHello.type` is wrong.**

   RFC 9849 says: if `ECHClientHello.type` is not a valid `ECHClientHelloType`, the server must abort with `illegal_parameter` (`rfc9849.txt:1156-1157`).

   In rustls, the parser returns `InvalidContentType` for an unknown ECH type in `src/msgs/client_hello.rs:427`. That error is then mapped to `decode_error` in `src/error/mod.rs:957`.

   This is a concrete alert-semantics mismatch.

5. **`public_name` validation does not fully implement the RFC 6.1.7 filtering rules.**

   RFC 9849 says clients should ignore configurations whose `public_name` is not valid preferred-name syntax, or whose final label could be interpreted as an IPv4 literal, including all-numeric final labels and `0x...` forms (`rfc9849.txt:1016-1031`).

   The current parsing path performs `DnsName::try_from(...)` in `src/msgs/server_hello.rs:343-350`, but `EchConfig::new_for_configs` in `src/client/ech.rs:156-201` does not implement the additional RFC 6.1.7 filtering rules before selecting a configuration.

   The repository's own BOGO expectations also acknowledge at least part of this gap: `bogo/config.json.in:126` says `we allow underscore names`.

jellevandenhooff:
So that's what it's like to be on the receiving end of AI content! @icebluey, please double check the AI output before you comment; this PR implements the server side so the client side feedback is not relevant. The shared-mode server is called out in the description of the PR, though perhaps it should be configurable.

I've marked the PR as a draft. I am not quite happy yet with the ECH state machine and some of the code around that, and I also want to think a bit more about what API exactly makes sense for configuration, status, and the proxy. Feedback of course still very welcome.

iadev09:
I’m looking at this primarily as a real-world user trying to evaluate whether server-side ECH could become production-usable. I had been following the related issues and PRs for a while, and I was honestly surprised that such an important gap had remained open for so long. I’ve also noticed that @djc seems to be paying close attention to this area, which is encouraging. At the same time, rustls is too critical a security component for a server-side ECH design and implementation to be rushed.

So it was genuinely exciting to see this work come alive again at almost exactly the moment when I had just finished preparing `rustls 0.24` bindings, was following [#1980](https://github.com/rustls/rustls/issues/1980), and was looking at earlier work like [#2104](https://github.com/rustls/rustls/pull/2104) while thinking about what it would take to implement in our own fork.

From the outside, it is easy to read the current work as three fairly distinct layers, even if they continue to evolve together in one PR.

The first is the minimum server-side ECH core:
- `server: add ECH server module`
- `tls13: extract server_ech_confirmation_secret as a free function`
- `server: integrate ECH into TLS 1.3 handshake`
- `server: add ECH HelloRetryRequest support`

The second is the policy/API side:
- `server: add ECH rewind on cert resolution failure`
- and the remaining configuration/status decisions

The third is the split-mode/proxy side:
- `server: add ECH split-mode proxy`
- `client: skip fake CCS for QUIC retries`

Your “not quite happy yet” note actually feels very informative. To me, the discomfort seems less about the feature itself and more about the goal becoming less clear as several concerns are being shaped at once.

That makes me think the answer is not just in syntax, but in a more Rust-native design shape: tighter scope, clearer boundaries, explicit state, and types carrying more of the invariants.

I’d especially be interested in @djc’s view on the right path here from the rustls side. Even a bit of maintainer guidance on whether this should settle as one unit or as a few distinct layers would already help a lot.

ctz:
My view on this is that implementing all possible options (in this case, both shared and split mode) is usually not a good idea and a symptom of:

- some sort of completionist objective, which generally doesn't lead to good software. I don't want to be maintaining unused features, as it's a waste of time.
- failure to understand the problem-space or make an informed decision, which leads to a scatter-gun approach to problem solving (very common in RFCs, which standardise a ton of options in an unprincipled way)
- some sort of advanced usage, which is something that could be explained up-front

jellevandenhooff:
I appreciate the concern of not wanting to do to much. Like I wrote above, my (advanced) use case is to have a split mode proxy fronting backend TLS servers. The way the RFC is written makes this less daunting than you might think since it describes the server side protocol as a combination of front-end ECH decryption and back-end ECH acknowledgment: If you implement the simple use case of single-server ECH you should get split-server ECH almost for free. In the latest version of the PR I have now structured the ECH module to have a front-end and back-end split and I think it looks quite nice.

All parts do not have to land together, but I think trying to use the code in both use-cases is helpful in making sure the API is forward-compatible. I'll make sure to split the PR before asking for an in-depth review.

As always, (early) feedback on the code is quite welcome.

jellevandenhooff:
I've dropped split mode and the proxy code. I split out some of the infrastructure and prep changes into their own commits (the raw hello reader, the bogo tests).

At this point the core implementation in `ech.rs` I think is pretty much good. It is structured top-down; the byte wrangling is gone and instead uses the raw hello reader; the state machine is split into a front-end and back-end and looks clean; the public API is simpler and does not expose a state enum anymore and instead just ECH accepted or not along with the key ID and outer SNI for monitoring. The integration with the rest of the code is a slim 100-ish line commit.

Bogo tests still pass except for the inner mode support.

The old branch with proxy support is at https://github.com/jellevandenhooff/rustls/pull/2 and would be a natural follow-up.

iadev09:
Looking back at earlier discussions (https://github.com/rustls/rustls/issues/1980)  #2104), this was already noted as something that likely needs maintainer involvement. At the time it seemed mostly limited by bandwidth, but it’s been ~0.02 centuries since, which is starting to feel worth revisiting.

More broadly, this still feels like a real gap in the Rust ecosystem — there is still no widely usable, first-class server-side ECH support today. 🦀

jellevandenhooff:
I’ve read that PR and the comments, and have tried to incorporate that feedback as well as the feedback here into the changes. Compared to the earlier PR the changes here require less invasive changes to the rest of the code and have more documentation explaining what and why. Any future feedback I will happily incorporate.

jellevandenhooff:
I split the ECH implementation in several commits, which I think makes reviewing easier, and I added a high-level description of the structure of the changes in the PR message at the top, and added commit messages.

The PR should now be ready for review. Thanks.

frochet:
Hey,

I am chiming in this PR to discuss a potential issue; this is more a conceptual review of the proposal rather than a review of the code, with hopefully good suggestions to solve the stated problem.

So, if I understand correctly what you aim to do: you want a reverse proxy that does not terminate the TLS or QUIC connection, and forward the decrypted ECH towards the real server behind the frontend server.

The remark from @djc is meaningful:

> Do you have a (serious) use case for this? What are you trying to achieve?

To what you answer:

> My specific use case is that I would like to offer people personal (sub-)domains that I reverse proxy to their personal servers (think self-hosting), without me decrypting their TLS, and without anyone inspecting SNI being able to see who is trying to access anyone else's page.µ

which dwelves on your original comment:

> (And the reason I am excited about split mode is because it splits responsibilities and secrecy nicely: the backends do not get the ECH keys, so they cannot decode other requests going into the proxy; meanwhile, the proxy does not decrypt the actual traffic.)

Another question of importance link to @djc's question would be: what is your threat model? If people getting (sub-)domains trust your reverse proxy, then terminating the TLS or QUIC connection at the reverse proxy, and having long-lived connections towards people's (sub-) domains does the job fine, and this PR is not needed in the sense that it does not bring any new use case.

However, if people getting (sub-)domains do not trust your reverse proxy, which is kinda what I suppose you mean by "the proxy does not decrypt the actual traffic", then the use case you have cannot be secure as of now. That is, your people would not have any guarantee that your reverse proxy is not MITM'ing the connections. One way to do that would be for your proxy to request certificates for your people's (sub-) domains and intercepting the challenge towards these domains. You'd get valid cryptographic material that your reverse proxy *could* use to terminate the connection, and opening another channel towards the people's (sub-)domain. Neither the backend servers, nor the clients would see a problem with that.

So, in my view, the current PR either does not admit a use case or could lead people building misleading, insecure reverse proxy with incorrect guarantees.

There might be a way to mitigate the insecurity problem, but it seems like a long road. We would need a new method to acquire a certificate that could only be challenged to the backend server and its owner. One idea I can think of would be DNS extensions: tell CAs that the domain can only be validated through a DNS record (or higher tier validation, like EV certs). That works assuming you are not as well in control of the domain(s) behind your reverse proxy (so sub-domain case is likely off the table). It is a long road because Browsers need to support this extension and refuse any certificate behind such a reverse proxy that was not acquired through a DNS record challenge. This would need standardization as well, and have people on board with this idea. I am afraid that without this little dance you would end up to nothing useful or secure.

jellevandenhooff:
@frochet you’re wrong; the ECH RFC split mode is explicitly designed to have a proxy do the ECH handshake while the backend servers do the TLS termination. Backend servers cannot decrypt other ECH handshakes because they do not have the ECH keys and the proxy cannot MITM the actual TLS connection because it does not have certificates for the actual domains.

That’s about an earlier version of the PR anyway. In response to feedback I removed split mode support from the API and instead this PR now implements ECH where a single server does both the ECH decryption and the TLS termination.

frochet:
> @frochet you’re wrong; the ECH RFC split mode is explicitly designed to have a proxy do the ECH handshake while the backend servers do the TLS termination.

I agree. My statements do not contradict this.

> Backend servers cannot decrypt other ECH handshakes because they do not have the ECH keys

I agree, but I did not discuss this. Only the the client-facing server can decrypt ECH.

> and the proxy cannot MITM the actual TLS connection because it does not have certificates for the actual domains.

I disagree. If the backend domain configures their domain to point towards the client-facing server (the reverse proxy), then this server, as an active attacker, can request new valid certificates to CAs for the backend domain and _use_ these certificates to terminate the connection. Ask yourself:  how could the split mode prevent that? If a domain owner configures the domain to point to another machine that it does not trust, then no possible privacy guarantee can be possibly formulated given how origin validation works on the Internet.

Coming back to my original question: what is your threat model? I think it is alright for a passive honest-but-curious reverse proxy; in that case the privacy stands, but arguably, this is not a serious threat model for the Internet.

Skimming through the ECH RFC, it appears completely unclear what are the security guarantees, quoting section 10.1:

>    Passive and active attackers can exist anywhere in the network,
>    including between the client and client-facing server, as well as
>    between the client-facing and backend servers when running ECH in
>    split mode.  However, for split mode in particular, ECH makes two
>    additional assumptions:
> 
>    1.  The channel between each client-facing and each backend server is
>        authenticated such that the backend server only accepts messages
>        from trusted client-facing servers.  The exact mechanism for
>        establishing this authenticated channel is out of scope for this
>        document. 

The text seems to exclude the reverse proxy as an attacker, and the point "1." says 'trusted'; which means that indeed that no privacy guarantee is formulated from the reverse proxy.

Further bellow, the RFC says this:

>   Given this threat model, the primary goals of ECH are as follows.
> 
>    1.  Security preservation.  Use of ECH does not weaken the security
>        properties of TLS without ECH.
> 

In split mode, this seems true only if we assume that the reverse proxy is trusted.

My understanding of this situation is as follows: security-wise, against active attackers, the split-mode brings not added value compared to the shared mode. I am unsure whether the RFC fumbled it or if there is more subtle nuance I am not seeing. Maybe I should move this discussion to the TLS working group mailing list.

> 
> That’s about an earlier version of the PR anyway. In response to feedback I removed split mode support from the API and instead this PR now implements ECH where a single server does both the ECH decryption and the TLS termination.

Oh, sorry, I missed this information. I would however kindly advice precautions with security claims from potential  future split mode usage. If anything is unclear, please let me know, I am happy to further elaborate.

jellevandenhooff:
Off list seems better, yes. It’s true that a malicious proxy could try to acquire certificates through a SNI challenge with  letsencrypt, but you can prevent that with CAA records and an account binding (see https://letsencrypt.org/docs/caa/). Additionally, such certificates would show up in the CT log so you would have evidence of the proxy misbehaving.

iadev09:
@jellevandenhooff

I know split mode has been removed from this PR (which makes sense), but I’m still curious about the earlier design:

Was split mode inherently limited to SNI/DNS boundaries, or could it support single-domain indirection while keeping the privay of the backend identity  — similar in spirit to a `tiny URL`-style entrypoint acting as a proxy?

jellevandenhooff:
The ECH proxy can see the SNI information (and all other information in the hello) but nothing else. What you can do, with a wildcard certificate, stick information in the subdomain (like the bsky domains), and the browser will reuse the connection if it resolves to the same IP.

yaroslavros:
Split mode would be very valuable. It is a very viable way to implement privacy-preserving proxies.
Hopefully it will be brought back.

jellevandenhooff:
@yaroslavros I agree split mode is valuable. Restoring it is a natural extension of this PR.

jellevandenhooff:
rebased

iadev09:
@jellevandenhooff

I think @ctz’s earlier comment still captures the main design risk here:

> implementing all possible options (in this case, both shared and split mode) is usually not a good idea

My outside read is that keeping this PR focused on shared-mode server ECH gives it the clearest merge path, while split/proxy support can be handled in a later PR.

jellevandenhooff:
I don’t quite know how to think about code contributions in the time of LLMs. I sympathize with maintainers getting a lot of contributions of unknown quality.

The path for now is I think someone wanting to use this in their own codebase doing so with a fork and/or a more concrete use case and perhaps sponsoring the work in rustls.

iadev09:
To me, this is mostly about holding both views at once: your concrete need, and the public API shape rustls can review and own confidently.

@djc has also pointed later parallel server-side ECH work back here, so this does not read to me like fork-only work.

rootful:
How is the progress on this so far? I'm working on a Rust HTTP(S) daemon and have been wanting to implement ECH, and have considered resorting to forking rustls for this until I stumbled upon this PR.

djc:
> How is the progress on this so far? I'm working on a Rust HTTP(S) daemon and have been wanting to implement ECH, and have considered resorting to forking rustls for this until I stumbled upon this PR.

I don't think we'll get to this before 0.24 is released (tentatively planned for ~3 months from now).

johalun:
Thank you all for working on this. I very much would like this feature as well.

PR #2995, opened 2026-03-21
msgs: simplify deframer implementation

djc:
A series of smallish changes that IMO help make this code a little more approachable, trying to avoid bouncing around too much between different representations and methods.


djc:
For the clippy failure, I can splice in a `struct Deframed { message, bounds }` type if this PR is otherwise acceptable.

PR #2996, opened 2026-03-21
Update pinned webpki alpha

djc:
Should fix the remaining security alerts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -346,7 +346,7 @@ dependencies = [
  "rustls 0.24.0-dev.0",
  "rustls-aws-lc-rs",
  "rustls-ring",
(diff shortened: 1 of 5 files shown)


djc:
I have to stop for a bit, but this needs to more work to untangle the rustls-post-quantum stuff. Maybe rustls-post-quantum finally needs to jump over to rustls 0.24?

(Feel free to take the baton.)

ctz (approved):
(for the first commit -- if the second looks good to you, please squash)

ctz:
After this is landed I think it would be good to also move the ml-dsa code from rustls-aws-lc-rs to rustls-post-quantum

djc:
✅ for the second commit, though I reversed their order.

PR #3000, opened 2026-03-26
fix: big-endian u16 in msgs enum test helper get16

cuiweixie:
## Summary

The `get16` helper in `rustls/src/msgs/enums.rs` test module incorrectly combined wire bytes using a right shift on the high byte, which decodes multi-byte extension types wrong (for example `0x3374` became `0x0074`).

## Changes

- Add a regression test using `ExtensionType::NextProtocolNegotiation` (`0x3374`), which distinguishes correct big-endian assembly from the buggy shift direction.
- Fix `get16` to use `(high << 8) | low`.

## Testing

```bash
cargo test -p rustls get16_decodes_big_endian_u16
```

Verify the regression: the first commit adds the test with the old `get16` and the test fails; the second commit fixes `get16` and the test passes.


ctz:
Thanks for the report, fixed in #3001

PR #3009, opened 2026-03-31
client: allow skipping selected ALPN validation

### LINKED ISSUE #3007: Support unoffered application protocol
**Checklist**
* [X] I've searched the issue tracker for similar requests

I'm unable to connect to specific servers with non-standard ALPN negotiation

Simply do not check the returned ALPN protocol was offered.

Finding a different crate or forking this one.

**Additional context**
I know this technically violates spec. I'm trying to connect to Apple's APNS. See [rustpush](https://github.com/[OpenBubbles/rustpush](https://github.com/OpenBubbles/rustpush)). For their apns-pack-v1 protocol, they return an ALPN that contains extra metadata. However, as it is not listed in the list, rustls bails out with SelectedUnofferedApplicationProtocol.

I don't really see any problems with simply not checking this on rustls's end. Any clients will have to check the negotiated ALPN manually to determine which protocol to speak anyways, and can validate or parse it when they check. Or at the bare minimum, provide a flag clients can pass to not validate returning ALPN.

TaeHagen:
Closes #3007

### DIFF
--- a/rustls-test/tests/api/api.rs
+++ b/rustls-test/tests/api/api.rs
@@ -151,12 +151,27 @@ fn connection_level_alpn_protocols() {
 }
 
 #[test]
-fn server_selects_unoffered_alpn() {
-    let client_config = Arc::new(make_client_config(
-        KeyType::Rsa2048,
-        &provider::DEFAULT_PROVIDER,
-    ));
-    let mut client = client_config
+fn server_selects_unoffered_alpn_checked() {
+    let result = unoffered_alpn_test(true);
+    assert_eq!(
+        result.err(),
+        Some(PeerMisbehaved::SelectedUnofferedApplicationProtocol.into())
+    );
+}
+
+#[test]
+fn server_selects_unoffered_alpn_unchecked() {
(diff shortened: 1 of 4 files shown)


rustls-test/tests/api/api.rs
```diff
@@ -178,10 +175,23 @@ fn server_selects_unoffered_alpn() {
...
+        );
+    } else {
+        assert_ne!(
+            client.process_new_packets().err(),
+            Some(PeerMisbehaved::SelectedUnofferedApplicationProtocol.into())
+        );
+    }
+}
+
+#[test]
```
djc: Nit: suggest making this two different tests, and keeping the `#[test]` functions above the helper. The helper could just yield the `Result` value so that the assertions can live directly in the test functions.

cameronaaron:
LGTM

rustls-test/tests/api/api.rs
TaeHagen: Fixed

djc (approved):

djc:
Thanks!

PR #3012, opened 2026-04-01
Customizable cipher suite selection

### LINKED ISSUE #2252: Support prioritizing ChaCha20 for clients without AES hardware acceleration

djc:
@brian-pane I ended up significantly reworking your PR, please have a look.

I dropped the ChaCha20-specific implementation for now, since I retained preference for the client order as the default (in which case we automatically pick ChaCha20 if the client has it first). I guess we could customize `PreferServerOrder` to prefer ChaCha20 if the client lists it first?

Fixes #2252; obsoletes #2987.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -43,7 +43,8 @@ use rustls::pki_types::{
 };
 use rustls::server::danger::{ClientIdentity, ClientVerifier, SignatureVerificationInput};
 use rustls::server::{
-    self, ClientHello, ServerConfig, ServerConnection, ServerSessionKey, WebPkiClientVerifier,
+    self, ClientHello, PreferClientOrder, PreferServerOrder, ServerConfig, ServerConnection,
+    ServerSessionKey, WebPkiClientVerifier,
 };
 use rustls::{Connection, DistinguishedName, HandshakeKind, RootCertStore, compress};
 use rustls_aws_lc_rs::hpke;
@@ -1647,7 +1648,11 @@ fn make_server_cfg(opts: &Options, key_log: &Arc<KeyLogMemo>) -> Arc<ServerConfi
     cfg.max_fragment_size = opts.max_fragment;
     cfg.send_tls13_tickets = 1;
     cfg.require_ems = opts.require_ems;
-    cfg.ignore_client_order = opts.server_preference;
+    cfg.cipher_suite_selector = match opts.server_preference {
+        true => &PreferServerOrder,
+        false => &PreferClientOrder,
+    };
+
     if opts.export_traffic_secrets {
         cfg.key_log = key_log.clone();
     }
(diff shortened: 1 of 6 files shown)


bogo/src/main.rs
```diff
@@ -1647,7 +1648,6 @@ fn make_server_cfg(opts: &Options, key_log: &Arc<KeyLogMemo>) -> Arc<ServerConfi
     cfg.max_fragment_size = opts.max_fragment;
     cfg.send_tls13_tickets = 1;
     cfg.require_ems = opts.require_ems;
-    cfg.ignore_client_order = opts.server_preference;
```
ctz, line 1650: belongs in earlier commit?

rustls/src/server/config.rs
```diff
@@ -689,3 +694,114 @@ impl ConfigBuilder<ServerConfig, WantsServerCert> {
...
+    fn select(
+        &self,
+        client: &mut dyn Iterator<Item = &'static T>,
+        server: &[&'static T],
+    ) -> Option<&'static T>;
+}
+
+/// A filter that chooses the cipher suite to use for a TLS session.
+pub trait CipherSuiteSelector: Debug + Send + Sync {
+    /// Choose a cipher suite, given the server's and client's order lists of supported options.
```
ctz: I think this doc could be expanded, eg:

- what does it mean to return `None`.
- that `client` is reduced to supported items but the order is from the client.
- that `server` is from the provider (could be renamed to `provider` maybe?)

ctz:
```suggestion
    /// Choose a cipher suite, given the server's and client's ordered lists of supported options.
```

djc: I think `server` better helps establish the mental model of negotiation between client and server.

Otherwise, will expand docs.

bogo/src/main.rs
djc: Have squashed all commits to avoid intermediate broken bogo tests (and temporarily commented out test).

(Hmm, I guess the broken bogo tests may have just been the PQC ones that are only locally broken?)

brian-pane:
> I guess we could customize PreferServerOrder to prefer ChaCha20 if the client lists it first?

Yes, that would fix https://github.com/rustls/rustls/issues/2252

djc:
> > I guess we could customize PreferServerOrder to prefer ChaCha20 if the client lists it first?
> 
> Yes, that would fix https://github.com/rustls/rustls/issues/2252

So that's not here but at least after this you can customize the behavior. Honestly not sure if this is important enough that we should carry an implementation of it.

PR #3014, opened 2026-04-02
Add server-side Encrypted Client Hello (ECH) support per RFC 9849

MavenRain:

PR #3015, opened 2026-04-03
Support id-RSASSA-PSS certificates (rsa_pss_pss_* signature schemes)

### LINKED ISSUE #1161: `InvalidCertificateSignatureType` encountered in TLS 1.3 handshake
I encountered a strange `InvalidCertificateSignatureType` when attempting to establish a connection between an instance of `rustls` running as a server and another instance running as a client. I wrote a test which reproduces the issue in a fork of this repository [here](https://github.com/mdcarr941/rustls/blob/main/rustls/tests/repro.rs).

Tracing through with the debugger, I found that the signature algorithm negotiated by the client and server was RSASSA-PSS with SHA-512, even though the Subject Public Key Info in the server certificate specifies RSASSA-PSS with SHA-256. Working backwards from `emit_certificate_verify_tls13` in `src/server/tls13.rs`, it seems that the `schemes` argument is determined using the `signature_algorithms` extension in the ClientHello and the algorithms supported by the server's current cipher suite and protocol version. I can't find anywhere that the Subject Public Key Info in the server's certificate is taken into account.

I may be mistaken, but if this is the case then this implementation is not in compliance with the spec. From section 4.1.3 of RFC 8446: "In addition, the signature algorithm MUST be compatible with the key in the sender's end-entity certificate."

It seems like the minimal way to address this is to filter the `schemes` argument using the algorithm in Subject Public Key Info, if the server is configured to use certificate authentication. This could be done in `handle_client_hello` just before calling `emit_certificate_verify`. I'd be down to write the PR but I'm not familiar with webpki, so if you could give me some pointers on parsing the server cert I'd appreciate it.

MavenRain:
Add SignatureScheme variants for rsa_pss_pss_sha256/384/512 (0x0809-0x080b) and wire them through both the aws-lc-rs and ring crypto providers for verification and signing. Certificates whose SPKI uses the id-RSASSA-PSS OID (1.2.840.113549.1.1.10) are now supported alongside the existing rsaEncryption OID variants.

Resolves #1161.


djc:
You have not responded yet to my questions on your [previous PR](https://github.com/rustls/rustls/pull/3014). Please stop submitting large PRs unless you can explain your use cases and what human effort has gone into this PR other than prompting an LLM. If you continue submitting PRs without otherwise personally interacting with the maintainers, we will block you from further participation.

MavenRain:
> You have not responded yet to my questions on your [previous PR](https://github.com/rustls/rustls/pull/3014). Please stop submitting large PRs unless you can explain your use cases and what human effort has gone into this PR other than prompting an LLM. If you continue submitting PRs without otherwise personally interacting with the maintainers, we will block you from further participation.

My apologies again . . . I am getting to these comments just now after seeing email notifications from earlier today.

I did respond on the other PR and closed it.  I will close my other PRs in this repo.  I very much apologize for any inconvenience they have caused you.

PR #3026, opened 2026-04-10
Drop provider-example crate

djc:
Now that there are a number of existing outside provider crates, we no longer need this example.


cpu (commented):
I think there's still a reference in `README`, see heading "Custom provider": https://github.com/rustls/rustls?tab=readme-ov-file#custom-provider

Also in CI, for the `daily-tests.yml` and `build.yml`:

https://github.com/rustls/rustls/blob/f02ec1768cc627ecfec3e0197467b1721468f5f1/.github/workflows/build.yml#L102-L109

https://github.com/rustls/rustls/blob/f02ec1768cc627ecfec3e0197467b1721468f5f1/.github/workflows/daily-tests.yml#L124-L125

cpu:
Also some rustdoc in `rustls/src/crypto/mod.rs`: https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#example-code

cpu (approved):

ctz (approved):

PR #3034, opened 2026-04-21
Default require_ems based on CryptoProvider FIPS status

janrueth:
Previously, require_ems was defaulted solely based on cfg!(feature = "fips"), which is a compile-time check tied to the fips cargo feature. The fips feature unconditionally pulls in aws-lc-rs as the cryptographic provider, making it impossible for third-party FIPS-compliant CryptoProvider implementations (e.g., those backed by BoringSSL) to get correct FIPS policy defaults without also pulling in aws-lc-rs — which may conflict with their own crypto backend.

This change also considers the runtime CryptoProvider::fips() status when defaulting require_ems in both ClientConfig and ServerConfig builders. If the configured provider reports itself as FIPS-compliant, require_ems is now automatically set to true, ensuring that ClientConfig::fips() and ServerConfig::fips() return the correct result without requiring the fips cargo feature.

This is backward-compatible: existing users of the fips feature see no behavior change, while third-party providers now work correctly out of the box.

### DIFF
--- a/rustls/src/client/builder.rs
+++ b/rustls/src/client/builder.rs
@@ -162,6 +162,9 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
         self,
         client_auth_cert_resolver: Arc<dyn ResolvesClientCert>,
     ) -> ClientConfig {
+        #[cfg(feature = "tls12")]
+        let require_ems = self.provider.fips();
+
         ClientConfig {
             provider: self.provider,
             alpn_protocols: Vec::new(),
@@ -176,7 +179,7 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
             enable_secret_extraction: false,
             enable_early_data: false,
             #[cfg(feature = "tls12")]
-            require_ems: cfg!(feature = "fips"),
+            require_ems,
             time_provider: self.time_provider,
             cert_compressors: compress::default_cert_compressors().to_vec(),
             cert_compression_cache: Arc::new(compress::CompressionCache::default()),
--- a/rustls/src/manual/fips.rs
+++ b/rustls/src/manual/fips.rs
@@ -1,7 +1,8 @@
 /*! # Using rustls with FIPS-approved cryptography
 
-To use FIPS-approved cryptography with rustls, you should take
-these actions:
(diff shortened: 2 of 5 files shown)


rustls/src/client/builder.rs
```diff
@@ -162,6 +162,9 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
         self,
         client_auth_cert_resolver: Arc<dyn ResolvesClientCert>,
     ) -> ClientConfig {
+        #[cfg(feature = "tls12")]
+        let require_ems = cfg!(feature = "fips") || self.provider.fips();
```
djc: Don't see much point in outlining this from its previous location?

Also, can we just use `self.provider.fips()`? Seems sufficient for this... (If so, need to adjust the documentation accordingly.)

rustls/src/manual/fips.rs
```diff
@@ -40,6 +40,17 @@ assert!(client_config.fips());
 But maybe your application has an error handling
 or health-check strategy better than panicking.
 
+# Using a third-party FIPS `CryptoProvider`
```
djc: Not sure it makes sense to add a section on this -- probably makes more sense to review the current documentation in this module to review whether it still makes sense and/or can be clarified?

(I also don't think it's useful to call out "backed by BoringSSL" here.)

djc (commented):
I think this broadly makes sense. Fixed some pre-existing CI issues here:

- https://github.com/rustls/rustls/pull/3035

rustls/src/client/builder.rs
janrueth: provider moves into the type so this needs to happen outside. Happy to just use the provider.fips()

rustls/src/manual/fips.rs
janrueth: Will do!

janrueth: Removed this section and just clarified at the top that it needs a FIPS-approved provider and that rustls ships with one and kept the rest as is.

djc (approved):
Thanks!

ctz (approved):

ctz:
Published in 0.23.40 -- thanks!

PR #3036, opened 2026-04-22
Fix nightly clippy warning, take updates

ctz:


openssl-tests/src/raw_key_openssl_interop.rs
```diff
@@ -145,6 +145,7 @@ mod client {
 }
 
 mod server {
+    #![allow(clippy::std_instead_of_core)] // awaits core::io::ErrorKind in stable
```
djc: `expect()` instead of `allow`?

djc (approved):

openssl-tests/src/raw_key_openssl_interop.rs
ctz: Unfortunately `core::io::ErrorKind` is only present in nightly, so `expect` would fail here on stable. For the same reason, we can't just move the import to `core`. Bit painful!

djc: Fair enough. If it is stable-on-nightly, maybe mention "MSRV 1.97" or something in the comment?

ctz: Have added a mention of 1.97 here, no MSRV as this is test code.

PR #3042, opened 2026-04-26
Correct ECH inner hello padding

### LINKED ISSUE #3043: ECH inner SNI length leaked via missing padding append
(originally reported as GHSA-gjg2-6f7h-x33r by @walterhpearce. the maintainers decided to deal with this issue outside the security response, as the security impact was judged as marginal given the 32-byte padding that is in place, and an expectation that -- depending on other inner hello contents -- a typical server's anonymity set would be not split by this padding.)

### Summary
(Summary entirely human written) 

This finding was found as part of Project Glasswing (Utilizing the AI Model Mythos for vulnerability detection). Rust Foundation staff have manually validated the report. 

[RFC 9849 §6.1.3](https://www.rfc-editor.org/rfc/rfc9849.html#name-recommended-padding-scheme) states that padding should be done in reference to `server_name` extension padding *AND* rounded whole message padding. However, at [rustls/src/client/ech.rs:726](https://github.com/rustls/rustls/blob/main/rustls/src/client/ech.rs#L726) we can see that:
1. The server_name padding is calculated
2. This padding is fed into the calculated rounded padding
3. The buffer is extended for the calculated rounded padding

However, RFC 9849  states that both the server_name extension padding *and* the final rounded caulcation should both be applied. You can see that in the reference code section, where the padding is calculated, fed into the rounding function, and then applied in totality.

Relevant RFC Section
```
In contrast, clients do not know the longest SNI value in the client-facing server's anonymity set without server input. Clients SHOULD use the ECHConfig's maximum_name_length field as follows, where M is the maximum_name_length value.

If the ClientHelloInner contained a "server_name" extension with a name of length D, add max(0, M - D) bytes of padding.

If the ClientHelloInner did not contain a "server_name" extension (e.g., if the client is connecting to an IP address), add M + 9 bytes of padding. This is the length of a "server_name" extension with an M-byte name.

Finally, the client SHOULD pad the entire message as follows:

Let L be the length of the EncodedClientHelloInner with all the padding computed so far.

Let N = 31 - ((L - 1) % 32) and add N bytes of padding.

This rounds the length of EncodedClientHelloInner up to a multiple of 32 bytes, reducing the set of possible lengths across all clients
```

(LLM Generated Text Follows)

## Details
encode_inner_hello() computes the RFC 9849 §6.1.3 SNI-compensation padding (max_name_len - len(server_name)) at lines 707-722, but the resulting padding_len is only fed into the rounding formula on line 726 where the variable is immediately shadowed; the SNI pad bytes are never appended to encoded_hello. Only the 0-31 byte rounding pad is appended (line 727). Because encoded_hello.len() + padding_len is constant across SNI values, the rounding pad N is constant — so the final plaintext length is encoded_hello.len() + N, which still varies one-for-one with the inner server_name length (SNI is encoded uncompressed in the inner hello, lines 650-657). The under-padded buffer is HPKE-sealed verbatim at line 463, so the encrypted_client_hello extension length on the wire reveals the inner SNI length.

## Impact
Passive observer infers protected inner server name length, weakening ECH privacy

## Reproduction steps
1. A passive on-path observer (ISP, corporate middlebox, nation-state tap) records a rustls client's ClientHelloOuter and measures the encrypted_client_hello extension payload length. Subtracting the fixed AEAD tag length and the constant baseline for the client's extension set yields the byte length of the inner server_name. When the ECH client-facing server fronts multiple origins with distinguishable name lengths (e.g. 'cdn.ex' vs 'internal-admin-console.corp.example'), the observer identifies which protected origin the client is contacting — exactly the traffic-analysis leak the RFC 9849 padding scheme is specified to prevent.

## Recommended fix
The EncodedClientHelloInner length after padding must be independent of the inner server_name length per RFC 9849 §6.1.3 — append the SNI-compensation padding bytes before computing and appending the 32-byte-multiple rounding pad.

ctz:


rustls/src/client/ech.rs
```diff
@@ -714,10 +714,11 @@ impl EchState {
             // "This is the length of a "server_name" extension with an L-byte name."
             _ => max_name_len + 9,
         };
+        encoded_hello.extend(vec![0; name_padding_len]);
```
djc: Here (and in the pre-existing case), I think there's probably a way to do this without allocating? Maybe with `iter::repeat_n()`?

rustls/src/client/ech.rs
```diff
@@ -904,4 +908,160 @@ mod tests {
...
+            None,
+            None,
+        )
+    }
+
+    fn dns_name_of_len(mut len: usize) -> DnsName<'static> {
+        let mut s = String::new();
+        let labels = len.div_ceil(63);
+        for _ in 0..labels {
+            let chars = cmp::min(len, 63);
```
djc: Nit: suggest writing this as `Ord::min()`, which doesn't need an import.

djc (approved):
LGTM, nice work.

djc (approved):

cpu (approved):

PR #3044, opened 2026-04-28
0.23.40: backport ECH padding fixes to 0.23

ctz:
Proposed release notes:

* ECH: correct implementation of RFC-recommended padding scheme for SNI names in the inner client hello.
* ECH: correct padding of inner client hello when SNI is disabled.
* Alter default of `ServerConfig::require_ems` based on provider's FIPS status. Prior to this change, the default followed the `fips` crate feature, which was less helpful for users of external FIPS-approved providers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2512,7 +2512,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.39"
+version = "0.23.40"
 dependencies = [
  "aws-lc-rs",
  "base64",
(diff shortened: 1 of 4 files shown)


cpu (approved):
> Prepare 0.23.40

Can you also propose some release notes for review? 🙇

djc (approved):

ctz:
> > Prepare 0.23.40
> 
> Can you also propose some release notes for review? 🙇

Added, PTAL

cpu:
> ECH: implement RFC-recommended padding of SNI names in the inner client hello.

Would it be more accurate to describe this as fixing the RFC-recommended padding? I think we did at least attempt to implement it initially right?

Otherwise LGTM.

djc:
LGTM.

PR #3045, opened 2026-04-29
Use GHA-provided public ARM runners

ctz:


djc:
Why not latest then, for consistency with the others?

ctz:
> Why not latest then, for consistency with the others?

Doesn't exist unfortunately - https://github.com/orgs/community/discussions/149392

djc (approved):

PR #3051, opened 2026-05-10
client: add Codec impls to Tls{12,13}ClientSessionValue (enables disk-persisted ClientSessionStore)

swamig:
## Motivation

`ServerSessionValue` is `Codec`-roundtrippable for use in server-side ticket caches. The symmetric client-side types `Tls13ClientSessionValue` and `Tls12ClientSessionValue` aren't, which means custom `ClientSessionStore` implementations cannot persist sessions to disk.

This is a real production constraint: the in-memory `ClientSessionMemoryCache` is the only `ClientSessionStore` that ships with rustls, and it's process-local. Long-running clients pay a fresh 1-RTT TLS handshake on every cold start, even when a recent ticket would otherwise have enabled 0-RTT. For QUIC media transports, IoT clients with intermittent connectivity, and mobile apps that get backgrounded — this is a meaningful UX regression vs. what TLS 1.3 0-RTT was designed to deliver.

## What this PR adds

- `Codec<'_>` impls on `Tls13ClientSessionValue`, `Tls12ClientSessionValue`, and the shared `ClientSessionCommon`.
- Mirror the existing `ServerSessionValue::encode/read` pattern in `msgs/persist.rs`.
- Cipher suites encoded as the underlying `CipherSuite` u16 and resolved back via `CryptoProvider::get_default()` on read (returns `Err(InvalidMessage::UnknownProtocolVersion)` if the suite isn't in the active provider — safe fallback to 1-RTT).
- Round-trip unit tests in `msgs/persist.rs`.

## One open design question for maintainers

`ClientSessionCommon` carries `Weak<dyn ServerCertVerifier>` and `Weak<dyn ResolvesClientCert>` references that the existing `compatible_config()` uses for pointer-equality identity checks. These cannot survive process restart — the upstream `Weak` would always be dangling.

Two options for handling this:

**A) Add a `from_disk: bool` flag (what's in this PR).** `compatible_config()` returns true unconditionally for disk-loaded values. Operationally, the application is expected to keep its verifier policy stable across runs. This is what makes disk persistence actually useful.

**B) Skip persistence of the verifier weak refs and require callers to re-set them post-deserialization.** Cleaner type-wise but pushes complexity to consumers and would require breaking `Codec::read` into a two-phase deserialize / re-attach API.

I implemented (A) here because it's the minimum change to make persistence usable and the operational constraint (keep verifier policy stable across runs) is reasonable for production deployments. Open to either; happy to switch if (B) is preferred.

## What this PR is NOT

- Not adding any new public API surface beyond `Codec` on these types (and the helper impl on `ClientSessionCommon`).
- Not changing the `ClientSessionStore` trait.
- Not adding external-PSK support (separate concern).
- Not changing default behavior — `ClientSessionMemoryCache` is unaffected.

## Security considerations

Persisted client session values contain sensitive material (resumption secrets, ticket bytes). Downstream consumers MUST:
- Encrypt at rest if the host filesystem is untrusted.
- Honour ticket lifetimes and discard expired entries.
- Treat 0-RTT data as replayable; gate non-idempotent operations behind handshake completion.

These already apply to in-memory storage. This PR makes the bytes-roundtrip mechanically possible.

## Tests

- Round-trip tests added under `#[cfg(test)]` in `msgs/persist.rs` (`tls13_client_session_value_roundtrip`, `tls12_client_session_value_roundtrip`).
- End-to-end validation in a downstream consumer: a sled-backed `ClientSessionStore` is opened on disk, the client connects (1-RTT, ticket lands), the `Client` is dropped, a new `Client` is constructed reading the same sled path, and the second connect attempts 0-RTT. Test passes in ~1.4s.

## Roadmap fit

External-PSK, `Tls13ClientSessionValue` serialization, and disk-persistent session caching aren't currently on the public roadmap. If maintainers don't want this in core, happy to:
- (a) gate it behind a `serializable-client-session` cargo feature
- (b) maintain it downstream

Hoping for (a) — many ecosystem consumers would benefit. We're currently maintaining a vendored fork with this change, and would much rather drop the fork once this lands.

## Backwards compatibility

`Codec` impls are additive. `ClientSessionMemoryCache` doesn't use `Codec` (it stores values directly in `Vec`), so this PR is invisible to existing users.

Filed against `rel-0.23` since that's where my downstream consumer lives. Happy to forward-port to `main` (0.24.0-dev) once the design question is resolved — the persist module's been refactored on main but the same primitives still exist.

### DIFF
--- a/rustls/src/msgs/persist.rs
+++ b/rustls/src/msgs/persist.rs
@@ -5,6 +5,7 @@ use pki_types::{DnsName, UnixTime};
 use zeroize::Zeroizing;
 
 use crate::client::ResolvesClientCert;
+use crate::crypto::CryptoProvider;
 use crate::enums::{CipherSuite, ProtocolVersion};
 use crate::error::InvalidMessage;
 use crate::msgs::base::{MaybeEmpty, PayloadU8, PayloadU16};
@@ -225,6 +226,15 @@ pub struct ClientSessionCommon {
     server_cert_chain: Arc<CertificateChain<'static>>,
     server_cert_verifier: Weak<dyn ServerCertVerifier>,
     client_creds: Weak<dyn ResolvesClientCert>,
+    /// Set when this value was reconstructed from a `Codec::read` (i.e. loaded from
+    /// a disk-backed `ClientSessionStore`). In that case we have no live `Arc` to
+    /// the original verifier / client-cert resolver, so the `Weak::ptr_eq` check in
+    /// `compatible_config` would always fail and resumption would silently downgrade
+    /// to a fresh handshake. When `true`, `compatible_config` skips the ptr-equality
+    /// checks; the caller is responsible for ensuring the configured verifier is
+    /// equivalent to the one that produced the ticket. See `crates/vendor/rustls`
+    /// in the workspace for the disk-resumption rationale.
+    pub(crate) from_disk: bool,
 }
 
 impl ClientSessionCommon {
@@ -245,6 +255,7 @@ impl ClientSessionCommon {
             server_cert_chain: Arc::new(server_cert_chain),
             server_cert_verifier: Arc::downgrade(server_cert_verifier),
             client_creds: Arc::downgrade(client_creds),
+            from_disk: false,
         }
     }
 
@@ -253,6 +264,11 @@ impl ClientSessionCommon {
         server_cert_verifier: &Arc<dyn ServerCertVerifier>,
         client_creds: &Arc<dyn ResolvesClientCert>,
     ) -> bool {
+        if self.from_disk {
+            // Disk-loaded sessions can't carry a live verifier/resolver `Arc`, so
+            // the ptr_eq check would always fail. Trust the caller's configuration.
+            return true;
+        }
         let same_verifier = Weak::ptr_eq(
             &Arc::downgrade(server_cert_verifier),
             &self.server_cert_verifier,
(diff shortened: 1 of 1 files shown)


djc:
IIRC we did some of this stuff already on `main`. Can you look for the relevant PRs and review the implementation as well as related discussion?

swamig:
Thanks for the pointer! I see now — `f7e70079` and `f3055531` from February already do this exact work, and the approach is significantly cleaner than what I'd put together. Specifically:

- Public `from_slice` / `encode` on the value types instead of exposing `pub(crate) Codec`
- Restructured `ClientSessionCommon` carrying `peer_identity: Arc<Identity>` instead of `Weak<dyn ServerCertVerifier>` — which eliminates the verifier-identity-bypass workaround my PR needed

Closing this as a duplicate. Two follow-up questions if you have a moment:

1. Is the restructured `ClientSessionCommon` slated for a 0.23.x backport, or only landing in 0.24? My downstream consumer is pinned to quinn 0.11.x → rustls 0.23 for now, and we're maintaining a vendored fork to use the feature today.
2. Any sense of when 0.24 might release? If it's near-term I'll just wait; if it's months out, would you accept a backport PR against `rel-0.23` that pulls just the serialization API surface (without the larger `ClientSessionCommon` restructure, which I assume is too invasive for a stable backport)?

Thanks for taking the time to point me at the prior art.

swamig:
Superseded by f7e70079 + f3055531 on main. See comment thread.

djc:
For now we have no plans to backport, but I think we'd be willing to entertain that if you can do the legwork. I don't remember what other differences from `main` are prerequisites.

The tentative plan is to release 0.24 towards the end of July.

PR #3062, opened 2026-05-22
chore: improve rustls maintenance path

oab24413gmai:
Summary:
- Add focused unit tests to ChunkVecBuffer in rustls/src/vecbuf.rs covering edge cases: appending an empty chunk (should be a no-op and not affect len/is_empty), reading into a zero-length destination slice, consume() with exactly len bytes (boundary), and pop() behavior on an empty buffer. Tests live in the same file's mod tests block, keeping the change a single small additive edit with no behavior change.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/rustls/src/vecbuf.rs
+++ b/rustls/src/vecbuf.rs
@@ -263,6 +263,42 @@ mod tests {
     }
 
     #[test]
+    fn append_empty_chunk_is_no_op() {
+        let mut cvb = ChunkVecBuffer::new(None);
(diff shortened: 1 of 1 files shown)


cpu:
I don't think the commit message for this change is particularly strong (what's the "maintenance path"?). 

What motivated you down the road of working on this particular test coverage? It doesn't seem to shift the needle on overall coverage metrics.

ctz:
Thank you, but no.

PR #3064, opened 2026-05-23
implement rfc 9149 ticket_request extension

### LINKED ISSUE #2205: Add support for RFC 9149 (TLS Ticket Requests)
[RFC 9149](https://www.rfc-editor.org/rfc/rfc9149.html) describes an extension to request specific number of TLS session tickets (NewSessionTicket messages) from the server.

Without this extension servers have to guess how many tickets to send. OpenSSL default unless `SSL_set_num_tickets` is called is two, nginx seems to send two tickets, postfix sends one if session is not resumed and zero if session is resumed. This may result in storing a lot of tickets on the client e.g. if the client only uses at most one ticket, or not being able to resume session if the number of tickets goes low.

With RFC 9149 extension client can request more tickets when it runs out of them or request less tickets if it already has many fresh tickets stored.

I don't know which existing servers support the extension, but even if it is only used between Rustls clients and Rustls servers it's an improvement.

hz2:
allow clients to request a specific number of tls 1.3 session tickets via the ticket_request extension. servers cap the requested count at their configured maximum and echo the expected count back in encrypted extensions.

Fixes https://github.com/rustls/rustls/issues/2205

### DIFF
--- a/rustls-test/tests/api/resume.rs
+++ b/rustls-test/tests/api/resume.rs
@@ -13,7 +13,7 @@ use rustls::crypto::{CertificateIdentity, Identity};
 use rustls::enums::ProtocolVersion;
 use rustls::error::{ApiMisuse, Error, PeerMisbehaved};
 use rustls::server::ServerSessionKey;
-use rustls::{ClientConfig, Connection, HandshakeKind, ServerConfig, ServerConnection};
+use rustls::{ClientConfig, Connection, HandshakeKind, ServerConfig, ServerConnection, VecInput};
 use rustls_test::{
     ClientConfigExt, ClientStorage, ClientStorageOp, ErrorFromPeer, KeyType, ServerConfigExt,
     do_handshake, do_handshake_until_error, make_client_config, make_client_config_with_auth,
@@ -30,23 +30,40 @@ fn client_only_attempts_resumption_with_compatible_security() {
 
     let server_config = make_server_config(kt, &provider);
     for version_provider in ALL_VERSIONS {
+        let mut client_input = VecInput::default();
+        let mut server_input = VecInput::default();
         let base_client_config = make_client_config(kt, &version_provider);
         let (mut client, mut server) =
             make_pair_for_configs(base_client_config.clone(), server_config.clone());
-        do_handshake(&mut client, &mut server);
+        do_handshake(
+            &mut client_input,
+            &mut client,
+            &mut server_input,
+            &mut server,
+        );
         assert_eq!(client.handshake_kind(), Some(HandshakeKind::Full));
 
         // base case
         let (mut client, mut server) =
             make_pair_for_configs(base_client_config.clone(), server_config.clone());
-        do_handshake(&mut client, &mut server);
+        do_handshake(
+            &mut client_input,
+            &mut client,
+            &mut server_input,
+            &mut server,
+        );
         assert_eq!(client.handshake_kind(), Some(HandshakeKind::Resumed));
 
         // allowed case, using `clone`
         let client_config = ClientConfig::clone(&base_client_config);
         let (mut client, mut server) =
             make_pair_for_configs(client_config.clone(), server_config.clone());
-        do_handshake(&mut client, &mut server);
+        do_handshake(
+            &mut client_input,
+            &mut client,
+            &mut server_input,
+            &mut server,
+        );
         assert_eq!(client.handshake_kind(), Some(HandshakeKind::Resumed));
 
         // disallowed case: unmatching `client_auth_cert_resolver`
@@ -61,7 +78,12 @@ fn client_only_attempts_resumption_with_compatible_security() {
 
         let (mut client, mut server) =
             make_pair_for_configs(client_config.clone(), server_config.clone());
-        do_handshake(&mut client, &mut server);
+        do_handshake(
+            &mut client_input,
+            &mut client,
+            &mut server_input,
+            &mut server,
+        );
         assert_eq!(client.handshake_kind(), Some(HandshakeKind::Full));
 
         // disallowed case: unmatching `verifier`
@@ -79,7 +101,12 @@ fn client_only_attempts_resumption_with_compatible_security() {
 
         let (mut client, mut server) =
             make_pair_for_configs(client_config.clone(), server_config.clone());
-        do_handshake(&mut client, &mut server);
+        do_handshake(
+            &mut client_input,
+            &mut client,
+            &mut server_input,
+            &mut server,
+        );
         assert_eq!(client.handshake_kind(), Some(HandshakeKind::Full));
     }
 }
@@ -93,6 +120,8 @@ fn resumption_combinations() {
             (ProtocolVersion::TLSv1_2, provider::DEFAULT_TLS12_PROVIDER),
             (ProtocolVersion::TLSv1_3, provider::DEFAULT_TLS13_PROVIDER),
         ] {
+            let mut client_input = VecInput::default();
+            let mut server_input = VecInput::default();
             let resumption_data = format!("resumption data {kt:?} {version:?}");
             let client_config = make_client_config(*kt, &version_provider);
             let (mut client, mut server) =
@@ -100,7 +129,12 @@ fn resumption_combinations() {
             server
                 .set_resumption_data(resumption_data.as_bytes())
                 .unwrap();
-            do_handshake(&mut client, &mut server);
+            do_handshake(
+                &mut client_input,
+                &mut client,
+                &mut server_input,
+                &mut server,
+            );
 
             let expected_kx = expected_kx_for_version(version);
 
@@ -123,7 +157,12 @@ fn resumption_combinations() {
 
             let (mut client, mut server) =
                 make_pair_for_configs(client_config.clone(), server_config.clone());
-            do_handshake(&mut client, &mut server);
+            do_handshake(
+                &mut client_input,
+                &mut client,
+                &mut server_input,
+                &mut server,
+            );
 
             assert_eq!(client.handshake_kind(), Some(HandshakeKind::Resumed));
             assert_eq!(server.handshake_kind(), Some(HandshakeKind::Resumed));
@@ -191,13 +230,21 @@ fn test_client_tls12_no_resume_after_server_downgrade() {
         ServerConfig::builder(provider::DEFAULT_TLS12_PROVIDER.into()).finish(KeyType::Ed25519);
     server_config_2.session_storage = Arc::new(rustls::server::NoServerSessionStorage {});
 
+    let mut client_input = VecInput::default();
+    let mut server_input = VecInput::default();
+
     dbg!("handshake 1");
     let mut client_1 = client_config
         .connect("localhost".try_into().unwrap())
         .build()
         .unwrap();
     let mut server_1 = ServerConnection::new(server_config_1).unwrap();
-    do_handshake(&mut client_1, &mut server_1);
+    do_handshake(
+        &mut client_input,
+        &mut client_1,
+        &mut server_input,
+        &mut server_1,
+    );
 
     assert_eq!(client_storage.ops().len(), 7);
     println!("hs1 storage ops: {:#?}", client_storage.ops());
@@ -220,7 +267,12 @@ fn test_client_tls12_no_resume_after_server_downgrade() {
         .build()
         .unwrap();
     let mut server_2 = ServerConnection::new(Arc::new(server_config_2)).unwrap();
-    do_handshake(&mut client_2, &mut server_2);
+    do_handshake(
+        &mut client_input,
+        &mut client_2,
+        &mut server_input,
+        &mut server_2,
+    );
     println!("hs2 storage ops: {:#?}", client_storage.ops());
     assert_eq!(client_storage.ops().len(), 9);
 
@@ -247,9 +299,18 @@ fn test_tls13_client_resumption_does_not_reuse_tickets() {
     server_config.send_tls13_tickets = 5;
     let server_config = Arc::new(server_config);
 
+    let mut client_input = VecInput::default();
+    let mut server_input = VecInput::default();
+
     // first handshake: client obtains 5 tickets from server.
     let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
-    do_handshake_until_error(&mut client, &mut server).unwrap();
+    do_handshake_until_error(
+        &mut client_input,
+        &mut client,
+        &mut server_input,
+        &mut server,
+    )
+    .unwrap();
 
     let ops = shared_storage.ops_and_reset();
     println!("storage {ops:#?}");
@@ -270,8 +331,10 @@ fn test_tls13_client_resumption_does_not_reuse_tickets() {
     // connectivity uncertainty.
     for _ in 0..5 {
         let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
-        transfer(&mut client, &mut server);
-        server.process_new_packets().unwrap();
+        transfer(&mut client, &mut server_input);
+        server
+            .process_new_packets(&mut server_input)
+            .unwrap();
 
         let ops = shared_storage.ops_and_reset();
         assert!(matches!(ops[0], ClientStorageOp::TakeTls13Ticket(_, true)));
@@ -279,8 +342,10 @@ fn test_tls13_client_resumption_does_not_reuse_tickets() {
 
     // 6th subsequent handshake: cannot be resumed; we ran out of tickets
     let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
-    transfer(&mut client, &mut server);
-    server.process_new_packets().unwrap();
+    transfer(&mut client, &mut server_input);
+    server
+        .process_new_packets(&mut server_input)
+        .unwrap();
 
     let ops = shared_storage.ops_and_reset();
     println!("last {ops:?}");
@@ -299,9 +364,17 @@ fn tls13_stateful_resumption() {
     server_config.session_storage = storage.clone();
     let server_config = Arc::new(server_config);
 
+    let mut client_input = VecInput::default();
+    let mut server_input = VecInput::default();
+
     // full handshake
     let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
-    let (full_c2s, full_s2c) = do_handshake(&mut client, &mut server);
+    let (full_c2s, full_s2c) = do_handshake(
+        &mut client_input,
+        &mut client,
+        &mut server_input,
+        &mut server,
+    );
     assert_eq!(client.tls13_tickets_received(), 2);
     assert_eq!(storage.puts(), 2);
     assert_eq!(storage.gets(), 0);
@@ -320,7 +393,12 @@ fn tls13_stateful_resumption() {
 
     // resumed
     let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
-    let (resume_c2s, resume_s2c) = do_handshake(&mut client, &mut server);
+    let (resume_c2s, resume_s2c) = do_handshake(
+        &mut client_input,
+        &mut client,
+        &mut server_input,
+        &mut server,
+    );
     assert!(resume_c2s > full_c2s);
     assert!(resume_s2c < full_s2c);
     assert_eq!(storage.puts(), 4);
@@ -340,7 +418,12 @@ fn tls13_stateful_resumption() {
 
     // resumed again
     let (mut client, mut server) = make_pair_for_arc_configs(&client_config, &server_config);
-    let (resume2_c2s, resume2_s2c) = do_handshake(&mut client, &mut server);
+    let (resume2_c2s, resume2_s2c) = do_handshake(
+        &mut client_input,
+        &mut client,
+        &mut server_input,
+        &mut server,
+    );
     assert_eq!(resume_s2c, resume2_s2c);
     assert_eq!(resume_c2s, resume2_c2s);
     assert_eq!(storage.puts(), 6);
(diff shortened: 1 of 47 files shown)


rustls/src/client/config.rs
```diff
@@ -138,6 +138,21 @@ pub struct ClientConfig {
...
+    /// the [RFC 9149] `ticket_request` extension.
+    ///
+    /// This is a pair of `(new_session_count, resumption_count)`:
+    /// - `new_session_count`: tickets desired when the server negotiates a new connection.
+    /// - `resumption_count`: tickets desired when the server resumes using a presented ticket.
+    ///
+    /// Set to `None` to disable sending the extension (the default).
+    ///
+    /// A reasonable starting point is `Some((2, 1))` — request 2 tickets on a new
+    /// connection and 1 on resumption.
```
cpu: This seems counter to the advice in [RFC 9149](https://datatracker.ietf.org/doc/html/rfc9149#section-3) where it says:

> A client starting a new connection SHOULD set new_session_count to the desired number of session tickets and resumption_count to 0. Once a client's ticket cache is primed, a resumption_count of 1 is a good choice that allows the server to replace each ticket with a new ticket without over-provisioning the client with excess tickets.

rustls/src/client/config.rs
```diff
@@ -138,6 +138,21 @@ pub struct ClientConfig {
...
+    ///
+    /// [RFC 9149]: https://datatracker.ietf.org/doc/html/rfc9149
+    pub send_ticket_request: Option<(u8, u8)>,
```
cpu: I don't think a tuple is very ergonomic for the user-facing API here. I'd personally prefer to see a struct with named fields to provide clear disambiguation between the two u8 values.

rustls-test/tests/api/resume.rs
```diff
@@ -765,3 +765,92 @@ impl rustls::server::StoresServerSessions for ServerStorage {
         true
     }
 }
+
+#[test]
```
cpu: I think the test coverage here is missing a few angles that might be interesting:

* HRR isn't considered
* There's no distinction between new_session_count and resumption_count AFAICT
* I don't think there's a test that `None` in the config results in the extension being omitted (maybe harder to test?)

rustls/src/server/tls13.rs
```diff
@@ -718,6 +729,12 @@ mod client_hello {
             config,
         )?;
 
+        if hello.ticket_request.is_some() {
+            extensions.ticket_request = Some(ServerTicketRequestHint {
+                expected_count: send_tickets as u8,
```
cpu: I think this is safe based on a pre-existing clamp of `st.send_tickets` <= 255, but I think it might be safer long-term to use something like `send_tickets.min(u8::MAX as usize) as u8` here?

rustls/src/client/config.rs
cpu: How does this interact when a user sets `ClientConfig.Resumption` to `Resumption::disabled()` but also species a `Some(_)` `send_ticket_request`? It seems like the type system should prevent this kind of confusing misconfiguration.

In general it also feels a bit odd that the other aspects of resumption are configured separately from this.

cpu (commented):
Can you speak more to your motivation for adding this? AFAICT this RFC hasn't seen broad implementation (e.g. I don't think OpenSSL, BoringSSL, or Go have implemented it). The BoringSSL omission is particularly relevant because it would be swell to have BoGo test coverage for a protocol feature like this.

On that topic I think your branch is missing some requirements from RFC 9149 w.r.t client behaviour, for example enforcing the MUST NOT [from §3](https://www.rfc-editor.org/rfc/rfc9149.html#section-3):

> Servers MUST NOT send the "ticket_request" extension in any handshake message, including ServerHello or HelloRetryRequest messages. A client MUST abort the connection with an "illegal_parameter" alert if the "ticket_request" extension is present in any server handshake message.

Without committing to a desire to land this, here's some other additional review feedback I had from taking a look at the branch.

hz2:
Thanks for the thorough review!

> Can you speak more to your motivation for adding this? AFAICT this RFC hasn't seen broad implementation (e.g. I don't think OpenSSL, BoringSSL, or Go have implemented it). The BoringSSL omission is particularly relevant because it would be swell to have BoGo test coverage for a protocol feature like this.

Fair enough, I noticed that OpenSSL acknowledges the extension but note they don't explicitly enforce/interpret it. That being said, this would be only for rustls-to-rustls apps for now. The benefit is that extension becomes safe to send since TLS 1.3 requires servers to ignore unknown clienthello extensions. For end-to-end apps, e.g. internal meshes, this is immediately useful.

However, totally understand that there isn't broader adoption and not wanting to be stuck with the maintenance burden. Though, thought this was relatively small server-side logic?

rustls/src/client/config.rs
hz2: Ya fair point, will clean this up

cpu:
I can't speak for the other maintainers but I think for the benefit of rustls<-> rustls users it seems reasonable to pursue further. As you mention the implementation isn't particularly invasive 👍

rustls/src/client/config.rs
hz2: went with suppressing the extension when resumption is disabled (detected via `tls12_resumption == Disabled`)

this is the only constructor that sets that state, so it catches the `Resumption::disabled()` case without needing to change the type hierarchy. open to moving the config onto `Resumption` itself if you think that's cleaner?

rustls/src/server/tls13.rs
hz2: makes sense, added

rustls-test/tests/api/resume.rs
hz2: > HRR isn't considered

Good call out, added a test for this that verifies extension survives the hello retry and server still honors the count

along with other comment + `Resumption::disabled()` suppresses the extension

lmk if there's more coverage i'm missing

rustls/src/client/config.rs
hz2: thanks for keeping me honest on this!

djc:
> For end-to-end apps, e.g. internal meshes, this is immediately useful.

@cpu asked for a motivation, but I'm not seeing an application-level use case or constraint that this helps with here.

hz2:
> @cpu asked for a motivation, but I'm not seeing an application-level use case or constraint that this helps with here.

Are you asking what _I personally_ intend to use it for? Or are the [use cases in the RFC](https://www.rfc-editor.org/rfc/rfc9149.html#name-use-cases) not sufficient for why it would be helpful at the application level?

Rather than doing the default 2 tickets from the first connection after a server restart, a client with a pool of $n$ connections can request $n$ tickets, allowing the remaining $n-1$ connections resume instead of doing full handshakes

rustls/src/server/tls13.rs
```diff
@@ -235,6 +235,15 @@ mod client_hello {
                 st.send_tickets = st.config.send_tls13_tickets;
             }
 
+            if let Some(ticket_req) = &input.client_hello.ticket_request {
+                let requested = if resuming.is_some() {
+                    ticket_req.resumption_count
+                } else {
+                    ticket_req.new_session_count
+                };
+                st.send_tickets = st.send_tickets.min(requested as usize);
```
ctz: I think this is probably hinting that `ServerConfig::send_tls13_tickets` is too low-resolution to make this feature useful?

Say we have a client turning up requesting 4 tickets via this extension. For the sake of argument say it is doing happy eyeballs across both TCP + QUIC, and IPv4 + IPv6.

- by default, we have `ServerConfig::send_tls13_tickets` = 2, and therefore we ignore the client request
- or, we have `ServerConfig::send_tls13_tickets` >= 4, and therefore we are sending a ton of tickets to every client, and the client's request is also ignored.

In other words, this extension can only be used to reduce the number of requested tickets.

I am wondering if `ServerConfig::send_tls13_tickets` should allow expressing a "default" number (as currently) and also an upper bound.

djc: > I am wondering if `ServerConfig::send_tls13_tickets` should allow expressing a "default" number (as currently) and also an upper bound.

Seems reasonable. Maybe as a `Range`, or would it be confusing if the `Range::start` is the default?

hz2: > I am wondering if `ServerConfig::send_tls13_tickets` should allow expressing a "default" number (as currently) and also an upper bound.

you're right, the extension can only count down from the default. the most compelling use case (client asking for more to prime a pool or race connections) doesn't work unless the admin raises the default for everyone, which is counter-productive.

splitting into a default and an upper bound sounds clean to me

> Seems reasonable. Maybe as a `Range`, or would it be confusing if the `Range::start` is the default?

- `Range<usize>` isn't `Copy` which seems a bit awkward for a public config field
- half-open `start..end` semantics don't really map to to "default and max", it would to be documented that `end` is inclusive-as-max while `start` is the default

a few alternatives that come to mind:

1. keep `send_tls13_tickets` as-is (the default) and add a separate `max_tls13_tickets` field for the requestable ceiling
2. a small struct `Tls13Tickets { default, max }` replacing the current field. the clearer intent but it's a breaking change to a public field

more inclined towards the former since it doesn't break anyone and the two knobs read clearly at the call site?

djc: I'd suggest a `struct Tls13Tickets { default, max }` approach for `main` and approach and a separate `max_tls13_tickets` for `rel-0.23` if you want to backport this.

hz2: done

djc:
We're looking for a clean [commit history](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#commit-history). In this case, I think it would be good to have two commits:

- One to update/extend the `ServerConfig` around tickets
- One to introduce support for the new extension

rustls/src/server/config.rs
```diff
@@ -300,6 +298,51 @@ impl ServerConfig {
...
+    ///
+    /// `max` is clamped to be at least `default`, so an ordinary client is
+    /// never sent fewer tickets than the maximum.
+    pub fn new(default: usize, max: usize) -> Self {
+        Self {
+            default,
+            max: max.max(default),
+        }
+    }
+}
```
djc: Let's get rid of this. Having two positional arguments of the same type doesn't really help read what happens.

rustls/src/msgs/handshake.rs
```diff
@@ -394,6 +394,46 @@ pub(crate) struct ServerExtensionsInput {
...
+        self.new_session_count.encode(bytes);
+        self.resumption_count.encode(bytes);
+    }
+
+    fn read(r: &mut Reader<'_>) -> Result<Self, InvalidMessage> {
+        let new_session_count = u8::read(r)?;
+        let resumption_count = u8::read(r)?;
+        Ok(Self {
+            new_session_count,
+            resumption_count,
```
djc: Nit: let's inline the bindings here.

rustls/src/msgs/handshake.rs
```diff
@@ -394,6 +394,46 @@ pub(crate) struct ServerExtensionsInput {
...
+    pub(crate) expected_count: u8,
+}
+
+impl Codec<'_> for ServerTicketRequestHint {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.expected_count.encode(bytes);
+    }
+
+    fn read(r: &mut Reader<'_>) -> Result<Self, InvalidMessage> {
+        let expected_count = u8::read(r)?;
```
djc: Nit: inline the binding.

rustls/src/msgs/handshake.rs
```diff
@@ -394,6 +394,46 @@ pub(crate) struct ServerExtensionsInput {
     pub(crate) transport_parameters: Option<TransportParameters>,
 }
 
+/// RFC 9149: ClientTicketRequest extension payload.
+#[derive(Clone, Copy, Debug, PartialEq)]
```
djc: Nit: I think this belongs in the `client_hello` module.

rustls/src/msgs/handshake.rs
```diff
@@ -394,6 +394,46 @@ pub(crate) struct ServerExtensionsInput {
...
+        Ok(Self {
+            new_session_count,
+            resumption_count,
+        })
+    }
+}
+
+/// RFC 9149: ServerTicketRequestHint extension payload.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub(crate) struct ServerTicketRequestHint {
```
djc: While this belongs into the `server_hello` module.

rustls/src/server/config.rs
```diff
@@ -334,6 +334,17 @@ impl Tls13Tickets {
...
         }
     }
+
+    /// Resolve how many tickets to send given an optional client request.
+    ///
+    /// A request (from the RFC 9149 `ticket_request` extension) is bounded by
+    /// `max`; absent a request, the `default` is used.
+    pub(super) fn resolve(&self, requested: Option<usize>) -> usize {
+        match requested {
+            Some(n) => n.min(self.max),
```
djc: Nit: please spell this as `Ord::min(n, self.max)`.

rustls/src/server/tls13.rs
```diff
@@ -718,6 +737,12 @@ mod client_hello {
             config,
         )?;
 
+        if hello.ticket_request.is_some() {
+            extensions.ticket_request = Some(ServerTicketRequestHint {
+                expected_count: send_tickets.min(u8::MAX as usize) as u8,
```
djc: Nit: this casting approach is silly. Use `Ord::min()` here, too.

rustls/src/server/tls13.rs
```diff
@@ -232,7 +232,24 @@ mod client_hello {
...
+                    .client_hello
+                    .ticket_request
+                    .as_ref()
+                    .map(|ticket_req| {
+                        usize::from(if resuming.is_some() {
+                            ticket_req.resumption_count
+                        } else {
+                            ticket_req.new_session_count
+                        })
+                    });
```
djc: Suggest passing `Option<ClientTicketRequest>` into `resolve()` directly.

rustls/src/client/hs.rs
```diff
@@ -664,6 +664,15 @@ fn emit_client_hello_for_retry(
...
         });
+
+        if let Some(ticket_req) = &config.send_ticket_request {
+            if config.resumption.tls12_resumption != Tls12Resumption::Disabled {
+                exts.ticket_request = Some(ClientTicketRequest {
+                    new_session_count: ticket_req.new_session_count,
+                    resumption_count: ticket_req.resumption_count,
+                });
+            }
+        }
```
djc: The extension is only relevant to TLS 1.3. If the client doesn't want to support 1.3, we should not send it, and I don't think blocking this based on `config.resumption.tls12_resumption` makes sense.

On the other hand, if we're (potentially) doing TLS 1.3, perhaps we we should just always send it?

rustls/src/server/config.rs
```diff
@@ -300,6 +298,62 @@ impl ServerConfig {
...
+/// How many TLS 1.3 session tickets the server sends after a handshake.
+///
+/// Because TLS 1.3 tickets are single-use, sending more than one allows a
+/// client to perform multiple resumptions.
+///
+/// This has two parts:
+///
+/// - `default`: the number of tickets sent to an ordinary client.
+/// - `max`: an upper bound on the number sent, used to cap requests made
+///   via the [RFC 9149] `ticket_request` extension.
```
djc: Let's drop this part, since we're just rehashing field docstrings.

djc:
The RFC also has this requirement:

> A client MUST abort the connection with an "illegal_parameter" alert if the "ticket_request" extension is present in any server handshake message.

(I'm a little confused by the notion that servers "MUST also send the "ticket_request" extension in the EncryptedExtensions message" but "MUST NOT send the "ticket_request" extension in any handshake message". Does EncryptedExtensions not count as a handshake message?)

hz2:
> The RFC also has this requirement:
> 
> > A client MUST abort the connection with an "illegal_parameter" alert if the "ticket_request" extension is present in any server handshake message.
> 
> (I'm a little confused by the notion that servers "MUST also send the "ticket_request" extension in the EncryptedExtensions message" but "MUST NOT send the "ticket_request" extension in any handshake message". Does EncryptedExtensions not count as a handshake message?)

i read it the same way at first, but i think it resolves once you line up [§3](https://www.rfc-editor.org/rfc/rfc9149.html#section-3) with the [iana table](https://www.rfc-editor.org/rfc/rfc9149.html#name-iana-considerations). 

the MUST NOT in [§3](https://www.rfc-editor.org/rfc/rfc9149.html#name-ticket-requests) enumerates the messages it means: "in any handshake message, __including ServerHello or HelloRetryRequest messages__". 

EncryptedExtensions is deliberately not in that list. and the [iana registry entry](https://www.rfc-editor.org/rfc/rfc9149.html#name-iana-considerations) is the authoritative bit:

> | 58 | ticket_request | CH, EE | N | Y |

so `ticket_request` is only ever valid in ClientHello and EncryptedExtensions. prohibition is really about the cleartext*handshake messages (ServerHello/HRR), where it could be observed or tampered with; EE is the one server message it's required to appear in. so no contradiction, just slightly loose wording. 

our impl matches this: we only ever emit it in EE, and the client rejects it anywhere else

hz2:
worth surfacing while we're on conformance: 
[§3](https://www.rfc-editor.org/rfc/rfc9149.html#name-ticket-requests) says the client MUST abort with `illegal_parameter` if the extension shows up in a server handshake message. we do abort in all the right places, but via the existing generic paths, which send `unsupported_extension` rather than `illegal_parameter`:

- in ServerHello: caught by `only_contains(ALLOWED_PLAINTEXT_EXTS)` -> `UnexpectedCleartextExtension`
- in HRR: caught by `UnknownHelloRetryRequestExtension`
- both map to `unsupported_extension` (110), not `illegal_parameter` (47)

honestly `unsupported_extension` reads like the more natural alert here,  and it's what rustls already uses for this whole class of "extension in the  wrong place" errors. 

do you want me to special-case `ticket_request` to send `illegal_parameter`  to match the letter of 9149, or are you comfortable keeping the existing generic behavior? 

happy either way, just want to avoid carving out a one-off if you'd rather not.

cpu:
> do you want me to special-case ticket_request to send illegal_parameter to match the letter of 9149, or are you comfortable keeping the existing generic behavior?

I don't know if anyone else feels strongly but I think it's reasonable to not special case this since both alerts are fatal and its largely cosmetic (especially without BoGo coverage mandating a specific choice). I think RFC 9149 should have used `unsupported_extension` for this case to begin with and I agree with you that it's a better fit for this condition than `illegal_parameter`.

Edit: I submitted an erratum for RFC 9149 to suggest using `unsupported_extension`. I'm not sure it merits errata-status but figured it was worth a shot.

rustls/src/client/config.rs
```diff
@@ -138,6 +138,16 @@ pub struct ClientConfig {
     /// [FIPS 140-3 IG.pdf]: https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf
     pub require_ems: bool,
 
+    /// Request a specific number of TLS 1.3 session tickets from the server
+    /// via the [RFC 9149] `ticket_request` extension.
+    ///
+    /// Set to `None` to disable sending the extension (the default).
+    ///
+    /// The extension is not sent when resumption is disabled.
```
djc: I think this is just sort of repeating the previous line?

rustls/src/client/config.rs
```diff
@@ -138,6 +138,16 @@ pub struct ClientConfig {
     /// [FIPS 140-3 IG.pdf]: https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf
     pub require_ems: bool,
 
+    /// Request a specific number of TLS 1.3 session tickets from the server
```
djc: Nit: would be nice to tighten this initial sentence to a version that fits on one line. (Could probably drop "from the server", for example.)

rustls/src/client/config.rs
```diff
@@ -605,6 +615,36 @@ pub enum Tls12Resumption {
...
+    /// Tickets desired when the server resumes using a presented ticket.
+    ///
+    /// A value of 1 is a good default for primed caches. Clients racing
+    /// multiple connections may want a higher value.
+    pub resumption_count: u8,
+}
+
+impl TicketRequest {
+    /// Create a new `TicketRequest` with the given counts.
+    pub fn new(new_session_count: u8, resumption_count: u8) -> Self {
```
djc: I think we should drop this (two positional arguments of the same type, low utility), and accordingly replace the `#[non_exhaustive]` declaration in favor of the `expect(clippy::exhaustive_structs)` or whatever it's called.

rustls/src/server/config.rs
```diff
@@ -308,6 +309,22 @@ pub struct Tls13Tickets {
...
+                let n = usize::from(if resuming {
+                    req.resumption_count
+                } else {
+                    req.new_session_count
+                });
+                Ord::min(n, self.max)
+            }
+            None => self.default,
+        }
+    }
```
djc: This looks a little messy. Suggest something more like this:

```rust
let Some(req) = requested else {
    return self.default;
};

Ord::min(usize::from(match resuming {
     true => req.resumption_count,
     false => req.new_session_count,
}), self.max)
```

rustls/src/msgs/server_hello.rs
```diff
@@ -20,6 +20,24 @@ use crate::crypto::hpke::{HpkeKem, HpkeSymmetricCipherSuite};
 use crate::enums::{CertificateType, ProtocolVersion};
 use crate::error::InvalidMessage;
 
+/// RFC 9149: ServerTicketRequestHint extension payload.
```
djc: Nit: we're following top-down style, so this should go below `ServerHelloPayload` (probably somewhere near the bottom, below larger/higher-complexity item groups).

rustls/src/msgs/client_hello.rs
```diff
@@ -28,6 +28,27 @@ use crate::log::warn;
 use crate::msgs::enums::ServerNameType;
 use crate::verify::DistinguishedName;
 
+/// RFC 9149: ClientTicketRequest extension payload.
```
djc: Nit: we're following top-down style, so this should go below `ServerHelloPayload` (probably somewhere near the bottom, below larger/higher-complexity item groups).

djc (approved):
Mostly looks good modulo a few nits. Thanks for the iterations!

djc (approved):

djc:
Needs a rebase, otherwise LGTM. Thanks!

ctz (approved):
Thanks!

djc:
I tried to push a rebased version of this branch back to your repo, but failed. New PR:

- https://github.com/rustls/rustls/pull/3126

hz2:
> I tried to push a rebased version of this branch back to your repo, but failed.

weird, thanks for getting it through and all the help throughout everyone!

cheers!

cpu:
Thanks for sticking it out through all of our feedback :-)

djc:
@hz2 FYI if you wanted to submit a backport with this to the `rel-0.23` branch I'd be open to reviewing it.

PR #3065, opened 2026-05-26
Preserve negotiated TLS version when continuing from server Acceptor

iadev09:
When continuing a server-side handshake from `server::Acceptor`, `SideCommonOutput` forwards `OutputEvent::ProtocolVersion` to `common.outputs`, but does not update the receive/send negotiated-version state.

That differs from the normal receive path, where `CaptureAppData` records the negotiated version before forwarding the output event. For TLS 1.3, that state is needed by the record layer when handling compatibility CCS after continuing from `Acceptor`.

This patch mirrors the existing `CaptureAppData` handling in `SideCommonOutput`.

### DIFF
--- a/rustls-test/tests/api/io.rs
+++ b/rustls-test/tests/api/io.rs
@@ -14,12 +14,14 @@ use rustls::enums::{ContentType, HandshakeType, ProtocolVersion};
 use rustls::error::{
     AlertDescription, ApiMisuse, Error, InvalidMessage, PeerIncompatible, PeerMisbehaved,
 };
-use rustls::{ClientConfig, Connection, ServerConfig, ServerConnection};
+use rustls::server::Acceptor;
+use rustls::{ClientConfig, Connection, HandshakeKind, ServerConfig, ServerConnection};
 use rustls_test::{
     ClientConfigExt, KeyType, OtherSession, ServerConfigExt, TestNonBlockIo, check_fill_buf,
     check_fill_buf_err, check_read, check_read_and_close, check_read_err, do_handshake, encoding,
-    make_client_config, make_client_config_with_auth, make_disjoint_suite_configs, make_pair,
-    make_pair_for_arc_configs, make_pair_for_configs, make_server_config,
+    make_client_config, make_client_config_with_auth, make_client_config_with_kx_groups,
+    make_disjoint_suite_configs, make_pair, make_pair_for_arc_configs, make_pair_for_configs,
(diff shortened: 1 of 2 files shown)


cpu:
Without giving much thought to the fix itself, I think it should come with test coverage since it seems to be addressing a bug/oversight.

iadev09:
Thanks, agreed, and sorry I missed adding test coverage in the initial patch.

I added `test_acceptor_continues_tls13_hrr_with_compatibility_ccs` in `rustls-test/tests/api/io.rs`. The test drives a TLS 1.3 handshake through `server::Acceptor`, forces HelloRetryRequest, and then continues the handshake through the client's compatibility CCS record.

I checked this red/green locally. On this PR branch, it passes:

```sh
cargo test -p rustls-test --test api test_acceptor_continues_tls13_hrr_with_compatibility_ccs --features ring --no-default-features -- --nocapture
```

I also verified the test fails on the parent commit with only the test applied:

```text
InappropriateMessage { expect_types: [Handshake], got_type: ChangeCipherSpec }
```

This came up while comparing delayed server-config flows, where a TCP/TLS `LazyConfigAcceptor`-style path exposed that `server::Acceptor` continuation could reject TLS 1.3 compatibility CCS records.

This keeps the coverage in rustls' own API tests and does not depend on QUIC or tokio-rustls.

ctz (approved):
Thanks!

rustls-test/tests/api/io.rs
```diff
@@ -1577,6 +1578,54 @@ fn test_acceptor() {
     assert_eq!(alert_content, expected);
 }
 
+#[test]
+fn test_acceptor_continues_tls13_hrr_with_compatibility_ccs() {
+    use rustls::server::Acceptor;
```
cpu: nit-picking: I think we'd typically prefer this added as a top-level import even if it's only used by this one test.

cpu (approved):
Thanks for adding the test coverage. LGTM.

rustls-test/tests/api/io.rs
iadev09: Thanks, moved `Acceptor` to a top-level import.

iadev09:
Thanks, appreciate the review and approval!

djc:
Last thing: please squash all of your commits; we use the merge queue to rebase and so we prefer not to bypass the merge queue in order to squash during merges.

iadev09:
Thanks, squashed.

Issue #1218, opened 2023-03-15
QUIC support for `Acceptor`


daxpedda:
Currently the `Acceptor` API can't be used with QUIC. It's missing something like [`new_quic()`](https://docs.rs/rustls/0.20.8/rustls/server/struct.ServerConnection.html#method.new_quic) among other things.

I actually wanted to work on this myself quite some time ago but got distracted with a bunch of other stuff. Maybe I will get to it again, thought I would leave that here considering #1216.

This was briefly discussed a bit more then a year ago on matrix:
https://matrix.to/#/!kZVoAlfRzfachIZeCA:ralith.com/$bu5HQXTF-eJWBqQhl0RrxlWQc0x9a7tUD-OTDSR7eDQ?via=mozilla.org&via=gitter.im&via=matrix.org

djc:
Given that this would be new API, I think we can postpone this past the upcoming 0.21 release. But would be great if you can work on this after #1216 is merged!

cpu:
@daxpedda We've stabilized the new API from #1216 and released v0.21.0. Are you able to work on this feature request?

daxpedda:
If you are interested in working on this yourself please go ahead, it will take me some time to get back to this.

ctz:
Closed via #3070 -- thank you @iadev09 !

iadev09:
Thank you [ctz](https://github.com/ctz) for your great work — for a better world.

PR #3070, opened 2026-05-30
Add QUIC support for server Acceptor

iadev09:
This adds a QUIC-compatible server `Acceptor` path.

The goal is to let QUIC integrations feed TLS handshake bytes into rustls until
`ClientHello` is available, inspect rustls' `ClientHello`, and then continue the
QUIC TLS session with a selected `ServerConfig`.

This keeps `ClientHello` parsing inside rustls. QUIC integrations remain
responsible for packet/CID/ACK handling and only use rustls to consume and
continue the QUIC TLS stream.

This is related to rustls/rustls#1218.

One boundary note: this PR intentionally keeps QUIC packet/CID/ACK handling out
of rustls.

The reason is that rustls only sees the QUIC TLS stream bytes. It does not own
the QUIC packet number spaces, connection IDs, peer path state, ACK generation,
anti-amplification accounting, or packet retransmission timers. Those are QUIC
transport responsibilities.

This matters for multi-packet ClientHello cases. A QUIC implementation using
this acceptor still needs to keep processing Initial packets and ACKing received
Initial packet numbers while waiting for enough CRYPTO data to expose
ClientHello. Rustls can report that more TLS bytes are needed, but it cannot
drive that transport state itself.

A Quinn-side prototype using this boundary is being discussed in
quinn-rs/quinn#2024.

So the intended split is:

- rustls: consume QUIC TLS stream bytes, expose ClientHello, continue the TLS
  session with the selected ServerConfig
- QUIC implementation: own packet/CID/ACK/timer state while feeding bytes into
  rustls

Tests cover:

- exposing `ClientHello` through the QUIC acceptor path
- continuing with a `ServerConfig` selected from `ClientHello`
- terminal read errors
- QUIC early-data size validation

### DIFF
--- a/rustls-test/tests/api/quic.rs
+++ b/rustls-test/tests/api/quic.rs
@@ -242,6 +242,248 @@ fn test_quic_handshake() {
 }
 
 #[test]
+fn test_quic_acceptor() {
+    let kt = KeyType::Rsa2048;
+    let provider = provider::DEFAULT_TLS13_PROVIDER;
+    let mut client_config = make_client_config(kt, &provider);
+    client_config.alpn_protocols = vec![b"h3".into()];
+    let client_config = Arc::new(client_config);
+    let mut server_config = make_server_config(kt, &provider);
+    server_config.alpn_protocols = vec![b"h3".into()];
+    let server_config = Arc::new(server_config);
+    let client_fips = client_config.fips();
+    let server_fips = server_config.fips();
+    let client_params = &b"client params"[..];
+    let server_params = &b"server params"[..];
+
+    let mut client = quic::ClientConnection::new(
+        client_config,
+        quic::Version::V1,
+        server_name("localhost"),
+        client_params.into(),
+    )
+    .unwrap();
+    assert_eq!(client.fips(), client_fips);
+
+    let mut acceptor = quic::Acceptor::new(quic::Version::V1);
+    assert!(acceptor.accept().unwrap().is_none());
+
+    let mut client_initial = Vec::new();
+    assert!(
+        client
+            .write_hs(&mut client_initial)
+            .is_none()
+    );
+    assert!(client_initial.len() > 8);
+
+    acceptor
+        .read_hs(&client_initial[..8])
+        .unwrap();
+    assert!(acceptor.accept().unwrap().is_none());
+
+    acceptor
+        .read_hs(&client_initial[8..])
+        .unwrap();
+    let accepted = acceptor.accept().unwrap().unwrap();
+    assert_eq!(
+        acceptor.accept().err(),
+        Some(ApiMisuse::AcceptorPolledAfterCompletion.into())
+    );
+    assert_eq!(
+        acceptor.read_hs(&[]).err(),
+        Some(ApiMisuse::AcceptorPolledAfterCompletion.into())
+    );
+    {
+        let client_hello = accepted.client_hello();
+        assert_eq!(
+            client_hello
+                .server_name()
+                .unwrap()
+                .as_ref(),
+            "localhost"
+        );
+        assert_eq!(
+            client_hello
+                .alpn()
+                .unwrap()
+                .collect::<Vec<_>>(),
+            vec![b"h3".as_slice()]
+        );
+    }
+
+    let mut server = accepted
+        .into_connection(server_config, server_params.into())
+        .unwrap();
+    assert_eq!(server.fips(), server_fips);
+    assert_eq!(server.quic_transport_parameters(), Some(client_params));
+
+    step(&mut server, &mut client)
+        .unwrap()
+        .unwrap();
+    step(&mut client, &mut server)
+        .unwrap()
+        .unwrap();
+    step(&mut server, &mut client)
+        .unwrap()
+        .unwrap();
+    step(&mut client, &mut server)
+        .unwrap()
+        .unwrap();
+
+    assert!(!client.is_handshaking());
+    assert!(!server.is_handshaking());
+    assert_eq!(client.quic_transport_parameters(), Some(server_params));
+}
+
+#[test]
+fn test_quic_acceptor_continues_with_server_config_chosen_from_client_hello() {
+    let kt = KeyType::Rsa2048;
+    let provider = provider::DEFAULT_TLS13_PROVIDER;
+    let mut client_config = make_client_config(kt, &provider);
+    client_config.alpn_protocols = vec![b"h3".into()];
+    let client_config = Arc::new(client_config);
+
+    let mut h3_server_config = make_server_config(kt, &provider);
+    h3_server_config.alpn_protocols = vec![b"h3".into()];
+    let h3_server_config = Arc::new(h3_server_config);
+
+    let mut incompatible_server_config = make_server_config(kt, &provider);
+    incompatible_server_config.alpn_protocols = vec![b"hq-interop".into()];
+    let incompatible_server_config = Arc::new(incompatible_server_config);
+
+    let client_params = &b"client params"[..];
+    let server_params = &b"server params"[..];
+
+    let mut client = quic::ClientConnection::new(
+        client_config,
+        quic::Version::V1,
+        server_name("localhost"),
+        client_params.into(),
+    )
+    .unwrap();
+
+    let mut acceptor = quic::Acceptor::new(quic::Version::V1);
+    let mut client_initial = Vec::new();
+    assert!(
+        client
+            .write_hs(&mut client_initial)
+            .is_none()
+    );
+
+    acceptor
+        .read_hs(&client_initial)
+        .unwrap();
+    let accepted = acceptor.accept().unwrap().unwrap();
+
+    let selected_config = {
+        let client_hello = accepted.client_hello();
+        assert_eq!(
+            client_hello
+                .server_name()
+                .map(|name| name.as_ref()),
+            Some("localhost")
+        );
+        assert_eq!(
+            client_hello
+                .alpn()
+                .unwrap()
+                .collect::<Vec<_>>(),
+            vec![b"h3".as_slice()]
+        );
+
+        match client_hello
+            .server_name()
+            .map(|name| name.as_ref())
+        {
+            Some("localhost") => h3_server_config,
+            _ => incompatible_server_config,
+        }
+    };
+
+    let mut server = accepted
+        .into_connection(selected_config, server_params.into())
+        .unwrap();
+
+    step(&mut server, &mut client)
+        .unwrap()
+        .unwrap();
+    step(&mut client, &mut server)
+        .unwrap()
+        .unwrap();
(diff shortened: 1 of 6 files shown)


djc:
Why is this in draft state? Does it make sense for me to spend time reviewing this?

iadev09:
Yes, I think it makes sense to review this now.

I opened it as draft because the broader discussion in quinn-rs/quinn#2024 has two coupled pieces:

- the rustls-side QUIC `Acceptor` boundary
- the Quinn-side consumer/API shape, currently depending on rustls 0.24 work and an Quinn 

The callback solution for compablitiy with legacy and staged api with rustls 0.24  in same discussion threads cause complexity. 

I have also made some downstream benchmarks as well. i haven't seen any incompatibility with the discussed conditions yet.

This solution (Quic -side) is smaller than TCP.  Because  QUIC already provides ordered plaintext crypto stream bytes.

djc:
Please rebase once #3074 is merged.

rustls/src/server/hs.rs
```diff
@@ -406,7 +406,34 @@ impl ChooseConfig {
...
+                .server_certificate_types
+                .as_deref(),
+            client_cert_types: client_hello
+                .client_certificate_types
+                .as_deref(),
+            cipher_suites: &client_hello.cipher_suites,
+            certificate_authorities: client_hello
+                .certificate_authority_names
+                .as_deref(),
+            named_groups: client_hello.named_groups.as_deref(),
```
djc: Please update this to use the constructor.

rustls/src/quic.rs
```diff
@@ -346,6 +332,155 @@ impl Debug for ServerConnection {
...
+}
+
+impl Debug for Accepted {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("quic::Accepted")
+            .finish_non_exhaustive()
+    }
+}
+
+fn check_server_config(config: &ServerConfig) -> Result<(), Error> {
```
djc, line 477: I'd like this extraction to be in a separate commit.

rustls-test/tests/api/quic.rs
```diff
@@ -239,6 +241,306 @@ fn test_quic_handshake() {
...
+        .unwrap()
+        .unwrap();
+
+    assert!(!client.is_handshaking());
+    assert!(!server.is_handshaking());
+    assert_eq!(client.quic_transport_parameters(), Some(server_params));
+    assert_eq!(server.quic_transport_parameters(), Some(client_params));
+}
+
+#[test]
```
djc, line 426: It's not obvious to me that these add substantial coverage (at least for things in `check_server_config()` -- for other parts, maybe there's other ways to deduplicate the logic?), so I'm not sure having all these tests is worth it -- maybe prune them down a bit?

rustls/src/server/connection.rs
```diff
@@ -423,14 +422,7 @@ pub struct Accepted {
...
     pub fn client_hello(&self) -> ClientHello<'_> {
-        let client_hello = self.choose_config.client_hello();
-        let server_name = client_hello
-            .server_name
-            .as_ref()
-            .and_then(ServerNamePayload::to_dns_name_normalized)
-            .map(Cow::Owned);
-
-        let ch = ClientHello::new(client_hello, None, server_name, None);
+        let ch = self.choose_config.client_hello();
```
djc, line 425: Let's use a separate commit for this refactoring.

rustls/src/quic.rs
```diff
@@ -346,6 +332,155 @@ impl Debug for ServerConnection {
...
+    /// Supply more handshake data with [`Acceptor::read_hs()`] and call this function again.
+    ///
+    /// Returns `Ok(Some(accepted))` if the connection has been accepted. Call
+    /// [`Accepted::into_connection()`] to continue. Do not call this function again.
+    pub fn accept(&mut self) -> Result<Option<Accepted>, Error> {
+        let Some(mut connection) = self.inner.take() else {
+            return Err(ApiMisuse::AcceptorPolledAfterCompletion.into());
+        };
+
+        let state = core::mem::replace(
```
djc: Nit: import `mem` directly.

djc (commented):
Looking pretty good, some nits to address.

rustls/src/server/hs.rs
iadev09: updated to use constructor.

rustls/src/quic.rs
iadev09: splited in [Extract QUIC server config validation](https://github.com/rustls/rustls/pull/3070/changes/01355a26d3a64642edfe4d0d06bad409aa6f8b11)

rustls-test/tests/api/quic.rs
iadev09: Pruned extra validation tests.

rustls/src/server/hs.rs
```diff
@@ -406,7 +406,18 @@ impl ChooseConfig {
...
+        let server_name = client_hello
+            .server_name
+            .as_ref()
+            .and_then(ServerNamePayload::to_dns_name_normalized)
+            .map(Cow::Owned);
+
+        ClientHello::new(client_hello, None, server_name, None)
+    }
+
+    fn client_hello_payload(&self) -> &ClientHelloPayload {
```
djc: Is there any point in retaining this as a separate method?

rustls/src/server/connection.rs
iadev09: Split into   [Share ClientHello construction for server Accepted](https://github.com/rustls/rustls/pull/3070/changes/c6b50aabe894fa17ea937bcd2c2726dd435d0959)

rustls/src/quic.rs
iadev09: done

rustls/src/quic.rs
```diff
@@ -332,6 +333,135 @@ impl Debug for ServerConnection {
...
+    }
+
+    /// Check if a `ClientHello` message has been received.
+    ///
+    /// Returns `Ok(None)` if the complete `ClientHello` has not yet been received.
+    /// Supply more handshake data with [`Acceptor::read_hs()`] and call this function again.
+    ///
+    /// Returns `Ok(Some(accepted))` if the connection has been accepted. Call
+    /// [`Accepted::into_connection()`] to continue. Do not call this function again.
+    pub fn accept(&mut self) -> Result<Option<Accepted>, Error> {
```
djc, line 391: Why are we not yielding an `AcceptedAlert` here?

rustls/src/quic.rs
```diff
@@ -332,6 +333,135 @@ impl Debug for ServerConnection {
...
+        };
+
+        self.connection.core.state =
+            Ok(self
+                .choose_config
+                .use_config(config, exts, &mut output)?);
+
+        Ok(ServerConnection {
+            inner: self.connection,
+        })
```
djc, line 466: Can we deduplicate more logic between `server::Accepted` and `quic::Accepted`? I think some parts are missing/different here. (Like `self.connection.fips = config.fips();`.)

rustls/src/quic.rs
iadev09: QUIC does not send TLS alert records. A TLS alert is carried out-of-band as a QUIC CONNECTION_CLOSE error.
AcceptedAlert would be empty.

rustls/src/quic.rs
iadev09: yes.  I have added FIPS propagation for the QUIC connection state.

rustls/src/server/hs.rs
iadev09: inlined  into `ChooseConfig::client_hello()`.

rustls/src/quic.rs
```diff
@@ -332,6 +346,137 @@ impl Debug for ServerConnection {
...
+    /// Supply more handshake data with [`Acceptor::read_hs()`] and call this function again.
+    ///
+    /// Returns `Ok(Some(accepted))` if the connection has been accepted. Call
+    /// [`Accepted::into_connection()`] to continue. Do not call this function again.
+    pub fn accept(&mut self) -> Result<Option<Accepted>, Error> {
+        let Some(mut connection) = self.inner.take() else {
+            return Err(ApiMisuse::AcceptorPolledAfterCompletion.into());
+        };
+
+        let state = mem::replace(
```
djc: You can append a `?` to this expression to yield out the error directly, and hoist the `Ok(..)` wrapper out of the final `match` expression.

rustls/src/quic.rs
```diff
@@ -332,6 +346,137 @@ impl Debug for ServerConnection {
...
+    /// Returns `Ok(None)` if the complete `ClientHello` has not yet been received.
+    /// Supply more handshake data with [`Acceptor::read_hs()`] and call this function again.
+    ///
+    /// Returns `Ok(Some(accepted))` if the connection has been accepted. Call
+    /// [`Accepted::into_connection()`] to continue. Do not call this function again.
+    pub fn accept(&mut self) -> Result<Option<Accepted>, Error> {
+        let Some(mut connection) = self.inner.take() else {
+            return Err(ApiMisuse::AcceptorPolledAfterCompletion.into());
+        };
+
```
djc, line 395: Compared to the server `Acceptor`, this is missing the `connection.process_new_packets()` call. Is that intentional?

rustls/src/quic.rs
```diff
@@ -355,14 +500,16 @@ fn check_server_config(config: &ServerConfig) -> Result<(), Error> {
...
 struct ConnectionCommon<Side: SideData> {
     core: ConnectionCore<Side>,
+    fips: FipsStatus,
     deframer_buffer: VecInput,
     quic: Quic,
 }
 
 impl<Side: SideData> ConnectionCommon<Side> {
-    fn new(core: ConnectionCore<Side>, quic: Quic) -> Self {
+    fn new(core: ConnectionCore<Side>, fips: FipsStatus, quic: Quic) -> Self {
```
djc, line 506: Again, this should be a separate commit.

djc (approved):

rustls/src/quic.rs
iadev09: cleanup done:

rustls/src/quic.rs
iadev09: splitted into commit, [Track QUIC connection FIPS status](https://github.com/rustls/rustls/pull/3070/changes/a2edd3d0b40ed3fa61f78c1c1b955499a87b5383)

rustls/src/quic.rs
iadev09: Yes. 
process_new_packets() comment is reply-only:   QUIC read_hs() already calls it, unlike TCP read_tls().

ctz (approved):
👍🏽

iadev09:
🤝

PR #3080, opened 2026-06-05
Introduce initial split-mode API

### LINKED ISSUE #2761: Consider a reader-writer split for the unbuffered API

ctz:
This is a small slice of #2953 attached to the existing API, rather than replacing everything.

fixes #2795 fixes #2761

### DIFF
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -10,6 +10,7 @@ use super::hs::ClientHelloInput;
 use crate::client::EchStatus;
 use crate::common_state::{CommonState, ConnectionOutputs, EarlyDataEvent, Event, Protocol, Side};
 use crate::conn::private::SideOutput;
+use crate::conn::split::SplitConnection;
 use crate::conn::{
     Connection, ConnectionCommon, ConnectionCore, IoState, KeyingMaterialExporter, Reader,
     SideCommonOutput, SideData, Writer,
@@ -37,6 +38,25 @@ impl fmt::Debug for ClientConnection {
 }
 
 impl ClientConnection {
+    /// Split a post-handshake connection into a [`SplitConnection`].
+    ///
+    /// This allows the two directions (transmit and receive) of the connection to be progressed
+    /// separately (including by different threads, which would allow dedicating a CPU core for each
+    /// direction rather than one per connection; this can dramatically improve performance for
+    /// full-duplex protocols).
+    ///
+    /// It also separates out the [`ConnectionOutputs`] which gives the application direct control
+    /// of how long this is kept.
+    ///
+    /// This fails if:
+    ///
+    /// - the handshake is not complete. Check with [`Connection::is_handshaking()`].
+    /// - there is any buffered application data.  Check with [`Connection::reader()`].
+    /// - there is any buffered TLS data to send.  Obtain it first with [`Connection::write_tls()`].
+    pub fn split(self) -> Result<SplitConnection<ClientSide>, Error> {
+        self.inner.split()
+    }
+
     /// Returns an `io::Write` implementer you can write bytes to
     /// to send TLS1.3 early data (a.k.a. "0-RTT data") to the server.
     ///
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -12,7 +12,7 @@ use crate::common_state::{CommonState, ConnectionOutputs, EarlyDataEvent, Event,
 use crate::conn::private::SideOutput;
 use crate::conn::{
     Connection, ConnectionCommon, ConnectionCore, IoState, KeyingMaterialExporter, Reader,
-    SideCommonOutput, Writer,
+    SideCommonOutput, SideData, Writer,
 };
 #[cfg(doc)]
 use crate::crypto;
@@ -417,7 +417,7 @@ impl ClientConnectionData {
 #[derive(Debug)]
 pub struct ClientSide;
 
-impl crate::conn::SideData for ClientSide {}
+impl SideData for ClientSide {}
 
 impl crate::conn::private::Side for ClientSide {
     type Data = ClientConnectionData;
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -212,6 +212,31 @@ impl ConnectionOutput for ConnectionOutputs {
     }
 }
 
+impl fmt::Debug for ConnectionOutputs {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self {
+            negotiated_version,
+            handshake_kind,
+            suite,
+            negotiated_kx_group,
+            alpn_protocol,
+            peer_identity,
+            extended_master_secret,
+            exporter: _,
+            early_exporter: _,
+        } = self;
+        f.debug_struct("ConnectionOutputs")
+            .field("negotiated_version", negotiated_version)
+            .field("handshake_kind", handshake_kind)
+            .field("suite", suite)
+            .field("negotiated_kx_group", negotiated_kx_group)
+            .field("alpn_protocol", alpn_protocol)
+            .field("peer_identity", peer_identity)
+            .field("extended_master_secret", extended_master_secret)
+            .finish_non_exhaustive()
+    }
+}
+
 /// Send an alert via `output` if `error` specifies one.
 pub(crate) fn maybe_send_fatal_alert(send: &mut dyn SendOutput, error: &Error) {
     let Ok(alert) = AlertDescription::try_from(error) else {
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -32,6 +32,9 @@ mod send;
 use send::DEFAULT_BUFFER_LIMIT;
 pub(crate) use send::{SendOutput, SendPath};
 
+pub(crate) mod split;
+use split::SplitConnection;
+
 use crate::crypto::cipher::OutboundPlain;
 
 /// A trait generalizing over buffered client or server connections.
@@ -621,6 +624,20 @@ impl<Side: SideData> ConnectionCommon<Side> {
                 .has_received_close_notify,
         }
     }
+
+    pub(crate) fn split(self) -> Result<SplitConnection<Side>, Error> {
+        // `SplitConnection` cannot be used to progress a handshake.
+        if self.is_handshaking() {
+            return Err(ApiMisuse::SplitDuringHandshake.into());
+        }
+
+        // We are about to drop `Buffers`
+        if !self.buffers.is_empty() {
+            return Err(ApiMisuse::SplitWithPendingBuffers.into());
+        }
+
+        SplitConnection::try_from(self.core)
+    }
 }
 
 impl<Side: SideData> ConnectionCommon<Side> {
@@ -697,6 +714,12 @@ impl Buffers {
             sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
             has_seen_eof: false,
         }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.received_plaintext.is_empty()
+            && self.deframer_buffer.filled().is_empty()
+            && self.sendable_plaintext.is_empty()
     }
 }
 
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -142,6 +142,11 @@ impl ReceivePath {
                 // "Any data received after a closure alert has been received MUST be ignored."
                 // -- <https://datatracker.ietf.org/doc/html/rfc8446#section-6.1>
                 // This is data that has already been accepted in `read_tls`.
+
+                // First, discard actually-processed bytes.
+                input.discard(self.deframer.take_discard());
+
+                // Then the rest of any input data.
                 let entirety = input.slice_mut().len();
                 input.discard(entirety);
                 break;
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -5,8 +5,7 @@ use core::ops::Range;
 use super::SendOutput;
 use crate::SideData;
 use crate::common_state::{
-    ConnectionOutput, ConnectionOutputs, Event, Output, OutputEvent, Side, UnborrowedPayload,
-    maybe_send_fatal_alert,
+    ConnectionOutput, Event, Output, OutputEvent, Side, UnborrowedPayload, maybe_send_fatal_alert,
 };
 use crate::conn::StateMachine;
 use crate::conn::private::SideOutput;
@@ -50,6 +49,24 @@ impl ReceivePath {
             seen_consecutive_empty_fragments: 0,
             tls13_tickets_received: 0,
         }
+    }
+
+    pub(crate) fn process_recv_traffic<Side: SideData>(
+        &mut self,
+        input: &mut dyn TlsInputBuffer,
+        state: &mut Result<Side::State, Error>,
+        send: &mut dyn SendOutput,
+    ) -> Result<Option<UnborrowedPayload>, Error> {
+        self.process_new_packets::<Side>(
+            input,
+            state,
+            &mut JoinOutput {
+                outputs: &mut Discard,
+                quic: None,
+                send,
+                side: &mut Discard,
+            },
+        )
     }
 
     pub(super) fn process_new_packets<'a, 'm, Side: SideData>(
@@ -504,10 +521,20 @@ impl<'m> Output<'m> for CaptureAppData<'_, '_, 'm> {
 }
 
 pub(super) struct JoinOutput<'a> {
-    pub(super) outputs: &'a mut ConnectionOutputs,
+    pub(super) outputs: &'a mut dyn ConnectionOutput,
     pub(super) quic: Option<&'a mut dyn QuicOutput>,
     pub(super) send: &'a mut dyn SendOutput,
     pub(super) side: &'a mut dyn SideOutput,
+}
+
+struct Discard;
+
+impl ConnectionOutput for Discard {
+    fn handle(&mut self, _ev: OutputEvent<'_>) {}
+}
+
+impl SideOutput for Discard {
+    fn emit(&mut self, _ev: Event<'_>) {}
 }
 
 /// Tracking technically-allowed protocol actions
--- a/rustls/src/conn/split.rs
+++ b/rustls/src/conn/split.rs
@@ -1,0 +1,508 @@
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use core::fmt;
+use core::ops::{DerefMut, Range};
+use std::sync::MutexGuard;
+
+use crate::client::ClientSide;
+use crate::common_state::UnborrowedPayload;
+use crate::conn::{ConnectionCore, ReceivePath, SendOutput, SendPath};
+use crate::crypto::cipher::{MessageEncrypter, OutboundPlain};
+use crate::enums::ProtocolVersion;
+use crate::error::{AlertDescription, ErrorWithAlert};
+use crate::lock::Mutex;
+use crate::msgs::{AlertLevel, Delocator, Message, TlsInputBuffer};
+use crate::sync::Arc;
+use crate::tls13::key_schedule::KeyScheduleTrafficSend;
+use crate::{ConnectionOutputs, Error, SideData};
+
+/// A post-handshake connection which has been split by direction.
+///
+/// Typically you will immediately destructure this type, and give the components
+/// to different threads/handlers to progress separately.
+#[expect(clippy::exhaustive_structs)]
+#[derive(Debug)]
+pub struct SplitConnection<Side: SideData> {
+    /// The ability to encrypt data to be sent.
+    pub send: SendTraffic,
+    /// The ability to decrypt received data.
+    pub receive: ReceiveTraffic<Side>,
+    /// Facts about the connection established during the handshake.
+    pub outputs: ConnectionOutputs,
+}
+
+impl<Side: SideData> TryFrom<ConnectionCore<Side>> for SplitConnection<Side> {
+    type Error = Error;
+
+    fn try_from(conn: ConnectionCore<Side>) -> Result<Self, Error> {
+        let send = Arc::new(Mutex::new(conn.common.send));
+        let state = conn.state?;
+
+        Ok(Self {
+            send: SendTraffic(send.clone()),
+            receive: ReceiveTraffic {
+                state,
+                recv: conn.common.recv,
+                send,
+                pending_flush_sender: false,
+            },
+            outputs: conn.common.outputs,
+        })
+    }
+}
+
+/// The send-side of a connection, after a successful handshake.
+///
+/// You can use this object to send data to the peer.
+pub struct SendTraffic(pub(crate) Arc<Mutex<SendPath>>);
+
+impl SendTraffic {
+    /// Write application data to the peer.
+    ///
+    /// The TLS data to send to the peer is returned. This data should then
+    /// be communicated to the peer, in order.
+    pub fn write(&mut self, application_data: OutboundPlain<'_>) -> Vec<Vec<u8>> {
+        let mut inner = self.0.lock().unwrap();
+        inner.maybe_refresh_traffic_keys();
+        inner.send_appdata_encrypt(application_data);
+        inner.sendable_tls.take()
+    }
+
+    /// Conclude sending traffic by sending a `close_notify` alert.
+    ///
+    /// The alert is written into a Vec which is returned along with any pending data.
+    /// This data should then be communicated to the peer.
+    ///
+    /// This is the final possible operation with a [`SendTraffic`].
+    pub fn close(mut self) -> Vec<Vec<u8>> {
+        let mut inner = self.0.lock().unwrap();
+        inner.send_close_notify();
+        drop(inner);
+        self.take_data()
+    }
+
+    /// Obtain any pending data to write to the peer.
+    ///
+    /// Any such pending data will be output with any call to [`SendTraffic::write()`]
+    /// so there is no need to call this function if you have recently written data
+    /// using this.
+    ///
+    /// The TLS data to send to the peer is returned. This data should then
+    /// be communicated to the peer.
+    ///
+    /// This is useful to handle a [`ReceiveTrafficState::FlushSender`] event, but
+    /// where you don't have any plaintext to send.
+    pub fn take_data(&mut self) -> Vec<Vec<u8>> {
+        let mut inner = self.0.lock().unwrap();
+        inner.maybe_refresh_traffic_keys();
+        inner.sendable_tls.take()
+    }
+
+    /// Sends a TLS1.3 `key_update` message to refresh a connection's keys.
+    ///
+    /// The main reason to call this manually is to roll keys when it is known
+    /// a connection will be idle for a long period.
+    ///
+    /// rustls implicitly and automatically refreshes traffic keys when needed
+    /// according to the selected cipher suite's cryptographic constraints. There
+    /// is therefore no need to call this manually to avoid cryptographic keys
+    /// "wearing out".
+    ///
+    /// This call refreshes our encryption keys. Once the peer receives the message,
+    /// it refreshes _its_ encryption and decryption keys and sends a response.
+    /// Once we receive that response, we refresh our decryption keys to match.
+    /// At the end of this process, keys in both directions have been refreshed.
+    ///
+    /// Note that this process does not happen synchronously: this call just
+    /// arranges that the `key_update` message will be included in the next
+    /// `write()` output.
+    ///
+    /// This returns an error if a version prior to TLS1.3 is negotiated.
+    ///
+    /// # Usage advice
+    /// Note that other implementations (including rustls) may enforce limits on
+    /// the number of `key_update` messages allowed on a given connection to prevent
+    /// denial of service. Therefore, this should be called sparingly.
+    pub fn refresh_traffic_keys(&mut self) -> Result<(), Error> {
+        self.0
+            .lock()
+            .unwrap()
+            .refresh_traffic_keys()
+    }
+}
+
+impl fmt::Debug for SendTraffic {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("SendTraffic")
+            .finish_non_exhaustive()
+    }
+}
+
+/// The receive-side of a connection, after a successful handshake.
+///
+/// You can use this object to receive data from the peer.
+pub struct ReceiveTraffic<Side: SideData> {
+    pub(crate) state: Side::State,
+    pub(crate) recv: ReceivePath,
+    pub(crate) send: Arc<Mutex<SendPath>>,
+    pub(crate) pending_flush_sender: bool,
+}
+
+impl<Side: SideData> ReceiveTraffic<Side> {
+    /// Receive application data from the peer.
+    ///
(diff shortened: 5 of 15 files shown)


rustls/src/client/connection.rs
```diff
@@ -37,6 +38,22 @@ impl fmt::Debug for ClientConnection {
 }
 
 impl ClientConnection {
+    /// Split a post-handshake connection into a [`SplitConnection`].
+    ///
+    /// The [`SplitConnection`] is typically deconstructed by the caller.
+    /// [`SplitConnection::send`] is the send/encryption side of the connection.
+    /// [`SplitConnection::receive`] is the receive/decryption side of the connection.
+    /// [`SplitConnection::outputs`] are facts about the connection learned during the handshake.
```
djc: Suggest moving these docs to `SplitConnection`, and referencing those as a whole via an intra-doc link. Duplicating that stuff here seems low-value.

rustls/src/server/connection.rs
```diff
@@ -45,6 +47,22 @@ impl ServerConnection {
         })
     }
 
+    /// Split a post-handshake connection into a [`SplitConnection`].
+    ///
+    /// The [`SplitConnection`] is typically deconstructed by the caller.
+    /// [`SplitConnection::send`] is the send/encryption side of the connection.
+    /// [`SplitConnection::receive`] is the receive/decryption side of the connection.
+    /// [`SplitConnection::outputs`] are facts about the connection learned during the handshake.
```
djc: Same as for client.

rustls/src/conn/mod.rs
```diff
@@ -621,6 +626,33 @@ impl<Side: SideData> ConnectionCommon<Side> {
...
+                .is_empty()
+            || !self
+                .buffers
+                .sendable_plaintext
+                .is_empty()
+        {
+            return Err(ApiMisuse::SplitWithPendingBuffers.into());
+        }
+
+        split::SplitConnection::try_from(self.core)
```
djc: Nit: import `SplitConnection` directly?

rustls/src/conn/mod.rs
```diff
@@ -32,6 +32,11 @@ mod send;
 use send::DEFAULT_BUFFER_LIMIT;
 pub(crate) use send::{SendOutput, SendPath};
 
+/// Types related to "split" mode.
+///
+/// See [`split::SplitConnection`] for more information.
+pub mod split;
```
djc: Would suggest keeping this module private for now, only exporting `SplitConnection` directly.

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// Note that other implementations (including rustls) may enforce limits on
+    /// the number of `key_update` messages allowed on a given connection to prevent
+    /// denial of service. Therefore, this should be called sparingly.
+    ///
+    /// rustls implicitly and automatically refreshes traffic keys when needed
+    /// according to the selected cipher suite's cryptographic constraints. There
+    /// is therefore no need to call this manually to avoid cryptographic keys
+    /// "wearing out".
+    ///
+    /// The main reason to call this manually is to roll keys when it is known
```
djc: The last sentence here is the most helpful, maybe put it up front and make it more explicit when to call this (as in, before or after going idle for a long period). Can you concisely articulate what problem this is solving?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+        inner.maybe_refresh_traffic_keys();
+        inner.sendable_tls.pop()
+    }
+
+    /// Conclude sending traffic by sending a `close_notify` alert.
+    ///
+    /// The alert is written into a Vec which is returned.
+    ///
+    /// This is the final possible operation with a [`SendTraffic`].
+    pub fn close(mut self) -> Vec<u8> {
```
djc: Should `SendTraffic` impl `Drop`?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+///
+/// You can use this object to send data to the peer.
+pub struct SendTraffic(pub(crate) Arc<Mutex<SendPath>>);
+
+impl SendTraffic {
+    /// Write application data to the peer.
+    ///
+    /// The TLS data to send to the peer is returned. This data should then
+    /// be communicated to the peer, in order.
+    pub fn write(&mut self, application_data: OutboundPlain<'_>) -> Result<Vec<Vec<u8>>, Error> {
```
djc: I'd suggest to just take `&[u8]` and wrap it in `OutboundPlain`.

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// Obtain any pending data to write to the peer.
+    ///
+    /// Any such pending data will be output with any call to [`SendTraffic::write()`]
+    /// so there is no need to call this function if you have recently written data
+    /// using this.
+    ///
+    /// The TLS data to send to the peer is returned. This data should then
+    /// be communicated to the peer.
+    ///
+    /// This is useful to handle a [`ReceiveTrafficState::WakeSender`] event, but
```
djc: This seems a little obscure. Could we have `write()` take `Option` instead?

rustls/src/conn/receive.rs
```diff
@@ -52,6 +51,24 @@ impl ReceivePath {
         }
     }
 
+    pub(crate) fn process_recv_traffic<Side: SideData>(
```
djc, line 54: Why not inline this?

rustls/src/conn/receive.rs
```diff
@@ -490,12 +507,22 @@ impl<'m> Output<'m> for CaptureAppData<'_, '_, 'm> {
...
 
+struct Discard;
+
+impl ConnectionOutput for Discard {
+    fn handle(&mut self, _ev: OutputEvent<'_>) {}
+}
+
+impl SideOutput for Discard {
+    fn emit(&mut self, _ev: Event<'_>) {}
+}
```
djc, line 538: And presumably this is safe because `SideOutput` and `ConnectionOutput` are only used during the `Handshake`? Would still be nice to enforce that at compile time.

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// A [`ReceiveTrafficState`] is returned on success:
+    ///
+    /// - [`ReceiveTrafficState::Available`] if an application data message
+    ///   has been received.
+    /// - [`ReceiveTrafficState::WakeSender`] if the previous operation may have
+    ///   caused some data to become sendable. The application should service
+    ///   the send-side of the connection.
+    /// - [`ReceiveTrafficState::Await`] if more IO is required.
+    /// - [`ReceiveTrafficState::CloseNotify`] if the peer has cleanly
+    ///   closed the receive direction of the connection.
```
djc: Recommend moving this documentation to `ReceiveTrafficState`.

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+        self.ensure_locked();
+        let Self::Locked(guard) = self else {
+            unreachable!();
+        };
+        guard
+    }
+}
+
+impl SendOutput for SendAdapter<'_> {
+    fn negotiated_version(&mut self, version: crate::enums::ProtocolVersion) {
```
djc: Nit: import `ProtocolVersion`?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+/// for the remainder of the processing. This means that, for example,
+/// a sequence of sent messages is not interleaved with others from another
+/// thread.
+enum SendAdapter<'a> {
+    Unlocked(&'a Mutex<SendPath>),
+    Locked(MutexGuard<'a, SendPath>),
+}
+
+impl<'a> SendAdapter<'a> {
+    fn ensure_locked(&mut self) {
```
djc: This only has one caller... can't we inline it?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+/// [`Await`]: ReceiveTrafficState::Await
+/// [`WakeSender`]: ReceiveTrafficState::WakeSender
+/// [`Available`]: ReceiveTrafficState::Available
+/// [`CloseNotify`]: ReceiveTrafficState::CloseNotify
+#[expect(clippy::exhaustive_enums)]
+pub enum ReceiveTrafficState<'a, Side: SideData> {
+    /// More input is required.
+    ///
+    /// Collect it into your input buffer, and then call [`ReceiveTraffic::read()`] again.
+    Await(ReceiveTraffic<Side>),
```
djc: Don't love the variant name. `BlockedRead`/`ReadBlocked` or something? `ReadMore`?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+
+    rt: ReceiveTraffic<Side>,
+}
+
+impl<Side: SideData> ReceivedApplicationData<'_, Side> {
+    /// Finish processing this received data.
+    ///
+    /// This yields the discard value that should now be applied to the originating
+    /// buffer, and the next [`ReceiveTraffic`] state.
+    pub fn into_next(self) -> (usize, ReceiveTraffic<Side>) {
```
djc: Should this take the `&mut impl TlsInputBuffer` instead of yielding the `usize`?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+/// [`CloseNotify`]: ReceiveTrafficState::CloseNotify
+#[expect(clippy::exhaustive_enums)]
+pub enum ReceiveTrafficState<'a, Side: SideData> {
+    /// More input is required.
+    ///
+    /// Collect it into your input buffer, and then call [`ReceiveTraffic::read()`] again.
+    Await(ReceiveTraffic<Side>),
+
+    /// The sender may have new data to send.
+    WakeSender(WakeSender<Side>),
```
djc: "Wake" has a fairly specific meaning in async Rust, maybe we should pick something different for this name?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+impl<'a> SendAdapter<'a> {
+    fn ensure_locked(&mut self) {
+        if let Self::Unlocked(m) = self {
+            *self = Self::Locked(m.lock().unwrap());
+        }
+    }
+
+    fn as_locked<'b>(&'b mut self) -> &'b mut MutexGuard<'a, SendPath> {
+        self.ensure_locked();
+        let Self::Locked(guard) = self else {
```
djc: How reliable is having locked as a signal for whether the sender needs to be woken? Maybe we should track that explicitly based on which `SendOutput` methods were called?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// data from the peer. The error may be accompanied by a TLS alert,
+    /// which can be obtained from the returned [`ErrorWithAlert`] and sent
+    /// to the peer. Following this, the underlying IO medium should be
+    /// closed by the application.
+    pub fn read<'a>(
+        mut self,
+        received_tls: &'a mut impl TlsInputBuffer,
+    ) -> Result<ReceiveTrafficState<'a, Side>, ErrorWithAlert> {
+        let mut send = SendAdapter::Unlocked(&self.send);
+        let mut state = Ok(self.state);
```
djc: Would it be cleaner to destruct `Self` into its constituent parts?

djc (commented):
Nice and small! Here's an initial round of feedback.

rustls/src/conn/split.rs
ctz: It is an overestimate for sure, but quite simple to reason about so seems like a good starting point.

rustls/src/conn/split.rs
ctz: trying out `ServiceSender`, wdyt?

rustls/src/conn/split.rs
ctz: trying `ReadMore`

rustls/src/conn/split.rs
ctz: This way allows one function to service `write()` and `writev()`-shaped callers.

rustls/src/conn/split.rs
ctz: Sorry, which would do what? It couldn't do this operation, as drop has no way to emit the alert.

rustls/src/conn/split.rs
ctz: Having applications doing `write(None)` seems a bit mysterious. Then again, I don't really like the name of this function either.

rustls/src/conn/mod.rs
ctz: Would like to avoid having all these types in the crate root, so now doing the export in another module in `lib.rs` (also, 6 types need to be exported).

rustls/src/conn/receive.rs
ctz: Attempting to keep as much detail as possible private to `mod receive`

rustls/src/conn/split.rs
ctz: That would be nice, but `self` borrows it, so the caller cannot have both `self` and `&mut impl TlsInputBuffer`

ctz: An alternative would be keeping the original buffer (as  `&mut dyn TlsInputBuffer` ) inside `ReceivedApplicationData`, alongside the bounds of the received message as an `UnborrowedPayload`. That would make `into_next(self)` do the work itself.

ctz: I have tried this approach.

rustls/src/conn/split.rs
djc: I think making it more explicit would actually make it easier to reason about? The combination of implicit plus conservative makes it not that attractive IMO.

rustls/src/conn/split.rs
djc: Fair enough. Can't wait for `!Destruct`!

rustls/src/conn/split.rs
ctz: Have added this as a fixup commit, WDYT?

djc: I like it, what about you?

ctz: I don't mind it, I guess it makes it more obvious what is happening.

rustls/src/common_state.rs
```diff
@@ -212,6 +212,29 @@ impl ConnectionOutput for ConnectionOutputs {
     }
 }
 
+impl fmt::Debug for ConnectionOutputs {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self {
```
cpu, line 217: building on this rev gives me:

> error[E0027]: pattern does not mention field `extended_master_secret`
>    --> rustls/src/common_state.rs:217:13

It looks like there's a fix for this later-on in https://github.com/rustls/rustls/pull/3080/changes/01239a595900611d4732734d75fbb14c4e6fa4c0 that should be lifted out of there and into this commit.

rustls/src/client/connection.rs
```diff
@@ -37,6 +38,17 @@ impl fmt::Debug for ClientConnection {
 }
 
 impl ClientConnection {
+    /// Split a post-handshake connection into a [`SplitConnection`].
+    ///
+    /// This fails if:
+    ///
+    /// - the handshake is not complete. Check with [`Connection::is_handshaking()`].
+    /// - there is any buffered application data.  Check with [`Connection::reader()`].
+    /// - there is any buffered TLS data.  Service [`Connection::write_tls()`].
```
cpu: Will "Service" as a verb be understandable to most readers? Can we be more explicit? Maybe this isn't meaningful feedback since we're using the same verb as part of the new API surface.

(Same feedback applies to the equiv comment in `rustls/src/server/connection.rs` if you're swayed)

rustls/src/conn/mod.rs
```diff
@@ -621,6 +624,33 @@ impl<Side: SideData> ConnectionCommon<Side> {
...
+            .is_empty()
+            || !self
+                .buffers
+                .deframer_buffer
+                .filled()
+                .is_empty()
+            || !self
+                .buffers
+                .sendable_plaintext
+                .is_empty()
```
cpu: It might be nice to lift this logic into a `Buffers.is_empty()` both to make this `if` less gnarly and to centralize the logic so it'd be more obvious to update with the introduction of any additional bits in `Buffers`.

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// where you don't have any plaintext to send.
+    pub fn take_data(&mut self) -> Option<Vec<u8>> {
+        let mut inner = self.0.lock().unwrap();
+        inner.maybe_refresh_traffic_keys();
+        inner.sendable_tls.pop()
+    }
+
+    /// Conclude sending traffic by sending a `close_notify` alert.
+    ///
+    /// The alert is written into a Vec which is returned.
```
cpu: Should this also have a "This data should then be communicated to the peer." remark to match `write()` + `take_data()`?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+impl fmt::Debug for SendTraffic {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("SendTraffic")
+            .finish_non_exhaustive()
+    }
+}
+
+/// The receive-side of a connection, after a successful handshake.
```
cpu, line 141: nit: the send-side doesn't make the same comment about "after a successful handshake". Worth updating for consistency?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+                rt: Self {
+                    state,
+                    recv,
+                    send,
+                    pending_service_sender,
+                },
+            }));
+        }
+
+        // If we locked the sender during that, it is still locked and we can provide
```
cpu: having a hard time parsing the "during that" part of this comment. Can you reword?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+        self.as_locked(false)
+            .set_encrypter(cipher, max_messages);
+    }
+
+    fn update_key_schedule(&mut self, schedule: Box<KeyScheduleTrafficSend>) {
+        self.as_locked(false)
+            .update_key_schedule(schedule);
+    }
+
+    fn send_alert(&mut self, level: crate::msgs::AlertLevel, desc: AlertDescription) {
```
cpu: nit: add `AlertLevel` to the `crate::msgs::{}` import block on L14?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+        self.as_locked(true)
+            .send_alert(level, desc)
+    }
+
+    fn start_traffic(&mut self) {
+        self.as_locked(false).start_traffic();
+    }
+
+    fn send_msg(&mut self, m: crate::msgs::Message<'_>, must_encrypt: bool) {
```
cpu: nit: add `Message` to the `crate::msgs::{}` import block on L14?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+///
+/// You can use this object to send data to the peer.
+pub struct SendTraffic(pub(crate) Arc<Mutex<SendPath>>);
+
+impl SendTraffic {
+    /// Write application data to the peer.
+    ///
+    /// The TLS data to send to the peer is returned. This data should then
+    /// be communicated to the peer, in order.
+    pub fn write(&mut self, application_data: OutboundPlain<'_>) -> Result<Vec<Vec<u8>>, Error> {
```
cpu: How come this returns a `Result<_>` if the only error site is a poison mutex and we `unwrap()`? It seems like `take_data()` is pretty symmetric in that sense, but doesn't return a `Result<_>`.

More generally it seems like we're unwrapping poison away across all the mutex interactions. Is that the right choice? Could it be justified/mentioned in the commit message?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+        let mut inner = self.0.lock().unwrap();
+        inner.maybe_refresh_traffic_keys();
+        inner.send_appdata_encrypt(application_data);
+        Ok(inner.sendable_tls.take())
+    }
+
+    /// Obtain any pending data to write to the peer.
```
cpu, line 84: I think this should probably clarify that it `pop()`'s one chunk. As-written it seems a bit like you could imagine it returning all of the pending data.

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,458 @@
...
+        // safety: state consumed only on error.
+        let state = state.unwrap();
+
+        if let Some(unborrowed) = received_plain {
+            let pending_discard = recv.deframer.take_discard();
+            let UnborrowedPayload::Unborrowed(range) = unborrowed else {
+                return Err(Error::Unreachable("decrypted data should be borrowed").into());
+            };
+            drop(send_adapter);
+            return Ok(ReceiveTrafficState::Available(ReceivedApplicationData {
```
cpu, line 201: should this early return also be considering populating the `pending_service_sender` hint from checking `send_adapter` that's done below on L210...L214, or is that orthogonal for some reason that might be worth documenting?

rustls/src/client/connection.rs
```diff
@@ -37,6 +38,17 @@ impl fmt::Debug for ClientConnection {
 }
 
 impl ClientConnection {
+    /// Split a post-handshake connection into a [`SplitConnection`].
```
cpu, line 41: For both this & the `ServerConnection` comment I feel like it might be helpful to try and give some colour on _when_ & _why_ you would reach for this API. What kind of workloads? What is the advantage relative to other choices? What are the disadvantages?

cpu (commented):
Here's some initial feedback (mostly on the docs side). I think the general API design is really nice.

I'm not sure what's up with the failing [openssl-tests CI job](https://github.com/rustls/rustls/actions/runs/27757794732/job/82409115872). Seems like maybe the GH actions cache got itself into a situation where the cached openssl-3.4.0 binary is bogus? It fails consistently across retries. Maybe needs a manual cache purge or a push that changes the cache-key and causes a re-fetch?

rustls/src/conn/split.rs
ctz: Updated this (stale) comment

rustls/src/conn/split.rs
ctz: Thanks, this points out a shortcoming elsewhere (that, if `take_data()` would return `Some()`, then `close()` returns that rather than the alert.)

rustls/src/conn/split.rs
ctz: Removed the unused `Result` here. The approach of unwrapping a poison mutex I think is the right one; it transfer a prior panic within use of a mutex into other uses of it. (Naturally, the mutex is internal and so any panic is not good and 100% our problem, but at least this approach avoids any idea that any panic-to-rustls::Error mapping is possible.)

rustls/src/conn/split.rs
ctz: Addressed this.

rustls/src/conn/split.rs
ctz: I think it should, yes!

ctz: This is now addressed, which needed a change in the API to avoid that notification being delayed until the _next_ read.

Now items contained in `ReceiveTrafficState` run through subsequent state changes also (so produce further `ReceiveTrafficState` states). This API is now more general than the actual behaviour of the library (for example: if we receive one buffer containing some application data followed by a close_notify, the application data is received and retired first, and the caller needs to present the remaining data in the buffer again. There's a new test `key_update_alongside_data` which demonstrates that this works.

ctz:
> I'm not sure what's up with the failing [openssl-tests CI job](https://github.com/rustls/rustls/actions/runs/27757794732/job/82409115872). Seems like maybe the GH actions cache got itself into a situation where the cached openssl-3.4.0 binary is bogus? It fails consistently across retries. Maybe needs a manual cache purge or a push that changes the cache-key and causes a re-fetch?

https://github.com/rustls/rustls/pull/3113 -- it's a result of having a branch existing over the switch of runners.

rustls/src/client/connection.rs
ctz: Reworded.

rustls/src/conn/split.rs
ctz: now trying `FlushSender` instead of `ServiceSender`

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,471 @@
...
+        inner.maybe_refresh_traffic_keys();
+        match inner.sendable_tls.take() {
+            v if !v.is_empty() => Some(v),
+            _ => None,
+        }
+    }
+
+    /// Conclude sending traffic by sending a `close_notify` alert.
+    ///
+    /// The alert is written into a Vec which is returned. This data should
```
cpu: Should this be updated now that it's returning a `Vec<Vec<_>>` with both any pending data, and the alert?

cpu (approved):
LGTM with the latest revisions 👍

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,471 @@
...
+        }
+    }
+
+    /// Conclude sending traffic by sending a `close_notify` alert.
+    ///
+    /// The alert is written into a Vec which is returned along with any pending data.
+    /// This data should then be communicated to the peer.
+    ///
+    /// This is the final possible operation with a [`SendTraffic`].
+    pub fn close(mut self) -> Vec<Vec<u8>> {
```
djc: Nit: since this calls `take_data()`, it should be defined above it?

rustls/src/conn/split.rs
```diff
@@ -0,0 +1,471 @@
...
+    /// Any such pending data will be output with any call to [`SendTraffic::write()`]
+    /// so there is no need to call this function if you have recently written data
+    /// using this.
+    ///
+    /// The TLS data to send to the peer is returned. This data should then
+    /// be communicated to the peer.
+    ///
+    /// This is useful to handle a [`ReceiveTrafficState::FlushSender`] event, but
+    /// where you don't have any plaintext to send.
+    pub fn take_data(&mut self) -> Option<Vec<Vec<u8>>> {
```
djc: Instead of wrapping the return value in `Option`, I think we could yield an empty outer `Vec`? `Vec<Vec<u8>>` is ugly enough, though I guess it will go away with buffer externalization before release anyway?

djc (approved):
I think this is good enough (especially if it unblocks some of the other work on buffer externalization). Would still be nice to come up with a plan for open comments.

rustls/src/conn/receive.rs
ctz: So I think what this will take is something like splitting `trait StateMachine` into `trait HandshakeStateMachine` and `trait DataStateMachine` (with the latter being an unforgeable output of the former). Then we can clarify the `Output` trait type on both (for example, the `Handshake` one would not have `received_plaintext()`, the `Data` one would not have `start_traffic()` or anything involving `OutputEvent`)

However, higher level code which is currently uniform (thanks to the single `trait StateMachine`) would either have to abstract over the two, or duplicate over it.

djc: > However, higher level code which is currently uniform (thanks to the single `trait StateMachine`) would either have to abstract over the two, or duplicate over it.

Like what?

ctz: `ReceivePath::process_new_packets`

rustls/src/conn/split.rs
ctz: Open to future changes to this interface in later PRs.

rustls/src/conn/receive.rs
ctz: I suggest this is something to investigate later.

bal-e:
Woah, cool! Thanks for making this happen, I'm excited to see how this API develops further :D

PR #3081, opened 2026-06-05
Remove redundant clones

FranciscoTGouveia:
If I am not overlooking anything, these clones can be safely removed since they aren't used anywhere else later.

CC @nunoplopes


djc (approved):
Thanks! I thought there was a clippy lint for this...

djc:
Ah, we do enable `redundant_clone = "warn"` but it has some false negatives.

PR #3083, opened 2026-06-05
Expose whether Extended Master Secret was negotiated

mike-marcacci:
Adds `ConnectionOutputs::extended_master_secret() -> Option<bool>`, reporting whether the Extended Master Secret extension (RFC 7627) was negotiated.

A TLS exporter is only a sound channel binding (RFC 9266 tls-exporter) on TLS 1.3, or on TLS 1.2 with Extended Master Secret; without EMS the TLS 1.2 exporter is vulnerable to the triple-handshake attack.

Currently, rustls exposes the exporter and protocol_version(), but not EMS status, so a caller doing channel binding currently has no way to tell whether a TLS 1.2 exporter is safe to use.

See [this concrete use case in reqwest](https://github.com/seanmonstar/reqwest/pull/3048).

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -97,6 +97,7 @@ pub struct ConnectionOutputs {
     negotiated_kx_group: Option<&'static dyn SupportedKxGroup>,
     alpn_protocol: Option<ApplicationProtocol<'static>>,
     peer_identity: Option<Identity<'static>>,
+    extended_master_secret: Option<bool>,
     pub(crate) exporter: Option<Box<dyn Exporter>>,
     pub(crate) early_exporter: Option<Box<dyn Exporter>>,
 }
@@ -150,6 +151,17 @@ impl ConnectionOutputs {
         self.negotiated_version
     }
 
+    /// Whether the Extended Master Secret extension was negotiated.
+    ///
+    /// Returns:
+    /// - `None` until the handshake reaches the point where this is known.
+    /// - `None` for TLS 1.3, where the extension does not apply.
+    /// - `Some(true) for TLS 1.2 if the extension was negotiated.
+    /// - `Some(false) otherwise.
+    pub fn extended_master_secret(&self) -> Option<bool> {
+        self.extended_master_secret
+    }
+
     /// Which kind of handshake was performed.
     ///
     /// This tells you whether the handshake was a resumption or not.
@@ -183,6 +195,7 @@ impl ConnectionOutput for ConnectionOutputs {
             OutputEvent::CipherSuite(suite) => self.suite = Some(suite),
             OutputEvent::EarlyExporter(exporter) => self.early_exporter = Some(exporter),
             OutputEvent::Exporter(exporter) => self.exporter = Some(exporter),
+            OutputEvent::ExtendedMasterSecret(ems) => self.extended_master_secret = Some(ems),
             OutputEvent::HandshakeKind(hk) => {
                 assert!(self.handshake_kind.is_none());
                 self.handshake_kind = Some(hk);
@@ -278,6 +291,7 @@ pub(crate) enum OutputEvent<'a> {
     CipherSuite(SupportedCipherSuite),
     EarlyExporter(Box<dyn Exporter>),
     Exporter(Box<dyn Exporter>),
+    ExtendedMasterSecret(bool),
     HandshakeKind(HandshakeKind),
     KeyExchangeGroup(&'static dyn SupportedKxGroup),
     PeerIdentity(Identity<'static>),
--- a/rustls-test/tests/api/api.rs
+++ b/rustls-test/tests/api/api.rs
@@ -824,6 +824,32 @@ fn test_tls13_exporter() {
 }
 
 #[test]
+fn test_extended_master_secret_reporting() {
+    // TLS 1.2: rustls always offers the Extended Master Secret extension, so a
+    // rustls<->rustls 1.2 handshake negotiates it and reports `Some(true)` on
+    // both ends.
+    let provider = provider::DEFAULT_TLS12_PROVIDER;
+    let client_config = make_client_config(KeyType::Rsa2048, &provider);
+    let server_config = make_server_config(KeyType::Rsa2048, &provider);
+    let (mut client, mut server) = make_pair_for_configs(client_config, server_config);
+    assert_eq!(client.extended_master_secret(), None);
+    assert_eq!(server.extended_master_secret(), None);
+    do_handshake(&mut client, &mut server);
+    assert_eq!(client.extended_master_secret(), Some(true));
+    assert_eq!(server.extended_master_secret(), Some(true));
+
+    // TLS 1.3 has no Extended Master Secret extension (the property is intrinsic
+    // to its key schedule), so it is reported as `None`.
+    let provider = provider::DEFAULT_TLS13_PROVIDER;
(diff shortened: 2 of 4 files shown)


cpu (approved):
This seems pretty reasonable to me.

If you're looking to use this in downstream projects it would probably make sense to try and backport it to [rel-0.23](https://github.com/rustls/rustls/tree/rel-0.23), but since I think the shape of the code is fairly different there I would wait until you get a +1 from djc/ctz before starting that.

cpu:
> mike-marcacci added a commit to eidola-ai/eidola that referenced this pull request 

As a small side-note, mentioning PRs in commit messages on GitHub is really noisy and creates a bunch of linkages whenever you push. I'd recommend avoiding it if you don't mind.

djc:
Alternatively, should we fail the secret exporter for 1.2 connections that don't use EMS?

Or is the exported secret still useful/safe to use, just unsound as a channel binding?

ctz:
Consider using simply https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html#structfield.require_ems and https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.require_ems instead of checking after the fact?

We have some documentation advising why 3shake is not a problem for rustls (as one or both parties of a connection) --
https://docs.rs/rustls/latest/rustls/manual/_02_tls_vulnerabilities/index.html#3shake

mike-marcacci:
Thanks for all the thoughtful comments folks!

@cpu good recommendation to avoid PR links in commits; I'll respect that going forward

@djc that's a really good question. I'm not sufficiently familiar with the breadth of uses beyond channel binding to make a well informed call on this, although my impression is that EMS might not be a concern for all of them.

@ctz depending on the use case, `require_ems` may not be a fit. I'm working to [upstream channel binding into a project](https://github.com/tinfoilsh/cvmimage/pull/160) that I suspect needs to support TLS 1.2 without EMS, but with lesser guarantees. Exposing the per-connection state allows the best course of action to be chosen by the caller.

Your link RE 3shake in rustls is extremely helpful, and something I didn't realize. However, I think the EMS status is still relevant for channel binding, at least as far as compliance with RFC 9266 is concerned: without EMS, nothing in the exporter is actually bound to the session hash; with EMS, the master secret itself is bound to the session hash, and therefore so is the exporter. I'm admittedly a bit out of my depth evaluating the *practical* implications of this in potential adversarial models outside of 3shake.

For my direct use case, I will be requiring TLS 1.3, and [this](https://github.com/seanmonstar/reqwest/pull/3048) is perfectly adequate. However, to extend this back to TLS 1.2, I'd want an EMS check to feel confident (absent a thorough analysis of the threat by someone more qualified than myself). I think it's perfectly defensible if you'd rather avoid exposing this as part of discouraging continued reliance on TLS 1.2 at all. If that's the case, feel free to close this PR: this is more of an enhancement than a blocker for me.

PS: This is the first time in my life I've commented to THREE folks with 3-letter usernames in the same message 😆

mike-marcacci:
@djc I got curious about your question and spent a bit diving in deeper.

Looking up other potential uses (none of which I have low-level experience with), EMS is currently considered a requirement for all, even if known vulnerabilities (like 3shake) are mitigated by other choices made in rustls. This includes PostgreSQL's SCRAM-SHA-256-PLUS, DTLS-SRTP (RFC 5764), Token Binding (RFC 8471), EAP-TLS.

I checked the behavior in golang, and [they completely restrict access to key-material exports for non-EMS, non-1.3 connections](https://github.com/golang/go/blob/e95518b5404abb9241836f2e95338151766bb79e/src/crypto/tls/prf.go#L247-L252).

So this would be a more opinionated change, but not without good rationale and not without the company of other prominent implementations. I'll let you maintainers make the call between approaches: let me know if you'd rather see a PR with that approach and I can put it together.

cpu:
> I checked the behavior in golang, and [they completely restrict access to key-material exports for non-EMS, non-1.3 connections](https://github.com/golang/go/blob/e95518b5404abb9241836f2e95338151766bb79e/src/crypto/tls/prf.go#L247-L252).

At least in the Go case, I think that's merited because unlike Rustls there's opt-in renegotiation support for clients ([`RenegotiationSupport`](https://pkg.go.dev/crypto/tls#RenegotiationSupport)). As further comparison, Go doesn't expose whether EMS was negotiated or not. 

I don't feel strongly, but it seems less obvious to me that Rustls should fail export for TLS 1.2 w/o EMS when there's no renegotiation risk.

> depending on the use case, require_ems may not be a fit. I'm working to https://github.com/tinfoilsh/cvmimage/pull/160 that I suspect needs to support TLS 1.2 without EMS, but with lesser guarantees. Exposing the per-connection state allows the best course of action to be chosen by the caller.

I feel like the path forward hinges on how convincing this aspect is. You're linking a Go project (`cvmimage`) so I'm not sure I understand fully. Where's the Rust code that needs the flexibility described? Is it just the `reqwest` work in https://github.com/seanmonstar/reqwest/pull/3048 ? Perhaps we should check with @seanmonstar that there's even downstream interest in 3048 before making any changes here?

mike-marcacci:
Ah yes, so I can disambiguate my path here a bit better:
1. I'm specifically working on [Eidola](https://github.com/eidola-ai/eidola) (fully rust), which I'm currently running on [Tinfoil Containers](https://tinfoil.sh/containers), which use an evolving attestation scheme with both hardware attestation (SEV-SNP, TDX) and WebPKI.
2. For their next version, I'm proposing the addition of channel binding with that `cvmimage` PR, and I've made [a corresponding change on my end](https://github.com/eidola-ai/eidola/pull/146).
3. For my change to work, I needed [`reqwest` to expose channel binding keying material](https://github.com/seanmonstar/reqwest/pull/3048).
4. While working on that, I realized that in general it wasn't possible to determine the soundness of a TLS 1.2 exporter, which seemed like quite a foot-gun. (I limited that value to TLS 1.3 connections in the `reqwest` PR, but it could be safely expanded to 1.2 if this addition landed in `rustls`.)

For my *specific* use-case, all I need is the `reqwest` change, since I can happily limit connections to TLS 1.3.

So this is really a speculative change for other clients that would want to support TLS 1.2 while confirming the soundness of channel binding. It seemed like an easy gap to fix, but this might not clear the "need threshold" for a change, and that's totally fine with me too.

ctz (approved):

rustls/src/common_state.rs
```diff
@@ -150,6 +151,16 @@ impl ConnectionOutputs {
         self.negotiated_version
     }
 
+    /// Whether the Extended Master Secret extension was negotiated.
+    ///
+    /// Returns `None` until the handshake reaches the point where this is known,
+    /// and `None` for TLS 1.3, where the extension does not apply. For TLS 1.2
+    /// and earlier, this returns `Some(true)` if the extension was negotiated
+    /// and `Some(false)` otherwise.
```
djc: Nit: maybe structure this as a bullet list to make it easier to understand? Also, suggest to get rid of the "and earlier" in "1.2 and earlier", since we don't support earlier (and are unlikely to ever support it).

djc (approved):

cpu (approved):

djc:
Please rebase and squash.

rustls/src/common_state.rs
mike-marcacci: Fixed

PR #3084, opened 2026-06-08
backport root_hint_subjects signature from rustls 0.24

stormshield-gt:


djc:
I'm not sure this is semver-compatible enough...

djc:
(One potential alternative way forward would be if you want to start porting Quinn to a rustls 0.24 git commit or alpha release...)

djc:
Oh, actually it looks like that work was already done in

- https://github.com/quinn-rs/quinn/pull/2671

stormshield-gt:
Sorry for the noise, I meant to make this PR to my fork, as indeed it's not backward compatible.

I think having a light fork of rustls 0.23 might be cheaper to maintain for us, as 0.24 has no stability guaranty for now (and that's totally fine)

Issue #3087, opened 2026-06-09
crates.io links to roadmap and benchmarks lead to a 404


Erik-Sovereign:
**Checklist**
* [x] I've searched the issue tracker for similar bugs.

**Describe the bug**
If you go to https://crates.io/crates/rustls and click on any link, for example on roadmap it will lead to https://github.com/rustls/rustls/blob/HEAD/rustls/ROADMAP.md, while the correct URL would be https://github.com/rustls/rustls/blob/main/ROADMAP.md.

**To Reproduce**
See description.

**Applicable Version(s)**
I assume we are only interested in the live version of crates.io.

**Expected behavior**
The links should lead to the correct pages so that `rustls` has a nicer public image.

**Additional context**
I visit the website with the Brave browser.

ctz:
I believe the fictious URL is a cargo bug ( https://github.com/rust-lang/crates.io/issues/9886 & https://github.com/rust-lang/crates.io/issues/9927 ) where cargo destroys information during publishing a crate that is needed for crates.io to produce the correct link.

PR #3088, opened 2026-06-09
Fix internal rustls links to make them compatible with crates.io

Erik-Sovereign:
For the underlying issue look here: https://github.com/rustls/rustls/issues/3087.


ctz (approved):
Thanks!

cpu (approved):
Maybe also worth doing against `rel-0.23`? I assume the `crates.io` copy won't update until a release is published and I'd wager we might do that for 0.23.x before cutting 0.24 from `main`?

Erik-Sovereign:
Am I right in believing only you can now merge that PR? Because I see no such button :D

PR #3095, opened 2026-06-18
Fail on empty handshake/alert messages

djc:
Can add a test if this passes existing tests.


ctz (approved):
Yup, a test would be good for this!

cpu (approved):

Issue #1343, opened 2023-07-05
v0.21.3 broken on 32bit platforms


cpu:
The update to webpki 0.101.0 that was bundled into the recently published 0.21.3 release introduced a 32bit platform support regression (reported in https://github.com/rustls/webpki/issues/112).

Filing this issue so folks that bump into the problem can find a marker in this repository that explains we're aware + working on it.

cpu:
This should now be resolved with the publication of webpki v0.101.1

My apologies for the regression.

PR #3105, opened 2026-06-23
Demote cross tests from CI to a nightly job

ctz:
These tests are both quite expensive, but also have in the past not really been very useful in surfacing issues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -341,45 +341,6 @@ jobs:
           # `cargo minimal-versions` uses unstable cargo features
           RUSTC_BOOTSTRAP: 1
 
-  cross:
-    name: cross-target testing
-    runs-on: ubuntu-latest
-    if: github.event_name != 'merge_group'
-    strategy:
-      matrix:
-        target:
-          # 32-bit Android (Linux) targets:
-          - armv7-linux-androideabi
-          - i686-linux-android
-          - thumbv7neon-linux-androideabi
-          # Other standard 32-bit (Linux) targets
-          - i586-unknown-linux-gnu
-          - i686-unknown-linux-gnu
-          # exotic Linux targets:
-          - riscv64gc-unknown-linux-gnu
-          - s390x-unknown-linux-gnu
-          # additional target(s):
-          # NOTE: This could have some overlap with 64-bit ARM-style CPU on macOS CI host;
-          # may have similar ARM-style CPU overlap with standard Windows & possibly Linux in the future
-          - aarch64-linux-android
-    steps:
-      - name: Checkout sources
-        uses: actions/checkout@v7
-        with:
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Should we just completely ditch them instead?

cpu (approved):
This seems reasonable, but the nightly CI notification story still sucks for if this does break.

Since we [have a step in RELEASING.md](https://github.com/rustls/rustls/blob/main/RELEASING.md) about manually invoking the nightly CI before cutting a release this seems OK overall. I mainly want to avoid a repeat of something like https://github.com/rustls/rustls/issues/1343 (though that specific instance was via `webpki`).

cpu:
> Should we just completely ditch them instead?

I would be in favor of keeping the i686 cross-compile. I somewhat frequently see Fedora + other downstream distros flagging broken crates/tests for that arch and since we've done that ourselves in the past I think there's evidence it has value.

djc:
> I would be in favor of keeping the i686 cross-compile. I somewhat frequently see Fedora + other downstream distros flagging broken crates/tests for that arch and since we've done that ourselves in the past I think there's evidence it has value.

Could we just do cargo check without full cross-compilation, instead just using --target for some 32-bit target?

It's not like we have substantial architecture-specific code in rustls itself.

ctz:
> > Should we just completely ditch them instead?
> 
> I would be in favor of keeping the i686 cross-compile. I somewhat frequently see Fedora + other downstream distros flagging broken crates/tests for that arch and since we've done that ourselves in the past I think there's evidence it has value.

I've readded this, though as a trivial cross build of the core crate. That is very fast (6s) though less thorough -- no tests, no checks on the provider crates (as they both have significant non-rust dependencies).

djc:
> > > Should we just completely ditch them instead?
> 
> > 
> 
> > I would be in favor of keeping the i686 cross-compile. I somewhat frequently see Fedora + other downstream distros flagging broken crates/tests for that arch and since we've done that ourselves in the past I think there's evidence it has value.
> 
> 
> 
> I've readded this, though as a trivial cross build of the core crate. That is very fast (6s) though less thorough -- no tests, no checks on the provider crates (as they both have significant non-rust dependencies).

Do you think that's enough defense that we can drop the scheduled extensive cross tests? Feels like the provider crypto is not really our system to test.

ctz:
> Do you think that's enough defense that we can drop the scheduled extensive cross tests? Feels like the provider crypto is not really our system to test.

I think there's some slight but marginal value to recording how to cross-build this stuff somewhere. And GHA seems preferable to documentation.

But, if it becomes very painful or regularly breaks for reasons beyond our control, I think it's a good candidate to stop doing. As you say, it's not really ours to test.

PR #3106, opened 2026-06-23
Retire badssl.com tests

ctz:
The upstream server(s) are just unmaintained, and nowadays these tests
aren't doing much but checking we have nice error messages in common
broken-server cases.


cpu (approved):
Seems reasonable. Do you want to squash the commit history to avoid the tweak+revert commits?

Also there are a couple `badssl.com` examples [in the README.md](https://github.com/rustls/rustls#client-example-program) we might want to remove since they won't produce the described output in the current state right?

ctz:
> Seems reasonable. Do you want to squash the commit history to avoid the tweak+revert commits?

Oops, didn't mean to push that!

cpu (approved):

Issue #3109, opened 2026-06-25
Don't construct `EncodedMessage` in `MessageEncrypter`, `MessageDecrypter` impls


tgeoghegan:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
Each implementation of trait `MessageEncrypter` in a crypto provider implements the function:

```rust
    /// Encrypt the given TLS message `msg`, using the sequence number
    /// `seq` which can be used to derive a unique [`Nonce`].
    fn encrypt(
        &mut self,
        msg: EncodedMessage<OutboundPlain<'_>>,
        seq: u64,
    ) -> Result<EncodedMessage<OutboundOpaque>, Error>;
```

As a result, each crypto provider must make a decision about what content type and protocol version to put in the returned `EncodedMessage<OutboundOpaque>`. Some of them copy the content type and protocol version from the input `msg`, but some of them force those values to `ContentType::ApplicationData` and `ProtocolVersion::TLSv1_2`, respectively.

The correct content type or protocol version to emit has nothing to do with which crypto implementation is in use, so this interface seems to unnecessarily introduce an opportunity for providers to make mistakes.

Symmetrically, we have trait `MessageDecrypter`:

```rust
    fn decrypt<'a>(
        &mut self,
        msg: EncodedMessage<InboundOpaque<'a>>,
        seq: u64,
    ) -> Result<EncodedMessage<&'a [u8]>, Error>;
```

TLS 1.2 implementations ignore everything in `msg` except the payload, while TLS 1.3 implementations have to remember to use `EncodedMessage::into_tls13_unpadded_message`.

**Describe the solution you'd like**
If the interface were changed to take bytes of plaintext instead of `EncodedMessage<OutboundPlain>` and return bytes of ciphertext instead of `EncodedMessage<OutboundOpaque>`, then the calling context (`crypto::cipher::record_layer::EncryptionState::encrypt_outgoing`) could take responsibility for correctly constructing records no matter what provider is in use. So perhaps change the trait method to:

```rust
fn encrypt(&mut self, plaintext: &[u8], seq: u64) -> Result<Vec<u8>, Error>;
```

Or if we consider the proposition in #3068 and the broad push toward client owned buffers, then we could do decryption in-place and avoid allocating:

```rust
[...]

PR #3110, opened 2026-06-25
WIP unmergeable DTLS implementation

tgeoghegan:
Per discussion in #40, I'm posting this draft PR so that we can start discussing these changes.

Summary of the significant changes here and the motivations:

#### `OpaqueOutbound` now has field `EncodingContext`

Why? We need various pieces of information to make correct decisions about how to encode messages:
- The version field should _sometimes_ reflect the actual protocol in use (particularly when working with test vectors) but _sometimes_ needs to claim (D)TLS 1.2 for backward compatibility, except that when the message is the initial ClientHello you need to claim (D)TLS 1.0.
- Whether the payload is encrypted informs whether a DTLS 1.3 unified header or a DTLS 1.2-style full header is used
- For DTLS, we need to know the epoch and sequence of the record.

The idea is that `EncodedMessage` structures contain the _actual_ protocol in use, and `EncodingContext` contains enough information for `EncodedMessage<OpaqueOutbound>::encode` to make the right decisions about what bytes to put on the wire.

I opted to put DTLS epoch and sequence into EncodingContext instead of a field on `EncodedMessage` because it winds up "contaminating" less of the codebase with DTLS epoch and sequence, which really is only something that the send and receive paths need to consider.

#### trait `MessageEncrypter`

Because `MessageEncrypter::encrypt` implementations are responsible for constructing `EncodedMessage<OpaqueOutbound>` values, there are implications here for crypto providers. 

`EncodingContext` gets passed through trait `MessageEncrypter` methods, and  plaintext message version is preserved by `MessageEncrypter::encrypt` implementations instead of occasionally forcing TLS 1.2.

In passing, I notice that every provider has to make decisions about what protocol version to write into outbound messages, which just seems like an opportunity for providers to get it wrong. I wrote up #3109 to discuss changes to the crypto traits, which are orthogonal to implementing DTLS.

#### Message deframing+coalescing and message fragmentation

- Receive path invokes special logic for coalescing DTLS messages
- Send path conspires with encryption state to keep track of DTLS epoch and sequence in `EncryptState::write_seq`, which is interpreted to be 16 bits of epoch and 48 bits of sequence (this still needs work).
- `msgs::deframer::Deframer` learns how to deframe DTLS messages and then coalesce them (reorder and reassemble DTLS handshake fragments).
    - In keeping with rustls' philosophy of letting the linker discard unused code, we should refactor this into `Deframer` and `DtlsDeframer`.
- `msgs::fragmenter::MessageFragmenter` learns how to fragment DTLS handshake messages into multipl records.
    - Right now we plumb a "max fragment size" parameter all the way from `client::ClientConfig` or `server::ServerConfig` into the fragmenter and use that as the maximum size of a datagram. Path MTU discovery in practice is notoriously complex (you do it by sending smaller and smaller messages until one gets through) so we will have to decide if this happens inside or outside rustls and revisit this interface.
    - Right now, we have both `fragment_dtls_handshake_message` (takes a single `handshake_payload: &[u8]`) and `fragment_dtls_handshake_message_flight` (takes a flight of message as a slice `[(HandshakeType, Vec<u8>)]`). The former should be deleted: fragmenting a single handshake message can be realized by passing a slice of one element to `fragment_dtls_handshake_message_flight`.

#### Miscellaneous

- The state machines need more knowledge of what transport (e.g. TCP vs. QUIC vs. UDP) is in use or what protocol version eventually gets negotiated so they can cosntruct messages appropriately. This means `protocol` fields are added to various structs and other times a `struct HandshakeState` gets passed into a function instead of `struct HashshakeHash`.
- `HandshakeFlight` changes: we can't just concatenate encoded messages into one big `Vec<u8>`, because the DTLS fragmenter needs to know the handshake type of each message in the flight. So now we store a `Vec<(HandshakeType, Vec<u8>)>`.
- `mod datagram` is a stab at an API for wrapping DTLS around something that looks like a UDP socket. As discussed in the DTLS issue, this will change a lot as we refine the design, but for now it's useful for high level end to end testing of DTLS.

#### Not done yet

- Transcript hashing is broken because the two sides have different views of wire messages: it's either that one side is hashing messages with epoch and sequence fields and the other isn't, or that one side is hashing messages with the actual `ProtocolVersion` in use but the other is hashing the backward compatibility claim.
- ECH
- 0-RTT
- Replay detection and rejection
- Handshake retransmission state machine
- Epoch handling
- Record number (de)protection
- Path MTU discovery, negotiation
- RFC 9146 Connection ID handling
- HelloRetry cookie exchange


rustls/src/msgs/fragmenter.rs
tgeoghegan, line 1: Besides removing `fragment_dtls_handshake_message`, I would refactor `fragment_dtls_handshake_message_flight` onto a new `DtlsMessageFragmenter` struct, which I think would better separate concerns and make it easier for the linker to not include DTLS specific code.

rustls/src/msgs/deframer/mod.rs
tgeoghegan, line 1: All the DTLS-specific methods here could be refactored into a distinct `DtlsDeframer` structure.

rustls/src/conn/send.rs
```diff
@@ -278,41 +348,84 @@ impl SendOutput for SendPath {
...
-            has_sent_fatal_alert: false,
-            has_sent_close_notify: false,
-            message_fragmenter: MessageFragmenter::default(),
-            sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
-            queued_key_update_message: None,
-            refresh_traffic_keys_pending: false,
-            negotiated_version: None,
-            tls13_key_schedule: None,
+    fn send_msg(&mut self, m: Message<'_>, must_encrypt: bool, is_retry_req: bool) {
+        match (self.protocol, &m.payload) {
```
tgeoghegan, line 402: This `match` is ugly, but I think we should be able to collapse the `(Protocol::Udp, MessagePayload::Handshake)` and `(Protocol::Udp, MessagePayload::HandshakeFlight)` cases together. The noisy `Iterator::map` calls can also go away if we revisit the return type of `fragment_dtls_handshake_message_flight` so that it can be passed right into `self.send_messages`.

rustls/src/conn/receive.rs
```diff
@@ -258,6 +283,32 @@ impl ReceivePath {
                 return Err(PeerMisbehaved::MessageInterleavedWithHandshakeMessage.into());
             }
 
+            if let Some(EpochAndSequence {
```
tgeoghegan: One important question is where to do the DTLS retransmission state machine (on the sender side) and where to do anti-replay and epoch handling (on the receive side). This incomplete stab at anti-replay is a hint to my answer: `conn::receive::ReceivePath` and `conn::send::SendPath`. But perhaps we want to introduce a further layer in between the send and receive paths and the TLS handshake state machine where these DTLS specific concerns happen?

PR #3112, opened 2026-06-25
Introduce `EncodableVersion`

tgeoghegan:
rustls needs various pieces of information to make correct decisions about how to encode records:

- the version field should _sometimes_ reflect the actual protocol in use (particularly when working with test vectors) but otherwise needs to claim to be TLS 1.2 for backward compatibility...
- ...except when the message is the initial ClientHello, in which case clients MAY claim TLS 1.0 (and rustls chooses to).

Once DTLS arrives, there's more pertinent contextual information:

- If DTLS 1.3 is negotiated and the record is encrypted, then a unified header should be used, otherwise a full DTLS 1.2 style header is used
- We need to know the epoch and sequence numbers of the record.

Up until now, rustls has been making decisions like lying about protocol version all the way up in the handshake state machine, but this approach becomes untenable as the encoding decisions become more complex.

In this PR, structures `Message` and `EncodedMessage` now have a field `version: EncodableVersion`, which is an enum that captures both the actual protocol version in use (when it's known) and enough contextual information to make the right decisions about what bytes to put on the wire.

`Message` and `EncodedMessage` structures are now expected to contain the actual protocol version in use in their version field, which means that field can now be used to make decisions about encoding or other behavior instead of indicating TLS 1.2 unconditionally.

The exception is `ClientHello` messages, because at that stage of the handshake, no protocol version is in use yet at all, but we have to put something in the structure.

This plumbing will become particularly valuable as DTLS 1.2 and 1.3 get implemented.

There are implications here for the crypto provider implementations, which now are expected to preserve whatever version was indicated in an `EncodedMessage<OutboundPlain>` in the emitted
`EncodedMessage<&[u8]>`. See #3109 for a potential further refinement to that interface.

For some context, see #40, #3110


cpu:
FWIW it looks like CI has a handful of true-positive failures to sort out for this branch.

tgeoghegan:
Sorry for sending obviously broken code. I've validated locally that these tests pass with the latest few commits.

rustls-ring/src/tls13.rs
```diff
@@ -214,9 +216,7 @@ impl MessageEncrypter for Tls13MessageEncrypter {
 
         Ok(EncodedMessage {
             typ: ContentType::ApplicationData,
-            // Note: all TLS 1.3 application data records use TLSv1_2 (0x0303) as the legacy record
-            // protocol version, see https://www.rfc-editor.org/rfc/rfc8446#section-5.1
-            version: ProtocolVersion::TLSv1_2,
```
djc: Suggest maybe plumbing through the `EncodingContext` should be in a separate commit from making these changes that actually use it.

rustls-ring/src/tls13.rs
```diff
@@ -197,11 +198,12 @@ struct Tls13MessageDecrypter {
 impl MessageEncrypter for Tls13MessageEncrypter {
     fn encrypt(
         &mut self,
+        encoding_context: EncodingContext,
```
djc: This should be the last argument. I'd name it `cx` (for local variables at least) since there are no other relevant contexts in play (and the type disambiguates enough).

rustls/src/common_state.rs
```diff
@@ -258,7 +258,7 @@ pub(crate) trait Output<'m> {
 
     fn output(&mut self, ev: OutputEvent<'_>);
 
-    fn send_msg(&mut self, m: Message<'_>, must_encrypt: bool);
+    fn send_msg(&mut self, m: Message<'_>, must_encrypt: bool, is_retry_req: bool);
```
djc: Adjacent positional parameters of the same type are a bit of a hazard, suggest rethinking this.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -13,7 +13,8 @@ use crate::msgs::{Codec, HEADER_SIZE, MAX_FRAGMENT_LEN, Reader, hex, read_opaque
 pub struct EncodedMessage<P> {
     /// The content type of this message.
     pub typ: ContentType,
-    /// The protocol version of this message.
+    /// The protocol version of this message. The actual protocol version that
```
djc: Nit: for all docstrings, keep the first line as a single sentence, then add more details in further paragraphs.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -13,7 +13,8 @@ use crate::msgs::{Codec, HEADER_SIZE, MAX_FRAGMENT_LEN, Reader, hex, read_opaque
...
+    /// gts encoded on the wire may differ.
```
djc: "gts"

rustls/src/crypto/cipher/messages.rs
```diff
@@ -299,68 +330,74 @@ impl<'a> From<&'a [u8]> for OutboundPlain<'a> {
...
 /// It results from encryption and is used for io write.
 #[derive(Clone, Debug)]
-pub struct OutboundOpaque(Vec<u8>);
+pub struct OutboundOpaque {
+    /// Amount of space reserved for the record header.
+    header_size: usize,
+    /// Encoded payload of the record.
+    payload: Vec<u8>,
+    /// Contextual information needed to encode the record.
+    encoding_context: EncodingContext,
```
djc: Changing this from a tuple struct to a named field struct should be in a separate commit.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -472,6 +509,42 @@ impl DerefMut for InboundOpaque<'_> {
     }
 }
 
+/// Contextual information for encoding messages.
+#[derive(Clone, Copy, Debug, Default)]
+#[non_exhaustive]
+pub struct EncodingContext {
+    /// Whether this message is the initial handshake message, e.g., not a
```
djc: Nit: these docstrings need tweaking (see elsewhere).

rustls/src/crypto/cipher/messages.rs
```diff
@@ -472,6 +509,42 @@ impl DerefMut for InboundOpaque<'_> {
...
+    pub is_initial_handshake: bool,
+    /// Whether to preserve the version in the EncodedMessage instead of
+    /// downgrading for compatibility. This will only be used in tests loading
+    /// messages from test vectors.
+    pub preserve_version: bool,
+}
+
+impl EncodingContext {
+    /// Set this context's `is_initial_handshake` value.
+    pub fn is_initial_handshake(mut self, is: bool) -> Self {
```
djc: I think it is unidiomatic to have a setter/builder method name start with `is_`.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -472,6 +509,42 @@ impl DerefMut for InboundOpaque<'_> {
...
+    }
+
+    /// The size of the record layer header for this message.
+    // version parameter is not yet used, but it will matter when we need to
+    // distinguish between TLS and DTLS 1.2 and DTLS 1.3, so we wire it up now
+    // <https://github.com/rustls/rustls/issues/40>
+    fn header_size(&self, _version: ProtocolVersion) -> usize {
+        // TLS 1.2 and 1.3 messages use a full record header
+        HEADER_SIZE
+    }
```
djc, line 599: I'd prefer to wire it up later if it unused now/here.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -164,9 +176,28 @@ impl EncodedMessage<OutboundOpaque> {
...
+        ) {
+            (true, _) => self.version,
+            // <https://datatracker.ietf.org/doc/html/rfc8446#section-5.1>:
+            // "This value MUST be set to 0x0303 for all records generated
+            //  by a TLS 1.3 implementation ..."
+            (false, false) => ProtocolVersion::TLSv1_2,
+            // "... other than an initial ClientHello (i.e., one not
+            // generated after a HelloRetryRequest), where it MAY also be
+            // 0x0301 for compatibility purposes"
+            (false, true) => ProtocolVersion::TLSv1_0,
```
djc: It feels like potentially a three-way enum makes more sense for this than two `bool`s?

djc (commented):
Here's an initial round of feedback.

rustls-ring/src/tls13.rs
tgeoghegan: Done (as part of a commit rebased since your last review). I chose to leave the struct field `OutboundOpaque::encoding_context` in the more verbose form because there I think the verbosity helps, since field accesses don't always have the type name right there. But I'm happy to change it to `OutboundOpaque::cx` if you prefer.

rustls/src/crypto/cipher/messages.rs
tgeoghegan: Fixed eda3f9a95ce9af756de12469b5327bdecaba5b2f

rustls/src/crypto/cipher/messages.rs
tgeoghegan: Fixed eda3f9a95ce9af756de12469b5327bdecaba5b2f

rustls/src/crypto/cipher/messages.rs
tgeoghegan: I removed the unused parameter and the comment referencing #40.

rustls/src/crypto/cipher/messages.rs
tgeoghegan: I rebased so that 3b7cced1fdc7f16677f800d90344b649d1a79d9c introduces `struct EncodingContext` and makes this change to `OutboundOpaque`.

rustls-ring/src/tls13.rs
tgeoghegan: I rebased so that 4ed83af90b6a1c55480cbb49793485a1ef1e8946 plumbs `EncodingContext` everywhere it needs to go (thus far), separately from 3b7cced1fdc7f16677f800d90344b649d1a79d9c which defines the struct and c5833fb5973f98639eb3a3c7fdfe63e593a31758 which uses it to make decisions.

tgeoghegan (commented):
I rebased according to djc's review comments, so that we have one commit that introduces the new `struct EncodingContext` and changes `struct OutboundOpaque`, then another that plumbs `EncodingContext` where it needs to go but without using it for decisions yet, and a final commit that makes the actual logic changes.

I still need to address djc's other comments about avoiding the hazard of adjacent boolean parameters to a function as well as putting an enum inside `EncodingContext`.

djc:
> I still need to address djc's other comments about avoiding the hazard of adjacent boolean parameters to a function as well as putting an enum inside `EncodingContext`.

Flipped this to draft until that part is done. 👍

tgeoghegan:
There's new test failures here because I had to rebase onto `main` to get `src/conn/split.rs`.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -472,6 +488,54 @@ impl DerefMut for InboundOpaque<'_> {
...
+    fn header_size(&self) -> usize {
+        // TLS 1.2 and 1.3 messages use a full record header
+        HEADER_SIZE
+    }
+}
+
+/// How to encode a [`ProtocolVersion`] when encoding an [`EncodedMessage`].
+#[derive(Clone, Copy, Debug, Default)]
+#[non_exhaustive]
+pub enum VersionEncoding {
```
tgeoghegan: a53611fadb6bc5506fa82fd9fe646576d469200b addresses @djc's remaining two pieces of feedback here, which are (1) using a three-valued enum in `EncodedMessage<OutboundOpaque>::encode` is nicer than matching on a pair of booleans and (2) two adjacent `bool` arguments in `common_state::Output::send_msg` is a usability hazard.
    
These problems are addressed by introducing `enum VersionEncoding`, a value of which is (for now) the sole field in `struct EncodingContext`.

You might ask why we bother having an enum inside of `struct EncodingContext` instead of just having the enum: the reason is that in the near future, we'll add fields that are orthogonal to the `VersionEncoding` such as whether the message is encrypted (needed so we can decide what kind of header to use in DTLS 1.3) and the epoch and sequence numbers.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -472,6 +488,54 @@ impl DerefMut for InboundOpaque<'_> {
...
+    /// Preserve whatever version is in the enclosing `EncodedMessage`.
+    ///
+    /// This is intended for use in tests, such as ones that load encoded
+    /// messages from test vectors and want them preserved byte-for-byte.
+    Preserve,
```
tgeoghegan: Since this is only intended for use in tests, it would be nice to tag it `#[cfg(test)]` so the compiler can help ensure it's never used in production configurations. But unfortunately it's needed in `msgs::fuzzing::fuzz_message`, which is used in `rustls-fuzz`, which is not built in a test configuration.

We could introduce a whole new feature to `rustls` for this, but I don't think it's worth it. I'm open to suggestions for how to restrict utilization of this item.

rustls/src/crypto/cipher/messages.rs
tgeoghegan: We now match on a single enum value. See https://github.com/rustls/rustls/pull/3112#discussion_r3538517360

rustls/src/crypto/cipher/messages.rs
tgeoghegan: I changed this to `fn with_version_encoding(mut self, ve: VersionEncoding) -> Self` which I think is more idiomatic.

rustls/src/common_state.rs
tgeoghegan: We now take an `enum VersionEncoding` value here, so that callers can be more explicit about what they want and to avoid the risk of mixing up bools. See https://github.com/rustls/rustls/pull/3112#discussion_r3538517360

rustls/src/crypto/cipher/messages.rs
djc: I'd really prefer to avoid adding `encoding_context` in this commit, postponing that to a later commit, making this commit very strictly about renaming the inner field to `payload`.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -355,20 +362,6 @@ impl<'a> Extend<&'a u8> for OutboundOpaque {
...
-        value.payload.extend(content);
-        value
-    }
-}
-
-impl<const N: usize> From<&[u8; N]> for OutboundOpaque {
-    fn from(content: &[u8; N]) -> Self {
-        Self::from(&content[..])
-    }
-}
```
djc: I think replacing these impls with a named constructor could also be a separate commit. (And it might be worth splitting some of these "boring" refactoring commits into a separate PR, so that we can merge them early and avoid painful rebases down the line.)

rustls-aws-lc-rs/src/tls12.rs
```diff
@@ -3,9 +3,9 @@ use alloc::boxed::Box;
 use aws_lc_rs::{aead, tls_prf};
 use pki_types::FipsStatus;
 use rustls::crypto::cipher::{
-    AeadKey, EncodedMessage, InboundOpaque, Iv, KeyBlockShape, MessageDecrypter, MessageEncrypter,
-    NONCE_LEN, Nonce, OutboundOpaque, OutboundPlain, Tls12AeadAlgorithm, UnsupportedOperationError,
-    make_tls12_aad,
+    AeadKey, EncodedMessage, EncodingContext, InboundOpaque, Iv, KeyBlockShape, MessageDecrypter,
```
djc: For plumbing `EncodingContext` everywhere it needs to go,  the commit message mentions:

> The struct still has no fields, and we still don't consult it for any
decisions except for header size.

If there's no data, how can we consult it for header size?

rustls-test/src/lib.rs
```diff
@@ -1369,6 +1369,7 @@ impl RawTls {
 
     pub fn encrypt_and_send(
         &mut self,
+        is_initial_handshake: bool,
```
djc: So I like to order arguments from most specific/special/interesting to least. I think by that measure, this should probably be last (and please be mindful of this for other argument additions).

rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -10,7 +10,7 @@ use rustls::crypto::cipher::{
 };
 use rustls::crypto::tls13::{Hkdf, HkdfExpander, OkmBlock, OutputLengthError};
 use rustls::crypto::{self, CipherSuite};
-use rustls::enums::{ContentType, ProtocolVersion};
+use rustls::enums::ContentType;
```
djc, line 12: "clean up rebase" should be squashed.

rustls/src/client/tls13.rs
```diff
@@ -470,7 +470,7 @@ pub(super) fn emit_fake_ccs(sent_tls13_fake_ccs: &mut bool, output: &mut dyn Out
 
     output.send_msg(
         Message {
-            version: ProtocolVersion::TLSv1_2,
+            version: ProtocolVersion::TLSv1_3,
```
djc: IMO this should be in the same commit that adds logic for encoding them correctly. The goal here is that each commit could pass CI (even if you don't always have to check that, because it would be too slow).

rustls/src/client/tls12.rs
```diff
@@ -481,7 +481,7 @@ fn emit_certificate(
     };
 
     transcript.add_message(&cert);
-    output.send_msg(cert, false, false);
+    output.send_msg(cert, false, VersionEncoding::UseBackwardCompatible);
```
djc: I don't think we want the intermediate state in this PR, so this should be squashed with the previous commit that introduced these?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -524,6 +510,32 @@ impl EncodingContext {
...
+    Use1_0,
+    /// Use the backward compatible protocol version.
+    ///
+    /// For example, if the `version` value of the enclosing [`EncodedMessage`]
+    /// is `ProtocolVersion::TLSv1_3`, `ProtocolVersion::TLSv1_2` is used
+    /// instead ([1]).
+    ///
+    /// [1]: https://datatracker.ietf.org/doc/html/rfc8446#section-5.1
+    #[default]
+    UseBackwardCompatible,
```
djc: Bikeshedding these names a bit, maybe more like `Tests`, `ClientHello`, `Compatible` to (a) tie them more clearly to the why rather than the how and (b) make the last one more concise.

rustls-aws-lc-rs/src/tls12.rs
tgeoghegan: As of this PR, rustls still only does TLS 1.2 and 1.3, and thus the header size is always just `msgs::HEADER_SIZE`. In the not-too-distant future, though, the header size will vary, because encrypted DTLS 1.3 records use a variable-length unified header. But I acknowledge that for now, the usage of `EncodingContext::header_size` is pointless and can be deferred to a later PR.

rustls-test/src/lib.rs
tgeoghegan: Noted, I will do my best to be more careful about this.

tgeoghegan:
Thank you for the detailed and patient review, Dirkjan. Going forward I'll be careful about making the PR mergeable before requesting review, in the sense that commits are rebased and squashed so that they look like what we'd want to see on `main`. Apologies for this PR having been unnecessarily difficult to review so far.

Based on your feedback, my plan is to make a separate PR with the "boring refactor stuff". It will consist of several commits that will:

- make `OpaqueOutbound` into a struct with fields

If/when that lands, I'll rebase this PR on top of those changes, and it will be about:

- introduce `struct EncodingContext` with fields `enum VersionEncoding`
- add an `EncodingContext` field to `OpaqueOutbound` and then add `cx: EncodingContext` parameters to methods where needed
- using `EncodingContext` for encoding decisions in `EncodedMessage<OutboundOpaque>::encode`
- changing sites where `EncodedMessage`s are constructed to indicate the real protocol version in use

In the meantime, I'll downgrade this PR to a draft, as it won't be ready for re-review until the rebase I described happens.

(plan updated in light of djc's comment below)

djc:
I think introducing `EncodingContext` should remain in this PR (doesn't count as boring enough). By boring I meant naming the field and moving the impls to named methods.

rustls-test/src/lib.rs
```diff
@@ -1371,10 +1372,15 @@ impl RawTls {
         &mut self,
         msg: &EncodedMessage<Payload<'_>>,
         peer_input: &mut VecInput,
+        ve: VersionEncoding,
```
tgeoghegan: I put this argument last because it's the least specific value passed to this function, in that a single `VersionEncoding` value can be used across many different messages.

rustls/src/client/hs.rs
```diff
@@ -824,18 +824,7 @@ fn emit_client_hello_for_retry(
...
-            //  by a TLS 1.3 implementation ..."
-            Some(_) => ProtocolVersion::TLSv1_2,
-            // "... other than an initial ClientHello (i.e., one not
-            // generated after a HelloRetryRequest), where it MAY also be
-            // 0x0301 for compatibility purposes"
-            //
-            // (retryreq == None means we're in the "initial ClientHello" case)
-            None => ProtocolVersion::TLSv1_0,
-        },
+        version: ProtocolVersion::TLSv1_2,
```
tgeoghegan: As I discuss in 79bcc0920772abac9d551b7ab8d0f6b3bc9a3821, it isn't strictly correct to say that this message is `ProtocolVersion::TLSv1_2`. In some sense, no protocol version is yet in use, because the client hasn't yet seen what versions the server will offer. But we have to put something here, and it'll end up being 1.2 on the wire, so this seems appropriate.

An alternative would be to make the `version` field be `Option<ProtocolVersion>`, set `None` here and have the encoding layer emit `ProtocolVersion::TLSv1_2`, but it's not obvious to me that it's any better.

rustls/src/crypto/cipher/messages.rs
tgeoghegan: I agree with renaming to indicate purpose rather than behavior. I landed on `TestVectors`, `Compatible` and `InitialClientHello`.

`TestVectors` is a bit more specific than `Tests`, and in fact there's lots of tests where we do indicate one of the other behaviors. And I went with `InitialClientHello` because the MAY in [RFC 8446 section 5.1](https://datatracker.ietf.org/doc/html/rfc8446#section-5.1) applies only to the initial `ClientHello`, not retries.

rustls/src/crypto/cipher/mod.rs
```diff
@@ -163,6 +164,7 @@ pub trait MessageEncrypter: Send + Sync {
         &mut self,
         msg: EncodedMessage<OutboundPlain<'_>>,
         seq: u64,
+        cx: EncodingContext,
```
tgeoghegan: I placed this last as it's the least specific argument, in that a single `EncodingContext` value is reusable across many calls to `MessageEncrypter::encrypt`.

tgeoghegan:
I rebased into four commits, separating plumbing from logic.

- The first defines `struct EncodingContext` and `enum VersionEncoding`
- The second wires `EncodingContext` arguments where they need to go, but only sets the default value and never consults it
- The third wires `VersionEncoding` arguments where they need to go, but only sets the default value and never consults it
- The final commit starts using `EncodingContext` values to make encoding decisions. Meaningful values of `VersionEncoding` and `EncodingContext` are now used.

tgeoghegan:
Rebased to resolve some conflicts with recent changes on `main`.

djc:
Okay, I suggest one more round of commit history surgery: let's squash commits 1 and 2 to introduce `EncodingContext`, but keep it empty. Then squash commit 3 and 4 to introduce `VersionEncoding` and start using them to make decisions. If there no/few uses of `VersionEncoding::default()` left (and thus, maybe, `EncodingContext::default()`), maybe we should drop the `Default` impl(s)?

@ctz do you want to take a look at this?

tgeoghegan:
Rebased per your instructions.

> If there no/few uses of `VersionEncoding::default()` left (and thus, maybe, `EncodingContext::default()`), maybe we should drop the `Default` impl(s)?

I removed both `Default` impls, but we are left with `EncodingContext::new(ve: VersionEncoding) -> Self`. The tricky thing is that there's a place in `rustls-test` where we construct an `EncodingContext`, and since the struct is `non_exhaustive`, we have to construct it via an associated function. The current approach has the benefit of being explicit everywhere about the version encoding to use, though it's a bit more verbose than either `EncodingContext::default()` would be, or a hypothetical `EncodingContext::new()` that takes no parameters and sets `version_encoding: Compatible`.

And in passing, I should remind reviewers that if we _did_ opt to implement a parameter-less `::new()`, `clippy` would gripe about not having a corresponding `Default` impl.

tgeoghegan:
This needs a rebase, and in fact needs to be rethought quite a bit in light of #3127 removing `struct OutboundOpaque` altogether. I'm downgrading to draft while I iterate here.

djc:
> Rebased per your instructions.
> 
> > If there no/few uses of `VersionEncoding::default()` left (and thus, maybe, `EncodingContext::default()`), maybe we should drop the `Default` impl(s)?
> 
> I removed both `Default` impls, but we are left with `EncodingContext::new(ve: VersionEncoding) -> Self`. The tricky thing is that there's a place in `rustls-test` where we construct an `EncodingContext`, and since the struct is `non_exhaustive`, we have to construct it via an associated function. The current approach has the benefit of being explicit everywhere about the version encoding to use, though it's a bit more verbose than either `EncodingContext::default()` would be, or a hypothetical `EncodingContext::new()` that takes no parameters and sets `version_encoding: Compatible`.

That sounds like the right trade-off.

> And in passing, I should remind reviewers that if we _did_ opt to implement a parameter-less `::new()`, `clippy` would gripe about not having a corresponding `Default` impl.

Yes, I think we prefer `Default` over parameter-less `new()` (dropping the latter when we have the former).

> This needs a rebase, and in fact needs to be rethought quite a bit in light of #3127 removing `struct OutboundOpaque` altogether. I'm downgrading to draft while I iterate here.

Sorry for the bumpy ride! I think we're in a phase of the release cycle where some more large changes are coming down the pike.

tgeoghegan:
Rebased. This wasn't as tricky as I'd feared. The big difference is that instead of having an `EncodingContext` field on `OpaqueOutbound`, we just pass an `EncodingContext` to `encode_record_header`.

Other options would be to reintroduce `OpaqueOutbound` so that it could wrap the encrypted payload and the encoding context, or to stick the encoding context in `OutboundPlain`, but those options aren't clearly better to me.

> Sorry for the bumpy ride! I think we're in a phase of the release cycle where some more large changes are coming down the pike.

It's fine. I actually think this change is a little bit better now that it's rebased, because `EncodingContext` no longer needs to be threaded through the crypto providers.

@ctz I'd be grateful for review on this. I'm happy to answer any questions.

tgeoghegan:
"Generate pre-release documentation" appears to be failing due to timeouts and HTTP 502s connecting to rfc-editor.org.

djc:
> @ctz I'd be grateful for review on this. I'm happy to answer any questions.

So is this Ready for review (again) now?

tgeoghegan:
Oops, yes. I was waiting to get a reasonably clean set of test results before marking it ready.

On that topic: the codecov checks fail because overall coverage goes down by 0.08% and only 80.25% of the diff is exercised by tests. But it can be hard to hit such goals when many lines of code in a change are not executable (e.g. a line consisting of only `}`). Should I do something about those?

edit: actually, I think there's something fishy going on with these items in `server::test` and `client::test`, orthogonally to this change. I wrote up #3183 for this.

tgeoghegan:
With #3183 fixed, the codecov checks now pass. Thanks for that quick fix!

cpu:
That was a good catch, thank you.

rustls/src/client/hs.rs
```diff
@@ -874,7 +863,17 @@ fn emit_client_hello_for_retry(
...
+
+    output.send_msg(
+        ch,
+        false,
+        if retryreq.is_none() {
+            // Not a retry means this is the initial client hello
+            VersionEncoding::InitialClientHello
+        } else {
+            VersionEncoding::Compatible
+        },
```
djc: Nit: write this as a `match`?

```rust
match &retryreq {
     Some(_) => VersionEncoding::Compatible,
     None => VersionEncoding::InitialClientHello,
}
```

djc (approved):
LGTM, thanks!

rustls/src/crypto/cipher/messages.rs
```diff
@@ -534,10 +539,62 @@ impl DerefMut for InboundOpaque<'_> {
...
+    /// [1]: https://datatracker.ietf.org/doc/html/rfc8446#section-5.1
+    InitialClientHello,
+    /// Use the backward compatible protocol version.
+    ///
+    /// For example, if the `version` value of the enclosing [`EncodedMessage`] is
+    /// `ProtocolVersion::TLSv1_3`, `ProtocolVersion::TLSv1_2` is used instead ([1]).
+    ///
+    /// [1]: https://datatracker.ietf.org/doc/html/rfc8446#section-5.1
+    Compatible,
+}
```
ctz: I think I'm going to want to understand the utility of this going forward. Right now it seems like the encoded message contains one thing, and then the encoding step mostly ignores that -- it seems it would be better to record that in the types involved. Now we have per-direction types, why couldn't the outgoing's version field be something like:

```rust
enum EncodableVersion {
  Literal(ProtocolVersion),
  InitialClientHello(Protocol), // encodes as TLS1.0 or DTLS1.0
  Legacy(Protocol), // encodes as TLS1.2 or DTLS1.2
}
```

I had a look at #3110 to try and understand this myself and couldn't really identify this type there.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -534,10 +539,62 @@ impl DerefMut for InboundOpaque<'_> {
...
+/// How to encode a [`ProtocolVersion`] when encoding an [`EncodedMessage`].
+#[derive(Clone, Copy, Debug)]
+#[non_exhaustive]
+pub enum VersionEncoding {
+    /// Encode the message without changing anything.
+    ///
+    /// This preserves whatever version is in the enclosing `EncodedMessage`. This is intended for
+    /// use in tests that load encoded messages from test vectors and want them preserved
+    /// byte-for-byte.
+    TestVectors,
```
ctz: Notwithstanding my other comment, I would call this something else. If it genuinely is testing-only, then let's align the tests with what the library needs, rather than the other way around.

rustls/src/crypto/cipher/messages.rs
tgeoghegan: Yes, #3110 doesn't yet reflect the improvements that have been made in the course of this PR, so its `struct EncodingContext` is a bag of booleans instead of using this enum. I also haven't rebased it on the recent substantial changes to main like split connections or the crypto provider interface change, so that PR's utility as a glimpse into the future has decayed. I'm working on rebasing it and cleaning it up now.

I like the idea of an `EncodableVersion` that describes how to encode itself. The difficulty I anticipate though is that the naive thing we'd do is change the type of field `EncodedMessage::version` to be `EncodableVersion`, but we'd have to make that make sense on the send side. I suppose we could always parse incoming messages as `EncodableVersion::Literal)`.

rustls/src/crypto/cipher/messages.rs
tgeoghegan: Sure. The use case for this variant is tests where a message is read from a test vector or the fuzzing corpus and we want to round-trip it and ensure it doesn't get mangled by version encoding heuristics. The alternative to this enum variant would be versions of methods like `EncodedMessage<OutboundPlain<'_>>::to_unencrypted_bytes` and `Message::into_wire_bytes` that preserve version. I'll give this a try and see what it looks like.

djc: I think what @ctz meant was that we could change the tests to have a different expectation, so that we can drop this variant (rather than moving it elsewhere).

tgeoghegan:
Joe's ideas of removing the `TestVector` variant and introducing `enum EncodableVersion` wind up being good ones. But the tricky thing is that the latter invalidates the premise of this PR: if the logic for encoding protocol version is embedded into `EncodedMessage.version`, then we don't need `struct EncodingContext` anymore, at least not immediately. It will _eventually_ be needed for (1) tracking whether the payload is encrypted so we can choose between a unified and full header in DTLS 1.3 and (2) tracking DTLS epoch and sequence number. But if we add it now, it'll be empty and unused until more of DTLS lands, which I think we want to avoid.

So this PR should be re-worked to do nothing but change field `EncodedMessage.version` from `ProtocolVersion` to `EncodableVersion`. Before I try to merge `EncodingContext` to `main`, I'd like to make more progress on implementing DTLS epoch and sequence, including anti-replay and handshake retransmission, to make sure I've captured all the necessary functionality.

For now, I'll make this a draft again.

ctz:
> Joe's ideas of removing the `TestVector` variant and introducing `enum EncodableVersion` wind up being good ones. But the tricky thing is that the latter invalidates the premise of this PR: if the logic for encoding protocol version is embedded into `EncodedMessage.version`, then we don't need `struct EncodingContext` anymore, at least not immediately. It will _eventually_ be needed for (1) tracking whether the payload is encrypted so we can choose between a unified and full header in DTLS 1.3 and (2) tracking DTLS epoch and sequence number. But if we add it now, it'll be empty and unused until more of DTLS lands, which I think we want to avoid.

How about a type that contains everything needed (now, and to be extended in the future) to completely determine the content of the header? Sorry if I've concentrated on the version before now!

tgeoghegan:
> How about a type that contains everything needed (now, and to be extended in the future) to completely determine the content of the header? Sorry if I've concentrated on the version before now!

Right, that's what `EncodingContext` did in previous versions of the change. My expectation is that I'll bring it back in a later PR because I believe it's the right way to handle the other bits of info we'll need for DTLS. Unlike `EncodableVersion`, I don't think it's a good idea to put `EncodingContext` into structures `Message` and `EncodedMessage`, because `EncodingContext` is only meaningful on the outbound path and I don't want to force every place that instantiates `Message` or `EncodedMessage` to create an `EncodingContext` value.

Additionally, one of the pieces of information that will go into `EncodingContext` is DTLS epoch and sequence numbers. I don't think it would be a good idea to make the handshake state machines (which are where many `Message` and `EncodedMessage` values get constructed) responsible for choosing epoch and sequence, since those can't be known until messages get fed into the fragmenter. This is why earlier iterations of this change had the handshake state machines sending just `VersionEncoding` down, and then `SendPath` would construct `EncodingContext`. 

So for those reasons, I think it makes sense to split out the protocol version and represent it as `EncodedVersion` as this PR currently does.

ctz (approved):
Thanks for the clarifications!

tgeoghegan:
Rebased on the latest `main`.

rustls/src/crypto/cipher/mod.rs
```diff
@@ -345,13 +346,13 @@ pub fn make_tls13_aad(payload_len: usize) -> [u8; 5] {
...
     typ: ContentType,
-    vers: ProtocolVersion,
+    vers: EncodableVersion,
     len: usize,
 ) -> [u8; TLS12_AAD_SIZE] {
     let mut out = [0; TLS12_AAD_SIZE];
     put_u64(seq, &mut out[0..]);
     out[8] = typ.into();
-    put_u16(vers.into(), &mut out[9..]);
+    put_u16(vers.encode().into(), &mut out[9..]);
```
ctz: Used in a decryption context, the value here ought to be the one received on the wire. When called with `EncodableVersion::Legacy` the value received is ignored and fixed as TLS1.2. Surprised we don't have a test for this actually! Working on that...

ctz (changes requested):

rustls/src/crypto/cipher/mod.rs
ctz: #3198

tgeoghegan: Great catch, thank you. The way I resolved this was to change `make_tls12_aad` take `ProtocolVersion` rather than `EncodableVersion`, and make the calling context decide how to interpret its `EncodableVersion` value. On the encrypt path, we pass `version.encode()`, which means we'll always be setting `ProtocolVersion::TLSv1_2`. On the decrypt path, we pass `version.version()`, thus preserving whatever was encoded into the received message.

I don't love this solution, however, because it's one more thing that every single crypto provider has to remember to get right. That said, I think changing the `Encrypter/Decrypter` interfaces as discussed in #3109 would help here: we could make it `record_layer`'s responsibility to correctly construct AADs `EncodedMessage` values and only have to get it right once.

tgeoghegan: Fixed in 85e6a8d6ca6bc0b0047134c9b74085b0b6cab71e.

tgeoghegan:
Rebased on main. @ctz could you please re-review and let me know if any more work is needed here?

tgeoghegan:
Test failures! [One of them](https://github.com/rustls/rustls/actions/runs/31626132008/job/94212844126) is `clippy-nightly`, which I believe is tripping a new lint in code I didn't touch.

[The other](https://github.com/rustls/rustls/actions/runs/31626132008/job/94212844157) is a build failure on Windows, due to what looks like a transient failure to connect to a remote host, perhaps an artifact store. I re-pushed the branch just to make CI run again, because I don't think that failure will reproduce, and I can't otherwise make test jobs re-run.

ctz (approved):

djc (approved):

PR #3114, opened 2026-06-30
Update FIPS information for rustls-aws-lc-rs

ctz:
The first commit alone is going to be sensitive (in terms of correctness) to the actual version of aws-lc-fips-sys (see https://github.com/aws/aws-lc-rs/releases/tag/v1.17.1 for explanation.)

Pinning the aws-lc-rs version seems egregious at where we sit in the stack.

~~The second commit addresses that, but in a way which doesn't seem palatable. Probably the best way is to ask for an API for this information from `aws-lc-rs` -- thoughts?~~ -- see #3116

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ mio = { version = "1", features = ["net", "os-poll"] }
 num-bigint = "0.4.4"
 once_cell = { version = "1.16", default-features = false, features = ["alloc", "race"] }
 openssl = "0.10"
-pki-types = { package = "rustls-pki-types", version = "1.14", features = ["alloc"] }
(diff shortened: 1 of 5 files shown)


djc:
> The second commit addresses that, but in a way which doesn't seem palatable. Probably the best way is to ask for an API for this information from `aws-lc-rs` -- thoughts?

I think that would be good. In an ideal world, they would take a dependency on rustls-pki-types and have a `const FIPS_STATUS: FipsStatus;`. Absent that, maybe a `const FIPS: Option<&'static str>`-like thing?

ctz:
Filed https://github.com/aws/aws-lc-rs/issues/1166

ctz:
Dropped the second commit and filed #3116 to reconsider it later.

djc (approved):

PR #3115, opened 2026-07-01

ctz:
Fixes https://github.com/rustls/rustls/actions/runs/28467290939/job/84370248475

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
 
 [[package]]
 name = "alloc-stdlib"
-version = "0.2.2"
+version = "0.2.4"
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Some sizable regressions in no_crypto performance results...

And it looks like we should avoid the latest zlib-rs for now?

- https://github.com/trifectatechfoundation/zlib-rs/issues/536

ctz:
> Some sizable regressions in no_crypto performance results...

Seems like this is 100% in zeroize.

djc:
> > Some sizable regressions in no_crypto performance results...
> 
> Seems like this is 100% in zeroize.

For this PR, hold that back?

ctz:
> > Some sizable regressions in no_crypto performance results...
> 
> Seems like this is 100% in zeroize.

https://github.com/RustCrypto/utils/issues/1504

PR #3127, opened 2026-07-07
Caller-provided MessageEncrypter output buffer support

### LINKED ISSUE #3068: Allow in-place `MessageEncrypter::encrypt`
https://rustls.dev/docs/crypto/cipher/trait.MessageEncrypter.html#tymethod.encrypt currently takes a borrowed immutable input and produces an owned output. This means it is required that it does a copy, an allocation and highly likely a buffer clear.

Instead, we could choose another option, such as:

- a mutable input buffer, with a requirement for some extra space at the end (enough to include a tag, any other framing, and any prefix -- in this case the encryption primitive would have to do in-place into a moved & overlapping buffer.
- a mutable input buffer where the caller has pre-arranged space for any and all prefix, framing and tag.

In the short term this would move overhead out of this (public) interface, into its (internal) caller. Longer term could chase through this, eliminating the overhead from the library as a whole.

cpu:
This PR reworks `MessageEncrypter::encrypt()` to write encrypted records into a caller-provided `&mut [u8]` instead of returning a freshly-allocated `OutboundOpaque`, mirroring how `MessageDecrypter::decrypt()` already operates in-place over a borrowed buffer. The plaintext input stays borrowed and immutable as both `*ring*` and `aws-lc-rs` only offer in-place sealing, so one plaintext copy is inherent regardless. The new contract removes the per-record allocation and gives callers control over where ciphertext lands.

Implementations size their output with `encrypted_payload_len()`, write whatever framing their suite needs (explicit nonce, padding, tag) anywhere within the `out` buffer (a small `EncryptBuffer` cursor keeps the provider impls near-identical to their previous form), and return an `EncodedMessage` whose payload borrows the written prefix. Encoding the record header becomes the internal caller's job, via `encode_record_header()`. With nothing producing `OutboundOpaque` any more, the type is removed from the public API. 

Inside the buffered send path we feed encryption with recycled spent record buffers, and the split-mode API gains `SendTraffic::write_into()`, which fragments and encrypts application data directly into a caller's buffer with no internal queueing or allocation. The buffered API gains code to reuse previous buffers to also benefit from the new API design, but this may be too speculative or better to remove in favour of a larger plan to rearchitect the buffered API in terms of the new split-mode unbuffered API. With that in mind I left those commits at the end, easy to drop from the series.

Resolves https://github.com/rustls/rustls/issues/3068

### DIFF
--- a/rustls-aws-lc-rs/src/tls12.rs
+++ b/rustls-aws-lc-rs/src/tls12.rs
@@ -3,9 +3,9 @@ use alloc::boxed::Box;
 use aws_lc_rs::{aead, tls_prf};
 use pki_types::FipsStatus;
 use rustls::crypto::cipher::{
-    AeadKey, EncodedMessage, InboundOpaque, Iv, KeyBlockShape, MessageDecrypter, MessageEncrypter,
-    NONCE_LEN, Nonce, OutboundOpaque, OutboundPlain, Tls12AeadAlgorithm, UnsupportedOperationError,
-    make_tls12_aad,
+    AeadKey, EncodedMessage, EncryptBuffer, InboundOpaque, Iv, KeyBlockShape, MessageDecrypter,
+    MessageEncrypter, NONCE_LEN, Nonce, OutboundPlain, Tls12AeadAlgorithm,
+    UnsupportedOperationError, make_tls12_aad,
 };
 use rustls::crypto::kx::{ActiveKeyExchange, KeyExchangeAlgorithm, SharedSecret};
 use rustls::crypto::tls12::{Prf, PrfSecret};
@@ -302,13 +302,14 @@ impl MessageDecrypter for GcmMessageDecrypter {
 }
 
 impl MessageEncrypter for GcmMessageEncrypter {
-    fn encrypt(
+    fn encrypt<'a>(
         &mut self,
         msg: EncodedMessage<OutboundPlain<'_>>,
         seq: u64,
-    ) -> Result<EncodedMessage<OutboundOpaque>, Error> {
+        out: &'a mut [u8],
+    ) -> Result<EncodedMessage<&'a [u8]>, Error> {
         let total_len = self.encrypted_payload_len(msg.payload.len());
-        let mut payload = OutboundOpaque::with_capacity(total_len);
+        let mut payload = EncryptBuffer::new(out, total_len)?;
 
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).to_array()?);
         let aad = aead::Aad::from(make_tls12_aad(seq, msg.typ, msg.version, msg.payload.len()));
@@ -327,7 +328,7 @@ impl MessageEncrypter for GcmMessageEncrypter {
         Ok(EncodedMessage {
             typ: msg.typ,
             version: msg.version,
-            payload,
+            payload: payload.into_written(),
         })
     }
 
@@ -392,27 +393,32 @@ impl MessageDecrypter for ChaCha20Poly1305MessageDecrypter {
 }
 
 impl MessageEncrypter for ChaCha20Poly1305MessageEncrypter {
-    fn encrypt(
+    fn encrypt<'a>(
         &mut self,
         msg: EncodedMessage<OutboundPlain<'_>>,
         seq: u64,
-    ) -> Result<EncodedMessage<OutboundOpaque>, Error> {
+        out: &'a mut [u8],
+    ) -> Result<EncodedMessage<&'a [u8]>, Error> {
         let total_len = self.encrypted_payload_len(msg.payload.len());
-        let mut payload = OutboundOpaque::with_capacity(total_len);
+        let mut payload = EncryptBuffer::new(out, total_len)?;
 
         let nonce =
             aead::Nonce::assume_unique_for_key(Nonce::new(&self.enc_offset, seq).to_array()?);
         let aad = aead::Aad::from(make_tls12_aad(seq, msg.typ, msg.version, msg.payload.len()));
         payload.extend_from_chunks(&msg.payload);
 
-        self.enc_key
-            .seal_in_place_append_tag(nonce, aad, &mut payload)
-            .map_err(|_| Error::EncryptError)?;
+        match self
+            .enc_key
+            .seal_in_place_separate_tag(nonce, aad, payload.as_mut())
+        {
+            Ok(tag) => payload.extend_from_slice(tag.as_ref()),
+            Err(_) => return Err(Error::EncryptError),
+        }
 
         Ok(EncodedMessage {
             typ: msg.typ,
             version: msg.version,
-            payload,
+            payload: payload.into_written(),
         })
     }
 
--- a/rustls-aws-lc-rs/src/tls13.rs
+++ b/rustls-aws-lc-rs/src/tls13.rs
@@ -4,8 +4,8 @@ use aws_lc_rs::hkdf::KeyType;
 use aws_lc_rs::{aead, hkdf, hmac};
 use pki_types::FipsStatus;
 use rustls::crypto::cipher::{
-    AeadKey, EncodedMessage, InboundOpaque, Iv, MessageDecrypter, MessageEncrypter, Nonce,
-    OutboundOpaque, OutboundPlain, Tls13AeadAlgorithm, UnsupportedOperationError, make_tls13_aad,
+    AeadKey, EncodedMessage, EncryptBuffer, InboundOpaque, Iv, MessageDecrypter, MessageEncrypter,
+    Nonce, OutboundPlain, Tls13AeadAlgorithm, UnsupportedOperationError, make_tls13_aad,
 };
 use rustls::crypto::tls13::{Hkdf, HkdfExpander, OkmBlock, OutputLengthError};
 use rustls::crypto::{self, CipherSuite};
@@ -219,29 +219,34 @@ struct AeadMessageDecrypter {
 }
 
 impl MessageEncrypter for AeadMessageEncrypter {
-    fn encrypt(
+    fn encrypt<'a>(
         &mut self,
         msg: EncodedMessage<OutboundPlain<'_>>,
         seq: u64,
-    ) -> Result<EncodedMessage<OutboundOpaque>, Error> {
+        out: &'a mut [u8],
+    ) -> Result<EncodedMessage<&'a [u8]>, Error> {
         let total_len = self.encrypted_payload_len(msg.payload.len());
-        let mut payload = OutboundOpaque::with_capacity(total_len);
+        let mut payload = EncryptBuffer::new(out, total_len)?;
 
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).to_array()?);
         let aad = aead::Aad::from(make_tls13_aad(total_len));
         payload.extend_from_chunks(&msg.payload);
         payload.extend_from_slice(&msg.typ.to_array());
 
-        self.enc_key
-            .seal_in_place_append_tag(nonce, aad, &mut payload)
-            .map_err(|_| Error::EncryptError)?;
+        match self
+            .enc_key
+            .seal_in_place_separate_tag(nonce, aad, payload.as_mut())
+        {
+            Ok(tag) => payload.extend_from_slice(tag.as_ref()),
+            Err(_) => return Err(Error::EncryptError),
+        }
 
         Ok(EncodedMessage {
             typ: ContentType::ApplicationData,
             // Note: all TLS 1.3 application data records use TLSv1_2 (0x0303) as the legacy record
             // protocol version, see https://www.rfc-editor.org/rfc/rfc8446#section-5.1
             version: ProtocolVersion::TLSv1_2,
-            payload,
+            payload: payload.into_written(),
         })
     }
 
@@ -280,27 +285,32 @@ struct GcmMessageEncrypter {
 }
 
 impl MessageEncrypter for GcmMessageEncrypter {
-    fn encrypt(
+    fn encrypt<'a>(
         &mut self,
         msg: EncodedMessage<OutboundPlain<'_>>,
         seq: u64,
-    ) -> Result<EncodedMessage<OutboundOpaque>, Error> {
+        out: &'a mut [u8],
+    ) -> Result<EncodedMessage<&'a [u8]>, Error> {
         let total_len = self.encrypted_payload_len(msg.payload.len());
-        let mut payload = OutboundOpaque::with_capacity(total_len);
+        let mut payload = EncryptBuffer::new(out, total_len)?;
 
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).to_array()?);
         let aad = aead::Aad::from(make_tls13_aad(total_len));
         payload.extend_from_chunks(&msg.payload);
         payload.extend_from_slice(&msg.typ.to_array());
 
-        self.enc_key
-            .seal_in_place_append_tag(nonce, aad, &mut payload)
-            .map_err(|_| Error::EncryptError)?;
+        match self
+            .enc_key
+            .seal_in_place_separate_tag(nonce, aad, payload.as_mut())
+        {
+            Ok(tag) => payload.extend_from_slice(tag.as_ref()),
+            Err(_) => return Err(Error::EncryptError),
+        }
 
         Ok(EncodedMessage {
             typ: ContentType::ApplicationData,
             version: ProtocolVersion::TLSv1_2,
-            payload,
+            payload: payload.into_written(),
         })
     }
 
--- a/rustls/src/conn/send.rs
+++ b/rustls/src/conn/send.rs
@@ -91,8 +91,10 @@ impl SendPath {
             }
 
             self.perhaps_write_key_update();
-            self.sendable_tls
-                .append(self.encrypt_state.encrypt_outgoing(m));
+            let record = self
+                .encrypt_state
+                .encrypt_outgoing(m, self.sendable_tls.take_spare());
+            self.sendable_tls.append(record);
         }
     }
 
@@ -241,7 +243,7 @@ impl SendPath {
         let message = EncodedMessage::<Payload<'static>>::from(Message::build_key_update_notify());
         self.queued_key_update_message = Some(
             self.encrypt_state
-                .encrypt_outgoing(message.borrow_outbound()),
+                .encrypt_outgoing(message.borrow_outbound(), Vec::new()),
         );
 
         if let Some(mut ks) = self.tls13_key_schedule.take() {
@@ -329,7 +331,7 @@ impl Default for SendPath {
             has_sent_fatal_alert: false,
             has_sent_close_notify: false,
             message_fragmenter: MessageFragmenter::default(),
-            sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
+            sendable_tls: ChunkVecBuffer::new_recycling(Some(DEFAULT_BUFFER_LIMIT)),
             queued_key_update_message: None,
             refresh_traffic_keys_pending: false,
             negotiated_version: None,
--- a/rustls/src/conn/send.rs
+++ b/rustls/src/conn/send.rs
@@ -91,11 +91,8 @@ impl SendPath {
             }
 
             self.perhaps_write_key_update();
-            self.sendable_tls.append(
-                self.encrypt_state
-                    .encrypt_outgoing(m)
-                    .encode(),
-            );
+            self.sendable_tls
+                .append(self.encrypt_state.encrypt_outgoing(m));
         }
     }
 
@@ -190,8 +187,7 @@ impl SendPath {
         let message = EncodedMessage::<Payload<'static>>::from(Message::build_key_update_notify());
         self.queued_key_update_message = Some(
             self.encrypt_state
-                .encrypt_outgoing(message.borrow_outbound())
-                .encode(),
+                .encrypt_outgoing(message.borrow_outbound()),
         );
 
         if let Some(mut ks) = self.tls13_key_schedule.take() {
--- a/rustls/src/conn/send.rs
+++ b/rustls/src/conn/send.rs
@@ -7,7 +7,7 @@ use crate::crypto::cipher::{
 use crate::enums::{ContentType, ProtocolVersion};
 use crate::error::{AlertDescription, Error};
 use crate::log::{debug, error};
-use crate::msgs::{AlertLevel, Message, MessageFragmenter};
+use crate::msgs::{AlertLevel, HEADER_SIZE, Message, MessageFragmenter};
 use crate::tls13::key_schedule::KeyScheduleTrafficSend;
 use crate::vecbuf::ChunkVecBuffer;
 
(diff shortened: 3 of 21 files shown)


cpu:
> This branch has conflicts that must be resolved

I'm out of time to continue this work for now but wanted to put it up for feedback. I'll resolve conflicts alongside review feedback.

rustls/src/conn/send.rs
```diff
@@ -126,6 +126,60 @@ impl SendPath {
         }
     }
 
+    /// Encrypt application data from `payload` directly into `out`.
+    ///
+    /// Any records already queued inside the connection (for example, a
+    /// pending `key_update`) are written to `out` first.
+    ///
+    /// Records are written while `out` has space for them. The returned
+    /// [`WrittenInto`] indicates how much of `payload` was consumed and how many
```
ctz, line 137: I guess a more complex and "optimistic" option here would be finding a reduced prefix of `payload` which _can_ fit into `out`.

That would have quite different semantics. At the moment, assuming nothing is pending, if `out.len() < payload.len() + overhead` then this function returns `WrittenInto { 0, 0 }`.

(Just a note, not suggesting this changes.)

rustls/src/crypto/cipher/messages.rs
```diff
@@ -286,6 +287,50 @@ impl<'a> OutboundPlain<'a> {
...
+        remaining: usize,
+    },
+}
+
+impl<'a> Iterator for Chunks<'a> {
+    type Item = &'a [u8];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            Self::Single(chunk) => chunk.take(),
```
djc: Nit: reduce rightward drift by yielding the `Multiple` fields out of the `match`?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -226,22 +227,22 @@ impl<'a> OutboundPlain<'a> {
...
+    /// Iterate over the payload's chunks of bytes, in order.
+    ///
+    /// Empty chunks are not yielded.
+    pub fn chunks(&self) -> impl Iterator<Item = &[u8]> + '_ {
+        match self {
+            Self::Single(chunk) => Chunks::Single((!chunk.is_empty()).then_some(*chunk)),
+            Self::Multiple { chunks, start, end } => Chunks::Multiple {
+                chunks: chunks.iter(),
+                skip: *start,
+                remaining: end - start,
```
djc, line 242: I did not find it obvious from reading the docs for `end` in `OutboundPlain::Multiple` that it is actually the total number of bytes across chunks; "The end cursor into the last chunk" reads like it might be relative to the start of the last chunk. Maybe it would be nice to clarify that in its docstring.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -286,6 +287,50 @@ impl<'a> OutboundPlain<'a> {
...
+                    return None;
+                }
+
+                let chunk = chunks.next()?;
+                if chunk.len() <= *skip {
+                    *skip -= chunk.len();
+                    continue;
+                }
+
+                let chunk = &chunk[*skip..];
```
djc: Nit: maybe try to write this as `split_at_checked()`?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -364,6 +409,71 @@ impl<const N: usize> From<&[u8; N]> for OutboundOpaque {
     }
 }
 
+/// A fixed-size buffer into which a [`MessageEncrypter`][] writes an encrypted message payload.
+///
+/// This wraps the output buffer passed to [`MessageEncrypter::encrypt()`][], tracking how
+/// much of it has been written as the append methods fill it front-to-back. It writes
+/// into caller-owned memory and cannot grow, instead the append methods panic if a write
+/// would overflow the buffer and it is the caller's responsibility to avoid this.
```
djc: Is this necessary? Are these invariants easy to enforce in the caller? Could we employ fallibility instead of panicking?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -364,6 +409,71 @@ impl<const N: usize> From<&[u8; N]> for OutboundOpaque {
...
+            .into()),
+        }
+    }
+
+    /// Append bytes from a slice.
+    ///
+    /// Panics if the write would extend beyond the buffer wrapped by [`Self::new()`].
+    pub fn extend_from_slice(&mut self, slice: &[u8]) {
+        self.buf[self.used..self.used + slice.len()].copy_from_slice(slice);
+        self.used += slice.len();
```
djc: If we decide to make this fallible instead, maybe `split_at_mut_checked()`?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -364,6 +409,71 @@ impl<const N: usize> From<&[u8; N]> for OutboundOpaque {
...
+    }
+
+    /// Append bytes from an `OutboundPlain`'s chunks.
+    ///
+    /// Panics if the write would extend beyond the buffer wrapped by [`Self::new()`].
+    pub fn extend_from_chunks(&mut self, chunks: &OutboundPlain<'_>) {
```
djc, line 387: Nit: this should go above `extend_from_slice()`.

rustls/src/crypto/cipher/messages.rs
```diff
@@ -364,6 +409,71 @@ impl<const N: usize> From<&[u8; N]> for OutboundOpaque {
...
+/// A fixed-size buffer into which a [`MessageEncrypter`][] writes an encrypted message payload.
+///
+/// This wraps the output buffer passed to [`MessageEncrypter::encrypt()`][], tracking how
+/// much of it has been written as the append methods fill it front-to-back. It writes
+/// into caller-owned memory and cannot grow, instead the append methods panic if a write
+/// would overflow the buffer and it is the caller's responsibility to avoid this.
+///
+/// [`MessageEncrypter`]: crate::crypto::cipher::MessageEncrypter
+/// [`MessageEncrypter::encrypt()`]: crate::crypto::cipher::MessageEncrypter::encrypt()
+pub struct EncryptBuffer<'a> {
```
djc, line 361: Not sure this commit meaningfully stands apart from the next one.

rustls/src/crypto/cipher/record_layer.rs
```diff
@@ -26,22 +28,38 @@ impl EncryptionState {
...
     ///
     /// `plain` is a TLS message we'd like to send.  This function
     /// panics if the requisite keying material hasn't been established yet.
-    pub(crate) fn encrypt_outgoing(
-        &mut self,
-        plain: EncodedMessage<OutboundPlain<'_>>,
-    ) -> EncodedMessage<OutboundOpaque> {
+    pub(crate) fn encrypt_outgoing(&mut self, plain: EncodedMessage<OutboundPlain<'_>>) -> Vec<u8> {
         assert!(self.pre_encrypt_action(0) != Some(PreEncryptAction::Refuse));
+        let encrypter = self.message_encrypter.as_mut().unwrap();
```
djc, line 74: Why is this `unwrap()` safe?

rustls/src/crypto/cipher/record_layer.rs
```diff
@@ -26,22 +28,38 @@ impl EncryptionState {
...
-            .unwrap()
-            .encrypt(plain, seq)
-            .unwrap()
+
+        let needed = HEADER_SIZE + encrypter.encrypted_payload_len(plain.payload.len());
+        let mut record = vec![0u8; needed];
+        let out_ptr = record.as_ptr();
+        let encrypted = encrypter
+            .encrypt(plain, seq, &mut record[HEADER_SIZE..])
+            .unwrap();
```
djc, line 83: And this one?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -164,15 +164,33 @@ impl EncodedMessage<OutboundPlain<'_>> {
...
 
+/// Encode a TLS record header.
+///
+/// `typ`, `version` and `len` describe the record's payload.  `len` is
+/// debug-asserted to fit the header's 16-bit length field, and truncated to
+/// 16 bits in release builds.
+pub(crate) fn encode_record_header(
+    typ: ContentType,
+    version: ProtocolVersion,
+    len: usize,
```
djc: Maybe this should take `u16` and the caller should take responsibility for ensuring it's short enough?

rustls/src/crypto/cipher/messages.rs
```diff
@@ -164,15 +164,33 @@ impl EncodedMessage<OutboundPlain<'_>> {
...
         encoded_payload
     }
 }
 
+/// Encode a TLS record header.
+///
+/// `typ`, `version` and `len` describe the record's payload.  `len` is
+/// debug-asserted to fit the header's 16-bit length field, and truncated to
+/// 16 bits in release builds.
+pub(crate) fn encode_record_header(
```
djc, line 158: Looks like this could be a separate commit?

rustls/src/conn/send.rs
```diff
@@ -300,4 +354,19 @@ pub(crate) trait SendOutput {
     fn send_msg(&mut self, m: Message<'_>, must_encrypt: bool);
 }
 
+/// The outcome of encrypting application data into a caller-provided buffer.
+///
+/// Returned by [`SendTraffic::write_into()`].
+///
+/// [`SendTraffic::write_into()`]: crate::split::SendTraffic::write_into()
+#[expect(clippy::exhaustive_structs)]
```
djc: Since this is an output type, I think it could be `non_exhaustive` instead?

djc (commented):
Here's a first round of feedback.

rustls/src/crypto/cipher/record_layer.rs
cpu: I can add a safety comment but I wanted to point out that it's not something I added, it's pre-existing: https://github.com/rustls/rustls/blob/1364a77f00e6e8d67582bb80999e8b5880ce61a7/rustls/src/crypto/cipher/record_layer.rs#L40-L42

rustls/src/crypto/cipher/record_layer.rs
cpu: Ditto here, pre-existing from `main`'s `EncryptionState::encrypt_outgoing()`: https://github.com/rustls/rustls/blob/1364a77f00e6e8d67582bb80999e8b5880ce61a7/rustls/src/crypto/cipher/record_layer.rs#L43-L44

rustls/src/crypto/cipher/record_layer.rs
djc: Fair enough, could ignore for now.

@ctz do you think we could/should do `Error::Unreachable` for these? Also looks like we're unwrapping on `MessageEncrypter::encrypt()` failures which doesn't feel like the right thing to do here.

Wonder if it's useful/interesting to hoist the `Option` here out of `EncryptionState` into `SendPath` so we can be more deliberate about this, but it requires a fair amount of work.

rustls/src/crypto/cipher/messages.rs
cpu: Good idea. Done.

rustls/src/crypto/cipher/messages.rs
cpu: >  Maybe it would be nice to clarify that in its docstring.

Yeah, I see what you mean. I've updated the docs to add more specificity here.

rustls/src/crypto/cipher/messages.rs
cpu: Done.

rustls/src/crypto/cipher/messages.rs
cpu: Fixed.

rustls/src/crypto/cipher/messages.rs
cpu: Fair, squished them together.

rustls/src/crypto/cipher/messages.rs
cpu: Sure. Implemented in a commit on top of the separate extraction of the helper.

rustls/src/crypto/cipher/messages.rs
cpu: Extracted up-front into its own commit (and then did the usize->u16 conversion on top of that).

rustls/src/conn/send.rs
cpu: Good call. Switched to non_exhaustive.

rustls/src/crypto/cipher/messages.rs
cpu: I don't think that it's strictly necessary, we _could_ make this fallible. But, I don't think it's the right call (_loosely held opinion_). 

`EncryptBuffer::new()` already returns an error for the one condition a caller can genuinely get wrong where the provided buffer is too small. AIUI an append overflow later means the encrypter implementation contradicted its own `encrypted_payload_len()` and I think merits a panic instead of complicating the logic with another `Result`.

I tried to make this case clearer in the doc strings. If you or ctz don't buy it and want me to revisit you can unresolve this conversation thread and I'll try switching to a fallible arrangement.

rustls/src/crypto/cipher/messages.rs
cpu: Decided not to make it fallible for now, [see my reply in the other thread](https://github.com/rustls/rustls/pull/3127#discussion_r3597762167). If we revisit that I'll unresolve this too.

rustls/src/crypto/cipher/record_layer.rs
```diff
@@ -26,22 +28,39 @@ impl EncryptionState {
...
         self.write_seq += 1;
-        self.message_encrypter
-            .as_mut()
-            .unwrap()
-            .encrypt(plain, seq)
-            .unwrap()
+
+        let needed = HEADER_SIZE + encrypter.encrypted_payload_len(plain.payload.len());
+        let mut record = vec![0u8; needed];
+        let out_ptr = record.as_ptr();
```
djc: Nit: can we inline the `as_ptr()` call into the `debug_assert_eq!()` call to make it more obvious that it is not used as part of the normal execution?

rustls/src/vecbuf.rs
```diff
@@ -96,11 +96,17 @@ impl ChunkVecBuffer {
...
+        let mut chunks = Vec::from(mem::take(&mut self.chunks));
+        // slice off `prefix_used` if needed (uncommon)
+        let prefix = mem::take(&mut self.prefix_used);
+        chunks[0].drain(0..prefix);
+        chunks
     }
 
     pub(crate) fn take_one_vec(&mut self) -> Vec<u8> {
-        let Some(mut first) = self.chunks.pop_front() else {
+        // `pop()` (unlike `pop_front()`) slices off `prefix_used`
```
djc: Nit: I think the "(unlike `pop_front()`)" parenthetical probably doesn't make sense to keep around?

rustls/src/conn/send.rs
```diff
@@ -329,7 +331,7 @@ impl Default for SendPath {
             has_sent_fatal_alert: false,
             has_sent_close_notify: false,
             message_fragmenter: MessageFragmenter::default(),
-            sendable_tls: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
+            sendable_tls: ChunkVecBuffer::new_recycling(Some(DEFAULT_BUFFER_LIMIT)),
```
djc, line 334: I don't think we should add this? It seems to add more complexity, and I think we're going to kill `sendable_tls` soon?

cpu: Yeah, that's what I was trying to hint at in the PR description:

> The buffered API gains code to reuse previous buffers to also benefit from the new API design, but this may be too speculative or better to remove in favour of a larger plan to rearchitect the buffered API in terms of the new split-mode unbuffered API. With that in mind I left those commits at the end, easy to drop from the series.

I'm happy to drop this commit if ctz is on the same page as you. Just also want to note that without the reuse it might show up as a perf regression until `sendable_tls` goes away.

rustls/src/crypto/cipher/record_layer.rs
cpu: Unfortunately that doesn't work because of the mutable borrow in `encrypt(plain, seq, &mut out[HEADER_SIZE..])` that we need to get the `encrypted.payload` in order to do the asserts. We need to pull out the `*const u8` `out_ptr` and `usize` `out_len` before that borrow. I could `#[cfg(debug)]` gate them, but it felt like overkill to me. 

Open to other suggestions.

rustls/src/vecbuf.rs
cpu: Removed.

rustls/src/crypto/cipher/record_layer.rs
djc: `#[cfg(debug)]` sounds okay to me, otherwise maybe a comment on the same line?

cpu: Added `#[cfg(debug_assertions)]`. I also had to add a `#[cfg(debug_assertions)]` scope around the `debug_assert_eq!` or release builds would error on compile from the missing var defs.

rustls/src/conn/send.rs
ctz: I think we should keep it purely to avoid this PR oscillating performance too far, which would both over-state the performance loss of this PR, and then over-state the win of any follow-on?

(Any idea where the benchmarks have gone on this PR?)

cpu:
@rustls-benchmarking bench

rustls/src/conn/send.rs
cpu: > (Any idea where the benchmarks have gone on this PR?)

Sorry, I'm out of practice. I've been sitting around waiting for it to run but I forgot that I opened this from a fork. I think it wasn't going to run until there was a PR approval from a maintainer. I gave it a manual comment kick.

iainmcgin:
In case it's useful for the benchmark question: I built a transport on `write_into` and measured it end to end, serving HTTP under hyper with the load generator on a separate host, so that only the server's CPU is in the measurement. Server CPU per response byte, one pinned core, medians of three interleaved runs:

| body | buffered `writer()`/`write_tls()` | split + `write()` | split + `write_into()` |
|---|---|---|---|
| 16 KiB | 0.883 | 0.874 | 0.828 |
| 64 KiB | 0.497 | 0.468 | 0.432 |
| 256 KiB | 0.457 | 0.389 | 0.376 |
| 1 MiB | 0.480 | 0.446 | 0.397 |
| 8 MiB | 0.419 | 0.405 | 0.353 |

ns per byte, lower is better. `write_into()` comes out 3-13% cheaper than split mode's existing `write()` at every size, and the buffered path is slowest at all five.

All three are built on this branch, so all of them get the record-buffer reuse from the last commit.

rustls/src/conn/send.rs
djc: Fair enough.

rustls/src/vecbuf.rs
```diff
@@ -96,11 +96,17 @@ impl ChunkVecBuffer {
...
+        // slice off `prefix_used` if needed (uncommon)
+        let prefix = mem::take(&mut self.prefix_used);
+        chunks[0].drain(0..prefix);
+        chunks
     }
 
     pub(crate) fn take_one_vec(&mut self) -> Vec<u8> {
-        let Some(mut first) = self.chunks.pop_front() else {
+        // `pop()` slices off `prefix_used`
+        let Some(mut first) = self.pop() else {
```
djc, line 127: Tiny nit: by caller-before-callee logic, this should be moved up now.

rustls/src/conn/split.rs
```diff
@@ -69,6 +71,28 @@ impl SendTraffic {
...
+    /// internally. Encrypted records are written straight into the caller's
+    /// buffer, along with any internally-queued records (such as a pending
+    /// `key_update`).
+    ///
+    /// Records are written while `out` has space for them. The returned
+    /// [`WrittenInto`] indicates how much of `application_data` was consumed and
+    /// how many bytes were written to `out`. Once the written bytes have been
+    /// communicated to the peer, call again with the unconsumed remainder of
+    /// `application_data`.
+    pub fn write_into(
```
djc: Should this be named something more like `write_tls_into()` to make it more obvious that the output stream contains TLS records?

rustls/src/crypto/cipher/record_layer.rs
```diff
@@ -32,9 +32,27 @@ impl EncryptionState {
...
+    ) -> Vec<u8> {
+        // The buffer is overwritten in full below, so a reused buffer's
+        // existing initialized length is kept as-is and only newly-grown space
+        // needs zero-filling. A fresh buffer is allocated pre-zeroed.
         let needed = HEADER_SIZE + self.encrypted_len(plain.payload.len());
-        let mut record = vec![0u8; needed];
+        if record.capacity() == 0 {
+            record = vec![0u8; needed];
+        } else if record.len() < needed {
+            record.resize(needed, 0);
```
djc: I think we could collapse these branches?

rustls/src/vecbuf.rs
```diff
@@ -20,6 +20,11 @@ pub(crate) struct ChunkVecBuffer {
 
     /// The total upper limit (in bytes) of this object.
     limit: Option<usize>,
+
+    /// Fully-consumed chunks retained for reuse via `take_spare()`.
+    ///
+    /// `None` means recycling is disabled and spent chunks are freed.
+    spare: Option<Vec<Vec<u8>>>,
```
djc, line 31: Why would someone want to disable recycling?

rustls/src/vecbuf.rs
cpu: Moved.

rustls/src/conn/split.rs
cpu: Renamed.

rustls/src/crypto/cipher/record_layer.rs
cpu: I reworked slightly and added a comment to justify why there's a branch.

rustls/src/vecbuf.rs
cpu: I added a comment to help clarify this. Re-use is only valuable for the sendable tls path where we construct the chunk vec buff w/ `new_recycling()`. For the plaintext path keeping consumed chunks would be counter productive, they're never reused.

rustls-aws-lc-rs/src/tls12.rs
```diff
@@ -388,27 +389,29 @@ impl MessageDecrypter for ChaCha20Poly1305MessageDecrypter {
...
         let nonce =
             aead::Nonce::assume_unique_for_key(Nonce::new(&self.enc_offset, seq).to_array()?);
         let aad = aead::Aad::from(make_tls12_aad(seq, msg.typ, msg.version, msg.payload.len()));
         payload.extend_from_chunks(&msg.payload);
 
         self.enc_key
-            .seal_in_place_append_tag(nonce, aad, &mut payload)
+            .seal_in_place_separate_tag(nonce, aad, payload.as_mut())
+            .map(|tag| payload.extend_from_slice(tag.as_ref()))
             .map_err(|_| Error::EncryptError)?;
```
ctz: nit: where we have both a `map` and `map_err` on the same value, I think it's better to have a two-arm match.

rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -219,29 +219,31 @@ struct AeadMessageDecrypter {
...
 
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).to_array()?);
         let aad = aead::Aad::from(make_tls13_aad(total_len));
         payload.extend_from_chunks(&msg.payload);
         payload.extend_from_slice(&msg.typ.to_array());
 
         self.enc_key
-            .seal_in_place_append_tag(nonce, aad, &mut payload)
+            .seal_in_place_separate_tag(nonce, aad, payload.as_mut())
+            .map(|tag| payload.extend_from_slice(tag.as_ref()))
```
ctz: ditto (not going to tag the remainder)

rustls/src/conn/split.rs
```diff
@@ -69,6 +71,28 @@ impl SendTraffic {
...
+    ///
+    /// Records are written while `out` has space for them. The returned
+    /// [`WrittenInto`] indicates how much of `application_data` was consumed and
+    /// how many bytes were written to `out`. Once the written bytes have been
+    /// communicated to the peer, call again with the unconsumed remainder of
+    /// `application_data`.
+    pub fn write_tls_into(
+        &mut self,
+        application_data: OutboundPlain<'_>,
+        out: &mut [u8],
```
ctz, line 94: Just a thought about how a caller sizes this buffer. I think the current shape is good. The issue would be adding a "tell me the length I need" function, as this could race the availability of messages from another thread.

ctz (approved):

rustls/src/conn/split.rs
cpu: Do you think I should add some kind of a warning comment about avoiding this in a future refactor?

rustls-aws-lc-rs/src/tls12.rs
cpu: Fixed throughout.

rustls-aws-lc-rs/src/tls13.rs
cpu: Fixed throughout.

cpu:
> [CIFuzz / Fuzzing (pull_request)](https://github.com/rustls/rustls/actions/runs/29928808779/job/88959866608?pr=3127) Failing after 3m

Looks unrelated:

> 161.5 E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/p/python3.8/libpython3.8-minimal_3.8.10-0ubuntu1~20.04.18_amd64.deb  Connection failed [IP: 185.125.190.81 80]

rustls/src/conn/split.rs
cpu: Added some coverage/justification on this to the rustdoc.

rustls/src/crypto/cipher/record_layer.rs
cpu: I have to resolve this thread to allow merging. Ctz: if you think there's a change to be made here as follow-up could you file an issue to track it explicitly? ty

PR #3128, opened 2026-07-08
Miscellaneous documentation improvements

ctz:
Mostly further chasing recent changes into doc strings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,9 +96,11 @@ jobs:
 
       - name: cargo build (debug; no-std)
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -97,8 +97,10 @@ jobs:
       - name: cargo build (debug; no-std)
         run: |
           cargo build --locked --lib -p rustls $(admin/all-features-except std,brotli rustls)
+          cargo build --locked --lib -p rustls-util
```
djc: I think we could/should make all of these `check` instead of `build`?

djc (approved):
Have some early feedback!

PR #3129, opened 2026-07-08
Make `OpaqueOutbound` into a struct with fields

tgeoghegan:
To prepare for `OutboundOpaque` gaining a field carrying encoding information, which in turn will eventually be consulted by `EncodedMessage<OutboundOpaque>::encode`, `OutboundOpaque` becomes a conventional struct with fields instead of a tuple struct wrapping `Vec<u8>`.

Additionally, we remove the `From<&[u8]>` and `From<&[u8; N]>` implementations, replacing them with a named constructor `OutboundOpaque::from_byte_slice` and updating callsites accordingly. As `OutboundOpaque` gains more fields, it will no longer be possible to implement these traits.

### DIFF
--- a/rustls/src/crypto/cipher/messages.rs
+++ b/rustls/src/crypto/cipher/messages.rs
@@ -54,7 +54,7 @@ impl<'a> EncodedMessage<Payload<'a>> {
         EncodedMessage {
             typ: self.typ,
             version: self.version,
-            payload: OutboundOpaque::from(self.payload.bytes()),
+            payload: OutboundOpaque::from_byte_slice(self.payload.bytes()),
         }
     }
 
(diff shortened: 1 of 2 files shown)


tgeoghegan:
This PR was split out of #3112. More discussion can be found in #40, #3110.

djc:
> Additionally, we remove the From<&[u8]> and From<&[u8; N]> implementations, replacing them with a named constructor OutboundOpaque::from_byte_slice and updating callsites accordingly. As OutboundOpaque gains more fields, it will no longer be possible to implement these traits.

In the future, please keep this as separate commits, since it is semantically separate change.

djc (approved):

djc:
(Sorry for picking all the nits, I'm hoping that will smooth out the rest of the contribution process but I realize it might make this stage extra slow/frustrating.)

tgeoghegan:
> (Sorry for picking all the nits, I'm hoping that will smooth out the rest of the contribution process but I realize it might make this stage extra slow/frustrating.)

It's perfectly okay! I'm a guest in your house and am happy to learn your way of doing things. Thank you for taking the time to point these things out to me.

PR #3133, opened 2026-07-10
Switch log feature backend from log to tracing

### LINKED ISSUE #3122: Proposal: migrate from log to tracing (with log compatibility feature)
**Checklist**

rustls logs through the `log` crate, which is fine for simple apps but doesn't hold up well in async contexts where rustls usually runs, where the `tracing` crate is a lot more popular (tokio, hyper, reqwest, tonic). 

Switch rustls from `log` to `tracing`, using `tracing`'s `"log"` feature so it's not a breaking change:

```toml
[dependencies]
tracing = { version = "0.1", features = ["log"] }
```

With `"log"` enabled, `tracing` falls back to emitting `log` records when there's no `tracing` subscriber, so existing users with a `log`-only backend wouldn't notice anything changed.

The actual payoff is being able to add spans, for instance one per handshake, around individual flights, around cert validation and so forth. That could be a follow-up after the initial swap.

- **`tracing-log` bridge (current recommendation)**: What maintainers have suggested in #2330. It gets the events into a `tracing` subscriber but with limited span context AFAIK.
- **Depend on both `log` and `tracing`**: More complexity, doesn't solve the span problem, don't think it's worth it.
- **Drop `log` entirely**: Breaking change, needs a major bump. The `"log"` feature avoids that.

Happy to put up a draft PR if there's interest :)

FalkWoldmann:
Closes #3122.

Not sure if I should keep the `log` shim

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,6 +93,7 @@ subtle = { version = "2.5.0", default-features = false }
 time = { version = "0.3.6", default-features = false }
 tikv-jemallocator = "0.7"
 tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"] }
+tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
 webpki = { package = "rustls-webpki", version = "=0.104.0-alpha.7", features = [
   "alloc",
 ], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,7 +93,7 @@ subtle = { version = "2.5.0", default-features = false }
 time = { version = "0.3.6", default-features = false }
 tikv-jemallocator = "0.7"
 tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"] }
-tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
+tracing = { version = "0.1", default-features = false }
 webpki = { package = "rustls-webpki", version = "=0.104.0-alpha.7", features = [
   "alloc",
 ], default-features = false }
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -14,9 +14,9 @@ autotests = false
 exclude = ["src/testdata", "tests/**"]
 
 [features]
-default = ["log", "webpki"]
+default = ["tracing", "webpki"]
 brotli = ["dep:brotli", "dep:brotli-decompressor"]
-log = ["dep:tracing"]
+tracing = ["dep:tracing"]
 webpki = ["dep:webpki"]
 zlib = ["dep:zlib-rs"]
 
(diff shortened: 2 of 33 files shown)


Cargo.toml
```diff
@@ -90,6 +90,7 @@ subtle = { version = "2.5.0", default-features = false }
 time = { version = "0.3.6", default-features = false }
 tikv-jemallocator = "0.7"
 tokio = { version = "1.34", features = ["io-util", "macros", "net", "rt"] }
+tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
```
djc: Do we actually need to enable `std`? What for?

I think we should not enable `log` here, but leave that for to downstream applications.

rustls/Cargo.toml
```diff
@@ -16,18 +16,18 @@ exclude = ["src/testdata", "tests/**"]
 [features]
 default = ["log", "webpki"]
 brotli = ["dep:brotli", "dep:brotli-decompressor"]
-log = ["dep:log"]
+log = ["dep:tracing"]
```
djc: Suggest that we rename the feature to `tracing` as well, in a separate commit.

rustls/src/lib.rs
```diff
@@ -318,10 +320,8 @@ use crate::crypto::CryptoProvider;
 #[allow(unused_extern_crates)]
 extern crate test;
 
-// log for logging (optional).
 #[cfg(feature = "log")]
-#[expect(clippy::single_component_path_imports)]
-use log;
+use tracing as log;
```
djc: I think this should be come `tracing`, as well as the module below, but that should happen in a separate commit.

djc (commented):
Nice!

djc:
> Not sure if I should keep the `log` shim

As I understand it tracing has a way of effectively disabling logging by setting a static log level, so perhaps that means we don't need it anymore? @ctz thoughts?

rustls/Cargo.toml
FalkWoldmann: replace rename or add an additional feature?

djc: I think it should be renamed.

djc:
(Don't forget to update `fuzz/Cargo.lock`.)

rustls/Cargo.toml
ctz: Yep, I think a feature which controls a dependency should be named the same as the dependency. Please observe the top level "Crate features" documentation when doing this change.

Cargo.toml
FalkWoldmann: Not sure. We need at least `liballoc`. Should I remove the feature? Not enabling `log` would mean a breaking change for consumers that rely on `log`, no?

djc: It's fine for this to be a breaking change IMO.

Why do we need `alloc`? We should enable only what we really need.

djc:
Please rebase this so that only the two initial commits remain -- other changes should be squashed into the relevant originating commit.

Cargo.toml
FalkWoldmann: Not sure tbh. At least this is stated in the docs: https://docs.rs/tracing-core/latest/tracing_core/#crate-feature-flags
> Note:tracing-core’s no_std support requires liballoc.

Having some look into the code I think the `std` feature is only needed for some specialized use cases but I'm definitely no expert on the crate!
 If a consumer of `rustls` needs these, they can always just include tracing themself and enable the `std` feature.

ctz (commented):
Seems agreeable but a couple of CI issues.

Issue #3119, opened 2026-07-04
Dynamically set ECH configuration


Oakchris1955:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
While it is possible to enable ECH with the `ConfigBuilder::with_ech` method, once a `ClientConfig` has been constructed, that configuration cannot be changed. I have opened an issue over at seanmonstar/reqwest#3035 and this inability of dynamically providing an ECH configuration (dynamically as in with a `ClientConfig` struct) is currently preventing us from supporting ECH, since in `reqwest`, a HTTP `Client` may be used for connections to URLs with different domains.

**Describe the solution you'd like**
Publicly expose the `ech_mode` field of the `ClientConfig` struct (may not be possible due to ECH requiring a TLS version of 1.3 or newer and this might conflict with any previously-defined configuration)

**Describe alternatives you've considered**
Create a new method for `ClientConfig` that sets the `ech_mode` private field of the struct, raising an error if the `ClientConfig` uses an unsupported TLS version (< 1.3).

**Additional context**
Related reqwest issue: seanmonstar/reqwest#3035

djc:
I think we could do something similar here like we did for ALPN?

- https://github.com/rustls/rustls/pull/2438

(For 0.24, not sure if it makes sense to keep the config-level ECH setup or if that should always be per-connection.)

ctz:
I think, on balance, that `hpke_suites: &[&'static dyn Hpke],` should be in `ClientConfig` but everything else should be deferred to connection construction.

djc:
@Oakchris1955 would you be able to submit a PR? (`main` currently targets 0.24, but you can backport to `rel-0.23` if that's useful.)

Oakchris1955:
> I think, on balance, that `hpke_suites: &[&'static dyn Hpke],` should be in `ClientConfig` but everything else should be deferred to connection construction.

Is there any advantage this design would have over providing a HPKE suite on connection construction or would it add unnecessary complexity?

djc:
The selection of HPKE suites with an implementation available in a program won't change over the lifetime of a process, so keeping that in the `ClientConfig` makes for better ergonomics and (very slightly) better performance.

Oakchris1955:
[...]

PR #3136, opened 2026-07-11
Dynamically set ECH config

Oakchris1955:
Related to #3119 

I've introduced some new methods and enums: `with_ech_hpke_suites` on `ConfigBuilder` to provide a list of supported HPKE suites and signal the use of ECH (use TLS 1.3 or error out, etc.). `EchMode` has be kept and can be passed to `ClientConnectionBuilder` using the `with_ech_mode` method (any suites in there will override what has been set by `with_ech_hpke_suites`, so someone can pass an empty array to `with_ech_hpke_suites` and then pass the suite of their choice using this method. On the other hand, if these suites stay the same for the lifetime of a program (as they should for most of them), they can be passed using `with_ech_hpke_suites` and then these standalone parameters (`EchParams`) can be passed to `ClientConnectionBuilder` using the `with_ech_params` method.

Notes:
- I implemented a simple `random_usize_range` function to randomly obtain a `usize` within a range which uses [rejection sampling](https://dimitri.xyz/random-ints-from-random-bits/), however I worry that the running time from it might be pretty bad. Originally I had used a remainder division method, but this can introduce a bias towards small numbers, which becomes worse as the upper limit of the range increases. Perhaps a hybrid approach would be better?
- `EchParams` is a name I chose to merely distinguish it from `EchMode`, feel free to change it to a more descriptive one.
- There was a suggestion in #3119 to backport these changes to the 0.23 branch, but since these changes are breaking, I doubt that can be done without violating semantic versioning.

### DIFF
--- a/bogo/src/main.rs
+++ b/bogo/src/main.rs
@@ -7,6 +7,7 @@
 use core::any::Any;
 use core::fmt::{Debug, Formatter};
 use core::hash::Hasher;
+use core::slice;
 use std::borrow::Cow;
 use std::io::{self, Read, Write};
 use std::sync::{Arc, Mutex};
@@ -19,11 +20,10 @@ use nix::sys::signal::{self, Signal};
 use nix::unistd::Pid;
 use rustls::client::danger::{HandshakeSignatureValid, ServerIdentity, ServerVerifier};
 use rustls::client::{
-    self, ClientConfig, ClientConnection, ClientSessionKey, CredentialRequest, EchConfig,
-    EchGreaseConfig, EchMode, EchStatus, Resumption, Tls12Resumption, Tls13Session,
-    WebPkiServerVerifier,
+    self, ClientConfig, ClientConnection, ClientSessionKey, CredentialRequest, EchStatus,
+    Resumption, Tls12Resumption, Tls13Session, WebPkiServerVerifier,
 };
-use rustls::crypto::hpke::{Hpke, HpkePublicKey};
+use rustls::crypto::hpke::Hpke;
 use rustls::crypto::kx::NamedGroup;
 use rustls::crypto::{
     Credentials, CryptoProvider, Identity, SelectedCredential, SignatureScheme, Signer, SigningKey,
@@ -112,10 +112,19 @@ pub fn main() {
                     .unwrap()
                     .to_owned();
                 let mut output = Vec::new();
-                let sess = config
-                    .connect(server_name)
-                    .build(&mut output)
-                    .unwrap();
+                let conn_cfg = config.connect(server_name);
+
+                let conn_cfg = if let Some(ech_config_list) = &opts.ech_config_list {
+                    conn_cfg
+                        .with_ech(slice::from_ref(ech_config_list))
+                        .unwrap_or_else(|_| quit(":INVALID_ECH_CONFIG_LIST:"))
+                } else if opts.enable_ech_grease {
+                    conn_cfg.with_ech_grease().unwrap()
+                } else {
+                    conn_cfg
+                };
+
+                let sess = conn_cfg.build(&mut output).unwrap();
                 exec(&opts, sess, output, &key_log, i);
             }
             SideConfig::Server(config) => {
@@ -1867,19 +1876,10 @@ fn make_client_cfg(opts: &Options, key_log: &Arc<KeyLogMemo>) -> Arc<ClientConfi
             .into(),
         );
 
-        if let Some(ech_config_list) = &opts.ech_config_list {
-            let ech_mode: EchMode = EchConfig::new(ech_config_list.clone(), ALL_HPKE_SUITES)
-                .unwrap_or_else(|_| quit(":INVALID_ECH_CONFIG_LIST:"))
-                .into();
-
-            ech_cfg.with_ech(ech_mode)
+        if opts.ech_config_list.is_some() {
+            ech_cfg.with_ech_hpke_suites(ALL_HPKE_SUITES)
         } else if opts.enable_ech_grease {
-            let ech_mode = EchMode::Grease(EchGreaseConfig::new(
-                GREASE_HPKE_SUITE,
-                HpkePublicKey(GREASE_25519_PUBKEY.to_vec()),
-            ));
-
-            ech_cfg.with_ech(ech_mode)
+            ech_cfg.with_ech_hpke_suites(&[GREASE_HPKE_SUITE])
         } else {
             cfg
         }
@@ -2400,11 +2400,6 @@ impl compress::CertCompressor for RandomAlgorithm {
 }
 
 static GREASE_HPKE_SUITE: &dyn Hpke = hpke::DH_KEM_X25519_HKDF_SHA256_AES_128;
-
-const GREASE_25519_PUBKEY: &[u8] = &[
-    0x67, 0x35, 0xCA, 0x50, 0x21, 0xFC, 0x4F, 0xE6, 0x29, 0x3B, 0x31, 0x2C, 0xB5, 0xE0, 0x97, 0xD8,
-    0xD0, 0x58, 0x97, 0xCF, 0x5C, 0x15, 0x12, 0x79, 0x4B, 0xEF, 0x1D, 0x98, 0x52, 0x74, 0xDC, 0x5E,
-];
 
 // nb. hpke::ALL_SUPPORTED_SUITES omits fips-incompatible options,
 // this includes them. bogo fips tests are activated by -fips-202205
--- a/examples/src/bin/ech-client.rs
+++ b/examples/src/bin/ech-client.rs
@@ -41,7 +41,7 @@ use hickory_resolver::net::runtime::TokioRuntimeProvider;
 use hickory_resolver::proto::rr::rdata::svcb::{SvcParamKey, SvcParamValue};
 use hickory_resolver::proto::rr::{RData, RecordType};
 use hickory_resolver::{Resolver, TokioResolver};
-use rustls::client::{EchConfig, EchGreaseConfig, EchMode, EchStatus};
+use rustls::client::EchStatus;
 use rustls::crypto::hpke::Hpke;
 use rustls::pki_types::pem::PemObject;
 use rustls::pki_types::{CertificateDer, EchConfigListBytes, ServerName};
@@ -85,17 +85,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
         .with_ansi(stderr().is_terminal())
         .init();
 
-    let ech_mode = match server_ech_configs.is_empty() {
-        false => EchMode::from(
-            server_ech_configs
-                .into_iter()
-                .find_map(|list| EchConfig::new(list, ALL_SUPPORTED_SUITES).ok())
-                .ok_or("no supported ECH configs")?,
-        ),
-        true => {
-            let (public_key, _) = GREASE_HPKE_SUITE.generate_key_pair()?;
-            EchMode::from(EchGreaseConfig::new(GREASE_HPKE_SUITE, public_key))
-        }
+    let ech_hpke_suites: &[&dyn Hpke] = match server_ech_configs.is_empty() {
+        false => ALL_SUPPORTED_SUITES,
+        true => &[GREASE_HPKE_SUITE],
     };
 
     let root_store = match args.cafile {
@@ -116,7 +108,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Construct a rustls client config with a TLS1.3-only provider, and ECH enabled.
     #[cfg_attr(not(debug_assertions), expect(unused_mut))]
     let mut config = ClientConfig::builder(rustls_aws_lc_rs::DEFAULT_TLS13_PROVIDER.into())
-        .with_ech(ech_mode)
+        .with_ech_hpke_suites(ech_hpke_suites)
         .with_root_certificates(root_store)
         .with_no_client_auth()?;
 
@@ -133,9 +125,17 @@ async fn main() -> Result<(), Box<dyn Error>> {
     for i in 0..args.num_reqs {
         trace!("\nRequest {} of {}", i + 1, args.num_reqs);
         let mut output = Vec::new();
-        let mut conn = config
-            .connect(server_name.clone())
-            .build(&mut output)?;
+        let mut conn = config.connect(server_name.clone());
+
+        conn = if !server_ech_configs.is_empty() {
+            conn.with_ech(&server_ech_configs)
+                .expect("no compatible ECH configuration found")
+        } else {
+            conn.with_ech_grease()
+                .expect("Hpke suite couldn't generate placeholder public key")
+        };
+
+        let mut conn = conn.build(&mut output)?;
         // The "outer" server that we're connecting to.
         let sock_addr = (args.outer_hostname.as_str(), args.port)
             .to_socket_addrs()?
--- a/rustls/src/builder.rs
+++ b/rustls/src/builder.rs
@@ -2,8 +2,8 @@ use alloc::format;
 use core::fmt;
 use core::marker::PhantomData;
 
-use crate::client::EchMode;
 use crate::crypto::CryptoProvider;
+use crate::crypto::hpke::Hpke;
 use crate::sync::Arc;
 use crate::time_provider::TimeProvider;
 #[cfg(doc)]
@@ -158,7 +158,7 @@ impl<Side: ConfigSide, State: fmt::Debug> fmt::Debug for ConfigBuilder<Side, Sta
 /// For more information, see the [`ConfigBuilder`] documentation.
 #[derive(Clone, Debug)]
 pub struct WantsVerifier {
-    pub(crate) client_ech_mode: Option<EchMode>,
+    pub(crate) client_ech_hpke_suites: Arc<[&'static dyn Hpke]>,
 }
 
 /// Helper trait to abstract [`ConfigBuilder`] over building a [`ClientConfig`] or [`ServerConfig`].
--- a/rustls/src/client/config.rs
+++ b/rustls/src/client/config.rs
@@ -1,14 +1,13 @@
 use alloc::vec::Vec;
 use core::any::Any;
-use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::marker::PhantomData;
+use core::{fmt, iter};
 
 #[cfg(feature = "webpki")]
 use pki_types::PrivateKeyDer;
 use pki_types::{FipsStatus, ServerName, UnixTime};
 
-use super::ech::EchMode;
 use super::handy::{ClientSessionMemoryCache, FailResolveClientCert, NoClientSessionStorage};
 use super::{Tls12Session, Tls13Session};
 use crate::builder::{ConfigBuilder, WantsVerifier};
@@ -16,6 +15,7 @@ use crate::client::connection::ClientConnectionBuilder;
 use crate::common_state::Protocol;
 #[cfg(doc)]
 use crate::crypto;
+use crate::crypto::hpke::Hpke;
 use crate::crypto::kx::NamedGroup;
 use crate::crypto::{CipherSuite, CryptoProvider, SelectedCredential, SignatureScheme, hash};
 #[cfg(feature = "webpki")]
@@ -40,12 +40,6 @@ use crate::{DistinguishedName, DynHasher, KeyLog, compress};
 ///
 /// These must be created via the [`ClientConfig::builder()`] or [`ClientConfig::builder_with_details()`]
 /// function.
-///
-/// Note that using [`ConfigBuilder<ClientConfig, WantsVersions>::with_ech()`] will produce a common
-/// configuration specific to the provided [`crate::client::EchConfig`] that may not be appropriate
-/// for all connections made by the program. In this case the configuration should only be shared
-/// by connections intended for domains that offer the provided [`crate::client::EchConfig`] in
-/// their DNS zone.
 ///
 /// # Defaults
 ///
@@ -183,8 +177,11 @@ pub struct ClientConfig {
     /// a cache that does no caching.
     pub cert_compression_cache: Arc<compress::CompressionCache>,
 
-    /// How to offer Encrypted Client Hello (ECH). The default is to not offer ECH.
-    pub(super) ech_mode: Option<EchMode>,
+    /// List of supported HPKE suites to use when offering Encrypted Client Hello (ECH).
+    ///
+    /// This is optional: If this is empty, attempting to configure ECH will result
+    /// in a [`ApiMisuse::NoEchHpkeSuites`] error.
+    pub(super) ech_hpke_suites: Arc<[&'static dyn Hpke]>,
 }
 
 impl ClientConfig {
@@ -212,7 +209,7 @@ impl ClientConfig {
     ) -> ConfigBuilder<Self, WantsVerifier> {
         ConfigBuilder {
             state: WantsVerifier {
-                client_ech_mode: None,
+                client_ech_hpke_suites: Arc::default(),
             },
             provider,
             time_provider,
@@ -229,6 +226,7 @@ impl ClientConfig {
             config: self.clone(),
             name: server_name,
             alpn_protocols: None,
+            ech_mode: None,
         }
     }
 
@@ -242,17 +240,21 @@ impl ClientConfig {
     ///
     /// This is different from [`CryptoProvider::fips()`]: [`CryptoProvider::fips()`]
     /// is concerned only with cryptography, whereas this _also_ covers TLS-level
-    /// configuration that NIST recommends, as well as ECH HPKE suites if applicable.
+    /// configuration that NIST recommends.
+    ///
+    /// ECH HPKE suites' FIPS status will be taken into account: the minimum FIPS status
+    /// of both the [`CryptoProvider`] and of the ECH HPKE suites will be returned.
     pub fn fips(&self) -> FipsStatus {
         if !self.require_ems {
             return FipsStatus::Unvalidated;
         }
 
-        let status = self.domain.provider.fips();
-        match &self.ech_mode {
-            Some(ech) => Ord::min(status, ech.fips()),
-            None => status,
-        }
+        self.ech_hpke_suites
+            .iter()
+            .map(|suite| suite.fips())
+            .chain(iter::once(self.domain.provider.fips()))
+            .min()
+            .unwrap()
     }
 
     /// Return the crypto provider used to construct this client configuration.
@@ -675,7 +677,7 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
         ConfigBuilder {
             state: WantsClientCert {
                 verifier,
-                client_ech_mode: self.state.client_ech_mode,
+                client_ech_hpke_suites: self.state.client_ech_hpke_suites,
             },
             provider: self.provider,
             time_provider: self.time_provider,
@@ -683,18 +685,15 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
         }
     }
 
-    /// Enable Encrypted Client Hello (ECH) in the given mode.
+    /// Pass a slice of HPKE suites to use in case ECH is configured later on
+    /// with [`ClientConnectionBuilder::with_ech`] or [`ClientConnectionBuilder::with_ech_grease`].
     ///
     /// This requires TLS 1.3 as the only supported protocol version to meet the requirement
     /// to support ECH.  At the end, the config building process will return an error if either
     /// TLS1.3 _is not_ supported by the provider, or TLS1.2 _is_ supported.
-    ///
-    /// The `ClientConfig` that will be produced by this builder will be specific to the provided
-    /// [`crate::client::EchConfig`] and may not be appropriate for all connections made by the program.
-    /// In this case the configuration should only be shared by connections intended for domains
-    /// that offer the provided [`crate::client::EchConfig`] in their DNS zone.
-    pub fn with_ech(mut self, mode: EchMode) -> Self {
-        self.state.client_ech_mode = Some(mode);
+    pub fn with_ech_hpke_suites(mut self, hpke_suites: &[&'static dyn Hpke]) -> Self {
+        self.state.client_ech_hpke_suites = Arc::from(hpke_suites);
+
         self
     }
 
@@ -712,7 +711,7 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
 #[derive(Clone)]
 pub struct WantsClientCert {
     verifier: Arc<dyn ServerVerifier>,
-    client_ech_mode: Option<EchMode>,
+    client_ech_hpke_suites: Arc<[&'static dyn Hpke]>,
 }
 
 impl ConfigBuilder<ClientConfig, WantsClientCert> {
@@ -747,7 +746,11 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
     ) -> Result<ClientConfig, Error> {
         self.provider.consistency_check()?;
 
-        if self.state.client_ech_mode.is_some() {
+        if !self
+            .state
+            .client_ech_hpke_suites
+            .is_empty()
+        {
             match (
                 self.provider
                     .tls12_cipher_suites
@@ -783,7 +786,7 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
             cert_decompressors: compress::default_cert_decompressors().to_vec(),
             cert_compressors: compress::default_cert_compressors().to_vec(),
             cert_compression_cache: Arc::new(compress::CompressionCache::default()),
-            ech_mode: self.state.client_ech_mode,
+            ech_hpke_suites: self.state.client_ech_hpke_suites,
         })
     }
 }
@@ -828,7 +831,7 @@ pub(super) mod danger {
             ConfigBuilder {
                 state: WantsClientCert {
                     verifier,
-                    client_ech_mode: self.cfg.state.client_ech_mode,
+                    client_ech_hpke_suites: self.cfg.state.client_ech_hpke_suites,
                 },
                 provider: self.cfg.provider,
                 time_provider: self.cfg.time_provider,
--- a/rustls/src/client/connection.rs
+++ b/rustls/src/client/connection.rs
@@ -2,12 +2,13 @@ use alloc::vec::Vec;
 use core::fmt;
 use core::ops::Deref;
 
-use pki_types::{FipsStatus, ServerName};
+use pki_types::{EchConfigListBytes, FipsStatus, ServerName};
 
 use super::config::ClientConfig;
 use super::hs::ClientHelloInput;
 use crate::TlsInputBuffer;
 use crate::client::EchStatus;
+use crate::client::ech::{EchConfig, EchMode};
 use crate::common_state::{CommonState, ConnectionOutputs, EarlyDataEvent, Event, Protocol, Side};
 use crate::conn::private::SideOutput;
 use crate::conn::split::SplitConnection;
@@ -19,7 +20,7 @@ use crate::conn::{
 use crate::crypto;
 use crate::crypto::cipher::OutboundPlain;
 use crate::enums::ApplicationProtocol;
-use crate::error::Error;
+use crate::error::{ApiMisuse, Error, RejectedEch};
 use crate::msgs::ClientExtensionsInput;
 use crate::quic::QuicOutput;
 use crate::suites::ExtractedSecrets;
@@ -165,10 +166,12 @@ impl Deref for ClientConnection {
 /// Builder for [`ClientConnection`] values.
 ///
 /// Create one with [`ClientConfig::connect()`].
+#[derive(Debug)]
 pub struct ClientConnectionBuilder {
     pub(crate) config: Arc<ClientConfig>,
     pub(crate) name: ServerName<'static>,
     pub(crate) alpn_protocols: Option<Vec<ApplicationProtocol<'static>>>,
+    pub(crate) ech_mode: Option<EchMode>,
 }
 
 impl ClientConnectionBuilder {
@@ -178,12 +181,97 @@ impl ClientConnectionBuilder {
         self
     }
 
+    /// Provide a slice of [`EchConfigListBytes`] configurations to use to
+    /// connect using ECH. The slice's elements will be iterated until an ECH
+    /// configuration that is compatible with the HPKE provider's supported
+    /// suites is found.
+    ///
+    /// # Errors
+    ///
+    /// One of the provided ECH configurations must be compatible with the HPKE provider's supported
+    /// suites or an [`EncryptedClientHelloError::NoCompatibleConfig`](crate::error::EncryptedClientHelloError::NoCompatibleConfig)
+    /// error will be returned.
+    ///
+    /// If no ECH HPKE suites were provided with
+    /// [`ConfigBuilder::with_ech_hpke_suites`](crate::builder::ConfigBuilder::with_ech_hpke_suites),
+    /// [`ApiMisuse::NoEchHpkeSuites`] will be returned instead.
+    pub fn with_ech(
+        mut self,
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
+    ) -> Result<Self, Error> {
+        if self.config.ech_hpke_suites.is_empty() {
+            return Err(Error::ApiMisuse(ApiMisuse::NoEchHpkeSuites));
+        }
+
+        self.ech_mode = Some(EchMode::from_ech_config_list(
+            ech_config_list_slice,
+            &self.config.ech_hpke_suites,
+        )?);
+
+        Ok(self)
+    }
+
+    /// Configure the builder to use ECH GREASE.
+    ///
+    /// ECH GREASE is a mechanism to make a non-ECH connection appear as if it is one.
+    /// This way, the extent to which ECH connections stick out is reduced and
+    /// thus, network ossification is mitigated.
+    ///
+    /// If your client uses ECH but doesn't have a configuration for a server it
+    /// wants to connect to, it is recommended per ECH's RFC 9849 to instead
+    /// GREASE the connection for the reasons listed above.
+    ///
+    /// # Note
+    ///
+    /// In order to not stick out, this method will use the first one of the
+    /// provided HPKE and stick with it. ECH's RFC 9849 says that "The selection SHOULD vary",
+    /// however, since both Google's BoringSSL & Mozilla's NSS apparently don't
+    /// do that, we follow similar behavior in order to blend in with them.
+    ///
+    /// # Errors
+    ///
+    /// This method will error if the HPKE provider fails to generate a placeholder public key.
+    ///
+    /// If no ECH HPKE suites were provided with
+    /// [`ConfigBuilder::with_ech_hpke_suites`](crate::builder::ConfigBuilder::with_ech_hpke_suites),
+    /// [`ApiMisuse::NoEchHpkeSuites`] will be returned instead.
+    pub fn with_ech_grease(mut self) -> Result<Self, Error> {
+        if self.config.ech_hpke_suites.is_empty() {
+            return Err(Error::ApiMisuse(ApiMisuse::NoEchHpkeSuites));
+        }
+
+        // just pick the first HPKE suite and stick with it
+        //
+        // the RFC says that the suites should vary to prevent fingerprinting,
+        // but both BoringSSL + NSS seem to do what we do here, so in order to
+        // not stick out, we just copy that behavior
+        self.ech_mode = Some(EchMode::grease_from_suite(self.config.ech_hpke_suites[0])?);
+
+        Ok(self)
+    }
+
+    /// Retrying ECH using a retry config from a server's previous rejection
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the server provided no retry configurations in [`RejectedEch`], or if
+    /// none of the retry configurations are compatible with the HPKE provider's supported suites.
+    pub fn with_ech_for_retry(mut self, rejection: RejectedEch) -> Result<Self, Error> {
+        self.ech_mode = Some(EchMode::Enable(EchConfig::for_retry(
+            rejection,
+            &self.config.ech_hpke_suites,
+        )?));
+
+        Ok(self)
+    }
+
     /// Finalize the builder and create the `ClientConnection`.
     pub fn build(self, tls: &mut Vec<u8>) -> Result<ClientConnection, Error> {
         let Self {
             config,
             name,
             alpn_protocols,
+            ech_mode,
         } = self;
 
         let alpn_protocols = alpn_protocols.unwrap_or_else(|| config.alpn_protocols.clone());
@@ -194,6 +282,7 @@ impl ClientConnectionBuilder {
                 ClientExtensionsInput::from_alpn(alpn_protocols),
                 None,
                 Protocol::Tcp,
+                ech_mode,
                 tls,
             )?,
         })
@@ -269,6 +358,7 @@ impl ConnectionCommon<ClientSide> {
         extra_exts: ClientExtensionsInput,
         quic: Option<&mut dyn QuicOutput>,
         protocol: Protocol,
+        ech_mode: Option<EchMode>,
         tls: &mut Vec<u8>,
     ) -> Result<Self, Error> {
         let mut common_state = CommonState::new(Side::Client, config.fips());
@@ -284,7 +374,8 @@ impl ConnectionCommon<ClientSide> {
             tls,
         };
 
-        let input = ClientHelloInput::new(name, &extra_exts, protocol, &mut output, config)?;
+        let input =
+            ClientHelloInput::new(name, &extra_exts, protocol, &mut output, config, ech_mode)?;
         let state = input.start_handshake(extra_exts, &mut output)?;
 
         Ok(Self::new(state, data, common_state))
--- a/rustls/src/client/ech.rs
+++ b/rustls/src/client/ech.rs
@@ -3,7 +3,7 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::iter;
 
-use pki_types::{DnsName, EchConfigListBytes, FipsStatus, ServerName};
+use pki_types::{DnsName, EchConfigListBytes, ServerName};
 use subtle::ConstantTimeEq;
 
 use super::config::ClientConfig;
@@ -35,7 +35,7 @@ use crate::tracing::{debug, trace, warn};
 /// Controls how Encrypted Client Hello (ECH) is used in a client handshake.
 #[non_exhaustive]
 #[derive(Clone, Debug)]
-pub enum EchMode {
+pub(crate) enum EchMode {
     /// ECH is enabled and the ClientHello will be encrypted based on the provided
     /// configuration.
     Enable(EchConfig),
@@ -48,12 +48,45 @@ pub enum EchMode {
 }
 
 impl EchMode {
-    /// Returns true if the ECH mode will use a FIPS approved HPKE suite.
-    pub fn fips(&self) -> FipsStatus {
-        match self {
-            Self::Enable(ech_config) => ech_config.suite.fips(),
-            Self::Grease(grease_config) => grease_config.suite.fips(),
+    /// Create a new [`EchMode`] from a list of supported suites and a list of
+    /// ECH configurations to try.
+    ///
+    /// Each suite-configuration pair will be tried, beginning at the first
+    /// ECH configuration, until a valid [`EchConfig`] is constructed
+    ///
+    /// # Errors
+    ///
+    /// If all pairs have been exhausted, an [`EncryptedClientHelloError::NoCompatibleConfig`]
+    /// error will be raised
+    pub(crate) fn from_ech_config_list(
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
+        suites: &[&'static dyn Hpke],
+    ) -> Result<Self, Error> {
+        for ech_config_list in ech_config_list_slice {
+            match EchConfig::new(ech_config_list, suites) {
+                Ok(ech_config) => return Ok(ech_config.into()),
+                Err(Error::InvalidEncryptedClientHello(
+                    EncryptedClientHelloError::NoCompatibleConfig,
+                )) => {
+                    // silently reject NoCompatibleConfig errors
+                }
+                Err(err) => return Err(err),
+            }
         }
+
+        // if we have yet to return, no compatible ECH config has been found
+        Err(Error::InvalidEncryptedClientHello(
+            EncryptedClientHelloError::NoCompatibleConfig,
+        ))
+    }
+
+    /// Create a new GREASE [`EchMode`] from a supported suite.
+    ///
+    /// # Errors
+    ///
+    /// This method will error if the HPKE provider fails to generate a placeholder public key.
+    pub(crate) fn grease_from_suite(suite: &'static dyn Hpke) -> Result<Self, Error> {
+        Ok(EchGreaseConfig::new(suite)?.into())
     }
 }
 
@@ -73,7 +106,7 @@ impl From<EchGreaseConfig> for EchMode {
 ///
 /// Note: differs from the protocol-encoded EchConfig (`EchConfigMsg`).
 #[derive(Clone, Debug)]
-pub struct EchConfig {
+pub(crate) struct EchConfig {
     /// The selected EchConfig.
     pub(crate) config: EchConfigPayload,
 
@@ -92,16 +125,16 @@ impl EchConfig {
     /// be base64 decoded to yield the `EchConfigListBytes` you provide to rustls.
     ///
     /// One of the provided ECH configurations must be compatible with the HPKE provider's supported
-    /// suites or an error will be returned.
+    /// suites or an [`EncryptedClientHelloError::InvalidConfigList`] error will be returned.
     ///
     /// See the [`ech-client.rs`] example for a complete example of fetching ECH configs from DNS.
     ///
     /// [`ech-client.rs`]: https://github.com/rustls/rustls/blob/main/examples/src/bin/ech-client.rs
-    pub fn new(
-        ech_config_list: EchConfigListBytes<'_>,
+    pub(crate) fn new(
+        ech_config_list: &EchConfigListBytes<'_>,
         hpke_suites: &[&'static dyn Hpke],
     ) -> Result<Self, Error> {
-        let ech_configs = Vec::<EchConfigPayload>::read_bytes(&ech_config_list).map_err(|_| {
+        let ech_configs = Vec::<EchConfigPayload>::read_bytes(ech_config_list).map_err(|_| {
             Error::InvalidEncryptedClientHello(EncryptedClientHelloError::InvalidConfigList)
         })?;
 
@@ -112,7 +145,7 @@ impl EchConfig {
     ///
     /// Returns an error if the server provided no retry configurations in `RejectedEch`, or if
     /// none of the retry configurations are compatible with the supported `hpke_suites`.
-    pub fn for_retry(
+    pub(crate) fn for_retry(
         rejection: RejectedEch,
         hpke_suites: &[&'static dyn Hpke],
     ) -> Result<Self, Error> {
@@ -201,7 +234,7 @@ impl EchConfig {
 
 /// Configuration for GREASE Encrypted Client Hello.
 #[derive(Clone, Debug)]
-pub struct EchGreaseConfig {
+pub(crate) struct EchGreaseConfig {
     pub(crate) suite: &'static dyn Hpke,
     pub(crate) placeholder_key: HpkePublicKey,
 }
@@ -213,14 +246,16 @@ impl EchGreaseConfig {
     /// but doesn't have a real ECH configuration to use for the remote server. In this case
     /// a placeholder or "GREASE"[^0] extension is used.
     ///
-    /// Returns an error if the HPKE provider does not support the given suite.
+    /// # Errors
+    ///
+    /// Returns an error if the HPKE provider fails to generate a placeholder public key.
     ///
     /// [^0]: <https://www.rfc-editor.org/rfc/rfc8701>
-    pub fn new(suite: &'static dyn Hpke, placeholder_key: HpkePublicKey) -> Self {
-        Self {
+    pub(crate) fn new(suite: &'static dyn Hpke) -> Result<Self, Error> {
+        Ok(Self {
             suite,
-            placeholder_key,
-        }
+            placeholder_key: suite.generate_key_pair()?.0,
+        })
     }
 
     /// Build a GREASE ECH extension based on the placeholder configuration.
@@ -996,20 +1031,7 @@ mod tests {
         provider.secure_random = &CountingRandom;
 
         let config = ClientConfig::builder(Arc::new(provider))
-            .with_ech(EchMode::Enable(EchConfig {
-                config: EchConfigPayload::V18(EchConfigContents {
-                    key_config: HpkeKeyConfig {
-                        config_id: 0,
-                        kem_id: MockHpke::SUITE.kem,
-                        public_key: vec![0; 32].into(),
-                        symmetric_cipher_suites: vec![MockHpke::SUITE.sym],
-                    },
-                    maximum_name_length: 255,
-                    public_name: DnsName::try_from("public.example.com").unwrap(),
-                    extensions: vec![],
-                }),
-                suite: &MockHpke,
-            }))
+            .with_ech_hpke_suites(&[&MockHpke])
             .with_root_certificates(roots)
             .with_no_client_auth()
             .unwrap();
@@ -1048,6 +1070,24 @@ mod tests {
         let mut first_flight = Vec::new();
         let mut conn = config
             .connect(server_name)
+            .with_ech(&[{
+                let mut bytes = Vec::new();
+                vec![EchConfigPayload::V18(EchConfigContents {
+                    key_config: HpkeKeyConfig {
+                        config_id: 0,
+                        kem_id: MockHpke::SUITE.kem,
+                        public_key: vec![0; 32].into(),
+                        symmetric_cipher_suites: vec![MockHpke::SUITE.sym],
+                    },
+                    maximum_name_length: 255,
+                    public_name: DnsName::try_from("public.example.com").unwrap(),
+                    extensions: vec![],
+                })]
+                .encode(&mut bytes);
+
+                EchConfigListBytes::from(bytes)
+            }])
+            .unwrap()
             .build(&mut first_flight)
             .unwrap();
 
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -34,7 +34,8 @@ mod connection;
 pub use connection::{ClientConnection, ClientConnectionBuilder, ClientSide, WriteEarlyData};
 
 mod ech;
-pub use ech::{EchConfig, EchGreaseConfig, EchMode, EchStatus};
+use ech::EchMode;
+pub use ech::EchStatus;
 
 mod handy;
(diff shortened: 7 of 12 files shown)


Oakchris1955:
I got everything but the BoGo test suite working locally. Once I get that working too, I will force push to this branch.

Oakchris1955:
The tests should pass now, can someone approve the workflows?

rustls/src/builder.rs
```diff
@@ -159,7 +160,7 @@ impl<Side: ConfigSide, State: fmt::Debug> fmt::Debug for ConfigBuilder<Side, Sta
 /// For more information, see the [`ConfigBuilder`] documentation.
 #[derive(Clone, Debug)]
 pub struct WantsVerifier {
-    pub(crate) client_ech_mode: Option<EchMode>,
+    pub(crate) client_ech_hpke_suites: Option<Box<[&'static dyn Hpke]>>,
```
djc: I think we could get rid of the `Option` wrapper here in favor of having an empty slice? The `Box<[..]>` wrapper also feels a little weird here; we don't have much of that going on. I think `Arc<[..]>` would probably be better?

rustls/src/client/config.rs
```diff
@@ -670,26 +640,27 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
...
+                client_ech_hpke_suites: self.state.client_ech_hpke_suites,
             },
             provider: self.provider,
             time_provider: self.time_provider,
             side: PhantomData,
         }
     }
 
-    /// Enable Encrypted Client Hello (ECH) in the given mode.
+    /// Enable Encrypted Client Hello (ECH) in the given mode by specifying which HPKE suites to use.
```
djc: I think this docstring should probably be revised substantially to make it more about just the HPKE suites which are necessary but not sufficient to enable ECH in connections using this `ClientConfig`.

rustls/src/client/config.rs
```diff
@@ -742,7 +713,11 @@ impl ConfigBuilder<ClientConfig, WantsClientCert> {
     ) -> Result<ClientConfig, Error> {
         self.provider.consistency_check()?;
 
-        if self.state.client_ech_mode.is_some() {
+        if self
```
djc: I think this should probably be moved to the connection builder.

rustls/src/client/connection.rs
```diff
@@ -210,12 +215,49 @@ impl ClientConnectionBuilder {
         self
     }
 
+    /// Specify how to connect using ECH
+    ///
+    /// One of the provided ECH configurations must be compatible with the HPKE provider’s supported
+    /// suites or an error will be returned.
```
djc: Should mention the need for HPKE suites. I think this should probably be fallible?

rustls/src/client/connection.rs
```diff
@@ -210,12 +215,49 @@ impl ClientConnectionBuilder {
...
+    }
+
+    /// Specify how to connect using ECH. Does nothing if [`with_ech_hpke_suites`] hasn't been invoked
+    ///
+    /// Unlike [`with_ech_params`](Self::with_ech_params), this allows one to also provide their own
+    /// HPKE suite, whether that has already been provided previously
+    /// with [`with_ech_hpke_suites`] or not
+    ///
+    /// [`with_ech_hpke_suites`]: crate::ConfigBuilder::with_ech_hpke_suites
+    pub fn with_ech_mode(mut self, ech_mode: EchMode) -> Self {
```
djc: I think we should drop this?

rustls/src/client/ech.rs
```diff
@@ -33,6 +34,16 @@ use crate::tls13::key_schedule::{
...
+///
+/// If you have already specified which HPKE suites to be used for ECH through
+/// [`ConfigBuilder::with_ech_hpke_suites`], check [`EchParams`]
+///
+/// [`ClientConnectionBuilder`]: crate::client::connection::ClientConnectionBuilder
+/// [`with_ech_mode`]: crate::client::connection::ClientConnectionBuilder::with_ech_mode
+/// [`ConfigBuilder::with_ech_hpke_suites`]: crate::ConfigBuilder::with_ech_hpke_suites
 #[non_exhaustive]
 #[derive(Clone, Debug)]
 pub enum EchMode {
```
djc: Maybe this could/should become private now?

rustls/src/client/ech.rs
```diff
@@ -55,6 +66,39 @@ impl EchMode {
...
         }
     }
+
+    /// Create a new [`EchMode`] from an [`EchParams`] enum, a list of supported suites
+    /// and a source of cryptographically secure randomness.
+    ///
+    /// It is highly recommended to use this method if the `params` argument can also
+    /// be of variant [`Grease`](EchParams::Grease), since this method will securely
+    /// randomly choose a suite from the provided ones, per RFC 9849 §6.2.1
+    pub fn from_params(
```
djc: This should be private.

djc (commented):
Needs some design review on the future of `EchMode`, I think.

Oakchris1955:
Having only one way to provide an ECH config instead of two (`ECHMode` and `ECHParams`) would certainly increase code coverage and make the API more simple. I retained the ability of passing an ECH configuration through the `ECHMode` struct for easier migration, although removing it (and perhaps renaming the `ECHParams` struct to `ECHMode`) would resolve most of the comments left in the code review.

djc:
> Having only one way to provide an ECH config instead of two (`ECHMode` and `ECHParams`) would certainly increase code coverage and make the API more simple. I retained the ability of passing an ECH configuration through the `ECHMode` struct for easier migration, although removing it (and perhaps renaming the `ECHParams` struct to `ECHMode`) would resolve most of the comments left in the code review.

I don't think easier migration is a strong argument for this, but let's see what @ctz and/or @cpu think.

Oakchris1955:
Removed the whole `from_params` API, ECH configuration is now passed only through `EchParams` (and the with the `with_ech_hpke_suites` method of course). Force-pushed because I had forgotten to update some documentation comments.

Oakchris1955:
I don't want to rush this, but I believe it would be good if we could get this merged before `0.24`, since the whole API redesign is a breaking change.

djc:
If this is now ready for review, can you mark it as such?

Oakchris1955:
Of course. However, there are still a couple of issues that need to be reviewed. There are these reviews of ours that I haven't resolved yet, I am not sure if the random suite selection for GREASE is [FIPS-compliant](https://github.com/rustls/rustls/issues/3119#issuecomment-4893021596) and the random selection algorithm should probably be replaced with something more secure.

cpu:
I will try to take a pass through this week.

Oakchris1955:
Can someone approve the workflows? The tests seem to pass on my machine, but I want to make sure that I fixed the code coverage.

Oakchris1955:
Fixed clippy error

Oakchris1955:
Either I messed something up really bad or we caught an ICE.

Oakchris1955:
Nope, this is an ICE, other people have reported it too: https://github.com/rust-lang/rust/issues/159815. So, what's the plan now?

cpu:
> So, what's the plan now?

It's not a required status check in our CI configuration. The 0.24 release is also not imminent and we're not racing a clock. If you're happy with the diff & commit history the plan now is to wait patiently for review feedback or follow-up in ~a week if you haven't heard back. Likely the ICE will be fixed before then, if not we wouldn't block the merge on it.

rustls/src/client/connection.rs
Oakchris1955: > I think this should probably be fallible

I believe this is an API restriction. If the `with_*` methods weren't consumable (`mut self`) but instead borrowing (`&mut self`), that would be more easier to implement.

For example, the `ech-client.rs` example could become:

```rs
let mut conn = config.connect(server_name.clone());
ech_params.iter()
            .find_map(|ech_mode| {
                conn
                    .with_ech_params(ech_mode.clone())
                    .ok()
            }).expect("no compatible ECH configuration found").build()?;
```

Another option would be to make the method take a slice of `EchParams` and iterate through them until it finds one compatible with the HPKE provider's suites. If it doesn't find anything, then return an error (essentially move what the `ech-client.rs` example is currently doing within the `with_ech_params` method)

rustls/src/client/ech.rs
```diff
@@ -48,12 +49,37 @@ pub enum EchMode {
 }
 
 impl EchMode {
-    /// Returns true if the ECH mode will use a FIPS approved HPKE suite.
-    pub fn fips(&self) -> FipsStatus {
```
djc, line 52: Why don't we need this anymore? Presumably a connection using ECH will still need to be able to ascertain whether it's FIPS.

rustls/src/client/ech.rs
```diff
@@ -48,12 +49,37 @@ pub enum EchMode {
...
+        suites: &[&'static dyn Hpke],
+        secure_random: &dyn SecureRandom,
+    ) -> Result<Self, Error> {
+        if suites.is_empty() {
+            return Err(Error::InvalidEncryptedClientHello(
+                EncryptedClientHelloError::NoCompatibleConfig,
+            ));
+        };
+
+        let ech_mode = match params {
```
djc: Nit: no need to bind here, we can just yield `Ok(match params { .. })`.

rustls/src/client/ech.rs
```diff
@@ -69,11 +95,32 @@ impl From<EchGreaseConfig> for EchMode {
...
+    /// ECH is enabled and the ClientHello will be encrypted based on the provided
+    /// configuration.
+    Enable(EchConfigListBytes<'a>),
+
+    /// No ECH configuration is available but the client should act as though it were.
+    ///
+    /// This is an anti-ossification measure, sometimes referred to as "GREASE"[^0].
+    /// [^0]: <https://www.rfc-editor.org/rfc/rfc8701>
+    Grease(HpkePublicKey),
+}
```
djc: This seems to be somewhat unrelated to dynamically configuring ECH? Can it be in a separate commit?

rustls/src/client/mod.rs
```diff
@@ -362,15 +360,21 @@ struct ClientHelloDetails {
     sent_extensions: Vec<ExtensionType>,
     extension_order_seed: u16,
     offered_cert_compression: bool,
+    ech_info: Option<ech::EchMode>,
```
djc: Nit: import `EchMode` directly.

rustls/src/crypto/mod.rs
```diff
@@ -528,6 +530,30 @@ pub(crate) mod rand {
...
+        let new_range = range.start..end;
+
+        loop {
+            let n = usize::from_be_bytes(random_array(secure_random)?);
+
+            if new_range.contains(&n) {
+                return Ok(n % range.end);
+            }
+        }
+    }
```
djc: This should just be fixed now.

rustls/src/client/ech.rs
Oakchris1955: This is mainly intended as a way of dynamically configuring ECH by using only one method. This enum could be removed, but then there would need to be 2 methods to dynamically configure ECH, one that accepts an EchConfigListBytes and another for GREASE that accepts EchConfigListBytes.

rustls/src/crypto/mod.rs
Oakchris1955: I believe it is better to trust the experts at this and copy the [relevant function](https://docs.rs/rand/0.10.2/src/rand/distr/uniform_int.rs.html#183) from the rand crate. There is both an unbiased and a biased function that both more or less use the same algorithm, but [from what I can see](https://github.com/rust-random/rand/issues/494#issuecomment-1434997349), when it comes to the biased one, "anything involving less than ~1 year of CPU time is unaffected", so we may as well get away with it. We would have to add a notice to comply with the MIT license `rand` is licensed under to use these algorithms, but that would certainly be better than *whatever what I wrote* is.

rustls/src/client/ech.rs
Oakchris1955: This was used in `ClientConfig::fips`, but now that we don't know which one of the provided suites will be used (since that is determined later on when provided the `EchParams` struct), I modified that method so that it directly returns the provider's FIPS status. Now that I am looking into it, I should have also modified `ConnectionCore::for_client` and moved the FIPS-related logic there, while also keeping that method.

rustls/src/client/connection.rs
djc: I'm not sure what you're arguing here, can you be more explicit?

I don't think changing the builder methods to be non-consuming will be an option, but it feels like errors here will be fatal anyway so it doesn't really matter?

Oakchris1955: > it feels like errors here will be fatal anyway so it doesn't really matter

Just because an `EchParams` returns an error when passed to the `with_ech_params` method doesn't mean that that error is fatal. It merely means that the provided `EchParams` wasn't compatible with the HPKE provider's supported suites.

The version of the `ech-client` example that is currently in the main branch utilizes a smart way of iterating through all the available ECH configurations:

```rs
server_ech_configs
                .into_iter()
                .find_map(|list| EchConfig::new(list, ALL_SUPPORTED_SUITES).ok())
                .ok_or("no supported ECH configs")?
```

This is no longer possible, since the `EchConfig` struct is now visible only within the crate and the only way to pass an ECH configuration to the `ClientConnectionBuilder` is through the `with_ech_params` method. This effectively disallows a client to cycle through a list of ECH configurations until it finds a supported one, since while the `with_ech_params` method will error out if the provided one is unsupported and these errors can be handled in a manner similar to the one seen above, the `with_ech_params` method is consuming and will consume the `ClientConnectionBuilder`, leaving no choice for the client but to recreate one, no matter the time or resource cost.

To keep that functionality, we can change the `with_ech_params` method so that it instead accepts a slice of `EchMode`s and cycles through that slice's elements one-by-one until it finds a configuration that is supported by the HPKE provider's suites.

cpu:
It'd be nicer to review this sort of branch if the commit history were tidied (e.g. so style nits aren't applied separately to new code introduced in earlier commits).

I'm perhaps biased since I came up with the original API design, but I'm not a huge fan of this approach. Without going too deep on the code itself I think there are specific issues to talk about:

1. I think the way the GREASE key + suite are decoupled under this new design is incorrect. Right now the `EchParams::Grease(HpkePublicKey)` variant requires the caller to generate a public key for a specific KEM, but `EchMode::from_params()` is randomly picking a suite from the whole config list. I think this will error for any KEM mismatch at `build()` time as an opaque `Error::Other` but non-deterministically based on the selection made. We end up with a weird circularity in the API where you need to pick a suite to generate the grease pubkey but you don't know which suite the library will pick.

2. I think we have some test coverage gaps and this branch would break ECH for QUIC. [`ClientConnection::new_with_alpn()` is hard-coded to provide `None`](https://github.com/Oakchris1955/rustls/blob/eb862e6bf70903970a64888fd14101eafd77bdb0/rustls/src/quic.rs#L116) as the `ech_info: Option<EchMode>`.

3. I think it's not ideal that you can configure ECH HPKE suites on the `ClientConfig` but forget to call `with_ech_parms()` on a specific connection and the behaviour is that we silently send an unencrypted hello. Maybe when `ech_hpke_suites` are configured non-empty and the connection doesn't specify any ECH config params we should error? Or GREASE? Or something else? (It's also a bit suboptimal that you could forget to add suites to `ClientConfig`, but try to use `with_ech_params()` on a connection)

4. I think `EchGreaseConfig` is dead API surface now? It's `pub` with a `pub new` constructor but the only thing that could consume it on this branch is `pub(crate)`.

5. It feels pretty awkward that we have things split between an `EchMode` enum for Enable/Grease but then use a builder method for `with_ech_for_retry()`. Would it be better to use `with_ech()`, `with_ech_grease()` and `with_ech_for_retry()` and removing the enum?

Pulling back from the above, are we sure this is a road we need to go down? I thought the idea was that our `ClientConfig`'s are cheap to construct/clone/etc and so there's not a tremendous advantage to re-using one between domains (and IIRC there are some session store implications that might also mean it's not a great idea).

rustls/src/crypto/mod.rs
cpu: I think we're probably overthinking this. 

AIUI BoringSSL + NSS both fix the KEM/KDF and only choose between the AEAD randomly (Both switching between AES-128-GCM and ChaCha20). Probably we could do similar letting FIPS status help guide the choice?

djc:
> Pulling back from the above, are we sure this is a road we need to go down? I thought the idea was that our `ClientConfig`'s are cheap to construct/clone/etc and so there's not a tremendous advantage to re-using one between domains (and IIRC there are some session store implications that might also mean it's not a great idea).

Even if `ClientConfig`s are cheap, I feel like we've pulled back from that design a bit, for example with ALPN. Given that ECH configs are basically per-host (as I understand it), it feels weird to store a per-host config in a supposedly-wider scope `ClientConfig`.

(I obviously still agree we need to come up with better API for matching up the ECH config with the configured HPKE suites, but that feels like it's "just" API design work?)

Oakchris1955:
Replying on what @cpu said:

> It'd be nicer to review this sort of branch if the commit history were tidied (e.g. so style nits aren't applied separately to new code introduced in earlier commits).

I could rebase everything into one or multiple commits, just tell me what you believe is better.

1. Regarding GREASE, I just noticed that the Hpke traits contains a generate_key_pair method. Perhaps once a Hpke suite has been chosen, that method can be used to create a suitable public key.

2. If there are indeed coverage gaps, we could write some tests to cover them. Regarding the None variant, I am not familiar with QUIC, so I just put it there as a placeholder.

3. That API design just seemed more logical to me, but I guess the specifics of it could be further discussed.

4. I am not sure what you are arguing here, that the pub methods should be made pub(crate) or that EchGreaseConfig's functionality should be integrated elsewhere?

5. I agree here, the EchParams::Grease enum variant may currently only error if the SecureRandom source errors, while the Enable variant may only error if it hasn't found a compatible Hpke suite. This means that with the current design, the same function may return two very different errors depending only on the enum variants and I don't why this should be kept that way.

As for the ClientConfig cloning/construction argument, @djc's words are more than enough to cover my opinion too.

P.S. Regarding KEM/KDF fixing, RFC 9849 says that "The selection **SHOULD** vary". While that wording doesn't express a requirement, it reflects a strong suggestion to do so, since using the same Hpke suite over and over again may enable fingerprinting of a client.

cpu:
> I could rebase everything into one or multiple commits,

I think one commit is best. I'm not sure there are helpful intermediate bits to pull out (except perhaps for something like adding the missing ECH+QUIC test coverage up-front before doing API changes).

> If there are indeed coverage gaps, we could write some tests to cover them. Regarding the None variant, I am not familiar with QUIC, so I just put it there as a placeholder.

Fair, but ideally if you're not certain about something like this, or just using a placeholder, it would be great to call that out explicitly during self-review, or mark with a TODO. If we had merged this branch as-is, it would have broken ECH for QUIC (though again, we should have had test coverage here to help backstop the review!).

> I am not sure what you are arguing here, that the pub methods should be made pub(crate) or that EchGreaseConfig's functionality should be integrated elsewhere?

If we're keeping the current design the types should be `pub(crate)` if there's not a sensible `pub` API to use them with.

> P.S. Regarding KEM/KDF fixing, RFC 9849 says that "The selection SHOULD vary". While that wording doesn't express a requirement, it reflects a strong suggestion to do so, since using the same Hpke suite over and over again may enable fingerprinting of a client.

If the two largest browser implementations are doing something then I would be in favor of matching it whether or not the RFC says it's correct, especially if the goal is to blend in.

Oakchris1955:
> If we had merged this branch as-is, it would have broken ECH for QUIC

I normally litter my code with FIXMEs in such cases, I don't know how I glanced over that one.

> If the two largest browser implementations are doing something then I would be in favor of matching it whether or not the RFC says it's correct, especially if the goal is to blend in.

Fair, but I believe that this behavior should be mentioned somewhere, either as a comment in the source code or within the docs.

cpu:
> Regarding GREASE, I just noticed that the Hpke traits contains a generate_key_pair method. Perhaps once a Hpke suite has been chosen, that method can be used to create a suitable public key.

That might be reasonable. I can't remember if there was a specific strong reason that I wanted to allow the caller to control generation of the GREASE keypair and I believe for the supported algs the keygen step is very cheap.

Oakchris1955:
Squashed all commits into one.

rustls/src/crypto/mod.rs
Oakchris1955: Regarding that, what exactly do you mean? Select the first available Hpke suite and stick with it or use a hardcoded HPKE suite?

rustls/src/client/config.rs
```diff
@@ -238,17 +227,16 @@ impl ClientConfig {
     ///
     /// This is different from [`CryptoProvider::fips()`]: [`CryptoProvider::fips()`]
     /// is concerned only with cryptography, whereas this _also_ covers TLS-level
-    /// configuration that NIST recommends, as well as ECH HPKE suites if applicable.
+    /// configuration that NIST recommends.
+    ///
+    /// ECH HPKE suites are not applicable, since in this stage we have yet to pick
+    /// one and there may be a plethora of them to choose from.
```
ctz: This function should evaluate all suites and return the minimum. Someone who cares about the return value of this function wants absolutely zero non-FIPS cryptography that is even theoretically usable in their config.

rustls/src/client/ech.rs
```diff
@@ -47,8 +49,39 @@ pub enum EchMode {
...
+        };
+
+        Ok(match params {
+            EchParams::Enable(ech_config_list) => {
+                let ech_config = EchConfig::new(ech_config_list, suites)?;
+
+                Self::Enable(ech_config)
+            }
+            EchParams::Grease(key) => {
+                let random_index = random_usize_range(secure_random, 0..suites.len())?;
```
ctz: Let's choose just the first and we can lose `random_usize_range`. Remember that ECH GREASE is not adversarial -- there is no requirement that someone can't tell you're doing it. If there were, even choosing uniformly wouldn't meet that, because real uses won't have a uniformly random choice of suite.

ctz (commented):
Thanks for working on this!

rustls/src/client/config.rs
Oakchris1955:
```rs
self.ech_hpke_suites
    .iter()
    .map(|suite| suite.fips())
    .chain(iter::once(self.domain.provider.fips()))
    .min()
    .unwrap()
```
seems to work like a charm on my machine

Oakchris1955:
My bad for force-pushing twice, I just had forgotten to include a small change in a documentation comment in the first force-push

Oakchris1955:
Some changes I made to try and resolve as many of the aforementioned issues as possible:
- `EchParams` struct has been removed in favor of the `with_ech()` and `with_ech_grease()` methods of `ClientConnectionBuilder`
- `ClientConnectionBuilder::with_ech` now accepts a slice. This way, multiple `EchConfigListBytes` can be passed to it and each one of them will be tried until one compatible with one of the HPKE suites is found (or if that doesn't happen, an error will be raised)
- For GREASE, we just pick the first provided suite, so that means that the `random_usize_range` function is no longer needed and has been removed. Furthermore, there is no need for the user to provide a placeholder public key for GREASE; that is generated using the selected suite's `generate_key_pair()` method
- When the `fips()` method is called on `ClientConfig`, the minimum FIPS status of both the `CryptoProvider` and of the ECH HPKE suites will be returned. Also, now that this method is back, I changed the `test_client_fips_service_indicator_includes_ech_hpke_suite` almost back to how it was.
- Renamed a bunch of internal ECH-related variables (`ech_info` -> `ech_mode`) and also `EchConfig::new` now takes an `ech_config_list` by reference

Can we run the workflows once more? I don't know if this is ready to be merged, since as @cpu mentioned, there seems to be a test coverage gap regarding QUIC, but I would like to see if there is anything else I missed regarding coverage.

djc:
> Can we run the workflows once more?

I think you just need to rebase to deal with the conflicts.

Oakchris1955:
Rebased

Oakchris1955:
Removed the unused GREASE public key and fixed that one clippy warning.

rustls/src/client/connection.rs
```diff
@@ -288,8 +342,13 @@ impl ConnectionCore<ClientSide> {
         extra_exts: ClientExtensionsInput,
         quic: Option<&mut dyn QuicOutput>,
         protocol: Protocol,
+        ech_mode: Option<EchMode>,
     ) -> Result<Self, Error> {
-        let mut common_state = CommonState::new(Side::Client, config.fips());
+        let fips = match &ech_mode {
+            Some(ech) => Ord::min(config.fips(), ech.fips()),
+            None => config.fips(),
+        };
```
ctz: Should no longer be required, now `config.fips()` is unconditionally including it?

Oakchris1955:
It also seems as if the `with_ech_for_retry` method is missing coverage, and from what I can see from the current main branch through codecov, the `EchConfig::for_retry` (which is now not publicly exposed and only used by `with_ech_for_retry`) was missing coverage too

Oakchris1955:
Fixed clippy warning, applied ctz's suggestion and also removed `EchMode::fips` since that apparently isn't used (should that be retained and marked as `dead_code`?)

rustls/src/client/connection.rs
Oakchris1955: Removed

Oakchris1955:
I think I got you all covered, however there are still test coverage gaps regarding `with_ech_for_retry` and QUIC.

ctz:
> It also seems as if the `with_ech_for_retry` method is missing coverage, and from what I can see from the current main branch through codecov, the `EchConfig::for_retry` (which is now not publicly exposed and only used by `with_ech_for_retry`) was missing coverage too

Yeah I can confirm we have a pre-existing coverage gap for this. I suggest not trying to address that (or QUIC) in this PR.

Oakchris1955:
Rebased commit to fix the issue with the documentation link and also applied the suggestion by the nightly format check. Codecov will still report a coverage decrease, but as we have already discussed, that is due to the fact that retrying ECH wasn't covered by any preexisting tests and now that things have been moved around, codecov registers that as a coverage decrease. I am still worried however about breaking ECH for QUIC.

bogo/src/main.rs
```diff
@@ -7,6 +7,7 @@
 use core::any::Any;
```
cpu, line 7: _Writing a comment here since I can't comment directly on the commit message_

I think the `!` in the commit headline should be removed (maybe it was an autosquash artifact?)

rustls/src/client/ech.rs
```diff
@@ -47,12 +47,39 @@ pub enum EchMode {
...
+    /// If all pairs have been exhausted, an [`EncryptedClientHelloError::NoCompatibleConfig`]
+    /// error will be raised
+    pub(crate) fn from_ech_config_list(
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
+        suites: &[&'static dyn Hpke],
+    ) -> Result<Self, Error> {
+        for ech_config_list in ech_config_list_slice {
+            if let Ok(ech_config) = EchConfig::new(ech_config_list, suites) {
+                return Ok(ech_config.into());
+            };
```
cpu: It seems a bit unfortunate here that we don't distinguish between an _invalid_ ECH config and one that we just don't support. I think we should be forwarding `InvalidConfigList` errors to the caller.

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
...
+    /// Specify how to connect using ECH
+    ///
+    /// # Errors
+    ///
+    /// One of the provided ECH configurations must be compatible with the HPKE provider’s supported
+    /// suites or an [`EncryptedClientHelloError::NoCompatibleConfig`](crate::error::EncryptedClientHelloError::NoCompatibleConfig)
+    /// error will be returned.
+    pub fn with_ech(
+        mut self,
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
```
cpu, line 200: I'm not sure I understand why this takes multiple `EchConfigListBytes` as input. AIUI you'd only ever have one in hand from an HTTPS lookup (and it's plural individual configurations within that single serialized list).

In the rare event you did have multiple in hand it seems better to lift the responsibility of looping through them to the caller and keeping our API simpler.

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
         self
     }
 
+    /// Specify how to connect using ECH
```
cpu: This feels pretty terse for public API docs and doesn't offer much that the function name didn't already convey.

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
...
+    ) -> Result<Self, Error> {
+        self.ech_mode = Some(EchMode::from_ech_config_list(
+            ech_config_list_slice,
+            &self.config.ech_hpke_suites,
+        )?);
+
+        Ok(self)
+    }
+
+    /// Connect using ECH GREASE
```
cpu: I think we should explain more about what this here as well.

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
...
+    /// # Errors
+    ///
+    /// This methods will error if the HPKE provider fails to generate a placeholder public key.
+    pub fn with_ech_grease(mut self) -> Result<Self, Error> {
+        // just pick the first HPKE suite and stick with it
+        //
+        // the RFC says that the suites should vary to prevent fingerprinting,
+        // but both BoringSSL + NSS seem to do what we do here, so in order to
+        // not stick out, we just copy that behavior
+        self.ech_mode = Some(EchMode::grease_from_suite(self.config.ech_hpke_suites[0])?);
```
cpu, line 248: I think this would panic if `ech_hpke_suites` wasn't set to a slice with at least one element, and it should probably return an `Err` instead.

Can you add test coverage for this at the same time as fixing it to return an `ApiMisuse` error?

rustls/src/client/ech.rs
```diff
@@ -47,12 +47,39 @@ pub enum EchMode {
...
+    ///
+    /// Each suite-configuration pair will be tried, beginning at the first
+    /// ECH configuration, until a valid [`EchConfig`] is constructed
+    ///
+    /// # Errors
+    ///
+    /// If all pairs have been exhausted, an [`EncryptedClientHelloError::NoCompatibleConfig`]
+    /// error will be raised
+    pub(crate) fn from_ech_config_list(
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
```
cpu, line 62: same feedback here RE: taking a slice of lists. Why not just a single list?

rustls/src/client/config.rs
```diff
@@ -179,8 +173,10 @@ pub struct ClientConfig {
     /// a cache that does no caching.
     pub cert_compression_cache: Arc<compress::CompressionCache>,
 
-    /// How to offer Encrypted Client Hello (ECH). The default is to not offer ECH.
-    pub(super) ech_mode: Option<EchMode>,
+    /// List of supported HPKE suites to use when offering Encrypted Client Hello (ECH).
+    ///
+    /// This is optional: If this is empty, we don't offer ECH
```
cpu: Should this also mention GREASE? In particular it might be worth a mention that we always take the 0-th element for GREASE.

rustls/src/client/config.rs
```diff
@@ -238,17 +235,21 @@ impl ClientConfig {
     ///
     /// This is different from [`CryptoProvider::fips()`]: [`CryptoProvider::fips()`]
     /// is concerned only with cryptography, whereas this _also_ covers TLS-level
-    /// configuration that NIST recommends, as well as ECH HPKE suites if applicable.
+    /// configuration that NIST recommends.
+    ///
+    /// ECH HPKE suites' FIPS status will be taken into account: the minimum FIPS status
+    /// of both the [`CryptoProvider`] and of the ECH HPKE suites will be returned
```
cpu: nit: missing a trailing `.` on this sentence.

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
         self
     }
 
+    /// Specify how to connect using ECH
+    ///
+    /// # Errors
+    ///
+    /// One of the provided ECH configurations must be compatible with the HPKE provider’s supported
```
cpu: nit: the curly apostrophe for "provider's" should just be a normal one (maybe an LLM added this?)

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
...
+        )?);
+
+        Ok(self)
+    }
+
+    /// Connect using ECH GREASE
+    ///
+    /// # Errors
+    ///
+    /// This methods will error if the HPKE provider fails to generate a placeholder public key.
```
cpu: typo: "This methods will error" -> "This method will error"

rustls/src/client/ech.rs
```diff
@@ -47,12 +47,39 @@ pub enum EchMode {
...
+        Err(Error::InvalidEncryptedClientHello(
+            EncryptedClientHelloError::NoCompatibleConfig,
+        ))
+    }
+
+    /// Create a new GREASE [`EchMode`] from a supported suite.
+    ///
+    /// # Errors
+    ///
+    /// This methods will error if the HPKE provider fails to generate a placeholder public key.
```
cpu: Same typo here RE: "methods" -> "method"

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
...
+
+        Ok(self)
+    }
+
+    /// Retrying ECH using a retry config from a server's previous rejection
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the server provided no retry configurations in [`RejectedEch`], or if
+    /// none of the retry configurations are compatible with the HPKE provider’s supported suites.
```
cpu: same comment here RE the apostrophe.

rustls/src/quic.rs
```diff
@@ -115,6 +115,7 @@ impl ClientConnection {
             exts,
             Some(&mut quic),
             Protocol::Quic(version),
+            None,
```
cpu, line 123: I know there's a question about test coverage, but there needs to be a code change merged here whether or not we add the required test coverage in this branch (that would be my preference, as a commit-up-front) or as a follow-up.

I think it's pointing to a real API surface gap. IMO we need a way to thread a non-`None` value through here before this is ready to merge.

rustls/src/client/config.rs
```diff
@@ -671,26 +672,22 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
...
+                client_ech_hpke_suites: self.state.client_ech_hpke_suites,
             },
             provider: self.provider,
             time_provider: self.time_provider,
             side: PhantomData,
         }
     }
 
-    /// Enable Encrypted Client Hello (ECH) in the given mode.
+    /// Enable Encrypted Client Hello (ECH) in the given mode by specifying which HPKE suites to use.
```
cpu: I don't think it's accurate to say this "enables" ECH, and remains a bit of a foot gun that didn't exist before.

AFAICT if you call `with_ech_hpke_suites()` on a `Config`, but _don't_ remember to do the per-connection configuration we silently don't enable ECH at all and send the true SNI in plaintext. I'd like to find a way to avoid letting folks fall into this trap.

rustls/src/client/connection.rs
```diff
@@ -194,12 +196,63 @@ impl ClientConnectionBuilder {
...
+    /// suites or an [`EncryptedClientHelloError::NoCompatibleConfig`](crate::error::EncryptedClientHelloError::NoCompatibleConfig)
+    /// error will be returned.
+    pub fn with_ech(
+        mut self,
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
+    ) -> Result<Self, Error> {
+        self.ech_mode = Some(EchMode::from_ech_config_list(
+            ech_config_list_slice,
+            &self.config.ech_hpke_suites,
+        )?);
```
cpu, line 209: Can you add test coverage for a scenario where `with_ech()` is called on a connection, but the `Config` hasn't been populated without supported suites? I think it will correctly return an error in this case but it'd be nice to see that confirmed & tightened against regressions.

cpu (commented):
Thanks for the iterations! I think this is looking better, but I still have some concerns that would prevent me from +1'ing for merge.

cpu:
Can you also do a pass resolving old feedback threads that you believe are addressed? It makes it easier to understand what's still in flight vs what's already resolved. We can always un-resolve a thread if there's more discussion needed.

bogo/src/main.rs
Oakchris1955: The `!` is used to indicate breaking changes per the Conventional Commits guidelines. It doesn't seem as if there are any specific commit message guidelines, but I use them a lot and I see some other people here do too. Then again, I can remove the `!` from the commit message if you want.

cpu: > The ! is used to indicate breaking changes per the Conventional Commits guidelines.

Ah! TIL.

> I can remove the ! from the commit message if you want.

Yeah I think that's best. We don't use this convention or need to call out the breaking changes on a per-commit basis. All of `main` is semver incompatible with the `rel-0.23` branch.

rustls/src/client/connection.rs
Oakchris1955: After banging my head for 15 minutes trying to find where that apostrophe came from since I haven't used LLMs for anything else than generating a formatted date time string from Javascript's Date API (which is cursed btw), I believe I copy-pasted that excerpt from here: <https://docs.rs/rustls/0.23.43/rustls/client/struct.EchConfig.html>

Oakchris1955:
Of course. I will do a greater dive on your comments and check older comments I might have missed tomorrow.

rustls/src/client/connection.rs
cpu: Ohhhhh interesting! It's Rustdoc doing a fancy translation [of the plain `'` in-tree](https://github.com/rustls/rustls/blob/3d975293347887eda1a3c7b270fd8a3604a66cfc/rustls/src/client/ech.rs#L93-L94). That makes sense :+1: Thanks for digging in (and apologies for suggesting it might have been an LLM artifact).

I don't feel super strongly but I think we should stick to the normal one in our doc comments.

Oakchris1955: > Thanks for digging in (and apologies for suggesting it might have been an LLM artifact).

With all the slop you (as in, open-source maintainers in general) have been dealing with, I don't blame you.

> I don't feel super strongly but I think we should stick to the normal one in our doc comments.

I hadn't noticed it till you pointed it out, so yeah, I agree

rustls/src/client/ech.rs
Oakchris1955: I believe it would be better if we just silently rejected `NoCompatibleConfig` errors and returned everything else to the client.

rustls/src/client/connection.rs
Oakchris1955: I made that choice because there might be use cases where having multiple `EchConfigListBytes` might be a realistic scenario and one might want to iterate through them one-by-one until a right one is found, like for example with the `ech-client.rs` example. [Here](https://github.com/rustls/rustls/commit/02f9ad6af4460e0ae6cb62208eeae4de7d3a99d7#diff-b5f5542063147ea698119bcccf95631ad7a5ac118c192003570e4e71a56b1ccfR131) is what that would look like without that design; since `ClientConnectionBuilder` doesn't implement `Clone`, one would have to create the same connection builder each time, all that just because the `with_ech` failed and consumed the previous builder with it.

rustls/src/client/connection.rs
Oakchris1955: If `with_ech_grease` returns an `ApiMisuse` error if `ech_hpke_suites` is empty, wouldn't it make more sense if `with_ech` did so too?

rustls/src/client/config.rs
Oakchris1955: What about
```rs
/// Pass a slice of HPKE suites to use in case ECH is configured later on
/// with [`ClientConnectionBuilder::with_ech`] or [`ClientConnectionBuilder::with_ech_grease`].
```

Oakchris1955:
> Can you also do a pass resolving old feedback threads that you believe are addressed? It makes it easier to understand what's still in flight vs what's already resolved. We can always un-resolve a thread if there's more discussion needed.

I think only djc's comment [here](https://github.com/rustls/rustls/pull/3136#discussion_r3569141268) has been left unaddressed from older threads, mainly because I didn't understand what alternative design he was proposing.

rustls/src/client/config.rs
Oakchris1955: I believe it would be better if we just documented the `with_ech_hpke_suites`, `with_ech` and `with_ech_grease` methods, mention the 0-th element behavior on `with_ech_grease` and just update this to say that an `ApiMisuse` error will be returned if one attempts to configure ECH while this is empty.

rustls/src/quic.rs
Oakchris1955: Other than making `EchMode` public, a solution would be to use a builder API like `ClientConnectionBuilder` (we could also make an `EchMode`-like enum exclusive to QUIC that can be passed to the `new` and `new_with_alpn` methods)

rustls/src/client/connection.rs
cpu: > I made that choice because there might be use cases where having multiple EchConfigListBytes might be a realistic scenario and one might want to iterate through them one-by-one until a right one is found, like for example with the ech-client.rs example.

Ah! I had forgotten that there could be multiple HTTPS records, each with their own ECH config list (and that's what ech-client.rs is handling today). With that framing it does seem more reasonable :+1:

rustls/src/client/connection.rs
cpu: Yes I think they both should.

rustls/src/client/ech.rs
cpu: Resolving based on conversation in the other thread.

rustls/src/client/connection.rs
Oakchris1955: I have added test coverage for both cases above, but I had to derive the `Debug` trait for `ClientConnectionBuilder` or else I couldn't `unwrap_err` the returned value (comparing the Result doesn't work here since `ClientConnectionBuilder` doesn't implement `PartialEq` and I figured that would be a larged overhead than just deriving `Debug`)

Oakchris1955:
Implemented the changes discussed above, now gonna rebase so that there are no merge conflicts

Oakchris1955:
Is the `Codec` trait only implemented for vectors and not for slices? I need to change the `ech_rejected_by_hello_retry_request_conceals_inner_psk` test a bit due to the API changes and to do that I need to convert a `EchConfigPayload` to `EchConfigListBytes`. To do that, I must put the `EchConfigPayload` in question in a vector and then call the `.encode` method from the `Codec` trait on that vector to get the bytes, returning `EchConfigListBytes::from(bytes)`.

Oakchris1955:
Rebased & fixed a clippy warning

djc:
> Is the `Codec` trait only implemented for vectors and not for slices? I need to change the `ech_rejected_by_hello_retry_request_conceals_inner_psk` test a bit due to the API changes and to do that I need to convert a `EchConfigPayload` to `EchConfigListBytes`. To do that, I must put the `EchConfigPayload` in question in a vector and then call the `.encode` method from the `Codec` trait on that vector to get the bytes, returning `EchConfigListBytes::from(bytes)`.

IIRC we can't really implement `Codec` for slices because you can't decode into a slice. This is one reason I've seen other projects split `Codec` into `Encode` and `Decode` traits...

Oakchris1955:
What's up with clippy?

cpu:
> What's up with clippy?

https://github.com/rust-lang/rust-clippy/issues/17529 - It's removed from our required checks for now as a workaround.

Oakchris1955:
Increase coverage on `EchMode::from_ech_config_list` & `From<EchGreaseConfig> for EchMode`

Oakchris1955:
> > Is the `Codec` trait only implemented for vectors and not for slices? I need to change the `ech_rejected_by_hello_retry_request_conceals_inner_psk` test a bit due to the API changes and to do that I need to convert a `EchConfigPayload` to `EchConfigListBytes`. To do that, I must put the `EchConfigPayload` in question in a vector and then call the `.encode` method from the `Codec` trait on that vector to get the bytes, returning `EchConfigListBytes::from(bytes)`.
> 
> IIRC we can't really implement `Codec` for slices because you can't decode into a slice. This is one reason I've seen other projects split `Codec` into `Encode` and `Decode` traits...

I should probably open a new issue for that, but is there any reason to have a unified `Codec` trait over `Encode` and `Decode` traits or is it just that no one thought it?

Oakchris1955:
The only coverage decrease detected is from the `with_ech_for_retry` method, but as we already said, that already existed and Codecov picked up on it just now. Is there anything else than the breakage & gap coverage regarding QUIC that has to be addressed?

Oakchris1955:
May I suggest that we simply add an enum with two variants so that we can configure ECH in QUIC?

Oakchris1955:
Rebased. I'd like some feedback on my idea for QUIC so that I can implement it and close the current API gap.

djc:
> Rebased. I'd like some feedback on my idea for QUIC so that I can implement it and close thew current API gap.

Did you mean this idea?

> Other than making `EchMode` public, a solution would be to use a builder API like `ClientConnectionBuilder` (we could also make an `EchMode`-like enum exclusive to QUIC that can be passed to the `new` and `new_with_alpn` methods)

I think it makes sense to have a builder API for QUIC clients, too. Should probably be a separate PR, though.

Sorry, I understand the lack of feedback here might be frustrating but this is relatively low priority for the project. We will get to it, though!

Oakchris1955:
With the new QUIC API, supporting ECH is more straightforward. I think I fixed the breakage @cpu was talking about, since the `ConnectionCommon::for_client` method under `ClientConnectionBuilder::build_quic` now doesn't take a `None` parameter for ECH, but the same `ech_mode` as under the `build` method, as it can be seen [here](https://github.com/rustls/rustls/pull/3136/changes#diff-5e838d374e93a58a3859c5b6c7a968cb52c5874aecb721826c7fb1476042c93cR343). The tests seems to pass on my end, and while there are some comments left by @cpu that I haven't marked as resolved, they all seem to have been dealt with.

djc:
> The tests seems to pass on my end, and while there are some comments left by @cpu that I haven't marked as resolved, they all seem to have been dealt with.

Please go ahead and resolve any comments that you're fairly certain you have dealt with/don't have open questions.

Oakchris1955:
Should I fix the format issue or no? Other than that (regarding the CI), that coverage gap is from `with_ech_for_retry` (https://github.com/rustls/rustls/pull/3136#issuecomment-5255640565).

ctz:
Yes please! Run `cargo +nightly fmt-unstable` and then squash the change into the commit that introduced it.

rustls/src/client/config.rs
Oakchris1955: Can you please elaborate?

rustls/src/client/ech.rs
Oakchris1955: Forgot to resolve that, working on it.

Oakchris1955: By the looks of it, we shouldn't be calling `config.fips()` under `ConnectionCommon::for_client`, since that takes into account all the ECH HPKE suites, meaning that a single unvalidated suite there would make the connection's FIPS status appear as unvalidated, even if that connection doesn't even use ECH or uses a FIPS-validated suite.

If I am correct, that seems like a coverage gap. Should I write some test cases for this?

cpu: > If I am correct, that seems like a coverage gap. Should I write some test cases for this?

Sounds like a good idea to me.

rustls/src/client/config.rs
```diff
@@ -183,8 +177,11 @@ pub struct ClientConfig {
     /// a cache that does no caching.
     pub cert_compression_cache: Arc<compress::CompressionCache>,
 
-    /// How to offer Encrypted Client Hello (ECH). The default is to not offer ECH.
-    pub(super) ech_mode: Option<EchMode>,
+    /// List of supported HPKE suites to use when offering Encrypted Client Hello (ECH).
+    ///
+    /// This is optional: If this is empty, attempting to configure ECH will result
+    /// in a [`ApiMisuse::NoEchHpkeSuites`] error.
```
cpu, line 183: I think this user facing documentation should be on the builder method, not on the `pub(super)` built config field.

I'd also propose rewriting slightly as:

```
    /// Optional list of supported HPKE suites to use when offering Encrypted Client Hello (ECH).
    ///
    /// If this is empty, attempting to configure ECH for a specific connection 
    /// created with this configuration will result in a [`ApiMisuse::NoEchHpkeSuites`] error.
```

(wrapping might need adjustment, drafting in GH comments)

rustls/src/client/config.rs
```diff
@@ -675,26 +677,23 @@ impl ConfigBuilder<ClientConfig, WantsVerifier> {
...
         }
     }
 
-    /// Enable Encrypted Client Hello (ECH) in the given mode.
+    /// Pass a slice of HPKE suites to use in case ECH is configured later on
+    /// with [`ClientConnectionBuilder::with_ech`] or [`ClientConnectionBuilder::with_ech_grease`].
     ///
     /// This requires TLS 1.3 as the only supported protocol version to meet the requirement
     /// to support ECH.  At the end, the config building process will return an error if either
     /// TLS1.3 _is not_ supported by the provider, or TLS1.2 _is_ supported.
```
cpu, line 693: See my comment on `ClientConfig.ech_hpke_suites` - I think we should be describing the API error that will occur if you configure ECH on a connection without having used `with_ech_hpke_suites()` first here.

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
         self
     }
 
+    /// Provide a slice of [`EchConfigListBytes`] configurations to use to
+    /// connect using ECH. The slice's elements will be iterated until an ECH
+    /// configuration that is compatible with the HPKE provider's supported
+    /// suites is found.
```
cpu, line 189: I think this bit of docs needs tweaking. Ideally the headline should be a single non-broken line.

I also think the way it discusses a "HPKE provider" is maybe not quite right since that's not really a user-facing concept. The concern is specifically that the ClientConfig has been configured with a compatible HPKE suite using the new API and I think the text should better reflect that. We only touch on that aspect in the end w.r.t the NoEchHpkeSuites error.

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+    /// thus, network ossification is mitigated.
+    ///
+    /// If your client uses ECH but doesn't have a configuration for a server it
+    /// wants to connect to, it is recommended per ECH's RFC 9849 to instead
+    /// GREASE the connection for the reasons listed above.
+    ///
+    /// # Note
+    ///
+    /// In order to not stick out, this method will use the first one of the
+    /// provided HPKE and stick with it. ECH's RFC 9849 says that "The selection SHOULD vary",
```
cpu, line 229: grammar nit: "first one of the provided HPKE" reads awkwardly. Maybe "first provided HPKE suite"?

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+    /// however, since both Google's BoringSSL & Mozilla's NSS apparently don't
```
cpu, line 230: The "apparently" here reads oddly.

In general I don't think our choice to not vary, quoting RFC 9849, or citing other implementations makes a lot of sense for API documentation like this. We should just say what we do (select the first provided HPKE suite) and leave it at that.

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+    /// do that, we follow similar behavior in order to blend in with them.
+    ///
+    /// # Errors
+    ///
+    /// This method will error if the HPKE provider fails to generate a placeholder public key.
```
cpu, line 235: Another place we're talking about the HPKE "provider" that I think should be adjusted to be discussing the first selected client configuration HPKE suite failing to generate a placeholder pubkey.

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+    pub fn with_ech_grease(mut self) -> Result<Self, Error> {
+        if self.config.ech_hpke_suites.is_empty() {
+            return Err(Error::ApiMisuse(ApiMisuse::NoEchHpkeSuites));
+        }
+
+        // just pick the first HPKE suite and stick with it
+        //
+        // the RFC says that the suites should vary to prevent fingerprinting,
+        // but both BoringSSL + NSS seem to do what we do here, so in order to
+        // not stick out, we just copy that behavior
```
cpu, line 249: Citing BSSL + NSS here as an internal fn comment seems fine 👍

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+        self.ech_mode = Some(EchMode::grease_from_suite(self.config.ech_hpke_suites[0])?);
+
+        Ok(self)
+    }
+
+    /// Retrying ECH using a retry config from a server's previous rejection
```
cpu, line 255: I think it might be clearer to describe this as configuring the connection for an ECH retry vs leading with the "retrying" verb.

rustls/src/client/ech.rs
```diff
@@ -48,12 +48,45 @@ pub enum EchMode {
...
+        ech_config_list_slice: &[EchConfigListBytes<'_>],
+        suites: &[&'static dyn Hpke],
+    ) -> Result<Self, Error> {
+        for ech_config_list in ech_config_list_slice {
+            match EchConfig::new(ech_config_list, suites) {
+                Ok(ech_config) => return Ok(ech_config.into()),
+                Err(Error::InvalidEncryptedClientHello(
+                    EncryptedClientHelloError::NoCompatibleConfig,
+                )) => {
+                    // silently reject NoCompatibleConfig errors
```
cpu, line 71:
```suggestion
                    // Continue past `NoCompatibleConfig` errors to try the  next `ech_config_list`.
```

rustls/src/client/ech.rs
```diff
@@ -48,12 +48,45 @@ pub enum EchMode {
...
+        Err(Error::InvalidEncryptedClientHello(
+            EncryptedClientHelloError::NoCompatibleConfig,
+        ))
+    }
+
+    /// Create a new GREASE [`EchMode`] from a supported suite.
+    ///
+    /// # Errors
+    ///
+    /// This method will error if the HPKE provider fails to generate a placeholder public key.
```
cpu, line 87: another HPKE provider instance to reword.

rustls/src/client/ech.rs
```diff
@@ -213,14 +246,16 @@ impl EchGreaseConfig {
     /// but doesn't have a real ECH configuration to use for the remote server. In this case
     /// a placeholder or "GREASE"[^0] extension is used.
     ///
-    /// Returns an error if the HPKE provider does not support the given suite.
+    /// # Errors
+    ///
+    /// Returns an error if the HPKE provider fails to generate a placeholder public key.
```
cpu, line 251: provider -> suite

rustls/src/error/mod.rs
```diff
@@ -1629,6 +1629,12 @@ pub enum ApiMisuse {
     /// ECH attempted with a configuration that also supports TLS1.2.
     EchForbidsTls12Support,
 
+    /// ECH attempted without any provided HPKE suites to use when offering ECH
+    NoEchHpkeSuites,
+
+    /// The received plaintext buffer is full; read out some plaintext before receiving more.
+    ReceivedPlaintextBufferFull,
```
cpu, line 1636: What's up with this new error variant? It doesn't seem related to ECH and AFAICT has no users in this branch.

examples/src/bin/ech-client.rs
```diff
@@ -133,9 +125,17 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-            .connect(server_name.clone())
-            .build(&mut output)?;
+        let mut conn = config.connect(server_name.clone());
+
+        conn = if !server_ech_configs.is_empty() {
+            conn.with_ech(&server_ech_configs)
+                .expect("no compatible ECH configuration found")
+        } else {
+            conn.with_ech_grease()
+                .expect("Hpke suite couldn't generate placeholder public key")
```
cpu, line 135: nit: Hpke -> HPKE

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the server provided no retry configurations in [`RejectedEch`], or if
+    /// none of the retry configurations are compatible with the HPKE provider's supported suites.
+    pub fn with_ech_for_retry(mut self, rejection: RejectedEch) -> Result<Self, Error> {
```
cpu, line 261: I think this is missing the `if self.config.ech_hpke_suites.is_empty() ` check right?

rustls/src/client/config.rs
```diff
@@ -242,17 +240,21 @@ impl ClientConfig {
...
-        match &self.ech_mode {
-            Some(ech) => Ord::min(status, ech.fips()),
-            None => status,
-        }
+        self.ech_hpke_suites
+            .iter()
+            .map(|suite| suite.fips())
+            .chain(iter::once(self.domain.provider.fips()))
+            .min()
+            .unwrap()
```
cpu, line 257: would be nice to see a safety comment for this `unwrap()` since it's an addition that wasn't there before, but I also wonder whether we could rewrite this closer to `CryptoProvider::fips()` where we fold with `Ord::min()` and avoid both the unwrap and the iter import.

rustls/src/client/connection.rs
```diff
@@ -180,12 +183,97 @@ impl ClientConnectionBuilder {
...
+    /// # Errors
+    ///
+    /// One of the provided ECH configurations must be compatible with the HPKE provider's supported
+    /// suites or an [`EncryptedClientHelloError::NoCompatibleConfig`](crate::error::EncryptedClientHelloError::NoCompatibleConfig)
+    /// error will be returned.
+    ///
+    /// If no ECH HPKE suites were provided with
+    /// [`ConfigBuilder::with_ech_hpke_suites`](crate::builder::ConfigBuilder::with_ech_hpke_suites),
+    /// [`ApiMisuse::NoEchHpkeSuites`] will be returned instead.
+    pub fn with_ech(
```
cpu, line 200: I think with `EchConfig::new()` becoming `pub(crate)` we've lost some useful user-facing docs (for example, explaining where config list bytes are sourced from).

rustls/src/client/ech.rs
```diff
@@ -48,12 +48,45 @@ pub enum EchMode {
...
+        ))
+    }
+
+    /// Create a new GREASE [`EchMode`] from a supported suite.
+    ///
+    /// # Errors
+    ///
+    /// This method will error if the HPKE provider fails to generate a placeholder public key.
+    pub(crate) fn grease_from_suite(suite: &'static dyn Hpke) -> Result<Self, Error> {
+        Ok(EchGreaseConfig::new(suite)?.into())
```
cpu, line 89: this is a single line helper fn with a single caller. It should be inlined instead.

cpu (commented):
Here's some additional feedback. I think some of the HPKE provider text nits might be pre-existing, but I think it makes sense to standardize the description throughout as part of this rework to provide suites up-front.

I'm still concerned that my earlier feedback about failing open to plaintext SNI if you configure your `ClientConfig` with HPKE suites but fail to do the per-connection ECH configuration hasn't been addressed. It seems like a fairly dangerous footgun.

I think we should also be thinking about how a crate like `tokio-rustls` that owns the connection construction is going to adapt to this proposed API. With the old API design you gave them a ECH-enabled client config and everything worked, I'm not sure I understand how you would do the same using today's `tokio-rustls` API where the connection is built internally... The `connect_with()` API hook gives you a `&mut ClientConnection` and at that point it's too late. Mirroring all of the `with_ech()`, `with_ech_grease()`, and `with_ech_for_retry()` API surface is kinda yucky and `tokio-rustls` is just one consumer to think about.

cpu:
> Yes please! Run cargo +nightly fmt-unstable and then squash the change into the commit that introduced it.

There's also a merge conflict to resolve.

Oakchris1955:
> I'm still concerned that my earlier feedback about failing open to plaintext SNI if you configure your ClientConfig with HPKE suites but fail to do the per-connection ECH configuration hasn't been addressed. It seems like a fairly dangerous footgun.

Other than adding a "Warning" section in the documentation comment of `with_ech_hpke_suites`, I don't see much that we could do. `rustls`' users can simply use the `ClientConnection::ech_status` method to check if ESNI has been offered.

> I think we should also be thinking about how a crate like tokio-rustls that owns the connection construction is going to adapt to this proposed API. With the old API design you gave them a ECH-enabled client config and everything worked, I'm not sure I understand how you would do the same using today's tokio-rustls API where the connection is built internally... The connect_with() API hook gives you a &mut ClientConnection and at that point it's too late. Mirroring all of the with_ech(), with_ech_grease(), and with_ech_for_retry() API surface is kinda yucky and tokio-rustls is just one consumer to think about.

If anything, that's another reason to change the API, since ECH still hasn't gained much traction and any API changes to it will be minimally annoying. I also conducted a [quick search on Github](https://github.com/search?q=%22.with_ech\(%22%20lang%3Arust&type=code) for any code using the `.with_ech` method from the old API and I got less than 150 results. `rustls` seems to be following semantic versioning more or less, so a breaking change for 0.24 shouldn't be an issue IMO.

I gave a quick look at `tokio-rustls` and I believe such issues could be solved by a builder API. Over at #3261, I was said to not create a mirrored `ClientConnectionBuilder`, but from what I see, something similar has happened with `tokio-rustls` regarding `TlsConnector` and `TlsConnectorWithAlpn`. I don't know if there was a particular reason for that API design, but, as a said, a client connection builder API would solve both the issue with ECH and with the mirrored TLS connectors.

Then again, I am open to alternative API designs, as long as they allow one to dynamically configure ECH.

Oakchris1955:
Force-pushed to resolve merge conflict. Please don't run CI yet as I will probably force-push again to resolve some comments.

P.S. the second force-push was because I had forgotten to run `cargo +nightly fmt-unstable`

rustls/src/client/connection.rs
Oakchris1955: Something like:

```rs
/// Provide a slice of [`EchConfigListBytes`] configurations to use to
/// connect using ECH. The slice's elements will be iterated until an ECH
/// configuration that is compatible with one of the HPKE suites provided
/// with [`ConfigBuilder::with_ech_hpke_suites`].
```
?

rustls/src/client/connection.rs
Oakchris1955: I also quote RFC 9849 above that documentation comment, but that it more of an advice to whoever is configuring ECH to chose GREASE instead of plaintext SNI, per the RFC. Should I remove that section too?

rustls/src/client/connection.rs
Oakchris1955: How about:

```rs
/// This method will error if the selected HPKE suite (the first one provided
/// with [`ConfigBuilder::with_ech_hpke_suites`]) fails to generate a placeholder
/// public key.
```
?

cpu:
> Other than adding a "Warning" section in the documentation comment of with_ech_hpke_suites, I don't see much that we could do. rustls' users can simply use the ClientConnection::ech_status method to check if ESNI has been offered.

I'd rather arrive at a solution that adds a more immediate guardrail than just documentation. Checking `ech_status()` is nice, but in the same category of things that someone is likely to forget to do (and wouldn't it also be after the point SNI has been leaked on the wire?).

Is there a way to arrange the connection creation so that if the config specified ECH HPKE suites you have to explicitly opt-out of using ECH using something like `without_ech()`, or the construction errors if none of the other `with_ech()` variants have been used (grease, retry, etc)? I think that would address my concern about failing open accidentally but continues the proliferation of new API surface both here, and in downstream crates that own conn creation internally.

rustls/src/error/mod.rs
Oakchris1955: Seems to have been added to the main branch with c18c486 and removed with f7abe98. I must have picked it up with a rebase between its addition and removal. Removing it.

rustls/src/client/connection.rs
Oakchris1955: Seems like it. By the control flow, it seems to return `NoCompatibleConfig` in such cases, while it should return `NoEchHpkeSuites`. This would have been caught, had it not been for the already-existing coverage gap there. At the very least, I will add a test case to ensure it returns `NoEchHpkeSuites`

Oakchris1955: Regarding that, it seems as if I can't create a dummy `RejectedEch` struct within `rustls-test`, due to [E0639](https://doc.rust-lang.org/error_codes/E0639.html). Any recommendations?

rustls/src/client/config.rs
Oakchris1955: Turned into a for-loop:

```rs
let mut status = self.domain.provider.fips();

for suite in self.ech_hpke_suites.iter() {
    status = Ord::min(status, suite.fips())
}

status
```

rustls/src/client/connection.rs
Oakchris1955: Mirrored relevant documentation to `ClientConnectionBuilder::with_ech`

rustls/src/client/ech.rs
Oakchris1955: I disagree here. `EchMode` already has a `from_ech_config_list` to construct an `Enable` variant, and `grease_from_suite` is the `Grease` counterpart. It may be a single-line function, but at least to me, having both of these constructors in one place seems more neat to me.

Oakchris1955:
> Is there a way to arrange the connection creation so that if the config specified ECH HPKE suites you have to explicitly opt-out of using ECH using something like without_ech(), or the construction errors if none of the other with_ech() variants have been used (grease, retry, etc)?

I don't see why we couldn't add a simple check at the `build` and `build_quic` methods so that, if `config.ech_hpke_suites.is_empty() && self.ech_mode.is_none()` or something along those lines, an `ApiMisuse::EchNotConfigured` error is raised. I could also use if-let chains, since `rustls` uses the 2024 Rust edition.

> I think that would address my concern about failing open accidentally but continues the proliferation of new API surface both here, and in downstream crates that own conn creation internally.

If that is such a big concern , I think that it needs to be discussed more thoroughly before I implement anything else. If you want a second opinion, I guess you can ping djc and ctz.

Oakchris1955:
> I don't see why we couldn't add a simple check at the build and build_quic methods so that, if config.ech_hpke_suites.is_empty() && self.ech_mode.is_none() or something along those lines, an ApiMisuse::EchNotConfigured error is raised. I could also use if-let chains, since rustls uses the 2024 Rust edition.

I will soon rebase so that `with_ech_hpke_suites` returns a `Result` for that very reason.

Oakchris1955:
Resolved doc-related comments, created two new `ApiMisuse` error enums (`EmptyEchHpkeSuitesListFound` for when passing an empty slice to `with_ech_hpke_suites`, and `EchNotConfigured` for when is called `with_ech_hpke_suites` but `with_ech` or `with_ech_grease` isn't), rewrote the method chain within `ClientConfig::fips` as a for-loop.

It seems that there are two issues need addressing:
- Within `ConnectionCommon::for_client`, I am not sure if calling `config.fips()` and passing the result to `CommonState::new` is correct, since it will take into account all provided HPKE suites, whereas by that step, we have already selected one. This way, all the algorithms we use could be FIPS-approved, by `config.fips` could make that appear not being the case.
- @cpu's concern's about adapting to the new API. To me, that doesn't seem like much of an issue, as I discussed previously [here](https://github.com/rustls/rustls/pull/3136#issuecomment-5741950892), but I believe a second opinion wouldn't be harmful. @djc @ctz

Oakchris1955:
Fixed issues with documentation and the testing suite

PR #3140, opened 2026-07-13
Unify Acceptor API with new server state handshake API

ctz:
Subsequent plans continuing along this line:

- [x] also do this to `quic::Acceptor`, and thereby externalise buffers for that API #3152 
- [ ] add `ServerHandshake::VerifyClientCert` which follows the shape of `ServerHandshake::Accepted` (for resolution of `ServerConfig`) and something like https://github.com/rustls/rustls/pull/2713 (a prior client-side attempt at something similar) and externalises use of the cert verification trait
- [ ] follow the same for client handshakes
- [ ] insert this and the components of `SplitConnection` under `ServerConnection`?

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -62,6 +62,14 @@ impl StateMachine for ClientState {
         true
     }
 
+    fn is_traffic(&self) -> bool {
+        matches!(
+            self,
+            Self::Tls12(tls12::Tls12State::Traffic(..))
+                | Self::Tls13(tls13::Tls13State::Traffic(..) | tls13::Tls13State::QuicTraffic(..))
+        )
+    }
+
     fn handle_decrypt_error(&mut self) {
         if let Self::Tls12(tls12::Tls12State::Finished(e)) = self {
             e.handle_decrypt_error();
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -872,6 +872,7 @@ use private::SideOutput;
 pub(crate) trait StateMachine: Sized {
     fn handle<'m>(self, input: Input<'m>, output: &mut dyn Output<'m>) -> Result<Self, Error>;
     fn wants_input(&self) -> bool;
+    fn is_traffic(&self) -> bool;
     fn handle_decrypt_error(&mut self);
     fn into_external_state(
         self,
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -168,6 +168,10 @@ impl<'a, 'm, Side: SideData> MessageIter<'a, 'm, Side> {
     pub(super) fn input(&mut self) -> &mut dyn TlsInputBuffer {
         self.input
     }
+
+    pub(crate) fn state(&self) -> &Result<Side::State, Error> {
+        self.state
+    }
 }
 
 pub(crate) struct ReceivePath {
--- a/rustls/src/server/connection.rs
+++ b/rustls/src/server/connection.rs
@@ -1,6 +1,5 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
-use core::fmt::{Debug, Formatter};
 use core::ops::Deref;
 use core::{fmt, mem};
 use std::io;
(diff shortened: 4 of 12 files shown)


rustls/src/client/hs.rs
```diff
@@ -62,6 +62,14 @@ impl StateMachine for ClientState {
         true
     }
 
+    fn traffic(&self) -> bool {
```
djc: Nit: maybe `is_traffic()`?

rustls/src/conn/mod.rs
```diff
@@ -10,6 +10,7 @@ use pki_types::FipsStatus;
 use crate::common_state::{
     CommonState, ConnectionOutput, ConnectionOutputs, Event, Output, OutputEvent, UnborrowedPayload,
 };
+use crate::conn::receive::FinishCondition;
```
djc: Nit: move this to the local imports?

rustls/src/conn/receive.rs
```diff
@@ -447,6 +448,26 @@ impl ReceivePath {
     }
 }
 
+pub(crate) trait FinishCondition {
```
djc: I forget, did you have any other uses for this in mind? If not, should we just dumb it down to a `traffic_only: bool` argument on `process_new_packets()`?

Or, could we tie it into `SendOutput` being `Discard`?

rustls/src/server/connection.rs
```diff
@@ -208,129 +209,187 @@ impl Deref for ServerConnection {
...
     }
 }
 
-impl Acceptor {
-    /// Check if a `ClientHello` message has been received.
+impl TryFrom<ConnectionCore<ServerSide>> for ServerHandshake {
+    type Error = Error;
+
+    fn try_from(mut inner: ConnectionCore<ServerSide>) -> Result<Self, Error> {
+        const MISUSED: Error = Error::Unreachable("Forgot to restore state");
```
djc: Nit: aren't most of these messages all-lowercase?

rustls/src/server/connection.rs
```diff
@@ -208,129 +209,187 @@ impl Deref for ServerConnection {
...
-            )));
-        };
+    /// - a [`ServerHandshake::Receive`] if more data is required.
+    /// - a [`ServerHandshake::Accepted`] if a whole `ClientHello` has been received, requiring
+    ///   and a choice of [`ServerConfig`] is required to continue.
+    /// - a [`ServerHandshake::Complete`] if the handshake is complete.
+    pub fn process(
+        mut self,
+        input: &mut dyn TlsInputBuffer,
+        output: &mut Vec<Vec<u8>>,
```
djc, line 306: Maybe we should hold off on landing this until after 

- https://github.com/rustls/rustls/pull/3117

lands?

rustls/src/server/connection.rs
```diff
@@ -208,129 +209,187 @@ impl Deref for ServerConnection {
...
-///
-///     // Proceed with handling the ServerConnection.
-/// }
-/// #   Ok(())
-/// # }
-/// ```
-pub struct Acceptor {
-    inner: Option<ConnectionCommon<ServerSide>>,
+/// An in-progress TLS server handshake.
+#[non_exhaustive]
```
djc, line 220: Do we dare make this exhaustive?

rustls/src/server/connection.rs
```diff
@@ -208,129 +209,187 @@ impl Deref for ServerConnection {
...
+    /// A complete `ClientHello` has been received.
+    ///
+    /// The handshake can be progressed by choosing a [`ServerConfig`] based on
+    /// [`Accepted::client_hello()`] and providing it to [`Accepted::choose_config()`].
+    Accepted(Accepted),
+
+    /// The handshake is complete.
+    ///
+    /// Now see [`SplitConnection`] to continue the connection.
+    Complete(SplitConnection<ServerSide>),
```
djc, line 235: Why does this make sense? IIRC we can fuse but not split, so yielding a `ServerConnection` (that can optionally be split by the caller) would be more expressive, right?

rustls/src/conn/receive.rs
ctz: Fallen away in preference to the new iterator.

rustls/src/server/connection.rs
ctz: I would like to have this API underneath `ServerConnection`

The thing I envision there is some internal portion of that type being:

```rust
enum Inner {
  Handshake(ServerHandshake),
  Traffic(SplitConnection<ServerSide>),
}
```

rustls/src/server/connection.rs
ctz: I would say, probably never.

rustls/src/server/connection.rs
djc: Okay, I think that makes sense.

rustls/src/server/connection.rs
```diff
@@ -208,132 +209,215 @@ impl Deref for ServerConnection {
...
-/// # }
-/// ```
-pub struct Acceptor {
-    inner: Option<ConnectionCommon<ServerSide>>,
+/// An in-progress TLS server handshake.
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum ServerHandshake {
+    /// More data needs to be received to make progress.
+    Receive(Receive),
```
djc: I feel like "Receive" does not do a great job of describing what the server's state condition is. Alternative name suggestions:

- `Receiving`
- `NeedInput` or `NeedsInput`
- `Blocked` (maybe evoking `WouldBlock`)

rustls/src/server/connection.rs
```diff
@@ -208,132 +209,215 @@ impl Deref for ServerConnection {
...
+
+            state => {
+                inner.state = Ok(state);
+                Self::Receive(Receive { inner })
+            }
+        })
+    }
+}
+
+/// More data needs to be received to make progress.
```
djc, line 280: "See `process()`."?

djc (approved):

rustls/src/server/connection.rs
ctz: Tried `NeedsInput` here

PR #3141, opened 2026-07-13
Account for encryption overhead in `max_fragment_size` fragmentation

### LINKED ISSUE #991: MessageFragmenter doesn't take account of encryption overhead
If we set `ServerConfig::max_fragment_size` or `ClientConfig::max_fragment_size` it is expected that all the TLS messages are no larger than this.  However, this is only true during the handshake: after that the sizes are larger by the encryption overhead.

This means, setting `max_fragment_size` to `Some(64)`, the message sizes are:

```
64, 64, 9, 6, 32, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81, 81
```

bnjmnjrk:
Fixes #991.

`max_fragment_size` is meant to cap the **wire size** of each TLS record (5-byte header + payload). Encrypted records were exceeding that limit because fragmentation only limited plaintext, not AEAD overhead.

This subtracts the cipher-specific encryption overhead (via `encrypted_record_overhead()`) before fragmenting encrypted traffic. Unencrypted handshake messages are unchanged.

- `test_server_mtu_reduction` now asserts `<= 64` without the old `+ 20` workaround
- New unit test for `max_plaintext_for_encrypted

### DIFF
--- a/rustls/src/conn/send.rs
+++ b/rustls/src/conn/send.rs
@@ -65,15 +65,27 @@ impl SendPath {
         );
     }
 
+    fn max_plaintext_fragment(&self, encrypt: bool) -> usize {
+        if encrypt {
+            self.message_fragmenter.max_plaintext_for_encrypted(
+                self.encrypt_state.encrypted_record_overhead(),
+            )
+        } else {
+            self.message_fragmenter.max_plaintext()
+        }
+    }
+
     /// Like send_msg_encrypt, but operate on an appdata directly.
     pub(crate) fn send_appdata_encrypt(&mut self, payload: OutboundPlain<'_>) -> usize {
         let len = payload.len();
+        let max_plaintext = self.max_plaintext_fragment(true);
         self.send_messages(
             self.message_fragmenter
-                .fragment_payload(
+                .fragment_payload_with_max(
                     ContentType::ApplicationData,
                     ProtocolVersion::TLSv1_2,
                     payload,
+                    max_plaintext,
                 ),
         );
         len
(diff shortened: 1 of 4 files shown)


rustls/src/conn/send.rs
```diff
@@ -65,15 +65,27 @@ impl SendPath {
         );
     }
 
+    fn max_plaintext_fragment(&self, encrypt: bool) -> usize {
```
djc, line 68: Nit: per the style guide, move callees below callers.

rustls/src/crypto/cipher/record_layer.rs
```diff
@@ -77,6 +77,11 @@ impl EncryptionState {
             .unwrap_or_default()
     }
 
+    /// Number of bytes added to a plaintext fragment by record protection.
```
djc, line 80: Nit: move this just above `encrypted_len()`.

rustls/src/msgs/fragmenter.rs
```diff
@@ -40,23 +40,59 @@ impl MessageFragmenter {
...
+        self.max_frag
+    }
+
+    /// Maximum plaintext per fragment when records will be encrypted.
+    ///
+    /// `encrypted_overhead` is the number of bytes added to the plaintext by the
+    /// record protection layer (for example, the TLS 1.3 content-type byte and AEAD tag).
+    pub(crate) fn max_plaintext_for_encrypted(&self, encrypted_overhead: usize) -> usize {
+        self.max_frag.saturating_sub(encrypted_overhead)
+    }
```
djc, line 53: Having both methods on `MessageFragmenter` doesn't seem helpful.

rustls/src/msgs/fragmenter.rs
```diff
@@ -40,23 +40,59 @@ impl MessageFragmenter {
...
+        msg: &'a EncodedMessage<Payload<'_>>,
+        max_plaintext: usize,
+    ) -> impl ExactSizeIterator<Item = EncodedMessage<OutboundPlain<'a>>> + 'a {
+        self.fragment_payload_with_max(
+            msg.typ,
+            msg.version,
+            msg.payload.bytes().into(),
+            max_plaintext,
+        )
+    }
```
djc, line 89: I don't think we should add more methods for this.

cpu:
@bnjmnjrk Are you intending to address the review feedback received so far, or should we close this?

ctz:
Thanks for the contribution!

PR #3142, opened 2026-07-13
conn: extract packet processing from ReceivePath

djc:
Part of a branch from a few months ago working towards dropping `Buffers::received_plaintext`.

### DIFF
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -316,7 +316,7 @@ impl ReceivePath {
 
             let message = unborrowed.reborrow(&Delocator::new(buffer));
             self.deframer
-                .input_message(message.version, message.payload, bounds);
+                .input_message(message.version, bounds, buffer);
             self.deframer.coalesce(buffer)?;
         }
     }
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -186,8 +186,6 @@ impl ReceivePath {
 
     /// Pull a message out of the deframer and send any messages that need to be sent as a result.
     fn deframe<'b>(&mut self, buffer: &'b mut [u8]) -> Result<Option<Decrypted<'b>>, Error> {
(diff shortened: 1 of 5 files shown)


rustls/src/msgs/deframer/mod.rs
```diff
@@ -251,8 +257,14 @@ impl Deframer {
                 payload: delocator.slice_from_range(&first.bounds),
             };
 
+            let iter = DissectHandshakeIter {
+                version: first.version,
+                payload: msg.payload,
+                bounds: first.bounds.start..first.bounds.end,
```
ctz, line 259: Is this expression just `first.bounds`?

rustls/src/conn/receive.rs
```diff
@@ -89,7 +89,84 @@ impl ReceivePath {
         while st.wants_input() {
             let buffer = input.slice_mut();
             let locator = Locator::new(buffer);
-            let res = self.deframe(buffer);
+            let mut want_close_before_decrypt = false;
```
ctz: IMO this change makes `MessageIter::next()` (nee `process_new_packets()`) too long for one function.

ctz (approved):

rustls/src/msgs/deframer/mod.rs
djc: Yes, but `Range` is not `Copy` even if its `T` is `Copy`... so this avoids the `clone()` syntax.

rustls/src/conn/receive.rs
djc: Fair, I've dropped those commits for now.

rustls/src/msgs/deframer/mod.rs
ctz: Got it. Roll on `cheap_clone()` or whatever they intend to call it!

djc: They're planning to replace `Range` with a different `Range` type that does implement `Copy`, AIUI (changing which of the types the syntax refers to at an edition boundary).

https://rust-lang.github.io/rfcs/3550-new-range.html

PR #3132, opened 2026-07-09
backport rfc 9149 ticket_request extension to rel-0.23

hz2:
## summary

backport of #3126 (merged to main) adapted for the rel-0.23 API surface.

the main difference from the `main` implementation: instead of replacing
`send_tls13_tickets: usize` with `Tls13Tickets { default, max }` (breaking),
this adds a new `max_tls13_tickets: usize` field that defaults to 0 (disabled).
when non-zero and a client sends the ticket_request extension, the server
sends `min(requested, max_tls13_tickets)` tickets.

no breaking changes to the public API.

## what's here

- `ExtensionType::TicketRequest` (0x003a)
- `ClientTicketRequest` / `ServerTicketRequestHint` wire format types
- `ClientConfig::send_ticket_request: Option<TicketRequest>` (client opt-in)
- `ServerConfig::max_tls13_tickets: usize` (server opt-in, 0 = disabled)
- server honors the request when max_tls13_tickets > 0, echoes expected_count in EncryptedExtensions

PR #3143, opened 2026-07-13
Prepare 0.23.42

djc:
For

- https://github.com/rustls/rustls/pull/3132

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -279,7 +279,7 @@ dependencies = [
  "bitflags 2.13.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.10.5",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -2681,7 +2681,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.41"
+version = "0.23.42"
 dependencies = [
  "aws-lc-rs",
  "base64",
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -275,7 +275,7 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.41"
+version = "0.23.42"
 dependencies = [
  "log",
  "once_cell",
--- a/rustls/Cargo.toml
+++ b/rustls/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls"
-version = "0.23.41"
+version = "0.23.42"
 edition = "2021"
 rust-version = "1.71"
 license = "Apache-2.0 OR ISC OR MIT"


ctz (approved):

cpu:
I think this should have proposed release notes as part of the PR since this was adding user-facing API surface per the [RELEASING.md](https://github.com/rustls/rustls/blob/main/RELEASING.md) documentation. 

I tried to backfill some into [the release](https://github.com/rustls/rustls/releases/tag/v%2F0.23.42). Further edits welcome.

djc:
IMO this is pretty niche, even if it's user facing, and it's hard to miss in the short release notes.

cpu:
I don't think it's all that more niche than things we've mentioned in other release notes ("Alter default of ServerConfig::require_ems based on provider's FIPS status.", "allow skipping selected ALPN validation", etc).

djc:
Yes, I would probably not write up separate release notes for those either. (Especially if there aren't many other PRs in the release.)

cpu:
I guess I see the point of proposing release notes (or proposing there's none worth adding) is that it allows us to have this discussion. For example the FIPS one came from https://github.com/rustls/rustls/pull/3044 and was approved as-is.

djc:
> I guess I see the point of proposing release notes (or proposing there's none worth adding) is that it allows us to have this discussion. For example the FIPS one came from https://github.com/rustls/rustls/pull/3044 and was approved as-is.

I didn't write anything in this PR and it was still approved, does that count as an implicit discussion?

I want to keep these bugfix releases cheap, and if there's a small list of PRs it's pretty easy to chase down what changed already. IMO writing up custom notes is mainly worth it for larger releases to call out user-visible behavior changes.

cpu:
> I didn't write anything in this PR and it was still approved, does that count as an implicit discussion?

Ideally ctz or myself would have asked about release notes before approving.

It's a minor thing but I think if we have a release process written down it's to make releases consistent and we should try to follow the documented procedure, or fix what's documented when it doesn't match how we want to do releases in practice.

> I want to keep these bugfix releases cheap, and if there's a small list of PRs it's pretty easy to chase down what changed already. IMO writing up custom notes is mainly worth it for larger releases to call out user-visible behavior changes.

That's fair, but it also took me very little time to write the custom release notes here (and I would have appreciated them as a consumer of the release for rustls-ffi). I don't think asking for proposed release notes would make the process particularly expensive (especially for releases like this one where there were few changes to dig through), but if it's a blocker I'm also happy to try and suggest release notes so you don't have to.

djc:
> > I didn't write anything in this PR and it was still approved, does that count as an implicit discussion?
> 
> Ideally ctz or myself would have asked about release notes before approving.
> 
> It's a minor thing but I think if we have a release process written down it's to make releases consistent and we should try to follow the documented procedure, or fix what's documented when it doesn't match how we want to do releases in practice.

I think this is the relevant section from `RELEASING.md`:

> Come up with text detailing headline changes for this release. General guidelines:
> 
> 💚 include any breaking changes.
> 💚 include any major new headline features.
> 💚 include any major, user-visible bug fixes.
> 💚 include any new API deprecations.
> 💚 emphasise contributions from outside the maintainer team.
> ❌ omit any internal build, process or test improvements.
> ❌ omit any minor or user-invisible bug fixes.
> ❌ omit any changes to dependency versions (unless these cause breaking changes).

IMO this change is not breaking and is not major, so the decision not to include it seems in line with these guidelines?

> > I want to keep these bugfix releases cheap, and if there's a small list of PRs it's pretty easy to chase down what changed already. IMO writing up custom notes is mainly worth it for larger releases to call out user-visible behavior changes.
> 
> That's fair, but it also took me very little time to write the custom release notes here (and I would have appreciated them as a consumer of the release for rustls-ffi). I don't think asking for proposed release notes would make the process particularly expensive (especially for releases like this one where there were few changes to dig through), but if it's a blocker I'm also happy to try and suggest release notes so you don't have to.

I think the added value of the custom release notes for a case like this is very small, since a small percentages of users are likely to look at the changelog for semver-compatible releases in the first place and the changes included are still legible via the list of PRs (in this case, linking back to earlier PRs and the issue). I'm fine if you want to asynchronously add to the release notes, but I would prefer not to block on it or add round trips for it, since I want to prioritize shipping changes rather than meta-documenting them.

ctz:
I think a new TLS protocol feature should count as a major change (as, theoretically, it can cause interop problems that causes someone to look into their dependency changes closely.) Sorry for missing that!

ctz:
> as, theoretically, it can cause interop problems that causes someone to look into their dependency changes closely.)

Though, somewhat unlikely as it is not activated by default for clients.

cpu:
> unlikely as it is not activated by default for clients.

Right, and that's the sort of detail worth capturing in the notes IMO.

I think ctz and I are on the same page that implementing a new RFC with client/server API surface is probably worth release note mention and we can try and provide that perspective for discussion on future prep PRs when applicable.

I don't want to rules lawyer our own release notes guidance and if it's OK to asynchronously add release notes where I think they might be worthwhile I'm happy to call this sorted.

Thanks for the discussion

PR #3147, opened 2026-07-14
quic: externalise input buffers

ctz:


rustls/src/quic.rs
```diff
@@ -43,7 +44,10 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     /// Consume unencrypted TLS handshake data.
     ///
     /// Handshake data obtained from separate encryption levels should be supplied in separate calls.
-    fn read_hs(&mut self, plaintext: &[u8]) -> Result<(), Error>;
+    ///
+    /// How much of the input buffer `plaintext` is consumed is recorded by a call to
+    /// [`TlsInputBuffer::discard()`].  Unconsumed data should be presented again on the next call.
+    fn read_hs(&mut self, plaintext: &mut dyn TlsInputBuffer) -> Result<(), Error>;
```
djc: Pre-existing, but `plaintext` is actually a pretty bad name for this?

djc (approved):
Nice!

rustls/src/quic.rs
ctz: Renamed simply to `input`

PR #3149, opened 2026-07-14
conn: yield received plaintext from a lending iterator

djc:
@ctz early feedback?

### DIFF
--- a/examples/src/bin/tlsserver-mio.rs
+++ b/examples/src/bin/tlsserver-mio.rs
@@ -241,6 +241,7 @@ impl OpenConnection {
         if let Err(err) = self
             .tls_conn
             .process_new_packets(&mut self.input)
+            .handle_all(&mut Vec::new())
         {
             error!("cannot process packet: {err:?}");
 
@@ -253,34 +254,33 @@ impl OpenConnection {
 
     fn try_plain_read(&mut self) {
         // Read and process all available plaintext.
-        if let Ok(io_state) = self
+        let mut received_plaintext = Vec::new();
+        let iter = self
             .tls_conn
-            .process_new_packets(&mut self.input)
-        {
-            if let Some(mut early_data) = self.tls_conn.early_data() {
-                let mut buf = Vec::new();
-                early_data
-                    .read_to_end(&mut buf)
-                    .unwrap();
-
-                if !buf.is_empty() {
-                    debug!("early data read {:?}", buf.len());
-                    self.incoming_plaintext(&buf);
-                    return;
-                }
+            .process_new_packets(&mut self.input);
+        match iter.handle_all(&mut received_plaintext) {
+            Ok(_) => {}
+            Err(error) => {
+                error!("cannot read plaintext: {error:?}");
+                self.closing = true;
+                return;
             }
+        }
 
-            if io_state.plaintext_bytes_to_read() > 0 {
-                let mut buf = vec![0u8; io_state.plaintext_bytes_to_read()];
+        if let Some(mut early_data) = self.tls_conn.early_data() {
+            let mut buf = Vec::new();
+            early_data
+                .read_to_end(&mut buf)
+                .unwrap();
 
-                self.tls_conn
-                    .reader()
-                    .read_exact(&mut buf)
-                    .unwrap();
-
-                debug!("plaintext read {:?}", buf.len());
+            if !buf.is_empty() {
+                debug!("early data read {:?}", buf.len());
                 self.incoming_plaintext(&buf);
             }
+        }
+
+        if !received_plaintext.is_empty() {
+            self.incoming_plaintext(&received_plaintext);
         }
     }
 
--- a/rustls-test/tests/api/io.rs
+++ b/rustls-test/tests/api/io.rs
@@ -4,12 +4,10 @@
 #![allow(clippy::std_instead_of_core)] // awaits core::io::IoSlice in stable (1.98)
 
 use core::fmt::Debug;
-use std::borrow::Cow;
 use std::io::{self, BufRead, IoSlice, Read, Write};
 use std::sync::Arc;
 
 use pki_types::DnsName;
-use rustls::crypto::CryptoProvider;
 use rustls::enums::{ContentType, HandshakeType, ProtocolVersion};
 use rustls::error::{AlertDescription, Error, InvalidMessage, PeerIncompatible, PeerMisbehaved};
 use rustls::server::ServerHandshake;
@@ -18,8 +16,8 @@ use rustls::{
 };
 use rustls_test::{
     ClientConfigExt, KeyType, MultiTest, OtherSession, ServerConfigExt, TestNonBlockIo,
-    check_fill_buf, check_fill_buf_err, check_read, check_read_and_close, check_read_err,
-    do_handshake, encoding, make_client_config, make_client_config_with_auth,
+    check_fill_buf, check_fill_buf_err, check_iter, check_read, check_read_err, do_handshake,
+    do_handshake_collecting, encoding, make_client_config, make_client_config_with_auth,
     make_client_config_with_kx_groups, make_disjoint_suite_configs, make_pair,
     make_pair_for_arc_configs, make_pair_for_configs, make_server_config,
     make_server_config_with_kx_groups, make_server_config_with_mandatory_client_auth, server_name,
@@ -39,18 +37,25 @@ fn buffered_client_data_sent() {
         assert_eq!(0, server.writer().write(b"").unwrap());
         assert_eq!(5, client.writer().write(b"hello").unwrap());
 
-        do_handshake(
+        // The client's buffered plaintext may be delivered as part of the final handshake
+        // flight, so collect it during the handshake as well as afterwards.
+        let mut server_received = Vec::new();
+        do_handshake_collecting(
             &mut client_input,
             &mut client,
+            &mut Vec::new(),
             &mut server_input,
             &mut server,
+            &mut server_received,
         );
+
         transfer(&mut client, &mut server_input);
         server
             .process_new_packets(&mut server_input)
+            .handle_all(&mut server_received)
             .unwrap();
 
-        check_read(&mut server.reader(), b"hello");
+        assert_eq!(&server_received, b"hello");
     }
 }
 
@@ -64,18 +69,24 @@ fn buffered_server_data_sent() {
         assert_eq!(0, server.writer().write(b"").unwrap());
         assert_eq!(5, server.writer().write(b"hello").unwrap());
 
-        do_handshake(
+        // The server's buffered plaintext may be delivered as part of the final handshake
+        // flight, so collect it during the handshake as well as afterwards.
+        let mut client_received = Vec::new();
+        do_handshake_collecting(
             &mut client_input,
             &mut client,
+            &mut client_received,
             &mut server_input,
             &mut server,
+            &mut Vec::new(),
         );
         transfer(&mut server, &mut client_input);
         client
             .process_new_packets(&mut client_input)
+            .handle_all(&mut client_received)
             .unwrap();
 
-        check_read(&mut client.reader(), b"hello");
+        assert_eq!(&client_received, b"hello");
     }
 }
 
@@ -101,24 +112,33 @@ fn buffered_both_data_sent() {
                 .unwrap()
         );
 
-        do_handshake(
+        // Buffered plaintext on both sides may be delivered as part of the final handshake
+        // flight, so collect it during the handshake as well as afterwards.
+        let mut client_received = Vec::new();
+        let mut server_received = Vec::new();
+        do_handshake_collecting(
             &mut client_input,
             &mut client,
+            &mut client_received,
             &mut server_input,
             &mut server,
+            &mut server_received,
         );
 
         transfer(&mut server, &mut client_input);
         client
             .process_new_packets(&mut client_input)
+            .handle_all(&mut client_received)
             .unwrap();
+
         transfer(&mut client, &mut server_input);
         server
             .process_new_packets(&mut server_input)
+            .handle_all(&mut server_received)
             .unwrap();
 
-        check_read(&mut client.reader(), b"from-server!");
-        check_read(&mut server.reader(), b"from-client!");
+        assert_eq!(&client_received, b"from-server!");
+        assert_eq!(&server_received, b"from-client!");
     }
 }
 
@@ -145,18 +165,22 @@ fn server_respects_buffer_limit_pre_handshake() {
         12
     );
 
-    do_handshake(
+    let mut client_received = Vec::new();
+    do_handshake_collecting(
         &mut client_input,
         &mut client,
+        &mut client_received,
         &mut server_input,
         &mut server,
+        &mut Vec::new(),
     );
+
     transfer(&mut server, &mut client_input);
     client
         .process_new_packets(&mut client_input)
+        .handle_all(&mut client_received)
         .unwrap();
-
-    check_read(&mut client.reader(), b"01234567890123456789012345678901");
+    assert_eq!(&client_received, b"01234567890123456789012345678901");
 }
 
 #[test]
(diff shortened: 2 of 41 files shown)


rustls/src/conn/mod.rs
```diff
@@ -77,7 +75,10 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     ///
     /// Success from this function comes with some sundry state data
     /// about the connection.
-    fn process_new_packets(&mut self, input: &mut dyn TlsInputBuffer) -> Result<IoState, Error>;
+    fn process_new_packets<'a, 'm>(
```
ctz, line 63: probably needs to come out of the trait?

ctz (commented):
I suggest we do the first two commits separately, as aren't really controversial

rustls/src/conn/mod.rs
djc: I parametrized the trait with a `SideData`, but that also gets a little messy -- see current push -- but probably still better than taking it out of the crate, so that downstreams can still parametrize over connections?

bogo and rustls-util need a fair amount of reworking when reading needs to follow directly on processing.

examples/src/bin/tlsserver-mio.rs
djc: This currently yields a non-trivial error:

```rust
error[E0499]: cannot borrow `self.tls_conn` as mutable more than once at a time
   --> examples/src/bin/tlsserver-mio.rs:261:43
    |
257 |            if let Ok(mut iter) = self
    |   _______________________________-
    |  |_______________________________|
258 | ||             .tls_conn
    | ||_____________________- first mutable borrow occurs here
259 | |              .process_new_packets(&mut self.input)
    | |__________________________________________________- a temporary with access to the first borrow is created here ...
260 |            {
261 |                if let Some(mut early_data) = self.tls_conn.early_data() {
    |                                              ^^^^^^^^^^^^^ second mutable borrow occurs here
...
288 |        }
    |        - ... and the first borrow might be used here, when that temporary is dropped and runs the destructor for type `Result<PlaintextIter<'_, '_, ServerSide>, rustls::Error>`
```

We can't get at early data before `process_new_packets()`, but I think we probably also don't want to only allow access to early data after 1-RTT plaintext has been received. Maybe we live with that for this commit and figure out a way of splicing the early data into the `PlaintextIter` stream?

ctz: I think connection/state type that makes it absolutely clear that a given `PlaintextIter` contains early data would be best. It's going to have to be clear that a given `PlaintextIter` doesn't mix early and normal data, and that we're not inflicting the impossible concept of "receiving early data" on clients.

djc: For now I've worked around it in the example, keeping the early data access separate from the new interface. Would like to follow up on that in a separate PR to get this large change out of the way first.

rustls/src/conn/mod.rs
djc: Alternatively, we could give up on `dyn Connection` and use an associated type for the `Side` instead? So far we use `impl Connection` more anyway.

rustls/src/conn/mod.rs
```diff
@@ -468,125 +395,124 @@ impl ConnectionCore<ServerSide> {
     }
 }
 
-/// Common items for buffered, std::io-using connections.
-pub(crate) struct Buffers {
-    pub(crate) received_plaintext: ChunkVecBuffer,
-    pub(crate) sendable_plaintext: ChunkVecBuffer,
-    pub(crate) has_seen_eof: bool,
+/// Lending iterator over decrypted plaintext chunks.
+pub struct PlaintextIter<'a, 'm, Side: SideData> {
```
djc: Maybe this is not the best name giving the fact that driving this to completion is necessary to make progress even if you're not interested in the plaintext payloads (for example, during the handshake)?

Could instead be more like:

- `InputMessages`
- `InputProcessor`
- `InputHandler`
- `MessageProcessor`

rustls/src/conn/mod.rs
```diff
@@ -468,125 +395,124 @@ impl ConnectionCore<ServerSide> {
...
-        Self {
-            received_plaintext: ChunkVecBuffer::new(Some(DEFAULT_RECEIVED_PLAINTEXT_LIMIT)),
-            sendable_plaintext: ChunkVecBuffer::new(Some(DEFAULT_BUFFER_LIMIT)),
-            has_seen_eof: false,
+impl<'a, 'm, Side: SideData> PlaintextIter<'a, 'm, Side> {
+    pub(crate) fn new(
+        input: &'m mut dyn TlsInputBuffer,
+        buffers: &'a mut Buffers,
+        core: &'a mut ConnectionCore<Side>,
+    ) -> Result<Self, Error> {
```
djc: Noticed during self-review that this doesn't actually need to be fallible. Will rework that in a bit, but shouldn't change the structure significantly.

rustls/src/conn/mod.rs
```diff
@@ -468,125 +395,124 @@ impl ConnectionCore<ServerSide> {
...
-    /// If there are no bytes to read, this returns `Err(ErrorKind::WouldBlock.into())`.
-    ///
-    /// You may learn the number of bytes available at any time by inspecting
-    /// the return of [`Connection::process_new_packets()`].
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let len = self.received_plaintext.read(buf)?;
-        if len > 0 || buf.is_empty() {
-            return Ok(len);
+impl<'a, 'm, Side: SideData + private::Side> Drop for PlaintextIter<'a, 'm, Side> {
+    fn drop(&mut self) {
```
ctz, line 465: Does this mean `mem::forget` on a `PlaintextIter` (or any type that wraps it, like a future) breaks the connection?

ctz: In any case, it would be good to affix the behaviour of this in some tests, to ensure that the connection does something expected (ie, does not panic, or do any serious cryptography failure) if the iterator is:

- not dropped
- not exhausted: dropped after 0..n-1 items are consumed (should be possible to come back and obtain them later, to allow flow control to work)

rustls/src/conn/mod.rs
```diff
@@ -468,125 +394,132 @@ impl ConnectionCore<ServerSide> {
...
-    pub(crate) received_plaintext: ChunkVecBuffer,
-    pub(crate) sendable_plaintext: ChunkVecBuffer,
-    pub(crate) has_seen_eof: bool,
-}
+/// Driver for handling messages from the `TlsInputBuffer`.
+///
+/// Must be driven to completion to make progress, by calling either [`Self::read_all()`] or
+/// repeatedly calling [`Self::next_chunk()`] until it returns `None`.
+#[must_use]
+pub struct MessageHandler<'a, 'm, Side: SideData> {
```
ctz, line 387: Can this iterator operate in terms of `ReceiveTrafficState`, or is that not a goal?

djc: I'm not sure what you mean? My goal here was to drop `Buffers::received_plaintext`.

rustls/src/conn/mod.rs
djc: So we do two things here:

```rust
input.discard(recv.deframer.take_discard());
```

I think not doing this is harmless -- it means the input buffer is not cleaned up, which means someone adding more data to it will fill it up sooner, but that's certainly survivable.

```rust
        // Release unsent buffered plaintext.
        if send.may_send_application_data
            && !self
                .buffers
                .sendable_plaintext
                .is_empty()
        {
            send.send_buffered_plaintext(&mut self.buffers.sendable_plaintext);
        }
```

Not entirely sure what this is for, honestly? But this too, seems like it can be skipped without ill effects.

examples/src/bin/tlsserver-mio.rs
djc: Or, go back to something like

- https://github.com/rustls/rustls/pull/2669

and replace the trait with a parametrized `Connection` type.

rustls/src/conn/mod.rs
ctz: ie, #2669? I'm wondering if some of the discussion on rustdoc and R-A is different now, or still substantially the same?

rustls/src/conn/mod.rs
ctz: Agree on the first one. Skipping the second one omits sending buffered plaintext, so would stall a client-talks-first protocol (ie, HTTP1 would silently stop working)

rustls/src/conn/mod.rs
ctz: To be clear, this isn't an outright strong objection. Just more like actually making a usable API in this way seems to be exceptionally hard for some reason.

rustls-util/src/stream.rs
```diff
@@ -261,22 +314,24 @@ where
...
     #[test]
     fn stream_can_be_created_for_connection_and_tcpstream() {
-        type _Test<'a> = Stream<'a, ClientConnection, TcpStream>;
+        type _Test<'a> = Stream<'a, ClientConnection, ClientSide, TcpStream>;
     }
 
     #[test]
     fn streamowned_can_be_created_for_client_and_tcpstream() {
-        type _Test = StreamOwned<ClientConnection, TcpStream>;
+        type _Test = StreamOwned<ClientConnection, ServerSide, TcpStream>;
```
ctz: this is wrong. (can the bound be put on the struct?)

rustls-util/src/stream.rs
```diff
@@ -261,22 +314,24 @@ where
...
     }
 
     #[test]
     fn streamowned_can_be_created_for_server_and_tcpstream() {
-        type _Test = StreamOwned<ServerConnection, TcpStream>;
+        type _Test = StreamOwned<ServerConnection, ClientSide, TcpStream>;
```
ctz: ditto ^

rustls-test/tests/api/io.rs
```diff
@@ -2238,17 +2272,14 @@ fn client_close_notify() {
...
         transfer(&mut client, &mut server_input);
-        let io_state = server
-            .process_new_packets(&mut server_input)
-            .unwrap();
-        assert!(io_state.peer_has_closed());
-        check_read_and_close(&mut server.reader(), b"from-client!");
+        let iter = server.process_new_packets(&mut server_input);
+        let mut received = Vec::with_capacity(16);
+        let state = iter.handle_all(&mut received).unwrap();
+        assert!(state.peer_has_closed());
```
ctz, line 2279: `assert_eq!(received, b"from-client!");` here or similar -- it is important that `peer_has_closed()` does not overtake appdata.

in addition tests that do not consume the whole of `iter` should see no `peer_has_closed()` here for the same reason.

rustls/src/conn/mod.rs
```diff
@@ -78,7 +72,10 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     ///
     /// Success from this function comes with some sundry state data
     /// about the connection.
```
ctz: docs

rustls/src/conn/mod.rs
```diff
@@ -469,125 +395,132 @@ impl ConnectionCore<ServerSide> {
...
-    fn consume(&mut self, amt: usize) {
-        self.received_plaintext
-            .consume_first_chunk(amt)
+    fn is_empty(&self) -> bool {
+        self.sendable_plaintext.is_empty()
     }
 }
 
-const UNEXPECTED_EOF_MESSAGE: &str = "peer closed connection without sending TLS close_notify: \
-https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof";
```
ctz, line 579: Update docs as referenced here.

I think they can be much simpler, and cover basically that callers should do the check that `stream.rs` does, and how `MessageIter` expresses clean closure.

rustls/src/conn/mod.rs
ctz: in addition the "Getting started" section of the front page docs would benefit from updating.

rustls/src/conn/mod.rs
```diff
@@ -469,125 +395,132 @@ impl ConnectionCore<ServerSide> {
...
-/// Common items for buffered, std::io-using connections.
-pub(crate) struct Buffers {
-    pub(crate) received_plaintext: ChunkVecBuffer,
-    pub(crate) sendable_plaintext: ChunkVecBuffer,
-    pub(crate) has_seen_eof: bool,
-}
+/// Driver for handling messages from the `TlsInputBuffer`.
+///
+/// Must be driven to completion to make progress, by calling either [`Self::handle_all()`] or
+/// repeatedly calling [`Self::next_payload()`] until it returns `None`.
```
ctz, line 382: Would be good to cover backpressure here.

rustls/src/conn/mod.rs
```diff
@@ -469,125 +395,132 @@ impl ConnectionCore<ServerSide> {
...
+            send.send_buffered_plaintext(&mut self.buffers.sendable_plaintext);
         }
+    }
+}
 
-        self.check_no_bytes_state()
-            .map(|()| len)
+impl<S: SideData> Debug for MessageHandler<'_, '_, S> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("PlaintextIter")
```
ctz:
```suggestion
        f.debug_struct("MessageHandler")
```

ctz (approved):

rustls-util/src/stream.rs
djc: What do you mean by putting the bound on the struct?

djc: Fixed.

rustls-util/src/stream.rs
djc: Fixed.

rustls/src/conn/mod.rs
djc: Fixed.

rustls-util/src/stream.rs
ctz: Bare in mind first my confidence with rust type golf is low. Can `trait Connection` require each impl exposes its `Side: SideData` value as an associated item? Then it doesn't need to be restated here, and any references to it can be `C::Side` instead.

rustls/src/conn/mod.rs
djc: Updated.

rustls/src/conn/mod.rs
djc: Done.

rustls/src/conn/mod.rs
djc: Added some notes about this.

rustls-test/tests/api/io.rs
djc: > `assert_eq!(received, b"from-client!");` here or similar -- it is important that `peer_has_closed()` does not overtake appdata.

Done.

> in addition tests that do not consume the whole of `iter` should see no `peer_has_closed()` here for the same reason.

Did you see concrete instances of this?

rustls-util/src/stream.rs
djc: > Bare in mind first my confidence with rust type golf is low. Can `trait Connection` require each impl exposes its `Side: SideData` value as an associated item? Then it doesn't need to be restated here, and any references to it can be `C::Side` instead.

It could but I think it would preclude the use of `dyn Connection` which we're using in some places, which makes it a larger change at the least (and probably triggers some of the monomorphization things you tend to dislike).

rustls/src/conn/mod.rs
djc: @ctz did you have thoughts on this? (Current name is `MessageHandler`.)

djc:
With your approval, going to merge this. Happy to take further feedback of course, and noting to

- Check whether we could/should make the `Side` an associated type on `Connection`

next.

rustls/src/conn/mod.rs
djc: Second one is going away soon due to

- https://github.com/rustls/rustls/pull/3193

so this no longer needs to block.

PR #3151, opened 2026-07-15
docs: note cargo feature behavior invariant

### LINKED ISSUE #940: Document and verify how Rustls feature flags work
The default behavior of Rustls (any crate) must not change based on whether or not a feature flag is enabled. In response to a feature flag being enabled, we may only make additional APIs to enable different/additional functionality if that API is used. In particular, a library using Rustls has no control over whether a different library or the application using the library has enabled a feature or not. 

This has come up a few times:
* People have asked for SSLKEYLOGFILE functionality to be enabled by default if a feature flag is enabled.
* #709 asks for webpki-roots to be enabled by default if a feature flag is enabled.

We should document this in our design documentation so we can easily refer to it in one place, instead of debating it over and over.

cachebag:
Features may expose opt-in APIs but must not change default behavior, since Cargo unifies features across the dependency graph.

PS. Really appreciate the work being done here it's very intriguing. 

Closes #940

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,6 +35,16 @@ Some ideas and guidelines for contributions:
 - I run `cargo outdated` prior to major releases; but PRs to update specific
   dependencies are welcome.
 
+## Crate features
+
+Enabling a Cargo feature must not change rustls' default behavior. Cargo unifies enabled features
+across the entire dependency graph, so a library using rustls cannot control whether another
+library or the application enabled a feature.
+
+A feature may expose additional APIs for opt-in functionality, but that functionality must be
+enabled through those APIs by the application. For example, a feature must not automatically
+enable key logging or select a set of trust anchors.
+
 ## Commit history
 
 We prefer to keep the commit history clean and easy to follow. As such, we prefer small commits


rustls/src/manual/features.rs
```diff
@@ -10,6 +10,16 @@ and are dependent on the provider used.
 [`rustls-aws-lc-rs`]: https://crates.io/crates/rustls-aws-lc-rs
 [`rustls-ring`]: https://crates.io/crates/rustls-ring
 
+## Cargo features
```
djc: I think it's good to document this but I don't think this is the best place for this documentation? There's a "Crate features" section in the crate root docstring that I think would be more appropriate.

rustls/src/lib.rs
```diff
@@ -271,6 +271,14 @@
...
 //! they mean.
 //!
+//! Enabling a Cargo feature must not change rustls' default behavior. Cargo unifies enabled
+//! features across the entire dependency graph, so a library using rustls cannot control whether
+//! another library or the application enabled a feature.
+//!
+//! A feature may expose additional APIs for opt-in functionality, but that functionality must be
+//! enabled through those APIs by the application. For example, a feature must not automatically
+//! enable key logging or select a set of trust anchors.
+//!
```
ctz: Sorry to give you the run around, but I think this is best put in CONTRIBUTING.md. The docs in lib.rs are user-facing and high-value. This is more like "how the project should use crate features" targeted at maintainers and  contributors. I would suggest this text can go -- as is -- in a new `# Crate features` section of that file?

cachebag: No problem, thanks. That makes sense.

ctz (approved):
Thanks for the patience!

PR #3152, opened 2026-07-15
Align QUIC Acceptor with TCP

ctz:
Adds a QUIC version of `ServerHandshake`

### DIFF
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -1,9 +1,8 @@
 use alloc::boxed::Box;
 use alloc::collections::VecDeque;
 use alloc::vec::Vec;
-use core::fmt::{self, Debug};
-use core::mem;
 use core::ops::{Deref, DerefMut};
+use core::{fmt, mem};
 
 use pki_types::{DnsName, FipsStatus, ServerName};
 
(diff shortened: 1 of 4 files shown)


rustls/src/quic.rs
```diff
@@ -623,16 +622,12 @@ pub(crate) struct Quic {
     pub(crate) version: Version,
     /// QUIC transport parameters received from the peer during the handshake
     pub(crate) params: Option<Vec<u8>>,
-    pub(crate) hs_queue: VecDeque<(bool, Vec<u8>)>,
+    pub(crate) events: Vec<QuicEvent>,
```
djc, line 625: Once we get here, maybe `take_events(&mut Vec<QuicEvent>)` should instead be `events() -> impl Iterator<Item = QuicEvent>`, from `self.events.drain(..)`?

djc:
Good stuff!

rustls/src/quic.rs
ctz: Good idea, thanks!

djc (approved):

djc:
Not sure the `take_` prefix still makes sense here. It's kind of an implementation detail if we're yielding an iterator?

PR #3154, opened 2026-07-17
Switch to 2024 edition

djc:
Since we already have 1.85 as MSRV, might as well take this?

### DIFF
--- a/rustls/src/ticketer.rs
+++ b/rustls/src/ticketer.rs
@@ -268,14 +268,14 @@ mod tests {
 
     #[derive(Debug)]
     struct FakeTicketer {
-        gen: u8,
+        generation: u8,
     }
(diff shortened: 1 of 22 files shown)


rustls/src/ticketer.rs
```diff
@@ -268,41 +268,41 @@ mod tests {
 
     #[derive(Debug)]
     struct FakeTicketer {
-        gen: u8,
+        r#gen: u8,
```
ctz: i would call this `generation` rather than using raw idents

ctz (approved):

rustls/src/ticketer.rs
djc: Fair enough, done.

PR #3160, opened 2026-07-20
Prove `ExpectCertificateVerify` comes after chain verification

ctz:
It is a logic error to verify a signature with an unidentified key.

This is a structural robustness improvement, with no runtime effect.

### DIFF
--- a/rustls/src/server/tls12.rs
+++ b/rustls/src/server/tls12.rs
@@ -32,7 +32,7 @@ use crate::suites::PartiallyExtractedSecrets;
 use crate::sync::Arc;
 use crate::tls12::{self, ConnectionSecrets, Tls12CipherSuite};
 use crate::tls13::key_schedule::KeyScheduleTrafficSend;
-use crate::verify::{ClientIdentity, SignatureVerificationInput};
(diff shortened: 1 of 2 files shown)


cpu (approved):
Have you considered changing the zero-sized `PeerVerified` marker into something that carries the `Identity<'static>` of the peer that was verified? This seems like an improvement either way, but I wonder if there's value in more tightly binding _which_ identity was verified to the marker type.

ctz:
> Have you considered changing the zero-sized `PeerVerified` marker into something that carries the `Identity<'static>` of the peer that was verified? This seems like an improvement either way, but I wonder if there's value in more tightly binding _which_ identity was verified to the marker type.

Will look into doing this as a follow-up.

ctz:
> Have you considered changing the zero-sized `PeerVerified` marker into something that carries the `Identity<'static>` of the peer that was verified? This seems like an improvement either way, but I wonder if there's value in more tightly binding _which_ identity was verified to the marker type.

Looked at this and, because we want to keep the marker all the way to the end of the connection, it would mean keeping a separate copy of the certificate chain. That is not currently needed otherwise.

PR #3162, opened 2026-07-21
reject trailing data in decompressed certificate messages

Mounika2456:
The TLS1.3 compressed-certificate handlers decompress the peer's certificate
into a buffer sized to its declared uncompressed length, then parse it with
`CertificatePayloadTls13::read`. That returns as soon as the context and
certificate_list are consumed and drops anything after them. Running the new
test against the current code:

    left:  Err(Server(PeerMisbehaved(NoCertificatesPresented)))
    right: Err(Server(InvalidMessage(TrailingData("read_bytes"))))

`decompress()` fills the output buffer exactly, so a peer can decompress to a
valid Certificate followed by arbitrary trailing bytes and have them accepted.
The uncompressed path rejects the same bytes in
`HandshakeMessagePayload::read_version` via `expect_empty`. Switch both sites
(client reading the server certificate, server reading the client certificate)
to `read_bytes`, which requires the buffer to be fully consumed.

### DIFF
--- a/rustls-test/tests/api/compress.rs
+++ b/rustls-test/tests/api/compress.rs
@@ -260,6 +260,65 @@ fn test_cert_decompression_by_server_produces_invalid_cert_payload() {
 }
 
 #[test]
+fn test_cert_decompression_by_client_rejects_trailing_data() {
+    let provider = provider::DEFAULT_PROVIDER;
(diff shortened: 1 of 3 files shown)


djc (approved):
Good catch, this makes sense to me.

ctz (approved):
Good spot, thank you!

PR #3163, opened 2026-07-21
conn: simplify message writing path

djc:
Happy to rebase on top of 

- #3127

if that makes more sense.


djc:
Hmm, weird that this makes no-crypto benches slower.

ctz (approved):
Nice!

djc:
I've dropped the last 4 commits for now, since more work is needed to deal with the introduction of `write_appdata_into()` from #3127, which reduplicates a bunch of code here.

PR #3164, opened 2026-07-21
Rework and clarify checks for excess data

### LINKED ISSUE #642: Parsing should be refactored to automatically reject unparsed trailing input.
See #640. It seems like it is very easy to accidentally skip calling `if rd.any_left()` after parsing some input. If that check is skipped then unparsed garbage may remain, and/or this might be indicative that the earlier parsing was done incorrectly.

This is one of the main problems I tried to address in `untrusted`; https://docs.rs/untrusted/0.6.2/untrusted/struct.Reader.html. We should do something similar to `untrusted` to automate away this checking.

ctz:
Following on from #3162 this fixes additional similar issues, as found by Opus 4.8. Then refactors exhaustive reading following the closure-based approach of `untrusted` to improve robustness. All actual changes here are by hand.

Recommend reviewing without whitespace.

fixes #642

### DIFF
--- a/rustls/src/client/ech.rs
+++ b/rustls/src/client/ech.rs
@@ -24,7 +24,7 @@ use crate::msgs::{
     ClientExtensions, ClientHelloPayload, Codec, EchConfigContents, EchConfigPayload, Encoding,
     EncryptedClientHello, EncryptedClientHelloOuter, ExtensionType, HandshakeAlignedProof,
     HandshakeMessagePayload, HandshakePayload, HelloRetryRequest, HpkeKeyConfig, Message,
-    MessagePayload, PresharedKeyBinder, PresharedKeyOffer, Random, Reader, ServerHelloPayload,
+    MessagePayload, PresharedKeyBinder, PresharedKeyOffer, Random, ServerHelloPayload,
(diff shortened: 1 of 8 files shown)


djc (approved):
Pretty good. Don't love the callback-based API, but not completely sure if there's a better way in this case...

djc:
Maybe we could have an `AllReader(Reader<'a>)` with `Deref` and `DerefMut` to `Reader` but which overrides a few inherent methods? Might be a bit too convoluted.

cpu (approved):

ctz:
> Maybe we could have an `AllReader(Reader<'a>)` with `Deref` and `DerefMut` to `Reader` but which overrides a few inherent methods? Might be a bit too convoluted.

Not sure I understand how this would work. Seems like it either needs "must move" types or some kind of fallible drop, as a venue to ensure all uses are terminated by a are-we-at-the-end check?

djc:
> Not sure I understand how this would work. Seems like it either needs "must move" types or some kind of fallible drop, as a venue to ensure all uses are terminated by a are-we-at-the-end check?

Yup, I guess that's true. Maybe someday when we have `!Destruct`.

PR #3166, opened 2026-07-22
reject trailing data in v18 ech config contents

Mounika2456:
EchConfigPayload::read takes each config's body from a u16 length-delimited sub-reader, but the recognized-version (V18) branch parses EchConfigContents without checking that sub-reader is drained. A config whose declared length runs past the encoded ECHConfigContents is accepted and the extra bytes are dropped. This is reachable both when a client parses an EchConfigList from DNS and when it parses the server's retry_configs. The recent EchConfigList change tightened the outer list buffer, but the per-config contents were left unchecked.

Route the V18 branch through contents.all(...) so leftover bytes return TrailingData, matching how the sibling Unknown branch and the other message readers already treat their length-delimited regions. Keeping the check on the contents sub-reader means the bound lives next to the field whose length prefix defines it.

### DIFF
--- a/rustls/src/msgs/handshake_test.rs
+++ b/rustls/src/msgs/handshake_test.rs
@@ -861,6 +861,29 @@ fn test_echconfig_serialization() {
     assert_round_trip_eq(ECHCONFIG_LIST_WITH_UNSUPPORTED);
 }
 
+#[test]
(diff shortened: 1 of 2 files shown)


ctz (approved):
Thanks!

djc (approved):

PR #3167, opened 2026-07-22
Bind together `PeerVerified` proof and identity

ctz:
Follow up from #3160 

I think this is OK, but also could be a bit over-wrought?

### DIFF
--- a/rustls/src/client/mod.rs
+++ b/rustls/src/client/mod.rs
@@ -16,7 +16,7 @@ use crate::msgs::{
 use crate::sync::Arc;
 use crate::tls13::Tls13ProtocolSuite;
 use crate::tracing::{debug, trace};
-use crate::verify::DistinguishedName;
+use crate::verify::{DistinguishedName, VerifiedIdentity};
 #[cfg(feature = "webpki")]
 pub use crate::webpki::{
     ServerVerifierBuilder, VerifierBuilderError, WebPkiServerVerifier,
@@ -52,8 +52,7 @@ pub(crate) use tls13::TLS13_HANDLER;
 pub mod danger {
     pub use super::config::danger::{DangerousClientConfig, DangerousClientConfigBuilder};
     pub use crate::verify::{
-        HandshakeSignatureValid, PeerVerified, ServerIdentity, ServerVerifier,
-        SignatureVerificationInput,
+        HandshakeSignatureValid, ServerIdentity, ServerVerifier, SignatureVerificationInput,
     };
 }
 
@@ -217,7 +216,7 @@ impl Deref for Tls13Session {
 #[derive(Clone)]
 pub(crate) struct Tls13ClientSessionInput {
     pub(crate) suite: Tls13ProtocolSuite,
-    pub(crate) peer_identity: Identity<'static>,
+    pub(crate) peer_identity: VerifiedIdentity<'static>,
     pub(crate) quic_params: Option<SizedPayload<'static, u16, MaybeEmpty>>,
 }
 
@@ -262,7 +261,7 @@ impl Tls12Session {
         session_id: SessionId,
         ticket: Arc<SizedPayload<'static, u16, MaybeEmpty>>,
         master_secret: &[u8; 48],
-        peer_identity: Identity<'static>,
+        peer_identity: VerifiedIdentity<'static>,
         time_now: UnixTime,
         lifetime: Duration,
         extended_ms: bool,
@@ -306,7 +305,7 @@ pub struct ClientSessionCommon {
     pub(crate) ticket: Arc<SizedPayload<'static, u16>>,
     pub(crate) epoch: u64,
     lifetime: Duration,
-    peer_identity: Arc<Identity<'static>>,
+    peer_identity: Arc<VerifiedIdentity<'static>>,
 }
 
 impl ClientSessionCommon {
@@ -314,7 +313,7 @@ impl ClientSessionCommon {
         ticket: Arc<SizedPayload<'static, u16>>,
         time_now: UnixTime,
         lifetime: Duration,
-        peer_identity: Identity<'static>,
+        peer_identity: VerifiedIdentity<'static>,
     ) -> Self {
         Self {
             ticket,
@@ -346,7 +345,7 @@ impl<'a> Codec<'a> for ClientSessionCommon {
             ticket: Arc::new(SizedPayload::read(r)?.into_owned()),
             epoch: u64::read(r)?,
             lifetime: Duration::from_secs(u64::read(r)?),
-            peer_identity: Arc::new(Identity::read(r)?.into_owned()),
+            peer_identity: Arc::new(VerifiedIdentity::assertion(Identity::read(r)?.into_owned())),
         })
     }
 }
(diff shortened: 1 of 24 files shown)


rustls/src/verify.rs
```diff
@@ -412,23 +422,49 @@ impl FinishedMessageVerified {
...
+    pub(crate) fn as_marker(&self) -> PeerVerified {
+        PeerVerified(())
+    }
+}
+
+impl<'a> From<VerifiedIdentity<'a>> for Identity<'a> {
+    fn from(val: VerifiedIdentity<'a>) -> Self {
+        val.0
+    }
+}
```
djc, line 462: Nit: I suppose this should live closer to `Identity` rather than `VerifiedIdentity`.

rustls/src/client/tls13.rs
```diff
@@ -613,17 +613,17 @@ impl ExpectEncryptedExtensions {
 
                 // We *don't* reverify the certificate chain here: resumption is a
                 // continuation of the previous session in terms of security policy.
-                let cert_verified = verify::PeerVerified::assertion();
+                let peer_identity =
+                    verify::VerifiedIdentity::assertion(resuming_session.peer_identity().clone());
```
djc: Pre-existing nit: style says we should drop the `verified::` qualification.

djc (approved):
Looks okay to me... so rather than holding on to `VerifiedIdentity`, the insight here compared to yesterday is that we keep it as long as we'd hold the identity anyway? Maybe `as_marker()` should be more like a split thing, that yields both the marker and the (unverified) identity at the same time?

All the cloning still seems ugly to me, but I guess the performance penalty might be acceptable?

rustls/src/verify.rs
ctz: Put it next to `Identity` and added an extra accessor to keep the inside private.

rustls/src/client/tls13.rs
ctz: Fixed in a prior commit.

djc:
> Maybe `as_marker()` should be more like a split thing, that yields both the marker and the (unverified) identity at the same time?

Did you miss this?

ctz:
> > Maybe `as_marker()` should be more like a split thing, that yields both the marker and the (unverified) identity at the same time?
> 
> Did you miss this?

I did!

Since the value in the handshake ends up being moved into a `OutputEvent::PeerIdentity` (and that contains the `VerifiedIdentity`) deconstructing it doen't seem to help?

ctz:
> All the cloning still seems ugly to me, but I guess the performance penalty might be acceptable?

I think this is performance neutral since all the clones of owned values existed before.

rustls/src/client/mod.rs
```diff
@@ -51,8 +51,7 @@ pub(crate) use tls13::TLS13_HANDLER;
 pub mod danger {
     pub use super::config::danger::{DangerousClientConfig, DangerousClientConfigBuilder};
     pub use crate::verify::{
-        HandshakeSignatureValid, PeerVerified, ServerIdentity, ServerVerifier,
```
cpu, line 55: Should this be re-exporting `VerifiedIdentity` like in server/mod.rs?

rustls/src/verify.rs
```diff
@@ -412,23 +422,58 @@ impl FinishedMessageVerified {
...
+    /// Convert the value into an owned one.
+    ///
+    /// This is a straight move if the value is already owned.
+    pub fn into_owned(self) -> VerifiedIdentity<'static> {
+        VerifiedIdentity(self.0.into_owned())
+    }
+
+    pub(crate) fn as_marker(&self) -> PeerVerified {
+        PeerVerified(())
+    }
```
cpu, line 451: Is this and the `PeerVerified` type still needed? AFAICT there are only two callers of `as_marker()` in client `ExpectFinished` and both are throwing it into a `_` prefixed binding in a way that I'm not sure is adding anything.

cpu (commented):
I was expecting to see some code that has a check that the verified identity a verifier returns matches the identity presented on the wire, but I don't see that happening anywhere (?). Am I misunderstanding something?

Related to that I think there might be an asymmetry now with the client/server impls w.r.t how the TLS signature verification inputs are constructed. The server carries forward the verifier's returned `VerifiedIdentity` but drops the wire provided `peer_identity` that was the input and `ExpectCertificateVerify` uses the `VerifiedIdentity`, but on the client-side we check the signature against the on-the-wire identity, and then carry forward a `VerifiedIdentity`. I'm not 100% sure this is meaningful except perhaps in the case a verifier impl is busted but it seemed worth noting?

rustls/src/verify.rs
ctz: The purpose is to prove that `ExpectTraffic` was only reached via a path which previously had a `VerifiedIdentity`

ctz:
> I was expecting to see some code that has a check that the verified identity a verifier returns matches the identity presented on the wire, but I don't see that happening anywhere (?). Am I misunderstanding something?

There's indeed some new latitude introduced here that a verifier can return different identities to the one presented on the wire. I guess one example of where that could be useful is a typical webpki server certificate verifier which returns a found path. FWIW not planning to do that in any of the default verifiers.

An alternative to this is to _not_ have that latitude, retain `PeerVerified` in the public API, and have `VerifiedIdentity` as a private type that is constructed from the `Identity` and `PeerVerified` instead?

> Related to that I think there might be an asymmetry now with the client/server impls w.r.t how the TLS signature verification inputs are constructed. The server carries forward the verifier's returned `VerifiedIdentity` but drops the wire provided `peer_identity` that was the input and `ExpectCertificateVerify` uses the `VerifiedIdentity`, but on the client-side we check the signature against the on-the-wire identity, and then carry forward a `VerifiedIdentity`. I'm not 100% sure this is meaningful except perhaps in the case a verifier impl is busted but it seemed worth noting?

Good spot!

cpu:
> There's indeed some new latitude introduced here that a verifier can return different identities to the one presented on the wire. I guess one example of where that could be useful is a typical webpki server certificate verifier which returns a found path.

Ah ok yes, I understand what you're saying now. Thanks :+1: 

> An alternative to this is to not have that latitude, retain PeerVerified in the public API, and have VerifiedIdentity as a private type that is constructed from the Identity and PeerVerified instead?

As a middle ground, what about a quick assertion that the DER of the EE returned from the verifier is byte-for-byte equal to the one we read off the wire? Maybe I'm still overthinking this, OK if you want to push back.

rustls/src/client/mod.rs
cpu: (Bump on this: was it unresolved because you don't think it should be API surface or just overlooked?)

cpu (approved):
There are conflicts to resolve + a couple Q's from ^ but I think they're minor. The updates from my last review pass LGTM.

rustls/src/client/mod.rs
ctz: the `server::danger` export was undesirable and has been removed (would prefer for a given type to appear once in the public API ideally)

ctz:
> As a middle ground, what about a quick assertion that the DER of the EE returned from the verifier is byte-for-byte equal to the one we read off the wire? Maybe I'm still overthinking this, OK if you want to push back.

I think this would be overdoing it really. Ultimately returning a different identity to that which was asserted would be something that would not happen accidentally, and verifiers are high-assurance bits of code.

PR #3169, opened 2026-07-23
Align `ClientConfig` with `ServerConfig::crypto_provider()`

ctz:
We have `ClientConfig::provider()` plus `ServerConfig::crypto_provider()`. Correct the inconsistency.

(Don't much mind which way this goes.)


djc (approved):
I have a slight preference for `provider()`, since I value conciseness and there's no conflict with another type of provider in the context of rustls.

ctz:
> I have a slight preference for `provider()`, since I value conciseness and there's no conflict with another type of provider in the context of rustls.

Swapped!

PR #3170, opened 2026-07-23
Improve multi-dimension testing

ctz:
The impetus for this change is that I found the Accepter-type API had no testing in cases where client auth is enabled, and broke that in some other work. That is quite silly because it is trivial to test.

Elsewhere we have a hodge-podge of things, like nested loops over protocol versions.

This makes the tests a little slower and more wasteful, but 2.5s for --release still seems acceptable to me.

### DIFF
--- a/rustls-test/src/lib.rs
+++ b/rustls-test/src/lib.rs
@@ -549,6 +549,11 @@ impl MultiTest {
             key_types,
         }
     }
+
+    pub fn require_client_auth(mut self) -> Self {
+        self.anon_client = false;
+        self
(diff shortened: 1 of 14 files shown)


djc (approved):
Nice!

I'd squash the changes in the last commit back into the commits that rendered each helper unnecessary.

ctz:
> I'd squash the changes in the last commit back into the commits that rendered each helper unnecessary.

Done that and moved the remainder of that commit to the top to demonstrate they were preexisting

PR #3172, opened 2026-07-23
Tighten up rejecting messages w/ misplaced extensions

cpu:
While reviewing [a similar fix](https://github.com/golang/go/commit/e88582c5e1e6a12916a23fe57695b4de9347485f) for Go TLS I spotted that outside of HRR + Certificate messages we were lax about rejecting messages containing extensions that are not spec'd as belonging in those message types.

[RFC 9846 section 4.3](https://www.rfc-editor.org/rfc/rfc9846.html#name-extensions) (formerly RFC 8446 section 4.2) requires:

> If an implementation receives an extension which it recognizes and which is not specified for the message in which it appears, it MUST abort the handshake with an "illegal_parameter" alert.

This branch attempts to do that for the other main message types, along the way splitting up the server extensions type to model TLS 1.3 encrypted extensions separately.

WDYT?


cpu:
@rustls-benchmarking bench

rustls/src/msgs/handshake.rs
```diff
@@ -1227,7 +1227,10 @@ impl Codec<'_> for NewSessionTicketExtensions {
         let mut sub = r.sub(len)?;
 
         while sub.any_left() {
-            out.read_one(&mut sub, |unknown| checker.check(unknown))?;
+            // The only extension specified for NewSessionTicket is
+            // `early_data`, which is modelled as a field above; there are
+            // no specified-but-unprocessed types.
+            out.read_one(&mut sub, |unknown| checker.check_unprocessed(unknown, &[]))?;
```
djc: I don't understand from this documentation why we're not passing early data to `check_unprocessed()` here?

rustls/src/msgs/handshake.rs
```diff
@@ -1436,3 +1438,18 @@ impl DuplicateExtensionChecker {
...
 }
+
+/// Recognized but unprocessed extensions specified for CertificateRequest.
+///
+/// rustls does not process these, and they are ignored if received in a
+/// CertificateRequest message.
+///
+/// See RFC 9846 section 4.3 Table 1, plus `signed_certificate_timestamp` per
+/// its IANA "TLS 1.3" registry entry (carried over from RFC 8446 section 4.2).
+const UNPROCESSED_CERTIFICATE_REQUEST_EXTS: &[ExtensionType] = &[
```
djc: Nit: maybe attach this to an appropriate type (maybe `CertificateRequestExtensions`) instead of having a verbose name?

rustls/src/msgs/server_hello.rs
```diff
@@ -166,6 +166,15 @@ extension_struct! {
 }
 
 impl ServerExtensions<'_> {
+    /// Every extension type present in this message.
+    pub(crate) fn received_types(&self) -> impl Iterator<Item = ExtensionType> + '_ {
```
djc: Nit: I think this should be below `into_owned()`, as it is a simpler getter.

rustls/src/server/hs.rs
```diff
@@ -312,6 +321,15 @@ impl<'a> ExtensionProcessing<'a> {
     }
 }
 
+/// Extension values negotiated identically for TLS 1.2 ServerHello and
+/// TLS 1.3 EncryptedExtensions, prior to placement in the right message.
```
djc: Nit: shorter first line?

rustls/src/msgs/server_hello.rs
```diff
@@ -334,14 +334,28 @@ impl<'a> Codec<'a> for EncryptedExtensions<'a> {
...
+                checker.check_unprocessed(unknown, UNPROCESSED_ENCRYPTED_EXTS)
+            })?;
         }
 
         out.unknown_extensions = checker.0;
         Ok(out)
     }
 }
 
+/// Recognized extensions that are specified for EncryptedExtensions, but which
```
djc: Nit: one line?

rustls/src/msgs/server_hello.rs
```diff
@@ -334,14 +334,28 @@ impl<'a> Codec<'a> for EncryptedExtensions<'a> {
...
+/// rustls does not process: these are ignored if received.
+///
+/// See RFC 9846 section 4.3 Table 1, plus `max_fragment_length` per its IANA
+/// "TLS 1.3" registry entry (carried over from RFC 8446 section 4.2).
+const UNPROCESSED_ENCRYPTED_EXTS: &[ExtensionType] = &[
```
djc: Nit: attach?

djc (approved):
Good stuff!

rustls/src/msgs/handshake.rs
cpu: I'll clarify this in the comment since it's hard to see with the macro logic. The reason it's not passed here is that we _do_ process the early data extension before this point, feeding the ext value to the struct field associated with the ext type. The arg to `check_unprocessed()` is only for extensions that:

1. We recognize (e.g. have an ext type defined for)
2. That are spec'd for the message type we have in hand
3. That we _don't_ implement processing for

ctz (approved):

rustls/src/msgs/handshake.rs
cpu: Clarified.

rustls/src/msgs/handshake.rs
cpu: Good idea. Done.

rustls/src/msgs/server_hello.rs
cpu: Moved.

rustls/src/server/hs.rs
cpu: Done.

rustls/src/msgs/server_hello.rs
cpu: Done.

rustls/src/msgs/server_hello.rs
cpu: Done.

PR #3173, opened 2026-07-24
Correct interactions between TLS1.2 and QUIC

ctz:
This PR addresses two disclosed private reports that I have opted to treat as normal bugs.

Found and verified with rust-in-peace (https://github.com/scadastrangelove/rust-in-peace) by @scadastrangelove

These are summarised as:

- A `CryptoProvider` mixing QUIC-capable and -incapable TLS1.3 suites can panic if the peer chooses the QUIC-incapable one. This affects clients but not servers in 0.23, but both in 0.24 (current main) due to 207f809627d4b02b238b49e29d80bd5e39783234. This PR effectively backs that out and tests it.
- A QUIC client would incorrectly accept a TLS1.2 server hello. A full handshake would require a trusted server that talks TLS1.2 in QUIC.

Probably will drop the final commit, as responding to internal inconsistency with misbehaviour is never preferable to an assertion failure.

### DIFF
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -156,7 +156,10 @@ impl ExpectServerHello {
 
         let suite = <CryptoProvider as Borrow<[&'static T]>>::borrow(self.input.config.provider())
             .iter()
-            .find(|cs| cs.common().suite == server_hello.cipher_suite)
+            .find(|cs| {
+                cs.common().suite == server_hello.cipher_suite
+                    && cs.usable_for_protocol(self.input.protocol)
+            })
             .ok_or(PeerMisbehaved::SelectedUnofferedCipherSuite)?;
 
         match self.suite {
--- a/rustls/src/client/hs.rs
+++ b/rustls/src/client/hs.rs
@@ -1006,6 +1006,14 @@ fn prepare_resumption<'a>(
         suite.can_resume_from(tls13.suite)?;
     }
 
+    // Similarly, if the suite is not usable for the protocol.
+    if !tls13
+        .suite
+        .usable_for_protocol(protocol)
+    {
+        return None;
+    }
+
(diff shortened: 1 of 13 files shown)


rustls/src/client/hs.rs
```diff
@@ -192,7 +192,16 @@ impl ExpectServerHello {
...
+        let tls13_supported = self
+            .input
+            .protocol
+            .supports_version(ProtocolVersion::TLSv1_3)
+            && config.supports_version(ProtocolVersion::TLSv1_3);
+        let tls12_supported = self
+            .input
+            .protocol
+            .supports_version(ProtocolVersion::TLSv1_2)
+            && config.supports_version(ProtocolVersion::TLSv1_2);
```
djc: Suggest to reformulate the `supports_version()` interface so that it can encompass more of the repetition here? As in, it could be a method on `input` that takes a `&config` too.

djc (approved):
Nice! Sorry for regressing this...

rustls/src/client/hs.rs
ctz: did that, albeit via adding a `Protocol` to the `supports_version()` function on config types. this happens in a separate commit on the server type.

djc:
Thinking more about this, should we instead have a different `QuicCipherSuite` type (wrapping `Tls13CipherSuite`) which has its own field in `CryptoProvider`? Feels like that might be more principled/rely less on run-time checks?

ctz:
Yeah I agree that would be a better arrangement. However, for the time being, probably not in this PR which will want backporting?

djc:
> Yeah I agree that would be a better arrangement. However, for the time being, probably not in this PR which will want backporting?

Sure, that makes sense.

ctz:
Merging with prior approval as I'd like to move this along.

PR #3177, opened 2026-07-27
rustls-post-quantum: drop unused webpki dependency, prepare 0.3.0-dev.2

cpu:
Nothing in the crate's sources, examples, or benches directly ref's webpki at this point.

I don't think it needs proposed release notes. I'm only bumping the version so I can take this crate as a dev-dep in `webpki` for the ML-DSA sig. algs for `x509-limbo` purposes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1984,7 +1984,7 @@ dependencies = [
 
 [[package]]
 name = "rustls-post-quantum"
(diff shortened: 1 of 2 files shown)


djc (approved):

cpu:
Published rustls-post-quantum v0.3.0-dev.2 at registry `crates-io`

PR #3178, opened 2026-07-28
Expose `ServerHandshake` state for client cert verification

ctz:
This works towards #850 along the style of #2713.

Future plans for this path:

- [ ] follow the same for client handshakes & server cert verification
- [ ] implement the internals of `ServerConnection` using `ServerHandshake` and `SplitConnection`?

Draft: docs and testing gaps.

### DIFF
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -316,7 +316,7 @@ impl<'a, 'm, Side: SideData> MessageHandler<'a, 'm, Side> {
         core: &'a mut ConnectionCommon<Side>,
     ) -> Self {
         Self {
-            iter: MessageIter::new(input, tls, None, core),
+            iter: MessageIter::new(input, tls, None, core, true),
             done: false,
         }
     }
@@ -573,8 +573,17 @@ pub(crate) mod private {
 use private::SideOutput;
 
 pub(crate) trait StateMachine: Sized {
+    /// Advance the state machine using `input` and emitting data to `output`.
     fn handle<'m>(self, input: Input<'m>, output: &mut dyn Output<'m>) -> Result<Self, Error>;
+
+    /// Return true if the current state requires input to be provided via `handle()`.
     fn wants_input(&self) -> bool;
+
+    /// Advance the state machine using no input if possible.
+    ///
+    /// This should return `Ok(self)` otherwise.
+    fn handle_without_input(self) -> Result<Self, Error>;
+
     fn is_traffic(&self) -> bool;
     fn handle_decrypt_error(&mut self);
     fn into_external_state(
--- a/rustls/src/conn/receive.rs
+++ b/rustls/src/conn/receive.rs
@@ -30,6 +30,7 @@ pub(crate) struct MessageIter<'a, 'm, Side: SideData, Send: SendOutput + 'a> {
     pub(super) recv: &'a mut ReceivePath,
     pub(super) state: &'a mut Result<Side::State, Error>,
     pub(super) output: JoinOutput<'a, Send>,
+    pub(super) advance: bool,
 }
 
 impl<'a, 'm, Side: SideData> MessageIter<'a, 'm, Side, SendPath> {
@@ -38,6 +39,7 @@ impl<'a, 'm, Side: SideData> MessageIter<'a, 'm, Side, SendPath> {
         tls: &'a mut Vec<u8>,
         quic: Option<&'a mut dyn QuicOutput>,
         conn: &'a mut ConnectionCommon<Side>,
+        advance: bool,
     ) -> Self {
         Self {
             input,
@@ -50,6 +52,7 @@ impl<'a, 'm, Side: SideData> MessageIter<'a, 'm, Side, SendPath> {
                 send: &mut conn.common.send,
                 side: &mut conn.side,
             },
+            advance,
         }
     }
 }
@@ -61,6 +64,7 @@ impl<'a, 'm, 's, Side: SideData> MessageIter<'a, 'm, Side, SendAdapter<'s>> {
         state: &'a mut Result<Side::State, Error>,
         recv: &'a mut ReceivePath,
         output: JoinOutput<'a, SendAdapter<'s>>,
+        advance: bool,
     ) -> Self {
         Self {
             input,
@@ -68,6 +72,7 @@ impl<'a, 'm, 's, Side: SideData> MessageIter<'a, 'm, Side, SendAdapter<'s>> {
             recv,
             state,
             output,
+            advance,
         }
     }
 }
@@ -156,6 +161,17 @@ impl<'a, 'm, Side: SideData, Send: SendOutput + 'a> MessageIter<'a, 'm, Side, Se
                     *self.state = Err(e.clone());
                     return Some(Err(e));
                 }
+            }
+
+            if self.advance && !st.wants_input() {
+                st = match st.handle_without_input() {
+                    Ok(st) => st,
+                    Err(err) => {
+                        maybe_send_fatal_alert(self.output.send, &err, self.tls);
+                        *self.state = Err(err.clone());
+                        return Some(Err(err));
+                    }
+                };
             }
 
             if self.recv.has_received_close_notify {
--- a/rustls/src/quic.rs
+++ b/rustls/src/quic.rs
@@ -596,7 +596,13 @@ impl<Side: SideData> QuicCommon<Side> {
             .input_quic(input.slice_mut())?;
 
         let mut tls = Vec::new();
-        let mut iter = MessageIter::new(input, &mut tls, Some(&mut self.quic), &mut self.common);
+        let mut iter = MessageIter::new(
+            input,
+            &mut tls,
+            Some(&mut self.quic),
+            &mut self.common,
+            true,
+        );
         let result = match iter.next() {
             Some(Ok(_)) | None => Ok(()),
             Some(Err(e)) => Err(e),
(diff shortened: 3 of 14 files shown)


rustls/src/server/tls12.rs
```diff
@@ -549,45 +549,70 @@ impl ExpectCertificate {
...
+}
+
+impl From<Box<ExpectCertificate>> for ServerState {
+    fn from(value: Box<ExpectCertificate>) -> Self {
+        Self::Tls12(Tls12State::Certificate(value))
+    }
+}
+
+// --- Verify the client's identity
+struct AwaitVerifyClientIdentity {
```
djc: Nit: can we do `AwaitingClientIdentifyVerification` instead, or another more humane word order?

rustls/src/server/tls12.rs
```diff
@@ -549,45 +549,70 @@ impl ExpectCertificate {
...
+    hs: HandshakeState,
+    randoms: ConnectionRandoms,
+    suite: &'static Tls12CipherSuite,
+    server_kx: GroupAndKeyExchange,
+    peer_identity: Identity<'static>,
+}
+
+impl AwaitVerifyClientIdentity {
+    fn use_verifier_trait(self) -> Result<ServerState, Error> {
```
djc: Maybe `with_config()` or something referencing the source of the trait impl would make for a better name?

rustls/src/conn/mod.rs
```diff
@@ -872,8 +872,17 @@ pub(crate) mod private {
...
+    /// Advance the state machine using `input` and emitting data to `output`.
     fn handle<'m>(self, input: Input<'m>, output: &mut dyn Output<'m>) -> Result<Self, Error>;
+
+    /// Return true if the current state requires input to be provided via `handle()`.
     fn wants_input(&self) -> bool;
+
+    /// Advance the state machine using no input if possible.
+    ///
+    /// This should return `Ok(self)` otherwise.
+    fn advance(self) -> Result<Self, Error>;
```
djc: Here too, I'd like to see a name that at least hints at where the decision making comes from in case of a lack of external input.

rustls/src/conn/receive.rs
```diff
@@ -159,6 +175,17 @@ impl<'a, 'm, Side: SideData> MessageIter<'a, 'm, Side> {
...
+            if self.advance && !st.wants_input() {
+                st = match st.advance() {
+                    Ok(st) => st,
+                    Err(err) => {
+                        maybe_send_fatal_alert(self.output.send, &err);
+                        *self.state = Err(err.clone());
+                        return Some(Err(err));
+                    }
+                };
+            }
```
djc: Do we need both? Why, and does that merit a comment?

djc (approved):
Do you have a plan for splitting `ServerConfig` into something with just the transition trait impls vs the other stuff that even direct state machine manipulation needs?

ctz:
> Do you have a plan for splitting `ServerConfig` into something with just the transition trait impls vs the other stuff that even direct state machine manipulation needs?

I have thought about it, and not really. The devil is in the details, because `ClientVerifier` does more than just this interface -- aside from a bunch of configuration accessors, it also has signature verification which is or should be trivially quick and not amenable to this style of dispatch. The same goes for `ServerVerifier`. So we'd also need to split these traits up and configure them separately.

rustls/src/server/connection.rs
```diff
@@ -209,6 +210,25 @@ pub enum ServerHandshake {
...
+    /// - Call [`VerifyClientIdentity::asserted_identity()`] to obtain the peer's presented identity,
+    ///   verify that outside the library (perhaps asynchronously), and then continue the handshake with
+    ///   [`VerifyClientIdentity::into_continue()`].
+    ///
+    ///   If the verification fails, the error can be passed into [`VerifyClientIdentity::into_continue()`] to follow
+    ///   a uniform error handling path.
+    ///
+    /// - Abandon the handshake by discarding this object.
+    ///
+    /// [`ClientVerifier::verify_identity()`]: crate::verify::ClientVerifier::verify_identity
```
djc: Not sure if pre-existing on other variants, but suggest not duplicating this on both the variant and the inner type?

rustls/src/server/hs.rs
```diff
@@ -775,6 +787,46 @@ impl From<Box<ExpectClientHello>> for ServerState {
...
+        self.inner.asserted_identity()
+    }
+
+    /// Progress the handshake by calling the verifier trait synchronously.
+    pub(crate) fn use_verifier_trait(self) -> Result<ServerState, Error> {
+        self.inner.with_config()
+    }
+
+    /// Progress the handshake by incorporating an external verification.
+    pub(crate) fn into_continue(
```
djc: Don't love this method name. Was thinking maybe `verified()` but I guess `verified(verified)` is a bit much...

rustls/src/server/connection.rs
```diff
@@ -388,6 +412,88 @@ impl fmt::Debug for Accepted {
...
+/// - Abandon the handshake by discarding this object.
+///
+/// The returned object is a further [`ServerHandshake`].  Commonly this will be a
+/// [`ServerHandshake::NeedsInput`] which will accept and process further data.
+///
+/// [`ClientVerifier::verify_identity()`]: crate::verify::ClientVerifier::verify_identity
+pub struct VerifyClientIdentity {
+    // invariant: `inner.state` is `Err(_)` and requires restoring
+    inner: ConnectionCommon<ServerSide>,
+    verify_identity: hs::VerifyClientIdentity,
```
djc, line 424: Nit: import `VerifyClientIdentity`.

rustls/src/server/connection.rs
```diff
@@ -388,6 +412,88 @@ impl fmt::Debug for Accepted {
...
+            verification_result.and_then(|verified| {
+                self.verify_identity
+                    .into_continue(verified)
+            }),
+            tls,
+        )
+    }
+
+    /// Inspect the identity that the client has provided.
+    pub fn asserted_identity(&self) -> Result<ClientIdentity<'static, '_>, Error> {
```
djc: Nit: suggest moving this below `next()`?

Also wonder about this name. Feels like `asserted` here has a slightly different name compared to how we use it in something like `HandshakeSignatureValid::assertion()`?

What do you think about using "declare" or "proclaim" or "advertise" instead? (Open question -- not sure I like any of these better.)

rustls/src/server/connection.rs
```diff
@@ -388,6 +412,88 @@ impl fmt::Debug for Accepted {
...
+        self.verify_identity.asserted_identity()
+    }
+
+    fn next(
```
djc, line 463: Maybe attach this to `ConnectionCommon`, as like `ConnectionCommon::advance()`?

rustls/src/quic.rs
```diff
@@ -357,6 +361,25 @@ pub enum ServerHandshake {
     /// [`Accepted::client_hello()`] and providing it to [`Accepted::choose_config()`].
     Accepted(Accepted),
 
+    /// The client's presented identity must be verified.
```
djc, line 364: As for TCP.

rustls/src/quic.rs
```diff
@@ -536,6 +563,72 @@ fn check_server_config(config: &ServerConfig) -> Result<(), Error> {
...
+///   If the verification fails, the error can be passed into [`Self::into_continue()`] to follow
+///   a uniform error handling path.
+///
+/// - Abandon the handshake by discarding this object.
+///
+/// The returned object is a further [`ServerHandshake`].  Commonly this will be a
+/// [`ServerHandshake::NeedsInput`] which will accept and process further data.
+///
+/// [`ClientVerifier::verify_identity()`]: crate::verify::ClientVerifier::verify_identity
+pub struct VerifyClientIdentity {
```
djc, line 572: It feels like there must be a way to abstract over this and the TCP variant? Have been contemplating a public `Quic` generic argument to `ConnectionCommon` and using a `Tcp` type (which could potentially be a `()` newtype wrapper).

Doesn't need to be addressed here, but something to think about.

djc (commented):
I think this is right and yet it feels like more boilerplate than is necessary?

I wonder if we could slightly demodularize this so that there's a step-by-step potentially asynchronous path and a monolithic path based on the current traits, but you can't switch between them at each step? The caller could always use the trait stuff outside...

rustls/src/server/connection.rs
ctz: This is avoiding two types named `VerifyClientIdentity` being in scope, one internal and one external.

djc: Ah, fair.

rustls/src/server/hs.rs
ctz: going to try `continue_with()`. another thought is `ready()` to reflect `std::task::Poll`?

rustls/src/server/connection.rs
ctz: i'll try `presented_identity()` here which is already used in comments

ctz: keeping it above `next()` as it is more public

djc: I like "presented", nice.

rustls/src/quic.rs
ctz: Yes that should be doable in the future.

rustls/src/server/connection.rs
ctz: Since the types returned between this and the QUIC one are different, and the bodies are slightly different also, I think it doesn't save much.

PR #3180, opened 2026-07-29
Remove stale `std` feature references

### LINKED ISSUE #3134: Reinstate `std` feature/fix documentation
https://github.com/rustls/rustls/pull/2905 removed this, depending on where 0.24 ends up it may or may not need to come back.

If it doesn't need to come back (it would be ideal if it doesn't!), instead:

- update crate feature docs
- update `[package.metadata.docs.rs]`

Joshuaisikah:
Rustls no longer has the std but its still refrenced in package.metadata.docs.rs this  breaks the docks.rs-style build 
To fix this I removed the stale references from both places.   
Closes #3134


djc (approved):
Thanks!

PR #3181, opened 2026-07-29
Use more checked arithmetic

ctz:
(Not fixing any reachable issues)


djc (approved):
LGTM.

I investigated `clippy::arithmetic_side_effects` a little while ago, but it has a pretty bad signal/noise ratio...

PR #3182, opened 2026-07-29
0.23 backports

ctz:
0.23.43 proposed release notes:

- **Bug fix**: corrected reachable panic in debug mode (or with `overflow-checks`) when decrypting a ticket of specific lengths with `Rfc5077Ticketer` (and therefore `rustls::crypto::aws_lc_rs::Ticketer`).  This happens pre-authentication.

  This is not used in the default configuration, but using stateless resumption is common and this affects the aws-lc-rs provider. The ticketer associated with the *ring* provider is not affected.
- **Bug fix**: A QUIC client using a `CryptoProvider` that mixes QUIC-capable and -incapable TLS1.3 suites can panic if the server chooses the QUIC-incapable suite.
- **Bug fix**: A QUIC client would incorrectly accept a TLS1.2 server hello. A full handshake would require a trusted server that talks TLS1.2 in QUIC.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -279,7 +279,7 @@ dependencies = [
  "bitflags 2.13.1",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.10.5",
  "log",
  "prettyplease",
  "proc-macro2",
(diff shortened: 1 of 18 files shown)


djc (approved):

cpu (approved):
Thanks for handling this.

PR #3185, opened 2026-07-29
Track QUIC suite state in the type system

djc:
Followup from

- #3173 

to try to anchor the difference in the type system. This tries to propagate tracking outward, though it doesn't go as far as adding a separate `quic_cipher_suites` field in `CryptoProvider`. To be continued maybe...

### DIFF
--- a/rustls/src/tls13/mod.rs
+++ b/rustls/src/tls13/mod.rs
@@ -82,7 +82,7 @@ impl Tls13CipherSuite {
     /// Returns a `quic::Suite` for the ciphersuite, if supported.
     pub fn quic_suite(&'static self) -> Option<crate::quic::Suite> {
         self.quic
-            .map(|quic| crate::quic::Suite { suite: self, quic })
+            .map(|quic| crate::quic::Suite { inner: self, quic })
     }
 }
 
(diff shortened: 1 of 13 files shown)


rustls/src/tls13/mod.rs
```diff
@@ -5,11 +5,31 @@ use pki_types::FipsStatus;
...
 use crate::crypto::{self, SignatureScheme, hash};
 use crate::enums::ProtocolVersion;
+use crate::quic;
 use crate::suites::{CipherSuiteCommon, Suite, SupportedCipherSuite};
 use crate::version::Tls13Version;
 
 pub(crate) mod key_schedule;
 
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum Tls13ProtocolSuite {
```
ctz, line 15: Not sure if it's a good idea, but this commit would be smaller if this type Deref'd to `Tls13CipherSuite`?

ctz (approved):

rustls/src/tls13/mod.rs
djc: I tried that but it gets a little annoying because a `Deref::Target` takes on its parents lifetime and we kind of want `&'static` for all/some of these.

PR #3186, opened 2026-07-29
grease the outer psk when a hello retry request rejects ech

Mounika2456:
New test, before the fix:

    ---- ech_rejected_by_hello_retry_request_conceals_inner_psk ----
    assertion `left != right` failed
      left: [105, 110, 110, 101, 114, 32, 110, 97, 109, 101, ...]
     right: [105, 110, 110, 101, 114, 32, 110, 97, 109, 101, ...]

Both sides are the stored ticket, taken from the second ClientHelloOuter's `pre_shared_key`
identity. A HelloRetryRequest that rejects ECH leaves `emit_client_hello_for_retry` in
`(EchStatus::Rejected, Some(ech_state))`, which the catch-all arm swallows, so `ech_hello()`
and the `grease_psk()` inside it never run for the retry. That hello then offers the real
ticket for the concealed inner name in cleartext, with the all-zero placeholder binder, since
`fill_in_psk_binder` is skipped while `ech_state` is `Some`.

### DIFF
--- a/rustls/src/client/ech.rs
+++ b/rustls/src/client/ech.rs
@@ -572,6 +572,41 @@ impl EchState {
         self.inner_hello_transcript = inner_transcript_buffer;
     }
 
+    // See https://datatracker.ietf.org/doc/html/rfc9849#name-grease-psk
+    pub(super) fn grease_psk(&self, psk_offer: &mut PresharedKeyOffer) -> Result<(), Error> {
+        for ident in psk_offer.identities.iter_mut() {
+            // "For each PSK identity advertised in the ClientHelloInner, the
+            // client generates a random PSK identity with the same length."
+            let Some(identity) = ident.identity.as_mut() else {
+                unreachable!();
+            };
+            self.secure_random.fill(identity)?;
+
+            // "It also generates a random, 32-bit, unsigned integer to use as
+            // the obfuscated_ticket_age."
+            let mut ticket_age = [0_u8; 4];
+            self.secure_random
+                .fill(&mut ticket_age)?;
+            ident.obfuscated_ticket_age = u32::from_be_bytes(ticket_age);
+        }
+
+        // "Likewise, for each inner PSK binder, the client generates a random string
+        // of the same length."
+        psk_offer.binders = psk_offer
+            .binders
+            .iter()
+            .map(|old_binder| {
+                // We can't access the wrapped binder PresharedKeyBinder's PayloadU8 mutably,
+                // so we construct new PresharedKeyBinder's from scratch with the same length.
+                let mut new_binder = vec![0; old_binder.as_ref().len()];
+                self.secure_random
+                    .fill(&mut new_binder)?;
+                Ok::<PresharedKeyBinder, Error>(PresharedKeyBinder::from(new_binder))
+            })
+            .collect::<Result<_, _>>()?;
+        Ok(())
+    }
+
     // 5.1 "Encoding the ClientHelloInner"
     fn encode_inner_hello(
         &mut self,
@@ -736,41 +771,6 @@ impl EchState {
             .add_message(&inner_hello_msg);
 
         Ok(encoded_hello)
-    }
-
-    // See https://datatracker.ietf.org/doc/html/rfc9849#name-grease-psk
-    fn grease_psk(&self, psk_offer: &mut PresharedKeyOffer) -> Result<(), Error> {
-        for ident in psk_offer.identities.iter_mut() {
-            // "For each PSK identity advertised in the ClientHelloInner, the
-            // client generates a random PSK identity with the same length."
-            let Some(identity) = ident.identity.as_mut() else {
-                unreachable!();
-            };
-            self.secure_random.fill(identity)?;
-
-            // "It also generates a random, 32-bit, unsigned integer to use as
-            // the obfuscated_ticket_age."
-            let mut ticket_age = [0_u8; 4];
-            self.secure_random
-                .fill(&mut ticket_age)?;
-            ident.obfuscated_ticket_age = u32::from_be_bytes(ticket_age);
-        }
-
-        // "Likewise, for each inner PSK binder, the client generates a random string
-        // of the same length."
-        psk_offer.binders = psk_offer
-            .binders
-            .iter()
-            .map(|old_binder| {
-                // We can't access the wrapped binder PresharedKeyBinder's PayloadU8 mutably,
-                // so we construct new PresharedKeyBinder's from scratch with the same length.
-                let mut new_binder = vec![0; old_binder.as_ref().len()];
-                self.secure_random
-                    .fill(&mut new_binder)?;
-                Ok::<PresharedKeyBinder, Error>(PresharedKeyBinder::from(new_binder))
-            })
-            .collect::<Result<_, _>>()?;
-        Ok(())
     }
 
     fn server_hello_conf(
--- a/rustls/src/client/ech.rs
+++ b/rustls/src/client/ech.rs
@@ -22,8 +22,8 @@ use crate::msgs::{
     ClientExtensions, ClientHelloPayload, Codec, EchConfigContents, EchConfigPayload, Encoding,
     EncryptedClientHello, EncryptedClientHelloOuter, ExtensionType, HandshakeAlignedProof,
     HandshakeMessagePayload, HandshakePayload, HelloRetryRequest, HpkeKeyConfig, Message,
-    MessagePayload, PresharedKeyBinder, PresharedKeyOffer, Random, ServerHelloPayload,
-    ServerNamePayload, SizedPayload,
+    MessagePayload, PresharedKeyBinder, PresharedKeyIdentity, PresharedKeyOffer, Random,
+    ServerHelloPayload, ServerNamePayload, SizedPayload,
 };
 use crate::tls13::Tls13CipherSuite;
 use crate::tls13::key_schedule::{
@@ -340,6 +340,9 @@ pub(crate) struct EchState {
     enable_sni: bool,
     // The extensions sent in the inner hello.
     sent_extensions: Vec<ExtensionType>,
+    // The GREASE PSK identities offered in the first outer hello, if any. A retry hello
+    // re-offers the same identities.
+    grease_psk_identities: Option<Vec<PresharedKeyIdentity>>,
 }
 
 impl EchState {
@@ -384,6 +387,7 @@ impl EchState {
             enc,
             enable_sni,
             sent_extensions: Vec::new(),
+            grease_psk_identities: None,
         })
     }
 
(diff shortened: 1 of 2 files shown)


rustls/src/client/ech.rs
```diff
@@ -751,7 +751,7 @@ impl EchState {
     }
 
     // See https://datatracker.ietf.org/doc/html/rfc9849#name-grease-psk
-    fn grease_psk(&self, psk_offer: &mut PresharedKeyOffer) -> Result<(), Error> {
+    pub(super) fn grease_psk(&self, psk_offer: &mut PresharedKeyOffer) -> Result<(), Error> {
```
cpu: I think this should be reordered below `transcript_hrr_update()` and before `encode_inner_hello()` now that it's `pub(super)` to maintain a `pub` > `pub(crate)` > `pub(super)` > mod private ordering.

rustls/src/client/ech.rs
```diff
@@ -961,6 +976,159 @@ mod tests {
...
+            .unwrap();
+        conn.process_new_packets(&mut input)
+            .unwrap();
+
+        // we continue with a second outer hello, which must conceal the ticket too
+        let second = client_hello_written_by(&mut conn);
+        assert_ne!(psk_identity(&second), TICKET);
+    }
+
+    fn mock_ech_config() -> EchConfig {
```
cpu: I think since this has only one caller it should be in-lined instead of a helper fn.

rustls/src/client/ech.rs
```diff
@@ -961,6 +976,159 @@ mod tests {
...
+                },
+                maximum_name_length: 255,
+                public_name: DnsName::try_from("public.example.com").unwrap(),
+                extensions: vec![],
+            }),
+            suite: &MockHpke,
+        }
+    }
+
+    fn client_hello_written_by(conn: &mut ClientConnection) -> ClientHelloPayload {
```
cpu: ditto here w.r.t not making a helper for a single test to use.

rustls/src/client/ech.rs
```diff
@@ -961,6 +976,159 @@ mod tests {
...
+
+    fn psk_identity(hello: &ClientHelloPayload) -> &[u8] {
+        hello
+            .preshared_key_offer
+            .as_ref()
+            .unwrap()
+            .identities[0]
+            .identity
+            .bytes()
+    }
```
cpu: This has two callers, but the logic is fairly trivial. I would either A) in-line it and eat the small amount of duplication B) make this something that's scoped internal to the one test that uses it, as a closure or an inner fn.

rustls/src/client/ech.rs
```diff
@@ -961,6 +976,159 @@ mod tests {
...
+
+    fn roots() -> RootCertStore {
```
cpu: another single-caller helper fn.

rustls/src/client/ech.rs
```diff
@@ -961,6 +976,159 @@ mod tests {
...
+        let mut r = RootCertStore::empty();
+        r.add(CertificateDer::from_slice(include_bytes!(
+            "../../../test-ca/rsa-2048/ca.der"
+        )))
+        .unwrap();
+        r
+    }
+
+    const TICKET: &[u8] = b"inner name resumption ticket";
```
cpu: It feels slightly cleaner to me to have `SingleTicketStore` wrap a `&[u8]` for the static ticket and let the test callers construct the store with the value they want.

If you want to keep this as-is I would vote to make this const associated with `SingleTicketStore` in an `impl SingeTicketStore { ... }` block positioned after `struct SingleTicketStore` and before `impl ClientSessionStore for SingleTicketStore`.

If you don't make either of those changes and keep this as-is, then the `const` should at least be defined below `impl ClientSessionstore { ... }` to maintain our top-down-ordering + style.

rustls/src/client/hs.rs
```diff
@@ -816,6 +816,14 @@ fn emit_client_hello_for_retry(
...
             }
         }
+        // If a hello retry request rejected our ECH offer, we carry on with the outer hello.
+        // There is no inner hello to encrypt, but the outer one still must not offer the
+        // inner name's ticket.
+        (EchStatus::Rejected, Some(ech_state)) => {
+            if let Some(psk_offer) = chp_payload.preshared_key_offer.as_mut() {
+                ech_state.grease_psk(psk_offer)?;
+            }
+        }
```
cpu, line 824: I agree that we should be making sure the HRR response doesn't offer the inner ticket but I don't think we should be generating a fresh GREASE'd PSK for the response.

I think doing that would be "sticking out" in RFC 9849 terminology because an attacker that can trigger HRR can distinguish between real-ECH-with-GREASE-PSK from GREASE-ECH-with-real-PSK.

That's why we carry the `prev_ech_ext` in `ClientHelloInput` in the arm above. I would suspect the right thing here is to generate the identities and ages initially, and only re-randomize the binders across HRR (?)

rustls/src/client/ech.rs
```diff
@@ -961,6 +976,159 @@ mod tests {
...
+
+    /// Offers `TICKET` as a TLS1.3 session for any name.
+    #[derive(Debug)]
+    struct SingleTicketStore;
```
cpu: Do we need this? Couldn't you use the public API surface of `ClientSessionMemoryCache` to `insert_tls13_ticket()`?

cpu (commented):
Thanks for the PR. I have some questions/feedback. Please also refer to https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md since some of the feedback is already covered there.

rustls/src/client/hs.rs
cpu: (@ctz would be grateful for your spot-check on this line of thinking)

Mounika2456: Agreed on the sticking out concern. I checked RFC 9849: it pins down copying the encrypted_client_hello extension itself across HRR but is silent on the GREASE PSK, while a genuine offer repeats its identity and age on retry and only recomputes the binders. So regenerating everything is observable by an attacker that can trigger a retry.

Reworked along the lines you suggested: EchState now keeps the GREASE identities and ages from the first hello, and grease_psk() re-offers them on retry, generating fresh binders only. That also covers the accepted-HRR path, where ech_hello() previously built a completely fresh GREASE PSK for the second outer hello and had the same distinguisher. The test now also asserts the retry hello repeats the first hello's identity, and the TLS-ECH-Client* bogo tests pass with the retained behavior.

rustls/src/client/ech.rs
Mounika2456: Done, it now sits between transcript_hrr_update() and encode_inner_hello().

rustls/src/client/ech.rs
Mounika2456: Inlined.

rustls/src/client/ech.rs
Mounika2456: Done, it's an inner fn of the test now (it has two call sites there).

rustls/src/client/ech.rs
Mounika2456: Went with B, it's an inner fn scoped to the test.

rustls/src/client/ech.rs
Mounika2456: Inlined.

rustls/src/client/ech.rs
Mounika2456: Moot now that SingleTicketStore is gone; TICKET is a const scoped to the test.

rustls/src/client/ech.rs
Mounika2456: Done, the test now uses ClientSessionMemoryCache and insert_tls13_ticket() with the real key (config_hash plus the inner name). One snag I hit on the way: sizes up to 8 compute max_servers = 1, and LimitedCache with capacity 1 evicts the entry it just inserted (the oldest.len() == oldest.capacity() check fires right after the first insert), so take_tls13_ticket() never finds anything. The test uses new(256) to steer clear. Looks like a standalone quirk, happy to file it separately.

Mounika2456:
Pushed the rework, all comments addressed inline.

rustls/src/client/ech.rs
```diff
@@ -340,12 +347,17 @@ pub(crate) struct EchState {
     enable_sni: bool,
     // The extensions sent in the inner hello.
     sent_extensions: Vec<ExtensionType>,
+    // The GREASE PSK identities offered in the first outer hello, if any. A retry hello
+    // re-offers the same identities, as a genuine PSK offer would; only the binders are
+    // regenerated.
```
cpu: I think this could be shortened, maybe something like:

> The GREASE PSK identities offered in the first outer hello, if any. A retry hello re-offers the same identities

rustls/src/client/ech.rs
```diff
@@ -949,6 +990,142 @@ mod tests {
...
+            .read(&mut hrr.into_wire_bytes().as_slice())
+            .unwrap();
+        conn.process_new_packets(&mut input)
+            .unwrap();
+
+        // we continue with a second outer hello: it must conceal the ticket too, and
+        // must re-offer the first hello's GREASE identity as a genuine PSK offer would
+        let second = client_hello_written_by(&mut conn);
+        assert_ne!(psk_identity(&second), TICKET);
+        assert_eq!(psk_identity(&second), psk_identity(&first));
```
cpu, line 1087: I think that because of [the fake RNG](https://github.com/rustls/rustls/blob/3d975293347887eda1a3c7b270fd8a3604a66cfc/rustls/src/crypto/test_provider.rs#L43-L55) given by `crypto/test_provider.rs` this would continue to pass even if we regressed saving the initial `grease_psk_identities` into the `EchState`. 

If we generated a second fresh GREASE identity from scratch inappropriately, it would end up wuith the same `b"Rand"` fill by happenstance.

Maybe we should use a tweaked `TEST_PROVIDER` here that implements `SecureRandom` with some kind of counter instead of a fixed fill?

cpu (commented):
Thanks, I think this is looking better.

In addition to the new feedback I added it looks like you have some true positive CI failures to address after your rebase.

Also please resolve previous review conversations once you think they're addressed. We can unresolve them if there's any additional back/forth needed and it makes things clearer for other reviewers.

rustls/src/client/ech.rs
Mounika2456: Done, took your wording.

rustls/src/client/ech.rs
Mounika2456: Good catch, the fixed fill does hide exactly that regression. The test now swaps a counter-backed SecureRandom into TEST_PROVIDER (scoped to this test), so consecutive fills differ. I verified it does its job by temporarily making grease_psk() regenerate on retry: the test fails under the counter and passed under the fixed fill.

Mounika2456:
Rebased onto main and fixed the fallout in the test: Tls13ClientSessionInput now takes Tls13ProtocolSuite::Tcp and a VerifiedIdentity, and the connection driving moved to the build(&mut tls) / process_new_packets(input, &mut tls) API. Full suite, clippy and both fmt configs are green locally.

ctz (approved):

cpu (approved):
Thanks!

rustls/src/client/ech.rs
```diff
@@ -738,41 +787,6 @@ impl EchState {
         Ok(encoded_hello)
     }
 
-    // See https://datatracker.ietf.org/doc/html/rfc9849#name-grease-psk
-    fn grease_psk(&self, psk_offer: &mut PresharedKeyOffer) -> Result<(), Error> {
```
cpu, line 742: Apologies for the late feedback but would you mind splitting this up so that:

1. The first commit on this branch changes `grease_psk` from `fn` to `pub(crate) fn` and moves it up to its new location without changing the body of the function
2. The second commit then does all the other work that's in this commit, modifying `grease_psk()`'s function body to consider `self.grease_psk_identities` and so on.

I think we're ready to merge afterwards.

Mounika2456: Done, split as described: the first commit just moves grease_psk up and makes it pub(super) with the body untouched, the second carries the functional change. The resulting tree is identical to what you approved, and the ech tests are still green.

djc (approved):

djc:
Thanks -- if you want to be able to use this soon, would you be able to submit a backport to the `rel-0.23` branch?

PR #3190, opened 2026-07-31
conn: inline SendPath::send_buffered_plaintext()

djc:


rustls/src/conn/mod.rs
```diff
@@ -259,10 +259,12 @@ impl<Side: SideData> ConnectionCommon<Side> {
                 .sendable_plaintext
                 .is_empty()
```
ctz: Seems like this condition could be dropped now?

ctz (approved):

rustls/src/conn/mod.rs
djc: Yup, good call -- done.

PR #3191, opened 2026-07-31
Only allow one outstanding `KeyUpdateRequest` at time

ctz:
This is one of the TLS1.3-bis functional changes -- see https://www.rfc-editor.org/info/rfc9846/#section-4.7.3-4

### DIFF
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -76,6 +76,14 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
 
     /// Sends a TLS1.3 `key_update` message into `tls` to refresh a connection's keys.
     ///
+    /// The main reason to call this manually is to roll keys when it is known
+    /// a connection will be idle for a long period.
+    ///
+    /// rustls implicitly and automatically refreshes traffic keys when needed
+    /// according to the selected cipher suite's cryptographic constraints.  There
+    /// is therefore no need to call this manually to avoid cryptographic keys
+    /// "wearing out".
+    ///
     /// This call refreshes our encryption keys. Once the peer receives the message,
     /// it refreshes _its_ encryption and decryption keys and sends a response.
     /// Once we receive that response, we refresh our decryption keys to match.
@@ -88,14 +96,6 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     /// Note that other implementations (including rustls) may enforce limits on
     /// the number of `key_update` messages allowed on a given connection to prevent
     /// denial of service.  Therefore, this should be called sparingly.
-    ///
-    /// rustls implicitly and automatically refreshes traffic keys when needed
-    /// according to the selected cipher suite's cryptographic constraints.  There
-    /// is therefore no need to call this manually to avoid cryptographic keys
-    /// "wearing out".
-    ///
-    /// The main reason to call this manually is to roll keys when it is known
-    /// a connection will be idle for a long period.
     fn refresh_traffic_keys(&mut self, tls: &mut Vec<u8>) -> Result<(), Error>;
 
     /// Writes a `close_notify` warning alert into `tls`.
--- a/rustls/src/conn/mod.rs
+++ b/rustls/src/conn/mod.rs
@@ -96,6 +96,9 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     /// Note that other implementations (including rustls) may enforce limits on
     /// the number of `key_update` messages allowed on a given connection to prevent
     /// denial of service.  Therefore, this should be called sparingly.
+    ///
+    /// rustls only allows one outstanding request at a time; this function succeeds
+    /// but sends nothing if a request is already in-flight.
     fn refresh_traffic_keys(&mut self, tls: &mut Vec<u8>) -> Result<(), Error>;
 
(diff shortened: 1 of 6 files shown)


rustls/src/conn/send.rs
```diff
@@ -313,14 +323,36 @@ impl Default for SendPath {
...
+/// State machine for TLS1.3 key updates triggered by peer.
+///
+/// This sits at [`Self::Idle`] for TLS1.2 connections.
+enum KeyUpdateResponseState {
+    /// Nothing is happening.
+    Idle,
+
+    /// A key update response is awaiting sending.
+    Queued(Vec<u8>),
+}
```
djc, line 283: What do these buy us over `bool` and `Option<Vec<u8>>`?

(The `KeyUpdateRequestState` enum makes more sense after removing the last commit.)

rustls/src/conn/send.rs
```diff
@@ -21,8 +22,8 @@ pub(crate) struct SendPath {
     pub(crate) has_sent_close_notify: bool,
     message_fragmenter: MessageFragmenter,
     pub(crate) sendable_tls: ChunkVecBuffer,
-    queued_key_update_message: Option<Vec<u8>>,
-    pub(crate) refresh_traffic_keys_pending: bool,
+    key_update_request: KeyUpdateRequestState,
+    key_update_response: KeyUpdateResponseState,
```
djc: I find this a little harder to read because both the field names and the type names are so similar... maybe something like `outgoing_key_update` and `incoming_key_update`?

rustls/src/conn/split.rs
```diff
@@ -520,11 +520,11 @@ impl SendOutput for SendAdapter<'_> {
             .negotiated_version(version);
     }
 
-    fn ensure_key_update_queued(&mut self) {
+    fn queue_key_update_response(&mut self) {
```
djc: Nit: IMO `_response` is not that obvious and also superfluous.

rustls/src/conn/mod.rs
```diff
@@ -172,6 +172,9 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     /// Note that other implementations (including rustls) may enforce limits on
     /// the number of `key_update` messages allowed on a given connection to prevent
     /// denial of service.  Therefore, this should be called sparingly.
+    ///
+    /// Rustls only allows one outstanding request at a time; this function succeeds
```
djc: Nit: I thought we settled on uncapitalized "rustls" as the canonical name?

rustls/src/conn/send.rs
ctz: renamed both the types and fields

rustls/src/conn/mod.rs
ctz: Fixed; I think I flip-flop day to day so it is quite inconsistent unfortunately!

rustls/src/conn/send.rs
ctz: I think giving names to what the various states mean is valuable.

cpu (approved):

rustls/src/conn/send.rs
```diff
@@ -19,8 +19,8 @@ pub(crate) struct SendPath {
     /// If we signaled end of stream.
     pub(crate) has_sent_close_notify: bool,
     message_fragmenter: MessageFragmenter,
-    queued_key_update_message: Option<Vec<u8>>,
-    pub(crate) refresh_traffic_keys_pending: bool,
+    key_update_local: KeyUpdateLocal,
```
djc, line 22: How about `WriteKeyUpdate` and `ReadKeyUpdate`? Both for a more intuitive word order and because it feels like these keys govern edges rather than vertices.

rustls/src/conn/send.rs
```diff
@@ -168,7 +168,7 @@ impl SendOutput for SendPath {
         self.negotiated_version = Some(version);
     }
 
-    fn ensure_key_update_queued(&mut self) {
+    fn queue_key_update(&mut self) {
```
djc: Nit: is it worth clarifying in the name if this is about the local/remote or read/write key?

rustls/src/conn/send.rs
```diff
@@ -185,6 +192,12 @@ impl SendOutput for SendPath {
         }
     }
 
+    fn note_key_update_response(&mut self) {
```
djc, line 195: Nit: plus direction/side?

djc (approved):

rustls/src/conn/send.rs
ctz: I think this is more confusing really. It would mean `ReadKeyUpdate::Queued()` containing a message that is associated with updating the write key (which is all a key update message can ever do, in addition to requesting the peer does the same to their own write key.)

rustls/src/conn/send.rs
ctz: -> `queue_requested_key_update()` (from the standard naming `KeyUpdateRequest::update_requested`)

rustls/src/conn/send.rs
ctz: to align these with the standard naming this would be `note_key_update_update_not_requested` which is a bit far for me. Any `response` here is only precisely that, which is sent to acknowledge our previous `KeyUpdate`

PR #3193, opened 2026-07-31
Move to externally buffered TLS output

djc:
The last commit here is pretty large, and it undoes some recently carefully designed API (from #3127) that took `&mut [u8]` to take `&mut Vec<u8>` instead. I tried to cut it up into smaller pieces but did not meaningfully succeed as a lot of stuff is intertwined in the current output API. There's still more to do after this, but I think this is a decent way to digest the large changes to get rid of the internal data buffers, which simplifies quite a bit of the internals as well.

I made the library changes and got Claude to help out with converting all the other crates. Keeping it draft for my comprehensive self-review of those changes and to get early feedback from CI.

## Next steps

- Probably use a `TlsOutputBuffer` trait instead of `Vec<u8>`? For one thing, draining from the front of a `Vec` is relatively inefficient. Is this still a practical in-between step?
- This removes the tricky `Drop` interactions from 
  - #3149


cpu:
>  it undoes some recently carefully designed API that took &mut [u8] to take &mut Vec<u8> instead.

Can you provide some justification for why? There's no motivation in the PR desc or any commit messages.

djc:
> > it undoes some recently carefully designed API that took &mut [u8] to take &mut Vec instead.
> 
> Can you provide some justification for why? There's no motivation in the PR desc or any commit messages.

`SendPath::write_appdata_into()` needs to rely on internal API like in `EncryptionState` and I wanted to minimize the number of code paths in there. I think it could make sense to bring it back, though I wonder if passing in `&mut Vec<u8>` doesn't also provide a more ergonomic API at limited cost.

djc:
> > it undoes some recently carefully designed API that took &mut [u8] to take &mut Vec instead.
> 
> Can you provide some justification for why? There's no motivation in the PR desc or any commit messages.

Trying to articulate this more, I think passing in `&mut [u8]` for this level of API is a bad trade-off?

Benefits:

- Hoists allocation out of the library
- Usable in no-alloc scenarios

Costs:

- Requires the caller to pre-allocate
- Requires the caller to pre-initialize
- Requires the caller to loop if they allocated (and initialized) too little, generally externalizing a bunch of bookkeeping

I can splice this into a commit message if y'all agree and think that's useful.

djc:
I think this is ready for review now.

ctz:
> * Probably use a `TlsOutputBuffer` trait instead of `Vec<u8>`? For one thing, draining from the front of a `Vec` is relatively inefficient. Is this still a practical in-between step?

I think moving to the trait is always a possibility, but to admit the idea that providers can avoid a memset by writing directly into a vec's capacity (by pointer) the trait impl would need a surprisingly large and complex surface?

Anyway, if someone wants to write from the vec into a socket and then track/discard written data, they can use `VecDeque` (free conversion from `Vec`). The only issue with `Vec` is the bloat problem, but hopefully this can be thought about carefully at only a few places above this crate.

rustls/src/error/mod.rs
```diff
@@ -1712,6 +1704,12 @@ pub enum ApiMisuse {
         /// The buffer length actually provided
         provided: usize,
     },
+
+    /// Plaintext cannot be encrypted before the handshake is complete.
+    WriteTlsBeforeHandshakeComplete,
+
+    /// Plaintext cannot be decrypted after the send path has been closed.
```
ctz:
```suggestion
    /// Plaintext cannot be encrypted after the send path has been closed.
```

rustls-util/src/lib.rs
```diff
@@ -31,12 +35,11 @@ pub use crate::stream::{Stream, StreamOwned};
 ///
 /// [`is_handshaking()`]: rustls::CommonState::is_handshaking
 /// [`wants_read()`]: rustls::Connection::wants_read
-/// [`wants_write()`]: rustls::Connection::wants_write
-/// [`write_tls()`]: rustls::Connection::write_tls
 /// [`process_new_packets()`]: rustls::Connection::process_new_packets
 pub fn complete_io(
     io: &mut (impl io::Read + io::Write),
     input: &mut VecInput,
+    output: &mut Vec<u8>,
```
ctz, line 44: Eventually this might be better represented in a different way, eg something that is replacing the artist formerly known as `ChunkVecBuffer`

(doesn't need to be in this PR)

rustls/src/client/connection.rs
```diff
@@ -219,13 +213,34 @@ impl ClientConnectionBuilder {
...
     early_data: &'a mut EarlyData,
     common: &'a mut CommonState,
 }
 
 impl<'a> WriteEarlyData<'a> {
+    /// Encrypt early data as TLS records and encode them into `tls`.
+    ///
+    /// Yields the number of bytes of `plaintext` that were consumed.  This may be less than
+    /// the length of `plaintext` if the server has limited the amount of early data that
+    /// may be sent.
```
ctz, line 217: Maybe a note to not retry?

rustls/src/conn/split.rs
ctz: Not returned.

rustls/src/conn/split.rs
```diff
@@ -109,38 +103,9 @@ impl SendTraffic {
     ///
     /// The TLS data to send to the peer is returned. This data should then
```
ctz: No longer returned.

rustls/src/conn/split.rs
```diff
@@ -109,38 +103,9 @@ impl SendTraffic {
...
     /// be communicated to the peer, in order.
```
ctz: Since `take_data()` is gone, this function needs to explain about the utility of calling with `OutboundPlain::new_empty()`

rustls/src/error/mod.rs
```diff
@@ -1768,55 +1766,44 @@ pub use other_error::OtherError;
 
 /// An [`Error`] along with the (possibly encrypted) alert to send to
 /// the peer.
-pub struct ErrorWithAlert {
+///
+/// This borrows the output buffer passed to the operation that failed;
+/// the alert (if one is to be sent) has been appended to that buffer.
+pub struct ErrorWithAlert<'a> {
```
ctz, line 1763: Doesn't the caller have this already, as they passed it in? I think uses of this type can simply become `Error`?

rustls/src/common_state.rs
```diff
@@ -44,22 +44,15 @@ impl CommonState {
...
 
-    /// Returns true if the caller should call [`Connection::write_tls`] as soon as possible.
-    ///
-    /// [`Connection::write_tls`]: crate::Connection::write_tls
-    pub fn wants_write(&self) -> bool {
-        !self.send.sendable_tls.is_empty()
-    }
-
     /// Queues a `close_notify` warning alert to be sent in the next
     /// [`Connection::write_tls`] call.  This informs the peer that the
```
ctz: Out of date docs.

rustls/src/conn/mod.rs
```diff
@@ -28,26 +29,15 @@ pub(crate) use receive::{Input, MessageIter, ReceivePath, TrafficTemperCounters}
...
-    /// Writes TLS messages to `wr`.
-    ///
-    /// On success, this function returns `Ok(n)` where `n` is a number of bytes written to `wr`
-    /// (after encoding and encryption).
-    ///
-    /// After this function returns, the connection buffer may not yet be fully flushed. The
-    /// [`Self::wants_write()`] function can be used to check if the output buffer is
-    /// empty.
-    fn write_tls(&mut self, wr: &mut dyn io::Write) -> Result<usize, io::Error>;
+    /// Writes the application data from `plaintext` into TLS records and appends them to `tls`.
```
ctz, line 37: This comment should explain the constraints on when it can be called.

rustls/src/conn/mod.rs
ctz: Stale docs

rustls/src/crypto/cipher/record_layer.rs
ctz: Stale docs
```suggestion
   /// The result including framing is appended to `output`.
```

rustls/src/conn/split.rs
ctz: stale docs

rustls/src/conn/mod.rs
```diff
@@ -172,14 +121,14 @@ pub trait Connection: Debug + Deref<Target = ConnectionOutputs> {
     ///
     /// The main reason to call this manually is to roll keys when it is known
     /// a connection will be idle for a long period.
-    fn refresh_traffic_keys(&mut self) -> Result<(), Error>;
+    fn refresh_traffic_keys(&mut self, tls: &mut Vec<u8>) -> Result<(), Error>;
 
     /// Queues a `close_notify` warning alert to be sent in the next [`Self::write_tls`] call.
```
ctz: Stale.

rustls/src/conn/mod.rs
```diff
@@ -252,35 +202,38 @@ impl<Side: SideData> ConnectionCommon<Side> {
...
+
+    pub(crate) fn write_tls(
+        &mut self,
+        plaintext: OutboundPlain<'_>,
+        tls: &mut Vec<u8>,
+    ) -> Result<(), Error> {
+        if plaintext.is_empty() {
+            return Ok(());
+        } else if !self.send.may_send_application_data {
+            return Err(ApiMisuse::WriteTlsBeforeHandshakeComplete.into());
```
ctz, line 155: Needs a corresponding docs update to `is_handshaking()` ("During this time plaintext written to the connection is buffered in memory")

ctz (approved):
Directionally good, mostly docs comments.

rustls/src/client/connection.rs
djc: What do you mean?

rustls/src/error/mod.rs
djc: I thought about that, but wanted to express in the type system that the caller should not just yield the error to the caller but do something more here. We could/should maybe use that in more places?

djc:
Thanks for the docs review -- got a bit hasty towards the end due to the start of my vacation.

rustls/src/conn/split.rs
```diff
@@ -236,10 +182,11 @@ impl<Side: SideData> ReceiveTraffic<Side> {
     /// which can be obtained from the returned [`ErrorWithAlert`] and sent
     /// to the peer. Following this, the underlying IO medium should be
     /// closed by the application.
-    pub fn read<'a>(
+    pub fn read<'a, 't>(
         self,
         input: &'a mut impl TlsInputBuffer,
-    ) -> Result<ReceiveTrafficState<'a, Side>, ErrorWithAlert> {
+        tls: &'t mut Vec<u8>,
```
ctz, line 188: Unfortunately there cannot be a `tls` here. It would be the send-side buffer. This is the receive-side object, it really cannot have the send-side buffer.

Because of this, stuff like `send_alert` (formerly a stand-alone messaging passing interface between the two sides) now doesn't actually work. It would write consecutive messages to different thread's send buffers with no way for an application to send them in the correct order.

PR #3194, opened 2026-08-03
Move from `log` to `tracing`

### LINKED ISSUE #3122: Proposal: migrate from log to tracing (with log compatibility feature)

ctz:
This picks up #3133 and runs with it, removing `log` and `env_logger` throughout from tests, examples, and documentation. Thanks to @FalkWoldmann for the work here!

fixes #3133 fixes #3122


rustls-test/src/lib.rs
```diff
@@ -2446,3 +2447,104 @@ pub mod encoding {
...
+}
+
+impl CountingSubscriber {
+    pub fn new() -> Self {
+        let mut subs = Self {
+            inner: Arc::new(CountingInner::default()),
+            guard: None,
+        };
+        subs.guard = Some(subscriber::set_default(subs.inner.clone()));
+        subs
```
djc: Nit: can this be written more like

```rust
let inner = Arc::new(CountingInner::default());
Self {
     guard: Some(subscriber::set_default(inner.clone())),
     inner,
}
```

?

rustls-test/src/lib.rs
```diff
@@ -2446,3 +2447,104 @@ pub mod encoding {
...
+        self.inner.data.lock().unwrap().clone()
+    }
+}
+
+#[derive(Debug)]
+struct CountingInner {
+    data: Mutex<CountingData>,
+}
+
+impl Default for CountingInner {
```
djc: Nit: I think this can be derived?

djc (approved):
LGTM!

PR #3195, opened 2026-08-03
docs: fix nightly docs search and external links

ctz:
Due to our abuse of rustdoc, search on https://rustls.dev/docs/ links to results like https://rustls.dev/rustls/client/fn.verify_identity_signed_by_trust_anchor.html

Despite use of --no-deps, rustdoc still generates links to dependencies for external links, these get picked up by lychee when they appear on (eg) inherent impls.

This breaks all existing links, however this is the zen of nightly documentation (removing/renaming a type also has this effect.)

(will fix docs CI in #3194)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -72,8 +72,9 @@ jobs:
           # lockfile causes deployment step to go wrong, due to permissions
           rm -f target/doc/.lock
           # move the result into website root
-          mv target/doc/* target/website/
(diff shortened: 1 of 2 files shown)


ctz:
https://github.com/rustls/rustls/actions/runs/30814260383 worked and currently https://rustls.dev/docs/ shows the desired output

cpu (approved):

cpu (approved):

PR #3198, opened 2026-08-06
Fix malleability of TLS1.3 record headers

ctz:
This started as a testing gap found while reviewing #3112 

While this is malleability of network data, the header fields are not actually consulted for anything.

Unfortunately cannot be easily backported.

### DIFF
--- a/rustls-aws-lc-rs/src/tls13.rs
+++ b/rustls-aws-lc-rs/src/tls13.rs
@@ -228,8 +228,9 @@ impl MessageEncrypter for AeadMessageEncrypter {
         let total_len = self.encrypted_payload_len(msg.payload.len());
         let mut payload = EncryptBuffer::new(out, total_len)?;
 
+        let typ = ContentType::ApplicationData;
         let nonce = aead::Nonce::assume_unique_for_key(Nonce::new(&self.iv, seq).to_array()?);
-        let aad = aead::Aad::from(make_tls13_aad(total_len));
+        let aad = aead::Aad::from(make_tls13_aad(typ, TLS13_LEGACY_RECORD_VERSION, total_len));
         payload.extend_from_chunks(&msg.payload);
         payload.extend_from_slice(&msg.typ.to_array());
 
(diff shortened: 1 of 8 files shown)


rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -294,8 +296,10 @@ impl MessageEncrypter for GcmMessageEncrypter {
         let total_len = self.encrypted_payload_len(msg.payload.len());
         let mut payload = EncryptBuffer::new(out, total_len)?;
 
+        let version = ProtocolVersion::TLSv1_2; // sic
```
cpu: nit: I think "sic" is a bit jargon-y and missing broader context to make it understandable here outside the context of this commit.

cpu (approved):
Nice 👍

rustls-aws-lc-rs/src/tls13.rs
ctz: extracted const with the pre-existing comment

djc:
I'm still a little confused about what this is fixing. It seems to me that the previous definition of `make_tls13_aad()` makes more sense given the description in RFC 9846 section 5.2... In particular, it seems weird to have `let typ = ContentType::ApplicationData;` in an encrypter that can also be used for handshake packets.

ctz:
> I'm still a little confused about what this is fixing. It seems to me that the previous definition of `make_tls13_aad()` makes more sense given the description in RFC 9846 section 5.2... 

The previous definition doesn't contain any of the data off the wire, which is what the AAD is meant to contain. The most important part of section 5.2 is "the additional data input is the record header".

> In particular, it seems weird to have `let typ = ContentType::ApplicationData;` in an encrypter that can also be used for handshake packets.

The outer `ContentType` that appears on the wire, aka `opaque_type`, is fixed as `application_data` irrespective of the actual underlying type (which is encrypted). The reason for all this is middleboxes. But even though it is fixed that doesn't mean we can construct our own header for received packets just to get the AAD right.

djc:
Okay, with fresh eyes I better understand what you're getting at -- thanks.

PR #3199, opened 2026-08-07
conn: use associated type for Connection's side

djc:
Generally seems like an improvement.

I think we were already monomorphizing in practice due to taking the `SideData` as a generic argument anyway.


cpu (approved):
Seems like a nice improvement to me 👍

ctz (approved):
Nice!

PR #3200, opened 2026-08-09
Push crypto algorithm slices down into backing modules

djc:


djc:
Filed

- https://github.com/rust-lang/rust-clippy/issues/17529

for the nightly clippy failures.

ctz (approved):

PR #3201, opened 2026-08-09
reject zero-length ResponderID in ocsp status_request

Mounika2456:
`CertificateStatusRequest::read`, one zero-length ResponderID in an OCSP `status_request` (`01 00 02 00 00 00 00`):

    Ok(Ocsp(OcspCertificateStatusRequest { responder_ids: [ResponderId()], extensions:  }))

RFC 6066 has `opaque ResponderID<1..2^16-1>`, so an empty entry is not a legal encoding. `ResponderId` wrapped `SizedPayload<u16, MaybeEmpty>`, which skips the length check; the sibling opaque vectors here (`PresharedKeyIdentity.identity`, `PresharedKeyBinder`, `DistinguishedName`) all use `NonEmpty` with the RFC bound cited. Switch `ResponderId` to `NonEmpty` so an empty entry is refused with `IllegalEmptyList`.

### DIFF
--- a/rustls/src/msgs/client_hello.rs
+++ b/rustls/src/msgs/client_hello.rs
@@ -708,7 +708,10 @@ impl Codec<'_> for OcspCertificateStatusRequest {
     }
 }
 
-wrapped_payload!(pub(crate) struct ResponderId, SizedPayload<u16, MaybeEmpty>,);
+wrapped_payload!(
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks, this looks good! Not sure whether the OpenSSL interop tests are failing because of this change?

ctz (approved):

PR #3203, opened 2026-08-10
aws-lc-rs 1.18: stable MLDSA & out-of-place AEAD encrypter support

cpu:
This branch is picking up the AEAD API optimization discussed in https://github.com/aws/aws-lc-rs/issues/1182

I'm playing with an idea to reduce some of the duplication in the `MessageEncrypter` impls but I'm not confident it'll be a productive avenue and in the mean time we don't need to block the optimization work on that.

### DIFF
--- a/rustls-aws-lc-rs/src/lib.rs
+++ b/rustls-aws-lc-rs/src/lib.rs
@@ -45,7 +45,7 @@ use rustls::crypto::{
     CryptoProvider, GetRandomFailed, KeyProvider, SecureRandom, SigningKey, TicketProducer,
     TicketerFactory,
 };
-use rustls::error::{Error, OtherError};
+use rustls::error::{ApiMisuse, Error, OtherError};
 #[cfg(feature = "std")]
 use rustls::ticketer::TicketRotator;
 
@@ -248,6 +248,21 @@ mod ring_shim {
         agreement::agree_ephemeral(priv_key, peer_key, (), |secret| {
             Ok(SharedSecret::from(secret))
         })
+    }
+}
+
+/// The region of `out` that a `len`-byte sealed record payload will occupy.
+///
+/// If `out` is shorter than `len` bytes, this returns [`ApiMisuse::EncryptBufferTooSmall`].
+fn record_region(out: &mut [u8], len: usize) -> Result<&mut [u8], Error> {
+    let provided = out.len();
+    match out.get_mut(..len) {
+        Some(record) => Ok(record),
+        None => Err(ApiMisuse::EncryptBufferTooSmall {
+            required: len,
+            provided,
+        }
+        .into()),
     }
 }
 
--- a/rustls-aws-lc-rs/src/tls12.rs
+++ b/rustls-aws-lc-rs/src/tls12.rs
@@ -16,7 +16,7 @@ use rustls::version::TLS12_VERSION;
 use rustls::{CipherSuiteCommon, ConnectionTrafficSecrets, Tls12CipherSuite};
 use zeroize::Zeroizing;
 
-use crate::MAX_FRAGMENT_LEN;
+use crate::{MAX_FRAGMENT_LEN, record_region};
 
 /// The TLS1.2 cipher suite configuration that an application should use by default.
 ///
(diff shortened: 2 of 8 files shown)


cpu:
@rustls-benchmarking bench

cpu:

These look like the positive results I was expecting :racehorse:

rustls-aws-lc-rs/src/lib.rs
```diff
@@ -251,6 +252,33 @@ mod ring_shim {
...
+/// The payload's single contiguous chunk, or `None` if it is fragmented.
+///
+/// An empty payload is a treated as a single empty chunk.
+fn single_chunk<'p>(payload: &'p OutboundPlain<'_>) -> Option<&'p [u8]> {
+    let mut chunks = payload.chunks();
+    let first = chunks.next().unwrap_or(&[]);
+    match chunks.next() {
+        Some(_) => None,
+        None => Some(first),
+    }
```
djc: Seems like this could just be

```rust
match payload {
     OutboundPlain::Single(inner) => Some(inner),
     _ => None,
}
```

rustls-aws-lc-rs/src/lib.rs
```diff
@@ -251,6 +252,33 @@ mod ring_shim {
...
+}
+
+/// The region of `out` that a `len`-byte sealed record payload will occupy.
+///
+/// If `out` is shorter than `len` bytes, this returns [`ApiMisuse::EncryptBufferTooSmall`].
+fn record_region(out: &mut [u8], len: usize) -> Result<&mut [u8], Error> {
+    let provided = out.len();
```
djc, line 258: Nit: inline the `provided` binding?

rustls-aws-lc-rs/src/tls12.rs
```diff
@@ -334,26 +334,47 @@ impl MessageEncrypter for GcmMessageEncrypter {
...
-        match self.enc_key.seal_in_place_separate_tag(
-            nonce,
-            aad,
-            &mut payload.as_mut()[GCM_EXPLICIT_NONCE_LEN..],
-        ) {
-            Ok(tag) => payload.extend_from_slice(tag.as_ref()),
-            Err(_) => return Err(Error::EncryptError),
-        }
+
+        let payload = match super::single_chunk(&msg.payload) {
```
djc: Nit: import `single_chunk()` and `record_region()`?

djc (approved):
Cool!

rustls-aws-lc-rs/src/lib.rs
cpu: I needed to use this binding because of the mutable borrow of `out` in the `match` input.

rustls-aws-lc-rs/src/tls12.rs
cpu: Done.

rustls-aws-lc-rs/src/lib.rs
cpu: Oh yeah! That's much simpler. Thank you. Fixed up.

djc:
(Maybe also sort `single_chunk()` below `record_region()` now? It could potentially even become an `OutboundPlain` method.)

rustls/src/crypto/cipher/messages.rs
```diff
@@ -254,6 +254,17 @@ impl<'a> OutboundPlain<'a> {
...
     }
 
+    /// The payload's single contiguous chunk, or `None` if it is fragmented.
+    ///
+    /// An empty payload is treated as a single empty chunk, and a [`Self::Multiple`]
+    /// payload yields `None` even when its chunks happen to form a contiguous whole.
+    pub fn single_chunk(&self) -> Option<&'a [u8]> {
+        match *self {
+            Self::Single(chunk) => Some(chunk),
+            Self::Multiple { .. } => None,
```
ctz, line 264: Seems fine but note the "Warning: OutboundPlain does not guarantee that the simplest variant is used." warning at the top of this code. This is merely a missed opportunity rather than a correctness issue.

ctz (approved):

rustls/src/crypto/cipher/messages.rs
cpu: >  This is merely a missed opportunity rather than a correctness issue.

Indeed, and that's what I was hoping to hint at with this function also documenting itself with "a `Self::Multiple` payload yields `None` even when its chunks happen to form a contiguous whole." :+1:

PR #3204, opened 2026-08-11
enforce handshake size limit for non-coalesced fragments

Mounika2456:
`coalesce()` returns `Ok(())` for a single oversized fragment:

        assert_eq!(deframer.coalesce(&mut input), Err(HandshakePayloadTooLarge));
        left:  Ok(())
        right: Err(HandshakePayloadTooLarge)

The 64KB `MAX_HANDSHAKE_SIZE` limit is only enforced over spans that get re-dissected while fragments are merged. A message that needs no coalescing skips the check via the early `return Ok(())`. Over QUIC this is reachable pre-auth: `input_quic()` hands the whole application-reassembled CRYPTO buffer to `input_message()`, which produces one complete span whose declared u24 length can be up to the 16MB wire limit, and `coalesce()` then accepts it, so the oversized message is parsed and `into_owned()`-copied. The TCP path already rejects these while coalescing, so this also brings QUIC in line with it.

Moving the check to the loop-exit path covers every span, including an incomplete one whose header length is already known (so an oversized message is rejected as soon as its header is seen, without buffering the body). The existing `coalesce_rejects_excess_size_message` test only exercises the split-header coalescing path, so the added test delivers a single oversized fragment.

### DIFF
--- a/rustls/src/msgs/deframer/mod.rs
+++ b/rustls/src/msgs/deframer/mod.rs
@@ -29,6 +29,9 @@ pub(crate) struct Deframer {
     /// Spans covering individual handshake payloads, in order of receipt.
     spans: VecDeque<FragmentSpan>,
 
+    /// Whether any span declared a size exceeding `MAX_HANDSHAKE_SIZE`.
+    too_large: bool,
+
     /// Prefix of the buffer that has been processed so far.
     ///
     /// `processed` may exceed `discard`, that means we have parsed
@@ -151,6 +154,7 @@ impl Deframer {
         };
 
         for span in iter {
+            self.too_large |= span.size.unwrap_or_default() > MAX_HANDSHAKE_SIZE;
             self.spans.push_back(span);
         }
     }
@@ -213,6 +217,12 @@ impl Deframer {
         // for a pair where the first is not complete.  move
         // the second down towards the first, then reparse the contents.
         loop {
+            // this covers spans that never require coalescing, eg. a single
+            // complete QUIC handshake message with an excessive declared size
+            if self.too_large {
+                return Err(InvalidMessage::HandshakePayloadTooLarge);
+            }
+
             let limit = self.spans.len().saturating_sub(1);
             let iter = self.spans.iter();
             let Some(index) = iter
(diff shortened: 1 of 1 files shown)


rustls/src/msgs/deframer/mod.rs
```diff
@@ -259,17 +271,9 @@ impl Deframer {
                 bounds: first.bounds.start..first.bounds.end,
             };
 
-            let mut too_large = false;
```
djc, line 262: Seems like this removes the size checks from the multi-fragment path, is that intentional? We should probably have test coverage for both...

djc:
Good find!

rustls/src/msgs/deframer/mod.rs
Mounika2456: Yes, that's intentional. The loop only ever exits through that first branch (the two `remove` else-arms can't trigger since `index < len - 1`), and the check there runs over every span once all merging is done, so the multi-fragment path still hits it. The only behavioral difference is the error surfaces at loop exit instead of mid-merge.

Both paths are covered: the existing `coalesce_rejects_excess_size_message` exercises the split-header multi-fragment case, and the new test covers the single-fragment case. To make sure they both actually depend on the relocated check, I neutered it locally and both tests fail without it.

rustls/src/msgs/deframer/mod.rs
```diff
@@ -220,7 +220,19 @@ impl Deframer {
                 .take(limit)
                 .find_map(|(i, span)| (!span.is_complete()).then_some(i))
             else {
-                return Ok(());
+                // No further fragments need coalescing.  Enforce the size limit over
+                // every span: one that never required coalescing (a single complete
+                // QUIC handshake message, or a single record whose declared length
+                // already exceeds the limit) is not covered by the check in the loop
+                // body below.
```
djc: Nit: this looks like a typically verbose generated comment. Suggest editing it down to one or at most two lines?

(If you rebase as well, the nightly clippy CI issue should go away.)

djc (approved):
Thanks! I guess we should probably backport this to `rel-0.23` as well...

rustls/src/msgs/deframer/mod.rs
djc: Ah yeah, you're right -- sorry for getting confused.

djc:
There is a bit of a performance penalty here. Maybe it would be better after all to keep track of spans that are too large while adding them, instead of checking each span after the fact before returning?

ctz (approved):

cpu (approved):

rustls/src/msgs/deframer/mod.rs
Mounika2456: Trimmed, and rebased onto main.

Mounika2456:
Reworked it as you suggested: the deframer now carries a too_large flag that gets set when a span with an oversized declared length is added (in input_message and during re-dissection), and coalesce() just checks the flag on entry. That removes the per-call scan over all spans, so the hot path is a single bool check. Both size-limit tests still fail if the check is neutered, so coverage holds for the single-fragment and split-header paths.

Happy to open the rel-0.23 backport once this lands.

ctz (approved):

ctz:
Thanks!

PR #3205, opened 2026-08-11
rustls-aws-lc-rs: refactor encrypter/decrypters, share more between protocols

cpu:
This is a follow-up to https://github.com/rustls/rustls/pull/3203 trying to dedupe/tighten up some of the `aws-lc-rs` provider code. I think it might also be helpful as a partial step towards something like https://github.com/rustls/rustls/issues/3109. I'm sort of imagining next steps involving lifting some of the `seal_record()` & `Framing` bits into something that's broader than a single provider. Happy to take that on next if folks are interested.

### DIFF
--- a/rustls-aws-lc-rs/src/lib.rs
+++ b/rustls-aws-lc-rs/src/lib.rs
@@ -40,6 +40,8 @@ use alloc::sync::Arc;
 #[cfg(feature = "std")]
 use core::time::Duration;
 
+use aws_lc_rs::aead;
+use aws_lc_rs::error::Unspecified;
 use pki_types::{FipsStatus, PrivateKeyDer};
 use rustls::crypto::{
     CryptoProvider, GetRandomFailed, KeyProvider, SecureRandom, SigningKey, TicketProducer,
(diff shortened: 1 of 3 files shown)


cpu:
@rustls-benchmarking bench

rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -324,99 +327,86 @@ impl MessageDecrypter for AeadMessageDecrypter {
     }
 }
 
-struct GcmMessageEncrypter {
-    enc_key: aead::TlsRecordSealingKey,
-    iv: Iv,
+/// A TLS1.3 sealing key, dispatching to whichever aws-lc-rs key type backs it.
+enum SealKey {
```
djc, line 331: Nit: `SealingKey`, to align with the crypto provider?

rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -324,99 +327,86 @@ impl MessageDecrypter for AeadMessageDecrypter {
...
+            Self::TlsRecord(key) => key.algorithm(),
+        }
     }
 }
 
-struct GcmMessageDecrypter {
-    dec_key: aead::TlsRecordOpeningKey,
-    iv: Iv,
+/// A TLS1.3 opening key, dispatching to whichever aws-lc-rs key type backs it.
+enum OpenKey {
```
djc, line 387: Nit: `OpeningKey`.

rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -324,99 +327,86 @@ impl MessageDecrypter for AeadMessageDecrypter {
...
-
-                payload.into_written()
-            }
-        };
+        nonce: aead::Nonce,
+        aad: aead::Aad<A>,
+        in_plaintext: &[u8],
+        out_ciphertext: &mut [u8],
+        extra_in: &[u8],
+        extra_out_and_tag: &mut [u8],
```
djc, line 344: Could/should these be wrapped in a named field struct?

rustls-aws-lc-rs/src/tls13.rs
```diff
@@ -324,99 +327,86 @@ impl MessageDecrypter for AeadMessageDecrypter {
...
-                {
-                    Ok(tag) => payload.extend_from_slice(tag.as_ref()),
-                    Err(_) => return Err(Error::EncryptError),
-                }
-
-                payload.into_written()
-            }
-        };
+        nonce: aead::Nonce,
+        aad: aead::Aad<A>,
```
djc, line 340: Nit: can we write `aead::Aad<impl AsRef<[u8]>>` instead?

rustls-aws-lc-rs/src/lib.rs
```diff
@@ -253,6 +254,84 @@ mod ring_shim {
...
+    None,
+    /// These bytes prefix the ciphertext on the wire, unencrypted.
+    ///
+    /// This is used by TLS1.2 GCM's explicit nonce.
+    UnencryptedPrefix(&'a [u8]),
+    /// These bytes follow the ciphertext on the wire, encrypted.
+    ///
+    /// This is used by TLS1.3's inner content type byte.
+    EncryptedTail(&'a [u8]),
+}
```
djc, line 333: Could we instead use something like this to unify the 1.2 and 1.3 encrypter and decrypter traits in rustls itself? Maybe like

- https://github.com/rustls/rustls/issues/3109

djc:
+236/-229 is a little disappointing for something that tries to share more code. Just one heuristic, of course...

rustls-aws-lc-rs/src/lib.rs
cpu: Did you read the PR description?

> I think it might also be helpful as a partial step towards something like https://github.com/rustls/rustls/issues/3109. I'm sort of imagining next steps involving lifting some of the seal_record() & Framing bits into something that's broader than a single provider. Happy to take that on next if folks are interested.

djc: > Did you read the PR description?
> 
> > I think it might also be helpful as a partial step towards something like #3109. I'm sort of imagining next steps involving lifting some of the seal_record() & Framing bits into something that's broader than a single provider. Happy to take that on next if folks are interested.

I did, although admittedly only after my review...

Feels to me like it's better to attack that problem directly rather than via this route, because the constraints might lead you in a different direction?

tgeoghegan: See my latest note in #3109 as well as my draft PR https://github.com/tgeoghegan/rustls/pull/2 for a sketch of where I think the crypto providers can go. I hope that helps inform whether this change is a step in that direction.

cpu: Since you're further along I'm going to close this and you can run with your sketch.

PR #3206, opened 2026-08-11
Account for encryption overhead in `max_fragment_size` fragmentation

### LINKED ISSUE #991: MessageFragmenter doesn't take account of encryption overhead

ctz:
This takes #3141 forward, and simplifies it significantly.

fixes #3141 fixes #991

### DIFF
--- a/rustls/src/crypto/cipher/mod.rs
+++ b/rustls/src/crypto/cipher/mod.rs
@@ -178,8 +178,13 @@ pub trait MessageEncrypter: Send + Sync {
         out: &'a mut [u8],
     ) -> Result<EncodedMessage<&'a [u8]>, Error>;
 
-    /// Return the length of the ciphertext that results from encrypting plaintext of
-    /// length `payload_len`
+    /// Return the length of the ciphertext that results from encrypting plaintext of length `payload_len`.
+    ///
+    /// For a zero `payload_len` this should return the _minimum_ overhead for any
+    /// message.  Then, to fragment a long message into chunks of length `F`,
+    /// Rustls will first set `A := encrypted_payload_len(0)` and then supply the
+    /// message to [`Self::encrypt()`] in chunks of length `F - A`.  Each `encrypt()`
+    /// is then free to pad or otherwise transform the length at its option.
     fn encrypted_payload_len(&self, payload_len: usize) -> usize;
 }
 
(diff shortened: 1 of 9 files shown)


rustls/src/crypto/cipher/mod.rs
```diff
@@ -179,7 +179,13 @@ pub trait MessageEncrypter: Send + Sync {
     ) -> Result<EncodedMessage<&'a [u8]>, Error>;
 
     /// Return the length of the ciphertext that results from encrypting plaintext of
-    /// length `payload_len`
+    /// length `payload_len`.
```
djc: Nit: if you're going to fix this up, also try to tighten up the first sentence into a single line?

rustls/src/msgs/fragmenter.rs
```diff
@@ -31,8 +33,13 @@ impl MessageFragmenter {
         typ: ContentType,
         version: ProtocolVersion,
         payload: OutboundPlain<'a>,
+        encryption_overhead: usize,
     ) -> impl ExactSizeIterator<Item = EncodedMessage<OutboundPlain<'a>>> + use<'a> {
-        Chunker::new(payload, self.max_frag).map(move |payload| EncodedMessage {
+        let max_plaintext = self
+            .max_frag
+            .saturating_sub(encryption_overhead)
+            .max(1);
```
djc: This `max()` feels a little weird. Maybe `checked_sub()` would be better than `saturating_sub()` in this case, with an explicit alternative?

rustls/src/msgs/fragmenter.rs
```diff
@@ -19,28 +19,14 @@ impl Default for MessageFragmenter {
 }
 
 impl MessageFragmenter {
-    /// Take `msg` and fragment it into new messages with the same type and version.
```
djc, line 22: While you're in here, maybe

- Rename it to just `Fragmenter`? Doesn't seem likely to get confused.
- Move the `Default` impl below the inherent impl.

rustls/src/crypto/cipher/mod.rs
ctz: ok, makes it looks weird with the prevailing wrapping here though

rustls/src/msgs/fragmenter.rs
ctz: added an earlier commit which expresses the problem here, using `NonZeroUsize`

djc (approved):

PR #3208, opened 2026-08-12
util: create SSLKEYLOGFILE with owner-only permissions

### LINKED ISSUE #910: SSLKEYLOGFILE should be created with restrictive file permissions.
In many cases `$SSLKEYLOGFILE` will point into the user's `$HOME` and in many Unixy operating systems, files in `$HOME` are readadable by everybody by default. `$SSLKEYLOGFILE` is contains (potentially) very sensitive information, we should open it with restrictive permissions, which means we should probably apply more restrictive file permissions when creating the file, when the OS allows us to do so.

SebTardif:
## Motivation

`KeyLogFile` creates `$SSLKEYLOGFILE` with the default file creation mode
(`0o666` masked by umask). With a typical umask of `0o022` the file is
world-readable (`0o644`). The file holds TLS traffic secrets.

Closes #910

## Change

On Unix, open with `OpenOptionsExt::mode(0o600)` when creating the file.
Existing files keep their current permissions (mode applies only on create).

## Testing

Unit test asserts a newly created path has mode `0o600`.

Without the fix the test fails with mode `0o644` under a normal umask.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,6 +124,11 @@ too_many_arguments = "allow"
 #   helpful
 new_without_default = "allow"
 
+# - double_must_use: async-trait expands async methods to Pin<Box<dyn Future>>,
+#   which is already must_use, and also emits a bare #[must_use]. Nightly
+#   clippy 1.99+ flags that combination; not actionable on our source.
+double_must_use = "allow"
+
 [workspace.lints.rust]
 elided_lifetimes_in_paths = "warn"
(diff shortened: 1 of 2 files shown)


ctz (approved):
Thanks!

djc (approved):

djc:
Should we backport this to 0.23?

ctz:
> Should we backport this to 0.23?

Seems like a good candidate, yes.

PR #3209, opened 2026-08-13
server: stop buffering early data

djc:
Instead, `MessageHandler` grows a `next_early_payload()` method that exclusively yields early data payloads.

Deletes the `ChunkVecBuffer`.


rustls/src/conn/receive.rs
```diff
@@ -173,9 +176,11 @@ impl<'a, 'm, Side: SideData, Send: SendOutput + 'a> MessageIter<'a, 'm, Side, Se
                 break;
             }
 
-            if let Some(payload) = plaintext.take() {
-                *self.state = Ok(st);
-                return Some(Ok(payload));
+            if st.is_traffic() || allow_early_data {
```
ctz, line 179: I think a caller with `allow_early_data: true` actually wants a stronger guarantee here? (ie, `only_early_data_never_normal_application_data`) That any data they receive was certainly received as early data, rather than getting some mix of early and then normal data.

Sorry if we already have a test for this, but if not I think we should have one (a client that sends early data and normal data, and a server which receives both but with the two streams kept totally separate)

rustls/src/conn/mod.rs
```diff
@@ -168,8 +170,9 @@ impl<Side: SideData> NeedsInput<Side> {
             &mut self.inner,
             MessageIterMode::Handshake,
         );
+
         let result = loop {
-            match iter.next() {
+            match iter.next(false) {
```
djc, line 175: This drops early data on the floor. I figured that might be okay for now until

- https://github.com/rustls/rustls/pull/3278

rustls/src/conn/receive.rs
```diff
@@ -187,12 +187,21 @@ impl<'a, 'm, Side: SideData, Send: SendOutput + 'a> MessageIter<'a, 'm, Side, Se
                 break;
             }
 
-            if let Some(payload) = plaintext.take() {
-                *self.state = Ok(st);
-                return Some(Ok(payload));
+            match plaintext.take() {
+                Some(DataKind::Traffic(payload)) => {
```
ctz: `if !early_only` on this arm? A caller of `next_early_data()` should not receive normal app data?

rustls/src/conn/receive.rs
```diff
@@ -76,7 +76,7 @@ impl<'a, 'm, 's, Side: SideData> MessageIter<'a, 'm, Side, SendAdapter<'s>> {
 }
 
 impl<'a, 'm, Side: SideData, Send: SendOutput + 'a> MessageIter<'a, 'm, Side, Send> {
-    pub(crate) fn next(&mut self) -> Option<Result<UnborrowedPayload, Error>> {
+    pub(crate) fn next(&mut self, early_only: bool) -> Option<Result<UnborrowedPayload, Error>> {
```
ctz, line 79: `early_only` could be const i guess

ctz (approved):

rustls/src/conn/receive.rs
djc: That would cause `MessageIter::next()` to be monomorphized, including two copies. Not sure it's a great trade-off?

rustls/src/conn/mod.rs
djc: @ctz just checking, did you see this?

rustls/src/conn/receive.rs
djc: Good idea -- added.

rustls/src/conn/receive.rs
ctz: fair!

rustls/src/conn/mod.rs
ctz: yep, that should be fine. that PR will need quite a bit of rework anyhow

Issue #911, opened 2022-01-07
Use `secure_getenv` or equivalent to read $SSLKEYLOGFILE


briansmith:
According to https://gitlab.com/gnutls/gnutls/-/issues/802, gnutls uses `secure_getenv` for this feature.

The risk for a a setuid/setgid executable in this feature is that SSLKEYLOGFILE might have been (in fact, probably was) inherited from another user's environment, so its value is controlled by another user. Thus, this feature could be abused, if used by a setuid/setgid executable, to overwrite any file, or to log the keys anywhere that the executable can write to, at the choosing of another user. On the one hand, that's the fault of the setuid/setgid executable for not clearing its environment, but this seems like a pretty big footgun.

OpenSSL uses command-line flags and never looks at this environment variable.

See also #899.

PR #3212, opened 2026-08-13
util: document SSLKEYLOGFILE risks for elevated processes

SebTardif:
## Summary

Addresses #911 with documentation rather than a `secure_getenv`-style filter.

Maintainer feedback on the earlier approach was that the platform-specific / `unsafe` cost was not worth it for a debug-oriented util: prefer not enabling key logging outside debugging, and let apps that need stronger env handling supply their own `KeyLog`.

## Changes

- Document setuid/elevated inheritance risk on `KeyLogFile`
- `admin/coverage`: run `rustls-util` unit tests (not a workspace default-member)

## Non-goals

No platform privilege probes and no new `unsafe` in `KeyLogFile`.

### DIFF
--- a/admin/coverage
+++ b/admin/coverage
@@ -14,6 +14,8 @@ cargo test --locked --all-features # this presumably enables `fips`
 cargo test -p rustls-ring --locked
 cargo test -p rustls-aws-lc-rs --locked
 cargo test -p rustls-test --locked --features aws-lc-rs # aws-lc-rs, but no fips
+# rustls-util isn't a workspace default-member, test it explicitly.
+cargo test -p rustls-util --locked --all-features
 
 # ensure both zlib and brotli are tested, irrespective of their order
 cargo test --locked $(admin/all-features-except zlib rustls)
--- a/rustls-util/src/key_log_file.rs
+++ b/rustls-util/src/key_log_file.rs
@@ -87,6 +87,23 @@ impl Debug for KeyLogFileInner {
 ///
 /// If such a file cannot be opened, or cannot be written then
 /// this does nothing but logs errors at warning-level.
+///
+/// # Security
+///
+/// Key material is extremely sensitive. Prefer not enabling `KeyLog`
+/// outside of local debugging. On Unix, files created by this type use
+/// owner-only permissions (`0o600`).
+///
+/// This type reads `SSLKEYLOGFILE` from the process environment with a
+/// normal environment lookup. In a setuid/setgid (or otherwise elevated)
+/// process, that variable may have been inherited from an untrusted parent
+/// and could point at an attacker-chosen path. Applications that raise
+/// privileges should clear sensitive environment variables, avoid compiling
+/// key-log support into production builds, or supply their own [`KeyLog`]
+/// implementation.
+///
+/// This util is intentionally small. It does not attempt a full
+/// `secure_getenv`-style environment filter.
 pub struct KeyLogFile(Mutex<KeyLogFileInner>);
 
 impl KeyLogFile {


rustls-util/src/key_log_file.rs
```diff
@@ -12,6 +12,106 @@ use rustls::KeyLog;
 #[cfg(feature = "tracing")]
 use tracing::warn;
 
+/// Name of the environment variable consulted by [`KeyLogFile`].
+const SSLKEYLOGFILE_VAR: &str = "SSLKEYLOGFILE";
```
cpu: Recommend taking a peek at https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#style-guide

For example this const should be ordered at the bottom.

djc:
Please account for our AI policy and adjust your PR(s) accordingly:

https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#ai-policy

cpu:
I think the cost of this new code (both w.r.t platform specific complexity & the new `unsafe` usage) isn't offset by the value.

As a general practice I think we want to discourage configuring a `key_log` on your client/server configs at all outside of debugging settings in which extreme hardening against the environment doesn't seem merited. Not compiling any of this support into your production artifacts at all is the best defense IMO.

For the folks that do have a use-case that requires this level of diligence they can always provide their own `dyn KeyLog` impl externally.

How do you feel about reworking this PR to instead update the docs on our simple util impl to reference this risk and saying that's sufficient? I assume you're mostly tackling this because it was an open issue and you're keen to let an LLM help and not because of a specific operational concern you had as a user?

SebTardif:
@cpu That makes sense. I reworked the PR to docs only:

- `KeyLogFile` stays the small debug helper (no platform probes / no new `unsafe`)
- Document the setuid/inherited-`SSLKEYLOGFILE` risk on the type and in the howto manual
- Point people at clearing env, not shipping key-log in production, or a custom `KeyLog`

I left the small `admin/coverage` change that runs `rustls-util` unit tests (it is not a default-member, so those tests were invisible to codecov). Happy to split that out if you would rather keep this PR docs-only.

On the open issue: happy for #911 to be closed as docs / wontfix once this lands, if that matches how you want to track it.

SebTardif:
@djc Thanks. I have re-read the AI policy and rewritten the PR around the maintainer direction above (docs-only, no secure-env implementation). I will keep review replies in my own words.

rustls-util/src/key_log_file.rs
SebTardif: Addressed by dropping the implementation (and that const) in favor of docs-only per the thread on the PR.

rustls/src/manual/howto.rs
```diff
@@ -116,7 +116,9 @@ With the file from `SSLKEYLOGFILE` it is possible to use [Wireshark] or another
 decrypt the post-handshake messages, following [these instructions][curl-sslkeylogfile].
 
 Remember this allows plaintext decryption and should only be done in testing contexts
-where no sensitive data (API keys, etc) are being shared.
+where no sensitive data (API keys, etc) are being shared. Do not enable key logging
+in setuid/setgid or otherwise elevated processes: `SSLKEYLOGFILE` is read from the
+environment and may have been inherited from an untrusted parent.
```
ctz: This is advice about how to debug TLS, platform-specific advice on production security matters doesn't really belong here.

rustls-util/src/key_log_file.rs
```diff
@@ -167,27 +191,25 @@ mod tests {
...
         assert!(inner.file.is_some(), "key log file should open");
 
-        let mode = std::fs::metadata(&path)
+        let mode = fs::metadata(&path)
             .expect("metadata")
             .permissions()
             .mode()
             & 0o777;
-        let _ = std::fs::remove_file(&path);
+        let _ = fs::remove_file(&path);
```
ctz, line 210: I think all these changes are just fiddling with imports, so don't belong in this commit.

djc:
Should probably rebase and squash first?

rustls/src/manual/howto.rs
SebTardif: @ctz

> This is advice about how to debug TLS, platform-specific advice on production security matters doesn't really belong here.

Agreed. I dropped the howto change. The setuid note stays only on `KeyLogFile`.

SebTardif:
@djc

> Should probably rebase and squash first?

Rebased onto current `main` and dropped the abandoned secure-env series. The branch is now two commits: rustdoc on `KeyLogFile`, and the `admin/coverage` line that runs `rustls-util` tests.

rustls-util/src/key_log_file.rs
SebTardif: @ctz

> I think all these changes are just fiddling with imports, so don't belong in this commit.

Dropped. After the rebase, this commit only adds the rustdoc on `KeyLogFile`.

rustls-util/src/key_log_file.rs
```diff
@@ -87,6 +87,25 @@ impl Debug for KeyLogFileInner {
...
+/// process, that variable may have been inherited from an untrusted parent
+/// and could point at an attacker-chosen path. Applications that raise
+/// privileges should clear sensitive environment variables, avoid compiling
+/// key-log support into production builds, or supply their own [`KeyLog`]
+/// implementation.
+///
+/// This util is intentionally small. It does not attempt a full
+/// `secure_getenv`-style environment filter.
+///
+/// Related discussion: <https://github.com/rustls/rustls/issues/911>
```
cpu: nit: I'd prefer not to cite the issue. I don't think the discussion raises anything that isn't already covered by the text above and someone that _really_ wants the extra context can find it via `git blame` the same way they would backtrack any other change to a PR/issue.

admin/coverage
```diff
@@ -14,6 +14,8 @@ cargo test --locked --all-features # this presumably enables `fips`
 cargo test -p rustls-ring --locked
 cargo test -p rustls-aws-lc-rs --locked
 cargo test -p rustls-test --locked --features aws-lc-rs # aws-lc-rs, but no fips
+# Not a default-member; unit tests (e.g. KeyLogFile) otherwise never hit coverage.
```
cpu: IMO I think this is another place where the LLM comment style isn't great. There's really nothing of note about the `KeyLogFile` tests in particular that rises to the level of needing a call out here.

How about:
> rustls-util isn't a workspace default-member, test it explicitly.

rustls-util/src/key_log_file.rs
SebTardif: @cpu

> I'd prefer not to cite the issue.

Dropped the link. The rustdoc now stops at the risk and the `secure_getenv` note.

admin/coverage
SebTardif: @cpu

> rustls-util isn't a workspace default-member, test it explicitly.

Used that wording.

cpu (approved):

ctz (approved):

ctz:
Thanks!

PR #3217, opened 2026-08-19
examples: collect plaintext in one process_new_packets call

Dongdia:
Collect plaintext with a single process_new_packets().handle_all() call instead of calling process_new_packets() twice, which discarded plaintext into a throwaway Vec and left nothing for the second call.


ctz (approved):
Thanks!

PR #3218, opened 2026-08-19
Introduce `ClientHandshake`

ctz:
This works further towards #850 following the route established in #3178.


rustls/src/conn/mod.rs
```diff
@@ -134,6 +134,71 @@ pub trait Connection: fmt::Debug + Deref<Target = ConnectionOutputs> {
...
+    /// On success, this returns a handshake object specifying what to do to progress
+    /// the connection.  If this contains another [`NeedsInput`] object then obtaining more
+    /// input (eg, from a socket or other source) is certainly necessary.
+    pub fn process(
+        mut self,
+        input: &mut dyn TlsInputBuffer,
+        tls: &mut Vec<u8>,
+    ) -> Result<Side::Handshake, Error> {
+        let mut iter = MessageIter::new(input, tls, None, &mut self.inner, false);
+        let r = loop {
```
djc, line 163: Nit: rename `r` to `result`? (Pre-existing, I guess.)

rustls/src/conn/mod.rs
```diff
@@ -199,6 +201,98 @@ impl<S: SideData> fmt::Debug for NeedsInput<S> {
...
+
+impl<Side: SideData> fmt::Debug for VerifyPeerIdentity<Side> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("VerifyPeerIdentity")
+            .finish_non_exhaustive()
+    }
+}
+
+/// Trait to maintain static unreachablity of per-protocol-version code.
+pub(crate) trait VerifyPeerIdentityInternal<Side: SideData>: Send + Sync {
```
djc, line 287: Nit: the `Internal` suffix feels rather ugly. Maybe `VerifySide` or something? (Also seems pre-existing.)

It feels like there are more layers than necessary here? Why do we need to abstract over the protocol version here?

rustls/src/quic.rs
```diff
@@ -632,8 +633,19 @@ impl VerifyClientIdentity {
 
     fn next(
         mut inner: QuicCommon<ServerSide>,
-        result: Result<ServerState, Error>,
+        advance: impl FnOnce(&mut dyn Output<'_>) -> Result<ServerState, Error>,
```
djc, line 636: Ugh, is there no way to work around callback-based API?

rustls/src/conn/receive.rs
```diff
@@ -202,6 +202,16 @@ impl<'a, 'm, Side: SideData, Send: SendOutput + 'a> MessageIter<'a, 'm, Side, Se
...
 
+pub(crate) enum MessageIterMode {
+    /// Iterate through all available messages.
+    Normal,
+
+    /// Handshake mode.
+    ///
+    /// Stop iteration when when external input is needed to progress handshake.
+    Handshake,
+}
```
djc, line 213: Naming nits: maybe more specific names like `Iterate { All, Handshake }` or `MessageKind { Any, Handshake }`?

djc (commented):
It feels like the state-based API is pretty unwieldy, perhaps more in the implementation than in the public API. I wonder if we could improve on that...

djc (approved):

rustls/src/conn/mod.rs
ctz: If the version-specific code is statically reachable from the `pub(crate) enum ClientState` type then all version-specific code will be included.

Have renamed a bit.

rustls/src/quic.rs
ctz: Did this by keeping the "continue" and "use the config" parts separate

Issue #2498, opened 2025-06-16
ClientHello fingerprinting (JA3/JA4)


amitnos123:
I'm trying to implement in my project JA4 fingerprinting.

From going around the repository issues,
I found you referenced [Acceptor struct](https://docs.rs/rustls/latest/rustls/server/struct.Acceptor.html) as a way to access ` client_hello` data.

I need the ` client_hello` to implement JA4.
I wanted confirmation, I did understand correctly.

Thank you for your time.
This is the first time I'm working with low level around SSL.

djc:
rustls doesn't support the fingerprinting use case and doesn't intend to in the near future.

Consider using https://github.com/3andne/craftls (or something like it) instead.

ctz:
> I need the ` client_hello` to implement JA4.

This won't be sufficient, and I don't think we plan to extend `client_hello` to expose sufficient data to make it possible. I would suggest calculating JA4 is a specialised task that warrants its own parser. See -- previously --https://github.com/rustls/rustls/issues/1815#issuecomment-1969237810

amitnos123:
@ctz 
Thank you for the reference.
I did some research on my part.
I do agree rustls doesn't expose enough data via ClientHello to generate JA4.
To generate JA4, I would need:

- TLS version	
- Cipher Suites
- ALPN
- SNI
- Supported Groups (Curves)	
- EC Point Formats

Which not all are exposed.

The [#1815]( https://github.com/rustls/rustls/issues/1815#issuecomment-1969237810) does help me find a solution for my needs.

Thank you for your help!

Issue #1815, opened 2024-02-28
`ServerHelloPayload` re-export inquery/request


5Dev24:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
Prior to the 0.22.x release(s) (0.21.10 and before), I used `rustls::internal::msgs::handshake::ServerHelloPayload` to extract information in relation to the connection made by rustls at a low level. Accessing this data was simple by taking it from a `rustls::msgs:message::Message` containing the handshake. This struct is not longer re-exported and all the fields were made pub(crate) too.

I do understand that the internal classes are not part of the stable interface from rustls. However, it would be appreciated if this could be made public again instead of having to maintain a fork with such a minor change.

**Describe the solution you'd like**
Make the fields of the `ServerHelloPayload` pub again and re-export it once more in the lib.rs.

**Describe alternatives you've considered**
I've considered and made a fork the re-exports this struct. It's a simple patch fork that is just applied after each release. Major drawback being that `cargo outdated` won't catch using an outdated version and yanked, bad versions won't automatically affect the fork.

**Additional comments**
If there's an underlying reason that it wasn't re-exported anymore, I couldn't find the discussion made on this repository. If you know there was one or know how to better look for it, please let me know.

If there's an alternative way to access this information through the stable interface, I'd love to hear it as it was difficult to find the previously mentioned struct as-is.

I don't mind making the small PR to change this, but I'd like to hear what others think before doing so.

Thank you for your time and consideration.

djc:
What were you trying to do with the ServerHelloPayload data? What goal are you trying to achieve?

djc:
I think we'd probably merge a PR that makes the fields public again, want to submit one?

Out of curiosity, what are you using the JA3 fingerprints for?

ctz:
I think you are better off writing your own parser that outputs the JA3 pre-hash format, or relying on a crate that has TLS message parsing as its goal (`tls_parser`). I would like to land #1475 in our next release (not the one that is imminent tomorrow) and this will break any code that computes JA3 (because it does not maintain extension order in its internal representation).

5Dev24:
@djc I can do that. They are used to fingerprint tls connections made to servers, that is all. Just gives a better view of what infrastructure you may be connected to if the hash is a known one.

[...]

PR #3219, opened 2026-08-20
Expose original ClientHello bytes from Accepted

iadev09:
This PR exposes the original ClientHello bytes already retained for the handshake transcript through a borrowed, zero-cost getter, with no new public types.

The typed ClientHello view drops unknown extensions and extension order at parse time.  The encoded message is already retained for the transcript, so offer a borrowed view of the exact received bytes from both the TCP and QUIC acceptor states.

This makes the suggested “write your own parser” approach generally usable (#1815, #2498, quinn-rs/quinn#2644).

### DIFF
--- a/rustls-test/tests/api/quic.rs
+++ b/rustls-test/tests/api/quic.rs
@@ -242,6 +242,7 @@ fn test_quic_acceptor() {
         let mut client_initial = flatten_events(&mut client);
         assert!(client_initial.len() > 8);
         println!("client_initial: {client_initial:x?}");
+        let expected_hello = client_initial.clone();
 
         let ServerHandshake::NeedsInput(needs_input) = needs_input
             .process(&mut SliceInput::new(&mut client_initial[..8]), &mut vec![])
@@ -256,6 +257,8 @@ fn test_quic_acceptor() {
         else {
             panic!("unexpected state after full hello");
         };
+
+        assert_eq!(accepted.client_hello_bytes(), expected_hello);
 
         let client_hello = accepted.client_hello();
         assert_eq!(
--- a/rustls/src/server/connection.rs
+++ b/rustls/src/server/connection.rs
@@ -358,6 +358,18 @@ impl Accepted {
         ch
     }
 
+    /// Get the raw, wire-encoded `ClientHello`, including information not exposed by
+    /// [`Accepted::client_hello()`], such as extension order and unknown extensions.
+    ///
+    /// # Warning
+    ///
+    /// External parsers may interpret the `ClientHello` differently from rustls.
+    /// Do not use externally parsed values for security-sensitive decisions that
+    /// must agree with rustls's interpretation.
+    pub fn client_hello_bytes(&self) -> &[u8] {
+        self.choose_config.client_hello_bytes()
+    }
+
     /// Choose a [`ServerConfig`] to progress the handshake.
     ///
     /// Output to send to the peer is appended to `output`.  Typically, this is the `ServerHello`,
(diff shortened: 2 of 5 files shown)


iadev09:
I chose not to add an artificial panic test or restructure the code solely to cover the `unreachable!()` branch.

rustls-test/tests/api/quic.rs
```diff
@@ -257,6 +258,8 @@ fn test_quic_acceptor() {
             panic!("unexpected state after full hello");
         };
 
+        assert_eq!(accepted.client_hello_bytes(), expected_hello);
```
cpu, line 261: I think it'd be nice to have a similar assertion added to `large_client_hello_acceptor()` so we get cheap coverage of  fragmented input.

rustls/src/server/connection.rs
```diff
@@ -358,6 +358,20 @@ impl Accepted {
...
+    /// Get the wire-encoded bytes of the `ClientHello` for this connection.
+    ///
+    /// This is the complete TLS handshake message as received from the client,
+    /// starting with the four-byte handshake message header, and is identical
+    /// to the bytes hashed into the handshake transcript.  If the client offered
+    /// Encrypted Client Hello, this is the outer `ClientHello` as received.
+    ///
+    /// This can be fed to an external parser to observe parts of the message
+    /// that are not covered by [`Accepted::client_hello()`], such as the
+    /// order of extensions or extensions unknown to this library.
```
cpu: I think here (and in the matching doc comment) we should maybe have some warning about avoiding assumptions that however the caller parses the raw hello being 1:1 with how _we_ parse it internally. I wouldn't want to see callers using this API for load bearing security decisions since it could introduce the opportunity for the two parsers to disagree...

cpu (commented):
Here's some initial feedback. I'm not 100% sure this is something we want to offer, and I'm likely to be swayed pro/con based on how other reviewers feel so i'd prefer to hold off on approving until then.

rustls-test/tests/api/quic.rs
iadev09: Added to large_client_hello_acceptor()

rustls/src/server/connection.rs
iadev09: Added the warning to both doc comments. I also shortened the docs so it stands out more.

iadev09:
The rustls `Acceptor` boundary was previously discussed in quinn-rs/quinn#2644.  My motivation here is to expose the retained bytes as a lossless view, so downstreams are not pushed toward riskier workarounds outside rustls.

cpu:
> The rustls Acceptor boundary was previously discussed in https://github.com/quinn-rs/quinn/pull/2644.

But it seems to me djc's position at the end of that discussion isn't supportive of this API. Am I reading this wrong:

> The use cases listed in the PR description also seem to be serviceable from a rustls ClientHello, such as the one exposed via the Acceptor

ctz:
At the current time, the externalised buffers arrangement we have for 0.24 makes this significantly less interesting I think? As the caller, you have the buffer that contains the ClientHello, and rustls has edited it for you to arrange that it is reassembled and contiguous.

iadev09:
That's a fair reading within the scope of that discussion, and I included the link as context, not as evidence of support. Since the API would live in rustls, I opened this PR to raise the question here.

iadev09:
> At the current time, the externalised buffers arrangement we have for 0.24 makes this significantly less interesting I think? As the caller, you have the buffer that contains the ClientHello, and rustls has edited it for you to arrange that it is reassembled and contiguous.

Thanks, that is very helpfull. I hadn't accounted for the caller-owned input buffer in 0.24.  I'll close the PR, gather up my paperwork, and take this back to Quinn.

PR #3220, opened 2026-08-21
Restore access to QUIC server state before handshake completion

iadev09:
After #3152, continuing a QUIC server handshake with a "selected"  `ServerConfig` leaves it in `NeedsInput`  which did not expose connection outputs, SNI, peer transport parameters, or 0-RTT keys before completion. 

QUIC needs the transport parameters and for resumed sessions, the 0-RTT keys before that point. 

This exposes the existing state through `NeedsInput`, with API coverage for full and resumed handshakes.


ctz (approved):
Thanks!

PR #3221, opened 2026-08-21
Eliminate `ErrorWithAlert` and re-split receive path

ctz:
See https://github.com/rustls/rustls/pull/3193#discussion_r3831371688 for background.

### DIFF
--- a/rustls/src/conn/split.rs
+++ b/rustls/src/conn/split.rs
@@ -1,7 +1,7 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::fmt;
-use core::ops::{DerefMut, Range};
+use core::ops::Range;
 use std::sync::MutexGuard;
 
 use super::receive::{Discard, JoinOutput};
@@ -13,7 +13,7 @@ use crate::conn::{
 };
 use crate::crypto::cipher::{OutboundPlain, RecordEncrypter};
 use crate::enums::ProtocolVersion;
-use crate::error::{AlertDescription, ErrorWithAlert};
+use crate::error::AlertDescription;
 use crate::lock::Mutex;
 use crate::msgs::{AlertLevel, Delocator, Message};
 use crate::sync::Arc;
@@ -184,14 +184,12 @@ impl<Side: SideData> ReceiveTraffic<Side> {
     ///
     /// An error from this function permanently breaks the ability to receive
     /// data from the peer. The error may be accompanied by a TLS alert,
-    /// which can be obtained from the returned [`ErrorWithAlert`] and sent
-    /// to the peer. Following this, the underlying IO medium should be
-    /// closed by the application.
-    pub fn read<'a, 't>(
+    /// which is appended to `tls`.
+    pub fn read<'a>(
         self,
         input: &'a mut impl TlsInputBuffer,
-        tls: &'t mut Vec<u8>,
-    ) -> Result<ReceiveTrafficState<'a, Side>, ErrorWithAlert<'t>> {
+        tls: &mut Vec<u8>,
+    ) -> Result<ReceiveTrafficState<'a, Side>, Error> {
         let Self {
             state,
             mut recv,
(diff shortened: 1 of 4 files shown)


rustls/src/conn/split.rs
```diff
@@ -208,19 +236,17 @@ impl<Side: SideData> ReceiveTraffic<Side> {
...
-            MessageIter::<Side, _>::receive(input, tls, &mut state, &mut recv, output, true);
+        let mut iter = MessageIter::<Side, _>::receive(
+            input,
+            &mut tls_unused,
+            &mut state,
+            &mut recv,
+            output,
+            true,
+        );
         let received_plain = match iter.next() {
```
cpu, line 247: WDYT about adding a `debug_assert!` in between L246 and L247 that asserts `tls_unused.is_empty()`?

It seems like it might be nice to catch a future direct write into `tls` that isn't handled through the `SendAdapter`.

cpu (approved):
This looks good to me.

I noticed a small gap in the test coverage. How about shoring it up a little bit with something like: https://github.com/cpu/rustls/commit/809603dab1cba06747d86cefd506001a6aa7a474 ?

djc:
I wonder if a better fix here would be to keep `ErrorWithAlert`, giving it its own buffer?

Have also been wondering if we should use something like this in more places, notably on receive-side methods of the connection.

ctz:
> I wonder if a better fix here would be to keep `ErrorWithAlert`, giving it its own buffer?

The question is -- if the receiving thread gets this `ErrorWithAlert`, how does the sending thread obtain this buffer, and then order it with respect to other data it has to send on the same connection? Seems like a basically impossible task unless all the buffers we ever emit have some sequence number range attached to them?

djc:
> The question is -- if the receiving thread gets this `ErrorWithAlert`, how does the sending thread obtain this buffer, and then order it with respect to other data it has to send on the same connection? Seems like a basically impossible task unless all the buffers we ever emit have some sequence number range attached to them?

Okay, so you're saying, either we already have a `tls` and we can append alert to it directly, or we don't have `tls` because we're receive only and then ordering is too much of a PITA?

I still wonder if it is too easy for callers to forget to send out `tls` in error cases (just `?` it), and putting a little speed bump in front of them (for handshaking or full-duplex connections) would be nice.

ctz:
> > The question is -- if the receiving thread gets this `ErrorWithAlert`, how does the sending thread obtain this buffer, and then order it with respect to other data it has to send on the same connection? Seems like a basically impossible task unless all the buffers we ever emit have some sequence number range attached to them?
> 
> Okay, so you're saying, either we already have a `tls` and we can append alert to it directly, or we don't have `tls` because we're receive only and then ordering is too much of a PITA?
> 
> I still wonder if it is too easy for callers to forget to send out `tls` in error cases (just `?` it), and putting a little speed bump in front of them (for handshaking or full-duplex connections) would be nice.

It's true and also a problem that existed in prior rustls callers, see for example https://github.com/rustls/rustls/blob/rel-0.23/examples/src/bin/tlsserver-mio.rs#L235-L242 -- it's pretty clumsy and very easy to overlook. It's also a great candidate for not doing "proper" IO, as in this case, where if a write temporarily fails (eg, the send window is full) then some or all of the alert gets dropped on the floor. I'm hoping that having the output method unified like it is now will be a bit of an improvement.

djc:
Okay, but the question is, should we do something about it (here)?

rustls/src/conn/split.rs
ctz: I added this, but one statement below. That is because `receive()` is actually a constructor and `next()` does all the work.

ctz:
So I think we're in agreement that this PR is ok, but are talking about something like introducing something similar to `ErrorWithAlert` which steals contents of the output buffer on error in order that the error type (from `MessageHandler::next_payload` and associated) requires action to retrieve an `impl core::error::Error`?

djc:
> So I think we're in agreement that this PR is ok, but are talking about something like introducing something similar to `ErrorWithAlert` which steals contents of the output buffer on error in order that the error type (from `MessageHandler::next_payload` and associated) requires action to retrieve an `impl core::error::Error`?

Yup!

ctz:
I think the length of documentation needed to explain the relationship between the two buffers (the &mut one passed in, and the one alongside the error) probably makes me think it's not a very obvious design? Specifically, stuff like:

- the buffer alongside any error is your input one, so any external tracking of (eg) a written prefix is now against this new buffer
- if you throw this buffer away, you are not necessarily just throwing away the alert
- we moved a default into your buffer, so now any indices into it are illegal but only on error (same point as 1, really)

(Could also do it with a borrow, though we are fully on the way to line-noise rust and `MessageIter` would need yet another lifetime with an ordering against one of the existing ones?)

Though I agree that forcing people to consider it in their error path is a good goal.

PR #3223, opened 2026-08-24
Add convenience methods for ApplicationProtocol and ProtocolVersion inspection

Aditya-9-6:
Hi team! 👋

When building proxies and services on top of `rustls`, inspecting the negotiated ALPN protocol (e.g. checking if it's `h2` or `h3`) and TLS version is a very common task. Right now, callers usually have to write verbose `matches!()` blocks or manually unpack the underlying bytes.

This PR adds a few lightweight, zero-cost helper methods to make working with `ApplicationProtocol`, `ProtocolVersion`, and `ConnectionOutputs` more ergonomic:

### What's Added:
- **`ApplicationProtocol`:**
  - `as_bytes()` & `as_str()` — quick ways to view the protocol name as raw bytes or a UTF-8 slice without allocations.
  - `is_http2()`, `is_http3()`, and `is_http()` — quick predicates for common HTTP protocol checks.
- **`ProtocolVersion` & `ConnectionOutputs`:**
  - `is_tls13()` and `is_tls12()` — clean predicates to check the negotiated version directly from connection outputs.
  - `is_dtls()` — helper on `ProtocolVersion`.

### Quality & Tests:
- All methods are marked `#[inline]` for zero runtime overhead.
- Added comprehensive unit tests in `enums.rs` and `common_state.rs` for both positive and negative branches.
- Code is formatted cleanly with `cargo fmt`.

Feedback is welcome—happy to adjust if you'd like any naming or signature tweaks!

### DIFF
--- a/rustls/src/common_state.rs
+++ b/rustls/src/common_state.rs
@@ -150,6 +150,20 @@ impl ConnectionOutputs {
         self.negotiated_version
     }
 
+    /// Returns `true` if TLS 1.3 was negotiated for this connection.
+    #[inline]
+    pub fn is_tls13(&self) -> bool {
+        self.protocol_version()
+            .is_some_and(|v| v.is_tls13())
+    }
+
+    /// Returns `true` if TLS 1.2 was negotiated for this connection.
+    #[inline]
+    pub fn is_tls12(&self) -> bool {
+        self.protocol_version()
+            .is_some_and(|v| v.is_tls12())
+    }
+
     /// Whether the Extended Main Secret extension was negotiated.
     ///
     /// Returns:
@@ -442,3 +456,28 @@ impl<'a, const TLS13: bool> HandshakeFlight<'a, TLS13> {
 
 pub(crate) type HandshakeFlightTls12<'a> = HandshakeFlight<'a, false>;
 pub(crate) type HandshakeFlightTls13<'a> = HandshakeFlight<'a, true>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_connection_outputs_version_helpers() {
+        let mut outputs = ConnectionOutputs::default();
+        assert_eq!(outputs.protocol_version(), None);
+        assert!(!outputs.is_tls13());
+        assert!(!outputs.is_tls12());
+
+        outputs.negotiated_version = Some(ProtocolVersion::TLSv1_3);
+        assert!(outputs.is_tls13());
+        assert!(!outputs.is_tls12());
+
+        outputs.negotiated_version = Some(ProtocolVersion::TLSv1_2);
+        assert!(!outputs.is_tls13());
+        assert!(outputs.is_tls12());
+
+        outputs.negotiated_version = Some(ProtocolVersion::SSLv3);
+        assert!(!outputs.is_tls13());
+        assert!(!outputs.is_tls12());
+    }
+}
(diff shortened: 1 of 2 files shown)


rustls/src/common_state.rs
```diff
@@ -150,6 +150,20 @@ impl ConnectionOutputs {
         self.negotiated_version
     }
 
+    /// Returns `true` if TLS 1.3 was negotiated for this connection.
+    #[inline]
+    pub fn is_tls13(&self) -> bool {
```
ctz, line 155: I think the callers of this would be clearer writing `conn.protocol_version == Some(ProtocolVersion::TLSv1_3)`

The same goes for the TLS1.2 one.

rustls/src/enums.rs
```diff
@@ -67,6 +67,39 @@ impl<'a> ApplicationProtocol<'a> {
             Self::Other(data) => ApplicationProtocol::Other(Cow::Owned(data.to_vec())),
         }
     }
+
+    /// Returns the raw byte slice representation of this application protocol.
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] {
```
ctz, line 73: I don't think it's a great idea to give a second name to the AsRef.

rustls/src/enums.rs
```diff
@@ -67,6 +67,39 @@ impl<'a> ApplicationProtocol<'a> {
...
+    }
+
+    /// Returns `true` if this protocol is any version of HTTP (`http/0.9`, `http/1.0`, `http/1.1`, `h2`, `h3`).
+    #[inline]
+    pub fn is_http(&self) -> bool {
+        matches!(
+            self,
+            Self::Http09 | Self::Http10 | Self::Http11 | Self::Http2 | Self::Http3
+        )
+    }
```
ctz, line 102: All these are better and clearer inlined into the caller.

rustls/src/enums.rs
```diff
@@ -190,6 +223,26 @@ enum_builder! {
...
+    pub const fn is_tls13(self) -> bool {
+        self.0 == Self::TLSv1_3.0
+    }
+
+    /// Returns `true` if this protocol version is TLS 1.2.
+    #[inline]
+    pub const fn is_tls12(self) -> bool {
+        self.0 == Self::TLSv1_2.0
+    }
+
```
ctz, line 238: These would be better inlined into the caller.

rustls/src/enums.rs
```diff
@@ -190,6 +223,26 @@ enum_builder! {
...
+    /// Returns `true` if this protocol version is any DTLS version.
+    #[inline]
+    pub const fn is_dtls(self) -> bool {
+        matches!(self.0, 0xFEFF | 0xFEFD | 0xFEFC)
```
ctz, line 242: Doesn't seem useful as rustls cannot negotiate DTLS.

ctz (commented):
Apologies but I don't think this is very useful overall.

Aditya-9-6:
Thank you for taking the time to review this PR and for the feedback! 
That makes perfect sense. I agree that convenience wrappers like `is_http` or `is_tls13` add unnecessary API surface when callers can write clear inline comparisons against `ProtocolVersion::TLSv1_3` directly. Also, good catch on DTLS—it indeed makes no sense to include it here since rustls does not negotiate it.

Since the existing traits (`AsRef`) and direct enum matches cover these use cases cleanly, I will close this PR. 
Thanks again for the guidance!

PR #3224, opened 2026-08-24
Generalise internals of handshake APIs over their transport

ctz:
This de-duplicates the internals of TCP/QUIC handshake types. Currently it's on top of #3218 but a different order would work.


rustls/src/conn/handshake.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Transport state that persists across handshake steps.
+    type State;
+
+    /// The sink for handshake messages and key changes, if this transport has one.
+    ///
+    /// `None` means the handshake is encoded as TLS records into the caller's buffer.
+    fn quic(state: &mut Self::State) -> Option<&mut dyn QuicOutput>;
+}
+
+pub(crate) struct TcpTransport;
```
djc: Could we just call this `Tcp`? (And similar for `QuicTransport`.)

djc (commented):
Do you think this could be split into smaller commits? The current version is pretty hard to review IMO.

ctz:
> Do you think this could be split into smaller commits? The current version is pretty hard to review IMO.

I have tried this.

rustls/src/quic.rs
```diff
@@ -670,6 +680,17 @@ impl fmt::Debug for VerifyClientIdentity {
     }
 }
 
+/// The QUIC transport: TLS handshake messages and key changes delivered as [`QuicEvent`]s.
+pub(crate) struct Quic;
+
+impl Transport for Quic {
+    type State = QuicState;
```
djc, line 639: Why have this indirection between `Quic` and `QuicState`?

rustls/src/conn/handshake.rs
```diff
@@ -0,0 +1,99 @@
...
+//! here, parameterised by [`Transport`].
+
+use alloc::vec::Vec;
+
+use super::{ConnectionCommon, MessageIter, MessageIterMode, ReceivePath, SideData};
+use crate::TlsInputBuffer;
+use crate::error::Error;
+use crate::quic::QuicOutput;
+
+pub(crate) struct NeedsInputCore<Side: SideData, T: Transport>(pub(crate) Core<Side, T>);
```
djc, line 102: What is the value of this wrapper?

rustls/src/quic.rs
```diff
@@ -442,17 +439,22 @@ impl NeedsInput {
     /// they cannot be fully trusted until then. Reliance on them should be minimized.
     /// Any tampering with the parameters will cause the handshake to fail.
     pub fn quic_transport_parameters(&self) -> Option<&[u8]> {
-        self.inner.quic_transport_parameters()
+        self.0
```
djc, line 436: The `.0.0` stuff makes me think this is not the best design.

djc (commented):
Here's some feedback on the first more substantive commit of the branch.

I think this is generally the right direction and I think it might be neat to end up with an API where we expose `Connection<ClientSide, Tcp>` to downstreams. I was thinking we might also attach things like whether a fake CCS needs to be emitted to the `Transport` trait (or private supertrait of it).

Actually, I was wondering if some sort of high-level state becomes another parameter to `Connection` such that we can distinguish `Connection<ClientSide, Tcp, Traffic>` from `Connection<ClientSide, Udp, Handshaking>` (or maybe that becomes more like `Connection<S, T>` and `Handshaking<S, T>`), and a way to distinguish between the phased handshake approach and the monolithic handshake API.

rustls/src/conn/handshake.rs
ctz: Good point; not much. Sliced it off leaving the `Core` part there.

rustls/src/quic.rs
ctz: Simplified this.

ctz:
This is ready for a second look now.

rustls/src/quic.rs
djc: But `Transport` still has an associated `State` type -- why not use the `Transport`-implementing type itself directly?

ctz: I was kind of expecting more stuff to hang off this eventually? But have done that for now.

rustls/src/conn/handshake.rs
```diff
@@ -0,0 +1,79 @@
...
+        );
+
+        result?;
+        Ok(Self { inner, transport })
+    }
+}
+
+/// The aspects of a handshake that depend on the underlying transport.
+pub(crate) trait Transport: Sized {
+    /// The sink for handshake messages and key changes, if this transport has one.
```
cpu, line 67: ultra-nit: it feels like this docstring should maybe explicitly say "The sink for _quic_ handshake messages and key changes ..." ?

rustls/src/quic.rs
cpu, line 617: I think all the callers of this are now providing `MessageIterMode::All` - could we drop the param here?

rustls/src/quic.rs
```diff
@@ -381,17 +376,23 @@ impl NeedsInput {
...
+            .transport
+            .params
+            .as_ref()
+            .map(|v| v.as_ref())
```
cpu, line 383: Could this & `QuicCommon::quic_transport_parameters()` both share a single `Quic::params()` helper instead of digging through the `transport.params`?

cpu (approved):
I think this looks good, but I probably have less meaningful input on this compared to djc :-)

rustls/src/quic.rs
```diff
@@ -626,40 +636,7 @@ impl<Side: SideData> QuicCommon<Side> {
...
-            inner: suite,
-            quic: suite.quic?,
-        };
-
-        Some(DirectionalKeys::new(
-            suite,
-            self.quic.early_secret.as_ref()?,
-            self.quic.version,
-        ))
-    }
```
djc, line 650: Nit: would be nice to keep this in a separate commit.

djc (commented):
What is the value of the transport-specific wrappers like `Accepted(AcceptedCore<Tcp>)`? Why not just expose `AcceptedCore` as `Accepted`, keeping the transport parameter public?

ctz:
> What is the value of the transport-specific wrappers like `Accepted(AcceptedCore<Tcp>)`? Why not just expose `AcceptedCore` as `Accepted`, keeping the transport parameter public?

General aversion to lots of generic parameters, and one type which has lots if impls split across instantiations, really. Will experiment a bit more though.

djc:
> > What is the value of the transport-specific wrappers like `Accepted(AcceptedCore<Tcp>)`? Why not just expose `AcceptedCore` as `Accepted`, keeping the transport parameter public?
> 
> General aversion to lots of generic parameters, and one type which has lots if impls split across instantiations, really. Will experiment a bit more though.

I feel like adding more layers doesn't really make that stuff easier to reason about, though?

ctz:
For me it's mainly about limiting the complexity of the public API, even if that comes the cost of internal complexity.

rustls/src/quic.rs
ctz: added a prior commit

djc (approved):

djc:
> For me it's mainly about limiting the complexity of the public API, even if that comes the cost of internal complexity.

I don't think I agree? But obviously you get to make the final call.

In my mind there is value in letting some of the internals shine through on the outside, because I think that makes it easier for users to reason through what's happening on the inside. Plus, the wrappers make it impossible to abstract over things outside the crate. And in this case, it feels like there is substantial boilerplate and bouncing around when trying to understand how this code works. With good documentation and well-defined abstractions, I think making the public API less opaque makes it more powerful and actually easier to understand.

ctz:
Added a sketch of what this would look like for one of the types. Naturally it would need squashing into the earlier commits. WDYT?

djc (commented):
The sketch looks nice to me -- deduplicating all that interface sure seems good. The `SideData` methods look somewhat ugly, but maybe we can find a better way for that?

ctz:
> The sketch looks nice to me -- deduplicating all that interface sure seems good.

I have done this uniformly and squashed the changes into the relevant earlier commits.

> The `SideData` methods look somewhat ugly, but maybe we can find a better way for that?

Something like https://github.com/rustls/rustls/commit/be7dc2359d5a1cc4cbcfd8b5fe21ee1ad28d5cc4 ? From the boost school of API design.

ctz:
Going to merge this. Happy to progress the last point separately.

rustls/src/conn/buffered.rs
```diff
@@ -0,0 +1,487 @@
...
+///
+/// This drives a handshake (see [`ClientHandshake`] and [`ServerHandshake`]) to completion
+/// as data is received, and then the resulting [`SplitConnection`], behind a single object.
+///
+/// Encrypt data destined for the peer using [`Self::write()`].
+/// Process received data from the peer using [`Self::read_tls()`].
+///
+/// [`ClientHandshake`]: crate::client::ClientHandshake
+/// [`ServerHandshake`]: crate::server::ServerHandshake
+pub struct Connection<Side: SideData> {
```
djc, line 25: What is the purpose of this? Is this the same reason the PR is +500 lines even though it aimed to share more code?

ctz: Oops. This is something that was in my working tree and made it in during a rebase

PR #3225, opened 2026-08-25
Clarify naming of records and messages

### LINKED ISSUE #3111: Rename `EncodedMessage` to `Record`
**Checklist**

`crypto::cipher::messages::EncodedMessage` gets used _a lot_ throughout rustls. But since TLS is a layered protocol, I think `Message` is somewhat ambiguous: there's messages within the handshake protocol, then there's records (which is what an `EncodedMessage` actually represents) and then there's packets in the transport underneath (such as TCP, QUIC or UDP).

`Encoded` is also misleading: if the message were encoded, it'd be an opaque sequence of bytes. In fact we will eventually call `EncodedMessage<OpaqueOutbound>::encode` to get just that, which further suggests the name isn't right.

An `EncodedMessage` is a parsed, structured representation of a [TLS record](https://datatracker.ietf.org/doc/html/rfc8446#section-5.1). I propose renaming it to `Record`. I think that would make it easier for readers to keep track of what they're looking at.

DanBondarenko:
Closes #3111.

Introduces consistent naming, distinguishing TLS records from messages. Uses "record" if it corresponds to `TLSPlaintext`/`TLSCiphertext` and the payload has not
been interpreted. Keeps "message" if the payload's been interpreted depending on the record's content type.

Breaks public API. No changes in behaviour.

Changes split into four commits [as requested](https://github.com/rustls/rustls/issues/3111#issuecomment-5045671276). The QUIC commit drops MessageEncrypter/Decrypter from `PacketKey` limit docs (QUIC doesn't use these).

Deliberately left alone `MessageIter` (coalesces handshake messages) and `Deframer::input_message` (tracks handshake message boundaries).

### DIFF
--- a/rustls/src/tls12.rs
+++ b/rustls/src/tls12.rs
@@ -8,7 +8,7 @@ use zeroize::Zeroizing;
 
 use crate::common_state::{Protocol, Side};
 use crate::conn::{ConnectionRandoms, Exporter};
-use crate::crypto::cipher::{AeadKey, MessageDecrypter, MessageEncrypter, Tls12AeadAlgorithm};
+use crate::crypto::cipher::{AeadKey, RecordDecrypter, RecordEncrypter, Tls12AeadAlgorithm};
 use crate::crypto::kx::{ActiveKeyExchange, KeyExchangeAlgorithm};
 use crate::crypto::tls12::PrfSecret;
 use crate::crypto::{self, SignatureScheme, hash};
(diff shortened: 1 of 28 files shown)


rustls/src/tls12.rs
```diff
@@ -398,7 +398,7 @@ fn join_randoms(first: &[u8; 32], second: &[u8; 32]) -> [u8; 64] {
     randoms
 }
 
-type MessageCipherPair = (Box<dyn MessageDecrypter>, Box<dyn MessageEncrypter>);
+type MessageCipherPair = (Box<dyn RecordDecrypter>, Box<dyn RecordEncrypter>);
```
ctz:
```suggestion
type RecordCipherPair = (Box<dyn RecordDecrypter>, Box<dyn RecordEncrypter>);
```

ctz (approved):

rustls/src/tls12.rs
DanBondarenko: Done. Rebased it such that this change is part of the MessageEncrypter/Decrypter commit.

ctz:
Thanks!

PR #3227, opened 2026-08-26
Rework accesses to `SideData::Data` types

ctz:
This PR moves towards exposing `ClientConnectionData` and `ServerConnectionData` and unifying accessors for them. As a final step, it exposes them in `SplitConnection`.


djc:
Have not reviewed in detail, but why would we want this?

ctz:
> Have not reviewed in detail, but why would we want this?

Otherwise `SplitConnection` discards the side-specific data. It means that, for example, someone using that API cannot determine if early data was accepted, or what happened with ECH, etc.

djc:
> > Have not reviewed in detail, but why would we want this?
> 
> Otherwise `SplitConnection` discards the side-specific data. It means that, for example, someone using that API cannot determine if early data was accepted, or what happened with ECH, etc.

IIRC `SplitConnection` can only be created after the handshake anyway, right? So someone can inspect those things before calling `split()`? And dropping handshake outputs could substantially be construed as a feature reducing memory usage for longer-lived connection objects?

ctz:
> > > Have not reviewed in detail, but why would we want this?
> > 
> > 
> > Otherwise `SplitConnection` discards the side-specific data. It means that, for example, someone using that API cannot determine if early data was accepted, or what happened with ECH, etc.
> 
> IIRC `SplitConnection` can only be created after the handshake anyway, right?

That is one way. The other way is `ServerHandshake` -- that route has no way to access any of this data without this PR.

> So someone can inspect those things before calling `split()`? And dropping handshake outputs could substantially be construed as a feature reducing memory usage for longer-lived connection objects?

This should be at the control of the caller.

rustls/src/client/connection.rs
```diff
@@ -476,12 +466,35 @@ impl SideOutput for ClientConnectionData {
...
+pub struct ClientConnectionData {
     early_data: Option<EarlyData>,
     ech_status: EchStatus,
 }
 
+impl ClientConnectionData {
+    /// Return the connection's Encrypted Client Hello (ECH) status.
+    pub fn ech_status(&self) -> EchStatus {
+        self.ech_status
+    }
```
djc, line 471: Nit: move the trivial getter below the more complex early data helper?

Could/should we just make these fields public with the struct becoming `non_exhaustive`?

rustls/src/server/connection.rs
```diff
@@ -421,18 +421,36 @@ impl ConnectionCommon<ServerSide> {
...
-    pub(crate) fn received_resumption_data(&self) -> Option<&[u8]> {
+    /// Retrieves the resumption data supplied by the client, if any.
+    ///
+    /// Returns `Some` if and only if a valid resumption ticket has been received from the client.
+    pub fn received_resumption_data(&self) -> Option<&[u8]> {
         self.received_resumption_data.as_deref()
     }
 
-    pub(crate) fn server_name(&self) -> Option<&DnsName<'static>> {
+    /// Retrieves the server name, if any, used to select the certificate and
```
djc, line 424: Nit: single-line first sentence?

rustls/src/client/connection.rs
```diff
@@ -110,6 +110,11 @@ impl ClientConnection {
             .recv
             .tls13_tickets_received
     }
+
+    /// Returns data learned during the connection, specific to being a client.
+    pub fn client_data(&self) -> &ClientConnectionData {
```
djc, line 101: The `client_` prefix feels superfluous to me?

rustls/src/server/connection.rs
```diff
@@ -133,6 +133,11 @@ impl ServerConnection {
             None
         }
     }
+
+    /// Returns data learned during the connection, specific to being a server.
+    pub fn server_data(&self) -> &ServerConnectionData {
```
djc, line 108: (And same for the server.)

djc (approved):
Seems like a nice modularity improvement!

Suggest we rename the types to drop the `Connection` infix (like `ClientConnectionData` -> `ClientData`).

rustls/src/client/connection.rs
cpu, line 80: I think this `is_early_data_accepted` ref should be pointing to `ClientConnectionData::is_early_data_accepted()` (maybe make it a proper rustdoc intra-link so that it would break in CI if it goes stale?)

rustls/src/server/connection.rs
```diff
@@ -420,19 +406,37 @@ impl ConnectionCommon<ServerSide> {
 }
 
 /// State associated with a server connection.
-#[derive(Default)]
-pub(crate) struct ServerConnectionData {
+#[derive(Debug, Default)]
```
cpu, line 409: I think it might be a good idea to implement `Debug` manually to omit the raw `received_resumption_data`. I think that also better matches the way `EarlyDataState` omits the `Accepted` `ChunkVecBuffer`.

rustls/src/quic.rs
cpu, line 581: I think only `Accepted::choose_config()` kept the `debug_assert()` for `tls.is_empty()`, but `VerifyClientIdentity::continue_with()` and `NeedsInput::read_hs()` dropped it. Was that intentional?

rustls/src/conn/mod.rs
```diff
@@ -143,13 +139,11 @@ pub trait Connection: fmt::Debug + Deref<Target = ConnectionOutputs> {
 /// More data needs to be supplied to make progress.
 ///
 /// Provide the data to [`Self::process()`].
-pub struct NeedsInput<Side: SideData>(pub(crate) Core<Side, Tcp>);
+pub struct NeedsInput<Side: SideData> {
+    pub(crate) inner: ConnectionCommon<Side>,
+}
 
 impl<Side: SideData> NeedsInput<Side> {
-    pub(crate) fn new(inner: ConnectionCommon<Side>) -> Self {
```
cpu, line 149: Adding a comment here just as a place to anchor it. I think there's another instance of this kind of mismatched param name in `rustls/src/common_state.rs` for `maybe_send_fatal_alert()` where it talks about a non-existent `output` param.

cpu (approved):

rustls/src/conn/mod.rs
```diff
@@ -702,6 +702,10 @@ impl<'q> Output<'_> for SideCommonOutput<'_, 'q> {
 /// Data specific to the peer's side (client or server).
 #[expect(private_bounds)]
 pub trait SideData: private::Side + Sized {
+    /// Type holding data learned during the connection, specific to this side.
+    #[expect(private_bounds)]
+    type Data: SideOutput + fmt::Debug;
```
djc, line 707: Also, should we just transpose this to get rid of `ClientSide` and `ServerSide` and use `ClientData` and `ServerData` as the marker types directly? (As in, have `ClientData` implement `SideData` instead of redirecting via `SideData::Data`.)

rustls/src/conn/mod.rs
ctz: Dealing with these en-bloc in #3287

rustls/src/quic.rs
ctz: These changes fell out of this PR.

rustls/src/client/connection.rs
ctz: > Could/should we just make these fields public with the struct becoming `non_exhaustive`?

Probably not, as it forces a specific internal representation?

djc: Fair enough.

rustls/src/conn/mod.rs
ctz: Going to leave this as a future simplification, as I'm not sure about merging two types with quite different roles.

PR #3229, opened 2026-08-26
Fix QUIC handshakes requiring HelloRetryRequest

iadev09:
QUIC does not use the TLS record layer, so it has no means to carry a `ChangeCipherSpec` record ([RFC 9001 §8.4 — Prohibit TLS Middlebox Compatibility Mode](https://www.rfc-editor.org/rfc/rfc9001.html#section-8.4)).

In downstream QUIC handshakes requiring a `HelloRetryRequest`, the fake CCS corrupts the framing of the second `ClientHello`.

This PR:

- Centralizes the QUIC check in `emit_fake_ccs()`, preventing fake CCS emission from all client QUIC paths, including the second `ClientHello` following a `HelloRetryRequest`.
- Removes the redundant caller-side checks and the now-unused `KeyScheduleEarlyClient::is_quic()`.
- Adds dedicated QUIC handshake coverage for the `HelloRetryRequest` path.

### DIFF
--- a/rustls-test/tests/api/quic.rs
+++ b/rustls-test/tests/api/quic.rs
@@ -17,7 +17,8 @@ use rustls::server::Tls13Tickets;
 use rustls::{CipherSuiteCommon, HandshakeKind, SliceInput, Tls13CipherSuite, VecInput};
 use rustls_test::{
     ClientStorage, KeyType, MultiTest, do_handshake, encoding, make_client_config,
-    make_pair_for_arc_configs, make_server_config, server_name,
+    make_client_config_with_kx_groups, make_pair_for_arc_configs, make_server_config,
+    make_server_config_with_kx_groups, server_name,
 };
 
 use super::provider;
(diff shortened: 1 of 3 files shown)


rustls/src/client/hs.rs
```diff
@@ -872,7 +872,9 @@ fn emit_client_hello_for_retry(
         payload: MessagePayload::handshake(chp),
     };
 
-    if retryreq.is_some() {
+    // RFC 9001 §8.4 prohibits TLS middlebox compatibility mode in QUIC:
+    // <https://www.rfc-editor.org/rfc/rfc9001.html#section-8.4>
+    if retryreq.is_some() && !input.protocol.is_quic() {
```
cpu: I think we should handle this closer to how 0.23 did and have `emit_fake_ccs()` perform this check itself using `Output.quic().is_some()` and then dropping the `ClientHandler.handle_server_hello()` and `derive_early_traffic_secret()` caller-side checks (in addition to this new one).

It seems preferable to me to have the skip logic central to the emitting function instead of duplicated in the callers.

rustls-test/tests/api/quic.rs
```diff
@@ -212,6 +213,57 @@ fn test_quic_handshake() {
...
+        quic::Version::V1,
+        b"server params".to_vec(),
+    )
+    .unwrap();
+
+    quic_transfer(&mut client, &mut server).unwrap();
+    quic_transfer(&mut server, &mut client).unwrap();
+
+    let client_retry_flight = client.events().collect::<Vec<_>>();
+    assert_eq!(client_retry_flight.len(), 1);
```
cpu: nit: I think this `assert_eq!` is redundant with the `assert!(matches!(...))` below.

cpu (commented):
Thanks for the fix. For other reviewers, AFAICT this is specifically a regression on `main`.

rustls/src/client/hs.rs
iadev09: Agreed, this is the right boundary. I had overlooked the existing client-side handling as well. Thanks! Since the QUIC check now lives in `emit_fake_ccs()`, the client no longer needs separate caller-side checks. This also left `KeyScheduleEarlyClient::is_quic()` unused, so I removed it.

rustls-test/tests/api/quic.rs
iadev09: Removed.

cpu (approved):
Thanks.

I think the PR description has fallen out-of-sync with the implementation if you wouldn't mind editing it to match.

iadev09:
> Thanks.
> 
> I think the PR description has fallen out-of-sync with the implementation if you wouldn't mind editing it to match.

I've updated the PR description to match the implementation. Thanks!

djc (approved):
Nice, thanks!

djc:
> Thanks for the fix. For other reviewers, AFAICT this is specifically a regression on `main`.

Did you figure out when/where this regressed?

ctz:
> > Thanks for the fix. For other reviewers, AFAICT this is specifically a regression on `main`.
> 
> Did you figure out when/where this regressed?

This was f04f958dab4a2c3bded91cc9e4fd46b3d46405a6 -- sorry!

cpu:
> Did you figure out when/where this regressed?

https://github.com/rustls/rustls/commit/f04f958d removed the guard internal to `emit_fake_ccs()` but missed adding the QUIC check on all the call-sites.

The natural follow-up question is why we moved that check out and I think the fix we landed on here was only possible after e5f08a24 moved the QUIC state into `Output` so we could do that check again inside `emit_fake_ccs()`.

PR #3230, opened 2026-08-27
Refuse secret extraction with pending send data

cpu:
Follow-up to #3221 I bumped into while reviewing.

Data awaiting the next send-side write (an alert parked in the split send half's aside buffer, or a TLS1.3 `key_update` response queued in the send path) has consumed send sequence numbers.

Previously both `dangerous_extract_secrets()` and the split `dangerous_into_kernel_connection()` silently discarded it, leaving the extracted `tx` secrets ahead of what the peer receives. Subsequently every record protected with those secrets will fail to decrypt.

This commit guards both conversion paths, returning an `ApiMisuse` error directing the application to write out pending data first.

### DIFF
--- a/rustls-test/tests/api/crypto.rs
+++ b/rustls-test/tests/api/crypto.rs
@@ -6,6 +6,7 @@ use std::borrow::Cow;
 use std::sync::{Arc, Mutex};
 
 use rustls::crypto::{Credentials, CryptoProvider};
+use rustls::error::ApiMisuse;
 use rustls::{
     ClientConfig, ClientConnection, Connection, ConnectionTrafficSecrets, Error, KeyLog,
     ServerConfig, ServerConnection, SupportedCipherSuite, VecInput,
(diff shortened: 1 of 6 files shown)


ctz (approved):
Good spot, thanks!

djc:
With external outgoing buffers, should we add more documentation about callers having to make sure the contents of those buffers have been sent before using the extracted secrets?

cpu:
>  should we add more documentation about callers having to make sure the contents of those buffers have been sent before using the extracted secrets?

I'm not sure I understand what you think is missing. This branch adds that documentation for the `SplitConnection` API (and the older `Connection` API) for extracting secrets. Where would you want to see additional documentation added?

djc:
> > should we add more documentation about callers having to make sure the contents of those buffers have been sent before using the extracted secrets?
> 
> I'm not sure I understand what you think is missing. This branch adds that documentation for the `SplitConnection` API (and the older `Connection` API) for extracting secrets. Where would you want to see additional documentation added?

I mean that the only thing we can raise errors for is queued key updates. I'm talking about outgoing TLS bytes that have been pushed into the caller's `tls` buffer before the secret extraction is activated, which, because we don't (now) own the outgoing TLS buffer, we can't check. The documentation you added here doesn't really seem to cover that?

ctz:
> > > should we add more documentation about callers having to make sure the contents of those buffers have been sent before using the extracted secrets?
> > 
> > 
> > I'm not sure I understand what you think is missing. This branch adds that documentation for the `SplitConnection` API (and the older `Connection` API) for extracting secrets. Where would you want to see additional documentation added?
> 
> I mean that the only thing we can raise errors for is queued key updates. I'm talking about outgoing TLS bytes that have been pushed into the caller's `tls` buffer before the secret extraction is activated, which, because we don't (now) own the outgoing TLS buffer, we can't check. The documentation you added here doesn't really seem to cover that?

Elsewhere we have documentation like:

```rust
    /// The TLS data to send to the peer is written into `tls`. This data should then be
    /// communicated to the peer.
```

I think that is sufficient? At the "edge" of the library, if we give the application some bytes to send to the peer and they don't do that, the connection is going to break.

djc:
> > > > should we add more documentation about callers having to make sure the contents of those buffers have been sent before using the extracted secrets?
> > > 
> > > 
> > > I'm not sure I understand what you think is missing. This branch adds that documentation for the `SplitConnection` API (and the older `Connection` API) for extracting secrets. Where would you want to see additional documentation added?
> > 
> > 
> > I mean that the only thing we can raise errors for is queued key updates. I'm talking about outgoing TLS bytes that have been pushed into the caller's `tls` buffer before the secret extraction is activated, which, because we don't (now) own the outgoing TLS buffer, we can't check. The documentation you added here doesn't really seem to cover that?
> 
> Elsewhere we have documentation like:
> 
> ```rust
>     /// The TLS data to send to the peer is written into `tls`. This data should then be
>     /// communicated to the peer.
> ```
> 
> I think that is sufficient? At the "edge" of the library, if we give the application some bytes to send to the peer and they don't do that, the connection is going to break.

It is technically sufficient. Practically, adding one line to this "special" secret extraction APIs' documentation seems like it might help someone?

cpu:
>  adding one line to this "special" secret extraction APIs' documentation seems like it might help someone?

I tweaked the docs additions. WDYT?

ctz (approved):

djc:
> > adding one line to this "special" secret extraction APIs' documentation seems like it might help someone?
> 
> I tweaked the docs additions. WDYT?

Nice, thanks!

PR #3235, opened 2026-08-31
Verify server certificate against correct name on ECH rejection

ctz:
(And incrementally improve ECH testing generally.)


bogo/src/main.rs
```diff
@@ -739,7 +742,7 @@ impl Options {
         self.support_tls12 && self.version_allowed(ProtocolVersion::TLSv1_2)
     }
 
-    fn expected_server_name(&self) -> Option<ServerName<'static>> {
+    fn expected_server_names(&self) -> Option<Vec<ServerName<'static>>> {
```
djc: Should we simplify `Option<Vec<_>>` to just `Vec<_>` or is the `None` case materially different from empty `Vec` in this case?

djc (approved):

rustls/src/client/ech.rs
```diff
@@ -507,6 +508,12 @@ impl EchState {
             }
             false => {
                 trace!("ECH rejected by server");
+
+                // "If the server rejects ECH, the client proceeds with the handshake, authenticating
+                // for ECHConfig.contents.public_name"
+                // -- <https://www.rfc-editor.org/info/rfc9849/#section-6.1.6>
+                *server_name = self.outer_name.clone().into();
```
djc: Should this be backported?

ctz: Yep I think this is a good candidate.

djc (approved):

PR #3237, opened 2026-08-31
aws-lc-rs: merge ML-DSA support from rustls-post-quantum

### LINKED ISSUE #3233: Promote ML-DSA from `rustls-post-quantum` to `rustls-aws-lc-rs`
Seems like this is stable enough to have as a default now. eg, it's in go 1.27 by default.

(This probably leaves `rustls-post-quantum` empty, wherein it can go back to being a trivial re-export of `rustls-aws-lc-rs` pending any future use we find for it.)

djc:
Fixes #3233.


rustls/src/tls12.rs
```diff
@@ -111,9 +111,13 @@ impl Suite for Tls12CipherSuite {
     /// Return true if this suite is usable for a key only offering `sig_alg`
     /// signatures.
     fn usable_for_signature_scheme(&self, scheme: SignatureScheme) -> bool {
+        let Some(alg) = scheme.algorithm() else {
+            return false;
+        };
+
         self.sign
             .iter()
-            .any(|s| s.algorithm() == scheme.algorithm())
```
djc, line 116: This felt a little unprincipled and I was wondering if it had something to do with the bogo failures for ML-DSA in 1.2. It doesn't seem to help... should we do something like this?

rustls-aws-lc-rs/src/verify.rs
```diff
@@ -71,6 +74,18 @@ pub(crate) static SUPPORTED_SIG_ALGS: WebPkiSupportedAlgorithms =
                 SignatureScheme::RSA_PKCS1_SHA256,
                 &[RSA_PKCS1_2048_8192_SHA256],
             ),
+            (
+                SignatureScheme::ML_DSA_44,
+                &[ML_DSA_44, ML_DSA_65, ML_DSA_87],
```
ctz: surely not?

djc: Nope, already fixed it locally.

djc:
I've fixed the bogo issues here, but asked some design questions on Discord.

rustls/src/client/tls12.rs
```diff
@@ -672,14 +672,21 @@ impl ExpectCertificateRequest {
         // We ignore certreq.certtypes as a result, since the information it contains
         // is entirely duplicated in certreq.sigschemes.
 
+        let signature_schemes = certreq
```
ctz, line 677: Would be good if the commit message explains what's going on here; what the goal is.

djc: Expanded the commit message and added a comment here.

djc:
Not sure why this results in a bunch of performance regressions? Maybe larger code size from dragging around more verification algorithms?

djc:
> Not sure why this results in a bunch of performance regressions? Maybe larger code size from dragging around more verification algorithms?

Claude analysis of profile difference from the worst benchmark regression:

> * More transcript hashing — the extra bytes push the hashed handshake data across a 64-byte block boundary, so each handshake costs an extra SHA-256 compression. That's why `SHA256_Final` went up ~4% while `SHA256_Update` stayed flat: the padding in `Final` now needs a second block. This is a length-alignment step, not proportional work, which is why it looks disproportionately large (+63.8k) for a few extra bytes.
> * More scheme iteration in `with_version` — the new "tls12: require known signature algorithm" filtering runs over a longer scheme list, and the compiler inlined `usable_for_signature_scheme` into it (hence that symbol vanishing). Net ~+6k.
> * More list parsing in `Vec<T>::read` (+20%) — the extra entries in the extension.

ctz:
One thing we could consider here is not including those in TLS1.2-only handshakes, as they don't have any associated ciphersuites.

djc:
> One thing we could consider here is not including those in TLS1.2-only handshakes, as they don't have any associated ciphersuites.

I guess? The audience that (1) cares about performance and (2) **only** enables 1.2 seems like it might be too small to be worth it.

djc:
> > One thing we could consider here is not including those in TLS1.2-only handshakes, as they don't have any associated ciphersuites.
> 
> I guess? The audience that (1) cares about performance and (2) **only** enables 1.2 seems like it might be too small to be worth it.

After I thought about this some more, ended up adding a commit for this, more for the correctness than for the performance.

rustls/src/client/hs.rs
```diff
@@ -612,6 +612,16 @@ fn emit_client_hello_for_retry(
...
     });
 
+    if config
+        .provider()
+        .tls13_cipher_suites
+        .is_empty()
+    {
+        if let Some(schemes) = &mut exts.signature_schemes {
+            schemes.retain(|scheme| scheme.algorithm().is_some());
+        }
```
ctz, line 622: Let's also do this here https://github.com/rustls/rustls/blob/c5115df733f2c951c7e1627e43f595770d64f5a1/rustls/src/server/tls12.rs#L497

ctz (approved):

rustls-aws-lc-rs/src/verify.rs
```diff
@@ -258,9 +250,29 @@ pub static RSA_PKCS1_2048_8192_SHA512: &dyn SignatureVerificationAlgorithm =
...
+};
+
+/// ML-DSA signatures using the [6, 5] matrix (security strength category 3).
+pub static ML_DSA_65: &dyn SignatureVerificationAlgorithm = &AwsLcRsVerificationAlgorithm {
+    public_key_alg_id: alg_id::ML_DSA_65,
+    signature_alg_id: alg_id::ML_DSA_65,
+    verification_alg: &signature::ML_DSA_65,
+};
+
+/// ML-DSA signatures using the [8. 7] matrix (security strength category 5).
```
ctz:
```suggestion
/// ML-DSA signatures using the [8, 7] matrix (security strength category 5).
```

djc: Thanks, fixed.

PR #3242, opened 2026-09-03
Apply Cargo suggestions from new Clippy lints

djc:
Probably need to fix up more stuff for the renamed example binaries...


ctz (approved):

fuzz/fuzzers/server-name.rs
ctz, line 1: this one will need an OSS fuzz change

djc: Ugh... probably not worth it, then. But I haven't quite figured out how to suppress these lints yet.

ctz: sorry, i just checked and it won't. it will need a rename in https://github.com/rustls/rustls-fuzzing-corpora and that's it

PR #3244, opened 2026-09-03
Explicitly indicate protocol version in use throughout handshake machinery

tgeoghegan:
This set of commits plumbs explicit `ProtocolVersion` values into a variety of places in the handshake machinery which previously could determine that based on either the context they're in or what crypto suite is in use.

These assumptions will not hold as DTLS gets implemented, since for example, module `rustls::client::tls12` will be used for both TLS 1.2 and DTLS 1.2, and `rustls::tls13::Tls13Ciphersuite` is used for DTLS 1.3 as well as TLS 1.3.

Individual commit messages discuss the changes in greater detail.

This PR contains no functional changes.

### DIFF
--- a/rustls/src/hash_hs.rs
+++ b/rustls/src/hash_hs.rs
@@ -3,11 +3,14 @@ use alloc::vec::Vec;
 use core::mem;
 
 use crate::crypto::{HashAlgorithm, hash};
+use crate::enums::ProtocolVersion;
 use crate::msgs::{Codec, HandshakeAlignedProof, HandshakeMessagePayload, Message, MessagePayload};
 
 /// Early stage buffering of handshake payloads.
 ///
-/// Before we know the hash algorithm to use to verify the handshake, we just buffer the messages.
+/// Before we know the hash algorithm to use to verify the handshake or the negotiated protocol
+/// version, we just buffer the messages.
+///
 /// During the handshake, we may restart the transcript due to a HelloRetryRequest, reverting
 /// from the `HandshakeHash` to a `HandshakeHashBuffer` again.
 #[derive(Clone)]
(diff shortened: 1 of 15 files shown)


rustls/src/hash_hs.rs
```diff
@@ -84,6 +92,9 @@ pub(crate) struct HandshakeHash {
 
     /// buffer for client-auth.
     client_auth: Option<Vec<u8>>,
+
+    /// The protocol version negotiated for the handshake.
+    version: ProtocolVersion,
```
tgeoghegan, line 97: As an alternative to adding this `version` field and using it to make runtime decisions, I considered if we could use generics so that the DTLS-specific transcript code could be left out of stream-TLS-only setups altogether (per ["linker friendliness"](https://github.com/rustls/rustls/blob/7124c5d7bd5c5f5f29b90264bff7929ba6b5df4b/CONTRIBUTING.md#linker-friendliness)).

The tricky thing is that while an application can statically decide whether it's doing stream TLS or datagram TLS at compile time, an endpoint can't know if it's going to end up negotiating DTLS 1.2 or 1.3 until runtime, and the transcript hashing mechanism to use depends on that information. That means the generic interface to `HashshakeHashBuffer::start_hash` has to take `ProtocolVersion` no matter what, and  DTLS endpoints will end up linking in both hash implementations anyway.

djc:
`ProtocolVersion`, because it represents bytes on the wire, models both the version and TLS-TCP vs DTLS. I think we should change most of the usage of `ProtocolVersion` instead to a different representation that only models the version, and model the protocol in use separately (along the lines of #3224).

tgeoghegan:
I like that idea, because I found myself writing code like `if version == ProtocolVersion::TLSv1_3 || version == ProtocolVersion::DTLSv1_3 {}` quite a lot. So we'd end up with methods like

```
client::hs::ExpectServerHello::with_version<T: Suite + 'static>(
        mut self,
        protocol: Protocol,
        version: VersionFamily,
        server_hello: &ServerHelloPayload,
        input: &Input<'_>,
        output: &mut dyn Output<'_>,
    ) -> Result<ClientState, Error>
```
and:
```
enum VersionFamily {
    Version1_2,
    Version1_3,
}
```

The names are just a first try. Then we'd construct `EncodableVersion` or `ProtocolVersion` values only when constructing messages to be transmitted. I'll give that a try. Perhaps I should also rebase on #3224 and try to merge this after that?

djc:
Would it perhaps make sense to follow through on 

- https://github.com/tgeoghegan/rustls/pull/2

before working on this?

tgeoghegan:
The crypto trait rework I envision in #3109 and that PR are not blockers to the DTLS work, but doing that would spare us from tediously duplicating a lot of new code across every provider. So, yes, I'm more than happy to go tackle that before trying to land this PR or other pieces that are more explicitly about DTLS.

tgeoghegan:
At @djc's suggestion, I've been working on changes to the record encryption/decryption crypto provider interfaces, with an eye toward making them easier to manage in the face of multiple protocol versions and transports with less code duplication. I'll share more about that in #3109 and a yet-to-be-published PR. In the meantime, to reduce the size of the rustls team's inbox, I'll make this PR a draft.

PR #3249, opened 2026-09-04
[0.23] Enable ML-DSA by default

djc:
Backports the following PRs:

- https://github.com/rustls/rustls/pull/3241
- https://github.com/rustls/rustls/pull/3234 (partially)
- https://github.com/rustls/rustls/pull/2764 (partially)
- https://github.com/rustls/rustls/pull/3237


ctz (approved):
Nice!

ctz:
Will need some release notes. I suggest that:

a) the various tightenings of parsing included here don't need to be included, it's not expected these are user-visible and if they are the errors are probably good enough
b) mention clearly changes to ML-DSA defaults, I think this is the only item to be mentioned in the notes

PR #3255, opened 2026-09-09
reject server hello that does not echo legacy_session_id

Mounika2456:
`ClientHandler::handle_server_hello` for TLS1.3 never looks at the ServerHello's
`legacy_session_id_echo`, so a server echoing nothing at all is accepted:

    client::test::test_client_rejects_server_hello_with_varied_session_id
    called `Result::unwrap_err()` on an `Ok` value: ()

RFC 9846 4.1.3 makes a mismatch a MUST-abort with `illegal_parameter`, and the
identical rule is already enforced one message earlier for HelloRetryRequest via
`IllegalHelloRetryRequestWithWrongSessionId`. In middlebox-compatibility mode the
client sends a random 32-byte `legacy_session_id`, so the check is live on every
TCP TLS1.3 handshake.

The TLS1.2 handler holds the other half of the same gap. Its guard tests
`st.input.resuming.is_none()`, but `resuming` is `Some` for any cached session,
including a TLS1.3 ticket, which is not a TLS1.2 resumption candidate. With one
of those cached the client still sends a compatibility session id and then
accepts a TLS1.2 server that echoes it, instead of raising
`ServerEchoedCompatibilitySessionId`. Testing `resuming_session` is what the
comment above the guard already describes.

The ServerHellos synthesised by the test suite relied on the absent check, so
they now echo the offered id; `encoding::echo_session_id` does that for the two
byte-level flights. BoGo is clean on ring and aws-lc-rs.

### DIFF
--- a/rustls/src/client/tls12.rs
+++ b/rustls/src/client/tls12.rs
@@ -118,18 +118,6 @@ mod server_hello {
                 return Err(PeerMisbehaved::AttemptedDowngradeToTls12WhenTls13IsSupported.into());
             }
 
-            // If we didn't have an input session to resume, and we sent a session ID,
-            // that implies we sent a TLS 1.3 legacy_session_id for compatibility purposes.
-            // In this instance since we're now continuing a TLS 1.2 handshake the server
-            // should not have echoed it back: it's a randomly generated session ID it couldn't
-            // have known.
-            if st.input.resuming.is_none()
-                && !st.input.session_id.is_empty()
-                && st.input.session_id == server_hello.session_id
-            {
-                return Err(PeerMisbehaved::ServerEchoedCompatibilitySessionId.into());
-            }
-
             let ClientHelloInput {
                 config,
                 session_key,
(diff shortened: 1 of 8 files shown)


rustls/src/client/tls12.rs
```diff
@@ -144,6 +132,18 @@ mod server_hello {
                     ClientSessionValue::Tls13(_) => None,
                 });
 
+            // If we didn't have a TLS 1.2 session to resume, and we sent a session ID,
```
djc, line 135: Why did this need to move?

rustls/src/error/mod.rs
```diff
@@ -1226,6 +1226,7 @@ pub enum PeerMisbehaved {
     SelectedUnofferedPsk,
     ServerEchoedCompatibilitySessionId,
     ServerHelloMustOfferUncompressedEcPoints,
+    ServerHelloWithWrongSessionId,
```
djc, line 1229: I wonder if this could be named more precisely? Like maybe `UnmatchedSessionId` or something.

djc (approved):

rustls/src/client/tls12.rs
Mounika2456: The guard now tests `resuming_session` instead of `st.input.resuming`, and that binding is only derived just below the old spot (the `and_then` consumes `st.input.resuming`), so the check has to come after it. Apart from that condition and the first line of the comment, the block is unchanged.

rustls/src/error/mod.rs
Mounika2456: Renamed to `UnmatchedSessionId` and folded that into the first commit. The bogo match arm reflowed under rustfmt with the shorter name, otherwise it's a pure rename.

djc:
Looks like there are some CI failures to address here.

Mounika2456:
The CI failure is `client_connection_rejects_record_containing_subsequent_messages`, added on main in 98456ae after this branch was cut. Its ServerHello carries an empty session id, so the new check trips before the pending-fragment check the test is looking for.

Rebased onto main and folded the fix into the first commit: `echo_session_id` now splices ids of differing lengths and adjusts the record and handshake lengths, so that test echoes the client's id the same way the other two flights do. The rebase also resolved a conflict with the `renegotiation_info` check that now sits next to the moved TLS 1.2 guard. Unit tests, the rustls-test suite, clippy and fmt are clean locally.

ctz (approved):
Thanks!

PR #3258, opened 2026-09-09
Remove unused once_cell dependency

Thomasdezeeuw:


djc (approved):

Thomasdezeeuw:
Didn't realise the Cargo.lock file was committed, so updated that as well.

cpu (approved):

cpu:
Can you squash your commits please

Thomasdezeeuw:
I didn't think that removing an unused dependency would require this many changes, but here we are. I've now checked the entire workspace, not just the rustls crate, and it should work.

PR #3259, opened 2026-09-10
Reject second ClientHello that withdraws PSK offer

oleksii-udovychenko:
[RFC 8446 §4.1.2](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.1.2) lists exhaustively what a client may change in its second ClientHello after a HelloRetryRequest. A `pre_shared_key` extension may be updated there (recomputing the binders, adjusting the obfuscated ticket age, and optionally dropping PSKs incompatible with the server's chosen cipher suite), but withdrawing the extension altogether is not one of the permitted changes.

We accepted that anyway. `ExpectClientHello` kept no record of whether the first ClientHello carried a PSK offer, so a second hello that omitted the extension was handled as if the client had never offered a PSK at all.

This PR:

- adds `PeerMisbehaved::MissingPskExtensionInSecondClientHello`, which maps to a `missing_extension` alert
- records the first hello's offer in `ExpectClientHello::offered_psk_before_retry` when emitting the HelloRetryRequest, and rejects a second hello that drops it in `client_hello::handle_client_hello`
- enables `Resume-Server-OmitAllPSKsOnSecondClientHello`, which was disabled with a `TODO` reason
- adds a server unit test for the same case

### DIFF
--- a/rustls/src/server/test.rs
+++ b/rustls/src/server/test.rs
@@ -23,11 +23,11 @@ use crate::crypto::{
     SingleCredential, TEST_PROVIDER, TLS13_TEST_SUITE, tls12, tls12_only,
 };
 use crate::enums::{CertificateType, ProtocolVersion};
-use crate::error::{Error, PeerIncompatible};
+use crate::error::{Error, PeerIncompatible, PeerMisbehaved};
 use crate::msgs::{
     ClientExtensions, ClientHelloPayload, Codec, Compression, HEADER_SIZE, HandshakeMessagePayload,
-    HandshakePayload, KeyShareEntry, Message, MessagePayload, Random, Reader, SessionId,
-    SupportedProtocolVersions,
+    HandshakePayload, KeyShareEntry, Message, MessagePayload, PresharedKeyIdentity,
+    PresharedKeyOffer, PskKeyExchangeModes, Random, Reader, SessionId, SupportedProtocolVersions,
 };
 use crate::pki_types::pem::PemObject;
 use crate::pki_types::{CertificateDer, FipsStatus, PrivateKeyDer};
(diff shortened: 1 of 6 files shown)


djc (approved):
Thanks! Is there bogo coverage for this?

oleksii-udovychenko:
Yes. `Resume-Server-OmitAllPSKsOnSecondClientHello` was in `DisabledTests` with the reason `TODO: we do not reject a second ClientHello that drops all PSKs`, and this PR removes that entry, so it runs now. 

It is the test the change came from: without the check the shim finishes a full handshake with a HelloRetryRequest instead of sending `missing_extension`, and the runner fails it on `wanted to see [Resumed] but got FullWithHelloRetryRequest`.

djc:
Ugh, not sure why I missed that. Thanks, I think this looks good but will wait for one more maintainer to have a look.

rustls/src/server/test.rs
```diff
@@ -323,6 +323,60 @@ fn server_chooses_ffdhe_group_for_client_hello(
         .expect("DHE not used");
 }
 
+#[test]
+fn second_client_hello_cannot_withdraw_psk_offer() {
+    // Per RFC 8446 section 4.1.2, dropping a PreSharedKey offer is not one of the
```
cpu: Nit: I think we should cite 9846 in new code where possible: https://www.rfc-editor.org/info/rfc9846/#section-4.2.2

rustls/src/server/tls13.rs
```diff
@@ -171,6 +171,17 @@ mod client_hello {
                 return Err(PeerMisbehaved::EarlyDataAttemptedInSecondClientHello.into());
             }
 
+            // RFC 8446 section 4.1.2 allows the second ClientHello to update a PreSharedKey
```
cpu: Ditto here

cpu (approved):
Thanks! Seems reasonable to me. I had two very minor comment nits to consider.

rustls/src/server/test.rs
oleksii-udovychenko: Updated.

rustls/src/server/tls13.rs
oleksii-udovychenko: Updated.

oleksii-udovychenko:
Pushed the RFC number update just now. Thanks!

cpu (approved):

PR #3260, opened 2026-09-12
reject a second ClientHello that changes the cipher suite

oleksii-udovychenko:
## What RFC 9846 requires

RFC 9846 [§4.2.4](https://www.rfc-editor.org/rfc/rfc9846.html#section-4.2.4): "Servers MUST ensure that they negotiate the same cipher suite when receiving a conformant updated ClientHello (if the server selects the cipher suite as the first step in the negotiation, then this will happen automatically)." The client cannot vary the offer either. [§4.2.2](https://www.rfc-editor.org/rfc/rfc9846.html#section-4.2.2) has it "MUST send the same ClientHello without modification, except as follows", and the list that follows covers `key_share` and `early_data`, not `cipher_suites`.

## Before the change

We choose the suite again for the second ClientHello, so a client that offers a different one is given it:

    ---- server::test::second_client_hello_cannot_change_cipher_suite stdout ----

    thread 'server::test::second_client_hello_cannot_change_cipher_suite' (249343) panicked at rustls/src/server/test.rs:444:5:
      left: 0xff13
     right: 0xff14

Left is the suite our HelloRetryRequest named, right is the one our ServerHello then negotiated.

We do not pick the suite as a first step, so nothing makes this happen automatically. `choose_suite_and_kx_group` runs again over whatever the second hello offers, and `ExpectClientHello` kept no record of the earlier choice to compare against. Everything else that has to hold still across a retry is already checked, `ServerNameDifferedOnRetry` and `HandshakeHashVariedAfterRetry` in `hs.rs` and `MissingPskExtensionInSecondClientHello` in `tls13.rs`. Only a broken or hostile client reaches this, since a conformant one rejects the outcome itself with `SelectedDifferentCipherSuiteAfterRetry`, and two suites with different hashes would already fail `HandshakeHashVariedAfterRetry`.

## After the change

`suite_before_retry` records the suite as the HelloRetryRequest is emitted, and `with_version` compares it as soon as `choose_suite_and_kx_group` returns. A second hello that withdraws that suite now fails with `CipherSuiteDifferedOnRetry`, which falls through to `illegal_parameter`.

There is no bogo test to enable. `HelloRetryRequest-CipherChange-TLS13` is a client test and passes already, and the runner cannot vary the cipher list between the two hellos: `SendCipherSuites` is applied in `createClientHello`, which builds only the first.

### DIFF
--- a/rustls/src/crypto/test_provider.rs
+++ b/rustls/src/crypto/test_provider.rs
@@ -81,6 +81,19 @@ pub(crate) const TLS13_TEST_SUITE: &Tls13CipherSuite = &Tls13CipherSuite {
     quic: None,
 };
 
+/// Differs from `TLS13_TEST_SUITE` only in its code point: same hash, same everything else.
+pub(crate) const TLS13_TEST_SUITE_ALT: &Tls13CipherSuite = &Tls13CipherSuite {
+    common: CipherSuiteCommon {
+        suite: CipherSuite(0xff14),
+        hash_provider: FAKE_HASH,
+        confidentiality_limit: u64::MAX,
+    },
+    protocol_version: crate::version::TLS13_VERSION,
+    hkdf_provider: &tls13::HkdfUsingHmac(FAKE_HMAC),
+    aead_alg: &Aead,
+    quic: None,
+};
+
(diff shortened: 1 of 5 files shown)


djc (approved):

ctz (approved):
Thanks!

Issue #3246, opened 2026-09-04
QUIC builder API


Oakchris1955:
**Checklist**
* [x] I've searched the issue tracker for similar requests

**Is your feature request related to a problem? Please describe.**
The current QUIC API makes it difficult to add configuration options not present in `ClientConfig` without causing breaking API causes. This is currently an issue with #3136.

**Describe the solution you'd like**
Use a builder API for `quic::ClientConnection`, like `client::ClientConnection`, so that by simply calling a method like `client_config.connect_quic(/* params */)`, a `quic::ClientConnectionBuilder` can be obtained and configured akin to `client::ClientConnectionBuilder`.

Note: the TLS-encoded transport parameters method parameter seems to be set to an empty vector in some test cases, so perhaps that could be moved away from `.connect_quic` and into the connection builder, leaving the former with just the server name and QUIC version parameters.

**Describe alternatives you've considered**
None

**Additional context**
https://github.com/rustls/rustls/pull/3136#issuecomment-5507379474

PR #3261, opened 2026-09-12
Builder API for QUIC

Oakchris1955:
Related to #3246

I don't know whether that is a desirable API change, but since it wasn't complex to implement, I did so.

The `connect_quic` method of `ClientConfig` takes a `params` parameters, which contains the TLS-encoded transport parameters to send. However, in some test case, it seems to be empty, and I am thinking that perhaps it can be removed from the builder constructor and have its own builder method.


ctz:
Why not add `fn ClientConnectionBuilder::build_quic()` (taking all mandatory QUIC parameters) instead of a parallel QUIC-specific `ClientConnectionBuilder`?

Oakchris1955:
I didn't think of that, but to be honest, that sounds like a better solution.

Oakchris1955:
Implemented aforementioned changes. I also implemented `From<QuicCommon<ClientSide>> for ClientConnection`, since making the `inner` field publicly accessible to the whole crate just for one constructor didn't seem sensible to me.

ctz (approved):

djc (approved):
Nice, thanks!

djc:
Would you mind rebasing so CI can pass?

Oakchris1955:
> Would you mind rebasing so CI can pass?

Done (forgot to update my local `main` branch for the previous rebase)

PR #3264, opened 2026-09-12
Sketch of desirable(?) `RecordEncrypter::encrypt` interface

ctz:
This is a sketch of https://github.com/rustls/rustls/issues/3068#issuecomment-4751820215 and relevant to discussing #3109

Note it doesn't actually address #3109, except trivially, because each encrypter still knows about the version and protocol it is for.


cpu:
How do you feel about the benchmark bot results?

ctz:
> How do you feel about the benchmark bot results?

I've optimised the rustls-fuzzing-provider encrypters. (This is 100% test code, but is a significant part of the `no_crypto` benchmarks).

PR #3265, opened 2026-09-13
0.23.45: Ensure handshake "alignment" check covers previously-received messages

ctz:
Release notes (and text for pending RUSTSEC)

**Bug fix**: TLS 1.3 handshake messages incorrectly accepted across encryption level boundaries

Rustls accepted TLS 1.3 handshake messages sent at the wrong encryption level when they followed a key-changing message in the same record. The handshake transcript is still authenticated, so a network-position attacker cannot use this to alter or complete a handshake; the practical effect is that a peer could send handshake messages that should be encrypted in plaintext without rustls rejecting the connection.

This issue is tracked as [GHSA-2mjx-qc3c-rqvc](https://github.com/rustls/rustls/security/advisories/GHSA-2mjx-qc3c-rqvc). This issue affects rustls versions 0.23.13 through 0.23.44 inclusive.

Note that this is functionally the same bug as [GO-2026-4340](https://pkg.go.dev/vuln/GO-2026-4340).


rustls/tests/api.rs
```diff
@@ -4206,6 +4208,161 @@ fn key_log_for_tls13() {
...
+        encrypter
+            .encrypt(remainder.borrow_outbound(), 0)
+            .unwrap()
+            .encode(),
+    );
+    output
+}
+
+fn hkdf_expand_label<T: From<[u8; N]>, const N: usize>(
+    expander: &dyn rustls::crypto::tls13::HkdfExpander,
```
djc: Nit: import `HkdfExpander` and `expand()`?

rustls/tests/api.rs
```diff
@@ -6614,6 +6769,102 @@ fn test_acceptor() {
     assert_eq!(alert_content, expected);
 }
 
+#[test]
+fn client_hello_acceptor_rejects_record_containing_subsequent_messages() {
+    use rustls::server::Acceptor;
```
djc: Nit: import this at the top of the module?

djc (approved):
Ugh, when/how did we regress it?

rustls/tests/api.rs
ctz: did that, it involved knock-on changes in some unrelated tests though which is what i was attempting with this.

PR #3266, opened 2026-09-13
Ensure handshake "alignment" check covers previously-received messages

ctz:
Incorporate test case from GHSA-2mjx-qc3c-rqvc POC, plus three more limited tests.


djc (approved):
Looks like my feedback from the 0.23.x PR is not applicable here.

PR #3270, opened 2026-09-15
Take semver-compatible updates

ctz:
Fixing https://github.com/rustls/rustls/actions/runs/34878662492/job/104092116587


djc (approved):

djc:
The cargo-deny-action invokes `cargo deny --all-features check`; you probably skipped the option?

PR #3273, opened 2026-09-16
Add serde Serialize and Deserialize to enum_builder

Narfinger:
This adds serde Serialize and Deserialize to all enums constructed with enum_builder gated behind a feature flag (serde).

Signed-off-by: Narfinger <Narfinger@users.noreply.github.com>


rustls/Cargo.toml
```diff
@@ -14,8 +14,9 @@ autotests = false
 exclude = ["src/testdata", "tests/**"]
 
 [features]
-default = ["tracing", "webpki"]
+default = ["tracing", "webpki", "serde"]
```
djc: We're definitely not going to enable this by default.

Narfinger: Sorry it wasn't my intention to put it as a default.

